PR #8320, opened 2019-08-18
Do not yield to UI from PauseAndLock

CookiePLMonster:
This PR fixes a rare crash in Dolphin, easiest to reproduce by spamming volume up/down controls continuously in game. With a little bit of luck (and patience), it was possible to crash on recursive stepping mutex lock. It seems to be a lot easier to reproduce it after opening and closing Config window before running the game, so Audio settings window gets created and registers its config callbacks.

The underlying reason behind this crash was an infinite recursion caused by yielding to UI thread (processing Qt events) from `Core::RunAsCPUThread` running in a callback from another Qt event. Yielding was done only when `s_state_cpu_idle_cvar` wait timed out, which turned out to be **extremely** rare, so removing it does not impact UI responsiveness negatively.

The exact sequence of events here was as follows:
1. Volume control hotkey is triggered.
2. `AudioPane::OnVolumeChanged` is called.
3. `AudioPane::SaveSettings` is called.
4. `InvokeConfigChangedCallbacks` is called.
5. `VideoConfig::Refresh` is called on CPU thread (acquires stepping mutex).
6. `CPU::PauseAndLock` waits for `s_state_cpu_idle_cvar` which times out.
7. Timing out causes Dolphin to yield to UI and call `processEvents`.
8. `AudioPane::OnVolumeChanged` is called **again**, and so are further calls.
9. `VideoConfig::Refresh` attempts to call on CPU thread **again**.
10. Stepping mutex is **not** recursive, so attempting to acquire it again results in an exception.

Now, would making a stepping lock recursive help? Most likely not, since it would likely be possible to yield to UI recursively, eventually causing a stack overflow. Not yielding to UI at all here is safer, since it almost never happened anyway.

### DIFF
--- a/Source/Core/Core/HW/CPU.cpp
+++ b/Source/Core/Core/HW/CPU.cpp
@@ -183,10 +183,7 @@ void Stop()
 
   while (s_state_cpu_thread_active)
   {
-    std::cv_status status =
-        s_state_cpu_idle_cvar.wait_for(state_lock, std::chrono::milliseconds(100));
-    if (status == std::cv_status::timeout)
-      Host_YieldToUI();
+    s_state_cpu_idle_cvar.wait(state_lock);
   }
 
   RunAdjacentSystems(false);
@@ -252,10 +249,7 @@ void EnableStepping(bool stepping)
 
     while (s_state_cpu_thread_active)
     {
-      std::cv_status status =
-          s_state_cpu_idle_cvar.wait_for(state_lock, std::chrono::milliseconds(100));
-      if (status == std::cv_status::timeout)
-        Host_YieldToUI();
+      s_state_cpu_idle_cvar.wait(state_lock);
     }
 
     RunAdjacentSystems(false);
@@ -303,10 +297,7 @@ bool PauseAndLock(bool do_lock, bool unpause_on_unlock, bool control_adjacent)
 
     while (s_state_cpu_thread_active)
     {
-      std::cv_status status =
-          s_state_cpu_idle_cvar.wait_for(state_lock, std::chrono::milliseconds(100));
-      if (status == std::cv_status::timeout)
-        Host_YieldToUI();
+      s_state_cpu_idle_cvar.wait(state_lock);
     }
 
     if (control_adjacent)
--- a/Source/Core/Core/HW/CPU.cpp
+++ b/Source/Core/Core/HW/CPU.cpp
@@ -87,7 +87,7 @@ static void ExecutePendingJobs(std::unique_lock<std::mutex>& state_lock)
 
 void Run()
 {
-  std::unique_lock<std::mutex> state_lock(s_state_change_lock);
+  std::unique_lock state_lock(s_state_change_lock);
   while (s_state != State::PowerDown)
   {
     s_state_cpu_cvar.wait(state_lock, [] { return !s_state_paused_and_locked; });
@@ -177,7 +177,7 @@ void Stop()
   // Change state and wait for it to be acknowledged.
   // We don't need the stepping lock because State::PowerDown is a priority state which
   // will stick permanently.
-  std::unique_lock<std::mutex> state_lock(s_state_change_lock);
+  std::unique_lock state_lock(s_state_change_lock);
   s_state = State::PowerDown;
   s_state_cpu_cvar.notify_one();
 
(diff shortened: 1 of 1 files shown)


JosJuice:
If you apply the same change to `EnableStepping` too, you can fix https://bugs.dolphin-emu.org/issues/11718.

CookiePLMonster:
Updated the PR with removal of any yielding in CPU.cpp - forementioned issue now seems fixed. I used the same "hotkey spam" method to take hundreds of screenshots as fast as code allows for, with no issues.

stenzek:
Have we verified this doesn't break booting-with-fullscreen-enabled in DX11? I know this was one of the reasons we had to yield - it could deadlock if the swap chain is created on a different thread to the message loop, and the messages aren't handled. Although, booting might be asynchronous anyway, can't remember.

CookiePLMonster:
That's a good point, I have not verified this with any other backends but
D3D12. It'd be pretty bad if we had to yield from there, but if really
needed maybe I can introduce a flag not to yield if it would cause
recursive Qt calls. Shouldn't be hard but I'm not a fan of how it sounds.

CookiePLMonster:
@stenzek How reproducible was this deadlock? I ticked "Use Fullscreen" and tried booting games in each backend around 4 to 5 times, with no issues.

Ebola16:
I'm still able to reproduce https://bugs.dolphin-emu.org/issues/11718 in this PR.

CookiePLMonster:
How easily? I tried to "bruteforce" it by hacking the screenshot hotkey to always return true, effectively making hundreds of screenshots as fast as Dolphin allows to - no issues.

Also, which backend did you use? I only checked screenshots against D3D12 I believe.

Ebola16:
It is easy for me to reproduce the problem with D3D11 by pressing the actual hotkey rapidly. I just tried the same method in D3D12 and the issue still occurs. I could test your build that spams screenshots without hotkey presses if you'd like.

CookiePLMonster:
> I could test your build that spams screenshots without hotkey presses if you'd like.

In `HotkeyScheduler.cpp` there is
```
// Screenshot
if (IsHotkey(HK_SCREENSHOT))
  emit ScreenShotHotkey();
```

All I did was comment out the condition here.

CookiePLMonster:
OK, I still froze now. I'll take a look, as it may be two unrelated issues.

Ebola16:
I commented out `if (IsHotkey(HK_SCREENSHOT))` in your PR and Dolphin still freezes or crashes after taking about as many screenshots as Master.

CookiePLMonster:
So far I never had it crash but I indeed was able to get it to freeze - it may actually be a separate issue to the yield one, so I'll investigate further.

CookiePLMonster:
So turns out freezes I had were missclicks on F9 vs F10 - now after I unbound it I was able to spam over 750 screenshots with both D3D11 and D3D12 and didn't freeze... what is going on???

Ebola16:
Capture

CookiePLMonster:
Those settings are same as mine... the only remaining idea I have is that maybe you are not using a Dual Core option and I am - I'll retest today later with dual core option **disabled**.

CookiePLMonster:
I tried it again with Dual Core option disabled, as well with some other options like Deterministic CPU Thread - still was unable to reproduce freezes or crashes.

I'm confused.

JosJuice:
Yes, I was able to reproduce it on master, but not on this PR once `EnableStepping` was changed.

CookiePLMonster:
> I'm not sure how to do it in Visual Studio, but would enabling the debugger and providing a dump during the freeze / crash help?

It would - ideally, you could also provide your Dolphin/DolphinD.exe with a PDB, so this dump can be analyzed with symbols.

Ebola16:
Dolphin debug build with the debugger attached frequently raised access violation exceptions during startup of Super Smash Bros. Brawl. I disabled breaking when that type of exception was hit. I was able to reproduce the problem but the debugger didn't raise an exception. I just paused frozen Dolphin from the debugger and dumped from there. Let me know if you need something more from me.

http://www.mediafire.com/file/sdwan4m63jcp44m/DolphinCPLM.rar/file

JosJuice:
Any objections to merging this? While it doesn't fix Ebola16's issue, it does fix things, and the logic seems sound to me. I tested booting to fullscreen with Direct3D 11 just now and experienced no problems.

CookiePLMonster:
> While it doesn't fix Ebola16's issue, it does fix things

We can always just give this issue another spin after merging, since I haven't had a chance to analyze this crash dump yet.

PR #8298, opened 2019-08-05
WinUpdater: Properly hide window on startup as intended

CookiePLMonster:
This PR fixes two minor issues in Windows Updater:

1) Updater window is now created hidden, as UpdaterCommon attempts to hide it anyway just after creating:
```cpp
UI::Init();
UI::SetVisible(false);
```

However, creating it visible and hiding still made it flicker for a split second, which is not really useful and not aesthetic.

2) `UI::Init` was tweaked so it waits for the UI thread to exit from `InitWindow`, so after the window has either been created or it's been attempted to be created. This avoids "losing" information from calls to eg. `SetVisible` which could be made **before** the window creates. This for example caused an issue where updater was stuck as always on top, even when the user selected "Update after quitting Dolphin". This rendered this option basically useless, because updater window was stuck on top until the user has quit Dolphin.

PR #8322, opened 2019-08-19
WinUpdater: Properly initialize COM on window thread

CookiePLMonster:
This PR fixes somewhat of a regression from #8298, even though it was not really related. Before my changes, COM has been implicitly initialized on window thread by something else, and so `CoCreateInstance` succeeded. After my changes, this implicit initialization stopped covering this specific case and `CoCreateInstance` started failing. This caused `ITaskbarList3` not to be created, which in turn broke the taskbar progress bar functionality.

Now I explicitly initialize and uninitialize COM on window thread, like it always should have been done. I also migrated `ITaskbarList3` to use WRL instead of a raw pointer to make code cleaner and easier to release all references properly before calling `CoUninitialize`.

EDIT:
Obligatory Raymond Chen explaining the phenomenon which **most likely** happened here:
https://devblogs.microsoft.com/oldnewthing/20130419-00/?p=4613

### DIFF
--- a/Source/Core/WinUpdater/WinUI.cpp
+++ b/Source/Core/WinUpdater/WinUI.cpp
@@ -11,6 +11,7 @@
 #include <CommCtrl.h>
 #include <ShObjIdl.h>
 #include <shellapi.h>
+#include <wrl/client.h>
 
 #include "Common/Event.h"
 #include "Common/ScopeGuard.h"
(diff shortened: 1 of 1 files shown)


Source/Core/WinUpdater/WinUI.cpp
```diff
@@ -206,6 +205,14 @@ void SetDescription(const std::string& text)
 
 void MessageLoop()
 {
+  HRESULT result = CoInitializeEx(nullptr, COINIT_MULTITHREADED);
```
stenzek: This could potentially be named `coinit_result` or something more meaningful, but it's a small function anyway.

CookiePLMonster:
I agree, but I opted for this name because all variable uses are literally
two lines apart, so no way to lose track of it.

PR #8323, opened 2019-08-20
Add safe texture cache to some uDraw games' ini files

Craftyawesome:
Fixes https://bugs.dolphin-emu.org/issues/11823


JMC47:
@dolphin-emu-bot rebuild

PR #8324, opened 2019-08-20
Enable adapter combo only if adapters are supported by current backend

CookiePLMonster:
`GeneralWidget::OnEmulationStateChanged` enabled/disabled the Adapter combo regardless whether the current backend actually supports it or not.


CookiePLMonster:
Idea how to engineer this better: either subclass this Qt class to overload `setEnabled` to check for adapters implicitly, or (if possible) install an event filter on `EnabledChange`.

PR #8326, opened 2019-08-21
ActionReplay: Make use of fmt where applicable

lioncash:
Migrates the ActionReplay logging over to using `fmt::format` instead of `StringFromFormat`.

### DIFF
--- a/Source/Core/Core/ActionReplay.cpp
+++ b/Source/Core/Core/ActionReplay.cpp
@@ -23,7 +23,6 @@
 
 #include <algorithm>
 #include <atomic>
-#include <cstdarg>
 #include <iterator>
 #include <mutex>
 #include <string>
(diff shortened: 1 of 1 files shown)


Source/Core/Core/ActionReplay.cpp
```diff
@@ -327,6 +325,12 @@ static void LogInfo(const char* format, ...)
   }
 }
 
+template <typename... Args>
+static void LogInfo(std::string_view format, const Args&... args)
```
CookiePLMonster: Shouldn't this be `Args&&... args` and then perfectly forwarding `args...` to the call?

lioncash: This wouldn't result in any benefit as `make_format_args` uses as `const Args&...` as its parameter type

PR #8164, opened 2019-06-08
Game INI updates and additions

Miksel12:
An update to the Game INI's:
-Adds a few games.
-Disables Dual Core for a bunch of games that crashed with Dual Core enabled in a way that makes it unplayable/unfinishable or caused severe issues.
-Enables MMU for Crunchyroll Channel so it can boot and for Interactive Multi Game Demo Disc with demo's which require MMU.
-Disables defered EFB for games that crashed or showed weird behaviour with defered EFB enabled.
-Sets TextureCacheAccuracy to safe for a few games that need it.
-Sets EFB to texture and ram for a few games that need it.
-Removes Skip EFB access for some games that don't need it anymore.
-Removes Wiimote.Shake and Wiimote.Swing as those are no longer functional. 

I also removed TextureCacheAccuracy = Safe from NSMBW because accuracy on fast shows no issues.

PR #8328, opened 2019-08-23
Dualcore INI changes

Miksel12:
Dualcore changes taken from #8164.
I have re-evaluated all the dualcore changes and only included the most important INI changes based on the wiki, issue tracker and forum. 
Dualcore has been enabled for 'Baten Kaitos Eternal Wings and the Lost Ocean' as tests on wiki show that it can safely be enabled.

### DIFF
--- a/Data/Sys/GameSettings/E62.ini
+++ b/Data/Sys/GameSettings/E62.ini
@@ -2,6 +2,7 @@
 
 [Core]
 # Values set here will override the main Dolphin settings.
+CPUThread = False
 
 [OnLoad]
 # Add memory patches to be loaded once on boot here.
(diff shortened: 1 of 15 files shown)


JMC47:
Something I feel strongly about for down the road is saying why a feature is disabled in the INI so that when we eventually change timings, we know what to test.

Miksel12:
Probably not a bad idea, especially for obscure titles.

JMC47:
Fake completion should never be used outside of netplay.  Other tha nthat, this LGTM.  Please use SyncGPU or Single Core instead.  If neither works, then we just leave it broken.  Fake-completion is egregious.

JosJuice:
I don't see a problem with setting fake-completion if the game has been tested to work with it, though I don't think I really would mind setting single core instead.

JMC47:
If Fake-completion is significantly faster and verified to work, I guess we could set it to that for now.  Maybe when dualcore is fixed, dualcore will just work.

Miksel12:
I changed all the Fake-completion to single core.

PR #8329, opened 2019-08-23
WiimoteEmu: Add option to use system battery level

blaahaj:
Only supports Windows and Mac right now.

### DIFF
--- a/Source/Core/Core/HW/WiimoteEmu/EmuSubroutines.cpp
+++ b/Source/Core/Core/HW/WiimoteEmu/EmuSubroutines.cpp
@@ -16,6 +16,7 @@
 #include "Core/Analytics.h"
 #include "Core/Core.h"
 #include "Core/HW/WiimoteCommon/WiimoteHid.h"
+#include "Core/HW/WiimoteEmu/SystemBattery.h"
 #include "Core/HW/WiimoteEmu/WiimoteEmu.h"
 #include "Core/HW/WiimoteReal/WiimoteReal.h"
 #include "InputCommon/ControllerEmu/ControlGroup/Attachments.h"
@@ -240,12 +241,34 @@ void Wiimote::HandleRequestStatus(const OutputReportRequestStatus&)
   // Max battery level seems to be 0xc8 (decimal 200)
   constexpr u8 MAX_BATTERY_LEVEL = 0xc8;
 
-  m_status.battery = u8(std::lround(m_battery_setting.GetValue() / 100 * MAX_BATTERY_LEVEL));
-
   if (Core::WantsDeterminism())
   {
     // One less thing to break determinism:
     m_status.battery = MAX_BATTERY_LEVEL;
+  }
+  else if (std::optional<float> battery;
+           m_system_battery_setting.GetValue() && (battery = GetSystemBatteryLevel()))
+  {
+    // The HOME Menu and Metroid Prime 3 battery bars work like:
+    // 84+ => 4/4, 67+ => 3/4, 51+ => 2/4, 3+ 1/4, 2 or less => 0/4
+    // Something that fits those values is more useful than just multiplying
+    // by MAX_BATTERY_LEVEL (otherwise e.g. 50% battery status shows up as 4/4)
+    // The following math is not based on real battery behaviour, it just makes
+    // good-looking numbers..
+    constexpr float LOW = 0.3f;
+    if (*battery > LOW)
+    {
+      *battery = std::exp(*battery) / M_E;
+    }
+    else
+    {
+      *battery *= std::exp(LOW) / M_E / LOW;
+    }
+    m_status.battery = u8(std::lround(*battery * 100));
+  }
+  else
+  {
+    m_status.battery = u8(std::lround(m_battery_setting.GetValue() / 100 * MAX_BATTERY_LEVEL));
   }
 
   // Less than 0x20 triggers the low-battery flag:
--- a/Source/Core/Core/HW/WiimoteEmu/SystemBattery.cpp
+++ b/Source/Core/Core/HW/WiimoteEmu/SystemBattery.cpp
@@ -1,0 +1,102 @@
+// Copyright 2019 Dolphin Emulator Project
+// Licensed under GPLv2+
+// Refer to the license.txt file included.
+
+#include "Core/HW/WiimoteEmu/SystemBattery.h"
+
+#include "Common/ScopeGuard.h"
+
+#if defined(_WIN32)
+#include <Windows.h>
+#elif defined(__APPLE__)
+#include <CoreFoundation/CFDictionary.h>
+#include <IOKit/ps/IOPSKeys.h>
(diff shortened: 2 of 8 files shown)


CookiePLMonster:
@dolphin-emu-bot rebuild

Source/Core/Core/HW/WiimoteEmu/EmuSubroutines.cpp
```diff
@@ -240,13 +241,26 @@ void Wiimote::HandleRequestStatus(const OutputReportRequestStatus&)
...
+  else if (m_system_battery_setting.GetValue() && (battery = GetSystemBatteryLevel()))
+  {
+    // HOME Menu treats this as a percentage, if weirdly rounded.
+    // 100 shows as 4/4, 75 as 3/4, 51 (not 50) as 2/4.
+    m_status.battery = u8(std::lround(*battery * 100));
+  }
+#endif
+  else
+  {
+    m_status.battery = u8(std::lround(m_battery_setting.GetValue() / 100 * MAX_BATTERY_LEVEL));
```
blaahaj: Note: I have left this code as it is, but I think it is wrong. MAX_BATTERY_LEVEL is way beyond 100% if we use HOME menu as a reference, so even if you set the m_battery_setting to 50% currently, it still shows up as 4/4 bars.

jordan-woyak: The Remote hardware definitely produces values around 0xc8 (decimal 200) for new batteries.

Perhaps the HOME menu has been designed to show 4/4 bars for partially drained batteries to comfort players. 4/4 is more of a "you're all good" rather than a "it rounds up to 100%".

Having said that, I don't know the best way to implement this feature when the visual does not match the underlying value nicely.

blaahaj: I think I will check some other Wii software that displays the battery, to see if they use the value differently.

altimumdelta: It may be designed to show a visual change only when the level comes below the standardized value of the required 1.5V AA batteries, which usually come at 1.65V when new/fully charged and that would produce a total of 3.3V which seems to be the optimal value considering the Wii Remote IR camera is labeled as _3.3V_

For example it could be something like: 

- 200 = 3.3 VDC
- 100 = 3.0 VDC
- 0 = 2.7 VDC

blaahaj: I tried Metroid Prime 3 and it seems to display the battery value bars exactly the same as the HOME menu, at least for 100, 75, 51 and 50.

You might be onto something with the voltage thing. Another option would be for the settings to have 200% as the maximum and default value, corresponding to the integer 200.

blaahaj: I could adjust the existing setting in this PR, or make another PR, if you think it's a good idea.

Source/Core/Core/HW/WiimoteEmu/SystemBattery.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+  // Not on battery power, treat as 100%
+  if (power_status.ACLineStatus == 1)
+  {
+    return 1.0;
+  }
+
+  if (power_status.BatteryLifePercent != 255)
+  {
+    return (float)power_status.BatteryLifePercent / 100.0;
```
Warepire: You can avoid the cast if you declare the divisor as `100.0f`.

blaahaj: Oh right, if one side of the division is a floating-point type, I don't need to cast the other.

Source/Core/Core/HW/WiimoteEmu/EmuSubroutines.cpp
blaahaj: I was wrong, it's not really percentage-based.

Source/Core/Core/HW/WiimoteEmu/EmuSubroutines.cpp
```diff
@@ -240,13 +241,26 @@ void Wiimote::HandleRequestStatus(const OutputReportRequestStatus&)
...
+#ifdef SYSTEM_BATTERY_LEVEL_AVAILABLE
+  std::optional<float> battery;
+#endif
   if (Core::WantsDeterminism())
   {
     // One less thing to break determinism:
     m_status.battery = MAX_BATTERY_LEVEL;
   }
+#ifdef SYSTEM_BATTERY_LEVEL_AVAILABLE
+  else if (m_system_battery_setting.GetValue() && (battery = GetSystemBatteryLevel()))
```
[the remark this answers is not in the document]
blaahaj: Yes, but it is most elegant if written this way. I don't like that it would otherwise fetch the system battery even if you turned that feature off. Alternatively, I would have to add another m_system_battery_setting_check.

Source/Core/Core/HW/WiimoteEmu/EmuSubroutines.cpp
jordan-woyak: I don't like the "hacky math".

I think it would make more sense to map the system's 0-100% level directly to the 0-200 value produced by a Wii Remote.

Why must we hack the value attempting to produce linear output on the "4 bars" when that is not even how the real system works?

blaahaj: Well, yes, you could do that, but it would also make the feature completely useless.

blaahaj: > Why must we hack the value attempting to produce linear output on the "4 bars" when that is not even how the real system works?

Battery voltage doesn't decline linearly with battery capacity. I should probably be trying to emulate that behaviour rather than doing this fudge though.

jordan-woyak: I'd be happier with emulating a battery voltage decline.

blaahaj: I looked up an example AA battery voltage curve and made a function that roughly fit it, but it is no better than the identity function in terms of making a given percentage produce a sensible UI result. Now I am wondering what the Wiimote byte actually means, and how the HOME Menu interprets it.

jordan-woyak:
I had forgotten about this PR. There have been some relevant changes to Dolphin since the last update here.
I figured out a linear fit for the Wii Remote battery level in the Wii Menu #8908.
Devices can provide a "Battery" input #8614 #8474.
These inputs can be connected to the "Battery" level setting of emulated Wii Remotes #8417.

Perhaps you should expose the system battery level as an input of the default "Keyboard and Mouse" device?
This would make it consistent with XInput, DSU, and Real Wiimote battery level passthrough.
I'd like to not add another setting when this functionality is already almost in-place elsewhere.

PR #8330, opened 2019-08-24
VolumeVerifier: Add Redump.org integration

JosJuice:
https://bugs.dolphin-emu.org/issues/10867

~~The caching behavior is not especially good right now (if you fail to connect it will cache the fail and not retry again until you restart Dolphin...), but other than that~~ this should be ready for review.

image


Source/Core/DiscIO/VolumeVerifier.cpp
```diff
@@ -39,6 +49,276 @@
...
+      request.Get("http://redump.org/datfile/" + platform + "/serial,version",
+                  {{"User-Agent", Common::scm_rev_str}});
+
+  if (!result)
+  {
+    *status_out = DownloadStatus::Fail;
+    return;
+  }
+
+  if (result->size() > 1 && (*result)[0] == '<' && (*result)[1] == '!')
```
BhaaLseN: While redump does present a `DOCTYPE` declaration (for XHTML 1.1), I wonder if checking for the open angle bracket would be enough (so it still works when they do). But then again, nothing mandates that it really needs to be starting with that. It could also have blanks, a BOM or other stuff that doesn't affect the DOM (altho, the majority _would_ indeed start with a `<` there when it's not whitespace).

Should we perhaps be checking the `Content-type` header instead (assuming we can somehow get it from the response without jumping too many hoops); or check for the PK3 Zip header instead (assuming everything else is _probably_ plain-text-ish)?

...the more I think about this (and type into the comment box), the more I think my comment is useless and won't being any change to this section of code...mainly because all the alternatives are just as bad.

Source/Core/DiscIO/VolumeVerifier.cpp
```diff
@@ -39,6 +49,276 @@
...
+std::once_flag RedumpVerifier::m_gc_download_flag;
+std::once_flag RedumpVerifier::m_wii_download_flag;
+RedumpVerifier::DownloadStatus RedumpVerifier::m_gc_download_status;
+RedumpVerifier::DownloadStatus RedumpVerifier::m_wii_download_status;
+
+void RedumpVerifier::Start(const Volume& volume)
+{
+  // We use GetGameTDBID instead of GetGameID so that Datel discs will be represented by an empty
+  // string, which matches Redump not having any serials for Datel discs.
+  m_game_id = volume.GetGameTDBID();
```
BhaaLseN: Does it even make sense to trigger a download etc. when the game id is already empty?
Or do we just yolo it since Datel discs are really rare enough to be barely worth mentioning in code complexity?

Source/Core/DiscIO/VolumeVerifier.cpp
```diff
@@ -39,6 +49,276 @@
...
+    return;
+  }
+
+  File::IOFile(GetPathForPlatform(platform), "wb").WriteBytes(result->data(), result->size());
+}
+
+std::vector<u8> RedumpVerifier::ReadDatfile(const std::string& platform)
+{
+  unzFile file = unzOpen(GetPathForPlatform(platform).c_str());
+  Common::ScopeGuard file_guard{[&] { unzClose(file); }};
```
BhaaLseN: I mean, we do ignore the result of `unzClose`, but should be really be calling that when `file` is `nullptr`? According to a quick google this would return a parameter error (but not blow up or anything); so it _practically_ works but is _technically_ wrong.

Source/Core/DiscIO/VolumeVerifier.cpp
JosJuice: This is worth doing, because if the user has a good dump that is in redump then we will be able to confirm that it is a good dump. We just won't be able to know whether dumps that don't match a redump dump exactly are bad or not.

Source/Core/DiscIO/VolumeVerifier.cpp
JosJuice: > While redump does present a `DOCTYPE` declaration (for XHTML 1.1), I wonder if checking for the open angle bracket would be enough (so it still works when they do).

So it still works when they do what?

Source/Core/DiscIO/VolumeVerifier.cpp
BhaaLseN: How does Redump know if it doesn't have a game id?

JosJuice:
I've fixed the caching now.

Source/Core/DiscIO/VolumeVerifier.cpp
BhaaLseN: Remove the `DOCTYPE` (altho, the only thing they should be switching to at this point is potentially HTML5, which also has one), add copyright comment or whatever (which _also_ starts with `<!`) or...dunno, some other CDN shenanigans that leave the content same-ish but minutely mess with the first two bytes.

I suppose we should just drop the what-if and deal with it if/when it happens.

JosJuice: Yeah. I think the chance of them changing how the error message is written is larger than them making the page not start with `<!`, really. But redump doesn't tend to get a lot of changes anyway...

Hopefully they'll make the Wii datfile public soon, and then we can just remove this piece of code.

Source/Core/DiscIO/VolumeVerifier.cpp
JosJuice: I'm not quite sure what you mean by that, but Redump's serial field is empty for all Datel discs (since it actually is based on the serial printed on the disc label rather than any data stored on the disc and since Datel discs don't have any serial printed on the disc label).

JosJuice:
```
(12:28:32) Bh44L: JosJuice: if the datel disc has no game id, what are you using to find the matching candidate in the redump file?
(12:28:38) JosJuice: Just the hashes
(12:28:46) Bh44L: oh
(12:29:02) Bh44L: so its basically a reverse brute-force, gotcha
(12:30:01) Bh44L: i guess you can mark those two as resolved then
(12:30:12) JosJuice: Okay
```

Source/Core/DiscIO/VolumeVerifier.h
```diff
@@ -36,6 +36,72 @@ namespace DiscIO
...
+
+  std::string m_game_id;
+  u16 m_revision;
+  u8 m_disc_number;
+  u64 m_size;
+
+  std::future<std::vector<PotentialMatch>> m_future;
+  Result m_result;
+
+  static DownloadState m_gc_download_state;
```
lioncash: These can be `static inline` to avoid the need to redefine them in the cpp file.

JosJuice: I'm getting build errors on non-Windows now. Can this be resolved without removing the `inline`?

Source/Core/Common/MinizipUtil.h
```diff
@@ -0,0 +1,43 @@
...
+
+#include "Common/CommonTypes.h"
+#include "Common/ScopeGuard.h"
+
+namespace Common
+{
+// Since minzip doesn't provide a way to unzip a file of a length > 65535,
+// we have to implement this ourselves
+template <typename ContiguousContainer>
+static bool ReadCurrentFileUnlimited(unzFile file, ContiguousContainer* destination)
```
Warepire: After the move to Common, this function name bears little resemblance of the function purpose. Could it be renamed to something like `DeflateZipFile`?

It should also probably be noted in the comments that the function expects the container to be able to hold the requested data before calling this function.

mbc07:
Does it work for Wii games too? AFAIK Redump.org doesn't have a public DAT for Wii ISOs...

JosJuice:
It currently doesn't work, but the code is written so that it automatically will start working once the datfile becomes public. Thankfully, having a datfile for Wii games is less important than for GameCube games since Wii games have a lot of verification data built in.

Source/Core/DiscIO/VolumeVerifier.h
JosJuice: I've reverted the `inline` change for now.

Source/Core/DiscIO/VolumeVerifier.cpp
```diff
@@ -39,6 +49,291 @@
...
+    c -= 'a' - 'A';
+  if (c >= 'A')
+    c -= 'A' - ('9' + 1);
+  c -= '0';
+
+  if (c >= 0x10)
+    return 0;  // Error
+
+  return c;
+}
```
Pokechu22: It feels a bit complicated to have this, but [according to StackOverflow](https://stackoverflow.com/q/17261798) there isn't really a better way.

PR #8331, opened 2019-08-24
Add enc_temp_folder to .gitignore

JosJuice:
Not quite sure what this is, but Visual Studio creates it occasionally.


Antidote:
Appears to be related to this https://blogs.msdn.microsoft.com/zainnab/2010/06/30/autorecover/

PR #8332, opened 2019-08-25
DSPLLE: Put DSP thread in idle state when it's paused

CookiePLMonster:
Fixes a deadlock when accessing Graphics config **immediately** after booting up the game when using a LLE DSP.

DSP thread is considered "idle" when it signals `s_ppc_event` and waits for `s_dsp_event`, without putting it in this state when `m_dsp_thread_mutex` is locked it was possible to create a deadlock between a DSP thread, emulation thread and Qt thread.

### DIFF
--- a/Source/Core/Core/HW/DSPLLE/DSPLLE.cpp
+++ b/Source/Core/Core/HW/DSPLLE/DSPLLE.cpp
@@ -336,8 +336,16 @@ u32 DSPLLE::DSP_UpdateRate()
 void DSPLLE::PauseAndLock(bool do_lock, bool unpause_on_unlock)
 {
   if (do_lock)
+  {
     m_dsp_thread_mutex.lock();
+  }
   else
+  {
     m_dsp_thread_mutex.unlock();
+
+    // Signal the DSP thread so it can perform any outstanding work now (if any)
+    s_ppc_event.Wait();
+    s_dsp_event.Set();
+  }
 }
 }  // namespace DSP::LLE
--- a/Source/Core/Core/HW/DSPLLE/DSPLLE.cpp
+++ b/Source/Core/Core/HW/DSPLLE/DSPLLE.cpp
@@ -98,22 +98,24 @@ void DSPLLE::DSPThread(DSPLLE* dsp_lle)
     const int cycles = static_cast<int>(dsp_lle->m_cycle_count.load());
     if (cycles > 0)
     {
-      std::lock_guard<std::mutex> dsp_thread_lock(dsp_lle->m_dsp_thread_mutex);
-      if (g_dsp_jit)
+      std::unique_lock dsp_thread_lock(dsp_lle->m_dsp_thread_mutex, std::try_to_lock);
+      if (dsp_thread_lock)
       {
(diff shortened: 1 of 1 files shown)


JosJuice:
Should we signal `s_dsp_event` when unlocking the mutex on the PPC side? Not sure if it matters or not, but that way the DSP thread will get to do any work it failed to do as soon as possible.

CookiePLMonster:
> Should we signal `s_dsp_event` when unlocking the mutex on the PPC side? Not sure if it matters or not, but that way the DSP thread will get to do any work it failed to do as soon as possible.
> 
> EDIT: If we do this, I believe we also have to wait for `s_ppc_event` before setting `s_dsp_event`, because otherwise `DSP_Update` might consume `s_ppc_event` early and update `m_cycle_count` _before_ `DSPThread` overwrites it with 0.

I guess it makes sense. As for the edited part, do you think it maybe makes sense to restructure DSP thread to atomically exchange cycle count as being "consumed" and then run in a loop without going to sleep, to account for such cases? Looking at the code this may be non-trivial however, so maybe it makes more sense to subtract `cycles` on completion instead of storing 0.

JosJuice:
Yes, subtracting the cycles atomically seems like a good idea.

CookiePLMonster:
> Yes, subtracting the cycles atomically seems like a good idea.

Sadly compare-exchange is not possible, as not all code paths which load will also consume. That said, two atomic operations ought to be just fine here too.

CookiePLMonster:
I applied those changes. With those in place, `s_ppc_event` seems a bit useless, unless we don't want `DSP_Update` to furiously keep submitting work while DSP thread is busy.

JosJuice:
It's probably a good idea to not let the CPU thread run too far ahead of the DSP thread, yes. In fact, I'm wondering if the minor runahead (just one extra iteration) we can get from doing `s_dsp_event.Set()` without `s_ppc_event.Wait()` matters... Does anyone who knows more about the DSP threading want to comment on this? I think I would be comfortable with merging this PR either if we add `s_ppc_event.Wait()` to `PauseAndLock` or if someone is more confident than me that it won't cause any problems.

CookiePLMonster:
> if we add `s_ppc_event.Wait()` to `PauseAndLock`

If that is to happen, I'll probably revert the `do...while` loop in DSP thread as it will not be needed.

CookiePLMonster:
Done, now we're not overly eager on processing outstanding data. Since this is an edge case either way, I don't think it was worth the additional code anyway.

CookiePLMonster:
@JosJuice One thing I just noticed now is that `DSPLLE::PauseAndLock` will now freeze if `s_ppc_event` takes a while to signal - do you think this is safe to do? Ideally it should be signalled asynchronously but I don't know how to do it cleanly.

JosJuice:
By freeze you just mean until it gets signaled, right? If the mutex isn't locked by the DSP thread, that should happen very quickly (if it hasn't already happened), so it sounds fine.

CookiePLMonster:
> By freeze you just mean until it gets signaled, right?

Yeah. If that's fine then good :)

PR #8333, opened 2019-08-25
Core: Generate screenshot name with timestamps

CookiePLMonster:
Screenshot names in format `GAMEID-YYYY-MM-DD_HH-MM-SS.png` are not only more intuitive for the user, but also remove the need for this hideous loop which kept re-checking every possible suffix starting with `GAMEID-1.png'. If the user had thousands of screenshots from the same name, I doubt checking those every time was healthy.

In an unlikely case of name collisions (eg. multiple screenshots within a second), an increasing number gets appended to the filename.

### DIFF
--- a/Source/Core/Core/Core.cpp
+++ b/Source/Core/Core/Core.cpp
@@ -13,6 +13,7 @@
 #include <variant>
 
 #include <fmt/format.h>
+#include <fmt/time.h>
 
 #ifdef _WIN32
 #include <windows.h>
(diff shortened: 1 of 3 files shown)


MayImilae:
As someone with a collection of screenshots in a separate folder on a hard drive, I have a lot of R3ME01-18(2) nonsense. This will be SO much better for me!

CookiePLMonster:
> but you could've moved the (kinda unrelated) `std::move` changes to another commit.

Sure, I'll split them into two commits in a few hours.

CookiePLMonster:
I actually had it suggested yesterday too and I agree. I'll change it later today.

PR #8230, opened 2019-07-06
NetPlay: Remove 'reduce polling rate' option

Techjar:
With the SI poll line count fixed, pretty much all games are polling twice per frame anyways, making this option superfluous. Since it's a bit of a gross hack and makes DTMs incompatible with console, let's just bin it.

This is necessarily based on #8236, so don't merge this before that one.

PR #8334, opened 2019-08-26
VideoInterface: start counting half-lines at 0 instead of 1

hosaka-corp:
Found it easier to reason about `VideoInterface::Update()` if we start counting half-lines at 0 instead of 1. 

- Re-organize VideoInterface::Update() to count half-lines starting at 0 instead of 1
- Use horizontal position when checking if we should assert some display interrupt. This should make them slightly more accurate to the hardware (granularity changes to half-lines, instead of only checking on vertical line boundaries)
- Add some more descriptive comments

### DIFF
--- a/Source/Core/Core/HW/VideoInterface.cpp
+++ b/Source/Core/Core/HW/VideoInterface.cpp
@@ -191,7 +191,7 @@ void Preset(bool _bNTSC)
   m_BorderHBlank.Hex = 0;
 
   s_ticks_last_line_start = 0;
-  s_half_line_count = 1;
+  s_half_line_count = 0;
   s_half_line_of_next_si_poll = num_half_lines_for_si_poll;  // first sampling starts at vsync
   s_current_field = FieldType::Odd;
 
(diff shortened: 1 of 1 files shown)


hosaka-corp:
Fixed to account for impending #8230

Source/Core/Core/HW/VideoInterface.cpp
```diff
@@ -761,30 +774,30 @@ void Update(u64 ticks)
...
   }
 
-  if (s_half_line_count & 1)
+  // Check if we need to assert IR_INT. Note that the granularity of our current horizontal
+  // position is limited to half-lines.
+
+  for (UVIInterruptRegister& reg : m_InterruptRegister)
   {
-    s_ticks_last_line_start = CoreTiming::GetTicks();
+    u32 h_thresh = (reg.HCT > m_HTiming0.HLW) ? 1 : 0;
```
BhaaLseN: No need to save on letters here; especially when your PR is meant to make it easier to reason about code ;)
```suggestion
    u32 horizontal_threshold = (reg.HCT > m_HTiming0.HLW) ? 1 : 0;
```
Not sure if thats the best name though (or whether my guesses of what `h` and `thresh` were meant to stand for were correct).

Source/Core/Core/HW/VideoInterface.cpp
```diff
@@ -317,7 +317,7 @@ void RegisterMMIO(MMIO::Mapping* mmio, u32 base)
   // MMIOs with unimplemented writes that trigger warnings.
   mmio->Register(
       base | VI_VERTICAL_BEAM_POSITION,
-      MMIO::ComplexRead<u16>([](u32) { return 1 + (s_half_line_count - 1) / 2; }),
+      MMIO::ComplexRead<u16>([](u32) { return 1 + (s_half_line_count) / 2; }),
```
adroitwhiz: are the parentheses around `s_half_line_count` necessary anymore?

PR #8336, opened 2019-08-26
Add Dolphin version and current video backend to shader compilation logs

Ebola16:
Implements https://bugs.dolphin-emu.org/issues/9992

This makes it easier to quickly identify what version of Dolphin and video backend produced shader compilation issues.

### DIFF
--- a/Source/Core/VideoBackends/D3DCommon/Shader.cpp
+++ b/Source/Core/VideoBackends/D3DCommon/Shader.cpp
@@ -10,8 +10,11 @@
 #include "Common/Logging/Log.h"
 #include "Common/MsgHandler.h"
 #include "Common/StringUtil.h"
+#include "Common/Version.h"
 
 #include "VideoBackends/D3DCommon/Shader.h"
+
+#include "VideoCommon/VideoBackendBase.h"
 #include "VideoCommon/VideoConfig.h"
 
 namespace D3DCommon
@@ -105,13 +108,15 @@ std::optional<Shader::BinaryData> Shader::CompileShader(D3D_FEATURE_LEVEL featur
   if (FAILED(hr))
   {
     static int num_failures = 0;
-    std::string filename = StringFromFormat(
-        "%sbad_%s_%04i.txt", File::GetUserPath(D_DUMP_IDX).c_str(), target, num_failures++);
+    std::string filename = VideoBackendBase::BadShaderFilename(target, num_failures++);
     std::ofstream file;
     File::OpenFStream(file, filename, std::ios_base::out);
     file.write(source.data(), source.size());
     file << "\n";
     file.write(static_cast<const char*>(errors->GetBufferPointer()), errors->GetBufferSize());
+    file << "\n";
+    file << "Dolphin Version: " + Common::scm_rev_str + "\n";
+    file << "Video Backend: " + g_video_backend->GetDisplayName();
     file.close();
 
     PanicAlert("Failed to compile %s:\nDebug info (%s):\n%s", filename.c_str(), target,
--- a/Source/Core/VideoCommon/VideoBackendBase.cpp
+++ b/Source/Core/VideoCommon/VideoBackendBase.cpp
@@ -10,10 +10,13 @@
 #include <string>
 #include <vector>
 
+#include "fmt/format.h"
+
 #include "Common/ChunkFile.h"
 #include "Common/CommonTypes.h"
 #include "Common/Event.h"
 #include "Common/Logging/Log.h"
+
 #include "Core/ConfigManager.h"
 #include "Core/Core.h"
 
@@ -67,6 +70,12 @@ extern "C" {
 __declspec(dllexport) DWORD NvOptimusEnablement = 1;
 }
 #endif
+
+std::string VideoBackendBase::BadShaderFilename(const char* shader_stage, int counter)
+{
+  return fmt::format("{}bad_{}_{}_{}.txt", File::GetUserPath(D_DUMP_IDX), shader_stage,
+                     g_video_backend->GetName(), counter);
+}
 
 void VideoBackendBase::Video_ExitLoop()
 {
--- a/Source/Core/VideoBackends/OGL/ProgramShaderCache.cpp
+++ b/Source/Core/VideoBackends/OGL/ProgramShaderCache.cpp
@@ -17,6 +17,7 @@
 #include "Common/Logging/Log.h"
 #include "Common/MsgHandler.h"
 #include "Common/StringUtil.h"
+#include "Common/Version.h"
 
 #include "Core/ConfigManager.h"
 
@@ -31,6 +32,7 @@
 #include "VideoCommon/Statistics.h"
 #include "VideoCommon/VertexLoaderManager.h"
 #include "VideoCommon/VertexShaderManager.h"
+#include "VideoCommon/VideoBackendBase.h"
 
 namespace OGL
 {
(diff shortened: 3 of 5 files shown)


stenzek:
While potentially useful for debugging, this will actually force drivers which use a hash of the shader code as an identifier to recompile all shaders every Dolphin build.

Ebola16:
Ok, all of these concerns should be fixed if I add this information to the filename instead. Is this better?

Source/Core/VideoBackends/D3DCommon/Shader.cpp
```diff
@@ -105,8 +108,9 @@ std::optional<Shader::BinaryData> Shader::CompileShader(D3D_FEATURE_LEVEL featur
   if (FAILED(hr))
   {
     static int num_failures = 0;
-    std::string filename = StringFromFormat(
-        "%sbad_%s_%04i.txt", File::GetUserPath(D_DUMP_IDX).c_str(), target, num_failures++);
+    std::string filename = StringFromFormat("%sbad_%s_%04i_", File::GetUserPath(D_DUMP_IDX).c_str(),
```
leoetlino: It might be worth extracting the filename generation logic to a separate function that you can reuse, instead of having to include Common/Version everywhere. Also, it's probably better not to mix StringFromFormat(...) and manual string concatenation. Just add the version and backend name part of the format string

Source/Core/VideoBackends/D3DCommon/Shader.cpp
```diff
@@ -105,8 +108,9 @@ std::optional<Shader::BinaryData> Shader::CompileShader(D3D_FEATURE_LEVEL featur
...
+                                            target, num_failures++) +
+                           Common::scm_rev_str + "_" + g_video_backend->GetDisplayName() + ".txt";
```
leoetlino: Not a big fan of using a global just to get the video backend name, when it is already known. Could you just replace this with "D3D" (and OGL/Vulkan below)?

Ebola16: I'd like to differentiate D3D11 vs D3D12. Is there a better way to accomplish this?

Source/Core/VideoBackends/D3DCommon/Shader.cpp
Ebola16: Not sure if I did this correctly but I think I properly used StringFromFormat in the latest commit. And the filename format isn't consistent across all cases (see the changes in this PR for OGL). Also, wouldn't we need to include Version.h anyway to avoid indirect includes?

leoetlino: >Also, wouldn't we need to include Version.h everywhere anyway to avoid indirect includes?

No, you wouldn't have to expose implementation details in the header.

Source/Core/VideoBackends/D3DCommon/Shader.cpp
leoetlino: Oh right, I forgot that this is used for both D3D11 and D3D12. I'm not sure if there's a better way to do that... Maybe @stenzek would know?

Ebola16:
@leoetlino Is this better?

Ebola16:
Ah, I forgot about @stenzek's review. This still forces shader recompilation every time. I'll look into that.

Ebola16:
I can't reproduce @stenzek's  concerns anymore but it's worth a check.

Ebola16:
@stenzek?

Source/Core/VideoCommon/VideoBackendBase.cpp
```diff
@@ -68,6 +72,12 @@ __declspec(dllexport) DWORD NvOptimusEnablement = 1;
 }
 #endif
 
+std::string VideoBackendBase::BadShaderFilename(const char* shader_stage, int counter)
+{
+  return fmt::format("{}bad_{}_{}_{}_{}.txt", File::GetUserPath(D_DUMP_IDX), shader_stage,
+                     Common::scm_rev_str, g_video_backend->GetDisplayName(), counter);
```
stenzek: Are you sure you want the display name here, instead of the name? This would generate filenames containing say, "Direct3D 12" instead of "D3D12".

Ebola16: I forgot about `GetName`. Changed.

Ebola16:
@stenzek I think I like adding Dolphin version and current video backend to just the filename more now. Since the bad shaders are likely to raise errors themselves, adding a log message would probably be redundant. This PR will help identify where bad shaders originate, which is useful if you're testing a build that generates a lot of them and then forget to clear them. I think this PR is ready for merge now!

stenzek:
By "log file" I was referring to the `bad_nnnn.txt` file. You could tack something like "Dolphin version NNN, video backend YYY" and any other relevant configuration options at the end of the file.

Putting the version in the filename has the potential to leave lots of files in the dump directory, especially if the user has a buggy driver and tries multiple Dolphin versions. At least with the current behavior, those files will be overwritten each time.

I think putting the backend name in the filename is a good idea though.

Ebola16:
My alphabetically sorted naming convention produces:
bad_vs_5_0_Dolphin [PS] 5.0-11429-dirty_D3D_0
bad_vs_5_0_Dolphin [PS] 5.0-11429-dirty_D3D_1
bad_vs_5_0_Dolphin [PS] 5.0-11429-dirty_D3D12_0
bad_vs_5_0_Dolphin [PS] 5.0-11429-dirty_D3D12_1
bad_vs_5_0_Dolphin [PS] 5.0-11429-dirty_OGL_0
bad_vs_5_0_Dolphin [PS] 5.0-11429-dirty_OGL_1

Moving the counter before the version and video backend could would split up the video backends, producing less desirable results. It also may be confused as part of the shader stage.
bad_vs_5_0_0_Dolphin [PS] 5.0-11429-dirty_D3D
bad_vs_5_0_0_Dolphin [PS] 5.0-11429-dirty_D3D12
bad_vs_5_0_0_Dolphin [PS] 5.0-11429-dirty_OGL
bad_vs_5_0_1_Dolphin [PS] 5.0-11429-dirty_D3D
bad_vs_5_0_1_Dolphin [PS] 5.0-11429-dirty_D3D12
bad_vs_5_0_1_Dolphin [PS] 5.0-11429-dirty_OGL

I think I prefer my naming convention for the reasons stated above.

Also, I was hoping to include the version since that would be immensely helpful for determining if a bad shader file is still relevant. When I encounter bad shaders (thankfully only on old builds), it's always an endless spam of them. Are there cases where bad shaders only appear a few times per Dolphin session? If not, does it really matter if we add to the endless spam? I suppose we could add an option to disable bad shader logs but that's probably for a separate PR.

Ebola16:
Oh wait, I just understood what you said. Give me a few moments to try something.

stenzek:
I'm less concerned about the ordering of the fields in the filename and more about the version itself being included. Let's say a bad driver causes 20 files to be generated, if a user tries 5 Dolphin versions thinking it's a bug on our end, they'll end up 100 dump files in the folder. If the version string is not included, there would still only be 20.

As someone who works on the shader generators, having to manually clean the folder all the time would also get annoying. With the current behavior, I know that the 0th or 1st file is always going to be the first bad shader from the current version. With the version in the filename, I'd first have to sort by the timestamp.

Hence why I suggested appending the version string and any other info the end of the file itself - that way the information is still there if needed for a bug report, etc, but doesn't clog up the folder with additional, duplicate files.

Ebola16:
@stenzek Updated with requested changes.

PR #8337, opened 2019-08-28
Log widget improvements

CookiePLMonster:
This PR largely overhauls LogWidget to drastically decrease its memory usage and negative impact on UI responsiveness.

In the worst case, spammy enough logs will cause LogWidget to lock up an entire Qt UI (as reported by @JMC47 and confirmed by my tests)). Additionally, both the widget itself and its log buffering functionality (designed so it never updates more than 200 lines in one go) were unbounded and could grow infinitely, taking GB's of RAM and maybe even causing Qt to run out of memory (I think that's why UI locked up eventually).

My changes include:
- **Most importantly**, removal of `QTextEdit` in favour of `QPlainTextEdit`. The former is a full WYSIWYG editor with support for images, tables and rich text features we **absolutely** don't need in a log viewer! `QPlainTextEdit` is closer in functionality to Notepad with syntax highlighting (whereas `QTextEdit` could be compared to WordPad), and Qt documentation recommends using it for log viewers.
- Removal of a synchronization mutex, because all calls to the buffer queue are done from the same thread.
- Reconstruction of `LogWidget::Log` to make as little memory allocations and string copies as reasonably possible.
- Removal of unused widgets (one of which was actually leaking) and class fields from `LogWidget`.
- Addition of `MAXIMUM_BLOCK_COUNT` constant, currently set to 5000. Both log viewer widget and the log queue are now limited to only store this many most recent messages. Not only this drastically improves performance, but also prevents memory usage from exploding with spammy enough logs. If there is a need to preview more than 5000 recent log lines, please log to file or to a console.
- Addition of additional constraints to widget updates. Now logs get appended to a Qt object only if widget is visible. This further improves UI responsiveness, making it completely fluid even with the most spammy logs, with hitches occuring **only** if Log tab is visible.

### DIFF
--- a/Source/Core/DolphinQt/Config/LogWidget.cpp
+++ b/Source/Core/DolphinQt/Config/LogWidget.cpp
@@ -46,7 +46,12 @@ LogWidget::LogWidget(QWidget* parent) : QDockWidget(parent), m_timer(new QTimer(
   ConnectWidgets();
 
   connect(m_timer, &QTimer::timeout, this, &LogWidget::UpdateLog);
-  m_timer->start(UPDATE_LOG_DELAY);
+  connect(this, &QDockWidget::visibilityChanged, [this](bool visible) {
+    if (visible)
+      m_timer->start(UPDATE_LOG_DELAY);
+    else
+      m_timer->stop();
+  });
 
   connect(&Settings::Instance(), &Settings::DebugFontChanged, this, &LogWidget::UpdateFont);
 
--- a/Source/Core/DolphinQt/Config/LogWidget.cpp
+++ b/Source/Core/DolphinQt/Config/LogWidget.cpp
@@ -19,8 +19,6 @@
 
 // Delay in ms between calls of UpdateLog()
 constexpr int UPDATE_LOG_DELAY = 100;
-// Maximum number of lines to show in log viewer
-constexpr int MAX_LOG_LINES = 5000;
 // Maximum lines to process at a time
 constexpr size_t MAX_LOG_LINES_TO_UPDATE = 200;
 // Timestamp length
@@ -70,16 +68,13 @@ void LogWidget::UpdateLog()
   std::vector<LogEntry> elements_to_push;
   {
     std::lock_guard lock(m_log_mutex);
-    if (m_log_queue.empty())
+    if (m_log_ring_buffer.empty())
       return;
 
-    elements_to_push.reserve(std::min(MAX_LOG_LINES_TO_UPDATE, m_log_queue.size()));
+    elements_to_push.reserve(std::min(MAX_LOG_LINES_TO_UPDATE, m_log_ring_buffer.size()));
 
-    for (size_t i = 0; !m_log_queue.empty() && i < MAX_LOG_LINES_TO_UPDATE; i++)
-    {
-      elements_to_push.push_back(std::move(m_log_queue.front()));
-      m_log_queue.pop();
-    }
+    for (size_t i = 0; !m_log_ring_buffer.empty() && i < MAX_LOG_LINES_TO_UPDATE; i++)
+      elements_to_push.push_back(std::move(m_log_ring_buffer.pop_front()));
   }
 
   for (auto& line : elements_to_push)
@@ -168,7 +163,7 @@ void LogWidget::ConnectWidgets()
 {
   connect(m_log_clear, &QPushButton::clicked, [this] {
     m_log_text->clear();
-    m_log_queue = {};
+    m_log_ring_buffer.clear();
   });
   connect(m_log_wrap, &QCheckBox::toggled, this, &LogWidget::SaveSettings);
   connect(m_log_font, static_cast<void (QComboBox::*)(int)>(&QComboBox::currentIndexChanged), this,
@@ -219,8 +214,8 @@ void LogWidget::Log(LogTypes::LOG_LEVELS level, const char* text)
     text_length--;
 
   std::lock_guard lock(m_log_mutex);
-  m_log_queue.emplace(std::piecewise_construct, std::forward_as_tuple(text, text_length),
-                      std::forward_as_tuple(level));
+  m_log_ring_buffer.emplace(std::piecewise_construct, std::forward_as_tuple(text, text_length),
+                            std::forward_as_tuple(level));
 }
 
 void LogWidget::closeEvent(QCloseEvent*)
--- a/Source/Core/DolphinQt/Config/LogWidget.cpp
+++ b/Source/Core/DolphinQt/Config/LogWidget.cpp
@@ -8,12 +8,13 @@
 #include <QComboBox>
 #include <QFont>
 #include <QFontDatabase>
-#include <QGroupBox>
+#include <QGridLayout>
+#include <QPlainTextEdit>
 #include <QPushButton>
 #include <QScrollBar>
-#include <QTextEdit>
 #include <QTimer>
-#include <QVBoxLayout>
+
+#include "Core/ConfigManager.h"
 
 #include "Common/FileUtil.h"
 #include "Common/StringUtil.h"
@@ -25,8 +26,10 @@
 
 // Delay in ms between calls of UpdateLog()
 constexpr int UPDATE_LOG_DELAY = 100;
+// Maximum number of lines to show in log viewer
+constexpr int MAX_LOG_LINES = 5000;
 // Maximum lines to process at a time
-constexpr int MAX_LOG_LINES = 200;
+constexpr size_t MAX_LOG_LINES_TO_UPDATE = 200;
 // Timestamp length
 constexpr int TIMESTAMP_LENGTH = 10;
 
@@ -75,9 +78,9 @@ void LogWidget::UpdateLog()
   int old_horizontal = hscroll->value();
   int old_vertical = vscroll->value();
 
-  for (int i = 0; !m_log_queue.empty() && i < MAX_LOG_LINES; i++)
+  for (size_t i = 0; !m_log_queue.empty() && i < MAX_LOG_LINES_TO_UPDATE; i++)
   {
-    m_log_text->append(m_log_queue.front());
+    m_log_text->appendHtml(m_log_queue.front());
     m_log_queue.pop();
   }
 
@@ -115,8 +118,7 @@ void LogWidget::UpdateFont()
 void LogWidget::CreateWidgets()
 {
   // Log
-  m_tab_log = new QWidget;
-  m_log_text = new QTextEdit;
+  m_log_text = new QPlainTextEdit;
   m_log_wrap = new QCheckBox(tr("Word Wrap"));
   m_log_font = new QComboBox;
   m_log_clear = new QPushButton(tr("Clear"));
@@ -124,7 +126,6 @@ void LogWidget::CreateWidgets()
   m_log_font->addItems({tr("Default Font"), tr("Monospaced Font"), tr("Selected Font")});
 
   auto* log_layout = new QGridLayout;
-  m_tab_log->setLayout(log_layout);
   log_layout->addWidget(m_log_wrap, 0, 0);
   log_layout->addWidget(m_log_font, 0, 1);
   log_layout->addWidget(m_log_clear, 0, 2);
@@ -136,6 +137,8 @@ void LogWidget::CreateWidgets()
   setWidget(widget);
 
   m_log_text->setReadOnly(true);
+  m_log_text->setUndoRedoEnabled(false);
+  m_log_text->setMaximumBlockCount(MAX_LOG_LINES);
 
   QPalette palette = m_log_text->palette();
   palette.setColor(QPalette::Base, Qt::black);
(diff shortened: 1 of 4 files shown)


Ebola16:
Fixes https://bugs.dolphin-emu.org/issues/11696

Source/Core/DolphinQt/Config/LogWidget.h
```diff
@@ -46,5 +46,7 @@ class LogWidget final : public QDockWidget, LogListener
   QPushButton* m_log_clear;
   QPlainTextEdit* m_log_text;
 
+  QTimer* m_timer;
```
BhaaLseN: Kinda funny that you removed this in the 2nd commit only to re-add it back in the last one ;)

CookiePLMonster: There's more to come! I'll be re-adding a mutex too :)

I think I'll rewrite history after testing is done and streamline such changes.

BhaaLseN: Oh, good to know; guess I approved early then

CookiePLMonster:
Idea: Since `m_log_queue` now has a fixed maximum size, I could replace it with a ring buffer (implemented on a `std::vector` for example) and remove **any** container allocations in the process, as it'd be preserved up front and never grown nor shrunk.

CookiePLMonster:
@Ebola16 @JMC47 I updated the PR with a removal of `QueueOnObject` calls. Logging workflow now uses a lock (again...) to put log lines in a buffer queue directly as opposed to queueing a Qt signals. I'm assuming those Qt signals had to use some synchronization internally anyway, and with sufficient log spam we were potentially sending thousands of those - so now it's done explicit by taking a lock and adding things to a queue, without clogging Qt signal queue.

`UpdateLog` is now move heavyweight, of course - but the upside is that `std::string -> `QString` conversions are now done **only** when log is actually displaying, whereas earlier they were done always, even if log lines were to be thrown away without being displayed ever. While I could perform this work without moving strings to a temporary `std::vector` before, I figured it makes sense to move them as then I can safely release the lock.

My profiling shows that this change improves logging performance further, but testing is appreciated.


There will be more coming (a ring buffer I mentioned earlier), so performance may still be in non-final state. Seeing how I already corrected two design decisions in consecutive commits, I think I'll rewrite history completely before removing the [WIP] tag.

Source/Core/DolphinQt/Config/LogWidget.cpp
```diff
@@ -189,40 +220,20 @@ void LogWidget::SaveSettings()
 
 void LogWidget::Log(LogTypes::LOG_LEVELS level, const char* text)
 {
-  // The text has to be copied here as it will be deallocated after this method has returned
-  std::string str(text);
+  // The text has to be copied into the lambda as it will be deallocated after this method has
```
lioncash: This comment is no longer relevant, given the lambda was removed from here.

Source/Core/DolphinQt/Config/LogWidget.h
```diff
@@ -5,6 +5,7 @@
 #pragma once
 
 #include <QDockWidget>
+#include <QPlainTextEdit>
```
lioncash: This can be replaced with a forward declaration (the existing forward declarations can also be trimmed of the types you removed as well).

Source/Core/DolphinQt/Config/LogWidget.h
```diff
@@ -43,12 +44,10 @@ class LogWidget final : public QDockWidget, LogListener
...
-  QVBoxLayout* m_main_layout;
-  QTextEdit* m_log_text;
-  QWidget* m_tab_log;
+  QPlainTextEdit* m_log_text;
 
   QTimer* m_timer;
 
   std::mutex m_log_mutex;
-  std::queue<QString> m_log_queue;
+  std::queue<std::pair<std::string, LogTypes::LOG_LEVELS>> m_log_queue;
```
lioncash: `<string>` needs to be included.

Source/Core/DolphinQt/Config/LogWidget.cpp
```diff
@@ -60,36 +73,50 @@ LogWidget::~LogWidget()
...
+  std::vector<std::pair<std::string, LogTypes::LOG_LEVELS>> elements_to_push;
+  {
+    std::lock_guard lock(m_log_mutex);
+    if (m_log_queue.empty())
+      return;
 
-  // If the vertical scrollbar is within 50 units of the maximum value, count it as being at the
-  // bottom
-  bool vscroll_bottom = vscroll->maximum() - vscroll->value() < 50;
+    elements_to_push.reserve(MAX_LOG_LINES);
```
BhaaLseN: Would this be more practical as `m_log_queue.size()` instead, since there can be at most `MAX_LOG_LINES` in there, with a good chance that there is less (for non-debug builds with non-spammy listeners)?

CookiePLMonster: `m_log_queue` size can be `MAXIMUM_BLOCK_COUNT` at most. `MAX_LOG_LINES` is the maximum amount of log lines to be outputted to the widget in one iteration.

That said, `std::min(MAX_LOG_LINES, m_log_queue.size())` is indeed a good idea here :)

CookiePLMonster: I should also probably rename those constants as this proves they're confusingly named.

BhaaLseN: Oops, mixing up constants, but yeah; you caught my drift 👍

CookiePLMonster:
Commit history has been completely remade to make more sense and to remove intermediate experiments. According to `git diff` the end result is 1:1 identical with what was submitted before.

This is now ready for final reviews!

Ebola16:
I didn't do any quantitative analysis but the latest version of this PR seems to be even better! My test case was the workaround for https://bugs.dolphin-emu.org/issues/10791 (DSP LLE recompiler when using Project M Launcher and its derivatives). This tends to spam a TON of error messages with certain builds. After exiting emulation, the error message spam stops significantly sooner in this PR when compared to Master.

Function-wise LGTM but I don't feel comfortable enough with my coding skills to review the code itself.

Source/Core/DolphinQt/Config/LogWidget.cpp
```diff
@@ -8,29 +8,32 @@
...
-#include "DolphinQt/QtUtils/QueueOnObject.h"
 #include "DolphinQt/Settings.h"
 
 // Delay in ms between calls of UpdateLog()
 constexpr int UPDATE_LOG_DELAY = 100;
 // Maximum lines to process at a time
-constexpr int MAX_LOG_LINES = 200;
+constexpr int MAX_LOG_LINES_TO_UPDATE = 200;
+// Maximum number of lines to show in log viewer
+constexpr int MAX_LOG_LINES = 5000;
```
Warepire: If you make this of type `size_t`, you don't need the specialization of `std::min()` on line 77

Source/Core/DolphinQt/Config/LogWidget.h
```diff
@@ -43,12 +44,99 @@ class LogWidget final : public QDockWidget, LogListener
...
+
+    reference front() noexcept
+    {
+      return const_cast<reference>(const_cast<const circular_buffer*>(this)->front());
+    }
+
+    // If buffer is empty, behaviour is undefined
+    void pop() noexcept
+    {
+      reinterpret_cast<pointer>(std::launder(&m_container[m_tail]))->~value_type();
```
Warepire: Note: C++20 marks std::launder as `[[nodiscard]]`. This would discard the return value.

CookiePLMonster: Laundered pointer is used for a method call (`~value_type()`) so it's not really discarded and `[[nodiscard]]` doesn't complain about it.

Source/Core/DolphinQt/Config/LogWidget.h
```diff
@@ -43,12 +44,99 @@ class LogWidget final : public QDockWidget, LogListener
...
 
   QTimer* m_timer;
 
+  // Ring buffer implementation, heavily based on:
+  // https://embeddedartistry.com/blog/2017/4/6/circular-buffers-in-cc#c-=
+  // Modified to be an almost complete drop-in replacement for std::queue
+  // It does not implement some methods compared to std::queue,
+  // and also implements capacity(), clear() and full(), which std::queue does not
+  template <class T>
+  class circular_buffer
```
lioncash: Our naming conventions for classes is `CamelCase`.

Source/Core/DolphinQt/Config/LogWidget.h
```diff
@@ -43,12 +44,99 @@ class LogWidget final : public QDockWidget, LogListener
...
+  public:
+    using value_type = T;
+    using size_type = std::size_t;
+    using reference = T&;
+    using const_reference = const T&;
+
+    explicit circular_buffer(std::size_t count)
+        : m_container(std::make_unique<storage[]>(count)), m_capacity(count)
+    {
+    }
```
lioncash: The destructor for this should really be invoking the destructor of the placed types should it not? Otherwise any type with a non-trivial destructor won't actually execute. If it exists within the buffer when its lifetime scope ends as far as I can tell, which is dangerous.

Source/Core/DolphinQt/Config/LogWidget.h
CookiePLMonster: I hear you, but I consciously decided against it here because I wanted it to mimick STL containers style as closely as possible.

lioncash: The commit introducing this should also really have rationale or an explanation for its introduction. It may be obvious to you, but not necessarily for others; especially without a justifying reason for it.

Source/Core/DolphinQt/Config/LogWidget.h
CookiePLMonster: Snap, you're absolutely right! I'll fix it, good catch!

Source/Core/DolphinQt/Config/LogWidget.h
CookiePLMonster: I pushed a fix for incomplete destructor and also included a rationale in the commit message which should be exhaustive enough. I have **not** renamed the class for reasons given above.

Source/Core/DolphinQt/Config/LogWidget.h
```diff
@@ -47,8 +48,97 @@ class LogWidget final : public QDockWidget, LogListener
...
+  private:
+    using storage = std::aligned_storage_t<sizeof(value_type), alignof(value_type)>;
+    using pointer = T*;
+    using const_pointer = const T*;
+
+    std::unique_ptr<storage[]> m_container;
+    std::size_t m_head = 0;
+    std::size_t m_tail = 0;
+    const std::size_t m_capacity;
+    bool m_full = false;
```
jordan-woyak: Maintaining this `m_full` member bothers me. `m_tail` and `m_full` can be replaced by a single `m_size` member. Also, it looks like `pop()` can be implemented in terms of `front()`. The tail calculation should then only be needed in one place (`front()`'s implementation). Or you could even reverse the storage order to only need the calculation on a push. Update `m_size` accordingly. `size()`'s implementation is then simply `return m_size;` instead of an else-if-ladder.

CookiePLMonster: I'll look into this.

CookiePLMonster: `circular_buffer` is now obsolete, turns out we had a (slightly bugged) `FixedSizeQueue` for the same thing. The only thing it doesn't have compared to `circular_buffer` is construction behaviour - `FixedSizeQueue` holds default-initialized elements, while `circular_buffer` held unallocated buffers like STL.

Source/Core/DolphinQt/Config/LogWidget.h
CookiePLMonster: `circular_buffer` is now replaced in favour of `FixedSizeQueue`, see below.

CookiePLMonster:
`circular_buffer` is now obsolete, turns out we had a (slightly bugged) `FixedSizeQueue` for the same thing. The only thing it doesn't have compared to `circular_buffer` is construction behaviour - `FixedSizeQueue` holds default-initialized elements, while `circular_buffer` held unallocated buffers like STL.

Source/UnitTests/Common/FixedSizeQueueTest.cpp
```diff
@@ -31,3 +31,83 @@ TEST(FixedSizeQueue, Simple)
...
+  EXPECT_EQ(5u, q.size());
+  EXPECT_EQ(1, q.pop_front());
+
+  EXPECT_EQ(4u, q.size());
+}
+
+class TestCaseClass
+{
+public:
+  static inline int num_objects = 0;  // Static fields the reason we cannot define it locally
```
Warepire: This comment is a lot of words without any meaning at all.

Warepire: Your test class is overly complicated. If you use member variables you don't need to reset them in the destructor. The class is re-created by googletest for each TEST_F.

PR #8338, opened 2019-08-29
Enable EFB access in INI

Miksel12:
Because of the EFB access optimization, games which were forced to skip EFB access in order to be playable don't need to be forced anymore. 
There are also reports of skip EFB access causing issues in Baten Kaitos: https://forums.dolphin-emu.org/Thread-gc-baten-kaitos-eternal-wings-and-the-lost-ocean?pid=493827#pid493827

Monster Hunter also doesn't need EFB to texture and ram.

### DIFF
--- a/Data/Sys/GameSettings/GKB.ini
+++ b/Data/Sys/GameSettings/GKB.ini
@@ -15,8 +15,5 @@ CPUThread = 0
 
 [Video_Hacks]
 ImmediateXFBEnable = False
-EFBAccessEnable = False
 EFBToTextureEnable = False
 DeferEFBCopies = False
-
-[Video_Settings]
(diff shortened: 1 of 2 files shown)


mbc07:
Are you sure about Monster Hunter? As far as I remember the character creation and the bloom is broken without EFB2RAM...

Miksel12:
Those effects use EFB access, and don't need EFB to ram.

JMC47:
I can confirm.  I'm nervous about enabling it because it was so slow before, but I believe this is the correct decision.

JMC47:
@dolphin-emu-bot rebuild

stenzek:
@dolphin-emu-bot rebuild

PR #8239, opened 2019-07-13
[WIP] SI: Rework SI command architecture

booto:
[WIP: Opening to generate builds for testers.]

This builds on #8236, but is separated out because this is potentially more dangerous, and it shouldn't hold up those changes.

This is a rework of the SI command architecture, causing it to be event based with distinct sending command/getting response stages.

PR #8339, opened 2019-08-29
SI: Reduce logging.

booto:
An earlier SI PR added logging for SI devices that return the incorrect number of bytes for certain commands. It turns out pretty much every SI device in dolphin returns the wrong number of bytes (#8239 mostly fixes this). For the time being, it results in loads of non-actionable log spam for users that should probably be suppressed.


Ebola16:
Conceals https://bugs.dolphin-emu.org/issues/11821

booto:
Yes, this was the point. The logs are useless for end-users and the issue is dealt with in a separate PR.

BhaaLseN:
Should we leave a comment next to it for later?

booto:
I don't think that's necessary.

PR #8340, opened 2019-08-30
DolphinQt: Don't overwrite >8x IR scale in ini, add maximum internal res option

stenzek:
This adds an ini-only setting under GFX.ini -> [Settings] -> MaxInternalResolution. Setting this will allow the user to select resolutions beyond the default 8x max scale in graphics options.

Also fixes the bug if a user had a >8xIR set, opening the graphics window would overwrite the value, and save it to the ini.

Requested by @MayImilae.

### DIFF
--- a/Source/Core/DolphinQt/Config/Graphics/EnhancementsWidget.cpp
+++ b/Source/Core/DolphinQt/Config/Graphics/EnhancementsWidget.cpp
@@ -26,6 +26,7 @@
 
 #include "VideoCommon/PostProcessing.h"
 #include "VideoCommon/VideoBackendBase.h"
+#include "VideoCommon/VideoCommon.h"
 #include "VideoCommon/VideoConfig.h"
 
 EnhancementsWidget::EnhancementsWidget(GraphicsWindow* parent)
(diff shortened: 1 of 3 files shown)


Source/Core/DolphinQt/Config/Graphics/EnhancementsWidget.cpp
```diff
@@ -48,21 +49,28 @@ void EnhancementsWidget::CreateWidgets()
...
+
+  // If the current scale is greater than the max scale in the ini, add sufficient options so that
+  // when the settings are saved we don't lose the user-modified value from the ini.
+  const int max_efb_scale =
+      std::max(Config::Get(Config::GFX_EFB_SCALE), Config::Get(Config::GFX_MAX_EFB_SCALE));
+  for (int scale = static_cast<int>(resolution_options.size()); scale <= max_efb_scale; scale++)
   {
-    m_ir_combo->addItem(tr("Custom"));
-    m_ir_combo->setCurrentIndex(m_ir_combo->count() - 1);
+    resolution_options.append(QString::asprintf("%dx Native (%dx%d)", scale,
```
CookiePLMonster:
```suggestion
resolution_options.append(QString::asprintf("%dx Native (%dx%d)", scale,
resolution_options.append(QString("%1x Native (%2x%3)").arg(QString::number(scale),
                                            QString::number(static_cast<int>(EFB_WIDTH) * scale),
                                            QString::number(static_cast<int>(EFB_HEIGHT) * scale));
```

Qt docs do not recommend using `asprintf` and recommend using `arg` instead.

JosJuice: The string should use `tr`.

MayImilae:
I've been using this PR exclusively. This is SO NICE, and fixes a genuine bug too!

jordan-woyak:
I don't like the idea of a setting that exposes additional settings in the UI. Can we eliminate the ini overwrites without the meta-setting?

MayImilae:
The reason for the meta-setting is that it allows power users to expose more options in the GUI (and the tremendous convenience boost it offers) without exposing it to the majority of users.

As an 8k display user, I can tell you, the situation right now is rubbish. Even if the bug of setting internal resolution to -1 is fixed, having to set the internal resolution in the INI exclusively is a paaaain. No one wants to deal with that! But we also shouldn't expose 12x native in the GUI, since the majority of our user base will not be helped by it at all and may be hurt by it (a number of users set it to maximum for no reason). This is a good middle ground: convenience for the bleeding edge, but also no effect on the majority of users.

jordan-woyak:
Like was mentioned in IRC, I think adding a "Custom" choice to the drop-down which exposes a numeric entry is a better solution.

stenzek:
I like the idea of a custom box as well. But I'd rather make that a separate change as it's much more invasive, and we'd need to figure out the best way to warn users that high internal resolutions may not function on their hardware. To do this, we could query the GPU for the maximum texture size, but that depends on having the device created to query features, and in the case of the GL backend refactoring initialization to do so.

PR #8341, opened 2019-08-30
ConsoleListenerWin: Properly output console logging in UTF-16

CookiePLMonster:
Minor fix, Windows builds used `OutputDebugStringA` for Console Logger which doesn't handle UTF-8 correctly (A functions assume a local (ACP) codepage). This change explicitly converts an input UTF-8 string to UTF-16 beforehand and passes it to `OutputDebugStringW' instead.

This will impact any logging with UTF-8 in it - the most possible scenario is user paths.

Before:
image

After:
image

### DIFF
--- a/Source/Core/Common/Logging/ConsoleListenerWin.cpp
+++ b/Source/Core/Common/Logging/ConsoleListenerWin.cpp
@@ -5,6 +5,7 @@
 #include <windows.h>
 
 #include "Common/Logging/ConsoleListener.h"
+#include "Common/StringUtil.h"
 
 ConsoleListener::ConsoleListener()
 {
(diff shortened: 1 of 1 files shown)


cremno:
On the other hand Win10 and debugger support are required to support Unicode `OutputDebugStringW`. While the name suggests otherwise the function was essentially a non-Unicode `OutputDebugStringA` call before.

I guess broken output on older Windows versions might be preferable than broken output on all Windows versions.

CookiePLMonster:
> On the other hand Win10 and debugger support are required to support Unicode `OutputDebugStringW`. While the name suggests otherwise the function was essentially a non-Unicode `OutputDebugStringA` call before.

I see this now, yeah. However, since the function itself exists since Windows XP it's essentially a non-issue IMO. If output is going to be broken - tough luck.

PR #8342, opened 2019-09-01
DolphinQt: GeckoCodeWidget remove double space

dreamsyntax:
One line change.
Removes the double space in the read-only GeckoCodeWidget view.
```
Before:
XXXXXXXX  XXXXXXXX

After:
XXXXXXXX XXXXXXXX
```

The before is annoying for those trying to share GeckoCodes.

### DIFF
--- a/Source/Core/DolphinQt/Config/GeckoCodeWidget.cpp
+++ b/Source/Core/DolphinQt/Config/GeckoCodeWidget.cpp
@@ -165,9 +165,11 @@ void GeckoCodeWidget::OnSelectionChanged()
   m_code_view->clear();
 
   for (const auto& c : code.codes)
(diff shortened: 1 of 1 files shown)


Source/Core/DolphinQt/Config/GeckoCodeWidget.cpp
```diff
@@ -165,7 +165,7 @@ void GeckoCodeWidget::OnSelectionChanged()
   m_code_view->clear();
 
   for (const auto& c : code.codes)
```
BhaaLseN: I realize this is pre-existing code, but would you mind adding curlies to this; since the contents of the `for` loop spans over multiple lines?

dreamsyntax: done

PR #8343, opened 2019-09-01
DolphinNoGUI: Add a FBDev platform

stenzek:
Also adds fbdev support for GLInterface, and is sufficient for running Dolphin on an ODROID-N2 with the ARM blob.

I still want to rip out our GLInterface at some point and replace it with some cross-platform library, but the actual fbdev stuff is pretty self contained and only ~100 lines.


JMC47:
Is this ready to merge?

PR #7740, opened 2019-01-26
DSP: ARAM cleanup experiments

booto:
Some experiments trying to clean up ARAM-related code that seems strange.

PR #8344, opened 2019-09-02
Revert "DSP: ARAM cleanup experiments"

JMC47:
Reverts dolphin-emu/dolphin#7740

Seems to be breaking Wii games.  Was missed due to the fact the Wii doesn't have ARAM so they were not tested as thoroughly.


BParks21:
@JMC47 Thank you all is well now lol :)

PR #8345, opened 2019-09-03
D3D12: D3D12-on-Windows-7 support

stenzek:
This PR integrates the newly-released 12on7 library for the D3D12 runtime on Windows 7.

Initial performance testing seems to be about the same for me, but I'm also using an NVIDIA GPU.

If you want to test this, you'll need to copy the files from `Externals/D3D12/bin/x64/*` to `Binaries/x64`.

The binaries themselves present a potential issue with the GPL license - are we allowed to distribute them as an excluded "system library"? The DLLs themselves are loaded dynamically, so worst case we can just ask users to download it themselves and add it.

Process for testing:

1. Download the nuget package from here: https://www.nuget.org/packages/Microsoft.Direct3D.D3D12On7
2. Open the nuget package with your favorite archiving software (7-zip works, or rename it to a .zip).
3. Extract `bin\x64\dxilconv7.dll` to the directory with `Dolphin.exe`
4. Create a directory named `12on7` in your Dolphin directory.
5. Extract `bin\x64\d3d12.dll` to that new `12on7` directory.
6. Run Dolphin, select Direct3D 12 in graphics configuration, have fun.


fantesykikachu:
Unfortunately not only is distributing the binaries a violation of the GPL so is distributing the headers, but then again we already include the xaudio 2.7 headers which is also a violation, and while you are dynamically linking against the dll you are still statically linking the headers which are not GPL compatible and are thus also a violation.

Triang3l:
If something (GPL in this case) creates a lot more problems (largely irrelevant to the project itself and its goals even) than it solves, maybe it'd be helpful to throw such trash away in the future?

mbc07:
Ha ha, that's just not how it works, and good luck trying to relicense a project as big as Dolphin. It has been done once in the past ([going from GPLv2 to GPLv2+](https://dolphin-emu.org/blog/2015/05/25/relicensing-dolphin/)) and was a monumental task to pull off...

JosJuice:
The GPL does solve problems for us, and like mbc07 said, we wouldn't be able to switch away from it at this point even if we wanted to.

Helios747:
If this gives Windows 7 logic ops it might be worth it. But otherwise I don't see the point if performance is largely the same. Especially if this is just some wrapper down to D3D11.

in regards to the GPL problem, if we decide that the benefits are there and worth merging, yeah I don't see why we can't have UI code that instructs users to go and grab the binary.

stenzek:
@Helios747 the header file may still pose a challenge. The presentation interface is different for 12on7 so we can't use it without the header.

Triang3l:
@Helios747 It exposes everything, even raytracing. Though need to be more careful about heap/committed resource allocations with it (same as with Vulkan's 4096 memory object limit, I think) since it apparently has to write all the existing allocations to every command list.

CookiePLMonster:
`__has_include` is a thing in C++17, and all API usage can be killed off with either `if constexpr` and SFINAE. This would make it possible to ask users to paste 12on7 headers to a specific directory in Dolphin's source directory instead of shipping it *and* it would make it compileable both with and without them. Does that solve the issue?

alessiot89:
You can still wrap everything in a DLL, public it in a different repo with a different license (e.g.: MIT), detect if the user is runnning under Windows 7 and make an on-demand download of the DLL and then link it at runtime.

shuffle2:
Do people really want win7 support? The _extended_ EOL is less than a month away (https://support.microsoft.com/en-us/help/13853/windows-lifecycle-fact-sheet). Of course that timeline doesn't really impact dolphin, however I'm honestly confused by people running even out-of-date win10 these days, much less win7. Just figured I'd poke about it :)

Helios747:
At this point I don't see any reason to merge this because yeah, Win 7 is extremely close to EOL. Don't see the point of merging in non trivial amounts of code to make dead operating systems work. I'm sorry to the users who have various valid reasons for avoiding later versions of Windows.

stenzek:
Agreed. The PR is still here for anyone who wants to build it for themselves.

mirh:
Why couldn't you just add the nuget package to the vcxproj? (then move bins around on packaging)
`<PackageReference Include="Microsoft.Direct3D.D3D12On7" Version="1.1.0" />`

JosJuice:
That doesn't do anything to solve the licensing problems as far as I understand.

mirh:
[This](https://www.nuget.org/packages/Microsoft.Direct3D.D3D12On7/1.1.0/License) is the license. 
As also explained in their [FAQ](https://microsoft.github.io/DirectX-Specs/d3d/D3D12onWin7.html#package-and-release-windows-7-version-of-d3d12-binaries-as-part-of-your-game), you are allowed (or better yet, actually *required*) to ship their binaries. 
I don't know in which shape or form that would contradict the GPL. Each object file is its own world, and unless you were explicitly pretending their IP to be yours I don't see the problem (unless the issue was with linking closed libraries? which.. uh, happens all the time?).

If instead the trouble was just with redistributing the headers in the repo (not dissimilarly from [XAudio2](https://github.com/dolphin-emu/dolphin/pull/8361#issuecomment-581027828)), as I was saying you can just let nuget handle that for you?
When you fire VS it automatically downloads, load and add the thing to the project and profit.

stenzek:
I am not a lawyer, but I believe it's more of an issue with linking GPL code (dolphin) to non free code (D3D12) and this one doesn't fall under the 'system library exception'

But that said, win7 is no longer supported by MS so I see no reason for us to support it either. Not that we're going to intentionally break it, but it's silly to add new functionality at this point IMO.

mirh:
Well, to be fair it seems to me even more sillier to trash this 95% already done PR. But it's your code, so have it. 

> and this one doesn't fall under the 'system library exception'

I don't really think "system library" only covers "system critical" stuff. Otherwise linking GDI to have a GUI/opengl already could be considered an extra. 
Then of course D3D12 is actually very much an optional on top of that (because even *functionally* the minimum API the OS requires you for most features is D3D11).. But wouldn't that also be true for normal W10 operations?

stenzek:
It took me longer to install/update win7 than to write the patch..

I don't see much chance of us merging this. Win7 holdouts will eventually migrate, just like people on XP did. But feel free to cherry-pick and include in your own build if you want the functionality.

PR #8236, opened 2019-07-09
SI: Minor bugfixes

booto:
- VI was using "lines" count from SI without converting to half lines.
- origin calibration structure for controllers is 10 bytes, not 12

PR #8347, opened 2019-09-04
Align frame advance and movies to full field boundaries

hosaka-corp:
Currently, frame-stepping and `Movie::FrameUpdate()` are dealt with at the beginning of each "active video field" (see `Core::Callback_VideoCopiedToXFB()`). 

This is [conceptually] inconsistent with accurate SI polling emulation (#8236 and other WIP branches) because SI polls may occur in the space between active video fields. Here's some log output from a game that does SI polling four times in one field:

```
// Current behavior
Core/HW/VideoInterface.cpp:809 I[SI]: new video interrupt, half_line=0000
Core/HW/VideoInterface.cpp:741 I[SI]: new field, half_line=0000 (start of full field)
Core/HW/VideoInterface.cpp:749 I[SI]: new poll, half_line=000f
VideoCommon/VideoBackendBase.cpp:73 I[SI]: Video_BeginField (start of active field)
!!!!!! Core/Core.cpp:843 I[SI]: <Movie::FrameUpdate() happens here>
VideoCommon/RenderBase.cpp:1336 I[SI]: Renderer::Swap() ended
Core/HW/VideoInterface.cpp:749 I[SI]: new poll, half_line=00ab
Core/HW/VideoInterface.cpp:749 I[SI]: new poll, half_line=0147
Core/HW/VideoInterface.cpp:749 I[SI]: new poll, half_line=01e3

// New behavior
Core/HW/VideoInterface.cpp:809 I[SI]: new video interrupt, half_line=0000
Core/HW/VideoInterface.cpp:741 I[SI]: new field, half_line=0000 (start of full field)
!!!!!! Core/Movie.cpp:208 I[SI]: <new Movie::FrameUpdate() happens here>
Core/HW/VideoInterface.cpp:749 I[SI]: new poll, half_line=000f
VideoCommon/VideoBackendBase.cpp:73 I[SI]: Video_BeginField called (start of active field)
Core/Core.cpp:843 I[SI]: Callback_VideoCopiedToXFB
VideoCommon/RenderBase.cpp:1336 I[SI]: Renderer::Swap() ended
Core/HW/VideoInterface.cpp:749 I[SI]: new poll, half_line=00ab
Core/HW/VideoInterface.cpp:749 I[SI]: new poll, half_line=0147
Core/HW/VideoInterface.cpp:749 I[SI]: new poll, half_line=01e3
```

This ensures that "a movie frame" (and the check for frame-stepping) starts at the same time as a "full field."

### DIFF
--- a/Source/Core/Core/Core.cpp
+++ b/Source/Core/Core/Core.cpp
@@ -839,9 +839,12 @@ void Callback_VideoCopiedToXFB(bool video_update)
 {
   if (video_update)
     s_drawn_frame++;
+}
 
+// Called at field boundaries in `VideoInterface::Update()`
+void FrameUpdate()
+{
   Movie::FrameUpdate();
-
   if (s_frame_step)
   {
     s_frame_step = false;
--- a/Source/Core/Core/HW/VideoInterface.cpp
+++ b/Source/Core/Core/HW/VideoInterface.cpp
@@ -733,6 +733,12 @@ static void EndField()
 // Run when: When a frame is scanned (progressive/interlace)
 void Update(u64 ticks)
 {
+  // If this half-line is at a field boundary, potentially deal with frame-stepping
+  // and/or update movie state before dealing with anything else
+
+  if (s_half_line_count == 0 || s_half_line_count == GetHalfLinesPerEvenField())
+    Core::FrameUpdate();
+
   // If an SI poll is scheduled to happen on this half-line, do it!
 
   if (s_half_line_of_next_si_poll == s_half_line_count)
--- a/Source/Core/Core/Core.h
+++ b/Source/Core/Core/Core.h
@@ -26,6 +26,7 @@ bool GetIsThrottlerTempDisabled();
 void SetIsThrottlerTempDisabled(bool disable);
 
 void Callback_VideoCopiedToXFB(bool video_update);
+void FrameUpdate();
 
 enum class State
 {


Source/Core/Core/Core.cpp
```diff
@@ -839,16 +839,21 @@ void Callback_VideoCopiedToXFB(bool video_update)
 {
   if (video_update)
     s_drawn_frame++;
+}
 
+// Called at field boundaries in `VideoInterface::Update()`
+bool FrameUpdate(void)
```
JosJuice: We typically don't put `void` in the parameter list.

Any particular reason why this returns a value?

Source/Core/Core/HW/VideoInterface.cpp
```diff
@@ -733,6 +733,12 @@ static void EndField()
 // Run when: When a frame is scanned (progressive/interlace)
 void Update(u64 ticks)
 {
+  // If this half-line at a field boundary, potentially deal with frame-stepping
```
JosJuice: I think there's a word missing here.

Source/Core/Core/Core.cpp
hosaka-corp: Nope! Leftover from checking something.

CookiePLMonster:
@dolphin-emu-bot rebuild

RisingFog:
This seems like a much better way of handling it according to your test results.

Source/Core/Core/HW/VideoInterface.cpp
```diff
@@ -733,6 +733,12 @@ static void EndField()
 // Run when: When a frame is scanned (progressive/interlace)
 void Update(u64 ticks)
 {
+  // If this half-line is at a field boundary, potentially deal with frame-stepping
+  // and/or update movie state before dealing with anything else
+
+  if ((s_half_line_count == 0) || (s_half_line_count == GetHalfLinesPerEvenField()))
```
JosJuice: Also, the inner parentheses here can be removed.

JosJuice:
@dolphin-emu-bot rebuild

vadosnaprimer:
This causes Dolphin to fail rendering (and dumping) the first frame, which leads to heavy AV desync in Resident Evil 4 (worth 300 frames).

phire:
Isn't this going to break the TAS workflow, which relies on Bypass XFB and frame stepping to stop emulation right as the frame finishes rendering (skipping any XFB latency)

JosJuice:
I was under the impression that the purpose of this was to improve TASing... Is there any particular reason why you would want to use immediately present XFB (assuming that's what you mean by bypass XFB) when TASing?

phire:
When TASing, you want to step, then apply input then step again and see your input applied. 

Anything that adds extra latency complicates this workflow. The more latency, the more re-records are required to rewind.

I'm not actually sure if the TASing community currently uses Immediately present XFB (I'm not a TASing expert, but I thought we kept that feature around partly for TAS reasons), and this PR probably improves things when Immediately present XFB is off. 

Maybe we should be switching between two different behaviours depending on XFB mode.

vadosnaprimer:
Is XFB mode related to improvements this PR does, or to the dumping bug?

JosJuice:
The XFB thing that phire and I are talking about isn't related to the dumping bug (as far as I know). It's about the process of recording the inputs for a TAS.

PR #8348, opened 2019-09-04
Disable full MMU by default and add it to global config GUI

JosJuice:
Requested by JMC.


Helios747:
+1, be nice to start claiming back some of the performance we've been chipping away at. Known MMU titles can have it on via gameini, which is gameini's job.

Users that want game glitches to work has this option.

JMC47:
If this is approved I'll close my other one disabling it in specific games.

PR #8349, opened 2019-09-04
Fix 2 DTK bugs

Pokechu22:
Bug 1: subcommand 1 of 0xE1 does not actually reset the current position; it just stops playback.  0xE2 returns the same value it would have before stopping playback (other than reporting that the stream is stopped).

Bug 2: subcommand 3 of 0xE2 was incorrectly dividing the length by 4; while this makes sense for positions (which are multiplied by 4 earlier to get a byte offset), it is not correct for lengths.

### DIFF
--- a/Source/Core/Core/HW/DVD/DVDInterface.cpp
+++ b/Source/Core/Core/HW/DVD/DVDInterface.cpp
@@ -991,11 +991,6 @@ void ExecuteCommand(u32 command_0, u32 command_1, u32 command_2, u32 output_addr
     {
       s_stop_at_track_end = false;
       s_stream = false;
-      s_audio_position = 0;
-      s_next_start = 0;
-      s_next_length = 0;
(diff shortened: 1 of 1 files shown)


Pokechu22:
[Hardware test](https://gist.github.com/Pokechu22/aa91cc1c88ea666476fee9c19208f22b), which should at least work on the Wii.

lioncash:
It'd probably be beneficial if the hardware test was integrated with our [hwtests](https://github.com/dolphin-emu/hwtests) in some capacity

Pokechu22:
I'll probably PR the hwtests repository later with a more complete set of automated tests for my bigger DI pull request; for now the test is manual so it probably doesn't need to be added.

JosJuice:
@dolphin-emu-bot rebuild

GlydetheSurfer:
where do i post issues with a game?

PR #5726, opened 2017-06-29
Implement minimal emulation of TMEM caching

mimimi085181:
This is a remake of https://github.com/dolphin-emu/dolphin/pull/3749

Full credit goes to phire.

Old message:
"If none of the texture registers have changed and TMEM hasn't been invalidated or changed in other ways, we can blindly reuse the old texture cache entries without rehashing.

Not only does this fix the bloom effect in Spyro: A Hero's Tail (The game abused texture cache) but it will also provide speedups for other games which use the same texture over multiple draw calls, especially when safe texture cache is in use."

Changed the pr per phire's instructions to only return the current texture(s), if none of the texture registers were changed. If any texture register was changed, fall back to the default hashing and rebuilding textures from memory.

PR #5718, opened 2017-06-28
Update to partial texture updates

mimimi085181:
Partial updates had a bug, that updated textures could be loaded by hash, ignoring the address(textures_by_hash). This could result in unexpected results, if a game had 2 textures where the base is identical, but they are stored in different memory locations and differ because of parts that are updated via partial updates. Since efb copies always write the same dummy data, both textures would have the same hash, and Dolphin would load one of them for both textures.

This pr also changes the logic about which textures are removed from the cache when they overlap new efb copies. Dolphin now keeps all textures that are partly overwritten and have the same stride as the efb copy. This is supposed to get efb2tex to the same texture as efb2ram, by applying the related efb copies as updates after each other, in the order of their creation.

The main purpose of this pr is to improve partial updates for hybrid xfb. Other than that, it does "fix" Spyro: A Hero's Tail, so efb2tex looks the same as efb2ram now. "fix", because the amount of bloom the game uses is high and it gets increased with higher IR, resulting at 3xIR in something i'd call hyperbloom... Anyways, this pr together with https://github.com/dolphin-emu/dolphin/pull/5726 gets efb2tex to the same output as efb2ram, while it's actually correct emulation, as far as i can see.

PR #8350, opened 2019-09-06
Extend Minimal TMEM cache implementation.

phire:
Now works with games that deliberately avoid invalidating TMEM because they know textures are too large to fit:

 * Sonic Riders
 * Metal Arms: Glitch in the System
 * Godzilla: Destroy All Monsters Melee
 * NHL Slapshot
 * Tak and the Power of Juju
 * Night at the Museum: Battle of the Smithsonian
 * 428: Fūsa Sareta Shibuya de

This is just a first pass, while it should work, I'm not happy with how the code is structured.
Mostly commiting for fifoci run and potential feedback.


Source/Core/VideoCommon/BPStructs.cpp
```diff
@@ -669,14 +669,16 @@ static void BPWritten(const BPCmd& bp)
...
   case BPMEM_TX_SETIMAGE3:
   case BPMEM_TX_SETIMAGE3_4:
     TextureCacheBase::InvalidateAllBindPoints();
     return;
+  case BPMEM_TX_SETIMAGE1:
+  case BPMEM_TX_SETIMAGE1_4:
+  case BPMEM_TX_SETIMAGE2:
+  case BPMEM_TX_SETIMAGE2_4:
+    TmemBase::ConfigurationChanged(bp.address, bp.newvalue);
+
```
Warepire: Missing return here is intentional?

phire: yeah...... no.

phire:
Rebased for clean fifo-ci run.

Source/Core/VideoCommon/TmemBase.cpp
```diff
@@ -0,0 +1,203 @@
...
+
+#include "VideoCommon/BPMemory.h"
+#include "VideoCommon/TmemBase.h"
+
+namespace TmemBase
+{
+// On TMEM configuration changed:
+// 1. invalidate stage.
+
+struct texture_unit_state
```
lioncash:
```suggestion
struct TextureUnitState
```

Source/Core/VideoCommon/TmemBase.cpp
```diff
@@ -0,0 +1,203 @@
...
+{
+  enum State
```
lioncash:
```suggestion
  enum class State
```

Source/Core/VideoCommon/TmemBase.cpp
```diff
@@ -0,0 +1,203 @@
...
+  State state;
+
+  int even_base;
+  int even_size;
+  int odd_base;
+  int odd_size;
+};
+static std::array<texture_unit_state, 8> s_unit;
+
+void ConfigurationChanged(int bp_addr, int config)
```
lioncash: bp_addr should likely be a `u32`, given its an address.

Source/Core/VideoCommon/TmemBase.cpp
```diff
@@ -0,0 +1,203 @@
...
+  {
+    unit_state.odd.hex = config;
+  }
+  unit_state.state = texture_unit_state::INVALID;
+}
+
+// On invalidate cache
+// 1. invalidate all
+
+void Invalidate(u32 param)
```
lioncash:
```suggestion
void Invalidate([[maybe_unused]] u32 param)
```

Source/Core/VideoCommon/TmemBase.cpp
```diff
@@ -0,0 +1,203 @@
...
+    unit.state = texture_unit_state::INVALID;
+  }
+}
+
+// On bind:
+// 1. store the memory region(s) bound
+// 2. if texture size is small enough to fit in region AND there is no overlap, mark as cached.
+// 3. If there is cached overlap, invalidate.
+
+static inline u32 calculate_unit_size(TexImage1 config)
```
lioncash: `inline` is superfluous here given static is already specified. This should also be named `CalculateUnitSize`

Source/Core/VideoCommon/TmemBase.cpp
```diff
@@ -0,0 +1,203 @@
...
+  }
+
+  // However, the registers allow a much larger amount of configurablity.
+  // Maybe other sizes are broken?
+  // Until hardware tests are done, this is a guess at the size algorithm
+
+  return 512 * (1 << config.cache_width) * (1 << config.cache_height);
+}
+
+static inline u32 calculate_unit_size(TexImage2 config)
```
lioncash: Ditto

Source/Core/VideoCommon/TmemBase.cpp
```diff
@@ -0,0 +1,203 @@
...
+static bool overlap(int a, int a_size, int b, int b_size)
+{
+  return (a <= b && (a + a_size) >= b) || (b <= a && (b + b_size) >= a);
+}
+
+static bool process_overlaps_single(u8 unit_id, int base, int size)
+{
+  bool has_overlap = false;
+  // We are just going to do a linear search
+  for (int i = 0; i < s_unit.size(); i++)
```
lioncash: `i` should be a `size_t`, otherwise this introduces a signed/unsigned mismatch.

Source/Core/VideoCommon/TmemBase.h
```diff
@@ -0,0 +1,16 @@
...
+#pragma once
+
+#include "Common/CommonTypes.h"
+
+
+namespace TmemBase
+{
+void Invalidate(u32 param);
+void ConfigurationChanged(int bp_addr, int config);
+void Bind(u8 unit, int num_blocks_width, int num_blocks_height, bool is_mipmapped, bool is_32_bit);
```
lioncash: These `u8`s should probably be `u32`. u32 is slightly more flexible when it comes to interface changes outside of this source file, given it's very easy to implicitly truncate to u8, causing warnings.

Source/Core/VideoCommon/TmemBase.h
```diff
@@ -0,0 +1,16 @@
+// Copyright 2010 Dolphin Emulator Project
```
lioncash: 2019

Source/Core/VideoCommon/TmemBase.cpp
```diff
@@ -0,0 +1,203 @@
...
+  TexImage1 even;
+  TexImage2 odd;
+  State state;
+
+  int even_base;
+  int even_size;
+  int odd_base;
+  int odd_size;
+};
+static std::array<texture_unit_state, 8> s_unit;
```
lioncash: Is there really no other place to place this (like in the actual cache)? This introduces more file-scope/global mutable state.

Source/Core/VideoCommon/CMakeLists.txt
```diff
@@ -88,6 +88,8 @@ add_library(videocommon
   TextureDecoder.h
   TextureDecoder_Common.cpp
   TextureDecoder_Util.h
+  TmemBase.cpp
+	TmemBase.h
```
BhaaLseN: This line needs more space ;)

Source/Core/VideoCommon/BPStructs.cpp
```diff
@@ -669,14 +669,17 @@ static void BPWritten(const BPCmd& bp)
...
+    TextureCacheBase::InvalidateAllBindPoints();
+    return;
   case BPMEM_TX_SETIMAGE1:
   case BPMEM_TX_SETIMAGE1_4:
   case BPMEM_TX_SETIMAGE2:
   case BPMEM_TX_SETIMAGE2_4:
-  case BPMEM_TX_SETIMAGE3:
-  case BPMEM_TX_SETIMAGE3_4:
-    TextureCacheBase::InvalidateAllBindPoints();
+    TmemBase::ConfigurationChanged(bp.address, bp.newvalue);
```
BhaaLseN: It feels odd that 0 and 3 do something differently to 1 and 2. Maybe it would be a good idea to leave a few comments on _why_ they're different?

Source/Core/VideoCommon/TextureCacheBase.cpp
```diff
@@ -1155,9 +1156,23 @@ class ArbitraryMipmapDetector
 TextureCacheBase::TCacheEntry* TextureCacheBase::Load(const u32 stage)
 {
   // if this stage was not invalidated by changes to texture registers, keep the current texture
+  // FIXME: if and only if it's small enough to fit in TMEM, otherwise we need to hash.
   if (IsValidBindPoint(stage) && bound_textures[stage])
   {
-    return bound_textures[stage];
+    TCacheEntry* entry = bound_textures[stage];
+    // If the TMEM configuration is such that this texture is more or less guarneteed to still
```
BhaaLseN:
```suggestion
    // If the TMEM configuration is such that this texture is more or less guaranteed to still
```

Source/Core/VideoCommon/TmemBase.cpp
```diff
@@ -0,0 +1,203 @@
...
+namespace TmemBase
+{
+// On TMEM configuration changed:
+// 1. invalidate stage.
+
+struct texture_unit_state
+{
+  enum State
+  {
+    INVALID, // Configuration has changed
```
BhaaLseN: Thoughts on putting the comment in the line before rather than at the end of the line? At least Visual Studio shows those in Intellisense if they're in the line before.

Source/Core/VideoCommon/TmemBase.cpp
```diff
@@ -0,0 +1,203 @@
...
+  int even_base;
+  int even_size;
+  int odd_base;
+  int odd_size;
+};
+static std::array<texture_unit_state, 8> s_unit;
+
+void ConfigurationChanged(int bp_addr, int config)
+{
+  // Extract bits encoding texture unit
```
BhaaLseN: Would you mind extending that comment with more info on what the bits in `bp_addr` mean (or, if it is easier: _why_ the shifting and masking is done)?

Source/Core/VideoCommon/TmemBase.cpp
```diff
@@ -0,0 +1,203 @@
...
+  }
+  unit_state.state = texture_unit_state::INVALID;
+}
+
+// On invalidate cache
+// 1. invalidate all
+
+void Invalidate(u32 param)
+{
+  // The excat arguments of Invalidate commands is currently unknown.
```
BhaaLseN:
```suggestion
  // The exact arguments of Invalidate commands is currently unknown.
```

Source/Core/VideoCommon/TmemBase.cpp
```diff
@@ -0,0 +1,203 @@
...
+  casted_config.hex = config.hex;
+  return calculate_unit_size(casted_config);
+}
+
+static bool overlap(int a, int a_size, int b, int b_size)
+{
+  return (a <= b && (a + a_size) >= b) || (b <= a && (b + b_size) >= a);
+}
+
+static bool process_overlaps_single(u8 unit_id, int base, int size)
```
BhaaLseN: Would there be any benefit in having a `struct` parameter instead of separate `base`/`size` and reuse the same as `TextureUnitState.even`/`TextureUnitState.odd`? That way you could move the `overlap` method into that struct; potentially as operator if it makes sense; plus get the added benefit of less ambiguity when faced with 2 or 4 `int` parameters (like, is it `base` or `size` first; or are they next to each other in the `overlap` case?)
It could also be the home for `CalculateUnitSize` (and perhaps others) if it makes sense in there.

phire: I'm really unhappy with this code and the way implementation details (finding overlaps) is mixed with the higher level concepts.

What I want to do is implement a generic [ interval tree](https://en.m.wikipedia.org/wiki/Interval_tree) implementation in common, which appears to be the correct data structure for storing and checking for overlaps.

Then I can just implement tmem in terms of the interval tree.

Source/Core/VideoCommon/TmemBase.cpp
phire: I feel really uneasy about how complicated texture cache is getting and just how complicated it is to understand.

This part is still WIP, this state will need to be stored somewhere, and it will need initialized to zero, and something will need to be done on DoState (either we invalidate the whole thing, or for correctness we save anything that's tagged as "cached")

phire:
FIFOci update:

Good news:

 * [Spyro is showing the correct amount of bloom again](https://fifoci.dolphin-emu.org/compare/4965047-4964251/)
 * [NHL Slap regression is fixed](https://fifoci.dolphin-emu.org/compare/4964185-4953585/) (I'm assuming the players being black is not part of the regression?
 
Average news:

* [Fortune Street is "different" again](https://fifoci.dolphin-emu.org/compare/4964208-4953631/). I don't think we ever worked out if this was correct before/after tmem.

Bad news:

 * [Sonic riders gravity boost effect is broken](https://fifoci.dolphin-emu.org/compare/4964250-4953630/).
 * [Underwater is broken in Spyro](https://fifoci.dolphin-emu.org/compare/4964271-4953579/)

JMC47:
Are we sure those last two are really broken?  Those are ancient fifologs.  And Sonic's gravity boost was already broken.

JMC47:
The two games mentioned as broken in those fifologs are *only* broken in EFB2Tex.  In EFB2RAM, they render more correctly than they did before.  It'd be nice for them to not black out or act weird in EFB2Tex, but that's a far, far, far less severe issue than we have now.

JMC47:
Blacked out player colors are because of the fifolog.

image

JMC47:
Any update on what's going on here?

tony971:
Any hope of progress on this? Quite a few games are affected by it.

JMC47:
Even though this has a regression in EFB2TEX, it would much, much, much better than the current situation if rebased/reviewed/merged.

JMC47:
Can this be rebased?  There's an issue report that I think this may affect but the buildbot builds are gone.

JMC47:
So, as a bit of a refresher, here's the current status of this Pull Request.

The EFB2RAM path of this Pull Request seems to work perfectly.  It even has handled new, previously unknown cases of tmem issues that weren't known at the time of its creation.  It has been verified to work in the rebased version of this PR.

The *main* thing this Pull Request fixes is a bunch of games with video playback issues.  It appears that whatever video player that Sega used has a lot of issues with the tmem implementation in master.  Both paths of this PR seem to work fine for this.  Verified games that are fixed include:

Sonic Riders
Metal Arms: Glitch in the System
Tak and the Power of Juju
Sonic Riders: Zero Gravity
All the games here - https://bugs.dolphin-emu.org/issues/11164
Probably more that haven't been reported.

However, the EFB2Tex version of this PR does have some issues that aren't present on the EFB2RAM version.

**Spyro: A Hero's Tail**

https://fifo.ci/compare/4964271-4953579/

In **EFB2TEX ONLY**, the water issues return.  However, using EFB2RAM you can get the correct effect.  This appears to be an invalidation thing.

**Sonic Riders: Zero Gravity**

Fourth Player's viewport appears to flicker in EFB2Tex.  EFB2RAM in this Pull Request works fine, however.  It is currently broken in master too so this isn't a big issue.  The fact that EFB2RAM fixes it is actually a benefit.

That's actually all that's known broken in the PR, two invalidation issues.

JMC47:
I know this is marked as draft and do not merge.  But at this point what currently exists here is far superior to the behavior in Master.

While the invalidation issues kind of suck, I'd rather have things 95% working so we can potentially work toward a release and then worry about the EFB2Tex path later.

Source/Core/VideoCommon/TmemBase.cpp
degasus: Or better, use a struct to parse this BP command and pass on this struct instead of the raw u32.

autofire372:
This needs a rebase/rewrite; the Windows build system changed in PR #9092.

tony971:
Just doing my annual bump

Source/Core/VideoCommon/TmemBase.cpp
```diff
@@ -0,0 +1,203 @@
...
+  }
+  else
+  {
+    unit_state.odd.hex = config;
+  }
+  unit_state.state = texture_unit_state::INVALID;
+}
+
+// On invalidate cache
+// 1. invalidate all
```
phire, line 55: Incomplete comment?

Source/Core/VideoCommon/TmemBase.cpp
```diff
@@ -0,0 +1,203 @@
...
+  // It appears to contain the TMEM address and a size.
+
+  // For simplicity, we will just invalidate everything
+  for (auto& unit : s_unit)
+  {
+    unit.state = texture_unit_state::INVALID;
+  }
+}
+
+// On bind:
```
phire, line 69: This comment is not in the right place

Probably belongs with `Bind()` which is 100 lines further down.

Source/Core/VideoCommon/TmemBase.cpp
phire: Sure, why not.

JMC47:
I went through the major cases again and they all behave the same as we remember.

Spyro: A Hero's Tail - Needs EFB2RAM but is fixed at least.
Sonic Riders and other FMV games - Fixed
NHL Slapshot - Fixed
Sonic Riders - Zero Grav - Fixed, but needs EFB2RAM for some effects.

BlueSwordM:
So, has there been any progress in regards to this pull request?

I've merged it into my own semi-custom build for Sonic Riders and Sonic Riders ZG, and it seems to work fine(which are the only games that I own which have the TMEM issues).

The only problem is that I don't have a lot of games to do regression testing, and as such, I'm not sure if it's ready to pull in or not.

Thank you for all the work.

phire:
Status report:

This PR was rapidly implemented as a prototype to prove the idea.

It is not production ready code. I need to take some time to rewrite it. I haven't gotten around to doing this yet.

JMC47:
Note that this is just cleanup.  Titles fixed by this pull request are still fixed, but there's no change in behavior otherwise.

phire:
Investigations have shown that the efb2tex regressions appear to come from https://github.com/dolphin-emu/dolphin/pull/5718/commits/cdbd9863b4a05db046a377c57485e92ea62ead84

Which was actually merged slightly after the original minimal TMEM implementation: #5726

It's just that TMEM has kind of hidden the bug.

JMC47:
I think my main concern is that we don't break savestates/TASing with this.

Source/Core/VideoCommon/TmemBase.cpp
phire: This change has been implemented in #10096

BlueSwordM:
Everything seems to work fine on my end with the new patch.

All games that rely on proper TMem emulation for cutscenes work.

Sonic Riders Zero Gravity works without the texture cache bug, which means `EFB2Texture` works as well, restoring full performance to the game.

JMC47:
It is an EFB related issue and a texture cache related issue.  But neither appear to be a bug with TMEM, just something masked by TMEM.

JMC47:
In the words of great Aristotle, "One swallow does not make a summer, neither does one fine day; similarly one day or brief time of happiness does not make a person entirely happy."  Translated to modern times, I'm pretty sure it would read, "Yolo."  

...Then again, I'm not a philosopher.

PR #8304, opened 2019-08-08
Qt/GCMemcardManager: Fix icon animations displaying incorrectly.

AdmiralCurtiss:
The previous code had some odd assumptions that caused it to only show, at max, the first three frames of each animation. It also didn't handle the animation speed or the ping-pong frame order correctly. This should all be fixed with this.

The performance of this I mentioned in IRC is marginally improved by this due to the timer running less often, but it's still pretty poor; further investigation into that is necessary.

PR #8302, opened 2019-08-07
GCMemcard: Read comments, banners, and icons via logical data offsets instead of physical ones.

AdmiralCurtiss:
Previously the code assumed that the data for the comments, banners, and icons was contiguous from the first block of the data's physical offset in the memory card, rather than referring to the logical position within the referenced save file, which is not only incorrect for the format but could also lead to out-of-bounds memory accesses. This should now be rectified.

Fixes:
https://bugs.dolphin-emu.org/issues/10600
https://bugs.dolphin-emu.org/issues/11820

Please scrutinize this, I'm not entirely sure if this is all correct. It seems to give correct results for all memory cards I have access to, but considering how long this old approach seemingly worked who knows.

~~The Icon/Banner code could probably use some additional rewriting on top of this to be honest, it's a mess. It also seems to disagree with itself on how the bnr_format should actually be interpreted...~~ I've done this now.

PR #8351, opened 2019-09-06
GCMemcard : copy/decode icon animation frames correctly

DacoTaco:
some save file icon's contain blank frames (luigi's mansion & pikmin for example).
when handling these frames there was a bug as there was no 'buffer += 32*32' when the frame was type CI8SHARED + it took the last available frame it found, not the first available frame.
i cleaned up the code and fixed the issue.

with dolphin's master (at this point) this should break the animations of those games (because of the current 3 frame limit bug) but with PR https://github.com/dolphin-emu/dolphin/pull/8304 it'll works fine ( and fixes the issue it has with these saves/icons )


BhaaLseN:
Just a heads up, this will collide with changes in #8302

DacoTaco:
my mistake of not having discussed it thoroughly i guess? ill look into it

DacoTaco:
i think its better to be addressed in the other PR

PR #8355, opened 2019-09-08
QtFifoPlayer: handle FIFO load before window creation

phire:
Fixes a bug where if you loaded a fifo before opening the fifo
player window (which you can do by dragging a .dff onto dolphin's
main window) then the player's widgets wouldn't be initilized
correctly.

Importantly, the object range widgets would be broken.

### DIFF
--- a/Source/Core/Core/FifoPlayer/FifoPlayer.cpp
+++ b/Source/Core/Core/FifoPlayer/FifoPlayer.cpp
@@ -170,7 +170,7 @@ std::unique_ptr<CPUCoreBase> FifoPlayer::GetCPUCore()
 
 void FifoPlayer::SetFileLoadedCallback(CallbackFunc callback)
 {
-  m_FileLoadedCb = callback;
+  m_FileLoadedCb = std::move(callback);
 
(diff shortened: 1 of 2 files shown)


Source/Core/Core/FifoPlayer/FifoPlayer.cpp
```diff
@@ -168,6 +168,17 @@ std::unique_ptr<CPUCoreBase> FifoPlayer::GetCPUCore()
   return std::make_unique<CPUCore>(this);
 }
 
+void FifoPlayer::SetFileLoadedCallback(CallbackFunc callback)
+{
+  m_FileLoadedCb = callback;
```
lioncash: This is pre-existing, so obvious non-blocker, but,

```suggestion
  m_FileLoadedCb = std::move(callback);
```

would be preferable here, since std::function is allowed to allocate to hold any captures that exceed its internal buffer, so this can allow avoiding reallocations.

phire: I'll do the same to both callbacks.

PR #8352, opened 2019-09-07
Support for motion controllers like the DualShock 4

rlnilsen:
### The CemuHook controller input protocol
This was done by implementing the protocol for controller input used by [CemuHook ](https://cemuhook.sshnuke.net/) in a new controller interface class (`CemuHookUDPServer`). The protocol is used for transmitting controller input (including motion input) to Dolphin from specific server applications written to support specific controller types. For example, the Sony Dualshock 4 is supported through the [DS4Windows](https://ryochan7.github.io/ds4windows-site/) application. **This is the only configuration I have tested**. See [this page](https://cemuhook.sshnuke.net/padudpserver.html) on the CemuHook site for more information and instructions on testing motion input before trying to use it with Dolphin. 

### Server configuration
First make sure you have tested your controller's motion input like in the link above. Then run and quit Dolphin once to create the settings file `UDPServer.ini` in the Dolphin config folder. In the settings file change `Enabled = False` to `Enabled = True`. Default values for the other settings (`IPAddress` and `Port`) may need changing depending on the server application used. With DS4Windows you just need to enable the UDP Server in DS4Windows' settings.

### Emulated Wii Remote configuration
2019-09-19 15_04_21-vs2017 (1) - Remote Viewer
In the GUI for configuring an Emulated Wii Remote, a new tab called `Motion Input` was added. The old `Motion Controls` tab was renamed to `Motion Simulation` to differentiate them. In the Device dropdown list select a `UDPServer` device for motion input support. `Point` settings are described below. `Accelerometer` and `Gyro` settings should be left alone.

### Pointer emulation
The Wiimote camera sensor (pointer) is emulated from motion input. There are however some inaccuracies in motion sensors making the pointer slowly drift horizontally. To help this, there are two options in the `Point` box on the `Motion Input` tab. One is `Recenter`, which lets you recenter the pointer on the screen with a button press. The other is `Total Yaw`, which horizontally limits the pointer movement like when a computer mouse pointer hits the screen edge. Try out both options to see which works best in which games. I have playtested very little, but for **Skyward Sword** it seems best to disable `Total Yaw` by setting it to `360` and mapping `Recenter` to `Pad S` just like the ingame recenter function. Then, when you need to recenter, point the _controller_ in the direction of the screen center and tap `Pad S` _twice_ without moving the controller.

### Funny story
[...]

PR #8356, opened 2019-09-09
WIP - ControllerInterface. Combine evdev devices with the same "uniq" name.

jordan-woyak:
Unfortunately newer Linux drivers for DS4 (Playstation 4) controllers split the device into three separate event nodes.

On my system I get:
"Sony Interactive Entertainment Wireless Controller"
"Sony Interactive Entertainment Wireless Controller Touchpad"
"Sony Interactive Entertainment Wireless Controller Motion Sensors"

Apparently this is due to some evdev policy: "A device must not mix regular directional axes and accelerometer axes on the same event node."

Our input system does allow use of multiple devices via "All Devices" or so-called "absolute path" control expressions, but this is less than desirable especially when making use of multiple DS4s.

This PR combines evdev devices that return the same non-null unique ID (via `libevdev_get_uniq`).

Mapping of raw acceleration/gyroscope inputs in PR #8352 is less horrible this way.

I'm not entirely happy with this implementation. It gets pretty ugly, especially the parts that keep the button/axis names in the same deterministic order, but there doesn't seem to be a clean way to do it as evdev enumerates devices in a seemingly random order.

### DIFF
--- a/Source/Core/InputCommon/ControllerInterface/evdev/evdev.cpp
+++ b/Source/Core/InputCommon/ControllerInterface/evdev/evdev.cpp
@@ -4,14 +4,13 @@
 
 #include <algorithm>
 #include <cstring>
-#include <fcntl.h>
-#include <libudev.h>
-#include <map>
 #include <memory>
 #include <string>
-#include <unistd.h>
 
+#include <fcntl.h>
+#include <libudev.h>
 #include <sys/eventfd.h>
+#include <unistd.h>
 
 #include "Common/Assert.h"
 #include "Common/Flag.h"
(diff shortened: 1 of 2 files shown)


Source/Core/InputCommon/ControllerInterface/evdev/evdev.cpp
```diff
@@ -186,91 +227,105 @@ void Shutdown()
...
   {
-    AddOutput(new RumbleEffect(m_fd, RumbleEffect::Motor::Strong));
-    AddOutput(new RumbleEffect(m_fd, RumbleEffect::Motor::Weak));
+    AddOutput(new RumbleEffect(fd, RumbleEffect::Motor::Strong));
+    AddOutput(new RumbleEffect(fd, RumbleEffect::Motor::Weak));
   }
 
   // TODO: Add leds as output devices
 
   // Was there some reasoning behind these numbers?
```
riking: Highly probable that the reasoning is: You need two axes for a joystick, and eight buttons for a NES controller.

mbc07:
Anything new on this PR? Seems to be the last missing piece to have DualShock 4 motion sensors working natively on Linux...

jordan-woyak:
@mbc07 I have some ideas on how to do it more cleanly.

jordan-woyak:
New PR #8473 .

PR #8358, opened 2019-09-09
Add option to report empty GC disc drive as closed (issue 11840)

blaahaj:
https://bugs.dolphin-emu.org/issues/11840

This gives important doot-doot-doot music and bouncy cube.

Only problem: when an empty disc drive is reported as closed, the IPL encounters a read error:

> The disc could not be read.
>
> Please read the NINTENDO GAMECUBE  
> Instruction Booklet for more information.

This means the menu does not load afterwards.

I don't know why that happens. I wasn't able to fix it. So setting is off by default.

### DIFF
--- a/Source/Core/Core/HW/DVD/DVDInterface.cpp
+++ b/Source/Core/Core/HW/DVD/DVDInterface.cpp
@@ -347,6 +347,15 @@ static void DTKStreamingCallback(DIInterruptType interrupt_type, const std::vect
   }
 }
 
+static bool ShouldLidBeOpen()
+{
+  // Disc Channel relies on cover being open when no disc is inserted.
+  // The Wii also has no physical cover. (What is this "Mini" you speak of?)
+  // Therefore, the behaviour is only customisable for the GameCube.
+  return (!IsDiscInside() &&
+          (SConfig::GetInstance().bWii || !Config::Get(Config::MAIN_GC_EMPTY_DRIVE_IS_CLOSED)));
+}
+
 void Init()
 {
   ASSERT(!IsDiscInside());
@@ -354,7 +363,7 @@ void Init()
   DVDThread::Start();
 
   s_DISR.Hex = 0;
-  s_DICVR.Hex = 1;  // Disc Channel relies on cover being open when no disc is inserted
+  s_DICVR.Hex = ShouldLidBeOpen();
   s_DICMDBUF[0] = 0;
   s_DICMDBUF[1] = 0;
   s_DICMDBUF[2] = 0;
@@ -400,7 +409,7 @@ void ResetDrive(bool spinup)
     // On the Wii, this can only happen if something other than a DVD is inserted into the disc
     // drive (for instance, an audio CD) and only after it attempts to read it.  Otherwise, it will
     // report the cover as opened.
-    SetDriveState(DriveState::CoverOpened);
+    SetDriveState(ShouldLidBeOpen() ? DriveState::CoverOpened : DriveState::NoMediumPresent);
   }
   else if (!spinup)
   {
@@ -584,7 +593,7 @@ bool AutoChangeDisc()
 static void SetLidOpen()
 {
   u32 old_value = s_DICVR.CVR;
-  s_DICVR.CVR = IsDiscInside() ? 0 : 1;
+  s_DICVR.CVR = ShouldLidBeOpen();
   if (s_DICVR.CVR != old_value)
     GenerateDIInterrupt(DIInterruptType::CVRINT);
 }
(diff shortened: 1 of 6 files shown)


mbc07:
Unless I misread the issue report, having an IPL dump, disabling "Skip Main Menu" option in Dolphin's settings then holding B button on GC Controller 1 when booting any GameCube game should achieve the same, no?

Techjar:
Yeah you misread the issue report. This applies when booting the IPL directly without a disc (Tools > Load GameCube Main Menu).

blaahaj:
Do you think “The disc could not be read.” prevents merging this? I do not expect that this error is a regression. I think it is likelier that it was simply hidden before, because Dolphin did not ever report the drive as empty and closed at the same time when the IPL runs. The new option I add does not default to on, in order that it would not break anything for users unless they should choose to edit the setting.

Pokechu22:
It's probably because the drive isn't returning the proper error code.  If I recall correctly from my earlier tests, the high byte should be 0x03000000 (`ERROR_NO_DISK_L`) and the low bytes should be 0x023a00 (`ERROR_NO_DISK_H`).  I think, at least.

Note that my tests were on the Wii, and were based on a rather silly edge case; the Wii _will_ report the cover being closed with no disc if you insert something other than a DVD in it, such as an audio CD.  This also causes the disc channel to be confused, same as in dolphin.  I don't have a gamecube (and as such, I don't have an IPL dump), so I can't say whether the same thing would work here.

You probably need to add another case here:

https://github.com/dolphin-emu/dolphin/blob/b29ba7119a6695f6302da4d0f59ea50d02ee65b7/Source/Core/Core/HW/DVD/DVDInterface.cpp#L718-L734

blaahaj:
> It's probably because the drive isn't returning the proper error code. If I recall correctly from my earlier tests, the high byte should be 0x03000000 (`ERROR_NO_DISK_L`) and the low bytes should be 0x023a00 (`ERROR_NO_DISK_H`). I think, at least.

If I am remembering right, I did exactly what you said and it did not work. But, I will try this again, it could have been another combination of bits.

Pokechu22:
Oh, I should also note that it should probably be returning `ERROR_COVER` for the high bits in the current case, and that I locally renamed `ERROR_COVER_H` to `ERROR_NO_DISK_H`, which is probably why you didn't run into that...

blaahaj:
@Pokechu22 The current code `s_error_code = ERROR_NO_DISK | ERROR_COVER_H;` is already those hex values.

Pokechu22:
Can you run it with the `DVD Interface (DVD)` log enabled, and then post the log file?  Also, can you post a screenshot of the disc could not be read screen?

blaahaj:
@Pokechu22 

```
34:40:891 Core/HW/DVD/DVDInterface.cpp:928 I[DVD]: Read DiscID 00000000
34:40:909 Core/HW/DVD/DVDInterface.cpp:1070 I[DVD]: DVDLowStopMotor  kill!
```

Screenshot


I apologise that I did not see your comment sooner.

Pokechu22:
I'm not entirely sure what to make of that, unfortunately.  It doesn't seem to have even requested the error, which doesn't match up with my expectations.

What happens if you boot the IPL with a Wii disc set as the default disc (both before and after this change)?

blaahaj:
> What happens if you boot the IPL with a Wii disc set as the default disc (both before and after this change)?

* With 'Report disc drive as closed when empty' - same error message
* Without 'Report disc drive as closed when empty' - same lack of doot doot doot

Apologise again for slow response.

blaahaj:
Rebased code.

It seems the error message was fixed?

blaahaj:
Now:

* With 'Report disc drive as closed when empty' - doot doot doot. No error message. ✔️ 
* Without 'Report disc drive as closed when empty' - no doot doot doot. No error message. ✔️

PAL & NTSC-U IPL tested.

blaahaj:
Could some one remove the " the WIP / do not merge "?

Source/Core/Core/HW/DVD/DVDInterface.cpp
```diff
@@ -347,14 +347,23 @@ static void DTKStreamingCallback(DIInterruptType interrupt_type, const std::vect
   }
 }
 
+static bool ShouldLidBeOpen()
+{
+  // Disc Channel relies on cover being open when no disc is inserted
+  // The Wii also has no physical cover
```
MayImilae, line 353: About that... 

wiimini

Of course we don't target the Wii Mini, so this is all semantics. Might be nice to include a nod to this in the comment though!

Anyway, I'll take a look at the UI for this shortly and do a follow up.

Rumi-Larry:
What happened with the pr? :(

Zopolis4:
What's the status of this?

AdmiralCurtiss:
Can you rebase this?

AdmiralCurtiss:
Rebased as #11570

PR #8359, opened 2019-09-11
Windows GUI: Use native font size from new API, not 120% of old API

jimbo1qaz:
Not sure if this code is actually better in practice, but it's more "correct" than before, I think. There's no visual difference at 9pt and 96dpi, and probably minimal difference at higher DPIs. Maybe it affects users who customize their font sizes?

Code was copied from Qt library's dev branch's QWindowsFontDatabase, which (unlike the 5.x branch) uses `SPI_GETNONCLIENTMETRICS` to get the right font on Vista and up.

I commented out the lines about `qFont.setWeight(QPlatformFontDatabase::weightFromInteger(logFont.lfWeight))` (so this code won't support bold). This touches more existing code than necessary and may not be the best fit in the codebase (unsure if this code is unfitting, proof-of-concept, or actually good encapsulation). I tested that it works at 100% and 175% DPI scaling.

If I divide by 96 instead of `GetDeviceCaps()`, both -lfHeight and Qt's "point size" are multiplied by the DPI scaling, resulting in erroneously large text.

maybe @spycrab can review?

### DIFF
--- a/Source/Core/DolphinQt/Main.cpp
+++ b/Source/Core/DolphinQt/Main.cpp
@@ -80,6 +80,45 @@ static bool QtMsgAlertHandler(const char* caption, const char* text, bool yes_no
   return false;
 }
 
+#ifdef _WIN32
+// Taken from https://github.com/qt/qtbase/blob/835c3e94f6089751421a19008d442faec9649ed8/src/platformsupport/fontdatabases/windows/qwindowsfontdatabase.cpp#L2064
+int defaultVerticalDPI()
+{
+  static int vDPI = -1;
+  if (vDPI == -1)
+  {
+    if (HDC defaultDC = GetDC(0))
+    {
+      vDPI = GetDeviceCaps(defaultDC, LOGPIXELSY);
+      ReleaseDC(0, defaultDC);
+    }
+    else
+    {
+      // FIXME: Resolve now or return 96 and keep unresolved?
+      vDPI = 96;
+    }
+  }
+  return vDPI;
+}
+
+// Taken from https://github.com/qt/qtbase/blob/835c3e94f6089751421a19008d442faec9649ed8/src/platformsupport/fontdatabases/windows/qwindowsfontdatabase.cpp#L2048
+QFont LOGFONT_to_QFont(const LOGFONT& logFont, int verticalDPI_In = 0)
+{
+  if (verticalDPI_In <= 0)
+    verticalDPI_In = defaultVerticalDPI();
+  QFont qFont(QString::fromWCharArray(logFont.lfFaceName));
+  qFont.setItalic(logFont.lfItalic);
+  //if (logFont.lfWeight != FW_DONTCARE)
+  //  qFont.setWeight(QPlatformFontDatabase::weightFromInteger(logFont.lfWeight));
+  const qreal logFontHeight = qAbs(logFont.lfHeight);
+  qFont.setPointSizeF(logFontHeight * 72.0 / qreal(verticalDPI_In));
+  qFont.setUnderline(logFont.lfUnderline);
+  qFont.setOverline(false);
+  qFont.setStrikeOut(logFont.lfStrikeOut);
+  return qFont;
+}
+#endif
+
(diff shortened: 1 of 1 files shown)


jimbo1qaz:
There hasn't been any comments yet. Will this be merged if I fix the lint errors, or is it still waiting for code review?

Warepire:
What would be needed to not need to copy code from Qt, and just use the actual infrastructure from Qt?

jimbo1qaz:
This code was copied from Qt dev (Qt 6 dev), and I believe it is not present in 5.x.

Warepire:
Looks like they're a part of Qt 5.11 to me: https://github.com/qt/qtbase/blob/5.11/src/platformsupport/fontdatabases/windows/qwindowsfontdatabase.cpp#L2026

jimbo1qaz:
Qt on Windows has [class QWindowsFontDatabase with a method to compute the system font](https://code.woboq.org/qt5/qtbase/src/platformsupport/fontdatabases/windows/qwindowsfontdatabase.cpp.html#_ZN20QWindowsFontDatabase17systemDefaultFontEv). It uses `#ifdef` to return the right font on qt 6, but wrong font on qt 5 for backwards-compatibility.

> Qt 5 by (Qt 4) legacy uses GetStockObject(DEFAULT_GUI_FONT) to obtain the default GUI font (typically "MS Shell Dlg 2, 8pt"). This has been long deprecated; the message font of the NONCLIENTMETRICS structure obtained by SystemParametersInfo(SPI_GETNONCLIENTMETRICS) should be used instead (see QWindowsTheme::refreshFonts(), typically "Segoe UI, 9pt"), which is larger.

-----------

<details>
<summary>Semi-relevant files in Qt's source code</summary>
Can I use the function, or is it private within Qt?

- https://code.woboq.org/qt5/qtbase/src/plugins/platforms/windows/qwindowsnativeinterface.h.html#_ZN23QWindowsNativeInterface14logFontToQFontEPKvi
- https://code.woboq.org/qt5/qtbase/src/plugins/platforms/windows/qwindowsnativeinterface.cpp.html#_ZN23QWindowsNativeInterface14logFontToQFontEPKvi
- https://code.woboq.org/qt5/qtbase/src/platformsupport/fontdatabases/windows/qwindowsfontdatabase_p.h.html#_ZN20QWindowsFontDatabase16LOGFONT_to_QFontERKii
- https://code.woboq.org/qt5/qtbase/src/platformsupport/fontdatabases/windows/qwindowsfontdatabase.cpp.html#_ZN20QWindowsFontDatabase16LOGFONT_to_QFontERKii

https://github.com/qt/qtbase/blob/13e0a36626bd75e631dd9536e795a494432b1945/src/plugins/platforms/windows/qwindowsintegration.h includes https://github.com/qt/qtbase/blob/13e0a36626bd75e631dd9536e795a494432b1945/src/platformsupport/fontdatabases/windows/qwindowsfontdatabase_p.h#L123 exposes LOGFONT_to_QFont().
</details>

`logFontToQFont` is a wrapper around `LOGFONT_to_QFont` and found in neither C:\Qt\... nor Dolphin Qt's headers.

`QWindowsFontDatabase` and `LOGFONT_to_QFont` is header-declared in `C:\Qt\5.13.0\msvc2017\include\QtFontDatabaseSupport\5.13.0\QtFontDatabaseSupport\private\qwindowsfontdatabase_p.h` (but not Dolphin's Qt submodule). But I wouldn't use it...

https://code.woboq.org/qt5/qtbase/src/platformsupport/fontdatabases/windows/qwindowsfontdatabase_p.h.html#43

```
//
//  W A R N I N G
//  -------------
//
// This file is not part of the Qt API. It exists purely as an
// implementation detail. This header file may change from version to
// version without notice, or even be removed.
//
// We mean it.
//
```

I think copying out the functions I use is a good thing.

jimbo1qaz:
On Stack Overflow, I found a quick and easy trick to accomplish this easily: https://stackoverflow.com/a/40380684

`qApp->setFont(QApplication::font("QMenu"));`

Should I submit a new pull request with just that code?

PR #8318, opened 2019-08-17
InputCommon: Dynamic Input Textures

iwubcode:
Per game input profiles make it easy to play a variety of Wii games.  However, due to the need to custom tailor input for most Wii games, it can be difficult to remember what input keys map to what.  It can be even more confusing when you haven't played a game in a while!  Some users have created texture packs that overwrite buttons in the UI with gamepad specific buttons but that can be frustrating because it forces a control scheme.  Enter "Dynamic Input Textures"!

Dynamic Input Textures allow texture creators to replace areas of a texture using a bunch of color boxes.  When the input keys are loaded, a custom texture gets generated with images representing the keys you have set!

Example [video](https://i.imgur.com/KW0lsPf.mp4)

Note: A big thank you to Jakey455 on the Dolphin forums for the XboxOne textures I used in my example

PR #8360, opened 2019-09-14
Resource packs: support dynamic input tex

iwubcode:
This builds on #8318 by adding support for Resource Packs.  It makes Resource Packs more generic so they deal with a file that is a particular "ResourceType" (currently just texture or dynamic input).

It also does two other things:

* In master, if you moved priority up/down of a pack.  It would uninstall the pack first but never reinstall it.  This seemed like an oversight and has been fixed
* A pack can be installed or uninstalled (or priority changed) when the game is running.  Texture only packs reload the textures and the cache is invalidated.  For dynamic texture packs, the textures are regenerated (and textures are reloaded / cache is invalidated).

### DIFF
--- a/Source/Core/UICommon/ResourcePack/ResourcePack.cpp
+++ b/Source/Core/UICommon/ResourcePack/ResourcePack.cpp
@@ -5,6 +5,8 @@
 #include "UICommon/ResourcePack/ResourcePack.h"
 
 #include <algorithm>
+#include <array>
+#include <utility>
 
 #include <unzip.h>
 
@@ -21,6 +23,12 @@
 namespace ResourcePack
 {
 constexpr char TEXTURE_PATH[] = HIRES_TEXTURES_DIR DIR_SEP;
+constexpr char DYNAMIC_INPUT_PATH[] = DYNAMICINPUT_DIR DIR_SEP;
+
+bool ResourcePack::ResourceFile::operator<(const ResourcePack::ResourceFile& file) const
+{
+  return input_path < file.input_path;
+}
 
 ResourcePack::ResourcePack(const std::string& path) : m_path(path)
 {
@@ -79,26 +87,52 @@ ResourcePack::ResourcePack(const std::string& path) : m_path(path)
 
   unzGoToFirstFile(file);
 
+  constexpr std::array<ResourceType, 2> resource_types = {ResourceType::TEXTURE,
+                                                          ResourceType::DYNAMIC_INPUT};
+
   do
   {
     std::string filename(256, '\0');
 
-    unz_file_info texture_info;
-    unzGetCurrentFileInfo(file, &texture_info, filename.data(), static_cast<u16>(filename.size()),
+    unz_file_info file_info;
+    unzGetCurrentFileInfo(file, &file_info, filename.data(), static_cast<u16>(filename.size()),
                           nullptr, 0, nullptr, 0);
 
-    if (filename.compare(0, 9, "textures/") != 0 || texture_info.uncompressed_size == 0)
-      continue;
-
-    // If a texture is compressed and the manifest doesn't state that, abort.
-    if (!m_manifest->IsCompressed() && texture_info.compression_method != 0)
+    for (auto resource_type : resource_types)
     {
-      m_valid = false;
-      m_error = "Texture " + filename + " is compressed!";
-      return;
-    }
+      std::string zip_directory_root;
+      std::string load_directory_root;
+      if (resource_type == ResourceType::TEXTURE)
+      {
+        zip_directory_root = "textures/";
+        load_directory_root = TEXTURE_PATH;
+      }
+      else
+      {
+        zip_directory_root = "dynamic_input/";
+        load_directory_root = DYNAMIC_INPUT_PATH;
+      }
 
-    m_textures.push_back(filename.substr(9));
+      if (filename.compare(0, zip_directory_root.size(), zip_directory_root) != 0 ||
+          file_info.uncompressed_size == 0)
+      {
+        continue;
+      }
+
+      // If a file is compressed and the manifest doesn't state that, abort.
+      if (!m_manifest->IsCompressed() && file_info.compression_method != 0)
+      {
+        m_valid = false;
+        m_error =
+            "File " + filename + " is compressed when the resource pack says it shouldn't be!";
+        return;
+      }
+
+      m_files.emplace(ResourceFile{filename,
+                                   load_directory_root + filename.substr(zip_directory_root.size()),
+                                   resource_type});
+      m_types_provided |= static_cast<u32>(resource_type);
+    }
   } while (unzGoToNextFile(file) != UNZ_END_OF_LIST_OF_FILE);
 }
 
@@ -127,9 +161,14 @@ const Manifest* ResourcePack::GetManifest() const
   return m_manifest.get();
 }
 
-const std::vector<std::string>& ResourcePack::GetTextures() const
+const std::set<ResourcePack::ResourceFile>& ResourcePack::GetFiles() const
 {
-  return m_textures;
+  return m_files;
+}
+
+u32 ResourcePack::GetResourceTypesSupported() const
+{
+  return m_types_provided;
 }
 
 bool ResourcePack::Install(const std::string& path)
@@ -140,18 +179,17 @@ bool ResourcePack::Install(const std::string& path)
     return false;
   }
 
-  auto file = unzOpen(m_path.c_str());
-  Common::ScopeGuard file_guard{[&] { unzClose(file); }};
+  auto resource_pack = unzOpen(m_path.c_str());
+  Common::ScopeGuard file_guard{[&] { unzClose(resource_pack); }};
 
-  for (const auto& texture : m_textures)
+  for (const auto& file : m_files)
   {
     bool provided_by_other_pack = false;
 
-    // Check if a higher priority pack already provides a given texture, don't overwrite it
+    // Check if a higher priority pack already provides a given file, don't overwrite it
     for (const auto& pack : GetHigherPriorityPacks(*this))
     {
-      if (std::find(pack->GetTextures().begin(), pack->GetTextures().end(), texture) !=
-          pack->GetTextures().end())
+      if (pack->GetFiles().count(file) != 0)
       {
         provided_by_other_pack = true;
         break;
@@ -161,15 +199,16 @@ bool ResourcePack::Install(const std::string& path)
     if (provided_by_other_pack)
       continue;
 
-    if (unzLocateFile(file, ("textures/" + texture).c_str(), 0) != UNZ_OK)
+    if (unzLocateFile(resource_pack, (file.input_path).c_str(), 0) != UNZ_OK)
     {
-      m_error = "Failed to locate texture " + texture;
+      m_error = "Failed to locate file " + file.input_path;
       return false;
     }
 
-    const std::string texture_path = path + TEXTURE_PATH + texture;
+    const std::string output_path = path + file.output_path;
+
     std::string m_full_dir;
-    SplitPath(texture_path, &m_full_dir, nullptr, nullptr);
+    SplitPath(output_path, &m_full_dir, nullptr, nullptr);
 
     if (!File::CreateFullPath(m_full_dir))
     {
@@ -177,21 +216,21 @@ bool ResourcePack::Install(const std::string& path)
       return false;
     }
 
-    unz_file_info texture_info;
-    unzGetCurrentFileInfo(file, &texture_info, nullptr, 0, nullptr, 0, nullptr, 0);
+    unz_file_info file_info;
+    unzGetCurrentFileInfo(resource_pack, &file_info, nullptr, 0, nullptr, 0, nullptr, 0);
 
-    std::vector<char> data(texture_info.uncompressed_size);
-    if (!Common::ReadFileFromZip(file, &data))
+    std::vector<char> data(file_info.uncompressed_size);
+    if (!Common::ReadFileFromZip(resource_pack, &data))
     {
-      m_error = "Failed to read texture " + texture;
+      m_error = "Failed to read " + file.input_path;
       return false;
     }
 
-    std::ofstream out(texture_path, std::ios::trunc | std::ios::binary);
+    std::ofstream out(output_path, std::ios::trunc | std::ios::binary);
 
     if (!out.good())
     {
-      m_error = "Failed to write " + texture;
+      m_error = "Failed to write " + output_path;
       return false;
     }
 
@@ -215,16 +254,14 @@ bool ResourcePack::Uninstall(const std::string& path)
 
   SetInstalled(*this, false);
 
-  for (const auto& texture : m_textures)
+  for (const auto& file : m_files)
   {
     bool provided_by_other_pack = false;
 
-    // Check if a higher priority pack already provides a given texture, don't delete it
+    // Check if a higher priority pack already provides a given file, don't delete it
     for (const auto& pack : GetHigherPriorityPacks(*this))
     {
-      if (::ResourcePack::IsInstalled(*pack) &&
-          std::find(pack->GetTextures().begin(), pack->GetTextures().end(), texture) !=
-              pack->GetTextures().end())
+      if (::ResourcePack::IsInstalled(*pack) && pack->GetFiles().count(file) != 0)
       {
         provided_by_other_pack = true;
         break;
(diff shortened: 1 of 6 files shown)


Source/Core/InputCommon/ImageOperations.cpp
```diff
@@ -0,0 +1,381 @@
...
+{
+  File::IOFile file;
+  file.Open(path, "rb");
+  std::vector<u8> buffer(file.GetSize());
+  file.ReadBytes(buffer.data(), file.GetSize());
+
+  ImagePixelData image;
+  std::vector<u8> data;
+  if (!Common::LoadPNG(buffer, &data, &image.width, &image.height))
+    return {};
```
lioncash: `return std::nullopt;`

The nullopt constructor and default constructor invocations don't always do the [same thing](https://godbolt.org/z/UHP5B9)

Source/Core/InputCommon/ImageOperations.cpp
```diff
@@ -0,0 +1,381 @@
...
+  return result;
+}
+
+std::vector<Rect> FindRectRegionsOfNeighboringPixels(const ImagePixelData& src)
+{
+  static const Pixel mask_color{0, 0, 0, 255};
+
+  std::vector<Rect> regions;
+  std::vector<bool> seen_pixels;
+  seen_pixels.resize(src.width * src.height);
```
lioncash: This can be collapsed into the constructor:

```suggestion
  std::vector<bool> seen_pixels(src.width * src.height);
```

or, this could be moved entirely into CalculateNeighboringPixels (see next review comment). It could also be dropped entirely from the interface, given it's not actually used, aside from being stored into in CalculateNeighboringPixels.

Source/Core/InputCommon/ImageOperations.cpp
```diff
@@ -0,0 +1,381 @@
...
+namespace
+{
+Pixel SampleNearest(const ImagePixelData& src, double u, double v)
+{
+  const u32 x = std::clamp(static_cast<u32>(u * src.width), 0u, src.width - 1);
+  const u32 y = std::clamp(static_cast<u32>(v * src.height), 0u, src.height - 1);
+  return src.pixels[x + y * src.width];
+}
+
+void CalculateNeighboringPixels(const ImagePixelData& src, std::vector<Point>& neighboring_pixels,
```
lioncash: These vectors can be returned as a pair, which allows them to be unpacked with structured bindings.

Source/Core/InputCommon/ImageOperations.cpp
```diff
@@ -0,0 +1,381 @@
...
+  }
+  CopyTo(resized, result, Rect{Point{0, 0}, resized.width, resized.height},
+         Rect{Point{left, top}, resized.width, resized.height});
+
+  return result;
+}
+
+std::vector<Rect> FindRectRegionsOfNeighboringPixels(const ImagePixelData& src)
+{
+  static const Pixel mask_color{0, 0, 0, 255};
```
lioncash:
```suggestion
  static constexpr Pixel mask_color{0, 0, 0, 255};
```

Source/Core/InputCommon/ImageOperations.cpp
```diff
@@ -0,0 +1,381 @@
...
+        result.pixels[y * src.width + x] = Pixel{0, 0, 0, 255};
+      }
+    }
+  }
+  return result;
+}
+
+ImagePixelData Resize(ResizeMode mode, const ImagePixelData& src, u32 new_width, u32 new_height)
+{
+  ImagePixelData result;
```
lioncash: This and the following three lines can be collapsed into:

```suggestion
  ImagePixelData result(new_width, new_height);
```

iwubcode:
Thank you @lioncash for your feedback.  Lots of good stuff.

Source/Core/InputCommon/ImageOperations.cpp
iwubcode: Very interesting, I wasn't aware of that.  Thank you!

Source/Core/InputCommon/ImageOperations.cpp
iwubcode: Unless I'm missing something from your suggestion, I don't think `seen_pixels` can move into `CalculateNeighboringPixels` because it is built up from multiple for loop passes.  The other vector (as you mention below) could be moved to be a return value instead though.

lioncash: Ah OK, I must have missed the read from the vector in the function when doing my initial review. Disregard the latter half of my comment.

Source/Core/InputCommon/DynamicInputTextureConfiguration.h
```diff
@@ -0,0 +1,48 @@
...
+{
+class DynamicInputTextureConfiguration
+{
+public:
+  explicit DynamicInputTextureConfiguration(const std::string& json_file);
+  ~DynamicInputTextureConfiguration();
+  void GenerateTextures(const IniFile::Section* sec);
+
+private:
+  struct DynamicInputTextureData
```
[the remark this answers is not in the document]
iwubcode: I like this idea but the vector's type but VisualStudio complains about this change.  I did a similar forward declaration of `DynamicInputTextureConfiguration` in `DynamicInputTextureManager`'s header but with this forward declaration it complains there about not having the size of `DynamicInputTextureConfiguration`.  I'm not too familiar with this style of forward declaration, I mostly forward declare pointer types, so I might be missing something.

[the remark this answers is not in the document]
lioncash: Can you provide both the header and source file with the changes you made that caused the compilation error?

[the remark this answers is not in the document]
iwubcode: @lioncash - please see the last commit on #8318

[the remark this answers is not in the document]
iwubcode: To give an example:

```
// a.h
struct B;
struct A
{
 A();
 ~A();
 B b;
};
void do_something(const& A a);

// a.cpp
#include "b.h"
A::A() = default;
A::~A() = default;
void do_something(const& A a)
{
  // ...
}

// b.h
struct B
{
  B();
  ~B();
  struct Imp;
  imp imp;
};

// B.cpp
struct B::Imp
{
  int a = 3;
};
B::B() = default;
B::~B() = default;
```

(didn't actually compile this but it portrays the basic idea) We can't compile "A" because "B" entire size isn't known from within A's cpp file

Source/Core/UICommon/ResourcePack/ResourcePack.cpp
```diff
@@ -19,6 +21,7 @@
 namespace ResourcePack
 {
 constexpr char TEXTURE_PATH[] = "Load/Textures/";
+constexpr char DYNAMIC_INPUT_PATH[] = "Load/DynamicInputTextures";
```
BhaaLseN: `TEXTURE_PATH` has a trailing slash, while this one does not. Does it make a difference to use either for consistency? I for one prefer folder paths to end in a slash so I don't have to worry about it elsewhere.

Source/Core/InputCommon/DynamicInputTextureConfiguration.h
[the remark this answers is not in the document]
iwubcode: I reverted the change (since it wasn't building as is), it only saves a map include.  Let me know though if I'm missing something.

Source/Core/InputCommon/InputCommon.vcxproj
```diff
@@ -35,6 +35,16 @@
     <Import Project="..\..\VSProps\PCHUse.props" />
   </ImportGroup>
   <PropertyGroup Label="UserMacros" />
+  <ItemDefinitionGroup Condition="'$(Configuration)|$(Platform)'=='Release|x64'">
+    <ClCompile>
+      <AdditionalIncludeDirectories>C:\dev\opencv\build\include;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
```
BhaaLseN: I don't think we should be committing a local path like `C:\dev\opencv` here. Also, are those two `ItemDefinitionGroup` now obsolete that you removed the OpenCV dependency?

iwubcode: Yes this whole block can be removed, thanks

Source/Core/InputCommon/DynamicInputTextureConfiguration.cpp
```diff
@@ -0,0 +1,251 @@
...
+
+namespace InputCommon
+{
+DynamicInputTextureConfiguration::DynamicInputTextureConfiguration(const std::string& json_file)
+{
+  std::ifstream json_stream;
+  File::OpenFStream(json_stream, json_file, std::ios_base::in);
+  if (!json_stream.is_open())
+  {
+    ERROR_LOG(PAD, "Failed to load dynamic input json file '%s'", json_file.c_str());
```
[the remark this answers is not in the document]
iwubcode: Do you have a suggestion for a logger?  It was a tossup between PAD and one of the graphics ones.  Since this is for inputs, I picked PAD

Source/Core/InputCommon/DynamicInputTextureConfiguration.cpp
```diff
@@ -0,0 +1,251 @@
...
+    m_valid = false;
+    return;
+  }
+  const std::string json_str((std::istreambuf_iterator<char>(json_stream)),
+                             (std::istreambuf_iterator<char>()));
```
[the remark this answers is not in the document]
iwubcode: Coming back to this, this is indeed cryptic and quite slow.  I will change it

Source/Core/InputCommon/ImageOperations.cpp
```diff
@@ -0,0 +1,380 @@
...
+      image.pixels[index] = pixel;
+    }
+  }
+
+  return image;
+}
+
+#ifdef _MSC_VER
+#pragma warning(push)
+#pragma warning(disable : 4611)
```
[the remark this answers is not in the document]
iwubcode: Done, thanks!

iwubcode:
A big thank you to @BhaaLseN  for reviewing.  I updated the code based on the feedback.

I also pulled in @spycrab 's original ResourcePack document which used to live outside the Dolphin repo (here:  https://gist.github.com/spycrab/9d05056755d8d7908bdb871a99d050bf) and then I updated it with the new dynamic input information (need to figure out how to link to another md file).

Source/Core/InputCommon/DynamicInputTextureConfiguration.cpp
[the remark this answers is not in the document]
BhaaLseN: Very good point. I'd rather use `VIDEO` (or not?) because it changes something graphics related (ie. the result is shown on screen; or _not_ when something is wrong). Guessing Screenshots/FrameDumps go to `VIDEO` for that reason.
Or we define a new one; perhaps specific to resource packs, perhaps generic enough to handle many different customization points?
I can live with either, but feel like `PAD` is the last one of them that I'd look at for a _texture_ related message.

CustomFileFormats/ResourcePacks.md
```diff
@@ -0,0 +1,114 @@
+# Dolphin Resource Pack Specification v3
```
BhaaLseN: Any reason you didnt just update `docs/ResourcerPacks.md` instead?

CustomFileFormats/DynamicInputTextures.md
```diff
@@ -0,0 +1,136 @@
+# Dolphin Dynamic Input Textures Specification (v1)
```
BhaaLseN: Just like `ResourcePacks.md`, this should probably go in the `docs/` folder.

CustomFileFormats/DynamicInputTextures.md
```diff
@@ -0,0 +1,136 @@
...
+|-------------------------|---------|-------|
+|``colorkey_image``       | **Yes** | v1    |
+|``fallback_image``       | **Yes** | v1    |
+|``emulated_controls``    | **Yes** | v1    |
+|``device``               | No      | v1    |
+|``host_controls``        | No      | v1    |
+
+*colorkey_image* - an image with colored boxes, will be used as a base image, so any non-input related sections can be upscaled/redrawn if desired
+
+*fallback_image* - after looking at the colorkey_image, if the color isn't mapped to a specific button, it will use this image as a reference for what to put in the color. This should be the same size/quality as the colorkey_image but without the colored boxes
```
BhaaLseN: I personally prefer identifiers to be monospaced, and at a glance it seems `ResourcePacks.md` also does this.
Would you mind updating them here?
```suggestion
*fallback_image* - after looking at the ``colorkey_image``, if the color isn't mapped to a specific button, it will use this image as a reference for what to put in the color. This should be the same size/quality as the ``colorkey_image`` but without the colored boxes
```

iwubcode: Did you mean instead of being italicized?  I had made one italicized to visually differentiate them from the table which is monospaced.  I can try monospacing everything though.  I wanted the descriptions in the table but they are much longer than the ones in ResourcePacks

CustomFileFormats/ResourcePacks.md
iwubcode: I saw `docs` but missed that `ResourcePacks.md` was in there.  Thanks!

CustomFileFormats/DynamicInputTextures.md
iwubcode: Will do, thank you!

CustomFileFormats/DynamicInputTextures.md
BhaaLseN: I mainly mean use of previously (emphasised or not) identifiers to set them apart from plain-text; whether thats italicized or monospaced or whatever is up to you (as long as it is consistent)

Source/Core/InputCommon/DynamicInputTextureConfiguration.cpp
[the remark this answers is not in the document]
iwubcode: Yeah, I originally thought about making my own category but for such a minor feature, I didn't know if it was worth it.  I don't really mind one way or other so I'll move it to `VIDEO`.

CustomFileFormats/DynamicInputTextures.md
iwubcode: Ok updated.  Please let me know if this isn't what you had in mind.

Source/Core/UICommon/ResourcePack/ResourcePack.cpp
```diff
@@ -78,26 +86,49 @@ ResourcePack::ResourcePack(const std::string& path) : m_path(path)
 
   unzGoToFirstFile(file);
 
+  std::array<ResourceType, 2> resource_types = {ResourceType::TEXTURE, ResourceType::DYNAMIC_INPUT};
```
BhaaLseN: Can (or should) this be `static`, or perhaps even `constexpr`?

Source/Core/UICommon/ResourcePack/ResourcePack.cpp
```diff
@@ -78,26 +86,49 @@ ResourcePack::ResourcePack(const std::string& path) : m_path(path)
...
+      }
+      else
+      {
+        zip_directory_root = "dynamic_input/";
+        load_directory_root = DYNAMIC_INPUT_PATH;
+      }
+
+      if (filename.compare(0, zip_directory_root.size(), zip_directory_root) != 0 ||
+          file_info.uncompressed_size == 0)
+        continue;
```
BhaaLseN: Since this `if` condition spans multiple lines, curlies should be used for the body.

docs/DynamicInputTextures.md
```diff
@@ -0,0 +1,139 @@
...
+|``emulated_controls``    | **Yes** | v1    |
+|``device``               | No      | v1    |
+|``host_controls``        | No      | v1    |
+
+*colorkey_image* - an image with colored boxes, will be used as a base image, so any non-input related sections can be upscaled/redrawn if desired.
+
+*fallback_image* - after looking at the ``colorkey_image``, if the color isn't mapped to a specific button, it will use this image as a reference for what to put in the color. This should be the same size/quality as the ``colorkey_image`` but without the colored boxes.
+
+*emulated_controls* - a series of emulated buttons that map to a given color rectangle.
+
```
BhaaLseN: Did you intentionally leave _two_ blank lines here between explainations, vs. one between every other?

iwubcode: Whoops, no that wasn't intentional

RinMaru:
is it possible to use better quality textures?

Rumi-Larry:
What happened with this pr?

iwubcode:
While this is perfectly mergeable as is.  I have recently had an idea that might affect this PR.  If in the future this PR is still relevant, I'll re-open.

PR #8361, opened 2019-09-14
Remove XAudio2

ligfx:
Audio on Windows is provided by the cubeb (shared mode) and WASAPI (exclusive mode) backends, both of which exhibit lower latency.

What to do about users who have XAudio2 configured? Dolphin could:
- Ignore the problem. Users who upgrade would need to change their backend manually.
- Automatically use the default backend instead if XAudio2 is configured
- Automatically use the default backend instead if _any_ unknown backend is configured

I went with the third option here, but happy to implement any of the others.


dreamsyntax:
After upgrading while my config was set to XAudio2

It now has "No Audio Output" as the Audio Backend, but sound still works as normal.
Is this intended?

JosJuice:
Thanks for the report. Fixed in PR #8392.

MayImilae:
@ligfx Would you mind giving me an overview of this? Like, this will be in the progress report, and the questions we need to answer in the progress report are roughly:

* Why was xaudio2 removed?
* How is this good/bad for Dolphin?
* How is this good/bad for users?
* Was anything particularly bad about xaudio2 support? Or anything particularly good about cubeb?
* Will users have a worse, better, or same experience?
* How will users be transitioned from xaudio2 and will they need to do anything?

I can answer most of these myself, but not all of them and I'm sure I'd be missing details. Would you mind telling me your answers to those questions?

MayImilae:
Thanks!

>XAudio2 has worse latency/performance than Cubeb/WASAPI. This ensures users are using a better audio backend.

Do you have any numbers for that? That would be nice to have!

CookiePLMonster:
> Do you have any numbers for that? That would be nice to have!

Looks like some tests were made for a progress report detailing WASAPI addition:
https://pl.dolphin-emu.org/blog/2018/06/03/dolphin-progress-report-april-and-may-2018/

JosJuice:
> Looks like some tests were made for a progress report detailing WASAPI addition:
https://pl.dolphin-emu.org/blog/2018/06/03/dolphin-progress-report-april-and-may-2018/

That's WASAPI in exclusive mode, which has even lower latency than the mode Cubeb uses.

CookiePLMonster:
> That's WASAPI in exclusive mode, which has even lower latency than the mode Cubeb uses.

Sure, but there are also figures about Cubeb vs XAudio2 which seems to be exactly what is being asked about ;)

JosJuice:
Ah! Yes, sorry, I didn't look at the actual graph.

mirh:
I don't want to be that guy, but just while you were writing these last comments, microsoft silently released **XAudio 2.9** ***for Windows 7*** (which also comes with its own headers, solving that other problem)
https://github.com/MicrosoftDocs/win32/blob/docs/desktop-src/xaudio2/xaudio2-redistributable.md

And I am now wondering if the performance advantage cubeb had (and has?) couldn't just be not using the IAudioClient3 new shiny W10 api.

PR #8362, opened 2019-09-14
Various string and other optimizations

GerbilSoft:
This PR has a few optimizations to reduce the total code size of the Dolphin emulator:
* VolumeVerifier.cpp: Use arrays of C strings for the disc IDs instead of unordered_map<string>. For small lists, use a linear search; for larger lists, use std::binary_search().
* CBoot::SetupWiiMemory(): Reworked `struct RegionSetting` to use C strings.
* PowerPC.cpp: Don't explicitly initialize ppcState. Explicit initialization results in ~4 MB of zeroes being added to the .data section. Without explicit initialization, this turns into an entry in the .bss section, which saves space on disk. The effective result is the same, since .bss entries are zeroed on load.
* GetSysMenuVersionString(): Use a char instead of std::string for the region letter.
* OGL/Render.cpp: Use C string functions and get rid of the std::string copies.
* GLExtensions.cpp: Use C string arrays instead of std::string arrays for extensions.
* InputConfig::LoadConfig(): Convert num[] to an array of C strings.
* Use std::istringstream or std::ostringstream instead of std::stringstream where possible.
* Use pre-increment for iterators instead of post-increment.
* IsSettingSaveable.cpp: Use std::array instead of std::vector; use pointers instead of copies of the ConfigLocation objects. This reduces the total size of the object file from ~16 KB to ~1 KB.

Note that a lot of these optimizations were done by using C strings instead of std::string. std::string can be useful because it allows for dynamic resizing and doesn't have buffer overflow issues, but C strings are better for constant data because they don't have the size and runtime overhead of std::string.

### DIFF
--- a/Source/Core/Common/GL/GLExtensions/GLExtensions.cpp
+++ b/Source/Core/Common/GL/GLExtensions/GLExtensions.cpp
@@ -2198,7 +2198,7 @@ static void InitExtensionList(GLContext* context)
     default:
     case 450:
     {
-      std::string gl450exts[] = {
+      static const char* const gl450exts[] = {
           "GL_ARB_ES3_1_compatibility",
           "GL_ARB_clip_control",
           "GL_ARB_conditional_render_inverted",
@@ -2216,7 +2216,7 @@ static void InitExtensionList(GLContext* context)
     }
     case 440:
     {
-      std::string gl440exts[] = {
+      static const char* const gl440exts[] = {
           "GL_ARB_buffer_storage",
           "GL_ARB_clear_texture",
           "GL_ARB_enhanced_layouts",
@@ -2232,7 +2232,7 @@ static void InitExtensionList(GLContext* context)
     }
     case 430:
     {
-      std::string gl430exts[] = {
+      static const char* const gl430exts[] = {
           "GL_ARB_ES3_compatibility",
           "GL_ARB_arrays_of_arrays",
           "GL_ARB_clear_buffer_object",
@@ -2260,7 +2260,7 @@ static void InitExtensionList(GLContext* context)
     }
     case 420:
     {
-      std::string gl420exts[] = {
+      static const char* const gl420exts[] = {
           "GL_ARB_base_instance",
           "GL_ARB_compressed_texture_pixel_storage",
           "GL_ARB_conservative_depth",
@@ -2280,7 +2280,7 @@ static void InitExtensionList(GLContext* context)
     }
     case 410:
     {
-      std::string gl410exts[] = {
+      static const char* const gl410exts[] = {
           "GL_ARB_ES2_compatibility",
           "GL_ARB_get_program_binary",
           "GL_ARB_separate_shader_objects",
@@ -2294,7 +2294,7 @@ static void InitExtensionList(GLContext* context)
     }
     case 400:
     {
-      std::string gl400exts[] = {
+      static const char* const gl400exts[] = {
           "GL_ARB_draw_indirect",
           "GL_ARB_gpu_shader5",
           "GL_ARB_gpu_shader_fp64",
@@ -2314,7 +2314,7 @@ static void InitExtensionList(GLContext* context)
     }
     case 330:
     {
-      std::string gl330exts[] = {
+      static const char* const gl330exts[] = {
           "GL_ARB_shader_bit_encoding",
           "GL_ARB_blend_func_extended",
           "GL_ARB_explicit_attrib_location",
@@ -2332,7 +2332,7 @@ static void InitExtensionList(GLContext* context)
     }
     case 320:
     {
-      std::string gl320exts[] = {
+      static const char* const gl320exts[] = {
           "GL_ARB_geometry_shader4",
           "GL_ARB_sync",
           "GL_ARB_vertex_array_bgra",
@@ -2350,7 +2350,7 @@ static void InitExtensionList(GLContext* context)
     case 310:
     {
       // Can't add NV_primitive_restart since function name changed
-      std::string gl310exts[] = {
+      static const char* const gl310exts[] = {
           "GL_ARB_draw_instanced",
           "GL_ARB_copy_buffer",
           "GL_ARB_texture_buffer_object",
@@ -2366,7 +2366,7 @@ static void InitExtensionList(GLContext* context)
     {
       // Quite a lot of these had their names changed when merged in to core
       // Disable the ones that have
-      std::string gl300exts[] = {
+      static const char* const gl300exts[] = {
           "GL_ARB_map_buffer_range",
           "GL_ARB_color_buffer_float",
           "GL_ARB_texture_float",
--- a/Source/Core/Core/ConfigLoaders/IsSettingSaveable.cpp
+++ b/Source/Core/Core/ConfigLoaders/IsSettingSaveable.cpp
@@ -22,134 +22,136 @@ bool IsSettingSaveable(const Config::ConfigLocation& config_location)
   if (config_location.system == Config::System::Logger)
     return true;
 
-  if (config_location.system == Config::System::Main && config_location.section == "NetPlay")
-    return true;
+  if (config_location.system == Config::System::Main)
+  {
+    if (config_location.section == "NetPlay" || config_location.section == "General")
+      return true;
+  }
 
-  if (config_location.system == Config::System::Main && config_location.section == "General")
-    return true;
-
-  const static std::vector<Config::ConfigLocation> s_setting_saveable{
+  static constexpr std::array<const Config::ConfigLocation*, 92> s_setting_saveable = {
       // Main.Core
 
-      Config::MAIN_DEFAULT_ISO.location,
-      Config::MAIN_MEMCARD_A_PATH.location,
-      Config::MAIN_MEMCARD_B_PATH.location,
-      Config::MAIN_AUTO_DISC_CHANGE.location,
-      Config::MAIN_DPL2_DECODER.location,
-      Config::MAIN_DPL2_QUALITY.location,
+      &Config::MAIN_DEFAULT_ISO.location,
+      &Config::MAIN_MEMCARD_A_PATH.location,
+      &Config::MAIN_MEMCARD_B_PATH.location,
+      &Config::MAIN_AUTO_DISC_CHANGE.location,
+      &Config::MAIN_DPL2_DECODER.location,
+      &Config::MAIN_DPL2_QUALITY.location,
 
       // Main.Display
-      Config::MAIN_FULLSCREEN_DISPLAY_RES.location,
-      Config::MAIN_FULLSCREEN.location,
-      Config::MAIN_RENDER_TO_MAIN.location,
-      Config::MAIN_RENDER_WINDOW_AUTOSIZE.location,
-      Config::MAIN_KEEP_WINDOW_ON_TOP.location,
-      Config::MAIN_DISABLE_SCREENSAVER.location,
+      &Config::MAIN_FULLSCREEN_DISPLAY_RES.location,
+      &Config::MAIN_FULLSCREEN.location,
+      &Config::MAIN_RENDER_TO_MAIN.location,
+      &Config::MAIN_RENDER_WINDOW_AUTOSIZE.location,
+      &Config::MAIN_KEEP_WINDOW_ON_TOP.location,
+      &Config::MAIN_DISABLE_SCREENSAVER.location,
 
       // Graphics.Hardware
 
-      Config::GFX_VSYNC.location,
-      Config::GFX_ADAPTER.location,
+      &Config::GFX_VSYNC.location,
+      &Config::GFX_ADAPTER.location,
 
       // Graphics.Settings
 
-      Config::GFX_WIDESCREEN_HACK.location,
-      Config::GFX_ASPECT_RATIO.location,
-      Config::GFX_CROP.location,
-      Config::GFX_SAFE_TEXTURE_CACHE_COLOR_SAMPLES.location,
-      Config::GFX_SHOW_FPS.location,
-      Config::GFX_SHOW_NETPLAY_PING.location,
-      Config::GFX_SHOW_NETPLAY_MESSAGES.location,
-      Config::GFX_LOG_RENDER_TIME_TO_FILE.location,
-      Config::GFX_OVERLAY_STATS.location,
-      Config::GFX_OVERLAY_PROJ_STATS.location,
-      Config::GFX_DUMP_TEXTURES.location,
-      Config::GFX_HIRES_TEXTURES.location,
-      Config::GFX_CACHE_HIRES_TEXTURES.location,
-      Config::GFX_DUMP_EFB_TARGET.location,
-      Config::GFX_DUMP_FRAMES_AS_IMAGES.location,
-      Config::GFX_FREE_LOOK.location,
-      Config::GFX_USE_FFV1.location,
-      Config::GFX_DUMP_FORMAT.location,
-      Config::GFX_DUMP_CODEC.location,
-      Config::GFX_DUMP_ENCODER.location,
-      Config::GFX_DUMP_PATH.location,
-      Config::GFX_BITRATE_KBPS.location,
-      Config::GFX_INTERNAL_RESOLUTION_FRAME_DUMPS.location,
-      Config::GFX_ENABLE_GPU_TEXTURE_DECODING.location,
-      Config::GFX_ENABLE_PIXEL_LIGHTING.location,
-      Config::GFX_FAST_DEPTH_CALC.location,
-      Config::GFX_MSAA.location,
-      Config::GFX_SSAA.location,
-      Config::GFX_EFB_SCALE.location,
-      Config::GFX_TEXFMT_OVERLAY_ENABLE.location,
-      Config::GFX_TEXFMT_OVERLAY_CENTER.location,
-      Config::GFX_ENABLE_WIREFRAME.location,
-      Config::GFX_DISABLE_FOG.location,
-      Config::GFX_BORDERLESS_FULLSCREEN.location,
-      Config::GFX_ENABLE_VALIDATION_LAYER.location,
-      Config::GFX_BACKEND_MULTITHREADING.location,
-      Config::GFX_COMMAND_BUFFER_EXECUTE_INTERVAL.location,
-      Config::GFX_SHADER_CACHE.location,
-      Config::GFX_WAIT_FOR_SHADERS_BEFORE_STARTING.location,
-      Config::GFX_SHADER_COMPILATION_MODE.location,
-      Config::GFX_SHADER_COMPILER_THREADS.location,
-      Config::GFX_SHADER_PRECOMPILER_THREADS.location,
-      Config::GFX_SAVE_TEXTURE_CACHE_TO_STATE.location,
+      &Config::GFX_WIDESCREEN_HACK.location,
+      &Config::GFX_ASPECT_RATIO.location,
+      &Config::GFX_CROP.location,
+      &Config::GFX_SAFE_TEXTURE_CACHE_COLOR_SAMPLES.location,
+      &Config::GFX_SHOW_FPS.location,
+      &Config::GFX_SHOW_NETPLAY_PING.location,
+      &Config::GFX_SHOW_NETPLAY_MESSAGES.location,
+      &Config::GFX_LOG_RENDER_TIME_TO_FILE.location,
+      &Config::GFX_OVERLAY_STATS.location,
+      &Config::GFX_OVERLAY_PROJ_STATS.location,
+      &Config::GFX_DUMP_TEXTURES.location,
+      &Config::GFX_HIRES_TEXTURES.location,
+      &Config::GFX_CACHE_HIRES_TEXTURES.location,
+      &Config::GFX_DUMP_EFB_TARGET.location,
+      &Config::GFX_DUMP_FRAMES_AS_IMAGES.location,
+      &Config::GFX_FREE_LOOK.location,
+      &Config::GFX_USE_FFV1.location,
+      &Config::GFX_DUMP_FORMAT.location,
+      &Config::GFX_DUMP_CODEC.location,
+      &Config::GFX_DUMP_ENCODER.location,
+      &Config::GFX_DUMP_PATH.location,
+      &Config::GFX_BITRATE_KBPS.location,
+      &Config::GFX_INTERNAL_RESOLUTION_FRAME_DUMPS.location,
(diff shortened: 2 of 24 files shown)


JosJuice:
@dolphin-emu-bot rebuild

Source/Core/VideoBackends/OGL/Render.cpp
```diff
@@ -144,53 +144,54 @@ static void APIENTRY ClearDepthf(GLfloat depthval)
 
 static void InitDriverInfo()
 {
-  std::string svendor = std::string(g_ogl_config.gl_vendor);
-  std::string srenderer = std::string(g_ogl_config.gl_renderer);
```
lioncash: These can be changed to std::string_view instances, eliminating the need for C-style string functions.

Source/Core/Core/ConfigLoaders/IsSettingSaveable.cpp
```diff
@@ -19,133 +19,135 @@ bool IsSettingSaveable(const Config::ConfigLocation& config_location)
...
       // UI.General
 
-      Config::MAIN_USE_DISCORD_PRESENCE.location,
+      &Config::MAIN_USE_DISCORD_PRESENCE.location,
 
   };
 
-  return std::find(s_setting_saveable.begin(), s_setting_saveable.end(), config_location) !=
-         s_setting_saveable.end();
+  for (auto iter = s_setting_saveable.cbegin(); iter != s_setting_saveable.cend(); ++iter) {
```
lioncash: This can collapsed into a std::any_of call.

GerbilSoft:
Applied all review changes. The C++ STL functionality does seem to add a few extra kilobytes compared to the C functions, but I can see how string_view is useful here.

Source/Core/VideoBackends/OGL/Render.cpp
```diff
@@ -144,54 +144,56 @@ static void APIENTRY ClearDepthf(GLfloat depthval)
...
     }
   }
-  else if (!strcmp(g_ogl_config.gl_vendor, "ATI Technologies Inc.") ||
-           !strcmp(g_ogl_config.gl_vendor, "Advanced Micro Devices, Inc."))
+  else if (svendor == "ATI Technologies Inc." || svendor == "Advanced Micro Devices, Inc.")
   {
     vendor = DriverDetails::VENDOR_ATI;
   }
-  else if (strstr(g_ogl_config.gl_version, "Mesa") != nullptr)
+  else if (std::string::npos != sversion.find("Mesa"))
```
lioncash: Our convention is to structure things like this as: `sversion.find("Mesa") != std::string::npos`. There's no benefit gained through the other method aside from looking slightly more obtuse to a reader. This is also more consistent with existing code. Ditto for other occurrences in the change.

GerbilSoft: The flipped version is how the code was originally before my changes. I'll unflip it anyway.

lioncash: Ah, OK, I didn't realize it was a revert. Thanks!

Source/Core/Core/Boot/Boot_BS2Emu.cpp
```diff
@@ -227,13 +227,22 @@ bool CBoot::EmulatedBS2_GC(const DiscIO::VolumeDisc& volume)
...
+      {"KOR", "NTSC", "KR", "LKH"}, // NTSC_K
+  };
+  const DiscIO::Region region = SConfig::GetInstance().m_region;
+  if (region < DiscIO::Region::NTSC_J || region >= DiscIO::Region::NTSC_K ||
+      region == DiscIO::Region::Unknown)
+  {
+    PanicAlertT("SetupWiiMemory: Invalid region value");
+    return false;
+  }
+  const RegionSetting& region_setting = region_settings[(int)region];
```
BhaaLseN:
```suggestion
  const RegionSetting& region_setting = region_settings[static_cast<int>(region)];
```

Source/Core/Core/Boot/Boot_BS2Emu.cpp
```diff
@@ -268,7 +277,10 @@ bool CBoot::SetupWiiMemory(IOS::HLE::IOSC::ConsoleType console_type)
     INFO_LOG(BOOT, "Using serial number: %s", serno.c_str());
   }
 
-  std::string model = "RVL-001(" + region_setting.area + ")";
+  std::string model = "RVL-001(";
+  model += region_setting.area;
+  model += ')';
```
BhaaLseN: (Why) Does this make a difference? I mean the splitting into multiple lines; not the use of `')'` (`char`) vs. `")"` (`char*`)

Source/Core/Core/ConfigLoaders/IsSettingSaveable.cpp
```diff
@@ -19,133 +19,132 @@ bool IsSettingSaveable(const Config::ConfigLocation& config_location)
   if (config_location.system == Config::System::Logger)
     return true;
 
-  if (config_location.system == Config::System::Main && config_location.section == "NetPlay")
-    return true;
-
-  if (config_location.system == Config::System::Main && config_location.section == "General")
-    return true;
+  if (config_location.system == Config::System::Main) {
```
BhaaLseN: That curly should be on the next line.

Source/Core/DiscIO/VolumeVerifier.cpp
```diff
@@ -344,38 +344,50 @@ bool VolumeVerifier::IsDebugSigned() const
...
 }
 
 bool VolumeVerifier::ShouldHaveInstallPartition() const
 {
-  const std::unordered_set<std::string> dragon_quest_x{"S4MJGD", "S4SJGD", "S6TJGD", "SDQJGD"};
-  return dragon_quest_x.find(m_volume.GetGameID()) != dragon_quest_x.end();
+  static constexpr std::array<std::string_view, 4> dqx = {
+      "S4MJGD", "S4SJGD", "S6TJGD", "SDQJGD"
+  };
+  const std::string& gameID = m_volume.GetGameID();
```
BhaaLseN: Our code style uses snake_case, so `game_id` please.

Source/Core/DiscIO/VolumeVerifier.cpp
```diff
@@ -344,38 +344,50 @@ bool VolumeVerifier::IsDebugSigned() const
...
 }
 
 bool VolumeVerifier::ShouldHaveMasterpiecePartitions() const
 {
-  const std::unordered_set<std::string> ssbb{"RSBE01", "RSBJ01", "RSBK01", "RSBP01"};
-  return ssbb.find(m_volume.GetGameID()) != ssbb.end();
+  static constexpr std::array<std::string_view, 4> ssbb = {
+      "RSBE01", "RSBJ01", "RSBK01", "RSBP01"
+  };
+  const std::string& gameID = m_volume.GetGameID();
```
BhaaLseN: Same down here.

Source/Core/Core/Boot/Boot_BS2Emu.cpp
JosJuice: You can't write `"RVL-001(" + region_setting.area`, because there is no `+` operator that takes two `const char*`. The way it's currently written, `"RVL-001("` gets converted to `std::string` by the first line, making it work.

JosJuice: With that in mind, if you want to write it on one line, you can write it as `std::string("RVL-001(") + region_setting.area + ')'` (or `"RVL-001(" + std::string(region_setting.area) + ')'`).

BhaaLseN: Oh, right, the type of `region_setting.area` also changed. Nvm this one then.

Source/Core/Common/GL/GLExtensions/GLExtensions.cpp
```diff
@@ -2198,7 +2198,7 @@ static void InitExtensionList(GLContext* context)
     default:
     case 450:
     {
-      std::string gl450exts[] = {
+      static const char *const gl450exts[] = {
```
CookiePLMonster: `s_extension_list` doesn't have a transparent comparator, so I think `s_extension_list[it]` still instantiates a `std::string` out of `it`. IMO you should definitely check if this is the case, as with a proper transparent comparator you could avoid constructing strings everywhere but when an entry is added to the map.

Source/Core/Core/Boot/Boot.h
```diff
@@ -26,10 +26,10 @@ class IOFile;
 
 struct RegionSetting
 {
-  const std::string area;
-  const std::string video;
-  const std::string game;
-  const std::string code;
+  char area[4];
```
CookiePLMonster: With how `CBoot::EmulatedBS2_GC` uses this structure, won't using `const char*` for those be "cleaner"? I'd probably be a bit paranoid about those fixed array sizes which are non-issue with a pointer type.

Source/Core/Core/ConfigLoaders/IsSettingSaveable.cpp
```diff
@@ -19,133 +19,133 @@ bool IsSettingSaveable(const Config::ConfigLocation& config_location)
...
 
-      Config::MAIN_USE_DISCORD_PRESENCE.location,
+      &Config::MAIN_USE_DISCORD_PRESENCE.location,
 
   };
 
-  return std::find(s_setting_saveable.begin(), s_setting_saveable.end(), config_location) !=
-         s_setting_saveable.end();
+  return std::any_of(s_setting_saveable.cbegin(), s_setting_saveable.cend(),
+    [&, config_location](const Config::ConfigLocation *location) { return *location == config_location; });
```
CookiePLMonster:
```suggestion
    [config_location](const Config::ConfigLocation *location) { return *location == config_location; });
```

Although I hope this passes a const reference and doesn't make a value type out of `config_location`.

Source/Core/DiscIO/VolumeVerifier.cpp
```diff
@@ -344,38 +344,47 @@ bool VolumeVerifier::IsDebugSigned() const
...
+      "RFNE01", "RFNJ01", "RFNK01", "RFNP01", "RFNW01",
+      "RFPE01", "RFPJ01", "RFPK01", "RFPP01", "RFPW01",
+      "RGWE41", "RGWJ41", "RGWP41", "RGWX41",
+      "RMCE01", "RMCJ01", "RMCK01", "RMCP01",
   };
 
-  return channel_discs.find(m_volume.GetGameID()) != channel_discs.end();
+  return std::binary_search(channel_discs.cbegin(), channel_discs.cend(),
+    m_volume.GetGameID().c_str(),
+    [](const std::string_view a, const std::string_view b) { return a < b; });
```
CookiePLMonster: Is this predicate really needed? Looks like something which is already defined by STL.

Source/Core/DiscIO/VolumeVerifier.cpp
```diff
@@ -344,38 +344,47 @@ bool VolumeVerifier::IsDebugSigned() const
...
 }
 
 bool VolumeVerifier::ShouldHaveInstallPartition() const
 {
-  const std::unordered_set<std::string> dragon_quest_x{"S4MJGD", "S4SJGD", "S6TJGD", "SDQJGD"};
-  return dragon_quest_x.find(m_volume.GetGameID()) != dragon_quest_x.end();
+  static constexpr std::array<std::string_view, 4> dqx = {"S4MJGD", "S4SJGD", "S6TJGD", "SDQJGD"};
+  const std::string& game_id = m_volume.GetGameID();
+  return std::any_of(dqx.cbegin(), dqx.cend(),
+                     [&, game_id](std::string_view x) { return x == game_id; });
```
CookiePLMonster:
```suggestion
                     [game_id](std::string_view x) { return x == game_id; });
```

Source/Core/DiscIO/VolumeVerifier.cpp
```diff
@@ -344,38 +344,47 @@ bool VolumeVerifier::IsDebugSigned() const
...
 }
 
 bool VolumeVerifier::ShouldHaveMasterpiecePartitions() const
 {
-  const std::unordered_set<std::string> ssbb{"RSBE01", "RSBJ01", "RSBK01", "RSBP01"};
-  return ssbb.find(m_volume.GetGameID()) != ssbb.end();
+  static constexpr std::array<std::string_view, 4> ssbb = {"RSBE01", "RSBJ01", "RSBK01", "RSBP01"};
+  const std::string& game_id = m_volume.GetGameID();
+  return std::any_of(ssbb.cbegin(), ssbb.cend(), [&, game_id](std::string_view x) { return x == game_id; });
```
CookiePLMonster:
```suggestion
  return std::any_of(ssbb.cbegin(), ssbb.cend(), [game_id](std::string_view x) { return x == game_id; });
```

Source/Core/DiscIO/VolumeVerifier.cpp
```diff
@@ -344,38 +344,47 @@ bool VolumeVerifier::IsDebugSigned() const
...
   // (unlike the other versions) and must not be added to this list.
-  const std::unordered_set<std::string> dual_layer_discs{
+  static constexpr std::array<std::string_view, 33> dual_layer_discs = {
       "R3ME01", "R3MP01", "R3OE01", "R3OJ01", "R3OP01", "RSBE01", "RSBJ01", "RSBK01", "RSBP01",
-      "RXMJ8P", "S59E01", "S59JC8", "S59P01", "S5QJC8", "SK8X52", "SAKENS", "SAKPNS", "SK8V52",
-      "SK8X52", "SLSEXJ", "SLSP01", "SQIE4Q", "SQIP4Q", "SQIY4Q", "SR5E41", "SR5P41", "SUOE41",
-      "SUOP41", "SVXX52", "SVXY52", "SX4E01", "SX4P01", "SZ3EGT", "SZ3PGT",
+      "RXMJ8P", "S59E01", "S59JC8", "S59P01", "S5QJC8", "SAKENS", "SAKPNS", "SK8V52", "SK8X52",
+      "SLSEXJ", "SLSP01", "SQIE4Q", "SQIP4Q", "SQIY4Q", "SR5E41", "SR5P41", "SUOE41", "SUOP41",
+      "SVXX52", "SVXY52", "SX4E01", "SX4P01", "SZ3EGT", "SZ3PGT",
```
CookiePLMonster: Since this had to be sorted to achieve proper functionality, how about a `std::is_sorted` assertion in Debug builds to ensure nobody breaks this in the future?

Source/Core/DiscIO/VolumeVerifier.cpp
```diff
@@ -344,38 +344,47 @@ bool VolumeVerifier::IsDebugSigned() const
...
   };
 
-  return dual_layer_discs.find(m_volume.GetGameID()) != dual_layer_discs.end();
+  return std::binary_search(
+      dual_layer_discs.cbegin(), dual_layer_discs.cend(), m_volume.GetGameID().c_str(),
+      [](const std::string_view a, const std::string_view b) { return a < b; });
```
CookiePLMonster: Is this predicate really needed? Looks like something which is already defined by STL.

Source/Core/InputCommon/InputConfig.cpp
```diff
@@ -58,10 +58,10 @@ bool InputConfig::LoadConfig(bool isGC)
 
     for (int i = 0; i < 4; i++)
     {
-      if (control_section->Exists(type + "Profile" + num[i]))
+      if (control_section->Exists(type + "Profile" + std::string(num[i])))
```
CookiePLMonster: Is this needed now that this type is `std::string_view`?

Source/Core/InputCommon/InputConfig.cpp
```diff
@@ -58,10 +58,10 @@ bool InputConfig::LoadConfig(bool isGC)
...
       {
         std::string profile_setting;
-        if (control_section->Get(type + "Profile" + num[i], &profile_setting))
+        if (control_section->Get(type + "Profile" + std::string(num[i]), &profile_setting))
```
CookiePLMonster: Is this needed now that this type is `std::string_view`?

Source/Core/Core/ConfigLoaders/IsSettingSaveable.cpp
GerbilSoft: I checked this with godbolt.org and if & isn't specified, it *does* get converted to value type.

Source/Core/DiscIO/VolumeVerifier.cpp
GerbilSoft: I checked and you're right, the predicate isn't needed; the default should work.

Source/Core/DiscIO/VolumeVerifier.cpp
GerbilSoft: Added locally; will be included in a cleaned-up rebase.

Source/Core/InputCommon/InputConfig.cpp
GerbilSoft: gcc-9.2.0 throws a ton of errors about no valid operator+() overload if this isn't there.

lioncash: operator+ was specifically excluded from handling conversions to std::string in the standard (rationale being that part of the purpose of std::string_view was to avoid allocations, and if avoiding allocations is desirable, then allocating conversions should be made explicitly).

CookiePLMonster:
Will this be rebased/history rewritten or is it good for another review?

GerbilSoft:
I did the rebase last night.

Source/Core/Core/ConfigLoaders/IsSettingSaveable.cpp
GerbilSoft: This is fixed in the rebase.

CookiePLMonster: This line doesn't show as outdated, though.

GerbilSoft: ...looks like I missed one of them. (I fixed the other ones.) Will fix this tonight.

CookiePLMonster:
Is this too much to ask to squash some of the changes? I noticed your later commits address concerns raised in earlier reviews, so PR ends up changing the same code multiple times.

CookiePLMonster:
@dolphin-emu-bot rebuild

Source/Core/DiscIO/VolumeVerifier.cpp
```diff
@@ -344,38 +345,47 @@ bool VolumeVerifier::IsDebugSigned() const
...
-  const std::unordered_set<std::string> channel_discs{
-      "RFNE01", "RFNJ01", "RFNK01", "RFNP01", "RFNW01", "RFPE01", "RFPJ01", "RFPK01", "RFPP01",
-      "RFPW01", "RGWE41", "RGWJ41", "RGWP41", "RGWX41", "RMCE01", "RMCJ01", "RMCK01", "RMCP01",
+  static constexpr std::array<std::string_view, 18> channel_discs = {
+      "RFNE01", "RFNJ01", "RFNK01", "RFNP01", "RFNW01",
+      "RFPE01", "RFPJ01", "RFPK01", "RFPP01", "RFPW01",
+      "RGWE41", "RGWJ41", "RGWP41", "RGWX41",
+      "RMCE01", "RMCJ01", "RMCK01", "RMCP01",
   };
+  assert(std::is_sorted(channel_discs.cbegin(), channel_discs.cend()));
```
BhaaLseN: Could this be a `static_assert` instead? According to the documentation, `std::is_sorted` is `constexpr` so it should work at compile time for that.

Source/Core/DiscIO/VolumeVerifier.cpp
```diff
@@ -344,38 +345,47 @@ bool VolumeVerifier::IsDebugSigned() const
...
-  return channel_discs.find(m_volume.GetGameID()) != channel_discs.end();
+  return std::binary_search(channel_discs.cbegin(), channel_discs.cend(),
+                            std::string_view(m_volume.GetGameID()));
 }
 
 bool VolumeVerifier::ShouldHaveInstallPartition() const
 {
-  const std::unordered_set<std::string> dragon_quest_x{"S4MJGD", "S4SJGD", "S6TJGD", "SDQJGD"};
-  return dragon_quest_x.find(m_volume.GetGameID()) != dragon_quest_x.end();
+  static constexpr std::array<std::string_view, 4> dqx = {"S4MJGD", "S4SJGD", "S6TJGD", "SDQJGD"};
```
BhaaLseN: With `ssbb` and `dual_layer_discs`, renaming this from `dragon_quest_x` to `dqx` tips the scale from "use the longer/readable name" to "use the abbreviation".

I tend to lean on the "prefer the readable one" side, but I wonder what everyone else thinks about this particular case.
Are we fine with the shorter names here (since they don't really add too much information for the two games, and one could just go and check our wiki for the game ids), or would we prefer the longer ones (mainly for consistency, to present a more uniform style across methods and locals; particularly in this file)?

Source/Core/DiscIO/VolumeVerifier.cpp
BhaaLseN: Same down here I suppose; can this be a `static_assert`?

Source/Core/DiscIO/VolumeVerifier.cpp
CookiePLMonster: It's constexpr only since C++20, do you think all supported compilers support this already?

JosJuice: If it's C++20, we shouldn't rely on it.

BhaaLseN: :(

Missed that, my bad.

Source/Core/DiscIO/VolumeVerifier.cpp
GerbilSoft: Changed it back to dragon_quest_x for now.

CookiePLMonster:
@dolphin-emu-bot rebuild

CookiePLMonster:
@dolphin-emu-bot rebuild

stenzek:
Could you please rebase on latest master?

CookiePLMonster:
Looks like the code doesn't work, is this due to rebase?

GerbilSoft:
Hi,

I got caught up with all sorts of stuff and forgot about this. Branch is rebased now.

GerbilSoft:
Removed the RegionSetting C string changes due to upstream commits that preserve region settings in setting.txt.

delroth:
Thank you for this PR! FYI in the future it helps a lot if you mark the resolved comments as resolved on GitHub, especially when reviews drag on through multiple rounds like this one.

PR #8363, opened 2019-09-15
Add Wii Mini system menu version detection

Fullmetal5:


JosJuice:
4608 is equal to `4096 & 512` (bitwise AND). What do you think about changing `switch (title_version & ~0xf)` to `switch (title_version & 0xff0)` instead of adding an extra case?

Fullmetal5:
Yeah, that seems fine however, do you think the mini's system menu should be distinguishable from regular one? Like M4.3U to indicate Mini or should it just be left alone.

JosJuice:
I think it's fine to leave it alone. But if we do want to have an indicator, I think adding something like `(Wii Mini)` at the end would be the best way to go. The meaning of just an M isn't super obvious.

CookiePLMonster:
Since effectively those three commits now change a single line only, I guess it makes sense to squash it and rewrite the commit message to be more accurate.

Fullmetal5:
Sorry about the wait, should be good now.

JosJuice:
@dolphin-emu-bot rebuild

PR #8151, opened 2019-05-31
Jit64/Jit64_Tables: Construct tables at compile-time

lioncash:
Utilizing constexpr, we can eliminate the need to construct the tables at runtime and just do all the work at compile-time. Making for less moving parts overall.

The general structure is more or less the same, however rather than one single initialization function, each table is built off an immediately executed lambda function. This is nice, since it narrows the scope of the table building logic down to the table that actually uses it.

PR #8364, opened 2019-09-18
PowerPC: use CPP to generate static instruction information and decoding tables

CrystalGamma:
This replaces all of Interpreter_Tables and most of PPCTables with ones that are built at compile-time and thus don't need initialization at runtime and so avoid mutable global variables.

What's left of Interpreter_Tables is used as a unit test for instruction decoding.

The new decoding process uses two tables (dispatch_table, which maps dispatch codes to either instruction forms (IForm) or subopcode fields, and tables_table, which describes how to map the value of a subopcode field to a dispatch code). Instructions using a primary opcode are decoded with a single random load.

After instruction form decoding, dispatch to interpreter or JIT functions or OpInfo lookup can be done using a flat (i. e. not multi-level) table, indexed by IForm. (For JIT code this has not been implemented for this PR but will be after it lands).

Since the new decoding process supports subopcode fields of non-10-bit size, the GekkoOPInfo entries for 'overflowing' variants of the integer arithmetic instructions have been merged with their normal versions, and the FL_SET_OE has been changed into FL_OE_BIT, accordingly.


lioncash:
Compile-time tables for Jit64 are also done in: https://github.com/dolphin-emu/dolphin/pull/8151 without the cpp processor (just MSVC needs to be upgraded on the build bots). This can easily be extended to the other tables as well. Is there a particular reason for using the preprocessor over a constexpr approach to the tables?

CrystalGamma:
Valid point. Using the preprocessor has the advantage that the number of supported instruction forms can be detected automatically (here via IForm::NUM_IFORMS) and thus lookup in a flat table (<256 entries currently) becomes possible, while the approach of #8151 leaves the two-level structure (with rigid 6-bit/10-bit opcode fields) in place which requires 64+5*1024-sized tables.

lioncash:
Sure, however the idea in 8151 was to convert the tables over first, and then introduce the structural changes in their own PR to reduce the amount of the overall diff, so I don't think it's fair to compare the two based on that.

CrystalGamma:
But changing the structure of the table would require not only rewriting the generation algorithm *again*, but also doing the same (or similar) changes I did here, wouldn't it?

CrystalGamma:
But I see your point … I'll just quickly push some changes that will hopefully satisfy MSVC's integer type checking and lint, and then I'll see how hard it is to make a constexpr function out of this generation logic.

CrystalGamma:
One thing that would be lost if I were to rewrite this as a (group of) constexpr function(s) is the ability to refer to instruction forms my name (e. g. IForm::rlwinmx) which comes naturally to generating the tables this way. That could be used in place of using raw opcode values in functions like JitArm64::fp_arith, once the JITs are changed to using IForms as well.

Tilka:
No need for this anymore, I think.

PR #8365, opened 2019-09-20
Update game settings for all PWAA games

teconmoon:
This pull request is for all 3 PWAA game's ini to fix visual issues with the games. This simply changes SafeTextureCacheColorSamples from 512 to 0 (Safe) to ensure the following:

- Resolve visual issues with evidence previews
- Proper scene transitions when unlocking a new case
- (1st game DLC only) Resolve display issues when testing areas with luminol. Previously, blood would not turn blue when sprayed with luminol until after you inspected it, and pressing - to switch between sides of the room would result in luminol results being overlayed on the wrong side of the room

### DIFF
--- a/Data/Sys/GameSettings/W2G.ini
+++ b/Data/Sys/GameSettings/W2G.ini
@@ -5,4 +5,4 @@
 [Video_Hacks]
 EFBEmulateFormatChanges = True
 [Video_Settings]
-SafeTextureCacheColorSamples = 512
+SafeTextureCacheColorSamples = 0
--- a/Data/Sys/GameSettings/W3G.ini
+++ b/Data/Sys/GameSettings/W3G.ini
@@ -5,4 +5,4 @@
 [Video_Hacks]
 EFBEmulateFormatChanges = True
 [Video_Settings]
-SafeTextureCacheColorSamples = 512
+SafeTextureCacheColorSamples = 0
--- a/Data/Sys/GameSettings/WGS.ini
+++ b/Data/Sys/GameSettings/WGS.ini
@@ -8,4 +8,4 @@
 [Video_Hacks]
 EFBEmulateFormatChanges = True
 [Video_Settings]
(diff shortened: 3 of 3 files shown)


teconmoon:
I've never done a GitHub pull request before so apologies if I haven't done this properly.

mimimi085181:
What's PWAA? Anyways, if the performance impact on those games is big enough to push some people below 100%, you could consider using the value 4096 instead. With 4096, Dolphin is hashing 8 times more data than with the medium(512) setting, but for bigger textures that's still far from hashing everything. Generally, you should only use a value that's a power of 2, so that the higher setting takes all samples that the lower setting takes and additional ones, otherwise you could run into textures that weren't an issue before, suddenly becoming an issue.

JosJuice:
PWAA is Phoenix Wright: Ace Attorney.

teconmoon:
Unfortunately, 4096 did not resolve the visual issues. Only 0 would.

All three games are ports of Nintendo DS released games, and are 2D games using absurdly small textures. Even on very low-spec machines, performance shouldn't suffer. Tested the entire 1st game on an Intel m3-7Y30 with Intel HD Graphics 615 and took no performance hits. (Not the best test machine but could be worse)

JosJuice:
LGTM if you squash the three commits into one. (This cannot be done with the GitHub web UI.)

teconmoon:
There's no Squash and merge option available for those who can potentially approve the pull request?

image

JosJuice:
We don't have it enabled for this repository. It doesn't create the same result as squashing the commits and then doing a regular merge.

BhaaLseN:
We don't use this workflow; we instead assume your history makes sense and try to preserve it.

You can [combine commits](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History#_squashing) if you run an interactive rebase locally on your clone.

PR #8366, opened 2019-09-21
Qt/RenderWidget: Account for devicePixelRatio when auto-adjusting window size

Techjar:
We want the game displayed at its actual resolution, not a scaled resolution. Fixes https://bugs.dolphin-emu.org/issues/11861

### DIFF
--- a/Source/Core/DolphinQt/RenderWidget.cpp
+++ b/Source/Core/DolphinQt/RenderWidget.cpp
@@ -51,7 +51,9 @@ RenderWidget::RenderWidget(QWidget* parent) : QWidget(parent)
     if (!Config::Get(Config::MAIN_RENDER_WINDOW_AUTOSIZE) || isFullScreen() || isMaximized())
       return;
 
-    resize(w, h);
+    const auto dpr = window()->windowHandle()->screen()->devicePixelRatio();
+
+    resize(w / dpr, h / dpr);
   });
 
   connect(&Settings::Instance(), &Settings::EmulationStateChanged, this, [this](Core::State state) {


BhaaLseN:
Do we handle moving the window between screens anywhere? If so, that place needs to be updated as well (assuming we declare PerMonitorDPI in our manifest)

Techjar:
There doesn't appear to be anything handling movement between screens.

MayImilae:
Tested! So this fixes it on the primary display, 1:1 exactly as it should be! However, if I move the window to another display or even open dolphin on another display, the ui is always scaled based on the primary display. So my primary display is 200% scaling, but my side monitors are 150% scaling. On the primary monitor 2x native is correctly set to 1280 pixels wide, but on the secondary displays, it is incorrectly 960 pixels wide. No matter what I did, changing multiple resolutions to auto-adjust multiple times, moving the render window between displays, or opening Dolphin on those screens, it was always 960 pixels wide.

Well, it's still an improvement, but, hmm.

Source/Core/DolphinQt/RenderWidget.cpp
```diff
@@ -51,7 +51,10 @@ RenderWidget::RenderWidget(QWidget* parent) : QWidget(parent)
     if (!Config::Get(Config::MAIN_RENDER_WINDOW_AUTOSIZE) || isFullScreen() || isMaximized())
       return;
 
-    resize(w, h);
+    const auto dpr =
+        QGuiApplication::screens()[QApplication::desktop()->screenNumber(this)]->devicePixelRatio();
```
spycrab: ``QDesktopWidget`` is deprecated. Please use ``window()->windowHandle()->screen()`` to obtain ``QScreen`` instead.

Techjar: Oh I totally missed your comment, sorry. I didn't know this was deprecated. I suppose we should fix the other usages of it too?

Techjar:
Pinging @MayImilae to check the behavior of this PR again, since it's now using a different function to get the `QScreen`.

Techjar:
We've verified that per-monitor DPI works now, so this should be good to merge! We some other odd bugs in testing but they aren't in any way related to this PR.

PR #8367, opened 2019-09-21
DiscIO: Add missing company IDs

JosJuice:


iwubcode:
Just curious but what Gust game is on Wii/GC?  I didn't realize there was one

Pokechu22:
@iwubcode https://wiki.dolphin-emu.org/index.php?title=Game_Sound_Station it looks like

JosJuice:
Yes, that wiki page is where I found that company ID.

Source/Core/DiscIO/Enums.cpp
```diff
@@ -555,48 +561,67 @@ const std::string& GetCompanyFromID(const std::string& company_id)
...
+      {"KY", "Medaverse Studios"},
       {"L3", "G-Mode"},
+      {"L8", "FujiSoft"},
       {"LG", "Black Bean Games"},
       {"LJ", "Legendo Entertainment"},
       {"LL", "HB Studios"},
       {"LP", "Left Field Productions"},
       {"LR", "Koch Media"},
       {"LT", "Legacy Interactive"},
+      {"LU", "Lexis Num\xc3\xa9rique"},  // We can't use a u8 prefix due to C++20's u8string
```
CookiePLMonster: What do you mean here?

JosJuice: It would be convenient to write this string as `u8"Lexis Numérique"`, but if I did that, compilation would fail when C++20 is supported since it would give us a `const char8_t[]` (instead of `const char[]`), which is not compatible with `std::string`.

CookiePLMonster: TIL! That's pretty nasty... Can `char` be converted to `char8_t` though? Sounds like it may be more future proof to migrate to u8string here (once applicable) and either let it convert implicitly or prefix them all with u8.

JosJuice: We would have to wait until all supported compilers have `u8string` support if we want to use `u8string`, which is going to take a while. And using `u8string` in this function doesn't really solve everything, since we have to convert it back to `string` at some point in the calling code anyway (barring a major move to `u8string` in large parts of Dolphin).

PR #8369, opened 2019-09-24
Add a git blame.ignoreRevsFile file

JosJuice:


BhaaLseN:
Thoughts on leaving a comment in the file itself that it needs to be added by hand before it works (using `git config blame.ignoreRevsFile` for example)?

Also, while we're at it; Did you by chance come across any support for `gitk` as well? I usually trace origins about as often (if not more) using `gitk` as I do with `git blame`.

Warepire:
> Also, while we're at it; Did you by chance come across any support for `gitk` as well? I usually trace origins about as often (if not more) using `gitk` as I do with `git blame`.

Additional question for the CLI people: Does `tig blame` handle it?

JosJuice:
> Thoughts on leaving a comment in the file itself that it needs to be added by hand before it works (using `git config blame.ignoreRevsFile` for example)?

Done.

> Also, while we're at it; Did you by chance come across any support for `gitk` as well? I usually trace origins about as often (if not more) using `gitk` as I do with `git blame`.

I haven't looked. (I usually use GitHub's web interface when I need to do a blame.)

Techjar:
Couldn't we make this apply by default by setting it in `.gitconfig` in the repository root?

.git-blame-ignore-revs
```diff
@@ -0,0 +1,11 @@
+# This file can be used for git blame using --ignore-revs-file
+# or by setting blame.ignoreRevsFile in the git cconfig.
```
Warepire:
```suggestion
# or by setting blame.ignoreRevsFile in the git config.
```

PR #8372, opened 2019-09-30
Vulkan: Exclusive fullscreen support via VK_EXT_full_screen_exclusive

stenzek:
This PR uses the new `VK_EXT_full_screen_exclusive` extension to provide application-level control of the display. You can check if the extension is being used by looking at the log (enable "Video Backend", set to info level) to see if there is a message about "exclusive fullscreen being used", and acquired/released. Scary diff line count is due to the headers update, the actual changes are quite small (see the last 3 commits).

Unfortunately, the extension is (for now) Windows-only, sorry Linux users.. I hope your compositor is smart enough to scan out Dolphin's buffers directly...

I know the NVIDIA driver "attempts" to use exclusive fullscreen when it can, so this may not make a difference on such systems. However, the extension is also supported by AMD, and I'm not sure if their driver performs this "optimization" implicitly.  It would be interesting to get some feedback from AMD users to see if it makes any difference there.

### DIFF
--- a/Source/Core/VideoBackends/Vulkan/CommandBufferManager.cpp
+++ b/Source/Core/VideoBackends/Vulkan/CommandBufferManager.cpp
@@ -371,13 +371,16 @@ void CommandBufferManager::SubmitCommandBuffer(u32 command_buffer_index,
                                      &present_image_index,
                                      nullptr};
 
-    res = vkQueuePresentKHR(g_vulkan_context->GetPresentQueue(), &present_info);
-    if (res != VK_SUCCESS)
+    m_last_present_result = vkQueuePresentKHR(g_vulkan_context->GetPresentQueue(), &present_info);
+    if (m_last_present_result != VK_SUCCESS)
     {
       // VK_ERROR_OUT_OF_DATE_KHR is not fatal, just means we need to recreate our swap chain.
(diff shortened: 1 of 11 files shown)


fantesykikachu:
I have an AMD card and after about 30 minutes of testing (most starting and stopping various games and alt-tabbing while in game) I didn't really notice any difference with this PR but then again I don't really notice any difference between the backends. (I did confirm that it does use VK_EXT_full_screen_exclusive)
It is also possible this would affect people with multi-monitor setups (with both AMD and Nvidea GPUs)

stenzek:
@fantesykikachu thanks for testing. I guess AMD's driver is also being smart and scanning the swap chain out directly then as well. Still, it doesn't hurt to be explicit, in case there's some overlay window which would otherwise prevent exclusive mode.

FWIW, with multi-monitor setups, I believe it should switch to exclusive mode on whichever monitor window has the Dolphin window in windowed mode prior to switching.

Mizoxman:
as an AMD user, imma just say that, with the current master, whether or not exclusive fullscreen "catches" on vulkan seems more or less random, sometimes catching correctly, other times just doing windowed fullscreen, DX11 on the other hand, successfully switches to exclusive fullscreen almost every time

PR #8373, opened 2019-09-30
DolphinQt/Graphics: Fix NVIDIA 3D Vision support handling.

jordan-woyak:
Previously when switching between backends that do/don't support "3D Vision" an attempt was made to remove the choice from the `QComboBox`.

This was problematic for a few reasons..
The wrong index was used (regression?) so the wrong item was removed.
Switching to an unsupported backend after having selected "3D Vision" activated the "Passive" stereo mode as it was the previous item in the list.
Adding additional items to the `StereoMode` enum after `Nvidia3DVision` (e.g. `OpenXR`) will have unmatched indices in the `QComboBox` which `GraphicsChoice` is not prepared for.

Now "NVIDIA 3D Vision" is always added to the stereo mode list. A popup message is shown when switching to an unsupported backend/3d-vision combo and stereoscopy is then disabled.

### DIFF
--- a/Source/Core/DolphinQt/Config/Graphics/EnhancementsWidget.cpp
+++ b/Source/Core/DolphinQt/Config/Graphics/EnhancementsWidget.cpp
@@ -20,6 +20,7 @@
 #include "DolphinQt/Config/Graphics/GraphicsSlider.h"
 #include "DolphinQt/Config/Graphics/GraphicsWindow.h"
 #include "DolphinQt/Config/Graphics/PostProcessingConfigWindow.h"
+#include "DolphinQt/QtUtils/ModalMessageBox.h"
 #include "DolphinQt/Settings.h"
 
 #include "UICommon/VideoUtils.h"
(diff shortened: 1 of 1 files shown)


mbc07:
What about always adding 3D Vision as the last option on the list? That way, it's guaranteed that by removing the last element of the list we're removing 3D Vision...

jordan-woyak:
@mbc07 Removing the item isn't the problem. And it can't be last in the list when I add OpenXR without breaking existing configs unless major changes are done to `GraphicsChoice`.

stenzek:
Will this allow a user to select 3D Vision even if it's unsupported? (I can see the check in LoadSettings, but not elsewhere)

To be honest, we could probably drop it completely. NVIDIA don't support it anymore, and I doubt there is many/any users.

jordan-woyak:
> 
> 
> Will this allow a user to select 3D Vision even if it's unsupported? (I can see the check in LoadSettings, but not elsewhere)
> 
> To be honest, we could probably drop it completely. NVIDIA don't support it anymore, and I doubt there is many/any users.

When 3D Vision is selected, with OpenGL for example, `LoadSettings` is invoked indirectly and the message is show + stereoscopy is disabled.

I have no objections to removing it entirely.

PR #8374, opened 2019-09-30
Rebase of 'Update game settings for all PWAA games'

AdmiralCurtiss:
See #8365, which was mysteriously closed, because I guess we intimidated them with the rebasing talk? Either way.


JMC47:
Seems fine to me.  If the game doesn't work without safe then I guess we can go with it.

PR #8376, opened 2019-10-01
WatchWidget: Fix integer comparison signedness warnings

tkln:
Fixes integer comparison signedness warnings.

### DIFF
--- a/Source/Core/DolphinQt/Debugger/WatchWidget.h
+++ b/Source/Core/DolphinQt/Debugger/WatchWidget.h
@@ -54,5 +54,5 @@ private:
 
   bool m_updating = false;
 
-  static constexpr size_t NUM_COLUMNS = 6;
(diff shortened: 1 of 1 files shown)


lioncash:
`NUM_COLUMNS` should likely be changed to an `int`. The change as is doesn't resolve signedness conversions, but rather moves them elsewhere. e.g. QTableWidget's `setItem` function expects an `int` for its row and column arguments, however this change now passes in an unsigned int.

CookiePLMonster:
Also I fail to see how this change fixes "enumeration value not handled in switch"? Seems like commit and PR refer to two different warnings.

tkln:
@CookiePLMonster True, that is for another commit which I somehow got mixed up with this PR.
@lioncash That's indeed a better solution. I've pushed a version that does just that.

CookiePLMonster:
@dolphin-emu-bot rebuild

PR #8377, opened 2019-10-02
Drop 3D Vision Support

stenzek:
This feature is no longer supported in NVIDIA drivers after major version 418. I doubt it had many users, but will leave this PR around for a little while in case anyone has any objections.

@jordan-woyak you mentioned in #8373 that you're adding another enum value for OpenXR. This could potentially cause config conflicts. However, if the number of 3D Vision users is basically zero, I don't think we need to worry too much about breaking configs, and keeps the config UI code simpler.


Source/Core/DolphinQt/HotkeyScheduler.cpp
```diff
@@ -496,21 +496,6 @@ void HotkeyScheduler::Run()
           Config::SetCurrent(Config::GFX_ENHANCE_POST_SHADER, "");
         }
       }
-
-      if (IsHotkey(HK_TOGGLE_STEREO_3DVISION))
```
CookiePLMonster: Doesn't it render this hotkey useless?

sonik-br:
Nvidia 3d vision still have a great community. They managed to get back the support for it on the driver and people has been making patches for the latest games to work with it.
I would love to have this feature back in Dolphin.

Please take a look at helixmod blog:
http://helixmod.blogspot.com/2017/05/3d-fix-manager.html

"Auto restores 3D Vision compatibility with Win 10 > 1809 and Geforce drivers beyond 425.31"

PR #8379, opened 2019-10-02
Vulkan: Optimizations for Mali with EFB2RAM on

stenzek:
Using the cached memory type appears to be slower on Mali drivers, with ~10-15% CPU spent in the __pi___inval_cache_range kernel function. This can improve the framerate by up to 30% in some cases.

I also noticed that the Adreno drivers place a coherent+cached heap before the coherent-only heap in the vulkan.gpuinfo.org reports, which the refactoring will handle. I've also set the "slow-cached-memory" bug for Adreno, if anyone has any high-end Adreno devices it'd be interesting to see if it makes any impact there as well.


JMC47:
I don't see much of a difference on my snapdragon.  It's hard to tell obviously but it seemed to be identical.  There was some flickering on transitions that don't happen on master, though.

Source/Core/Core/Config/GraphicsSettings.cpp
```diff
@@ -75,13 +75,15 @@ const ConfigInfo<bool> GFX_ENABLE_VALIDATION_LAYER{
 #if defined(ANDROID)
 const ConfigInfo<bool> GFX_BACKEND_MULTITHREADING{
     {System::GFX, "Settings", "BackendMultithreading"}, false};
+const ConfigInfo<int> GFX_COMMAND_BUFFER_EXECUTE_INTERVAL{
+    {System::GFX, "Settings", "CommandBufferExecuteInterval"}, 100};
```
weihuoya: default 100 on Android?

PR #8382, opened 2019-10-05
VerifyWidget: Handle Severity::None case in Verify()

tkln:
The case body is empty because VolumeVerifier doesn't actually report
problems with severity of None.

Fixes "warning: enumeration value ‘None’ not handled in switch [-Wswitch]"
warning reported by gcc.

### DIFF
--- a/Source/Core/DolphinQt/Config/VerifyWidget.cpp
+++ b/Source/Core/DolphinQt/Config/VerifyWidget.cpp
@@ -138,6 +138,8 @@ void VerifyWidget::Verify()
     case DiscIO::VolumeVerifier::Severity::High:
       severity = tr("High");
       break;
+    case DiscIO::VolumeVerifier::Severity::None:
+      break;
     }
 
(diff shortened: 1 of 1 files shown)


BhaaLseN:
I wonder if we should instead add a `default` case that does something like `severity = std::to_string(problem.severity)`; just to future-proof things a little...

CookiePLMonster:
I would go further and put an assertion/critical error there so if it ever happens, it explodes **loudly**.

tkln:
@CookiePLMonster I agree that it's often a good idea to have a default case where all unhandled values fall into and signal errors. However, in this case it seems a bit redundant since `problem.severity` is a strongly typed enum class which makes unhandled values detectable at compile time. I'm not sure if "future-proofing" like that is a good idea because a default case will mask those compiler warnings.

CookiePLMonster:
Indeed... can this warning be upgraded to an error **only** in this place maybe? I think that would be best, as then it would explode - but at compile time.

stenzek:
@dolphin-emu-bot rebuild

PR #8383, opened 2019-10-05
Externals: Eliminate the need to have the root "Externals" directory as an include path.

jordan-woyak:
The primary CMakeLists.txt made it look like `include_directories(Externals)` was just providing soundtouch headers but it's really pulling in way more.

I've moved a few "Externals" headers around so providing their includes isn't so horrible.


BhaaLseN:
Those now double `Externals/ed25519/ed25519`, `Externals/soundtouch/soundtouch` etc. look a bit weird; is this really necessary? I don't speak too much CMake, so I can't really tell...

iwubcode:
Being primarily a windows user, I can't speak to the CMake logic in Dolphin.  However, I use quite a bit of CMake in other places.  In "modern" CMake, `include_directories` in of itself is technically bad practice (since it is a global include).  I'd probably make all the externals individual targets that source in their include directories and drop the global include altogether.

As for this change, the only thing I don't personally care for is the way the discord include now works (from `#include <discord-rpc/include/discord_rpc.h>` to `#include <discord_rpc.h>`) but that's mostly down to personal preference.  Otherwise, this seems fine.

jordan-woyak:
> Those now double `Externals/ed25519/ed25519`, `Externals/soundtouch/soundtouch` etc. look a bit weird; is this really necessary? I don't speak too much CMake, so I can't really tell...

It looks weird to me too. I'm open to suggestions on better ways to do it. :P

> Being primarily a windows user, I can't speak to the CMake logic in Dolphin. However, I use quite a bit of CMake in other places. In "modern" CMake, `include_directories` in of itself is technically bad practice (since it is a global include). I'd probably make all the externals individual targets that source in their include directories and drop the global include altogether.
> 
> As for this change, the only thing I don't personally care for is the way the discord include now works (from `#include <discord-rpc/include/discord_rpc.h>` to `#include <discord_rpc.h>`) but that's mostly down to personal preference. Otherwise, this seems fine.

I don't think `discord-rpc/include/discord_rpc.h` is normal/correct. Other projects appear to use just `discord_rpc.h`.

iwubcode:
> I don't think discord-rpc/include/discord_rpc.h is normal/correct. Other projects appear to use just discord_rpc.h.

Yeah, I can't speak to the discord include itself.  I was more commenting on the style of include looking odd to me but that probably depends on the team.  Personally, I'm used to seeing `<libraryname>/<header>`.  However, if groups using discord use it like `#include <discord_rpc.h>` then I agree we should follow that trend.

PR #8385, opened 2019-10-06
DolphinQt: Add ability to override ResourcePack and Load directory

iwubcode:
If you have Dolphin installed on a SSD, you might not want to  have giant textures or giant zips installed in  your user directory.  Instead, you can delegate those folders to be off in another location.

Fixes:  https://bugs.dolphin-emu.org/issues/9970

### DIFF
--- a/Source/Core/DolphinQt/Settings/PathPane.cpp
+++ b/Source/Core/DolphinQt/Settings/PathPane.cpp
@@ -72,6 +72,17 @@ void PathPane::BrowseDump()
   }
 }
 
+void PathPane::BrowseLoad()
+{
+  QString dir = QDir::toNativeSeparators(QFileDialog::getExistingDirectory(
+      this, tr("Select Load Path"), QString::fromStdString(Config::Get(Config::MAIN_LOAD_PATH))));
+  if (!dir.isEmpty())
+  {
+    m_load_edit->setText(dir);
+    Config::SetBase(Config::MAIN_LOAD_PATH, dir.toStdString());
+  }
+}
+
 void PathPane::BrowseSDCard()
 {
   QString file = QDir::toNativeSeparators(QFileDialog::getOpenFileName(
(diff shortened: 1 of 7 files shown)


Source/Core/DolphinQt/Settings/PathPane.cpp
```diff
@@ -72,6 +72,29 @@ void PathPane::BrowseDump()
   }
 }
 
+void PathPane::BrowseLoad()
+{
+  QString dir = QDir::toNativeSeparators(QFileDialog::getExistingDirectory(
+      this, tr("Select Load Path"), QString::fromStdString(Config::Get(Config::MAIN_LOAD_PATH))));
```
BhaaLseN: "Load Path" seems a bit generic and nondescript to me (same goes for the setting names etc, btw), I don't suppose we have any more useful names for those?
"Resource pack path" on the other hand clearly tells me what it is.

iwubcode: Having a tooltip or something as an explanation would be ideal.  The "Load" path is for loading any data (today it is just textures but I already have another review that has different data).  I wasn't sure how to summarize it.  Maybe "LoadData"?

BhaaLseN: No idea. If I had a suggestion, I'd have included it already.

Source/Core/UICommon/UICommon.cpp
```diff
@@ -62,10 +62,25 @@ static void CreateDumpPath(const std::string& path)
   File::CreateFullPath(File::GetUserPath(D_DUMPTEXTURES_IDX));
 }
 
+static void CreateLoadPath(const std::string& path)
+{
+  if (!path.empty())
+    File::SetUserPath(D_LOAD_IDX, path + '/');
+  File::CreateFullPath(File::GetUserPath(D_HIRESTEXTURES_IDX));
```
Techjar: Why is this using 'Load/Textures/`?

iwubcode: This matches the way "DumpPath" does it.  Any subdirectories underneath get created when the root path changes

Techjar: Ah yeah I see it a few lines above. Makes sense.

Source/Core/UICommon/ResourcePack/ResourcePack.cpp
```diff
@@ -18,7 +18,7 @@
 
 namespace ResourcePack
 {
-constexpr char TEXTURE_PATH[] = "Load/Textures/";
+constexpr char TEXTURE_PATH[] = "Textures/";
```
Techjar: If we're following the mnemonic, this should be `HIRES_TEXTURES_DIR DIR_SEP`.

iwubcode: Makes sense.  Maybe I'll just remove this variable then?

Techjar: Well it is used in several different places, so I dunno.

iwubcode: Ah yeah, thought it was only used in two places, looks like three though.  I'll keep it

Techjar: Perhaps it would make more sense to drop it entirely here and move the `HIRES_TEXTURES_DIR` reference to ResourcePackManager.cpp? `path` only ever has that appended to it in `Install` and `Uninstall` anyways.

iwubcode: I'm changing that in #8360 so I went with the first option

Techjar: Ah, okay makes sense to keep it like this then.

Source/Core/UICommon/ResourcePack/ResourcePack.cpp
```diff
@@ -19,7 +20,7 @@
 
 namespace ResourcePack
 {
-constexpr char TEXTURE_PATH[] = "Load/Textures/";
+constexpr char TEXTURE_PATH[] = HIRES_TEXTURES_DIR DIR_SEP;
```
leoetlino: I'm sorry, I missed this change at first. Looks like this changes Load/Textures/ to Textures/, is that intended?

Techjar: Yes, because the Load directory is configurable.

leoetlino: Doh, you are right.

PR #8386, opened 2019-10-06
FixedSizeQueue: Work around GCC generating large amounts of debug info

stenzek:
The `storage = {};` statement in the `clear()` function causes GCC to blow up in memory usage when compiling LogWidget.cpp (approx. 16GB on my machine), and generate a 3.1GB assembly file in the `RelWithDebInfo` config. The `dolphin-emu` binary grows from ~300MB to ~900MB.

Not certain if this is a gcc bug or working as intended... I created a small repro program (albeit with a larger array size) which exhibits the same issue, and crashes gcc: https://pastebin.com/raw/iXrQEv0j

### DIFF
--- a/Source/Core/Common/FixedSizeQueue.h
+++ b/Source/Core/Common/FixedSizeQueue.h
@@ -21,7 +21,13 @@ public:
   void clear()
   {
     if constexpr (!std::is_trivial_v<T>)
-      storage = {};
+    {
+      // The clear of non-trivial objects previously used "storage = {}". However, this causes GCC
+      // to take a very long time to compile the file/function, as well as generating huge amounts
+      // of debug information (~2GB object file, ~600MB of debug info).
+      while (count > 0)
+        pop();
+    }
 
     head = 0;
     tail = 0;


Antidote:
Definitely a GCC bug, I'd report it upstream.

stenzek:
This is a regression from #8337, although as I said I'm not sure if the code is at fault here. Any thoughts, @CookiePLMonster?

CookiePLMonster:
I don't know enough about compilers but this indeed doesn't make any sense.

I know this is a lot to ask but can you pack and upload a compiled binary and debug info? I'd like to disassemble it.

stenzek:
> Edit: hold on with that, I'll first play around with compiler explorer. What is your exact GCC version?

gcc version 9.2.1 20190909 (Ubuntu 9.2.1-8ubuntu1)

Antidote:
I'm getting the same issue with `g++ (GCC) 9.1.0` on Arch

stenzek:
Looks like GCC is compiling the empty-initializer-list assignment into an unrolled loop initializing all elements, which generates a decent amount of code, but that's where the extra size in the debug info is coming from. clang appears to compile it as an actual loop...

CookiePLMonster:
If that is the case then I think it would be best to use `storage.fill` on this. Should hopefully not generate code as stupid.

CookiePLMonster:
Just dropping a link:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=59659

Source/Core/Common/FixedSizeQueue.h
```diff
@@ -21,7 +21,13 @@ class FixedSizeQueue
   void clear()
   {
     if constexpr (!std::is_trivial_v<T>)
-      storage = {};
+    {
+      // The clear of non-trivial objects previously used "storage = {}". However, this causes GCC
+      // to take a very long time to compile the file/function, as well as generating huge amounts
+      // of debug information (~2GB object file, ~600MB of debug info).
+      while (count > 0)
```
CookiePLMonster:
```suggestion
      storage.fill({});
```

I played around with Compiler Explorer and this is semantically equivalent to initializing with an empty initializer list and produces sane code.

jordan-woyak: Using `.fill` would be nicer if it does work around the bug.

stenzek: As I mentioned on IRC, filling would have the effect of always default constructing N (5000 in the log case) objects vs `count` objects. For the clear button this isn't a big deal but it might matter with other users of the class.

CookiePLMonster: In this case I'm cool with it.

PR #8389, opened 2019-10-06
Socket: Abort pending ops on close

sepalani:
It seems there was a network regression, I'm investigating to find the root cause of it.

Meanwhile, here is a quick fix which prevent closesocket to hang forever and instead return -1 on close.

Do not merge.

### DIFF
--- a/Source/Core/Core/IOS/Network/Socket.cpp
+++ b/Source/Core/Core/IOS/Network/Socket.cpp
@@ -170,6 +170,12 @@ s32 WiiSocket::CloseFd()
     ReturnValue = WiiSockMan::GetNetErrorCode(EITHER(WSAENOTSOCK, EBADF), "CloseFd", false);
   }
   fd = -1;
+
+  for (auto it = pending_sockops.begin(); it != pending_sockops.end();)
+  {
+    GetIOS()->EnqueueIPCReply(it->request, -SO_ENOTCONN);
+    it = pending_sockops.erase(it);
+  }
   return ReturnValue;
 }
 


sepalani:
Details regarding the issue can be found here: https://dolp.in/i11867/6
It seems to fix Ubisoft and Activision games having issues to go online.

I still need to write a hardware test to check what does the Wii since the error I returned is a blind guess. I also haven't tested for regressions, ATM.

sepalani:
According to this hardware test, the Wii returns SO_ENOTCONN when recvfrom is interrupted via close.
[wii-soread-close.zip](https://github.com/dolphin-emu/dolphin/files/4512008/wii-soread-close.zip)

Source/Core/Core/IOS/Network/Socket.cpp
```diff
@@ -170,6 +170,12 @@ s32 WiiSocket::CloseFd()
     ReturnValue = WiiSockMan::GetNetErrorCode(EITHER(WSAENOTSOCK, EBADF), "CloseFd", false);
   }
   fd = -1;
+  auto it = pending_sockops.begin();
+  while (it != pending_sockops.end())
```
leoetlino, line 174:
```cpp
for (auto it = pending_sockops.begin(); it != pending_sockops.end(); )
```

(I wouldn't move `it = pending_sockops.erase(it);` though, as it's less readable IMO)

JMC47:
With Leoetlino's approval, I'm willing to merge this once you're done with it.  Just ping me.

PR #8391, opened 2019-10-06
Qt/CMake: Fix Qt5_DIR being set to the wrong directory

JackBister:
Hello, I just checked out the repo for the first time, did `git submodule update --init` and tried to build using CMake on Windows (I use CLion), but got an error because Qt5_DIR was not pointing to the folder that was added by `git submodule update`. This commit fixes that so other Windows users who don't use Visual Studio shouldn't have the same problem in the future.


spycrab:
This fix has been included in #8401.

PR #8392, opened 2019-10-06
DolphinQt: When audio backend is invalid, show nothing in dropdown

JosJuice:
It used to show the first option, No Audio Output (but audio output would work correctly anyway since AudioCommon didn't use this logic.)

### DIFF
--- a/Source/Core/DolphinQt/Settings/AudioPane.cpp
+++ b/Source/Core/DolphinQt/Settings/AudioPane.cpp
@@ -181,12 +181,18 @@ void AudioPane::LoadSettings()
 
   // Backend
   const auto current = SConfig::GetInstance().sBackend;
+  bool selection_set = false;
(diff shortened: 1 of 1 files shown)


Source/Core/DolphinQt/Settings/AudioPane.cpp
```diff
@@ -181,12 +181,18 @@ void AudioPane::LoadSettings()
 
   // Backend
   const auto current = SConfig::GetInstance().sBackend;
+  bool selection_set = false;
```
Techjar: Can we not just set the index to -1 here and be done with it?

JosJuice: No, because the act of adding an item (or the act of adding the first item?) sets it back to 0.

Techjar: Oh I missed the `addItem` in the loop.

PR #8393, opened 2019-10-06
Support Windows 10 long paths

CookiePLMonster:
This PR makes Dolphin compatible with Windows 10 long paths. Windows 10 1607 and newer allows applications to lift MAX_PATH limitation via manifest, so with a few code changes I was able to remove MAX_PATH dependency from all non-Qt code paths and opt in to that.

I verified against games and settings located in a path which has over MAX_PATH characters - game showed in the list and booted, settings also seem to save.

The only function still using MAX_PATH (and not being a third party) is `CreateTempDir` but as far as I can tell that is only used in unit tests, so I didn't bother.

### DIFF
--- a/Source/Core/Common/CDUtils.cpp
+++ b/Source/Core/Common/CDUtils.cpp
@@ -26,6 +26,7 @@
 #include <IOKit/storage/IOMedia.h>
 #include <paths.h>
 #else
+#include <climits>
 #include <fcntl.h>
 #include <sys/ioctl.h>
 #include <sys/stat.h>
@@ -211,7 +212,7 @@ bool IsCDROMDevice(std::string device)
 #ifndef _WIN32
   // Resolve symbolic links. This allows symbolic links to valid
   // drives to be passed from the command line with the -e flag.
-  char resolved_path[MAX_PATH];
+  char resolved_path[PATH_MAX];
   char* devname = realpath(device.c_str(), resolved_path);
   if (!devname)
     return false;
--- a/Source/Core/Common/CommonFuncs.cpp
+++ b/Source/Core/Common/CommonFuncs.cpp
@@ -49,4 +49,27 @@ std::string GetLastErrorString()
                  MAKELANGID(LANG_NEUTRAL, SUBLANG_DEFAULT), error_message, BUFFER_SIZE, nullptr);
   return std::string(error_message);
 }
+
+// Obtains a full path to the specified module.
+std::optional<std::wstring> GetModuleName(void* hInstance)
+{
+  DWORD max_size = 50;  // Start with space for 50 characters and grow if needed
+  std::wstring name(max_size, L'\0');
+
+  DWORD size;
+  while ((size = GetModuleFileNameW(static_cast<HMODULE>(hInstance), name.data(), max_size)) ==
+             max_size &&
+         GetLastError() == ERROR_INSUFFICIENT_BUFFER)
+  {
+    max_size *= 2;
+    name.resize(max_size);
+  }
+
+  if (size == 0)
+  {
+    return std::nullopt;
+  }
+  name.resize(size);
+  return name;
+}
 #endif
--- a/Source/Core/Common/CompatPatches.cpp
+++ b/Source/Core/Common/CompatPatches.cpp
@@ -4,12 +4,14 @@
 
 #include <Windows.h>
 #include <functional>
+#include <optional>
 #include <string>
 #include <vector>
 #include <winternl.h>
 
 #include <fmt/format.h>
 
+#include "Common/CommonFuncs.h"
 #include "Common/CommonTypes.h"
 #include "Common/LdrWatcher.h"
 #include "Common/StringUtil.h"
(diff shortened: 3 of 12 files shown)


Source/Core/Common/CDUtils.cpp
```diff
@@ -27,6 +27,7 @@
 #include <paths.h>
 #else
 #include <fcntl.h>
+#include <limits.h>
```
lioncash: `<climits>`

Source/Core/Common/CommonFuncs.cpp
```diff
@@ -49,4 +49,28 @@ std::string GetLastErrorString()
                  MAKELANGID(LANG_NEUTRAL, SUBLANG_DEFAULT), error_message, BUFFER_SIZE, nullptr);
   return std::string(error_message);
 }
+
+// Obtains a full path to the specified module.
+std::optional<std::wstring> GetModuleName(void* hInstance)
+{
+  std::wstring name;
+  DWORD max_size = 50;  // Start with space for 50 characters and grow if needed
+  name.resize(max_size);
```
lioncash:
```cpp
std::wstring name(max_size, L'\0')
```

Source/Core/Common/FileUtil.cpp
```diff
@@ -639,19 +639,21 @@ std::string CreateTempDir()
...
 
-std::string GetTempFilenameForAtomicWrite(const std::string& path)
+std::string GetTempFilenameForAtomicWrite(std::string path)
 {
-  std::string abs = path;
 #ifdef _WIN32
-  TCHAR absbuf[MAX_PATH];
-  if (_tfullpath(absbuf, UTF8ToTStr(path).c_str(), MAX_PATH) != nullptr)
-    abs = TStrToUTF8(absbuf);
+  TCHAR* absbuf = _tfullpath(nullptr, UTF8ToTStr(path).c_str(), 0);
```
lioncash:
```suggestion
  std::unique_ptr<TCHAR*, decltype(&std::free)> absbuf{_tfullpath(nullptr, UTF8ToTStr(path).c_str(), 0), std::free};
```

Source/Core/Core/PowerPC/Jit64/JitAsm.cpp
```diff
@@ -16,6 +16,8 @@
 #include "Core/PowerPC/Jit64Common/Jit64PowerPCState.h"
 #include "Core/PowerPC/PowerPC.h"
 
+#include <limits.h>
```
lioncash:
```suggestion
#include <climits>
```

Also our current convention is to place standard includes after the corresponding header include for the cpp file (aka, the first one).

Source/Core/Common/FileUtil.cpp
```diff
@@ -672,22 +674,26 @@ std::string GetBundleDirectory()
...
-    GetModuleFileName(nullptr, dolphin_exe_path, ARRAYSIZE(dolphin_exe_path));
-    if (_tfullpath(dolphin_exe_expanded_path, dolphin_exe_path,
-                   ARRAYSIZE(dolphin_exe_expanded_path)) != nullptr)
-      dolphin_path = TStrToUTF8(dolphin_exe_expanded_path);
-    else
-      dolphin_path = TStrToUTF8(dolphin_exe_path);
+    auto dolphin_exe_path = GetModuleName(nullptr);
+    if (dolphin_exe_path)
+    {
+      TCHAR* dolphin_exe_expanded_path = _tfullpath(nullptr, dolphin_exe_path->c_str(), 0);
```
lioncash:
```
std::unique_ptr<TCHAR, decltype(&std::free)> dolphin_exe_expanded_path{_tfullpath(nullptr, dolphin_exe_path->c_str(), 0), std::free};
```

Source/Core/Common/CompatPatches.cpp
```diff
@@ -162,37 +164,26 @@ struct Version
...
-                                static_cast<DWORD>(path->size()));
-  if (!path_len)
-    return false;
-  auto error = GetLastError();
-  if (error == ERROR_SUCCESS)
-    return true;
-  if (error == ERROR_INSUFFICIENT_BUFFER)
-    goto retry;
-  return false;
+    return {};
```
lioncash:
```suggestion
    return std::nullopt;
```

Default initialization and the constructor taking std::nullopt_t are [not](https://godbolt.org/z/vOIZmE) behaviorally equivalent

CookiePLMonster: This page seems to say otherwise
https://en.cppreference.com/w/cpp/utility/optional/optional
as both `optional()` and `optional( std::nullopt_t )` are linked to the same description.

lioncash: They provide the same basic observable behavior, but the reality of the situation is that they currently, absolutely, do not behave the same way in how they accomplish that visible behavior (*especially* with [large](https://godbolt.org/z/tBkOGO) objects)

Please use std::nullopt in this case.

CookiePLMonster: I will, I was just curious about real life examples like the one you provided. This is a very good example towards it indeed.

BParks21:
I think this broke the auto updater. Before this the last build where the auto updater worked was 5.0-11165

CookiePLMonster:
Fixed by #8469.

PR #6034, opened 2017-09-07
Fix some Android UI crashes opening 'settings' menus

JonnyH:

PR #8394, opened 2019-10-08
Various DI improvements

Pokechu22:
An early draft of my improved DI work.  I still need to do some testing (and write automated hardware tests), but there [is documentation](https://wiibrew.org/wiki//dev/di) of what I've figured out so far.

The GPIO change fixes [bug 11818](https://bugs.dolphin-emu.org/issues/11818) (not DI related) and also implements GPIO-based ejection (which is used by the system menu when booting from the eject button in conjunction with the RTC flag; I don't think any games use it though.)

The RTC change should fix [bug 8115](https://bugs.dolphin-emu.org/issues/8115) (including the case when a channel other than the system menu is running which I reported as [bug 11803](https://bugs.dolphin-emu.org/issues/8115)).  Note that the trade-off is that currently, it *always* acts like the disc has changed and *never* uses the cache when first loading the system menu, though it should still use it when exiting from a game.  To be more accurate about it dolphin would need to track the current disc in the configuration file, probably.

Most of the other changes are related to errors reported by the disc drive, or various commands that were not implemented before.  I've made some attempts to organize these nicely into different commits, but it's still not the clearest yet.

I have only tested this with Wii games.  Testing it with gamecube games would be appreciated (in particular, I suspect that the stricter DTK changes will cause issues when starting from emulated BS2, but I haven't tested it).

Although this is a draft, I would like feedback on the implementation, since some of the code is rather hacky and I'm not completely sure how to do it better.

### DIFF
--- a/Source/Core/Core/Boot/Boot.cpp
+++ b/Source/Core/Core/Boot/Boot.cpp
@@ -244,6 +244,17 @@ bool CBoot::DVDRead(const DiscIO::VolumeDisc& disc, u64 dvd_offset, u32 output_a
   return true;
 }
 
+bool CBoot::DVDReadDiscID(const DiscIO::VolumeDisc& disc, u32 output_address)
+{
+  std::array<u8, 0x20> buffer;
+  if (!disc.Read(0, buffer.size(), buffer.data(), DiscIO::PARTITION_NONE))
+    return false;
+  Memory::CopyToEmu(output_address, buffer.data(), buffer.size());
+  // Clear ERROR_NO_DISKID_L, probably should check if that's currently set
+  DVDInterface::SetLowError(DVDInterface::ERROR_READY);
+  return true;
+}
+
 void CBoot::UpdateDebugger_MapLoaded()
 {
   Host_NotifyMapLoaded();
--- a/Source/Core/Core/Boot/Boot_BS2Emu.cpp
+++ b/Source/Core/Core/Boot/Boot_BS2Emu.cpp
@@ -22,6 +22,7 @@
 #include "Core/HW/DVD/DVDInterface.h"
 #include "Core/HW/EXI/EXI_DeviceIPL.h"
 #include "Core/HW/Memmap.h"
+#include "Core/IOS/DI/DI.h"
 #include "Core/IOS/ES/ES.h"
 #include "Core/IOS/ES/Formats.h"
 #include "Core/IOS/FS/FileSystem.h"
@@ -420,6 +421,12 @@ bool CBoot::EmulatedBS2_Wii(const DiscIO::VolumeDisc& volume)
   const auto console_type = volume.GetTicket(data_partition).GetConsoleType();
   if (!SetupWiiMemory(console_type) || !IOS::HLE::GetIOS()->BootIOS(tmd.GetIOSId()))
     return false;
+
+  auto di = std::static_pointer_cast<IOS::HLE::Device::DI>(
+      IOS::HLE::GetIOS()->GetDeviceByName("/dev/di"));
+
+  di->InitializeIfFirstTime();
+  di->ChangePartition(data_partition);
 
   DVDReadDiscID(volume, 0x00000000);
 
--- a/Source/Core/Core/Boot/Boot_BS2Emu.cpp
+++ b/Source/Core/Core/Boot/Boot_BS2Emu.cpp
@@ -19,6 +19,7 @@
 #include "Core/ConfigManager.h"
 #include "Core/Core.h"
 #include "Core/HLE/HLE.h"
+#include "Core/HW/DVD/DVDInterface.h"
 #include "Core/HW/EXI/EXI_DeviceIPL.h"
 #include "Core/HW/Memmap.h"
 #include "Core/IOS/ES/ES.h"
@@ -389,6 +390,9 @@ bool CBoot::EmulatedBS2_Wii(const DiscIO::VolumeDisc& volume)
     state->type = 0xff;
     state->discstate = 0x01;
   });
+
+  // The system menu clears the RTC flags
+  ExpansionInterface::g_rtc_flags.m_hex = 0;
 
   // While reading a disc, the system menu reads the first partition table
   // (0x20 bytes from 0x00040020) and stores a pointer to the data partition entry.
--- a/Source/Core/Core/Boot/Boot_BS2Emu.cpp
+++ b/Source/Core/Core/Boot/Boot_BS2Emu.cpp
@@ -210,7 +210,7 @@ bool CBoot::EmulatedBS2_GC(const DiscIO::VolumeDisc& volume)
 
   SetupGCMemory();
 
-  DVDRead(volume, /*offset*/ 0x00000000, /*address*/ 0x00000000, 0x20, DiscIO::PARTITION_NONE);
+  DVDReadDiscID(volume, 0x00000000);
 
   const bool ntsc = DiscIO::IsNTSC(SConfig::GetInstance().m_region);
 
@@ -406,7 +406,7 @@ bool CBoot::EmulatedBS2_Wii(const DiscIO::VolumeDisc& volume)
   if (!SetupWiiMemory(console_type) || !IOS::HLE::GetIOS()->BootIOS(tmd.GetIOSId()))
     return false;
 
-  DVDRead(volume, 0x00000000, 0x00000000, 0x20, DiscIO::PARTITION_NONE);  // Game Code
+  DVDReadDiscID(volume, 0x00000000);
 
   // This is some kind of consistency check that is compared to the 0x00
   // values as the game boots. This location keeps the 4 byte ID for as long
--- a/Source/Core/Core/Boot/Boot_BS2Emu.cpp
+++ b/Source/Core/Core/Boot/Boot_BS2Emu.cpp
@@ -212,6 +212,21 @@ bool CBoot::EmulatedBS2_GC(const DiscIO::VolumeDisc& volume)
 
   DVDReadDiscID(volume, 0x00000000);
 
+  bool streaming = Memory::Read_U8(0x80000008);
+  if (streaming)
+  {
+    u8 streaming_size = Memory::Read_U8(0x80000009);
+    // If the streaming buffer size is 0, then BS2 uses a default size of 10 instead.
+    // No known game uses a size other than the default.
+    if (streaming_size == 0)
+      streaming_size = 10;
+    DVDInterface::AudioBufferConfig(true, streaming_size);
+  }
+  else
+  {
+    DVDInterface::AudioBufferConfig(false, 0);
+  }
+
   const bool ntsc = DiscIO::IsNTSC(SConfig::GetInstance().m_region);
 
   // Setup pointers like real BS2 does
--- a/Source/Core/Core/HW/DVD/DVDInterface.cpp
+++ b/Source/Core/Core/HW/DVD/DVDInterface.cpp
@@ -187,7 +187,6 @@ void SetLidOpen();
 void UpdateInterrupts();
 void GenerateDIInterrupt(DIInterruptType _DVDInterrupt);
 
-void WriteImmediate(u32 value, u32 output_address, bool reply_to_ios);
 bool ExecuteReadCommand(u64 dvd_offset, u32 output_address, u32 dvd_length, u32 output_length,
                         const DiscIO::Partition& partition, ReplyType reply_type,
                         DIInterruptType* interrupt_type);
@@ -319,7 +318,7 @@ static void DTKStreamingCallback(const std::vector<u8>& audio_data, s64 cycles_l
   else
   {
     // There's nothing to read, so using DVDThread is unnecessary.
-    u64 userdata = PackFinishExecutingCommandUserdata(ReplyType::DTK, DIInterruptType::INT_TCINT);
+    u64 userdata = PackFinishExecutingCommandUserdata(ReplyType::DTK, DIInterruptType::TCINT);
     CoreTiming::ScheduleEvent(ticks_to_dtk, s_finish_executing_command, userdata);
   }
 }
@@ -340,7 +339,7 @@ void Init()
   s_finish_executing_command =
       CoreTiming::RegisterEvent("FinishExecutingCommand", FinishExecutingCommandCallback);
 
-  u64 userdata = PackFinishExecutingCommandUserdata(ReplyType::DTK, DIInterruptType::INT_TCINT);
+  u64 userdata = PackFinishExecutingCommandUserdata(ReplyType::DTK, DIInterruptType::TCINT);
   CoreTiming::ScheduleEvent(0, s_finish_executing_command, userdata);
 }
 
@@ -500,7 +499,7 @@ void SetLidOpen()
   u32 old_value = s_DICVR.CVR;
   s_DICVR.CVR = IsDiscInside() ? 0 : 1;
   if (s_DICVR.CVR != old_value)
-    GenerateDIInterrupt(INT_CVRINT);
+    GenerateDIInterrupt(DIInterruptType::CVRINT);
 }
 
 bool UpdateRunningGameMetadata(std::optional<u64> title_id)
@@ -576,8 +575,7 @@ void RegisterMMIO(MMIO::Mapping* mmio, u32 base)
                    s_DICR.Hex = val & 7;
                    if (s_DICR.TSTART)
                    {
-                     ExecuteCommand(s_DICMDBUF[0], s_DICMDBUF[1], s_DICMDBUF[2], s_DIMAR,
-                                    s_DILENGTH, false);
+                     ExecuteCommand(ReplyType::Interrupt);
                    }
                  }));
 
@@ -605,29 +603,59 @@ void GenerateDIInterrupt(DIInterruptType dvd_interrupt)
 {
   switch (dvd_interrupt)
   {
-  case INT_DEINT:
-    s_DISR.DEINT = 1;
+  case DIInterruptType::DEINT:
+    s_DISR.DEINT = true;
     break;
-  case INT_TCINT:
-    s_DISR.TCINT = 1;
+  case DIInterruptType::TCINT:
+    s_DISR.TCINT = true;
     break;
-  case INT_BRKINT:
-    s_DISR.BRKINT = 1;
+  case DIInterruptType::BRKINT:
+    s_DISR.BRKINT = true;
     break;
-  case INT_CVRINT:
-    s_DICVR.CVRINT = 1;
+  case DIInterruptType::CVRINT:
+    s_DICVR.CVRINT = true;
     break;
   }
 
   UpdateInterrupts();
 }
 
-void WriteImmediate(u32 value, u32 output_address, bool reply_to_ios)
+void SetInterruptEnabled(DIInterruptType interrupt, bool enabled)
 {
-  if (reply_to_ios)
-    Memory::Write_U32(value, output_address);
-  else
-    s_DIIMMBUF = value;
+  switch (interrupt)
+  {
+  case DIInterruptType::DEINT:
+    s_DISR.DEINTMASK = enabled;
+    break;
+  case DIInterruptType::TCINT:
+    s_DISR.TCINTMASK = enabled;
+    break;
+  case DIInterruptType::BRKINT:
+    s_DISR.BRKINTMASK = enabled;
+    break;
+  case DIInterruptType::CVRINT:
+    s_DICVR.CVRINTMASK = enabled;
+    break;
+  }
+}
+
+void ClearInterrupt(DIInterruptType interrupt)
+{
+  switch (interrupt)
+  {
+  case DIInterruptType::DEINT:
+    s_DISR.DEINT = false;
+    break;
+  case DIInterruptType::TCINT:
+    s_DISR.TCINT = false;
+    break;
+  case DIInterruptType::BRKINT:
+    s_DISR.BRKINT = false;
+    break;
+  case DIInterruptType::CVRINT:
+    s_DICVR.CVRINT = false;
+    break;
+  }
 }
 
 // Iff false is returned, ScheduleEvent must be used to finish executing the command
@@ -639,13 +667,13 @@ bool ExecuteReadCommand(u64 dvd_offset, u32 output_address, u32 dvd_length, u32 
   {
     // Disc read fails
     SetHighError(ERROR_NO_DISK_H);
-    *interrupt_type = INT_DEINT;
+    *interrupt_type = DIInterruptType::DEINT;
     return false;
   }
   else
   {
     // Disc read succeeds
-    *interrupt_type = INT_TCINT;
+    *interrupt_type = DIInterruptType::TCINT;
   }
 
   if (dvd_length > output_length)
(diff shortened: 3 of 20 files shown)


JMC47:
@dolphin-emu-bot rebuild

CookiePLMonster:
@dolphin-emu-bot rebuild

Source/Core/Core/IOS/DI/DI.cpp
```diff
@@ -30,70 +33,513 @@ void DI::DoState(PointerWrap& p)
...
+  case DVDLowUnencryptedRead:
+  {
+    u32 length = Memory::Read_U32(request.buffer_in + 4);
+    u32 position = Memory::Read_U32(request.buffer_in + 8);
+    u32 end = position + (length >> 2);  // a 32-bit offset
+    INFO_LOG(IOS_DI, "DVDLowUnencryptedRead: offset 0x%08x (byte 0x%09" PRIx64 "), length 0x%x",
+             position, (u64)position << 2, length);
+    // (start, end) as 32-bit offsets
+    // Older IOS versions only accept the first range.  Later versions added the extra ranges to
+    // check how the drive responds to out of bounds reads (an error 001 check).
```
JosJuice: Do you know which versions? You could add this to GetFeatures() in IOS/VersionInfo.cpp if you want Dolphin to handle this the way a given version of IOS would.

Source/Core/Core/HW/DVD/DVDInterface.cpp
```diff
@@ -805,51 +685,69 @@ void ExecuteCommand(u32 command_0, u32 command_1, u32 command_2, u32 output_addr
...
+    default:
+      ERROR_LOG(DVDINTERFACE, "Unknown 0xAD subcommand in %08x", command_0);
+      break;
+    }
+    break;
+  // Probably only used by Wii
+  case READ_DVD:
     ERROR_LOG(DVDINTERFACE, "DVDLowReadDvd");
     break;
   // Probably only used by Wii
```
JosJuice: You can remove these comments about GC vs Wii and direct access vs IOS access. I added them as a compromise for the code not showing which ones only are usable via direct access or IOS on a real console, but now the code does show that. (See PR #1634 for reference.)

Source/Core/Core/HW/DVD/DVDInterface.cpp
```diff
@@ -314,6 +319,8 @@ void DoState(PointerWrap& p)
 
   p.Do(s_disc_path_to_insert);
 
+  p.Do(s_rtc_flag);
```
JosJuice: Since you are making changes to the savestate system, you need to update the version number in State.cpp.

Source/Core/Core/HW/DVD/DVDInterface.cpp
```diff
@@ -358,6 +367,16 @@ void SetDisc(std::unique_ptr<DiscIO::VolumeDisc> disc,
 
   DVDThread::SetDisc(std::move(disc));
   SetLidOpen();
+  if (!has_disc)
+    // ERROR_COVER is used when the cover is open;
+    // ERROR_NO_DISK_L is used when the cover is closed but there is no disc.
+    // On the Wii, this can only happen if something other than a DVD is inserted into the disc
+    // drive (for instance, an audio CD) and only after it attempts to read it.  Otherwise, it will
+    // report the cover as opened.
+    SetLowError(ERROR_COVER);
```
JosJuice: Use braces here (and below as well), since there's more than one line.

Source/Core/Core/HW/DVD/DVDThread.cpp
```diff
@@ -342,20 +342,34 @@ static void FinishRead(u64 id, s64 cycles_late)
...
+                  request.dvd_offset, request.dvd_offset + request.length);
+    }
+    else
+    {
+      // Part of the error 001 check.
+      INFO_LOG(DVDINTERFACE, "Ignoring out of bounds test read (at 0x%" PRIx64 " - 0x%" PRIx64 ")",
+               request.dvd_offset, request.dvd_offset + request.length);
+    }
+
+    DVDInterface::SetHighError(DVDInterface::ERROR_BLOCK_OOB);
```
JosJuice: That we have reached this place in the code doesn't necessarily mean that we attempted to read out of bounds. We can also reach this under conditions such as having the game stored on a flash drive and yanking the flash drive out in the middle of gameplay. If you want to check whether we read out of bounds, you probably want to make use of this:

https://github.com/dolphin-emu/dolphin/blob/19ed64122c11f7cd36ff92b61e7af7bb772c01a8/Source/Core/DiscIO/Volume.h#L131

But keep in mind that it usually returns a too large value for WBFS and CISO as a result of those formats not containing size information. (The current approach has the same problem, though. Reads won't fail due to out of bounds when reading within the bounds of what `GetSize()` returns.)

Source/Core/Core/HW/DVD/DVDInterface.h
```diff
@@ -97,12 +116,144 @@ bool UpdateRunningGameMetadata(std::optional<u64> title_id = {});
...
+  struct
+  {
+    u32 CONFIG : 8;
+    u32 : 24;
+  };
+  UDICFG() { Hex = 0; }
+  UDICFG(u32 _hex) { Hex = _hex; }
+};
+
+extern UDISR s_DISR;
```
lioncash: These should probably not all be exposed, but rather have an interface built around them.

Source/Core/Core/HW/WII_IPC.h
```diff
@@ -35,6 +35,36 @@ enum StarletInterruptCause
...
+  DEBUG1 = 0x20000,
+  DEBUG2 = 0x40000,
+  DEBUG3 = 0x80000,
+  DEBUG4 = 0x100000,
+  DEBUG5 = 0x200000,
+  DEBUG6 = 0x400000,
+  DEBUG7 = 0x800000,
+};
+
+extern u32 gpio_out;
```
lioncash: Ditto

Source/Core/Core/HW/WII_IPC.h
```diff
@@ -35,6 +35,36 @@ enum StarletInterruptCause
   INT_CAUSE_IPC_STARLET = 0x80000000
 };
 
+enum GPIO : u32
```
lioncash: enum class and provide the necessary overloads.

Source/Core/Core/IOS/DI/DI.cpp
```diff
@@ -30,93 +34,618 @@ void DI::DoState(PointerWrap& p)
...
+    std::array<std::pair<u32, u32>, 3> valid_ranges = {
```
lioncash: constexpr

Source/Core/Core/IOS/DI/DI.cpp
```diff
@@ -30,93 +34,618 @@ void DI::DoState(PointerWrap& p)
...
+      {
+        return StartDMATransfer(0xA8000000, position, length, length, request);
+      }
+    }
+    WARN_LOG(IOS_DI, "DVDLowUnencryptedRead: trying to read from an illegal region!");
+    return std::make_optional(SECURITY_ERROR);
+  }
+  case DVDLowReportKey:
+  {
+    u8 param1 = Memory::Read_U8(request.buffer_in + 7);
```
lioncash: These can be const. Etc below as well.

Source/Core/Core/IOS/DI/DI.cpp
```diff
@@ -30,93 +34,618 @@ void DI::DoState(PointerWrap& p)
...
+    result = DRIVE_ERROR;
+    break;
+  default:
+    PanicAlert("IOS::HLE::Device::DI: Unexpected DVDInterface interrupt %d!", interrupt_type);
+    result = DRIVE_ERROR;
+    break;
+  }
+
+  auto di = IOS::HLE::GetIOS()->GetDeviceByName("/dev/di");
+  if (di)
```
lioncash: Both statement bodies should be braced here, since one travels more than one line.

Source/Core/Core/IOS/DI/DI.cpp
```diff
@@ -30,93 +34,618 @@ void DI::DoState(PointerWrap& p)
...
+    m_has_initialized = true;
+  }
+
+  if (request.in_vectors[0].size != 0x20)
+  {
+    ERROR_LOG(IOS_DI, "IOCtlV: Received bad input buffer size 0x%02x, should be 0x20",
+              request.in_vectors[0].size);
+    return GetDefaultReply(BAD_ARGUMENT);
+  }
+  u8 command = Memory::Read_U8(request.in_vectors[0].address);
```
lioncash:
```suggestion
  const u8 command = Memory::Read_U8(request.in_vectors[0].address);
```

Source/Core/Core/IOS/DI/DI.cpp
```diff
@@ -30,93 +34,618 @@ void DI::DoState(PointerWrap& p)
...
   }
   return GetDefaultReply(return_value);
 }
+
+void DI::ChangePartition(const DiscIO::Partition& partition)
+{
+  m_current_partition = partition;
+}
+
+const DiscIO::Partition& DI::GetCurrentPartition()
```
lioncash:
```suggestion
DiscIO::Partition DI::GetCurrentPartition()
```

A partition simply consists of a u64 value. It's fine to return by value.

Source/Core/Core/IOS/DI/DI.h
```diff
@@ -25,16 +31,108 @@ class DI : public Device
...
+
+  struct DICommand
+  {
+    DIIoctl command;
+    u8 padding[3];
+    std::array<u32, 7> args;
+  };
+  static_assert(sizeof(DICommand) == 0x20);
+
+  enum DIResult : s32
```
lioncash: enum class

Source/Core/Core/IOS/DI/DI.h
```diff
@@ -25,16 +31,108 @@ class DI : public Device
...
+  static const DiscIO::Partition& GetCurrentPartition();
+
   void DoState(PointerWrap& p) override;
 
+  IPCCommandResult Open(const OpenRequest& request) override;
   IPCCommandResult IOCtl(const IOCtlRequest& request) override;
   IPCCommandResult IOCtlV(const IOCtlVRequest& request) override;
 
-  void FinishIOCtl(DVDInterface::DIInterruptType interrupt_type);
+  enum DIIoctl : u8
```
lioncash: Ditto

Source/Core/Core/HW/DVD/DVDThread.cpp
```diff
@@ -342,20 +342,34 @@ static void FinishRead(u64 id, s64 cycles_late)
             (CoreTiming::GetTicks() - request.time_started_ticks) /
                 (SystemTimers::GetTicksPerSecond() / 1000000));
 
+  DVDInterface::DIInterruptType interrupt;
   if (buffer.size() != request.length)
   {
-    PanicAlertT("The disc could not be read (at 0x%" PRIx64 " - 0x%" PRIx64 ").",
-                request.dvd_offset, request.dvd_offset + request.length);
+    if (request.dvd_offset != 0x118280000 && request.dvd_offset != 0x1FB500000)
```
BhaaLseN: I wonder if we should invert this `if` for sanity (if check, else real-error). I'm assuming you found those values from RE of the related IOS?

Source/Core/Core/IOS/DI/DI.cpp
Pokechu22: I haven't fully tested all versions (I focused on the version dated Jun 3 2009 07:49:09 and included in IOS58 and IOS80), but here's what I wrote on WiiBrew:

> Versions of IOS prior to IOS30 only permitted reads in the first range[check]. Nintendo titles check on startup if they are running an IOS ≥ 30 (and ≤ 253) and if so, perform some DI checks; specifically, they attempt to read 0x20 bytes from 0x460a0000 (or from 0x7ed40000 if the byte at 0x8000319c is 0x81 — possibly related to dual-layer discs?). If this read attempt returns anything other than 2, the game will refuse to start with the message "An error has occurred. Press the Eject Button, remove the Game Disc, and turn off the power to the console. Please read the Wii Operations Manual for further instructions." If the drive error is anything other than 0x0052100 (OK/Logical block address out of range), the game will refuse to start with the message "Error #001, unauthorized device has been detected."

Source/Core/Core/HW/WiimoteEmu/Camera.cpp
```diff
@@ -105,6 +106,12 @@ void CameraLogic::Update(const Common::Matrix44& transform)
   std::array<CameraPoint, leds.size()> camera_points;
 
   std::transform(leds.begin(), leds.end(), camera_points.begin(), [&](const Vec3& v) {
+    if (!(IOS::gpio_out & IOS::GPIO::SENSOR_BAR))
```
jordan-woyak: You could skip the std::transform entirely if the sensor bar is off. Just `camera_points.fill(INVISIBLE_POINT)`.

Pokechu22: Done.  I did it this way initially because I didn't want a massive diff from the indentation change when I was initially experiencing, but that doesn't really matter.

Source/Core/Core/HW/WII_IPC.h
Pokechu22: Is the thing I implemented what you wanted?  I'm not *too* experienced with C++, so I don't know if the thing I made in dd5aa9b9ffa1df07e40d68c44e4dba9c4dce71f4 is what you want.

Source/Core/Core/HW/WII_IPC.h
```diff
@@ -35,6 +35,71 @@ enum StarletInterruptCause
...
+  DEBUG5 = 0x200000,
+  DEBUG6 = 0x400000,
+  DEBUG7 = 0x800000,
+};
+
+class GPIOBit
+{
+public:
+  GPIOBit(u32& bits, GPIO bit) : m_bits(bits), m_bit(bit) {}
+  operator bool const() { return m_bits & (u32)m_bit; }
```
lioncash:
```suggestion
  explicit operator bool const() { return m_bits & static_cast<u32>(m_bit); }
```

Source/Core/Core/HW/WII_IPC.h
```diff
@@ -35,6 +35,71 @@ enum StarletInterruptCause
...
+  GPIOBit& operator=(bool const rhs)
```
lioncash:
```suggestion
  GPIOBit& operator=(const bool rhs)
```

Source/Core/Core/HW/WII_IPC.h
```diff
@@ -35,6 +35,71 @@ enum StarletInterruptCause
...
+  {
+    if (rhs)
+      m_bits |= (u32)m_bit;
```
lioncash: static_cast, ditto below

Source/Core/Core/HW/WII_IPC.h
```diff
@@ -35,6 +35,71 @@ enum StarletInterruptCause
...
+
+class GPIOFlags
+{
+public:
+  constexpr GPIOFlags() {}
+  constexpr GPIOFlags(std::initializer_list<GPIO> bits)
+  {
+    for (auto bit : bits)
+    {
+      m_hex |= (u32)bit;
```
lioncash: static_cast

Source/Core/Core/HW/WII_IPC.h
```diff
@@ -35,6 +35,71 @@ enum StarletInterruptCause
...
+
+private:
+  u32& m_bits;
+  GPIO m_bit;
+};
+
+class GPIOFlags
+{
+public:
+  constexpr GPIOFlags() {}
```
lioncash:
```suggestion
  constexpr GPIOFlags() = default;
```

Source/Core/Core/HW/WII_IPC.h
Pokechu22: I'm not sure I understand what `explicit` is doing here; I would have assumed that that requires me to use `static_cast<bool>(IOS::gpio_out[IOS::GPIO::SENSOR_BAR])` but it doesn't.  What situation is it preventing, then?

CookiePLMonster:
@dolphin-emu-bot rebuild

CookiePLMonster:
@dolphin-emu-bot rebuild

Source/Core/Core/HW/WII_IPC.h
lioncash: It prevents implicit conversions to other types in various scenarios. For [example](https://godbolt.org/z/HEL-so).

Source/Core/Core/HW/WII_IPC.h
```diff
@@ -35,6 +36,71 @@ enum StarletInterruptCause
...
+  DEBUG5 = 0x200000,
+  DEBUG6 = 0x400000,
+  DEBUG7 = 0x800000,
+};
+
+class GPIOBit
+{
+public:
+  GPIOBit(u32& bits, GPIO bit) : m_bits(bits), m_bit(bit) {}
+  explicit operator bool() { return (m_bits & static_cast<u32>(m_bit)) != 0; }
```
lioncash: I missed this in my initial review, but this should likely also have a const specifier like so:

```suggestion
  explicit operator bool() const { return (m_bits & static_cast<u32>(m_bit)) != 0; }
```

Pokechu22: Ah, that's where that should go; I had `explicit operator bool const()` but GCC seemed to dislike it because `const bool` doesn't really mean much for a return type.

Source/Core/Core/HW/DVD/DVDThread.cpp
Pokechu22: Yeah, the values are found in IOS (in [DVDLowUnencryptedRead](https://wiibrew.org/wiki//dev/di#0x8D_DVDLowUnencryptedRead)), and also in various games.  I also have the same values here:

https://github.com/dolphin-emu/dolphin/blob/069e13f2ea96b24e49cda80feb4651ebd3f3fc19/Source/Core/Core/IOS/DI/DI.cpp#L250-L275

and they exist in current dolphin here (but are rejected in DVDInterface instead of making it to DVDThread):

https://github.com/dolphin-emu/dolphin/blob/d39555919d39357b49a7eb02f2d96b9a71bf128b/Source/Core/Core/HW/DVD/DVDInterface.cpp#L845-L873

I feel the real error case is more important since once a game starts, it's the only one that'll happen.  But it probably could still be swapped (especially if other error cases are added).

Source/Core/Core/Boot/Boot.cpp
```diff
@@ -244,6 +244,17 @@ bool CBoot::DVDRead(const DiscIO::VolumeDisc& disc, u64 dvd_offset, u32 output_a
   return true;
 }
 
+bool CBoot::DVDReadDiscID(const DiscIO::VolumeDisc& disc, u32 output_address)
+{
+  std::array<u8, 0x20> buffer;
+  if (!disc.Read(0, 0x20, buffer.data(), DiscIO::PARTITION_NONE))
+    return false;
+  Memory::CopyToEmu(output_address, buffer.data(), 0x20);
```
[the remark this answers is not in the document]
Pokechu22: Simply laziness &mdash; ideally, `DVDReadDiscID` wouldn't exist at all and it would instead just directly call code in DVDInterface (which would also manage the drive state change), but that doesn't work well with DVDInterface being async.  I'll change it to `buffer.size()` for now though.

Source/Core/Core/HW/DVD/DVDInterface.cpp
```diff
@@ -517,9 +530,13 @@ static void InsertDiscCallback(u64 userdata, s64 cyclesLate)
 }
 
 // Must only be called on the CPU thread
-void EjectDisc()
+// button_pressed should be true if this is user-triggered, and false if it is an eject from
+// emulated software.
+void EjectDisc(bool button_pressed)
```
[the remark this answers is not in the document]
Pokechu22: Makes sense.  I'm using this enum:

```c++
enum class EjectCause
{
  User,
  Software
};
```

Source/Core/Core/HW/DVD/DVDInterface.cpp
```diff
@@ -473,8 +355,8 @@ void Shutdown()
 void SetDisc(std::unique_ptr<DiscIO::VolumeDisc> disc,
              std::optional<std::vector<std::string>> auto_disc_change_paths = {})
 {
-  if (disc)
-    s_current_partition = disc->GetGamePartition();
+  bool had_disc = IsDiscInside();
+  bool has_disc = (bool)disc;
```
BhaaLseN: Thoughts on using `bool has_disc = static_cast<bool>(disc)` or the infamous cast-to-bool operator `bool has_disc = !!disc` instead of the C-style cast?

Source/Core/Core/HW/DVD/DVDInterface.cpp
```diff
@@ -930,70 +730,126 @@ void ExecuteCommand(u32 command_0, u32 command_1, u32 command_2, u32 output_addr
...
+      ERROR_LOG(DVDINTERFACE, "Attempted to play audio while audio is disabled!");
+      SetHighError(ERROR_AUDIO_BUF);
+      interrupt_type = INT_DEINT;
+      break;
     }
-    else
+
+    switch ((command_0 >> 16) & 0xFF)
     {
+    case 0x00:
```
BhaaLseN: Do we want to _name_ this case (as well as `0x01` further down)? Or at least leave a comment.
All the other cases either have an `enum` (or `constexpr`) name, a comment next to them; or just an `ERROR_LOG` that makes it immediately obvious what it means.
Main reason is that Dolphin might be the only place out there that _knows_ what the values mean, and serve as documentation for others.

Pokechu22:
I plan on addressing most of the TODO comments, but not all of them.  I probably will not address the DTK buffer size comment in this PR (since I don't think it affects games), and I don't plan on addressing the BCA TODO either (I want to handle that in a different PR).  I also don't plan on handling the GPIO `AVE` and `SLOT_LED` comment, since those are pretty far out of scope (I don't know how I'd indicate the slot LED, and I don't know anything about AVE at all).

I might handle the extra read in DVDLowReadDiskID (I've ordered a gamecube game that uses DTK, so I can do some testing regarding it).  I might handle the DVDLowGetCoverStatus resetting case, but that requires some hardware testing to understand what's going on with DICVR during the reset (I've done some basic testing, and it looks weird) and I don't think it affects games.  DVDLowSeek is a weird case since dolphin doesn't implement it currently, but I do plan on implementing it eventually (which will also necessitate doing something about the partition).  The two disc state TODOs are basically just cases where I need to do more testing and I plan on resolving them at some point.

The comment in `PerformDecryptingRead` is copied from `ExecuteReadCommand`.  That will require timing testing which I haven't done yet; I might or might not get to it.

The rest are code things that I plan on resolving before merge (or possibly have already resolved and haven't removed the comments on).

Source/Core/Core/IOS/DI/DI.cpp
```diff
@@ -30,93 +34,617 @@ void DI::DoState(PointerWrap& p)
...
+std::optional<DI::DIResult> DI::StartIOCtl(const IOCtlRequest& request)
 {
-  const u32 command_0 = Memory::Read_U32(request.buffer_in);
-  const u32 command_1 = Memory::Read_U32(request.buffer_in + 4);
-  const u32 command_2 = Memory::Read_U32(request.buffer_in + 8);
+  if (request.buffer_in_size != 0x20)
+  {
+    ERROR_LOG(IOS_DI, "IOCtl: Received bad input buffer size 0x%02x, should be 0x20",
+              request.buffer_in_size);
+    return std::make_optional(DIResult::SECURITY_ERROR);
```
lioncash: As far as I'm aware, you can simply return the enum value directly without the need for `std::make_optional`, Ditto elsewhere where applicable.

Source/Core/Core/HW/DVD/DVDInterface.cpp
Pokechu22: I find the `!!` thing somewhat silly looking, so I'd probably use static_cast instead.  I probably could also do just an implicit conversion (`bool has_disc = disc`) but that seems like it'd also be confusing.

Source/Core/Core/HW/DVD/DVDInterface.cpp
Pokechu22: I'll split the earlier log message into two messages for the cases.  It should be split anyways because the parameters are ignored in case 0x01 (stopping the stream).

Source/Core/Core/IOS/DI/DI.cpp
Pokechu22: Visual studio didn't seem to like that before, but it looks like it works now.  Possibly it didn't work before due to me using a regular `enum` instead of an `enum class`?

Pokechu22:
The gamecube games I ordered arrived, and I can now see that the DTK changes I have here are a bit too restrictive &mdash; DTK isn't configured when launching with emulated BS2, and also changing discs while on the IPL rejects DTK configuration for the second disc which causes issues even if neither game uses DTK (however, everything works fine if you start the IPL and then insert a disc or boot with "skip main menu" disabled).

I know how to fix the BS2 case (just add that logic into Boot_BS2Emu, which probably will involve messily duplicating more code for the time being, which is not great but at least a solution).  However,  I'm less sure how to solve the IPL one &mdash; I'm not sure when the drive gets back into a state that audio can be configured.  I'm currently enabling DTK configuration when the drive is reset, but I don't think that's something that happens normally on a gamecube.  Maybe it's instead configurable after changing discs, and that's something that's enforced in the drive itself without software needing to do a reset?

Pokechu22:
I've fixed both of the DTK not getting set issues, though I still haven't hardware tested it becoming configurable again.  I haven't fixed the crash, but it looks like [gamecube games just react poorly in general](https://bugs.dolphin-emu.org/issues/9019) to being ejected.  I'm not sure why that's happening as there should be an interrupt that's getting sent out, but I don't know enough about gamecube interrupts to diagnose it.

Source/Core/Core/Boot/Boot_BS2Emu.cpp
```diff
@@ -209,7 +211,20 @@ bool CBoot::EmulatedBS2_GC(const DiscIO::VolumeDisc& volume)
...
 
-  DVDRead(volume, /*offset*/ 0x00000000, /*address*/ 0x00000000, 0x20, DiscIO::PARTITION_NONE);
+  DVDReadDiscID(volume, 0x00000000);
+
+  bool streaming = Memory::Read_U8(0x80000008);
+  if (streaming)
+  {
+    u8 streaming_size = Memory::Read_U8(0x80000009);
+    if (streaming_size == 0)
+      streaming_size = 10;
```
BhaaLseN: What's so special about zero (and ten, for that matter)?

Source/Core/Core/HW/DVD/DVDInterface.cpp
```diff
@@ -485,8 +367,29 @@ void SetDisc(std::unique_ptr<DiscIO::VolumeDisc> disc,
...
 
+  // Assume that inserting a disc requires having an empty disc before
+  if (had_disc != has_disc)
+    s_rtc_flag |= RTC_FLAG_DISC_CHANGED;
+
   DVDThread::SetDisc(std::move(disc));
   SetLidOpen();
+
+  s_can_configure_dtk =
+      true;  // TODO: Hardware test needed to confirm that DTK can be configured again
```
BhaaLseN: This looks terrible, would you mind putting the comment in the line _before_ the statement?

Source/Core/Core/IOS/DI/DI.cpp
```diff
@@ -30,93 +34,617 @@ void DI::DoState(PointerWrap& p)
...
+  p.Do(m_last_length);
+}
+
+IPCCommandResult DI::Open(const OpenRequest& request)
+{
+  if (!m_has_initialized)
+  {
+    ResetDIRegisters();
+    m_has_initialized = true;
+  }
```
BhaaLseN: This exact snippet appears at least 3 times (from just scrolling through) and should likely be its own function.

Source/Core/Core/HW/DVD/DVDInterface.cpp
Pokechu22: ... jeeze, I didn't see that.  I typed it on one line, and then autoformatted and didn't notice it changed to that monstrosity.

Source/Core/Core/Boot/Boot_BS2Emu.cpp
Pokechu22: BS2 itself makes that check.  If byte 9 (streaming size) on the disc is set to 0, it treats the size as 10; otherwise, it uses that size.  I documented it on the [Wii Disc](https://wiibrew.org/wiki/Wii_Disc) and it was also already documented on the [BS2 page](https://wiibrew.org/wiki/System_Menu/BS2) for state 11, albeit not very explicitly:

> if not waiting for a disk, exit to state 50
> 
> if waiting, disable the waiting-for-disk and check streaming [0x80000008]
> 
> if streaming and no streaming buffer size defined [0x80000009], set to size 10
> 
> if streaming, setup via __DVDAudioBufferConfig
> 
> go to state 12 

To my understanding, no disc actually specifies a buffer size other than 0, so it probably doesn't come up in practice.  I'll add a comment explaining it.

It's worth noting that this check happens in both Wii and GC BS2, and on the Wii it only happens for Gamecube discs (to my understanding, Dolphin doesn't use the emulated wii bs2 when running GC discs).

There's some additional that prints a message out in the Wii Menu (separate from the code that configured the audio buffer which happens earlier).  In 4.3U it's located from 81379ab8 through 81379b54.  Tidied, it looks like this:

```C
OSReport("\n"); // 8169653f
OSReport("  Game Name ... %-4.4s\n",DiscHeader.GameID);      // 81645842, 80000000
OSReport("  Company ..... %-2.2s\n",DiscHeader.MakerID);     // 8164585a, 80000004
OSReport("  Disk # ...... %x\n",    DiscHeader.DiscNumber);  // 81645872, 80000006
OSReport("  Game ver .... %x\n",    DiscHeader.DiscVersion); // 81645886, 80000007
char * streamingEnabledStr = "OFF"; // 81696554
if (DiscHeader.DiscStreamingFlag != 0) { // 80000008
  streamingEnabledStr = "ON"; // 81696551
}
OSReport("  Streaming ... %s\n",streamingEnabledStr); // 816458db
if (DiscHeader.DiscStreamingFlag != 0) { // 80000008
  if (DiscHeader.DiscStreamingBufferSize == 0) { // 80000009
    OSReport("    Buf size .. 10(default)\n"); // 81645903
  }
  else {
    OSReport("    Buf size .. %d\n",DiscHeader.DiscStreamingBufferSize); // 816458ef, 80000009
  }
}
```

Source/Core/Core/IOS/DI/DI.cpp
Pokechu22: Done.

CookiePLMonster:
@dolphin-emu-bot rebuild

Pokechu22:
I got around to doing some more hardware testing on my Wii, and can now address the earlier TODOs about drive states:

* On the Wii, a disc change will result in the drive not immediately spinning until a reset with spinup.  Doing most commands in this time will result in `ERROR_CHANGE_DISK / ERROR_MEDIUM` (0x02062800).  Once it is spun up, `ERROR_NO_DISKID_L / ERROR_NO_DISKID_H` apply instead (0x05020401), until the ID is read.  So I think my implementation of it is correct.  No idea how it works on gamecube, though.
* DTK can be configured again after resetting the drive (which *must* happen for it to be in the spun-up state; at least on the wii as far as I can tell there is no way to get it back to that state without a reset)
* DTK can be configured multiple times until a read occurs (though I'm only configuring it with the same data in this test).  Once a read occurs, `ERROR_INV_PERIOD` is returned (0x00052402).
    * Reading the disc ID a second time will remove the ability to configure DTK, same as any other read.  I think the first read (when the drive is currently in `ERROR_NO_DISKID_L`) is special and doesn't count against DTK.
    * Starting audio playback also prohibits configuration of DTK.
    * Requesting the error and checking the audio status do not prohibit (re)configuration of DTK, though the audio status cannot be checked unless DTK is enabled (failing with ERROR_AUDIO_BUF (0x00052001).

Right now I'm not enforcing `ERROR_NO_DISKID_L` (I'll fix that), and there might be other commands (the Wii drive-exclusive ones in particular, but also seek) that can be used even when they're not enabled.  I don't think it's too important as most of those commands aren't implemented fully anyways, but it's worth noting that that could be expanded upon later.

Source/Core/Core/HW/DVD/DVDInterface.h
JosJuice: We already have these registers set up for MMIO (so that GameCube games can use them). Would it be cleaner to make IOS write to the emulated console's memory (triggering the MMIO handlers) instead of it doing what it does now? That way, it's even closer to the original hardware, and we don't have to make as many changes in what we expose.

Pokechu22: That would be ideal, actually.  Does that work?  I assumed `Memory::Write_U32` would only write to the main game memory and bypass MMIOs.

JosJuice: I believe `Memory::Write_U32` works for that, yes.

Pokechu22: I've started implementing something like that and have changed to writing the DICMDBUF values at least.  `Memory::Write_U32` didn't work, but `Memory::mmio_mapping->Write` does.  I'm still trying to figure out how to handle the other values; reading a value and writing it back is a bit awkward (especially given the r/z nature of some of the registers &mdash; IOS can do `DISR = DISR & ~(DEINT|TCINT|BRKINT) | DEINTMASK` but it'd be a lot uglier to do the same with the current struct).  Maybe a hybrid approach is easiest, where the bitfields do actually have dedicated functions in DVDInterface instead...

Pokechu22: I've implemented it; the only things that needed special access were the interrupts (which already had a bit of abstraction via `DIInterruptType`); everything else can just be treated as a `u32` since `/dev/di` only passes the value on to the emulated title.

I also did some tests and found that the alignment on `DIMAR` and `DILENGTH` is enforced; bits 0-4 are always zero regardless of what's written (even if no command is triggered).  However, my testing found that bits 26-31 were kept for both, while dolphin previously was clearing them for `DIMAR`.

Pokechu22:
I think that this is roughly at its final form now.  I might reorder commits a bit, but I don't think there's anything else I want to add.

JosJuice:
@dolphin-emu-bot rebuild

BhaaLseN:
Seems good code-wise; we'll figure out any edge-cases and inconsistencies once it is merged anyways.

Source/Core/Core/HW/DVD/DVDThread.cpp
Pokechu22: I guess I probably should reply to this: I don't plan on addressing this at the moment.  The main benefit of the change is that it returns an error to the game when an actual error occurs; making it match the actual error would be ideal but would require more complex changes to actually identify what went wrong.

I'll probably look into it for a separate pull request (probably along with completely async reads where it can wait longer than normal as an option).

Source/Core/Core/IOS/DI/DI.cpp
```diff
@@ -30,93 +52,658 @@ void DI::DoState(PointerWrap& p)
...
+DiscIO::Partition DI::GetCurrentPartition()
+{
+  auto di = IOS::HLE::GetIOS()->GetDeviceByName("/dev/di");
+  if (di)
+    return std::static_pointer_cast<DI>(di)->m_current_partition;
+  else
+    return DiscIO::PARTITION_NONE;
+}
+
+void DI::InitializeIfFirstTime()
```
leoetlino: Consider adding a comment to explain that this is the equivalent of Nintendo DI's initDvdDriverStage2, which is called the first time an open/ioctl/ioctlv IPC command is received. (Otherwise, people might be tempted to move this one-time initialisation code into the constructor, which would be inaccurate)

Source/Core/Core/Boot/Boot_BS2Emu.cpp
```diff
@@ -397,7 +416,14 @@ bool CBoot::EmulatedBS2_Wii(const DiscIO::VolumeDisc& volume)
   if (!SetupWiiMemory(console_type) || !IOS::HLE::GetIOS()->BootIOS(tmd.GetIOSId()))
     return false;
 
-  DVDRead(volume, 0x00000000, 0x00000000, 0x20, DiscIO::PARTITION_NONE);  // Game Code
+  auto di = std::static_pointer_cast<IOS::HLE::Device::DI>(
+      IOS::HLE::GetIOS()->GetDeviceByName("/dev/di"));
+
+  di->ResetDIRegisters();
```
leoetlino:
```suggestion
  di->InitializeIfFirstTime();
```

Source/Core/Core/Boot/Boot_BS2Emu.cpp
```diff
@@ -397,7 +416,14 @@ bool CBoot::EmulatedBS2_Wii(const DiscIO::VolumeDisc& volume)
...
+  di->m_has_initialized = true;
```
leoetlino:
```suggestion
```

Source/Core/Core/HW/DVD/DVDInterface.cpp
```diff
@@ -682,47 +647,108 @@ void GenerateDIInterrupt(DIInterruptType dvd_interrupt)
...
+// should issue a DEINT interrupt.
+bool CheckReadPreconditions()
+{
+  if (!IsDiscInside())  // Implies ERROR_COVER or ERROR_NO_DISK
+  {
+    ERROR_LOG(DVDINTERFACE, "No disc inside.");
+    SetHighError(ERROR_NO_DISK_H);
+    return false;
+  }
+  else if ((s_error_code & LOW_ERROR_MASK) == ERROR_CHANGE_DISK)
```
leoetlino:
```suggestion
  if ((s_error_code & LOW_ERROR_MASK) == ERROR_CHANGE_DISK)
```

Source/Core/Core/HW/DVD/DVDInterface.cpp
```diff
@@ -682,47 +647,108 @@ void GenerateDIInterrupt(DIInterruptType dvd_interrupt)
...
+  {
+    ERROR_LOG(DVDINTERFACE, "Disc changed (motor stopped).");
+    SetHighError(ERROR_MEDIUM);
+    return false;
+  }
+  else if ((s_error_code & LOW_ERROR_MASK) == ERROR_MOTOR_STOP_L)
```
leoetlino:
```suggestion
  if ((s_error_code & LOW_ERROR_MASK) == ERROR_MOTOR_STOP_L)
```

Source/Core/Core/HW/DVD/DVDInterface.cpp
```diff
@@ -682,47 +647,108 @@ void GenerateDIInterrupt(DIInterruptType dvd_interrupt)
...
+  {
+    ERROR_LOG(DVDINTERFACE, "Motor stopped.");
+    SetHighError(ERROR_MOTOR_STOP_H);
+    return false;
+  }
+  else if ((s_error_code & LOW_ERROR_MASK) == ERROR_NO_DISKID_L)
```
leoetlino:
```suggestion
  if ((s_error_code & LOW_ERROR_MASK) == ERROR_NO_DISKID_L)
```

Source/Core/Core/HW/DVD/DVDInterface.cpp
```diff
@@ -1142,6 +1166,35 @@ void ExecuteCommand(u32 command_0, u32 command_1, u32 command_2, u32 output_addr
   }
 }
 
+void PerformDecryptingRead(u32 position, u32 length, u32 output_address,
+                           const DiscIO::Partition& partition, ReplyType reply_type)
+{
+  DIInterruptType interrupt_type = DIInterruptType::TCINT;
+  s_can_configure_dtk = false;
+
+  bool command_handled_by_thread =
```
leoetlino:
```suggestion
  const bool command_handled_by_thread =
```

leoetlino:
Also needs a rebase ^^

Source/Core/Core/HW/WII_IPC.h
```diff
@@ -35,6 +36,71 @@ enum StarletInterruptCause
...
+  DEBUG1 = 0x20000,
+  DEBUG2 = 0x40000,
+  DEBUG3 = 0x80000,
+  DEBUG4 = 0x100000,
+  DEBUG5 = 0x200000,
+  DEBUG6 = 0x400000,
+  DEBUG7 = 0x800000,
+};
+
+class GPIOBit
```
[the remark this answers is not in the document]
Pokechu22: Done.  (I feel like there should be a check that the enum actually has bits for its constants, but I don't know if that's a thing that can be checked or how I'd do it).

Source/Core/Core/IOS/DI/DI.cpp
```diff
@@ -30,93 +52,658 @@ void DI::DoState(PointerWrap& p)
...
+
+void DI::ChangePartition(const DiscIO::Partition partition)
+{
+  m_current_partition = partition;
+}
+
+DiscIO::Partition DI::GetCurrentPartition()
+{
+  auto di = IOS::HLE::GetIOS()->GetDeviceByName("/dev/di");
+  if (di)
```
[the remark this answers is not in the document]
Pokechu22: I don't think a guard clause makes sense here; either case is equally important (that code is hit on Gamecube mode when IOS doesn't exist, since the partition is used for `UpdateRunningGameMetadata`).

Source/Core/Core/IOS/DI/DI.cpp
Pokechu22: Done.  One thing that's been bothering me, though, is the ioctl/ioctlv cases; can those actually be hit (on console)?  I feel like the device would need to be opened first before an ioctl could be issued, so there's no reason other than paranoia to also check in ioctl/ioctlv.

Source/Core/Core/IOS/DI/DI.h
```diff
@@ -25,16 +37,111 @@ class DI : public Device
...
+    DVDLowRequestError = 0xe0,
+    DVDLowAudioStream = 0xe1,
+    DVDLowRequestAudioStatus = 0xe2,
+    DVDLowStopMotor = 0xe3,
+    DVDLowAudioBufferConfig = 0xe4
+  };
+
+  enum class DIResult : s32
+  {
+    SUCCESS = 1,
```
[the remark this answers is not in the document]
Pokechu22: Would that also apply to `GPIO`, `DIInterruptType`, and `DICommand`?

Source/Core/Core/IOS/DI/DI.cpp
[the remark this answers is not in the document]
leoetlino: Ah, that makes sense. Still, I think the convention is to not have an else after a return (IIRC... it's been a while since my last PR :P)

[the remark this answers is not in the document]
Pokechu22: I don't see anything about that in the contributing guide (the only example with return is [for multiline conditionals](https://github.com/dolphin-emu/dolphin/blob/master/Contributing.md#general), and the other examples for else [are also about lines](https://github.com/dolphin-emu/dolphin/blob/master/Contributing.md#conditionals)).  I haven't spotted any existing code that goes either way, though, so that might still just be the convention and I haven't seen it yet.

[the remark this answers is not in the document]
leoetlino: Yeah, it's not a convention that's explicitly mentioned in the contributing guide. But I just remembered it from past PRs and a quick search led me to this comment by @lioncash: https://github.com/dolphin-emu/dolphin/pull/6034#discussion_r137761025

Older parts of the codebase routinely violate that convention, but that shouldn't be the case for new code.

Source/Core/Core/IOS/DI/DI.h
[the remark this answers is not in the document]
JosJuice: You can use uppercase for all the letters of an acronym. So there's no problem with the name `DIResult` or the other three names you mentioned, but `SUCCESS` should be `Success`.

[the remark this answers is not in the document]
Pokechu22: I'm asking about the members of `DICommand` and the like.  This is what `DICommand` currently looks like:

https://github.com/dolphin-emu/dolphin/blob/0ac8ceda411806e17886ec0fa58b5344775cfe8f/Source/Core/Core/HW/DVD/DVDInterface.h#L27-L52

(I'm guessing, though, that it's fine to leave GPIO and DIInterruptType members in all caps, since at least the common terminology for them is in all caps?)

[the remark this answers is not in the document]
JosJuice: Oh, the members? For `DICommand`, they should be like `Inquiry`, `ReportKey`, `ReadDVDMetadata`, etc. I would be fine with leaving the `DIInterruptType` members in all caps. I know less about the members in `GPIO`, but if that's how they usually are written, I guess that's fine? (Especially considering the ones that don't contain full words, like `EEP_CLK`.)

JosJuice:
@dolphin-emu-bot rebuild

Source/Core/Core/IOS/DI/DI.cpp
```diff
@@ -30,93 +52,679 @@ void DI::DoState(PointerWrap& p)
...
+    DICMDBUF0 = 0xA4000000 | (param1 << 16);
+    DICMDBUF1 = param2 & 0xFFFFFF;
+    DICMDBUF2 = 0;
+    return StartDMATransfer(0x20, request);
+  }
+  case DIIoctl::DVDLowSeek:
+  {
+    const u32 position = Memory::Read_U32(request.buffer_in + 4);  // 32-bit offset
+    INFO_LOG(IOS_DI, "DVDLowSeek: position 0x%08x, translated to 0x%08x", position,
+             position);  // TODO: do partition translation!
```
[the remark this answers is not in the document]
Pokechu22: I want to wait with it, since the actual seek command isn't implemented yet so there isn't much point in doing it yet.  The comment is mainly to note that it should be done here, or at least only done for Wii games and not Gamecube games.

Source/Core/Core/HW/WII_IPC.h
```diff
@@ -35,6 +36,36 @@ enum StarletInterruptCause
...
+  DEBUG1 = 0x20000,
+  DEBUG2 = 0x40000,
+  DEBUG3 = 0x80000,
+  DEBUG4 = 0x100000,
+  DEBUG5 = 0x200000,
+  DEBUG6 = 0x400000,
+  DEBUG7 = 0x800000,
+};
+
+extern Common::Flags<GPIO> gpio_out;
```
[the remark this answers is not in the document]
Pokechu22: The `Flags` thing already somewhat acts as getters and setters, and I don't know if it would be possible to do the same thing if I had getter/setter functions that returned the `Flags` on top of it.  (I also don't think reading it with MMIO makes sense for it here, since these represent actual pins on the board that physically control devices, instead of memory locations accessed by IOS; these are the GPIOs set by broadway and not the ones written by starlet).

I could create a function like this:

```C++
bool IsGPIOActive(GPIO gpio) {
    return gpio_out[gpio];
}
```

but that feels rather silly.

Source/Core/Core/HW/DVD/DVDInterface.cpp
Pokechu22: Sure, I can do that.  At this point I'm more or less completely sure that the list on gc-forever/[the old dvd-game wiki](https://web.archive.org/web/20070708050423/http://tmb.elitedvb.net/dvd-game/index.php/DICommands) is correct, with the one exception being the debug commands (with [0xFF](https://web.archive.org/web/20070328200145/http://tmb.elitedvb.net/dvd-game/index.php/CMDFF) at least not seeming to work as-is on the Wii, but [the same wiki](https://web.archive.org/web/20070328200323/http://tmb.elitedvb.net/dvd-game/index.php/CMDFE) claims that the Wii introduced new 0xFE debug commands, so I don't know what to do with that.)  (There is one other thing that I haven't yet tested: which commands can be used with dvd-video disabled.  But those aren't emulated yet, so it doesn't really matter.)

Source/Core/Core/HW/WII_IPC.h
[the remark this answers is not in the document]
JosJuice: In that case, please put a `g_` prefix in the name of this variable (and the RTC flag one).

Source/Core/Core/HW/DVD/DVDInterface.h
```diff
@@ -97,12 +124,23 @@ bool UpdateRunningGameMetadata(std::optional<u64> title_id = {});
...
 // Used by DVDThread
 void FinishExecutingCommand(ReplyType reply_type, DIInterruptType interrupt_type, s64 cycles_late,
                             const std::vector<u8>& data = std::vector<u8>());
 
+// Used by IOS HLE
+void SetInterruptEnabled(DIInterruptType interrupt, bool enabled);
+void ClearInterrupt(DIInterruptType interrupt);
+
+// Used by EXI_DeviceIPL
+extern u32 s_rtc_flag;
```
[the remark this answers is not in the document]
Pokechu22: I've tweaked this to also use the `Flags` template, and moved it into `EXI_DeviceIPL` since that's (presumably) where it's physically located and IMO it makes a bit more sense.  Let me know how that looks.

JMC47:
Wii System Menu 3.1E freezes on boot with this PR.

image

Pokechu22:
Should be fixed now.

Miksel12:
These commits cause Gamecube games to crash Dolphin when run with MIOS.

Pokechu22:
Can confirm that crash.  Possibly related to [my changes](https://github.com/dolphin-emu/dolphin/pull/8394/files#diff-d9ccb9c0478011db5fda751ee2490c5dR389-R409) to `DVDInterface::Reset`, which [MIOS calls](https://github.com/dolphin-emu/dolphin/blob/9596fe75f12b9199057e357dc6ae78f150b1981f/Source/Core/Core/IOS/MIOS.cpp#L37).  (I'm not sure if this crash is the same as [#11952](https://bugs.dolphin-emu.org/issues/11952) as that apparently bisected to a much earlier build, but that might just be a misleading bisect).

Miksel12:
I don't get any messages, Dolphin just crashes so i guess it is actually a different problem.

PR #8395, opened 2019-10-08
Improvements to WorkQueueThread

CookiePLMonster:
- Do not use a lambda for std::thread as invoke constructor exists
- Use simpler std::lock_guard wherever possible
- Do not require T to be default constructible
- Move T out of the queue instead of copying

### DIFF
--- a/Source/Core/Common/WorkQueueThread.h
+++ b/Source/Core/Common/WorkQueueThread.h
@@ -27,14 +27,14 @@ public:
     Shutdown();
     m_shutdown.Clear();
     m_function = std::move(function);
-    m_thread = std::thread([this] { ThreadLoop(); });
+    m_thread = std::thread(&WorkQueueThread::ThreadLoop, this);
   }
 
   template <typename... Args>
   void EmplaceItem(Args&&... args)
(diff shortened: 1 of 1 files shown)


Source/Core/Common/WorkQueueThread.h
```diff
@@ -59,14 +59,13 @@ class WorkQueueThread
...
       {
-        T item;
-        {
-          std::unique_lock<std::mutex> lg(m_lock);
-          if (m_items.empty())
-            break;
-          item = m_items.front();
-          m_items.pop();
-        }
+        std::unique_lock lg(m_lock);
```
jordan-woyak: You've changed the logic to hold this lock during the function call. Other than that the changes look good.

CookiePLMonster: Line 67 unlocks the mutex explicitly, so it's not held during the function call.

jordan-woyak: Oops. Dunno how I missed that. I guess the removal of the `{}`s scared me. :P

PR #8396, opened 2019-10-08
Qt/Mapping: Show current input states in advanced dialog.

jordan-woyak:
Input list in the "advanced" dialog now has a second column showing each input state in real-time.

This may help identify desired inputs in more advanced configurations.
e.g. when mapping raw accelerometer/gyroscope.

image

### DIFF
--- a/Source/Core/DolphinQt/Config/Mapping/IOWindow.cpp
+++ b/Source/Core/DolphinQt/Config/Mapping/IOWindow.cpp
@@ -229,8 +229,8 @@ void IOWindow::CreateMainLayout()
   m_devices_combo = new QComboBox();
   m_option_list = new QTableWidget();
   m_select_button = new QPushButton(tr("Select"));
-  m_detect_button = new QPushButton(tr("Detect"));
-  m_test_button = new QPushButton(tr("Test"));
+  m_detect_button = new QPushButton(tr("Detect"), this);
+  m_test_button = new QPushButton(tr("Test"), this);
   m_button_box = new QDialogButtonBox();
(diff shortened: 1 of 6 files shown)


Source/Core/DolphinQt/Config/Mapping/IOWindow.h
```diff
@@ -94,4 +99,18 @@ class IOWindow final : public QDialog
 
   ciface::Core::DeviceQualifier m_devq;
   Type m_type;
+  std::shared_ptr<ciface::Core::Device> m_selected_device;
+};
+
+class InputStateDelegate : public QItemDelegate
```
iwubcode: This class could be moved (along with the header above) to the cpp

jordan-woyak: Done.

Source/Core/DolphinQt/Config/Mapping/IOWindow.cpp
```diff
@@ -30,25 +35,44 @@
 
 constexpr int SLIDER_TICK_COUNT = 100;
 
-IOWindow::IOWindow(QWidget* parent, ControllerEmu::EmulatedController* controller,
+class InputStateDelegate : public QItemDelegate
+{
+public:
+  InputStateDelegate(IOWindow* parent);
```
BhaaLseN: Should this one be `explicit`?

jordan-woyak: Mmmhm. Done.

PR #8399, opened 2019-10-09
Qt: Game Tracker fixes

CookiePLMonster:
This PR improves Game Tracker functionality by addressing the following:

* Fixes a race condition when rapidly clicking Refresh button - previously, game list would continuously be populated over and over, leaving itself in broken state where most entries were duplicated.
* Enables/disables Refresh button when refresh is already in place. While code technically should support "nested" refresh jobs, I don't know if you can actually end up in such state - maybe if you press Refresh button **and** use Refresh Game List hotkey at once? No idea.

This PR is generally finished, but I am marking it WIP because I am not satisfied with how it is possible to see _"Dolphin could not find any games..."_ for a split second while refreshing! Ideally, I would like it to just show something like _"Refreshing, please wait..."_

### DIFF
--- a/Source/Core/DolphinQt/GameList/GameTracker.h
+++ b/Source/Core/DolphinQt/GameList/GameTracker.h
@@ -72,7 +72,8 @@ private:
     UpdateDirectory,
     UpdateFile,
     UpdateMetadata,
-    PurgeCache
+    PurgeCache,
+    BeginRefresh,
   };
 
   struct Command
--- a/Source/Core/DolphinQt/GameList/GameTracker.h
+++ b/Source/Core/DolphinQt/GameList/GameTracker.h
@@ -74,6 +74,7 @@ private:
     UpdateMetadata,
     PurgeCache,
     BeginRefresh,
+    EndRefresh,
   };
 
   struct Command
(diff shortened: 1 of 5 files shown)


Source/Core/DolphinQt/GameList/GameTracker.h
```diff
@@ -90,6 +92,7 @@ class GameTracker final : public QFileSystemWatcher
   Common::Event m_initial_games_emitted_event;
   bool m_initial_games_emitted = false;
   bool m_started = false;
+  u32 m_busy_count = 0;
```
lioncash: It may help to have a tiny documentation comment for this variable, given it's not obvious what it's used for just by reading it in isolation (and can prevent it being used incorrectly if code is ever changed in the future).

BhaaLseN: Should this perhaps even be an `atomic`? Or is this considered safe since only one thread (which is `m_load_thread`) should be changing it at any given time?

CookiePLMonster: > Or is this considered safe since only one thread (which is `m_load_thread`) should be changing it at any given time?

That is the case indeed, I ensured this only gets accessed from the load thread.

CookiePLMonster:
I now added a comment as requested and removed WIP tag, because I figured the issue I mentioned (_"Dolphin could not find games..."_) might as well be treated as a separate bug to fix, as:

- it happens both with and without this PR
- it may not be super trivial to fix

PR #8400, opened 2019-10-12
SignatureDB: Add missing selectors

sepalani:
Fix a DolphinQT regression.

### DIFF
--- a/Source/Core/DolphinQt/MenuBar.cpp
+++ b/Source/Core/DolphinQt/MenuBar.cpp
@@ -57,6 +57,13 @@
 
 QPointer<MenuBar> MenuBar::s_menu_bar;
 
+QString MenuBar::GetSignatureSelector() const
+{
+  return QStringLiteral("%1 (*.dsy);; %2 (*.csv);; %3 (*.mega)")
+      .arg(tr("Dolphin Signature File"), tr("Dolphin Signature CSV File"),
+           tr("WiiTools Signature MEGA File"));
+}
+
 MenuBar::MenuBar(QWidget* parent) : QMenuBar(parent)
 {
   s_menu_bar = this;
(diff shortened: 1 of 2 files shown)


Source/Core/DolphinQt/MenuBar.cpp
```diff
@@ -56,6 +56,10 @@
 #include "UICommon/GameFile.h"
 
 QPointer<MenuBar> MenuBar::s_menu_bar;
+const QString MenuBar::SIGNATURE_SELECTOR =
+    tr("Dolphin Signature File (*.dsy)") + QStringLiteral(";;") +
+    tr("Dolphin Signature CSV File (*.csv)") + QStringLiteral(";;") +
+    tr("WiiTools Signature MEGA File (*.mega)");
```
JosJuice: This would run at the beginning of the program execution, right? `tr` doesn't work at that point because the translation handler hasn't been set up yet.

sepalani: I forgot about that, I replaced it with a private member function.

Source/Core/DolphinQt/MenuBar.cpp
```diff
@@ -57,6 +57,13 @@
 
 QPointer<MenuBar> MenuBar::s_menu_bar;
 
+const QString MenuBar::GetSignatureSelector() const
```
lioncash:
```suggestion
QString MenuBar::GetSignatureSelector() const
```

const when returning by value doesn't do anything beneficial (and can even inhibit return value optimizations)

Source/Core/DolphinQt/MenuBar.cpp
```diff
@@ -57,6 +57,13 @@
 
 QPointer<MenuBar> MenuBar::s_menu_bar;
 
+QString MenuBar::GetSignatureSelector() const
+{
+  return tr("Dolphin Signature File (*.dsy)") + QStringLiteral(";;") +
```
spycrab:
```c++
return QStringLiteral("%1 (*.dsy);; %2 (*.csv);; %3 (*.mega)").arg(tr("Dolphin Signature File"), tr("Dolphin Signature CSV File"), tr("WiiTools Signature MEGA File"));
```
is probably preferable

sepalani: Done.

spycrab:
LGTM

PR #8404, opened 2019-10-14
Add /Source/Android/app/.cxx/ to .gitignore

JosJuice:
Recent versions of Android Studio seem to put stuff here.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,7 +8,8 @@ Source/Core/Common/scmrev.h
 /[Bb]uild*/
 /[Bb]inary/
 /obj/
-# Android cmake builds to here then copies to Source/Android.
+# Ignore files output by Android cmake build
+/Source/Android/app/.cxx/
 /libs/
 # Ignore various files created by visual studio/msbuild
 *.ipch


.gitignore
```diff
@@ -8,6 +8,7 @@ Source/Core/Common/scmrev.h
 /[Bb]uild*/
 /[Bb]inary/
 /obj/
+/Source/Android/app/.cxx/
```
lioncash: This should probably have an accompanying comment to provide context for the listing in the gitignore. Looks fine otherwise.

JosJuice: I don't really much other than that it gets created than building for Android, though... And it does say Android in the path already. Do you have a suggestion for what to write in the comment?

lioncash: A suitable comment would just mention that Android Studio creates files in that directory. Similar to the comment under it indicating that we ignore certain extensions because they're ignorable files creates by Visual Studio.

JosJuice: Does the current commit look good?

lioncash: Yup! 👌

PR #7711, opened 2019-01-15
Qt/Debugger Code Widget: Click address box to go to address.

TryTwo:
Previously had to change the value in the address box to go to address again.

Surprisingly delicate issue. Can easily break highlighting the address when trying to enable a click signal.  Looks like this method works great though. Not sure if there's another preferred method though.

Also, made it so the address size must be 8 for the code view to navigate to it.

PR #8405, opened 2019-10-14
Qt/Debugger/CodeWidget: Allow pressing 'enter' in address search box.

AdmiralCurtiss:
Pretty straightforward, connect the `returnPressed` signal to the same target as the `textChanged` one. This allows you to press Enter to go back to the address currently listed in the search box, which I feel like I need a surprising number of times while debugging...


AdmiralCurtiss:
FYI this probably addresses the same usability issue as #7711. I think I prefer this method of requiring an explicit keyboard key press.

PR #8406, opened 2019-10-15
CMake: Enable SDL by default on macOS.

jordan-woyak:
https://bugs.dolphin-emu.org/issues/11741 (and probably others)

Our IOKit code has problems with many controllers.
SDL seems to have fewer problems.

Let's see if our macOS buildbot can handle enabling SDL by default.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -52,8 +52,8 @@ option(OPROFILING "Enable profiling" OFF)
 # TODO: Add DSPSpy
 option(DSPTOOL "Build dsptool" OFF)
 
-# Enable SDL for default on operating systems that aren't OSX, Android, Linux or Windows.
-if(NOT APPLE AND NOT ANDROID AND NOT CMAKE_SYSTEM_NAME STREQUAL "Linux" AND NOT MSVC)
+# Enable SDL for default on operating systems that aren't Android, Linux or Windows.
+if(NOT ANDROID AND NOT CMAKE_SYSTEM_NAME STREQUAL "Linux" AND NOT MSVC)
   option(ENABLE_SDL "Enables SDL as a generic controller backend" ON)
 else()
   option(ENABLE_SDL "Enables SDL as a generic controller backend" OFF)


hebriel:
I'm underdisk. I made an error on the bug website so I repeat it here (we never know). With the Switch Pro Controller on macOS, using SDL fixed the problems! Big yes.

spycrab:
Are we sure we don't want to fix our IOKit backend instead?

jordan-woyak:
@spycrab I've seen so many reports of non-working controllers in Dolphin on macOS since I've been involved with the project. Things go unreported or stay broken for months, years, or they never get fixed. Once in a while someone will come around and fix up our macOS input backend a bit but we never have consistent people to work on any macOS-specific code. Even if you, or anyone, fixes up our IOKit code, I think it's silly to imagine it's going to have compatibility on par with SDL. We aren't going to get the level of testing that SDL has. Even if we did have people and testing to get our IOKit backend to a capable level why waste the time when SDL already exists? Why try to maintain more macOS-specific code when a ubiquitous library like SDL already works?

TLDR: Enable SDL (which uses IOKit). Kill our IOKit code. Eliminate 700 lines of macOS-specific code that we struggle to maintain. Profit.

spycrab:
By that logic we could just replace all of our input backends with SDL.

JMC47:
We have people who maintain Linux and Windows consistently.  macOS we do not and even the developers that do care if macOS works correctly are tired of all the separate problems it has.

SDL on Windows was causing crashes, which is why it was outright disabled there.

phire:
Was causing crashes on linux too, which is why it was disabled there.  
It's fixed now, as long as the user is running a new-enough distro.

It could be smart to standardise on SDL 2.0 for all operating systems, but it would require porting any extra features we need into SDL, which would probably require us to ship a custom version of SDL in externals, and take measures on linux to prevent distro-maintainers ripping our custom SDL out and using the distro packaged version of SDL.

hebriel:
Any example of such "extra features" ? (I'm not a dolphin dev)

jordan-woyak:
SDL isn't flawless (re: crashes) which is why I don't think we should solely rely on it for input on all platforms. I can't think of "extra features" we'd NEED that SDL doesn't provide but one problem is "nice" naming of axes. SDL has only axis numbers for generic "joysticks" so we'd lose all of our "Axis X"-type names (which would break existing configs on Windows). "Custom SDL" would be silly as that would be exactly the same as what we are doing now (maintaining three sets of OS-specific code) only much code and in an unfamiliar codebase all written in pure C.

If someone wants to fix our IOKit code, go ahead. In the meantime why not enable SDL on macOS so things work?

Helios747:
It would be ideal to fix our IOKit backend, but it's clear we don't have anybody that gives a shit about MacOS and hoping that somebody shows up to give one has proven to not be very effective. So I'll pick a revertable 2 LOC PR as a solution here.

hebriel:
macOS guy here, I still encounter the problem using 5.0-10993. No SDL interface appear in the device list, but personal build using -DENABLE_SDL=1 does work.

PR #8408, opened 2019-10-15
Qt: Move CPU Emulation Engine options to the Advanced tab

ethteck:
Solves https://bugs.dolphin-emu.org/issues/11743

I changed the CPU Emulation options to a combobox selection and moved it over to the Advanced tab.

### DIFF
--- a/Source/Core/DolphinQt/Config/SettingsWindow.cpp
+++ b/Source/Core/DolphinQt/Config/SettingsWindow.cpp
@@ -47,7 +47,7 @@ SettingsWindow::SettingsWindow(QWidget* parent) : QDialog(parent)
     wii_pane->OnEmulationStateChanged(state != Core::State::Uninitialized);
   });
 
-  m_tab_widget->addTab(new AdvancedPane(), tr("Advanced"));
+  m_tab_widget->addTab(GetWrappedWidget(new AdvancedPane, this, 125, 200), tr("Advanced"));
 
   // Dialog box buttons
   QDialogButtonBox* close_box = new QDialogButtonBox(QDialogButtonBox::Close);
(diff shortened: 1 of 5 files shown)


MayImilae:
Screenshot please!

ethteck:
Sure! Here's the new General tab:
General tab

And the Advanced tab:
Advanced tab

MayImilae:
Works well! Still though, there definitely needs to be more space between the CPU Emulation Engine and the CPU Clock Override, mainly because CPU Engine is SO much smaller than the enormous Clock Override.

So it might be best to have a CPU Options group and a Clock Override group.

ethteck:
Okay, I'll work on that tomorrow. While we're at it, is the "Enable Dual Core" option in the General tab also suitable to go under "CPU Options"?

MayImilae:
Probably best to leave that as is for now. Moving Dual Core will spur some discussion.

JosJuice:
I think dual core should be on the General tab like it currently is. It's something you have a lot more reason to adjust than the CPU emulation engine.

BhaaLseN:
Please rebase your branch instead of merging (`git fetch origin; git rebase origin/master move-cpu-emulation-engine-options` - replace `origin` with the name of the dolphin remote). This also gives you the option of [combining commits](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History#_squashing) if you use `git rebase -i` instead.

Helios747:
If May is happy with this layout (I am), I see no reason to hold this back. Would be nice if that third commit could be squashed into the first or second. First two commits are fine.

MayImilae:
LGTM!

ethteck:
I went ahead and squashed the third commit into the second.

Helios747:
Thanks!

PR #3807, opened 2016-04-30
[WIP] VideoCommon: Add epsilon to work around Nvidia rounding issue in texture lookup

endrift:
Fixes [ea-vp6](https://bugs.dolphin-emu.org/issues/7193) on Nvidia (at least on OS X). Also a terrible hack. ~~Probably breaks a lot of stuff.~~ I'm mostly posting this to see how much fifoci complains, and also to try and spur conversation about how to actually fix it.

---

This change is [Reviewable](https://reviewable.io/reviews/dolphin-emu/dolphin/3807)

PR #8409, opened 2019-10-17
Fixing video for EA games like Fifa 07

raelgc:
This will remove the glitch (something like jailbars) in EA videos (like the opening "EA Sports, it's in the game" in Fifa 07).

Tested in OpenGL renderer.

Discussion: https://bugs.dolphin-emu.org/issues/7193#note-25

### DIFF
--- a/Source/Core/VideoCommon/PixelShaderGen.cpp
+++ b/Source/Core/VideoCommon/PixelShaderGen.cpp
@@ -1198,7 +1198,7 @@ static void WriteStage(ShaderCode& out, const pixel_shader_uid_data* uid_data, i
         out.Write("\ttevcoord.xy = int2(0, 0);\n");
     }
     out.Write("\ttextemp = ");
-    SampleTexture(out, "float2(tevcoord.xy)", texswap, stage.tevorders_texmap, stereo, ApiType);
+    SampleTexture(out, "(float2(tevcoord.xy) + 0.05)", texswap, stage.tevorders_texmap, stereo, ApiType);
   }
   else
   {


stenzek:
Not a huge fan of the "hack", but outside of implementing sampling/filtering in the shader (and taking the performance hit) I haven't thought of a better idea..

If this were to go ahead, you'd want to update the ubershader as well.

raelgc:
@AdmiralCurtiss: It looks like the source of the issue is a rounding error. According the author of the working solution:

> The real fix for this is to integerize texture lookups, but wow that is out of my realm of expertise.

iwubcode:
Has this been tested against other games?  I'm just wondering if this breaks anything.  What about fifoci?

I know of a number of small graphical (potentially texture related?) game errors, I've always wondered if Dolphin had something wrong somewhere but wasn't sure where to look.  I'll have to see if this fixes any of them.

endrift:
Hello, original author of that hack here. There was a reason this never got merged. See also the original PR for this: #3807

raelgc:
Ok, so, how if I implement like suggested on #3807: a `0.25` rounding that will be tied to a `[Core]` option that can be enabled for some games (like `FPRF`)? Something titled "**Fix rounding issue in texture lookup**"?

dkacperski97:
So what is the status of this PR? It seems https://bugs.dolphin-emu.org/issues/4570 (mentioned in the original PR) has been fixed, but the changes didn't fix https://bugs.dolphin-emu.org/issues/7193.

Miksel12:
I think this pr should be closed as #9407 shows that the epsilon value needed depends on which game and which hardware you are running. Making an epsilon hack very hacky.

endrift:
Honestly I didn't realize this was still open.

JMC47:
Everyone is recommending this be closed.  This is still a mystery to this day.

PR #8410, opened 2019-10-18
CMake/Windows: Fix warnings about flags being overridden

spycrab:
Fixes the build log being spammed with [Warning D9025](https://docs.microsoft.com/en-us/cpp/error-messages/tool-errors/command-line-warning-d9025?view=vs-2019).

### DIFF
--- a/Source/CMakeLists.txt
+++ b/Source/CMakeLists.txt
@@ -15,8 +15,14 @@ set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
 
 if (MSVC)
-  # Set warning level to 4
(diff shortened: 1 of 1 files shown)


AdmiralCurtiss:
Depending on how old our CMake installation is it would be preferrable to set policy 0092 instead: https://cmake.org/cmake/help/latest/policy/CMP0092.html

If that's not possible, at least update the link to not point at a forwarder site.

spycrab:
Pretty sure we can't require CMake 3.15 at this point in time, I can add the policy to the comment though.

spycrab:
e.g. even Ubuntu 19.10 doesn't have 3.15 yet: https://packages.ubuntu.com/eoan/cmake

PR #8411, opened 2019-10-18
Windows: Make builds more reproducible

spycrab:
Hopefully helps us with whitelisting our updater...

### DIFF
--- a/Source/CMakeLists.txt
+++ b/Source/CMakeLists.txt
@@ -50,6 +50,9 @@ if (MSVC)
   add_subdirectory(PCH)
   add_definitions(/I${PCH_DIRECTORY})
   add_definitions(/Yu${PCH_PATH}) 
+
+	# Don't include timestamps in binaries
+	add_link_options(/Brepro)
 endif()
 
 # These aren't actually needed for C11/C++11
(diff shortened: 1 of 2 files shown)


BhaaLseN:
Somewhat sad that this has to be an undocumented flag. At least in Roslyn, theres the documented `/deterministic` switch and the project option `<Deterministic>true</Deterministic>`.
But it could be worth a try. +1 from my side unless someone specifically disagrees (or has a good reason to _not_ do this).

spycrab:
I'm not quite sure yet if this actually helped any. It looks to me as if at least one timestamp remains. Has to be looked into further

BhaaLseN:
I did a few builds/rebuilds and came out with the same md5sum; even though switched commits between them. So I guess...it _kinda_ works?

spycrab:
Well it might be better than nothing then at least.

mbc07:
Doesn't seem to work, updater binary still get different hashes after this PR. At least I successfully downloaded 5.0-11044, 5.0-11046 and 5.0-11048 without triggering false positive alerts on Windows Defender, but unsure if the problem was really fixed or if it's just that somebody else already submitted those binaries for analysis...

BhaaLseN:
We don't bake the sha1 in like we do with the main executable, do we? I'm not quite sure _why_ this happens tho...

PR #8412, opened 2019-10-18
Core/HW: Fix some explicitly defaulted but implicitly deleted warnings.

jordan-woyak:
These classes are not copyable or movable because they indirectly have std::atomic members.
Defaulting their constructors and move operators was producing warnings in clang.


lioncash:
It would likely be nicer to keep the explicitness and mark them as deleted, like the copy constructors and assignment operators are.

jordan-woyak:
> It would likely be nicer to keep the explicitness and mark them as deleted, like the copy constructors and assignment operators are.

Sounds good.

PR #8413, opened 2019-10-18
More descriptive names for ubershader modes

blaahaj:


phire:
The new names (descriptions?) seem sane to me.

MayImilae:
My initial response to this was _"oh gods no more!"_ but actually looking at it this is actually not bad. Then I looked at it more and was like, oh, yea, that's why we didn't go this route.

The reason these labels are the way they are is because that's the only way to have them be even remotely correct to what is going on. I'll give an example.

"Suspend while game is compiling". There's no way to make this right. For both Synchronous (Ubershaders) and regular Synchronous, what is going on is that Dolphin's CPU thread is waiting while the GPU thread catches up to maintain proper synchrony. _Outwardly_ this appears like Dolphin is stalling, but it isn't; Dolphin is still working as fast as it can, just the computer isn't fast enough to meet the timing targets and users see stutters. "Pause", "Stall", "Suspend", and "Halt" are all technically inaccurate terms and have a lot of misleading implications that Dolphin isn't doing anything during stutters. Saying "CPU thread waits for the GPU thread to finish shader compilation" is much too long. The most technically accurate thing I could think of that follows your attempt at natural language would be "Wait for Shader Compilation". Which is sort of right, I guess, but also very informative since it doesn't say what is waiting. Wait is a verb after all!

So with a back and forth of dozens of people over many weeks, we just said "Synchronous", in ubershaders or not, and Asynchronous, ubershaders or not (skip drawing is just added to make the inevitable visual glitches faster for users to spot), while relying on the description for the actual details. It's not perfect but, it works ok, and is accurate.

So yea, I'm against this. Sounds good on the surface but, really bad when one thinks about it.

phire:
> "Wait for Shader Compilation"

The more technical term would be "Block while compiling shaders". But I'm not sure if all users will understand this.

Techjar:
> The more technical term would be "Block while compiling shaders". But I'm not sure if all users will understand this.

Yeah "block" is a very computer science-y term that I doubt most people will understand.

blaahaj:
'Synchronous (Ubershaders)' is the one I dislike most. It sounds like it blocks like 'Synchronous', but it does not. It is not incorrect as it is synchronous, but it is ahead of time not just in time.

Techjar:
It does block though, it's just that the ubershader only has to be compiled once per pipeline configuration, rather than once per TEV configuration.

MayImilae:
"Sychronous (Ubershaders)" still blocks, but it behaves differently. Instead of hard stutters that come and go quickly, it can overtax your GPU resulting in a constant framerate decrease.

blaahaj:
Oh, I am sorry, I did not realise it was a runtime compilation.

phire:
> Yeah "block" is a very computer science-y term that I doubt most people will understand.

Then again, if we were ok with using "synchronous" and relying on the description to explain it, then we should be ok with Block. Even Wait or Suspend require extra description.

> It does block though, it's just that the ubershader only has to be compiled once per pipeline configuration

It (tries to) precompiles all pipeline configurations at start, so it doesn't block during runtime, there should always be a valid ubershader sitting there for you to use.  
And you better hope there is, the stutter for blocking to compile an ubershader is much larger than the stutter for a regular shader.

There are still edge cases where the driver might choose to compile a variant we didn't expect, and block, but dolphin doesn't block and in the general case the driver doesn't block either.

> "Sychronous (Ubershaders)" still blocks, but it behaves differently

Slow downs are not blocks. It didn't block, it just used an alternative path which was slower.

MayImilae:
>Slow downs are not blocks. It didn't block, it just used an alternative path which was slower.

Then we shouldn't use Block and stick with Synchronous?

Techjar:
> It (tries to) precompiles all pipeline configurations at start, so it doesn't block during runtime, there should always be a valid ubershader sitting there for you to use.

Ah, I thought it would only compile and cache a particular pipeline configuration when it was first encountered. Though I guess that would be annoying since the cache is invalidated whenever changes are made to the shader code.

blaahaj:
From the descriptions, 'Synchronous (Ubershaders)' does not sound more synchronous than 'Asynchronous (Ubershaders)'

phire:
> Then we should not change to Block? Synchronous still works well.

The option is regarding what to do when dolphin encounters are shader it hasn't compiled. 

In Ubershaders only, it never encounters a shader it hasn't compiled, completely bypassing the question of Synchronous vs Asynchronous,. It is neither.

MayImilae:
But it still operates synchronously, hence the slowdown if the PC can't keep up. As differs from asychronous, like skip drawing, where the game will just keep going regardless.

JosJuice:
The kind of slowdown you get with asynchronous ubershaders is the same kind as you get with either of the asynchronous options. It's from the GPU not running fast enough (and not anything related to compilation).

MayImilae:
Yea Asychronous Ubershaders was always a weird one.

blaahaj:
I changed 'Suspend' -> 'Stall', 'Only' -> 'Always', but these are minor

Techjar:
"Stall" is not any better.

blaahaj:
But, is it worse? 'Suspend' seems less momentary to me, is all

Techjar:
I think all words you could possibly use there are bad, as they don't remotely accurately describe what's going on.

blaahaj:
I know it is imperfect, but I think it is less mystifying than what it replaces

blaahaj:
| x  | **Fast** | **Slow** |
| -- | -- | -- |
| **Smooth**  | Asynchronous (Skip drawing) | Synchronous (Ubershaders) <br> Asynchronous (Ubershaders) |
| **Stuttery**  | Synchronous | 💡 future innovation |

Miksel12:
I don't know why the names should describe the effect of the option when there is a description at the bottom. With this change, the names are the same as the discription but just formulated in a few words. If the names of the options should discribe the option then why have a discription. 
I agree that in the current state, new users will likely not understand what these names mean but they can see their effect and functioning in the discription. This is also the case for many other options in the graphics config.

MayImilae:
We just spent the two hours or so discussing this topic in IRC. What we settled onto was:

Synchronous ➜ Traditional
Asynchronous (Ubershaders) ➜ Hybrid Ubershaders
Synchronous (Ubershaders) ➜ Exclusive Ubershaders
Asynchronous (Skip Drawing) ➜ Skip Drawing

This returns it pretty closely to the old arrangement but now with Skip Drawing. If you'd like to change this PR to reflect that you are welcome to, or we could make a new PR ourselves.

Personally I'd recommend letting one of us make the PR. There will bound to be more issues that come up, like the precise ordering of the options and maybe some description changes., and it's a LOT faster to hammer out those details in IRC vs github comments. 

Either way, thank you for bringing this back into our discussion. None of us were really happy with the old naming scheme, but we were so scared to get back into the discussions we kind of just sat on it. Since you forced us to look back into it, we came up with something a LOT better!

blaahaj:
That sounds like some improvement. But, I think “Traditional” is even vaguer than the current name.

MayImilae:
It's no more or less vague than saying "Ubershaders", I mean, what even are those. :P It's a name, basically, and it isn't _wrong_ like the existing names were. The descriptions will handle the heavy lifting by describing what these actually do.

blaahaj:
Maybe:

🔘 Ubershaders
┠  ⚫️ Exclusive
┖  ⚫️ Hybrid
⚫️ No ubershaders
┠ ⚫️ Skip drawing
┖ ⚫️ Stall

blaahaj:
Another:

🔘Use ubershaders exclusively
⚫️Compile specialised shaders, meanwhile
┠ ⚫️Stall
┠ ⚫️Skip drawing
┖ ⚫️Use ubershader

stenzek:
I think it's fine as-is.

blaahaj:
New PR obsolete this

PR #8416, opened 2019-10-19
Change the names for shader compilation settings

JosJuice:
See https://github.com/dolphin-emu/dolphin/pull/8413#issuecomment-544006931

### DIFF
--- a/Source/Core/DolphinQt/Config/Graphics/GeneralWidget.cpp
+++ b/Source/Core/DolphinQt/Config/Graphics/GeneralWidget.cpp
@@ -113,10 +113,10 @@ void GeneralWidget::CreateWidgets()
   auto* shader_compilation_layout = new QGridLayout();
 
   const std::array<const char*, 4> modes = {{
-      QT_TR_NOOP("Synchronous"),
-      QT_TR_NOOP("Synchronous (Ubershaders)"),
-      QT_TR_NOOP("Asynchronous (Ubershaders)"),
-      QT_TR_NOOP("Asynchronous (Skip Drawing)"),
+      QT_TR_NOOP("Specialized (Default)"),
+      QT_TR_NOOP("Exclusive Ubershaders"),
+      QT_TR_NOOP("Hybrid Ubershaders"),
+      QT_TR_NOOP("Skip Drawing"),
   }};
   for (size_t i = 0; i < modes.size(); i++)
   {
@@ -250,20 +250,22 @@ void GeneralWidget::AddDescriptions()
       QT_TR_NOOP("Shows chat messages, buffer changes, and desync alerts "
                  "while playing NetPlay.<br><br><dolphin_emphasis>If unsure, leave "
                  "this unchecked.</dolphin_emphasis>");
-  static const char TR_SHADER_COMPILE_SYNC_DESCRIPTION[] =
+  static const char TR_SHADER_COMPILE_SPECIALIZED_DESCRIPTION[] =
       QT_TR_NOOP("Ubershaders are never used. Stuttering will occur during shader "
                  "compilation, but GPU demands are low.<br><br>Recommended for low-end hardware. "
                  "<br><br><dolphin_emphasis>If unsure, select this mode.</dolphin_emphasis>");
-  static const char TR_SHADER_COMPILE_SYNC_UBER_DESCRIPTION[] = QT_TR_NOOP(
+  // The "very powerful GPU" mention below is by 2021 PC GPU standards
+  static const char TR_SHADER_COMPILE_EXCLUSIVE_UBER_DESCRIPTION[] = QT_TR_NOOP(
       "Ubershaders will always be used. Provides a near stutter-free experience at the cost of "
-      "high GPU performance requirements.<br><br><dolphin_emphasis>Only recommended "
-      "for high-end systems.</dolphin_emphasis>");
-  static const char TR_SHADER_COMPILE_ASYNC_UBER_DESCRIPTION[] = QT_TR_NOOP(
+      "very high GPU performance requirements.<br><br><dolphin_emphasis>Don't use this unless you "
+      "encountered stuttering with Hybrid Ubershaders and have a very powerful "
+      "GPU.</dolphin_emphasis>");
+  static const char TR_SHADER_COMPILE_HYBRID_UBER_DESCRIPTION[] = QT_TR_NOOP(
       "Ubershaders will be used to prevent stuttering during shader compilation, but "
       "specialized shaders will be used when they will not cause stuttering.<br><br>In the "
       "best case it eliminates shader compilation stuttering while having minimal "
       "performance impact, but results depend on video driver behavior.");
-  static const char TR_SHADER_COMPILE_ASYNC_SKIP_DESCRIPTION[] = QT_TR_NOOP(
+  static const char TR_SHADER_COMPILE_SKIP_DRAWING_DESCRIPTION[] = QT_TR_NOOP(
       "Prevents shader compilation stuttering by not rendering waiting objects. Can work in "
       "scenarios where Ubershaders doesn't, at the cost of introducing visual glitches and broken "
       "effects.<br><br><dolphin_emphasis>Not recommended, only use if the other "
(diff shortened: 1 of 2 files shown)


Source/Core/DolphinQt/Config/Graphics/GeneralWidget.cpp
```diff
@@ -110,10 +110,10 @@ void GeneralWidget::CreateWidgets()
   auto* shader_compilation_layout = new QGridLayout();
 
   const std::array<const char*, 4> modes = {{
-      QT_TR_NOOP("Synchronous"),
-      QT_TR_NOOP("Synchronous (Ubershaders)"),
-      QT_TR_NOOP("Asynchronous (Ubershaders)"),
-      QT_TR_NOOP("Asynchronous (Skip Drawing)"),
+      QT_TR_NOOP("Traditional"),
```
blaahaj: I like the other names. But I think 'Traditional' is worse than even 'Synchronous'. 'Synchronous' at least tries to explain it, but 'Traditional' only is meaningful if you know what the Tradition is. If it would be changed back, I would fully approve.

JosJuice: I guess reverting just that one to "Synchronous" would make sense in a way... It is the only option that compiles shaders synchronously during normal gameplay, after all. What do others think about this?

MayImilae: From what we discussed in IRC, basically *everything* is synchronous. Calling one option Synchronous is not accurate, so I'm against that. Traditional works because it is a valid name: it is the traditional method of shader compilation handling in Dolphin and elsewhere.

JosJuice: > From what we discussed in IRC, basically *everything* is synchronous.

That was when we were talking about slowdown caused by the GPU performing work. But the other options don't do synchronous **shader compilation**, and shader compilation is exactly what this option is about, which is why I think synchronous makes sense for this option.

CookiePLMonster: My first thought is that "Standard" would probably be nice - while it's similar to "Traditional", it is more obvious for the user that they are "normal" or "the most regular".

In an emulator like this, I can imagine people thinking "Traditional" is related to classic/retro modes, as in "looking like it used to on a real console and TV back in the day".

MayImilae: "Normal" was the original suggestion, but phire pointed out that it's actually pretty abnormal from a graphics standpooint, it's just the normal for us. Hence, Traditional.

blaahaj: It can make sense to an emulator programmer. But to reference assumed knowledge is not a substitute for an attempt at actual description.

jordan-woyak:
This is miles better than the current "Synchronous" naming.
I don't hate "Traditional" but I might suggest "Specialized".

stenzek:
I'm still not a fan of this because it doesn't communicate which shaders are being used, nor how they are being compiled. How does "traditional" mean "specialized shaders and stall emulation until compilation is finished?"

Extending what @jordan-woyak suggested, something like:
 - Specialized
 - Specialized+Ubershaders
 - Ubershaders
 - Specialized+Skip

Would better represent the behavior. At least it specifies which shaders are actually being used, even if the matter in which it affects emulation isn't. But I still prefer the current names.

JosJuice:
@stenzek Something that probably will be a problem with your suggestion is that people are likely to see that there's an option that just says Ubershaders and pick it simply because they've heard that ubershaders are good (even though this option is actually worse than Specialized + Ubershaders). I do think the naming makes sense on a technical level, but that specific thing has the potential to lead to a lot of trouble.

nbohr1more:
I would call it "pre-baked" rather than "specialized" and call Ubershaders "realtime emulated"

JosJuice:
But the specialized shaders aren't pre-baked. They're generated at runtime (barring a shader cache).

nbohr1more:
They are pre-baked in the sense that their code variants are all known a-priori. I guess you could call them "predicted"?

JosJuice:
No, that is exactly what they are not.

jordan-woyak:
@stenzek The problem with the current phrasing is the discrepancy in what's being described.
- A. Synchronous
  - "synchronous" describes waiting on specialized shader compilation.
- B. Synchronous (Ubershaders)
  - "synchronous" describes waiting on Ubershader compilation.
- C. Asynchronous (Ubershaders)
  - "asynchronous" describes not waiting on specialized shader compilation but we still have to wait for Ubershaders which are just as "synchronous" as in mode "B" yet we use the completely opposite word. **This is the problem!**
- D. Asynchronous (Skip Drawing)

The difference from mode "B" to mode "C" is use of specialized shaders after they compile in the background but the current phrasing implies there is less waiting in mode "C" which is not true.
If anything there is more waiting because of poor driver behavior.

"B" describes waiting on the thing in the parentheses, Ubershaders, while "C" describes not waiting on the thing NOT in the parentheses, specialized shaders, which aren't even mentioned. They describe different properties of the implementation and it's confusing. When I want to test the mode that uses just Ubershaders I have to stop and think every time about which of "B" or "C" is the one without specialized shaders.

Every mode is "synchronous" except "D". "C" has some "asynchronous" compilation going on but it is still very much "synchronous".

KatherineFtw:
Ok, so. Let me tackle this from a separate angle. As a user, what do I see for all of these? I see "Ubershaders" and three variants. By it's self, Ubershaders, Traditional, Sync and ASync, Specialized, Skip. Their all meaningless in informing me on what they all do, or very well what they are, mean, or if default is accurate/ok.

At the end of the day, realize that most people do not know, and probably don't care. 

I suggest it's handled like RPCS3's Options section of the wiki. It's helped me more than anything else suggested combined.

https://wiki.rpcs3.net/index.php?title=Help:Default_Settings

TL;DR: No amount of arguing over which jargon to use will actually resolve anything to non-jargon informed users.

stenzek:
@jordan-woyak B/C shouldn't involve any waiting for ubershaders as they are all compiled at startup.

Maybe another way of expressing the options would be better. For example, a slider with specialised shaders on the left, specialised+ubershaders in the middle, ubershaders on the right. So defining performance from fastest to slowest, but stutter from most to least. Skip drawing could be an additional checkbox as it's unrelated to ubershaders, and checking it would disable/grey out the slider.

@KatherineFtw having such documentation is great, but I fear that users won't read it. We're lucky if people read the in-dialog.text, let alone opening their browser and searching :(

Ebola16 (approved):
As long as the changes are technically correct, I'm ok with the current version of this PR. I had a problem with the previous naming scheme because the names were too similar. People could report a problem with "synchronous" shaders and really mean "Synchronous (Ubershaders)" but they only remembered the first word of the setting.

Source/Core/DolphinQt/Config/Graphics/GeneralWidget.cpp
blaahaj: Idea: 'Wait For Compile'?

MayImilae:
So pondering this some more, one option for this that we could do is to go with a more accurate name for specialized shaders and just stick (Default) beside it. Like this:

* Specialized (Default)
* Exclusive Ubershaders
* Hybrid Ubershaders
* Skip Drawing

JosJuice:
I would be fine with that. Does anyone have anything against it?

mbc07:
The new names LGTM. 

On a side note, since we're changing the naming, I would probably throw a note on the Exclusive Ubershaders description text to inform the users they should prefer Hybrid Ubershaders instead of Exclusive unless they really know what they're doing. I've seen quite a bunch of people straight up selecting "Exclusive" instead of "Hybrid" thinking it's better then complaining on the forums and on other communities that everything became slow...

JosJuice:
Hm, I wonder what the change should be in more concrete terms... What do you think about replacing "Only recommended for high-end systems" with "Don't use this unless you have a very powerful GPU"?

JMC47:
I'm good with those names.

JosJuice:
I've made the changes.

mbc07:
> What do you think about replacing "Only recommended for high-end systems" with "Don't use this unless you have a very powerful GPU"?

People with high end GPUs might still feel "encouraged" to select this, and that group is also the most likely to be running 6x+ IR with AA and other expensive settings, those coupled with Exclusive Ubershaders can bring even the mightiest of the cards to its knees. 

What about "Requires a very powerful GPU, use only if you still experience shader compilation stuttering with Hybrid Ubershaders"?

MayImilae:
> Hm, I wonder what the change should be in more concrete terms... What do you think about replacing "Only recommended for high-end systems" with "Don't use this unless you have a very powerful GPU"?

I'm ok with this, but I suggest putting in a comment dating it to 2021. What defines a powerful GPU will change and it will be convenient to have a date assigned to it for when it comes time to alter that.

JosJuice:
@mbc07 Would you be fine with "Don't use this unless you have a very powerful GPU and experience shader compilation stuttering with Hybrid Ubershaders"? I feel like "Requires a very powerful GPU" feels a bit redundant with the sentence that comes before.

@MayImilae What do you think would be a good way to include the 2021 wording?

JosJuice:
Ohh, wait, do you mean a source code comment for the 2021 thing? In that case I'll just throw it in there right away.

MayImilae:
Mmmhm, I meant source code comment for the 2021 thing just to give future devs an idea of what "powerful" meant there.

mbc07:
May's description LGTM

JosJuice:
The description is now:
> Ubershaders will always be used. Provides a near stutter-free experience at the cost of very high GPU performance requirements.
>
> **Don't use this unless you encountered stuttering with Hybrid Ubershaders and have a very powerful GPU.**

blaahaj:
3/4 options involve "specialized" shaders. However, only 1/4 options says "specialized"?

JMC47:
This is an improvement over the current situation.  Things we can do in the future is try to better describe what the modes do in the options since they do use mixes of the shaders.  There's no such thing as a Hybrid Ubershader - it's a simplification of a mode that combines ubershaders and specialized shaders.

Anyway, this has languished long enough.  Let's hit the green button and argue about more updates to it in a separate pull request.

PR #8418, opened 2019-10-20
Core: Replace usages of StringFromFormat with fmt where applicable

lioncash:
Migrates usages of StringFromFormat over to fmt.

### DIFF
--- a/Source/Core/Core/Analytics.cpp
+++ b/Source/Core/Core/Analytics.cpp
@@ -1,12 +1,13 @@
 #include "Core/Analytics.h"
 
 #include <array>
-#include <cinttypes>
-#include <mbedtls/sha1.h>
 #include <memory>
 #include <mutex>
 #include <string>
 #include <vector>
+
+#include <fmt/format.h>
+#include <mbedtls/sha1.h>
 
 #if defined(_WIN32)
 #include <windows.h>
@@ -21,7 +22,6 @@
 #include "Common/CPUDetect.h"
 #include "Common/CommonTypes.h"
 #include "Common/Random.h"
-#include "Common/StringUtil.h"
 #include "Common/Timer.h"
 #include "Common/Version.h"
 #include "Core/Config/MainSettings.h"
@@ -87,7 +87,7 @@ void DolphinAnalytics::GenerateNewIdentity()
 {
   const u64 id_high = Common::Random::GenerateValue<u64>();
   const u64 id_low = Common::Random::GenerateValue<u64>();
-  m_unique_id = StringFromFormat("%016" PRIx64 "%016" PRIx64, id_high, id_low);
+  m_unique_id = fmt::format("{:016x}{:016x}", id_high, id_low);
 
   // Save the new id in the configuration.
   SConfig::GetInstance().m_analytics_id = m_unique_id;
@@ -104,7 +104,7 @@ std::string DolphinAnalytics::MakeUniqueId(std::string_view data) const
   std::string out;
   for (int i = 0; i < 8; ++i)
   {
-    out += StringFromFormat("%02hhx", digest[i]);
+    out += fmt::format("{:02x}", digest[i]);
   }
   return out;
 }
--- a/Source/Core/Core/ConfigLoaders/NetPlayConfigLoader.cpp
+++ b/Source/Core/Core/ConfigLoaders/NetPlayConfigLoader.cpp
@@ -6,6 +6,8 @@
 
 #include <memory>
 
+#include <fmt/format.h>
+
 #include "Common/CommonPaths.h"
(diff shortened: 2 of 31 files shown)


Source/Core/Core/HW/GCMemcard/GCMemcardDirectory.cpp
```diff
@@ -627,15 +628,14 @@ void GCMemcardDirectory::FlushToFile()
 
           if (gci.IsGood())
           {
-            Core::DisplayMessage(
-                StringFromFormat("Wrote save contents to %s", m_saves[i].m_filename.c_str()), 4000);
+            Core::DisplayMessage(fmt::format("Wrote save contents to %s", m_saves[i].m_filename),
```
jordan-woyak: s/`%s`/`{}`/

Source/Core/Core/ConfigManager.cpp
```diff
@@ -496,12 +498,12 @@ void SConfig::LoadCoreSettings(IniFile& ini)
   core->Get("SlotB", (int*)&m_EXIDevice[1], ExpansionInterface::EXIDEVICE_NONE);
   core->Get("SerialPort1", (int*)&m_EXIDevice[2], ExpansionInterface::EXIDEVICE_NONE);
   core->Get("BBA_MAC", &m_bba_mac);
-  for (int i = 0; i < SerialInterface::MAX_SI_CHANNELS; ++i)
+  for (size_t i = 0; i < std::size(m_SIDevice); ++i)
   {
-    core->Get(StringFromFormat("SIDevice%i", i), (u32*)&m_SIDevice[i],
+    core->Get(fmt::format("SIDevice{}", i), (u32*)&m_SIDevice[i],
```
Warepire: Since you already edited this line, there's an old-school C-cast here.

Source/Core/Core/Analytics.cpp
```diff
@@ -87,7 +88,7 @@ void DolphinAnalytics::GenerateNewIdentity()
 {
   const u64 id_high = Common::Random::GenerateValue<u64>();
   const u64 id_low = Common::Random::GenerateValue<u64>();
-  m_unique_id = StringFromFormat("%016" PRIx64 "%016" PRIx64, id_high, id_low);
```
BhaaLseN: Do we still need the `cinttypes` include with this `PRIx64` gone?

Source/Core/Core/CoreTiming.cpp
```diff
@@ -396,8 +396,7 @@ std::string GetScheduledEventsSummary()
   std::sort(clone.begin(), clone.end());
   for (const Event& ev : clone)
   {
-    text += StringFromFormat("%s : %" PRIi64 " %016" PRIx64 "\n", ev.type->name->c_str(), ev.time,
```
BhaaLseN: Same down here for `cinttypes`

lioncash: Yes, it's still necessary for log calls within LogPendingEvents

Source/Core/Core/ConfigManager.cpp
lioncash: Done, I've added code that allows removing the cast entirely.

Source/Core/Core/ConfigLoaders/NetPlayConfigLoader.cpp
```diff
@@ -100,10 +102,10 @@ class NetPlayConfigLayerLoader final : public Config::ConfigLayerLoader
         layer->Set(Config::MAIN_GCI_FOLDER_A_PATH_OVERRIDE, path + "Card A");
         layer->Set(Config::MAIN_GCI_FOLDER_B_PATH_OVERRIDE, path + "Card B");
 
-        const std::string file = File::GetUserPath(D_GCUSER_IDX) + GC_MEMCARD_NETPLAY + "%c." +
-                                 m_settings.m_SaveDataRegion + ".raw";
-        layer->Set(Config::MAIN_MEMCARD_A_PATH, StringFromFormat(file.c_str(), 'A'));
-        layer->Set(Config::MAIN_MEMCARD_B_PATH, StringFromFormat(file.c_str(), 'B'));
+        const std::string file = fmt::format("{}{}{}.{{}}.raw", File::GetUserPath(D_GCUSER_IDX),
```
leoetlino: I think this should be

```suggestion
        const std::string file = fmt::format("{}{}{{}}.{}.raw", File::GetUserPath(D_GCUSER_IDX),
```

instead?

(IMO this is a bit hard to read... Any thoughts on having a small function/lambda that takes the memcard letter and returns the whole path?)

lioncash: Done

PR #8420, opened 2019-10-21
ButtonManager: Fix handling of empty device id.

phcoder:
Device id is "" on ChromeOS when using gamepad

### DIFF
--- a/Source/Core/InputCommon/ControllerInterface/Touch/ButtonManager.cpp
+++ b/Source/Core/InputCommon/ControllerInterface/Touch/ButtonManager.cpp
@@ -10,6 +10,7 @@
 
 #include "Common/FileUtil.h"
 #include "Common/IniFile.h"
+#include "Common/StringUtil.h"
 #include "Common/Thread.h"
 #include "InputCommon/ControllerInterface/Touch/ButtonManager.h"
 
@@ -630,7 +631,7 @@ void Init(const std::string& game_id)
       config << CONFIG_STRINGS[a] << "_" << pad_id;
       BindType type;
       int bindnum;
-      char dev[128];
+      char dev[128] {};
       bool hasbind = false;
       char modifier = '+';
       std::string value;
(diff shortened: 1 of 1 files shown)


CookiePLMonster:
@dolphin-emu-bot rebuild

Source/Android/jni/ButtonManager.cpp
```diff
@@ -586,13 +586,19 @@ void Init(const std::string& gameId)
       {
         hasbind = true;
         type = BIND_AXIS;
-        sscanf(value.c_str(), "Device '%127[^\']'-Axis %d%c", dev, &bindnum, &modifier);
+        if (std::string::npos != value.find("Device ''"))
```
CookiePLMonster: Although it's not mentioned in the Coding Style guide, Yoda conditions are not used in the codebase.

Source/Android/jni/ButtonManager.cpp
```diff
@@ -586,13 +586,19 @@ void Init(const std::string& gameId)
...
+          sscanf(value.c_str(), "Device ''-Axis %d%c", &bindnum, &modifier);
+        else
+          sscanf(value.c_str(), "Device '%127[^\']'-Axis %d%c", dev, &bindnum, &modifier);
       }
       else if (std::string::npos != value.find("Button"))
       {
         hasbind = true;
         type = BIND_BUTTON;
-        sscanf(value.c_str(), "Device '%127[^\']'-Button %d", dev, &bindnum);
+        if (std::string::npos != value.find("Device ''"))
```
CookiePLMonster: Ditto

Source/Android/jni/ButtonManager.cpp
phcoder: There is exactly the same code few lines higher:
      if (std::string::npos != value.find("Axis"))
Should I be consistent with the file or switch arguments?

CookiePLMonster: I guess you can keep it as is then.

CookiePLMonster:
@dolphin-emu-bot rebuild

Source/Android/jni/ButtonManager.cpp
JosJuice: Is there any particular reason why you're using `find` instead of `StringBeginsWith` (available in Common/StringUtil.h)?

phcoder: Changed. I will have to retest the patch

phcoder:
> This solution seems odd. I'd have to dig deeper into the code but I feel like it would be cleaner to just overwrite an empty device name with "Device" (or anything) where the device is created.

I see it as a problem with sscanf and just fixing it as near as possible to the cause. We can't change sscanf but we can add this fix on top of it.
But I don't really care which solution to use, I just want Dolphin to work well on Chromebook

PR #8421, opened 2019-10-21
EmulationActivity: Disable fullscreen immersive after pressing back

phcoder:
On Chromebook when in fullscreen immersive I can't open the toolbar.
I'm not sure if I swipe wrong or something but in any case it's not intuitive.
Showing toolbar when pressing back solves te problem neatly

### DIFF
--- a/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/activities/EmulationActivity.java
+++ b/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/activities/EmulationActivity.java
@@ -381,7 +381,12 @@ public final class EmulationActivity extends AppCompatActivity
       {
         backPressedOnce = true;
         Toast.makeText(this, "Press back again to exit", Toast.LENGTH_LONG).show();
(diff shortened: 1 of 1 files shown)


JosJuice:
I've been considering having the back button just open the menu directly, as a popup rather than a dropdown (and then there'd be a quit option in the menu, unlike now where you press back twice to quit). I think it already works that way on Android TV, but I could be misremembering. Do you think that would make more or less sense than what is in this pull request?

JosJuice:
@dolphin-emu-bot rebuild

phcoder:
Yes. Back button opening the menu would solve the problem as well. Any reason not to do it on all Android devices?

JosJuice:
No, I'm not aware of any reason.

JosJuice:
@dolphin-emu-bot rebuild

leoetlino:
What's the status of this PR? Do you intend to make the back button open the menu?

JosJuice:
I did it in PR #9004.

PR #8422, opened 2019-10-21
Fixup GameListModel::UpdateGame and columnCount

CookiePLMonster:
- dataChanged takes "first" and "last" indices inclusive, not exclusive
- columnCount should return 0 for valid parents

This fixes a non-fatal assertion in debug builds while generating gamelist cache. Delete your `Cache` directory and run a debug compilation of Dolphin to repro.

### DIFF
--- a/Source/Core/DolphinQt/GameList/GameListModel.cpp
+++ b/Source/Core/DolphinQt/GameList/GameListModel.cpp
@@ -28,7 +28,7 @@ GameListModel::GameListModel(QObject* parent) : QAbstractTableModel(parent)
   connect(&Settings::Instance(), &Settings::PathRemoved, &m_tracker, &GameTracker::RemoveDirectory);
   connect(&Settings::Instance(), &Settings::GameListRefreshRequested, &m_tracker,
           &GameTracker::RefreshAll);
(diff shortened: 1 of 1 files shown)


Source/Core/DolphinQt/GameList/GameListModel.cpp
```diff
@@ -199,6 +199,8 @@ int GameListModel::rowCount(const QModelIndex& parent) const
 
 int GameListModel::columnCount(const QModelIndex& parent) const
 {
+  if (parent.isValid())
+    return 0;
```
BhaaLseN: This looks really weird to me (as if it should always return zero from now on). _When_ is `parent` ever _not_ valid? Why is it being passed if it could ever be _not_ valid?

CookiePLMonster: No idea honestly. I based it solely on the documentation and a similar code in `rowCount`:

https://doc.qt.io/qt-5/qabstractitemmodel.html#columnCount

`Note: When implementing a table based model, columnCount() should return 0 when the parent is valid.`

Our model is table based, so I decided to follow the docs.

PR #8424, opened 2019-10-22
DolphinQt: Cleanup after moving CPU Emulation Engine to Advanced tab

JosJuice:


ethteck:
Ah, oops... Thanks, this is better.

PR #8425, opened 2019-10-23
VolumeVerifier: Improve handling of Datel Wii Freeloader discs

JosJuice:


Source/Core/DiscIO/VolumeVerifier.cpp
```diff
@@ -191,10 +193,13 @@ bool VolumeVerifier::CheckPartition(const Partition& partition)
...
-  CheckCorrectlySigned(
-      partition,
-      StringFromFormat(Common::GetStringT("The %s partition is not correctly signed.").c_str(),
-                       name.c_str()));
+  if (!m_is_datel)
+  {
+    CheckCorrectlySigned(
+        partition,
+        StringFromFormat(Common::GetStringT("The %s partition is not correctly signed.").c_str(),
+                         name.c_str()));
```
CookiePLMonster: Do you think it makes sense to also migrate this to fmt?

JosJuice: I'd say it's better to save that for a PR that migrates all of VolumeVerifier's `StringFromFormat` calls at once.

Source/Core/DiscIO/VolumeVerifier.cpp
```diff
@@ -541,9 +546,26 @@ bool VolumeVerifier::CheckPartition(const Partition& partition)
...
+      m_blocks.emplace_back(BlockToVerify{partition, offset, i});
+
+    m_block_errors.emplace(partition, 0);
+  }
+
   const DiscIO::FileSystem* filesystem = m_volume.GetFileSystem(partition);
   if (!filesystem)
   {
+    if (m_is_datel)
+      return true;  // Datel Wii Freeloader has an invalid FST in its only partition
```
lioncash: Comments should generally be placed above the respective line instead of off to the side for ease of refactoring.

PR #8426, opened 2019-10-23
VolumeVerifier: Improve region/game ID mismatch checking

JosJuice:


Source/Core/DiscIO/VolumeVerifier.cpp
```diff
@@ -539,20 +539,26 @@ void VolumeVerifier::CheckMisc()
   }
 
   const Region region = m_volume.GetRegion();
-  const Platform platform = m_volume.GetVolumeType();
+
+  constexpr std::string_view GAMECUBE_PLACEHOLDER_ID = "RELSAB";
+  constexpr std::string_view WII_PLACEHOLDER_ID = "RABAZZ";
```
BhaaLseN: I haven't seen those before, where are they from and when would they appear?

JosJuice: Games under development and other discs that haven't been released to retail.

PR #8316, opened 2019-08-17
WASAPI code improvements

CookiePLMonster:
This PR brings numerous improvements to WASAPI code in Dolphin, improving code readability as well as fixing numerous issues.

As a preparation, [WIL](https://github.com/microsoft/wil) has been added, since it's very useful for ensuring proper RAII-based destruction for several WinAPI types - in my case it was proper scoped calls to `CoUninitialize` and proper destruction of previously leaking (in some cases) `PROPVARIANT`.

All COM objects have been moved to use WRL, which ensures their proper release. Previously, there were numerous code paths where functions would return without cleaning up their COM objects, effectively leaking references.

`device->GetId(&device_id);` was present in several places, never using the return value nor freeing it - leaking memory for no reason.

Thread synchronization model for WASAPI Handler thread has been simplified by not detaching said thread. Not only it improved code readability, but also fixed a spinlock occuring if `WASAPIStream::SetRunning(true)` failed in **any** way - a following call to `WASAPIStream::SetRunning(false)` would then spin endlessly checking for `m_stopped`. Now this boolean is gone, and so is the problem.

`std::string_view` is now used where applicable. I opted for constructing a `std::string` out of `std::string_view` in `HandleWinAPI`, because it happens in code path which is taken relatively rarely - in a perfect case it's never taken.

I modified `SoundLoop` not to do volume adjustment unless absolutely necessary, saving on a relatively costly loop. For 100% volume, volume adjustment is not needed, and for 0% volume a `AUDCLNT_BUFFERFLAGS_SILENT` is now specified instead of zeroing the buffer. I also removed several nullptr checks which were not thread safe anyway, so only brought a false sense of safety.

PR #8427, opened 2019-10-23
Add Windows Implementation Libraries as an external library

CookiePLMonster:
Previously a part of #8316 but I would really like to use WIL in other PR's to get RAII wrappers around WinAPI resources.

Quoting the relevant part of this PR:

> As a preparation, [WIL](https://github.com/microsoft/wil) has been added, since it's very useful for ensuring proper RAII-based destruction for several WinAPI types - in my case it was proper scoped calls to `CoUninitialize` and proper destruction of previously leaking (in some cases) `PROPVARIANT`.

WIL also includes proper `unique_ptr`-like wrappers around eg. module handles, file handles, registry handles, `CoTaskMem` memory... etc.


Source/Core/AudioCommon/CMakeLists.txt
```diff
@@ -71,6 +71,11 @@ if(WIN32)
 
     WASAPIStream.cpp
     WASAPIStream.h
+  target_include_directories(audiocommon PRIVATE
```
Warepire: Maybe this part should remain with audiocommon or be moved somewhere more generic?

CookiePLMonster: I don't know if we have a "more generic" place for this, so I just removed it from that PR and will make it part of WASAPI PR.

CookiePLMonster:
As per the discussion on IRC I'll modify cmake to make WIL a dependency library so it's easier to use than what it is now. Will do it around Saturday.

CookiePLMonster:
Yesterday I spent way more time than I'd like trying to fix cmake project on Windows and looks like it's just fundamentally broken - PCH seems to be set up incorrectly and I failed to fix it fully.

Because of this I am unable to test and come up with a reasonable way to integrate WIL there. Any ideas how to proceed?

Tagging @lioncash since they're usually involved in cmake discussions.

CookiePLMonster:
As per discussion on IRC earlier, I now added WIL as a regular directory and not a submodule.

I did notice one problem with that however - now when it's **not** a submodule, it's not obvious what exact revision it contains...

shuffle2:
> I did notice one problem with that however - now when it's **not** a submodule, it's not obvious what exact revision it contains...
In the past dolphin would work around this by leaving a .txt file a level above, stating the version of the external resource, if changes were made, etc. Mainly to simplify process of updating the resource in the future.

CookiePLMonster:
Rebased on latest master.

PR #8428, opened 2019-10-23
InputCommon: Add hotkey support to input expressions.

jordan-woyak:
I've added support for hotkeys in our input expressions.

image

Fixes issue: https://bugs.dolphin-emu.org/issues/9320

Demonstration:
https://giant.gfycat.com/RapidScientificArgali.webm

Note the following improvements:
- Button combos are now detected and do not require use of the "advanced" dialog.
  - Simultaneous and successive presses also produce & and | operator usages.
- Hotkeys suppress other references to the same input.
  - Ugly `!Alt & Return` mappings are no longer required to prevent regular press activation.
- Hotkeys require pressing of modifers first.
  - e.g. Ctrl+O does not work unless you press Ctrl first.

The current syntax is of the form `@(Ctrl+Shift+R)` but I am open to suggestions.

Any input can be used as a "modifier". It is not limited to the typical ctrl, shift, alt.
This will make mapping button combinations on gamepads easier for users.

There is a disadvantage.. but I do not think it is horrendous.
`@(Ctrl+O)` Does not know `Shift` is a modifer unless a `@(Shift+O)` hotkey also exists.
In other words.. pressing Ctrl+Shift+O will present the "open file" dialog.
If you map something else to Shift+O, however, then it will not.


Source/Core/DolphinQt/Config/Mapping/MappingCommon.cpp
```diff
@@ -83,14 +83,24 @@ QString DetectExpression(QPushButton* button, ciface::Core::DeviceContainer& dev
...
+
+    if (!full_expression.isEmpty())
+      full_expression += QChar::fromLatin1('+');
+
+    full_expression += MappingCommon::GetExpressionForControl(
+        QString::fromStdString(input->GetName()), device_qualifier, default_device, quote);
+  }
+
+  if (detections.size() > 1)
+    return QString::fromUtf8("@(") + std::move(full_expression) + QString::fromUtf8(")");
```
lioncash:
```suggestion
    return QStringLiteral("@(%1)").arg(full_expression);
```

Is a slight bit more concise (and signals the strings are indeed intended to be non-translatable).

Source/Core/InputCommon/ControlReference/ExpressionParser.cpp
```diff
@@ -366,6 +399,80 @@ class VariableExpression : public Expression
   ControlState* m_value_ptr{};
 };
 
+class HotkeyExpression : public Expression
+{
+public:
+  HotkeyExpression(std::vector<std::unique_ptr<ControlExpression>> _inputs)
```
lioncash:
```suggestion
  HotkeyExpression(std::vector<std::unique_ptr<ControlExpression>> inputs)
```

Source/Core/InputCommon/ControlReference/ExpressionParser.cpp
```diff
@@ -366,6 +399,80 @@ class VariableExpression : public Expression
...
+  }
+
+  ControlState GetValue() const override
+  {
+    if (m_inputs.empty())
+      return 0;
+
+    const bool modifiers_pressed = std::all_of(m_inputs.begin(), std::prev(m_inputs.end()),
+                                               [](const std::unique_ptr<ControlExpression>& input) {
+                                                 // TODO: kill magic number.
```
lioncash: Was this intended to be resolved in this set of changes or a future one?

jordan-woyak: Yeah. Another PR moves the relevant constant to a header file. Didn't feel like dealing with merge conflicts.

Source/Core/InputCommon/ControlReference/ExpressionParser.cpp
```diff
@@ -21,6 +24,55 @@ namespace ciface::ExpressionParser
...
+      (*func)();
+    }
+  };
+
+  using Suppressor = std::unique_ptr<std::function<void()>, InvokingDeleter>;
+
+  bool IsSuppressed(Device::Input* input) const
+  {
+    // Input is suppressed if it exists in the map at all.
+    return m_suppressions.lower_bound({input, nullptr}) !=
```
iwubcode, line 48: I feel like I should know this but what are you doing here exactly?  How are you lower-bounding the pointer and why are you comparing against the lower-bound of the pair of pointers with another +1?

Source/Core/InputCommon/ControllerInterface/Device.cpp
```diff
@@ -249,18 +302,54 @@ bool DeviceContainer::HasConnectedDevice(const DeviceQualifier& qualifier) const
...
+                                  std::chrono::milliseconds initial_wait,
+                                  std::chrono::milliseconds confirmation_wait,
+                                  std::chrono::milliseconds maximum_wait) const
+    -> std::vector<InputDetection>
 {
   struct InputState
   {
-    ciface::Core::Device::Input& input;
-    ControlState initial_state;
+    InputState(ciface::Core::Device::Input* input_) : input{input_} { stats.Push(0.0); }
```
iwubcode, line 318: Minor but `explicit InputState` ?

iwubcode:
Very nice @jordan-woyak !  No real concerns with the code on an initial pass.

I did a bit of testing.  Seems like a massive step up.  For button combos, the hotkey vs "<key1>|<key2>" seems to give me what I expect part of the time.  I need to increase my dexterity :)

I plan on coming back and testing the hotkey suppression + giving the code another once over before the final review.

Source/Core/InputCommon/ControlReference/ExpressionParser.cpp
```diff
@@ -374,6 +471,90 @@ class VariableExpression : public Expression
...
+
+      // If some other hotkey suppressed us, require a release of final input to be ready again.
+      if (is_suppressed)
+        m_is_ready = false;
+
+      // Our modifiers are active. Pass through the final input.
+      if (m_is_ready)
+        return final_input_state;
+    }
+    else
```
Filoppi, line 515: Not sure I fully understand this yet, but shouldn't the HotkeyExpression be blocked if the "final_input" (the only non modifier input) is pressed before the modifiers? And stay blocked until it is reseased.
So that if you press O an then CTRL while O is still pressed, it wouldn't work. This doesn't seem to be the way now.

There is no need for time stamps on inputs, just make sure the modifiers were pressed in the same frame or any frame before (by frame, I mean a refresh of the cached input values from the control devices).

Source/Core/InputCommon/ControlReference/ExpressionParser.cpp
jordan-woyak: They're just sorted based on the pointer values. The `lower_bound` trick is to test if any pair in the map has `input` for the `first` value.

Source/Core/InputCommon/ControlReference/ExpressionParser.cpp
jordan-woyak: Yes. If the "final_input" is pressed before the modifiers then `m_is_ready` will not be set.

Filoppi: This is weird though, the different modifiers aren't required to be pressed within any order or time, while the final_input needs to be pressed at least one frame after the rest. I guess this is mostly for the Hotkeys and you'd be insane to use a modifier as a game control, but if Dolphin was running slow then it might be hard to get an input with an hotkey to pass. I think they should also be accepted if both the modifiers and the final_input were pressed in the same frame.

Another problem with this is this:
-Press O
-Press CTRL
-Release O
-Press O again
In this case, you'd expect the final CTRL+O to be accepted, but it wouldn't. That's why i had not looked inside the "if (modifiers_pressed)" condition, I expected the "final_input" blocking to be handled outside of it.

jordan-woyak: The final "CTRL+O" **will** be accepted in your example situation. `m_is_ready` will be set when "O" is released.

jordan-woyak: I would agree that pressing modifiers and the final input in the same frame could/should be accepted.
I think this might require an additional state variable.

Filoppi: More like wronglythinking

Filoppi: No, if you just move the code that sets `m_is_ready` outside of `modifiers_pressed`, and rename it `m_is_blocked`, it should work.
Basically you block everything if `final_input` is pressed before `modifiers_pressed`, until it's released.

jordan-woyak: I'll make the change when I get around to giving this a rebase.

Source/Core/DolphinQt/Config/Mapping/MappingCommon.cpp
```diff
@@ -102,10 +109,103 @@ void TestOutput(QPushButton* button, OutputReference* reference)
   QApplication::processEvents();
 
   reference->State(1.0);
-  Common::SleepCurrentThread(OUTPUT_TEST_TIME);
+  Common::SleepCurrentThread(OUTPUT_TEST_TIME_MS);
```
iwubcode, line 112: Not a big deal and not sure what advantages `SleepCurrentThread` has but if we used `std::this_thread::sleep_for(...)` we could use a chrono time, which would make `OUTPUT_TEST_TIME` consistent with the rest of the values in `MappingCommon`

jordan-woyak: I'll change it.

Source/Core/InputCommon/ControllerInterface/Device.cpp
```diff
@@ -301,27 +390,59 @@ DeviceContainer::DetectInput(u32 wait_ms, const std::vector<std::string>& device
...
+        auto& input_state = device_state.input_states[i];
+        input_state.Update();
+
+        if (input_state.IsPressed())
+        {
+          input_state.is_ready = false;
+
+          // Digital presses will evaluate as 1 here.
+          // Analog presses will evaluate greater than 1.
+          const auto smoothness =
```
iwubcode, line 423: Do you foresee a need for this 'smoothness' value or would a boolean / enumeration dictating whether the key is analog or digital be clearer?

iwubcode: (I saw you you already checked for 'digital' keys earlier)

jordan-woyak: I guess I left it as `smoothness` instead of an `is_analog` boolean because the check isn't foolproof. If you quickly pull your trigger so it's never seen as half-pressed then `smoothness` will be 1.
Also I might tweak input detect logic a bit more relating to `SPURIOUS_TRIGGER_COMBO_THRESHOLD`. If users are finding their button combos not mapping because of that logic, then `smoothness` could be increased to require a slow-pull to consider a combo "spurious".

Source/Core/InputCommon/ControlReference/ExpressionParser.cpp
```diff
@@ -222,12 +284,46 @@ class ControlExpression : public Expression
...
+}
+
+HotkeySuppressions::Suppressor
+HotkeySuppressions::MakeSuppressor(const Modifiers* modifiers,
+                                   const std::unique_ptr<ControlExpression>* final_input)
+{
+  for (auto& modifier : *modifiers)
+    ++m_suppressions[{(*final_input)->GetInput(), modifier->GetInput()}];
+
+  return Suppressor(std::make_unique<std::function<void()>>([this, modifiers, final_input]() {
```
iwubcode, line 320: This is fine but I was surprised Dolphin doesn't already have a scoped-lambda type thing.

iwubcode (approved):
Code LGTM

JMC47:
Well, I needed something big for the GUI section anyway and hotkeys are fun.

i30817:
I'm not sure i should ask this here, but since i don't have a account in the real bug report page a merged issue that touches this recently is the best i can do.

Anyway: would it be possible to make a way to 'lift a finger' automatically recenter the wii pointer?

I'm doing something where the touchpad is the wii pointer and a modifer(right touchpad button)+the touchpad is the tilt.

This works fine, but when i stop using the tilt shortcuts, the wii cursor shortcut activate and jump the mouse to where the 'cursor' is, while i'd prefer it was kept in the same position as it was move the tilt movement.

Part of the problem is that the cursor is not recentered after i release the modifier. 

I wonder if it's possible to add a 'hotkey' which is not part of these that triggers when a button is pressed or released and calls functions like 'recenter wii cursor' that have nothing to do with current interface functions.

jordan-woyak:
@i30817 Ask on the forums. Also I can't figure out what you're asking for.

i30817:
You're right, i probably should make sure it's actually a important issue in a game before i ask for this.

PR #8429, opened 2019-10-25
Create necessary folder when using Redump.org integration

JosJuice:
This was making it impossible to use the Redump.org integration without first manually creating a Redump folder in the Cache folder. https://bugs.dolphin-emu.org/issues/11885

### DIFF
--- a/Source/Core/Common/FileUtil.cpp
+++ b/Source/Core/Common/FileUtil.cpp
@@ -824,6 +824,7 @@ static void RebuildUserDirectories(unsigned int dir_index)
 
   case D_CACHE_IDX:
     s_user_paths[D_COVERCACHE_IDX] = s_user_paths[D_CACHE_IDX] + COVERCACHE_DIR DIR_SEP;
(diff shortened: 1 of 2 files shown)


BhaaLseN:
Should we really be returning `DownloadStatus::Success` when writing the file failed?

JosJuice:
That's probably not perfect... But returning `DownloadStatus::Fail` instead would make the GUI display "Failed to connect to Redump.org", which isn't true. So I'm not sure if there's a good solution for this unless we want to add an extra translatable string for the hopefully uncommon cases of folder permissions being messed up or the disk being full.

xerxespersrex:
This PR solves issue 11885 for me.

JosJuice:
Thanks for catching that – it should be fixed by the commit I just pushed.

xerxespersrex:
Yep, everything works perfectly now for me, it's created in ~/.cache/dolphin-emu/Redump/ like it should be and verification works fine.

PR #8430, opened 2019-10-25
InputCommon: Change "EOF" to "end of expression" in user facing string

JosJuice:


jordan-woyak:
I agree that "EOF" should probably be something better like: "end of expression".

BhaaLseN:
Even better! 👍

PR #8432, opened 2019-10-26
DolphinQt: Make HotkeyScheduler call UpdateInput when hotkeys are disabled.

jordan-woyak:
When emulation is not running our controller interface is updated by the hotkey scheduler.

A filter is installed to prevent hotkeys from triggering when configuring them.
This was preventing the mapping buttons from showing their activation by turning bold.

`UpdateInput` is now called before checking if hotkeys are filtered.

Fixes: https://bugs.dolphin-emu.org/issues/11835


Miksel12:
I can confirm that it fixes the issue.

PR #8433, opened 2019-10-26
DolphinQt: Call OnEmulationStateChanged when creating config window panes

JosJuice:
Otherwise UI elements won't be disabled correctly if the config window is first opened while a game is running.

### DIFF
--- a/Source/Core/DolphinQt/Settings/WiiPane.cpp
+++ b/Source/Core/DolphinQt/Settings/WiiPane.cpp
@@ -48,6 +48,7 @@ WiiPane::WiiPane(QWidget* parent) : QWidget(parent)
   LoadConfig();
   ConnectLayout();
   ValidateSelectionState();
+  OnEmulationStateChanged(Core::GetState() != Core::State::Uninitialized);
 }
 
 void WiiPane::CreateLayout()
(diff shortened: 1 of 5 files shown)


Techjar:
This seems like a hacky workaround. OnEmulationStateChanged should only be called when the state actually changes. Couldn't calling it under other circumstances potentially have other side effects?

JosJuice:
The OnEmulationStateChanged functions in these files have no side effects that we want to avoid. All they do is enable or disable UI elements based on the argument.

Source/Core/DolphinQt/Settings/WiiPane.cpp
```diff
@@ -92,6 +93,10 @@ void WiiPane::ConnectLayout()
   connect(m_wiimote_ir_sensitivity, &QSlider::valueChanged, this, &WiiPane::OnSaveConfig);
   connect(m_wiimote_speaker_volume, &QSlider::valueChanged, this, &WiiPane::OnSaveConfig);
   connect(m_wiimote_motor, &QCheckBox::toggled, this, &WiiPane::OnSaveConfig);
+
+  // Emulation State
+  connect(&Settings::Instance(), &Settings::EmulationStateChanged, this,
```
CookiePLMonster: `this` is not used here, since it's connecting a lambda.

PR #8436, opened 2019-10-27
VideoInterface: Don't crash when running Wii Freeloader

JosJuice:
Datel's Wii Freeloaders set `m_PictureConfiguration.WPL` to 0 for a while. Not sure if the fix in this commit is a proper fix or just a hack, since I'm not very familiar with this code.

With this change, it's possible to run a Wii Freeloader if you are running an old enough version of the Wii Menu, but the "coloured bars" that Datel reference in their documentation never show up. The screen just freezes for a few seconds instead.

### DIFF
--- a/Source/Core/Core/HW/VideoInterface.cpp
+++ b/Source/Core/Core/HW/VideoInterface.cpp
@@ -666,7 +666,9 @@ static void LogField(FieldType field, u32 xfb_address)
 static void BeginField(FieldType field, u64 ticks)
 {
   // Could we fit a second line of data in the stride?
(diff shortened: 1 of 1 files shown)


BhaaLseN:
Since we're not sure if this is correct; should we leave a comment there regarding known titles that set it to zero?

JosJuice:
Done.

MayImilae:
We're working on the Progress Report and having some trouble with some of the details of this one. Could you please ping either me or JMC  on IRC and answer a few questions? Thanks!

PR #8437, opened 2019-10-27
Disable render to main when using batch mode

JosJuice:
https://bugs.dolphin-emu.org/issues/11888

### DIFF
--- a/Source/Core/UICommon/CommandLineParse.cpp
+++ b/Source/Core/UICommon/CommandLineParse.cpp
@@ -22,7 +22,7 @@ class CommandLineConfigLayerLoader final : public Config::ConfigLayerLoader
 {
 public:
   CommandLineConfigLayerLoader(const std::list<std::string>& args, const std::string& video_backend,
-                               const std::string& audio_backend)
+                               const std::string& audio_backend, bool batch)
       : ConfigLayerLoader(Config::LayerType::CommandLine)
   {
     if (!video_backend.empty())
@@ -30,6 +30,11 @@ public:
 
     if (!audio_backend.empty())
       m_values.emplace_back(Config::MAIN_DSP_HLE.location, ValueToString(audio_backend == "HLE"));
+
+    // Batch mode hides the main window, and render to main hides the render window. To avoid a
+    // situation where we would have no window at all, disable render to main when using batch mode.
+    if (batch)
+      m_values.emplace_back(Config::MAIN_RENDER_TO_MAIN.location, ValueToString(false));
 
     // Arguments are in the format of <System>.<Section>.<Key>=Value
     for (const auto& arg : args)
@@ -98,7 +103,7 @@ std::unique_ptr<optparse::OptionParser> CreateParser(ParserOptions options)
     parser->add_option("-l", "--logger").action("store_true").help("Open the logger");
     parser->add_option("-b", "--batch")
         .action("store_true")
-        .help("Run Dolphin without the user interface (Requires --exec)");
+        .help("Run Dolphin without the user interface (Requires --exec or --nand-title)");
     parser->add_option("-c", "--confirm").action("store_true").help("Set Confirm on Stop");
   }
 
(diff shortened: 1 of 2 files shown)


iwubcode:
Interesting.  I wonder if this is incorrect usage though?

Typically if using batch, you're executing from the command line (say from a front-end).  I know for my front-end, I was using batch + render to main, to get around the multiple windows created.  I hadn't upgraded since #8308 came out but I feel like the whole point was to ensure batch runs headless (without the Qt nonsense).  Should we error instead?

iwubcode:
Maybe this should instead set `Settings::Instance().SetBatchModeEnabled` to true if the render-to-main isn't set?  (then add that to the command line description for 'batch')

JosJuice:
I think disallowing users from using batch when render to main is enabled would be a step backwards. After all, the main point of batch is that Dolphin exits when the game exits, and that works without problems even if you're using render to main.

iwubcode:
>After all, the main point of batch is that Dolphin exits when the game exits, and that works without problems even if you're using render to main.

Yeah that's fine but the command line description mentions going "headless" which isn't true if render-to-main is set..

JosJuice:
Oh, I see. I didn't realize that PR #8308 had changed the description for --batch.

JosJuice:
On IRC, @CookiePLMonster had the idea of instead disabling render to main when batch is used. I think this makes more sense than disabling batch when render to main is used. What do others think about this?

JosJuice:
Done. I also changed the behavior of specifying batch without specifying a game – see the commit message for details.

Source/Core/UICommon/CommandLineParse.cpp
```diff
@@ -31,21 +31,27 @@ class CommandLineConfigLayerLoader final : public Config::ConfigLayerLoader
     if (!audio_backend.empty())
       m_values.emplace_back(Config::MAIN_DSP_HLE.location, ValueToString(audio_backend == "HLE"));
 
+    if (batch)
+      m_values.emplace_back(Config::MAIN_RENDER_TO_MAIN.location, ValueToString(false));
+
     // Arguments are in the format of <System>.<Section>.<Key>=Value
-    for (const auto& arg : args)
+    if (!args.empty())
```
CookiePLMonster: This feels a bit redundant, for loop's body will just never be entered if the container is empty.

JosJuice: I actually get a crash without this and I'm not sure why.

CookiePLMonster: With `args` being `std::list`!?!? Sounds like something is severely wrong and you may just be working around some major breakage. Race somewhere?

JosJuice: I've made a proper fix by not calling `options.all("config")` when not allowed.

Source/Core/UICommon/CommandLineParse.cpp
```diff
@@ -114,13 +114,13 @@ std::unique_ptr<optparse::OptionParser> CreateParser(ParserOptions options)
...
-  {
-    const std::list<std::string>& config_args = options.all("config");
-    Config::AddLayer(std::make_unique<CommandLineConfigLayerLoader>(
-        config_args, static_cast<const char*>(options.get("video_backend")),
-        static_cast<const char*>(options.get("audio_emulation"))));
-  }
+    config_args = options.all("config");
+
+  Config::AddLayer(std::make_unique<CommandLineConfigLayerLoader>(
+      config_args, static_cast<const char*>(options.get("video_backend")),
```
CookiePLMonster: `std::move(config_args)` - even **if** it doesn't result in a move now, it's worth futureproofing like this because we may change `AddLayer` to support moves in the future.

CookiePLMonster:
LGTM but I definitely want to run a test with this PR prior to accepting.

Source/Core/UICommon/CommandLineParse.cpp
```diff
@@ -31,6 +31,9 @@ class CommandLineConfigLayerLoader final : public Config::ConfigLayerLoader
     if (!audio_backend.empty())
       m_values.emplace_back(Config::MAIN_DSP_HLE.location, ValueToString(audio_backend == "HLE"));
 
+    if (batch)
```
BhaaLseN: Do you think we should leave a comment on _why_ batch mode disables render-to-main?

PR #8438, opened 2019-10-28
Change Wii Remote tilt tooltip string based on translator feedback

JosJuice:


BhaaLseN:
I understand the first one, but hardly the second one. Then again, for the second one, I didn't really understand the original either ¯\\\_(ツ)\_/¯
(The translation; not the comment.)

PR #8439, opened 2019-10-28
Android: Native motion controls

JosJuice:
Accelerometer controls seem to work, but I haven't done a lot of testing yet. Left to do:

- [x] Figure out why the IR pointer is gone now
- [x] Adjust for phone orientation
- [x] Enable Sideways Wii Remote when using Horizontal Wii Remote
- [x] Disable IMU IR automatically when gyroscope or accelerometer is not present?
- [x] Add a setting for disabling IMU IR? (see PR #8440)
- [x] Add a setting for disabling motion controls entirely? Could be good for gamepad users
- [x] Add a setting for disabling screen rotation? Too easy to trigger when using motion controls


JosJuice:
Seems like the buildbots don't run for PRs marked as drafts, so let's undraft this. Still a WIP, though.

JosJuice:
@dolphin-emu-bot rebuild

jordan-woyak:
FYI, Sideways/Upright Wii Remote should be already handled. The "IMU" data is transformed appropriately in WiimoteEmu.cpp. You should just have to make sure the Accel and Gyro data is oriented with the screen.

JosJuice:
> FYI, Sideways/Upright Wii Remote should be already handled.

Yes. What I meant but probably wasn't super clear about was that I need to make it possible to change those settings on Android. They currently aren't accessible in the GUI.

> With the current way the IMU data is handled, users will need their phone still and level to use the "pointer". This might be annoying. We might need an option to not use IMU data for producing the transform for IR camera simulation.

Yeah, a setting might be a good idea, so that you can control the IR using the touch screen without motion controls affecting anything. But I'm wondering, how still does it need to be? (Just trying to figure out if my problems with getting the cursor to show up at all could be related to this.)

rlnilsen:
PR #8440 adds an "enable" checkbox for motion controlled cursor to Qt. Hope it can be useful to this PR.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/utils/MotionListener.java
```diff
@@ -0,0 +1,88 @@
...
+import android.hardware.SensorManager;
+
+import org.dolphinemu.dolphinemu.NativeLibrary;
+import org.dolphinemu.dolphinemu.NativeLibrary.ButtonType;
+
+public class MotionListener implements SensorEventListener
+{
+  private SensorManager mSensorManager;
+  private Sensor mAccelSensor;
+  private Sensor mGyroSensor;
```
BhaaLseN: Do we want to mark those as `final`? They're set in the ctor and _shouldn't_ change after the fact.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/activities/EmulationActivity.java
```diff
@@ -867,6 +980,37 @@ private void chooseController()
...
+  {
+    final int[] orientationValues = getResources().getIntArray(R.array.orientationValues);
+    int initialChoice = mPreferences.getInt("emulationActivityOrientation",
+            ActivityInfo.SCREEN_ORIENTATION_LANDSCAPE);
+    int initialIndex = -1;
+    for (int i = 0; i < orientationValues.length; i++)
+    {
+      if (orientationValues[i] == initialChoice)
+        initialIndex = i;
+    }
```
BhaaLseN: I'm surprised Java has no _simple_ `indexOf` (or `find` or whatever) method for such a purpose.

The most concise one I could find (aside from a bazillion libraries) is
```java
int initialIndex = java.util.Arrays.asList(orientationValues).indexOf(initialChoice);
```
Thoughts on switching to that (potentially with our own, bazillion-th-plus-one utility class)?

JosJuice: I was surprised too...

According to this Stack Overflow answer, I can't use your suggestion for arrays of ints, though I don't really understand why (something about `int` vs `Integer`?): https://stackoverflow.com/a/4962413

BhaaLseN: Sigh. Java.
But we _could_ just hide that in a util class somewhere, can't we? I'm almost certain we have one of those _somewhere_ already.
Or leave as-is, since I just don't know anymore...

JosJuice: We don't have a generic Java util class, just a util namespace that contains more specialized util classes (none of which this would fit into).

JosJuice:
The orientation lock commit is somewhat hacky right now due to an existing hack (see the comments in the commit), but merging PR #8452 would get rid of that existing hack.

JosJuice:
This PR should be ready to go now.

sfaludi1:
how do I implement this?

sfaludi1:
how do I get this to work on my phone?

JosJuice:
If you want to test this PR, you can download an APK at https://dl.dolphin-emu.org/prs/pr-8439-dolphin-latest.apk

Helios747:
I'm so excited to see how many users break their phones with this feature

PR #8440, opened 2019-10-29
Motion Input enhancements

rlnilsen:
Small enhancements to PR #8352.

#### Motion Input tab
- ~Apply default accelerometer and gyroscope bindings (as shown) when loading an old input configuration.
**Reason:** If there exists a `WiimoteNew.ini` from an older Dolphin version, those bindings will be blank, and motion input won't work. You would have to either manually set each of the 12 controls (which is not hard, just time consuming) or press the `Default` button and lose all other bindings.
**Details:** This was accomplished by adding versioning to input config files, including input profiles.~
- Add checkbox for enabling/disabling the motion controlled pointer. This is useful if you want to control the pointer by other means, like a mouse or the DualShock 4's touch pad. 

2019-11-01 03_38_00-vs2017 (1) - Remote Viewer - Copy

#### DSU Client config
- Enable `Server IP Address` and `Server Port` fields only when `Enable` is checked.

2019-10-29 21_39_37-vs2017 (1) - Remote Viewer

### DIFF
--- a/Source/Core/Core/HW/WiimoteEmu/Dynamics.cpp
+++ b/Source/Core/Core/HW/WiimoteEmu/Dynamics.cpp
@@ -325,6 +325,12 @@ void EmulateIMUCursor(std::optional<RotationalState>* state, ControllerEmu::IMUC
   // Avoid having to double dereference
   auto& st = *state;
 
+  if (!imu_ir_group->enabled)
+  {
+    st = std::nullopt;
+    return;
+  }
+
   auto accel = imu_accelerometer_group->GetState();
   auto ang_vel = imu_gyroscope_group->GetState();
 
--- a/Source/Core/InputCommon/ControllerEmu/ControlGroup/ControlGroup.h
+++ b/Source/Core/InputCommon/ControllerEmu/ControlGroup/ControlGroup.h
@@ -48,8 +48,16 @@ enum class GroupType
 class ControlGroup
 {
 public:
-  explicit ControlGroup(std::string name, GroupType type = GroupType::Other);
-  ControlGroup(std::string name, std::string ui_name, GroupType type = GroupType::Other);
+  enum class CanBeDisabled
+  {
+    No,
+    Yes,
+  };
+
+  explicit ControlGroup(std::string name, GroupType type = GroupType::Other,
+                        CanBeDisabled can_be_disabled = CanBeDisabled::No);
+  ControlGroup(std::string name, std::string ui_name, GroupType type = GroupType::Other,
+               CanBeDisabled can_be_disabled = CanBeDisabled::No);
   virtual ~ControlGroup();
 
   virtual void LoadConfig(IniFile::Section* sec, const std::string& defdev = "",
@@ -79,7 +87,9 @@ public:
   const std::string name;
   const std::string ui_name;
   const GroupType type;
+  const bool can_be_disabled;
 
+  bool enabled = true;
   std::vector<std::unique_ptr<Control>> controls;
   std::vector<std::unique_ptr<NumericSettingBase>> numeric_settings;
 };
--- a/Source/Core/DolphinQt/Config/ControllerInterface/DualShockUDPClientWidget.cpp
+++ b/Source/Core/DolphinQt/Config/ControllerInterface/DualShockUDPClientWidget.cpp
@@ -29,10 +29,12 @@ void DualShockUDPClientWidget::CreateWidgets()
 
   m_server_address = new QLineEdit(
       QString::fromStdString(Config::Get(ciface::DualShockUDPClient::Settings::SERVER_ADDRESS)));
+  m_server_address->setEnabled(m_server_enabled->isChecked());
 
   m_server_port = new QSpinBox();
   m_server_port->setMaximum(65535);
   m_server_port->setValue(Config::Get(ciface::DualShockUDPClient::Settings::SERVER_PORT));
+  m_server_port->setEnabled(m_server_enabled->isChecked());
 
   auto* description =
       new QLabel(tr("DSU protocol enables the use of input and motion data from compatible "
@@ -58,8 +60,10 @@ void DualShockUDPClientWidget::CreateWidgets()
 void DualShockUDPClientWidget::ConnectWidgets()
 {
   connect(m_server_enabled, &QCheckBox::toggled, this, [this] {
-    Config::SetBaseOrCurrent(ciface::DualShockUDPClient::Settings::SERVER_ENABLED,
-                             m_server_enabled->isChecked());
+    bool checked = m_server_enabled->isChecked();
+    Config::SetBaseOrCurrent(ciface::DualShockUDPClient::Settings::SERVER_ENABLED, checked);
+    m_server_address->setEnabled(checked);
+    m_server_port->setEnabled(checked);
   });
 
   connect(m_server_address, &QLineEdit::editingFinished, this, [this] {
--- a/Source/Core/DolphinQt/Config/Mapping/MappingWidget.cpp
+++ b/Source/Core/DolphinQt/Config/Mapping/MappingWidget.cpp
@@ -4,8 +4,10 @@
 
 #include "DolphinQt/Config/Mapping/MappingWidget.h"
 
+#include <QCheckBox>
 #include <QFormLayout>
 #include <QGroupBox>
+#include <QLabel>
 #include <QPushButton>
 #include <QTimer>
 
@@ -136,6 +138,28 @@ QGroupBox* MappingWidget::CreateGroupBox(const QString& name, ControllerEmu::Con
 
     if (setting_widget)
       form_layout->addRow(tr(setting->GetUIName()), setting_widget);
+  }
+
+  if (group->can_be_disabled)
+  {
+    QLabel* group_enable_label = new QLabel(tr("Enable"));
+    QCheckBox* group_enable_checkbox = new QCheckBox();
+    group_enable_checkbox->setChecked(group->enabled);
+    form_layout->insertRow(0, group_enable_label, group_enable_checkbox);
+    auto enable_group_by_checkbox = [group, form_layout, group_enable_label,
+                                     group_enable_checkbox] {
+      group->enabled = group_enable_checkbox->isChecked();
+      for (int i = 0; i < form_layout->count(); ++i)
+      {
+        QWidget* widget = form_layout->itemAt(i)->widget();
+        if (widget != nullptr && widget != group_enable_label && widget != group_enable_checkbox)
+          widget->setEnabled(group->enabled);
+      }
+    };
+    enable_group_by_checkbox();
+    connect(group_enable_checkbox, &QCheckBox::toggled, this, enable_group_by_checkbox);
+    connect(this, &MappingWidget::ConfigChanged, this,
+            [group_enable_checkbox, group] { group_enable_checkbox->setChecked(group->enabled); });
   }
 
   return group_box;
--- a/Source/Core/InputCommon/ControllerEmu/ControlGroup/ControlGroup.cpp
+++ b/Source/Core/InputCommon/ControllerEmu/ControlGroup/ControlGroup.cpp
@@ -15,13 +15,16 @@
 
 namespace ControllerEmu
 {
-ControlGroup::ControlGroup(std::string name_, const GroupType type_)
-    : name(name_), ui_name(std::move(name_)), type(type_)
+ControlGroup::ControlGroup(std::string name_, const GroupType type_, CanBeDisabled can_be_disabled_)
+    : name(name_), ui_name(std::move(name_)), type(type_),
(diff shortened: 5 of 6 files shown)


Source/Core/Core/HW/WiimoteEmu/Dynamics.cpp
```diff
@@ -325,6 +325,12 @@ void EmulateIMUCursor(std::optional<RotationalState>* state, ControllerEmu::IMUC
   // Avoid having to double dereference
   auto& st = *state;
 
+  if (!imu_ir_group->GetEnabled())
+  {
+    st = std::nullopt;
```
BhaaLseN: Maybe I'm a bit late to the party, but is there a reason `state` isn't just the return value?

rlnilsen: I've just followed the pattern of similar functions (`EmulateSwing`, `EmulateTilt`, `EmulateCursor`, `EmulateShake`). Monkey see, monkey do :-)

But I'm guessing it's probably because the `state` argument is both input and output.

jordan-woyak:
I think I liked the "Enable" checkbox you had better than the floating mystery checkbox in the title.

rlnilsen:
Since there are a few UI changes, perhaps @MayImilae and/or @JMC47 could take a look?

Also @mbc07 as the originator of most of these additions.

JMC47:
The UI additions look fine to me.

My one worry is when we add Nunchuck support to this, how it'll fit in, but that was discussed in the previous PR.

rlnilsen:
> I think I liked the "Enable" checkbox you had better than the floating mystery checkbox in the title.

@jordan-woyak is talking about a WIP version of this PR that looked like this:
2019-10-29 16_50_06-vs2017 (1) - Remote Viewer - Copy - Copy

JMC47:
I have to disagree, personally, but to each their own.  I don't care enough about the small details to really argue one way or the other.

iwubcode:
>Add Reset to Default buttons for Accelerometer and Gyroscope controls. If there exists a WiimoteNew.ini from an older Dolphin version, those controls will be blank, and motion input won't work. Without these new Reset to Default buttons, you would have to either manually set each of the 12 controls (which is not hard, just time consuming) or press the Default button and lose all other bindings.

I know it was already talked about but I didn't really voice my thoughts before.  I think this is a bit weird.  The reset buttons are not on any of our other sections and it's odd to me to include them here.

I do understand the reasoning, just wonder if there's a way to avoid it.  Here's just a stab at one potential alternative:

Instead of the default buttons, if you click the "Enable" button under DSU Client, it will notice if the input controller(s) has any empty assignments.  If so, it will ask the user if they wish to update them with the defaults.

mbc07:
Ok, let's review:
* Changes on Alternate Input Sources window LGTM
* Regarding the IMU Pointer, I think we're almost there:
  * I agree with @jordan-woyak that a "floating" checkbox on the header of the section is odd. Also it would be inconsistent as nowhere else on our UI we have a section with a checkbox on its header. Better having an individual checkbox inside the section like in the previous WIP.
  * However, for consistency, I think we could follow the same design pattern of the DSU Client window: The "Enable" (instead of "Enabled") checkbox should be the first item, followed by Recenter and Total Yaw.
  * To better communicate their relation to the user, Recenter and Total Yaw items should remain in disabled state until the "Enable" checkbox is selected.
  * While we're at it, perhaps consider renaming this group from "Point" to "IMU Pointer", "Emulated Pointer" or anything along those lines, to make clear it's not the same "Point" present on the "Motion Simulation" tab.
* Regarding the "Reset to Default" buttons added to "Accelerometer" and "Gyroscope" sections:
  * Yes, it was me who asked for them, but seeing how they look now that it's implemented, indeed they seem odd and out of place.
  * @iwubcode suggestion to this seems more reasonable, but perhaps it doesn't need to be that verbal. Unless I missed something, the only situation where these bindings will be blank is when an user is updating from an older version of Dolphin or when loading a saved profile made with old Dolphin versions.
  * So, based on that, make Dolphin check if the Accel/Gyro bindings are blank on the moment the user enable DSU Client; if yes, just silently set them to the default values and be done. If any of the bindings isn't blank, assume the user has previously manually configured them and do nothing.
  * Finally, get rid of the new "Reset to Default" buttons.

Source/Core/InputCommon/ControllerEmu/ControlGroup/ControlGroup.h
```diff
@@ -79,7 +81,9 @@ class ControlGroup
   const std::string name;
   const std::string ui_name;
   const GroupType type;
+  const bool disablable;
```
BhaaLseN: Maybe thats just me, but this word sounds terrible in my head :/
🚲 Am I alone, wishing this to be something more like `can_be_disabled` instead?

Feel free to ignore this one before we start hitting each other with paint buckets in disagreement over the shed's color.

iwubcode:
Personally I think versioning should not be specific to inputs, it should be available for all configuration and there should be a central version value.  Not sure where it would be best to go.  Maybe `IniFile`?

A couple of days ago we talked about having a config validator/fixer that happens during startup and that could do the changes to whatever config files we need to make modifications to.  We also did talk about a rollback/roll-forward feature but probably fine to start simple..

Source/Core/Core/HW/GCKeyboard.cpp
```diff
@@ -18,7 +18,9 @@
 
 namespace Keyboard
 {
-static InputConfig s_config("GCKeyNew", _trans("Keyboard"), "GCKey");
+static constexpr int CONFIG_VERSION = 1;
```
iwubcode: Unless I'm missing something, this value can just exist on the `InputConfig` class as a static and there isn't a need for the member variable.

rlnilsen: I guess it depends on if you want separate versions for each input config file.

iwubcode: That's a good point, I didn't think of that.

I think it just depends on what the team wants to do long-term.  Moving to a static will collapse us into one config version.  This makes it easy to fold into a more general purpose config version later or still specialize it for inputs ala `INPUT_CONFIG_VERSION`.  It also would line up with other parts of the code like: `STATE_VERSION`, `CACHE_REVISION`, and `GX_PIPELINE_UID_VERSION`.  That makes knowing when to change the number easy.  Make a config change?  Bump the version.

However, separate versions has pros too.  Would love to hear others thoughts on this.

rlnilsen:
I have now implemented most of the things discussed, but with a different solution for the accel/gyro bindings issue. Please see the updated PR description for details and screenshots.

@mbc07 suggested some new names for the `Point` group on the `Motion Input` tab to separate it from the group of same name on the `Motion Simulation` tab. I wasn't really able to find a name that I liked, so I left it alone.

> Personally I think versioning should not be specific to inputs, it should be available for all configuration and there should be a central version value.

There's some layering stuff going on with non input config files that I didn't want to deal with. I hope this implementation is good enough for the current needs, and that a more complete solution can be devised in the future if necessary.

mbc07:
I can't really comment about the code as I'm very unfamiliar with Dolphin's config system at this level, but all UI changes looks good to me...

JosJuice:
I noticed that the IR emulation added in the previous PR doesn't take the Sideways/Upright Wii Remote settings into account. In terms of what this would represent on real Wii Remote, it's as if the IR camera moves from one side of the Wii Remote to another when you enable Sideways Wii Remote. Should this be changed? There are some games that reply on the IR camera not pointing at the sensor bar when holding the Wii Remote sideways, like Metroid: Other M.

rlnilsen:
> I noticed that the IR emulation added in the previous PR doesn't take the Sideways/Upright Wii Remote settings into account.

It was done like this on purpose, after [this recommendation](https://github.com/dolphin-emu/dolphin/pull/8352#discussion_r322752856) from @jordan-woyak:
> I'd vote to not touch the IR data as it's just going to confuse people if they can't point normally

jordan-woyak:
@JosJuice Dolphin has never made sideways/upright affect IR data. We pretend the camera is always sticking out towards the TV in any orientation.

re: Metroid: Other M: The "Hide" mapping accomplishes the pointing vs. not pointing. I think it would confuse users to require they disable "Sideways Wii Remote" to do pointing. Consider any sideways remote game that has a pointer. e.g. Mario Kart Wii menus would be harder to navigate when playing without an extension.

Potentially we could make the "Upright" setting affect IR data. I think most games will not present a pointer with conflicting accelerometer data, anyways. But for consistency we've made neither orientation affect IR.

JosJuice:
Fair enough. I'm fine with leaving it the way it is unless we get comments about it from users.

Source/Core/Core/HW/GCKeyboard.cpp
rlnilsen: If my thinking is correct, moving between a common version and separate versions for input config files has no consequences. Therefore, until there is a clear need for one or the other, which we choose is irrelevant.

rlnilsen:
@BhaaLseN @iwubcode Is the code review satisfactory?

If so, I think this PR should be merge ready.

Source/Core/InputCommon/ControllerEmu/ControlGroup/ControlGroup.h
```diff
@@ -48,11 +48,19 @@ enum class GroupType
 class ControlGroup
 {
 public:
-  explicit ControlGroup(std::string name, GroupType type = GroupType::Other);
-  ControlGroup(std::string name, std::string ui_name, GroupType type = GroupType::Other);
+  enum CanBeDisabledOption
```
BhaaLseN: I think this should really be an `enum class`. Unless you can think of a better name, how about this?
```c++
enum class CanBeDisabled
{
  Yes,
  No,
};
```

rlnilsen:
After starting work on motion input support for the Nunchuk, I realized that my proposed solution for the accel/gyro bindings issue needs more thought. I therefore withdraw the related commits from this PR, to not delay the other issues fixed.

If there are no more code review issues, this PR should be merge ready.

Source/Core/InputCommon/ControllerEmu/ControlGroup/ControlGroup.h
```diff
@@ -79,7 +87,9 @@ class ControlGroup
   const std::string name;
   const std::string ui_name;
   const GroupType type;
+  const bool can_be_disabled;
 
+  bool enabled;
```
iwubcode: I'd set enabled to `true` here, then remove them from the constructors.  This ensure any future contructors will have `true` as the default.

iwubcode:
Just tested this.  Controller profiles don't seem to save the "enabled" state for point movement.  This is mostly an inconvienance but also might be surprising to some users (if the user is like me, it'd be when using game-specific input profiles)

rlnilsen:
> Controller profiles don't seem to save the "enabled" state for point movement.

It seems to work fine here. Like other settings, it is not written to file when it has the default value, which is `true`. Perhaps this is what you experienced?

rlnilsen:
> Controller profiles don't seem to save the "enabled" state for point movement.

Fixed, the UI was not updated after loading a profile.

(Crossing my fingers hoping this PR is ready now.)

iwubcode:
Thanks, the controller profiles now work well.  One minor thing, when I disable the point controls and close Dolphin, if I open Dolphin back up and go to the configuration, the controls are enabled despite the "enabled" checkbox not being checked.

rlnilsen:
>  One minor thing, when I disable the point controls and close Dolphin, if I open Dolphin back up and go to the configuration, the controls are enabled despite the "enabled" checkbox not being checked.

Fixed.

mbc07:
@rlnilsen I'm not sure if you follow our bug tracker so I'll just ask here. I have a small feature request which, apart from the UI side, should be pretty straightforward to implement: set the emulated Wiimote battery level to the same battery level reported by the DSU Server.

Right now users can already set the emulated Wiimote battery level to any fixed percentage they want, having an option to use the actual battery level reported by the controller connected to the DSU Server instead of the fixed value set by the user would be a very nice improvement...

jordan-woyak:
@mbc07 @rlnilsen I think battery "pass through" can be cleanly implemented with PR #8417. The battery level can be exposed as an `Input` which can then be supplied to the battery level setting.

rlnilsen:
> set the emulated Wiimote battery level to the same battery level reported by the DSU Server.

This could be useful for *all* input protocols reporting battery levels (DSU, XInput, others?). PR #8329  is possibly relevant.

> I think battery "pass through" can be cleanly implemented with PR #8417. The battery level can be exposed as an `Input` which can then be supplied to the battery level setting.

That seems like a good idea. The additional changes required in the DSU Client will be small.

I will however leave that to someone else. My goal was to get motion input support added to Dolphin. Seeing the Nunchuk motion input PR home, I feel that goal will have been met.

PR #8441, opened 2019-10-30
Fix versionCode dummy value

KenJPH:
When `git` isn't installed to the terminal the `getVersion()` and `getBuildVersionCode()` will use dummy values.

The problem is the dummy value for `getBuildVersionCode()` causes gradle to complain with the following error.
```
ERROR: android.defaultConfig.versionCode is set to 0, but it should be a positive integer.
See https://developer.android.com/studio/publish/versioning#appversioning for more information.
Affected Modules: app
```

Changed the dummy versionCode number to 1.


JosJuice:
@dolphin-emu-bot rebuild

PR #8443, opened 2019-10-30
RedumpVerifier: Fix handling of Datel Wii disc serials

JosJuice:


Source/Core/DiscIO/VolumeGC.cpp
```diff
@@ -74,13 +74,11 @@ std::string VolumeGC::GetGameID(const Partition& partition) const
 
 std::string VolumeGC::GetGameTDBID(const Partition& partition) const
 {
-  const std::string game_id = GetGameID(partition);
-
-  // Don't return an ID for Datel discs that are using the game ID of NHL Hitz 2002
-  if (game_id == "GNHE5d" && !GetBootDOLOffset(*this, partition).has_value())
+  // Don't return an ID for Datel discs
+  if (!GetBootDOLOffset(*this, partition).has_value())
     return "";
```
CookiePLMonster: I don't know if this should be the concern but NRVO won't kick in with multiple return paths like this.

Source/Core/DiscIO/VolumeWii.cpp
```diff
@@ -313,6 +313,10 @@ std::string VolumeWii::GetGameID(const Partition& partition) const
 
 std::string VolumeWii::GetGameTDBID(const Partition& partition) const
 {
+  // Don't return an ID for Datel discs
+  if (!GetBootDOLOffset(*this, partition).has_value())
+    return "";
```
CookiePLMonster: Same here, no NRVO. `std::string` is not big though so should be non issue.

Source/Core/DiscIO/VolumeVerifier.cpp
```diff
@@ -241,9 +241,9 @@ std::vector<RedumpVerifier::PotentialMatch> RedumpVerifier::ScanDatfile(const st
       continue;
 
     const std::string serials = game.child("serial").text().as_string();
-    if (serials.empty())
+    if (serials.empty() || StringBeginsWith(serials, "DS"))
```
CookiePLMonster: No access to `std::string::startswith` yet I assume?

JosJuice: No, we're on C++17 currently.

PR #8444, opened 2019-11-01
VI: Adjust start/end of ACV region for fields

booto:
This wasn't changed when halfline counts were changed to 0-based. Also
included is a diagram showing an understanding of how the values in the VI
timing registers maps to the video signal.


JosJuice:
Fixes https://bugs.dolphin-emu.org/issues/11889.

JosJuice:
This doesn't affect https://bugs.dolphin-emu.org/issues/11851 for me (the other regression caused by the same commit).

PR #8445, opened 2019-11-01
Add /dev/dolphin for homebrew to get information about Dolphin

Leseratte10:
I added some code to Dolphin, mainly a new device `/dev/dolphin` that can be used by Dolphin-aware software like homebrew to query certain information about Dolphin. 

Currently, that includes the real system time in ms (not the emulated system time), and the Dolphin version. Now I know from past conversations on the forum that there is no way for emulated software to detect real system time as that might confuse games when system time and emulated time run at different speeds. 

However, implementing this as an additional device /dev/dolphin means that no official game will ever know of this, as no official game would ever access /dev/dolphin. 

For Wiimmfi it would be pretty cool to let the game know that it lags (doesn't run at full 60fps), by comparing emulated time with this real time. Currently, when a player plays Mario Kart Wii online, they get a huge advantage when their Dolphin doesn't run at 60fps all time. If the game knew exactly how much it lags (that's the information that this PR gives the game) it can adapt the current racing timer, making that run at real-time. That would remove the advantage playing at <60fps gives you, and would allow players with slower computers to play on Wiimmfi as well. 

~Currently still a work in progress as I need to test it some more and might want to add another thing to the /dev/dolphin interface, but~ I wanted to create this PR to kinda ask the Dolphin team if there is a possibility for getting this merged at some point or if you'd have any objections against such an interface.

### DIFF
--- a/Source/Core/Core/BootManager.cpp
+++ b/Source/Core/Core/BootManager.cpp
@@ -212,6 +212,11 @@ void ConfigCache::RestoreConfig(SConfig* config)
 
 static ConfigCache config_cache;
 
+void SetEmulationSpeedReset(bool value)
+{
+  config_cache.bSetEmulationSpeed = value;
+}
+
 static GPUDeterminismMode ParseGPUDeterminismMode(const std::string& mode)
 {
   if (mode == "auto")
--- a/Source/Core/Core/IOS/DolphinDevice.cpp
+++ b/Source/Core/Core/IOS/DolphinDevice.cpp
@@ -1,0 +1,151 @@
+// Copyright 2019 Dolphin Emulator Project
+// Licensed under GPLv2+
+// Refer to the license.txt file included.
+
+#include <algorithm>
+#include <cstring>
+
+#include "Common/Logging/Log.h"
+#include "Common/Timer.h"
+#include "Common/scmrev.h"
+#include "Core/BootManager.h"
+#include "Core/ConfigManager.h"
+#include "Core/Core.h"
+#include "Core/HW/Memmap.h"
+#include "Core/IOS/DolphinDevice.h"
+
+namespace IOS::HLE::Device
+{
+namespace
+{
+enum
+{
+  IOCTL_DOLPHIN_GET_SYSTEM_TIME = 0x01,
+  IOCTL_DOLPHIN_GET_VERSION = 0x02,
+  IOCTL_DOLPHIN_GET_SPEED_LIMIT = 0x03,
+  IOCTL_DOLPHIN_SET_SPEED_LIMIT = 0x04,
+  IOCTL_DOLPHIN_GET_CPU_SPEED = 0x05,
+
+};
+
+IPCCommandResult GetSystemTime(const IOCtlVRequest& request)
+{
+  if (!request.HasNumberOfValidVectors(0, 1))
+  {
+    return DolphinDevice::GetDefaultReply(IPC_EINVAL);
+  }
+
+  if (request.io_vectors[0].size != 4)
+  {
+    return DolphinDevice::GetDefaultReply(IPC_EINVAL);
+  }
+
+  const u32 milliseconds = Common::Timer::GetTimeMs();
+
+  Memory::Write_U32(milliseconds, request.io_vectors[0].address);
+  return DolphinDevice::GetDefaultReply(IPC_SUCCESS);
+}
+
+IPCCommandResult GetVersion(const IOCtlVRequest& request)
+{
+  if (!request.HasNumberOfValidVectors(0, 1))
+  {
+    return DolphinDevice::GetDefaultReply(IPC_EINVAL);
+  }
+
+  const auto length = std::min(size_t(request.io_vectors[0].size), std::strlen(SCM_DESC_STR));
+
+  Memory::Memset(request.io_vectors[0].address, 0, request.io_vectors[0].size);
+  Memory::CopyToEmu(request.io_vectors[0].address, SCM_DESC_STR, length);
+
+  return DolphinDevice::GetDefaultReply(IPC_SUCCESS);
+}
+
+IPCCommandResult GetCPUSpeed(const IOCtlVRequest& request)
+{
+  if (!request.HasNumberOfValidVectors(0, 1))
+  {
+    return DolphinDevice::GetDefaultReply(IPC_EINVAL);
+  }
+
+  if (request.io_vectors[0].size != 4)
+  {
+    return DolphinDevice::GetDefaultReply(IPC_EINVAL);
+  }
+
+  const SConfig& config = SConfig::GetInstance();
+  const float oc = config.m_OCEnable ? config.m_OCFactor : 1.0f;
+
+  const u32 core_clock = u32(float(SystemTimers::GetTicksPerSecond()) * oc);
+
+  Memory::Write_U32(core_clock, request.io_vectors[0].address);
+
+  return DolphinDevice::GetDefaultReply(IPC_SUCCESS);
+}
+
+IPCCommandResult GetSpeedLimit(const IOCtlVRequest& request)
+{
+  // get current speed limit
+  if (!request.HasNumberOfValidVectors(0, 1))
+  {
+    return DolphinDevice::GetDefaultReply(IPC_EINVAL);
+  }
+
+  if (request.io_vectors[0].size != 4)
+  {
+    return DolphinDevice::GetDefaultReply(IPC_EINVAL);
+  }
+
+  const SConfig& config = SConfig::GetInstance();
+  const u32 speed_percent = config.m_EmulationSpeed * 100;
+  Memory::Write_U32(speed_percent, request.io_vectors[0].address);
+
+  return DolphinDevice::GetDefaultReply(IPC_SUCCESS);
+}
+
+IPCCommandResult SetSpeedLimit(const IOCtlVRequest& request)
+{
+  // set current speed limit
+  if (!request.HasNumberOfValidVectors(1, 0))
+  {
+    return DolphinDevice::GetDefaultReply(IPC_EINVAL);
+  }
+
+  if (request.in_vectors[0].size != 4)
+  {
+    return DolphinDevice::GetDefaultReply(IPC_EINVAL);
+  }
+
+  const float speed = float(Memory::Read_U32(request.in_vectors[0].address)) / 100.0f;
+  SConfig::GetInstance().m_EmulationSpeed = speed;
+  BootManager::SetEmulationSpeedReset(true);
+
+  return DolphinDevice::GetDefaultReply(IPC_SUCCESS);
+}
+
+}  // namespace
+
+IPCCommandResult DolphinDevice::IOCtlV(const IOCtlVRequest& request)
+{
+  if (Core::WantsDeterminism())
+  {
+    return DolphinDevice::GetDefaultReply(IPC_EACCES);
+  }
+
+  switch (request.request)
+  {
+  case IOCTL_DOLPHIN_GET_SYSTEM_TIME:
+    return GetSystemTime(request);
+  case IOCTL_DOLPHIN_GET_VERSION:
+    return GetVersion(request);
+  case IOCTL_DOLPHIN_GET_SPEED_LIMIT:
+    return GetSpeedLimit(request);
+  case IOCTL_DOLPHIN_SET_SPEED_LIMIT:
+    return SetSpeedLimit(request);
+  case IOCTL_DOLPHIN_GET_CPU_SPEED:
+    return GetCPUSpeed(request);
+  default:
+    return GetDefaultReply(IPC_EINVAL);
+  }
+}
+}  // namespace IOS::HLE::Device
--- a/Source/Core/Core/IOS/DolphinDevice.h
+++ b/Source/Core/Core/IOS/DolphinDevice.h
@@ -1,0 +1,18 @@
+// Copyright 2019 Dolphin Emulator Project
+// Licensed under GPLv2+
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include "Core/IOS/Device.h"
+
+namespace IOS::HLE::Device
+{
+class DolphinDevice final : public Device
+{
+public:
+  // Inherit the constructor from the Device class, since we don't need to do anything special.
+  using Device::Device;
+  IPCCommandResult IOCtlV(const IOCtlVRequest& request) override;
+};
+}  // namespace IOS::HLE::Device
--- a/Source/Core/Core/BootManager.h
+++ b/Source/Core/Core/BootManager.h
@@ -12,6 +12,7 @@ struct WindowSystemInfo;
 namespace BootManager
 {
 bool BootCore(std::unique_ptr<BootParameters> parameters, const WindowSystemInfo& wsi);
+void SetEmulationSpeedReset(bool value);
 
 // Synchronise Dolphin's configuration with the SYSCONF (which may have changed during emulation),
 // and restore settings that were overriden by per-game INIs or for some other reason.
--- a/Source/Core/Core/CMakeLists.txt
+++ b/Source/Core/Core/CMakeLists.txt
@@ -320,6 +320,8 @@ add_library(core
   IOS/Device.h
   IOS/DeviceStub.cpp
   IOS/DeviceStub.h
+  IOS/DolphinDevice.cpp
+  IOS/DolphinDevice.h
   IOS/IOS.cpp
   IOS/IOS.h
   IOS/IOSC.cpp
--- a/Source/Core/Core/Core.vcxproj
+++ b/Source/Core/Core/Core.vcxproj
@@ -197,6 +197,7 @@
     <ClCompile Include="HW\WiiSave.cpp" />
     <ClCompile Include="IOS\Device.cpp" />
     <ClCompile Include="IOS\DeviceStub.cpp" />
+    <ClCompile Include="IOS\DolphinDevice.cpp" />
     <ClCompile Include="IOS\IOS.cpp" />
     <ClCompile Include="IOS\IOSC.cpp" />
     <ClCompile Include="IOS\MIOS.cpp" />
@@ -473,6 +474,7 @@
     <ClInclude Include="HW\WII_IPC.h" />
     <ClInclude Include="IOS\Device.h" />
     <ClInclude Include="IOS\DeviceStub.h" />
+    <ClInclude Include="IOS\DolphinDevice.h" />
     <ClInclude Include="IOS\IOS.h" />
     <ClInclude Include="IOS\IOSC.h" />
     <ClInclude Include="IOS\MIOS.h" />
--- a/Source/Core/Core/Core.vcxproj.filters
+++ b/Source/Core/Core/Core.vcxproj.filters
@@ -707,6 +707,9 @@
     <ClCompile Include="IOS\DeviceStub.cpp">
       <Filter>IOS</Filter>
     </ClCompile>
+    <ClCompile Include="IOS\DolphinDevice.cpp">
+      <Filter>IOS</Filter>
+    </ClCompile>
     <ClCompile Include="IOS\DI\DI.cpp">
       <Filter>IOS\DI</Filter>
     </ClCompile>
@@ -1433,6 +1436,9 @@
       <Filter>IOS\USB\Bluetooth</Filter>
     </ClInclude>
     <ClInclude Include="IOS\DeviceStub.h">
+      <Filter>IOS</Filter>
+    </ClInclude>
+    <ClInclude Include="IOS\DolphinDevice.h">
       <Filter>IOS</Filter>
     </ClInclude>
     <ClInclude Include="IOS\DI\DI.h">
--- a/Source/Core/Core/IOS/IOS.cpp
+++ b/Source/Core/Core/IOS/IOS.cpp
@@ -30,6 +30,7 @@
 #include "Core/IOS/DI/DI.h"
 #include "Core/IOS/Device.h"
 #include "Core/IOS/DeviceStub.h"
+#include "Core/IOS/DolphinDevice.h"
 #include "Core/IOS/ES/ES.h"
 #include "Core/IOS/FS/FileSystem.h"
 #include "Core/IOS/FS/FileSystemProxy.h"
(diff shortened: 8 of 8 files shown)


Antidote:
I think this is a good idea, especially if it gets fleshed out more.

JMC47:
I'm not against this, it's a pretty nifty idea and it could potentially help with cheaters and unintentional disruptions from users lagging on games like Mario Kart Wii.

Leseratte10:
Just added the changes suggested by lioncash in the commit 422b911. I wasn't exactly sure how "This can likely be replaced with forward declarations." was meant, but I noticed that two of these included files weren't even needed and removed the include for these.

Source/Core/Core/IOS/DolphinDevice.cpp
```diff
@@ -0,0 +1,81 @@
+// Copyright 2019 Dolphin Emulator Project
+// Licensed under GPLv2+
+// Refer to the license.txt file included.
+
+#include <sys/time.h>
```
lioncash: This doesn't appear to be necessary (as far as I can tell). `<algorithm>` and `<cstring>` need to be included though to satisfy `std::min` and `std::strlen`

Source/Core/Core/IOS/DolphinDevice.cpp
```diff
@@ -0,0 +1,81 @@
...
+#include "Core/IOS/DolphinDevice.h"
+#include "Common/Logging/Log.h"
+#include "Common/scmrev.h"
+#include "Common/Timer.h"
+
+
+namespace IOS::HLE::Device
+{
+  
+  IPCCommandResult GetSystemTime(const IOCtlVRequest& request);
```
lioncash: There's a misunderstanding here when I meant they could be moved into the cpp file since they're implementation details. I was talking about the enum constants, not the actual function prototypes.

Leseratte10: They are implementation details? For /dev/es they are in the .h file as well (which I used as a "base"). Why would these be implementation details if custom code would need to know these to access the appropriate function?

lioncash: If they aren't used outside of the es code, they should also be in the cpp file as well. They're literally only used within `IOCtlV` inside the translation unit and not outside of it. By definition they're implementation details relative to other constructs within the codebase.

If other people need to know about things in particular, this is better served through a direct documentation comment rather than making reading code a requirement.

Leseratte10: Okay, I'll move them to the cpp file.

Leseratte10:
New commit 43e8d49, missing includes added (algorithm & cstring), unused include removed (sys/time.h, that was from before I knew about Common/timer.h), and enum moved to CPP file.

BhaaLseN:
You should consider running `clang-format` on your code; having it look and feel the same in every file helps by presenting a consistent style where everything looks familiar at some point.
Also, I don't see any vcxproj/filters changes, so this wont work on Windows as-is.

Leseratte10:
Ran clang-format and added the new files to the vcxproj files, didn't know Windows used something other than the CMake lists. Hoping this works on Windows now, I don't have a Windows environment to test this.

BhaaLseN:
Buildbot seems to be fine with that change. We're still looking into using CMake on Windows as well, but that is still some time out until its ready.

delroth:
Should probably be disabled in determinism mode just in case.

I very much like this change. One use case I have for this is when writing PPC tests for Dolphin, it could give us a nice interface to report test outputs.

Leseratte10:
So that determinism mode is for stuff like Netplay where the exact same output is preferred, right? So I could just make it return an error instead ot system time when in determinism mode. Or can that mode be enabled during online play as well and mess that up?
Would the version reporting need to be disabled as well?

delroth:
I would just not expose the device in this condition really. It's supposed to be a general interface for host access, which by nature will be 99% non-deterministic (the version number thing is likely the exception).

If I'm correct it should only be enabled during netplay and input recording / TAS, but there might be other conditions.

JosJuice:
> So that determinism mode is for stuff like Netplay where the exact same output is preferred, right?

Yes.

> Or can that mode be enabled during online play as well and mess that up?

Connecting to the internet is disabled when using deterministic mode, since the data you receive from servers on the internet isn't deterministic.

> Would the version reporting need to be disabled as well?

Having version reporting enabled when using deterministic mode isn't a problem in practice since we don't guarantee determinism across different versions of Dolphin. But it might make sense to just not expose the device at all when using deterministic mode.

Leseratte10:
Done. /dev/dolphin is now only exposed when Core::WantsDeterminism() is false.

JosJuice:
Hmm... There's also the case where determinism mode is enabled after emulation starts, though. (This only happen when starting a TAS from a savestate.) Is it possible to remove a device after it has been added?

Leseratte10:
I don't know. There is a function AddDevice, but I don't see RemoveDevice. And that might cause problems when you remove a device while it's opened by the game? 
Maybe it'd be better to just expose it all the time but then return IPC_EINVAL or IPC_EACCES when it is accessed in determinism mode.

Leseratte10:
`/dev/dolphin` is now created all the time, but `DolphinDevice::IOCtlV` checks `Core::WantsDeterminism()` and returns `IPC_EINVAL` if that is true.

Leseratte10:
In case you are interested, this is what it will look when this patch is added to Dolphin and the necessary server changes are rolled out to everyone: https://www.youtube.com/watch?v=NKtpx6g8TNg 

This is a live view of a Mario Kart Wii race, with Dolphin's speed limit set to 30%. The players are jumping all over the place, but the timer in the top right runs at real time, and the advantage players were getting with a slow Dolphin is gone.

Antidote:
One suggestion @jackoalan made for an addition to this is the ability to print directly to dolphin without the need to patch `OSReport` or `printf`

It was also suggested to add a pseudo terminal to the debug UI (not within the scope of this PR) where these messages are printed to without being burdened by other useful log messages in the log viewer.

And this is probably far outside the scope, but the ability to send and receive commands between dolphin and the running application for a REPL interface would also be incredibly useful, but could definitely be handled in a separate PR if this gets merged.

Leseratte10:
You mean to be able to print to the Dolphin log without Dolphin having to intercept the output of OSReport? What would be the point when the current solution works? Or is there some situation where the current process fails?

Might not be a bad idea to have a seperate terminal for debug output from own custom code that doesn't go into the Dolphin log, but I agree with you that that should be done in a seperate PR. 
As visible in the video above, this code seems to work so far, so I removed the "WIP" prefix from the PR, and if there aren't any other changes requested by the Dolphin team I guess this could be merged now. 

Could someone with the appropriate rights trigger the build bot again please?

Antidote:
The patching method can be unreliable, I've had issues with my own homebrew where `printf` was not properly hooked and I was unable to use the built in log, having a direct route to Dolphin without a potentially failed hook is desireable imho. I think having the hooks works extremely well for games using the official SDK, but it's hit or miss in my personal experience developing homebrew.

Source/Core/Core/IOS/DolphinDevice.cpp
```diff
@@ -0,0 +1,89 @@
...
+#include "Common/Logging/Log.h"
+#include "Common/Timer.h"
+#include "Common/scmrev.h"
+#include "Core/Core.h"
+#include "Core/HW/Memmap.h"
+#include "Core/IOS/DolphinDevice.h"
+
+namespace IOS::HLE::Device
+{
+IPCCommandResult GetSystemTime(const IOCtlVRequest& request);
```
lioncash: An anonymous namespace can be used here around the enum and the definition of these functions to allow the removal of these redundant prototypes.

Leseratte10: added anonymous namespace in bb0caef.

Leseratte10:
Added the anonymous namespace in bb0caef. Also, the Wiimmfi update to support this is now live, so everyone testing this PR should be able to replicate the timer changes seen in the video above (timer still in real-time even when Dolphin slows down).

JosJuice:
@dolphin-emu-bot rebuild

mbc07:
Might be slightly off topic but this PR make online races in Mario Kart Wii to be more fair when emulator users and console users are on the same match. However, it won't be effective if emulator users don't update Dolphin or use forks based on older versions.

How do you plan to make sure emulator users connecting to Wiimmfi are running an updated Dolphin version or fork, which have the changes from this PR? As far as I can tell there's no way to detect that server-side...

Leseratte10:
Well, when this PR is merged, the game will be able to detect the Dolphin version and send it to the server. The server can then send back an error message like "please update Dolphin to at least 5.0-xxxxx to play online" when it detects a Dolphin client that doesn't report its version number. 

Thanks for the rebuild, I'm going to get some testers to try this.

mbc07:
Ah, that clear things up. Last time I checked Wiimmfi servers couldn't tell if a client were a real console or Dolphin, especially if you used a dump of your own NAND...

Leseratte10:
Yeah, there has been a huge update like a year ago that gives the server much more control over the game, including checking the console type.

Leseratte10:
Just announced that beta in the Wiimmfi forum, that should give us a few beta testers: https://forum.wii-homebrew.com/index.php/Thread/59391-Dolphin-Beta-Testing/

Leseratte10:
Found out during testing that even with these changes a Dolphin player gets a "lag start" (starts the race a bit after the other players) when they lag during track selection and track loading. I'll try adding code to Dolphin to fix that as well. That lag start wouldn't be problematic in terms of the Dolphin player getting an advantage, but it's still stupid if a Dolphin player starts the race multiple seconds after the others. So, back to "work in progress".

JMC47:
Good stuff.  If testing ends up fine, I'm hugely in favor of this as I think it'd be great for a number of things.  Helping people play online + development uses just seems really cool.

Leseratte10:
Added code that A) allows the software to read the current CPU overclock setting and B) code that allows the game to read and set the Dolphin speed limit. 

In Mario Kart Wii online races, prior to the race, the clients syncronize their clocks to make sure they all start the race at the same time. When Dolphin now runs slower than fullspeed, it'll start the race later than others. 

With these changes, Wiimmfi can detect slowdowns, then force the game into 30fps mode (making it only render half of the frames, reducing the load), and if that still causes lagging, it'll set the speed limit to unlimited until either A) the race starts or B) Dolphin has made up the speed it lost. 

That method doesn't help much if the computer is still way too slow for 30fps, but if it's just little lag frames the game will be able to counteract that by speeding up Dolphin until the timers are in sync again. 

Can someone re-trigger the bot please?

Source/Core/Core/IOS/DolphinDevice.cpp
```diff
@@ -0,0 +1,150 @@
...
+    return DolphinDevice::GetDefaultReply(IPC_EINVAL);
+  }
+
+  if (request.io_vectors[0].size != 4)
+  {
+    return DolphinDevice::GetDefaultReply(IPC_EINVAL);
+  }
+
+  const SConfig& config = SConfig::GetInstance();
+  float oc = config.m_OCEnable ? config.m_OCFactor : 1.0f;
```
lioncash:
```suggestion
  const float oc = config.m_OCEnable ? config.m_OCFactor : 1.0f;
```

Source/Core/Core/IOS/DolphinDevice.cpp
```diff
@@ -0,0 +1,150 @@
...
+
+  unsigned int core_clock = ((float)SystemTimers::GetTicksPerSecond() * oc);
```
lioncash:
```suggestion
  const u32 core_clock = u32(float(SystemTimers::GetTicksPerSecond()) * oc);
```

Source/Core/Core/IOS/DolphinDevice.cpp
```diff
@@ -0,0 +1,150 @@
...
+  if (request.HasNumberOfValidVectors(0, 1))
+  {
+    // get current speed limit
+    if (request.io_vectors[0].size != 4)
+    {
+      return DolphinDevice::GetDefaultReply(IPC_EINVAL);
+    }
+
+    const SConfig& config = SConfig::GetInstance();
+    unsigned int speed_percent = config.m_EmulationSpeed * 100;
```
lioncash:
```suggestion
    const u32 speed_percent = config.m_EmulationSpeed * 100;
```

Source/Core/Core/IOS/DolphinDevice.cpp
```diff
@@ -0,0 +1,150 @@
...
+  }
+  else if (request.HasNumberOfValidVectors(1, 0))
+  {
+    // set current speed limit
+    if (request.in_vectors[0].size != 4)
+    {
+      return DolphinDevice::GetDefaultReply(IPC_EINVAL);
+    }
+
+    float speed = ((float)(Memory::Read_U32(request.in_vectors[0].address))) / 100.0f;
```
lioncash:
```suggestion
    const float speed = float(Memory::Read_U32(request.in_vectors[0].address)) / 100.0f;
```

Source/Core/Core/IOS/DolphinDevice.cpp
```diff
@@ -0,0 +1,150 @@
...
+  else if (request.HasNumberOfValidVectors(1, 0))
+  {
+    // set current speed limit
+    if (request.in_vectors[0].size != 4)
+    {
+      return DolphinDevice::GetDefaultReply(IPC_EINVAL);
+    }
+
+    const float speed = float(Memory::Read_U32(request.in_vectors[0].address)) / 100.0f;
+    SConfig::GetInstance().m_EmulationSpeed = (float)speed;
```
lioncash:
```suggestion
    SConfig::GetInstance().m_EmulationSpeed = speed;
```

I believe I may have missed this in a previous review (sorry!), but speed is already a float, so the cast can be removed here.

foolishsyrup:
What is the status of this?

JMC47:
Probably just waiting on reviews.  Everyone seems in favor of this, it's just getting people to review/approve it so it can be merged.

Leseratte10:
I put it back in WIP status because I'm trying to find a way to get rid of the lag start (happens when Dolphin runs too slow during track selection) as well. I didn't get around to implementing the necessary server changes yet (will probably do so on the weekend). 

Most likely that will not require any other changes to Dolphin, but to make sure I'd like to leave this PR open until this is done so in case it needs another feature in Dolphin I can just add it to this one. 

I'll let you know when the server changes are ready (and if what I'm planning for the track selection even works).

Leseratte10:
I just hacked the fix for the track load slowdown into my game manually (not yet rolled out on Wiimmfi yet) and it seems to work. Doesn't get rid of the slowdown completely, but helps a bit. 

So as soon as the latest changes get reviewed / approved by someone this PR will be ready to be merged.

JosJuice:
I'm a bit sceptical about letting the emulated software override the user's emulation speed like that... If we're going to do it, we should at least do it for the current session only (which I unfortunately suppose would be a bit hacky to implement since the emulation speed hasn't been migrated to the new config system).

Leseratte10:
I tested that, and if I remember correctly it did only override the speed for the current session (or until the user changes it in the config again). I can test that again to make sure if you want.

JosJuice:
It's possible that I'm misunderstanding the BootManager logic and that it in fact is already is for the current session only. If you want to re-test it, I would appreciate it.

Leseratte10:
Oh, damn, it looks like this is indeed not being reset. The settings menu shows 100% again, but even after restarting Dolphin the speed is still set to the value the game set it to (and it's displayed wrong in the settings), until I actively set it to something else and then back to 100% in the settings. Even though this setting will only be used temporarily and will set the speed back to 100% later, that will cause problems if someone exits the game or quits Dolphin while in a higher speed mode. 

Any idea how this could be implemented for the current session only?

JosJuice:
You'll probably want to set `bSetEmulationSpeed` in `BootManager.cpp` to true, or do something else involving `ConfigCache` in `BootManager.cpp`. The problem is how to do this while keeping the interface clean...

Leseratte10:
So I'd probably need to add a public function in BootManager.cpp that sets the `bSetEmulationSpeed` in the private struct?

Source/Core/Core/IOS/DolphinDevice.cpp
```diff
@@ -0,0 +1,152 @@
...
+  const float oc = config.m_OCEnable ? config.m_OCFactor : 1.0f;
+
+  const u32 core_clock = u32(float(SystemTimers::GetTicksPerSecond()) * oc);
+
+  Memory::Write_U32(core_clock, request.io_vectors[0].address);
+
+  return DolphinDevice::GetDefaultReply(IPC_SUCCESS);
+}
+
+IPCCommandResult SpeedLimit(const IOCtlVRequest& request)
```
leoetlino: IMO this should be split into two ioctls, GetSpeedLimit and SetSpeedLimit. This would make things clearer and consistent with existing interfaces.

Source/Core/Core/IOS/DolphinDevice.cpp
```diff
@@ -0,0 +1,152 @@
...
+  case IOCTL_DOLPHIN_GET_SYSTEM_TIME:
+    return GetSystemTime(request);
+  case IOCTL_DOLPHIN_GET_VERSION:
+    return GetVersion(request);
+  case IOCTL_DOLPHIN_SPEED_LIMIT:
+    return SpeedLimit(request);
+  case IOCTL_DOLPHIN_GET_CPU_SPEED:
+    return GetCPUSpeed(request);
+  default:
+    return DolphinDevice::GetDefaultReply(IPC_EINVAL);
```
leoetlino:
```suggestion
    return GetDefaultReply(IPC_EINVAL);
```

fancythedeveloper:
Well, hope this PR gets merged. Definitely will help Dolphin players play Wiimmfi w/o having to worry about Dolphin lagging (from dropping frames).

Source/Core/Core/IOS/DolphinDevice.cpp
```diff
@@ -0,0 +1,148 @@
...
+    {
+      return DolphinDevice::GetDefaultReply(IPC_EINVAL);
+    }
+
+    const float speed = float(Memory::Read_U32(request.in_vectors[0].address)) / 100.0f;
+    SConfig::GetInstance().m_EmulationSpeed = speed;
+    BootManager::SetEmulationSpeedReset(true);
+  }
+
+  return DolphinDevice::GetDefaultReply(IPC_EINVAL);
```
lioncash: Should this not be returning IPC_SUCCESS? Currently both the successful path and the error path both return the same error code, which is likely not intentional.

Source/Core/Core/IOS/DolphinDevice.cpp
```diff
@@ -0,0 +1,148 @@
...
+  const u32 core_clock = u32(float(SystemTimers::GetTicksPerSecond()) * oc);
+
+  Memory::Write_U32(core_clock, request.io_vectors[0].address);
+
+  return DolphinDevice::GetDefaultReply(IPC_SUCCESS);
+}
+
+IPCCommandResult GetSpeedLimit(const IOCtlVRequest& request)
+{
+  if (request.HasNumberOfValidVectors(0, 1))
```
lioncash: This can likely be inverted into:

```suggestion
  if (!request.HasNumberOfValidVectors(0, 1))
  {
    return DolphinDevice::GetDefaultReply(IPC_EINVAL);
  }
```

In order to unindent the bulk of the contents of the function and make the error case more consistent with the rest of the implementation functions surrounding this one.

Source/Core/Core/IOS/DolphinDevice.cpp
```diff
@@ -0,0 +1,148 @@
...
+    Memory::Write_U32(speed_percent, request.io_vectors[0].address);
+
+    return DolphinDevice::GetDefaultReply(IPC_SUCCESS);
+  }
+  return DolphinDevice::GetDefaultReply(IPC_EINVAL);
+}
+
+IPCCommandResult SetSpeedLimit(const IOCtlVRequest& request)
+{
+  if (request.HasNumberOfValidVectors(1, 0))
```
lioncash: This should likely also be inverted and return the error code as well.

Source/Core/Core/IOS/DolphinDevice.h
```diff
@@ -0,0 +1,16 @@
+// Copyright 2019 Dolphin Emulator Project
+// Licensed under GPLv2+
+// Refer to the license.txt file included.
+
```
leoetlino: missing `#pragma once`

Source/Core/Core/BootManager.cpp
```diff
@@ -212,6 +212,11 @@ void ConfigCache::RestoreConfig(SConfig* config)
 
 static ConfigCache config_cache;
 
+void SetEmulationSpeedReset(bool value)
```
leoetlino: Looks like there's a trailing space here?
```suggestion
void SetEmulationSpeedReset(bool value)
```

fancythedeveloper:
_someone better merge this now or one kid dyin' in the morn_

Leseratte10:
So now that this has been reviewed by @leoetlino what is the process to actually get this merged and into Dolphin? Will someone else need to review this PR as well? From my side this should now be done, all the other changes will now be server-side and not in Dolphin.

JosJuice:
> So now that this has been approved by @leoetlino what is the process to actually get this merged and into Dolphin? Will someone else need to review this PR as well?

Basically, yes. There's no exact process that's set in stone, but that's how it often works. Either way, there's nothing more you need to do with this PR for now unless someone leaves a comment about something to change.

JMC47:
Hey @Leseratte10 , I'm working on a writeup for this in the progress report.  Is there any way you could provide a screenshot/gif/something that shows it adjusting things in game so that I can show how cool it is?  And would you be willing to review what I'm writing to make sure it's accurate?

Thank you!

Leseratte10:
Hi JMC47, I'd definitely be willing to review the chapter for the progress report. I created a proof-of-concept video a while ago: https://www.youtube.com/watch?v=NKtpx6g8TNg 

It shows a Mario Kart Wii race with Dolphin being slowed down to 30% (as a result, all the players are jumping around) and the in-game timer is still running at full speed. Don't know if there would be a better way to show the time corrections in a video/GIF.

JMC47:
No, this works.  I guess I want to make sure I fully understand how this works.

dev/dolphin is a device that the game can communicate over and see certain things (like computer's actual clock?) and in turn control Dolphin to some degree as well?

Leseratte10:
That is correct. /dev/dolphin is a device that can be opened (by the game), like any of the original Wii devices (/dev/es, /dev/di, and so on) and can both receive information (Dolphin version, computer system time, Dolphin speed) and send / set information (Dolphin speed). The Wiimmfi-patched game can then use the computer system time to find out if it is lagging (if a certain time has passed on the computer clock but less time passed on the emulated clock there has been a slowdown), and if it is, it increases the in-game race timer until it caught up with real time again.

JMC47:
Thanks for the concise explanation.  I look forward to featuring it in the report, as this is a very unique situation with an awesome solution.

PR #8446, opened 2019-11-02
RedumpVerifier: Show an error when datfile lacks serials or versions

JosJuice:
This happens if someone manually downloads a regular datfile from redump.org and puts it where Dolphin stores datfiles. Dolphin needs "special" datfiles that contain fields for serials and versions.

Before this change, all discs (except Datel discs) would show up as "Unknown disc" when using a regular datfile.

### DIFF
--- a/Source/Core/DiscIO/VolumeVerifier.cpp
+++ b/Source/Core/DiscIO/VolumeVerifier.cpp
@@ -99,7 +99,7 @@ void RedumpVerifier::Start(const Volume& volume)
       ERROR_LOG(DISCIO, "Failed to fetch data from Redump.org, using old cached data instead");
       [[fallthrough]];
     case DownloadStatus::Success:
-      return ScanDatfile(ReadDatfile(system));
+      return ScanDatfile(ReadDatfile(system), system);
 
     case DownloadStatus::SystemNotAvailable:
       m_result = {Status::Error, Common::GetStringT("Wii data is not public yet")};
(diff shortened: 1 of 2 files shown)


Source/Core/DiscIO/VolumeVerifier.cpp
```diff
@@ -213,7 +213,8 @@ static std::vector<u8> ParseHash(const char* str)
   return hash;
 }
 
-std::vector<RedumpVerifier::PotentialMatch> RedumpVerifier::ScanDatfile(const std::vector<u8>& data)
+std::vector<RedumpVerifier::PotentialMatch> RedumpVerifier::ScanDatfile(const std::vector<u8>& data,
+                                                                        const std::string& system)
```
CookiePLMonster: I thought it's all about `std::string_view` now?

JosJuice: Using `std::string_view` in this specific case doesn't have any advantages (because the calling function doesn't do any substringing or similar) and actually has a disadvantage in the case where we need to call `c_str()` on it.

iwubcode: Just curious but what about making this a pass by value, then using `std::move` ?

JosJuice: I don't see why that would be faster unless we need to create a copy of the string for some reason (since that would let us skip making the copy).

iwubcode: If it is faster, it's a micro optimization.  No big deal, I just wanted to bring it up since @CookiePLMonster brought up string-view and it has been on my mind

PR #8447, opened 2019-11-03
[WIP] Attempt to fix av desync in segmented video

vadosnaprimer:
* Get latest Dolphin from master.
* Get this movie http://tasvideos.org/6501S.html
* Get this memory card http://files.tasvideos.org/common/SubmissionFiles/6501S/MemoryCardA_USA.7z
* Launch Dolphin
* Go to `Graphics -> Advanced`, disable `Dump at Internal Resolution`
* Enable `Movie -> Dump Frames`
* Replay the movie on _Resident Evil 4 (Disc 2)_
```
CRC32: 6C83A5FF
MD5: 2381ACD2199D6E7566932DF86901903D
SHA-1: C75F7936814636FFE03277F363FC3427C98602EE
```
* Stop the movie after the level starts
* Rename or move `framedump0.avi`
* Enable `Dump at Internal Resolution`
* Dump frames again.

If you splice the 2 dump segments that were done from _internal resolution_, you'll see that the level starts 68 frames earlier than in the single-segment dump that used the _window resolution_. It's because `s_last_frame_is_valid` gets reset, resulting in wrong frame duration timings and missing frames. This results in **heavy** audio-video desync if you dump the entire movie's audio and video and splice/mux everything together.

I tried to fix that by accounting for whether it's a new segment caused by video split, and now I consistently get 1 _extra frame_ on every split. I've been staring at the code for the entire day, and all the frame timings are now identical to the correct dump scenario (single-segment), but the output still doesn't match. I'm unable to figure out what else should be tweaked to get rid of the extra frame. Help!

### DIFF
--- a/Source/Core/VideoCommon/FrameDump.h
+++ b/Source/Core/VideoCommon/FrameDump.h
@@ -22,7 +22,7 @@ public:
     int savestate_index = 0;
   };
 
-  static bool Start(int w, int h, bool new_segment = false);
+  static bool Start(int w, int h);
   static void AddFrame(const u8* data, int width, int height, int stride, const Frame& state);
   static void Stop();
   static void DoState();
--- a/Source/Core/VideoCommon/FrameDump.h
+++ b/Source/Core/VideoCommon/FrameDump.h
@@ -22,7 +22,7 @@ public:
     int savestate_index = 0;
   };
 
-  static bool Start(int w, int h);
+  static bool Start(int w, int h, bool new_segment = false);
   static void AddFrame(const u8* data, int width, int height, int stride, const Frame& state);
   static void Stop();
   static void DoState();
--- a/Source/Core/VideoCommon/FrameDump.cpp
+++ b/Source/Core/VideoCommon/FrameDump.cpp
@@ -78,15 +78,18 @@ static bool AVStreamCopyContext(AVStream* stream, AVCodecContext* codec_context)
 #endif
 }
 
-bool FrameDump::Start(int w, int h)
+bool FrameDump::Start(int w, int h, bool new_segment)
 {
   s_pix_fmt = AV_PIX_FMT_RGBA;
 
   s_width = w;
   s_height = h;
-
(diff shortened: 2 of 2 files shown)


Source/Core/VideoCommon/FrameDump.h
```diff
@@ -22,7 +22,7 @@ class FrameDump
     int savestate_index = 0;
   };
 
-  static bool Start(int w, int h);
+  static bool Start(int w, int h, bool new_segment = false);
```
lioncash: This should likely be an enum rather than a bool so it's obvious at the callsite what behavior is being signified.

vadosnaprimer:
This should be better.

vadosnaprimer:
I found the problem, every last frame of a dump only has duration of 1 (since next frame's PTS is non-existent), even if it's supposed to repeat. The current approach uses just PTSs for every new frame, and the decoder then calculates how long each frame should last by looking at next frame's PTS. But when a video ends, there's no next frame, hence no duration info.

For a single-segment dump this problem is hard to notice, it's also non-apparent for games that render at 60fps. But if you have a game that renders at 30fps and changes resolution, every segment's last frame will last for 1 video frame instead of 2.

Now the solution would be to set `pkt.duration` explicitly I think, but we can't know it until the next frame has been emulated! I'm not sure how to resolve this.

vadosnaprimer:
Accidentally marked as ready, please ignore for now. It's still a WIP.

altimumdelta:
> Now the solution would be to set `pkt.duration` explicitly I think, but we can't know it until the next frame has been emulated! I'm not sure how to resolve this.

I'll start looking at this Framedumping thing again then; also while it's not required, it doesn't stop anyone from developing locally, it's still open what kind of ffmpeg integration the community feels would like to have along with the update to the ffmpeg external for windows, delayed dll or static.

vadosnaprimer:
I'm working on making it output separate frames, actually CFR.

PR #8448, opened 2019-11-03
cmake: set OpenGL_GL_PREFERENCE as a cached variable that can be modi…

aguinet:
…fied via the command line

This allows the user to set this to LEGACY. This can be useful for
people under Linux using optirun/primusrun with "old" nvidia drivers
that don't support GLVND properly.


aguinet:
Note: I had to set this to LEGACY on my system to have optirun works with dolphin!

CookiePLMonster:
@dolphin-emu-bot rebuild

PR #8451, opened 2019-11-04
Add motion input support to nunchuk

rlnilsen:
After discussions we ended up with the following UI layout. When (and only when) Nunchuk is selected as extension, two extra tabs appear: 'Extension Motion Simulation' and 'Extension Motion Input'.

2019-11-06 18_43_04-vs2017 (1) - Remote Viewer
2019-11-06 18_43_27-vs2017 (1) - Remote Viewer
2019-11-06 18_55_52-vs2017 (1) - Remote Viewer

#### To preserve the context of the discussion that took place, the original version of this post is preserved below

The UI layout was slightly adjusted to allow fitting five control groups on a 1024 pixel wide screen.
Added the same warning text and `Alternate Input Sources` button as on the `Motion Input` tab.

2019-11-04 18_48_38-vs2017 (1) - Remote Viewer

### DIFF
--- a/Source/Core/DolphinQt/Config/Mapping/WiimoteEmuExtensionMotionInput.cpp
+++ b/Source/Core/DolphinQt/Config/Mapping/WiimoteEmuExtensionMotionInput.cpp
@@ -1,0 +1,80 @@
+// Copyright 2019 Dolphin Emulator Project
+// Licensed under GPLv2+
+// Refer to the license.txt file included.
+
+#include "DolphinQt/Config/Mapping/WiimoteEmuExtensionMotionInput.h"
+
+#include <QGridLayout>
+#include <QGroupBox>
+#include <QHBoxLayout>
+#include <QLabel>
+#include <QPushButton>
+
+#include "Core/HW/Wiimote.h"
+#include "Core/HW/WiimoteEmu/Extension/Nunchuk.h"
+#include "Core/HW/WiimoteEmu/WiimoteEmu.h"
+
+#include "DolphinQt/Config/ControllerInterface/ControllerInterfaceWindow.h"
+
+#include "InputCommon/InputConfig.h"
+
+WiimoteEmuExtensionMotionInput::WiimoteEmuExtensionMotionInput(MappingWindow* window)
+    : MappingWidget(window)
+{
+  CreateNunchukLayout();
+  CreateMainLayout();
+}
+
+void WiimoteEmuExtensionMotionInput::CreateNunchukLayout()
+{
+  auto* layout = new QGridLayout();
+  m_nunchuk_box = new QGroupBox(tr("Nunchuk"), this);
+
+  auto* warning_layout = new QHBoxLayout();
+  auto* warning_label = new QLabel(
+      tr("WARNING: These controls are designed to interface directly with motion "
+         "sensor hardware. They are not intended for mapping traditional buttons, triggers or "
+         "axes. You might need to configure alternate input sources before using these controls."));
+  warning_label->setWordWrap(true);
+  auto* warning_input_sources_button = new QPushButton(tr("Alternate Input Sources"));
+  warning_layout->addWidget(warning_label, 1);
+  warning_layout->addWidget(warning_input_sources_button, 0, Qt::AlignRight);
+  connect(warning_input_sources_button, &QPushButton::clicked, this, [this] {
+    ControllerInterfaceWindow* window = new ControllerInterfaceWindow(this);
+    window->setAttribute(Qt::WA_DeleteOnClose, true);
+    window->setWindowModality(Qt::WindowModality::WindowModal);
+    window->show();
+  });
+  layout->addLayout(warning_layout, 0, 0, 1, -1);
+
+  layout->addWidget(CreateGroupBox(tr("Accelerometer"),
+                                   Wiimote::GetNunchukGroup(
+                                       GetPort(), WiimoteEmu::NunchukGroup::IMUAccelerometer)),
+                    1, 0);
+
+  m_nunchuk_box->setLayout(layout);
+}
+
+void WiimoteEmuExtensionMotionInput::CreateMainLayout()
+{
+  m_main_layout = new QHBoxLayout();
+
+  m_main_layout->addWidget(m_nunchuk_box);
+
+  setLayout(m_main_layout);
+}
+
+void WiimoteEmuExtensionMotionInput::LoadSettings()
+{
+  Wiimote::LoadConfig();
+}
+
+void WiimoteEmuExtensionMotionInput::SaveSettings()
+{
+  Wiimote::GetConfig()->SaveConfig();
+}
+
+InputConfig* WiimoteEmuExtensionMotionInput::GetConfig()
+{
+  return Wiimote::GetConfig();
+}
--- a/Source/Core/DolphinQt/Config/Mapping/WiimoteEmuExtensionMotionInput.h
+++ b/Source/Core/DolphinQt/Config/Mapping/WiimoteEmuExtensionMotionInput.h
@@ -1,0 +1,32 @@
+// Copyright 2019 Dolphin Emulator Project
+// Licensed under GPLv2+
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include "DolphinQt/Config/Mapping/MappingWidget.h"
+
+#include "Core/HW/WiimoteEmu/ExtensionPort.h"
+
+class QGroupBox;
+class QHBoxLayout;
+
+class WiimoteEmuExtensionMotionInput final : public MappingWidget
+{
+  Q_OBJECT
+public:
+  explicit WiimoteEmuExtensionMotionInput(MappingWindow* window);
+
+  InputConfig* GetConfig() override;
+
+private:
+  void LoadSettings() override;
+  void SaveSettings() override;
+
+  void CreateNunchukLayout();
+  void CreateMainLayout();
+
+  // Main
+  QHBoxLayout* m_main_layout;
+  QGroupBox* m_nunchuk_box;
+};
--- a/Source/Core/DolphinQt/Config/Mapping/WiimoteEmuExtensionMotionSimulation.h
+++ b/Source/Core/DolphinQt/Config/Mapping/WiimoteEmuExtensionMotionSimulation.h
@@ -1,0 +1,32 @@
+// Copyright 2019 Dolphin Emulator Project
+// Licensed under GPLv2+
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include "DolphinQt/Config/Mapping/MappingWidget.h"
+
+#include "Core/HW/WiimoteEmu/ExtensionPort.h"
+
+class QGroupBox;
+class QHBoxLayout;
+
+class WiimoteEmuExtensionMotionSimulation final : public MappingWidget
+{
+  Q_OBJECT
+public:
+  explicit WiimoteEmuExtensionMotionSimulation(MappingWindow* window);
+
+  InputConfig* GetConfig() override;
+
+private:
+  void LoadSettings() override;
+  void SaveSettings() override;
+
+  void CreateNunchukLayout();
+  void CreateMainLayout();
+
+  // Main
+  QHBoxLayout* m_main_layout;
+  QGroupBox* m_nunchuk_box;
+};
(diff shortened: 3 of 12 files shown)


JMC47:
Seems fine to me on an initial pass.

rlnilsen:
Oops, forgot the warning text and `Alternate Input Sources` button. Fixed.

rlnilsen:
Noticed some weirdness with the look of the spin boxes, seems like a possible QSS bug. Fixed by doing things differently (and simpler).

jordan-woyak:
I think selecting "Nunchuk" should make another tab with the "Accelerometer" mappings or there should be a button that opens another dialog with the mappings.

The amount of configuration available is already hard for users to handle and these additional mappings should rarely be used.

mbc07:
Controls magically appearing/disappearing might confuse users. I think it would be better having the Nunchuk's raw accelerometer bindings always visible on the "Motion Input" tab but in disabled state if the Extension setting of the main tab is set to anything other than Nunchuk...

jordan-woyak:
> Controls magically appearing/disappearing might confuse users. I think it would be better having the Nunchuk's raw accelerometer bindings always visible on the "Motion Input" tab but in disabled state if the Extension setting of the main tab is set to anything other than Nunchuk...

That sounds good.

rlnilsen:
> I think it would be better having the Nunchuk's raw accelerometer bindings always visible on the "Motion Input" tab but in disabled state if the Extension setting of the main tab is set to anything other than Nunchuk...

I think @MayImilae may [disagree](https://github.com/dolphin-emu/dolphin/pull/8352#issuecomment-541379697):

> I really really do not want you to put the Nunchuk accelerometer mappings into the Motion Input tab! Ok so, Dolphin has a design metaphor it is using for extensions that center around plugging a device into the Wii Remote. So like, for example, say a Wii Remote has nothing connected to it (extension is set to none), the extension tab is blank, just like a port would be in the physical setup. But plug in a Classic Controller (set the extension to "Classic Controller") and now the extension tab is filled with Classic Controller mappings. It mirrors the process of plugging a device into the Wii Remote, so the UX mirrors the experience of using the real thing: a classic design metaphor! Metaphors like this are the bread and butter of UX design are crucial for helping users understand their way around a UI. If you place the nunchuk motion input settings in with the wii remote input settings it will break that metaphor.

mbc07:
To be clear, the current design looks good to me, but if it ends being moved to the motion input tab like @jordan-woyak asked then I prefer the bindings being always visible but alternating between disabled/enabled state than having them magically appear/disappear on the Motion Input tab based on whether the Nunchuk is selected as an extension on the main tab...

rlnilsen:
@mbc07 Sorry, not the best quoting there.

iwubcode:
When I initially saw this, I too thought that the UI felt out of place.  The wiimote solution split the input into two tabs but now for the Nunchuk it's all in one tab.  That seems really confusing and unnecessary for users who will never touch that section.

Personally, I don't really mind what @mbc07 suggested but if that's not viable, I think the whole design needs to be rethought.  If the whole point is to mirror a wiimote with possible extensions, we've already kinda broken that by introducing two tabs that are "wiimote"s.

I'm not a UI expert but my idea would be to hide all the "raw" mappings under a raw sub-group that can be toggled.  I know I've seen this in UIs but can't think of what it was called.  This would be consistent for both the wiimote and nunchuk screens.

rlnilsen:
How the UI design ends up is not that important to me, but I would argue that putting the Nunchuk accelerometer controls as in the screenshot is consistent with existing UI design.

#### First, three facts about the UI _before any motion input support was added_:

Fact 1: Both the Wiimote UI and the extension UI _logically_ had these two sections (at least):
1. Non motion input
2. Motion simulation

Fact 2: The big difference between the Wiimote UI and the extension UI was in how they visually divided the sections. The Wiimote UI used tabs. The extension UI did not have visual dividers.

Fact 3: For both Wiimote and extension UIs, the sections had the same order, from left to right. Using tab ordering in the first case and position on screen in the second case.

#### Second, let's see what is the natural way of adding a section to each UI:

Wiimote UI: Add a new tab to the right of the `Motion Simulation` tab in the tab header. The `Motion Input` tab does this.

Extension UI: Add new control group(s) to the right of the rightmost control group in the motion simulation section. The `Accelerometer` group does this.

#### Third

I hope this makes sense.

rlnilsen:
@iwubcode 
> If the whole point is to mirror a wiimote with possible extensions, we've already kinda broken that by introducing two tabs that are "wiimote"s.

Do you mean the `Motion Simulation` and `Motion Input` tabs, in addition to the `General and Options` tab?

rlnilsen:
Following @MayImilae's [advice](https://github.com/dolphin-emu/dolphin/pull/8352#issuecomment-541379697), I really think we should keep all extension controls inside the `Extension` tab. I believe this limits us to two ways of organizing the Nunchuk controls: flat layout (like shown in the screenshot in the PR description) and tabbed layout (shown below). Which layout do you prefer?

2019-11-06 01_58_23-vs2017 (1) - Remote Viewer
2019-11-06 01_58_48-vs2017 (1) - Remote Viewer
2019-11-06 01_58_54-vs2017 (1) - Remote Viewer

mbc07:
I'm almost sure that having another set of tabs inside an existing tab is a no-no on the design guidelines of all OSes we support, and understandably as it sure looks very awkward. If having all bindings related to the extension grouped on the same tab is desired, the approach from the first post still looks clean and better than the other discussed solutions until now...

jordan-woyak:
How about just a "Configure Motion Input" button added to the Extension tab when Nunchuk is selected? It opens a new dialog where the usual warning and accelerometer mappings are.

Miksel12:
I think adding 'Extension Motion Simulation' and 'Extension Motion Input' tabs on the right side of the 'Extension' tab when the nunchuk is selected could also work. This would also in line with the Wiimote motion options and layout.

rlnilsen:
> 
> 
> How about just a "Configure Motion Input" button added to the Extension tab when Nunchuk is selected? It opens a new dialog where the usual warning and accelerometer mappings are.

Any suggestion exactly where to place the button?

jordan-woyak:
> > How about just a "Configure Motion Input" button added to the Extension tab when Nunchuk is selected? It opens a new dialog where the usual warning and accelerometer mappings are.
> 
> Any suggestion exactly where to place the button?

In other group under the "Shake" group maybe? There's a big space there :D

rlnilsen:
> I'm almost sure that having another set of tabs inside an existing tab is a no-no on the design guidelines of all OSes we support, and understandably as it sure looks very awkward.

It doesn't look quite as bad as I thought it would, though. Probably because there's the Nunchuk group box header inbetween the two tab rows. There's a lot of frames inside each other though, which I also think most guidelines shun.

iwubcode:
> Do you mean the Motion Simulation and Motion Input tabs, in addition to the General and Options tab?

Correct

> Following @MayImilae's advice, I really think we should keep all extension controls inside the Extension tab.

I'm still trying to decide what I think of that.  I like the symmetry but also agree with @mbc07  and think it looks a bit weird.  And what happens if a different extension is chosen?

> I think adding 'Extension Motion Simulation' and 'Extension Motion Input' tabs on the right side of the 'Extension' tab when the nunchuk is selected could also work. This would also be in line with the Wiimote motion options and layout.

Or remove Shake/Tilt/Swing and add two buttons: "Configure Motion Simulation" and "Configure Motion Input", with the former showing the shake/tilt/swing and the latter just showing the accelerometer?

With one button it seems like it might be weird to have it under shake but technically replace swing/tilt too...

> How about just a "Configure Motion Input" button added to the Extension tab when Nunchuk is selected? It opens a new dialog where the usual warning and accelerometer mappings are.

I did think of that but if a different extension is plugged in, what happens?

rlnilsen:
Here's a variant on the tabs-within-tabs layout of my previous suggestion.

2019-11-06 13_09_33-vs2017 (1) - Remote Viewer
2019-11-06 12_59_32-vs2017 (1) - Remote Viewer
2019-11-06 13_18_55-vs2017 (1) - Remote Viewer

Miksel12:
Definitely not a fan of that. With the tabs in tabs, the users can see all the options at once. This layout just says: 'Oh look, there are more options. You have to find out for yourself what it contains'. Besides, this layout isn't used anywhere else while the tabs in tabs can be found in Game Config under Game Properties. I'm also not a big fan of using a button to go to the Motion mapping as this isn't used anywhere else in dolphin.

JMC47:
I honestly liked the original idea that was posted more than these.

rlnilsen:
Here's @Miksel12's suggestion.

With no extension selected:

2019-11-06 18_46_52-vs2017 (1) - Remote Viewer

With Classic Controller selected as extension:

2019-11-06 18_47_48-vs2017 (1) - Remote Viewer

Finally with Nunchuk selected as extension:

2019-11-06 18_43_04-vs2017 (1) - Remote Viewer
2019-11-06 18_43_27-vs2017 (1) - Remote Viewer
2019-11-06 18_55_52-vs2017 (1) - Remote Viewer

mbc07:
Both @Miksel12 approach and "everything on the same tab" approach from the first post looks good to me...

iwubcode:
I think between the two I like the " @Miksel12 " approach best.

One variation on that is you could keep the current Nunchuk extension unchanged and add just a single tab "Extension Motion Input".  Though that won't match the Wiimote as well.  So sticking with what you have is probably best.

rlnilsen:
The latest UI layout discussed seems to be widely accepted. The code and PR description has been updated to this UI layout.

This PR should now be ready for code review.

mbc07:
Is there anything else holding this PR back? Emulated motion controls mapping would be finally feature complete on all platforms with the merge of this and #8473 PRs...

JMC47:
I'm fine with merging it as long as it's been tested.  I don't have a second motion controller to test it with.

mbc07:
Ok, I'll test this a bit more tomorrow and report back...

Source/Core/DolphinQt/Config/Mapping/WiimoteEmuExtensionMotionInput.cpp
```diff
@@ -0,0 +1,80 @@
+// Copyright 2017 Dolphin Emulator Project
```
jordan-woyak: 2019

Source/Core/DolphinQt/Config/Mapping/WiimoteEmuExtensionMotionInput.h
```diff
@@ -0,0 +1,32 @@
+// Copyright 2017 Dolphin Emulator Project
```
jordan-woyak: 2019

Source/Core/DolphinQt/Config/Mapping/WiimoteEmuExtensionMotionSimulation.h
```diff
@@ -0,0 +1,32 @@
+// Copyright 2017 Dolphin Emulator Project
```
jordan-woyak: 2019

mbc07:
> Ok, I'll test this a bit more tomorrow and report back...

Just tested this PR with a DualShock 4 set to the Wiimote and a separate DualShock 3 set to the Nunchuk and it worked fine on the games I tried.

rlnilsen:
I think this PR is ready to merge now.

praudmur:
Hello, folks!
I hope you haven't forgotten about this. I've been waiting for this fix since forever =)

rlnilsen:
The [Dolphin Progress Report for October](https://dolphin-emu.org/blog/2019/11/07/dolphin-progress-report-october-2019/) stated: "We expect Nunchuck Motion Input to be finished in the next *couple of weeks*" (my emphasis). That's *two months* ago now.

So if something is stopping this PR from being merged, please let me know.

For usability, the Switch JoyCons may not be far off the Wiimote + Nunchuk combo. That scenario is enabled by this PR used together with the "BetterJoyForCemu" DSU Server (see @mbc07's [guide](https://wiki.dolphin-emu.org/index.php?title=DSU_Client)). I'd really like to see that available to Dolphin's users.

mbc07:
Could we get this PR merged? All discussion points seems to be resolved by now..

JMC47:
Yeah, no one is discussing this and there are no more reviews.  I've tested this.  If anyone has anything to say against it, they've had plenty of time.

PR #8452, opened 2019-11-04
Android: Replace emulation rotation crash workaround with proper fix

JosJuice:
The workaround was added in 0446a58.

The underlying problem is that we must not destroy the surface while the video backend is initializing, otherwise the video backend may reference nullptr.

I've also cleaned up the logic for when to destroy the surface. Note that the comment in EmulationFragment.java about only being able to destroy the surface when emulation is running is not true anymore (due to de632fc, it seems like).


stenzek:
Seems annoying that we have to explicitly do this. My preference would be to just have starting block until everything is ready, but doing so could deadlock the Qt frontend as window messages have to be handled... Perhaps once I eventually get around to refactoring the frontend display and making the video thread own the window.

PR #8454, opened 2019-11-06
DolphinQt: Add accelerometer/gyroscope mapping indicators.

jordan-woyak:
These indicators will help in configuration and checking for proper orientation/scale.
It is currently a mystery if things are correct until you get in-game.

This is less of an issue when using the "Cemuhook" input backend (which has nicely named inputs) but more of an issue in Linux where evdev exposes the raw data.

Demonstration:
https://giant.gfycat.com/AllMediumFrillneckedlizard.webm
I've since added text to the gyroscope indicator with the red dot's angle from the upright position.

The accelerometer indicator shows the perceived orientation and g-force.
The colored dots provide a frame of reference.

The gyroscope indicator rotates in real time with the controller.
The red and blue dots are oriented differently so all three rotational axes can be followed and calibrated.
When the controller is stable for a second the gyroscope indicator resets.

### DIFF
--- a/Source/Core/DolphinQt/Config/Mapping/MappingIndicator.cpp
+++ b/Source/Core/DolphinQt/Config/Mapping/MappingIndicator.cpp
@@ -8,6 +8,8 @@
 #include <cmath>
 #include <numeric>
 
+#include <fmt/format.h>
+
 #include <QAction>
 #include <QDateTime>
 #include <QPainter>
@@ -118,6 +120,10 @@ double MappingIndicator::GetScale() const
 
 namespace
 {
+constexpr float SPHERE_SIZE = 0.7f;
+constexpr float SPHERE_INDICATOR_DIST = 0.85f;
+constexpr int SPHERE_POINT_COUNT = 200;
+
 // Constructs a polygon by querying a radius at varying angles:
 template <typename F>
 QPolygonF GetPolygonFromRadiusGetter(F&& radius_getter, double scale,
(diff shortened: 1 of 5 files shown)


Source/Core/DolphinQt/Config/Mapping/MappingIndicator.cpp
```diff
@@ -184,6 +190,22 @@ bool IsPointOutsideCalibration(Common::DVec2 point, ControllerEmu::ReshapableInp
   return current_radius > input_radius * ALLOWED_ERROR;
 }
 
+template <typename F>
+void GenerateFibonacciSphere(int point_count, F&& callback)
+{
+  const float golden_angle = MathUtil::PI * (3.f - std::sqrt(5.f));
```
BhaaLseN: I don't suppose this could be made `constexpr` at some point, could it? I'm pretty sure `MathUtil::PI` isn't `constexpr` right now, but I don't see any reason why it couldn't be (plus it is still usable at runtime in non-`constexpr` context if necessary).

jordan-woyak: `PI` is constexpr but `std::sqrt` isn't, unfortunately.

BhaaLseN: Awww, too bad. Maybe in a later standard.

barolo:
Can we get this merged? I'm trying to test DS4 gyro/accel and it' extremely confusing

Miksel12:
This doesn't seem to work with my DS4 with DSU, in the configure input screen the accel and gyro values are non zero and seemingly correct but the indicators show nothing.

jordan-woyak:
> This doesn't seem to work with my DS4 with DSU, in the configure input screen the accel and gyro values are non zero and seemingly correct but the indicators show nothing.

Yes, sorry, I based this on PR #8456 which will fix the problem.
A current workaround is enabling "Background Input".

Thanklang:
hello everyone

first thing sorry for my English language cuz is not good enough to speak, my problem is i cant use save state on dolphin emulator mmj and bounding box is not available

thank u so much

Thanklang:
my phone is sony Xperia 5

jordan-woyak:
@Thanklang We don't maintain "dolphin mmj". Also, this isn't a support forum.

PR #8455, opened 2019-11-06
Git-Ignore: Also ignore other derivatives of VS output directories

altimumdelta:
Someone already thought about it for the Build, but not the rest.

This helps with the trick of having  "_Branch Specific IDE Output Directories_", which can be created and managed with the following script I made originally for myself. https://gist.github.com/altimumdelta/1d43053f7373baae1066194b0f765a63

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,7 +6,7 @@ Thumbs.db
 Source/Core/Common/scmrev.h
 # Ignore files output by build
 /[Bb]uild*/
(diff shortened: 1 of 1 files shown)


leoetlino:
Please squash (or rather "fixup") your commits.

altimumdelta:
Yah, I did the other one in github directly, I'll figure it out.

PR #8459, opened 2019-11-08
DolphinQt: Hide unused buttons in the advanced mapping dialog.

jordan-woyak:
Fix regression from #8396.

Some widgets are conditionally added to the layout.

The unused ones were showing up in the top-left corner.

This hides the unused widgets.


BhaaLseN:
Why do we even add them in first place if we dont use/need them?

jordan-woyak:
> Why do we even add them in first place if we dont use/need them?

The widgets are referenced elsewhere and assumed to exist.
I think this is cleaner than throwing more "is input" checks throughout the code.

PR #8462, opened 2019-11-10
HW/WiimoteReal: Linux fix.

jordan-woyak:
Remove an old comment about blocking time.
We've use an infinite timeout here since: 02fc68ea5d1466998c7d59b2882a5f9e348de24b

The first argument to `select` was never updated in this commit.
It was causing `m_wakeup_pipe_r` to not be checked on my system. (sometimes)
As a result, writes would not occur unless there was something to read. (sometimes)

Apparently FD sets (and select) can't handle descriptors with values higher than 1024 anyways so I've changed the code to use `poll` instead which seems to be recommended.

### DIFF
--- a/Source/Core/Core/HW/WiimoteReal/IOLinux.cpp
+++ b/Source/Core/Core/HW/WiimoteReal/IOLinux.cpp
@@ -223,8 +223,6 @@ void WiimoteLinux::IOWakeup()
 // zero = error
 int WiimoteLinux::IORead(u8* buf)
 {
-  // Block select for 1/2000th of a second
-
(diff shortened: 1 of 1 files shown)


jordan-woyak:
> Any practical example of something fixed by this change?

When changing extensions a Wii remote sends one final report before disabling regular input reports.
The output report which re-enables reporting would not send because the wiimote thread was blocking on a read indefinitely. `select` was not looking at the wakeup pipe.
All communication would then halt until the next extension change.

This would all only happen "sometimes".. depending on the values of the descriptors and the timing of the read/written reports.

PR #8461, opened 2019-11-09
Android: Update target Android SdkVersion to 28

Simonx22:
Since 11/01/2019, Android requires SDK version 28 (aka Android 9). If you submit an update with SDK version 27 and below, Google will reject the update. Read more here: https://android-developers.googleblog.com/2019/02/expanding-target-api-level-requirements.html

I have tested this change on my phone and haven't encountered any issues.

PR #8463, opened 2019-11-10
Android:  update androidSupportVersion and compileSdkVersion to 28

Simonx22:
This is a suggested change by Android Studio. Depends on #8461.

I've tested this change on my device and didn't encounter any issues.

### DIFF
--- a/Source/Android/app/build.gradle
+++ b/Source/Android/app/build.gradle
@@ -1,7 +1,7 @@
 apply plugin: 'com.android.application'
 
 android {
-    compileSdkVersion 27
+    compileSdkVersion 28
 
     compileOptions {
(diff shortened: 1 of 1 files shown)


Adrvsh:
Hi, can I ask u guys some questions in here? I mean, I'm using dolphin emulator on my google pixel (android 10 with snapdragon 821) n i want to ask you about vulkan option, why it's so slow on my device.

For example I tested some wii/gc games with opengl and got 25-30 fps but when i switch to vulkan the fps is drop to 10-15 and sometimes it caused blank screen n makes the apps force close. Is it just me or anyone have a same problem with me?

I had this problem since when my device runnin on android pie until right now. I'm sorry for my long question guys.

JosJuice:
@dolphin-emu-bot rebuild

PR #8464, opened 2019-11-10
WiimoteEmu: Minor accuracy fixes.

jordan-woyak:
Minor edge case accuracy fixes tested on my Nintendo Wii Remote Plus.
The changes don't affect any games as far as I know.

### DIFF
--- a/Source/Core/Core/HW/WiimoteCommon/WiimoteConstants.h
+++ b/Source/Core/Core/HW/WiimoteCommon/WiimoteConstants.h
@@ -62,16 +62,22 @@ enum class AddressSpace : u8
   // FYI: The EEPROM address space is offset 0x0070 on i2c slave 0x50.
   // However attempting to access this device directly results in an error.
   EEPROM = 0x00,
-  // 0x01 is never used but it does function on a real wiimote:
-  I2CBusAlt = 0x01,
-  I2CBus = 0x02,
+  // I2CBusAlt is never used by games but it does function on a real wiimote.
+  I2CBus = 0x01,
+  I2CBusAlt = 0x02,
 };
 
 enum class ErrorCode : u8
 {
+  // Normal result.
   Success = 0,
+  // Produced by read/write attempts during an active read.
+  Busy = 4,
+  // Produced by using something other than the above AddressSpace values.
   InvalidSpace = 6,
+  // Produced by an i2c read/write with a non-responding slave address.
   Nack = 7,
+  // Produced by accessing invalid regions of EEPROM or the EEPROM directly over i2c.
   InvalidAddress = 8,
 };
 
--- a/Source/Core/Core/HW/WiimoteEmu/EmuSubroutines.cpp
+++ b/Source/Core/Core/HW/WiimoteEmu/EmuSubroutines.cpp
@@ -258,7 +258,12 @@ void Wiimote::HandleRequestStatus(const OutputReportRequestStatus&)
 
 void Wiimote::HandleWriteData(const OutputReportWriteData& wd)
 {
-  // TODO: Are writes ignored during an active read request?
+  if (m_read_request.size)
+  {
+    // FYI: Writes during an active read will occasionally produce a "busy" (0x4) ack.
+    // We won't simulate that as it often does work. Poorly programmed games may rely on it.
+    WARN_LOG(WIIMOTE, "WriteData: write during active read request.");
+  }
 
   u16 address = Common::swap16(wd.address);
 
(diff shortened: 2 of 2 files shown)


Source/Core/Core/HW/WiimoteEmu/EmuSubroutines.cpp
```diff
@@ -258,7 +258,12 @@ void Wiimote::HandleRequestStatus(const OutputReportRequestStatus&)
 
 void Wiimote::HandleWriteData(const OutputReportWriteData& wd)
 {
-  // TODO: Are writes ignored during an active read request?
+  if (m_read_request.size)
+  {
+    // FYI: Writes during an active read will occasionally produce a "busy" (0x4) ack.
+    // We won't simulate that as it often does work. Poorly programmed games may rely on it.
+    WARN_LOG(WIIMOTE, "WriteData: write during active read request.");
```
Warepire: Is the risk big enough to warrant statistics gathering?

jordan-woyak: Meh. Usually a real remote processes the write just fine in the case. Spamming writes during an active read fail about 22% of the time here. It seems to just be getting lucky with the timing. It shouldn't be a problem to just always work instead of working 78% of the time.

Source/Core/Core/HW/WiimoteEmu/EmuSubroutines.cpp
```diff
@@ -416,9 +421,11 @@ void Wiimote::HandleReadData(const OutputReportReadData& rd)
 {
   if (m_read_request.size)
   {
-    // There is already an active read request.
-    // A real wiimote ignores the new one.
-    WARN_LOG(WIIMOTE, "ReadData: ignoring read during active request.");
+    // There is already an active read being processed.
+    WARN_LOG(WIIMOTE, "ReadData: attempting read during active request.");
+
+    // My real wm+ sends a busy ack in this situation.
```
lioncash: "My" is ambiguous here given this is developed by more than one person.

Source/Core/Core/HW/WiimoteEmu/EmuSubroutines.cpp
```diff
@@ -258,7 +258,12 @@ void Wiimote::HandleRequestStatus(const OutputReportRequestStatus&)
 
 void Wiimote::HandleWriteData(const OutputReportWriteData& wd)
 {
-  // TODO: Are writes ignored during an active read request?
+  if (m_read_request.size)
+  {
+    // FYI: Writes during an active read will occasionally produce a "busy" (0x4) ack.
+    // We won't simulate that as it often does work. Poorly programmed games may rely on it.
```
iwubcode: Should this be "doesn't" ?

jordan-woyak: No. "it often does work." which is what is simulated here.

iwubcode: Hmm ok.  So you're saying we aren't sending a busy ack like for reads because on console it is so sporadic and a write will work anyway most of the time?

jordan-woyak: Yes. It the write will work "most of the time" on real hardware.

Source/Core/Core/HW/WiimoteCommon/WiimoteConstants.h
```diff
@@ -62,16 +62,22 @@ enum class AddressSpace : u8
...
 {
+  // Normal result.
   Success = 0,
+  // Produced by read/write attempts during an active read.
+  Busy = 4,
+  // Produced by using something other than the above AddressSpace values.
   InvalidSpace = 6,
+  // Produced by an i2c read/write with a non-responding slave address.
   Nack = 7,
+  // Produced by accessing protected regions of EEPROM or the EEPROM directly over i2c.
```
Pokechu22: I don't think "protected" is the right way of describing it; that also is returned when you read something that's definitely beyond the end of EEPROM if I recall correctly

Source/Core/Core/HW/WiimoteEmu/EmuSubroutines.cpp
Pokechu22: Perhaps "We won't simulate that as those reads still do often work, and poorly programmed games may rely on it despite it occasionally failing on hardware"?

That said, I don't think there are that many games that write to EEPROM, and I think the reading/writing is handled in Nintendo's SDK and that includes reading it back after writing to check, so it *probably* is fine to be strict.  I doubt a poorly programed game rolled its own EEPROM handling functions, but who knows...

Source/Core/Core/HW/WiimoteCommon/WiimoteConstants.h
jordan-woyak: I changed it to "invalid" now.

PR #8467, opened 2019-11-10
Make shader precompilation interruptable

CookiePLMonster:
This PR refactors shader precompilation ("Compile shaders before starting") to be interruptable. Previously, user was unable to close emulation gracefully during that step and had to either wait or forcibly terminate the process.

Additionally, UI window will now clear after finishing or aborting shader precompilation - so ImGui's message box will not linger until game renders something.

### DIFF
--- a/Source/Core/VideoCommon/AsyncShaderCompiler.cpp
+++ b/Source/Core/VideoCommon/AsyncShaderCompiler.cpp
@@ -9,6 +9,8 @@
 #include "Common/Logging/Log.h"
 #include "Common/Thread.h"
 
+#include "Core/Core.h"
+
 namespace VideoCommon
 {
 AsyncShaderCompiler::AsyncShaderCompiler()
@@ -65,17 +67,11 @@ bool AsyncShaderCompiler::HasCompletedWork()
   return !m_completed_work.empty();
 }
 
-void AsyncShaderCompiler::WaitUntilCompletion()
-{
-  while (HasPendingWork())
-    std::this_thread::sleep_for(std::chrono::milliseconds(1));
-}
-
-void AsyncShaderCompiler::WaitUntilCompletion(
+bool AsyncShaderCompiler::WaitUntilCompletion(
     const std::function<void(size_t, size_t)>& progress_callback)
 {
   if (!HasPendingWork())
-    return;
+    return true;
 
   // Wait a second before opening a progress dialog.
   // This way, if the operation completes quickly, we don't annoy the user.
@@ -85,11 +81,11 @@ void AsyncShaderCompiler::WaitUntilCompletion(
   {
     std::this_thread::sleep_for(std::chrono::milliseconds(CHECK_INTERVAL));
     if (!HasPendingWork())
-      return;
+      return true;
   }
 
   // Grab the number of pending items. We use this to work out how many are left.
-  size_t total_items = 0;
+  size_t total_items;
   {
     // Safe to hold both locks here, since nowhere else does.
     std::lock_guard<std::mutex> pending_guard(m_pending_work_lock);
@@ -100,6 +96,9 @@ void AsyncShaderCompiler::WaitUntilCompletion(
   // Update progress while the compiles complete.
   for (;;)
   {
+    if (Core::GetState() == Core::State::Stopping)
+      return false;
+
     size_t remaining_items;
     {
       std::lock_guard<std::mutex> pending_guard(m_pending_work_lock);
@@ -111,6 +110,7 @@ void AsyncShaderCompiler::WaitUntilCompletion(
     progress_callback(total_items - remaining_items, total_items);
     std::this_thread::sleep_for(CHECK_INTERVAL);
   }
+  return true;
 }
 
 bool AsyncShaderCompiler::StartWorkerThreads(u32 num_worker_threads)
--- a/Source/Core/VideoCommon/ShaderCache.cpp
+++ b/Source/Core/VideoCommon/ShaderCache.cpp
@@ -157,9 +157,12 @@ const AbstractPipeline* ShaderCache::GetUberPipelineForUid(const GXUberPipelineU
 
 void ShaderCache::WaitForAsyncCompiler()
 {
-  while (m_async_shader_compiler->HasPendingWork() || m_async_shader_compiler->HasCompletedWork())
+  bool running = true;
+
+  while (running &&
+         (m_async_shader_compiler->HasPendingWork() || m_async_shader_compiler->HasCompletedWork()))
   {
-    m_async_shader_compiler->WaitUntilCompletion([](size_t completed, size_t total) {
+    running = m_async_shader_compiler->WaitUntilCompletion([](size_t completed, size_t total) {
       g_renderer->BeginUIFrame();
 
       const float center_x = ImGui::GetIO().DisplaySize.x * 0.5f;
--- a/Source/Core/VideoCommon/ShaderCache.cpp
+++ b/Source/Core/VideoCommon/ShaderCache.cpp
@@ -159,35 +159,42 @@ void ShaderCache::WaitForAsyncCompiler()
 {
   bool running = true;
 
+  constexpr auto update_ui_progress = [](size_t completed, size_t total) {
+    g_renderer->BeginUIFrame();
+
+    const float center_x = ImGui::GetIO().DisplaySize.x * 0.5f;
+    const float center_y = ImGui::GetIO().DisplaySize.y * 0.5f;
+    const float scale = ImGui::GetIO().DisplayFramebufferScale.x;
+
+    ImGui::SetNextWindowSize(ImVec2(400.0f * scale, 50.0f * scale), ImGuiCond_Always);
+    ImGui::SetNextWindowPos(ImVec2(center_x, center_y), ImGuiCond_Always, ImVec2(0.5f, 0.5f));
+    if (ImGui::Begin(Common::GetStringT("Compiling Shaders").c_str(), nullptr,
+                     ImGuiWindowFlags_NoTitleBar | ImGuiWindowFlags_NoInputs |
+                         ImGuiWindowFlags_NoMove | ImGuiWindowFlags_NoSavedSettings |
+                         ImGuiWindowFlags_NoScrollbar | ImGuiWindowFlags_NoNav |
+                         ImGuiWindowFlags_AlwaysAutoResize | ImGuiWindowFlags_NoFocusOnAppearing))
+    {
+      ImGui::Text("Compiling shaders: %zu/%zu", completed, total);
+      ImGui::ProgressBar(static_cast<float>(completed) /
+                             static_cast<float>(std::max(total, static_cast<size_t>(1))),
+                         ImVec2(-1.0f, 0.0f), "");
+    }
+    ImGui::End();
+
+    g_renderer->EndUIFrame();
+  };
+
   while (running &&
          (m_async_shader_compiler->HasPendingWork() || m_async_shader_compiler->HasCompletedWork()))
   {
-    running = m_async_shader_compiler->WaitUntilCompletion([](size_t completed, size_t total) {
-      g_renderer->BeginUIFrame();
+    running = m_async_shader_compiler->WaitUntilCompletion(update_ui_progress);
 
-      const float center_x = ImGui::GetIO().DisplaySize.x * 0.5f;
-      const float center_y = ImGui::GetIO().DisplaySize.y * 0.5f;
-      const float scale = ImGui::GetIO().DisplayFramebufferScale.x;
-
-      ImGui::SetNextWindowSize(ImVec2(400.0f * scale, 50.0f * scale), ImGuiCond_Always);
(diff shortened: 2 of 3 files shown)


8times9:
Will fix https://bugs.dolphin-emu.org/issues/11599

CookiePLMonster:
@stenzek left comments on this ticket so I think he's the most appropriate person to review this.

Blackbird88:
I have tested this and it works 👍

iwubcode:
@CookiePLMonster - I completely forgot this was out there.  Just because, can you rebase this on latest so I can test?

CookiePLMonster:
@iwubcode Done! Even though it's been years, rebase thankfully wasn't hard. I verified that after rebase compilation can still be interrupted. It's not instant because each worker is still given the chance to finish work that it's performing (as the code joins on worker threads), but any other pending work is being discarded correctly.

Source/Core/VideoCommon/AsyncShaderCompiler.cpp
```diff
@@ -65,17 +67,23 @@ bool AsyncShaderCompiler::HasCompletedWork()
...
 {
   while (HasPendingWork())
+  {
+    if (interruptable && Core::GetState() == Core::State::Stopping)
+      return false;
+
     std::this_thread::sleep_for(std::chrono::milliseconds(1));
+  }
+  return true;
 }
```
Dentomologist, line 81: This function is never called, you can just delete it.

Source/Core/VideoCommon/AsyncShaderCompiler.cpp
```diff
@@ -65,17 +67,23 @@ bool AsyncShaderCompiler::HasCompletedWork()
...
 
-void AsyncShaderCompiler::WaitUntilCompletion(
-    const std::function<void(size_t, size_t)>& progress_callback)
+bool AsyncShaderCompiler::WaitUntilCompletion(
+    bool interruptable, const std::function<void(size_t, size_t)>& progress_callback)
```
Dentomologist: interruptable is always going to be true; we can skip the parameter and just always check Core::GetState()

Dentomologist (changes requested):

Source/Core/VideoCommon/ShaderCache.cpp
```diff
@@ -157,32 +157,43 @@ const AbstractPipeline* ShaderCache::GetUberPipelineForUid(const GXUberPipelineU
...
-        ImGui::ProgressBar(static_cast<float>(completed) /
-                               static_cast<float>(std::max(total, static_cast<size_t>(1))),
-                           ImVec2(-1.0f, 0.0f), "");
-      }
-      ImGui::End();
+  bool running = true;
 
-      g_renderer->EndUIFrame();
-    });
+  while (running &&
```
iwubcode, line 162: I propose this:

```
constexpr auto update_ui_progress = [](size_t completed, size_t total) {
	g_renderer->BeginUIFrame();

	const float center_x = ImGui::GetIO().DisplaySize.x * 0.5f;
	const float center_y = ImGui::GetIO().DisplaySize.y * 0.5f;
	const float scale = ImGui::GetIO().DisplayFramebufferScale.x;

	ImGui::SetNextWindowSize(ImVec2(400.0f * scale, 50.0f * scale), ImGuiCond_Always);
	ImGui::SetNextWindowPos(ImVec2(center_x, center_y), ImGuiCond_Always, ImVec2(0.5f, 0.5f));
	if (ImGui::Begin(Common::GetStringT("Compiling Shaders").c_str(), nullptr,
				ImGuiWindowFlags_NoTitleBar | ImGuiWindowFlags_NoInputs |
					ImGuiWindowFlags_NoMove | ImGuiWindowFlags_NoSavedSettings |
					ImGuiWindowFlags_NoScrollbar | ImGuiWindowFlags_NoNav |
					ImGuiWindowFlags_AlwaysAutoResize |
					ImGuiWindowFlags_NoFocusOnAppearing))
	{
		ImGui::Text("Compiling shaders: %zu/%zu", completed, total);
		ImGui::ProgressBar(static_cast<float>(completed) /
						static_cast<float>(std::max(total, static_cast<size_t>(1))),
					ImVec2(-1.0f, 0.0f), "");
	}
	ImGui::End();

	g_renderer->EndUIFrame();
}

while(m_async_shader_compiler->HasPendingWork() || m_async_shader_compiler->HasCompletedWork())
{
	if (!m_async_shader_compiler->WaitUntilCompletion(update_ui_progress))
		break;

	m_async_shader_compiler->RetrieveWorkItems();
}

// Just render nothing to clear the screen
g_renderer->BeginUIFrame();
g_renderer->EndUIFrame();
```

My reasoning but maybe I missed something:

1)  Breaking out the ui update into a lambda makes it a little readable imo.
2)  If you are cancelling, I don't think you need to wait on the other work items
3)  Passing the lambda for the clear is not necessary, just call that after the function returns.  Additionally, I don't think you need to call it every iteration, just the last one.

iwubcode: (also included @Dentomologist 's suggestion about removing `interruptable` )

CookiePLMonster: `StopWorkerThreads()` [joins on all worker threads](https://github.com/dolphin-emu/dolphin/blob/f4e2d369c075f44f037289a1ccb78274c7d5cb39/Source/Core/VideoCommon/AsyncShaderCompiler.cpp#L174-L177), thus waits for them to gracefully quit. This effectively means we wait for the currently started chunks of work to finish, then discard the remainder of that pending work.

I'll address the other points, I like the idea of this lambda being separate indeed.

iwubcode (approved):
Code changes LGTM.  Tested and confirmed working.

CookiePLMonster:
Any idea if fifoci result is a false negative? I have a hard time believing this change would change rendering.

iwubcode:
@CookiePLMonster - I think it is, pokechu's recent PR also has that issue.

Dentomologist:
Setup: I press escape once while the shader compilation progress bar is active and Confirm on Stop is disabled. 

Master: The progress bar finishes all remaining in-progress and queued shaders. The emulated game then freezes for several minutes with the "Shutting Down" toast up, before finally quitting.

This PR: The progress bar finishes all remaining in-progress and queued shaders. The game then closes after several seconds.

Fixing the hang is clearly an improvement and I was ready to approve this PR, but after digging into it I'm not sure it's behaving the way it's supposed to. 
- It doesn't prevent queued (as opposed to in-progress) shaders from being compiled.
- The check for whether Core's state is Stopping is never true. I commented it out without changing the behavior, and I added a logging statement and breakpoint before it returns that I've never seen triggered.

The conditions and behavior surrounding Core's state changes are complicated enough that I'm not certain of this, but I think that it won't switch to Core::State::Stopping until after (among other things) the shader queue is empty.

CookiePLMonster:
> * It doesn't prevent queued (as opposed to in-progress) shaders from being compiled.

I verified this on Direct3D11 and by the time the worker threads have stopped, the pending work queue still had jobs to do.



> * The check for whether Core's state is Stopping is never true. I commented it out without changing the behavior, and I added a logging statement and breakpoint before it returns that I've never seen triggered.

That's weird since in my testing that wasn't true - I'd put breakpoints on `return false` from the core checks and they were hit.

iwubcode:
@Dentomologist  - I'm not sure what you mean by 'queued' shaders?  Or how you are determining those.

> Master: The progress bar finishes all remaining in-progress and queued shaders

For me, after clicking "stop", master will hang until it completely gets through the _entire_ progress bar.  On this PR, it is a mere couple seconds for how long it stops.

I initially tried on D3D11 but just tried on D3D12 and Vulkan as well and had the same results.

I have not yet tried stopping when the game is in progress and the shaders changed.  I will do that later.

> The check for whether Core's state is Stopping is never true

That seems surprising to me, given my results.  However, that does match stenzek's original comment on the issue ticket.

Dentomologist:
I'm on Windows 10 1903 (my Windows Update is broken), and my 770 died recently so I'm using an integrated Intel HD Graphics 4600.

Specialized (default) Shaders, Compile Shaders Before Starting, Confirm on Stop disabled.

My testing has been with OpenGL, Direct3D 11, and Direct3D 12 and I haven't seen any difference between them. I deleted my cache as needed to make compilation happen. Of the games I have Twilight Princess seems to have the longest compilation times so I used that.

I'm not certain what behavior you're seeing on master and what you're expecting in this PR, so here's what happens for me in a bunch of situations.  Stop = hit escape once, Force Stop = hit escape twice.

On Master:

- Wait for Intro cutscene (after the Wiimote warning), Stop. 
- -	Result: Game closes normally.
- Wait for Intro cutscene, toggle Per-Pixel Lighting to trigger recompilation, Stop after progress bar appears. 
- -	Result: Compilation continues, finishes the current bar (roughly 30 seconds), then starts and finishes one or two new bars. The game closes normally afterward.
- Same as above, but Force Stop:
- -	Result: Same as above.
- Change settings before opening game, start game, Stop.
- -	Result: Compilation and game progress as if I hadn't hit Stop. Stopping during the intro works normally (I turned on Confirm on Stop for this to verify it was a Stop rather than Force Stop)
- Same as above, but Force Stop:
- - 	Result: Also no effect, even after hitting escape repeatedly.

On PR:

- Wait for Intro cutscene (after the Wiimote warning), Stop. 
- - Result: Game closes normally. (Same as Master)
- Wait for Intro cutscene, toggle Per-Pixel Lighting to trigger recompilation, Stop after progress bar appears. 
- - Result: Window title immediately says Stopping, but compilation continues, finishes the current bar (roughly 30 seconds), then starts and finishes one or two new bars. The game closes normally afterward. (Same as master except for title bar changing)
- Same as above, but Force Stop:
- - Result: Shader compilation is almost immediately interrupted (I presume any delay is from the in-progress unit from each worker thread), then game closes normally.
- Change settings before opening game, start game, Stop.
- - Result: Compilation and game progress as if I hadn't hit Stop. Stopping during the intro works normally (I turned on Confirm on Stop for this to verify it was a Stop rather than Force Stop)
- Same as above, but Force Stop:
- - Result: Also no effect, even after hitting escape repeatedly.

The only part thus far where I saw the new conditional triggered is when I Force Stopped the in-game recompilation.

------
And at this point I reread iwubcode's post and noticed `clicking "stop"` instead of hitting escape like I had been. 

I tried clicking stop during the initial compilation, and it worked!

...then I tried it during the in-game recompilation and it didn't.

Do all my results match what you were expecting? I think we should make in-game recompilation interruptible too, but before I review any further I'd like to make sure I understand what's supposed to be happening first.

iwubcode:
@Dentomologist - yeah.  I am noticing similar results as you.

Clicking stop when the game is updating the shaders will go through the entire progress bar before stopping.

Also, I notice that stopping at startup will sometimes make the progress disappear but there will still be a black screen for sometime after that.  Not sure what is going on there.

iwubcode (changes requested):
Cookie - it'd be nice if we could fix both of these issues.

CookiePLMonster:
> Clicking stop when the game has started running and the shader ompilation is occurring will go through the entire progress bar before stopping.

I'll check - does the emu actually show the progress bar when compiling in game? Or does it only happen when synchronous shaders with no ubershaders are active?

> Also, I notice that stopping at startup will sometimes make the progress disappear but there will still be a black screen for sometime after that. Not sure what is going on there. I was not seeing that before.

I think that's just finishing work that's already started by the time compilation was interrupted - I know there is some brief black screen time after the worker threads tear down but I think it's unrelated to shaders.

Dentomologist:
> I'll check - does the emu actually show the progress bar when compiling in game? Or does it only happen when synchronous shaders with no ubershaders are active?

For all four Shader Compilation models, the progress bar can show up when you change a setting that requires recompiling existing shaders. I don't know if it would show up naturally during gameplay as you encounter new shaders, since there's a ~1 second delay before the progress bar appears, but in principle it's possible if the game tries to compile enough shaders at once.

Re: the stop hotkey:

Hotkeys are processed in Core/DolphinQt/HotkeyScheduler.cpp. Escape triggers HK_STOP on line 220, but line 189 checks if Core::IsRunningAndStarted(). If it's not, which is the case until after initial shader compilation is complete, it skips checking other hotkeys except HK_PLAY_RECORDING and so HK_STOP does nothing. If it was processed normally it would send a signal to the connected MainWindow::RequestStop.

Meanwhile, Menubar buttons are processed normally and the Stop button also has a signal connected to MainWindow::RequestStop, which is why clicking Stop works in the beginning.

Since they both call the same function and clicking the stop button works before Core's fully started, it should be safe to move the hotkey check before the Core check so that escape will work during startup.

Dentomologist (approved):
After looking into it more handling the in-game recompilation case is going to be tricky enough to be worth its own PR. The hotkey isn't essential and would fit in better with another PR too, so no need to delay this one any further.

iwubcode (approved):
While the lack of hotkey support is unfortunate and might be confusing to users, I think it's reasonable to save that for another PR (given how long this has sat).  Code wise I think this seems reasonable and I tested again and it works as expected.

AdmiralCurtiss:
Good enough for me.

PR #8468, opened 2019-11-10
UI: Fixes for exit dialog

CookiePLMonster:
As per discussion on IRC, this PR makes "Do you want to exit emulation?" application modal and ensures only a single instance of it exists. This fixes numerous issues, eg.:

- Being able to spam dialogs by pressing Escape
- Being able to close emulation window by just pressing X twice

### DIFF
--- a/Source/Core/DolphinQt/MainWindow.cpp
+++ b/Source/Core/DolphinQt/MainWindow.cpp
@@ -30,6 +30,7 @@
 #include <qpa/qplatformnativeinterface.h>
 #endif
 
+#include "Common/ScopeGuard.h"
 #include "Common/Version.h"
 #include "Common/WindowSystemInfo.h"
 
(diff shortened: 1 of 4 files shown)


Source/Core/DolphinQt/MainWindow.cpp
```diff
@@ -785,6 +786,11 @@ bool MainWindow::RequestStop()
 
   if (SConfig::GetInstance().bConfirmStop)
   {
+    if (std::exchange(m_stop_confirm_showing, true))
+      return true;
+
+    Common::ScopeGuard confirmLock([this] { m_stop_confirm_showing = false; });
```
leoetlino:
```suggestion
    Common::ScopeGuard confirm_lock([this] { m_stop_confirm_showing = false; });
```

CookiePLMonster: Oversight, will fix!

PR #8469, opened 2019-11-11
Fix malformed manifests

CookiePLMonster:
Fixes a regression caused by #8393. 2016 manifest XML schema removed a 2005 `dpiAware` attribute, which caused Dolphin's manifest to silently ignore this attribute, but in the case of updater it exploded loudly instead.

This PR fixes both manifests. Sadly, since updater cannot launch it means that those unlucky people who updated during the past 1.5 days have to update manually.


Blackbird88:
I have tested this and it works 👍

JosJuice:
Going to do a quick merge of this so that we'll avoid having even more people update to a broken updater.

RinMaru:
>those unlucky people who updated during the past 1.5 days have to update manually.

yea thanks allot lol

PR #7864, opened 2019-03-07
Bumps compile API to 29 (Q) and removes deprecated calls and unnecessary casts

rafaeltoledo:
This will help the Android client to evolve with the latest libraries (as the legacy support libs will not be shipped anymore with the com.android.support package).

This PR also makes the app compliant with the new API requirements to start later this year:
https://android-developers.googleblog.com/2019/02/expanding-target-api-level-requirements.html

PR #8471, opened 2019-11-15
Android: migrate to AndroidX

Simonx22:
Support Libraries are outdated and AndroidX is recommended instead. Read more here: https://developer.android.com/jetpack/androidx

The AndroidX migration in #7864 is outdated so I just opened a new PR for the migration.


Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/services/GameFileCacheService.java
```diff
@@ -1,179 +1,180 @@
-package org.dolphinemu.dolphinemu.services;
```
JosJuice: Seems like the reason we have such a big diff here is because this file is using CRLF in master... My bad :(

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/viewholders/TvSettingsViewHolder.java.orig
```diff
@@ -0,0 +1,22 @@
+package org.dolphinemu.dolphinemu.viewholders;
```
JosJuice: This file should be deleted, right?

Simonx22: Whoops, no idea how that happened but I've deleted the file now

JosJuice:
@dolphin-emu-bot rebuild

PR #8473, opened 2019-11-17
ControllerInterface: Combine evdev devices with the same unique ID.

jordan-woyak:
Unfortunately newer Linux drivers for DS4 (Playstation 4) controllers split the device into three separate event nodes.

On my system I get:
"Sony Interactive Entertainment Wireless Controller"
"Sony Interactive Entertainment Wireless Controller Touchpad"
"Sony Interactive Entertainment Wireless Controller Motion Sensors"

Apparently this is due to some evdev policy: "A device must not mix regular directional axes and accelerometer axes on the same event node."

Our input system does allow use of multiple devices via "All Devices" or so-called "absolute path" control expressions, but this is less than desirable especially when making use of multiple DS4s.

This PR combines evdev devices that return the same non-null unique ID (via libevdev_get_uniq).

Mapping of DS4 acceleration, gyroscope, and touchpad inputs is now less horrible.

To prevent collisions of input names in combined devices I've made a few changes:
- If the `INPUT_PROP_BUTTONPAD` property is set then only X/Y axes are added with "Cursor X/Y" names. (Our default mouse mappings now work with the DS4 touchpad)
- Axes beyond `ABS_MISC` are no longer enumerated as they relate to multi-touch and require tracking and special processing of events for useful data.
- Buttons now take on their `libevdev_event_code_get_name` names in more situations rather than numbered "Button 0" names. (Existing normal button configurations will still work by overriding `IsMatchingName`. Touchpad inputs will need to be remapped but this is unavoidable as the names clashed.)

### DIFF
--- a/Source/Core/InputCommon/ControllerInterface/evdev/evdev.cpp
+++ b/Source/Core/InputCommon/ControllerInterface/evdev/evdev.cpp
@@ -251,8 +251,13 @@ void AddDeviceNode(const char* devnode)
 
     evdev_device->AddNode(devnode, fd, dev);
 
-    // Callbacks must be invoked as the device name and available inputs may change.
-    g_controller_interface.InvokeDevicesChangedCallbacks();
+    // Remove and re-add device as naming and inputs may have changed.
+    // This will also give it the correct index and invoke device change callbacks.
+    g_controller_interface.RemoveDevice([&evdev_device](const auto* device) {
+      return static_cast<const evdevDevice*>(device) == evdev_device.get();
+    });
+
+    g_controller_interface.AddDevice(evdev_device);
   }
   else
   {
--- a/Source/Core/InputCommon/ControllerInterface/evdev/evdev.cpp
+++ b/Source/Core/InputCommon/ControllerInterface/evdev/evdev.cpp
@@ -243,9 +243,12 @@ void AddDeviceNode(const char* devnode)
     return;
   }
 
-  auto evdev_device = FindDeviceWithUniqueID(libevdev_get_uniq(dev));
+  const auto uniq = libevdev_get_uniq(dev);
+  auto evdev_device = FindDeviceWithUniqueID(uniq);
   if (evdev_device)
   {
+    NOTICE_LOG(SERIALINTERFACE, "evdev combining devices with unique id: %s", uniq);
+
     evdev_device->AddNode(devnode, fd, dev);
 
     // Callbacks must be invoked as the device name and available inputs may change.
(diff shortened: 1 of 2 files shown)


Source/Core/InputCommon/ControllerInterface/evdev/evdev.cpp
```diff
@@ -35,42 +35,76 @@ class Input : public Core::Device::Input
...
 };
 
-class Button final : public Input
+class Button : public Input
 {
 public:
   Button(u8 index, u16 code, libevdev* dev) : Input(code, dev), m_index(index) {}
 
-  std::string GetName() const override
+  ControlState GetState() const final override
```
CookiePLMonster: Since the class is marked `final`, marking this method as such is redundant.

Source/Core/InputCommon/ControllerInterface/evdev/evdev.cpp
```diff
@@ -190,28 +301,21 @@ static void HotplugThreadFunc()
...
-        // We don't know the name for this device, so it is probably not an evdev device.
-        continue;
-      }
+      std::shared_ptr<evdevDevice> ptr;
 
-      const std::string& name = it->second;
-      g_controller_interface.RemoveDevice([&name](const auto& device) {
-        return device->GetSource() == "evdev" && device->GetName() == name && !device->IsValid();
-      });
+      const auto it = s_devnode_objects.find(devnode);
```
CookiePLMonster: `devnode` is a `const char*`, while `s_devnode_objects`'s key is `std::string`. This will construct a temporary `std::string` , unless you make `s_devnode_objects` use a transparent comparator (`std::less<>`).

Source/Core/InputCommon/ControllerInterface/evdev/evdev.cpp
```diff
@@ -35,42 +35,76 @@ class Input : public Core::Device::Input
...
+      if (name.find(remove_prefix) == 0)
+        return std::string(name.substr(std::strlen(remove_prefix)));
+
+      return std::string(name);
     }
-    // But controllers use codes above 0x100, and the standard label often doesn't match.
-    // We are better off with Button 0 and so on.
-    return "Button " + std::to_string(m_index);
+    else
+      return std::nullopt;
```
BhaaLseN: This `else` needs curlies, since the `if` has some.

Source/Core/InputCommon/ControllerInterface/evdev/evdev.cpp
jordan-woyak: This was also done before I touched the code. There's nothing too wrong with a temporary `std::string`.

lioncash:
The first commit should probably include most of the description put in the main PR comments, so that context is actually stored in the repository history.

PR #8474, opened 2019-11-17
ControllerInterface: Exposse DSU client battery level as an input.

jordan-woyak:
Tested with an Android DSU Server.

I've named the input "Battery". It should be visible in the "Advanced" mapping dialog.
If someone prefers "Battery Level" or something else I don't have a strong opinion.

This is mainly intended to be used with PR #8417 to pass-through the battery level to an emulated Wii remote.

### DIFF
--- a/Source/Core/InputCommon/ControllerInterface/Device.h
+++ b/Source/Core/InputCommon/ControllerInterface/Device.h
@@ -22,10 +22,17 @@ namespace ciface
 // range, used for periodic haptic effects though often ignored by devices
 // TODO: Make this configurable.
 constexpr int RUMBLE_PERIOD_MS = 10;
+
 // This needs to be at least as long as the longest rumble that might ever be played.
 // Too short and it's going to stop in the middle of a long effect.
 // Infinite values are invalid for ramp effects and probably not sensible.
 constexpr int RUMBLE_LENGTH_MS = 1000 * 10;
+
+// All inputs (other than accel/gyro) return 1.0 as their maximum value.
+// Battery inputs will almost always be mapped to the "Battery" setting which is a percentage.
+// If someone actually wants to map a battery input to a regular control they can divide by 100.
+// I think this is better than requiring multiplication by 100 for the most common usage.
+constexpr ControlState BATTERY_INPUT_MAX_VALUE = 100.0;
 
 namespace Core
 {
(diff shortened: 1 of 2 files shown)


Source/Core/InputCommon/ControllerInterface/Device.h
```diff
@@ -22,11 +22,18 @@ namespace ciface
...
 // TODO: Make this configurable.
 constexpr int RUMBLE_PERIOD_MS = 10;
+
 // This needs to be at least as long as the longest rumble that might ever be played.
 // Too short and it's going to stop in the middle of a long effect.
 // Infinite values are invalid for ramp effects and probably not sensible.
 constexpr int RUMBLE_LENGTH_MS = 1000 * 10;
 
+// All inputs (other than accel/gyro) return 1.0 as their maximum value.
+// Battery inputs will almost always be mapeed to the "Battery" setting which is a percentage.
```
BhaaLseN:
```suggestion
// Battery inputs will almost always be mapped to the "Battery" setting which is a percentage.
```

jordan-woyak: Oops. :P

AdmiralCurtiss:
Would it make more sense to define 1.0 as maximum battery on the emulated Wii remote side so we don't have to special case the input here?

jordan-woyak:
> Would it make more sense to define 1.0 as maximum battery on the emulated Wii remote side so we don't have to special case the input here?

I've mentioned in the comment above the `constexpr` defintion why I think 100 is a better decision.
No one is ever going to normally map this "Input" to a button press and if they do that can be smart enough to divide by 100.
People are going to map this to the "Battery" setting 99% of the time which will expect 100 as the maximum value. Multiplying by 100 will be annoying and confusing to users.

AdmiralCurtiss:
No, what I mean is, why have the 'Battery' setting expect 100 as the maximum value instead of 1.

jordan-woyak:
> No, what I mean is, why have the 'Battery' setting expect 100 as the maximum value instead of 1.

Oh.. :P

Well.. the setting already exists as a percentage. I think that's a more friendly interface for people not doing battery "pass through". It would require breaking configs or even more special-case handling to change it now. I think this is the lesser evil. I don't think it's too horrible to have a very specific kind of input take on a different maximum value. Accelerometer and gyroscope inputs return m/s/s and rad/s, similarly.

AdmiralCurtiss:
Ah, fair enough, I didn't realize that Battery setting already existed. In that case yeah this is probably fine.

rlnilsen:
@jordan-woyak I'm glad _you_ did this, as my brain is mush right now (and I don't know how long it will last). I've looked over the changes and there's nothing jumping out at me, although that might not be anything to go by.

mbc07:
Tested with both DS4 and Wiimote DSU Servers and works as expected. Sweet!

JMC47:
This needs to be rebased now.

PR #8479, opened 2019-11-19
Fixed #11874 (leading 0s ignored by debugger)

nokturnusmf:
Fixes both issues from the report.

For set value:
The original code converts the string into a value and then attempts to work out the size of the input by downcasting and comparing. This breaks for leading zeros as, for example, 0042 == (u8)0042. The fix calculates the correct length from the original string, taking into account the possibility that the string is prefixed by "0x".

For copy hex:
The code already attempts to take the first "length" characters, however incorrectly converts/pads the value and so copies the first "length" nonzero characters. This meant that copying the 01 of 01 23 would result in the string 12 being copied. The fix always fully pads the read value and then correctly takes the leftmost characters.

### DIFF
--- a/Source/Core/DolphinQt/Debugger/MemoryWidget.cpp
+++ b/Source/Core/DolphinQt/Debugger/MemoryWidget.cpp
@@ -488,12 +488,15 @@ void MemoryWidget::OnSetValue()
     const QByteArray bytes = m_data_edit->text().toUtf8();
 
     for (char c : bytes)
-      accessors->WriteU8(static_cast<u8>(c), addr++);
+      accessors->WriteU8(addr++, static_cast<u8>(c));
   }
   else
   {
     bool good_value;
-    u64 value = m_data_edit->text().toULongLong(&good_value, 16);
+    const QString text = m_data_edit->text();
+    const int length =
+        text.startsWith(QStringLiteral("0x"), Qt::CaseInsensitive) ? text.size() - 2 : text.size();
+    const u64 value = text.toULongLong(&good_value, 16);
 
     if (!good_value)
     {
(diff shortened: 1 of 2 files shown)


Source/Core/DolphinQt/Debugger/MemoryWidget.cpp
```diff
@@ -493,23 +493,25 @@ void MemoryWidget::OnSetValue()
   else
   {
     bool good_value;
-    u64 value = m_data_edit->text().toULongLong(&good_value, 16);
+    const auto& text = m_data_edit->text();
+    int length = text.startsWith(QStringLiteral("0x")) ? text.size() - 2 : text.size();
```
CookiePLMonster: What about uppercase `0X`?

Source/Core/DolphinQt/Debugger/MemoryWidget.cpp
```diff
@@ -493,23 +493,25 @@ void MemoryWidget::OnSetValue()
   else
   {
     bool good_value;
-    u64 value = m_data_edit->text().toULongLong(&good_value, 16);
+    const auto& text = m_data_edit->text();
```
lioncash: `text()` returns a QString by value, so taking it by const reference, while valid, is somewhat visually misleading.

Source/Core/DolphinQt/Debugger/MemoryWidget.cpp
lioncash: This and `value` can both be made const.

nokturnusmf:
@CookiePLMonster Yep, complete oversight.

@lioncash So it is, I'm not too familiar with Qt so just assumed.

I additionally noticed and fixed a typo in the code for setting ASCII values, the arguments to WriteU8 were swapped.

CookiePLMonster:
Looks good now, but please squash so those "PR feedback" commits disappear.

CookiePLMonster:
@dolphin-emu-bot rebuild

Helios747:
@dolphin-emu-bot rebuild

PR #8480, opened 2019-11-22
VolumeVerifier: Fix a copypaste error

JosJuice:
This was completely wrong, but I guess it happened to work correctly anyway due to timing reasons (MD5 is faster than SHA-1 + AES).


JosJuice:
I just realized the actual reason for why this worked before the change. When we overwrite the existing future, it actually blocks for that existing future to finish executing before the assignment completes. So there was no race condition, but instead, there was an unnecessary performance penalty.

PR #8481, opened 2019-11-22
Android: revert to AppCompat

Simonx22:
before:

image

after:

image


Simonx22:
#8482 seems like a better solution. Closing!

PR #8483, opened 2019-11-22
InputCommon: Make use of fmt where applicable

lioncash:
Continues the migration over to fmt


leoetlino:
Looks like the PR is based on an old-ish commit. Please rebase and fix the build issues that will most likely appear after doing so. (I think the only change you have to make is replace fmt/time.h with fmt/chrono.h.)

PR #8487, opened 2019-11-22
VideoCommon: Make use of fmt outside of shader generators

lioncash:
Migrates most of VideoCommon over to using fmt, with the exception being the shader generator code. The shader generators are quite large and have more corner cases to deal with in terms of conversion (shaders have braces in them, so we need to make sure to escape them).

Because of the large amount of code that would need to be converted, the conversion of VideoCommon will be in two parts:

- This change (which converts over the general case string formatting),
- A follow up change that will specifically deal with converting over the shader generators.

### DIFF
--- a/Source/Core/VideoCommon/BPStructs.cpp
+++ b/Source/Core/VideoCommon/BPStructs.cpp
@@ -8,9 +8,10 @@
 #include <cstring>
 #include <string>
 
+#include <fmt/format.h>
+
+#include "Common/CommonTypes.h"
 #include "Common/Logging/Log.h"
-#include "Common/StringUtil.h"
-#include "Common/Thread.h"
 #include "Core/ConfigManager.h"
 #include "Core/CoreTiming.h"
 #include "Core/FifoPlayer/FifoPlayer.h"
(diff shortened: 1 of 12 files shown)


Source/Core/VideoCommon/PostProcessing.cpp
```diff
@@ -286,8 +288,10 @@ void PostProcessingConfiguration::SaveOptionsConfiguration()
     {
       std::string value;
       for (size_t i = 0; i < it.second.m_integer_values.size(); ++i)
-        value += StringFromFormat("%d%s", it.second.m_integer_values[i],
-                                  i == (it.second.m_integer_values.size() - 1) ? "" : ", ");
+      {
+        value += fmt::format("{}{}", it.second.m_integer_values[i],
```
BhaaLseN: Would this be a potential candidate for an algorithm? `std::accumulate` comes to mind, or whatever else comes close to the C# `string.Join` method.

Source/Core/VideoCommon/ShaderGenCommon.cpp
```diff
@@ -76,7 +78,7 @@ std::string GetDiskShaderCacheFileName(APIType api_type, const char* type, bool
   {
     // We're using 21 bits, so 6 hex characters.
     ShaderHostConfig host_config = ShaderHostConfig::GetCurrent();
-    filename += StringFromFormat("-%06X", host_config.bits);
+    filename += fmt::format("{:<06X}", host_config.bits);
```
BhaaLseN: Is this really the same as the previous code? The previous one had a dash followed by a 6-digit hex number; but the open angle bracket is just alignment within the available space (as far as I can tell from the `fmt` docs). I'm guessing (without having run the code) that this is now missing the dash.

Source/Core/VideoCommon/BPStructs.cpp
```diff
@@ -921,19 +922,19 @@ void GetBPRegInfo(const u8* data, std::string* name, std::string* desc)
     const char* logicmodes[] = {"0",     "s & d",  "s & ~d",   "s",        "~s & d", "d",
```
BhaaLseN: Just wondering, would this make any difference being `constexpr` (and/or `std::array`, or anything other than `const char*[]`)?

lioncash: There would likely not be much of a difference here.

Source/Core/VideoCommon/PostProcessing.cpp
lioncash: I'm not sure using an algorithm here would make the readability much nicer. If the values were already strings, then the whole function could be collapsed down into a call to `JoinStrings()` in StringUtil, but they're `s32`s.

PR #8488, opened 2019-11-23
Simplify wstring to QString conversion

tinyredpanda:


CookiePLMonster:
@dolphin-emu-bot rebuild

CookiePLMonster:
While you're at it, `CompatPatches.cpp:239` converts `event.name` from `wstring` to `string` and that looks fairly useless. Do you want to take a look?

tinyredpanda:
I had a look and it seems that `fmt::format` doesn't support wide strings.
The compiler error message is
`'return': cannot convert from 'fmt::v5::basic_string_view<wchar_t>' to 'fmt::v5::internal::init<Context,fmt::v5::basic_string_view<char>,fmt::v5::internal::string_type>'`.

CookiePLMonster:
Indeed, but if you also provide a wide format string and use `MessageBoxW` (which is preferable over `MessageBoxA` anyway) then it compiles.

BhaaLseN:
Is this really the same? I remeber there being a reason why we needed `UTF16ToUTF8`...but that might be a legacy thing where this method didn't exist (in which case there are likely others that can also use the same treatment).

tinyredpanda:
I already checked for other wstrings that are (indirectly) converted to QString, it's just this one.

Source/Core/DolphinQt/Main.cpp
```diff
@@ -104,16 +102,16 @@ int main(int argc, char* argv[])
 
 #ifdef _WIN32
   // Get the default system font because Qt's way of obtaining it is outdated
-  NONCLIENTMETRICS metrics = {};
-  LOGFONT& logfont = metrics.lfMenuFont;
-  metrics.cbSize = sizeof(NONCLIENTMETRICS);
+  NONCLIENTMETRICSW metrics = {};
+  LOGFONTW& logfont = metrics.lfMenuFont;
+  metrics.cbSize = sizeof(NONCLIENTMETRICSW);
```
CookiePLMonster:
```suggestion
  metrics.cbSize = sizeof(metrics);
```

CookiePLMonster:
@dolphin-emu-bot rebuild

CookiePLMonster:
@dolphin-emu-bot rebuild

CookiePLMonster:
Please squash `Format fix` with the commit which touched `fmt::format` first.

CookiePLMonster:
@dolphin-emu-bot rebuild

CookiePLMonster:
@dolphin-emu-bot rebuild

PR #8494, opened 2019-11-24
Android: Fix displaying checkbox settings with no description

JosJuice:
https://bugs.dolphin-emu.org/issues/11904


Helios747:
Meh it's not like they'd read whats there anyways

PR #8495, opened 2019-11-26
Platform support for Windows-on-ARM64

stenzek:
What the title says. Enables Dolphin to run natively on Windows-on-ARM64 devices, such as the Surface Pro X. Most of the lines changed are in the vcxproj files...

I don't have a device to test, but some users have reported that these changes do work.

No Qt build yet as I've had issues compiling Qt for ARM64, so there's only NoGUI for now. If we do want to merge this, we'd want to split out the VS2019 update and the Win32 NoGUI platform.


Source/Core/Core/PowerPC/JitArm64/JitArm64_Integer.cpp
```diff
@@ -368,7 +368,17 @@ void JitArm64::cntlzwx(UGeckoInstruction inst)
 
   if (gpr.IsImm(s))
   {
-    gpr.SetImmediate(a, __builtin_clz(gpr.GetImm(s)));
+    const u32 imm = gpr.GetImm(s);
+#ifdef _MSC_VER
```
degasus: Please just avoid the intrinsic here. This can be done in pure C++

Source/Core/DolphinNoGUI/PlatformWin32.cpp
```diff
@@ -0,0 +1,281 @@
...
+  {
+  case WM_NCCREATE:
+  {
+    platform =
+        reinterpret_cast<PlatformWin32*>(reinterpret_cast<CREATESTRUCT*>(lParam)->lpCreateParams);
+    SetWindowLongPtr(hwnd, GWLP_USERDATA, reinterpret_cast<LONG_PTR>(platform));
+    return DefWindowProc(hwnd, msg, wParam, lParam);
+  }
+
+#if 0
```
degasus: dead code?

Externals/LZO/lzodefs.h
```diff
@@ -692,7 +692,7 @@
 #  else
 #    define LZO_INFO_ARCH           "arm"
 #  endif
-#elif defined(__arm__) || defined(_M_ARM)
+#elif defined(__arm__) || defined(_M_ARM) || defined(_M_ARM64)
```
cremno: What about updating the bundled LZO files instead? The current version seems to support arm64.

Source/VSProps/Base.props
```diff
@@ -95,6 +95,8 @@
       <AdditionalOptions>/Zo %(AdditionalOptions)</AdditionalOptions>
       <!--Treat sources as utf-8-->
       <AdditionalOptions>/utf-8 %(AdditionalOptions)</AdditionalOptions>
+      <!--Treat sources as utf-8-->
+      <AdditionalOptions>/utf-8 %(AdditionalOptions)</AdditionalOptions>
```
JosJuice: Aren't these two lines the same as the previous two lines?

Externals/Bochs_disasm/Bochs_disasm.vcxproj
```diff
@@ -1,23 +1,31 @@
...
+    </ProjectConfiguration>
     <ProjectConfiguration Include="Release|x64">
       <Configuration>Release</Configuration>
       <Platform>x64</Platform>
     </ProjectConfiguration>
   </ItemGroup>
   <PropertyGroup Label="Globals">
     <ProjectGuid>{8ADA04D7-6DB1-4DA4-AB55-64FB12A0997B}</ProjectGuid>
-    <WindowsTargetPlatformVersion>10.0.17134.0</WindowsTargetPlatformVersion>
+    <WindowsTargetPlatformVersion>10.0</WindowsTargetPlatformVersion>
```
BhaaLseN: Note: this requires a recent MSBuild (and/or Visual Studio 2019?) Version to work correctly. This basically means "use the latest available 10.0 SDK" while not limiting it to a specific build; but older versions of MSBuild/VS don't support this and simply fail with "Cannot find Windows SDK version 10.0" instead.
I _think_ this needs at least 16.3, but I don't remember exactly if it was _just_ any 16.x (since I've seen similar failures while attempting to do this with VS 2019 on the developer machine but VS 2017 build tools on the CI builders)

This ofc. goes for all the projects, not just this one.

Source/Core/DolphinQt/DolphinQt.vcxproj
```diff
@@ -473,10 +482,10 @@
     <ProjectReference Include="$(CoreDir)VideoBackends\D3D\D3D.vcxproj">
       <Project>{96020103-4ba5-4fd2-b4aa-5b6d24492d4e}</Project>
     </ProjectReference>
-    <ProjectReference Include="$(CoreDir)VideoBackends\OGL\OGL.vcxproj">
+    <ProjectReference Include="$(CoreDir)VideoBackends\OGL\OGL.vcxproj" Condition="'$(Platform)'!='ARM64'">
```
BhaaLseN: I get the feeling you wrote this one by hand, while all the others are VS generated. All the others could be more compact if they just did a Platform check instead of a combined Platform/Configuration one; but VS won't do that using the UI.

Source/Core/VideoCommon/VideoBackendBase.cpp
```diff
@@ -182,13 +189,17 @@ u16 VideoBackendBase::Video_GetBoundingBox(int index)
...
   // OGL > D3D11 > Vulkan > SW > Null
+#ifdef HAS_OPENGL
   g_available_video_backends.push_back(std::make_unique<OGL::VideoBackend>());
+#endif
 #ifdef _WIN32
   g_available_video_backends.push_back(std::make_unique<DX11::VideoBackend>());
   g_available_video_backends.push_back(std::make_unique<DX12::VideoBackend>());
 #endif
   g_available_video_backends.push_back(std::make_unique<Vulkan::VideoBackend>());
+#ifdef HAS_OPENGL
```
BhaaLseN: Uh, why is the Software Renderer dependant on OpenGL? That sounds...odd?

JosJuice: It uses OpenGL for scanout but not rendering.

Externals/Bochs_disasm/Bochs_disasm.vcxproj
stenzek: The builders have been updated to VS2019, so I don't think this should be a problem.

stenzek:
> With the SLN and everything updated to Visual Studio 2019, is there any particular reason to stick to the v142 toolset (rather than using v150 or higher)? This is mainly for binary compatibility if 3rd party components aren't updated to the new toolset yet; and I don't recall any binaries in there that we didn't precompile ourselves.

Isn't VS2019 `v142`?

Source/VSProps/Base.props
stenzek: Indeed. Not sure how that snuck in there.

Externals/Bochs_disasm/Bochs_disasm.vcxproj
BhaaLseN: Good to know. But now that you mention this; I _think_ we also need to update the developer requirements in our README. Or does this still build on VS 2017 without installing additional things?

BhaaLseN:
> 
> Isn't VS2019 `v142`?

Well yeah, not sure why I had v150 in my head; but you're right. The v14x series is simply binary compatible, while v140 came with VS 2015, vs141 with 2017 and v142 with 2019.
I _think_ during previews they had a v150 toolset with newer C++ Standard support; but I guess they figured out how to include that in the same runtime.

Source/Core/Core/PowerPC/JitArm64/JitArm64_Integer.cpp
shuffle2: what will generate `clz`?

Source/Core/VideoCommon/VideoBackendBase.cpp
```diff
@@ -17,14 +17,21 @@
 #include "Core/ConfigManager.h"
 #include "Core/Core.h"
 
+// OpenGL is not available on Windows-on-ARM64
+#if !defined(_WIN32) || !defined(_M_ARM64)
```
shuffle2: this doesn't seem like a good way/place to check this (e.g. for some non-windows platform that wants to disable ogl). but i guess not a big deal.

Source/Core/Common/Arm64Emitter.cpp
```diff
@@ -4164,7 +4168,7 @@ void ARM64XEmitter::ANDSI2R(ARM64Reg Rd, ARM64Reg Rn, u64 imm, ARM64Reg scratch)
 void ARM64XEmitter::AddImmediate(ARM64Reg Rd, ARM64Reg Rn, u64 imm, bool shift, bool negative,
                                  bool flags)
 {
-  switch ((negative << 1) | flags)
+  switch ((u8(negative) << 1) | u8(flags))
```
shuffle2: probably better to just use `if`s here.

Source/Core/Core/Core.vcxproj
```diff
@@ -539,20 +658,82 @@
...
-    <ClInclude Include="PowerPC\Jit64Common\BlockCache.h" />
-    <ClInclude Include="PowerPC\Jit64Common\EmuCodeBlock.h" />
-    <ClInclude Include="PowerPC\Jit64Common\FarCodeCache.h" />
-    <ClInclude Include="PowerPC\Jit64Common\Jit64AsmCommon.h" />
-    <ClInclude Include="PowerPC\Jit64Common\Jit64Constants.h" />
-    <ClInclude Include="PowerPC\Jit64Common\Jit64PowerPCState.h" />
-    <ClInclude Include="PowerPC\Jit64Common\TrampolineCache.h" />
-    <ClInclude Include="PowerPC\Jit64Common\TrampolineInfo.h" />
+    <ClInclude Include="PowerPC\Jit64Common\ConstantPool.h">
+      <ExcludedFromBuild Condition="'$(Configuration)|$(Platform)'=='Debug|ARM64'">true</ExcludedFromBuild>
```
shuffle2: all these duplicated lines can be reduced by just checking `$(Platform)`.

Source/Core/Core/PowerPC/JitArm64/JitArm64_Branch.cpp
```diff
@@ -160,12 +160,12 @@ void JitArm64::bcx(UGeckoInstruction inst)
   if (js.op->branchIsIdleLoop)
   {
     // make idle loops go faster
-    ARM64Reg WA = gpr.GetReg();
+    ARM64Reg WA2 = gpr.GetReg();
```
shuffle2: i assume this change is to quiet a shadowed variable warning. are you sure the behavior is correct? the following line uses `WA`.

Source/Core/Core/PowerPC/JitArm64/JitArm64_FloatingPoint.cpp
```diff
@@ -46,8 +46,12 @@ void JitArm64::fp_arith(UGeckoInstruction inst)
     VA = reg_encoder(fpr.R(a, type));
     if (use_b)
       VB = reg_encoder(fpr.R(b, type));
+    else
+      VB = {};
```
shuffle2: assuming these changes quiet uninitialized variable warning. consider just initializing in ctor / at decl instead.

Source/Core/DolphinQt/DolphinQt.vcxproj.user
```diff
@@ -3,6 +3,6 @@
   <PropertyGroup>
     <LocalDebuggerCommand>$(BinaryOutputDir)$(TargetFileName)</LocalDebuggerCommand>
     <DebuggerFlavor>WindowsLocalDebugger</DebuggerFlavor>
-    <ShowAllFiles>true</ShowAllFiles>
+    <ShowAllFiles>false</ShowAllFiles>
```
shuffle2: I would assume this is not wanted

shuffle2:
I would suggest not adding the NoUI project, and just waiting for / using arm64 Qt.
Besides my other comments and the numerous casts of questionable style, looks like nice addition.

shuffle2:
Is the Qt project actually working on win/arm64? I haven't found any bug tracker entry.

degasus:
@shuffle2 I've asked on IRC, and I was just told that Qt on ARM is working.

shuffle2:
@degasus yes I've heard the same thing (since my previous post here), although I haven't tried myself yet. Hopefully they will start posting installers, etc. on their downloads page, though?

I have a win/arm64 device, hopefully I can check this out soon.

Source/Core/VideoCommon/VideoBackendBase.cpp
stenzek: Agreed, we probably should move it to cmake (and make all backends selectable at compile time). But IMO that's a separate change.

Source/Core/DolphinQt/DolphinQt.vcxproj
stenzek: Done.

Source/Core/Core/PowerPC/JitArm64/JitArm64_Integer.cpp
stenzek: I found `_CountLeadingZeros()` in `intrin.h`. But yeah, could be worth adding a helper in `BitUtils` in the future.

stenzek:
When I tried cross-compiling Qt for ARM64, I had no luck. After hacking the scripts a bit (to force using x64 tools rather than ARM64), I was able to build a winrt version but not win32.

Maybe it only works when building natively or on the master version or something...

Source/Core/VideoCommon/VideoCommon.vcxproj
```diff
@@ -41,7 +49,10 @@
     <ClCompile Include="AbstractTexture.cpp" />
     <ClCompile Include="AsyncRequests.cpp" />
     <ClCompile Include="AsyncShaderCompiler.cpp" />
-    <ClCompile Include="FrameDump.cpp" />
+    <ClCompile Include="FrameDump.cpp">
```
shuffle2: a duplicate that slipped through :)

shuffle2:
Here's what I did to compile Qt for arm64 (host machine is x64, target arm64). (sorry for chickenscratch..)

ensure jom, perl and python are in PATH
python must be as 'python'; Qt does not detect 'py'
for reference i used jom 1.1.3, strawberry perl 5.30.1, python 3.7.4, vs2019 16.4.2

clone Qt git
checkout 5.12 (apparently dolphin uses 5.13.1 now, so could use that instead)
perl init-repository
cd ..
mkdir qt5_x86
mkdir qt5_arm64

open "Developer Command Prompt for VS 2019"
cd to qt5_x86 dir and make the host tools
..\qt5\configure -opensource -confirm-license -nomake tests -nomake examples -mp -developer-build
jom module-qtbase

open "Developer Command Prompt for VS 2019"
  yes a different one, to ensure env doesn't get wonky
cd to qt5_arm64 dir and make the target
..\qt5\configure -opensource -confirm-license -nomake tests -nomake examples -skip webengine -opengl es2 -mp -xplatform win32-arm64-msvc2017 -external-hostbindir c:\src\qt5_x86\qtbase\bin
jom
jom install

notes:
the jom stage failed for me near the end of the build. I just re-ran jom and the build completed fine. I suspect some parallel-build bug.
try building dolphin even if jom install fails...
-opengl es2 tells Qt to use ANGLE. Someone should probably look what makes the most sense for dolphin to use (we at least want Qt UI to be accelerated, and hopefully Software backend can be compatible).
ANGLE requires adding
<AdditionalIncludeDirectories>$(QtIncludeDir)QtANGLE;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
to QtCompile.props
and copying libGLESv2.dll from Qt bin dir to dolphin's Binary dir.

Then, setting QTDIR to the jom install output dir and enabling Dolphin vcxproj results in working build :)

Annotation 2019-12-17 201727

shuffle2:
On OGL:

> Note: As of Qt 5.5 this is the configuration used by the official, pre-built binary packages of Qt. It is strongly recommended to use it also in custom builds, especially for Qt binaries that are deployed alongside applications.

https://doc.qt.io/qt-5/windows-requirements.html
so perhaps this part of my config can be ignored and a default build just does the correct thing. But in that case, dolphin may want to explicitly choose `Qt::AA_UseOpenGLES` at runtime.

I haven't tried getting Software renderer to work, would you expect issues with the available API set?

Source/UnitTests/UnitTests.vcxproj
```diff
@@ -45,8 +53,9 @@
         The following libs are needed since we pull in pretty much the entire
         dolphin codebase.
         -->
-      <AdditionalLibraryDirectories>$(ExternalsDir)ffmpeg\lib;%(AdditionalLibraryDirectories)</AdditionalLibraryDirectories>
-      <AdditionalDependencies>avrt.lib;iphlpapi.lib;winmm.lib;setupapi.lib;opengl32.lib;glu32.lib;rpcrt4.lib;comctl32.lib;avcodec.lib;avformat.lib;avutil.lib;swresample.lib;swscale.lib;Shlwapi.lib;%(AdditionalDependencies)</AdditionalDependencies>
+      <AdditionalDependencies>avrt.lib;iphlpapi.lib;winmm.lib;setupapi.lib;rpcrt4.lib;comctl32.lib;Shlwapi.lib;%(AdditionalDependencies)</AdditionalDependencies>
+      <AdditionalDependencies Condition="'$(Platform)'=='x64'">opengl32.lib;avcodec.lib;avformat.lib;avutil.lib;swresample.lib;swscale.lib;%(AdditionalDependencies)</AdditionalDependencies>
```
shuffle2: excluding the ffmpeg libraries seems to only be required because dolphin doesn't build this External from source. 
Here is a third party repo including x64 and arm64 builds of ffmpeg: https://github.com/Aftnet/FFmpegWin Maybe dolphin could use it or something like it? I was never a fan of committing binaries to externals... :/

Externals/discord-rpc/src/discord-rpc.vcxproj
```diff
@@ -53,6 +61,24 @@
     </Midl>
     <Lib />
   </ItemDefinitionGroup>
+  <ItemDefinitionGroup Condition="'$(Configuration)|$(Platform)'=='Debug|ARM64'">
```
shuffle2: It looks like the changes to this file can also be reduced by just removing the `$(Platform)` component of the existing `ItemDefinitionGroup` for x64.

Source/Core/VideoBackends/D3D12/D3D12.vcxproj
```diff
@@ -41,12 +49,26 @@
       <ForcedIncludeFiles />
     </ClCompile>
   </ItemDefinitionGroup>
+  <ItemDefinitionGroup Condition="'$(Configuration)|$(Platform)'=='Release|ARM64'">
```
shuffle2: These blocks (as well as the ones in D3DCommon) can also be merged with existing x64 blocks.

Kinda related...why are these projects disabling pch?

Externals/LZO/lzodefs.h
shuffle2: Agree...wow, latest release is 2 years old, and the one in Externals is 9 years old :)

Source/UnitTests/UnitTests.vcxproj
```diff
@@ -56,7 +65,9 @@
     <ClCompile Include="$(ExternalsDir)gtest\src\gtest_main.cc" />
     <!--Lump all of the tests (and supporting code) into one binary-->
     <ClCompile Include="*.cpp" />
-    <ClCompile Include="*\*.cpp" />
+    <ClCompile Include="*\*.cpp">
```
shuffle2: Not quite sure what's going on here/why some subset is excluded. 
In any case, does gtest support running cross-compiled builds in some way? This project may be mostly useless (at least for CI) on arm64 if not.

Externals/discord-rpc/thirdparty/include/rapidjson/rapidjson.h
```diff
@@ -236,7 +236,7 @@
 #    define RAPIDJSON_ENDIAN RAPIDJSON_BIGENDIAN
 #  elif defined(__i386__) || defined(__alpha__) || defined(__ia64) || defined(__ia64__) || defined(_M_IX86) || defined(_M_IA64) || defined(_M_ALPHA) || defined(__amd64) || defined(__amd64__) || defined(_M_AMD64) || defined(__x86_64) || defined(__x86_64__) || defined(_M_X64) || defined(__bfin__)
 #    define RAPIDJSON_ENDIAN RAPIDJSON_LITTLEENDIAN
-#  elif defined(_MSC_VER) && defined(_M_ARM)
+#  elif defined(_MSC_VER) && (defined(_M_ARM) || defined(_M_ARM64))
```
shuffle2: This is also fixed 2 years ago in rapidjson mainline: https://github.com/Tencent/rapidjson/commit/f91405801f88533c609d95f2fcc2d88811544d35

Maybe just update the entire thing to a recent release?

Source/Core/Common/Arm64Emitter.cpp
```diff
@@ -2172,6 +2174,8 @@ void ARM64XEmitter::ABI_PopRegisters(BitSet32 registers, BitSet32 ignore_mask)
   ARM64Reg second;
```
shuffle2: consider init at decl instead. (or...ctor)

Source/UnitTests/UnitTests.vcxproj
shuffle2: p.s. ffmpeg was historically OK'd to be committed as binaries as it required the WDK to build, which doesn't make sense to require all dolphin devs to deal with when they probably don't care at all about it. However, judging from the build script in that repo, the WDK may not be required anymore. It's probably worth revisiting if we can just build ffmpeg from source and/or make it optional on windows.

shuffle2:
lzo and rapidjson are updated to versions supporting arm64 in #8524

Source/UnitTests/UnitTests.vcxproj
shuffle2: @RisingFog is the contributor of ffmpeg builds for dolphin, maybe they have some input here.
They have some franken-fork of ffmpeg here: https://github.com/RisingFog/dolphin-ffmpeg (mainline has more recent updates: https://github.com/hrydgard/ppsspp-ffmpeg ), however since the current version in Externals (from https://github.com/dolphin-emu/dolphin/pull/4989 ) is higher than the one in RisingFog's repo, it's not clear how related they are.

@RisingFog it would be nice to know how whatever's currently in dolphin was generated.

shuffle2: d'oh, missed that there's a dolphin branch there: https://github.com/RisingFog/dolphin-ffmpeg/commits/dolphin

there's also some compile log in the object files, like:
```
C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\BIN\amd64\cl.exe
-IC:\msys64\home\silen\dolphin-ffmpeg-dolphin -IC:\msys64\home\silen\dolphin-ffmpeg-dolphin\ -D_ISOC99_SOURCE -D_FILE_OFFSET_BITS=64 -D_LARGEFILE_SOURCE -D_WIN32_WINNT=0x0502 -DHAVE_AV_CONFIG_H -nologo -D_USE_MATH_DEFINES -D_CRT_SECURE_NO_WARNINGS
 -D_CRT_NONSTDC_NO_WARNINGS -MT -IC:\msys64\home\silen\dolphin-ffmpeg-dolphin\WindowsInclude -GS- -Z7 -W4 -wd4244 -wd4127 -wd4018 -wd4389 -wd4146 -wd4057 -wd4204 -wd4706 -wd4305 -wd4152 -wd4324 -we4013 -wd4100 -wd4214 -wd4307 -wd4273
 -wd4554 -wd4701 -O2 -SSAOptimizer- -c -I"C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\INCLUDE" -I"C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\ATLMFC\INCLUDE" -I"C:\Program Files (x86)\Windows Kits\10\include\10.0.10586.0\ucrt"
 -I"C:\Program Files (x86)\Windows Kits\NETFXSDK\4.6.1\include\um" -I"C:\Program Files (x86)\Windows Kits\10\include\10.0.10586.0\shared" -I"C:\Program Files (x86)\Windows Kits\10\include\10.0.10586.0\um" -I"C:\Program Files (x86)\Windows
 Kits\10\include\10.0.10586.0\winrt" -TC -X
```
however it still doesn't match what's in the repo (https://github.com/RisingFog/dolphin-ffmpeg/commit/6e0cf29eb02a1ef98c89d9432ca16dbf595011fa#diff-98ad90cfa64c27ce76009818fc95c8a5)

there is however this, which is maybe enough to repro:
```
--extra-cflags='-MT -IWindowsInclude -GS-' --toolchain=msvc --prefix=./Windows/x86_64 --arch=x86_64 --disable-avdevice --disable-programs --disable-avfilter --disable-postproc --disable-doc --disable-pthreads --enable-w32threads --disable-network --disable-everything --disable-encoders --disable-muxers --disable-hwaccels --disable-parsers --disable-protocols --enable-static --disable-shared --enable-muxer=avi --enable-encoder=ffv1 --enable-encoder=mpeg4 --enable-protocol=file
```

Source/Core/Common/ArmCPUDetect.cpp
```diff
@@ -60,6 +67,20 @@ void CPUInfo::Detect()
   Mode64bit = true;
   vendor = CPUVendor::ARM;
 
+#ifdef _WIN32
+  // Windows does not provide any mechanism for querying the system registers on ARMv8, unlike Linux
+  // which traps the register reads and emulates them in the kernel. There are environment variables
+  // containing some of the CPU-specific values, which we could use for a lookup table in the
+  // future. For now, assume all features are present as all known devices which are Windows-on-ARM
+  // compatible also support these extensions.
+  num_cores = 4;
```
shuffle2: My understanding is that `_SC_NPROCESSORS_CONF` returns number of currently available hardware threads, so you could just use `std::thread::hardware_concurrency` here?

Source/UnitTests/UnitTests.vcxproj
shuffle2: was able to build ffmpeg under wsl with a bit of a hack:

```diff
diff --git a/configure b/configure                                                                                                                                                                                                             index 34c2adb4a4..e49af5ee65 100755                                                                                                                                                                                                            --- a/configure                                                                                                                                                                                                                                +++ b/configure                                                                                                                                                                                                                                @@ -4607,7 +4607,7 @@ probe_cc(){                                                                                                                                                                                                                   elif $_cc -nologo- 2>&1 | grep -q Microsoft || { $_cc -v 2>&1 | grep -q clang && $_cc -? > /dev/null 2>&1; }; then                                                                                                                                 _type=msvc                                                                                                                                                                                                                                     _ident=$($_cc 2>&1 | head -n1 | tr -d '\r')                                                                                                                                                                                           -        _DEPCMD='$(DEP$(1)) $(DEP$(1)FLAGS) $($(1)DEP_FLAGS) $< 2>&1 | awk '\''/including/ { sub(/^.*file: */, ""); gsub(/\\/, "/"); if (!match($$0, / /)) print "$@:", $$0 }'\'' > $(@:.o=.d)'                                               +        _DEPCMD='$(DEP$(1)) $(DEP$(1)FLAGS) $($(1)DEP_FLAGS) $< 2>&1 | awk '\''/including/ { sub(/^.*file: */, ""); gsub(/\\/, "/"); gsub(/C:/, "/mnt/c/"); if (!match($$0, / /)) print "$@:", $$0 }'\'' > $(@:.o=.d)'                                 _DEPFLAGS='$(CPPFLAGS) $(CFLAGS) -showIncludes -Zs'                                                                                                                                                                                            _cflags_speed="-O2"                                                                                                                                                                                                                            _cflags_size="-O1"
```
hopefully there's a more proper way

shuffle2: I've built ffmpeg for x64 and arm64 and made a repo designed to function like the existing dolphin Qt submodule: https://github.com/shuffle2/ext-win-ffmpeg
The build source is here: https://github.com/shuffle2/FFmpeg/commits/dolphin
I would expect dolphin project to fork both these repos and effectively take them over. Hopefully it simplifies dealing with ffmpeg in the future.

I tested that dolphin x64 + arm64 builds against these build outputs.

shuffle2: btw, the configure outputs:
```
install prefix            ./ext-win-ffmpeg/ARM64
source path               .
C compiler                cl.exe
C library                 msvcrt
host C compiler           gcc
host C library            glibc
ARCH                      aarch64 (generic)
big-endian                no
runtime cpu detection     yes
NEON enabled              yes
VFP enabled               yes
debug symbols             yes
strip symbols             no
optimize for size         no
optimizations             yes
static                    yes
shared                    no
postprocessing support    no
network support           no
threading support         w32threads
safe bitstream reader     yes
texi2html enabled         no
perl enabled              yes
pod2man enabled           yes
makeinfo enabled          no
makeinfo supports HTML    no

External libraries:
schannel

External libraries providing hardware acceleration:
d3d11va                 dxva2

Libraries:
avcodec                 avformat                avutil                  swresample              swscale

Programs:

Enabled decoders:

Enabled encoders:
ffv1                    h263                    mpeg4

Enabled hwaccels:

Enabled parsers:

Enabled demuxers:

Enabled muxers:
avi

Enabled protocols:
file

Enabled filters:

Enabled bsfs:
null

Enabled indevs:

Enabled outdevs:

License: LGPL version 2.1 or later
```
```
install prefix            ./ext-win-ffmpeg/x64
source path               .
C compiler                cl.exe
C library                 msvcrt
ARCH                      x86 (generic)
big-endian                no
runtime cpu detection     yes
standalone assembly       yes
x86 assembler             yasm
MMX enabled               yes
MMXEXT enabled            yes
3DNow! enabled            yes
3DNow! extended enabled   yes
SSE enabled               yes
SSSE3 enabled             yes
AESNI enabled             yes
AVX enabled               yes
AVX2 enabled              yes
AVX-512 enabled           yes
XOP enabled               yes
FMA3 enabled              yes
FMA4 enabled              yes
i686 features enabled     yes
CMOV is fast              yes
EBX available             no
EBP available             no
debug symbols             yes
strip symbols             no
optimize for size         no
optimizations             yes
static                    yes
shared                    no
postprocessing support    no
network support           no
threading support         w32threads
safe bitstream reader     yes
texi2html enabled         no
perl enabled              yes
pod2man enabled           yes
makeinfo enabled          no
makeinfo supports HTML    no

External libraries:
schannel

External libraries providing hardware acceleration:
d3d11va                 dxva2

Libraries:
avcodec                 avformat                avutil                  swresample              swscale

Programs:

Enabled decoders:

Enabled encoders:
ffv1                    h263                    mpeg4

Enabled hwaccels:

Enabled parsers:

Enabled demuxers:

Enabled muxers:
avi

Enabled protocols:
file

Enabled filters:

Enabled bsfs:
null

Enabled indevs:

Enabled outdevs:

License: LGPL version 2.1 or later
```

shuffle2:
Still haven't heard anything about Qt's plans to host arm64 builds.

In any case, I think dolphin needs to decide how to shove it into a submodule: should it be in the existing Qt one (doubling the size) or separate (adding complexity)?

I still think it makes sense to drop the NoGUI addition.

shuffle2:
freenode/#qt
```
   shuffle2 | are there plans(with ETA/release target) for Qt to host 'official' binaries for the windows/arm64 platform?
   shuffle2 | and, is there a better place to ask such a question?
     V-ille | We are puzzled as to whether such platforms are relevant.
     V-ille | And to whom, and how many people that is.
```
so, dolphin would have to provide it's own Qt binaries for the foreseeable future.

Source/UnitTests/UnitTests.vcxproj
shuffle2: The only problematic file is x64EmitterTest. But for now this part of the PR is fine.

Source/UnitTests/UnitTests.vcxproj
RisingFog: @shuffle2 I was using msys to cross-compile the libraries, but this was some time ago.

shuffle2: @RisingFog thanks :) was mainly looking for configure options, which was helpfully embedded as a string into the object files, so nothing more to do here.

Externals/discord-rpc/src/discord-rpc.vcxproj
delroth: +1, doesn't seem like any of this is ARM64 specific.

Source/Core/VideoBackends/D3D12/D3D12.vcxproj
stenzek: No idea. Seems it's something that VS did at some point. I removed the whole conditional and it seems to work fine...

superheri:
This PR seems to create problems for updating for devices that already have the Windows version installed (on my case Intel processor)
image

My PC specs:
Win 10 1909
Intel I7-6700HQ
Nvidia GTX 980M

RinMaru:
same issue here

delroth:
This almost certainly has nothing to do with this PR. Can you open a bug and attach the updater log?

ShinyCookie:
This PR causes the same updater issue for me. In addition, updating manually and trying to launch causes this:
image

stenzek:
That sounds like you need to update your visual C++ runtime. https://aka.ms/vs/16/release/vc_redist.x64.exe

Still strange that this PR changed anything. None of the x64 parameters should've been affected.

superheri:
I'm sure this is this PR since I was able to update fine since this particuliar PR. When I see the changelog, only this one appears so it needs to be it.

delroth:
> This almost certainly has nothing to do with this PR. Can you open a bug and attach the updater log?

delroth:
Update issue was as expected unrelated, and fixed in https://github.com/dolphin-emu/sadm/commit/7e36c77e7fd692f68c00f6caadd57fd8ea3ba786

PR #8498, opened 2019-11-27
Update VS projects/solutions to VS2019

stenzek:
What the title says. The buildbots were updated a while back, so I don't see much reason to stay on the old toolset.

I'm assuming the plan is still to migrate to cmake, but at least we'll discover any issues with the new compiler this way, changing one thing a time?

### DIFF
--- a/Source/Core/DolphinQt/DolphinQt.vcxproj
+++ b/Source/Core/DolphinQt/DolphinQt.vcxproj
@@ -12,13 +12,13 @@
   </ItemGroup>
   <PropertyGroup Label="Globals">
     <ProjectGuid>{FA3FA62B-6F58-4B86-9453-4D149940A066}</ProjectGuid>
-    <WindowsTargetPlatformVersion>10.0.17134.0</WindowsTargetPlatformVersion>
+    <WindowsTargetPlatformVersion>10.0</WindowsTargetPlatformVersion>
     <ProjectName>Dolphin</ProjectName>
   </PropertyGroup>
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.Default.props" />
   <PropertyGroup Label="Configuration">
     <ConfigurationType>Application</ConfigurationType>
-    <PlatformToolset>v141</PlatformToolset>
+    <PlatformToolset>v142</PlatformToolset>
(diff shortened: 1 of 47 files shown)


Source/VSProps/Base.props
```diff
@@ -113,8 +113,9 @@
       4121  'symbol' : alignment of a member was sensitive to packing
       4324  Padding was added at the end of a structure because you specified a __declspec(align) value.
       4714  function 'function' marked as __forceinline not inlined
+      5054  operator '+': deprecated between enumerations of different types
```
lioncash: Should this warning *really* be disabled by default rather than being resolved? This is likely going to bite us down the road when it actually gets removed from the language.

stenzek: From what I remember, it was mainly in the Qt headers. I suppose we could just disable it in the `DolphinQt` project instead?

lioncash: Yeah, that sounds fine. At least that way, Core and other projects can still let the compiler complain loudly when it happens.

CookiePLMonster: There have been PR's pre-emptively fixing this in main code, so yes I also think disabling it only in Qt makes sense. Moreover, I think it's time we purge some suppressed warnings, I once started it but don't think I went far.

Source/Core/DolphinQt/DolphinQt.vcxproj
```diff
@@ -38,6 +38,12 @@
   </ImportGroup>
   <PropertyGroup Label="UserMacros">
   </PropertyGroup>
+  <ItemDefinitionGroup>
+    <ClCompile>
+      <!-- 5054  operator '+': deprecated between enumerations of different types -->
```
BhaaLseN: Should we leave a note here _why_ we do this rather than fixing it?

stenzek:
Unless anyone has any objections, I'll go ahead and merge the update. Will wait a couple more days until doing so.

xcstern-12:
> Unless anyone has any objections, I'll go ahead and merge the update. Will wait a couple more days until doing so.

so i cannot install VS2019 is it possible to use the older one on my end?

JosJuice:
No, you need VS2019 now.

PR #8499, opened 2019-11-27
VolumeVerifier: Detect broken Super Paper Mario

JosJuice:
https://bugs.dolphin-emu.org/issues/11900

### DIFF
--- a/Source/Core/DiscIO/VolumeVerifier.cpp
+++ b/Source/Core/DiscIO/VolumeVerifier.cpp
@@ -8,6 +8,7 @@
 #include <cinttypes>
 #include <future>
 #include <limits>
+#include <memory>
 #include <mutex>
 #include <optional>
 #include <string>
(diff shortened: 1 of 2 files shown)


Source/Core/DiscIO/VolumeVerifier.cpp
```diff
@@ -972,6 +973,36 @@ void VolumeVerifier::CheckMisc()
...
     }
+
+    if (StringBeginsWith(game_id_unencrypted, "R8P"))
+      CheckSuperPaperMario();
+  }
+}
+
+void VolumeVerifier::CheckSuperPaperMario()
+{
+  // https://bugs.dolphin-emu.org/issues/11900
```
BhaaLseN: I know (or at least hope!) our tracker is here to stay, but would you mind including a line or two describing the problem for the case where someone cannot access the tracker (or is simply too lazy to do so?)

JosJuice: I don't think such a comment would add anything that isn't already clear from the user-facing message and maybe the line of code that specifies which file we're checking. (Unless the fact that this happens with disc images created by currently existing versions of NKit is important to note in the source code?)

BhaaLseN: I mainly meant the fact that this game (and perhaps others?) will potentially read more data than specified by the size parameter, which _may_ result in issues if they're not zero. The rest is clear from the code and the message.

PR #8500, opened 2019-11-27
DolphinQt: Remove "Experimental" from "Export Wii Save"

JosJuice:
It's pretty well tested at this point.


Helios747:
Still mad that it made me redo a bunch of Trauma Center due to a bug (now fixed)

PR #8501, opened 2019-11-28
GCMemcard: Remove byteswapping functions and macros

lioncash:
ByteSwap can just be replaced with std::swap, and rather than use preprocessor macros for the swap functions, we can just use them directly.

### DIFF
--- a/Source/Core/Core/HW/EXI/EXI_DeviceMemoryCard.cpp
+++ b/Source/Core/Core/HW/EXI/EXI_DeviceMemoryCard.cpp
@@ -166,7 +166,9 @@ void CEXIMemoryCard::SetupGciFolder(u16 sizeMb)
   u32 CurrentGameId = 0;
   if (game_id.length() >= 4 && game_id != "00000000" &&
       SConfig::GetInstance().GetTitleID() != Titles::SYSTEM_MENU)
-    CurrentGameId = BE32((u8*)game_id.c_str());
+  {
+    CurrentGameId = Common::swap32(reinterpret_cast<const u8*>(game_id.c_str()));
+  }
 
   const bool shift_jis = region == DiscIO::Region::NTSC_J;
 
--- a/Source/Core/Core/HW/GCMemcard/GCMemcard.cpp
+++ b/Source/Core/Core/HW/GCMemcard/GCMemcard.cpp
@@ -8,6 +8,7 @@
 #include <cassert>
 #include <cinttypes>
 #include <cstring>
+#include <utility>
 #include <vector>
 
 #include "Common/BitUtils.h"
@@ -19,13 +20,6 @@
 #include "Common/MsgHandler.h"
 #include "Common/StringUtil.h"
 #include "Common/Swap.h"
-
-static void ByteSwap(u8* valueA, u8* valueB)
-{
-  u8 tmp = *valueA;
-  *valueA = *valueB;
-  *valueB = tmp;
-}
 
 static constexpr std::optional<u64> BytesToMegabits(u64 bytes)
 {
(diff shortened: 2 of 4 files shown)


Source/Core/Core/HW/EXI/EXI_DeviceMemoryCard.cpp
```diff
@@ -166,7 +166,9 @@ void CEXIMemoryCard::SetupGciFolder(u16 sizeMb)
   u32 CurrentGameId = 0;
   if (game_id.length() >= 4 && game_id != "00000000" &&
       SConfig::GetInstance().GetTitleID() != Titles::SYSTEM_MENU)
-    CurrentGameId = BE32((u8*)game_id.c_str());
+  {
+    CurrentGameId = Common::swap32(reinterpret_cast<const u8*>(game_id.c_str()));
```
CookiePLMonster: How does this work? Swapping endianness of a pointer?

Source/Core/Core/HW/GCMemcard/GCMemcard.cpp
```diff
@@ -1148,38 +1141,38 @@ void GCMemcard::Gcs_SavConvert(DEntry& tempDEntry, int saveType, u64 length)
...
-    ByteSwap(&tempDEntry.m_unused_1, &tempDEntry.m_banner_and_icon_flags);
+    std::swap(tempDEntry.m_unused_1, tempDEntry.m_banner_and_icon_flags);
 
     std::array<u8, 4> tmp;
-    memcpy(tmp.data(), &tempDEntry.m_image_offset, 4);
-    ByteSwap(&tmp[0], &tmp[1]);
-    ByteSwap(&tmp[2], &tmp[3]);
-    memcpy(&tempDEntry.m_image_offset, tmp.data(), 4);
+    std::memcpy(tmp.data(), &tempDEntry.m_image_offset, 4);
+    std::swap(tmp[0], tmp[1]);
```
CookiePLMonster: This isn't a big endian to little endian swap for sure, so what is it?

CookiePLMonster:
Less macros = more good, so always happy to see such changes!

Source/Core/Core/HW/EXI/EXI_DeviceMemoryCard.cpp
lioncash: The swap functions have [overloads](https://github.com/dolphin-emu/dolphin/blob/master/Source/Core/Common/Swap.h#L114) for operating on the data pointed to by  u8 pointers.

Source/Core/Core/HW/GCMemcard/GCMemcard.cpp
AdmiralCurtiss: I'm not 100% sure as this isn't spelled out, but I suspect this happens because some thirdparty Memory Card dumping hardware (Datel's MAX Drive, I think?) reads these fields in 16 bits words of the wrong endianness. But either way, that's just how the SAV format stores its data.

Source/Core/Core/HW/EXI/EXI_DeviceMemoryCard.cpp
CookiePLMonster: Oh, that is a bit counter-intuitive - so this in fact extracts first 4 characters of a string and byteswaps them, right?

Source/Core/Core/HW/GCMemcard/GCMemcard.cpp
CookiePLMonster: Fair enough, just wanted to point it out in an unlikely chance of a typo/misworded intention.

Source/Core/Core/HW/EXI/EXI_DeviceMemoryCard.cpp
lioncash: Yes.

CookiePLMonster: Cool - that said, I would personally prefer to see those not overloaded but rather as a family of `swapPtrXX` functions. I feel like right now it is needlessly easy to confuse `swap32(T)` and `swap32(T*)`.

Source/Core/Core/HW/GCMemcard/GCMemcard.cpp
```diff
@@ -1433,10 +1426,10 @@ s32 GCMemcard::FZEROGX_MakeSaveGameValid(const Header& cardheader, const DEntry&
   const auto [serial1, serial2] = cardheader.CalculateSerial();
 
   // set new serial numbers
-  *(u16*)&FileBuffer[1].m_block[0x0066] = BE16(BE32(serial1) >> 16);
-  *(u16*)&FileBuffer[3].m_block[0x1580] = BE16(BE32(serial2) >> 16);
-  *(u16*)&FileBuffer[1].m_block[0x0060] = BE16(BE32(serial1) & 0xFFFF);
-  *(u16*)&FileBuffer[1].m_block[0x0200] = BE16(BE32(serial2) & 0xFFFF);
```
AdmiralCurtiss: Type punning on assignment like that is undefined behavior, isn't it? Even if it isn't this looks pretty sketchy, maybe memcpy those bytes into the m_blocks instead?

lioncash: I'll save resolving this for a follow-up PR, given there's several other locations within the source file where this sort of type-punning is performed, and I'd rather resolve them all at once.

Source/Core/Core/HW/GCMemcard/GCMemcard.cpp
```diff
@@ -821,9 +815,8 @@ GCMemcardGetSaveDataRetVal GCMemcard::GetSaveData(u8 index, std::vector<GCMBlock
   if (!m_valid)
     return GCMemcardGetSaveDataRetVal::NOMEMCARD;
 
-  u16 block = DEntry_FirstBlock(index);
-  u16 BlockCount = DEntry_BlockCount(index);
-  // u16 memcardSize = BE16(hdr.m_size_mb) * MBIT_TO_BLOCKS;
+  const u16 block = DEntry_FirstBlock(index);
+  const u16 BlockCount = DEntry_BlockCount(index);
```
BhaaLseN: Wanna update this ones' naming convention while we're at it?

Source/Core/Core/HW/GCMemcard/GCMemcard.cpp
BhaaLseN: Is this worth updating the comment above to include a (speculative?) reason _why_ this is? It does say "swap byte pairs", but gives no reason for it.

Source/Core/Core/HW/GCMemcard/GCMemcard.cpp
AdmiralCurtiss: Mhm, fair enough. Looks fine to me then.

PR #8224, opened 2019-07-01
Write the entirety of the Wiimote EEPROM, in a per-Wiimote file

Pokechu22:
Previously, only Mii data was written.  Additionally, the file containing mii data was shared for all Wiimotes, which made it a lot less useful.

Additionally, the file was read/written on each Wiimote read, even though the whole EEPROM was kept in memory.  This was bad for performance and not particularly necessary (it did enforce that the data was properly shared between all Wiimotes, but that's not something I want).

-----

Some notes about this:

* If there is no `Wiimote#.bin` file, the previous `mii.bin` file will be read (if it exists) and used to populate the mii block.  However, since the `Wiimote#.bin` files are not written until a write to EEPROM happens, this will continue happening on each startup.  Also, the `mii.bin` file isn't ever deleted, which is probably fine but might get weird if all wiimotes have bin files with them since the mii file isn't used at that point.
* I think it would be neat to have files not tied directly to Wiimote numbers, since you might want to save more data per game (or data from different games or whatever).  But I've got no idea how to do that from a UI perspective.
* It would be neat to import (and _maybe_ export, though that may be dangerous) data to physical Wiimotes, but that does depend on having further management of the files than just tying it to the Wiimote number.
* This includes writing EEPROM for the balance board, which is ~~[currently unknown](https://wiibrew.org/wiki/Wii_Balance_Board#EEPROM_Memory).  Technically Dolphin allowed reads and writes to it before, but it would only interact with the same `mii.bin` file.  Now, a write to the EEPROM for the wii balance board will create a `Wiimote5.bin` file (I think), which might be incorrect.~~  correct based on hardware testing, and will go into a `BalanceBoard.bin` file.  That said, this only happens after a write, which presumably no actual game does.
* Data might be written too infrequently now.  It's written when `Reset()` is called while the eeprom is dirty, which should include disconnects; however, that only happens when the game is closed eventually.  If there's a crash or something in the middle, the file remains unwritten.

PR #8504, opened 2019-11-30
Core: Only reset wiimotes in Wii mode

stenzek:
This was causing the FifoCI runners to crash, as the Wiimotes were not initialized in the first place.

The whole `g_wiimote_sources` setup seems kinda janky in the first place...

### DIFF
--- a/Source/Core/Core/Core.cpp
+++ b/Source/Core/Core/Core.cpp
@@ -290,7 +290,9 @@ void Stop()  // - Hammertime!
     g_video_backend->Video_ExitLoop();
   }
 
-  Wiimote::ResetAllWiimotes();
+  if (_CoreParameter.bWii)
+    Wiimote::ResetAllWiimotes();
+
   ResetRumble();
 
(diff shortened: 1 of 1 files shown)


Pokechu22:
This code was added in #8224.  I don't remember if I actually tested it in raw Gamecube mode, but I did test it in the case of loading a Gamecube game from the Wii Menu.  To make sure, is `_CoreParameter.bWii` still set to true when stopping in that case?

stenzek:
`_CoreParameter.bWii` was false for fifologs in GC mode, I didn't check games in GC mode but I imagine it should also be the case.

Pokechu22:
Ah, looks like the reset was superfluous for the transition from Wii to GC, as they're reset in `MIOS::ReinitHardware` (which actually also sets `bWii` to false):

https://github.com/dolphin-emu/dolphin/blob/4d4a095c76501799767f70120fe86238f3fe8fb4/Source/Core/Core/IOS/MIOS.cpp#L30-L45

So this change shouldn't cause any issues.

PR #8506, opened 2019-12-03
PPCCache: Make arrays constexpr where applicable

lioncash:
We make use of std::array within the PPC cache code. While we're in there, we can make the initialization of two arrays constexpr and move them out of the save state data, shrinking the size of save states a little bits.

### DIFF
--- a/Source/Core/Core/PowerPC/PPCCache.cpp
+++ b/Source/Core/Core/PowerPC/PPCCache.cpp
@@ -4,7 +4,7 @@
 
 #include "Core/PowerPC/PPCCache.h"
 
-#include <cstring>
+#include <array>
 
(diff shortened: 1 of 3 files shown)


Source/Core/Core/PowerPC/PPCCache.cpp
```diff
@@ -15,74 +15,109 @@
...
   JitInterface::ClearSafe();
 }
 
 void InstructionCache::Init()
 {
-  memset(data, 0, sizeof(data));
-  memset(tags, 0, sizeof(tags));
-
+  data = {};
+  tags = {};
```
CookiePLMonster: You used `fill` in `InstructionCache::Reset`, why not here too?

PR #8507, opened 2019-12-03
Fix flickering in Metroid Prime: Trilogy.

SubsonicBeatz154:
Enabled SyncGPU, this fixes the flickering and black bar issues that have plagued the game for the longest of times. At a pretty substantial performance cost unfortunately... SyncGPU is the only way to fix those issues, the wiki is wrong and lists them as already having been fixed.

Set ImmediateXFBEnable and XFBToTextureEnable to true, this way if the user chooses to disable SyncGPU manually for performance reasons, they will get the black bar instead of the horrible flickering.

Also updated the game name in the GameINI to read as Metroid Prime: Trilogy instead of Metroid Prime Trilogy. Metroid Prime: Trilogy is the correct name according to the name in the game list.

None of these changes interfere with the effectiveness of the previous PR that fixed the loading screens.

### DIFF
--- a/Data/Sys/GameSettings/R3M.ini
+++ b/Data/Sys/GameSettings/R3M.ini
@@ -1,7 +1,8 @@
-# R3ME01, R3MP01 - Metroid Prime Trilogy
+# R3ME01, R3MP01 - Metroid Prime: Trilogy
 
 [Core]
 # Values set here will override the main Dolphin settings.
+SyncGPU = True
 
 [OnLoad]
 # Add memory patches to be loaded once on boot here.
@@ -14,6 +15,6 @@
 
 [Video_Hacks]
 EFBToTextureEnable = False
-ImmediateXFBEnable = False
-XFBToTextureEnable = False
+ImmediateXFBEnable = True
+XFBToTextureEnable = True
 DeferEFBCopies = False
--- a/Data/Sys/GameSettings/R3M.ini
+++ b/Data/Sys/GameSettings/R3M.ini
@@ -2,7 +2,6 @@
 
 [Core]
 # Values set here will override the main Dolphin settings.
-SyncGPU = True
 
 [OnLoad]
 # Add memory patches to be loaded once on boot here.


JosJuice:
We generally do not set `ImmediateXFBEnable` and `XFBToTextureEnable` to `True`, since they are hacks. If the game doesn't need them to be `False`, the lines should be omitted from the INI, so that we don't force the user to use settings that are less accurate than what they selected themselves. If someone wants to manually disable SyncGPU for whatever reason, then I don't mind if they also have to manually enable the XFB hacks.

Just for the reference, the name in the game list aren't always 100% correct. But adding the colon sounds fine to me.

SubsonicBeatz154:
ImmediateXFBEnable and XFBToTextureEnable are the only fix for the flickering when SyncGPU is disabled, that's why i enabled them. It instead causes the old black bar to appear, which most people I've seen will agree on being better than the flickering.

JosJuice:
Yes, but you are now enabling SyncGPU in the game INI. The game INI settings don't need to cater for people who choose to override the game INI settings.

SubsonicBeatz154:
Fair enough, i could remove SyncGPU. But ImmediateXFBEnable and XFBToTextureEnable at least should be kept enabled by default. The black bar vs the flickering is a night and day difference.

JosJuice:
Your PR is very unlikely to get merged if you force hacks to be enabled.

stenzek:
Does the game behave correctly with dual core disabled and EFB2RAM enabled (store to texture only disabled)? If it's needing SyncGPU in the first place, this suggests that the issue is dual core related...

SubsonicBeatz154:
From what I've seen, hacks being enabled in GameINI's are usually for compatibility. The flickering is MUCH worse in MP1 and MP2 than it is in MP3. Making those games almost unplayable.

SubsonicBeatz154:
Yes, the game does behave correctly with Dual Core disabled. @stenzek

Techjar:
This game is unstable on dual core, SyncGPU and any other settings only make the desync less likely to occur, if they have any effect at all. The only way to guarantee stability is to use single core mode.

SubsonicBeatz154:
From extensive play-testing on my end, SyncGPU completely fixes the desyncing issues.

Techjar:
Only by chance. I did similarly extensive testing and still got desyncs on dual core mode, so it's very much system-dependent. Additionally, setting `XFBToTextureEnable = True` causes crazy flashing loading screen. I'm aware that also enabling Immediate XFB "fixes" this, but it's a hack and not correct behavior, so it should not be set by default in GameINI.

stenzek:
If the performance hit is only minor versus SyncGPU, it's probably better to just force single core. SyncGPU is still not deterministic, it only prevents the GPU thread from getting too far ahead/behind.

SubsonicBeatz154:
The GameCube version of Metroid Prime 2 has SyncGPU enabled in it's GameINI, so i really don't know how that would get through and not this.

Techjar:
I believe that's been there for a long time, before we got more strict about GameINI policies.

SubsonicBeatz154:
Even when ImmediateXFBEnable and XFBToTextureEnable are both disabled. The same flickering still occurs. When and ImmediateXFBEnable and XFBToTextureEnable are both enabled, you get the old black bar bug. I thought having the old black bar back when desync's occur vs the flickering is better. Cause at least the black bar doesn't hurt your eyes.

SubsonicBeatz154:
AFAIK, the flicking with the default GameINI config was a side-effect of the loading screen fix PR that got merged a while back. Which isn't even required anymore as the problem seems to be gone on the emulation side of things now rather than hacks.

JosJuice:
Are you sure the loading screen problem is gone? Please test it with Immediately Present XFB off and Store XFB Copies to Texture Only on.

SubsonicBeatz154:
Oh, i guess i was wrong. The problem still persists when using Immediately Present XFB off and Store XFB Copies to Texture Only on.

SubsonicBeatz154:
With that setup, you'll get THIS horrible mess instead of the usual flickering...

Techjar:
Yes, you need XFB to RAM to avoid that issue.

SubsonicBeatz154:
I thought XFB to RAM was removed?

Techjar:
It was consolidated into one option like EFB to RAM, so it's either tecture only or texture+RAM (aka Hybrid XFB).

SubsonicBeatz154:
When both Immediately Present XFB and Store XFB Copies to Texture Only are on, you'll get the classic black bar instead of... Whatever that nonsense was. I personally think seeing this vs the flickering from the Fix loading screens on Metroid Prime Trilogy PR's config is better as a default config than it.

Techjar:
As we said, Immediate XFB is a hack for highly specific use cases and it shouldn't be enabled here.

SubsonicBeatz154:
Well, i apologize for the inconvenience then. I guess i don't share the same opinions as you guys...

SubsonicBeatz154:
Well, at least i now know what NOT to do when it comes to default GameINI modifications.

SubsonicBeatz154:
Again, i apologize. I'm new to GitHub.

PR #8508, opened 2019-12-03
VideoCommon/TextureConverterShaderGen: Convert over to fmt

lioncash:
Begins the conversion of the shader generators over to using fmt formatting specifiers, starting with the easiest one.

This introduces the `WriteFmt` function as a transitional API until all of them are converted over, where the name will be changed back to `Write`

This also has a benefit over the older StringFromFormat-based API in that all formatted data is appended to the existing buffer rather than creating a completely separate string and then appending it to the internal string buffer.

### DIFF
--- a/Source/Core/VideoCommon/ShaderGenCommon.h
+++ b/Source/Core/VideoCommon/ShaderGenCommon.h
@@ -6,10 +6,13 @@
 
 #include <cstdarg>
 #include <cstring>
+#include <iterator>
 #include <map>
 #include <string>
 #include <type_traits>
 #include <vector>
+
+#include <fmt/format.h>
 
 #include "Common/CommonTypes.h"
 #include "Common/StringUtil.h"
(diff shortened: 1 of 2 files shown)


Source/Core/VideoCommon/ShaderGenCommon.h
```diff
@@ -112,6 +117,13 @@ class ShaderCode : public ShaderGeneratorInterface
     va_end(arglist);
   }
 
+  // Writes format strings using fmtlib format strings.
+  template <typename... Args>
+  void WriteFmt(std::string_view format, Args&&... args)
+  {
+    fmt::format_to(std::back_inserter(m_buffer), format, std::forward<Args>(args)...);
```
CookiePLMonster: That looks so elegant, `fmt` is great.

Source/Core/VideoCommon/ShaderGenCommon.h
```diff
@@ -101,6 +104,8 @@ class ShaderCode : public ShaderGeneratorInterface
 public:
   ShaderCode() { m_buffer.reserve(16384); }
   const std::string& GetBuffer() const { return m_buffer; }
+
+  // Deprecated: Writes format strings using traditional printf format strings.
```
CookiePLMonster: Can you mark this as `[[deprecated]]`? Or will it fail compilations due to "treat warnings as errors"?

lioncash: Compilation will fail.

PR #8509, opened 2019-12-05
VideoCommon/FramebufferShaderGen: Minor clean up

lioncash:
Cleans up the framebuffer shader generator code in various ways, mostly by collapsing unnecessary repeated string literal insertions into string streams; making for a nicer transition over to fmt.

While in the same area, we can also:

- Make use of `std::string_view` to enforce the use of valid strings.
- Add missing initial source file comments.
- Remove an unused Config struct.

This isn't intended to perform any behavioral changes.

### DIFF
--- a/Source/Core/VideoCommon/FramebufferShaderGen.cpp
+++ b/Source/Core/VideoCommon/FramebufferShaderGen.cpp
@@ -20,7 +20,7 @@ APIType GetAPIType()
   return g_ActiveConfig.backend_info.api_type;
 }
 
-void EmitUniformBufferDeclaration(std::stringstream& ss)
+void EmitUniformBufferDeclaration(std::ostringstream& ss)
 {
   if (GetAPIType() == APIType::D3D)
     ss << "cbuffer PSBlock : register(b0)\n";
@@ -28,7 +28,7 @@ void EmitUniformBufferDeclaration(std::stringstream& ss)
     ss << "UBO_BINDING(std140, 1) uniform PSBlock\n";
 }
 
-void EmitSamplerDeclarations(std::stringstream& ss, u32 start = 0, u32 end = 1,
+void EmitSamplerDeclarations(std::ostringstream& ss, u32 start = 0, u32 end = 1,
                              bool multisampled = false)
 {
   switch (GetAPIType())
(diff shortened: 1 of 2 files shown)


Source/Core/VideoCommon/FramebufferShaderGen.cpp
```diff
@@ -51,17 +60,17 @@ static void EmitSamplerDeclarations(std::stringstream& ss, u32 start = 0, u32 en
   }
 }
 
-static void EmitSampleTexture(std::stringstream& ss, u32 n, const char* coords)
+void EmitSampleTexture(std::stringstream& ss, u32 n, const char* coords)
```
Warepire: This function appears to use `coords` in the same way as `EmitTextureLoad` which was upgraded to `std::string_view`, but this one wasn't. Is there a specific reason for that?

Source/Core/VideoCommon/FramebufferShaderGen.cpp
```diff
@@ -225,50 +238,52 @@ std::string GenerateScreenQuadVertexShader()
...
     for (u32 i = 0; i < num_colors; i++)
       ss << "  float4 color" << i << " : COLOR" << i << ";\n";
-    ss << "  float4 position : SV_Position;\n";
-    ss << "};\n";
-    ss << "struct GS_OUTPUT\n";
-    ss << "{";
+    ss << "  float4 position : SV_Position;\n"
+          "};\n";
+
+    ss << "struct GS_OUTPUT\n"
```
CookiePLMonster: Why is this using `ss << ` instead of making it one long string like you do above?

lioncash: Because the string above it is related to a completely different struct declaration. For visual clarity I've kept them apart from one another.

CookiePLMonster: Sure, but keeping a blank line between those would still keep the separation clear, still without the need to have a separate `<<`. IMO

lioncash: There is no difference here. The use of `<<` itself isn't an issue, especially given it's going to be moved over to fmt in a follow-up change, and I'd still prefer to keep the struct declarations separate from one another in that context as well.

PR #8510, opened 2019-12-05
VideoCommon/TextureConversionShader: Migrate generation off sprintf

lioncash:
Encapsulates all of the file-scope state into its own class, removing some global mutable state, while also converting the generator itself over to using fmt where applicable. This is one of the older shaders that still had a `WRITE` macro defined to perform the appending to a global string buffer (which essentially forwarded directly to `sprintf`)

This way, we don't need to care about inserting canary characters into a buffer and hope we don't overwrite them. In many cases, we didn't even need to use `std::sprintf`.

While we're in the same area, we can make minor clean-up related changes to the interfaces, which have been separated into their own commits.

### DIFF
--- a/Source/Core/VideoCommon/TextureConversionShader.cpp
+++ b/Source/Core/VideoCommon/TextureConversionShader.cpp
@@ -2,32 +2,26 @@
 // Licensed under GPLv2+
 // Refer to the license.txt file included.
 
-#include <array>
-#include <cmath>
-#include <cstdio>
+#include "VideoCommon/TextureConversionShader.h"
+
+#include <iterator>
 #include <map>
 #include <sstream>
 
-#include "Common/CommonFuncs.h"
+#include <fmt/format.h>
+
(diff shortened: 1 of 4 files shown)


Source/Core/VideoCommon/TextureConversionShader.cpp
```diff
@@ -2,1397 +2,1421 @@
...
+          break;
+        }
+      }
+      else
+      {
+        // Handle D3D depth inversion.
+        if (!g_ActiveConfig.backend_info.bSupportsReversedDepthRange)
+          Append("1.0 - (");
+        else
+          Append("()");
```
BhaaLseN: This looks weird. Everything else only writes an opening parenthesis, but no closing ones.
Then again, I have no idea about shader code 🤷‍♂

lioncash:
Closing this in favor of #8576 as it addresses the same thing, but in a more straightforward manner.

PR #8512, opened 2019-12-05
VideoCommon/IndexGenerator: Eliminate static state

lioncash:
Gets rid of the global state within the index generator and makes it a proper class. This can just be a member that sits within the vertex manager base class. By deglobalizing the state of the index generator we also get rid of the wonky dual-initializing that was going on within the OpenGL backend. (e.g. Initialize would be called once during initialization for general backend initialization, and then the OpenGL backend would call it again after checking whether or not primitive restart is supported).

Since the renderer is always initialized before the vertex manager, we now only call `IndexGenerator`'s `Init()` once throughout the execution lifecycle.

### DIFF
--- a/Source/Core/VideoCommon/IndexGenerator.cpp
+++ b/Source/Core/VideoCommon/IndexGenerator.cpp
@@ -17,9 +17,6 @@ namespace
 {
 constexpr u16 s_primitive_restart = UINT16_MAX;
 
-using PrimitiveFunction = u16*(*)(u16*, u32, u32);
-std::array<PrimitiveFunction, 8> s_primitive_table;
-
 template <bool pr>
 u16* WriteTriangle(u16* index_ptr, u32 index1, u32 index2, u32 index3)
 {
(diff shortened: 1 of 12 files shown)


Source/Core/VideoCommon/IndexGenerator.cpp
```diff
@@ -182,77 +133,122 @@ u16* IndexGenerator::AddFan(u16* Iptr, u32 numVerts, u32 index)
...
+void IndexGenerator::AddIndices(int primitive, u32 num_vertices)
+{
+  m_index_buffer_current =
+      m_primitive_table[primitive](m_index_buffer_current, num_vertices, m_base_index);
+  m_base_index += num_vertices;
+}
+
+void IndexGenerator::AddExternalIndices(const u16* indices, u32 num_indices, u32 num_vertices)
+{
+  std::memcpy(m_index_buffer_current, indices, sizeof(u16) * num_indices);
```
CookiePLMonster:
```suggestion
  std::memcpy(m_index_buffer_current, indices, sizeof(indices[0]) * num_indices);
```
If you ever need u32 indices it'll make migration a bit easier.

lioncash: This is unrelated to the actual changes being made.

CookiePLMonster: I respectfully disagree - you've changed destination from `index_buffer_current` to `m_index_buffer_current` so you "touched" this line in a meaningful way in this PR already.

lioncash: Yes. I've changed the line because we follow a naming convention for member variables in classes. This does not mean I also need to resolve everything else related to that line within an unrelated PR. The purpose of this change is to eliminate static state in the class, which it does. It does not try to address anything else with the interface.

lioncash: Yes. I've changed the line because we follow a naming convention for member variables in classes. This does not mean I also need to resolve everything else related to that line within an unrelated PR. The purpose of this change is to eliminate static state in the class, which it does. It does not try to address anything else with the interface.

PR #8513, opened 2019-12-05
VideoCommon/BoundingBox: Make interface for querying bounding box data

lioncash:
Rather than exposing the bounding box members directly, we can instead provide an interface for code to use. This makes it nicer to transition off of global data, as the interface function names are already in place when the stateful class is introduced, meaning only `BoundingBox::` generally needs to be replaced.

With this, we now also have a place for both the disabling of bounding box itself, but also a single location for notifying the pixel shader manager when bounding box is disabled or enabled.


CookiePLMonster:
You are typecasting `u16` to `Coordinate` in user code just to have `Coordinate` typecast to `u32` again internally. I'd suggest either making `Coordinate` a private implementation detail of BoundingBox.cpp or introducing overloads so `Coordinate` overloads call `[G/S]etCoordinate(u16)`. It's an enum class so it's all fine.

PR #8514, opened 2019-12-06
Common/Network: Make StringToMacAddress use a string_view 

lioncash:
This function only ever reads the contents of the string in a non-owning manner, so we can change the parameter over to being a string view.

### DIFF
--- a/Source/Core/Common/Network.cpp
+++ b/Source/Core/Common/Network.cpp
@@ -43,7 +43,7 @@ std::string MacAddressToString(const MACAddress& mac)
                      mac[4], mac[5]);
 }
 
-std::optional<MACAddress> StringToMacAddress(const std::string& mac_string)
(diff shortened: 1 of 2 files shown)


Source/Core/Common/Network.cpp
```diff
@@ -43,7 +41,7 @@ std::string MacAddressToString(const MACAddress& mac)
                      mac[4], mac[5]);
 }
 
-std::optional<MACAddress> StringToMacAddress(const std::string& mac_string)
+std::optional<MACAddress> StringToMacAddress(std::string_view mac_string)
 {
   if (mac_string.empty())
     return {};
```
CookiePLMonster: You yourself recommended `std::nullopt` a while back, would be a good fit there. Since this PR only touches a single function, I see no reason not to go for it together with `string_view` change.

PR #8515, opened 2019-12-06
Android: update to Gradle 3.5.3

Simonx22:
Android Studio recommends updating to Gradle 3.5.3


farouk327:
Hi the app crash every game launch since weihuoya put his hand on the emulator

PR #8516, opened 2019-12-08
Fix scrubbing unencrypted Wii disc images (for real this time)

JosJuice:
Unlike what my past self apparently thought, `%` is in fact not the "round down" operator.

https://bugs.dolphin-emu.org/issues/11906


Helios747:
oops

PR #8517, opened 2019-12-09
Fix Vulkan and D3D12 video backends crashing on boot

stenzek:
Apparently we weren't calling the base initialize method.. this was probably my mistake, and just exposed by #8512.


JosJuice:
Fixes the issue for me.

farouk327:
Bahhh
This thing still not working specially when you used weihuoya code

PR #8518, opened 2019-12-09
Core: Reset Wiimotes/rumble/memwatcher on emu thread

stenzek:
This was causing a race which was crashing the FifoCI runners. The main thread called Stop() which in turn called ResetAllWiimotes() while the emu thread was still exiting, also shutting down the Wiimote class.

By shifting the reset to the emu thread, all cleanup operations happen on the same thread where they were initialized.

The whole startup/shutdown situation is not an ideal design anyway (e.g. the `join()` in `Start()`...). But at least this fixes the "random" crashes.

I'd make `Stop()` simply block until the emu thread shuts down, but if any part of the shutdown requires window messages to be pumped, as `Stop()` is called from the UI thread it'll deadlock. So probably requires some more thinking.

### DIFF
--- a/Source/Core/Core/Core.cpp
+++ b/Source/Core/Core/Core.cpp
@@ -289,12 +289,6 @@ void Stop()  // - Hammertime!
 
     g_video_backend->Video_ExitLoop();
   }
-
-  if (_CoreParameter.bWii)
-    Wiimote::ResetAllWiimotes();
-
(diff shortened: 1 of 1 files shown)


stenzek:
Related to #8504, regression from #8224. I think this could potentially happen in the Qt frontend as well, but not consistently as it's a race.

PR #8522, opened 2019-12-17
FrameDump: Set first_frame if movie frame number <= 1

stenzek:
This appears to cause ffmpeg to create invalid video files on output if the emulator only runs for a single frame, breaking FifoCI.

Someone with frame dumping knowledge should probably review this, but I suspect the regression started with the VI changes a little while back...

### DIFF
--- a/Source/Core/VideoCommon/FrameDump.cpp
+++ b/Source/Core/VideoCommon/FrameDump.cpp
@@ -443,7 +443,7 @@ FrameDump::Frame FrameDump::FetchState(u64 ticks)
 {
   Frame state;
   state.ticks = ticks;
-  state.first_frame = Movie::GetCurrentFrame() < 1;
+  state.first_frame = Movie::GetCurrentFrame() <= 1;
   state.ticks_per_second = SystemTimers::GetTicksPerSecond();
   state.savestate_index = s_savestate_index;
   return state;


stenzek:
It looks like the first frame was being skipped in all fifologs... :/

e.g. https://fifoci.dolphin-emu.org/compare/5104335-5103853/

But, the single-frame logs are now fixed: https://fifoci.dolphin-emu.org/compare/5104276-5103897/

JosJuice:
I think this could lead to problems at resolution changes and when starting recording in the middle of emulation. (More specifically, this could lead to an incorrect `delta` value for the first frame of the new AVI file.) Have you checked these scenarios? Maybe changing `Movie::GetCurrentFrame() < 1` to `Movie::GetCurrentFrame() <= 1` could be an alternative to removing the check entirely. (It is a bit ugly that we're using `Movie` code here, though...)

stenzek:
Yeah, that was the other solution I was thinking.

RisingFog:
I feel that the `Movie::GetCurrentFrame() <= 1` check would cause less issues compared to how it is now

vadosnaprimer:
Sorry for missing, this doesn't help at all. The main test case I mentioned when it broke is broken in a similar way: Resident Evil 4 has about 280 duplicate frames at the very beginning, they are all missing, leading to insane av desync.

PR #8524, opened 2019-12-19
Update externals

shuffle2:
for arm64 compat

### DIFF
--- a/Externals/LZO/COPYING
+++ b/Externals/LZO/COPYING
@@ -1,0 +1,339 @@
+                    GNU GENERAL PUBLIC LICENSE
+                       Version 2, June 1991
+
+ Copyright (C) 1989, 1991 Free Software Foundation, Inc.,
+ 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+ Everyone is permitted to copy and distribute verbatim copies
+ of this license document, but changing it is not allowed.
+
+                            Preamble
+
+  The licenses for most software are designed to take away your
+freedom to share and change it.  By contrast, the GNU General Public
+License is intended to guarantee your freedom to share and change free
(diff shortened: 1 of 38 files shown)


stenzek:
@dolphin-emu-bot rebuild

stenzek:
Seems this doesn't build on non-windows.

shuffle2:
hm i only saw the lint error and just fixed that, was there more?

stenzek:
Yeah, bunch of SFML errors.. which sound familiar, I vaguely recall them from a while back.

@dolphin-emu-bot rebuild

shuffle2:
it seems clang-format as executed by VS doesn't result in something the linter is happy with ...

stenzek:
> `../Source/Core/Core/NetPlayClient.cpp:1558:12: error: ambiguous overload for 'operator>>' (operand types are 'sf::Packet' and 'lzo_uint {aka long unsigned int}')`

Might have to be explicit about the type here?

shuffle2:
Tools/lint.sh uses an out of date clang-format. WTF

shuffle2:
> > `../Source/Core/Core/NetPlayClient.cpp:1558:12: error: ambiguous overload for 'operator>>' (operand types are 'sf::Packet' and 'lzo_uint {aka long unsigned int}')`
> 
> Might have to be explicit about the type here?

I'm not really sure how sf::Packet (a nontrivial class) can be ambiguous with respect to...a scalar integer.

stenzek:
I was more wondering if you had to use u32/u64 (or the SFML typedefs for those).

shuffle2:
I mean..might as well try, but note the arg type to lzo1x_decompress is `lzo_uint`

stenzek:
@dolphin-emu-bot rebuild

PR #8525, opened 2019-12-19
HW/DSPHLE: Fix struct aliasing undefined behavior in AX ucode

Techjar:
Fixes no audio when compiled on VS2019 issue in Old_AXWii games (Wii Sports, Mario Party 8, WarioWare: Smooth Moves, probably a few others). ~~There are more undefined behaviors in AXWii code, I will address those soon.~~

Fixes https://bugs.dolphin-emu.org/issues/11781

### DIFF
--- a/Source/Core/Common/BitUtils.h
+++ b/Source/Core/Common/BitUtils.h
@@ -4,6 +4,7 @@
 
 #pragma once
 
+#include <array>
 #include <climits>
 #include <cstddef>
 #include <cstring>
@@ -238,6 +239,53 @@ template <typename T, typename PtrType>
 inline auto BitCastPtr(PtrType* ptr) noexcept -> BitCastPtrType<T, PtrType>
 {
   return BitCastPtrType<T, PtrType>{ptr};
+}
+
+// Similar to BitCastPtr, but specifically for aliasing structs to arrays.
+template <typename ArrayType, typename T,
+          typename Container = std::array<ArrayType, sizeof(T) / sizeof(ArrayType)>>
+inline auto BitCastToArray(const T& obj) noexcept -> Container
+{
+  static_assert(sizeof(T) % sizeof(ArrayType) == 0,
+                "Size of array type must be a factor of size of source type.");
+  static_assert(std::is_trivially_copyable<T>(),
+                "BitCastToArray source type must be trivially copyable.");
+  static_assert(std::is_trivially_copyable<Container>(),
+                "BitCastToArray array type must be trivially copyable.");
+
+  Container result;
+  std::memcpy(result.data(), &obj, sizeof(T));
+  return result;
+}
+
+template <typename ArrayType, typename T,
+          typename Container = std::array<ArrayType, sizeof(T) / sizeof(ArrayType)>>
+inline void BitCastFromArray(const Container& array, T& obj) noexcept
+{
+  static_assert(sizeof(T) % sizeof(ArrayType) == 0,
+                "Size of array type must be a factor of size of destination type.");
+  static_assert(std::is_trivially_copyable<Container>(),
+                "BitCastFromArray array type must be trivially copyable.");
+  static_assert(std::is_trivially_copyable<T>(),
+                "BitCastFromArray destination type must be trivially copyable.");
+
+  std::memcpy(&obj, array.data(), sizeof(T));
+}
+
+template <typename ArrayType, typename T,
+          typename Container = std::array<ArrayType, sizeof(T) / sizeof(ArrayType)>>
+inline auto BitCastFromArray(const Container& array) noexcept -> T
+{
+  static_assert(sizeof(T) % sizeof(ArrayType) == 0,
+                "Size of array type must be a factor of size of destination type.");
+  static_assert(std::is_trivially_copyable<Container>(),
+                "BitCastFromArray array type must be trivially copyable.");
+  static_assert(std::is_trivially_copyable<T>(),
+                "BitCastFromArray destination type must be trivially copyable.");
+
+  T obj;
+  std::memcpy(&obj, array.data(), sizeof(T));
+  return obj;
 }
 
 template <typename T>
--- a/Source/Core/Core/HW/DSPHLE/UCodes/AX.cpp
+++ b/Source/Core/Core/HW/DSPHLE/UCodes/AX.cpp
@@ -279,21 +279,6 @@ void AXUCode::HandleCommandList()
   }
 }
 
-void AXUCode::ApplyUpdatesForMs(int curr_ms, u16* pb, u16* num_updates, u16* updates)
-{
-  u32 start_idx = 0;
-  for (int i = 0; i < curr_ms; ++i)
-    start_idx += num_updates[i];
-
-  for (u32 i = start_idx; i < start_idx + num_updates[curr_ms]; ++i)
-  {
-    u16 update_off = Common::swap16(updates[2 * i]);
-    u16 update_val = Common::swap16(updates[2 * i + 1]);
-
-    pb[update_off] = update_val;
-  }
-}
-
 AXMixControl AXUCode::ConvertMixerControl(u32 mixer_control)
 {
   u32 ret = 0;
@@ -428,7 +413,7 @@ void AXUCode::ProcessPBList(u32 pb_addr)
 {
   // Samples per millisecond. In theory DSP sampling rate can be changed from
   // 32KHz to 48KHz, but AX always process at 32KHz.
-  const u32 spms = 32;
+  constexpr u32 spms = 32;
 
   AXPB pb;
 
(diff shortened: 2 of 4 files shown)


Techjar:
The buildbot build now has audio in Mario Party 8, so this seems to fix the issue.

shuffle2:
Could a warning or some compiler flag have caught this?

Source/Core/Common/BitUtils.h
```diff
@@ -4,6 +4,7 @@
 
 #pragma once
 
+#include <array>
```
Pokechu22: Is this include actually necessary?  I'm not completely familiar with the rules for templates but you're not directly using `std::array` in the new things in `BitUtils`.

Techjar:
As far as I know compilers don't have any warnings regarding strict aliasing.

Source/Core/Common/BitUtils.h
Techjar: The template still needs to know what an `std::array` is, and from what I can tell forward-declaring things in the `std` namespace is not allowed by the spec.

Pokechu22: ~~Won't it know at the point of use?  Doesn't all it need to know is that there is a data() function?~~

Oh, I missed the `typename Container = std::array<...>` bit, never mind.  I think that would require the import.

Source/Core/Common/BitUtils.h
```diff
@@ -240,6 +241,50 @@ inline auto BitCastPtr(PtrType* ptr) noexcept -> BitCastPtrType<T, PtrType>
   return BitCastPtrType<T, PtrType>{ptr};
 }
 
+// Similar to BitCastPtr, but specifically for aliasing structs to arrays.
+template <typename ArrayType, typename T,
+          typename Container =
+              std::array<ArrayType, (sizeof(T) + sizeof(ArrayType) - 1) / sizeof(ArrayType)>>
```
shuffle2: Is there really a case for supporting `sizeof(T) % sizeof(ArrayType) != 0` ? seems like it could cause surprising bugs in the future.

Techjar: Yeah I don't think there's any reasonable circumstances under which you'd use it like that. Should we just enforce alignment with `ArrayType` using an assert?

Source/Core/Common/BitUtils.h
```diff
@@ -240,6 +241,53 @@ inline auto BitCastPtr(PtrType* ptr) noexcept -> BitCastPtrType<T, PtrType>
...
+inline auto BitCastToArray(const T& obj) noexcept -> Container
+{
+  static_assert(sizeof(T) % sizeof(ArrayType) == 0,
+                "Size of array type must be a factor of size of source type.");
+  static_assert(std::is_trivially_copyable<T>(),
+                "BitCastToArray source type must be trivially copyable.");
+  static_assert(std::is_trivially_copyable<Container>(),
+                "BitCastToArray array type must be trivially copyable.");
+
+  Container result;
```
CookiePLMonster: This default constructs the `Container`, is this desirable? You may want to make this an aligned `std::byte` array.

Techjar: Uh, of course that's desirable. It's supposed to return an `std::array` of the type specified in the template.

CookiePLMonster: You memcpy data to it though, so essentially it default initialized for nothing.

Techjar: I don't believe you can avoid default initializing it.

Techjar: Good news is, default initialization is not zero initialization according to the docs, and leaves the newly allocated memory in some indeterminate state, which is ideal since we're writing over all of it anyways.

Source/Core/Core/HW/DSPHLE/UCodes/AXWii.h
```diff
@@ -37,18 +37,13 @@ class AXWiiUCode : public AXUCode
   int m_samples_aux3[6 * 3];
 
   // Are we implementing an old version of AXWii which still has updates?
-  bool m_old_axwii;
+  // bool m_old_axwii;
```
BhaaLseN: I think we want to get rid of this completely instead of just commenting it out, right?

Source/Core/Core/HW/DSPHLE/UCodes/AXWiiOld.cpp
```diff
@@ -0,0 +1,190 @@
...
+      break;
+    }
+  }
+}
+
+void AXWiiOldUCode::ProcessPBList(u32 pb_addr)
+{
+  // Samples per millisecond. In theory DSP sampling rate can be changed from
+  // 32KHz to 48KHz, but AX always process at 32KHz.
+  const u32 spms = 32;
```
BhaaLseN:
```suggestion
  constexpr u32 spms = 32;
```

Source/Core/Core/HW/DSPHLE/UCodes/AXWiiOld.cpp
```diff
@@ -0,0 +1,190 @@
...
+                          m_samples_auxB_left, m_samples_auxB_right, m_samples_auxB_surround,
+                          m_samples_auxC_left, m_samples_auxC_right, m_samples_auxC_surround,
+                          m_samples_wm0,       m_samples_aux0,       m_samples_wm1,
+                          m_samples_aux1,      m_samples_wm2,        m_samples_aux2,
+                          m_samples_wm3,       m_samples_aux3}};
+
+    ReadPB(pb_addr, pb, m_crc);
+
+    u32 updates_addr = HILO_TO_32(pb.updates.data);
+    u16* updates = (u16*)HLEMemory_Get_Pointer(updates_addr);
```
BhaaLseN:
```suggestion
    u16* updates = static_cast<u16*>(HLEMemory_Get_Pointer(updates_addr));
```
(or a different kind of C++ cast; but I'm fairly certain that this should be allowed)

Source/Core/Core/HW/DSPHLE/UCodes/AXWii.h
Techjar: Heh,  just noticed that as you commented.

Source/Core/Core/HW/DSPHLE/UCodes/AXWiiOld.cpp
Techjar: Suppose I'll change it in the other file too while we're at it.

BhaaLseN: It seems like a good fit, since we're touching the code anyways and it's not a huge refactoring.

Source/Core/Core/HW/DSPHLE/UCodes/AXWiiOld.cpp
Techjar: You're not wrong, but there's so much code in here that's using C-style casts that it seems a bit silly to change just one.

Techjar: Baby steps I suppose...

Source/Core/Core/HW/DSPHLE/UCodes/AXWiiOld.cpp
```diff
@@ -0,0 +1,190 @@
...
+  //     WARN_LOG(DSPHLE, "%04x", m_cmdlist[i]);
+  // WARN_LOG(DSPHLE, "-------------");
+
+  u32 curr_idx = 0;
+  bool end = false;
+  while (!end)
+  {
+    u16 cmd = m_cmdlist[curr_idx++];
+
+    switch (cmd)
```
Warepire: This whole switch statement looks quite duplicated, is there a way the common parts could be abstracted into a template?

Techjar: I don't think so. All the enum constants have an offset and there's some other subtle differences.

Warepire: (I know this discussion was ended on IRC yesterday, leaving it here just in case)

https://godbolt.org/z/p-3Dvf

Techjar:
I have reverted this PR to the original patch and will do the refactor in a separate PR.

delroth:
LGTM.

Source/Core/Common/BitUtils.h
```diff
@@ -240,6 +241,53 @@ inline auto BitCastPtr(PtrType* ptr) noexcept -> BitCastPtrType<T, PtrType>
...
+  std::memcpy(result.data(), &obj, sizeof(T));
+  return result;
+}
+
+template <typename ArrayType, typename T,
+          typename Container = std::array<ArrayType, sizeof(T) / sizeof(ArrayType)>>
+inline void BitCastFromArray(const Container array, T& obj) noexcept
```
leoetlino: Any reason `array` is not a const ref?

Techjar: I guess I just forgot.

PR #7823, opened 2019-02-25
Move shared Updater logic into UpdaterCommon

spycrab:
Deduplicates a lot of code.

Notes:
 * More code could probably be shared
 * Some of the implementation details of UpdaterCommon might be unsatisfactory at the moment

PR #8526, opened 2019-12-19
remove x86 targets from sln

shuffle2:
crept in from #7823


stenzek:
@dolphin-emu-bot rebuild

PR #8527, opened 2019-12-19
Update Readme.md

shuffle2:
Reword submodule statement to be more general.
Add pointer about dealing with clang-format.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -39,11 +39,11 @@ Dolphin can only be installed on devices that satisfy the above requirements. At
 ## Building for Windows
 
 Use the solution file `Source/dolphin-emu.sln` to build Dolphin on Windows.
-Visual Studio 2019 16.3 is a hard requirement. Other compilers might be
+Visual Studio 2019 16.3 or later is a hard requirement. Other compilers might be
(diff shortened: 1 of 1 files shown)


JosJuice:
Does the clang-format that's included with Visual Studio not work? (Different version than what's on the lint bot?)

shuffle2:
It produces output which differs from what the bot expects. lint.sh uses 7.0 so I assume it is some behavioral change between llvm releases. (which is lol, but whatever)

JosJuice:
These changes sound reasonable to me, then.

shuffle2:
```
/mnt/c/Program Files (x86)/Microsoft Visual Studio/2019/Community$ ./VC/Tools/Llvm/bin/clang-format.exe --version
clang-format version 9.0.0 (tags/RELEASE_900/final)
```
```
$ clang-format --version
clang-format version 8.0.0-3 (tags/RELEASE_800/final)
$ lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 19.04
Release:        19.04
Codename:       disco
```

PR #8528, opened 2019-12-19
Include <memory> for std::unique_ptr in DXContext.h

CaseyCarter:
`<array>` will no longer include `<memory>` in a future MSVC release.


CaseyCarter:
> Thanks for the change!

Anytime!

PR #8529, opened 2019-12-21
[WIP] HW/DSPHLE: Refactor Old AX Wii into a separate class

Techjar:


leoetlino:
What's the status of this PR, now that #8525 is merged?

Source/Core/Core/HW/DSPHLE/UCodes/AXStructs.h
```diff
@@ -113,6 +113,14 @@ struct PBUpdates
   u16 data_lo;
 };
 
+// Old AX Wii uses update data, but a block is 3ms.
```
BhaaLseN: It's been a while that I worked on the DSP, but this comment doesn't really make sense to me in its current form?

Source/Core/Core/HW/DSPHLE/UCodes/AXWiiOld.cpp
```diff
@@ -0,0 +1,190 @@
...
+      addr2_lo = m_cmdlist[curr_idx++];
+      MixAUXSamples(cmd - CMD_MIX_AUXA, HILO_TO_32(addr), HILO_TO_32(addr2), volume);
+      break;
+
+    case CMD_UPL_AUXA_MIX_LRSC:
+    case CMD_UPL_AUXB_MIX_LRSC:
+    {
+      volume = m_cmdlist[curr_idx++];
+      u32 addresses[6] = {
+          (u32)(m_cmdlist[curr_idx + 0] << 16) | m_cmdlist[curr_idx + 1],
```
BhaaLseN: I wonder if we should clean those up into proper casts while we're at it.

Source/Core/Core/HW/DSPHLE/UCodes/AXStructs.h
Techjar: It's not really clear to me either what update data is. If someone could provide a better explanation I'd be happy to put it in the comment.

PR #8273, opened 2019-07-30
Threading fixes for config layers

CookiePLMonster:
This PR refactors Config Layouts functionality a bit, adding a layer of thread safety to them, as well as improving overall clarity. This fixes several "random" crashes on switching the game on/off and accessing configuration menus at the same time.

1) First commit revisits layer manager in `Config` namespace to make it thread safe. API has been made a bit stricter (`Config::GetLayers` was thankfully never used, though!) and some implementation details have been moved from the header file to the .cpp file. Now configuration layers are stored in a `shared_ptr` and not an `unique_ptr`, which cleanly solves an issue of one thread reading those, with another thread deleting them on terminating the game.
On top of that, any functions accessing `s_layers` have been protected with a read-write lock. It might seem like I added some redundant scopes, but that's not the case - `InvokeConfigChangedCallbacks` **must not** be protected with this mutex, else deadlocks occur!
2) Second commit revisits `LayerMap`. ~Previously, it was defined as `std::map<ConfigLocation, std::optional<std::string>>` and presence of a `nullopt` value was treated equally to lack of a value.~ I was able to remove this redundancy by replacing some uses of `operator[]` with `find()`. This also allowed me to mask `Get` functions as const, which was not possible before due to an implicit construction in `operator[]`. Additionally, I believe this may be a memory/performance improvement, since config layers will now **not** store redundant, empty entries for options they were probed on.
EDIT: This has been partially reverted to again store optionals in a map, since it's needed to achieve correct behaviour of deleting config keys. Fixes around `operator[]` persisted.

### Races fixed

I was unable to reproduce this race in Release mode, most likely due to very specific timing required to reproduce it. However, I was able to reproduce it in Debug:
1) Open Graphics options.
2) Launch the game, and quickly terminate it.
3) Repeat 2) until Dolphin crashes.

There were numerous reasons for those crashes - either one thread was reading from a `Layer` which was already deleted by another thread (that's what usage of `std::shared_ptr` fixes), or map's iterators get corrupted because one thread was iterating through it while the other was adding/removing entries to it (that's what a read-write lock fixes).

**NOTE:** There seems to be at least one more race on starting the game (`GetDSPProcessor()` returning null and being used without any validation), but this will **not** be fixed by this PR. If I find a fix better than "add a null check beforehand", I'll submit another PR.

PR #8530, opened 2019-12-22
MacOS: Fixes configuration hang; bump MacOS SDK.

s-daveb:
Removed conditional use of std::mutex instead of std::shared_mutex on MacOS.

Because MacOS < 10.12 did not support std::shared_mutex, a previous commit
naïvely substituted std::mutex, which does not have the same behavior.

Reverses PR #8273, which substitues std::mutex for std::shared_mutex on
macOS, and results in several bugs that seem to only affect MacOS

- https://bugs.dolphin-emu.org/issues/11919
- https://bugs.dolphin-emu.org/issues/11842
- https://bugs.dolphin-emu.org/issues/11845

This change eliminates conditional code for MacOS in the core configuration
layer code and enables the use of modern language features that are more
secure and thread-safe.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,7 +5,10 @@ cmake_minimum_required(VERSION 3.10)
 set(CMAKE_OSX_ARCHITECTURES "x86_64")
 # Minimum OS X version.
 # This is inserted into the Info.plist as well.
-set(CMAKE_OSX_DEPLOYMENT_TARGET "10.10.0" CACHE STRING "")
+
+# MacOS prior to 10.12 did not fully support C++17, which is used to
+# handle configuration options
+set(CMAKE_OSX_DEPLOYMENT_TARGET "10.12.0" CACHE STRING "")
 
 project(dolphin-emu)
 
--- a/Readme.md
+++ b/Readme.md
@@ -15,7 +15,7 @@ Please read the [FAQ](https://dolphin-emu.org/docs/faq/) before using Dolphin.
 * OS
     * Windows (7 SP1 or higher is officially supported, but Vista SP2 might also work).
     * Linux.
-    * macOS (10.10 Yosemite or higher).
+    * macOS (10.12 Sierra or higher).
     * Unix-like systems other than Linux are not officially supported but might work.
 * Processor
     * A CPU with SSE2 support.
--- a/Source/Core/Common/Config/Config.cpp
+++ b/Source/Core/Common/Config/Config.cpp
@@ -5,11 +5,7 @@
 #include <algorithm>
 #include <list>
 #include <map>
-#if __APPLE__
-#include <mutex>
-#else
 #include <shared_mutex>
-#endif
 
 #include "Common/Config/Config.h"
 
@@ -21,19 +17,10 @@ static Layers s_layers;
 static std::list<ConfigChangedCallback> s_callbacks;
 static u32 s_callback_guards = 0;
 
-// Mac supports shared_mutex since 10.12 and we're targeting 10.10,
-// so only use unique locks there...
-#if __APPLE__
-static std::mutex s_layers_rw_lock;
-
-using ReadLock = std::unique_lock<std::mutex>;
-using WriteLock = std::unique_lock<std::mutex>;
-#else
 static std::shared_mutex s_layers_rw_lock;
 
 using ReadLock = std::shared_lock<std::shared_mutex>;
 using WriteLock = std::unique_lock<std::shared_mutex>;
-#endif
 
 void AddLayerInternal(std::shared_ptr<Layer> layer)
 {


stenzek:
@dolphin-emu-bot rebuild

pizuz:
Is there a reason why there are no macOS checks in this branch? I'd really like to test this.

stenzek:
Probably buildbot issues.

Yep, been stuck for ~3 hours apparently.

JosJuice:
Thanks for figuring out the issue!

s-daveb:
I encountered similar build issues at first. I believe the MacOS Buildbot  needs to have its SDK upgraded.

pizuz:
Looks like this won't even be necessary, since Buildbot is already using the 10.14 SDK. Looks like this argument is to blame:

`/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.14.sdk -mmacosx-version-min=10.10.0`

This results in XCode throwing this error:

`error: 'shared_mutex' is unavailable: introduced in macOS 10.12`

s-daveb:
> > I believe the MacOS Buildbot needs to have its SDK upgraded.
> 
> Looks like this won't even be necessary, since Buildbot is already using the 10.14 SDK. Looks like this argument is to blame:
> 
> `/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.14.sdk -mmacosx-version-min=10.10.0`
> 
> This results in XCode throwing this error:
> 
> `error: 'shared_mutex' is unavailable: introduced in macOS 10.12`

From what I understand of the [CMake documentation](https://cmake.org/cmake/help/v3.11/variable/CMAKE_OSX_DEPLOYMENT_TARGET.html), "CMake uses this variable value for the -mmacosx-version-min flag or their respective target platform equivalents. "

I'm not sure how buildbot is configured, but if there is an environment or commandline parameter hardcoding CMAKE_OSX_DEPLOYMENT_TARGET to 10.10, that could explain the failure.

stenzek:
@dolphin-emu-bot rebuild

stenzek:
We might have to clear the cache on the buildbot for the new default to take effect?

pizuz:
While you‘re at it, why not indeed bump the SDK to 10.15 and MoltenVK to the current release for full Metal 3 support? I‘m not sure if this will make any difference performance-wise, but one might at least give it a try.

s-daveb:
I thought about it, and I realize dolphin is a performance-heavy application that probably doesn’t fare well on older-macs-that-can’t-run-Catalina, but ultimately I decided to bump it only to 10.12 because the compiler pointed out this is the minimum required version for C++17 support.

to;dr I am not part of the project’s development planning, so I bumped it conservatively. 

If anyone wants to go ahead and give me the green-light, I don’t mind making Dolphin a Catalina-only application. I just don’t know the reasons why Yosemite is still supported, and don’t want to step on people’s toes.

s-daveb:
@pizuz I thought about it, and I realize dolphin is a performance-heavy application that probably doesn’t fare well on older-macs-that-can’t-run-Catalina, but ultimately I decided to bump it only to 10.12 because the compiler pointed out this is the minimum required version for C++17 support.

to;dr I am not part of the project’s development planning, so I bumped it conservatively. 

If anyone wants to go ahead and give me the green-light, I don’t mind making Dolphin a Catalina-only application. I just don’t know the reasons why Yosemite is still supported, and don’t want to step on people’s toes.

JosJuice:
We've actually discouraged Dolphin users from updating to Catalina, due to the new notarization requirements. I agree that this PR should not set the required version to anything higher than 10.12.

s-daveb:
> We've actually discouraged Dolphin users from updating to Catalina, due to the new notarization requirements. I agree that this PR should not set the required version to anything higher than 10.12.

@JosJuice  @pizuz Ah, I understand now. From what I'm seeing on developer.apple.com, it seems applications built on older SDK's for the time being are grandfathered in until January 2020.

If push comes to shove, from what I'm seeing on the internet, a continuous integration system could also notarize the application in an automated way.

pizuz:
I'm not even sure if Dolphin actually needs to be compiled agains the 10.15 SDK in order to make use of the new functions in Metal 3 or if the current MoltenVK dylib alone takes care of that, since it's built against the 10.15 SDK.

stenzek:
WRT updating MoltenVK, unless there's a critical issue I'd rather not add megabytes to the repository size when it's not absolutely necessary. There's nothing stopping you from downloading the SDK and replacing it yourself, and this is probably a better way to go about it anyway, rather than including it in the repo (perhaps with an auto downloader).

gtalusan:
> If push comes to shove, from what I'm seeing on the internet, a continuous integration system could also notarize the application in an automated way.

https://github.com/rednoah/notarize-app is a project I use to notarize a Qt app from the command line.  All it needs is an app-level AppleID password and an AppleID that owns the application identifier of the underlying .pkg/.dmg.

FWIW, the notarizing process is a bit chatty over email.

stenzek:
Anyone have any objection to merging this and sorting out the buildbot issues later?

Like I said earlier, it *might* just be matter of clearing the cache/wiping the build directory.

PR #8531, opened 2019-12-22
There is a simulation inside the simulator for freely stored SD cards.

GDWGH:
I want to have a simulated, free-to-store SD card inside the simulator.


leoetlino:
This is not a support forum or a bug tracker.

PR #8532, opened 2019-12-22
VideoCommon/ShaderGenCommon: Convert helper functions over to fmt where applicable

lioncash:
These helper functions are fairly trivial to convert over: no fancy formatting, just simple renaming `Write` over to `WriteFmt` and converting formatting specifiers over to `{}`. This'll make it more straightforward when converting over the shader generators, since `std::string_view` can now be used with these functions without needing to call `.data()`.

While we're in the same area, we can also convert several template functions into normal functions, given the type of their argument is always the same type (`ShaderCode&`). This allows hiding details within the translation unit, allowing changes to the implementation of the functions without needing to recompile every single shader generator translation unit. This also uncovered quite a few indirect inclusions in the process, which have also been amended.

I've separated the formatting changes and the changes that migrate the template functions to regular functions to make reviewing the two changes nicer on the reader.

### DIFF
--- a/Source/Core/VideoCommon/ShaderCache.cpp
+++ b/Source/Core/VideoCommon/ShaderCache.cpp
@@ -15,14 +15,19 @@
 #include "VideoCommon/Statistics.h"
 #include "VideoCommon/VertexLoaderManager.h"
 #include "VideoCommon/VertexManagerBase.h"
+#include "VideoCommon/VideoCommon.h"
+#include "VideoCommon/VideoConfig.h"
 
-#include "imgui.h"
+#include <imgui.h>
 
 std::unique_ptr<VideoCommon::ShaderCache> g_shader_cache;
 
 namespace VideoCommon
 {
-ShaderCache::ShaderCache() = default;
+ShaderCache::ShaderCache() : m_api_type{APIType::Nothing}
+{
+}
+
 ShaderCache::~ShaderCache()
 {
   ClearCaches();
--- a/Source/Core/VideoCommon/ShaderGenCommon.h
+++ b/Source/Core/VideoCommon/ShaderGenCommon.h
@@ -7,7 +7,6 @@
 #include <cstdarg>
 #include <cstring>
 #include <iterator>
-#include <map>
 #include <string>
 #include <type_traits>
 #include <vector>
@@ -16,8 +15,8 @@
 
 #include "Common/CommonTypes.h"
 #include "Common/StringUtil.h"
-#include "VideoCommon/VideoCommon.h"
-#include "VideoCommon/VideoConfig.h"
+
+enum class APIType;
 
 /**
  * Common interface for classes that need to go through the shader generation path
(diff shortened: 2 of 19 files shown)


Source/Core/VideoCommon/ShaderCache.cpp
```diff
@@ -15,14 +15,19 @@
...
 
-#include "imgui.h"
+#include <imgui.h>
 
 std::unique_ptr<VideoCommon::ShaderCache> g_shader_cache;
 
 namespace VideoCommon
 {
-ShaderCache::ShaderCache() = default;
+ShaderCache::ShaderCache() : m_api_type{APIType::Nothing}
```
BhaaLseN: Feels a bit weird to go from direct member initialization and `default` ctor to one that's typed out just to have a forward-declared `enum class` in the header instead. Is this really worth it?

lioncash: I don't particularly see anything unusual about it in this case, given if anyone was concerned about the final state of the internal member variables after construction, then one would need to see what the constructor is doing anyways. I can revert this portion of the change if it's not desirable.

Source/Core/VideoCommon/ShaderGenCommon.h
```diff
@@ -278,29 +206,8 @@ inline void AssignVSOutputMembers(T& object, const char* a, const char* b, u32 t
...
-  }
-  else
-  {
-    if (!ssaa)
-      return "centroid";
-    else
-      return "sample";
-  }
-}
+const char* GetInterpolationQualifier(bool msaa, bool ssaa, bool in_glsl_interface_block = false,
```
CookiePLMonster:
```suggestion
std::string_view GetInterpolationQualifier(bool msaa, bool ssaa, bool in_glsl_interface_block = false,
```

CookiePLMonster: It's probably nearly irrelevant but I noticed that result of this function is passed to a `string_view` argument, so using a string view here may tidy things up a bit.

Source/Core/VideoCommon/ShaderCache.cpp
BhaaLseN: No, was mainly wondering since for the most part, the suggestion is to do the reverse (get rid of the ctor and move it to the members themselves; so it isn't forgotten if another ctor is added).
Just wanting to wrap my head around it.

Source/Core/VideoCommon/ShaderGenCommon.h
lioncash: I'm waiting until all relevant usage cases with the ShaderCode generators are converted over to WriteFmt, so that we're using an API that safely handles custom types like string_view before making this change.

lioncash:
@stenzek Whenever you have the time, can you look this over?

Source/Core/VideoCommon/ShaderCache.cpp
stenzek: I think in this case it's more readable to keep it in the header - it doesn't seem to remove any includes?

lioncash: It doesn't remove any includes in the header because it was relying on an indirect include previously

PR #8534, opened 2019-12-23
Properly check for missing netplay session name/region

JosJuice:


BParks21:
@JosJuice This update is failing and not being applied correctly.

JosJuice:
What do you mean?

BParks21:
It ends with a fatal error using the updater.

Mizoxman:
same issue, when this update tried to apply it downloaded over 2000 files and then had a fatal error. it installed just fine when I downloaded it manually though

BParks21:
It also download 2000 files for me. Where did it store these?

nf6429:
Same issue here, gets stuck on verify part of updater and then fatal error occurs.

PR #8535, opened 2019-12-27
Movie: Make checking for existing GC saves more reliable

JosJuice:
The old code only handled a raw memory card in slot A, not taking GCI folders into account at all.

### DIFF
--- a/Source/Core/Core/HW/EXI/EXI_DeviceMemoryCard.cpp
+++ b/Source/Core/Core/HW/EXI/EXI_DeviceMemoryCard.cpp
@@ -9,6 +9,7 @@
 #include <functional>
 #include <memory>
 #include <string>
+#include <utility>
 
 #include <fmt/format.h>
 
@@ -158,10 +159,34 @@ CEXIMemoryCard::CEXIMemoryCard(const int index, bool gciFolder) : card_index(ind
   SetCardFlashID(header.data(), card_index);
 }
 
+std::pair<std::string /* path */, bool /* migrate */>
+CEXIMemoryCard::GetGCIFolderPath(int card_index, AllowMovieFolder allow_movie_folder)
+{
+  std::string path_override =
+      Config::Get(card_index == 0 ? Config::MAIN_GCI_FOLDER_A_PATH_OVERRIDE :
+                                    Config::MAIN_GCI_FOLDER_B_PATH_OVERRIDE);
+
+  if (!path_override.empty())
+    return {std::move(path_override), false};
+
+  std::string path = File::GetUserPath(D_GCUSER_IDX);
+
+  const bool use_movie_folder = allow_movie_folder == AllowMovieFolder::Yes &&
+                                Movie::IsPlayingInput() && Movie::IsConfigSaved() &&
+                                Movie::IsUsingMemcard(card_index) &&
+                                Movie::IsStartingFromClearSave();
+
+  if (use_movie_folder)
+    path += "Movie" DIR_SEP;
+
+  const DiscIO::Region region = SConfig::ToGameCubeRegion(SConfig::GetInstance().m_region);
+  path = path + SConfig::GetDirectoryForRegion(region) + DIR_SEP +
+         fmt::format("Card {}", char('A' + card_index));
+  return {std::move(path), !use_movie_folder};
+}
+
 void CEXIMemoryCard::SetupGciFolder(u16 sizeMb)
 {
-  const DiscIO::Region region = SConfig::ToGameCubeRegion(SConfig::GetInstance().m_region);
-
   const std::string& game_id = SConfig::GetInstance().GetGameID();
   u32 CurrentGameId = 0;
   if (game_id.length() >= 4 && game_id != "00000000" &&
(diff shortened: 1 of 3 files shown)


Source/Core/Core/HW/EXI/EXI_DeviceMemoryCard.cpp
```diff
@@ -158,10 +158,39 @@ CEXIMemoryCard::CEXIMemoryCard(const int index, bool gciFolder) : card_index(ind
   SetCardFlashID(header.data(), card_index);
 }
 
-void CEXIMemoryCard::SetupGciFolder(u16 sizeMb)
+std::string CEXIMemoryCard::GetGCIFolderPath(int card_index, bool allow_movie_folder, bool* migrate)
```
Warepire: Can this return a `std::pair<std::string, bool>` to avoid the out-param?

JosJuice: It could, but then there would be nothing in the function declaration that tells you what the bool means, so I think it's better the way it is right now.

Warepire: That is not necessarily a problem.
`std::pair<std::string /* ... */, bool /* migrate */>` is a common way to define a pair, then you can use structured bindings to unpack the pair on assignment.

JosJuice: Just one question about structured bindings before I change this. Calling the modified function like this works:

```
const auto [strDirectoryName, migrate] = GetGCIFolderPath(card_index, true);
```

But is there a way to specify the type of the two variables separately, so that I don't have to use auto? That way you wouldn't have to look up the function to be able to know what types the two variables are.

Tilka: > But is there a way to specify the type of the two variables separately, so that I don't have to use auto?

No :(

JosJuice: I've made it use a structured binding.

Source/Core/Core/HW/EXI/EXI_DeviceMemoryCard.cpp
```diff
@@ -158,10 +159,33 @@ CEXIMemoryCard::CEXIMemoryCard(const int index, bool gciFolder) : card_index(ind
...
+                                Movie::IsConfigSaved() && Movie::IsUsingMemcard(card_index) &&
+                                Movie::IsStartingFromClearSave();
+
+  if (use_movie_folder)
+    path += "Movie" DIR_SEP;
+
   const DiscIO::Region region = SConfig::ToGameCubeRegion(SConfig::GetInstance().m_region);
+  path = path + SConfig::GetDirectoryForRegion(region) + DIR_SEP +
+         fmt::format("Card {}", char('A' + card_index));
+  return {path, !use_movie_folder};
```
CookiePLMonster:
```suggestion
  return {std::move(path), !use_movie_folder};
```

Source/Core/Core/HW/EXI/EXI_DeviceMemoryCard.cpp
```diff
@@ -158,10 +159,33 @@ CEXIMemoryCard::CEXIMemoryCard(const int index, bool gciFolder) : card_index(ind
   SetCardFlashID(header.data(), card_index);
 }
 
-void CEXIMemoryCard::SetupGciFolder(u16 sizeMb)
+std::pair<std::string /* path */, bool /* migrate */>
+CEXIMemoryCard::GetGCIFolderPath(int card_index, bool allow_movie_folder)
 {
+  const std::string path_override =
```
leoetlino:
```suggestion
  std::string path_override =
```

(required to avoid copying `path_override`)

Source/Core/Core/HW/EXI/EXI_DeviceMemoryCard.cpp
```diff
@@ -158,10 +159,33 @@ CEXIMemoryCard::CEXIMemoryCard(const int index, bool gciFolder) : card_index(ind
...
+      Config::Get(card_index == 0 ? Config::MAIN_GCI_FOLDER_A_PATH_OVERRIDE :
+                                    Config::MAIN_GCI_FOLDER_B_PATH_OVERRIDE);
+
+  if (!path_override.empty())
+    return {path_override, false};
```
leoetlino:
```suggestion
    return {std::move(path_override), false};
```

Source/Core/Core/HW/EXI/EXI_DeviceMemoryCard.h
```diff
@@ -34,6 +36,9 @@ class CEXIMemoryCard : public IEXIDevice
   static void Init();
   static void Shutdown();
 
+  static std::pair<std::string /* path */, bool /* migrate */>
+  GetGCIFolderPath(int card_index, bool allow_movie_folder);
```
leoetlino: What do you think of making `allow_movie_folder` an enum class for readability?

I feel like `GetGCIFolderPath(0, AllowMovieFolder::Yes)` would be a lot clearer than `GetGCIFolderPath(0, true)`.

Also, regarding the structured binding `auto` grievance, have you considered returning a proper struct? Then you could avoid auto and still make the variable const.

JosJuice: It is now an enum class. I placed it outside of `CEXIMemoryCard` (but inside of `ExpansionInterface`) so that the namespace qualifiers in Movie.cpp wouldn't get too wordy.

I can't think of a good name for such a struct.

leoetlino: Hmm yeah, it's hard to find a good name... `GCIFolderPath` would be the obvious candidate but it feels wrong with the `migrate` bool being in there :/ Well, I guess structured bindings are good enough.

PR #8537, opened 2019-12-28
Core/HW -> PowerPC/JIT: Fastmem arena construction

degasus:
This pull request splits the fastmem arena construction out of the memory mapper. So for the generic emulation, we won't allocate the huge junk of virtual memory.

However the JIT now creates them the fastmem arena if needed. So with disabled fastmem, we won't allocate it any more.

### DIFF
--- a/Source/Core/Core/HW/Memmap.cpp
+++ b/Source/Core/Core/HW/Memmap.cpp
@@ -42,6 +42,7 @@ namespace Memory
 // Store the MemArena here
 u8* physical_base = nullptr;
 u8* logical_base = nullptr;
+static bool is_fastmem_arena_initialized = false;
 
 // The MemArena class
 static Common::MemArena g_arena;
@@ -131,7 +132,7 @@ struct LogicalMemoryView
 // other devices, like the GPU, use other rules, approximated by
 // Memory::GetPointer.) This memory is laid out as follows:
 // [0x00000000, 0x02000000) - 32MB RAM
-// [0x02000000, 0x08000000) - Mirrors of 32MB RAM
+// [0x02000000, 0x08000000) - Mirrors of 32MB RAM (not handled here)
 // [0x08000000, 0x0C000000) - EFB "mapping" (not handled here)
 // [0x0C000000, 0x0E000000) - MMIO etc. (not handled here)
 // [0x10000000, 0x14000000) - 64MB RAM (Wii-only; slightly slower)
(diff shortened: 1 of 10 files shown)


Source/Core/Core/HW/Memmap.cpp
```diff
@@ -190,15 +200,14 @@ void Init()
...
 
+  // Create an anonymous view of the physical memory
   for (PhysicalMemoryRegion& region : physical_regions)
   {
     if ((flags & region.flags) != region.flags)
       continue;
 
-    u8* base = physical_base + region.physical_address;
-    *region.out_pointer = (u8*)g_arena.CreateView(region.shm_position, region.size, base);
+    *region.out_pointer = (u8*)g_arena.CreateView(region.shm_position, region.size);
```
BhaaLseN: What do you think of replacing those C-style casts with the appropriate C++ cast while we're at it?

Source/Core/Core/PowerPC/Jit64/Jit_LoadStore.cpp
```diff
@@ -304,7 +304,9 @@ void Jit64::dcbz(UGeckoInstruction inst)
     AND(32, R(RSCRATCH), Imm32(~31));
   }
 
-  if (MSR.DR)
+  bool emit_fast_path = MSR.DR && m_jit.jo.fastmem_arena;
```
stenzek: Can this be `jo.fastmem` instead as `fastmem` checks for `fastmem_arena` in the base class?

Source/Core/Core/PowerPC/Jit64Common/EmuCodeBlock.cpp
```diff
@@ -371,7 +371,7 @@ void EmuCodeBlock::SafeLoadToReg(X64Reg reg_value, const Gen::OpArg& opAddress,
 
   FixupBranch exit;
   const bool dr_set = (flags & SAFE_LOADSTORE_DR_ON) || MSR.DR;
-  const bool fast_check_address = !slowmem && dr_set;
+  const bool fast_check_address = !slowmem && dr_set && m_jit.jo.fastmem_arena;
```
stenzek: Same with these

Source/Core/Core/PowerPC/Jit64/Jit_LoadStore.cpp
degasus: I tried to keep this separate. jo.fastmem_arena might be turned on without the exception handlers. The checks for jo.fastmem requires both.

JMC47:
Is this ready to go?  I see pending questions that don't appear to be answered.  When they're taking care of, I'm willing to merge.

stenzek:
degasus explained the reasoning on IRC. I probably should've mentioned that in my review.

PR #8539, opened 2019-12-29
IOS/FS: Reimplement many functions in a more accurate way

leoetlino:
Some official titles rely on implementation details of Nintendo's FS sysmodule and will not work properly if those are changed. Notably, some games and older versions of the System Menu appear to be relying on the order of files returned by FS::ReadDirectory and will either fail to find their save data (for [Bolt](https://wiki.dolphin-emu.org/index.php?title=Bolt)) or outright crash (for the [System Menu](https://wiki.dolphin-emu.org/index.php?title=Wii_Menu)).

Some titles also actually expect filesystem metadata to be correct. One title that has been confirmed to do this is [DQX](https://wiki.dolphin-emu.org/index.php?title=Dragon_Quest_X), which generates paths based on the GID of files within its own title directory.

While it is easy to make workarounds for these issues -- and in fact we already do have some for the sysmenu and DQX, having hacks is obviously nonideal and adding yet another hack would be required to fix Bolt -- one that would be even uglier.

Furthermore, while it is currently unknown whether any official title cares about permissions, the lack of FS metadata means that we are unable to implement them if that turns out to be desirable or necessary.

By adding a FST, we can implement things correctly and solve all those problems without adding hacks. In fact, this PR even *gets rid* of two hacks in the FS code.

For simplicity, a binary format that is inspired from Nintendo's FST structure was chosen for serialization. It is not expected to ever receive an update.

Most of the new code that is added by this PR is additional checks or necessary logic which were missing from the previous implementation. I've also seized this opportunity to even implement permission checks, since obscure IOS features have a history of coming into play at unexpected times...

The last few commits are mostly unit test fixes or additions. All tests passed.

### Results

This PR fixes Bolt, which is now able to find its save files:

RLUE4Q_2019-12-29_19-42-55

Two hacks were removed.

Apart from DQX, the sysmenu and Bolt, this PR also happens to fix the [Photo Channel](https://wiki.dolphin-emu.org/index.php?title=Photo_Channel) complaining about corrupted system files on the initial launch.

Before (initial boot):

HAAA01_2019-12-29_19-40-00

After (initial boot):

HAAA01_2019-12-29_19-43-20

### An update on the NAND image backend

A long time ago I had planned to add another FS backend which would be using a NAND image/blob as the storage. While I have already written an implementation that has been tested, solves all the aforementioned issues and more, produces images that are fully compatible with IOS's FS driver, I feel like NAND images raise too many issues: savestate sizes, code complexity and maintenance cost.

Since many fixes and additions that are part of that branch (e.g. FS timings, utility structures, FST) have already been merged or submitted as part of this PR, I will likely not submit it.

### DIFF
--- a/Source/Core/Core/IOS/FS/FileSystem.cpp
+++ b/Source/Core/Core/IOS/FS/FileSystem.cpp
@@ -11,6 +11,17 @@
 
 namespace IOS::HLE::FS
 {
+bool IsValidPath(std::string_view path)
+{
+  return path == "/" || IsValidNonRootPath(path);
+}
+
+bool IsValidNonRootPath(std::string_view path)
+{
+  return path.length() > 1 && path.length() <= MaxPathLength && path[0] == '/' &&
+         path.back() != '/';
+}
+
 std::unique_ptr<FileSystem> MakeFileSystem(Location location)
 {
   const std::string nand_root =
--- a/Source/Core/Core/IOS/FS/FileSystem.cpp
+++ b/Source/Core/Core/IOS/FS/FileSystem.cpp
@@ -22,6 +22,13 @@ bool IsValidNonRootPath(std::string_view path)
          path.back() != '/';
 }
 
+SplitPathResult SplitPathAndBasename(std::string_view path)
+{
+  const auto last_separator = path.rfind('/');
+  return {std::string(path.substr(0, std::max<size_t>(1, last_separator))),
+          std::string(path.substr(last_separator + 1))};
+}
+
 std::unique_ptr<FileSystem> MakeFileSystem(Location location)
 {
   const std::string nand_root =
--- a/Source/Core/Core/IOS/FS/FileSystem.cpp
+++ b/Source/Core/Core/IOS/FS/FileSystem.cpp
@@ -84,12 +84,6 @@ Result<FileStatus> FileHandle::GetStatus() const
   return m_fs->GetFileStatus(*m_fd);
 }
 
-void FileSystem::Init()
-{
-  if (Delete(0, 0, "/tmp") == ResultCode::Success)
-    CreateDirectory(0, 0, "/tmp", 0, {Mode::ReadWrite, Mode::ReadWrite, Mode::ReadWrite});
-}
-
 Result<FileHandle> FileSystem::CreateAndOpenFile(Uid uid, Gid gid, const std::string& path,
                                                  Modes modes)
 {
--- a/Source/Core/Core/IOS/FS/FileSystem.h
+++ b/Source/Core/Core/IOS/FS/FileSystem.h
@@ -77,6 +77,16 @@ struct Modes
 {
   Mode owner, group, other;
 };
+inline bool operator==(const Modes& lhs, const Modes& rhs)
+{
+  const auto fields = [](const Modes& obj) { return std::tie(obj.owner, obj.group, obj.other); };
+  return fields(lhs) == fields(rhs);
+}
+
+inline bool operator!=(const Modes& lhs, const Modes& rhs)
+{
+  return !(lhs == rhs);
+}
 
 struct Metadata
 {
--- a/Source/Core/Core/IOS/FS/FileSystem.h
+++ b/Source/Core/Core/IOS/FS/FileSystem.h
@@ -216,9 +216,6 @@ public:
   virtual Result<NandStats> GetNandStats() = 0;
   /// Get usage information about a directory (used cluster and inode counts).
   virtual Result<DirectoryStats> GetDirectoryStats(const std::string& path) = 0;
-
-protected:
-  void Init();
 };
 
 template <typename T>
--- a/Source/Core/Core/IOS/FS/FileSystem.h
+++ b/Source/Core/Core/IOS/FS/FileSystem.h
@@ -7,6 +7,7 @@
 #include <memory>
 #include <optional>
 #include <string>
+#include <string_view>
 #include <vector>
 
 #ifdef _WIN32
@@ -110,6 +111,13 @@ struct FileStatus
   u32 offset;
   u32 size;
 };
+
+/// The maximum number of characters a path can have.
+constexpr size_t MaxPathLength = 64;
+
+/// Returns whether a Wii path is valid.
+bool IsValidPath(std::string_view path);
+bool IsValidNonRootPath(std::string_view path);
 
 class FileSystem;
 class FileHandle final
--- a/Source/Core/Core/IOS/FS/FileSystem.h
+++ b/Source/Core/Core/IOS/FS/FileSystem.h
@@ -112,12 +112,24 @@ struct FileStatus
   u32 size;
 };
 
+/// The maximum number of components a path can have.
+constexpr size_t MaxPathDepth = 8;
 /// The maximum number of characters a path can have.
 constexpr size_t MaxPathLength = 64;
 
 /// Returns whether a Wii path is valid.
 bool IsValidPath(std::string_view path);
 bool IsValidNonRootPath(std::string_view path);
+
+struct SplitPathResult
+{
+  std::string parent;
+  std::string file_name;
+};
+/// Split a path into a parent path and the file name. Takes a *valid non-root* path.
+///
+/// Example: /shared2/sys/SYSCONF => {/shared2/sys, SYSCONF}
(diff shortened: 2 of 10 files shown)


Source/Core/Core/IOS/FS/FileSystem.cpp
```diff
@@ -11,6 +11,24 @@
...
+
+bool IsValidNonRootPath(std::string_view path)
+{
+  return path.length() > 1 && path.length() <= MaxPathLength && path[0] == '/' &&
+         path.back() != '/';
+}
+
+SplitPathResult SplitPath(std::string_view path)
+{
+  const auto last_separator = path.find_last_of('/');
```
lioncash:
```suggestion
  const auto last_separator = path.rfind('/');
```

Source/Core/Core/IOS/FS/FileSystem.h
```diff
@@ -76,6 +77,16 @@ struct Modes
 {
   Mode owner, group, other;
 };
+inline bool operator==(const Modes& lhs, const Modes& rhs)
+{
+  auto fields = [](const Modes& obj) { return std::tie(obj.owner, obj.group, obj.other); };
```
lioncash:
```suggestion
  const auto fields = [](const Modes& obj) { return std::tie(obj.owner, obj.group, obj.other); };
```

Source/Core/Core/IOS/FS/FileSystem.h
```diff
@@ -111,6 +122,36 @@ struct FileStatus
...
+bool IsValidNonRootPath(std::string_view path);
+
+struct SplitPathResult
+{
+  std::string parent;
+  std::string file_name;
+};
+inline bool operator==(const SplitPathResult& lhs, const SplitPathResult& rhs)
+{
+  auto fields = [](const SplitPathResult& obj) { return std::tie(obj.parent, obj.file_name); };
```
lioncash:
```suggestion
  const auto fields = [](const SplitPathResult& obj) { return std::tie(obj.parent, obj.file_name); };
```

Source/Core/Core/IOS/FS/HostBackend/FS.cpp
```diff
@@ -159,179 +335,307 @@ void HostFileSystem::DoState(PointerWrap& p)
...
-  // if we use the host filesystem.
-  std::sort(entry.children.begin(), entry.children.end(),
-            [](const File::FSTEntry& one, const File::FSTEntry& two) {
-              return one.virtualName < two.virtualName;
+  // Sort files according to their order in the FST tree (issue 10234).
+  // The result should look like this:
+  //     [FilesNotInFst, ..., OldestFileInFst, ..., NewestFileInFst]
+  std::unordered_map<std::string_view, int> sort_keys;
+  for (size_t i = 0; i < entry->children.size(); ++i)
+    sort_keys[entry->children[i].name] = int(i);
```
lioncash:
```suggestion
    sort_keys.emplace(entry->children[i].name, int(i));
```

(or `insert_or_assign()`, whichever is preferable)

Source/Core/Core/IOS/FS/HostBackend/FS.cpp
```diff
@@ -159,179 +335,307 @@ void HostFileSystem::DoState(PointerWrap& p)
...
 
-  // NOTE(leoetlino): this is absolutely wrong, but there is no way to fix this properly
-  // if we use the host filesystem.
-  std::sort(entry.children.begin(), entry.children.end(),
-            [](const File::FSTEntry& one, const File::FSTEntry& two) {
-              return one.virtualName < two.virtualName;
+  // Sort files according to their order in the FST tree (issue 10234).
+  // The result should look like this:
+  //     [FilesNotInFst, ..., OldestFileInFst, ..., NewestFileInFst]
+  std::unordered_map<std::string_view, int> sort_keys;
```
lioncash: This could also reserve the memory ahead of time, given the entry's child sizes are known)

Source/Core/Core/IOS/FS/HostBackend/FS.cpp
```diff
@@ -44,13 +47,186 @@ static u64 ComputeTotalFileSize(const File::FSTEntry& parent_entry)
...
+
+void HostFileSystem::LoadFst()
+{
+  File::IOFile file{GetFstFilePath(), "rb"};
+  // Existing filesystems will not have a FST. This is not a problem,
+  // as the rest of HostFileSystem will use sane defaults.
+  if (!file)
+    return;
+
+  auto parse_entry = [&file](const auto& parse, size_t depth) -> std::optional<FstEntry> {
```
lioncash:
```suggestion
  const auto parse_entry = [&file](const auto& parse, size_t depth) -> std::optional<FstEntry> {
```

Source/Core/Core/IOS/FS/HostBackend/FS.cpp
```diff
@@ -159,179 +335,307 @@ void HostFileSystem::DoState(PointerWrap& p)
...
 
-ResultCode HostFileSystem::CreateFile(Uid, Gid, const std::string& path, FileAttribute, Modes)
+ResultCode HostFileSystem::CreateFileOrDirectory(Uid uid, Gid gid, const std::string& path,
+                                                 FileAttribute attr, Modes modes, bool is_file)
 {
-  std::string file_name(BuildFilename(path));
-  // check if the file already exist
-  if (File::Exists(file_name))
+  if (!IsValidNonRootPath(path) ||
+      std::any_of(path.begin(), path.end(), [](char c) { return c - ' ' > 0x5e; }))
```
lioncash: This should likely have a clarifying comment.

leoetlino: Ah yeah, definitely. This is what the FS sysmodule does:

```c
  if ( len > 0 )
  {
    while ( (((u8)path[i] - 0x20) & 0xFFu) <= 0x5E )
    {
      if ( ++i >= len )
        goto path_ok;
    }
    ret = FS_INVALID;
  }
```

...and now I've noticed the check is incorrect, since c is supposed to be unsigned so as to rely on modular arithmetic. Thanks for catching this! I'll replace this with std::isprint since that's what it's really checking.

JMC47:
I still want a NAND image backend.  I know there are people that would appreciate it.  I know this fixes the biggest issues, but there are some advantages to a NAND image backend like uh... savestates would be way bigger.  I mean, there has to be SOME advantage.

leoetlino:
@lioncash Fixed. I've added a new commit for IsPrintableCharacter.

@JMC47 One advantage is that you'd be able to use it for testing with IOS, since it's fully compatible with Nintendo's driver. And I guess another one is that it becomes impossible to have a filesystem that goes over the size limit. But IMO the drawbacks far outweigh those benefits.

Source/Core/Core/IOS/FS/HostBackend/FS.cpp
```diff
@@ -44,13 +47,186 @@ static u64 ComputeTotalFileSize(const File::FSTEntry& parent_entry)
...
+    std::memcpy(name.data(), new_name.data(), std::min(name.size(), new_name.length()));
+  }
+
+  /// File name
+  std::array<char, 12> name{};
+  /// File owner user ID
+  Common::BigEndianValue<Uid> uid{};
+  /// File owner group ID
+  Common::BigEndianValue<Gid> gid{};
+  /// Is file
```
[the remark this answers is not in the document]
leoetlino: Haha yeah, fixed!

boogerlad:
This breaks Mario Kart Wii initial saves for me. Unable to test subsequent saves.
5.0-11558 works fine but 5.0-11578 does not on Windows 10 x64

Techjar:
To clarify the above comment, with a fresh NAND I get this message after it attempts to create the initial save file:


I can't progress past this message, so I restart the game and am presented with this:


Pressing OK then attempts once again to create the initial save and fails, looping back to the first message, thereby rendering Mario Kart Wii unplayable unless you import a good save or have one from a previous build.

Techjar:
If I install the system menu, boot it and do the initial setup, Mario Kart Wii is able to successfully create its save.

JMC47:
inb4 this is just how the game would act without a properly setup NAND.

boogerlad:
Thanks for the clarification @Techjar 
I can confirm that as of 5.0-11578, the issue is resolved. Thank you @leoetlino

PR #8541, opened 2019-12-30
StringUtil: Make TryParse of floats handle comma and dot decimal separators.

jordan-woyak:
This should fix: https://bugs.dolphin-emu.org/issues/11899
(floats fail to parse on locales with non-dot decimal separators)

I also noticed signed integer and enum types were being handled by the overload intended for floats so I generalized the integer version of `TryParse`.

### DIFF
--- a/Source/Core/Common/StringUtil.h
+++ b/Source/Core/Common/StringUtil.h
@@ -6,7 +6,9 @@
 
 #include <cstdarg>
 #include <cstddef>
+#include <cstdlib>
 #include <iomanip>
+#include <locale>
 #include <sstream>
 #include <string>
 #include <type_traits>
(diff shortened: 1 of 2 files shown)


AdmiralCurtiss:
This instead causes problems if someone has thousand-separators. Shouldn't we just force the C locale for config files?

jordan-woyak:
@AdmiralCurtiss It looks like our "ToString" functions do not use thousand-separator format strings so this shouldn't be a problem.
I agree we should save in the "C" locale but we still need to handle loading of existing non-"C"-locale configs.

JosJuice:
Yeah, I can't recall Dolphin ever writing thousands separators to config files.

CookiePLMonster:
Decimal separators are rather rare I think and I'd be surprised if they were written in configs - for what it's worth, I think it's worth going as far as breaking some configs if it means achieving perfect consistency and making configs fully locale agnostic moving forward. It's a bit surprising they were following locales in the first place...

JosJuice:
Decimal separators are common in controller config INIs.

CookiePLMonster:
Ugh my bad, of course I meant thousands separators.

Source/Core/Common/StringUtil.h
```diff
@@ -47,20 +47,60 @@ std::string ArrayToString(const u8* data, u32 size, int line_len = 20, bool spac
...
+  // Set errno to a clean slate.
+  errno = 0;
+
+  // Read a u64 for unsigned types and s64 otherwise.
+  using ReadType = std::conditional_t<std::is_unsigned_v<T>, u64, s64>;
+  ReadType value;
 
-  N tmp;
+  if constexpr (std::is_unsigned_v<T>)
+    value = strtoull(str.c_str(), &end_ptr, 0);
```
leoetlino:
```suggestion
    value = std::strtoull(str.c_str(), &end_ptr, 0);
```

`<cstdlib>` needs to be included, too

Source/Core/Common/StringUtil.h
```diff
@@ -47,20 +47,60 @@ std::string ArrayToString(const u8* data, u32 size, int line_len = 20, bool spac
...
+  else
+    value = strtoll(str.c_str(), &end_ptr, 0);
```
leoetlino:
```suggestion
    value = std::strtoll(str.c_str(), &end_ptr, 0);
```

Source/Core/Common/StringUtil.h
```diff
@@ -47,20 +47,60 @@ std::string ArrayToString(const u8* data, u32 size, int line_len = 20, bool spac
...
+}
+
+template <typename T, std::enable_if_t<std::is_floating_point_v<T>>* = nullptr>
+bool TryParse(std::string str, T* const output)
+{
+  // Replace commas with dots.
+  std::istringstream iss(ReplaceAll(std::move(str), ",", "."));
+
+  // Use "classic" locale to force a "dot" decimal separator.
+  iss.imbue(std::locale::classic());
```
leoetlino: `<locale>` should be included

PR #8542, opened 2019-12-31
Fix Gecko codes loading

Itrimel:
Fixes https://bugs.dolphin-emu.org/issues/11809

I found on my system (Archlinux) that using a locale with UTF-8 was breaking loading of Gecko codes. Hopefully forcing the C locale on this string stream should be enough without breaking it on other systems.

### DIFF
--- a/Source/Core/Core/GeckoCodeConfig.cpp
+++ b/Source/Core/Core/GeckoCodeConfig.cpp
@@ -85,6 +85,11 @@ std::vector<GeckoCode> DownloadCodes(std::string gametdb_id, bool* succeeded)
     {
       std::istringstream ssline(line);
       std::string addr, data;
+
+      // Some locales (e.g. fr_FR.UTF-8) don't split the string stream on space
+      // Use the C locale to workaround this behavior
+      ssline.imbue(std::locale::classic());
+
       ssline >> addr >> data;
       ssline.seekg(0);
 
@@ -138,6 +143,10 @@ std::vector<GeckoCode> LoadCodes(const IniFile& globalIni, const IniFile& localI
     for (auto& line : lines)
     {
       std::istringstream ss(line);
+
+      // Some locales (e.g. fr_FR.UTF-8) don't split the string stream on space
+      // Use the C locale to workaround this behavior
+      ss.imbue(std::locale::classic());
 
       switch ((line)[0])
       {


Source/Core/Core/GeckoCodeConfig.cpp
```diff
@@ -139,6 +139,9 @@ std::vector<GeckoCode> LoadCodes(const IniFile& globalIni, const IniFile& localI
     {
       std::istringstream ss(line);
 
+      //Force ISO-8859-1 encoding, UTF-8 breaks formating
+      ss.imbue(std::locale("C"));
```
BhaaLseN: You should probably use this instead:
```suggestion
      ss.imbue(std::locale::classic());
```

Itrimel: Out of curiosity, what does it change ?

BhaaLseN: It's a specialized class rather than doing a lookup on the "C" locale (which internally probably does just the same in the end), so it _should_ not make a difference; but thats more in line with what we have everywhere else.

JosJuice:
Do we know why this fixes it? It doesn't intuitively make sense to me.

jordan-woyak:
FYI, The `DownloadCodes` function in this file also needs to be fixed.

I can reproduce the problem using `LANG=fr_FR.UTF-8` but I'm still trying to figure out why this is even a problem in the first place. We have other instances of `std::hex` in Dolphin that may need attention.

Itrimel:
While searching where this behavior could come from, I stumbled upon this : 
while using this commit (https://github.com/Itrimel/dolphin/commit/b9844e5a4e243ddddee4839d539ff9e41bbde872) on master and the locale `fr_FR.UTF-8`, the second part of all codes (which was before `00000000`) will turn into a number which can be changed, among other things, by adding or removing comments from the source file (I typically get something around `00007f50`).

So ... I am very lost about the true origin of this bug ...

jordan-woyak:
@Itrimel The second value is uninitialized memory after the first value fails to read.
Weird values are to be expected.

Why the first value fails to read is the mystery. I see no pattern between "good" and "bad" values (e.g. 18000058 fails but 18000024 parses fine, consistently). Reading a `u64` and afterwards casting to `u32` works around the problem for some reason. And I have no idea why certain locales fail and others always work.

I also can't reproduce the reproblem in a self contained test case, only within Dolphin.

Itrimel:
Well that's bad ... I can also fix DownloadCodes (so that the display just after download works), but if there are more places to fix I think it would be best for you to take over.

CookiePLMonster:
@dolphin-emu-bot rebuild

Itrimel:
Updated to better reflect jordan-woynak's findings in the comments and the commit message, and to add the fix to DownloadCodes function

PR #8545, opened 2020-01-04
WiimoteEmu: IMU pointing behavior improvements and code cleanup.

jordan-woyak:
- Replaced IMU cursor's hacky Euler angle math with 3x3 matrices.
- Enabled "Recenter" button for both yaw and pitch adjustment (previously just yaw was done).
- Added a couple of matrix functions and other minor cleanups.

Pointing now behaves when the controller is in odd orientations (e.g. tilted, sideways, upside-down).
There are fewer branches and special cases in the math and I think it's more readable.

### DIFF
--- a/Source/Core/Core/HW/WiimoteEmu/Dynamics.cpp
+++ b/Source/Core/Core/HW/WiimoteEmu/Dynamics.cpp
@@ -54,6 +54,29 @@ double CalculateStopDistance(double velocity, double max_accel)
   return velocity * velocity / (2 * std::copysign(max_accel, velocity));
 }
 
+Common::Matrix33 ComplementaryFilter(const Common::Vec3& accelerometer,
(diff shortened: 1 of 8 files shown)


Source/Core/Core/HW/WiimoteEmu/Dynamics.cpp
```diff
@@ -54,10 +54,35 @@ double CalculateStopDistance(double velocity, double max_accel)
...
+  {
+    return gyroscope;
+  }
+}
+
 }  // namespace
 
 namespace WiimoteEmu
 {
+IMUCursorState::IMUCursorState() : rotation{Common::Matrix33::Identity()}, recentered_pitch{}
```
lioncash: `recentered_pitch` can be initialized within the class definition.

JMC47:
Seems to help a lot if you're holding the Emulated Remote in really weird directions.

PR #8548, opened 2020-01-04
Core/WiimoteReal: Wii remote connection pool and source change cleanup.

jordan-woyak:
Replaced wide usages of `g_wiimote_sources` with get/set functions.
Made the unnamed enum type an enum class, `WiimoteSource`.

Connected real Wii remotes are now added to a "pool" when no open slot is available.
Wii remotes in the pool have LEDs 1 and 4 lit to make it apparent they are not active.
They are disconnected per normal after 5 seconds.
This allows reassigning sources and reordering remotes without the hassle of Bluetooth re-connections.

This is mainly in preparation for Wii remotes in controller interface.
My plan is an option on the Wii remote dialog, "Connect Wii Remotes for input configuration" (or something with a better name).
Enabling it will first fill any "Real" slots but additional remotes will be sent to controller interface.
Disabling the option will send spurious remotes to the pool.

I think this "pool" will make things less horrible for users that are playing around with options.
Making Wii remote connections can be not-fun, especially on Windows/macOS.
Holding on to a connection for a few seconds when between configurations will give users a better experience

Perhaps someone ambitious might make a UI for the pool.
Remotes could be identified with rumble and sent to a slot or controller interface.

This all of course only affects emulated Bluetooth.

### DIFF
--- a/Source/Core/Core/HW/Wiimote.cpp
+++ b/Source/Core/Core/HW/Wiimote.cpp
@@ -25,6 +25,38 @@
 // Limit the amount of wiimote connect requests, when a button is pressed in disconnected state
 static std::array<u8, MAX_BBMOTES> s_last_connect_request_counter;
 
+namespace WiimoteCommon
+{
+static std::array<std::atomic<WiimoteSource>, MAX_BBMOTES> s_wiimote_sources;
+
+WiimoteSource GetSource(unsigned int index)
+{
+  return s_wiimote_sources[index];
+}
+
+void SetSource(unsigned int index, WiimoteSource source)
+{
+  const WiimoteSource previous_source = s_wiimote_sources[index].exchange(source);
+
+  if (previous_source == source)
+  {
+    // No change. Do nothing.
+    return;
+  }
+
+  WiimoteReal::HandleWiimoteSourceChange(index);
+
+  // Reconnect to the emulator.
+  Core::RunAsCPUThread([index, previous_source, source] {
+    if (previous_source != WiimoteSource::None)
+      ::Wiimote::Connect(index, false);
+
+    if (source == WiimoteSource::Emulated)
+      ::Wiimote::Connect(index, true);
+  });
+}
+}  // namespace WiimoteCommon
+
 namespace Wiimote
 {
 static InputConfig s_config(WIIMOTE_INI_NAME, _trans("Wii Remote"), "Wiimote");
@@ -155,7 +187,7 @@ void Pause()
 // An L2CAP packet is passed from the Core to the Wiimote on the HID CONTROL channel.
 void ControlChannel(int number, u16 channel_id, const void* data, u32 size)
 {
-  if (WIIMOTE_SRC_EMU == g_wiimote_sources[number])
+  if (WiimoteCommon::GetSource(number) == WiimoteSource::Emulated)
   {
     static_cast<WiimoteEmu::Wiimote*>(s_config.GetController(number))
         ->ControlChannel(channel_id, data, size);
@@ -169,7 +201,7 @@ void ControlChannel(int number, u16 channel_id, const void* data, u32 size)
 // An L2CAP packet is passed from the Core to the Wiimote on the HID INTERRUPT channel.
 void InterruptChannel(int number, u16 channel_id, const void* data, u32 size)
 {
-  if (WIIMOTE_SRC_EMU == g_wiimote_sources[number])
+  if (WiimoteCommon::GetSource(number) == WiimoteSource::Emulated)
   {
     static_cast<WiimoteEmu::Wiimote*>(s_config.GetController(number))
         ->InterruptChannel(channel_id, data, size);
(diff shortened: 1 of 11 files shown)


Source/Core/Core/HW/Wiimote.h
```diff
@@ -44,14 +44,14 @@ enum
 
 #define WIIMOTE_INI_NAME "WiimoteNew"
 
-enum
+enum WiimoteSource
```
leoetlino: Thoughts on turning this into an enum class?

Source/Core/Core/HW/WiimoteReal/WiimoteReal.cpp
```diff
@@ -291,8 +291,7 @@ bool Wiimote::IsBalanceBoard()
 
     switch (InputReportID(buf[1]))
     {
-    case InputReportID::Status:
-    {
+    case InputReportID::Status: {
```
leoetlino: clang-format does *not* want the brace to be on its own line? o.O

If so, could we fix the configuration? It looks inconsistent with the rest of the code style...

Source/Core/Core/HW/Wiimote.cpp
```diff
@@ -182,24 +182,26 @@ void InterruptChannel(int number, u16 channel_id, const void* data, u32 size)
...
-    if (g_wiimote_sources[number] && NetPlay::IsNetPlayRunning())
+    if (source != WiimoteSource::WIIMOTE_SRC_NONE && NetPlay::IsNetPlayRunning())
       Wiimote::NetPlay_GetButtonPress(number, false);
     return false;
   }
 
   bool button_pressed = false;
 
-  if (WIIMOTE_SRC_EMU & g_wiimote_sources[number])
+  if (WIIMOTE_SRC_EMU == source)
```
leoetlino: `if (source == WIIMOTE_SRC_EMU)`
(I think we don't use Yoda conditions :P)

Source/Core/Core/HW/WiimoteReal/WiimoteReal.cpp
```diff
@@ -49,9 +47,49 @@ static std::mutex s_known_ids_mutex;
...
 
+static void ProcessWiimotePool()
+{
+  std::lock_guard<std::mutex> wm_lk(g_wiimotes_mutex);
+
+  for (auto it = g_wiimote_pool.begin(); it != g_wiimote_pool.end();)
+  {
+    if (it->IsExpired())
+    {
+      NOTICE_LOG(WIIMOTE, "Removing expired wiimote pool entry.");
```
BhaaLseN: Is this really a message that must always be visible in the log? If not, you should probably demote it to a very low level (like `INFO` or even `DEBUG`).
Reminder that `NOTICE` is "always visible" in Dolphin.

Source/Core/Core/HW/WiimoteReal/WiimoteReal.cpp
```diff
@@ -840,7 +852,12 @@ static void TryToConnectWiimote(std::unique_ptr<Wiimote> wm)
       return;
   }
 
-  NOTICE_LOG(WIIMOTE, "No open slot for real wiimote.");
+  NOTICE_LOG(WIIMOTE, "No open slot for real wiimote, adding it to the pool.");
```
BhaaLseN: Same down here wrt `NOTICE`

Source/Core/Core/NetPlayClient.cpp
```diff
@@ -1510,7 +1510,8 @@ bool NetPlayClient::StartGame(const std::string& path)
   }
 
   for (unsigned int i = 0; i < 4; ++i)
-    WiimoteReal::ChangeWiimoteSource(i, m_wiimote_map[i] > 0 ? WIIMOTE_SRC_EMU : WIIMOTE_SRC_NONE);
+    WiimoteCommon::SetSource(i,
```
BhaaLseN: This now spans multiple lines, and should have curlies for clarity.

Source/Core/DolphinQt/MainWindow.cpp
```diff
@@ -115,6 +115,7 @@
 
 #if defined(HAVE_XRANDR) && HAVE_XRANDR
 #include "UICommon/X11Utils.h"
+#undef None
```
BhaaLseN: I wonder if this should be commented; so we know where it comes from (as a reminder that we might be able to remove it later at some point). Thoughts?

Source/Core/Core/HW/Wiimote.cpp
jordan-woyak: Yoda conditional it already was.
I think there are still quite a few of them. I can change them if we really want.

Source/Core/Core/HW/WiimoteReal/WiimoteReal.cpp
jordan-woyak: I'll make it an `INFO`. I can never remember the difference.

Source/Core/DolphinQt/MainWindow.cpp
jordan-woyak: I added a comment.

iwubcode:
I haven't tested this but code wise, this looks really nice!

PR #8550, opened 2020-01-05
IOS/ES: Add SetUid exception for the Wii U Transfer Tool

leoetlino:
The ES sysmodule in IOS62 (v6430) has an exception for the
Wii U Transfer Tool in the SetUid function.

If the active title is the Wii U Transfer Tool, then calling SetUid
is always allowed. (The UID is still checked first, though.)

Fixes https://bugs.dolphin-emu.org/issues/10985

### DIFF
--- a/Source/Core/Core/IOS/ES/ES.cpp
+++ b/Source/Core/Core/IOS/ES/ES.cpp
@@ -168,13 +168,26 @@ static bool UpdateUIDAndGID(Kernel& kernel, const IOS::ES::TMDReader& tmd)
   return true;
 }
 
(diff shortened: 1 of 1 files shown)


AbnormalPoof:
can this fix also be applied to the wii u transfer tool's vwii counterpart? (named "Wii System Transfer")
it has the same issue

Pokechu22:
@AbnormalPoof: See https://bugs.dolphin-emu.org/issues/11985. The same fix (more or less) _can_ be applied, though perhaps with additional restrictions as to the specific IOS version.

However, from what I recall, the VWii components are signed using a different key from normal Wii titles, which would need to be added to dolphin, and that has other issues currently.

PR #8552, opened 2020-01-06
WiimoteEmu: Nunchuk and Classic Controller calibration accuracy improvements.

jordan-woyak:
Nunchuk and Classic Controller calibration hold values imposed by the stick "gate" on real hardware.
We incorrectly provided the full range beyond the stick gate.

The Classic Controller's trigger calibration is the neutral value to 8 bits of precision, not the maximum value.

These changes are probably mostly unnoticeable. Games seem to mostly ignore the calibration data.

### DIFF
--- a/Source/Core/Core/HW/WiimoteEmu/Extension/Classic.cpp
+++ b/Source/Core/Core/HW/WiimoteEmu/Extension/Classic.cpp
@@ -84,7 +84,7 @@ Classic::Classic() : Extension1stParty("Classic", _trans("Classic Controller"))
   }
 
   // sticks
-  constexpr auto gate_radius = ControlState(STICK_GATE_RADIUS) / LEFT_STICK_RADIUS;
+  constexpr auto gate_radius = ControlState(STICK_GATE_RADIUS) / CAL_STICK_RANGE;
   groups.emplace_back(m_left_stick =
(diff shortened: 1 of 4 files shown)


Source/Core/Core/HW/WiimoteEmu/Extension/Classic.cpp
```diff
@@ -177,26 +173,27 @@ void Classic::Reset()
...
+  // All values are to 8 bits of precision.
   m_reg.calibration = {{
       // Left Stick X max,min,center:
-      CAL_STICK_CENTER + CAL_STICK_RANGE,
-      CAL_STICK_CENTER - CAL_STICK_RANGE,
+      CAL_STICK_CENTER + STICK_GATE_RADIUS,
+      CAL_STICK_CENTER - STICK_GATE_RADIUS,
       CAL_STICK_CENTER,
       // Left Stick Y max,min,center:
       CAL_STICK_CENTER + CAL_STICK_RANGE,
```
mihai-vlad: I believe this should be:
```suggestion
       CAL_STICK_CENTER + STICK_GATE_RADIUS,
```

jordan-woyak: Whoops. Nice catch. Thanks.

PR #8553, opened 2020-01-06
Document the evdev "interesting" heuristic

phire:
Was checking over this old code, and saw a comment calling me out for a lack of documentation.

It might be half a decade late, but better late then never.


i30817:
Unfortunate these libraries don't have a tag mechanism. What's even the point of knowing how many axis something has if you can't tell it's a classic controller or a joystick or a mousewheel?

phire:
Not really an issue for dolphin, as dolphin doesn't automatically configure a controller. `interesting` just means it shows up in the list of potential controllers a user might want to set up. 

The alternative is having only a narrow number of controllers white-listed by layout, like xinput does on windows. "Does it look like an xbox350 controller? No, then we don't support it". By doing it this way, we get to support any controller some random manufacture might put out in the future.

PR #8554, opened 2020-01-09
Add an option to present duplicate frames

stenzek:
Currently, we do not display every second frame in 25fps/30fps games which run to vsync. This improves performance as there's less rendering for the GPU to perform, but when combined with vsync, could cause frame pacing issues.

This commit adds an option to force every frame generated by the console to be displayed to the host, which may improve pacing for these games.

### DIFF
--- a/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/ui/SettingsFragmentPresenter.java
+++ b/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/ui/SettingsFragmentPresenter.java
@@ -585,6 +585,7 @@ public final class SettingsFragmentPresenter
     Setting gpuTextureDecoding = gfxSection.getSetting(SettingsFile.KEY_GPU_TEXTURE_DECODING);
     Setting xfbToTexture = hacksSection.getSetting(SettingsFile.KEY_XFB_TEXTURE);
     Setting immediateXfb = hacksSection.getSetting(SettingsFile.KEY_IMMEDIATE_XFB);
+    Setting skipDuplicateXfbs = hacksSection.getSetting(SettingsFile.KEY_SKIP_DUPLICATE_XFBS);
     Setting fastDepth = gfxSection.getSetting(SettingsFile.KEY_FAST_DEPTH);
 
     sl.add(new HeaderSetting(null, null, R.string.embedded_frame_buffer, 0));
@@ -613,6 +614,9 @@ public final class SettingsFragmentPresenter
             R.string.xfb_copy_method, R.string.xfb_copy_method_description, true, xfbToTexture));
     sl.add(new CheckBoxSetting(SettingsFile.KEY_IMMEDIATE_XFB, Settings.SECTION_GFX_HACKS,
             R.string.immediate_xfb, R.string.immediate_xfb_description, false, immediateXfb));
+    sl.add(new CheckBoxSetting(SettingsFile.KEY_SKIP_DUPLICATE_XFBS, Settings.SECTION_GFX_HACKS,
+            R.string.skip_duplicate_xfbs, R.string.skip_duplicate_xfbs_description, true,
+            skipDuplicateXfbs));
 
     sl.add(new HeaderSetting(null, null, R.string.other, 0));
     sl.add(new CheckBoxSetting(SettingsFile.KEY_FAST_DEPTH, Settings.SECTION_GFX_SETTINGS,
--- a/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/utils/SettingsFile.java
+++ b/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/utils/SettingsFile.java
@@ -89,6 +89,7 @@ public final class SettingsFile
   public static final String KEY_GPU_TEXTURE_DECODING = "EnableGPUTextureDecoding";
   public static final String KEY_XFB_TEXTURE = "XFBToTextureEnable";
   public static final String KEY_IMMEDIATE_XFB = "ImmediateXFBEnable";
+  public static final String KEY_SKIP_DUPLICATE_XFBS = "SkipDuplicateXFBs";
   public static final String KEY_FAST_DEPTH = "FastDepthCalc";
   public static final String KEY_ASPECT_RATIO = "AspectRatio";
   public static final String KEY_SHADER_COMPILATION_MODE = "ShaderCompilationMode";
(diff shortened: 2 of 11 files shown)


JosJuice:
Normally the settings in the hacks tab make Dolphin go faster when you enable them, not when you disable them. Would it make sense to invert how the setting works, or is there a particular reason why it works the way it currently does?

JMC47:
Does this really belong as a hack?  Is it a significant performance difference.

stenzek:
> Normally the settings in the hacks tab make Dolphin go faster when you enable them, not when you disable them. Would it make sense to invert how the setting works, or is there a particular reason why it works the way it currently does?

That's a fair point. I'll invert it :)

> Does this really belong as a hack? Is it a significant performance difference.

I doubt it, except on very low-end GPUs.

stenzek:
I guess that raises another question - do we want to enable this option by default? It really only makes sense when vsync is on, but will improve pacing when it is, for a very small (if not difficult to measure on desktops) performance cost.

iwubcode:
Personally, I think we should have that option on by default, since the previous behavior was to skip duplicate frames.

Also, I'm a little surprised this helps performance.  AFAIK this was one of the reasons originally for why a user was seeing micro stuttering (see https://forums.dolphin-emu.org/Thread-30fps-micro-stuttering-frame-pacing-problems-investigation-ubershaders-at-last).  This was fixed with the hybrid xfb revamp.

JMC47:
It probably relates to the hack modes.  I have users that swear by this feature though.  I don't particularly see a huge difference, but the ones affected appear to be AMD graphics users?

stenzek:
Keeping the option on (i.e. skipping duplicate frames, or maintaining current behavior) by default probably makes sense anyway, since it's only really going to have an effect with vsync on *and* immediate XFB off. And we ship vsync off and immediate XFB on by default.

JMC47:
I don't feel like there's enough of an explanation as to when to use an option like this and I don't think it belongs in hacks really.  Maybe just put it next to Vsync?

stenzek:
Yeah, it's in hacks, but IMO it also makes sense to group it with the other XFB settings. Otherwise there'd be a single random XFB-related setting in another page :/

JMC47:
Could we only make it possible to be enabled when "Skip Presenting" is on?

JosJuice:
@stenzek Immediate XFB is off by default, not on.

stenzek:
Oh, TIL... thanks. I thought we had it on by default. Hmm, in that case perhaps it makes more sense to present all frames (-> disable the option) by default since the performance impact should be near zero at worst.

It also has another upside - the UI will render even during static loading screens. Which probably only affects netplay chat for now, but still.

stenzek:
I linked the new option to the immediate XFB setting - the checkbox will be disabled when immediate XFB is on.

JMC47:
This sounds good to me at this point.

JMC47:
@dolphin-emu-bot rebuild

BParks21:
I just wanna say thanks for adding this feature to dolphin! I've been using the emulator since 2014 and the only thing that prevented me from playing games like Luigi's Mansion, Mario Sunshine, Twilight Princess and Skyward Sword was the 30 fps frame pacing stutter. I'm so happy that enabling duplicate frames finally resolves the micro stutters due to poor frame pacing at 60Hz. Is this suppose to have any negative impact on native 60 fps games? I noticed Donkey Kong Country Returns hangs at startup when this is enabled. Followed by a lot of audio crackle and pop. As soon as i disable duplicate frames the game proceeds as normal.

stenzek:
@BParks21 not if the game runs at 60fps without any slowdown. The option will essentially have no effect then as every presented frame will be unique, so no performance impact on the host nor downsides. The DKC crash seems strange, I'll have to investigate.

For anyone else who is experimenting with this option to improve frame pacing, *please remember to enable vsync*. It will likely have less of an effect without it.

JMC47:
I can't reproduce your issue with Donkey Kong Country Returns.

BParks21:
Thanks for responding. Yes this option is having some bizarre negative impact on this particular game.

BParks21:
@JMC47 Ok I will test again and report back.

BParks21:
@JMC47 Yeah i can produce it every time on my end. I have vsync enabled and skip presenting duplicate frames unchecked.

stenzek:
Does disabling dual core make any difference? It's possible the option will mess with CPU timing in DC mode.

BParks21:
I'll test it.

BParks21:
@JMC47 Thats it! Dual core in combination with duplicate frames enabled causes the issue.

RinMaru:
> @JMC47 Thats it! Dual core in combination with duplicate frames enabled causes the issue.

Dual Core should be disabled by default its really not needed except on maybe fewer then 15 games of the GC and Wii Library unless your CPU is very slow

OCRBonk:
> 
> 
> I just wanna say thanks for adding this feature to dolphin! I've been using the emulator since 2014 and the only thing that prevented me from playing games like Luigi's Mansion, Mario Sunshine, Twilight Princess and Skyward Sword was the 30 fps frame pacing stutter. I'm so happy that enabling duplicate frames finally resolves the micro stutters due to poor frame pacing at 60Hz. Is this suppose to have any negative impact on native 60 fps games? I noticed Donkey Kong Country Returns hangs at startup when this is enabled. Followed by a lot of audio crackle and pop. As soon as i disable duplicate frames the game proceeds as normal.

FWIW if you were already running in Exclusive Full Screen and you have an Nvidia GPU you have had the option to force Dolphin to use 1/2 Refresh Rate Vsync when playing 30hz games. (provided they don't mix framerates and use 60FPS for any segment of gameplay or menus. Otherwise you'd probably get slowdown). This normally fixes frame pacing issues with 30FPS games on a 60hz display. (Make sure to set pre-rendered frames to 1 when doing so too. Otherwise there is a bit of a large input latency increase.)

stenzek:
> FWIW if you were already running in Exclusive Full Screen and you have an Nvidia GPU you have had the option to force Dolphin to use 1/2 Refresh Rate Vsync when playing 30hz games. (provided they don't mix framerates and use 60FPS for any segment of gameplay or menus. Otherwise you'd probably get slowdown). This normally fixes frame pacing issues with 30FPS games on a 60hz display. (Make sure to set pre-rendered frames to 1 when doing so too. Otherwise there is a bit of a large input latency increase.)

If you force it via the GPU control panel, I guess. But it's better for us to manage vsync ourselves, either by using `Present(2)`, or presenting every frame as it comes in like this PR does. The latter allows us to draw on-screen elements at the full refresh rate, e.g. imgui.

BParks21:
@OCRBonk I'm aware of that solution and am not a fan because a lot of games do have mix framerates and is problematic.

RinMaru:
this is causing some games to lock up unless you use frame advance. Vexx being one of them also Metroid Prime

PR #8557, opened 2020-01-13
Add a hack to avoid Error #001 when using WBFS

JosJuice:
Regression from PR #8394, which made the behavior more accurate. We need some sort of hack or heuristic for this since WBFS files don't store the original disc size.


JosJuice:
The hack now reads a value from the disc instead of reading a value from emulated RAM. This should make it possible for CleanRip to rip dual-layer WBFS disc images correctly and shouldn't affect licensed titles.

JosJuice:
Closing in favor of PR #8561. If anyone thinks this would be a better solution than that PR, post a comment and I can rebase and reopen this PR.

PR #8558, opened 2020-01-13
DiscIO: Add out of bounds checks for blob reading

JosJuice:
Fixes the same problem as PR #8557 but for other formats. (No hacks this time!)


JosJuice:
Now that PR #8561 has been merged, this isn't urgently needed anymore. However, I still think it would be nice to have proper bounds checking for all blob types, so I'm keeping this PR open.

PR #8559, opened 2020-01-14
DolphinQt: Properly hide Wii remote extension motion tabs when no extension is selected.

jordan-woyak:
When opening the dialog with extension set to "None" the tabs were incorrectly shown.
The tabs should only be shown when "Nunchuk" is selected (which triggers an event that unhides the tabs).


BlueSupernova22:
I encountered an error 001 Unauthorized device error using this version

PR #8560, opened 2020-01-14
DI MIOS crash fixes

Pokechu22:
This should the problem with launching Gamecube games from the Wii menu (through MIOS) that @Miksel12 mentioned on #8394.

My mistake was that `IOS::HLE::GetIOS()` can be empty, and is when launching with MIOS.  This was obscured by the existing checks, which assumed that `IOS::HLE::GetIOS()->GetDeviceByName("/dev/di")` could fail, but only with `GetDeviceByName` returning nothing.  As noted in the (existing!) comment about that behavior, it is hit from `UpdateRunningGameMetadata`, which MIOS calls:

https://github.com/dolphin-emu/dolphin/blob/9596fe75f12b9199057e357dc6ae78f150b1981f/Source/Core/Core/IOS/MIOS.cpp#L85

What tricked me was that `UpdateRunningGameMetadata` is not hit when launching Gamecube games directly (which might be a bug, perhaps explaining [issue 11503](https://bugs.dolphin-emu.org/issues/11503) but I'm not sure), so I never ran into this in my testing.  I do think I tested MIOS at some point, but it must have been before the partition changes that added the `IOS::HLE::GetIOS()` call.

-----

The second change, which is much more hacky, removes the call to `DVDInterface::Reset()` from MIOS.  Resetting it resets the DTK configuration, which causes games that try to use DTK to fail.  Removing the reset completely seems like it's probably incorrect (perhaps the way MIOS resets DI only resets some things but not DTK, but I'm not sure what that actually would look like), but it's better to have the games work.  (Furthermore, the normal gamecube IPL doesn't reset DI after it's finished setting things up to my understanding, so it seems odd for MIOS to do so).

### DIFF
--- a/Source/Core/Core/IOS/DI/DI.cpp
+++ b/Source/Core/Core/IOS/DI/DI.cpp
@@ -526,6 +526,16 @@ std::optional<DI::DIResult> DI::StartImmediateTransfer(const IOCtlRequest& reque
   return {};
 }
 
+static std::shared_ptr<DI> GetDevice()
+{
+  auto ios = IOS::HLE::GetIOS();
+  if (!ios)
+    return nullptr;
+  auto di = ios->GetDeviceByName("/dev/di");
+  // di may be empty, but static_pointer_cast returns empty in that case
+  return std::static_pointer_cast<DI>(di);
+}
+
 void DI::InterruptFromDVDInterface(DVDInterface::DIInterruptType interrupt_type)
 {
   DIResult result;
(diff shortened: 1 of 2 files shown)


JosJuice:
> What tricked me was that `UpdateRunningGameMetadata` is not hit when launching Gamecube games directly (which might be a bug, perhaps explaining issue 11503 but I'm not sure)

Assuming you mean when using the GC IPL, this is more of a missing feature than a bug. The problem is that we don't have a reliable way of detecting game launching on the GameCube – it's all handled on the PPC, without any equivalent of ES_Launch. (Though maybe we could just do something like treating a game as running as soon as you insert its disc when Dolphin is running in GameCube mode...)

If you mean when launching from the game list, that's because we don't need the UpdateRunningGameMetadata logic and just call SetRunningGameMetadata directly (from SetPathsAndGameMetadata).

Either way, I don't think this is related to issue 11503. Setting the game metadata shouldn't affect the emulated software other than loading/unloading game-specific settings and cheats.

PR #8561, opened 2020-01-14
Fix Error #001 (alternative solution)

JosJuice:
This is an alternative to PR #8557 and PR #8558. The way this PR solves the problem is essentially the same as what we had before PR #8394 (except the code we had back then only worked because it was broken).


JosJuice:
It doesn't really interact. Maybe I should clarify the comment I added to the code in this PR – what I meant by "non-decrypting read" in the comment is actually "non-partition read" (i.e. we're reading from PARTITION_NONE).

JosJuice:
Since this is a rather serious problem (most WBFS files are unplayable), I will be merging this PR in 24 hours unless someone objects (or unless someone merges it before then). I'm doing it with this PR rather than PR #8557 since this PR is the closest to the pre-#8394 behavior.

Tilka:
Not that I understand all the implications but lgtm.

PR #8564, opened 2020-01-15
Use retail console type for GC, instead of devkit

Pokechu22:
This removes some old hack where Gamecube games are marked as running on a devkit; [YAGCD §4.2.1.1.2](http://hitmen.c02.at/files/yagcd/yagcd/chap4.html#sec4.2.1.1.2) explains the different values for 0x8000002C.  Note that Wii titles already use the production version (except when actually running development titles):

https://github.com/dolphin-emu/dolphin/blob/1cc7ef356b9a33814efb290368150a1cceb0a3ed/Source/Core/Core/Boot/Boot_BS2Emu.cpp#L330-L331

I don't know too much of the context of the old hack, but blaming does give [some very old comments](https://github.com/dolphin-emu/dolphin/blame/feea7ebed50313e5986a36b1de8baa03158fff7d/Source/Core/Core/Src/Boot/Boot_BS2Emu.cpp#L75-L80) that claim it is related to DSP timing.  Since that's improved a lot in the past decade, I doubt this hack is no longer needed.

I have not tested this myself, as I do not own Ikaruga.

### DIFF
--- a/Source/Core/Core/Boot/Boot_BS2Emu.cpp
+++ b/Source/Core/Core/Boot/Boot_BS2Emu.cpp
@@ -174,10 +174,8 @@ void CBoot::SetupGCMemory()
   // Physical Memory Size (24MB on retail)
   PowerPC::HostWrite_U32(Memory::GetRamSizeReal(), 0x80000028);
 
-  // Console type - DevKit  (retail ID == 0x00000003) see YAGCD 4.2.1.1.2
-  // TODO: determine why some games fail when using a retail ID.
-  // (Seem to take different EXI paths, see Ikaruga for example)
-  const u32 console_type = static_cast<u32>(Core::ConsoleType::LatestDevkit);
+  // Console type - Retail (YAGCD 4.2.1.1.2)
+  const u32 console_type = static_cast<u32>(Core::ConsoleType::LatestProductionBoard);
   PowerPC::HostWrite_U32(console_type, 0x8000002C);
 
   // Fake the VI Init of the IPL (YAGCD 4.2.1.4)


jordan-woyak:
FYI, Ikaruga still boots and loads the memory card with the change.

leoetlino:
IIRC (it's been a very *long* while so I could be misremembering), one downside of setting the console type to retail is that some games will no longer print OSReports.

delroth:
Getting OSReport was the main original motivation IIRC. They're super useful and we should make sure this patch does not impact our ability to log them.

BhaaLseN:
For the most part (read as: your average Dolphin end-user) we don't really care too much about them.
For dev purposes, I do agree. But any dev is likely smart enough to just change that single line themselves.

We could opt to leave the DevKit setting as reminder-comment, but you probably hate commented out code as much as I do.

mbc07:
What about setting the console type to retail by default but changing it at runtime to devkit if, for example, the debugging GUI is enabled, or by setting a value to true in Dolphin.ini?

Pokechu22:
I assumed all games had stubbed out `OSReport` functions, which wouldn't do anything without HLE patches anyways (which is the case with the Wii IIRC).  Looking at the actual code (in "Alien Hominid") it seems like that's not the case and it's doing *something*.

That said, I also have seen that game can give different output based on the boot magic word at 0x80000020 (I know the DVD code gives a bit more debug logging about the disc header if it's set to jtag).  That is currently hardcoded at bootrom:

https://github.com/dolphin-emu/dolphin/blob/dd7f9ed5da142f91ba20da38c074033c7342af65/Source/Core/Core/Boot/Boot_BS2Emu.cpp#L171-L172

I did some testing, and it looks like OSReport does actually work directly when the console type is set to devkit.  Patching HLE functions (after first generating symbols from the signature database) previously caused OSReport output to show up twice (once from whatever it's doing internally, and once from the HLE patch).  With this change, there is no direct OSReport output immediately, but it still shows up fine after doing the HLE patch.

If we do want to have it be configurable, it would probably be best to allow setting both the boot type and the console type from the UI (probably as a dropdown with all of the options, not just a checkbox to switch between "retail 3" and "latest devkit HW").

shuffle2:
Yes, last time this came up, it was decided to leave the type indicator as devkit for reasons such as debug logs. Probably the comment about Ikaruga should be removed (the code path is different, but practically it doesn't matter anymore).

IIRC, HLE patching debug logs may also find other debug functions which have been *compiled* out (e.g. `int dbglog(char*,...) { return 0; }`), so it can be considered a superset of normal debug log functions.

There doesn't really seem to be a reason to change the default, unless there is some known compat issue which would be solved. Otherwise it just seems like adding Yet Another Option for no discernible benefit.

shuffle2:
> previously caused OSReport output to show up twice 

This sounds like a bug/regression in the HLE system.

Pokechu22:
Rebasing to fix conflict from #8712...

dreamsyntax:
I'm in favor of having a toggle option in Config->GameCube, but its not a huge deal

delroth:
It has a huge impact for any kind of Dolphin development workflow which involves looking at OSReport (this makes it significantly harder and more flaky -- our HLE patching subsystem is heuristic, and these heuristics break in a ton of situations), and I've yet to see someone point to any improvement caused by this change. This feels like breaking OSReport "just because".

delroth:
> Plus, people wanting to take advantage of debug messages have to enable log in the first place and will very likely work with symbol maps (dolphin generated or not) so it shouldn't be an issue for most.

This is objectively wrong. I extremely rarely go to the extent of using symbol maps with Dolphin. I've been using OSReport to help in debugging game issues since we've changed the console type to devkit years and years ago. I suspect that I'm not alone in that and that most of the developers who debug game issues are in the same situation. You're basically now asking everyone to take multiple extra steps to get basic debug information out of games, some of which have known flakiness issues (our symbol map generation sucks, combined with our HLE subsystem also not being great).

sepalani:
@delroth Feel free to correct me if I'm wrong but:

**From a "user" point of view**, it shouldn't change anything. Maybe the speed benefit from not logging `OSReport` when not stubbed, but that's negligible.

**From a "developer" point of view**, there is no issue to switch to the retail model as there is nothing to loose. I doubt that nowadays people developing for the GC/Wii use `OSReport`. They're also able to produce ELF files and symbols from them are loaded (and HLE hook'd) without issues. 

**From a "reverser" point of view** you're half-correct, though. Most games have debugging functions stubbed and are useless unless you're using HLE hooks. If the game provides ELF or symbol files it's a no-brainer. Otherwise, if you really plan to reverse or troubleshoot issues, you won't be relying on OSReport alone. You'll rely on many types of Dolphin log regarding what you're investigating, plus all the debug the game can offer like signature database (from RSO, MEGA files, etc.) that you'll probably save in a symbol map of some sort. You can also export/import symbol maps on more fitting tools like ghidra or ida for the reversing part.

In sum, you can get very basic debug information by using a debug GC model, sure. Regardless, the amount of debug information you get from that (and only from a small amount of games not stubbing them) is nowhere enough to troubleshoot medium to complex issues IMHO.

On top of that, the Wii mode seems to use the retail model rather than the debug model so it doesn't feel right to keep this behaviour inconsistent between GC and Wii games.

If this PR is really an issue for you, I can write a follow-up one with a dropdown allowing to select the Wii and GC model to be used.

JMC47:
I have to chime in, I didn't realize that this would become an issue with the OSReport and other debug messages.  I monitor those all the time when I am playing games and it has helped me with debugging issues.  I rarely use Dolphin's actual debugger because it's way beyond me.

While I think that it'd be good to be able to emulate a retail device, that does seem like a significant cost for casual testers that look at the logs a lot.

mbc07:
Will leave this suggestion here again

> What about setting the console type to retail by default but changing it at runtime to devkit if, for example, the debugging GUI is enabled?

delroth:
But still, what's the point?

If there was a single tangible benefit to end users with this change, I'd gladly accept the inconvenience of having to patch Dolphin any time I do any kind of debugging (though I'd probably send a patch to add a .ini option pretty quickly).

As far as I can tell, nobody has yet provided evidence that simulating a devkit instead of a retail console causes any issues. Dolphin can run most of the GC library, and you'd imagine we'd have figured out if anything was sensitive to that. At this point, I don't think we know of any example.

This is literally making things more difficult for a decently common use case without improving anything else in return.

Rumi-Larry:
> Will leave this suggestion here again
> 
> > What about setting the console type to retail by default but changing it at runtime to devkit if, for example, the debugging GUI is enabled?

How about the same functionality, but the devkit is the default?

Ebola16 (changes requested):
OSReport is extremely useful for modders with little programming knowledge and interest in learning. Once they learn what OSReports are expected with a particular game, they can get hints for why their code is behaving unexpectedly from those logs. I've been able to assist with troubleshooting some simple issues from those logs without telling the newbie modder to generate a symbol map. This is a niche use case but an important one, so I'm against any change that doesn't give the option to modify the `console_type` value from the UI.

As for what the default value for console type should be, I agree with @delroth in that it should probably be the one that outputs OSReports since changing this value doesn't offer any demonstrable benefits. If a benefit of chaining this value is found, I'm fine with making the default value retail console instead.

leoetlino:
Closing because this will not be merged.

PR #8565, opened 2020-01-15
Always display the horizontal scroll bar when log word wrap is off

Pokechu22:
Otherwise, a line that's too wide for the log widget will cause the horizontal scroll bar to appear, which reduces the vertical height, and causes the most recent line to be off screen.  Since that line is off screen, the log widget no longer scrolls as new lines appear, unless it's manually scrolled to the very bottom again.

### DIFF
--- a/Source/Core/DolphinQt/Config/LogWidget.cpp
+++ b/Source/Core/DolphinQt/Config/LogWidget.cpp
@@ -184,6 +184,8 @@ void LogWidget::LoadSettings()
   m_log_wrap->setChecked(settings.value(QStringLiteral("logging/wraplines")).toBool());
   m_log_text->setLineWrapMode(m_log_wrap->isChecked() ? QPlainTextEdit::WidgetWidth :
                                                         QPlainTextEdit::NoWrap);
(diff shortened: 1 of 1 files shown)


shuffle2:
This seems like behavior all users of Qt would have to deal with.
A few threads suggest manually moving the vertical scroll after each append.
(here's an example) https://www.qtcentre.org/threads/5983-QTextEdit-auto-scroll?s=2a1fa9396c24cb72ebed58177c9d79a7&p=31003#post31003

Pokechu22:
It already does autoscroll (as far as I can tell, without any manual intervention on Dolphin's behalf; perhaps Qt changed since that thread was created in 2007).  This PR just fixes a case where the autoscroll can break.

shuffle2:
yes, but i think manually scrolling to the end after lines are pushed would be a better fix for this issue. worth trying at least?

Pokechu22:
There is a difference &mdash; the current automatic scrolling behavior Qt gives allows following the log, but scrolling up to look at older output at which point it'll auto-scrolling until it's scrolled to the end again.  Force-scrolling to the end would break that (and if that is done, I'd want a checkbox to disable it, but that'd mean also disabling the Qt auto-scrolling when it's unchecked which I'm not is even possible).

PR #8569, opened 2020-01-18
InputCommon/evdev: Sort devices by their /dev/ name for consistent ordering.

jordan-woyak:
Fixes: https://bugs.dolphin-emu.org/issues/11917

Unfortunately udev seems to enumerate devices in a seemingly random order.
This is especially problematic for multi-port controller adapters.
Users expect our device list ordering to match the physical ports on their adapter.

This PR sorts evdev devices based on their /dev/ name.
I cannot find any other more reliable means of sorting.

### DIFF
--- a/Source/Core/InputCommon/ControllerInterface/evdev/evdev.cpp
+++ b/Source/Core/InputCommon/ControllerInterface/evdev/evdev.cpp
@@ -2,10 +2,13 @@
 // Licensed under GPLv2+
 // Refer to the license.txt file included.
 
+#include "InputCommon/ControllerInterface/evdev/evdev.h"
+
 #include <algorithm>
 #include <cstring>
 #include <map>
 #include <memory>
+#include <regex>
 #include <string>
 
 #include <fcntl.h>
@@ -21,7 +24,6 @@
 #include "Common/StringUtil.h"
 #include "Common/Thread.h"
 #include "InputCommon/ControllerInterface/ControllerInterface.h"
-#include "InputCommon/ControllerInterface/evdev/evdev.h"
 
 namespace ciface::evdev
 {
@@ -223,13 +225,17 @@ std::shared_ptr<evdevDevice> FindDeviceWithUniqueID(const char* unique_id)
   return nullptr;
 }
 
-void AddDeviceNode(const char* devnode)
+void AddDeviceNode(std::string devnode)
 {
+  // This node is already in use.
+  if (s_devnode_objects.count(devnode) != 0)
+    return;
+
   // Unfortunately udev gives us no way to filter out the non event device interfaces.
   // So we open it and see if it works with evdev ioctls or not.
 
   // The device file will be read on one of the main threads, so we open in non-blocking mode.
-  const int fd = open(devnode, O_RDWR | O_NONBLOCK);
+  const int fd = open(devnode.c_str(), O_RDWR | O_NONBLOCK);
   if (fd == -1)
   {
     return;
@@ -269,7 +275,7 @@ void AddDeviceNode(const char* devnode)
       g_controller_interface.AddDevice(evdev_device);
   }
 
-  s_devnode_objects.emplace(devnode, std::move(evdev_device));
+  s_devnode_objects.emplace(std::move(devnode), std::move(evdev_device));
 }
 
 static void HotplugThreadFunc()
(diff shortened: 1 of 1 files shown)


Source/Core/InputCommon/ControllerInterface/evdev/evdev.cpp
```diff
@@ -371,6 +378,21 @@ void Init()
...
 void PopulateDevices()
 {
+  // Unfortunately there doesn't seem to be a documented way to
+  // enumerate devices in a deterministic order.
+  // This is especially problematic for 4-port controller adapters.
+  // Sorting device nodes based on their /dev/ name seems to work.
+  std::map<std::string, std::string> devnodes;
+
+  // Pad the first occurance of digits for natural sorting,
+  // so /dev/input/event9 sorts before /dev/input/event10.
```
Warepire: Would it make more sense to sort by /dev/input/by-id/ or /dev/input/by-path/ ? According to what I can gather /dev/input/eventX is by order of discovery.

phire:
> According to what I can gather /dev/input/eventX is by order of discovery

I assume in the case of multi-controller adapters, the discovery order will be linear. 

> I cannot find any other more reliable means of sorting.



Have you tried `/sys/class/inputNN/phys`? You should be able to find the correct sysfs directory with `udev_device_get_syspath()`.  
I don't have any multi-input controller adaptors, but phys seems like it does what you want.

Another advantage is that it's stable by USB port and even USB hub ordering. Should have the same ordering on every single boot.

Source/Core/InputCommon/ControllerInterface/evdev/evdev.cpp
jordan-woyak: Order of discovery is desired. The first attached device should take the first slot and so on. A second device should never pop in before the first. I've updated the comment to make that more clear.

jordan-woyak:
@phire I mentioned in IRC, but I'll post here for everyone else. All 4 ports of my GC Adapter return the same "phys" string so it can't be used. It also wouldn't give us the discovery order.

phire:
This solution makes me uneasy; It's not a stable API, just a side effect of some internal kernel code.

1. A future Linux kernel update could break the initialization order and event number ordering consistency.
2. There are only 31 event device nodes. They will wrap around at some point.

It's really annoying that the phys strings are identical. 

One thing I did susggest on irc was using phys and then appending the event id as a fallback for when the phys strings was identical.

jordan-woyak:
@phire I do agree that relying on some probably unspecified kernel behavior isn't ideal but "phys" isn't want we want anyways. We don't want to sort by some irrelevant physical property. We **do** want discovery order.

The firstly attached PS4 controller should take the first slot and so on. A secondly attached PS4 controller should not jump to the first slot because its USB port or MAC address or something compares less than that of the other controller.

"Wrap around" is a problem but this should still handle the main situation I am trying to address.
If you completely fill, then remove some nodes creating fragmentation; and then attach a 4-port GC adapter, at least those four ports should be relatively sorted.

phire:
It's not an irrelevant physical property, it's the USB port tree node, meaning it stays stable based on what USB host/hub port it's been plugged into.

Referencing the original bug, the other usecase is identical controllers plugged into different USB ports. Enumeration order is less than ideal for that usecase as it's different depending on if the users plugged devices in, or rebooted with all the devices already plugged in.

While phys stays stable in both situations.

> If you completely fill, then remove some nodes creating fragmentation; and then attach a 4-port GC adapter, at least those four ports should be relatively sorted.

Or the first two ports could end up at 30 and 31, and the second two ports could be end up in lower gaps.

jordan-woyak:
phire, The USB port is stable but we do not care which port it's plugged in! We want discovery order for the reason I stated above: A secondly attached controller should not jump to the first slot just because it's in a lower USB port number.

Rebooting will change event nodes no doubt but it's not unreasonable to assign controllers to different slots after an entire reboot.

> Or the first two ports could end up at 30 and 31, and the second two ports could be end up in lower gaps.
That's not the behavior I've observed.

phire:
Oh, you are talking about runtime user plugging order, while the bug report you reference is talking about initialization order during boot and wanting stable controller assignments across boots.

These are different usecases.

jordan-woyak:
The logic affects both situations.

jordan-woyak:
Closing in favor of some different/better stuff I'm working on.

PR #3582, opened 2016-01-30
DVDInterface: Cleanup for setting/getting whether a disc is inside

JosJuice:
[Review on Reviewable](https://reviewable.io/reviews/dolphin-emu/dolphin/3582)

PR #8571, opened 2020-01-20
Fix gamecube games not noticing disc changes

Pokechu22:
Fixes [issue 9019](https://bugs.dolphin-emu.org/issues/9019).

The problems:

* Changing a disc reset the drive state completely (which was something I added in 55a88ba2ed).  This included the DI registers, meaning interrupts would no longer be sent (so the drive errors that were supposed to be sent in that case just weren't making it anywhere).  Now it only resets "internal" properties.  I don't know if this is 100% accurate behavior (and I only changed it for disc changes; other places that reset it were not changed)
* Although that allowed games to detect disc errors if they happened on e.g. request stream status, they still didn't care about errors on the read command (which are far more likely).  This happened to be because games look at `DILENGTH`, and if it's 0 they assume the read was successful even if `DEINT` was fired instead of `TCINT`.  I touched related code in 31105995591 but didn't actually think about the error case.  Note that YAGCD mentions that DILENGTH can indicate the amount of data left over for transfer, but it isn't explicit about that being used in errors.
  
  One possible change here is that it might be possible for commands (especially reads) to fail partway through, in which case DILENGTH would maybe be half-updated instead of the current all-or-nothing situation.  But all-or-nothing seems sufficient for most games.

* There were some other inaccuracies with drive state behavior, based on attempts at reverse-engineering the drive firmware.  There are some old half-archived notes [here](https://web.archive.org/web/20070708050203/http://tmb.elitedvb.net/dvd-game/index.php/StateMachines), which was enough for me to get started (but it's very hard to understand).  These changes required a savestate version bump.
  
  One thing this fixes (when combined with the other changes) is the AudioStream command, which gave the wrong error code in some states, which games wouldn't know what to do with and would then crash.

  A note: at least with Alien Hominid, if a game using DTK is ejected while it is streaming, upon reinsertion it'll start the stream from the beginning again.  I tested and confirmed this behavior on console (specifically my Wii).

### DIFF
--- a/Source/Core/Core/HW/DVD/DVDInterface.cpp
+++ b/Source/Core/Core/HW/DVD/DVDInterface.cpp
@@ -157,12 +157,12 @@ static u32 s_current_length;
 static u64 s_next_start;
 static u32 s_next_length;
 static u32 s_pending_samples;
-static bool s_can_configure_dtk = true;
 static bool s_enable_dtk = false;
 static u8 s_dtk_buffer_length = 0;  // TODO: figure out how this affects the regular buffer
 
 // Disc drive state
-static u32 s_error_code = 0;
+static DriveState s_drive_state;
+static DriveError s_error_code;
 
 // Disc drive timing
 static u64 s_read_buffer_start_time;
@@ -219,10 +219,10 @@ void DoState(PointerWrap& p)
   p.Do(s_next_start);
   p.Do(s_next_length);
   p.Do(s_pending_samples);
-  p.Do(s_can_configure_dtk);
   p.Do(s_enable_dtk);
   p.Do(s_dtk_buffer_length);
 
+  p.Do(s_drive_state);
   p.Do(s_error_code);
 
   p.Do(s_read_buffer_start_time);
@@ -383,31 +383,30 @@ void Reset(bool spinup)
   s_current_start = 0;
   s_current_length = 0;
   s_pending_samples = 0;
-  s_can_configure_dtk = true;
   s_enable_dtk = false;
   s_dtk_buffer_length = 0;
 
   if (!IsDiscInside())
   {
-    // ERROR_COVER is used when the cover is open;
-    // ERROR_NO_DISK_L is used when the cover is closed but there is no disc.
+    // CoverOpened is used when the cover is open;
+    // NoMediumPresent is used when the cover is closed but there is no disc.
     // On the Wii, this can only happen if something other than a DVD is inserted into the disc
     // drive (for instance, an audio CD) and only after it attempts to read it.  Otherwise, it will
     // report the cover as opened.
-    SetLowError(ERROR_COVER);
+    SetDriveState(DriveState::CoverOpened);
   }
   else if (!spinup)
   {
     // Wii hardware tests indicate that this is used when ejecting and inserting a new disc, or
     // performing a reset without spinup.
-    SetLowError(ERROR_CHANGE_DISK);
+    SetDriveState(DriveState::DiscChangeDetected);
   }
   else
   {
-    SetLowError(ERROR_NO_DISKID_L);
+    SetDriveState(DriveState::DiscIdNotRead);
   }
 
-  SetHighError(ERROR_NONE);
+  SetDriveError(DriveError::None);
 
   // The buffer is empty at start
   s_read_buffer_start_offset = 0;
@@ -705,32 +704,32 @@ void ClearInterrupt(DIInterruptType interrupt)
 }
 
 // Checks the drive state to make sure a read-like command can be performed.
-// If false is returned, SetHighError will have been called, and the caller
+// If false is returned, SetDriveError will have been called, and the caller
 // should issue a DEINT interrupt.
 static bool CheckReadPreconditions()
 {
-  if (!IsDiscInside())  // Implies ERROR_COVER or ERROR_NO_DISK
+  if (!IsDiscInside())  // Implies CoverOpened or NoMediumPresent
   {
     ERROR_LOG(DVDINTERFACE, "No disc inside.");
-    SetHighError(ERROR_NO_DISK_H);
+    SetDriveError(DriveError::MediumNotPresent);
     return false;
   }
-  if ((s_error_code & LOW_ERROR_MASK) == ERROR_CHANGE_DISK)
+  if (s_drive_state == DriveState::DiscChangeDetected)
   {
     ERROR_LOG(DVDINTERFACE, "Disc changed (motor stopped).");
-    SetHighError(ERROR_MEDIUM);
+    SetDriveError(DriveError::MediumChanged);
     return false;
   }
-  if ((s_error_code & LOW_ERROR_MASK) == ERROR_MOTOR_STOP_L)
+  if (s_drive_state == DriveState::MotorStopped)
   {
     ERROR_LOG(DVDINTERFACE, "Motor stopped.");
-    SetHighError(ERROR_MOTOR_STOP_H);
+    SetDriveError(DriveError::MotorStopped);
     return false;
   }
-  if ((s_error_code & LOW_ERROR_MASK) == ERROR_NO_DISKID_L)
+  if (s_drive_state == DriveState::DiscIdNotRead)
   {
     ERROR_LOG(DVDINTERFACE, "Disc id not read.");
-    SetHighError(ERROR_NO_DISKID_H);
+    SetDriveError(DriveError::NoDiscID);
     return false;
   }
   return true;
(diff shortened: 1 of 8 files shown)


Source/Core/Core/HW/DVD/DVDInterface.cpp
```diff
@@ -931,36 +940,45 @@ void ExecuteCommand(ReplyType reply_type)
...
     interrupt_type = DIInterruptType::DEINT;
     break;
 
   // Used by both GC and Wii
   case DICommand::RequestError:
-    INFO_LOG(DVDINTERFACE, "Requesting error... (0x%08x)", s_error_code);
-    s_DIIMMBUF = s_error_code;
-    SetHighError(0);
+  {
+    u8 drive_state;
```
lioncash: This should probably be a `u32`, otherwise the assignment to `result` is intermixing values of different signedness (`drive_state << 24` produces an `int`, not an unsigned value).

Source/Core/Core/HW/DVD/DVDInterface.cpp
```diff
@@ -931,36 +940,45 @@ void ExecuteCommand(ReplyType reply_type)
...
+    if (s_drive_state == DriveState::Ready)
+      drive_state = 0;
+    else
+      drive_state = static_cast<u8>(s_drive_state) - 1;
+
+    u32 result = (drive_state << 24) | static_cast<u32>(s_error_code);
```
lioncash:
```suggestion
    const u32 result = (drive_state << 24) | static_cast<u32>(s_error_code);
```

Source/Core/Core/HW/DVD/DVDInterface.cpp
Pokechu22: Changed.  I still have `drive_state = static_cast<u8>(s_drive_state) - 1` though; should that be changed to `u32` too or is `u8` fine?  (the enum itself is u8, and also it's already been checked that s_drive_state is not Ready i.e. 0 so that won't underflow).

lioncash: Changing the cast over to `u32` would likely be better for consistency, though I'm not hard pressed about it.

Pokechu22: According to [cppreference](https://en.cppreference.com/w/cpp/language/static_cast):

> 7) [Scoped enumeration](https://en.cppreference.com/w/cpp/language/enum)(C++11) type can be converted to an integer or floating-point type.
> <table><tr><td>When the target type is <code>bool</code> (possibly cv-qualified), the result is <code>false</code> if the original value is zero and <code>true</code> for all other values. For the remaining integral types, the result is  the value of the enum if it can be represented by the target type and  unspecified otherwise</td><td>(until C++20)</td></tr>
> <tr><td>The result is the same as <a href="https://en.cppreference.com/w/cpp/language/implicit_conversion">implicit conversion</a> from the enum's underlying type to the destination type.</td><td>(since C++20)</td></tr></table>

which I'm reading as the cast to `u8` followed by implicit conversion to `u32` is the same as just casting to `u32`.  Changed.

Pokechu22:
I've added some new fixes for resetting the drive from the emulated software (both on GameCube and Wii).  This allows removal of a hack where the drive would act as if it was spun up when reading the disc ID, even if it shouldn't have been in that state.  It also fixes [bug 12017](https://bugs.dolphin-emu.org/issues/12017) (caused by MaxPlay stopping the motor and then resetting the drive; since the reset wasn't implemented it would error since, well, the motor was stopped).

Note that I **have not** hardware tested the exact behavior of `PI_RESET_CODE`.  However, code definitely expects that clearing bit 4 resets DI.

<details>
<summary>Take a look at <code>DVDLowReset</code> (from Scooby-Doo! Mystery Mayhem).</summary>
<pre lang="C">
void DVDLowReset(void)
{
  uint uVar1;
  uint reset;
  undefined8 uVar3;
  undefined8 uVar4;
  write_volatile_4(DICVR,2);
  reset = read_volatile_4(PI_RESET_CODE);
  write_volatile_4(PI_RESET_CODE,reset & 0xfffffffb | 1);
  uVar3 = __OSGetSystemTime();
  uVar1 = DAT_800000f8 >> 2;
  do {
    uVar4 = __OSGetSystemTime();
  } while (((int)((ulonglong)uVar4 >> 0x20) -
            ((uint)((uint)uVar4 < (uint)uVar3) + (int)((ulonglong)uVar3 >> 0x20)) ^ 0x80000000) <
           (uint)((uint)uVar4 - (uint)uVar3 < (uVar1 / 0x1e848) * 0xc >> 3) + 0x80000000);
  write_volatile_4(PI_RESET_CODE,reset | 5);
  ResetOccurred = 1;
  uVar3 = __OSGetSystemTime();
  LastResetEnd._0_4_ = (int)((ulonglong)uVar3 >> 0x20);
  LastResetEnd._4_4_ = (int)uVar3;
  return;
}
</pre>
</details>

<details>
<summary>Datel's implementation is similar, but they use a busy loop (which Ghidra produces something more readable for).</summary>
<pre lang="C">
void DVDLowReset(void)
{
  uint reset;
  uint cvr;
  int i;
  write_volatile_4(DICVR,2);
  reset = read_volatile_4(PI_RESET_CODE);
  write_volatile_4(PI_RESET_CODE,reset & 0xfffffffb | 1);
  i = 0;
  do {
    i = i + 1;
  } while (i < 0x80000);
  write_volatile_4(PI_RESET_CODE,reset | 5);
  write_volatile_4(DISR,0x2a);
  cvr = read_volatile_4(DICVR);
  write_volatile_4(DICVR,cvr);
  DAT_802b9ae4 = FUN_8006f8fc();
  DAT_802b9ae0 = DAT_802b9ae4 >> 0x1f;
  DAT_80318ed4 = 1;
  return;
}
</pre>
</details>

I've also marked functions local to DVDInterface as `static` (as was originally mentioned in #3582).  Now's as good a time as any to do that.

Source/Core/Core/IOS/MIOS.cpp
```diff
@@ -37,7 +37,7 @@ static void ReinitHardware()
   // HACK However, resetting DI will reset the DTK config, which is set by the system menu
   // (and not by MIOS), causing games that use DTK to break.  Perhaps MIOS doesn't actually
   // reset DI fully, in such a way that the DTK config isn't cleared?
-  // DVDInterface::Reset();
+  // DVDInterface::ResetDrive(true);
```
Pokechu22, line 40: I'm less sure that MIOS is even supposed to reset DI in the first place.  It does make some writes to `PI_RESET_CODE`, but I haven't seen it actually write anything other than 7 (there is code that can do it, but I don't think it's live).  Of course, if it *did* write something else, that would be properly emulated because we actually run (the PPC part of) MIOS.  I don't see any writes to it from the ARM part of MIOS, nor from BC, but I might be missing them.

I also did a very basic hardware test of listening to the disc drive while launching a GameCube game; it didn't sound like it triggered a reset.

Pokechu22:
Rebasing to fix conflict with #8733 (i.e. don't change DILENGTH/DIMAR in the case of `NoReply` and `DTK`, and also only change it on successful transfers)...

Source/Core/Core/HW/DVD/DVDInterface.cpp
```diff
@@ -705,32 +704,32 @@ void ClearInterrupt(DIInterruptType interrupt)
 }
 
 // Checks the drive state to make sure a read-like command can be performed.
-// If false is returned, SetHighError will have been called, and the caller
+// If false is returned, SetDriveError will have been called, and the caller
 // should issue a DEINT interrupt.
 static bool CheckReadPreconditions()
 {
-  if (!IsDiscInside())  // Implies ERROR_COVER or ERROR_NO_DISK
+  if (!IsDiscInside())  // Implies CoverOpened or NoMediumPresent
```
JosJuice, line 710:
```suggestion
  if (!IsDiscInside())  // Implies CoverOpened or MediumNotPresent
```

Source/Core/Core/HW/DVD/DVDInterface.cpp
```diff
@@ -1121,22 +1143,31 @@ void ExecuteCommand(ReplyType reply_type)
     // The link is dead, but you can access the page using the Wayback Machine at archive.org.
 
     // This command can only be used immediately after reading the disc ID, before any other
-    // reads. Too early, and you get ERROR_NO_DISKID.  Too late, and you get ERROR_INV_PERIOD.
-    if (!s_can_configure_dtk)
+    // reads. Too early, and you get NO_DISKID.  Too late, and you get INV_PERIOD.
```
JosJuice, line 1141:
```suggestion
    // reads. Too early, and you get NoDiscID.  Too late, and you get InvalidPeriod.
```

Source/Core/Core/HW/DVD/DVDInterface.cpp
Pokechu22: This one is correct; it refers to `DriveState::NoMediumPresent`, not `DriveError::MediumNotPresent`.

GitToTheHub:
Great work, thank you!

PR #8572, opened 2020-01-20
Update Readme.md to remove mention of Vista.

jordan-woyak:
Qt no longer supports Vista.
https://doc.qt.io/qt-5.12/windows.html

Dolphin fails to run on Vista.
https://bugs.dolphin-emu.org/issues/11961

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -13,7 +13,7 @@ Please read the [FAQ](https://dolphin-emu.org/docs/faq/) before using Dolphin.
 ### Desktop
 
 * OS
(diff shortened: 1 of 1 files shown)


PF94:
Gamerappa from the dolphin issue tracker here, I have also realized that the readme.md hasen't been really changed that much since 5.0's release, other than some minor change in the Android section.

MayImilae:
Do we know when Qt dropped Vista support? I'm curious if we can determine exactly when Dolphin stopped working on Vista. Without, you know, actually installing vista and bisecting it, because bleh.

PR #8577, opened 2020-01-24
Common/Analytics: Replace mutex with shared_mutex and minor cleanups.

jordan-woyak:
It looks like original writer, delroth, was waiting on C++17.

### DIFF
--- a/Source/Core/Common/Analytics.h
+++ b/Source/Core/Common/Analytics.h
@@ -6,7 +6,7 @@
 
 #include <chrono>
 #include <memory>
-#include <mutex>
+#include <shared_mutex>
 #include <string>
 #include <string_view>
 #include <thread>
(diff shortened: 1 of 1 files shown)


Source/Core/Common/Analytics.h
```diff
@@ -59,19 +59,16 @@ class AnalyticsReportBuilder
...
 
   const AnalyticsReportBuilder& operator=(const AnalyticsReportBuilder& other)
   {
     if (this != &other)
     {
-      std::scoped_lock lk{m_lock, other.m_lock};
-      m_report = other.m_report;
+      std::string other_report = other.Get();
+      std::scoped_lock lk{m_lock};
+      m_report = std::move(other_report);
```
Tilka: If you only lock one mutex it probably makes more sense to use lock_guard instead of scoped_lock.

PR #8578, opened 2020-01-24
Common/ChunkFile: Use std::is_trivially_copyable directly.

jordan-woyak:
The `TriviallyCopyable` requirement no longer requires types be non-volatile.
https://en.cppreference.com/w/cpp/named_req/TriviallyCopyable
https://wg21.cmeerw.net/cwg/issue2094


jordan-woyak:
Silly outdated compilers are not ready!

PR #8580, opened 2020-01-24
DolphinQt: Add "File Path" column to Game Grid

AlexApps99:
The path is generated by finding a file path that contains the filename, then removing the path to that path.

Example:
File paths: `/home/foo/games`, `/home/foo/Desktop/wii`
Game path: `/home/foo/games/gcn/ntsc/fun game.gcn`
~~File name listed: `gcn/ntsc/fun game.gcn`~~
File name listed: `games/gcn/ntsc/fun game.gcn`

`dir.lastIndexOf` uses `-2` in the case that the file path ends with a trailing slash.

Fixes https://bugs.dolphin-emu.org/issues/10567

This is my first PR to Dolphin, so a thorough review would be awesome, ~~and let me know if this would be better as its' own column instead of replacing "File name".~~

### DIFF
--- a/Source/Core/DolphinQt/GameList/GameListModel.cpp
+++ b/Source/Core/DolphinQt/GameList/GameListModel.cpp
@@ -5,6 +5,7 @@
 #include "DolphinQt/GameList/GameListModel.h"
 
 #include <QDir>
+#include <QFileInfo>
 #include <QPixmap>
 
 #include "Core/ConfigManager.h"
@@ -144,21 +145,10 @@ QVariant GameListModel::data(const QModelIndex& index, int role) const
   case COL_FILE_PATH:
     if (role == Qt::DisplayRole || role == Qt::InitialSortOrderRole)
     {
-      QString file_path = QDir::cleanPath(QString::fromStdString(game.GetFilePath()));
-      for (QString dir : Settings::Instance().GetPaths())
-      {
-        dir = QDir::cleanPath(dir);
-        if (file_path.startsWith(dir))
-        {
-          int path_index = dir.lastIndexOf(QLatin1Char('/'), -2);
-          if (path_index > -1)
-          {
-            file_path = file_path.mid(path_index + 1);
-            break;
-          }
-        }
-      }
-      return QDir::toNativeSeparators(file_path);
+      QString file_path = QFileInfo(QString::fromStdString(game.GetFilePath())).canonicalPath();
+      if (!file_path.endsWith(QDir::separator()))
+        file_path.append(QDir::separator());
+      return file_path;
     }
     break;
   case COL_SIZE:
--- a/Source/Core/DolphinQt/GameList/GameListModel.cpp
+++ b/Source/Core/DolphinQt/GameList/GameListModel.cpp
@@ -4,6 +4,7 @@
 
 #include "DolphinQt/GameList/GameListModel.h"
 
+#include <QDir>
 #include <QPixmap>
 
 #include "Core/ConfigManager.h"
@@ -140,6 +141,26 @@ QVariant GameListModel::data(const QModelIndex& index, int role) const
     if (role == Qt::DisplayRole || role == Qt::InitialSortOrderRole)
       return QString::fromStdString(game.GetFileName());
     break;
+  case COL_FILE_PATH:
+    if (role == Qt::DisplayRole || role == Qt::InitialSortOrderRole)
+    {
+      QString file_path = QDir::cleanPath(QString::fromStdString(game.GetFilePath()));
+      for (QString dir : Settings::Instance().GetPaths())
+      {
+        dir = QDir::cleanPath(dir);
+        if (file_path.startsWith(dir))
+        {
+          int path_index = dir.lastIndexOf(QLatin1Char('/'), -2);
+          if (path_index > -1)
+          {
+            file_path = file_path.mid(path_index + 1);
+            break;
+          }
+        }
+      }
+      return QDir::toNativeSeparators(file_path);
+    }
+    break;
   case COL_SIZE:
     if (role == Qt::DisplayRole)
     {
@@ -186,6 +207,8 @@ QVariant GameListModel::headerData(int section, Qt::Orientation orientation, int
     return tr("Maker");
   case COL_FILE_NAME:
     return tr("File Name");
+  case COL_FILE_PATH:
+    return tr("File Path");
   case COL_SIZE:
     return tr("Size");
   case COL_TAGS:
--- a/Source/Core/Core/ConfigManager.cpp
+++ b/Source/Core/Core/ConfigManager.cpp
@@ -188,6 +188,7 @@ void SConfig::SaveGameListSettings(IniFile& ini)
   gamelist->Set("ColumnTitle", m_showTitleColumn);
   gamelist->Set("ColumnNotes", m_showMakerColumn);
   gamelist->Set("ColumnFileName", m_showFileNameColumn);
+  gamelist->Set("ColumnFilePath", m_showFilePathColumn);
   gamelist->Set("ColumnID", m_showIDColumn);
   gamelist->Set("ColumnRegion", m_showRegionColumn);
   gamelist->Set("ColumnSize", m_showSizeColumn);
@@ -464,6 +465,7 @@ void SConfig::LoadGameListSettings(IniFile& ini)
   gamelist->Get("ColumnTitle", &m_showTitleColumn, true);
   gamelist->Get("ColumnNotes", &m_showMakerColumn, true);
   gamelist->Get("ColumnFileName", &m_showFileNameColumn, false);
+  gamelist->Get("ColumnFilePath", &m_showFilePathColumn, false);
   gamelist->Get("ColumnID", &m_showIDColumn, false);
   gamelist->Get("ColumnRegion", &m_showRegionColumn, true);
   gamelist->Get("ColumnSize", &m_showSizeColumn, true);
--- a/Source/Core/Core/ConfigManager.h
+++ b/Source/Core/Core/ConfigManager.h
@@ -268,6 +268,7 @@ struct SConfig
   bool m_showTitleColumn;
   bool m_showMakerColumn;
   bool m_showFileNameColumn;
+  bool m_showFilePathColumn;
   bool m_showIDColumn;
   bool m_showRegionColumn;
   bool m_showSizeColumn;
--- a/Source/Core/DolphinQt/GameList/GameList.cpp
+++ b/Source/Core/DolphinQt/GameList/GameList.cpp
@@ -145,6 +145,7 @@ void GameList::MakeListView()
   hor_header->setSectionResizeMode(GameListModel::COL_COUNTRY, QHeaderView::Fixed);
   hor_header->setSectionResizeMode(GameListModel::COL_SIZE, QHeaderView::Fixed);
   hor_header->setSectionResizeMode(GameListModel::COL_FILE_NAME, QHeaderView::Interactive);
+  hor_header->setSectionResizeMode(GameListModel::COL_FILE_PATH, QHeaderView::Interactive);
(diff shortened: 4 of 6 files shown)


MayImilae:
Definitely make this a new column please.

AlexApps99:
Ok, do you have any other feedback?

MayImilae:
Nothing else for the moment. I'll give it a proper try tomorrow and give you additional feedback then, it's 2am here!

CookiePLMonster:
Looks good, but is it hidden by default? I imagine this is one of those columns you'd rather default to hidden and allow the user to show it later.

AlexApps99:
I think it is, but I haven't verified it

CookiePLMonster:
Please verify then. I'd also appreciate a screenshot as I'm curious how does this display.

AlexApps99:
Ok, deleted config from `~/.config/dolphin-emu/` and ran it.
It *is* hidden by default
image

AlexApps99:
Screenshots of it in action
image
image

AlexApps99:
The issue is that the path in Settings uses backslashes and the function to get the path uses forwardslashes.

I'll work on a fix right now

CookiePLMonster:
Hence my remark that you **probably** need to compare those as paths and not strings - it'd most likely save some headache with more edge cases and not just slashes.

AlexApps99:
Thanks for the advice, will do that

AlexApps99:
Is this implementation valid? It removes all symlinks/backslashes so the strings will be correctly formatted.

Otherwise, how would I check if a directory is contained in another? I couldn't find a function for this in `QFileInfo`'s docs

Your feedback is appreciated

CookiePLMonster:
Sorry, I might have been mistaken. Looks like the function you need here is `QDir::cleanPath`.

So IMO you could erase your last commit by undoing everything in it and just filtering paths with this function, then you should be able to compare everything as you did previously.

As for erasing the commit, you could undo the changes, apply `QDir::cleanPath` changes, then amend the commit and rename it, then force push (it's fine since it's your own branch).

AlexApps99:
Done

Source/Core/DolphinQt/GameList/GameListModel.cpp
```diff
@@ -143,12 +144,13 @@ QVariant GameListModel::data(const QModelIndex& index, int role) const
...
-      QString file_path = QString::fromStdString(game.GetFilePath());
-      for (const QString& dir : Settings::Instance().GetPaths())
+      QString file_path = QDir::cleanPath(QString::fromStdString(game.GetFilePath()));
+      for (QString dir : Settings::Instance().GetPaths())
       {
+        dir = QDir::cleanPath(dir);
         if (file_path.startsWith(dir))
         {
-          int path_index = dir.lastIndexOf(QRegExp(QStringLiteral("[\\/]")), -2);
+          int path_index = dir.lastIndexOf(QStringLiteral("/"), -2);
```
CookiePLMonster:
```suggestion
          int path_index = dir.lastIndexOf(QLatin1Char('/'), -2);
```

CookiePLMonster:
@dolphin-emu-bot rebuild

CookiePLMonster:
@dolphin-emu-bot rebuild

AlexApps99:
I gotta go to sleep now (1:30am here), thanks for all of the assistance so far @CookiePLMonster :)

CookiePLMonster:
Unfortunately it still seems to be invalid:

image

I will debug this in a few hours.

mbc07:
Shouldn't it display back slashes instead of forward slashes on Windows?

CookiePLMonster:
I don't think it matters. I'm not convinced you can get Qt to output a "platform natural" path, and that's what you'd need in this case.

AlexApps99:
cleanPath will always output forwardslashes, I could either check the platform and replace all forwardslashes with backslashes, or find a different function.

CookiePLMonster:
I think performing comparisons and slicing on "clean" paths is a good idea - but may want to use "native separators" path for solely for displaying it to the user.

AlexApps99:
Maybe I could put the processed path back into `QFileInfo` or `QDir` and get a platform-native relative path

CookiePLMonster:
No need - `QDir::toNativeSeparators` should do exactly what you need :)

AlexApps99:
Awesome, will add that as soon as I can :+1:

CookiePLMonster:
Excellent :)

image

Note to self: Verify that paths display consistently everywhere, as right now in Game Folders I can see paths with both styles. Ideally, they should be stored in "clean" format (so they are cross platform) but stored in "native" format!

CookiePLMonster:
@dolphin-emu-bot rebuild

CookiePLMonster:
Hmm now that's interesting, did lint bot spot missing brackets or incorrect indentation? Could be both, depending on what you intended this code to do.

AlexApps99:
Yeah, it was missing brackets... Oops

CookiePLMonster:
@dolphin-emu-bot rebuild

CookiePLMonster:
One more nitpick - I see you have edited an example of how it is supposed to work (or it's been there and I am just blind), and it would seem like Game Folder path is not to be included in the shown path at all. Yet, if you see the screens I posted above for me `dolphin-games` still shows in those paths, despite being part of both game folder paths.

Any idea what's up with that?

AlexApps99:
That is a typo, the path shortening behaviour works as I intended it to
If you'd prefer that I remove the base folder path from the columns, that's fine too

AlexApps99:
@CookiePLMonster should I keep the current functionality or change it to reflect the old example?

CookiePLMonster:
> @CookiePLMonster should I keep the current functionality or change it to reflect the old example?

For me the way it works now is good - let's wait for somebody else (preferably those who can merge) to comment on it too.

Source/Core/DolphinQt/GameList/GameListModel.cpp
```diff
@@ -140,6 +141,26 @@ QVariant GameListModel::data(const QModelIndex& index, int role) const
...
+          if (path_index > -1)
+          {
+            file_path = file_path.mid(path_index + 1);
+            break;
+          }
+        }
+      }
+      return QDir::toNativeSeparators(file_path);
+    }
+    break;
```
Tilka: It might not matter nowadays, but keep in mind that this function is called *a lot*, many times for each item. At some point we'll need some caching. I don't know whether we need caching yet. Maybe make a few hundred copies of a .dol file and put them in a deep path. Can you still scroll without maxing your CPU?

AlexApps99: It's a simple iteration over a string replacing backslashes with slashes, it would hardly put a dent in the CPU's performance

AlexApps99: Per game, it's called once per game directory preset until a match is found, and then only once more.

AlexApps99: Maybe the function to store/get preset paths should be platform independent, so it's only called once per directory addition, if performance is a concern

Tilka: I mean the function `data()`.

AlexApps99: Ok

AlexApps99: That's outside of the scope of this PR, considering this is my first, but perhaps I could give caching a shot for another PR?

CookiePLMonster: imo it should be handled separately, yes. I am pretty sure this could be cached by Qt itself by setting column's display role appropriately at creation time.

AlexApps99:
What else needs to be done to complete this PR?

CookiePLMonster:
> What else needs to be done to complete this PR?

I would say it's finished. At this point just sit tight and wait until somebody who
a) is involved with UI
b) has merge rights
reviews it.

Pokechu22:
I've found a slight weird behavior: if one game folder is nested within another, then the way the path is shortened depends on which one comes first in the list.  Listing like this:

F:\Test\Maindir, then F:\Test\Maindir\Subdir

gives this:

Maindir\a.dol, then Maindir\Subdir\b.dol

while listing like this:

F:\Test\Maindir\Subdir, then F:\Test\Maindir

gives this:

Maindir\a.dol, then Subdir\b.dol

This probably doesn't matter too much, though, since I suspect most people with a nested setup will just have search subfolders on.  (Also, the `F:\Test\` part _was_ successfully stripped on Windows, so that at least is fixed.)

AlexApps99:
I did that because of simplicity, if it's a problem then I could try sort the paths by length before comparing so the most nested paths are checked first.

Source/Core/DolphinQt/GameList/GameListModel.cpp
```diff
@@ -140,6 +141,26 @@ QVariant GameListModel::data(const QModelIndex& index, int role) const
     if (role == Qt::DisplayRole || role == Qt::InitialSortOrderRole)
       return QString::fromStdString(game.GetFileName());
     break;
+  case COL_FILE_PATH:
+    if (role == Qt::DisplayRole || role == Qt::InitialSortOrderRole)
+    {
+      QString file_path = QDir::cleanPath(QString::fromStdString(game.GetFilePath()));
```
stenzek: I'm a little concerned about the performance impact of the logic here when used with very large game lists. But, I guess if the column is hidden by default, it shouldn't get executed.

MayImilae:
Gave this a quick check. It's not enabled by default, and it all works as advertised!

However, I'm puzzled by the fact that it only shows one tree above the file name. Doesn't that greatly reduce its usefulness?

Here's an example, an issue that I've run into before that this PR could help with, but doesn't.

filepaths1

These two games appear exactly the same under this PR. However, if we take a peek at the full paths...

filepaths2-2
filepaths2-1

One is on my NAS, the other is local.

This is something that this PR could genuinely help with over just showing file names. I use my NAS for storing my games, but I do a local mirror of any game I'm playing through at the moment, with both paths accessible to Dolphin at all times. File names won't help with this, and neither would anything else in the GUI. Only file paths could help. But because this PR only goes up one level, it can't show that difference. It's no more useful than file names here.

Of course, I could rename the folder of my local mirror folder. But at that point, I could _also_ just rename the local mirrored version of the iso's file name, which would work without this PR at all, so...

IMO, I think this PR should just show the full file path. Yes it'll be long, but it's _useful_ in handling edge cases like the one I just outlined where file names are not sufficient. There are also other cases, like having games on a bulk storage drive and an SSD mirror that would this would also apply to. Just showing one folder up from the file name doesn't offer much.

As for the length, why not just show only the path and not the file name? Why have the file name present in the file path when there is a dedicated column for that? That would greatly cut down on the length of the file path column and uses can add the file name column if they want that information.

Source/Core/DolphinQt/GameList/GameListModel.cpp
AlexApps99: How do you suggest I improve the performance impact?

stenzek: Would probably require caching the string in the game list struct. But again, since it's not enabled by default, I don't think it matters all that much.

AlexApps99:
Those are good points you make, I'll use the full file path instead.

AlexApps99:
@MayImilae I've integrated the changes you have requested
image
Should I include the trailing slash? I'm unsure

CookiePLMonster:
Not to be "that guy", but I guess that if you're to show a full path after all then ` QFileInfo::path()` may be better than tokenizing the path on your own?

CookiePLMonster:
@dolphin-emu-bot rebuild

MayImilae:
@AlexApps99 Much better! And yea, the final slash might be nice for combination things like this.

AlexApps99:
Using `QFileInfo::path()` is a good idea, I'll use that soon

AlexApps99:
It now uses `QFileInfo`, and the trailing slash is included too.
image

PR #8582, opened 2020-01-24
VideoCommon/OSD: Process OSD messages even when they are disabled.

jordan-woyak:
When OSD messages were disabled message handling was completely skipped and `s_messages` filled up indefinitely.
Now timeouts are still processed and only the actual drawing is skipped.

### DIFF
--- a/Source/Core/VideoCommon/OnScreenDisplay.cpp
+++ b/Source/Core/VideoCommon/OnScreenDisplay.cpp
@@ -94,29 +94,31 @@ void AddMessage(std::string message, u32 ms, u32 rgba)
 
 void DrawMessages()
 {
-  if (!SConfig::GetInstance().bOnScreenDisplayMessages)
-    return;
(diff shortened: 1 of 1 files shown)


BhaaLseN:
Is there any particular reason why we don't simply skip _adding_ it to the message list in first place (up in `AddMessage`/`AddTypedMessage`)?
I'm also not too sure on why the timeouts need to be processed (vs. just clearing the collection), since no drawing is being done and as far as I can tell there is no callbacks (or other notifications) that occur when the message disappears.

jordan-woyak:
@BhaaLseN Clearing the collection when OSD messages are disabled would "work" but enabling OSD messages after having been disabled wouldn't show possibly still relevant messages. It's something really important to worry about, but I think processing them as normal and simply not drawing them is the correct behavior.

This way toggling the setting on and off when messages currently exist does the expected thing of hiding and showing them in real time while the setting is toggled.

Tilka:
I find the use of `continue` a bit irritating here. Why not draw first and then erase?

PR #8583, opened 2020-01-25
Readme.md Tweak Command Line section

Craftyawesome:
Adds newer backends to readme + more positive about hle accuracy.

The command list itelf is still outdated, though.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -141,12 +141,13 @@ see where it's stored) if you don't plan to reinstall Dolphin.
 * -A, --audio_emulation=<str> Low level (LLE) or high level (HLE) audio
 
 Available DSP emulation engines are HLE (High Level Emulation) and
-LLE (Low Level Emulation). HLE is fast but often less accurate while LLE is
-slow but close to perfect. Note that LLE has two submodes (Interpreter and
-Recompiler), which cannot be selected from the command line.
+LLE (Low Level Emulation). HLE is faster but less accurate whereas
+LLE is slower but close to perfect. Note that LLE has two submodes (Interpreter and Recompiler) 
+but they cannot be selected from the command line.
 
-Available video backends are "D3D" (only available on Windows) and
-"OGL". There's also "Software Renderer", which uses the CPU for rendering and
+Available video backends are "D3D" and "D3D12" (they are only available on Windows), "OGL", and "Vulkan".
+There's also "Null", which will not render anything, and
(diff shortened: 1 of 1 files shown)


Readme.md
```diff
@@ -141,12 +141,13 @@ see where it's stored) if you don't plan to reinstall Dolphin.
 * -A, --audio_emulation=<str> Low level (LLE) or high level (HLE) audio
 
 Available DSP emulation engines are HLE (High Level Emulation) and
-LLE (Low Level Emulation). HLE is fast but often less accurate while LLE is
+LLE (Low Level Emulation). HLE is fast but can be less accurate while LLE is
```
jordan-woyak: I think "often" is still a better word here.

Craftyawesome: Is HLE incorrect enough to be "often" less accurate than LLE, though?
Maybe something like
```suggestion
LLE (Low Level Emulation). HLE is fast but is less accurate than LLE, which is
```

Readme.md
```diff
@@ -141,12 +141,13 @@ see where it's stored) if you don't plan to reinstall Dolphin.
...
 
 Available DSP emulation engines are HLE (High Level Emulation) and
-LLE (Low Level Emulation). HLE is fast but often less accurate while LLE is
+LLE (Low Level Emulation). HLE is fast but is less accurate than LLE, which is
 slow but close to perfect. Note that LLE has two submodes (Interpreter and
 Recompiler), which cannot be selected from the command line.
 
-Available video backends are "D3D" (only available on Windows) and
-"OGL". There's also "Software Renderer", which uses the CPU for rendering and
+Available video backends are "D3D", "D3D12", (only available on Windows) "OGL", and "Vulkan".
```
leoetlino: One comma is misplaced:
 
```suggestion
Available video backends are "D3D", "D3D12" (only available on Windows), "OGL", and "Vulkan".
```

Rukario:
It'd probably read better with less use of the word "which",

"HLE is faster but less accurate whereas LLE is slower but close to perfect. Note that LLE has two submodes (Interpreter and Recompiler) but they cannot be selected from the command line."

And then

"Available video backends are "D3D" and "D3D12" (they are only available on Windows), "OGL", and "Vulkan"."

It reads better with an "and" instead of comma for the D3D modes.

leoetlino:
That does read better. @Craftyawesome would you mind making those changes?

PR #8584, opened 2020-01-25
VideoCommon: Tweak widescreen heuristic.

jordan-woyak:
Consider a projection "anamorphic" when the projection aspect ratio divided by the viewport aspect ratio is near 1.333... ((16/9) / (4/3))
Previously only 4:3 viewport and 16:9 projection situations were considered anamorphic.
This fixes the widescreen GeckoCode in "Gauntlet: Dark Legacy" which seems to maybe render the game to only an upper portion of the screen.

Keep separate counts for perspective and orthographic projections and compare them separately, apples to apples.
If either type of projection look overly anamorphic, go 16:9.
Otherwise, if orthographic projections were anamorphic and are now normal or perspective projections are normal, go 4:3.
This fixes Animal Crossing's inventory screen where perspective projections are temporarily not used.
It also fixes menus in SSX3 from flashing between 16:9 and 4:3.

Count and use the vertices for each projection type in addition to flush counts.
Also add up data for projections which weren't anamorphic or normal for an accurate total.

Only run the heuristic update on non-duplicate frames.

Consolidated duplicate aspect ratio calculation and cropping logic.

If widescreen hack is enabled with a manually configured aspect ratio the heuristic is now used to adjust from the detected aspect ratio to any target aspect ratio.
(Only in GameCube mode of course)

This fixes:
https://bugs.dolphin-emu.org/issues/11243
https://bugs.dolphin-emu.org/issues/11951
https://bugs.dolphin-emu.org/issues/11960

Some other games I tested that still work:
Splinter Cell
Alien Hominid
Mario Kart Double Dash
Burnout 2
F-Zero GX
Animal Crossing

### DIFF
--- a/Source/Core/VideoCommon/RenderBase.cpp
+++ b/Source/Core/VideoCommon/RenderBase.cpp
@@ -1155,19 +1155,6 @@ void Renderer::UpdateWidescreenHeuristic()
                                           g_ActiveConfig.aspect_mode == AspectMode::AnalogWide))
     return;
 
-  const auto& persp = flush_statistics.perspective;
-  const auto& ortho = flush_statistics.orthographic;
-
-  DEBUG_LOG(VIDEO,
-            "flush stats: persp: 4:3=%ld 16:9=%ld other=%ld | ortho: 4:3=%ld 16:9=%ld other=%ld",
-            persp.normal_flush_count, persp.anamorphic_flush_count, persp.other_flush_count,
-            ortho.normal_flush_count, ortho.anamorphic_flush_count, ortho.other_flush_count);
-
-  DEBUG_LOG(VIDEO,
-            "vertex stats: persp: 4:3=%ld 16:9=%ld other=%ld | ortho: 4:3=%ld 16:9=%ld other=%ld",
-            persp.normal_vertex_count, persp.anamorphic_vertex_count, persp.other_vertex_count,
-            ortho.normal_vertex_count, ortho.anamorphic_vertex_count, ortho.other_vertex_count);
-
   // Modify the threshold based on which aspect ratio we're already using:
   // If the game's in 4:3, it probably won't switch to anamorphic, and vice-versa.
   static constexpr u32 TRANSITION_THRESHOLD = 3;
@@ -1178,6 +1165,9 @@ void Renderer::UpdateWidescreenHeuristic()
   const auto looks_anamorphic = [](auto& counts) {
     return counts.anamorphic_vertex_count > counts.normal_vertex_count * TRANSITION_THRESHOLD;
   };
+
+  const auto& persp = flush_statistics.perspective;
+  const auto& ortho = flush_statistics.orthographic;
 
   const auto ortho_looks_anamorphic = looks_anamorphic(ortho);
 
--- a/Source/Core/VideoCommon/RenderBase.cpp
+++ b/Source/Core/VideoCommon/RenderBase.cpp
@@ -102,7 +102,7 @@ Renderer::Renderer(int backbuffer_width, int backbuffer_height, float backbuffer
   UpdateDrawRectangle();
   CalculateTargetSize();
 
-  m_aspect_wide = SConfig::GetInstance().bWii && Config::Get(Config::SYSCONF_WIDESCREEN);
+  m_is_game_widescreen = SConfig::GetInstance().bWii && Config::Get(Config::SYSCONF_WIDESCREEN);
 }
 
 Renderer::~Renderer() = default;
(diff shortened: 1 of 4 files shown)


slx7R4GDZM:
I'm still getting the same issue with Lost Kingdoms 2 and Animal Crossing.

jordan-woyak:
@slx7R4GDZM Can you please provide fifologs demonstrating the problem?

slx7R4GDZM:
Alright, [here](https://github.com/dolphin-emu/dolphin/files/4111631/FIFO.zip). The LK2 fifo looks completely black on my end no matter the settings. I'm not sure what that's about.
I just noticed while recording those though that enabling immediately present XFB stops the flickering between aspect ratios on both games.

jordan-woyak:
@slx7R4GDZM Thanks. I've fixed this and those fifologs now work correctly.

slx7R4GDZM:
@jordan-woyak LK2 looks good now from what I can tell but Animal Crossing still freaks out. I'm not sure if it'll help but [here](https://github.com/dolphin-emu/dolphin/files/4112725/fifo.zip)'s 2 more fifologs.

jordan-woyak:
@slx7R4GDZM Thanks.

I can't see a problem with "ac1.dff".

In "ac2.diff", when the menu pops up, literally nothing is rendered with a 16:9 perspective. All the 2D elements are drawn with a normal orthographic projection so there isn't any reason to consider it "widescreen".

Logic could be added to ignore orthographic (2D) projections after perspective (3D) projections were observed, but I'm worried this might break a 2D game with occasional 3D elements.

slx7R4GDZM:
>I can't see a problem with "ac1.dff".

Yeah, weird. It's happening in gameplay but not when rewatching the fifologs. Also the intro sequence doesn't flicker when immediately present XFB is enabled but the menus still do.
>In "ac2.diff", when the menu pops up, literally nothing is rendered with a 16:9 perspective. All the 2D elements are drawn with a normal orthographic projection so there isn't any reason to consider it "widescreen".

Huh, interesting. It [only happens sometimes](https://github.com/dolphin-emu/dolphin/files/4112759/ac3.zip) though which is the odd thing. Especially when you do it in the inventory menu.
Maybe the widescreen code for that game is just busted. The codes for the Lost Kingdoms games seem to work perfectly now so that's good.

jordan-woyak:
@slx7R4GDZM Whelp, I fixed your fifologs anyways. Just keep throwing them at me if you keep finding broken things. :P

slx7R4GDZM:
@jordan-woyak Now it works even with the code I previously tried. No more flickering or anything in AC.

Behavior on Lost Kingdoms 2 seems different. On the build of the PR before this one it used to be 4:3 on title menus, world map, videos, etc and only 16:9 on gameplay and loading screens. Now it will start off on 4:3 but once it enters gameplay everything else will stay widescreen and get stretched, even the menus.

jordan-woyak:
@slx7R4GDZM Yeah.. I broke something. :P I'll push a fix in a bit.

slx7R4GDZM:
@jordan-woyak 
>Should be good now.

This new behavior might be a consequence of getting Animal Crossing working properly. Seems like it still likes to stay in 16:9 after the loading screen.

JMC47:
If there's nothing in 4:3, I don't think it'd be right to switch back.  The way it'd work on a T.V. is that it'd just be stuck in 16:9.

iwubcode:
Very nice !  I did some testing on a handful of Wii/Gamecube games and was really happy with the results.  Will give the code a more thorough look later.

Source/Core/VideoCommon/RenderBase.cpp
```diff
@@ -1163,30 +1146,69 @@ void Renderer::EndUIFrame()
...
+
+void Renderer::Swap(u32 xfb_addr, u32 fb_width, u32 fb_stride, u32 fb_height, u64 ticks)
+{
+  if (SConfig::GetInstance().bWii)
+    m_is_game_widescreen = Config::Get(Config::SYSCONF_WIDESCREEN);
+
+  // If widescreen hack is disabled consider configured aspect ratio the game's aspect ratio.
+  // With widescreen hack enabled we only use it as the target aspect ratio,
+  // and use the widescreen heuristic (on GameCube) to adjust to this target.
+  if (!g_ActiveConfig.bWidescreenHack)
```
iwubcode: Does or should the widescreen hack affect Wii games?  Should this be an 'else if' with the above?

jordan-woyak: The widescreen hack can be relevant in Wii games.
It can be used to produce ultra-widescreen, for one.

JosJuice: You can also use it to make 4:3 Wii games be 16:9 – which it seems like this PR breaks, now that I'm looking! We need to set `m_is_game_widescreen = false` when `g_ActiveConfig.aspect_mode == AspectMode::Analog`, even if `Config::Get(Config::SYSCONF_WIDESCREEN) && g_ActiveConfig.bWidescreenHack`. This is the same problem as [issue 11237](https://bugs.dolphin-emu.org/issues/11237) (which is fixed in master).

iwubcode: Thanks @JosJuice .  Can you name an example of a 4:3 Wii game?  It's possible I have one but can't think of one off hand.

JosJuice: N64 VC games are maybe the most well-known ones. When it comes to disc games, I can only think of less popular titles (Anubis II, Tamagotchi: Party On...), but there is a complete list here if you want it: https://github.com/dolphin-emu/dolphin/search?l=INI&q=SuggestedAspectRatio

mbc07:
Right now I remember of WarioWare: Smooth Moves, which is a 4:3 only Wii game...

JosJuice:
Yes, but it's 2D only as far as I can recall, which would make it useless for testing the behavior of the widescreen hack. (Or are the 2D elements actually rendered in a 3D environment?)

mbc07:
The city map (where you select the stages) and various micro games are actually rendered in 3D and/or have 3D models on them AFAIK...

Source/Core/VideoCommon/RenderBase.cpp
iwubcode: Thanks!  Of that list, I also have Monster Lab.  So I can help test with that too.

jordan-woyak: @JosJuice the `suggested_aspect_mode` problem should now be fixed.

Tenome:
Thanks for working on this.

pyromancer5:
m8 plz ur dolphin 5.0-11841 by jordan-woyak — VideoCommon: Tweak widescreen heuristic. missed up YU GI OH FALSEBOUND KINGDOM its unplayable now :( the screen keep bugging up during combat it go wide then bk wide bk ...plz fix it

jordan-woyak:
@pyromancer5 File an issue report: https://bugs.dolphin-emu.org/projects/emulator/issues
Make sure to include a FIFO log!

PR #8586, opened 2020-01-26
Avoid using PDEP and PEXT on AMD Zen

Techjar:
For some unknown reason PDEP and PEXT are ridiculously slow on AMD Zen architecture, which is making DoubleToSingle an extremely costly operation on these CPUs, enough to cause severe slowdown in games. Unsure whether the usage in VertexLoaderX64 had any noticeable impact, but it's also been changed as a precautionary measure.

This should fix https://bugs.dolphin-emu.org/issues/11964

### DIFF
--- a/Source/Core/Common/CPUDetect.h
+++ b/Source/Core/Common/CPUDetect.h
@@ -42,6 +42,9 @@ struct CPUInfo
   bool bAVX2 = false;
   bool bBMI1 = false;
   bool bBMI2 = false;
+  // PDEP and PEXT are ridiculously slow on AMD Zen, so we have this flag to avoid using them there
+  // Zen 2 is also affected by this issue
+  bool bFastBMI2 = false;
   bool bFMA = false;
   bool bFMA4 = false;
   bool bAES = false;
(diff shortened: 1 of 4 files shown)


delroth:
https://uops.info/html-instr/PEXT_R64_R64_R64.html confirms the fact that PEXT is slow on Zen / Zen 2. PDEP is similarly slow, we should make sure we don't emit these if we have a better alternative.

andreas-abel:
> https://uops.info/html-instr/PEXT_R64_R64_R64.html confirms the fact that PEXT is slow on Zen / Zen 2. 

And this is actually just the best case. In the worst case, the latency is more than 289 cycles: https://twitter.com/uops_info/status/1202950247900684290

JMC47:
Hopefully this is the cause of users on AMD machines suddenly saying Dolphin is really slow in partiuclar cases...

MayImilae:
PDEP and PEXT latency issues are fixed on Zen 3. This might be worth another look, if we can target Zen 3 specifically.

Source: https://www.anandtech.com/show/16214/amd-zen-3-ryzen-deep-dive-review-5950x-5900x-5800x-and-5700x-tested/6

shuffle2:
I've been going over the CPUDetect code recently and realized that some dolphin jit code checks `bBMI2` while other code checks `bFastBMI2`. Is that intentional (for some paths on family 17h to still be using BMI2) or would it make sense to remove `bFastBMI2` and have `bBMI2` forced to `false` on family 17h?

PR #8591, opened 2020-01-27
DiscIO: Implement re-encryption of Wii partition data

JosJuice:
This was originally created as part of the WIA PR, but since it also is useful for DirectoryBlob, I'm submitting it as a separate PR. (Not that the accuracy improvement will matter for emulated software other than CleanRip unless you disable `VolumeWii`'s `ReadWiiDecrypted` code path...)

### DIFF
--- a/Source/Core/DiscIO/DirectoryBlob.cpp
+++ b/Source/Core/DiscIO/DirectoryBlob.cpp
@@ -387,12 +387,13 @@ bool DirectoryBlobReader::SupportsReadWiiDecrypted() const
   return m_is_wii;
 }
 
-bool DirectoryBlobReader::ReadWiiDecrypted(u64 offset, u64 size, u8* buffer, u64 partition_offset)
+bool DirectoryBlobReader::ReadWiiDecrypted(u64 offset, u64 size, u8* buffer,
+                                           u64 partition_data_offset)
 {
   if (!m_is_wii)
     return false;
 
-  auto it = m_partitions.find(partition_offset);
+  auto it = m_partitions.find(partition_data_offset);
   if (it == m_partitions.end())
     return false;
 
@@ -514,7 +515,8 @@ void DirectoryBlobReader::SetPartitions(std::vector<PartitionWithType>&& partiti
     SetPartitionHeader(partitions[i].partition, partition_address);
 
     const u64 partition_data_size = partitions[i].partition.GetDataSize();
-    m_partitions.emplace(partition_address, std::move(partitions[i].partition));
+    m_partitions.emplace(partition_address + PARTITION_DATA_OFFSET,
+                         std::move(partitions[i].partition));
     const u64 unaligned_next_partition_address = VolumeWii::EncryptedPartitionOffsetToRawOffset(
         partition_data_size, Partition(partition_address), PARTITION_DATA_OFFSET);
     partition_address = Common::AlignUp(unaligned_next_partition_address, 0x10000ull);
--- a/Source/Core/DiscIO/DirectoryBlob.cpp
+++ b/Source/Core/DiscIO/DirectoryBlob.cpp
@@ -27,8 +27,10 @@
 #include "Common/StringUtil.h"
 #include "Common/Swap.h"
 #include "Core/Boot/DolReader.h"
+#include "Core/IOS/ES/Formats.h"
 #include "DiscIO/Blob.h"
 #include "DiscIO/VolumeWii.h"
+#include "DiscIO/WiiEncryptionCache.h"
 
 namespace DiscIO
 {
@@ -70,6 +72,11 @@ DiscContent::DiscContent(u64 offset, u64 size, const u8* data)
 {
 }
 
+DiscContent::DiscContent(u64 offset, u64 size, DirectoryBlobReader* blob)
+    : m_offset(offset), m_size(size), m_content_source(blob)
+{
+}
+
 DiscContent::DiscContent(u64 offset) : m_offset(offset)
 {
 }
@@ -107,10 +114,20 @@ bool DiscContent::Read(u64* offset, u64* length, u8** buffer) const
       if (!file.Seek(offset_in_content, SEEK_SET) || !file.ReadBytes(*buffer, bytes_to_read))
         return false;
     }
-    else
+    else if (std::holds_alternative<const u8*>(m_content_source))
     {
       const u8* const content_pointer = std::get<const u8*>(m_content_source) + offset_in_content;
       std::copy(content_pointer, content_pointer + bytes_to_read, *buffer);
+    }
+    else
+    {
+      DirectoryBlobReader* blob = std::get<DirectoryBlobReader*>(m_content_source);
+      const u64 decrypted_size = m_size * VolumeWii::BLOCK_DATA_SIZE / VolumeWii::BLOCK_TOTAL_SIZE;
+      if (!blob->EncryptPartitionData(offset_in_content, bytes_to_read, *buffer, m_offset,
+                                      decrypted_size))
+      {
+        return false;
+      }
     }
 
     *length -= bytes_to_read;
@@ -131,6 +148,12 @@ void DiscContentContainer::Add(u64 offset, u64 size, const u8* data)
 {
   if (size != 0)
     m_contents.emplace(offset, size, data);
+}
+
+void DiscContentContainer::Add(u64 offset, u64 size, DirectoryBlobReader* blob)
+{
+  if (size != 0)
+    m_contents.emplace(offset, size, blob);
 }
 
 u64 DiscContentContainer::CheckSizeAndAdd(u64 offset, const std::string& path)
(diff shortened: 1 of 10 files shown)


Source/Core/DiscIO/WiiEncryptionCache.cpp
```diff
@@ -0,0 +1,72 @@
...
+  {
+    const std::array<u8, VolumeWii::GROUP_TOTAL_SIZE>* group =
+        EncryptGroup(offset, partition_data_offset, partition_data_decrypted_size, key);
+
+    if (!group)
+      return false;
+
+    const u64 offset_in_group = offset % VolumeWii::GROUP_TOTAL_SIZE;
+    const u64 bytes_to_read = std::min(VolumeWii::GROUP_TOTAL_SIZE - offset_in_group, size);
+    std::memcpy(out_ptr, group->data() + offset_in_group, bytes_to_read);
```
lioncash: `<cstring>` should be included.

Source/Core/DiscIO/WiiEncryptionCache.h
```diff
@@ -0,0 +1,43 @@
...
+// Licensed under GPLv2+
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <array>
+#include <memory>
+
+#include "Common/CommonTypes.h"
+#include "DiscIO/Blob.h"
```
lioncash: This can likely be replaced with a forward declaration.

Source/Core/DiscIO/VolumeWii.cpp
```diff
@@ -521,4 +525,127 @@ bool VolumeWii::CheckBlockIntegrity(u64 block_index, const Partition& partition)
...
+  for (size_t i = 0; i < BLOCKS_PER_GROUP; ++i)
+  {
+    h0_futures[i] = std::async(std::launch::async,
+                               [&unencrypted_data, &unencrypted_hashes, &read_promises, i]() {
+                                 if (!read_promises[i].get_future().get())
+                                   return false;
+
+                                 for (u32 j = 0; j < 31; ++j)
+                                 {
+                                   mbedtls_sha1_ret(unencrypted_data[i].data() + j * 0x400, 0x400,
```
BhaaLseN: There's a lot of magic going on from here on downwards; are there any useful structs to use with `offsetof` for this, or perhaps constants/`constexpr` values to help document where they come from/what they mean?

Source/Core/DiscIO/VolumeWii.cpp
```diff
@@ -521,4 +525,127 @@ bool VolumeWii::CheckBlockIntegrity(u64 block_index, const Partition& partition)
...
+  std::vector<std::array<u8, BLOCK_HEADER_SIZE>> unencrypted_hashes(BLOCKS_PER_GROUP);
+
+  std::array<std::promise<bool>, BLOCKS_PER_GROUP> read_promises;
+  std::array<std::future<bool>, BLOCKS_PER_GROUP> h0_futures;
+  std::array<std::future<bool>, BLOCKS_PER_GROUP / 8> h1_futures;
+  std::future<bool> h2_future;
+
+  for (size_t i = 0; i < BLOCKS_PER_GROUP; ++i)
+  {
+    h0_futures[i] = std::async(std::launch::async,
```
leoetlino: This could launch BLOCKS_PER_GROUP threads, since std::async does not necessarily use thread pools (as far as I can tell). Since you're doing the same for h1_futures and encryption_futures, this function would launch up to 3 * 0x40 = 192 threads, which is way too much.

JosJuice: Ah, I was hoping it would use some kind of thread pool :( It seems to not be using a too large amount of threads on my computer, but I can't be sure that this behavior isn't Windows-specific...

(For what it's worth, there are only 8 threads for the H1 hashes, and the encryption threads aren't launched until all the other tasks are done, so we wouldn't have more than 72 threads active at once, but that's still a relatively large amount.)

Do you know of a way to do something similar but without using too many threads? I could probably manually assign the tasks to threads if there's no other good option, but I'm not sure if we have a convenient way of knowing how many cores the system has, which would be good to use as a limit for how many threads we launch.

leoetlino: Unfortunately not using a thread pool manually, as I don't think `<async>` has anything like that :/

To get the number of cores, `std::thread::hardware_concurrency()` should suffice.

JosJuice: Ah, I didn't know that there was something like `std::thread::hardware_concurrency()`. I was imagining that we would need to write some code in `Common` that uses platform-dependent ifdefs...

JosJuice: I've made some changes to it. The hashing part still uses as many `std::async` calls as before (well, actually one less – the H2 calculation no longer uses a future), but there are much fewer of them active at once now (because the H0 hash calculations aren't started until the relevant data has been read from the disc). The encryption part uses roughly as many `std::async` calls as there are hardware threads. With this change, Dolphin's total thread count when running verification is now the same when using re-encryption as when not using re-encryption: 26 threads. (I'm not quite sure how that works, but I guess some sort of thread pool is involved on at least Windows after all.) Is this a good enough fix for your concerns?

JosJuice: I have now also reduced the number of `std::async` calls for the hashing part to 64.

leoetlino: Yep, that sounds a lot better. I suppose ideally we would have a thread pool implementation in Common but that's probably out of the scope of this PR.

Source/Core/DiscIO/VolumeWii.cpp
JosJuice: I've added a struct (in a new commit, since the change also affects code that's in master).

Source/Core/DiscIO/DirectoryBlob.cpp
```diff
@@ -439,6 +479,9 @@ void DirectoryBlobReader::SetNonpartitionDiscHeader(const std::vector<u8>& parti
   if (header_bin_bytes_read < 0x61)
     m_disc_header_nonpartition[0x61] = 0;
 
+  m_encrypted = std::all_of(m_disc_header_nonpartition.data() + 0x60,
```
degasus, line 482: This looks like a very complicated way to check if one u32 value equals zero. Is there a struct for this header?

Source/Core/DiscIO/VolumeWii.cpp
```diff
@@ -521,4 +527,143 @@ bool VolumeWii::CheckBlockIntegrity(u64 block_index, const Partition& partition)
...
+
+  // Wait for all the async tasks to finish
+  hash_futures.back().get();
+
+  if (error_occurred)
+    return false;
+
+  const size_t hardware_threads = std::thread::hardware_concurrency();
+  size_t threads = 1;
+  while (hardware_threads > threads && threads < BLOCKS_PER_GROUP)
```
degasus, line 632: hardware_threads does not have to be a power of two, so it is better to check for hardware_threads >= 2*threads

Source/Core/DiscIO/DirectoryBlob.cpp
JosJuice: > This looks like a very complicated way to check if one u32 value equals zero.

Hmm... Would it be safe to use `reinterpret_cast<u32*>` here? I don't know if the start of the vector is guaranteed to be correctly aligned.

> Is there a struct for this header?

No.

> Can you add a small comment why this flag is required?

Well, it's required in order to be handle both unencrypted and encrypted discs, and I'm not how useful a comment that just says something like that would be. The full details are available at http://wiibrew.org/wiki/Wii_Disc.

Source/Core/DiscIO/VolumeWii.cpp
JosJuice: Just to make sure that I'm understanding this correctly, is the effect of your suggestion that we would use 4 threads on a 6-core CPU instead of 8? If so, I don't see why that's better.

degasus: Both is as bad, indeed. I assume those tasks won't take more than 16ms, so 6 threads may finish before the other two have started. Why do you need this power of two constraint at all? Begin and end can be calcuated without a power of twp threads.

degasus: i * BLOCKS_PER_GROUP / threads .. (i + 1) * BLOCKS_PER_GROUP / threads

JosJuice: I suppose I did it because I figured the calculations needed for calculating start and end would be more complicated than that. I'll change it to not require a power of two.

Source/Core/DiscIO/WiiEncryptionCache.cpp
```diff
@@ -0,0 +1,79 @@
...
+  if (!m_cache)
+  {
+    m_cache = std::make_unique<std::array<u8, VolumeWii::GROUP_TOTAL_SIZE>>();
+    ASSERT(m_blob->SupportsReadWiiDecrypted());
+  }
+
+  const u64 group_offset_in_partition =
+      offset / VolumeWii::GROUP_TOTAL_SIZE * VolumeWii::GROUP_DATA_SIZE;
+  const u64 group_offset_on_disc =
+      partition_data_offset + Common::AlignDown(offset, VolumeWii::GROUP_TOTAL_SIZE);
```
degasus, line 39: I'd move this AlignDown to the caller EncryptGroups and create an assert which checks for group alignment. Providing an offset in bytes does not make much sense on reading full blocks.

PR #8594, opened 2020-01-30
IOS/FS: Fix CreateFullPath to not create directories that already exist

leoetlino:
Follow up to #8593.

This fixes CreateFullPath to not create directories when it is known
that they already exist, instead of calling CreateDirectory anyway
and checking if the error is AlreadyExists. (That doesn't work
now that we have an accurate implementation of CreateDirectory
that performs permission checks before checking for existence.)

I'm not sure what I was thinking when I wrote that function.

Also adds some tests for CreateFullPath.

### DIFF
--- a/Source/UnitTests/Core/IOS/FS/FileSystemTest.cpp
+++ b/Source/UnitTests/Core/IOS/FS/FileSystemTest.cpp
@@ -434,3 +434,21 @@ TEST_F(FileSystemTest, ReadDirectoryOrdering)
   ASSERT_EQ(result->size(), file_names.size());
   EXPECT_TRUE(std::equal(result->begin(), result->end(), file_names.rbegin()));
 }
+
+TEST_F(FileSystemTest, CreateFullPath)
+{
+  ASSERT_EQ(m_fs->CreateFullPath(Uid{0}, Gid{0}, "/tmp/a/b/c/d", 0, modes), ResultCode::Success);
+
+  // Parent directories should be created by CreateFullPath.
+  for (const std::string& path : {"/tmp", "/tmp/a", "/tmp/a/b", "/tmp/a/b/c"})
+    EXPECT_TRUE(m_fs->ReadDirectory(Uid{0}, Gid{0}, path).Succeeded());
+
+  // If parent directories already exist, the call should still succeed.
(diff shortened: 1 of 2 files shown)


Source/UnitTests/Core/IOS/FS/FileSystemTest.cpp
```diff
@@ -434,3 +434,21 @@ TEST_F(FileSystemTest, ReadDirectoryOrdering)
   ASSERT_EQ(result->size(), file_names.size());
   EXPECT_TRUE(std::equal(result->begin(), result->end(), file_names.rbegin()));
 }
+
+TEST_F(FileSystemTest, CreateFullPath)
```
iwubcode: Should any of these pathes be deleted after the test is finished?  Otherwise the next time it runs the test won't be operating on a clean slate

iwubcode:
Slightly off-topic but hoping we can eventually move to `std::filesystem`.  In my own experience, I found this custom solution filesystem stuff to be a bit lacking in certain situations.

Source/UnitTests/Core/IOS/FS/FileSystemTest.cpp
leoetlino: Yeah, they should. This is automatically done by the testing framework since we're using a fixture. Each test that is defined with TEST_F uses its own instance of FileSystemTest and (by extension) its own file system root, so the tests are independent.

leoetlino:
> Slightly off-topic but hoping we can eventually move to `std::filesystem`. In my own experience, I found this custom solution filesystem stuff to be a bit lacking in certain situations.

Well, the reason we have an abstraction layer over the Wii file system is that we *need* to keep filesystem metadata accurately (see #8539 for reasons). In the past various parts of Dolphin accessed the filesystem directly and that made it very easy to break things or implement things the wrong way (see all the FS and ES code that's been fixed over the last 3 years), while at the same time making it very difficult to have accurate implementations of certain commands. (How do you return file lists that are sorted in the correct order if the filesystem can be modified without the Wii filesystem code being aware of it -- or even without the actions being recorded?)

Forcing all accesses to go through the Wii FS layer ensures that Dolphin simply *cannot* do anything stupid with the filesystem or anything that would be obviously impossible on a console, since now both emulated software and the rest of Dolphin are mostly using the same code paths. It also allows filesystem stuff to be properly unit tested.

PR #8597, opened 2020-01-31
Hotkeys: Toggle Freelook Hotkey

SirMangler:
Straight to the point. A Hotkey to toggle Freelook.


Source/Core/Core/HotkeyManager.cpp
```diff
@@ -290,7 +291,7 @@ constexpr std::array<HotkeyGroupInfo, NUM_HOTKEY_GROUPS> s_groups_info = {
      {_trans("Controller Profile"), HK_NEXT_WIIMOTE_PROFILE_1, HK_PREV_GAME_WIIMOTE_PROFILE_4},
      {_trans("Graphics Toggles"), HK_TOGGLE_CROP, HK_TOGGLE_TEXTURES},
      {_trans("Internal Resolution"), HK_INCREASE_IR, HK_DECREASE_IR},
-     {_trans("Freelook"), HK_FREELOOK_DECREASE_SPEED, HK_FREELOOK_RESET},
```
MayImilae: You removed the reset hotkey instead of adding a toggle hotkey here.

Techjar: No this is correct, it sets up the hotkey ranges for the tabs. When a new hotkey is added after the end of a range you have to change the end of the range to include the new hotkey.

MayImilae: Oh ok, it certainly looks like a removal here. Consider that revolved then!

SirMangler:
> I guess this idea is fine? I mean, it's not necessary at all, you can just not touch the freelook combination keys, or keep freelook off in the GUI. And if you happen to want to use the shift+WADS key combination for something else _while keeping freelook on_, you can reassign the hotkeys. But, eh, the hotkeys area is a dump of options for advanced users anyway, and if you made this I assume you have a use case for it, so, eh.

The primary reason I made this in the first place is because a user needed it for a fighting game, but frequently needed to disable it for cutscenes (which need to relocate the camera). 

Toggling has a couple of advantages over having to manually re-adjust the camera every time, including: reverting back to your desired angle without having to manually set it again (for example an ideal angle for fighting games), allowing the game to take control over the camera (cutscenes), and possibly more than what immediately comes to mind. As you know, without the keybind you need to pause the game and open the settings window and disable the option, then afterwards do the same to enable the option, anytime the game needs the camera (which can include menus as well as cutscenes), or you want to resume the normal camera view. This gets tedious quickly.

This would also be generally handy for users trying to get decent screenshots but need to switch between regular camera and their desired camera angle for setup. I'm sure you get the idea.

This isn't for everyone, but like you mentioned, that's the nature of a lot of the hotkeys.

PR #8598, opened 2020-02-01
Android: More robust check for whether to sync channels

JosJuice:
This changes channel syncing to happen when the operating system is Android TV rather than when TvMainActivity is launched. (You can run TvMainActivity on a phone by specifying a launch activity manually in Android Studio, which I do sometimes for testing purposes. Without this change, you get an exception when channel syncing runs.)

I have tested this on my phone but not on Android TV.


leoetlino:
Any reason you moved it out of TvMainActivity instead of just adding the "is Android TV" check there? (I'm not familiar with the Android codebase ^^)

JosJuice:
This way, the channels will be synced if you launch the non-TV activity on a TV device, if you would like to do that for some reason.

leoetlino:
Ohh, that makes sense. This looks like a pretty safe change then...

PR #8601, opened 2020-02-04
DolphinQt/Mapping: Change "Dead Zone" color to a transparent black.

jordan-woyak:
The previously used "shadow" color is inconsistent across system color palettes and often looks bad against some of our indicator backgrounds.

Before:
Notice the hideous dark lines on the yellow background. The contrast is so high it draws attention from the much more relevant raw and adjusted input dots.
https://i.imgur.com/aUriQaT.png

After:
Consistent subtle dead zone.
https://i.imgur.com/FFbmO2B.png


jordan-woyak:
Hmm. I forgot we have a dead zone in the shake indicator on top of a themed color. I'll have to deal with that.

PR #8603, opened 2020-02-04
DolphinQt/Mapping: Tweak "Dead Zone" and "Gate" colors.

jordan-woyak:
Changed "dead zone" color to a transparent black or white depending on theme.
Inverted value of "gate" colors when active theme has a dark background.
Tweaked order of "shake" indicator axis colors so red is more often shown.

Indicators now look less horrible with dark themes and the dead zone is more subtle as it should be.

Before:
https://i.imgur.com/aUriQaT.png
https://i.imgur.com/aHCSYCj.png

After:
https://i.imgur.com/cWZu4IS.png
https://i.imgur.com/JR8anBW.png

### DIFF
--- a/Source/Core/DolphinQt/Config/Mapping/MappingIndicator.cpp
+++ b/Source/Core/DolphinQt/Config/Mapping/MappingIndicator.cpp
@@ -30,6 +30,7 @@
 
 namespace
 {
+const QColor STICK_GATE_COLOR = Qt::lightGray;
 const QColor C_STICK_GATE_COLOR = Qt::yellow;
 const QColor CURSOR_TV_COLOR = 0xaed6f1;
 const QColor TILT_GATE_COLOR = 0xa2d9ce;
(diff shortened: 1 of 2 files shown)


Source/Core/DolphinQt/Config/Mapping/MappingIndicator.cpp
```diff
@@ -96,9 +99,10 @@ QColor MappingIndicator::GetAltTextColor() const
   return palette().highlightedText().color();
 }
 
-QColor MappingIndicator::GetGateColor() const
+void MappingIndicator::AdjustGateColor(QColor* color)
```
iwubcode: My only comment (and it's really getting nitpicky) is this seems like a better place for a reference.  Maybe just me but when I see a pointer, I expect `nullptr` to be a viable input.

jordan-woyak: From https://github.com/dolphin-emu/dolphin/blob/master/Contributing.md
"Functions that specifically modify their parameters should have the respective parameter(s) marked as a pointer so that the variables being modified are syntaxically obvious."

iwubcode: Ah sorry, I was not aware of this.  Good to know.

MayImilae:
Looks good to me! Also, here are some macOS screenshots if you are curious! The dark theme is ignored by Dolphin in macOS so, only the normal light theme for this.

Before:
Screen Shot 2020-02-10 at 04 23 44

After:
Screen Shot 2020-02-10 at 04 25 42

PR #8604, opened 2020-02-05
Core/NetPlayServer: Fix session not being added to index on direct connection type

Techjar:
Order of operations issue. I also made `SetupIndex` not hard dependent on `m_dialog` just in case.

### DIFF
--- a/Source/Core/Core/NetPlayServer.cpp
+++ b/Source/Core/Core/NetPlayServer.cpp
@@ -101,8 +101,9 @@ NetPlayServer::~NetPlayServer()
 }
 
 // called from ---GUI--- thread
-NetPlayServer::NetPlayServer(const u16 port, const bool forward_port,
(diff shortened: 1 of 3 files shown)


Source/Core/Core/NetPlayServer.cpp
```diff
@@ -1936,6 +1932,10 @@ u16 NetPlayServer::GetPort() const
 void NetPlayServer::SetNetPlayUI(NetPlayUI* dialog)
 {
   m_dialog = dialog;
+
+  // Has to be here otherwise success/failure message won't appear
```
leoetlino: Is this the only place where the call to SetupIndex can be placed? It somewhat bothers me that this is done from a function that's called SetNetPlayUI...

Techjar: Yeah I don't like it here either, but I can't think of anywhere else that it can be called at the right time.

Techjar:
I came up with a different approach to the problem that's much cleaner.

PR #7178, opened 2018-06-28
Lossless image compression

Masterfireheart:
After a couple of passes I managed to shave off 654,586 bytes from most of the PNG images, and 16,998 bytes from the main installer ICO (I didn't bother compressing the other icons because they were for wxWidgets).

Unlike #3896, I used an actual encoder chain and genetic processing, not just some website, so the savings are _actually_ lossless :)

PR #8607, opened 2020-02-07
Themes: Run optipng on our png files.

jordan-woyak:
Optipng shaved off a total of 5650 bytes.
Better than nothing.


Tilka:
See also #7178.

jordan-woyak:
:P

PR #8609, opened 2020-02-08
DolphinQt: Give hotkeys their own "background input" setting.

jordan-woyak:
Setting is named "Hotkeys Require Window Focus".
Please give opinions on the naming in the UI+INI and whether the checked state should be enabled or disabled.
image
Fixes: https://bugs.dolphin-emu.org/issues/8262

### DIFF
--- a/Source/Core/DolphinQt/HotkeyScheduler.cpp
+++ b/Source/Core/DolphinQt/HotkeyScheduler.cpp
@@ -16,8 +16,10 @@
 #include "Common/Thread.h"
 
 #include "Core/Config/GraphicsSettings.h"
+#include "Core/Config/UISettings.h"
 #include "Core/ConfigManager.h"
 #include "Core/Core.h"
+#include "Core/Host.h"
 #include "Core/HotkeyManager.h"
 #include "Core/IOS/IOS.h"
 #include "Core/IOS/USB/Bluetooth/BTBase.h"
@@ -143,8 +145,8 @@ void HotkeyScheduler::Run()
 
     if (Core::GetState() != Core::State::Stopping)
     {
-      // Obey window focus before checking hotkeys.
-      Core::UpdateInputGate();
+      // Obey window focus (config permitting) before checking hotkeys.
+      Core::UpdateInputGate(Config::Get(Config::MAIN_FOCUSED_HOTKEYS));
 
       HotkeyManagerEmu::GetStatus();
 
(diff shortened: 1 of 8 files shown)


MayImilae:
I think the naming of this is pretty good. It says exactly what it does and it is very clear.

Btw, if this default on or off?

jordan-woyak:
The setting is enabled by default.

Source/Core/DolphinQt/HotkeyScheduler.cpp
```diff
@@ -143,8 +145,9 @@ void HotkeyScheduler::Run()
 
     if (Core::GetState() != Core::State::Stopping)
     {
-      // Obey window focus before checking hotkeys.
-      Core::UpdateInputGate();
+      // Obey window focus (config permitting) before checking hotkeys.
+      ControlReference::SetInputGate(Host_RendererHasFocus() ||
+                                     !Config::Get(Config::MAIN_FOCUSED_HOTKEYS));
```
Techjar: Doesn't this also need `&& !Host_UIBlocksControllerState()` like `UpdateInputGate`, otherwise weird behavior will happen?

jordan-woyak: Host_UIBlocksControllerState is only used when imgui grabs input if I remember correctly. I think hotkeys should still function in that situation.

Techjar: Ah, well I think it still needs to be checked, otherwise hotkeys will activate while typing into an ImGui text input. We don't currently have any of those, but we may eventually.

jordan-woyak: Fixed.

Techjar:
LGTM. Not sure I like the setting in the User Interface pane though, maybe move it to Render Window?

jordan-woyak:
> LGTM. Not sure I like the setting in the User Interface pane though, maybe move it to Render Window?

I don't think "Render Window" makes sense. This setting applies even when games aren't running.

Techjar:
Ah, fair enough I guess.

AdmiralCurtiss:
It seems this only considers the render window when checking for focus, which is very impractical for things like the debugger.

jordan-woyak:
@AdmiralCurtiss The window focus logic was already in place. This PR just exposes a separate "Background Input" setting.

PR #8611, opened 2020-02-08
HW/WiimoteEmu: Allow emulating the "Classic Controller Pro".

jordan-woyak:
The "Pro" version does not have analog triggers.
Apparently some games are aware of this and change their controls.
The mapping indicator changes appropriately when enabled.
image
Setting name good as just "Classic Controller Pro" ?

### DIFF
--- a/Source/Core/Core/HW/WiimoteEmu/EmuSubroutines.cpp
+++ b/Source/Core/Core/HW/WiimoteEmu/EmuSubroutines.cpp
@@ -200,7 +200,8 @@ void Wiimote::HandleExtensionSwap()
     }
   }
 
-  if (GetActiveExtensionNumber() != desired_extension_number)
+  if (GetActiveExtensionNumber() != desired_extension_number ||
+      GetActiveExtension()->IsResetNeeded())
   {
     // A different extension is wanted (either by user or by the M+ logic above)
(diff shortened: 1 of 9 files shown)


mbc07:
I think all controls except for L and R should be changed to disabled state when the user mark the "Classic Controller Pro" checkbox. Perhaps also add a tooltip when hovering the "Classic Controller Pro" saying that the Pro lacks analog triggers. Other than that LGTM...

jordan-woyak:
Now I'm wondering if I was misinformed about any games being aware of the "Pro" controller.
This PR might be unnecessary.

iwubcode:
> Apparently some games are aware of this and change their controls.

Haven't gone through the code yet but which games have this behavior?  Just curious!

jordan-woyak:
@iwubcode Apparently 007 Goldeneye, according to https://strategywiki.org/wiki/GoldenEye_007_(2010)/Controls#Classic_Controller_Style
But my testing is showing otherwise.
Now I'm not sure...

mbc07:
I think being able to emulate a Classic Controller or a Classic Controller Pro is a worthy addition just for the sake of emulating all Wiimote extensions, even if no commercial game actually makes a distinction between them...

jordan-woyak:
I think literally no game is actually is aware of the "controller pro" and UI should be improved if we decided to actually have this option.

PR #8614, opened 2020-02-08
InputCommon/XInput: Expose battery as input and cleanups.

jordan-woyak:
Moved things that didn't need to be in the header.
Replaced c-arrays with `std::array`.
Killed some `memset`/`cmp`s.
Cleaned up some comments.
Marked classes as `final`.

### DIFF
--- a/Source/Core/InputCommon/ControllerInterface/XInput/XInput.cpp
+++ b/Source/Core/InputCommon/ControllerInterface/XInput/XInput.cpp
@@ -10,42 +10,119 @@
 
 namespace ciface::XInput
 {
-class Battery : public Core::Device::Input
+struct ButtonDef
+{
+  const char* name;
+  WORD bitmask;
+};
+static constexpr std::array<ButtonDef, 15> named_buttons{{
+    {"Button A", XINPUT_GAMEPAD_A},
+    {"Button B", XINPUT_GAMEPAD_B},
+    {"Button X", XINPUT_GAMEPAD_X},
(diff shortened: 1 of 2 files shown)


iwubcode:
Just curious, what's the advantage of having battery as an input?  Is this so the user can expose the battery for alternative devices?

jordan-woyak:
@iwubcode The only useful purpose is mapping it to the battery setting of an emulated Wii Remote.
Many games will then give you alerts when your battery is low and such.

mbc07:
This is extremely useful now that PR #8417 got merged...

Source/Core/InputCommon/ControllerInterface/XInput/XInput.cpp
```diff
@@ -10,43 +10,120 @@
...
+  Device* m_parent;
+};
+
+class Battery final : public Core::Device::Input
 {
 public:
   Battery(const ControlState* level) : m_level(*level) {}
   std::string GetName() const override { return "Battery"; }
   ControlState GetState() const override { return m_level; }
+  bool IsDetectable() override { return false; }
```
BhaaLseN: Seems weird that this method isn't `const`, but I guess thats from the base class. Any particular reason `Input::IsDetectable` isn't `const`?

jordan-woyak: It should be. It's my my todo list. :P

PR #8616, opened 2020-02-09
InputCommon: Eliminate some duplicated button threshold logic.

jordan-woyak:
Added `GetState` function to `Control` towards allowing `control_ref` member to be private.

### DIFF
--- a/Source/Core/InputCommon/ControlReference/ControlReference.h
+++ b/Source/Core/InputCommon/ControlReference/ControlReference.h
@@ -4,10 +4,10 @@
 
 #pragma once
 
+#include <cmath>
 #include <memory>
 
 #include "InputCommon/ControlReference/ExpressionParser.h"
-#include "InputCommon/ControlReference/FunctionExpression.h"
 #include "InputCommon/ControllerInterface/Device.h"
 
 // ControlReference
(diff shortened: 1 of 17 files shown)


Source/Core/InputCommon/ControlReference/ControlReference.h
```diff
@@ -52,11 +52,17 @@ class ControlReference
 template <>
 inline bool ControlReference::GetState<bool>()
 {
-  return State() > ciface::ExpressionParser::CONDITION_THRESHOLD;
+  return std::lround(State()) > 0;
```
Tilka: I'd be careful with std::lround(), GCC is the only compiler that optimizes this to a movsx. Clang and MSVC both emit a float conversion and a call into the stdlib.

Tilka: But now I'm wondering why you convert to long.

jordan-woyak: @Tilka I'm just using `std::lround` to round to the nearest of 0 or 1.
That seemed cleaner to me than a 0.5 constant.

Tilka: Checking the value directly seems clearer to me, but I'm not the one who maintains this code^^

jordan-woyak: I'll add a comment so it's clear what's happening.

PR #8617, opened 2020-02-09
InputCommon: Clean up creation of inputs.

jordan-woyak:
Greatly simplifies creation of `Input` objects.
Eliminates `new` and `make_unique` calls all over the place.
`std::move` `std::string` arguments.

### DIFF
--- a/Source/Core/Core/HW/WiimoteEmu/Extension/Classic.cpp
+++ b/Source/Core/Core/HW/WiimoteEmu/Extension/Classic.cpp
@@ -6,6 +6,7 @@
 
 #include <array>
 #include <cassert>
+#include <string_view>
 
 #include "Common/BitUtils.h"
 #include "Common/Common.h"
@@ -37,7 +38,7 @@ constexpr std::array<u16, 9> classic_button_bitmasks{{
     Classic::BUTTON_HOME,
 }};
 
-constexpr std::array<const char*, 9> classic_button_names{{
+constexpr std::array<std::string_view, 9> classic_button_names{{
     "A",
     "B",
     "X",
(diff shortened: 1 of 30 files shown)


Source/Core/Core/HW/WiimoteEmu/WiimoteEmu.cpp
```diff
@@ -780,7 +777,7 @@ bool Wiimote::IsUpright() const
 void Wiimote::SetRumble(bool on)
 {
   const auto lock = GetStateLock();
-  m_motor->control_ref->State(on);
+  m_rumble->controls.front()->control_ref->State(on);
```
BhaaLseN: Since this is a collection, should it potentially enable all of them (vs. assuming there is just one in there)?

jordan-woyak: This particular collection always just has one element. I realize this is somewhat ugly but the few outputs are a bit of a special case compared to the inputs.

Source/Core/Core/HW/WiimoteEmu/Extension/Classic.cpp
```diff
@@ -78,9 +78,8 @@ Classic::Classic() : Extension1stParty("Classic", _trans("Classic Controller"))
   groups.emplace_back(m_buttons = new ControllerEmu::Buttons(_trans("Buttons")));
   for (const char* button_name : classic_button_names)
   {
-    const std::string& ui_name = (button_name == std::string("Home")) ? "HOME" : button_name;
-    m_buttons->controls.emplace_back(
-        new ControllerEmu::Input(ControllerEmu::DoNotTranslate, button_name, ui_name));
+    std::string ui_name = (button_name == std::string("Home")) ? "HOME" : button_name;
```
leoetlino:
```suggestion
    std::string ui_name = (button_name == std::string_view("Home")) ? "HOME" : button_name;
```

leoetlino: Actually, it might be worth making `classic_button_names` an array of string views to avoid having to calculate the string lengths all the time. Doing that also has the benefit of making `button_name == "Home"` just work.

jordan-woyak: Made it an array of string_view.

PR #8618, opened 2020-02-09
DolphinQt/Debugger: Fix crash with floating debugger windows

encounter:
QDockWidget::setFloating ends up calling showEvent in the implementation class.

In BreakpointWidget and CodeWidget, this leads to Update() and a null pointer dereference on widgets that haven't been created yet. Moving the CreateWidgets call before setFloating fixes this. (JITWidget and MemoryWidget already do this.)

The rest of the widgets that call setFloating have also been updated to avoid similar bugs if they're changed in the future.


encounter:
The crash can be reproduced with a Qt.ini containing:

```ini
[breakpointwidget]
floating=true

[codewidget]
floating=true

[debugger]
showbreakpoints=true
showcode=true
showregisters=true

[registerwidget]
floating=true
```

PR #8621, opened 2020-02-09
Fix VolumeVerifier not showing a problem for invalid partitions

JosJuice:


Source/Core/DiscIO/VolumeVerifier.cpp
```diff
@@ -549,10 +549,11 @@ bool VolumeVerifier::CheckPartition(const Partition& partition)
...
     }
 
-    // The loop above ends without breaking for discs that legitimately lack updates.
-    // No such discs have been released to end users. Most such discs are debug signed,
+    // The loop above ends without setting invalid_disc_header for discs that legitimately lack
+    // updates. No such discs have been released to end users. Most such discs are debug signed,
     // but there is apparently at least one that is retail signed, the Movie-Ch Install Disc.
-    return false;
+    if (!invalid_disc_header)
+      return false;
```
Tilka: The logic seems strange here, can you check it's correct?

JosJuice: It's correct. If we have a disc like the Movie-Ch Install Disc which legitimately lacks updates, we do not want to tell the user that "The UPDATE partition does not seem to contain valid data" (because it not containing valid data is normal in this case), but we also can't run the tests further down in this function that depend on the partition having a file system.

PR #8624, opened 2020-02-10
InputCommon: Only update setting expressions when the input gate is enabled.

jordan-woyak:
Previously, battery level inputs would read as 0 when window focus was lost.

### DIFF
--- a/Source/Core/InputCommon/ControllerEmu/Setting/NumericSetting.h
+++ b/Source/Core/InputCommon/ControllerEmu/Setting/NumericSetting.h
@@ -158,10 +158,13 @@ class SettingValue
 public:
   ValueType GetValue() const
   {
-    if (IsSimpleValue())
-      return m_value;
-    else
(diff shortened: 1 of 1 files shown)


Source/Core/InputCommon/ControllerEmu/Setting/NumericSetting.h
```diff
@@ -158,10 +158,13 @@ class SettingValue
 public:
   ValueType GetValue() const
   {
-    if (IsSimpleValue())
-      return m_value;
-    else
-      return m_input.GetState<ValueType>();
+    // Only update dynamic values when the input gate is enabled.
+    // Otherwise settings will all change to 0 when focused is lost.
```
Tilka: "focused" -> "focus"

jordan-woyak: I orignally had focus and that didn't sound right for some reason.. :P

PR #8625, opened 2020-02-10
VolumeVerifier: Handle overlapping blocks more efficiently

JosJuice:
The performance gains of doing this aren't too important since you normally wouldn't run into any disc image that has overlapping blocks (which by extension means overlapping partitions), but this change also lets us get rid of things like VolumeVerifier's mutex that used to exist just for the sake of handling overlapping blocks.

### DIFF
--- a/Source/Core/DiscIO/VolumeVerifier.cpp
+++ b/Source/Core/DiscIO/VolumeVerifier.cpp
@@ -1095,6 +1095,12 @@ void VolumeVerifier::Process()
     m_volume.GetTMD(PARTITION_NONE).GetContent(m_content_index, &content);
     bytes_to_read = Common::AlignUp(content.size, 0x40);
     content_read = true;
+
+    if (m_content_index + 1 < m_content_offsets.size() &&
+        m_content_offsets[m_content_index + 1] < m_progress + bytes_to_read)
+    {
+      excess_bytes = m_progress + bytes_to_read - m_content_offsets[m_content_index + 1];
+    }
   }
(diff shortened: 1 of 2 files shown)


Source/Core/DiscIO/VolumeVerifier.cpp
```diff
@@ -1070,10 +1069,19 @@ void VolumeVerifier::WaitForAsyncOperations() const
 bool VolumeVerifier::ReadChunkAndWaitForAsyncOperations(u64 bytes_to_read)
 {
   std::vector<u8> data(bytes_to_read);
+
+  const u64 bytes_to_copy = std::min(m_excess_bytes, bytes_to_read);
+  if (bytes_to_copy > 0)
+    std::memcpy(data.data(), m_data.data() + m_data.size() - m_excess_bytes, bytes_to_copy);
```
CookiePLMonster: Can destination and source overlap here?

JosJuice: No. The source is the class member `m_data` and the destination is the local variable `data`.

CookiePLMonster: Bliiiiiiiind.

JosJuice: If it makes you feel any better, I originally wrote this commit in August and couldn't be bothered debugging a critical problem with it until today, and it turned out that part of the problem was that I had written `m_data` for the destination instead of `data` :)

lioncash (approved):

PR #8626, opened 2020-02-10
WIP: WiimoteReal: Attempt to fix M+ detection problems.

jordan-woyak:
This is an attempt to fix: https://bugs.dolphin-emu.org/issues/7444

MotionPlus detection sometimes fails in Real Wii Remote mode for the same reason speaker data doesn't work well.  Not enabling "sniff mode" keeps the remote in a 100hz (vs 200hz) input loop.
MotionPlus initialization quickly cycles the extension port from connected->disconnected->connected. At 100hz that signal is sometimes missed.

This hack-fix requests two status reports when it sees a game send the M+ activation signal and overwrites the first response to show the "disconnected" state.

I can almost never reproduce this issue myself but having just dealt with M+ in PR #8575 I can see how a trick like this might be needed.

Here's one report that it works around the problem. Yay.
https://www.reddit.com/r/DolphinEmulator/comments/iwa9h2/how_do_i_get_dolphin_to_recognize_motion_plus/

### DIFF
--- a/Source/Core/Core/HW/WiimoteReal/WiimoteReal.cpp
+++ b/Source/Core/Core/HW/WiimoteReal/WiimoteReal.cpp
@@ -9,6 +9,7 @@
 #include <queue>
 #include <unordered_set>
 
+#include "Common/BitUtils.h"
 #include "Common/ChunkFile.h"
 #include "Common/CommonTypes.h"
 #include "Common/Config/Config.h"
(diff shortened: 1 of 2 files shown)


iwubcode:
Just curious, what issue is occurring with M+ detection?

jordan-woyak:
@iwubcode I'll update the PR description with some more info.

JMC47:
Considering how simple this is, is there any reason not to merge it?

jordan-woyak:
> Considering how simple this is, is there any reason not to merge it?

I was hoping to get some better confirmations that it actually successfully works around the problem.

JMC47:
I used to be able to reproduce the problem, I'll give it a go.

Source/Core/Core/HW/WiimoteReal/WiimoteReal.cpp
```diff
@@ -24,6 +25,7 @@
 #include "Core/HW/Wiimote.h"
 #include "Core/HW/WiimoteCommon/DataReport.h"
 #include "Core/HW/WiimoteCommon/WiimoteHid.h"
+#include "Core/HW/WiimoteEmu/MotionPlus.h"
```
Rumi-Larry, line 28: The only thing is that now this include looks out of place, perhaps a comment would be good.

JosJuice: I don't agree.

PR #8627, opened 2020-02-11
CMake: Fix building ARM64 on Windows

spycrab:
* Adds two new configurations to ``CMakeSettings.json``: ``Release (arm64)`` and ``Debug (arm64)``
* Skips over ``External/ffmpeg`` for ARM64 builds
* Outputs ARM64 binaries in ``Binary/ARM64`` now
* Fixes a bunch of other build and linking errors

cc: @stenzek


CMakeSettings.json
```diff
@@ -29,6 +29,52 @@
...
+      "buildCommandArgs": "",
+      "buildRoot": "${workspaceRoot}\\Build\\${name}",
+      "cmakeCommandArgs": "",
+      "variables": [
+        {
+          "name": "Qt5_DIR",
+          "value": "${workspaceRoot}\\Externals\\Qt\\Qt5.14.1\\msvc2019_arm64\\lib\\cmake\\Qt5"
+        },
+        {
+          "name": "CMAKE_SYSTEM_NAME",
```
stenzek: Does this need to be explicitly set, or can it come from the toolchain?

CMakeSettings.json
```diff
@@ -29,6 +29,52 @@
...
+          "value": "Windows"
+        },
+        {
+          "name": "CMAKE_SYSTEM_PROCESSOR",
```
stenzek: Ditto

CMakeLists.txt
```diff
@@ -200,9 +205,11 @@ elseif(_ARCH_64 AND CMAKE_SYSTEM_PROCESSOR MATCHES "x86_64|amd64|AMD64")
   add_definitions(-D_M_X86_64=1)
   check_and_add_flag(HAVE_SSE2 -msse2)
 elseif(CMAKE_SYSTEM_PROCESSOR STREQUAL "aarch64")
+  if (NOT MSVC)
```
stenzek: The `_M_ARM` could probably be dropped, I can't see any references to it in the source.

CMakeSettings.json
spycrab: Yes it has to be, otherwise the other argument gets ignored as well.

CMakeSettings.json
spycrab: See above.

PR #8628, opened 2020-02-13
VideoCommon: Allow texture folders to be determined by a <gameid>.txt

iwubcode:
Currently, in your "Load" folder, you have a bunch of 3 letter or possibly more folders that reference some game.  Wouldn't it be nice if you instead could have a nice native language name instead?  Instead of "SLS", you could have "The Last Story".

Well now you can!

You only need to put a "SLS.txt" file in your directory and that's enough to tell Dolphin that is a directory of textures for "The Last Story".

This resolves two issues, neither were made on the issue tracker as far as I know:

* A user on the forum made a post in feature requests and they wanted the ability to have textures be read from subfolders.  They referenced a pack that might be stored on github where downloading automatically adds a subfolder.
* Admentus currently has to add multiple symlinks for his Majora's Mask / Ocarina of Time project so that all the textures can be shared.  Now he can just add multiple <gameid>.txt to the folder and all will be picked up on load.

### DIFF
--- a/Source/Core/VideoCommon/HiresTextures.cpp
+++ b/Source/Core/VideoCommon/HiresTextures.cpp
@@ -17,6 +17,7 @@
 
 #include <fmt/format.h>
 
+#include "Common/CommonPaths.h"
 #include "Common/File.h"
 #include "Common/FileSearch.h"
 #include "Common/FileUtil.h"
@@ -86,26 +87,40 @@ void HiresTexture::Update()
   }
 
   const std::string& game_id = SConfig::GetInstance().GetGameID();
-  const std::string texture_directory = GetTextureDirectory(game_id);
+  const std::set<std::string> texture_directories = GetTextureDirectories(game_id);
   const std::vector<std::string> extensions{".png", ".dds"};
 
-  const std::vector<std::string> texture_paths =
-      Common::DoFileSearch({texture_directory}, extensions, /*recursive*/ true);
-
-  const std::string code = game_id + "_";
-
-  for (auto& path : texture_paths)
+  for (const auto& texture_directory : texture_directories)
   {
-    std::string filename;
-    SplitPath(path, nullptr, &filename, nullptr);
+    const auto texture_paths =
+        Common::DoFileSearch({texture_directory}, extensions, /*recursive*/ true);
 
-    if (filename.substr(0, s_format_prefix.length()) == s_format_prefix)
+    bool failed_insert = false;
+    for (auto& path : texture_paths)
     {
-      const size_t arb_index = filename.rfind("_arb");
-      const bool has_arbitrary_mipmaps = arb_index != std::string::npos;
-      if (has_arbitrary_mipmaps)
-        filename.erase(arb_index, 4);
-      s_textureMap[filename] = {path, has_arbitrary_mipmaps};
+      std::string filename;
+      SplitPath(path, nullptr, &filename, nullptr);
+
+      if (filename.substr(0, s_format_prefix.length()) == s_format_prefix)
+      {
+        const size_t arb_index = filename.rfind("_arb");
+        const bool has_arbitrary_mipmaps = arb_index != std::string::npos;
+        if (has_arbitrary_mipmaps)
+          filename.erase(arb_index, 4);
+
+        const auto [it, inserted] =
+            s_textureMap.try_emplace(filename, DiskTexture{path, has_arbitrary_mipmaps});
+        if (!inserted)
+        {
+          failed_insert = true;
+        }
+      }
+    }
+
+    if (failed_insert)
+    {
+      ERROR_LOG(VIDEO, "One or more textures at path '%s' were already inserted",
+                texture_directory.c_str());
     }
   }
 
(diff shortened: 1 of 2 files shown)


Source/Core/VideoCommon/HiresTextures.cpp
```diff
@@ -445,7 +445,52 @@ std::string HiresTexture::GetTextureDirectory(const std::string& game_id)
...
+  {
+    const std::string region_free_directory =
+        File::GetUserPath(D_HIRESTEXTURES_IDX) + game_id.substr(0, 3);
+
+    if (!File::Exists(region_free_directory))
+    {
+      // If there's no region specific directory, look for any directories with
+      // a '<gameid>.txt' in their folder
+      const auto root = File::ScanDirectoryTree(File::GetUserPath(D_HIRESTEXTURES_IDX), false);
+      std::string found_directory = "";
```
lioncash:
```suggestion
      std::string found_directory;
```

Source/Core/VideoCommon/HiresTextures.cpp
```diff
@@ -445,7 +445,52 @@ std::string HiresTexture::GetTextureDirectory(const std::string& game_id)
...
+
+          const auto match_gameid = [game_id](const std::string& filename) {
+            std::string basename;
+            SplitPath(filename, nullptr, &basename, nullptr);
+            return basename == game_id || basename == game_id.substr(0, 3);
+          };
+
+          if (std::any_of(files.begin(), files.end(), match_gameid))
+          {
+            if (found_directory != "")
```
lioncash:
```suggestion
            if (!found_directory.empty())
```

Source/Core/VideoCommon/HiresTextures.cpp
```diff
@@ -445,7 +445,52 @@ std::string HiresTexture::GetTextureDirectory(const std::string& game_id)
...
+                        "Skipping texture path '%s' because one was already found for gameid '%s'.",
+                        entry.physicalName.c_str(), game_id.c_str());
+              continue;
+            }
+            found_directory = entry.physicalName;
+          }
+        }
+      }
+
+      if (found_directory != "")
```
lioncash:
```suggestion
      if (!found_directory.empty())
```

Source/Core/VideoCommon/HiresTextures.cpp
```diff
@@ -445,7 +445,52 @@ std::string HiresTexture::GetTextureDirectory(const std::string& game_id)
...
+    const std::string region_free_directory =
+        File::GetUserPath(D_HIRESTEXTURES_IDX) + game_id.substr(0, 3);
+
+    if (!File::Exists(region_free_directory))
+    {
+      // If there's no region specific directory, look for any directories with
+      // a '<gameid>.txt' in their folder
+      const auto root = File::ScanDirectoryTree(File::GetUserPath(D_HIRESTEXTURES_IDX), false);
+      std::string found_directory = "";
+      for (auto entry : root.children)
```
lioncash:
```suggestion
      for (const auto& entry : root.children)
```

Admentus64:
Works exactly like it says it does.

I noticed the following, bug or intended:

The texture pack does not load if the name of the texture pack folder is set to the first three characters of the GameID, even if a .txt file is added within the folder.

Testing with GameID: PZLE01
WORKS                      "PZLE01" folder with "PZL.txt" within
WORKS                      "Zelda 64 4K" folder with "PZL.txt" within
WORKS                      "Zelda 64 4K" folder with "PZLE01.txt" within
DOES NOT WORK     "PZL" folder with "PZLE01.txt" within
DOES NOT WORK     "PZL" folder with "PZL.txt" within

I can confirm that the .txt file for PZLE01 (NTSC-U) can either be named "PZL.txt" (Global) or "PZLE01.txt" (NTSC-U). Naming it "PZLP01.txt" (PAL) obviously does not load the texture pack.

Basicially you have to give the texture pack folder any name other than the first three characters of the GameID.

I don't know whatever this was intended or not, but it can seem a bit confusing when a texture pack will not load because of the texture pack folder name itself.

Tested with the following GameID's for the Zelda 64 4K Texture Pack (https://forums.dolphin-emu.org/Thread-zelda-64-uhd-v4-1-2020-01-22):

- PZL (Ocarina of Time and Majora's Mask for the GameCube)
- REL (Extracted Collector's Edition .TGC file)
- D43 (Ocarina of Time and Master Quest for the GameCube)
- NAC (Ocarina of Time for the Virtual Console)
- NAR (Majora's Mask for the GameCube)
- NAG (custom patched Dawn & Dusk rom hack as part of the texture pack)
- NGZ (custom patched injected practice rom)

This eliminates the need for symbolic links through Windows. It also cleans up the "Load" folder with unnecessary symbolic link folders.

A nice bonus, texture packs can now be given any name the user or texture pack author would like, which makes it easier to recognize the texture pack at first glance.

My only suggestion for improvement would be the following:

Use a single .txt file, where each new line entry contains the GameID.

Ex. "Load.txt" with the following lines
PZL
REL
D43
NAC
NAR
NAG

Which cleans up some space within the texture pack folder itself (aka, less files). But this is just extremely nitpicking from me. The solution as it stands now is already by far a great improvement and probably doesn't really need that much refining.

Having a similar concept for texture file hashes itself would be great as well (aka, one texture file that can be re-used for multiple different texture hashes), but I realize that's out of the scope for this PR.

iwubcode:
Thank you @Admentus64 for all the testing / feedback.

> The texture pack does not load if the name of the texture pack folder is set to the first three characters of the GameID, even if a .txt file is added within the folder.

This was actually intentional.  I prioritize the original behavior over the '.txt' solution.  Maybe I shouldn't though.  I'd be curious what other devs think?

iwubcode:
Originally, I was worried about multiple texture packs overriding each other.  That's why it now errors if there's more than one pack with a <gameid>.txt.

I thought this scenario would be rather uncommon.  Until I thought about dynamic input textures.  It's possible there will be a pack with <gameid>.txt and then dynamic input texture pack which generates GAMEID folder.  This means that Dolphin would read GAMEID folder and completely ignore the pack.

That's going to make this a bit more complicated.  But because of that, I will mark this WIP for now.

iwubcode:
@Admentus64 - your 'not' working cases should now work.

**All** - I now support loading multiple texture folders, as long as they match the gameid specified.  I log an error if there are any textures that were already handled by another pack.  This change is to make things work nicely with dynamic input textures and resource packs.  In the near future I plan to update the resource packs to properly handle ordering in this scenario.

Admentus64:
@iwubcode 

I can confirm all cases I described before once again and now they all now fully work. Thus the texture pack will also load if the folder name is set to the first three characters of the GameID, with or without a corresponding .txt file.

In addition I tried dropping a PZL.txt into multiple folders while prefetching textures:

- Only PZL folder: 6219 MB
- Two additional texture pack folders (which are not related): 8583 MB

In other words, you can also load textures from multiple texture pack folders. Generally it would be a bad idea for users to start dropping txt files in every folder they have (through honestly, that would be the own fault for doing so), but it actually has a nice feature to it as well, since texture pack creators could essentially split their texture pack into multiple folders.

I don't see any issues anymore with this.

iwubcode:
Updated the code at @jordan-woyak 's request.  This removes the "ScanDirectoryTree" call and instead attempts to use some string trickery to get better performance.

Let me know if/when this is ready to merge, I can squash the commits.

Source/Core/VideoCommon/HiresTextures.cpp
```diff
@@ -439,15 +453,47 @@ bool HiresTexture::LoadTexture(Level& level, const std::vector<u8>& buffer)
...
+  };
+
+  // Look for any other directories that might be specific to the given gameid
+  const auto root_directory = File::GetUserPath(D_HIRESTEXTURES_IDX);
+  const auto files = Common::DoFileSearch({root_directory}, {".txt"}, true);
+  for (const auto& file : files)
+  {
+    if (match_gameid(file))
+    {
+      const auto directory_path = file.substr(root_directory.size());
```
leoetlino: Could you please add a comment that explains what this does and why? It isn't immediately obvious what this does (get the top directory, not just the parent directory). Something similar to your explanation on IRC would be fine

iwubcode: Thanks for reviewing @leoetlino .  Please take a look at the comment when you have time and see if you think it's appropriate.

iwubcode:
Thanks @leoetlino .  Squashed the commits

PR #8629, opened 2020-02-14
Common/FileSearch: Portable DoFileSearch std::filesystem implementation.

jordan-woyak:
Instead of the Windows specific `CompareStringOrdinal`, case-insensitive file-extension matching is now performed by a `std::basic_regex` which will use the current locale.
Note that Qt appears to use the current locale to decode filenames by default so this seems sensible: https://doc.qt.io/archives/qt-4.8/qfile.html#setDecodingFunction

Used `std::error_code` versions of `std::filesystem` features to prevent exceptions from being thrown if say.. a directory happens to be deleted during a search.

Added `directory_options::follow_directory_symlink` flag to the recursive search.

Normalized paths to ensure proper uniqueness function.

### DIFF
--- a/Source/CMakeLists.txt
+++ b/Source/CMakeLists.txt
@@ -20,6 +20,19 @@ set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
+# Make sure std::filesystem usage actually compiles and links. *rolls eyes at android*
+check_cxx_source_compiles("
+	#include <filesystem>
+	int main() { std::filesystem::current_path(); }"
+	STD_FILESYSTEM_OK)
+
+if(STD_FILESYSTEM_OK)
+  message(STATUS "Environment supports std::filesystem.")
+  add_definitions(-DHAS_STD_FILESYSTEM)
+else()
+  message(STATUS "Environment does not support std::filesystem.")
+endif()
+
 if (MSVC)
 	# TODO: Use https://cmake.org/cmake/help/latest/policy/CMP0092.html instead (once we can require CMake >= 3.15)
   # Taken from http://www.cmake.org/Wiki/CMake_FAQ#Dynamic_Replace.
--- a/Source/Core/Common/FileSearch.cpp
+++ b/Source/Core/Common/FileSearch.cpp
@@ -4,16 +4,16 @@
 
 #include <algorithm>
 #include <functional>
+#include <regex>
 
 #include "Common/CommonPaths.h"
 #include "Common/FileSearch.h"
+#include "Common/Logging/Log.h"
 #include "Common/StringUtil.h"
 
-#ifdef _MSC_VER
-#include <Windows.h>
+#ifdef HAS_STD_FILESYSTEM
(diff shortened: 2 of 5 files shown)


JosJuice:
I think it would make more sense to use the C locale for case-insensitive extension matching.

Many programs would have a reason to use the user's locale for matching on the filename itself (minus the extension), since the file name might contain words in the user's language, but the extension isn't really in the user's language. I would be surprised if I ever saw the extension .İSO anywhere, for instance.

The Qt thing that you mentioned is rather different since it is used for converting between encodings, not just for case insensitivity.

shuffle2:
How does the perf compare

jordan-woyak:
> How does the perf compare

I'm like 92% sure this function is I/O bound.
Maybe tomorrow I'll test `CompareStringOrdinal` against `std::regex`.

jordan-woyak:
> How does the perf compare

@shuffle2 It's only 40 times slower. #nbd
https://i.imgur.com/C4x5LfW.png

shuffle2:
yea...this function was optimized because people tell dolphin to use like... C:\ (or some other drive root) as the iso path, which causes dolphin to parse every filename on a large directory structure.
(and some other reasons, probably)

jordan-woyak:
@shuffle2 I'm sure the actual filesystem access is the bottleneck.
Even if `std::regex` is 40 times slower than `CompareStringOrdinal` it still ran 700,000 comparisons in a second.

shuffle2:
It wasn't, I profiled it at the time.

jordan-woyak:
@shuffle2 I'll try this later. But perhaps your profiling was of the old `DoFileSearch` implementation that adds every single file to a tree structure which is then processed.
Your efficient `DoFileSearch` logic still here. I've just changed the comparison function.

JosJuice:
FWIW, we are no longer using this code for the game list except on Android, so you can't test this by just adding C:\ as a game directory.

shuffle2:
well, whatever, then

shuffle2:
image
sigh

leoetlino:
Any reason for not just using strcasecmp (which is defined to stricmp on Windows in CommonFuncs)? Seems even simpler than generating a regex and probably much more efficient too. I agree with @JosJuice that file extensions are unlikely to be anything other than ASCII characters anyway.

> Many programs would have a reason to use the user's locale for matching on the filename itself (minus the extension), since the file name might contain words in the user's language, but the extension isn't really in the user's language. I would be surprised if I ever saw the extension .İSO anywhere, for instance.
>
> The Qt thing that you mentioned is rather different since it is used for converting between encodings, not just for case insensitivity.

JosJuice:
> Any reason for not just using strcasecmp (which is defined to stricmp on Windows in CommonFuncs)?

`fs::path::value_type` is `wchar_t` on Windows.

leoetlino:
Can't we turn it into a string by calling fs::path::string, though?

JosJuice:
You would need to use our utility function `PathToString` in order for it to work on C++20, but yes. This does however somewhat ruin the point of c5fa470 (which shuffle2 was referring to earlier), since you now have to spend time converting from UTF-16 to UTF-8.

jordan-woyak:
This PR is dead.

PR #8630, opened 2020-02-15
Common: Avoid std::function overhead in ScopeGuard

leoetlino:
So far in all our uses of ScopeGuard, the type of the callable is
usually just a lambda or a function pointer, so there is no need
to rely on std::function's type erasure.

While the cost of using std::function is probably negligible, it still
causes some unnecessary overhead that can be avoided by making
ScopeGuard a templated class. Thanks to class template argument
deduction in C++17 most existing usages do not even need to be changed.

See https://godbolt.org/z/KcoPni for a comparison between
a ScopeGuard that uses std::function and one that doesn't

### DIFF
--- a/Source/Core/Common/ScopeGuard.h
+++ b/Source/Core/Common/ScopeGuard.h
@@ -4,17 +4,15 @@
 
 #pragma once
 
-#include <functional>
+#include <optional>
 
 namespace Common
 {
+template <typename Callable>
 class ScopeGuard final
 {
 public:
-  template <class Callable>
-  ScopeGuard(Callable&& finalizer) : m_finalizer(std::forward<Callable>(finalizer))
-  {
-  }
+  ScopeGuard(Callable&& finalizer) : m_finalizer(std::forward<Callable>(finalizer)) {}
 
   ScopeGuard(ScopeGuard&& other) : m_finalizer(std::move(other.m_finalizer))
   {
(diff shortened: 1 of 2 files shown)


CookiePLMonster:
Very good change, I had it in mind more than once in the past so glad to see it done. However, I am pretty sure `m_finalizer` doesn't need to be a `std::optional` and you can "null" it by assigning `{}` to it.

leoetlino:
Unfortunately that doesn't work for lambdas.

```
../Source/Core/Common/ScopeGuard.h:24:9: error: could not convert ‘((Common::ScopeGuard<Core::EmuThread(std::unique_ptr<BootParameters>, WindowSystemInfo)::<lambda()> >*)this)->Common::ScopeGuard<Core::EmuThread(std::unique_ptr<BootParameters>, WindowSystemInfo)::<lambda()> >::m_finalizer’ from ‘Core::EmuThread(std::unique_ptr<BootParameters>, WindowSystemInfo)::<lambda()>’ to ‘bool’
   24 |     if (m_finalizer)
      |         ^~~~~~~~~~~
      |         |
      |         Core::EmuThread(std::unique_ptr<BootParameters>, WindowSystemInfo)::<lambda()>
```
...
```
../Source/Core/Common/ScopeGuard.h:27:19: error: use of deleted function ‘Core::EmuThread(std::unique_ptr<BootParameters>, WindowSystemInfo)::<lambda()>& Core::EmuThread(std::unique_ptr<BootParameters>, WindowSystemInfo)::<lambda()>::operator=(const Core::EmuThread(std::unique_ptr<BootParameters>, WindowSystemInfo)::<lambda()>&)’
   27 |       m_finalizer = {};
      |       ~~~~~~~~~~~~^~~~
../Source/Core/Core/Core.cpp:524:71: note: a lambda closure type has a deleted copy assignment operator
  524 |   Common::ScopeGuard controller_guard{[init_controllers, init_wiimotes] {
```

PR #8631, opened 2020-02-15
InputCommon: Change default yaw/pitch to reach all corners

chungy:
The old default did not allow pointing (with the mouse) to all parts
of the screen, and notably, even the System Menu has controls that
were out-of-reach with them.

With this, yaw is changed to a default of 19° and pitch changed to a
default of 12°.  This overshoots the screen in either 16:9 or 4:3
aspect ratios, but allows the complete screen to be pointed to in
either one.

### DIFF
--- a/Source/Core/InputCommon/ControllerEmu/ControlGroup/Cursor.cpp
+++ b/Source/Core/InputCommon/ControllerEmu/ControlGroup/Cursor.cpp
@@ -33,7 +33,7 @@ Cursor::Cursor(std::string name, std::string ui_name)
 
   AddInput(Translate, _trans("Relative Input Hold"));
 
-  // Default values are optimized for "Super Mario Galaxy 2".
+  // Default values are optimized for the Wii System Menu.
   // This seems to be acceptable for a good number of games.
 
   AddSetting(&m_vertical_offset_setting,
@@ -50,7 +50,7 @@ Cursor::Cursor(std::string name, std::string ui_name)
               _trans("°"),
               // i18n: Refers to emulated wii remote movements.
               _trans("Total rotation about the yaw axis.")},
-             15, 0, 360);
+             20, 0, 360);
 
   AddSetting(&m_pitch_setting,
              // i18n: Refers to an amount of rotational movement about the "pitch" axis.


jordan-woyak:
Like comment on line 36 says, the defaults have been optimized for "Super Mario Galaxy 2" which seems to work decently for a handful of popular games.
I don't think optimizing the defaults for the sparingly-used system menu makes sense.

Are there any significant games this changes makes better for?

An ideal solution might be a database of ideal yaw/pitch angles of each game for automatic adjustment.

chungy:
> the defaults have been optimized for "Super Mario Galaxy 2" which seems to work decently for a handful of popular games.
> Are there any significant games this change makes better for?

Even in Super Mario Galaxy 1 (I don't have 2), the defaults (15/15) are too constrained to actually use the in-game UI.

>  I don't think optimizing the defaults for the sparingly-used system menu makes sense.

I used it as an example, though the system menu is a necessity for managing saves, I use it frequently as such...

Miksel12:
I don't have any issues with Super Mario Galaxy 1/2 when using 15/15 but I do use 20/15 as yaw and pitch because I did encounter some issues with the default values, including with the Wii menu. So I do agree with the increase in yaw degree as not being able to reach something with the pointer is really annoying but the pointer being off screen can be easily avoided by the users, just like with an actually Wii mote. With pitch to 12 degrees, I can't reach the edge in Super Mario Galaxy so I would keep that at 15, and once again, I'd rather have my pointer off screen than not being able to reach edge of the screen.

chungy:
> With pitch to 12 degrees, I can't reach the edge in Super Mario Galaxy so I would keep that at 15, and once again, I'd rather have my pointer off screen than not being able to reach edge of the screen.

Fair point, I didn't realize that different software can interpret pointing so differently.

leoetlino:
Could you rebase this please?

FWIW, 15/15 appears to work for me in SMG as well... but I agree that not being able to use the UI in some titles is pretty annoying and probably worse than having the pointer be slightly off screen.

jordan-woyak:
The comment about SMG2 on line 36 needs to be updated. You could mention yaw has been increased for the Wii Menu or whatever. Also this needs a rebase.

Source/Core/InputCommon/ControllerEmu/ControlGroup/Cursor.cpp
```diff
@@ -33,7 +33,7 @@ Cursor::Cursor(std::string name, std::string ui_name)
 
   AddInput(Translate, _trans("Relative Input Hold"));
 
-  // Default values are optimized for "Super Mario Galaxy 2".
+  // Default values are optimized for the Wii System Menu.
```
jordan-woyak, line 36: Thanks for changing the comment but it isn't really accurate. :P
Defaults of 20/15 might allow reaching every area of the screen but it certainly isn't "optimized" for the system menu.
There is extreme overshoot.

PR #8632, opened 2020-02-17
Windows GUI: Use QMenu font (Segoe UI) for entire application

jimbo1qaz:
On Windows, Qt's default system font (MS Shell Dlg 2) is outdated.

Dolphin previously used over 15 lines of code to compute a font closer to the proper font, but with an approximately correct font size. Using the QMenu font directly is both more concise and more elegant (in my opinion).

Successor to #8359.

Credit to https://stackoverflow.com/a/40380684 for discovering this method. Is it worth crediting in the source code? I think not?

### DIFF
--- a/Source/Core/DolphinQt/Main.cpp
+++ b/Source/Core/DolphinQt/Main.cpp
@@ -101,27 +101,12 @@ int main(int argc, char* argv[])
   QApplication app(argc, argv);
 
 #ifdef _WIN32
-  // Get the default system font because Qt's way of obtaining it is outdated
-  NONCLIENTMETRICSW metrics = {};
-  LOGFONTW& logfont = metrics.lfMenuFont;
-  metrics.cbSize = sizeof(metrics);
-
-  if (SystemParametersInfoW(SPI_GETNONCLIENTMETRICS, sizeof(metrics), &metrics, 0))
-  {
-    // Sadly Qt 5 doesn't support turning a native font handle into a QFont so this is the next best
-    // thing
-    QFont font = QApplication::font();
-    font.setFamily(QString::fromStdWString(logfont.lfFaceName));
-
-    font.setItalic(logfont.lfItalic);
-    font.setStrikeOut(logfont.lfStrikeOut);
-    font.setUnderline(logfont.lfUnderline);
-
-    // The default font size is a bit too small
-    font.setPointSize(QFontInfo(font).pointSize() * 1.2);
-
-    QApplication::setFont(font);
-  }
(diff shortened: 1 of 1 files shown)


CookiePLMonster:
@dolphin-emu-bot rebuild

CookiePLMonster:
Please provide before/after comparison screenshots.

jimbo1qaz:
> Please provide before/after comparison screenshots.

It seems to be visually indistinguishable in the English locale, for fonts at the default size. There may be slight differences in other locales, or parts of the UI using differently sized fonts (but I can't find any, even in the GC/Wiimote controller viewers).

Master:
master

PR:
pr

TBH I'm mostly doing this as a proof of concept. Dolphin already has modern fonts on Windows, so this is primarily a code cleanup. However many other Qt-based programs are still stuck on MS Shell Dlg 2, and would benefit more from this method. (Credit to https://stackoverflow.com/a/40380684 for discovering this method.)

MayImilae:
Wow that looks exactly the same. I guess it isn't something I need to worry about!

Source/Core/DolphinQt/Main.cpp
```diff
@@ -101,27 +101,11 @@ int main(int argc, char* argv[])
...
-    // The default font size is a bit too small
-    font.setPointSize(QFontInfo(font).pointSize() * 1.2);
-
-    QApplication::setFont(font);
-  }
+  // On Windows, Qt's default system font (MS Shell Dlg 2) is outdated.
+  // Interestingly, the QMenu font is correct and probably comes from lfMenuFont
+  // (Segoe UI on English computers).
+  // So use it for the entire application.
+  QApplication::setFont(QApplication::font("QMenu"));
```
spycrab: Please add a line about this being obsolete once we switch to [Qt 6 (which will release this year)](https://github.com/qt/qtbase/blob/35262678c31a4aea0df1516c565cf9a2e808b369/src/platformsupport/fontdatabases/windows/qwindowsfontdatabasebase.cpp#L760) (You don't need to add this link).

jimbo1qaz: New commit, or amend and force push?

spycrab: amend and force push please :)

jimbo1qaz:
`QApplication::font("QMenu")` traces down to https://code.woboq.org/qt5/qtbase/src/plugins/platforms/windows/qwindowstheme.cpp.html#505 which uses `lfMenuFont`. I don't know if I can really document this in the code though. (Sidenote: Qt's font handling is *special*, extremely complex, and has some unexpected properties.)

PR #8633, opened 2020-02-17
Qt/TAS: Improve TAS windows

spycrab:
- Use the Dolphin Icon in TAS Windows
- Prevent deformation of the stick widget (Bug #11988)
- Improve visual appearance
- Set a reasonable minimum size

**Before**:  
Before
Before
**After**:
After
After

### DIFF
--- a/Source/Core/DolphinQt/TAS/IRWidget.cpp
+++ b/Source/Core/DolphinQt/TAS/IRWidget.cpp
@@ -12,11 +12,16 @@
 
 #include "Common/CommonTypes.h"
 
+constexpr int PADDING = 1;
+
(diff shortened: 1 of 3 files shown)


exurd:
The circles in the GameCube window seem to have off-centre crosses.
I've aligned the circles before the change and after to see if the crosses got misaligned at some point, but there was no difference.
When aligned, the crosses match while the circles don't.
It's still using the crosses from before.

spycrab:
@exurd Yes, this PR does not intend to fix this bug. I might take a stab at doing that later. Is there already a bug report for this?

spycrab:
It might be fixed now not quite sure though.

PR #8634, opened 2020-02-17
Fix Sonic Colors/Unleashed gui (again)

MastaG:
See: https://bugs.dolphin-emu.org/issues/11897
When running Sonic using OpenGL ES the gui elements are not rendered at all.

I've confirmed this on two mobiles phones (Samsung Galaxy S9 and Oneplus 7t Pro).
But it can also easily be reproduced on Linux by setting "PreferGLES = True" in GFX.ini.

It renders fine using Vulkan and regular OpenGL without this hack.
However only OpenGL ES seems to run most games at acceptable speeds on Android phones.

### DIFF
--- a/Source/Core/VideoCommon/VertexShaderManager.cpp
+++ b/Source/Core/VideoCommon/VertexShaderManager.cpp
@@ -5,6 +5,7 @@
 #include "VideoCommon/VertexShaderManager.h"
 
 #include <array>
+#include <cfloat>
 #include <cmath>
 #include <cstring>
 #include <iterator>
@@ -371,7 +372,8 @@ void VertexShaderManager::SetConstants()
       g_fProjectionMatrix[12] = 0.0f;
       g_fProjectionMatrix[13] = 0.0f;
 
-      g_fProjectionMatrix[14] = -1.0f;
+      // Hack to fix depth clipping precision issues (such as Sonic Adventure UI)
+      g_fProjectionMatrix[14] = -(1.0f + FLT_EPSILON);
       g_fProjectionMatrix[15] = 0.0f;
 
       g_stats.gproj = g_fProjectionMatrix;
@@ -397,7 +399,8 @@ void VertexShaderManager::SetConstants()
       g_fProjectionMatrix[13] = 0.0f;
 
       g_fProjectionMatrix[14] = 0.0f;
-      g_fProjectionMatrix[15] = 1.0f;
+      // Hack to fix depth clipping precision issues (such as Sonic Unleashed UI)
+      g_fProjectionMatrix[15] = 1.0f + FLT_EPSILON;
 
       g_stats.g2proj = g_fProjectionMatrix;
       g_stats.proj = rawProjection;


Source/Core/VideoCommon/VertexShaderGen.cpp
```diff
@@ -421,11 +421,8 @@ ShaderCode GenerateVertexShaderCode(APIType api_type, const ShaderHostConfig& ho
   {
     // Since we're adjusting z for the depth range before the perspective divide, we have to do our
     // own clipping. We want to clip so that -w <= z <= 0, which matches the console -1..0 range.
-    // We adjust our depth value for clipping purposes to match the perspective projection in the
-    // software backend, which is a hack to fix Sonic Adventure and Unleashed games.
-    out.Write("float clipDepth = o.pos.z * (1.0 - 1e-7);\n");
-    out.Write("float clipDist0 = clipDepth + o.pos.w;\n");  // Near: z < -w
-    out.Write("float clipDist1 = -clipDepth;\n");           // Far: z > 0
```
stenzek: This will break desktop GL. `clipDist0`/`clipDist1` is referenced on both line 428 and 515.

stenzek:
I wonder if this is due to the lack of depth clamp in GLES. If there's an extension available now which provides this feature, that sounds preferable to a GLES-only hack.

Source/Core/VideoCommon/VertexShaderGen.cpp
MastaG: You're right. That part of the revert wasn't required at all.
Fixed it.

MastaG:
I honestly don't know stenzek.

I was just annoyed by the fact that I needed to use "custom" Android builds (which don't have most of their sources available), just to be able to play the Sonic Wii games using OpenGL ES.
Like I said Vulkan is way too slow (both on my Samsung S9 and Oneplus 7T Pro).

But perhaps you can guard the hack with some IF statement so it only applies when using OpenGL ES because desktop GL and Vulkan are not affected and render the gui in Sonic games just fine.

stenzek:
@dolphin-emu-bot rebuild

MastaG:
Anyways I forced push again using only the depth clipping hack and fixed the #include that lint was complaining about.

Furthermore I've tested and verified both desktop OpenGL and OpenGL ES to be working and rendering the gui in Sonic like it should.

Also OpenGL ES was tested again both on Linux (using PreferGLES=True) and using a custom build with Android Studio for my Oneplus 7T Pro.

leo60228:
GL_EXT_depth_clamp seems to be identical to ARB_depth_clamp. I can't find any information on driver support, though.

leo60228:
It wasn't added to Mesa until February 2019, so most likely very poor.

leo60228:
It wasn't added to Mesa until February 2019, so most likely very poor.

MastaG:
Hi, any update on this?
Regarding driver support, to my knowledge Android GLES drivers are binary only (and usually not updated until they ship the same GPU in a new device).

And who's going to play Wii games on Mesa's GLES renderer when most Mesa users are using x86 setups with GPU's which have full OpenGL available.

GLES is used in mobile devices only and that's where the projection hack is required to play the Sonic games.

MastaG:
@stenzek Could you please consider merging this taking my last comment in mind.

leo60228:
I was just using Mesa as an example. I wouldn't be surprised if it was the first to implement the extension.

MastaG:
Yeah that's why using a GLES extension that was added to Mesa for fixing the issue is not a solution.. since basically nobody is using Dolphin on Mesa with GLES anyways.

Also all existing mobile platforms aren't getting updated driver blobs anyways.

Only think I can think of would be some guarding in my commit so it only applies it when the GLES renderer is in use.

Perhaps somebody could give some hint on that since I'm not very familiar with the Dolphin code base (and C++ for that matter).

I'd really like to have this fixed so we can play the Sonic games on the official Android builds :)

Miksel12:
This is fixed by https://github.com/dolphin-emu/dolphin/pull/9591

JosJuice:
Closing in favor of the other PR, since the implementation in the other PR only activates when the depth clamp extension isn't present.

PR #8635, opened 2020-02-19
Add android gamepad profile loading and saving

Palidino:
This adds an "Open Gamepad Profile" and "Save Gamepad Profile" in the main menu under "Refresh Library" and "Open File."

I suspect this isn't the best way to do it, and I've never worked on an Android app before and had some trouble maneuvering around the menu system to add these options deeper in the settings instead of the main menu.  Hopefully, this will at least get some inspiration rolling to add controller profiles to the Android version.

If anyone wants to take parts from this to make a better system, please feel free to do so.

**How it works:**
Configure your emulated controls for Game Cube, Wii, etc with a specific gamepad. You can configure the controls for all the different emulated controllers for saving to a single profile, such as if you want to save a standard config for an Xbox controller with GC, WII remote, classic controller, etc.

Select save gamepad profile, choose a location, and name it. This saves all the Android-based controller settings into the file.

Select load gamepad profile and all your current emulated controls will be cleared and set to what is in the file you pick.

Sample contents of a saved profile:
```
[Profile]
InputL_0 = Device 'xyz'-Button 104

```
In emulated input, it'll be displayed as the file's name followed by the input (EX: xbox_controller.ini profile loaded, displays as "xbox_controller:  Button 104").

**Uses**
In my specific case, I have an LG G8X. I have my second screen to use when I'm out and about. I then have a gamepad that wraps around my phone for in bed. On top of this, I have an Xbox controller for when I'm playing on a TV. With this pull request, I'm able to create a profile for each controller, and then just quickly load that profile when I swap to a different controller, opposed to previously having to reconfigure each button whenever I change controllers.

### DIFF
--- a/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/model/Settings.java
+++ b/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/model/Settings.java
@@ -163,7 +163,8 @@ public class Settings
   {
     if (TextUtils.isEmpty(gameId))
     {
-      view.showToastMessage("Saved settings to INI files");
+      if (view != null)
+        view.showToastMessage("Saved settings to INI files");
 
       for (Map.Entry<String, List<String>> entry : configFileSectionsMap.entrySet())
       {
--- a/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/fragments/CustomFilePickerFragment.java
+++ b/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/fragments/CustomFilePickerFragment.java
@@ -21,7 +21,7 @@ import java.util.Set;
 public class CustomFilePickerFragment extends FilePickerFragment
 {
   private static final Set<String> extensions = new HashSet<>(Arrays.asList(
-          "gcm", "tgc", "iso", "ciso", "gcz", "wbfs", "wad", "dol", "elf", "dff"));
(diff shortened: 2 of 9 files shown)


JosJuice:
While I would like to see this feature on Android, I think there's too much complexity the way it currently is implemented. And that's not your fault – the underlying problem here is that Android has its own weird way of handling controller mappings instead of saving them in INI files like we do on PC. Had they been saved in INI files, you would essentially just need to copy those INI files in order to get profiles. But the way it is now, this PR adds what looks like another INI reader+writer, and defines a new controller profile format that is incompatible with the existing one that is in use on PC.

JosJuice:
@dolphin-emu-bot rebuild

jordan-woyak:
Yeah. I think we should remove the extra level of indirection in the android input system before adding profile support.

Palidino:
> Yeah. I think we should remove the extra level of indirection in the android input system before adding profile support.

I agree. I went into this thinking that adding loading and saving would much more simple, only to discover that profiles were designed completely different (or non-existent really) from the desktop versions. Unfortunately, I felt completely rewriting the input system on Android to match desktops was outside the scope of what I felt comfortable trying to do, so I opted to just transfer the current Android system into their own file outside of Dolphin.ini.

leoetlino:
I'm sorry, but as others have already said, the Android input system should really be fixed before we add even more code -- especially when it will just get removed at some point.

PR #8637, opened 2020-02-22
TitleDatabase: Added custom developer ("maker") database option.

MrMendelli:
Functions identically to `titles.txt`, but uses `publishers.txt` instead. I also think that _"Maker"_ should be changed to _"Publisher"_ in the UI. This will make more sense to end-users, but the code should stick to _"maker"_ as it's the official metadata tag and would require more work to rename everything. In accordance, the title database load option has been re-labeled to `Use Built-In Database of Metadata`. Note that the ability to load the title database is already broken if a custom title file is present, it seems that the custom titles aren't unloaded if the `Use Built-In Database of Titles` option is enabled. Unless I'm mistaken, the title database strings should be predominant if the option is enabled for all titles even if a custom titles file is present. Furthermore, the entire metadata load implementation would have to be completely re-written to fix this bug.

Comparison

### DIFF
--- a/Source/Core/Core/TitleDatabase.cpp
+++ b/Source/Core/Core/TitleDatabase.cpp
@@ -67,6 +67,9 @@ TitleDatabase::TitleDatabase()
   if (m_user_title_map.empty())
     m_user_title_map = LoadMap(load_directory + "titles.txt");
 
+  // Custom developer ("maker") index
+  m_user_developer_map = LoadMap(load_directory + "developers.txt");
+
   // Pre-defined databases (one per language)
   AddLazyMap(DiscIO::Language::Japanese, "ja");
   AddLazyMap(DiscIO::Language::English, "en");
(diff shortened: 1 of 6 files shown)


Source/Core/Core/TitleDatabase.cpp
```diff
@@ -91,8 +94,26 @@ TitleDatabase::TitleDatabase()
...
+
+{
+  auto it = m_user_developer_map.find(gametdb_id);
+  if (it != m_user_developer_map.end())
+    return it->second;
+
+  if (!SConfig::GetInstance().m_use_builtin_title_database)
+    return EMPTY_STRING;
+
+  it = m_base_map.find(gametdb_id);
```
JosJuice: Isn't this the map that contains game names, meaning that a game name could show up in the Developer column if you don't have the game in developers.txt?

JosJuice:
@dolphin-emu-bot rebuild

Source/Core/Core/TitleDatabase.cpp
```diff
@@ -91,8 +94,26 @@ TitleDatabase::TitleDatabase()
 
 TitleDatabase::~TitleDatabase() = default;
 
+const std::string& TitleDatabase::GetDeveloper(const std::string& gametdb_id) const
+
```
JosJuice: Could you remove the empty line here and the empty line you added to the function below? It seems like lint isn't complaining about it, but we never put an empty line here.

Source/Core/Core/TitleDatabase.cpp
MrMendelli: Thank you for the clarification, I will re-work this to the appropriate category of publisher and my oversights for using custom metadata or not.

Source/Core/Core/TitleDatabase.cpp
MrMendelli: Yes.

PR #8639, opened 2020-02-22
CMake: Disable libcurl from Externals on Unix

spycrab:
Because of the current way we ship libcurl (preconfigured) the CA_BUNDLE path is hardcoded to one path regardless of the actual path appropriate for the system (Bug #11772)

[Fixes Bug 11772](https://bugs.dolphin-emu.org/issues/11772).


spycrab:
I'd be willing to change this in a way where libcurl from Externals is only ignored for Linux too.

Tilka:
> Now still uses Externals curl on distributions that actually have that certificate file installed

Why?

spycrab:
I mostly wanted to go the path of least resistance to get this fixed because some people might object to basically making our libcurl Externals useless on anything but Windows. Plus the fact that this is only a temporary solution that should be replaced by a proper External with CMake later on.

CMakeLists.txt
```diff
@@ -685,15 +685,21 @@ else()
   include_directories(Externals/mbedtls/include)
 endif()
 
-find_package(CURL)
-if(CURL_FOUND)
-  message(STATUS "Using shared libcurl")
-  include_directories(${CURL_INCLUDE_DIRS})
+# The libcurl we currently use has a harcoded certificate location
+# It can only be used on distributions where SSL certificates actually are at that path.
+if (NOT UNIX OR EXISTS "/etc/ssl/certs/ca-certificates.crt")
```
JosJuice: This checks whether the file exists on the system that is building Dolphin, right? That's going to be a problem when cross-compiling for Android.

CMakeLists.txt
```diff
@@ -685,15 +685,21 @@ else()
...
+    add_subdirectory(Externals/curl)
+    set(CURL_LIBRARIES curl)
+    include_directories(BEFORE Externals/curl/include)
+  endif()
 else()
-  message(STATUS "Using static libcurl from Externals")
-  add_subdirectory(Externals/curl)
-  set(CURL_LIBRARIES curl)
-  include_directories(BEFORE Externals/curl/include)
+  find_package(CURL REQUIRED)
```
JosJuice: You might want to put `message(STATUS "Using shared libcurl")` here. Also, why is `include_directories` used above but not here?

CMakeLists.txt
spycrab: I guess I could check [CMAKE_CROSSCOMPILING](https://cmake.org/cmake/help/latest/variable/CMAKE_CROSSCOMPILING.html)?

CMakeLists.txt
spycrab: I *somehow* missed that.

Tilka:
> path of least resistance to get this fixed

But on Linux distributions that have the bundle at the expected path, you changed libcurl from shared to static even though the shared version worked, no? I don't understand.

spycrab:
No as I understand CMake operator precedence
> NOT UNIX OR EXISTS "..."
Has a precedence like this
> (NOT UNIX) OR (EXISTS "...")

And in that case it falls back to our static libcurl if need be.

dreamsyntax:
We can close this; Redundant after https://github.com/dolphin-emu/dolphin/pull/13228
The bug is now marked as fixed. https://bugs.dolphin-emu.org/issues/11772

PR #8640, opened 2020-02-23
Qt: Re-enable Dark Mode support

spycrab:
The Qt version we use now should support Dark Mode properly.

### DIFF
--- a/Source/Core/DolphinQt/Info.plist.in
+++ b/Source/Core/DolphinQt/Info.plist.in
@@ -50,7 +50,5 @@
     <true/>
     <key>CSResourcesFileMapped</key>
     <true/>
-    <key>NSRequiresAquaSystemAppearance</key>
-    <true/>
 </dict>
 </plist>


8times9:
image
The toolbar text now displays correctly in dark mode

spycrab:
@8times9 Does the rest of the interface look okay too? (Especially TAS and the Controller config Windows for Wiimotes / Gamecube Controllers)

JosJuice:
This only affects macOS, right?

spycrab:
Yes

MayImilae:
Here are a bunch of screenshots!

[pr8640-Screenshots.zip](https://github.com/dolphin-emu/dolphin/files/4241170/pr8640-Screenshots.zip)

So two main issues I spotted. The C stick is black against dark grey, with only a little tiny bit of yellow peeking through. That's not great. For all other joysticks, the calibration bounds for the joystick, the dashed lines, are black against dark grey as well, and nearly invisible. Other than those two things, everything seems good. 

Screen Shot 2020-02-23 at 03 43 49

spycrab:
@MayImilae It seems like that code already uses palettes instead of hard coded colors, so fixing this will be a bit more difficult. Would you prefer this PR be on hold until we fix that or should we go ahead and merge it anyway?

MayImilae:
I believe billiard is already looking into the indicators? Maybe ask him, he may already have it fixed in a pr.

spycrab:
I checked myself on Catalina. Contrast looks good now.

PR #8641, opened 2020-02-23
Dark Mode for Android 

Notjailbrokenguy:
You guys should add dark mode for android. I think that should be cool, wouldn’t it?


JosJuice:
This is not a place for feature requests. If you have a feature request, I would suggest posting it on the forums. This GitHub repo is for posting code that you have written.

PR #8642, opened 2020-02-23
Qt/CodeViewWidget: Implement branch arrows.

AdmiralCurtiss:
Back in WX, the debugger would draw arrows to indicate branches, like this:
https://user-images.githubusercontent.com/4522237/75113262-83629280-564c-11ea-857e-b86459b47c08.png
https://user-images.githubusercontent.com/4522237/75113263-83629280-564c-11ea-865e-e672b01c57a6.png

I found this pretty useful at quickly determining conditionals and loops, but during the transition to Qt this feature was lost. This PR is a implementation of this feature for Qt.

Screenshots:
https://user-images.githubusercontent.com/4522237/75113260-82316580-564c-11ea-9057-bf7dc6cc169a.png
https://user-images.githubusercontent.com/4522237/75202872-eb0fff00-576c-11ea-869c-0a9ca0b208ef.png

One thing I'm not sure here is if I have to make sure `BranchDisplayDelegate::paint()` and `CodeViewWidget::Update()` don't run at the same time in different threads, or if those two are both only already allowed on the GUI thread anyway. Does anyone know?

### DIFF
--- a/Source/Core/DolphinQt/Debugger/CodeViewWidget.cpp
+++ b/Source/Core/DolphinQt/Debugger/CodeViewWidget.cpp
@@ -36,9 +36,16 @@ constexpr double SCROLL_FRACTION_DEGREES = 15.;
 
 constexpr size_t VALID_BRANCH_LENGTH = 10;
 
+constexpr int CODE_VIEW_COLUMN_BREAKPOINT = 0;
+constexpr int CODE_VIEW_COLUMN_ADDRESS = 1;
+constexpr int CODE_VIEW_COLUMN_INSTRUCTION = 2;
+constexpr int CODE_VIEW_COLUMN_PARAMETERS = 3;
+constexpr int CODE_VIEW_COLUMN_DESCRIPTION = 4;
+constexpr int CODE_VIEW_COLUMNCOUNT = 5;
+
 CodeViewWidget::CodeViewWidget()
 {
-  setColumnCount(5);
+  setColumnCount(CODE_VIEW_COLUMNCOUNT);
   setShowGrid(false);
   setContextMenuPolicy(Qt::CustomContextMenu);
   setSelectionMode(QAbstractItemView::SingleSelection);
@@ -86,7 +93,7 @@ void CodeViewWidget::FontBasedSizing()
   const int rowh = fm.height() + 1;
   verticalHeader()->setMaximumSectionSize(rowh);
   horizontalHeader()->setMinimumSectionSize(rowh + 5);
-  setColumnWidth(0, rowh + 5);
+  setColumnWidth(CODE_VIEW_COLUMN_BREAKPOINT, rowh + 5);
   Update();
 }
 
@@ -189,11 +196,11 @@ void CodeViewWidget::Update()
           Resources::GetScaledThemeIcon("debugger_breakpoint").pixmap(QSize(rowh - 2, rowh - 2)));
     }
 
-    setItem(i, 0, bp_item);
-    setItem(i, 1, addr_item);
-    setItem(i, 2, ins_item);
-    setItem(i, 3, param_item);
-    setItem(i, 4, description_item);
+    setItem(i, CODE_VIEW_COLUMN_BREAKPOINT, bp_item);
+    setItem(i, CODE_VIEW_COLUMN_ADDRESS, addr_item);
+    setItem(i, CODE_VIEW_COLUMN_INSTRUCTION, ins_item);
+    setItem(i, CODE_VIEW_COLUMN_PARAMETERS, param_item);
+    setItem(i, CODE_VIEW_COLUMN_DESCRIPTION, description_item);
 
     if (addr == GetAddress())
     {
@@ -559,7 +566,7 @@ void CodeViewWidget::mousePressEvent(QMouseEvent* event)
   switch (event->button())
   {
   case Qt::LeftButton:
-    if (column(item) == 0)
+    if (column(item) == CODE_VIEW_COLUMN_BREAKPOINT)
       ToggleBreakpoint();
     else
       SetAddress(addr, SetAddressUpdate::WithUpdate);
--- a/Source/Core/DolphinQt/Debugger/CodeViewWidget.cpp
+++ b/Source/Core/DolphinQt/Debugger/CodeViewWidget.cpp
@@ -14,8 +14,10 @@
 #include <QKeyEvent>
 #include <QMenu>
 #include <QMouseEvent>
+#include <QPainter>
 #include <QResizeEvent>
 #include <QScrollBar>
+#include <QStyledItemDelegate>
 #include <QTableWidgetItem>
 #include <QWheelEvent>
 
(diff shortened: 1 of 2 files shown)


Source/Core/DolphinQt/Debugger/CodeViewWidget.cpp
```diff
@@ -90,6 +158,11 @@ void CodeViewWidget::FontBasedSizing()
   Update();
 }
 
+u32 CodeViewWidget::AddressForRow(int row) const
+{
+  return m_address - ((rowCount() / 2) * 4) + row * 4;
```
BhaaLseN: Thoughts on documenting those numbers (either as comment or `constexpr` if it makes sense to reuse them elsewhere)? Dividing by two and multiplying by four (is this the same four on both multiplies btw?) looks like magic and I cannot see a reason why either is necessary and/or yields the correct address here.

AdmiralCurtiss: That line was copied out of the row loop in `CodeViewWidget::Update()` so I'm not entirely sure what it's trying to do either. I think it's trying to align the address to be divisible by 4?

AdmiralCurtiss: Alright, I thought about it and ended up with:

~~~
  // m_address is defined as the center row of the table, so we have rowCount/2 instructions above
  // it; an instruction is 4 bytes long on GC/Wii so we increment 4 bytes per row
  const u32 rowZeroAddress = m_address - ((rowCount() / 2) * 4);
  return rowZeroAddress + row * 4;
~~~

Does that make sense to you?

BhaaLseN: Seems better than before 👍

Source/Core/DolphinQt/Debugger/CodeViewWidget.cpp
```diff
@@ -30,6 +32,68 @@
...
+             const QModelIndex& index) const override
+  {
+    QStyledItemDelegate::paint(painter, option, index);
+
+    painter->save();
+
+    painter->setPen(m_parent->palette().shadow().color());
+
+    const u32 addr = m_parent->AddressForRow(index.row());
+    const u32 x_offset_in_branch_for_vertical_line = 10;
```
Warepire:
```suggestion
    constexpr u32 x_offset_in_branch_for_vertical_line = 10;
```

Source/Core/DolphinQt/Debugger/CodeViewWidget.cpp
```diff
@@ -202,6 +284,7 @@ void CodeViewWidget::Update()
   }
 
   resizeColumnsToContents();
+  setColumnWidth(5, static_cast<int>(WIDTH_PER_BRANCH_ARROW * m_branches.size()));
```
Warepire: Thoughts about naming the columns via e.g. an enum to make the code more obvious?

AdmiralCurtiss: Might as well. Do we prefer an enum or several constexpr ints for values that are going to be used as integers?

Warepire: For this I usually go for an unnamed enum in a namespace, but I don't know what the project prefers.

JosJuice: We prefer enum classes over plain enums, which I suppose would exclude unnamed enums.

Source/Core/DolphinQt/Debugger/CodeViewWidget.cpp
```diff
@@ -51,6 +123,7 @@ CodeViewWidget::CodeViewWidget()
   horizontalHeader()->hide();
 
   setFont(Settings::Instance().GetDebugFont());
+  setItemDelegateForColumn(5, new BranchDisplayDelegate(this));
```
BhaaLseN:
```suggestion
  setItemDelegateForColumn(CODE_VIEW_COLUMN_BRANCH_ARROWS, new BranchDisplayDelegate(this));
```

Source/Core/DolphinQt/Debugger/CodeViewWidget.h
```diff
@@ -38,6 +44,10 @@ class CodeViewWidget : public QTableWidget
   void ToggleBreakpoint();
   void AddBreakpoint();
 
+  u32 AddressForRow(int row) const;
+
+  std::vector<CodeViewBranch> m_branches;
```
lioncash: This should likely remain private, and instead make `BranchDisplayDelegate` a friend of this class.

This would also allow for safely forward declaring `CodeViewBranch` as well, allowing the definition to be completely hidden.

Source/Core/DolphinQt/Debugger/CodeViewWidget.h
```diff
@@ -15,6 +15,12 @@ class QMouseEvent;
 class QResizeEvent;
 class QShowEvent;
 
+struct CodeViewBranch
+{
+  u32 m_src_addr;
```
lioncash: `m_` can be dropped here, given its just a data struct with no invariants.

Source/Core/DolphinQt/Debugger/CodeViewWidget.cpp
```diff
@@ -86,10 +159,18 @@ void CodeViewWidget::FontBasedSizing()
...
-  setColumnWidth(0, rowh + 5);
+  setColumnWidth(CODE_VIEW_COLUMN_BREAKPOINT, rowh + 5);
   Update();
 }
 
+u32 CodeViewWidget::AddressForRow(int row) const
+{
+  // m_address is defined as the center row of the table, so we have rowCount/2 instructions above
+  // it; an instruction is 4 bytes long on GC/Wii so we increment 4 bytes per row
+  const u32 rowZeroAddress = m_address - ((rowCount() / 2) * 4);
```
lioncash:
```suggestion
  const u32 row_zero_address = m_address - ((rowCount() / 2) * 4);
```

AdmiralCurtiss:
I've added a special case for branch instructions that set the link register to reduce visual noise, which turns this: https://user-images.githubusercontent.com/4522237/75113261-82c9fc00-564c-11ea-804d-61dca888cf76.png
into something much more readable-at-a-glance: https://user-images.githubusercontent.com/4522237/75202872-eb0fff00-576c-11ea-869c-0a9ca0b208ef.png

Since instructions that set the link register usually jump to another function, and thus very rarely actually have the target visible, this seems like a good idea to me.

I'll see if I can do something about using the horizontal space better, too.

Source/Core/DolphinQt/Debugger/CodeViewWidget.cpp
```diff
@@ -174,8 +282,15 @@ void CodeViewWidget::Update()
 
     if (hex_str.length() == VALID_BRANCH_LENGTH && desc != "---")
     {
-      description_item->setText(tr("--> %1").arg(QString::fromStdString(
-          PowerPC::debug_interface.GetDescription(GetBranchFromAddress(addr)))));
+      u32 branch_addr = GetBranchFromAddress(addr);
+      CodeViewBranch& branch = m_branches.emplace_back();
+      branch.src_addr = addr;
+      branch.dst_addr = branch_addr;
+      branch.is_link = ins.ends_with("l") || ins.ends_with("la") || ins.ends_with("l+") ||
```
jordan-woyak: `ends_with` is a C++20 feature. I have feeling at least one of our build bots won't like it.
Consider making a little `IsLink` function anyways to hide away these magic strings.

JosJuice: We have `StringEndsWith` in `StringUtil.h` if you want a replacement for `ends_with`.

AdmiralCurtiss:
Alright, I tried to use the horizontal space better, looks pretty good now if you ask me: https://user-images.githubusercontent.com/4522237/75385411-ff283d80-58df-11ea-9155-3db9a41efe2f.png

I'm not entirely sure I like how the indentation calculation code is written (in particular the lambda-in-lambda there...), but I suppose it does the job.

Source/Core/DolphinQt/Debugger/CodeViewWidget.cpp
```diff
@@ -189,25 +307,101 @@ void CodeViewWidget::Update()
...
+      continue;
+
+    const u32 arrow_first_visible_addr =
+        std::clamp(arrow_addr_lower, first_visible_addr, last_visible_addr);
+    const u32 arrow_last_visible_addr =
+        std::clamp(arrow_addr_higher, first_visible_addr, last_visible_addr);
+    const u32 arrow_first_visible_row = (arrow_first_visible_addr - first_visible_addr) / 4;
+    const u32 arrow_last_visible_row = (arrow_last_visible_addr - first_visible_addr) / 4;
+
+    const std::optional<u32> free_column = [&]() -> std::optional<u32> {
```
leoetlino:
```suggestion
    const std::optional<u32> free_column = [&] -> std::optional<u32> {
```

(and maybe use auto here, since the return type is already explicitly written out?)

Source/Core/DolphinQt/Debugger/CodeViewWidget.cpp
```diff
@@ -189,25 +307,101 @@ void CodeViewWidget::Update()
...
+      for (u32 column = 0; column < columns; ++column)
+      {
+        const bool column_is_free = [&]() {
```
leoetlino:
```suggestion
        const bool column_is_free = [&] {
```

Source/Core/DolphinQt/Debugger/CodeViewWidget.cpp
AdmiralCurtiss: The parens here seem to be mandatory when explicitly defining the return type. I'll use the `auto` though.

leoetlino:
>One thing I'm not sure here is if I have to make sure BranchDisplayDelegate::paint() and CodeViewWidget::Update() don't run at the same time in different threads, or if those two are both only already allowed on the GUI thread anyway. Does anyone know?

Looks like CodeViewWidget::Update is only called from the UI thread, so there shouldn't be any issues

PR #8643, opened 2020-02-23
InputCommon/DolphinQt: Mapping related race fixes and cleanups.

jordan-woyak:
Eliminate racy `UpdateInput` call within `DetectInput`.
(HotkeyManager/Core will already be doing this)

Split separate mapping indicator types into their own classes.

The "Point" indicator now uses the entire bounding box for drawing the "red dot".
The reduced box seemed to confuse some people into thinking it needed some kind of calibration.
image

Ensure the "state lock" is held when updating indicators.
Previously it was not held on a redraw invoked from Qt.

Cleaned up indicator code to eliminate some redundant logic and scalar multiplication all over the place.

Minor change to indicator colors for better contrast on some dark themes.
image

### DIFF
--- a/Source/Core/DolphinQt/Config/Mapping/MappingIndicator.cpp
+++ b/Source/Core/DolphinQt/Config/Mapping/MappingIndicator.cpp
@@ -105,14 +105,16 @@ void MappingIndicator::AdjustGateColor(QColor* color)
     color->setHsvF(color->hueF(), color->saturationF(), 1 - color->valueF());
 }
 
-MappingIndicator::MappingIndicator(ControllerEmu::ControlGroup* group) : m_group(group)
+MappingIndicator::MappingIndicator()
 {
   // TODO: Make these magic numbers less ugly.
-  int required_height = 106;
+  const int required_height = 106;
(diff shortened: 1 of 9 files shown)


Source/Core/DolphinQt/Config/Mapping/MappingIndicator.cpp
```diff
@@ -105,14 +105,16 @@ void MappingIndicator::AdjustGateColor(QColor* color)
...
 
-MappingIndicator::MappingIndicator(ControllerEmu::ControlGroup* group) : m_group(group)
+MappingIndicator::MappingIndicator()
 {
   // TODO: Make these magic numbers less ugly.
-  int required_height = 106;
-
-  if (group && ControllerEmu::GroupType::MixedTriggers == group->type)
-    required_height = 64 + 1;
+  const int required_height = 106;
```
BhaaLseN: Any reason you don't make this (and the ones further below) a `constexpr`?

jordan-woyak: I suppose I will. :)

PR #8644, opened 2020-02-23
DiscIO: Rework the implementation of TGC reading

JosJuice:
Fixes https://bugs.dolphin-emu.org/issues/10654.

To quote the documenation file included with the program tgctogcm:

> TGC's are miniaturized .gcm images with a 32kB header.
> The embedded gcm contains some bogus data, namely:
> -FST Location (0x424 in gcm)
> -DOL Location (0x420 in gcm)
> -FST File offsets (all files are offset/spoofed by a certain amount)

Dolphin has been handling the values at 0x420 and 0x424 by simply overwriting them with a working value (just like tgctogcm does), but it has used a different approach for the file offsets in the FST. Instead of changing the offsets that are stored in the FST, Dolphin changed where the files actually are placed on the virtual disc. My hope was that this would make the loading times more accurate to how they are when running a TGC file as part of a larger disc. However, there are TGC files where we would need to move files backwards on the disc in order to do this (this is what [issue 10654](https://bugs.dolphin-emu.org/issues/10654) is about), so the approach we have been using is flawed.

This change makes Dolphin overwrite offsets in the FST instead, like tgctogcm does. Other than making Dolphin handle the affected TGC files correctly, this change also makes it so that unnecessary padding data isn't written if you use Dolphin to convert a TGC file to an ISO file. ~This feature is not actually implemented in Dolphin as of now, but I'm planning to add it in the near future as part of a larger feature.~

### DIFF
--- a/Source/Core/DiscIO/TGCBlob.cpp
+++ b/Source/Core/DiscIO/TGCBlob.cpp
@@ -4,59 +4,43 @@
 
 #include "DiscIO/TGCBlob.h"
 
+#include <algorithm>
 #include <memory>
 #include <string>
+#include <type_traits>
 #include <utility>
+#include <vector>
 
 #include "Common/File.h"
 #include "Common/Swap.h"
 
 namespace
 {
-template <typename T>
-struct Interval
-{
-  T start;
-  T length;
-
-  T End() const { return start + length; }
-  bool IsEmpty() const { return length == 0; }
-};
-
-template <typename T>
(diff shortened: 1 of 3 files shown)


JMC47:
Has this been tested by anyone?  If not, what testing does it need?

PPLToast:
I can confirm this fixes playing a demo tgc that crashed unless converted to gcm. Same for extracting files that otherwise got corrupted.

JosJuice:
I did a similar amount of testing to PPLToast when I created the PR. I'm not aware of anything additional that needs to be tested, unless you have something in particular.

PR #8645, opened 2020-02-23
Add option to force texture filtering off

leo60228:
I saw this requested on Reddit. I thought it would look horrible, but I think I might actually prefer it to bilinear filtering.
Menu (ignore the titlebar, that's a bug with my screenshot tool):
Menu
Default:
Default
Force:
Force
None:
None

### DIFF
--- a/Source/Core/Common/StringUtil.h
+++ b/Source/Core/Common/StringUtil.h
@@ -14,7 +14,12 @@
 #include <type_traits>
 #include <vector>
 
+#include "Common/CommonFuncs.h"
 #include "Common/CommonTypes.h"
+
+#ifndef _WIN32
+#include <strings.h>
+#endif
 
 #ifdef _MSC_VER
 #include <filesystem>
@@ -59,6 +64,17 @@ template <typename T, std::enable_if_t<std::is_integral_v<T> || std::is_enum_v<T
 bool TryParse(const std::string& str, T* output, int base = 0)
 {
   char* end_ptr = nullptr;
+
+  if (!strcasecmp("true", str.c_str()))
+  {
+    *output = static_cast<T>(1);
+    return true;
+  }
+  else if (!strcasecmp("false", str.c_str()))
+  {
+    *output = static_cast<T>(0);
+    return true;
+  }
 
   // Set errno to a clean slate.
   errno = 0;
--- a/Source/Core/Core/Config/GraphicsSettings.cpp
+++ b/Source/Core/Core/Config/GraphicsSettings.cpp
@@ -102,8 +102,8 @@ const Info<bool> GFX_PREFER_GLES{{System::GFX, "Settings", "PreferGLES"}, false}
 
 // Graphics.Enhancements
 
-const Info<bool> GFX_ENHANCE_FORCE_FILTERING{{System::GFX, "Enhancements", "ForceFiltering"},
-                                             false};
+const Info<ForceFilteringMode> GFX_ENHANCE_FORCE_FILTERING{
+    {System::GFX, "Enhancements", "ForceFiltering"}, ForceFilteringMode::Default};
 const Info<int> GFX_ENHANCE_MAX_ANISOTROPY{{System::GFX, "Enhancements", "MaxAnisotropy"}, 0};
 const Info<std::string> GFX_ENHANCE_POST_SHADER{
     {System::GFX, "Enhancements", "PostProcessingShader"}, ""};
--- a/Source/Core/DolphinQt/Config/Graphics/EnhancementsWidget.cpp
+++ b/Source/Core/DolphinQt/Config/Graphics/EnhancementsWidget.cpp
@@ -77,11 +77,12 @@ void EnhancementsWidget::CreateWidgets()
 
   m_pp_effect = new ToolTipComboBox();
   m_configure_pp_effect = new QPushButton(tr("Configure"));
+  m_force_texture_filtering = new GraphicsChoice({tr("Default"), tr("Force Bilinear/Trilinear"),
+                                                  tr("Force Nearest Neighbor")},
+                                                 Config::GFX_ENHANCE_FORCE_FILTERING);
   m_scaled_efb_copy = new GraphicsBool(tr("Scaled EFB Copy"), Config::GFX_HACK_COPY_EFB_SCALED);
   m_per_pixel_lighting =
       new GraphicsBool(tr("Per-Pixel Lighting"), Config::GFX_ENABLE_PIXEL_LIGHTING);
-  m_force_texture_filtering =
-      new GraphicsBool(tr("Force Texture Filtering"), Config::GFX_ENHANCE_FORCE_FILTERING);
   m_widescreen_hack = new GraphicsBool(tr("Widescreen Hack"), Config::GFX_WIDESCREEN_HACK);
   m_disable_fog = new GraphicsBool(tr("Disable Fog"), Config::GFX_DISABLE_FOG);
   m_force_24bit_color =
@@ -102,14 +103,16 @@ void EnhancementsWidget::CreateWidgets()
   enhancements_layout->addWidget(m_pp_effect, 4, 1);
   enhancements_layout->addWidget(m_configure_pp_effect, 4, 2);
 
-  enhancements_layout->addWidget(m_scaled_efb_copy, 5, 0);
-  enhancements_layout->addWidget(m_per_pixel_lighting, 5, 1);
-  enhancements_layout->addWidget(m_force_texture_filtering, 6, 0);
-  enhancements_layout->addWidget(m_widescreen_hack, 6, 1);
-  enhancements_layout->addWidget(m_disable_fog, 7, 0);
-  enhancements_layout->addWidget(m_force_24bit_color, 7, 1);
-  enhancements_layout->addWidget(m_disable_copy_filter, 8, 0);
-  enhancements_layout->addWidget(m_arbitrary_mipmap_detection, 8, 1);
+  enhancements_layout->addWidget(new QLabel(tr("Texture Filtering:")), 5, 0);
+  enhancements_layout->addWidget(m_force_texture_filtering, 5, 1, 1, -1);
+
+  enhancements_layout->addWidget(m_scaled_efb_copy, 6, 0);
+  enhancements_layout->addWidget(m_per_pixel_lighting, 6, 1);
+  enhancements_layout->addWidget(m_widescreen_hack, 7, 0);
+  enhancements_layout->addWidget(m_disable_fog, 7, 1);
+  enhancements_layout->addWidget(m_force_24bit_color, 8, 0);
+  enhancements_layout->addWidget(m_disable_copy_filter, 8, 1);
+  enhancements_layout->addWidget(m_arbitrary_mipmap_detection, 9, 0);
 
   // Stereoscopy
   auto* stereoscopy_box = new QGroupBox(tr("Stereoscopy"));
@@ -354,8 +357,8 @@ void EnhancementsWidget::AddDescriptions()
   static const char TR_FORCE_TEXTURE_FILTERING_DESCRIPTION[] =
       QT_TR_NOOP("Filters all textures, including any that the game explicitly set as "
                  "unfiltered.<br><br>May improve quality of certain textures in some games, but "
-                 "will cause issues in others.<br><br><dolphin_emphasis>If unsure, leave this "
-                 "unchecked.</dolphin_emphasis>");
+                 "will cause issues in others.<br><br><dolphin_emphasis>If unsure, select "
+                 "Default.</dolphin_emphasis>");
   static const char TR_DISABLE_COPY_FILTER_DESCRIPTION[] = QT_TR_NOOP(
       "Disables the blending of adjacent rows when copying the EFB. This is known in "
       "some games as \"deflickering\" or \"smoothing\".<br><br>Disabling the filter has no "
@@ -383,6 +386,9 @@ void EnhancementsWidget::AddDescriptions()
   m_pp_effect->SetTitle(tr("Post-Processing Effect"));
   m_pp_effect->SetDescription(tr(TR_POSTPROCESSING_DESCRIPTION));
 
+  m_force_texture_filtering->SetTitle(tr("Texture Filtering"));
+  m_force_texture_filtering->SetDescription(tr(TR_FORCE_TEXTURE_FILTERING_DESCRIPTION));
+
   m_scaled_efb_copy->SetDescription(tr(TR_SCALED_EFB_COPY_DESCRIPTION));
 
   m_per_pixel_lighting->SetDescription(tr(TR_PER_PIXEL_LIGHTING_DESCRIPTION));
@@ -392,8 +398,6 @@ void EnhancementsWidget::AddDescriptions()
   m_disable_fog->SetDescription(tr(TR_REMOVE_FOG_DESCRIPTION));
 
   m_force_24bit_color->SetDescription(tr(TR_FORCE_24BIT_DESCRIPTION));
-
-  m_force_texture_filtering->SetDescription(tr(TR_FORCE_TEXTURE_FILTERING_DESCRIPTION));
 
   m_disable_copy_filter->SetDescription(tr(TR_DISABLE_COPY_FILTER_DESCRIPTION));
 
--- a/Source/Core/Core/Config/GraphicsSettings.h
+++ b/Source/Core/Core/Config/GraphicsSettings.h
@@ -9,6 +9,7 @@
 
 enum class AspectMode : int;
 enum class ShaderCompilationMode : int;
+enum class ForceFilteringMode : int;
 enum class StereoMode : int;
 enum class FreelookControlType : int;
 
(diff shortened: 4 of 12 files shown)


Tilka:
What happens with existing configs? Do we translate true/false to 1/0?

leo60228:
I wasn't sure how to implement (as in program, not design) that. That would be the best way of doing it, but for now it resets to Default.

Source/Core/Common/StringUtil.h
```diff
@@ -58,6 +58,11 @@ bool TryParse(const std::string& str, T* output)
 {
   char* end_ptr = nullptr;
 
+  if (str == "true")
+    return 1;
+  else if (str == "false")
+    return 0;
```
JosJuice: This function isn't supposed to return the parsed value, it's supposed to return whether the parse succeeded. The parsed value needs to be stored in the `output` parameter.

Also, the code you added only works when the string is all lowercase, which is a problem since Dolphin uses `True` and `False` when writing to INI files. It should be case insensitive to match `TryParse(const std::string&, bool*)`.

leo60228: Oops. I couldn't test since I was developing over ssh, will fix.

MayImilae:
I can't really agree with the UI and UX approach in this PR. By placing it in the enhancements tab, changing Force Texture Filtering to off is being elevated to the same level as extremely important options like internal resolution, anisotropic filtering, antialiasing, etc etc. This does not belong on that level. Setting texture filtering to force nearest neighbor (more on that later) is an _extremely_ niche option that does not belong on the same level as things users will want to keep configured and accessible.

Now you probably put it here because this is where Force Texture Filtering was. But that was a tiny checkmark button that took up very little real estate on the screen, in the midst of lots of other little buttons. Even though it is location in a prime location, it's mindshare impact is relatively minimal. However, this PR turns it into a dropdown, which takes up WAY more space and promotes is puts it right up in the same level as the critical options like internal resolution. This is really inappropriate for a niche feature.

Also I don't really think Force Texture Filtering belonged here in the first place. Outside of a handful of games where it genuinely improves things (New Super Mario Bros Wii, Skyward Sword, to name two), it basically does nothing but cause minor visual bugs. But it was small, in the midst of a number of other misc options, and had been there forever, so I never bothered to complain.

Anyway, I believe this should be moved to advanced, if we decide to include it at all.

-----

As for the naming of the options, that gets a little messy, as texture filtering is more or less a made up term from marketing that's stuck. Ok so, "texture filtering" the term is defined as the remapping algorithm the GPU uses to remap the texture's pixels to the screen's pixels during rendering. Unless a texture is 1:1 pixel tied to screen space (which you can do btw, screenspace huds are a thing!), its pixels HAVE to be remapped by something, and thus, filtered. That includes anything in 3D; scaling and rotation of textured polygons absolutely require texture filtering. Bilinear, Trilinear, etc etc are all algorithms to handle that remapping, but nearest neighbor is also a way to handle that remapping, it's just a more basic way of doing it. It sounds weird, but nearest neighbor is also texture filtering. The only time you get no texture filtering whatsoever is when the texture lives in screen space and is not scaled or manipulated in any way; but that is not nearest neighbor, that's a texture in screen space.

The confusion comes from marketing of very very early 3D technologies, where "texture filtering" was used as a marketing name for bilinear. Once trilinear came out they just called it the real name, trilinear filtering, but for a while marketing tried to posit that bilinear is "texture filtering", and nearest neighbor was not. But that's basically never the case, and we're kind of stuck with this not great name for it and that confusion to this day.

So, to that end, I believed the naming scheme used here is not correct. You currently have it set to:

*Default
*Force
*None

What that actually is, laid out accurately is:

*Texture Filtering of each texture is either trilinear filtering (or anisotropy if an aniso setting is enabled in Dolphin) or nearest neighbor as controlled by the game.
*Remove all game exclusions, forcing Bilinear or Trilinear (or anisotropy) for all textures
*Force Nearest Neighbor for all textures

How you describe those options doesn't anywhere near indicate what is going on. I'd much prefer something like this:

* Default
* Force Trilinear for all textures
* Force Nearest Neighbor for all textures.

With descriptors below in the Description area:

Default: Allows the game to set whatever texture filtering method they want for each texture. Anisotropy will replace Trilinear if configured.
Force Trilinear for all textures: Ignores game texture settings and forces Trilinear (or anisotropy if set) for all textures.
Forces Nearest Neighbor for all textures: Ignores game texture settings and forces nearest neighbor filtering for all textures. Disables anisotropy.

So about that trilinear there. So the GameCube has Trilinear filtering. However I'm not sure whether Bilinear or Trilinear is more common and how games can set and switch between them. I want to confirm with @stenzek about all of that and make sure the word(s) used there is correct.

mbc07:
> How you describe those options doesn't anywhere near indicate what is going on. I'd much prefer something like this:
> * Default
> * Force Trilinear for all textures
> * Force Nearest Neighbor for all textures.

I would drop that "for all textures", though. It would just make the drop-down list even larger especially when you account for translations and I think the suggested description text would already explain better what is going on. I think "Default", "Force Trilinear" (or whatever it should be) and "Force Nearest Neighbor" should be enough...

MayImilae:
That would be fine with me, honestly. The description can carry a bit more of the load in describing how they work, and it's not wrong at least.

As for Bilinear vs Trilinear, it's both. According to stenzek, apparently mipmapped things are trilinear, and everything else is bilinear (except for textures explicitly set as nearest neighbor). However, most things are mipmapped so, trilinear is more common, from my understanding at least.

JMC47:
@dolphin-emu-bot rebuild

JMC47:
So, disabling texture filtering does make a lot of games look nicer at higher resolutions, particularly games that have precision issues with text.

I do think the best solution would have a heuristic to try to detect HUD/UI stuff, and allow users to force it, use the heuristic, or disable it.

leo60228:
@JMC47 Do you have any ideas for how such a heuristic would work?

leo60228:
I've tried a few different ideas (contrast, attempt to detect a background color, etc.) and my best attempt was based on whether the texture has any transparent pixels, which... *shouldn't* be great, but using the UI and environment folders from Bighead's Mario Galaxy texture pack (it was the only png textures I had on my computer) it correctly classifies all but a few environment textures and ~90% of UI textures.

leo60228:
As it turns out, eyeballing it didn't work that great, so I calculated it properly. It's correct for 89.4% of environment textures and 98.9% of UI textures when treating "transparent" as "any transparency," and 91.0% of environment textures and 90.2% of UI textures when treating "transparent" as "full transparency."

leo60228:
I'm not really convinced that a heuristic that works for even 75% of games is *possible*, especially without machine learning or something similar.

JMC47:
Yeah, I don't know if a Heuristic would work, but the fact that this removes visual garbage around text (See: Starfox Assault) at higher resolutions makes it kind of interesting to experiment with.

leo60228:
So this overall algorithm seems to work very well but needs some tuning:
1. Count the number of pixels of every brightness
2. Iterate over all 2-element windows in that count
3. If the absolute difference of the two elements is over *a*/*b* of the larger element, increment a counter
4. If the counter is between *c* and *d*, then use nearest neighbor scaling, otherwise use the standard scaling algorithm
*a*, *b*, *c*, and *d* need tuning, but currently I'm using 15, 16, 1, and 50. The current values of *a* and *b* successfully filter out bell curves (the vast majority of environment textures in my experience), but an approximately equal percentage of environment and UI textures have a single edge, making choosing *c* difficult. *d* exists solely because of displacement maps.

Mizoxman:
I recall playing a game years ago that had a "trilinear nearest" option for texture filtering, so it used bilinear/trilinear when the texture was scaled down, but when you got close enough to it for it to be scaled larger than its largest mipmap it would be nearest-neighbor scaled. Maybe something like that would work?

Tenome:
Nice feature, hope it gets merged.

iMagic16:
any update on this feature? would love to see this usable!

JMC47:
It would need to be rebased, fixed for Lint, and reviewed in order to be considered.  I'm honestly pretty happy with it though as a feature.

leo60228:
Remembered this PR and I'm happy to rebase it. I'm still not entirely sure about heuristics, though, especially since I'm not really sure what infrastructure Dolphin has for image processing.

Phantop:
Ooh I was about to try to rebase this myself (but lack the git skills to do so). Really hope this gets merged; in particular, Animal Crossing looks significantly better with this.

Source/Core/Common/StringUtil.h
```diff
@@ -60,6 +65,17 @@ bool TryParse(const std::string& str, T* output, int base = 0)
...
+  {
+    *output = static_cast<T>(1);
+    return true;
+  }
+  else if (!strcasecmp("false", str.c_str()))
+  {
+    *output = static_cast<T>(0);
+    return true;
+  }
+
```
AdmiralCurtiss, line 78: ymmv but I feel this is pretty ugly to have in the middle of what is supposed to be a generic integer parsing function. At the very least I'd pass a flag that defaults to false whether true/false should convert to valid integers.

Source/Core/Core/Config/GraphicsSettings.cpp
```diff
@@ -102,8 +102,7 @@ const Info<bool> GFX_PREFER_GLES{{System::GFX, "Settings", "PreferGLES"}, false}
 
 // Graphics.Enhancements
 
-const Info<bool> GFX_ENHANCE_FORCE_FILTERING{{System::GFX, "Enhancements", "ForceFiltering"},
-                                             false};
+const Info<int> GFX_ENHANCE_FORCE_FILTERING{{System::GFX, "Enhancements", "ForceFiltering"}, 0};
```
AdmiralCurtiss, line 105: This should probably use an enum instead for clarity, like eg. the ShaderCompilationMode a few lines up.

Source/Core/DolphinQt/Config/Graphics/EnhancementsWidget.cpp
```diff
@@ -77,11 +77,11 @@ void EnhancementsWidget::CreateWidgets()
 
   m_pp_effect = new ToolTipComboBox();
   m_configure_pp_effect = new QPushButton(tr("Configure"));
+  m_force_texture_filtering = new GraphicsChoice({tr("Default"), tr("Force"), tr("None")},
+                                                 Config::GFX_ENHANCE_FORCE_FILTERING);
```
AdmiralCurtiss, line 81: MayImilae already commented on this but I also think the names used here are pretty poor.

AdmiralCurtiss (changes requested):

Source/Core/Common/StringUtil.h
```diff
@@ -14,8 +14,13 @@
 #include <type_traits>
 #include <vector>
 
+#include "Common/CommonFuncs.h"
 #include "Common/CommonTypes.h"
 
+#ifndef _WIN32
+#include <strings.h>
+#endif
```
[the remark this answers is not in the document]
Pokechu22: It's *not* included on Windows (since it's #if<ins><b>n</b></ins>def).  Also note that it's string<ins><b>s</b></ins>.h, not string.h; [strings.h defines non-standard functions such as strcasecmp](https://stackoverflow.com/a/4291328).

It works on Windows because `CommonFuncs.h` provides a `#define strcasecmp _stricmp`.

Source/Core/DolphinQt/Config/Graphics/EnhancementsWidget.cpp
Phantop: Yeah. I personally have no issue with the current names but changing to "Force Trilinear/Nearest Neighbor" seems to be the nicest way of doing this (in my opinion).

iMagic16:
any update on this feature? would love to see this in main! The fork works fantastic albeit confusing wording.

Source/Core/DolphinQt/Config/Graphics/EnhancementsWidget.cpp
SpongeeJumper: For what it's worth "Default", "Force Trilinear/Bilinear", and "Force Nearest Neighbor" would make sense to me.

SpongeeJumper:
I just want to add my raised hand to the "please merge this feature" crowd. I even tried to compile the fork myself but troubleshooting git and visual studio to get it to actually build proved beyond my capabilities. I want pixelated Animal Crossing textures!

I played the fan-translation of the N64 version a little bit, with a graphics plugin (Glide64mk2) that has a nearest neighbor filter option (although they call it "force point sampled", which might be more correct IDK), and I can confirm that it (subjectively) looks much better that way. If the translated rom hack hadn't kept crashing on me I might've stuck with it despite the lack of added features from the Gamecube version.

JMC47:
@dolphin-emu-bot rebuild

leo60228:
(probably should've built the rebased version before pushing, oops. still going to address the other issues)

JMC47:
Make sure you combine fixup commits into other commits, so that the only commits that exist are the main changes that you want split up for easier review.

I do think this would be a good change, as there are games that are broken *even on console* due to using the wrong filtering mode.  See:  Blaster Master WiiWare, and Sexy Poker for easy examples.

Source/Core/Common/StringUtil.h
leo60228: I tried to do that, but I felt like the changes necessary were a bit invasive.

AdmiralCurtiss:
I'm still against the `true`/`false` parsing to `1`/`0` in the generic integer parsing function -- I'd rather just make the setting not backwards-compatible with the old config. Can I get some input on that from other devs?

Otherwise this looks fine to me. The setting description might need to be reworded.

iwubcode:
> I'm still against the true/false parsing to 1/0 in the generic integer parsing function

I agree with you.  However, I'm going to go a step further, I think I'd prefer breaking config entirely:

```
const Info<bool> GFX_ENHANCE_FORCE_FILTERING{{System::GFX, "Enhancements", "ForceFiltering"}, false};
```

to

```
const Info<ForceFilteringMode> GFX_ENHANCE_TEX_FILTERING_TYPE{{System::GFX, "Enhancements", "TextureFilteringType"}, TextureFilteringMode::Default};
```

MayImilae:
I'm still not happy about this being in the enhanced tab. It really should be in hacks or advanced imo.

Pokechu22:
IMO enhancements is the proper tab for it (for the same reason why forcing texture filtering is there and forcing anisotropic filtering is there (compare #9975)).  Most of the arguments I can think of for this belonging in the hacks tab would also apply to the other enhancements; particularly disable fog, disable copy filter, per-pixel lighting, force 24-bit color, and widescreen hack.

MayImilae:
Well it's kind of a mixture. Force texture filtering is an enhancement (extrapolating from how the console works), while disabling texture filtering is hack (that's not how the console works at all). The reason I think the whole selection should be in hacks is because of the enlarged real estate that the dropdown has puts more emphasis on it.

Pokechu22:
Forcing texture filtering is probably more likely to break things (at least, it is known to break FMVs using Bink Video (on the Wii), and I think also EA VP6), which depend on getting a specific pixel's value without interpolation.  I'm not sure if force-disabling it breaks anything (it probably makes some indirect texture effects look worse such as the wind waker map and the color filtering in EA Sports Active, but I don't think they'd completely break).

I feel like the hacks tab is more for things that increase performance at the cost of accuracy, while the enhancements tab is for things that adjust rendering to one's preference at the cost of accuracy (regardless of the hackyness of the implementation).  Both tabs currently have about the same amount of stuff vertically, and putting it in the hacks tab would either require putting it in a new section (increasing the height of hacks even more) or putting it in the "other" section (which seems suboptimal).

JMC47:
Regardless of personal preference, most users will use this like an enhancement.

Phantop:
> I'm not sure if force-disabling it breaks anything (it probably makes some indirect texture effects look worse such as the wind waker map and the color filtering in EA Sports Active, but I don't think they'd completely break).

Just want to note that forcing nearest neighbor does introduce some issues in certain scenarios. For example, Animal Crossing City Folk's grass does not render correctly. I do still think this is better as an enhancement regardless.

AdmiralCurtiss:
> Well it's kind of a mixture. Force texture filtering is an enhancement (extrapolating from how the console works), while disabling texture filtering is hack (that's not how the console works at all). The reason I think the whole selection should be in hacks is because of the enlarged real estate that the dropdown has puts more emphasis on it.

I don't understand why one of the two should be considered a 'hack' and the other not. As I understand it this is just a flag on a texture that the game can set to 'yes' or 'no' in each rendering instance, and all this option does is force that flag to be either of the two instead of what the game intended to set. How is one way different than the other?

Avasam:
Just my two cents as a user, is that
> I feel like the hacks tab is more for things that increase performance at the cost of accuracy, while the enhancements tab is for things that adjust rendering to one's preference at the cost of accuracy

is how I feel. I would be looking in the "Enhancements" tab for something may improve the looks of a game depending on my personal preferences. Especially since it has to do with texture and scaling, it fits really well with Internal Resolution, AA, AF, etc.

--- 
> As for the naming of the options, [...]

Really interesting post about marketing naming btw.

Phantop:
What's the status on the possibility of this being merged? I would be willing to step up and make changes that would be necessary to do so. This feature is very useful for Animal Crossing in particular, and I'd love to have it end up in Dolphin properly.

iwubcode:
Well first of all, this PR would have to be rebased and its conflicts resolved.

If we are attempting to target specific elements,  just personally I think this would be better as  a Graphics Mod but that hasn't been merged yet either.

AdmiralCurtiss:
-> #11276

PR #8646, opened 2020-02-24
[WIP] Externals: Replace libraries with Git submodules

AlexApps99:
Yes you heard that right, I'm working on replacing the libraries in Externals with Git submodules.
Feedback, recommendations for downgrading libraries, and buildsystem help would all be appreciated.

It may take me a while to finish this, but I believe that it's a step in the right direction that will make updating the libraries easier and more convenient to manage.

## Todo:
- [x] Remove old libraries
- [x] Replace with submodules
- [ ] Coax CMake into building again
- [ ] Attempt to get rid of binaries in Externals (`gettext`, `MoltenVK`)


i30817:
Did you evaluate the other 2 alternatives in this space, git subrepo and git subtree? Just curious.

AlexApps99:
Git subrepo is unofficial so I did not evaluate it, it would be very difficult to persuade developers to install an external tool just to checkout code.
Git subtree would drastically increase the filesize of this git repo, as it would have to embed the history of all the externals instead of pointing to the remote url itself

spycrab:
The biggest hurdle atm is probably that we need a vcxproj file for all of our Externals on Windows at the moment.

AlexApps99:
Why is that? (I don't use Windows)

spycrab:
We haven't fully switched over to CMake yet. mainly the buildbot still needs to be migrated.

AlexApps99:
That's a shame
I guess this PR will have to be put on hold until that's finished

Tilka:
Can't the vcxproj files live directly in Externals/ while they're still necessary?

AlexApps99:
That does seem possible, I'll look into it tomorrow while I try to get it building

spycrab:
I think that would be rather messy though, to have two versions to worry about? The main benefits I see here vanish once you have to maintain both of these tbh :/. It might be okay temporarily though.

AlexApps99:
How soon would the CMake switchover be completed?

spycrab:
I'm not sure, I will probably migrate the build instructions for Windows this week. When the buildbot will be migrated? Not sure.

stenzek:
While the idea is sound, I don't think it works particularly well for Dolphin. Qt as a submodule is okay, since it's only for Windows.

Some issues I can with using submodules are below. I think delroth has already mentioned some of them before, but I may have missed some.

 - It's annoying when cloning and takes longer
 - Not all externals are required depending on how you're building, which means you're cloning a bunch of potentially unnecessary trees
 - Some of the externals contain modifications from upstream code. Using the upstream repositories prevents us from keeping these patches in-tree, potentially requiring ugly workarounds
 - More importantly, if the upstream git repository disappears, new clones break
 - Creates issues when bisecting, both if the upstream repository changes, and having to delete/re-checkout the submodule when the bisect cuts through the PR which introduced them

leo60228:
> Git subrepo is unofficial so I did not evaluate it, it would be very difficult to persuade developers to install an external tool just to checkout code.
The entire point of git-subrepo is that this isn't necessary...

AlexApps99:
> > Git subrepo is unofficial so I did not evaluate it, it would be very difficult to persuade developers to install an external tool just to checkout code.
> 
> The entire point of git-subrepo is that this isn't necessary...

@leo60228 What do you mean? Git subrepo isn't a part of Git (https://github.com/ingydotnet/git-subrepo), so how would it not be necessary?

@stenzek those are good points you make, I'll need to look into filtering which submodules are cloned to see whether this is worthwhile

leo60228:
You *only* need git-subrepo if you want to update the subrepo or push commits in it to upstream.

AlexApps99:
My point still stands, why would the developers prefer an external tool over Git's built in solution?

leo60228:
Only a very small minority of developers would need it, and git submodules are absolutely *horrible* for developers and users.

AlexApps99:
I'll look into it some more

fantesykikachu:
Another option would be to use cmake's ExternalProject and FetchContent functions. This would allow for only fetching modules that are required for a given build. It does however require the full transition to cmake first. It can also be written so as to not use submodules which is preferable for me because I use git without bash.

orbea:
At least on linux I think it would be really unfortunate to rely on git submodules when it never has before. This can increase the size of shallow clones significantly and has several other complications.

AlexApps99:
These are all good points, I will close this until the time comes for a restructuring of Externals.

PR #8650, opened 2020-02-25
HW/WiimoteEmu: Camera logic cleanups.

jordan-woyak:
Added Get/Setters. Added Comments. Renamed some variables.

Added more checks for camera being in a valid state.
Send 0xFF-filled data in more situations.

This fixes the flash of poorly positioned IR points when changing the sensor bar sensitivity in the Wii menu.

### DIFF
--- a/Source/Core/Core/HW/WiimoteEmu/Camera.cpp
+++ b/Source/Core/Core/HW/WiimoteEmu/Camera.cpp
@@ -19,14 +19,14 @@ namespace WiimoteEmu
 {
 void CameraLogic::Reset()
 {
-  reg_data = {};
+  m_reg_data = {};
 
   m_is_enabled = false;
 }
 
 void CameraLogic::DoState(PointerWrap& p)
 {
-  p.Do(reg_data);
+  p.Do(m_reg_data);
 
   // FYI: m_is_enabled is handled elsewhere.
 }
@@ -39,7 +39,7 @@ int CameraLogic::BusRead(u8 slave_addr, u8 addr, int count, u8* data_out)
   if (!m_is_enabled)
     return 0;
 
-  return RawRead(&reg_data, addr, count, data_out);
+  return RawRead(&m_reg_data, addr, count, data_out);
 }
 
 int CameraLogic::BusWrite(u8 slave_addr, u8 addr, int count, const u8* data_in)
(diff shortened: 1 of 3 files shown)


Source/Core/Core/HW/WiimoteEmu/WiimoteEmu.cpp
```diff
@@ -516,8 +516,15 @@ void Wiimote::SendDataReport()
...
-      m_i2c_bus.BusRead(CameraLogic::I2C_ADDR, camera_data_offset, rpt_builder.GetIRDataSize(),
-                        rpt_builder.GetIRDataPtr());
+      u8* ir_data = rpt_builder.GetIRDataPtr();
+      const u8 ir_size = rpt_builder.GetIRDataSize();
+
+      if (ir_size != m_i2c_bus.BusRead(CameraLogic::I2C_ADDR, camera_data_offset, ir_size, ir_data))
+      {
+        // This happens when IR reporting is enabled but the camera hardware is disabled.
+        // It commonly occurs when changing IR sensitivity.
+        std::fill_n(ir_data, ir_size, 0xff);
```
lioncash:
```suggestion
        std::fill_n(ir_data, ir_size, u8{0xff});
```

Otherwise this usually can't be optimized if the compiler would otherwise do so, e.g. by [using memset](https://github.com/gcc-mirror/gcc/blob/master/libstdc%2B%2B-v3/include/bits/stl_algobase.h#L910) or something else (this is also technically implicitly truncating an int to a u8 by itself anyways)

Source/Core/Core/HW/WiimoteEmu/Camera.cpp
```diff
@@ -86,141 +101,100 @@ void CameraLogic::Update(const Common::Matrix44& transform)
...
   {
-    u16 x;
-    u16 y;
+    IRBasic::IRObject position;
     u8 size;
   };
 
   // 0xFFFFs are interpreted as "not visible".
-  constexpr CameraPoint INVISIBLE_POINT{0xffff, 0xffff, 0xff};
+  const CameraPoint INVISIBLE_POINT = {{0xffff, 0xffff}, 0xff};
```
lioncash: Perhaps changes should be made to keep this constexpr, otherwise technically a static runtime constructor may need to be emitted by compilers here.

jordan-woyak: `Common::TVec2` isn't currently `constexpr`. I have some local changes `constexpr`ing it so I won't change that now.
I'll just eliminate `INVISIBLE_POINT`. It's only used in one place.

PR #8652, opened 2020-02-26
Qt: Use Builtin QKeySequence Shortcuts

sonic2kk:
There are a number of shortcuts built into Qt within the `QKeySequence` class. These have shortcuts for a number of different platforms and for Linux, respective desktop environments. This PR replaces the manually set keys with these builtins for opening, finding and quitting. It also adds a shortcut for opening the Configuration pane, using `QKeySequence::Preferences`.

### DIFF
--- a/Source/Core/DolphinQt/MenuBar.cpp
+++ b/Source/Core/DolphinQt/MenuBar.cpp
@@ -493,7 +493,8 @@ void MenuBar::AddViewMenu()
 void MenuBar::AddOptionsMenu()
 {
   QMenu* options_menu = addMenu(tr("&Options"));
-  options_menu->addAction(tr("Co&nfiguration"), this, &MenuBar::Configure);
+  options_menu->addAction(tr("Co&nfiguration"), this, &MenuBar::Configure,
+                          QKeySequence::Preferences);
   options_menu->addSeparator();
   options_menu->addAction(tr("&Graphics Settings"), this, &MenuBar::ConfigureGraphics);
   options_menu->addAction(tr("&Audio Settings"), this, &MenuBar::ConfigureAudio);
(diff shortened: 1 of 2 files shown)


Source/Core/DolphinQt/MenuBar.cpp
```diff
@@ -212,7 +212,8 @@ void MenuBar::AddFileMenu()
   file_menu->addSeparator();
 
   m_exit_action =
-      file_menu->addAction(tr("E&xit"), this, &MenuBar::Exit, QKeySequence(Qt::ALT + Qt::Key_F4));
+      file_menu->addAction(tr("E&xit"), this, &MenuBar::Exit);
+  m_exit_action->setShortcuts({ QKeySequence::Quit, QKeySequence(Qt::ALT + Qt::Key_F4) });
```
BhaaLseN: Any reason you didn't use `QKeySequence::Quit` for this while you were at it?

sonic2kk: It is in use, it uses both `QKeySequence::Quit` and the existing Alt+F4 shortcut. Granted, on my system (Arch + KDE Plasma) Alt+F4 still works anyway. Unless I'm misunderstanding?

BhaaLseN: Well, yeah, missed the word "_just_" in there. But now that you mention it, you're right; macOS uses Command+Q instead of Alt+F4 (or something like that).

Can probably leave it like that, forget that I said anything.

sonic2kk: macOS does indeed use Command+Q, which this adds support for. It's also common practice for Qt applications to use Ctrl+Q. I'm not sure what it's like on Windows with non-Qt applications, but on Linux a heck of a lot are starting to implement this, both Gtk and Qt alike.

PR #8654, opened 2020-02-27
Qt: Fix Hotkey Controller Profile display with boxes for each Wiimote

3t13nn3:
According to https://bugs.dolphin-emu.org/issues/11646 the Hotkey Controller Profile GUI needed to have an overhaul. 
Then @imStudd and I decided to solve this issue by putting theses options over boxes for each Wiimote. Separating Wiimotes looks way better and is more intuitive than displaying all Wiimotes options into one column.

Here is the old display:
Hotkeys Controller Profile

That is the new one:
Screenshot_2020-02-26_22-56-42


MayImilae:
Design wise, this is a nice little improvement!

CookiePLMonster:
@dolphin-emu-bot rebuild

Source/Core/DolphinQt/Config/Mapping/HotkeyControllerProfile.cpp
```diff
@@ -16,10 +16,27 @@ HotkeyControllerProfile::HotkeyControllerProfile(MappingWindow* window) : Mappin
...
-      tr("Controller Profile"), HotkeyManagerEmu::GetHotkeyGroup(HKGP_CONTROLLER_PROFILE)));
+  m_main_layout = new QGridLayout;
+
+  m_main_layout->addWidget(
+      CreateGroupBox(tr("Wii Remote %1").arg(1), 
+                     HotkeyManagerEmu::GetHotkeyGroup(HKGP_CONTROLLER_PROFILE_1)),
+      0, 0);
+
+  m_main_layout->addWidget(
+      CreateGroupBox(tr("Wii Remote %2").arg(2),
```
jordan-woyak: Perhaps this was a find/replace mistake. `%2` is probably meant to be `%1`.
Same problem below with 3 and 4.

Source/Core/DolphinQt/Config/Mapping/HotkeyControllerProfile.h
```diff
@@ -22,5 +22,5 @@ class HotkeyControllerProfile final : public MappingWidget
   void CreateMainLayout();
 
   // Main
-  QHBoxLayout* m_main_layout;
+  QGridLayout* m_main_layout;
```
jordan-woyak: It looks like this doesn't need to be a member variable as it's only used in `CreateMainLayout`.
You can just declare a local variable there and completely kill the above forward-declaration and include.
I realize you didn't add this member but now is a good time to fix it. :P

Source/Core/DolphinQt/Config/Mapping/HotkeyControllerProfile.cpp
3t13nn3: Then, I guess I have to switch 1 & 2, and 3 & 4 ?

Source/Core/DolphinQt/Config/Mapping/HotkeyControllerProfile.h
3t13nn3: OK, will change it to a local variable.

Source/Core/DolphinQt/Config/Mapping/HotkeyControllerProfile.cpp
jordan-woyak: Hmm? `%1` is the format string placeholder for the first value supplied to the `.arg` function.

3t13nn3: Oh sure.

Source/Core/DolphinQt/Config/Mapping/HotkeyControllerProfile.cpp
```diff
@@ -16,10 +16,27 @@ HotkeyControllerProfile::HotkeyControllerProfile(MappingWindow* window) : Mappin
...
+                     HotkeyManagerEmu::GetHotkeyGroup(HKGP_CONTROLLER_PROFILE_2)),
+      0, 1);
+
+  m_main_layout->addWidget(
+      CreateGroupBox(tr("Wii Remote %3").arg(3),
+                     HotkeyManagerEmu::GetHotkeyGroup(HKGP_CONTROLLER_PROFILE_3)),
+      1, 0);
+
+  m_main_layout->addWidget(
+      CreateGroupBox(tr("Wii Remote %4").arg(4),
```
AdmiralCurtiss: Are you sure this is the right way to format these? I'd assume the %1 is a placeholder for the argument and then the .arg() fills in the argument, so all of these should read "Wii Remote %1". Or just use literal strings and get rid of the .arg().

3t13nn3: As @jordan-woyak said upper, it must be 1% for the string format.

Source/Core/DolphinQt/Config/Mapping/HotkeyControllerProfile.cpp
```diff
@@ -16,10 +16,28 @@ HotkeyControllerProfile::HotkeyControllerProfile(MappingWindow* window) : Mappin
 
 void HotkeyControllerProfile::CreateMainLayout()
 {
-  m_main_layout = new QHBoxLayout();
-
-  m_main_layout->addWidget(CreateGroupBox(
-      tr("Controller Profile"), HotkeyManagerEmu::GetHotkeyGroup(HKGP_CONTROLLER_PROFILE)));
+  QGridLayout* m_main_layout;
+  m_main_layout = new QGridLayout;
```
AdmiralCurtiss:
```suggestion
  QGridLayout* m_main_layout = new QGridLayout;
```

3t13nn3: Done.

AdmiralCurtiss:
@dolphin-emu-bot rebuild

Source/Core/DolphinQt/Config/Mapping/HotkeyControllerProfile.h
```diff
@@ -6,7 +6,7 @@
 
 #include "DolphinQt/Config/Mapping/MappingWidget.h"
 
-class QHBoxLayout;
+class QGridLayout;
```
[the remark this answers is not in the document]
3t13nn3: Done.

Source/Core/Core/HotkeyManager.cpp
```diff
@@ -83,22 +83,22 @@ constexpr std::array<const char*, 133> s_hotkey_labels{{
...
-    _trans("Next Game Profile for Wii Remote 4"),
-    _trans("Previous Game Profile for Wii Remote 4"),
+    _trans("Next Profile"),
+    _trans("Previous Profile"),
+    _trans("Next Game Profile"),
+    _trans("Previous Game Profile"),
+    _trans("Next Profile "),
+    _trans("Previous Profile "),
+    _trans("Next Game Profile "),
+    _trans("Previous Game Profile "),
```
AdmiralCurtiss: Just noticed one more thing, there's extra spaces at the end of these strings. You can even see that in your screenshot.

Maybe also clean up your commit message while you're at it, all those extra fix lines are really not necessary.

3t13nn3: Done.

Source/Core/DolphinQt/Config/Mapping/HotkeyControllerProfile.cpp
```diff
@@ -16,10 +16,27 @@ HotkeyControllerProfile::HotkeyControllerProfile(MappingWindow* window) : Mappin
...
 void HotkeyControllerProfile::CreateMainLayout()
 {
-  m_main_layout = new QHBoxLayout();
-
-  m_main_layout->addWidget(CreateGroupBox(
-      tr("Controller Profile"), HotkeyManagerEmu::GetHotkeyGroup(HKGP_CONTROLLER_PROFILE)));
+  QGridLayout* m_main_layout = new QGridLayout;
+
+  m_main_layout->addWidget(
+      CreateGroupBox(tr("Wii Remote %1").arg(1),
```
Techjar: At this point using .arg() is superfluous, we should just put the number directly in the string since it's a compile-time constant. This would also allow translators to handle the numbers appropriately for their target language.

Source/Core/DolphinQt/Config/Mapping/HotkeyControllerProfile.cpp
```diff
@@ -16,10 +16,27 @@ HotkeyControllerProfile::HotkeyControllerProfile(MappingWindow* window) : Mappin
 
 void HotkeyControllerProfile::CreateMainLayout()
 {
-  m_main_layout = new QHBoxLayout();
-
-  m_main_layout->addWidget(CreateGroupBox(
-      tr("Controller Profile"), HotkeyManagerEmu::GetHotkeyGroup(HKGP_CONTROLLER_PROFILE)));
+  QGridLayout* m_main_layout = new QGridLayout;
```
jordan-woyak: Please remove m_ from this variable name.
You can also declare the type as const auto.

3t13nn3: Done.

3t13nn3:
So actually, my teammate and I have discover that naming many options with same names poses a conflict problem. In fact, assigned value are mixed once we quit and restart Dolphin.
Furthermore, we found in ```~/.config/dolphin-emu/Hotkeys.ini``` (the file that stock our inputs) that stocking depends on the name of the option.
So to avoid this, it seems we have to name each option differently.

https://github.com/dolphin-emu/dolphin/pull/8654/commits/1303a853af6dd671d2f300e50b271a06c6111487#diff-8d3cfd109d3bb0eeb1cd0e968d40d086

We were using ```Next Profile``` as the next profile string but we need to change it. For that, we thought about:
- return to ```Next Profile for Wii Remote 1```
- change to ```Next Profile 1```
- change to ```1 Next Profile```

Give us your opinion, so we can change.

If you find an other solution to avoid renaming each options differently, tell us !

Techjar:
The config key and the text label being tied to the same string seems like a fundamental design problem. Perhaps we should hold off on this change and fix that first?

Source/Core/DolphinQt/Config/Mapping/HotkeyControllerProfile.cpp
jordan-woyak: I can see the advantage of using the same string so it only has to be translated once.
I would be fine with either solution.

Techjar:
#8655 should address the name conflict issue.

3t13nn3:
Nice. We are waiting for you changes to be commited and we are done with this PR.

jordan-woyak:
Do we care about breaking existing configs?

Techjar:
I don't think these hotkeys are that commonly used, and they don't have any default bindings, so in this case it's probably okay.

Techjar:
This needs to be rebased now that #8655 is merged.

3t13nn3:
Update our PR, changes are ok.

BhaaLseN:
Something went wrong there; you have a lot of changes in there that don't seem to belong. Can you please take a look?

3t13nn3:
> Something went wrong there; you have a lot of changes in there that don't seem to belong. Can you please take a look?

Seems better now

3t13nn3:
Seems better now.

JMC47:
Seems to look fine to me.

AdmiralCurtiss:
Commits should be squashed though.

PR #8655, opened 2020-02-29
Core/HotkeyManager: Fix group names in config

Techjar:
For some reason every button group was shoved into the same config group called "Keys" which caused buggy behavior once someone tried to have hotkeys with the same name in different groups. This fixes that, and converts existing configs to the new group names so they don't get reset.

### DIFF
--- a/Source/Core/Core/HotkeyManager.cpp
+++ b/Source/Core/Core/HotkeyManager.cpp
@@ -13,6 +13,8 @@
 
 #include "Common/Common.h"
 #include "Common/CommonTypes.h"
+#include "Common/FileUtil.h"
+#include "Common/IniFile.h"
 #include "Common/StringUtil.h"
 
 #include "InputCommon/ControllerEmu/Control/Input.h"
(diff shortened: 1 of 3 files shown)


Source/Core/Core/HotkeyManager.cpp
```diff
@@ -437,3 +437,24 @@ void HotkeyManager::LoadDefaults(const ControllerInterface& ciface)
   set_key_expression(HK_UNDO_LOAD_STATE, NON + " & `F12`");
   set_key_expression(HK_UNDO_SAVE_STATE, SHIFT + " & `F12`");
 }
+
+// This override exists to load the old "Keys" group so pre-existing configs don't break.
```
lioncash: This and the TODO should likely go into the header, so it's more visible when looking at the interface.

LGTM otherwise.

Techjar: I'm in the process of reworking this PR, it's currently broken.

Source/Core/Core/HotkeyManager.cpp
```diff
@@ -246,6 +285,7 @@ void Initialize()
 
   // load the saved controller config
   s_config.LoadConfig(true);
```
jordan-woyak: Looks like you could change these 2 lines to a call to `LoadConfig`.

Techjar: Oh yeah, silly me.

Source/Core/Core/HotkeyManager.cpp
```diff
@@ -237,6 +239,43 @@ bool IsPressed(int id, bool held)
   return false;
 }
 
+// This function exists to load the old "Keys" group so pre-existing configs don't break.
+// TODO: Remove this at a future date when we're confident most configs are migrated.
+static void LoadLegacyConfig(ControllerEmu::EmulatedController* controller)
```
iwubcode: Can we / should we remove the old section from the configuration?

Source/Core/Core/HotkeyManager.cpp
```diff
@@ -237,6 +239,43 @@ bool IsPressed(int id, bool held)
...
+    {
+      auto sec = inifile.GetOrCreateSection("Hotkeys1");
+
+      {
+        std::string defdev;
+        sec->Get("Device", &defdev, "");
+        controller->SetDefaultDevice(defdev);
+      }
+
+      for (auto& cg : controller->groups)
```
leoetlino: nitpick: `group` rather than `cg`? and `control` instead of `c`?

Source/Core/Core/HotkeyManager.cpp
Techjar: No I think we should leave it there to keep the config functional on old Dolphin versions.

PR #8657, opened 2020-03-01
Qt: Tweak Toolbar Text

sonic2kk:
Currently the Toolbar has some unnecessary abbreviations, which I believe are holdovers from the wxWidgets UI. Also, the text for the Settings menu says "Config", however in code this is referred to as "Settings". It is also ambiguous as the "Configuration" screen for the graphics backend shares the same name on the UI.

"FullScr", "ScrShot" and "Config" have been changed to say "Fullscreen", "Screenshot" and "Settings". The MenuBar text and Settings dialog window have also been updated to reflect the change.

Master branch:
Screenshot_20200301_235717

Pull Request:
Screenshot_20200301_235826

### DIFF
--- a/Source/Core/DolphinQt/MenuBar.cpp
+++ b/Source/Core/DolphinQt/MenuBar.cpp
@@ -503,8 +503,7 @@ void MenuBar::AddViewMenu()
 void MenuBar::AddOptionsMenu()
 {
   QMenu* options_menu = addMenu(tr("&Options"));
-  options_menu->addAction(tr("Co&nfiguration"), this, &MenuBar::Configure,
-                          QKeySequence::Preferences);
+  options_menu->addAction(tr("&Settings"), this, &MenuBar::Configure, QKeySequence::Preferences);
   options_menu->addSeparator();
   options_menu->addAction(tr("&Graphics Settings"), this, &MenuBar::ConfigureGraphics);
   options_menu->addAction(tr("&Audio Settings"), this, &MenuBar::ConfigureAudio);
--- a/Source/Core/DolphinQt/ToolBar.cpp
+++ b/Source/Core/DolphinQt/ToolBar.cpp
@@ -122,12 +122,12 @@ void ToolBar::MakeActions()
   m_pause_play_action = addAction(tr("Play"), this, &ToolBar::PlayPressed);
 
   m_stop_action = addAction(tr("Stop"), this, &ToolBar::StopPressed);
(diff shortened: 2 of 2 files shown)


Techjar:
Seems you made a bit of a mess of the commits.

MayImilae:
So the Fullscr and ScrShot etc text was done deliberately to allow the icons to have even spacing. This was something that was debated and considered pretty carefully in the past, so please do not change it.

sonic2kk:
That seems like such an odd way to handle such a feature, but that isn't up to me. I guess I'll revert those.

orbea:
Personally I think the master looks better, the length of text fits much better with the width of the icons.

mbc07:
Honestly, having the button labels intentionally abbreviated so they have a consistent width seems a moot point. As far as I could tell, English seems to be the only language in Dolphin where some of the button labels were deliberately abbreviated to achieve a consistent width.

Those button labels are all marked as translatable strings and, in fact, I randomly changed Dolphin's language to the other translations that ships with Dolphin and as I suspected, the button width weren't consistent at all, except in eastern languages where their own, um, glyphs (is that the correct term?) actually results in button labels that actually are narrower than the button icons.

With that said, I do think "Fullscreen", "Screenshot" and "Settings" looks better, even if the button width isn't consistent anymore. This would also make the English button labels actually more consistent with the other translations that ships in Dolphin, where most don't use any abbreviation at all...

leoetlino:
Could you rebase this and get rid of the unncessary commits (e.g. the merge commits) with squashs/fixups please?

leoetlino:
@MayImilae Is there any reason to use abbreviations other than doing it for even spacing?

MayImilae:
So, I'd like for the abbreviations to go away too, but they are there for a reason - to maintain the icon spacing. This PR makes the spacing uneven and is just gross, imo. Other options are to make the icons bigger and/or the text smaller and just, bleh to those too.

However, it occurred to me while looking at this, that there is a possible real solution now though. Now that we have the [really good tooltips](https://dolphin-emu.org/blog/2020/12/10/dolphin-progress-report-october-2020/#50-13163-remove-description-box-in-graphics-tabs-and-use-custom-tooltips-instead-by-iwubcode), we could just remove the text entirely and replace it with our new tooltips! That would work great, and be flexible for all languages and everything!

But yea this PR is not that.

JMC47:
This pull request kind of fell through the cracks as it is.  We can revisit this was the new tooltips if you want but I don't see it making a huge difference at the moment.  I do agree that the abbreviations are a bit weird and we'd probably want them gone.

PR #8658, opened 2020-03-02
Externals: Update glslang.

orbea:
This updates `Externals/glslang` to upstream commit https://github.com/KhronosGroup/glslang/commit/56364b6b602696c021349794a8d39744a1052afc which is listed in the `scripts/known_good.json` file from the Vulkan-ValidationLayers repo for version `1.2.131.2` of the Vulkan-sdk, see https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/sdk-1.2.131.2/scripts/known_good.json.

in my experience this is as close as it gets for a stable glslang version, see https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/1082.

I also made sure to include the missing `LICENSE.txt` and tried to omit files that were previously omitted and are not needed.

Please review the additional commit for build fixes. This includes updating some code that seems to have been copied from `Externals/glslang/StandAlone/ResourceLimits.cpp` and adding `SPIRV/SpvPostProcess.cpp` to `Externals/glslang/CMakeLists.txt`.

I tested this briefly with vulkan and found no regressions with radv + aco using a RX Vega 56 with the mesa git master.

My main goal behind this is to fix the build with `clang-10` which fails to build older glslang versions like the one currently included.


orbea:
I added an alternative build fix in a third commit that I think should make it easier when updating glslang again in the future or if a system version of glslang is used (Not yet supported). I'm not really sure what the best course of action is so please advise and I will squash this PR as needed.

orbea:
I went with the second fix to use the glslang defaults as suggested to be better by Stenzek on irc.

JosJuice:
@dolphin-emu-bot rebuild

orbea:
@dolphin-emu-bot rebuild

orbea:
CI now passes and I think this is ready for review and testing. Please review the `vcxproj` changes carefully, I'm not familiar with these files.

CMakeLists.txt
```diff
@@ -539,9 +539,11 @@ else()
   message(STATUS "Using static fmt from Externals")
   add_subdirectory(Externals/fmt EXCLUDE_FROM_ALL)
 endif()
-add_subdirectory(Externals/glslang)
 add_subdirectory(Externals/imgui)
 
+include_directories(Externals/glslang)
```
stenzek, line 544: Does this have to be added as a project-wide include? Seems a little overkill.

orbea: Its been so long I don't recall why I added this, but it seems to compile/work fine without it so removed, thanks!

orbea: Now that I see the travis output I remembered. It finds the system glslang on my system and still builds. I'll see if I can fix it...

orbea: I added the include to `Source/Core/VideoBackends/Vulkan/CMakeLists.txt` and `Externals/glslang/CMakeLists.txt` instead, it should be good now.

PR #8659, opened 2020-03-03
Android: Fix obscured Dolphin version

Ebola16:
Fixes https://bugs.dolphin-emu.org/issues/11905

Partial revert of #7864 (@rafaeltoledo )

Before:
Screenshot_20200303-115830_Dolphin Emulator
Screenshot_20200303-115820_Dolphin Emulator

This PR:
Screenshot_20200303-135925_Dolphin Emulator
Screenshot_20200303-135933_Dolphin Emulator

### DIFF
--- a/Source/Android/app/src/main/res/values/styles.xml
+++ b/Source/Android/app/src/main/res/values/styles.xml
@@ -2,7 +2,7 @@
 <resources>
 
     <!-- Inherit from the material theme -->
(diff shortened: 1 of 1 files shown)


JosJuice:
I don't really understand the reason why the text gets cut off, but a partial revert to fix the problem sounds fine, so I will merge this in 7 days unless it turns out we actually have another active Android developer and they have an objection.

PR #8661, opened 2020-03-05
Tweaks to build against gcc-10.

trofi:


JosJuice:
@dolphin-emu-bot rebuild

Source/Core/Common/StringUtil.h
```diff
@@ -8,6 +8,7 @@
 #include <cstddef>
 #include <cstdlib>
 #include <iomanip>
+#include <limits> /* std::numeric_limits */
```
jordan-woyak: I think these comments are unnecessary.
We don't do this anywhere else and no one is going to keep them updated as header requirements change.

JosJuice: Agreed. This pull request LGTM if the comments are removed.

trofi: Sounds good. Re-pushed patches without a comment.

JosJuice:
@dolphin-emu-bot rebuild

PR #8662, opened 2020-03-06
Externals: Update MoltenVK to SDK version 1.2.131.2

stenzek:
Apparently it's broken on macOS at the moment...

I don't currently have access to a Mac to debug, so let's try this.


pizuz:
I'm still getting a black screen, unfortunately.

stenzek:
Superseded by #8728.

PR #8663, opened 2020-03-06
Resources: Make dolphin_logo.png square

spycrab:
Using non-square resolutions causes icon display issues on some platforms.

Before:
Screenshot_20200306_183356

After:
Screenshot_20200306_190414


JosJuice:
Does this affect the layout of the About window?

spycrab:
@JosJuice not that I can tell

Before:
image

After:
image

JosJuice:
Ah, the Dolphin logo takes up less space there than I recall it doing. I suppose there's no problem, then.

MayImilae:
To answer your question from IRC, there's not really any particular reason dolphin_logo.png was made non-square. It is just not square because it worked at the time with those that tested it on Linux for me.  If it is causing issues it should definitely be addressed!

PR #8665, opened 2020-03-07
Qt/CodeViewWidget: Make columns resizable by the user and set sensible defaults.

AdmiralCurtiss:
These were fixed-size and auto-calculated every time you scrolled up or down, making them jump wildly when you're in the vicinity of long symbol names or `rlwinm` instructions. With this they don't, and you can resize them at will.

A minor issue is that Qt is pretty awful at eliding strings in table cells, which means if you have a branch target to a long symbol, even if there's enough space for 90% of the symbol name, it will cut off the entire symbol. This kinda sucks but I can't find a way to work around it without completely reimplementing the text painting logic, so I guess I'll live with it.

### DIFF
--- a/Source/Core/DolphinQt/Debugger/CodeViewWidget.cpp
+++ b/Source/Core/DolphinQt/Debugger/CodeViewWidget.cpp
@@ -21,6 +21,7 @@
 #include <QTableWidgetItem>
 #include <QWheelEvent>
 
+#include "Common/GekkoDisassembler.h"
 #include "Common/StringUtil.h"
 #include "Core/Core.h"
 #include "Core/Debugger/PPCDebugInterface.h"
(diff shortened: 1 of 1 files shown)


Source/Core/DolphinQt/Debugger/CodeViewWidget.cpp
```diff
@@ -182,6 +190,18 @@ void CodeViewWidget::FontBasedSizing()
   verticalHeader()->setMaximumSectionSize(rowh);
   horizontalHeader()->setMinimumSectionSize(rowh + 5);
   setColumnWidth(CODE_VIEW_COLUMN_BREAKPOINT, rowh + 5);
+  setColumnWidth(CODE_VIEW_COLUMN_ADDRESS, fm.width(QStringLiteral("80000000")));
+
+  // this is the longest diassembled instruction i could find, so use it as a reference for how long
```
leoetlino:
```suggestion
  // This is the longest disassembled instruction I could find, so use it as a reference for how long
```
I think you should probably avoid "I" in comments though as it's not clear who wrote them (without having to git blame...)

AdmiralCurtiss: Mhm, yeah, that's fair. I think I'll try to figure out what the actual longest instruction is and then rephrase that to something better.

AdmiralCurtiss: Alright so I figured out the longest instruction and parameter set, but it turns out that using those values makes the columns noticeably too large for the typical case. I've rephrased the comment though.

leoetlino:
Hmm, the logic looked fine to me but in practice this appears to set column widths to values that are slightly too low. For instance the last two characters of addresses are not shown and replaced with ellipses instead...

AdmiralCurtiss:
That's curious. Here's how it looks to me (without having touched the columns), which I'd call wide enough: https://user-images.githubusercontent.com/4522237/76804509-baa40980-67dc-11ea-8261-45002c84fd33.png

Could it be that we're measuring with a different font than what we end up drawing with, somehow?

leoetlino:
Hmm, probably not, since `fm` uses the debug font, right? Maybe QFontMetrics doesn't work as well on Linux.

AdmiralCurtiss:
Alright, I tried a few different fonts and I can definitely reproduce this here on Windows too. Some fonts and font point sizes work fine, others end up too narrow. Maybe there's some column margins that need to be considered too? Definitely odd that it works fine with some fonts then though.

(Also I can't figure out what the default font was so I can't go back to it now, it was not the one selected by default. Oh well.)

leoetlino:
Well, in that case, do we want to add a small number to the value given by QFontMetrics? I think it's better for columns to be slightly too wide than too narrow, since the latter makes values unreadable without manual resizing. It does sound a bit hacky, but I can't think of a better way to fix the issue.

AdmiralCurtiss:
Yeah I can't figure out anything better either. Small extra it is.

PR #8666, opened 2020-03-11
Fix Vulkan/MoltenVK on macOS yet again

stenzek:
Well, this was a fun debugging experience. At least it wasn't as borked as the last time.

Basically, the auto-fill-background stuff changing out paint engines broke on Qt 5.14. I'm swapping this out with a better method - letting the graphics backend take care of clearing the screen instead. The downside is there will be a slightly longer delay before it is cleared as the graphics API needs to initialize, but it's much cleaner code-wise.

It also now passes the layer to MoltenVK instead of the view, which prevents `[NSView layer]` from being called off the main thread. Which currently is just a warning, but with the way Apple seems to be going at some point in the future it'll probably become an error, and cause Dolphin to crash, so better safe than sorry.

... hopefully this stays working for a while this time. I don't like debugging Mac issues :(


pizuz:
Working as advertised. The delay is barely noticeable.

gilcel:
Well done! I confirm that it fixes issue #12002.
Thanks for this!

PR #8668, opened 2020-03-12
RSO: Auto-detect RSO location in RAM

sepalani:
This PR adds an auto-detect features to the Generate Symbols from > RSO modules.

Ready to be reviewed.

### DIFF
--- a/Source/Core/Core/HW/AddressSpace.cpp
+++ b/Source/Core/Core/HW/AddressSpace.cpp
@@ -68,8 +68,8 @@ Accessors::iterator Accessors::end() const
   return nullptr;
 }
 
-std::optional<u32> Accessors::Search(u32 haystack_start, u8* needle_start, u32 needle_size,
-                                     bool forwards) const
+std::optional<u32> Accessors::Search(u32 haystack_start, const u8* needle_start,
+                                     std::size_t needle_size, bool forwards) const
 {
   return std::nullopt;
 }
@@ -91,7 +91,7 @@ struct EffectiveAddressSpaceAccessors : Accessors
   void WriteU64(u32 address, u64 value) override { PowerPC::HostWrite_U64(value, address); }
   float ReadF32(u32 address) const override { return PowerPC::HostRead_F32(address); };
 
-  bool Matches(u32 haystack_start, u8* needle_start, u32 needle_size) const
+  bool Matches(u32 haystack_start, const u8* needle_start, std::size_t needle_size) const
   {
     u32 page_base = haystack_start & 0xfffff000;
     u32 offset = haystack_start & 0x0000fff;
@@ -121,7 +121,7 @@ struct EffectiveAddressSpaceAccessors : Accessors
         return false;
       }
 
-      u32 chunk_size = std::min(0x1000 - offset, needle_size);
+      std::size_t chunk_size = std::min<std::size_t>(0x1000 - offset, needle_size);
       if (memcmp(needle_start, page_ptr + offset, chunk_size) != 0)
       {
         return false;
@@ -134,7 +134,7 @@ struct EffectiveAddressSpaceAccessors : Accessors
     return (needle_size == 0);
   }
 
-  std::optional<u32> Search(u32 haystack_start, u8* needle_start, u32 needle_size,
+  std::optional<u32> Search(u32 haystack_start, const u8* needle_start, std::size_t needle_size,
                             bool forward) const override
   {
     u32 haystack_address = haystack_start;
@@ -190,7 +190,7 @@ struct AuxiliaryAddressSpaceAccessors : Accessors
 
   iterator end() const override { return DSP::GetARAMPtr() + GetSize(); }
 
-  std::optional<u32> Search(u32 haystack_offset, u8* needle_start, u32 needle_size,
+  std::optional<u32> Search(u32 haystack_offset, const u8* needle_start, std::size_t needle_size,
                             bool forward) const override
   {
     if (!IsValidAddress(haystack_offset))
@@ -264,7 +264,7 @@ struct CompositeAddressSpaceAccessors : Accessors
     return it->accessors->WriteU8(address, value);
   }
 
-  std::optional<u32> Search(u32 haystack_offset, u8* needle_start, u32 needle_size,
+  std::optional<u32> Search(u32 haystack_offset, const u8* needle_start, std::size_t needle_size,
                             bool forward) const override
   {
     for (const AccessorMapping& mapping : m_accessor_mappings)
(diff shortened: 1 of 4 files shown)


Source/Core/DolphinQt/MenuBar.cpp
```diff
@@ -1218,6 +1224,74 @@ void MenuBar::GenerateSymbolsFromRSO()
   }
 }
 
+void MenuBar::GenerateSymbolsFromRSOAuto()
+{
+  const std::array<std::string, 2> search_for = {".elf", ".plf"};
```
lioncash:
```suggestion
  constexpr std::array<std::string_view, 2> search_for = {".elf", ".plf"};
```

Source/Core/DolphinQt/MenuBar.cpp
```diff
@@ -1196,6 +1197,11 @@ void MenuBar::GenerateSymbolsFromSignatureDB()
 
 void MenuBar::GenerateSymbolsFromRSO()
 {
+  int ret =
```
lioncash:
```suggestion
  const int ret =
```

Source/Core/DolphinQt/MenuBar.cpp
```diff
@@ -1218,6 +1224,74 @@ void MenuBar::GenerateSymbolsFromRSO()
...
+  AddressSpace::Accessors* accessors = AddressSpace::GetAccessors(AddressSpace::Type::Effective);
+
+  // Find filepath to elf/plf commonly used by RSO modules
+  for (const auto& str : search_for)
+  {
+    u32 next = 0;
+    while (true)
+    {
+      auto found_addr =
+          accessors->Search(next, (const u8*)str.data(), static_cast<u32>(str.size() + 1), true);
```
lioncash: reinterpret_cast

Source/Core/DolphinQt/MenuBar.cpp
```diff
@@ -1218,6 +1224,74 @@ void MenuBar::GenerateSymbolsFromRSO()
...
+
+      if (!found_addr.has_value())
+        break;
+      next = *found_addr + 1;
+
+      // Get the beginning of the string
+      found_addr = accessors->Search(*found_addr, (const u8*)"", 1, false);
```
lioncash: reinterpret_cast

Source/Core/DolphinQt/MenuBar.cpp
```diff
@@ -1218,6 +1224,74 @@ void MenuBar::GenerateSymbolsFromRSO()
...
+      if (!found_addr.has_value())
+        continue;
+
+      // Get the string reference
+      u32 ref_addr = *found_addr + 1;
```
lioncash:
```suggestion
      const u32 ref_addr = *found_addr + 1;
```

Source/Core/DolphinQt/MenuBar.cpp
```diff
@@ -1218,6 +1224,74 @@ void MenuBar::GenerateSymbolsFromRSO()
...
+      const std::array<u8, 4> ref = {(ref_addr >> 24) & 0xFF, (ref_addr >> 16) & 0xFF,
```
lioncash: These should likely contain a cast around the values, rather than a mask, as this is implicitly truncating `u32` values down to `u8` as far as the type system cares.

Source/Core/Core/HW/AddressSpace.cpp
```diff
@@ -68,7 +68,7 @@ Accessors::iterator Accessors::end() const
   return nullptr;
 }
 
-std::optional<u32> Accessors::Search(u32 haystack_start, u8* needle_start, u32 needle_size,
+std::optional<u32> Accessors::Search(u32 haystack_start, const u8* needle_start, u32 needle_size,
```
lioncash: You may want to also make `needle_size` be a `size_t`, rather than a `u32`, as it wouldn't require casting from external usages that need to interact with standard containers, etc. Ditto for the other member functions that accept a size.

Source/Core/DolphinQt/MenuBar.cpp
```diff
@@ -1218,6 +1224,74 @@ void MenuBar::GenerateSymbolsFromRSO()
...
+                                     (ref_addr >> 8) & 0xFF, ref_addr & 0xFF};
+      found_addr = accessors->Search(ref_addr, ref.data(), 4, false);
```
lioncash:
```suggestion
      found_addr = accessors->Search(ref_addr, ref.data(), ref.size(), false);
```

lioncash:
Commit message for the second commit should also provide some form of description, rationale, or some form of explanation. Currently it can just be interpreted as "Implemented thing", which isn't as helpful of a description as it could be.

Source/Core/DolphinQt/MenuBar.cpp
```diff
@@ -1196,6 +1197,11 @@ void MenuBar::GenerateSymbolsFromSignatureDB()
 
 void MenuBar::GenerateSymbolsFromRSO()
 {
+  int ret =
+      ModalMessageBox::question(this, tr("RSO auto-detection"), tr("Auto-detect RSO modules?"));
```
BhaaLseN: I wonder if this is the best string to be presented to a user...but then again, this seems pretty advanced and whoever uses it likely knows what RSO means.
Yet, would it hurt to leave an `i18n` comment for translators with the spelled out wording for RSO?

sepalani:
All of the above comments are addressed.

I also modified the code a bit to add RSO module names to the dropdown.
image

Source/Core/DolphinQt/MenuBar.cpp
```diff
@@ -1196,6 +1197,12 @@ void MenuBar::GenerateSymbolsFromSignatureDB()
 
 void MenuBar::GenerateSymbolsFromRSO()
 {
+  // i18n: RSO refers to a proprietary format for shared object (like DLL file).
```
leoetlino:
```suggestion
  // i18n: RSO refers to a proprietary format for shared objects (like DLL files).
```

Source/Core/DolphinQt/MenuBar.cpp
```diff
@@ -1218,6 +1225,79 @@ void MenuBar::GenerateSymbolsFromRSO()
...
+      const u32 ref_addr = *found_addr + 1;
+      const std::array<u8, 4> ref = {static_cast<u8>(ref_addr >> 24),
+                                     static_cast<u8>(ref_addr >> 16),
+                                     static_cast<u8>(ref_addr >> 8), static_cast<u8>(ref_addr)};
+      found_addr = accessors->Search(ref_addr, ref.data(), ref.size(), false);
+      if (!found_addr.has_value() || *found_addr < 16)
+        continue;
+
+      // Go to the beginning of the RSO header
+      matches.push_back(std::make_pair(*found_addr - 16, PowerPC::HostGetString(ref_addr, 128)));
```
leoetlino: I think you can avoid make_pair by using emplace_back?

```suggestion
      matches.emplace_back(*found_addr - 16, PowerPC::HostGetString(ref_addr, 128));
```

sepalani:
@leoetlino Done.

Source/Core/DolphinQt/MenuBar.cpp
```diff
@@ -1218,6 +1225,79 @@ void MenuBar::GenerateSymbolsFromRSO()
...
+  std::vector<std::pair<u32, std::string>> matches;
+
+  // Find filepath to elf/plf commonly used by RSO modules
+  for (const auto& str : search_for)
+  {
+    u32 next = 0;
+    while (true)
+    {
+      auto found_addr = accessors->Search(next, reinterpret_cast<const u8*>(str.data()),
+                                          static_cast<u32>(str.size() + 1), true);
```
lioncash: This cast doesn't seem to be necessary any longer with the change over to size_t in the interface.

sepalani:
@lioncash Fixed.

PR #8669, opened 2020-03-14
Breakpoints: Fix crash after clearing all memory breakpoints

container1234:
This PR fixes crash after clearing memory breakpoints by pressing Clear button.

### DIFF
--- a/Source/Core/Core/PowerPC/BreakPoints.cpp
+++ b/Source/Core/Core/PowerPC/BreakPoints.cpp
@@ -196,6 +196,15 @@ void MemChecks::Remove(u32 address)
   });
 }
 
+void MemChecks::Clear()
+{
+  Core::RunAsCPUThread([&] {
+    m_mem_checks.clear();
+    JitInterface::ClearCache();
+    PowerPC::DBATUpdated();
+  });
+}
+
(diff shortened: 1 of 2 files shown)


Source/Core/Core/PowerPC/BreakPoints.cpp
```diff
@@ -196,6 +196,15 @@ void MemChecks::Remove(u32 address)
   });
 }
 
+void MemChecks::Clear()
+{
+  Core::RunAsCPUThread([&] {
+    m_mem_checks.clear();
+    JitInterface::ClearCache();
+    PowerPC::DBATUpdated();
```
leoetlino: Could you explain why this is required?

container1234: The bug doesn't occur when memory point is deleted by pressing Delete button (not Clear button), so I compared MemChecks::Remove and MemChecks::Clear.
I don't know how it works internally but compared to Memchecks::Remove, MemChecks::Clear lacks JitInterface::ClearCache and PowerPC::DBATUpdated.

leoetlino: Ok, that makes sense.

PR #8671, opened 2020-03-15
BTReal: Improve error handling for device opening

leoetlino:
This commit attempts to improve error handling for device opening by
reducing panic alert spam when opening one or several devices fails.
Currently, Dolphin shows a panic alert for every device that we fail
to open, and another panic alert at the end if no usable device was
found. That is certainly a bit excessive -- we should only keep the
very last panic alert (the one that is shown if everything fails)
and we can just put the error for the last device open operation there.

This also changes the PanicAlert to a CriticalAlert to ensure the
message is visible even if the user has disabled regular panic alerts.

The message has also been reworded and should hopefully be clearer.

### DIFF
--- a/Source/Core/Core/IOS/USB/Bluetooth/BTReal.cpp
+++ b/Source/Core/Core/IOS/USB/Bluetooth/BTReal.cpp
@@ -16,6 +16,7 @@
 #include <utility>
 #include <vector>
 
+#include <fmt/format.h>
 #include <libusb.h>
 
 #include "Common/ChunkFile.h"
@@ -81,6 +82,7 @@ IPCCommandResult BluetoothReal::Open(const OpenRequest& request)
   if (!m_context.IsValid())
     return GetDefaultReply(IPC_EACCES);
 
+  m_last_open_error.clear();
   m_context.GetDeviceList([this](libusb_device* device) {
     libusb_device_descriptor device_descriptor;
     libusb_get_device_descriptor(device, &device_descriptor);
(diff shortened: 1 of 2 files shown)


Source/Core/Core/IOS/USB/Bluetooth/BTReal.cpp
```diff
@@ -547,7 +558,8 @@ bool BluetoothReal::OpenDevice(libusb_device* device)
   const int ret = libusb_open(m_device, &m_handle);
   if (ret != 0)
   {
-    PanicAlertT("Failed to open Bluetooth device: %s", libusb_error_name(ret));
+    m_last_open_error = StringFromFormat(
```
iwubcode: I know @lioncash did some work to switch over some uses of `StringFromFormat` to `fmt`, would that be worthwhile to do where you're using `StringFromFormat` here and below?

leoetlino: Good idea! I've replaced both uses of StringFromFormat with fmt.

Source/Core/Core/IOS/USB/Bluetooth/BTReal.cpp
```diff
@@ -547,7 +559,8 @@ bool BluetoothReal::OpenDevice(libusb_device* device)
   const int ret = libusb_open(m_device, &m_handle);
   if (ret != 0)
   {
-    PanicAlertT("Failed to open Bluetooth device: %s", libusb_error_name(ret));
+    m_last_open_error = fmt::format(Common::GetStringT("Failed to open Bluetooth device: %s"),
```
JosJuice: Does `fmt::format` support `%s`-style formatting? All the other places where we use it use `{}`.

leoetlino: ...oh no, I forgot to change the `%s`s. Thanks for catching this!

iwubcode: It does support them to a degree, it mentions it under the `printf` style formatting: https://fmt.dev/latest/api.html#printf-api but whether you have to include the other header, not sure.

PR #8672, opened 2020-03-16
Back up Wii setting.txt and SYSCONF while emulating

JosJuice:
When booting a Wii game, Dolphin can overwrite certain settings in the SYSCONF file, such as turning off PAL60 for NTSC games. Normally, these settings get reverted at the end of emulation, but this does not happen if Dolphin crashes or force quits in some other way. (Personally, I have a tendency to use Visual Studio's Stop Debugging button, which kills the process...)

Dolphin also overwrites certain values in setting.txt when booting a Wii game. Unlike with SYSCONF, we currently make no effort to preserve the original values in this file.

This change fixes both of these problems by copying SYSCONF and setting.txt to the Backup folder when booting a Wii game, and then copying them back either when launching Dolphin (in case the previous run of Dolphin crashed) or when ending emulation.

### DIFF
--- a/Source/Core/Core/WiiRoot.cpp
+++ b/Source/Core/Core/WiiRoot.cpp
@@ -16,6 +16,8 @@
 #include "Common/FileUtil.h"
 #include "Common/Logging/Log.h"
 #include "Common/NandPaths.h"
+#include "Common/StringUtil.h"
+#include "Core/CommonTitles.h"
 #include "Core/ConfigManager.h"
 #include "Core/HW/WiiSave.h"
 #include "Core/IOS/ES/ES.h"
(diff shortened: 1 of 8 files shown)


leoetlino:
Won't this cause setting changes that are done during emulation (e.g. updating the Wii remote speaker volume with the HOME menu, or changing console settings in the System Menu) to be always reverted?

JosJuice:
Yes, you're right. I didn't think of that being a valid way to change settings. I will make it so that the SYSCONF backup gets deleted instead of restored at emulation end.

leoetlino:
I wonder whether it'd be possible to improve the existing SYSCONF restore logic, so that it also works if Dolphin was closed ungracefully: 

https://github.com/dolphin-emu/dolphin/blob/361364260a3e335e23be99ad0642d5095ebb60fd/Source/Core/Core/BootManager.cpp#L458-L486

JosJuice:
That would require us to save the list of overridden settings to disk in some way, right? But in exchange, we wouldn't have to save an extra copy of the SYSCONF file itself like I do in this PR.

leoetlino:
Yeah, and one benefit is that changes won't be lost even if Dolphin crashes. (Right now, I believe that changes will still be lost if Dolphin crashes, since the entire SYSCONF is restored.) But it's more complex and I don't know if it's worth handling that (somewhat) rare situation.

JosJuice:
Yes, I agree with all of those points.

Source/Core/Core/WiiRoot.cpp
```diff
@@ -32,6 +34,48 @@ namespace FS = IOS::HLE::FS;
...
 static std::string s_temp_wii_root;
 
+static std::string FileNameFromPath(const std::string& path)
+{
+  std::string file_name, extension;
+  SplitPath(path, nullptr, &file_name, &extension);
+  return file_name + extension;
+}
+
+static bool CopyBackupFile(const std::string& path_from, const std::string& path_to)
```
leoetlino: Why is this required? File::Copy should already fail if the source file doesn't exist. And the destination file is overwritten if it already exists (bFailIfExists=false is passed to [CopyFile](https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-copyfile) and ofstream is opened in non-append mode)

Source/Core/Core/WiiRoot.cpp
```diff
@@ -32,6 +34,48 @@ namespace FS = IOS::HLE::FS;
 
 static std::string s_temp_wii_root;
 
+static std::string FileNameFromPath(const std::string& path)
```
leoetlino: Is it worth putting this in StringUtil? (Also, it's a shame we still cannot use std::filesystem because of Android :/)

Source/Core/Core/WiiRoot.cpp
JosJuice: Ah, I didn't know it would be overwritten if it exists. As for the check for whether the source file exists, I thought it would be nice to avoid getting ERROR_LOGs on every normal boot of Dolphin. I can remove the target deletion, but do you think I should remove the source existence check as well?

leoetlino: Ah, that makes sense. I agree with keeping the existence check then.

Source/Core/Core/WiiRoot.cpp
JosJuice: Done.

PR #8673, opened 2020-03-16
Boot: Do a better job of preserving certain parts of settings.txt

JosJuice:
This is a part of fixing https://bugs.dolphin-emu.org/issues/11930.

PR #8672 should be merged before this one, otherwise Wiimmfi users may run into a situation where a 7-day wait isn't necessary at first but becomes necessary later after playing a game from a different region, which I think would be rather confusing for them.

EDIT: By request from @Leseratte10 (see comments below), do not merge this PR before PR #8680.

### DIFF
--- a/Source/Core/Core/Boot/Boot_BS2Emu.cpp
+++ b/Source/Core/Core/Boot/Boot_BS2Emu.cpp
@@ -242,6 +242,30 @@ bool CBoot::EmulatedBS2_GC(const DiscIO::VolumeDisc& volume)
   return RunApploader(/*is_wii*/ false, volume);
 }
 
+static DiscIO::Region CodeRegion(char c)
+{
+  switch (c)
+  {
+  case 'J':  // Japan
+  case 'T':  // Taiwan
+    return DiscIO::Region::NTSC_J;
+  case 'B':  // Brazil
+  case 'M':  // Middle East
+  case 'R':  // Argentina
+  case 'S':  // ???
+  case 'U':  // USA
+  case 'W':  // ???
+    return DiscIO::Region::NTSC_U;
+  case 'A':  // Australia
+  case 'E':  // Europe
+    return DiscIO::Region::PAL;
+  case 'K':  // Korea
+    return DiscIO::Region::NTSC_K;
+  default:
+    return DiscIO::Region::Unknown;
+  }
+}
+
 bool CBoot::SetupWiiMemory(IOS::HLE::IOSC::ConsoleType console_type)
 {
   static const std::map<DiscIO::Region, const RegionSetting> region_settings = {
@@ -267,11 +291,25 @@ bool CBoot::SetupWiiMemory(IOS::HLE::IOSC::ConsoleType console_type)
     {
       gen.SetBytes(std::move(data));
       serno = gen.GetValue("SERNO");
+
+      bool region_matches = false;
       if (SConfig::GetInstance().bOverrideRegionSettings)
+      {
+        region_matches = true;
+      }
+      else
+      {
+        const std::string code = gen.GetValue("CODE");
+        if (code.size() >= 2 && CodeRegion(code[1]) == SConfig::GetInstance().m_region)
+          region_matches = true;
+      }
+
+      if (region_matches)
       {
         region_setting = RegionSetting{gen.GetValue("AREA"), gen.GetValue("VIDEO"),
                                        gen.GetValue("GAME"), gen.GetValue("CODE")};
       }
+
       gen.Reset();
     }
   }
--- a/Source/Core/Core/Boot/Boot_BS2Emu.cpp
+++ b/Source/Core/Core/Boot/Boot_BS2Emu.cpp
@@ -278,6 +278,7 @@ bool CBoot::SetupWiiMemory(IOS::HLE::IOSC::ConsoleType console_type)
 
   Common::SettingsHandler gen;
   std::string serno;
+  std::string model = "RVL-001(" + region_setting.area + ")";
   CreateSystemMenuTitleDirs();
   const std::string settings_file_path(Common::GetTitleDataPath(Titles::SYSTEM_MENU) +
                                        "/" WII_SETTING);
@@ -291,6 +292,7 @@ bool CBoot::SetupWiiMemory(IOS::HLE::IOSC::ConsoleType console_type)
     {
       gen.SetBytes(std::move(data));
       serno = gen.GetValue("SERNO");
+      model = gen.GetValue("MODEL");
 
       bool region_matches = false;
       if (SConfig::GetInstance().bOverrideRegionSettings)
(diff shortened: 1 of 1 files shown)


Source/Core/Core/Boot/Boot_BS2Emu.cpp
```diff
@@ -242,6 +242,28 @@ bool CBoot::EmulatedBS2_GC(const DiscIO::VolumeDisc& volume)
   return RunApploader(/*is_wii*/ false, volume);
 }
 
+DiscIO::Region CodeRegion(char c)
```
lioncash: This should likely be made internally linked

Source/Core/Core/Boot/Boot_BS2Emu.cpp
```diff
@@ -267,11 +290,22 @@ bool CBoot::SetupWiiMemory(IOS::HLE::IOSC::ConsoleType console_type)
...
       if (SConfig::GetInstance().bOverrideRegionSettings)
       {
         region_setting = RegionSetting{gen.GetValue("AREA"), gen.GetValue("VIDEO"),
                                        gen.GetValue("GAME"), gen.GetValue("CODE")};
       }
+      else
+      {
+        const std::string code = gen.GetValue("CODE");
+        if (code.size() >= 2 && CodeRegion(code[1]) == SConfig::GetInstance().m_region)
+          region_setting.code = code;
```
lioncash: `code` can be made non-const and moved here:

```suggestion
          region_setting.code = std::move(code);
```

Leseratte10:
I just ran a script through the Wiimmfi database to grab all the existing serial number prefixes that ever connected to Wiimmfi, and it looks like the letters R and S can also exist as second char in the serial number. R is Argentina (NTSC-U). I don't know what country S is. From the logs it looks a lot like NTSC-U, but I'm not 100% sure.

JosJuice:
I've added those two as NTSC-U.

JosJuice:
For curiosity's sake, I checked the setting.txt file of my Middle Eastern Wii, and it has the following:

AREA=USA
MODEL=RVL-001(UAE)
DVD=0
MPCH=0x7FFE
CODE=LMH
SERNO=[redacted]
VIDEO=NTSC
GAME=US

So the text in parentheses in MODEL isn't necessarily the same as AREA. Interesting...

Leseratte10:
Yeah, it looks like "UAE" wouldn't be a valid value for AREA. Or the tool I'm using doesn't know that UAE would be valid. The "SettingEdit" I used to look at the settings.txt file only has JPN, USA, EUR, AUS, BRA, TWN, ROC, KOR, HKG, ASI, LTN and SAF in the dropdown for AREA. The "Model" value is a free text field. Maybe Nintendo decided on the UAE model region after they've already finalized the list of valid / possible AREAs?

Although it could just be that UAE is missing from the tool by accident, but then why would an UAE wii not have UAE as AREA ...

Source/Core/Core/Boot/Boot_BS2Emu.cpp
```diff
@@ -242,6 +242,30 @@ bool CBoot::EmulatedBS2_GC(const DiscIO::VolumeDisc& volume)
...
+  case 'K':  // Korea
+    return DiscIO::Region::NTSC_K;
+  default:
+    return DiscIO::Region::Unknown;
+  }
+}
+
 bool CBoot::SetupWiiMemory(IOS::HLE::IOSC::ConsoleType console_type)
 {
   static const std::map<DiscIO::Region, const RegionSetting> region_settings = {
```
Leseratte10: Can these serial mappings be changed? "LJ" isn't a valid prefix for JPN, and "LE" isn't a valid one for PAL. Japanese and PAL serial number prefixes have three chars. I'd suggest LJH and LEH (so it's identical to KOR for which it is LKH).

Leseratte10:
Can the default serial mappings for PAL and JPN changed? "LJ" isn't a valid prefix for JPN, and "LE" isn't a valid one for PAL. Japanese and PAL serial number prefixes have three chars. I'd suggest LJH and LEH (so it's identical to KOR for which it is LKH).

Source/Core/Core/Boot/Boot_BS2Emu.cpp
Leseratte10: Can these serial mappings be changed? "LJ" isn't a valid prefix for JPN, and "LE" isn't a valid one for PAL. Japanese and PAL serial number prefixes have three chars. I'd suggest LJH and LEH (so it's identical to KOR for which it is LKH).

Leseratte10: Or, maybe, for out-of-region games, just take the serial prefix from the settings.txt and just edit the 2nd character to "match" the region?

JosJuice: > Can these serial mappings be changed? As far as I know, "LJ" isn't a valid prefix for JPN, and "LE" isn't a valid one for PAL. Japanese and PAL serial number prefixes have three chars. I'd suggest LJH and LEH (so it's identical to KOR for which it is LKH).

I have now created PR #8678 for this.

> Or, maybe, for out-of-region games, just take the serial prefix from the settings.txt and just edit the 2nd character to "match" the region?

We need values to fall back to in case there isn't already a setting.txt file. (Most users don't import a NAND.)

Also, I don't see how this would solve the problem of having the wrong number of characters.

Leseratte10: That would solve the problem for the users who did import a NAND, but the solution in #8678 is probably better anyways because it applies to everyone.

Leseratte10:
I just tested this PR, but there is still a bug somewhere. 

After importing the NAND backup, the settings are looking correct. Printing "decrypted" to the console results in an output like this: 

    AREA=EUR
    MODEL=RVL-001(EUR)
    DVD=0
    MPCH=0x7FFE
    CODE=LEH
    
    SERNO=123456789
    VIDEO=PAL
    GAME=EU

Then, when I quit the game and start it again, it prints this instead: 

    AREA=EUR
    MODEL=RVL-001(EUR)
    DVD=0
    MPCH=0x7FFE
    CODE=LEH
    SERNO=12345

Then, in any subsequent runs, it prints this: 

    AREA=EUR
    MODEL=RVL-001(EUR)
    DVD=0
    MPCH=0x7FFE
    CODE=LEH
    SERNO=1234
    VIDEO=
    GAME=

So the last part of the serial number, and the VIDEO and GAME entries, are just gone. 

Is it running into a size limit somewhere?

Leseratte10:
@JosJuice Okay, either I am reading this piece of code wrong (I don't think that's the case), or there is a fairly stupid bug in the `SettingsHandler::Decrypt` function. 

Take a look at this code block that does the decryption: 

    const u8* str = m_buffer.data();
    while (*str != 0)
    {
      if (m_position >= m_buffer.size()) {
        return;  
      }
        
      decoded.push_back((u8)(m_buffer[m_position] ^ m_key));
      m_position++;
      str++;
      m_key = (m_key >> 31) | (m_key << 1);
    }

Am I understanding this correctly that "str" is a pointer to the encoded / obfuscated data? This means that "str" is pointing to binary data, not to a string. 
In turn, this means that when the encoded binary data contains a NULL byte, the code stops and thinks it's done. 

When I replace the `while (*str != 0)` loop with a `while (1)` loop and only rely on the if condition to return at some point, everything is working just fine. 

That would mean that currently, while decrypting a setting.txt, there is a 1/256 chance for each byte in the file, that the decryption just randomly stops because the encrypted value is a 0.

Leseratte10:
I just opened issue 12019 for this decryption bug.

Leseratte10:
Thanks for merging #8680 and fixing issue 12019, now this should be the last PR that needs to be merged to make NAND import work properly, at least for all the serial numbers and other identifiers.

PR #8676, opened 2020-03-18
TitleDatabase: Added custom publisher database option.

MrMendelli:
Copied and updated from #8637.

Functions identically to `titles.txt`, but uses `publishers.txt` instead. I also think that _"Maker"_ should be changed to _"Publisher"_ in the UI. I believe this will make more sense to the end-users, but the code should stick to _"maker"_ as it's the official tag and would require. It would also more work to rename everything id it is changed. In accordance, the title database load option has been re-labeled to `Use Built-In Database of Metadata`. Note that the ability to load the title database is already broken if a custom title file is present, it seems that the custom titles aren't unloaded if the `Use Built-In Database of Titles` option is enabled. The discovery of this bug was made during development of this update and it is unknown how far back it goes. Unless I'm mistaken, the title database strings should be predominant if the option is enabled for all titles even if a custom titles file is present. Furthermore, the entire metadata load implementation would have to be completely re-written to fix this bug.

Comparison

### DIFF
--- a/Source/Core/Core/ConfigManager.cpp
+++ b/Source/Core/Core/ConfigManager.cpp
@@ -150,7 +150,7 @@ void SConfig::SaveInterfaceSettings(IniFile& ini)
   interface->Set("LanguageCode", m_InterfaceLanguage);
   interface->Set("ExtendedFPSInfo", m_InterfaceExtendedFPSInfo);
   interface->Set("ShowActiveTitle", m_show_active_title);
-  interface->Set("UseBuiltinTitleDatabase", m_use_builtin_title_database);
+  interface->Set("UseBuiltinTitleDatabase", m_use_builtin_metadata_database);
   interface->Set("ThemeName", theme_name);
   interface->Set("PauseOnFocusLost", m_PauseOnFocusLost);
   interface->Set("DebugModeEnabled", bEnableDebugging);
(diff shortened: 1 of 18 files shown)


JosJuice:
> In accordance, the title database load option has been re-labeled to `Use Built-In Database of Metadata`.

This PR doesn't make us ship with any developer metadata in the Sys folder, so I don't see a reason to rename the setting.

> Unless I'm mistaken, the title database strings should be predominant if the option is enabled for all titles even if a custom titles file is present.

No, the intent is that the custom titles have the highest priority. What's the point of allowing custom titles if the titles in the Sys folder override them for almost every game?

JosJuice:
For the reference, this comment of mine from the old PR still applies:

> Not sure if we should "invent" a new format (even though it uses the same layout as titles.txt)... When it comes to titles, you can get a file containing them by e.g. going to GameTDB.com and downloading wiitdb.txt, but there is no such place where you can get a list of developers/publishers in the format you are using here as far as I know. On the other hand, GameTDB.com offers a wiitdb.zip file downloading containing various metadata including the developer and publisher. Maybe it would make more sense to add support for that format?

JosJuice:
@dolphin-emu-bot rebuild

MrMendelli:
I'm not looking to change or re-invent the system, this was just a spur-of-the-moment idea I had because I wanted the same system for titles for publishers as well. I have spoke with other users and I am not the only person that wants this feature. I think the entire system could use some cleaning up, but I am not capable of doing so at this time.

AlexApps99:
I do agree with Jos that instead of depending on the user to supply a text file of all the publishers it should be something left to a database, such as GameTDB.

MrMendelli:
> 
> 
> I do agree with Jos that instead of depending on the user to supply a text file of all the publishers it should be something left to a database, such as GameTDB.

The embedded publisher would still appear if the user doesn't provide a custom file. This is optional just like custom titles. Just like custom titles, the string embedded in the game dump can still be used. All this does is allow custom strings for games if the user adds them within the file I have added in my PR.

PR #8680, opened 2020-03-20
SettingsHandler: Always decode the whole settings.txt file

Leseratte10:
Fixes https://bugs.dolphin-emu.org/issues/12019
Related to https://bugs.dolphin-emu.org/issues/11930 and to #8673 and #8670

The encrypted settings.txt can contain nullbytes - it's a binary file. The old code assumed a null byte means "end of file". That's not the case - for some files, the current code stops parsing in the middle of the file. (That took me a while to figure out ...)

This approach decrypts the junk data at the end of the file, too, but because the valid data always ends with a newline, all the junk ends up on new line(s) at the end, meaning, the parsing code should just ignore that junk. The decrypted data unfortunately doesn't end with a null byte, and we don't know the length of the decrypted data, so I don't see any way to detect and remove that junk data. 

Please merge this prior to #8673.

### DIFF
--- a/Source/Core/Common/SettingsHandler.cpp
+++ b/Source/Core/Common/SettingsHandler.cpp
@@ -71,10 +71,8 @@ std::string SettingsHandler::GetValue(std::string_view key) const
 void SettingsHandler::Decrypt()
 {
   const u8* str = m_buffer.data();
-  while (*str != 0)
+  while (m_position < m_buffer.size())
   {
-    if (m_position >= m_buffer.size())
-      return;
     decoded.push_back((u8)(m_buffer[m_position] ^ m_key));
     m_position++;
     str++;


JosJuice:
Decrypting junk at the end doesn't sound perfect to me, but I can't think of a better way... I wonder how actual Wii software handles it?

I'm fine with merging this if nobody can think of a "cleaner" solution.

JosJuice:
@dolphin-emu-bot rebuild

Leseratte10:
Maybe the Wii is just assuming that the file has a certain amount of lines / entries? We could do that as well, but that might cause other issues. Decrypting junk, while not being an ideal solution, should at least not cause problems.

JosJuice:
As long as we're not unlucky enough that the junk happens to decode into something valid, that is. But the chances of that happening are much much smaller than the chance of the valid encrypted data containing a null byte, at least.

Leseratte10:
The junk decrypting into something valid wouldn't be a problem either. 
The parsing code (when querying for something like SERNO) stops after the first occurrence. That means, as long as the file data is valid (all parameters are present) the data afterwards is ignored anyways, even if it was syntactically valid.

Leseratte10:
I just tested three different NAND backups again (one of which did work before, two of which did not work before due to the null byte) and all of them do import properly with the proper serial number.
Is anything still missing before this (and then, #8673) can be merged or is this just waiting for someone else to review / check?

JosJuice:
This PR is just waiting for someone else to review. I was planning to merge it in a day or two from now if nobody does.

PR #8684, opened 2020-03-22
[Don't Merge, Meta one] Cleanup warnings

howard0su:
Clean up warnings in various projects exclude the Externals folder.

### DIFF
--- a/Source/Core/Common/ChunkFile.h
+++ b/Source/Core/Common/ChunkFile.h
@@ -312,6 +312,7 @@ private:
       break;
 
     case MODE_WRITE:
+#pragma GCC diagnostic ignored "-Wmaybe-uninitialized"
       memcpy(*ptr, data, size);
       break;
 
--- a/Source/Core/Core/HW/GCMemcard/GCMemcard.cpp
+++ b/Source/Core/Core/HW/GCMemcard/GCMemcard.cpp
@@ -926,9 +926,8 @@ GCMemcardRemoveFileRetVal GCMemcard::RemoveFile(u8 index)  // index in the direc
 
   // TODO: Deleting a file via the GC BIOS sometimes leaves behind an extra updated directory block
   // here that has an empty file with the filename "Broken File000" where the actual deleted file
-  // was. Determine when exactly this happens and if this is neccessary for anything.
-
-  memset(static_cast<void*>(&(UpdatedDir.m_dir_entries[index])), 0xFF, DENTRY_SIZE);
+  // was. Determine when exactly this happens and if this is necessary for anything.
+  UpdatedDir.m_dir_entries[index].m_gamecode = DEntry::UNINITIALIZED_GAMECODE;
   UpdatedDir.m_update_counter = UpdatedDir.m_update_counter + 1;
   UpdateDirectory(UpdatedDir);
 
@@ -1581,9 +1580,11 @@ std::pair<u32, u32> Header::CalculateSerial() const
   return std::make_pair(serial1, serial2);
 }
 
-DEntry::DEntry()
+DEntry::DEntry() :
+  m_gamecode(UNINITIALIZED_GAMECODE),
+  m_unused_1(0xff),
+  m_unused_2({0xff, 0xff})
 {
-  memset(this, 0xFF, DENTRY_SIZE);
 }
 
 std::string DEntry::GCI_FileName() const
--- a/Source/Core/Core/HW/GCMemcard/GCMemcard.cpp
+++ b/Source/Core/Core/HW/GCMemcard/GCMemcard.cpp
@@ -928,7 +928,7 @@ GCMemcardRemoveFileRetVal GCMemcard::RemoveFile(u8 index)  // index in the direc
   // here that has an empty file with the filename "Broken File000" where the actual deleted file
   // was. Determine when exactly this happens and if this is neccessary for anything.
 
-  memset(&(UpdatedDir.m_dir_entries[index]), 0xFF, DENTRY_SIZE);
+  memset(static_cast<void*>(&(UpdatedDir.m_dir_entries[index])), 0xFF, DENTRY_SIZE);
   UpdatedDir.m_update_counter = UpdatedDir.m_update_counter + 1;
   UpdateDirectory(UpdatedDir);
 
@@ -1108,6 +1108,8 @@ GCMemcardExportFileRetVal GCMemcard::ExportGci(u8 index, const std::string& file
     return GCMemcardExportFileRetVal::FAIL;
   case GCMemcardGetSaveDataRetVal::NOMEMCARD:
     return GCMemcardExportFileRetVal::NOMEMCARD;
+  case GCMemcardGetSaveDataRetVal::SUCCESS:
+    break;
   }
   gci.Seek(DENTRY_SIZE + offset, SEEK_SET);
   for (unsigned int i = 0; i < size; ++i)
--- a/Source/Core/Core/IOS/IOSC.cpp
+++ b/Source/Core/Core/IOS/IOSC.cpp
@@ -420,9 +420,9 @@ static CertECC MakeBlankEccCert(const std::string& issuer, const std::string& na
 {
   CertECC cert{};
   cert.signature.type = SignatureType(Common::swap32(u32(SignatureType::ECC)));
-  std::strncpy(cert.signature.issuer, issuer.c_str(), 0x40);
+  memcpy(cert.signature.issuer, issuer.c_str(), 0x40);
   cert.header.public_key_type = PublicKeyType(Common::swap32(u32(PublicKeyType::ECC)));
-  std::strncpy(cert.header.name, name.c_str(), 0x40);
+  memcpy(cert.header.name, name.c_str(), 0x40);
   cert.header.id = Common::swap32(key_id);
   cert.public_key = Common::ec::PrivToPub(private_key);
   return cert;
--- a/Source/Core/Core/IOS/IOSC.cpp
+++ b/Source/Core/Core/IOS/IOSC.cpp
@@ -420,9 +420,9 @@ static CertECC MakeBlankEccCert(const std::string& issuer, const std::string& na
 {
   CertECC cert{};
   cert.signature.type = SignatureType(Common::swap32(u32(SignatureType::ECC)));
-  memcpy(cert.signature.issuer, issuer.c_str(), 0x40);
+  std::strncpy(cert.signature.issuer, issuer.c_str(), 0x40 - 1);
   cert.header.public_key_type = PublicKeyType(Common::swap32(u32(PublicKeyType::ECC)));
-  memcpy(cert.header.name, name.c_str(), 0x40);
+  std::strncpy(cert.header.name, name.c_str(), 0x40 - 1);
   cert.header.id = Common::swap32(key_id);
   cert.public_key = Common::ec::PrivToPub(private_key);
   return cert;
--- a/Source/Core/Core/PowerPC/JitCommon/JitCache.h
+++ b/Source/Core/Core/PowerPC/JitCommon/JitCache.h
@@ -17,17 +17,9 @@
 
 class JitBase;
 
-// A JitBlock is block of compiled code which corresponds to the PowerPC
-// code at a given address.
-//
-// The notion of the address of a block is a bit complicated because of the
-// way address translation works, but basically it's the combination of an
-// effective address, the address translation bits in MSR, and the physical
-// address.
-struct JitBlock
+#pragma pack(push, 1)
+struct JitBlockData
 {
-  bool OverlapsPhysicalRange(u32 address, u32 length) const;
-
   // A special entry point for block linking; usually used to check the
   // downcount.
   u8* checkedEntry;
@@ -49,6 +41,22 @@ struct JitBlock
   // The number of PPC instructions represented by this block. Mostly
   // useful for logging.
   u32 originalSize;
+  // This tracks the position if this block within the fast block cache.
+  // We allow each block to have only one map entry.
+  size_t fast_block_map_index;
+};
+#pragma pack(pop)
+
+// A JitBlock is block of compiled code which corresponds to the PowerPC
+// code at a given address.
+//
+// The notion of the address of a block is a bit complicated because of the
+// way address translation works, but basically it's the combination of an
+// effective address, the address translation bits in MSR, and the physical
+// address.
+struct JitBlock : public JitBlockData
+{
+  bool OverlapsPhysicalRange(u32 address, u32 length) const;
 
   // Information about exits to a known address from this block.
   // This is used to implement block linking.
@@ -73,10 +81,6 @@ struct JitBlock
     u64 ticStart;
     u64 ticStop;
   } profile_data = {};
-
-  // This tracks the position if this block within the fast block cache.
-  // We allow each block to have only one map entry.
-  size_t fast_block_map_index;
 };
 
 typedef void (*CompiledCode)();
--- a/Source/Core/Core/PowerPC/JitCommon/JitCache.h
+++ b/Source/Core/Core/PowerPC/JitCommon/JitCache.h
@@ -17,7 +17,6 @@
 
 class JitBase;
 
-#pragma pack(push, 1)
 struct JitBlockData
 {
   // A special entry point for block linking; usually used to check the
@@ -45,7 +44,6 @@ struct JitBlockData
   // We allow each block to have only one map entry.
   size_t fast_block_map_index;
 };
-#pragma pack(pop)
 
 // A JitBlock is block of compiled code which corresponds to the PowerPC
 // code at a given address.
--- a/Source/Core/Core/State.cpp
+++ b/Source/Core/Core/State.cpp
@@ -342,7 +342,7 @@ static void CompressAndDumpState(CompressAndDumpState_args save_args)
 
   // Setting up the header
   StateHeader header;
-  strncpy(header.gameID, SConfig::GetInstance().GetGameID().c_str(), 6);
+  memcpy(header.gameID, SConfig::GetInstance().GetGameID().c_str(), 6);
   header.size = g_use_compression ? (u32)buffer_size : 0;
   header.time = Common::Timer::GetDoubleTime();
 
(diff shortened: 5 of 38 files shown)


Source/Core/Common/ChunkFile.h
```diff
@@ -312,6 +312,7 @@ class PointerWrap
       break;
 
     case MODE_WRITE:
+#pragma GCC diagnostic ignored "-Wmaybe-uninitialized"
```
lioncash: This will not work on Windows and doesn't actually resolve the warning.

Source/Core/Core/HW/GCMemcard/GCMemcard.cpp
```diff
@@ -928,7 +928,7 @@ GCMemcardRemoveFileRetVal GCMemcard::RemoveFile(u8 index)  // index in the direc
   // here that has an empty file with the filename "Broken File000" where the actual deleted file
   // was. Determine when exactly this happens and if this is neccessary for anything.
 
-  memset(&(UpdatedDir.m_dir_entries[index]), 0xFF, DENTRY_SIZE);
+  memset(static_cast<void*>(&(UpdatedDir.m_dir_entries[index])), 0xFF, DENTRY_SIZE);
```
lioncash: This should probably not actually be silenced like this and instead have the initializtion and/or invalidation be done properly within a member function.

Source/Core/Core/IOS/IOSC.cpp
```diff
@@ -420,9 +420,9 @@ static CertECC MakeBlankEccCert(const std::string& issuer, const std::string& na
 {
   CertECC cert{};
   cert.signature.type = SignatureType(Common::swap32(u32(SignatureType::ECC)));
-  std::strncpy(cert.signature.issuer, issuer.c_str(), 0x40);
+  memcpy(cert.signature.issuer, issuer.c_str(), 0x40);
```
lioncash: These are copying into a string buffer. strncpy is appropriate here. This doesn't actually resolve any potential warning and just hides it instead. This isn't adequate at all.

Source/Core/Core/PowerPC/JitCommon/JitCache.h
```diff
@@ -17,17 +17,9 @@
...
 
-// A JitBlock is block of compiled code which corresponds to the PowerPC
-// code at a given address.
-//
-// The notion of the address of a block is a bit complicated because of the
-// way address translation works, but basically it's the combination of an
-// effective address, the address translation bits in MSR, and the physical
-// address.
-struct JitBlock
+#pragma pack(push, 1)
```
lioncash: Why is packing necessary here? This can also result in more inefficient code being generated. It also makes it way easier for memory alignment violations to occur.

Source/Core/Core/State.cpp
```diff
@@ -342,7 +342,7 @@ static void CompressAndDumpState(CompressAndDumpState_args save_args)
 
   // Setting up the header
   StateHeader header;
-  strncpy(header.gameID, SConfig::GetInstance().GetGameID().c_str(), 6);
+  memcpy(header.gameID, SConfig::GetInstance().GetGameID().c_str(), 6);
```
lioncash: Again, this is copying into a string buffer, using the `str` variant of copying is correct here.

Source/Core/InputCommon/ControllerEmu/ControlGroup/AnalogStick.cpp
```diff
@@ -56,14 +56,14 @@ ControlState AnalogStick::GetGateRadiusAtAngle(double ang) const
   return m_stick_gate->GetRadiusAtAngle(ang);
 }
 
-OctagonAnalogStick::OctagonAnalogStick(const char* name, ControlState gate_radius)
-    : OctagonAnalogStick(name, name, gate_radius)
+OctagonAnalogStick::OctagonAnalogStick(const char* _name, ControlState _gate_radius)
```
lioncash: We prefer to prefix names with the underscore after the identifier like so: `name_`

Ditto for all other occurrences

Source/Core/VideoBackends/Software/Rasterizer.cpp
```diff
@@ -320,11 +320,11 @@ void DrawTriangleFrontFace(const OutputVertexData* v0, const OutputVertexData* v
     scissorTop = 0;
 
   s32 scissorRight = bpmem.scissorBR.x - xoff - 341;
-  if (scissorRight > EFB_WIDTH)
+  if (scissorRight > (s32)EFB_WIDTH)
```
lioncash: We prefer either `static_cast` or `s32(thing)` forms of casting over C-style casting.

Source/Core/VideoCommon/BPStructs.cpp
```diff
@@ -43,7 +43,7 @@ static const float s_gammaLUT[] = {1.0f, 1.7f, 2.2f, 1.0f};
 
 void BPInit()
 {
-  memset(&bpmem, 0, sizeof(bpmem));
+  memset(static_cast<void*>(&bpmem), 0, sizeof(bpmem));
```
lioncash: This is, again, hiding an issue rather than taking the time to fix it properly.

Source/Core/VideoCommon/ShaderCache.cpp
```diff
@@ -183,7 +183,7 @@ template <typename SerializedUidType, typename UidType>
 static void SerializePipelineUid(const UidType& uid, SerializedUidType& serialized_uid)
 {
   // Convert to disk format. Ensure all padding bytes are zero.
-  std::memset(&serialized_uid, 0, sizeof(serialized_uid));
+  std::memset(static_cast<void*>(&serialized_uid), 0, sizeof(serialized_uid));
```
lioncash: Ditto

Source/Core/VideoCommon/VideoBackendBase.cpp
```diff
@@ -277,8 +277,8 @@ void VideoBackendBase::DoState(PointerWrap& p)
 
 void VideoBackendBase::InitializeShared()
 {
-  memset(&g_main_cp_state, 0, sizeof(g_main_cp_state));
-  memset(&g_preprocess_cp_state, 0, sizeof(g_preprocess_cp_state));
+  memset(static_cast<void*>(&g_main_cp_state), 0, sizeof(g_main_cp_state));
```
lioncash: Ditto

Source/Core/Common/ChunkFile.h
howard0su: Sure. In order to clean up this, it may require significant rewrite of this file.

Source/Core/Core/IOS/IOSC.cpp
howard0su: I believe signature issuer is not a NULL-terminated string. am I right?

Source/Core/Core/PowerPC/JitCommon/JitCache.h
howard0su: okay. I will remove this.

Source/Core/Core/State.cpp
howard0su: std::array<char, 6> gameID;  // The Game ID
This is not a string. GameID is not null terminate string at all.

Source/Core/InputCommon/ControllerEmu/ControlGroup/AnalogStick.cpp
howard0su: I use this style because in ControlGroup subfolder, there is exiting code like this. If you prefer other way, I can submit a seperate patch to cleanup them.

Source/Core/VideoBackends/Software/Rasterizer.cpp
howard0su: okay

Source/Core/VideoCommon/BPStructs.cpp
howard0su: okay. I will fix it.

Source/Core/Core/HW/GCMemcard/GCMemcard.cpp
howard0su: > While I appreciate the gusto on trying to clean up some of these warnings, some of these changes seem to just silence warnings rather than properly fixing the underlying cause of said warnings. Particularly with ones that cast the destination type in memset calls to `void*`.
> 
> The commit messages also don't state what warnings were actually being resolved within them.

Your comments are fair. Let me split the patch and fix some places as you suggested.

Source/Core/Core/IOS/IOSC.cpp
howard0su: you are right. I will fix.

Source/Core/Core/State.cpp
lioncash: `gameID` isn't a std::array. `std::string` also provides copying to buffers as part of its interface:

```cpp
SConfig::GetInstance().GetGameID().copy(header.gameID, 6);
```

Source/Core/DiscIO/DiscExtractor.cpp
```diff
@@ -325,8 +325,8 @@ bool ExportDOL(const Volume& volume, const Partition& partition, const std::stri
 std::optional<u64> GetFSTOffset(const Volume& volume, const Partition& partition)
 {
   const Platform volume_type = volume.GetVolumeType();
-  if (!IsDisc(volume_type))
-    return {};
+  if (IsDisc(volume_type))
```
lioncash: Is removing the unary not here intentional?

howard0su: no. typo. good catch

Source/Core/Core/State.cpp
howard0su: Check this: SConfig is a different class here.

struct StateHeader
{
  char gameID[6];
  u32 size;
  double time;
};

 // Setting up the header
  StateHeader header;
  memcpy(header.gameID, SConfig::GetInstance().GetGameID().c_str(), 6);

lioncash: I don't particularly see how that's relevant. Regardless of what it is (copying characters into a character buffer), this is directly [provided](https://en.cppreference.com/w/cpp/string/basic_string/copy) by `std::string`, so using `memcpy` or `strnlen` isn't even necessary here.

howard0su: are u confused by the src and dest?
char *strncpy(char *dest, const char *src, size_t n);

Source/Core/InputCommon/ControllerEmu/ControlGroup/AnalogStick.cpp
BhaaLseN: Why did you have to change this anyways? Members should have the `m_` prefix already and not cause any shadowing with the parameters.
In case I did miss some, use a suffix as lioncash suggested.

howard0su: sure. I can make a big change to add a prefix m_ to the fields in the class of ControlGroup

Source/Core/Core/State.cpp
JosJuice: `strncpy` would be correct here. We cannot use `memcpy`, because there is a possibility that the memory pointed to by `SConfig::GetInstance().GetGameID().c_str()` is less than 6 bytes, making us read invalid memory. As for `std::string::copy`, doesn't this function not write a null terminator to the destination? That is something we would want in case the game ID is shorter than 6 bytes. (Or has the entire `gameID` array been set to null bytes at some prior point?)

Source/Core/InputCommon/ControllerEmu/ControlGroup/AnalogStick.cpp
BhaaLseN: Please don't. I only see `name` being used there, so I don't quite see the need to rename the others (like `gate_radius`) though?

Source/Core/InputCommon/ControllerInterface/Xlib/XInput2.cpp
```diff
@@ -131,8 +131,6 @@ void KeyboardMouse::SelectEventsForDevice(Window window, XIEventMask* mask, int
 KeyboardMouse::KeyboardMouse(Window window, int opcode, int pointer, int keyboard)
     : m_window(window), xi_opcode(opcode), pointer_deviceid(pointer), keyboard_deviceid(keyboard)
 {
-  memset(&m_state, 0, sizeof(m_state));
```
BhaaLseN: Any reason you didn't just move this to the declaration and do
```c++
private:
  //...
  State m_state{};
```
in `Xinput2.h` instead?

Source/Core/VideoBackends/Software/SetupUnit.cpp
```diff
@@ -25,7 +25,6 @@ void SetupUnit::Init(u8 primitiveType)
 
 OutputVertexData* SetupUnit::GetVertex()
 {
-  memset(m_VertWritePointer, 0, sizeof(*m_VertWritePointer));
```
BhaaLseN: Doesn't this change the behavior if you remove it?
What was the warning you're trying to fix here?

Source/Core/InputCommon/ControllerEmu/ControlGroup/AnalogStick.cpp
howard0su: so only change name and ui_name here?

Source/Core/Core/State.cpp
howard0su: GameID is not a STRING. It is just handled as a string in the SConfig. When we convert the string back to its native format, we should do some special handling here. Strncpy doesn't make sense as we are copy string to a non-string buffer.

Maybe I should introduce a new type gameid_t to abstract this out.

Source/Core/InputCommon/ControllerEmu/ControlGroup/AnalogStick.cpp
BhaaLseN: Only the ones necessary to resolve the shadowing.

Source/Core/Core/State.cpp
JosJuice: `gameID` is a string buffer.

Source/Core/InputCommon/ControllerInterface/Xlib/XInput2.cpp
howard0su: good point. fixed

Source/Core/VideoBackends/Software/SetupUnit.cpp
howard0su: first, m_vertWritePointer is not a simple structure so that we cannot do memset.

the usage of this function is only one place:
    // transform this vertex so that it can be used for rasterization (outVertex)
    OutputVertexData* outVertex = m_setup_unit.GetVertex();
    TransformUnit::TransformPosition(&m_vertex, outVertex);
    outVertex->normal = {};
    if (VertexLoaderManager::g_current_components & VB_HAS_NRM0)
    {
      TransformUnit::TransformNormal(
          &m_vertex, (VertexLoaderManager::g_current_components & VB_HAS_NRM2) != 0, outVertex);
    }
    TransformUnit::TransformColor(&m_vertex, outVertex);
    TransformUnit::TransformTexCoord(&m_vertex, outVertex, m_tex_gen_special_case);

So all fields is override in the following code path. So I believe we should remove the memset completely. I test the software video rendering, it works fine.

howard0su:
Maybe it is easy for me to split this change into smaller ones.

Source/Core/Core/State.cpp
howard0su: (lldb) p header->gameID
(char [6]) $2 = "SBLE5G"

You can find the gameID is a char[6] that doesn't have space for '\0'. Do you think this is a string still?

JosJuice: It is a string buffer which contains strings that are not necessarily null terminated. There's no rule that says that bytes in memory representing a string have to end with a null pointer – it is a convention and not a requirement. Though of course, if the string is not null terminated, you have to be careful with what functions you pass the string to. In our case, we only pass it to `strncmp` while making sure to specify the size as 6, so it's fine.

But regardless of whether you consider `gameID` to contain a string, we clearly cannot use `memcpy` for the reason I described in my first comment, otherwise there is a risk of undefined behavior.

howard0su: make sense. Let me think how to fix this.

lioncash: I quite literally gave you a solution ten hours ago. You can even dehardcode the 6 in the process.

```cpp
SConfig::GetInstance().GetGameID().copy(header.gameID, std::size(header.gameID));
```

JosJuice: Just make sure to also change `StateHeader header;` to `StateHeader header{};` if you use that solution.

Source/Core/VideoCommon/CPMemory.h
```diff
@@ -255,6 +257,46 @@ struct CPState final
   bool bases_dirty;
   VertexLoaderBase* vertex_loaders[8];
   int last_id;
+
+  CPState & operator=(std::nullptr_t)
```
lioncash: This seems to be the default behavior of aggregate initialization.

Why doesn't e.g.:

```cpp
CPState state;

// ...

state = {};
```

work?

Source/Core/VideoCommon/CPMemory.h
```diff
@@ -255,6 +257,46 @@ struct CPState final
...
+      last_id = 0;
+      for(auto &l : vertex_loaders)
+      {
+          l = nullptr;
+      }
+
+      return *this;
+  }
+
+  CPState & operator=(const CPState& new_state)
```
lioncash: This is performing the same behavior as the default copy constructor, so I don't see why this whole function needs to be hardcoded in.

Source/Core/VideoCommon/VideoBackendBase.cpp
```diff
@@ -277,8 +277,10 @@ void VideoBackendBase::DoState(PointerWrap& p)
 
 void VideoBackendBase::InitializeShared()
 {
-  memset(&g_main_cp_state, 0, sizeof(g_main_cp_state));
-  memset(&g_preprocess_cp_state, 0, sizeof(g_preprocess_cp_state));
+  // Initialize cp state
+  g_main_cp_state = nullptr;
```
lioncash:
```suggestion
  g_main_cp_state = {};
  g_preprocess_cp_state = {};
```

Source/Core/VideoCommon/CPMemory.h
howard0su: yes. You are right.

howard0su:
start splitting this PR. Keep this open for comments. I split the code already reviewed into new PRs. please help check.

leoetlino:
Closing this since this PR has already been split and you can still comment on closed PRs if necessary.

PR #8685, opened 2020-03-22
Add back previous value comparison

MisterSquishy:
## description
adds back the previous value comparison in cheats manager:
Screen Shot 2020-03-22 at 10 50 16

this is available in the latest stable version but it seems the cheats manager has been completely rewritten since then. i didn't do too much git detective work but here's where the original functionality was added: https://github.com/dolphin-emu/dolphin/pull/3412 -- i tried to somewhat follow the pattern there but the `Result` struct appears to be completely different now, plus we're always looking up by memory address now so it seemed easier and more efficient to create a new store mapping addresses to values

## testing
i verified the functionality works by using it to find the memory location that stores selected character ID on the super mario strikers character selection screen. i also confirmed this address was right by editing the value and see the cursor move accordingly

### note
i noticed a fair number of PPC invalid read crashes while testing. it's not clear to me that this is caused by my changes since nothing i added incurs additional memory reads, but im open to the possibility since i know im not particularly diligent about cleaning up the `m_prev_values` vector. i dont have debugging set up particularly well (probably evident that im new to the C++ game) but im happy to dig in further if this isnt a known issue with the cheats manager.


JosJuice:
@dolphin-emu-bot rebuild

JosJuice:
Thanks for working on this, but I believe this also has been implemented in PR #8311.

MisterSquishy:
Sweeeeeeeeeeet any idea why that hasn't merged?

JosJuice:
It takes a while for PRs to get reviewed and merged sometimes, unfortunately. I don't know of any reason why to not merge that PR, so hopefully it will be merged sooner or later.

PR #8686, opened 2020-03-22
Debugger: Add a Thread widget

sepalani:
~This a very early WIP PR.~ **This PR is ready to be reviewed.** I mostly would like feedback regarding the features that might be wanted for this widget and also the design since I'm really not a UI/UX person.

The goal of this new Widget is to display user thread information (not kernel/IOS thread). Regarding the pieces of information I can get:
 - Using the [Dolphin OS Globals](http://hitmen.c02.at/files/yagcd/yagcd/chap4.html#sec4.2.1)
    * current/defaut thread address
    * thread queue address
 - Using OSThread structure
    * thread's context (saved registers)
    * thread's stack
    * thread's state and attr
    * thread's priority
    * thread's error value
    * thread's data (often used by 3rd party libraries to store the thread name)
    * thread's callstack 
    * ...
 - Using LWPThread structure
    * Planned but out-of-scope of this PR 

Screenshot:
image

Addresses can be right-clicked to toggle a context menu:
image

Any suggestions are welcome.


Source/Core/DolphinQt/Debugger/ThreadWidget.cpp
```diff
@@ -0,0 +1,147 @@
...
+void ThreadWidget::showEvent(QShowEvent* event)
+{
+  Update();
+}
+
+void ThreadWidget::CreateWidgets()
+{
+  m_state = new QGroupBox(tr("State"));
+
+  auto create_line_edit = []() {
```
lioncash:
```suggestion
  const auto create_line_edit = [] {
```

Source/Core/DolphinQt/Debugger/ThreadWidget.cpp
```diff
@@ -0,0 +1,147 @@
...
+{
+  // TODO
+}
+
+void ThreadWidget::Update()
+{
+  if (!isVisible() || Core::GetState() != Core::State::Paused)
+    return;
+
+  auto get_address_as_text = [](u32 addr) {
```
lioncash:
```suggestion
  const auto get_address_as_text = [](u32 addr) {
```

Source/Core/DolphinQt/Debugger/ThreadWidget.h
```diff
@@ -0,0 +1,41 @@
...
+
+#pragma once
+
+#include <QDockWidget>
+
+#include "Common/CommonTypes.h"
+
+class QGroupBox;
+class QLineEdit;
+class QCloseEvent;
```
lioncash: These should be alphabetically organized

Source/Core/DolphinQt/Settings.cpp
```diff
@@ -358,6 +358,21 @@ void Settings::SetRegistersVisible(bool enabled)
...
 }
 
+bool Settings::IsThreadsVisible() const
+{
+  return GetQSettings().value(QStringLiteral("debugger/showthreads")).toBool();
+}
+
+void Settings::SetThreadsVisible(bool enabled)
+{
+  if (IsThreadsVisible() != enabled)
```
lioncash: This can be inverted for an early return and to unindent the contained code.

sepalani:
Still WIP, the active threads group box was added.

image

Data structures were reversed based on debug messages and assembly code from the GameCube ELF of Finding Nemo and the Wii ELF of Rayman Raving Rabbids.

The next steps will be to add the thread context group box and then add the context menu on table cells.

Source/Core/Core/Debugger/Thread.cpp
```diff
@@ -0,0 +1,159 @@
...
+
+// Context offsets based on the following functions:
+//  - OSSaveContext
+//  - OSSaveFPUContext
+//  - OSDumpContext
+//  - OSClearContext
+//  - OSExceptionVector
+void Dolphin_Debugger::Thread::ReadContext(u32 addr, OSContext* ctx)
+{
+  for (int i = 0; i < 32; i += 1)
```
lioncash, line 19: the array sizes can be queried from the arrays themselves, eliminating the need for magic numbers here.

It's also our convention to use `i++` or `++i` when incrementing by 1 within an iteration expression

Source/Core/Core/Debugger/Thread.h
```diff
@@ -0,0 +1,112 @@
...
+using OSMutexQueue = Queue<OSMutex>;
+using OSMutexLink = Link<OSMutex>;
+
+using ThreadList = std::vector<std::pair<u32, OSThread>>;
+
+struct OSContext
+{
+  enum class State : u16
+  {
+    HAS_FPU = 1,
```
lioncash, line 45: `HasFPU` and `HasException`

Source/Core/Core/Debugger/Thread.h
```diff
@@ -0,0 +1,112 @@
...
+{
+  OSThreadQueue thread_queue;  // Threads waiting to own the mutex
+  u32 owner_addr;              // Thread owning the mutex
+  s32 lock_count;              // Mutex lock count
+  OSMutexLink link;            // Used to traverse the thread's mutex queue
+  // OSLockMutex uses it to insert the acquired mutex at the end of the queue
+};
+
+void ReadContext(u32 addr, OSContext* ctx);
+void DumpContext(OSContext* ctx);
```
lioncash, line 104: Given this is dumping the context, this parameter can probably be const qualified. Ditto for `DumpMutex` and `DumpThread`

Source/Core/Core/Debugger/Thread.h
```diff
@@ -0,0 +1,112 @@
...
+void ReadMutex(u32 addr, OSMutex* mutex);
+void DumpMutex(OSMutex* mutex);
+void ReadThread(u32 addr, OSThread* thread);
+void DumpThread(OSThread* thread);
+bool IsThreadValid(OSThread* thread);
```
lioncash, line 109:
```suggestion
bool IsThreadValid(const OSThread* thread);
```

sepalani:
Compared to last time I added the thread's context and the thread's callstack:
image

Feature-wise I'm almost done. I need to do the dump functions and add dropdown menus on right-click.

I'm also planning to move the Threads functions to the PowerPC::debug_interface which IMO is more fitting. Then, I'll work on the UI/UX part and it should be ready to go.

sepalani:
This PR is ready to be reviewed.

Since last time:
 - I moved the Thread part to the DebugInterface
 - I added the context menu on right-click
 - I had to fix the WatchWidget to update the watch list on AddWatch

Source/Core/Common/Debug/OSThread.cpp
```diff
@@ -0,0 +1,205 @@
...
+
+// Context offsets based on the following functions:
+//  - OSSaveContext
+//  - OSSaveFPUContext
+//  - OSDumpContext
+//  - OSClearContext
+//  - OSExceptionVector
+void Common::Debug::OSContext::Read(u32 addr, OSContext* ctx)
+{
+  for (int i = 0; i < 32; i++)
```
lioncash, line 20: This loop and the ones below it can just query the array's size instead of hardcoding the value. `int` shouldn't be used here either since the functions expect an address, so this will cause sign mismatch warnings.

Source/Core/DolphinQt/Debugger/ThreadWidget.cpp
```diff
@@ -0,0 +1,481 @@
...
+  // Thread's context group
+  const Common::Debug::PartialContext empty_context;
+  UpdateThreadContext(&empty_context);
+}
+
+void ThreadWidget::UpdateThreadContext(const Common::Debug::PartialContext* context)
+{
+  const auto get_hex = [](const std::optional<u32>& value) {
+    if (!value)
+      return QStringLiteral("");
```
lioncash, line 350:
```suggestion
      return QString{};
```

Source/Core/DolphinQt/Debugger/ThreadWidget.cpp
```diff
@@ -0,0 +1,481 @@
...
+    return QStringLiteral("%1").arg(*value, 8, 16, QLatin1Char('0'));
+  };
+  const auto get_hex_idx = [](const auto& table, std::size_t index) {
+    if (!table || index >= table->size())
+      return QStringLiteral("");
```
lioncash, line 355:
```suggestion
      return QString{};
```

Source/Core/DolphinQt/Debugger/ThreadWidget.cpp
```diff
@@ -0,0 +1,481 @@
...
+    return QStringLiteral("%1").arg(table->at(index), 8, 16, QLatin1Char('0'));
+  };
+  const auto get_f64_idx = [](const auto& table, std::size_t index) {
+    if (!table || index >= table->size())
+      return QStringLiteral("");
```
lioncash, line 360:
```suggestion
      return QString{};
```

sepalani:
@lioncash 
Done.

Source/Core/Common/Debug/OSThread.h
```diff
@@ -0,0 +1,151 @@
...
+  u16 dummy;
+  State state;
+  std::array<u32, 8> gqr;
+  u32 psf_padding;
+  std::array<double, 32> psf;
+
+  static void Read(u32 addr, OSContext* ctx);
+};
+
+static_assert(std::is_standard_layout_v<OSContext>);
```
lioncash, line 63: Missed this in the previous review, but this is probably meant to be a `std::is_trivially_copyable_v`, not checking for standard layout.

Source/Core/Common/Debug/OSThread.h
```diff
@@ -0,0 +1,151 @@
...
+
+  s32 error;                    // errno value
+  std::array<u32, 2> specific;  // Pointers to data (can be used to store thread names)
+
+  static constexpr u32 STACK_MAGIC = 0xDEADBABE;
+  static void Read(u32 addr, OSThread* thread);
+  static bool IsValid(const OSThread* thread);
+};
+
+static_assert(std::is_standard_layout_v<OSThread>);
```
lioncash, line 103: Ditto

Source/Core/Common/Debug/OSThread.h
```diff
@@ -0,0 +1,151 @@
...
+  s32 GetErrno() const override;
+  std::string GetSpecific() const override;
+  bool IsValid() const override;
+
+private:
+  u32 m_address = 0;
+  OSThread m_thread;
+};
+
+static_assert(std::is_standard_layout_v<OSMutex>);
```
lioncash, line 147: Ditto. These should probably be moved up underneath `OSMutex` as well, not under `OSThreadView`

Source/Core/Common/Debug/OSThread.h
sepalani: IIRC, offsetof is UB if the struct/class isn't standard_layout. But you're also correct since I wanted to check if it's trivially copyable too.

sepalani: @lioncash Done.

sepalani:
Rebased on latest master, still ready to be reviewed & merged.

Source/Core/DolphinQt/Debugger/ThreadWidget.cpp
```diff
@@ -0,0 +1,481 @@
...
+  }
+  m_thread_table->resizeColumnsToContents();
+  m_thread_table->resizeRowsToContents();
+
+  // Thread's context group
+  const Common::Debug::PartialContext empty_context;
+  UpdateThreadContext(&empty_context);
+}
+
+void ThreadWidget::UpdateThreadContext(const Common::Debug::PartialContext* context)
```
leoetlino, line 346: I'd pass a const-ref instead of a pointer here. Pointer semantics don't seem useful here.

And then you can just do `UpdateThreadContext({})` in the other functions to reset.

```suggestion
void ThreadWidget::UpdateThreadContext(const Common::Debug::PartialContext& context)
```

Source/Core/DolphinQt/Debugger/ThreadWidget.cpp
```diff
@@ -0,0 +1,481 @@
...
+{
+  const auto get_hex = [](const std::optional<u32>& value) {
```
leoetlino, line 348: same comment about get -> format

sepalani:
@leoetlino 
Done.

Source/Core/Common/Debug/OSThread.cpp
```diff
@@ -0,0 +1,204 @@
+// Copyright 2020 Dolphin Emulator Project
+// Licensed under GPLv2+
+// Refer to the license.txt file included.
+
+#include "Common/Debug/OSThread.h"
+
+#include <algorithm>
+#include "fmt/format.h"
```
leoetlino, line 8: Sorry for missing this last time, but this should be

```suggestion
#include <fmt/format.h>
```

Source/Core/Common/Debug/Threads.h
```diff
@@ -0,0 +1,63 @@
...
+  virtual bool IsSuspended() const = 0;
+  virtual bool IsDetached() const = 0;
+  virtual s32 GetBasePriority() const = 0;
+  virtual s32 GetEffectivePriority() const = 0;
+  virtual u32 GetStackStart() const = 0;
+  virtual u32 GetStackEnd() const = 0;
+  virtual std::size_t GetStackSize() const = 0;
+  virtual s32 GetErrno() const = 0;
+  virtual std::string
+  GetSpecific() const = 0;  // Implementation specific, used to store arbitrary data
```
leoetlino, line 57: Could you put this comment on its own line, before `virtual std::string...`? Putting it here causes an awkward line break.

sepalani:
@leoetlino 
Done.

PR #8687, opened 2020-03-23
Jit: remove warning -Winvalid-offsetof

howard0su:
Remove the warning:
warning: offsetof within non-standard-layout type ‘JitBlock’ is conditionally-supported
JitBlock contains non-trival types now. Split the fields with trival
types that needs to be access from JIT code into JitBlockData structure.

### DIFF
--- a/Source/Core/Core/PowerPC/JitCommon/JitCache.h
+++ b/Source/Core/Core/PowerPC/JitCommon/JitCache.h
@@ -11,23 +11,17 @@
 #include <map>
 #include <memory>
 #include <set>
+#include <type_traits>
 #include <vector>
 
 #include "Common/CommonTypes.h"
 
 class JitBase;
 
-// A JitBlock is block of compiled code which corresponds to the PowerPC
-// code at a given address.
-//
-// The notion of the address of a block is a bit complicated because of the
-// way address translation works, but basically it's the combination of an
-// effective address, the address translation bits in MSR, and the physical
-// address.
-struct JitBlock
+// offsetof is only conditionally supported for non-standard layout types,
+// so this struct needs to have a standard layout.
+struct JitBlockData
 {
-  bool OverlapsPhysicalRange(u32 address, u32 length) const;
-
   // A special entry point for block linking; usually used to check the
   // downcount.
   u8* checkedEntry;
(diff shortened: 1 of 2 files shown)


BhaaLseN:
You should probably leave a comment next to `JitBlockData` that it exists to be a trivial type, and should stay that way. Otherwise we might run into a situation where this happens again or the two are merged (when someone doesn't know about this)

lioncash:
Even better yet, add:

```cpp
static_assert(std::is_standard_layout_v<JitBlockData>, "JitBlockData must have a standard layout");
```

underneath its definition.

howard0su:
My goal is to enable "warning as error" at the near future as well. So that we don't regress.

Source/Core/Core/PowerPC/JitCommon/JitCache.h
```diff
@@ -17,17 +17,11 @@
...
-// code at a given address.
-//
-// The notion of the address of a block is a bit complicated because of the
-// way address translation works, but basically it's the combination of an
-// effective address, the address translation bits in MSR, and the physical
-// address.
-struct JitBlock
+// A JitBlockData is data block of compiled code, which has to be in a
+// fixed layout so that JIT code is able to access the fields.
+// All the fields has to be trival types.
```
lioncash: This should really just be saying that the struct itself needs to be standard layout. Types within it can be non-trivial and offsetof will work completely fine.

CookiePLMonster:
@dolphin-emu-bot rebuild

CookiePLMonster:
LGTM once lint issues get fixed.

CookiePLMonster:
@dolphin-emu-bot rebuild

Source/Core/Core/PowerPC/JitCommon/JitCache.h
```diff
@@ -17,17 +17,11 @@
...
 
-// A JitBlock is block of compiled code which corresponds to the PowerPC
-// code at a given address.
-//
-// The notion of the address of a block is a bit complicated because of the
-// way address translation works, but basically it's the combination of an
-// effective address, the address translation bits in MSR, and the physical
-// address.
-struct JitBlock
+// A JitBlockData is data block of compiled code, which has to be in a
```
leoetlino, line 20:
```suggestion
// offsetof is only conditionally supported for non-standard layout types,
// so this struct needs to have a standard layout.
```

CookiePLMonster:
@dolphin-emu-bot rebuild

PR #8688, opened 2020-03-23
Remove warning of -Wstringop-truncation

howard0su:
Fix the right length when strncpy. And use specific version of copy
to copy GameID.


Source/Core/Core/IOS/IOSC.cpp
```diff
@@ -420,9 +420,9 @@ static CertECC MakeBlankEccCert(const std::string& issuer, const std::string& na
 {
   CertECC cert{};
   cert.signature.type = SignatureType(Common::swap32(u32(SignatureType::ECC)));
-  std::strncpy(cert.signature.issuer, issuer.c_str(), 0x40);
+  std::strncpy(cert.signature.issuer, issuer.c_str(), 0x40 - 1);
```
leoetlino: Might as well use std::string::copy here then?

PR #8689, opened 2020-03-23
Remove warnings of -Wsign-compare

howard0su:
Cast the variable to the coresponding type.

### DIFF
--- a/Source/Core/DolphinQt/Settings/AdvancedPane.cpp
+++ b/Source/Core/DolphinQt/Settings/AdvancedPane.cpp
@@ -177,10 +177,10 @@ void AdvancedPane::Update()
   const bool enable_custom_rtc_widgets = SConfig::GetInstance().bEnableCustomRTC && !running;
 
   const std::vector<PowerPC::CPUCore>& available_cpu_cores = PowerPC::AvailableCPUCores();
-  for (int i = 0; i < available_cpu_cores.size(); ++i)
+  for (size_t i = 0; i < available_cpu_cores.size(); ++i)
   {
     if (available_cpu_cores[i] == SConfig::GetInstance().cpu_core)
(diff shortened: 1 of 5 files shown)


stenzek:
@dolphin-emu-bot rebuild

Source/Core/DolphinQt/Settings/AdvancedPane.cpp
```diff
@@ -177,7 +177,7 @@ void AdvancedPane::Update()
   const bool enable_custom_rtc_widgets = SConfig::GetInstance().bEnableCustomRTC && !running;
 
   const std::vector<PowerPC::CPUCore>& available_cpu_cores = PowerPC::AvailableCPUCores();
-  for (int i = 0; i < available_cpu_cores.size(); ++i)
+  for (size_t i = 0; i < available_cpu_cores.size(); ++i)
   {
     if (available_cpu_cores[i] == SConfig::GetInstance().cpu_core)
       m_cpu_emulation_engine_combobox->setCurrentIndex(i);
```
lioncash: This change in the loop variable type will cause an implicit truncation to `int` here.

stenzek: Indeed, missed that one.

howard0su: since available_cpu_cores # is small. either approach is fine. I can change int i to uint i. or cast size() to int. both are pretty safe. thoughts?

lioncash: it's fine to change this line to:

```cpp
m_cpu_emulation_engine_combobox->setCurrentIndex(int(i));
```

We'll never have CPU backends exceed the max value of an int.

CookiePLMonster:
@dolphin-emu-bot rebuild

PR #8694, opened 2020-03-23
Cleanup warnings of -Wswitch

howard0su:
Add default branch to the switch-case.

### DIFF
--- a/Source/Core/VideoCommon/FramebufferShaderGen.cpp
+++ b/Source/Core/VideoCommon/FramebufferShaderGen.cpp
@@ -7,6 +7,8 @@
 #include <sstream>
 #include <string_view>
 
+#include "Common/Logging/Log.h"
+
 #include "VideoCommon/FramebufferManager.h"
 #include "VideoCommon/TextureDecoder.h"
 #include "VideoCommon/VertexShaderGen.h"
@@ -570,6 +572,15 @@ std::string GenerateTextureReinterpretShader(TextureFormat from_format, TextureF
   }
   break;
 
+  case TextureFormat::I4:
+  {
+    ss << "  float4 temp_value = ";
+    EmitTextureLoad(ss, 0, "coords");
+    ss << ";\n"
+          "  raw_value = uint(temp_value.r * 15.0);\n";
+  }
+  break;
+
   case TextureFormat::IA4:
   {
     ss << "  float4 temp_value = ";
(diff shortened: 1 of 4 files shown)


Source/Core/VideoCommon/FramebufferShaderGen.cpp
```diff
@@ -570,6 +570,14 @@ std::string GenerateTextureReinterpretShader(TextureFormat from_format, TextureF
   }
   break;
 
+  case TextureFormat::I4:
+  {
+    ss << "  float4 temp_value = ";
+    EmitTextureLoad(ss, 0, "coords");
+    ss << ";\n"
+          "  raw_value = uint(temp_value.r * 15.0);\n";
+  }
```
lioncash: This seems to be missing a break.

howard0su: you are right. Can you let me know how to test this code path? this switch missing many formats. and there is also no code to dealing with same src and dest format.

stenzek: There's only a handful of games which rely on this "reinterpreting-texture-data-from-one-format-to-another" behavior. The expression you've added seems fine, but difficult to know if anything will actually use it.

As for the existing formats, fifoci should catch any regressions there.

Source/Core/VideoCommon/FramebufferShaderGen.cpp
```diff
@@ -605,6 +613,9 @@ std::string GenerateTextureReinterpretShader(TextureFormat from_format, TextureF
           "  }\n";
   }
   break;
+
+  default:
+    return "{}\n";
```
stenzek: Could add please add a `WARN_LOG(VIDEO, "From format %u not supported", static_cast<u32>(from_format));` here in case this does get executed?

Source/Core/VideoCommon/FramebufferShaderGen.cpp
```diff
@@ -654,6 +665,8 @@ std::string GenerateTextureReinterpretShader(TextureFormat from_format, TextureF
           "  }\n";
   }
   break;
+  default:
+    return "{}\n";
```
stenzek: And same here.

CookiePLMonster:
@dolphin-emu-bot rebuild

PR #8695, opened 2020-03-23
HW: cleanup warnings of -Wtype-limits

howard0su:
address is unsigned, it is always greater or equal than zero.
Remove the asert of zero compare.


CookiePLMonster:
@dolphin-emu-bot rebuild

PR #8697, opened 2020-03-23
Input: Cleanup warnings of -Wshadow

howard0su:
Add prefix m_ to the fields name/ui_name/type
in ControlGroup class.

### DIFF
--- a/Source/Core/InputCommon/ControllerEmu/ControlGroup/ControlGroup.h
+++ b/Source/Core/InputCommon/ControllerEmu/ControlGroup/ControlGroup.h
@@ -89,9 +89,9 @@ public:
     return std::copysign(std::max(T{0}, std::abs(input) - deadzone) / (T{1} - deadzone), input);
   }
 
-  const std::string name;
-  const std::string ui_name;
-  const GroupType type;
+  const std::string m_name;
+  const std::string m_ui_name;
+  const GroupType m_type;
   const bool can_be_disabled;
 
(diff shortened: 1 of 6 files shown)


Source/Core/InputCommon/ControllerEmu/ControlGroup/ControlGroup.h
```diff
@@ -89,9 +89,9 @@ class ControlGroup
     return std::copysign(std::max(T{0}, std::abs(input) - deadzone) / (T{1} - deadzone), input);
   }
 
-  const std::string name;
-  const std::string ui_name;
-  const GroupType type;
+  const std::string m_name;
+  const std::string m_ui_name;
+  const GroupType m_type;
```
stenzek: This seems inconsistent. If you're going to change the member names, it's better to update all of them.

howard0su: @BhaaLseN mentioned #8684 that he prefers to only change the fields impacted.

leoetlino: Using the m_ prefix feels... wrong for (public) *struct* member variables. I'd just add a _ suffix to whatever constructor arguments require them.

leoetlino:
This seems to be obsolete; the warnings don't show up on master for me.

PR #8701, opened 2020-03-23
Wiimote: Cleanup warnings of -Wmissing-variable-declarations

howard0su:
While I am here, make g_wiimote_scanner as static as well.

### DIFF
--- a/Source/Core/Core/HW/WiimoteReal/WiimoteReal.cpp
+++ b/Source/Core/Core/HW/WiimoteReal/WiimoteReal.cpp
@@ -69,9 +69,9 @@ struct WiimotePoolEntry
 
 // Connected wii remotes are placed here when no open slot is set to "Real".
 // They are then automatically disconnected after some time.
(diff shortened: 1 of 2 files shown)


JosJuice:
@dolphin-emu-bot rebuild

Source/Core/Core/HW/WiimoteReal/WiimoteReal.cpp
```diff
@@ -69,9 +69,9 @@ struct WiimotePoolEntry
 
 // Connected wii remotes are placed here when no open slot is set to "Real".
 // They are then automatically disconnected after some time.
-std::vector<WiimotePoolEntry> g_wiimote_pool;
+static std::vector<WiimotePoolEntry> g_wiimote_pool;
```
stenzek: If it's static, you probably should change the prefix to `s_`.

stenzek:
@dolphin-emu-bot rebuild

PR #8704, opened 2020-03-24
SettingsHandler: Don't output null bytes

JosJuice:
https://bugs.dolphin-emu.org/issues/12019, take two.

### DIFF
--- a/Source/Core/Common/SettingsHandler.cpp
+++ b/Source/Core/Common/SettingsHandler.cpp
@@ -6,6 +6,7 @@
 
 #include "Common/SettingsHandler.h"
 
+#include <algorithm>
 #include <cstddef>
 #include <ctime>
 #include <iomanip>
@@ -79,12 +80,10 @@ void SettingsHandler::Decrypt()
     m_key = (m_key >> 31) | (m_key << 1);
   }
 
-  // Decryption done. Now get rid of all CR in the output.
-  // The decoded file is supposed to contain Windows line endings
-  // (CR-LF), but sometimes also contains CR-LF-LF endings which
-  // confuse the parsing code, so let's just get rid of all CR
-  // line endings.
-
+  // The decoded data normally uses CRLF line endings, but occasionally
+  // (see the comment in WriteLine), lines can be separated by CRLFLF.
+  // To handle this, we remove every CR and treat LF as the line ending.
+  // (We ignore empty lines.)
   decoded.erase(std::remove(decoded.begin(), decoded.end(), '\x0d'), decoded.end());
 }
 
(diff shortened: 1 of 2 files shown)


JosJuice:
We might want to revert PR #8680 at some point, but I would like to wait a little with doing that to let people migrate any null-containing setting.txt they have.

Leseratte10:
I just built this PR, and all four NANDs I tested worked properly. When playing out-of-region games I needed to enable the "Allow Mismatched Region Settings" setting, but when that was enabled, all the identifiers were identical to these from the NAND dump, including the serial number, no matter which region the game was from. So it looks like this PR completely fixes the issue, at least for all the NANDs I have available to test.

Leseratte10:
Just poking this PR again, is there a chance to get this merged soon?

Source/Core/Common/SettingsHandler.cpp
```diff
@@ -96,22 +95,33 @@ void SettingsHandler::Reset()
...
+  // a null byte, add an LF right before the line to prod the values into being different,
+  // just like Nintendo does. Due to the chosen key, LF itself never encodes into a null byte.
+  const auto begin = m_buffer.cbegin() + old_position;
+  const auto end = m_buffer.cbegin() + m_position;
+  if (std::find(begin, end, 0) != end)
+  {
+    m_key = old_key;
+    m_position = old_position;
+    WriteByte('\n');
+    WriteLine(str);
```
leoetlino, line 123: Hmm, is it possible for a string to always contain a null byte after encoding, no matter how many line feeds are added to change the key?

JosJuice: I don't think that's likely, but I haven't proven it to be impossible (unlike LF never turning into a null byte, which I did prove with a small program).

Leseratte10: I think that theoretically that could be possible, but I think it's highly unlikely. But in case it does happen, it might be a good idea to add some kind of fallback so that it only tries 32 times (all possible key bits, adding up to 31 newlines) and then prints an error to the log instead of trying again in an endless loop like it would now. 

Also, no idea if this is already implemented, if we assume that such a thing can happen it might also be a good idea to make sure that the encoded file doesn't exceed the maximum file size.

JosJuice: > Also, no idea if this is already implemented, if we assume that such a thing can happen it might also be a good idea to make sure that the encoded file doesn't exceed the maximum file size.

This is implemented. `WriteByte` does nothing if the file is full. This also means that we never will get an infinite loop, because if we recurse many times, `begin` will eventually be equal to `end`, making it impossible to recurse further.

Leseratte10: Oh, right, missed that part.

PR #8705, opened 2020-03-27
Update Mario Party Gecko Codes

NoraHanegan:
I authored the previous commits


JMC47:
@dolphin-emu-bot rebuild

leoetlino:
Please do not remove the first line of each GameINI (which contains a comment with the game name) and the line breaks between sections.

I don't feel strongly about removing the line breaks between cheat codes, but IMO it'd be best to leave those in for readability and consistency reasons.

CrankySupertoon:
Will do. The comments were meant to be there but yeah I’ll add back the lines. Just an extra * right?

PR #8706, opened 2020-03-29
Qt/CodeViewWidget: Don't use deprectaed QFontMetrics methods

spycrab:


leoetlino:
Looks like the FreeBSD and Ubuntu builders need to be updated to use a newer Qt?

leoetlino:
Fixed by #8789

PR #8709, opened 2020-03-30
Android: Split General settings

Ebola16:
- Android: Split General settings: General settings was a merge of a few submenus. I split them up to match the Win build, which should make users do less searching.

- Android: Reduce Submenu height by 25 percent: This will make users do less scrolling as more submenus are added. I also dropped submenu descriptions, as they were only used with the Stereoscopy submenu. I think unified Dolphin settings outweigh a definition of stereoscopy.

Screenshot for @MayImilae 
Screenshot_20200330-065131_Dolphin Emulator


Ebola16:
Can we get this merged too? This change is needed for some future PRs. Nobody has complained about the design change and it's easily reversible if somebody really doesn't like it. Also mentioning @JMC47 since this has already been approved.

Ebola16:
Merge reminder?

Ebola16:
Is there any reason why this isn't getting merged? It's needed for future PRs.

PR #8711, opened 2020-03-30
Set Insert SD Card default value to true

Ebola16:
Enabling the SD card is useful for loading / saving data for games such as Super Smash Bros. Brawl. I assume most users won't encounter any problems with this setting enabled. Was there a reason why the default was set to false? Also, the virtual SD card is created regardless of this setting.

### DIFF
--- a/Source/Core/Core/Config/MainSettings.cpp
+++ b/Source/Core/Core/Config/MainSettings.cpp
@@ -74,7 +74,7 @@ ConfigInfo<bool> GetInfoForSimulateKonga(u32 channel)
   return {{System::Main, "Core", fmt::format("SimulateKonga{}", channel)}, false};
 }
 
-const ConfigInfo<bool> MAIN_WII_SD_CARD{{System::Main, "Core", "WiiSDCard"}, false};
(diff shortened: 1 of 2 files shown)


mbc07:
> Also, the virtual SD card is created regardless of this setting.

I'm 95% sure Dolphin only generates `sd.raw` after you enable the option and boot a Wii game or if you enable the option while a Wii game is already running. Unless something changed, that has been the behavior since the feature was introduced, basically...

Ebola16:
> I'm 95% sure Dolphin only generates `sd.raw` after you enable the option and boot a Wii game or if you enable the option while a Wii game is already running. Unless something changed, that has been the behavior since the feature was introduced, basically...

Just checked again on 5.0-11824. Indeed the virtual SD card is created regardless of the Insert SD Card setting.

JosJuice:
Yes, it does seem like an SD card file gets created by IOS HLE as soon as the emulated software makes any use of SD card functionality, regardless of the setting. I don't know if this is the desired behavior, though.

JMC47:
I don't see any reason not to let it be plugged in if it's getting created.

PR #8712, opened 2020-03-30
Boot: Add ConsoleType enum class

sepalani:
This PR adds a ConsoleType enum class and replace the magic number (console type) used in boot.

### DIFF
--- a/Source/Core/Core/Core.h
+++ b/Source/Core/Core/Core.h
@@ -37,6 +37,59 @@ enum class State
   Starting,
 };
 
+// Console type values based on:
+//  - YAGCD 4.2.1.1.2
+//  - OSInit (GameCube ELF from Finding Nemo)
+//  - OSReportInfo (Wii ELF from Rayman Raving Rabbids)
+enum class ConsoleType : u32
+{
+  // 0x0XXXXXXX Retail units - Gamecube
+  HW1 = 1,
+  HW2 = 2,
+  LatestProductionBoard = 3,
+  Reserved = 4,
+
+  // 0x0XXXXXXX Retail units - Wii
+  PreProductionBoard0 = 0x10,    // Pre-production board 0
+  PreProductionBoard1 = 0x11,    // Pre-production board 1
+  PreProductionBoard2_1 = 0x12,  // Pre-production board 2-1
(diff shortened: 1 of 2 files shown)


JosJuice:
Please also update this line so that we don't have any hardcoded values in it: https://github.com/dolphin-emu/dolphin/blob/0cab317efb617d7a81dd12956b77fe6c48a5a1a2/Source/Core/Core/Boot/Boot_BS2Emu.cpp#L376

sepalani:
I'll try to find details about the `0x23` model. I assume there might be a model `0x21` and `0x22` as well.

Source/Core/Core/Core.h
```diff
@@ -37,6 +37,59 @@ enum class State
...
+  Reserved = 4,
+
+  // 0x0XXXXXXX Retail units - Wii
+  PreProductionBoard0 = 0x10,    // Pre-production board 0
+  PreProductionBoard1 = 0x11,    // Pre-production board 1
+  PreProductionBoard2_1 = 0x12,  // Pre-production board 2-1
+  PreProductionBoard2_2 = 0x20,  // Pre-production board 2-2
+  RVL_Retail1 = 0x21,
+  RVL_Retail2 = 0x22,
+  RVL_Retail3 = 0x23,
```
Pokechu22, line 59: Out of curiosity, how do these correspond to actual known hardware revisions (regular Wii RVL-001, Wii Family Edition RVL-101, and the Wii Mini RVL-201, and I guess also vWii)?

At least for the System Menu, I don't see any of these numbers hardcoded at all (it just uses `Retail %d` for any console type not specially handled if it's less than `0x10000000`, so 0x21 would give `Retail 33` to my understanding).  I haven't actually tested this though.  Also interestingly, System Menu 2.0U doesn't have RVA1 (though System Menu 4.3U does).

leoetlino:
Could you rebase this please?

Source/Core/Core/Core.h
sepalani: I only own1st-gen Wii, so I don't know if it does correspond or not to other hardware revisions. Though, it would be interesting to test, especially since BlueBomb can be used on a Wii Mini.

sepalani:
@leoetlino 
Rebased.

PR #8714, opened 2020-03-30
DolphinQt: Run tasks that use progress dialogs on separate threads

JosJuice:


JosJuice:
Apparently this not only makes the GUI more responsive (which was my intention with this PR) but also fixes a deadlock that was likely to happen if VolumeVerifier got a panic alert on one of its helper threads (most easily triggered by attempting to verify a correctly signed vWii WAD).

PR #8715, opened 2020-03-31
DolphinQt: Fix the panic alert deadlock (a.k.a. "Question" issue)

JosJuice:


iwubcode:
Is there an issue report for this?  Code seems fine

JosJuice:
https://bugs.dolphin-emu.org/issues/12031 is (at least partially) about this. There is no issue report with good reproduction steps, though. Most of the reports are just topics scattered around the forum. The best reproduction steps I have are that you need to get a panic alert from the CPU thread while in exclusive fullscreen. I was using Direct3D 11, in case it matters. I triggered the panic alert by booting Metroid Prime (GC) in single core.

e2zippo:
I'm still getting this, just tried to launch Muramasa Demon Blade using 5.0-11834.

JosJuice:
@e2zippo Does this happen with fullscreen on, off or both?

e2zippo:
> @e2zippo Does this happen with fullscreen on, off or both?
Just tried, happens with both

e2zippo:
> @e2zippo Does this happen with fullscreen on, off or both?

Happens with both!

DaveACG:
> > @e2zippo Does this happen with fullscreen on, off or both?
> 
> @JosJuice happens with both.

Same here. MP Gamecube, as well as Metroid Prime Trilogy (MP Gamecube versions hang after about 30 secs of functionality, MP Trilogy gives the question mark dialog when launching). Fullscreen or window. Backend doesn't seem to matter - DX11/12/Vulcan.

e2zippo:
Any updates on this, it's still happening, maybe it should be reopened, or could it be something else causing it?

JosJuice:
No updates, unfortunately. I don't know what's causing the problem in the remaining cases.

PRs can't be reopened – once they're merged, they're merged. I would suggest opening an issue report at https://bugs.dolphin-emu.org/projects/emulator. Then, if we figure out a solution for the remaining cases of this problem, a new PR can be opened.

DaveACG:
Sorry, forgot about this post until I got an email reminder of a reply. I actually did find what appears to the source of it - the **skip duplicate frames** option, specifically when it's left disabled. Below is a copy of my post from the dolphin forums:
__________

Below is a post I made on the Reddit dolphin forum, been experiencing this for a while and just now I believe I've found the culprit - the "Skip presenting duplicate frames" option.

This option was greatly appreciated when it appeared as it made 30fps games tolerable, Wind Waker/Luigi's Mansion's constant stutters before were virtually eliminated when this was disabled, so I just left it as such. Finally was able to narrow down it being the cause of my Metroid Prime troubles and why the builds after this feature were introduced all had the issue, guess most people just aren't using it that much. With it enabled (which is how it was before this option was introduced in Dolphin) the MP series plays normally now. Disabled it, boom back to the behavior I describe below.

My post on reddit:
________

Anyone else getting problems launching Metroid Prime (Gamecube) and MP Trilogy (Wii) with the releases over the past two months? I eventually went back to 11516 and they work perfectly with that release, but anything around the last two beta releases over the past couple of months causes the following:

Metroid Prime Gamecube - Starting the game full screen, video never progresses. If you switch the Windowed you can get in the game, but it will eventually hang after you've been in the game for 30sec-1 minute.

Metroid Prime Trilogy - Hangs on load 9 times out of 10, with "Question?" empty dialog, has to be killed through task manager.

Backend doesn't matter - tried Vulkan/D3D 11+12/OpenGl. Disabled any texture packs + other enhancements with no change. Verified integrity of iso.

Specs:

GTX 1660

Latest Nvidia Drivers

i5 9400f

16gb

JosJuice:
@DaveACG Reproduced. It seems like your problem happens when you are using dual core and a panic alert is shown from the GPU thread. This PR fixed a deadlock where the GUI thread was waiting for the CPU thread when the CPU thread was waiting for the GUI thread. The problem I managed to reproduce now is essentially the same situation but with the GPU thread instead of the CPU thread. I don't have any idea right away for how to fix this, but let me think about it a little.

JosJuice:
Submitted an attempt at a fix as PR #8763.

PR #8088, opened 2019-05-09
Allow interacting with the render window while configuring controllers

Pokechu22:
This change enables interacting with the render window while configuring controllers.  This is useful for e.g. trying out different configurations without having to close and reopen the configuration window (note that the parent controller settings window is _not_ modal).  Note that as-is it does _not_ allow interacting when "render to main window" is enabled.

I've done this by making the controller configuration window `WindowModal`, which allows alt+tabing to the render window, but prohibits interaction with parent windows (controller settings window and the main dolphin window).  `exec` acts the same as if `ApplicationModal` were used and makes it impossible to switch to _any_ window, including the render window.  (It also blocks until the window is closed, which I assume is why a temporary variable could be used).

The main reason to even make it modal in the first place is that it prevents opening a second window by going to the configuration dialog and clicking the button again.  An alternative approach would be to do the same thing that is done for the hotkey window:

https://github.com/dolphin-emu/dolphin/blob/5ca993330729935ada1b7c2f027be77994e5fcdd/Source/Core/DolphinQt/MainWindow.cpp#L1092-L1103

This requires an array here, though, and I think it would get complicated.  Additionally, though it allows interacting with the parent windows (i.e. the game when "render to main window" is enabled), the actual dialog is still always on top.  But if that's preferable behavior, I can change to it.

PR #8716, opened 2020-04-01
Delete properties dialog on close

Pokechu22:
Fixes the game file being locked until Dolphin closes.  The properties dialog *technically* didn't leak before this, since Qt will close it when its parent window closes, but that only happens when dolphin itself exits and that's way longer than it needs to be.  (See also: https://github.com/dolphin-emu/dolphin/pull/8088#issuecomment-491452256)

### DIFF
--- a/Source/Core/DolphinQt/GameList/GameList.cpp
+++ b/Source/Core/DolphinQt/GameList/GameList.cpp
@@ -431,6 +431,9 @@ void GameList::OpenProperties()
     return;
 
   PropertiesDialog* properties = new PropertiesDialog(this, *game);
+  // Since the properties dialog locks the game file, it's important to free it as soon as it's
+  // closed so that the file can be moved or deleted.
+  properties->setAttribute(Qt::WA_DeleteOnClose, true);
 
   connect(properties, &PropertiesDialog::OpenGeneralSettings, this, &GameList::OpenGeneralSettings);
 


Source/Core/DolphinQt/Config/PropertiesDialog.cpp
```diff
@@ -32,6 +32,7 @@ PropertiesDialog::PropertiesDialog(QWidget* parent, const UICommon::GameFile& ga
                           QString::fromStdString(game.GetGameID()),
                           QString::fromStdString(game.GetLongName())));
   setWindowFlags(windowFlags() & ~Qt::WindowContextHelpButtonHint);
+  setAttribute(Qt::WA_DeleteOnClose, true);
```
lioncash, line 35: This should probably have a comment accompanying it explaining why this behavior is needed.

JosJuice: Alternatively, maybe the code would be clear enough on its own without needing a comment if we move this line to where the `PropertiesDialog` gets allocated? That is how every other usage of `Qt::WA_DeleteOnClose` in Dolphin seems to be done.

Pokechu22: Added the comment, but I think what JosJuice mentions is better; I'll do that now.

Pokechu22: Changed the comment and moved it to the allocation point.  (For reference, the previous comment was `The properties dialog is not re-used, and maintains a lock on the game file until it's deleted, preventing the file from being moved or deleted.`)

PR #8717, opened 2020-04-01
VertexManagerBase: Skip drawing objects with mismatched xf/bp stages

stenzek:
Hardware tests have shown that if the number of texgens/channels do not match, you get garbage rendering. Presumably because the output registers from the XF stage are fed into the incorrect input registers for TEV/BP.

Currently, this causes Dolphin to crash/generate invalid shaders with an assertion failure in the hardware backends. Instead, we log an error.

Perhaps in the future we should just spit out all texgens/colors anyway from both stages, and let cross-stage optimization take care of DCE'ing it away. But doing so would require changing the UIDs and invalidating everyone's shader caches.

### DIFF
--- a/Source/Core/VideoCommon/VertexManagerBase.cpp
+++ b/Source/Core/VideoCommon/VertexManagerBase.cpp
@@ -344,6 +344,17 @@ void VertexManagerBase::Flush()
 
   m_is_flushed = true;
 
+  if (xfmem.numTexGen.numTexGens != bpmem.genMode.numtexgens ||
+      xfmem.numChan.numColorChans != bpmem.genMode.numcolchans)
+  {
+    ERROR_LOG(VIDEO,
+              "Mismatched configuration between XF and BP stages - %u/%u texgens, %u/%u colors. "
+              "Skipping draw. Please report on the issue tracker.",
+              xfmem.numTexGen.numTexGens, bpmem.genMode.numtexgens.Value(),
+              xfmem.numChan.numColorChans, bpmem.genMode.numcolchans.Value());
+    return;
+  }
+
 #if defined(_DEBUG) || defined(DEBUGFAST)
   PRIM_LOG("frame%d:\n texgen=%u, numchan=%u, dualtex=%u, ztex=%u, cole=%u, alpe=%u, ze=%u",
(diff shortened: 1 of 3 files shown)


iwubcode:
Just curious, what games were affected by this?

stenzek:
According to the discussion on IRC, Super Paper Mario at least.

Source/Core/VideoCommon/VertexManagerBase.cpp
```diff
@@ -344,6 +344,17 @@ void VertexManagerBase::Flush()
 
   m_is_flushed = true;
 
+  if (xfmem.numTexGen.numTexGens != bpmem.genMode.numtexgens ||
```
iwubcode, line 347: Does it hurt that this block (which can early return) is _after_ `m_is_flushed = true` ?

stenzek: Yeah, we want to skip the object, i.e. set it as flushed.

Sintendo:
Would it make sense to add this as a new GameQuirk and have analytics report this?

stenzek:
> Would it make sense to add this as a new GameQuirk and have analytics report this?

Yep, sounds like a good idea!

JosJuice:
Is this related to https://bugs.dolphin-emu.org/issues/10180?

stenzek:
@JosJuice yes, I believe so.

Source/Core/VideoCommon/VertexManagerBase.cpp
```diff
@@ -344,6 +345,31 @@ void VertexManagerBase::Flush()
...
+    // eventually simulate the behavior we have test cases for it.
+    if (xfmem.numTexGen.numTexGens != bpmem.genMode.numtexgens)
+    {
+      DolphinAnalytics::Instance().ReportGameQuirk(
+          GameQuirk::MISMATCHED_GPU_TEXGENS_BETWEEN_XF_AND_BP);
+    }
+    if (xfmem.numChan.numColorChans != bpmem.genMode.numcolchans)
+    {
+      DolphinAnalytics::Instance().ReportGameQuirk(
+          GameQuirk::MISMATCHED_GPU_TEXGENS_BETWEEN_XF_AND_BP);
```
sebastiaangoossens, line 367: Shouldn't this be MISMATCHED_GPU_COLORS_BETWEEN_XF_AND_BP?

PR #8718, opened 2020-04-01
Android TV: Fix crash when long pressing a game

JosJuice:
Also fixes the same crash when accessing the game details (which only can be accessed after long pressing a game). The problem was that we were not using a theme that had an AppCompat theme as a parent.

After fixing the crashes, I discovered that the game details dialog unfortunately uses white on white on Android TV, and I don't know how to fix this in a clean way. But I guess that can be left for another PR?


ds22x:
Hopefully this will get merged soon, even if it does have white on white dialogue currently.

PR #8719, opened 2020-04-01
AudioPane: Do not enable DPL II quality slider with HLE audio on init

CookiePLMonster:
This PR is a de facto follow up to #8664, since there was one more case where DPL II audio quality slider was enabled when it shouldn't be - if Audio panel was opened with HLE audio selected and DPL II checkbox checked (as a "leftover" from using LLE audio, most likely), slider would be enabled while the checkbox itself was disabled. Now it's back in sync, as only this one code path seemed to be out of sync (for this reason candidate for refactor maybe?)

image


FlorianPons:
Maybe linked to bug  #12036

CookiePLMonster:
Indeed it is, although it looks like in this case the user actually observed this wrong behaviour and expected it to work differently.

PR #8720, opened 2020-04-01
Android: Fix ATV on API 28+

webgeek1234:
This should fix #11944. But as I cannot get the gradle config in the project to cooperate with the android-sdk on my dev machine, I'm going to have to rely on the pr builder to test.

### DIFF
--- a/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/utils/Analytics.java
+++ b/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/utils/Analytics.java
@@ -44,7 +44,7 @@ public class Analytics
     sPrefsEditor.putBoolean(analyticsAsked, true);
     sPrefsEditor.apply();
 
-    new AlertDialog.Builder(context)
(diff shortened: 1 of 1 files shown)


webgeek1234:
Or... apparently I can't even do that. Can someone manually trigger the build or am I between a rock and a hard place here?

webgeek1234:
Nope, this isn't enough. It gives a different error, but still appcompat related. Asking around to see if I can find someone that understands this stuff.

webgeek1234:
So, this is getting more fun. Looks like the analytics alert dialog usage fails on atv because it uses appcompat, which isn't normally usable on leanback.

webgeek1234:
This revision is verified working on a SHIELD Android TV running Android 9.0.

javelinanddart:
Since there's no CLA of any kind, just wanted to put this here:

I accept the use of this code in this project and the licensing that comes with it.

JosJuice:
Ah, so the problem is when showing the alert... I see. I was assuming that it happened on activity launch all this time.

I'll do a quick merge of this and then amend PR #8718 to make the analytics alert use the theme defined in that PR (which also is an AppCompat theme). Thanks for fixing the issue!

PR #8721, opened 2020-04-03
Allow modifying MEM1 and MEM2 sizes at runtime via Dolphin.ini

Minty-Meeo:
Changing those enumerated values in the Memory namespace (REALRAM_SIZE, EXRAM_SIZE, etc) to global u32s seems to have actually boosted performance a bit.  A more thorough test is worth doing.

### DIFF
--- a/Source/Core/Core/HW/Memmap.cpp
+++ b/Source/Core/Core/HW/Memmap.cpp
@@ -31,6 +31,9 @@
 #include "Core/HW/WII_IPC.h"
 #include "Core/PowerPC/JitCommon/JitBase.h"
 #include "Core/PowerPC/PowerPC.h"
+#include <OptionParser.h>
+#include <UICommon/UICommon.h>
+#include <UICommon/CommandLineParse.h>
 #include "VideoCommon/CommandProcessor.h"
 #include "VideoCommon/PixelEngine.h"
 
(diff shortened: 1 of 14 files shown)


Source/Core/Core/HW/Memmap.cpp
```diff
@@ -57,6 +60,73 @@ u8* m_pL1Cache;
 u8* m_pEXRAM;
 u8* m_pFakeVMEM;
 
+static u32 get_MEM1Size()
+{
+  u32 MEM1_Size;
+  // I need to initialize some stuff that is normally initialized in
+  // int main(), because this funtion will be run before any of that.
+  optparse::Values options_temp;
```
JosJuice, line 68: Does default initializing `optparse::Values` actually produce any useful value, or is this essentially a more complicated way of passing an empty string to `SetUserDirectory`?

I'm rather skeptical to the whole approach of calling `SetUserDirectory`/`CreateDirectories`/`SConfig::Init` more than once.

AdmiralCurtiss:
Yeah I agree, this initialization is sketchy as hell. Is refactoring so those values are not initialized statically out of the question?

Also, you should not try to pass off non-constants as constants and rename all the relevant now-modifiable values to match the naming guidelines, `g_ram_size` or whatever.

Source/Core/Core/HW/Memmap.cpp
Minty-Meeo: I did little research into the workings of SetUserDirectory.  If all it needs is an empty string in order to set up the User Directory correctly (being able to locate the Dolphin.ini), I'll do that instead.

Minty-Meeo:
> Is refactoring so those values are not initialized statically out of the question?
Other statically initialized values (crucial ones, from the looks of it) need these values, and I didn't want to rewrite Dolphin as we know it.

Source/Core/Core/HW/Memmap.cpp
JosJuice: That depends on what you mean by correctly. If you give it an empty string, then it will locate Dolphin.ini in the location where it would be if you don't override the user directory with command-line arguments. This is fine most of the time but would be bad in the case where you do override the user directory with command-line arguments.

PR #8722, opened 2020-04-03
Configurable MEM1 and MEM2 sizes at runtime via Dolphin.ini

Minty-Meeo:
Changed several enums from Memmap.h to be static vars and implemented Get functions to query them. This seems to have boosted speed a bit in some titles? The new variables and some previously statically initialized items are now initialized via Memory::Init() and the new AddressSpace::Init(). s_ram_size_real and the new s_exram_size_real in particular are initialized from new OnionConfig values "MAIN_MEM1_SIZE" and "MAIN_MEM2_SIZE", only if "MAIN_RAM_OVERRIDE_ENABLE" is true.

GUI features have been added to Config > Advanced to adjust the new OnionConfig values.

A check has been added to State::doState to ensure savestates with memory configurations different from the current settings aren't loaded. The STATE_VERSION is now 115.

FIFO Files have been updated from version 4 to version 5, now including the MEM1 and MEM2 sizes from the time of DFF creation. FIFO Logs not using the new features (OnionConfig MAIN_RAM_OVERRIDE_ENABLE is false) are still backwards compatible. FIFO Logs that do use the new features have a MIN_LOADER_VERSION of 5. Thanks to the order of function calls, FIFO logs are able to automatically configure the new OnionConfig settings to match what is needed. This is a bit hacky, though, so I also threw in a failsafe for if the conditions that allow this to work ever go away.

I took the liberty of adding a log message to explain why the core fails to initialize if the MIN_LOADER_VERSION is too great.

Some IOS code has had the function "RAMOverrideForIOSMemoryValues" appended to it to recalculate IOS Memory Values from retail IOSes/apploaders to fit the extended memory sizes.  Worry not, if MAIN_RAM_OVERRIDE_ENABLE is false, this function does absolutely nothing.

A hotfix in DolphinQt/MenuBar.cpp has been implemented for RAM Override.

### DIFF
--- a/Source/Core/Core/FifoPlayer/FifoDataFile.cpp
+++ b/Source/Core/Core/FifoPlayer/FifoDataFile.cpp
@@ -11,12 +11,18 @@
 #include <vector>
 
 #include "Common/File.h"
+#include "Common/MsgHandler.h"
+#include "Core/Config/MainSettings.h"
+#include "Core/HW/Memmap.h"
 
 enum
 {
   FILE_ID = 0x0d01f1f0,
-  VERSION_NUMBER = 4,
+  VERSION_NUMBER = 5,
   MIN_LOADER_VERSION = 1,
+  // This value is only used if the DFF file was created with overridden RAM sizes.
+  // If the MIN_LOADER_VERSION ever exceeds this, it's alright to remove it.
+  MIN_LOADER_VERSION_FOR_RAM_OVERRIDE = 5,
 };
 
 #pragma pack(push, 1)
@@ -39,7 +45,11 @@ struct FileHeader
   u32 flags;
   u64 texMemOffset;
   u32 texMemSize;
-  u8 reserved[40];
+  // These are for overriden RAM sizes.  Otherwise the FIFO Player
+  // will crash and burn with mismatched settings.  See PR #8722.
+  u32 mem1_size;
+  u32 mem2_size;
+  u8 reserved[32];
 };
 static_assert(sizeof(FileHeader) == 128, "FileHeader should be 128 bytes");
 
@@ -129,7 +139,11 @@ bool FifoDataFile::Save(const std::string& filename)
   FileHeader header;
   header.fileId = FILE_ID;
   header.file_version = VERSION_NUMBER;
-  header.min_loader_version = MIN_LOADER_VERSION;
+  // Maintain backwards compatability so long as the RAM sizes aren't overridden.
+  if (Config::Get(Config::MAIN_RAM_OVERRIDE_ENABLE))
+    header.min_loader_version = MIN_LOADER_VERSION_FOR_RAM_OVERRIDE;
+  else
+    header.min_loader_version = MIN_LOADER_VERSION;
 
   header.bpMemOffset = bpMemOffset;
   header.bpMemSize = BP_MEM_SIZE;
@@ -150,6 +164,9 @@ bool FifoDataFile::Save(const std::string& filename)
   header.frameCount = (u32)m_Frames.size();
 
   header.flags = m_Flags;
+
+  header.mem1_size = Memory::GetRamSizeReal();
+  header.mem2_size = Memory::GetExRamSizeReal();
 
   file.Seek(0, SEEK_SET);
   file.WriteBytes(&header, sizeof(FileHeader));
@@ -198,8 +215,20 @@ std::unique_ptr<FifoDataFile> FifoDataFile::Load(const std::string& filename, bo
 
   if (header.fileId != FILE_ID || header.min_loader_version > VERSION_NUMBER)
   {
+    CriticalAlertT(
+        "The DFF's minimum loader version (%d) exceeds the version of this FIFO Player (%d)",
+        header.min_loader_version, VERSION_NUMBER);
     file.Close();
     return nullptr;
+  }
+
+  // Official support for overridden RAM sizes was added in version 5.
+  if (header.file_version < 5)
+  {
+    // It's safe to assume FIFO Logs before PR #8722 weren't using this
+    // obscure feature, so load up these header values with the old defaults.
+    header.mem1_size = Memory::MEM1_SIZE_RETAIL;
+    header.mem2_size = Memory::MEM2_SIZE_RETAIL;
   }
 
   auto dataFile = std::make_unique<FifoDataFile>();
@@ -209,8 +238,34 @@ std::unique_ptr<FifoDataFile> FifoDataFile::Load(const std::string& filename, bo
 
   if (flagsOnly)
   {
+    // Force settings to match those used when the DFF was created.  This is sort of a hack.
+    // It only works because this function gets called twice, and the first time (flagsOnly mode)
+    // happens to be before HW::Init().  But the convenience is hard to deny!
+    Config::SetCurrent(Config::MAIN_RAM_OVERRIDE_ENABLE, true);
+    Config::SetCurrent(Config::MAIN_MEM1_SIZE, header.mem1_size);
+    Config::SetCurrent(Config::MAIN_MEM2_SIZE, header.mem2_size);
+
     file.Close();
     return dataFile;
+  }
+
+  // To make up for such a hacky thing, here is a catch-all failsafe in case if the above code
+  // stops working or is otherwise removed.  As it is, this should never end up running.
+  // It should be noted, however, that Dolphin *will still crash* from the nullptr being returned
+  // in a non-flagsOnly context, so if this code becomes necessary, it should be moved above the
+  // prior conditional.
+  if (header.mem1_size != Memory::GetRamSizeReal() ||
+      header.mem2_size != Memory::GetExRamSizeReal())
+  {
+    CriticalAlertT("Emulated memory size mismatch!\n"
+                   "Current  | MEM1 %08X (%3dMB)    MEM2 %08X (%3dMB)\n"
+                   "DFF      | MEM1 %08X (%3dMB)    MEM2 %08X (%3dMB)",
+                   Memory::GetRamSizeReal(), Memory::GetRamSizeReal() / 0x100000,
+                   Memory::GetExRamSizeReal(), Memory::GetExRamSizeReal() / 0x100000,
+                   header.mem1_size, header.mem1_size / 0x100000, header.mem2_size,
+                   header.mem2_size / 0x100000);
+    file.Close();
+    return nullptr;
   }
 
   u32 size = std::min<u32>(BP_MEM_SIZE, header.bpMemSize);
@@ -237,6 +292,10 @@ std::unique_ptr<FifoDataFile> FifoDataFile::Load(const std::string& filename, bo
     file.Seek(header.texMemOffset, SEEK_SET);
     file.ReadArray(dataFile->m_TexMem, size);
   }
+
+  // idk what else these could be used for, but it'd be a shame to not make them available.
+  dataFile->m_ram_size_real = header.mem1_size;
+  dataFile->m_exram_size_real = header.mem2_size;
 
   // Read frames
   for (u32 i = 0; i < header.frameCount; ++i)
--- a/Source/Core/Core/HW/AddressSpace.cpp
+++ b/Source/Core/Core/HW/AddressSpace.cpp
@@ -234,6 +234,7 @@ struct AccessorMapping
 
 struct CompositeAddressSpaceAccessors : Accessors
 {
+  CompositeAddressSpaceAccessors() = default;
   CompositeAddressSpaceAccessors(std::initializer_list<AccessorMapping> accessors)
       : m_accessor_mappings(accessors.begin(), accessors.end())
   {
@@ -303,6 +304,7 @@ private:
 
 struct SmallBlockAccessors : Accessors
 {
+  SmallBlockAccessors() = default;
   SmallBlockAccessors(u8** alloc_base_, u32 size_) : alloc_base{alloc_base_}, size{size_} {}
 
   bool IsValidAddress(u32 address) const override
@@ -366,14 +368,11 @@ struct NullAccessors : Accessors
 
 static EffectiveAddressSpaceAccessors s_effective_address_space_accessors;
 static AuxiliaryAddressSpaceAccessors s_auxiliary_address_space_accessors;
-static SmallBlockAccessors s_mem1_address_space_accessors{&Memory::m_pRAM, Memory::REALRAM_SIZE};
-static SmallBlockAccessors s_mem2_address_space_accessors{&Memory::m_pEXRAM, Memory::EXRAM_SIZE};
-static SmallBlockAccessors s_fake_address_space_accessors{&Memory::m_pFakeVMEM,
-                                                          Memory::FAKEVMEM_SIZE};
-static CompositeAddressSpaceAccessors s_physical_address_space_accessors_gcn{
-    {0x00000000, &s_mem1_address_space_accessors}};
-static CompositeAddressSpaceAccessors s_physical_address_space_accessors_wii{
-    {0x00000000, &s_mem1_address_space_accessors}, {0x10000000, &s_mem2_address_space_accessors}};
+static SmallBlockAccessors s_mem1_address_space_accessors;
+static SmallBlockAccessors s_mem2_address_space_accessors;
+static SmallBlockAccessors s_fake_address_space_accessors;
+static CompositeAddressSpaceAccessors s_physical_address_space_accessors_gcn;
+static CompositeAddressSpaceAccessors s_physical_address_space_accessors_wii;
 static NullAccessors s_null_accessors;
 
 Accessors* GetAccessors(Type address_space)
@@ -411,6 +410,16 @@ Accessors* GetAccessors(Type address_space)
   }
 
   return &s_null_accessors;
+}
+
+void Init()
+{
+  s_mem1_address_space_accessors = {&Memory::m_pRAM, Memory::GetRamSizeReal()};
+  s_mem2_address_space_accessors = {&Memory::m_pEXRAM, Memory::GetExRamSizeReal()};
+  s_fake_address_space_accessors = {&Memory::m_pFakeVMEM, Memory::GetFakeVMemSize()};
+  s_physical_address_space_accessors_gcn = {{0x00000000, &s_mem1_address_space_accessors}};
+  s_physical_address_space_accessors_wii = {{0x00000000, &s_mem1_address_space_accessors},
+                                            {0x10000000, &s_mem2_address_space_accessors}};
 }
 
 }  // namespace AddressSpace
--- a/Source/Core/Core/HW/HW.cpp
+++ b/Source/Core/Core/HW/HW.cpp
@@ -10,6 +10,7 @@
 #include "Core/ConfigManager.h"
 #include "Core/Core.h"
 #include "Core/CoreTiming.h"
+#include "Core/HW/AddressSpace.h"
 #include "Core/HW/AudioInterface.h"
 #include "Core/HW/CPU.h"
 #include "Core/HW/DSP.h"
@@ -41,7 +42,8 @@ void Init()
   SerialInterface::Init();
   ProcessorInterface::Init();
   ExpansionInterface::Init();  // Needs to be initialized before Memory
-  Memory::Init();
+  Memory::Init();              // Needs to be initialized before AddressSpace
+  AddressSpace::Init();
   DSP::Init(SConfig::GetInstance().bDSPHLE);
   DVDInterface::Init();
   GPFifo::Init();
--- a/Source/Core/Core/HW/Memmap.cpp
+++ b/Source/Core/Core/HW/Memmap.cpp
@@ -10,6 +10,7 @@
 #include "Core/HW/Memmap.h"
 
 #include <algorithm>
+#include <array>
 #include <cstring>
 #include <memory>
 
@@ -18,6 +19,7 @@
 #include "Common/Logging/Log.h"
 #include "Common/MemArena.h"
 #include "Common/Swap.h"
+#include "Core/Config/MainSettings.h"
 #include "Core/ConfigManager.h"
 #include "Core/HW/AudioInterface.h"
 #include "Core/HW/DSP.h"
@@ -56,6 +58,70 @@ u8* m_pRAM;
 u8* m_pL1Cache;
 u8* m_pEXRAM;
 u8* m_pFakeVMEM;
+
+// s_ram_size is the amount allocated by the emulator, whereas s_ram_size_real
+// is what will be reported in lowmem, and thus used by emulated software.
+// Note: Writing to lowmem is done by IPL. If using retail IPL, it will
+// always be set to 24MB.
+static u32 s_ram_size_real;
+static u32 s_ram_size;
+static u32 s_ram_mask;
+static u32 s_fakevmem_size;
+static u32 s_fakevmem_mask;
+static u32 s_L1_cache_size;
+static u32 s_L1_cache_mask;
+static u32 s_io_size;
+// s_exram_size is the amount allocated by the emulator, whereas s_exram_size_real
+// is what gets used by emulated software.  If using retail IOS, it will
+// always be set to 64MB.
+static u32 s_exram_size_real;
+static u32 s_exram_size;
+static u32 s_exram_mask;
+
+u32 GetRamSizeReal()
+{
+  return s_ram_size_real;
+}
+u32 GetRamSize()
+{
+  return s_ram_size;
+}
+u32 GetRamMask()
+{
+  return s_ram_mask;
+}
+u32 GetFakeVMemSize()
+{
+  return s_fakevmem_size;
+}
+u32 GetFakeVMemMask()
+{
+  return s_fakevmem_mask;
+}
+u32 GetL1CacheSize()
+{
+  return s_L1_cache_size;
+}
+u32 GetL1CacheMask()
+{
+  return s_L1_cache_mask;
+}
+u32 GetIOSize()
+{
+  return s_io_size;
+}
+u32 GetExRamSizeReal()
+{
+  return s_exram_size_real;
+}
+u32 GetExRamSize()
+{
+  return s_exram_size;
+}
+u32 GetExRamMask()
+{
+  return s_exram_mask;
+}
 
 // MMIO mapping object.
 std::unique_ptr<MMIO::Mapping> mmio_mapping;
@@ -154,15 +220,9 @@ struct LogicalMemoryView
 //
 // Dolphin doesn't emulate the difference between cached and uncached access.
 //
-// TODO: The actual size of RAM is REALRAM_SIZE (24MB); the other 8MB shouldn't
-// be backed by actual memory.
+// TODO: The actual size of RAM is 24MB; the other 8MB shouldn't be backed by actual memory.
 // TODO: Do we want to handle the mirrors of the GC RAM?
-static PhysicalMemoryRegion physical_regions[] = {
-    {&m_pRAM, 0x00000000, RAM_SIZE, PhysicalMemoryRegion::ALWAYS},
-    {&m_pL1Cache, 0xE0000000, L1_CACHE_SIZE, PhysicalMemoryRegion::ALWAYS},
-    {&m_pFakeVMEM, 0x7E000000, FAKEVMEM_SIZE, PhysicalMemoryRegion::FAKE_VMEM},
-    {&m_pEXRAM, 0x10000000, EXRAM_SIZE, PhysicalMemoryRegion::WII_ONLY},
-};
+static std::array<PhysicalMemoryRegion, 4> physical_regions;
 
 static std::vector<LogicalMemoryView> logical_mapped_entries;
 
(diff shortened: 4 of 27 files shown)


JosJuice:
@dolphin-emu-bot rebuild

AdmiralCurtiss:
The comments about the static initialization from #8721 still apply here.

AdmiralCurtiss:
This also looks like it would horribly break if you tried to load a savestate with a different ram size than the current configuration.

JMC47:
Savestates need to have something added to protect against memory sizes.

Source/Core/Core/HW/Memmap.cpp
```diff
@@ -57,6 +60,71 @@ u8* m_pL1Cache;
 u8* m_pEXRAM;
 u8* m_pFakeVMEM;
 
+static u32 get_MEM1Size()
+{
+  u32 MEM1_Size;
+  // I need to initialize some stuff that is normally initialized in
+  // int main(), because this funtion will be run before any of that.
+  UICommon::SetUserDirectory("");
```
lioncash, line 68: This is kind of a large code smell. I don't think we should be calling UI specifics from Core.

Source/Core/Core/HW/Memmap.cpp
```diff
@@ -57,6 +60,71 @@ u8* m_pL1Cache;
 u8* m_pEXRAM;
 u8* m_pFakeVMEM;
 
+static u32 get_MEM1Size()
```
lioncash, line 63:
```suggestion
static u32 GetMEM1Size()
```

Source/Core/Core/HW/Memmap.cpp
```diff
@@ -57,6 +60,71 @@ u8* m_pL1Cache;
...
+{
+  u32 MEM1_Size;
```
lioncash, line 65: `mem1_size`, this should also be placed closer to its usage point.

Source/Core/Core/HW/Memmap.cpp
```diff
@@ -57,6 +60,71 @@ u8* m_pL1Cache;
...
+  else
+  {
+    // Retail MEM1 Size
+    MEM1_Size = 0x01800000;
+  }
+
+  SConfig::Shutdown();
+  return MEM1_Size;
+}
+static u32 get_MEM2Size()
```
lioncash, line 86: Same comments apply to this function as the ones for the one above it.

Source/Core/Core/HW/AddressSpace.cpp
```diff
@@ -303,6 +304,7 @@ struct CompositeAddressSpaceAccessors : Accessors
 
 struct SmallBlockAccessors : Accessors
 {
+  SmallBlockAccessors() : alloc_base() {}
```
lioncash, line 307: `alloc_base` can just be assigned nullptr directly in the class (along with the size being set to zero). This constructor can then just be defaulted.

Source/Core/Core/HW/AddressSpace.cpp
```diff
@@ -234,6 +234,7 @@ struct AccessorMapping
 
 struct CompositeAddressSpaceAccessors : Accessors
 {
+  CompositeAddressSpaceAccessors() : m_accessor_mappings() {}
```
lioncash, line 237:
```suggestion
  CompositeAddressSpaceAccessors() = default;
```

Source/Core/Core/HW/Memmap.cpp
```diff
@@ -189,6 +204,39 @@ static u32 GetFlags()
 
 void Init()
 {
+  auto GetMEM1Size = []()->u32 {
```
lioncash, line 207:
```suggestion
  const auto GetMEM1Size = []() -> u32 {
```

Source/Core/Core/HW/Memmap.cpp
```diff
@@ -189,6 +204,39 @@ static u32 GetFlags()
...
+    if (SConfig::GetInstance().m_RAMOverrideEnable == true)
+      // Configurable MEM1 Size
+      return SConfig::GetInstance().m_MEM1Size;
+    else
+      // Retail MEM1 Size
+      return (u32)0x01800000;
```
lioncash, line 213:
```suggestion
      return 0x01800000U;
```

Source/Core/Core/HW/Memmap.cpp
```diff
@@ -189,6 +204,39 @@ static u32 GetFlags()
 
 void Init()
 {
+  auto GetMEM1Size = []()->u32 {
+    if (SConfig::GetInstance().m_RAMOverrideEnable == true)
+      // Configurable MEM1 Size
```
lioncash, line 209: This if should be braced, since the body spans more than one line.

Source/Core/Core/HW/Memmap.cpp
```diff
@@ -189,6 +204,39 @@ static u32 GetFlags()
...
+      return SConfig::GetInstance().m_MEM1Size;
+    else
```
lioncash, line 211: This else can be removed.

Source/Core/Core/HW/Memmap.cpp
```diff
@@ -189,6 +204,39 @@ static u32 GetFlags()
 
 void Init()
 {
+  auto GetMEM1Size = []()->u32 {
+    if (SConfig::GetInstance().m_RAMOverrideEnable == true)
```
lioncash, line 208:
```suggestion
    if (SConfig::GetInstance().m_RAMOverrideEnable)
```

Source/Core/Core/HW/Memmap.cpp
```diff
@@ -189,6 +204,39 @@ static u32 GetFlags()
...
+      // Configurable MEM1 Size
+      return SConfig::GetInstance().m_MEM1Size;
+    else
+      // Retail MEM1 Size
+      return (u32)0x01800000;
+  };
+  auto GetMEM2Size = []()->u32 {
```
lioncash, line 215: Same review comments for `GetMEM1Size` can be applied here as well.

Source/Core/DolphinQt/Settings/AdvancedPane.cpp
```diff
@@ -208,6 +269,36 @@ void AdvancedPane::Update()
...
+    int MEM1_Size = SConfig::GetInstance().m_MEM1Size / 0x100000;
+    return tr("%1MB (MEM1)").arg(QString::number(MEM1_Size));
+  }());
+
+  m_mem2_override_slider->setEnabled(enable_ram_override_widgets && !running);
+  m_mem2_override_slider_label->setEnabled(enable_ram_override_widgets && !running);
+
+  {
+    const QSignalBlocker blocker(m_mem2_override_slider);
+    int MEM2_Size = SConfig::GetInstance().m_MEM2Size / 0x100000;
```
lioncash, line 293:
```suggestion
    const u32 mem2_size = SConfig::GetInstance().m_MEM2Size / 0x100000;
```

Source/Core/DolphinQt/Settings/AdvancedPane.cpp
```diff
@@ -208,6 +269,36 @@ void AdvancedPane::Update()
...
+    m_mem2_override_slider->setValue(MEM2_Size);
+  }
+
+  m_mem2_override_slider_label->setText([] {
+    int MEM2_Size = SConfig::GetInstance().m_MEM2Size / 0x100000;
```
lioncash, line 298:
```suggestion
    const u32 mem2_size = SConfig::GetInstance().m_MEM2Size / 0x100000;
```

Source/Core/DolphinQt/Settings/AdvancedPane.cpp
```diff
@@ -208,6 +269,36 @@ void AdvancedPane::Update()
...
+  m_mem1_override_slider_label->setEnabled(enable_ram_override_widgets && !running);
+
+  {
+    const QSignalBlocker blocker(m_mem1_override_slider);
+    int MEM1_Size = SConfig::GetInstance().m_MEM1Size / 0x100000;
+    m_mem1_override_slider->setValue(MEM1_Size);
+  }
+
+  m_mem1_override_slider_label->setText([] {
+    int MEM1_Size = SConfig::GetInstance().m_MEM1Size / 0x100000;
```
lioncash, line 284:
```suggestion
    const u32 mem1_size = SConfig::GetInstance().m_MEM1Size / 0x100000;
```

Source/Core/DolphinQt/Settings/AdvancedPane.cpp
```diff
@@ -208,6 +269,36 @@ void AdvancedPane::Update()
...
   }());
 
+  m_ram_override_checkbox->setEnabled(!running);
+
+  m_mem1_override_slider->setEnabled(enable_ram_override_widgets && !running);
+  m_mem1_override_slider_label->setEnabled(enable_ram_override_widgets && !running);
+
+  {
+    const QSignalBlocker blocker(m_mem1_override_slider);
+    int MEM1_Size = SConfig::GetInstance().m_MEM1Size / 0x100000;
```
lioncash, line 279:
```suggestion
    const u32 mem1_size = SConfig::GetInstance().m_MEM1Size / 0x100000;
```

Source/Core/DolphinQt/Settings/AdvancedPane.cpp
```diff
@@ -154,6 +193,27 @@ void AdvancedPane::ConnectLayout()
...
+
+  connect(m_mem1_override_slider, &QSlider::valueChanged, [this](int slider_value) {
+    const u32 MEM1_Size = m_mem1_override_slider->value() * 0x100000;
+    SConfig::GetInstance().m_MEM1Size = MEM1_Size;
+    Config::SetBaseOrCurrent(Config::MAIN_MEM1_SIZE, MEM1_Size);
+    Update();
+  });
+
+  connect(m_mem2_override_slider, &QSlider::valueChanged, [this](int slider_value) {
+    const u32 MEM2_Size = m_mem2_override_slider->value() * 0x100000;
```
lioncash, line 211:
```suggestion
    const u32 mem2_size = m_mem2_override_slider->value() * 0x100000;
```

Source/Core/DolphinQt/Settings/AdvancedPane.cpp
```diff
@@ -154,6 +193,27 @@ void AdvancedPane::ConnectLayout()
...
 
+  m_ram_override_checkbox->setChecked(SConfig::GetInstance().m_RAMOverrideEnable);
+  connect(m_ram_override_checkbox, &QCheckBox::toggled, [this](bool enable_ram_override) {
+    SConfig::GetInstance().m_RAMOverrideEnable = enable_ram_override;
+    Config::SetBaseOrCurrent(Config::MAIN_RAM_OVERRIDE_ENABLE, enable_ram_override);
+    Update();
+  });
+
+  connect(m_mem1_override_slider, &QSlider::valueChanged, [this](int slider_value) {
+    const u32 MEM1_Size = m_mem1_override_slider->value() * 0x100000;
```
lioncash, line 204:
```suggestion
    const u32 mem1_size = m_mem1_override_slider->value() * 0x100000;
```

Source/Core/Core/HW/Memmap.cpp
Minty-Meeo: In my opinion, the else makes the flow of the function more clear at a glance, that is why I left it in.  If you still disagree, I can remove it.

Source/Core/Core/HW/Memmap.h
```diff
@@ -37,23 +37,21 @@ extern u8* m_pEXRAM;
...
-  L1_CACHE_MASK = L1_CACHE_SIZE - 1,
-  IO_SIZE = 0x00010000,
-  EXRAM_SIZE = 0x04000000,
-  EXRAM_MASK = EXRAM_SIZE - 1,
-};
+// g_ram_size is the amount allocated by the emulator, whereas g_ram_size_real is
+// what will be reported in lowmem, and thus used by emulated software.
+// Note: Writing to lowmem is done by IPL. If using retail IPL, it will
+// always be set to 24MB.
+extern u32 g_ram_size_real;
```
lioncash, line 44: I meant to bring this up in the previous review, but these should likely be hidden from external view and be static file-scope variables and instead have functions that return by value to query the values of these variables. This way other subsystems can query, but not modify these values unintentionally.

Source/Core/Core/HW/Memmap.cpp
```diff
@@ -154,15 +174,10 @@ struct LogicalMemoryView
...
+// TODO: The actual size of RAM is g_ram_size_real (24MB); the other 8MB shouldn't
 // be backed by actual memory.
 // TODO: Do we want to handle the mirrors of the GC RAM?
-static PhysicalMemoryRegion physical_regions[] = {
-    {&m_pRAM, 0x00000000, RAM_SIZE, PhysicalMemoryRegion::ALWAYS},
-    {&m_pL1Cache, 0xE0000000, L1_CACHE_SIZE, PhysicalMemoryRegion::ALWAYS},
-    {&m_pFakeVMEM, 0x7E000000, FAKEVMEM_SIZE, PhysicalMemoryRegion::FAKE_VMEM},
-    {&m_pEXRAM, 0x10000000, EXRAM_SIZE, PhysicalMemoryRegion::WII_ONLY},
-};
+static PhysicalMemoryRegion physical_regions[4];
```
lioncash, line 180:
```suggestion
static std::array<PhysicalMemoryRegion, 4> physical_regions;
```

(`<array>` may need to be included)

Minty-Meeo: The original implementation did not use std::array.  What benefits does this bring to the table?  Is this just a general improvement that we ought to do while editing Memmap.cpp?

AdmiralCurtiss: If you're already changing the line, might as well make it better if it's trivial to do so, y'know? Or do you want a general list of reasons why std::array is better than a C-style array?

Minty-Meeo: Moreso the latter.  I'm still a learning a lot about C++, and have never used std::array before.

lioncash: - [std::array](https://en.cppreference.com/w/cpp/container/array) cannot have implicit array to pointer decay, It needs to be explicitly done by using it's `.data()` member function where necessary.

- `std::array` has a `size()` member function, so you can query the size of an array without needing to use `sizeof(array[0]) / sizeof(array)` nonsense.

- It's a wrapper around a C array. This means that implementations can optionally provide bounds checking in debug builds (controllable via a define) to catch out of bounds accesses that may occur across refactoring (currently all three main implementations—MSVC, libstdc++, and libc++—provide this).

TL;DR it's a more strongly-typed and type-safe version of a built-in array.

Source/Core/Core/HW/Memmap.h
lioncash: This still applies, just to make sure it isn't forgotten about.

Minty-Meeo: I'm not a fan of the disconnect between the actual 11 variables and the 11 query functions.  I'd say just don't write bad code, but if this is the last thing keeping you...

AdmiralCurtiss: Most of these aren't even read outside of Memmap.cpp as far as I can tell so you don't need to bother exposing them at all.

lioncash: > Most of these aren't even read outside of Memmap.cpp as far as I can tell so you don't need to bother exposing them at all.

This.

Historically, global variables have been a massive pain in Dolphin and a source of frustration for refactoring. If they don't need to be exposed, then they shouldn't be. Especially if they're mutable global variables. It turns a "you shouldn't do that" kind of mistake vector into a "you cannot do that". 

"just don't write bad code", is (IMO) a terrible viewpoint to have especially when the language already provides the necessary facilities to allow one to completely eliminate such code from being written in the first place directly at the interface boundary.

Source/Core/Core/BootManager.cpp
```diff
@@ -95,6 +95,9 @@ struct ConfigCache
   float m_EmulationSpeed;
   float m_OCFactor;
   bool m_OCEnable;
+  bool m_RAMOverrideEnable;
```
leoetlino, line 98: This isn't necessary if you use the new config system (the Common/Config and Core/Config stuff). Ditto for all the other additions to BootManager

Source/Core/Core/ConfigManager.cpp
```diff
@@ -242,6 +242,9 @@ void SConfig::SaveCoreSettings(IniFile& ini)
   core->Set("EmulationSpeed", m_EmulationSpeed);
   core->Set("Overclock", m_OCFactor);
   core->Set("OverclockEnable", m_OCEnable);
+  core->Set("RAMOverrideEnable", m_RAMOverrideEnable);
```
leoetlino, line 245: ditto for ConfigManager

Source/Core/Core/HW/Memmap.cpp
```diff
@@ -57,6 +58,69 @@ u8* m_pL1Cache;
...
+// is what gets used by emulated software.  This is usually 64MB.
+static u32 s_exram_size_real;
+static u32 s_exram_size;
+static u32 s_exram_mask;
+
+u32 GetRamSizeReal()
+{
+  return s_ram_size_real;
+}
+u32 GetRamSize()
```
leoetlino, line 83: I'm surprised clang-format is happy with the missing line breaks between these functions... Oh well.

Source/Core/Core/HW/Memmap.cpp
```diff
@@ -189,6 +248,34 @@ static u32 GetFlags()
 
 void Init()
 {
+  const auto GetMEM1Size = []() -> u32 {
```
leoetlino, line 251: Our naming convention is to use snake_case for variables, even if they're lambda functions.

```suggestion
  const auto get_mem1_size = []() -> u32 {
```

(also I'm not sure explicitly specifying the return type is necessary?)

Source/Core/Core/HW/Memmap.cpp
```diff
@@ -189,6 +248,34 @@ static u32 GetFlags()
...
+    if (SConfig::GetInstance().m_RAMOverrideEnable)
+      return SConfig::GetInstance().m_MEM1Size;
+    return 0x01800000U;
+  };
+  const auto GetMEM2Size = []() -> u32 {
```
leoetlino, line 256:
```suggestion
  const auto get_mem2_size = []() -> u32 {
```

Source/Core/Core/HW/Memmap.cpp
Minty-Meeo: I'm surprised clang-format is happy turning what was beautifully described in 11 lines into a mess that takes 44.

Source/Core/Core/HW/Memmap.cpp
Minty-Meeo: If lambdas are to use snake_case, this should be stated in the Contributing.md document.

Source/Core/Core/BootManager.cpp
Minty-Meeo: I don't understand what you're saying isn't necessary here.  I don't even know where the new config system is.

phire:
Like savestatese, Fifologs should also have protection against playing when the memory size is too small.

Source/Core/Core/HW/HW.cpp
```diff
@@ -42,6 +43,7 @@ void Init()
   ProcessorInterface::Init();
   ExpansionInterface::Init();  // Needs to be initialized before Memory
   Memory::Init();
+  AddressSpace::Init();
```
phire, line 46: This initializer depends on `Memory::Init()`

Ideally we should have a comment whenever there are initialization order dependencies.

Source/Core/Core/HW/Memmap.cpp
lioncash: Completely unnecessary response. We have code-base wide formatting to be consistent across the board and not have individual files with bespoke formatting choices that differ based on what programmer decided to code the file at that given point in time. Please direct snark like this elsewhere in the future, because it's not useful or helpful in a review.

Source/Core/Core/BootManager.cpp
JosJuice: The new config system is in Common/Config and Core/Config, as leoetlino mentioned. If you want to know how to add a new setting to it, you can use e.g. 3c9eb37 as a reference. (You only need to look at the changes in Core/Config.)

Minty-Meeo: Ooooooooh that's a folder you were referencing.  I was confused.  Alright, I'll look into this.

Minty-Meeo: Quick question, does the new config system have a cache to restore the "trampled on values" like the old system has?  I'm taking advantage of that in the fix for the FIFO player.

JosJuice: Actually, it has something better: Layers. At boot, instead of overwriting the values in the base layer, the value changes are added to a different layer. The new config system takes both layers into account when you want to fetch the value of a setting (or not – it is possible to e.g. fetch the value of the base layer specifically). Then when emulation ends, the layer that was used for overriding settings for this emulation session is discarded, making those overridden values go back to normal.

JMC47:
When I adjust MEM1 size, Some Wii Games aren't booting.  MEM2 didn't seem to cause problems.

Tested on Fire Emblem: Radiant Dawn.

Minty-Meeo:
I don't know what causes that, but it seems to be in all Wii games.  Were the Wii Dev units able to use more than 24MB of MEM1?

JMC47:
I think super smash bros. Brawl will print out the memarena in the logs.  I don't know if that's just MEM1, but maybe it'd give us a clue if it does include mem2

JMC47:
Now things appear to be working.  Obviously I can't test MEM2, but with MEM1, the few cases I know where extra memory affects games, they behaved as expected.

Source/Core/Core/IOS/VersionInfo.cpp
```diff
@@ -9,334 +9,339 @@
...
-         0x935E0000,
-         0x935E0000,
-         0x93600000,
-         HOLLYWOOD_REVISION,
-         RAM_VENDOR_MIOS,
-         PLACEHOLDER,
-         PLACEHOLDER,
-         PLACEHOLDER,
-     }}};
+std::array<MemoryValues, 41> ios_memory_values;
```
leoetlino, line 17: Maybe consider updating SetupMemory (in IOS) instead? These values are supposed to be dumped directly from IOS kernels.

JMC47:
@lioncash Are all of your qualms with this taken care of?

Source/Core/Core/IOS/IOS.cpp
```diff
@@ -158,18 +155,77 @@ static bool SetupMemory(u64 ios_title_id, MemorySetupType setup_type)
...
   Memory::Write_U8(0xAD, ADDR_APPLOADER_FLAG);
   Memory::Write_U16(0xBEEF, ADDR_DEVKIT_BOOT_PROGRAM_VERSION);
   Memory::Write_U32(target_imv->sysmenu_sync, ADDR_SYSMENU_SYNC);
+
+  RamOverrideForIOSMemoryValues(setup_type);
+
   return true;
 }
 
+// We're gonna make our own IOS Memory Values, with blackjack and hookers!
```
AdmiralCurtiss, line 173: Please don't write comments like that.

Minty-Meeo: ok

Source/Core/Core/IOS/IOS.cpp
```diff
@@ -158,18 +155,77 @@ static bool SetupMemory(u64 ios_title_id, MemorySetupType setup_type)
...
+    u32 ios_reserved_size;
+
+    {
+      const u32 ipc_buffer_begin = Memory::Read_U32(ADDR_IPC_BUFFER_BEGIN);
+      const u32 ipc_buffer_end = Memory::Read_U32(ADDR_IPC_BUFFER_END);
+      const u32 ios_reserved_begin = Memory::Read_U32(ADDR_IOS_RESERVED_BEGIN);
+      const u32 ios_reserved_end = Memory::Read_U32(ADDR_IOS_RESERVED_END);
+
+      // Inferring them
+      ipc_buffer_size = ipc_buffer_end - ipc_buffer_begin;
```
leoetlino, line 189: Why not just:

```cpp
const u32 ipc_buffer_size = ipc_buffer_end - ipc_buffer_begin;
```

or even 

```cpp
const u32 ipc_buffer_size =
  Memory::Read_U32(ADDR_IPC_BUFFER_END) - Memory::Read_U32(ADDR_IPC_BUFFER_BEGIN);
```

I don't think it's worth introducing a new scope *and* making some variables non-const.

Source/Core/Core/IOS/IOS.cpp
```diff
@@ -158,18 +155,77 @@ static bool SetupMemory(u64 ios_title_id, MemorySetupType setup_type)
...
   Memory::Write_U8(0xAD, ADDR_APPLOADER_FLAG);
   Memory::Write_U16(0xBEEF, ADDR_DEVKIT_BOOT_PROGRAM_VERSION);
   Memory::Write_U32(target_imv->sysmenu_sync, ADDR_SYSMENU_SYNC);
+
+  RamOverrideForIOSMemoryValues(setup_type);
+
   return true;
 }
 
+// We're gonna make our own IOS Memory Values.
```
leoetlino, line 173: I don't think this comment is necessary or adds much information that is not already in the function name.

Source/Core/DolphinQt/Settings/AdvancedPane.cpp
```diff
@@ -96,6 +96,44 @@ void AdvancedPane::CreateLayout()
   cpu_clock_override_description->setWordWrap(true);
   clock_override_layout->addWidget(cpu_clock_override_description);
 
+  auto* ram_override = new QGroupBox(tr("Memory Override"));
+  auto* ram_override_layout = new QVBoxLayout();
+  ram_override->setLayout(ram_override_layout);
+  main_layout->addWidget(ram_override);
+
+  m_ram_override_checkbox = new QCheckBox(tr("Enable Emulated Memory Override"));
```
leoetlino, line 104: "Override Emulated Memory Size" would be clearer, I think.

Source/Core/Core/FifoPlayer/FifoDataFile.cpp
```diff
@@ -199,20 +216,56 @@ std::unique_ptr<FifoDataFile> FifoDataFile::Load(const std::string& filename, bo
...
     file.Close();
     return dataFile;
   }
 
+  // To make up for such a hacky thing, here is a catch-all failsafe in case if the above code
+  // stops working or is otherwise removed.  As it is, this should never end up running.
+  if (header.mem1_size != Memory::GetRamSizeReal() ||
+      header.mem2_size != Memory::GetExRamSizeReal())
+  {
+    ERROR_LOG(BOOT,
```
JosJuice, line 257: Due to how serious this is if it happens, I would suggest making this use `PanicAlert` instead of `ERROR_LOG`.

Source/Core/Core/IOS/IOS.cpp
```diff
@@ -158,18 +155,68 @@ static bool SetupMemory(u64 ios_title_id, MemorySetupType setup_type)
...
   Memory::Write_U8(0xAD, ADDR_APPLOADER_FLAG);
   Memory::Write_U16(0xBEEF, ADDR_DEVKIT_BOOT_PROGRAM_VERSION);
   Memory::Write_U32(target_imv->sysmenu_sync, ADDR_SYSMENU_SYNC);
+
+  RamOverrideForIOSMemoryValues(setup_type);
+
   return true;
 }
 
+void RamOverrideForIOSMemoryValues(MemorySetupType setup_type)
```
JosJuice, line 173: For consistency (`IOS` is capitalized), `RAM` should be capitalized.

leoetlino: Unrelated, but we should probably get rid of the capitalization inconsistencies at some point.

Source/Core/Core/FifoPlayer/FifoDataFile.cpp
Minty-Meeo: Should I similarly change the alert I added for if the min_loader_version is too great?

JosJuice: I suppose that makes sense, yes.

leoetlino: Maybe even CriticalAlertT (T to make it translatable), so that this is always visible even if the user has disabled panic alerts?

Minty-Meeo:
About the conflicting file.  The only "conflict" it sees is an unused #include I removed.  I'm not sure why it is just now finding issue with that, it never did before.

JosJuice:
It's because that line of code was changed in master 4 hours ago. You'll need to rebase on latest master and fix the conflict.

RinMaru:
Messing with these settings has caused some issues. Tak and the power of Juju will crash mid game with Mem1/2 options modified so i dont think it should be on by default

JosJuice:
@RinMaru It isn't on by default, and turning it on for random games isn't expected to work.

PR #8726, opened 2020-04-06
Respect Core::WantsDeterminism for DNS servers on Linux

xperia64:
Previously Core::WantsDeterminism was only checked for pulling in the DNS servers on WIN32.
Same check should apply to Linux too.


JosJuice:
@dolphin-emu-bot rebuild

PR #8728, opened 2020-04-07
Vulkan: Switch from vkCreateMacOSSurfaceMVK() to vkCreateMetalSurfaceEXT() and Update MoltenVK

stenzek:
Since we are creating the swap chain off the UI thread, we can't use anything which accesses the underlying NSView object. We create and set the Metal layer on the UI thread before the video backend is initialized. This extension is both compatible with MoltenVK and gfx-portability for accepting a layer at surface creation.

Also updates MoltenVK to the latest version since the older version probably doesn't support the new surface extension.

### DIFF
--- a/Externals/MoltenVK/version.txt
+++ b/Externals/MoltenVK/version.txt
@@ -1,3 +1,3 @@
-libMoltenVK.dylib from vulkansdk-macos-1.1.108.0, renamed to libvulkan.dylib
+MoltenVK/macOS/dynamic/libMoltenVK.dylib from vulkansdk-macos-1.2.131.2, renamed to libvulkan.dylib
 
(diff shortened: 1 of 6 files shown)


stenzek:
Quickly tested with both MoltenVK and libportability and it appears to work fine.

gilcel:
Works great and is faster too than previous versions with Vulkan! Well done!
Now we just need to fix the annoying crash with "GPU Texture Decoding" when enabled on  macOS ;-)

gilcel:
A new big update for MoltenVK:
Release for Vulkan 1.1 Support and Vulkan SDK 1.2.154
https://github.com/KhronosGroup/MoltenVK/releases/tag/v1.1.0

PR #8730, opened 2020-04-07
Core: Skip duplicate frames when using frame advance

JosJuice:
https://bugs.dolphin-emu.org/issues/12025

It used to be the case that frame advance skipped duplicate frames (i.e. it would take 30 frame advances to get through one second of emulated time in a 30 fps game), but this broke in PR #8347. Skipping duplicate frames making TASing less annoying.

### DIFF
--- a/Source/Core/Core/Core.cpp
+++ b/Source/Core/Core/Core.cpp
@@ -77,6 +77,7 @@
 #include "InputCommon/ControllerInterface/ControllerInterface.h"
 #include "InputCommon/GCAdapter.h"
 
+#include "VideoCommon/AsyncRequests.h"
 #include "VideoCommon/Fifo.h"
 #include "VideoCommon/OnScreenDisplay.h"
 #include "VideoCommon/RenderBase.h"
@@ -104,6 +105,7 @@ static std::thread s_cpu_thread;
 static bool s_request_refresh_info = false;
 static bool s_is_throttler_temp_disabled = false;
 static bool s_frame_step = false;
+static std::atomic<bool> s_stop_frame_step;
 
 #ifdef USE_MEMORYWATCHER
 static std::unique_ptr<MemoryWatcher> s_memory_watcher;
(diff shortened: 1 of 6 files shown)


phire:
I'll point out that this isn't going to work so well in any XFB mode other than Immediate Present. 

Not sure if that's an issue you care about.

JosJuice:
My testing was done without Immediately Present XFB, and I didn't run into any problems. What type of issue would it be more specifically?

phire:
This PR makes an assumption that if the game copies something to a XFB, then that XFB will be displayed on screen in the next frame.

While this might be true in the general case, there are a lot of edge cases.

1. A frame might finish rendering just before the start of the field. The interrupt handler takes a short while to fire so in a race condition the XFB might not be displayed until a full frame later.

2. The game might be rendering partial XFBs. If it finishes one XFB section before the field boundary, it might still be waiting on a second section before it calls the XFB finished and displays it.

3. The game might be locked to 30fps. Just because an XFB is ready this field, doesn't mean the game will display it this field. For pacing issues.

4. The game might have more than 2 XFBs allocated. It might race ahead and finish 3 XFBs now and not display them until later.

phire:
Oh..  thought of some more edge cases.

Homebrew often draws directly to XFB. These updates will be skipped over.

Retail games occasionally do this too. Some games render loading screens or cutscenes directly to XFB.

phire:
So, we already have code that knows if an XFB is duplicate or not.

https://github.com/dolphin-emu/dolphin/blob/b0e040431a633ea3b0486673592e0de753110823/Source/Core/VideoCommon/RenderBase.cpp#L1203

You should pipe that though to the `FrameUpdate` callback.

JosJuice:
I've pushed a new version that should be better, though I'm not sure if it solves every problem you mentioned.

Source/Core/Core/Core.cpp
```diff
@@ -858,23 +860,35 @@ void VideoThrottle()
 
 // --- Callbacks for backends / engine ---
 
-// Should be called from GPU thread when a frame is drawn
+// Called from Renderer::Swap (GPU thread) when a frame is presented to the user's screen
 void Callback_VideoCopiedToXFB(bool video_update)
```
phire, line 864: This function is now seriously misnamed. 

Can we get it updated to something like `Callback_FrameDrawnToHostScreen` or `Callback_FramePresented`

And expand the explanatory comment to say it filters duplicates.

phire: Also, should we be deleting the video_update argument.

We are always calling this function with video_update = true.

JosJuice: Done (in a separate commit, because the cause for this needing to be renamed is not this PR).

For the reference, further in the past we used to call this function with false for duplicate frames, but there is no reason to do that anymore since there now is no code in this function that runs when `video_update` is false.

PR #8731, opened 2020-04-08
Fix crash when using DSP LLE with CPU interpreter (or fastmem off)

Pokechu22:
Fixes [#11980](https://bugs.dolphin-emu.org/issues/11980).  `Memory::physical_base` does not exist when fastmem is disabled.

Note that I remove some SSSE3 code (originally added 9 years ago in 559fb7434e404e81bf2c742f076a630cc6999d78).  My machine is too old to support SSSE3, so I haven't tested the performance impact, but I don't think it's likely to make a difference in this one case (since it's not used by DSP HLE).  If it actually does matter, then `CopyFromEmuSwapped` and `CopyToEmuSwapped` should be changed to use it instead.

### DIFF
--- a/Source/Core/Core/DSP/DSPHWInterface.cpp
+++ b/Source/Core/Core/DSP/DSPHWInterface.cpp
@@ -11,7 +11,6 @@
 
 #include "Common/CPUDetect.h"
 #include "Common/CommonTypes.h"
-#include "Common/Hash.h"
 #include "Common/Intrinsics.h"
 #include "Common/Logging/Log.h"
 #include "Common/MemoryUtil.h"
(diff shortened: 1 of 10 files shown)


Source/Core/Core/DSP/DSPHWInterface.cpp
```diff
@@ -283,22 +282,15 @@ u16 gdsp_ifx_read(u16 addr)
...
-  u16* dst = g_dsp.iram + (dsp_addr / 2);
-
-  const u8* code = &g_dsp.cpu_ram[addr & 0x0fffffff];
-  g_dsp.iram_crc = Common::HashEctor(code, size);
-
   Common::UnWriteProtectMemory(g_dsp.iram, DSP_IRAM_BYTE_SIZE, false);
-  memcpy(dst, code, size);
-  for (size_t i = 0; i < size / 2; i++)
-    dst[i] = Common::swap16(dst[i]);
+  Host::DMAToDSP(g_dsp.iram + dsp_addr / 2, addr, size);
```
leoetlino, line 286: Is not the mask not needed anymore?

Pokechu22: Assuming you're referring to `g_dsp.cpu_ram[addr & 0x0fffffff]`, I don't think it is; `CopyFromEmuSwapped` uses [`Memory::GetPointer`](https://github.com/dolphin-emu/dolphin/blob/9ffa72ad1fb8574ecda9e120d8a7fb930d053b01/Source/Core/Core/HW/Memmap.cpp#L536-L553) and it does its own masking (though with a different mask).

Perhaps [`Memory::GetPointerForRange`](https://github.com/dolphin-emu/dolphin/blob/9ffa72ad1fb8574ecda9e120d8a7fb930d053b01/Source/Core/Core/HW/Memmap.cpp#L463-L475) would be better, but it's not public.

PR #8733, opened 2020-04-09
Adjust s_DIMAR/s_DILENGTH behavior (fixes Baten Kaitos music)

JosJuice:
https://bugs.dolphin-emu.org/issues/11997

The problem seemed to be that `s_DILENGTH` would get set to 0 at times when it shouldn't. Simply not changing it in case of `NoReply` or `DTK` seems to fix the problem. However, we can actually go one step further in accuracy and use `data.size()` to change `s_DIMAR` and `s_DILENGTH` as partial reads (`NoReply` commands) complete, instead of jumping directly to 0 when the whole read completes.

### DIFF
--- a/Source/Core/Core/HW/DVD/DVDInterface.cpp
+++ b/Source/Core/Core/HW/DVD/DVDInterface.cpp
@@ -1233,8 +1233,17 @@ void SetHighError(u32 high_error)
 void FinishExecutingCommand(ReplyType reply_type, DIInterruptType interrupt_type, s64 cycles_late,
                             const std::vector<u8>& data)
 {
-  s_DIMAR += s_DILENGTH;
-  s_DILENGTH = 0;
+  // The data parameter contains the requested data iff this was called from DVDThread, and is
+  // empty otherwise. DVDThread is the only source of ReplyType::NoReply and ReplyType::DTK.
+
+  u32 transfer_size = 0;
+  if (reply_type == ReplyType::NoReply)
+    transfer_size = static_cast<u32>(data.size());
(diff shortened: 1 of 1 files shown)


AdmiralCurtiss:
Can confirm that this fixes the music issue in Baten Kaitos.

Source/Core/Core/HW/DVD/DVDInterface.cpp
```diff
@@ -1233,8 +1233,17 @@ void SetHighError(u32 high_error)
...
                             const std::vector<u8>& data)
 {
-  s_DIMAR += s_DILENGTH;
-  s_DILENGTH = 0;
+  // The data parameter contains the requested data iff this was called from DVDThread, and is
+  // empty otherwise. DVDThread is the only source of ReplyType::NoReply and ReplyType::DTK.
+
+  u32 transfer_size = 0;
+  if (reply_type == ReplyType::NoReply)
+    transfer_size = static_cast<u32>(data.size());
```
BhaaLseN, line 1241: Can this value ever exceed the current value of `s_DILENGTH` (making it potentially negative after the subtraction down there)?

JosJuice: Only if the emulated software overwrites `s_DILENGTH` in the middle of a read, which doesn't make much sense. I'm not sure what happens on hardware if you do that.

BhaaLseN: Good enough for me then.

PR #8735, opened 2020-04-10
Android: Reload Wii Remote settings upon saving them

Ebola16:
Fixes https://bugs.dolphin-emu.org/issues/12043.

@JosJuice thanks for the help.

### DIFF
--- a/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/model/Settings.java
+++ b/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/model/Settings.java
@@ -181,6 +181,7 @@ public class Settings
 
       // Notify the native code of the changes
       NativeLibrary.ReloadConfig();
+      NativeLibrary.ReloadWiimoteConfig();
     }
     else
     {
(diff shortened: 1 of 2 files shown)


JosJuice:
It seems a bit fishy to me that we call `UICommon::Init` twice, which is a function that does quite a lot of things, some of which maybe shouldn't be done twice... It would be better to just reload the controller settings, if that is possible (I haven't looked into it yet).

Ebola16:
@JosJuice I think I found a better way.

JosJuice:
Yes, that looks better.

Just one more thing to consider: Is it better to do this when starting emulation or right after saving the settings? I did the latter for the "normal" settings, and I guess it would make sense to use the same approach for both so there's no inconsistency, though I'm not sure if there are any particular reasons why my approach would be better.

Ebola16:
@JosJuice changed to your suggestion. Let's run less code at emulation start.

GjGamesBr:
Hey Ebola16
Can You add a Multimapping function on dolphin emulator for android plzzz, so the people can add one button for 2 or more things
this will help on guitar hero & rock band series because mapping 5 buttons on strum and mapping these 5 buttons on the color buttons, will make the tap notes and will become no necessary strum

Ebola16:
@GjGamesBr If your feature request doesn't already exist on our issue tracker (https://bugs.dolphin-emu.org/projects/emulator/issues), can you file it? I haven't worked with that part of the code yet and I don't own those games so another dev might be able to better help you there.

PR #8736, opened 2020-04-10
Fix bug 11920

Pokechu22:
https://bugs.dolphin-emu.org/issues/11920 - DSP LLE Interpreter does not have a DSP thread, so trying to wait would hang.  (DSP LLE recompiler also has no thread when in determinism mode).


lioncash:
Commit message should probably indicate the PR rationale/description in the commit itself so anyone running a blame doesn't need to look up everything to get a semblance of context for the change. As it is, it kind of reads as "Fix thing", which by itself is quite vague.

Looks good other than that though.

PR #8737, opened 2020-04-11
IOS KD: Added WC24 Download IOCTL to prevent errors

techydude0713:
All this PR does is write a non-zero return code from IOCTL_NWC24_DOWNLOAD_NOW_EX to indicate to the channel that downloading from WC24 failed.

This is to allow the Nintendo Channel to exit gracefully without the corrupted file message. Other channels might be able to benefit, but I haven't found any others that used the ioctl yet.

See https://bugs.dolphin-emu.org/issues/11978 for more details.

### DIFF
--- a/Source/Core/Core/IOS/Network/KD/NetKDRequest.cpp
+++ b/Source/Core/Core/IOS/Network/KD/NetKDRequest.cpp
@@ -143,6 +143,13 @@ IPCCommandResult NetKDRequest::IOCtl(const IOCtlRequest& request)
     // if ya set the IOS version to a very high value this happens ...
     INFO_LOG(IOS_WC24, "NET_KD_REQ: IOCTL_NWC24_REQUEST_SHUTDOWN - NI");
     break;
+          
+   case IOCTL_NWC24_DOWNLOAD_NOW_EX:
+    // Writing a non-zero return value to this ioctl prevents Nintendo Channel crashing.
+    // For more information, see https://bugs.dolphin-emu.org/issues/11978.
+    INFO_LOG(IOS_WC24, "NET_KD_REQ: IOCTL_NWC24_DOWNLOAD_NOW_EX - NI");
+    WriteReturnValue(NWC24::WC24_ERR_FATAL, request.buffer_out);
+    break;
 
   default:
     request.Log(GetDeviceName(), Common::Log::IOS_WC24);
--- a/Source/Core/Core/IOS/Network/KD/NetKDRequest.cpp
+++ b/Source/Core/Core/IOS/Network/KD/NetKDRequest.cpp
@@ -144,12 +144,60 @@ IPCCommandResult NetKDRequest::IOCtl(const IOCtlRequest& request)
     INFO_LOG(IOS_WC24, "NET_KD_REQ: IOCTL_NWC24_REQUEST_SHUTDOWN - NI");
     break;
           
-   case IOCTL_NWC24_DOWNLOAD_NOW_EX:
-    // Writing a non-zero return value to this ioctl prevents Nintendo Channel crashing.
-    // For more information, see https://bugs.dolphin-emu.org/issues/11978.
-    INFO_LOG(IOS_WC24, "NET_KD_REQ: IOCTL_NWC24_DOWNLOAD_NOW_EX - NI");
-    WriteReturnValue(NWC24::WC24_ERR_FATAL, request.buffer_out);
-    break;
+ case IOCTL_NWC24_DOWNLOAD_NOW_EX:
+    {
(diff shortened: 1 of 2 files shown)


JMC47:
I know Leoetlino has been busy, but they're the best person to review this imo.

Source/Core/Core/IOS/Network/KD/NetKDRequest.cpp
```diff
@@ -143,6 +143,13 @@ IPCCommandResult NetKDRequest::IOCtl(const IOCtlRequest& request)
     // if ya set the IOS version to a very high value this happens ...
     INFO_LOG(IOS_WC24, "NET_KD_REQ: IOCTL_NWC24_REQUEST_SHUTDOWN - NI");
     break;
+          
+   case IOCTL_NWC24_DOWNLOAD_NOW_EX:
+    // Writing a non-zero return value to this ioctl prevents Nintendo Channel crashing.
+    // For more information, see https://bugs.dolphin-emu.org/issues/11978.
+    INFO_LOG(IOS_WC24, "NET_KD_REQ: IOCTL_NWC24_DOWNLOAD_NOW_EX - NI");
+    WriteReturnValue(NWC24::WC24_ERR_FATAL, request.buffer_out);
```
leoetlino, line 151: Is this the correct return value? Do you have a hwtest or a disassembly of the relevant code in KD?

techydude0713: No. I just forced the return code to an error.
I don't have a hwtest, and I'm not sure how to get one.

leoetlino: A hwtest is a small homebrew you write and you run on a real console to test its behaviour. I suspect it would be easier to figure out what the proper error code is or how to signal that nothing was downloaded by disassembling the KD module. Either would be fine; I don't feel comfortable with merging this otherwise.

techydude0713:
From an outside view, I believe I understand what is supposed to happen.  

**Results:**
When IOCtl Download_Now is requested, what it should do is open wc24dl.bin and search for the offset of the link and jumps to it. It can be found by 0x800 +(0x200 * InBuff Offset (6-7)) + 0xb4. The IOCtl must find the link and initiate a HTTP transmission to acquire the file from the link. Once downloaded, it is decompressed by the IOCtl, and the contents are written into wc24dl.vff. The return value of this method is zero. The title will trigger the IOCtl every time a file needs to be downloaded individually. 

For the Nintendo Channel, there is an encryption key ("wc24pubk.mod") that is required to decrypt the decompression. No other title I know of does this.

**Proof:**
1. Dolphin does not download anything, while the Wii downloads the required file (according to Wireshark).
2. In the IOctl, I placed a breakpoint right before the return value of 0 is given, and I replaced Dolphin's VFF with a valid VFF that was created from the Wii. The channel was functioning after resuming the execution.

larsenv:
@techydude0713 I'm late to this. I am the main developer for RiiConnect24, so I really appreciate this. The Nintendo Channel doesn't work on Dolphin but we have it working on an actual Wii.

There are many other apps and games that use wc24pubk.mod:

* Animal Crossing: City Folk
* Check Mii Out Channel
* Minna no Theater
* My Pokémon Ranch
* Pokémon Mystery Dungeon
* Super Smash Bros. Brawl
* Tokyo Friend Park Wii
* Wii Shop Channel
* Wii Speak Channel
* and more...

I assume the mod file extension stands for "modulus". The wc24pubk.mod contains an AES-128-OFB key for decryption and an RSA-2048 key to sign it. The AES IV is in the simple WC24 container format. See https://github.com/RiiConnect24/wii-tools on how to encrypt it.

Hilariously, for games that use wc24pubk.mod, if the RSA signature isn't valid, the content will still download to wc24dl.vff. Games will only read for the VFF and process it (like DLC for Animal Crossing: City Folk), so in some cases you don't have to worry about it being properly signed at all.

Finally, @sepalani said he's interested in doing some work on WC24 stuff for Dolphin, that would both benefit this project and RiiConnect24!

zurgeg:
@techydude0713 iirc in order to run an hwtest there's a repo somewhere in https://github.com/dolphin-emu, I think it's dolphin-emu/hwtests, but anyway, you should be able to run that (I think on a real Wii). 
Anyway, I think that we should make a service that runs in the background for standby connection, the only problem with this is that would force users to install the system menu and some IOSes, I think if we gave users the option to run it in the background when they were installed through an emulated system update maybe? I don't know a lot about the way standby connection works and the way the Dolphin Team would want to implement it.

techydude0713:
Apologies for the two year wait. I finally got a prototype of this IOCtl semi-working (needs review and some improvements), and will upload it to a new PR with a fresh branch.

PR #8738, opened 2020-04-11
Replace the compress/uncompress actions with a convert dialog

JosJuice:
The reason why I decided to make these changes is because I need them for adding WIA writing support in PR #8538, but I'm opening this as a separate PR since it does have some benefits on its own:

* You can now convert from not only ISO and GCZ but also from TGC and extracted disc images (but you can still only convert **to** ISO and GCZ)
* You can configure the block size when converting to GCZ (tradeoff between size and performance)
* You can choose whether to scrub

I understand if this dialog with its additional options might be considered excessive for as long as we only support converting to ISO and GCZ. If we want to wait with merging this until the WIA PR is close to done, I'm fine with that.

**Don't merge this PR before ~#8591~, #8644, and ~#8740~**, because those PRs improve the results when converting from extracted discs, TGC, and all formats other than ISO, respectively.

image


Source/Core/DolphinQt/ConvertDialog.h
```diff
@@ -0,0 +1,43 @@
+// Copyright 2020 Dolphin Emulator Project
+// Licensed under GPLv2+
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <memory>
+
+#include <QCheckBox>
```
lioncash, line 9: This and `<QComboBox>` can be turned into forward declarations.

JosJuice:
I've made this PR try to filter out GCZ block sizes that trigger the bugs fixed by #8740. The reason is because someone might create a GCZ file on one version of Dolphin and then convert it back to ISO using an older one (even though that's uncommon...)

PR #8739, opened 2020-04-12
Android: Separate GameCube overlay refresh logic

Ebola16:
Wii overlay changes will be made in a subsequent PR.

Other minor changes:
- Renamed emulation menu's "Configure Controls" to "Overlay Controls", since the former could be misinterpreted as applying to all players.
- Added a "None" controller. It is primarily used for disabled controllers and real controllers (as an overlay is useless if Player 1 is using a real controller).

### DIFF
--- a/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/overlay/InputOverlay.java
+++ b/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/overlay/InputOverlay.java
@@ -24,12 +24,15 @@ import android.view.MotionEvent;
 import android.view.SurfaceView;
 import android.view.View;
 import android.view.View.OnTouchListener;
+import android.widget.Toast;
 
 import org.dolphinemu.dolphinemu.NativeLibrary;
 import org.dolphinemu.dolphinemu.NativeLibrary.ButtonState;
 import org.dolphinemu.dolphinemu.NativeLibrary.ButtonType;
 import org.dolphinemu.dolphinemu.R;
 import org.dolphinemu.dolphinemu.activities.EmulationActivity;
+import org.dolphinemu.dolphinemu.features.settings.model.Settings;
+import org.dolphinemu.dolphinemu.features.settings.utils.SettingsFile;
 
 import java.util.ArrayList;
 import java.util.HashSet;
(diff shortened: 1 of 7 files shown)


Ebola16:
#8735 has been merged so this is ready for review.

Ebola16:
Any feedback on this? I'd like to get this merged as it's needed for future PRs.

JosJuice:
If I'm reading the code right, enabling/disabling a GameCube controller in the settings activity or simply just running a GameCube game will make it so that the overlay controller used for Wii games becomes a GameCube controller. It doesn't make sense to me that Dolphin would change the controller type like that automatically. Or did I misread the code?

Ebola16:
I think I found a way to accomplish the same thing without touching SharedPreferences.

Ebola16:
Fixed a typo that caused problems with the Nunchuk overlay.

Ebola16:
Did I take care of @JosJuice's concerns?

Ebola16:
Is there anything holding this PR back?

Ebola16:
Can this PR be merged? I don't think there's any drawbacks to it anymore.

Ebola16:
Can this PR be merged?

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/overlay/InputOverlay.java
```diff
@@ -692,24 +701,62 @@ public void refreshControls()
...
-      {
-        addGameCubeOverlayControls(orientation);
-      }
-      else if (mPreferences.getInt("wiiController", 3) == 4)
+      if (EmulationActivity.isGameCubeGame())
       {
-        addClassicOverlayControls(orientation);
+        switch (NativeLibrary
+                .GetConfig(SettingsFile.FILE_NAME_DOLPHIN + ".ini", Settings.SECTION_INI_CORE,
+                        SettingsFile.KEY_GCPAD_PLAYER_1, DISABLED_GAMECUBE_CONTROLLER))
```
JosJuice, line 708: To match the C++ code, the default should be `EMULATED_GAMECUBE_CONTROLLER`. (For the reference, this only applies to the first player. The other players do indeed default to `DISABLED_GAMECUBE_CONTROLLER`.)

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/overlay/InputOverlay.java
```diff
@@ -692,24 +701,62 @@ public void refreshControls()
...
+
+          case EMULATED_GAMECUBE_CONTROLLER:
+            addGameCubeOverlayControls(orientation);
+            break;
+
+          case GAMECUBE_ADAPTER:
+            break;
+
+          default:
+            throw new IllegalArgumentException("Invalid SIDevice0 value");
```
JosJuice, line 726: This should not be an exception. There are other values than the three handled ones that are valid, even if the overlay doesn't support them.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/overlay/InputOverlay.java
```diff
@@ -692,24 +701,62 @@ public void refreshControls()
...
+
+          case OVERLAY_WIIMOTE_CLASSIC:
+            addClassicOverlayControls(orientation);
+            break;
+
+          case OVERLAY_NONE:
+            break;
+
+          default:
+            throw new IllegalArgumentException("Invalid wiiController value");
```
JosJuice, line 755: Same here.

Ebola16:
Wii Overlay PR is almost complete and this is required (likely along with #8803 to give users the option to disable GC Player 1 emulated controller when enabling the emulated wii remote). Please merge these PRs.

PR #8741, opened 2020-04-12
DolphinQt: Added checkbox to filter out NetPlay sessions that are already in-game.

cristian64:
A checkbox has been added to the **NetPlay Session Browser** dialog, so that in-game sessions can be filtered out from the list.

HideInGameSessions


Techjar:
Honestly don't know why we didn't think of adding this before.

PR #8742, opened 2020-04-13
IOS/USBHost: Skip starting threads when determinism is enabled

Techjar:
The threads can't actually be started when determinism is enabled, as the behavior would not be deterministic, but Open() still tries to start the threads and wait, resulting in a deadlock when booting certain games and homebrew in NetPlay.


JMC47:
Fixes Netplay in Certain Wii Titles and Homebrew.

PR #7753, opened 2019-01-28
Move a significant amount of video backend logic to VideoCommon

stenzek:
**Huge scary pull request ahead!**

This pull request moves a decent amount of logic from the video backends to VideoCommon. It's something I've been putting off for a while now, but finally found the motivation/time to do.

In addition to the maintainability improvements, as all backends are now considered equal, it adds things like:
- Post-processing to D3D
- GPU texture decoding to D3D
- Vulkan-style EFB cache to D3D and OpenGL
- Vulkan-style EFB poke batching to D3D and OpenGL

Some things makes easier in the future:
- Better state tracking in OpenGL -> performance improvements (since all state is now controlled via pipelines, we know what it is, and it hasn't been changed by random functions)
- Post-processing of XFB copies, rather than just the scanout (something I started a long time ago but was put off because of how much duplication there was in backends)
- Multi-threading of the backends
- Who doesn't like fewer lines of code for the same functionality
- Re-implementing D3D12? :) (if we can keep it to 5k lines of code or so, in line with the other backends)

Unfortunately, due to it being backend-specific, it does remove the **stencil buffer bounding box emulation**. I figured now that we have Vulkan/Metal on macOS, this removes one of the main uses for it.

I'm sorry for the size of this PR, but with most of the changes touching the same code, if I separated it, I'd be writing a lot of code just to throw it out again. I started doing that, but it got old quick :(

A rough work list:
- [x] Use abstract textures for EFB emulation
- [x] Move FramebufferManager to common and create implementation using abstract classes
- [x] Move TextureCache functionality to common (EFB copies, palette texture conversion)
- [x] Remove a lot of duplicated logic in Vulkan (largest backend by far...)
- [x] Re-implement GPU texture decoding
- [x] Implement texel buffers on OpenGL and Vulkan (for palette texture conversion)

PR #8743, opened 2020-04-13
VideoBackends: Disable GPU Texture Decoding under MoltenVK

Techjar:
It's broken and causes spectacular artifacts and crashes. Here's an example of what those look like: 

I don't know if the problem has been fixed upstream in MoltenVK. If someone else wants to look into that, by all means, I lack the in-depth knowledge of the video backends to do that myself. In the meantime this seems like a reasonable solution to most of the reports of crashes when using Vulkan on MacOS, due to users not being aware the feature is broken.

### DIFF
--- a/Source/Core/VideoCommon/DriverDetails.cpp
+++ b/Source/Core/VideoCommon/DriverDetails.cpp
@@ -117,7 +117,10 @@ constexpr BugInfo m_known_bugs[] = {
     {API_VULKAN, OS_ALL, VENDOR_ARM, DRIVER_ARM, Family::UNKNOWN, BUG_SLOW_CACHED_READBACK_MEMORY,
      -1.0, -1.0, true},
     {API_VULKAN, OS_ALL, VENDOR_QUALCOMM, DRIVER_QUALCOMM, Family::UNKNOWN,
-     BUG_SLOW_CACHED_READBACK_MEMORY, -1.0, -1.0, true}};
+     BUG_SLOW_CACHED_READBACK_MEMORY, -1.0, -1.0, true},
+    {API_VULKAN, OS_OSX, VENDOR_ALL, DRIVER_PORTABILITY, Family::UNKNOWN,
+     BUG_BROKEN_GPU_TEXTURE_DECODING, -1.0, -1.0, true},
+};
 
 static std::map<Bug, BugInfo> m_bugs;
 
--- a/Source/Core/VideoBackends/Vulkan/VulkanContext.cpp
+++ b/Source/Core/VideoBackends/Vulkan/VulkanContext.cpp
@@ -346,6 +346,10 @@ void VulkanContext::PopulateBackendInfoFeatures(VideoConfig* config, VkPhysicalD
   // with depth clamping. Fall back to inverted depth range for these.
   if (DriverDetails::HasBug(DriverDetails::BUG_BROKEN_REVERSED_DEPTH_RANGE))
     config->backend_info.bSupportsReversedDepthRange = false;
+
+  // GPU Texture Decoding is broken under MoltenVK.
+  if (DriverDetails::HasBug(DriverDetails::BUG_BROKEN_GPU_TEXTURE_DECODING))
+    config->backend_info.bSupportsGPUTextureDecoding = false;
 }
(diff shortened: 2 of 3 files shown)


MayImilae:
Here are some logs of the GPU Texture Decoding + MoltenVK crash, for anyone interested.

* Dolphin's logs (verbosity notice, everything checked) - https://hastebin.com/pakaqipexu.php

* macos crash report - https://hastebin.com/xajexepezu.php

pizuz:
The artifacts have been fixed in MoltenVK since 1.0.35. See also:

https://bugs.dolphin-emu.org/issues/11819

The crash is actually a regression from #7753.

Techjar:
Well okay, but that is such a large PR there's absolutely no way to narrow down a cause without some good old fashioned debugging, of which I lack the knowledge to do in this area. Until someone with the necessary knowledge has the motivation to actually debug it, at least quelling the support threads on the forum seems reasonable.

pizuz:
Do you have a Fifolog of that particular scene?

gilcel:
@Techjar: Which macOS version do you use and which Mac / GPU ? I tried with the latest Dolphin on a Mac mini 2012 with IGP Intel HD 4000 and a Macbook Air 2013 with Intel HD 5000 and it crashes immediately when enabling GPU Texture Decoding...I don't even get this screen with artifacts, as reported here: https://bugs.dolphin-emu.org/issues/11819

pizuz:
@gilcel: The glitches were probably produced prior to #7753. Meanwhile MoltenVK fixed the glitch in 1.0.35. you can still observe it in Wind Waker, because it manages to display the Nintendo and Dolby ProLogic logo before crashing.

Techjar:
Yeah I was informed that MoltenVK no longer causes those artifacts, it just crashes.

pizuz:
@Techjar: I was just mentioning this, because you listed the glitches as one of the reasons for opening this PR.

Source/Core/VideoCommon/DriverDetails.cpp
```diff
@@ -117,7 +117,9 @@ constexpr BugInfo m_known_bugs[] = {
     {API_VULKAN, OS_ALL, VENDOR_ARM, DRIVER_ARM, Family::UNKNOWN, BUG_SLOW_CACHED_READBACK_MEMORY,
      -1.0, -1.0, true},
     {API_VULKAN, OS_ALL, VENDOR_QUALCOMM, DRIVER_QUALCOMM, Family::UNKNOWN,
-     BUG_SLOW_CACHED_READBACK_MEMORY, -1.0, -1.0, true}};
+     BUG_SLOW_CACHED_READBACK_MEMORY, -1.0, -1.0, true},
+    {API_VULKAN, OS_OSX, VENDOR_ALL, DRIVER_PORTABILITY, Family::UNKNOWN,
+     BUG_BROKEN_GPU_TEXTURE_DECODING, -1.0, -1.0, true}};
```
leoetlino, line 122: Could you add a trailing comma? Helps make future diffs cleaner.

pizuz:
The current MoltenVK release 1.0.42 (for Vulkan SDK 1.2.141) fixes the crash. However there seems to be an issue anyway, because you have to start a game twice in order for it to actually display correct textures.

Screenshot 2020-06-03 at 22 18 59

MayImilae:
Still crashing for me in latest master, though it gets further than before. Either way, crashing or at best hella glitching is means we should just disable gpu texture decoding in moltenvk. At least in the UI.

Source/Core/VideoCommon/DriverDetails.cpp
Techjar: Like that?

JMC47:
How does this Pull Request affect the GUI?  Is the checkbox just grayed out?

JosJuice:
Yes, the checkbox is grayed out when using Vulkan on macOS.

StLouisCPhT:
Needs more work - does not gray out the checkbox on my 13" 2020 MBP.


Screen Shot 2020-09-13 at 10 52 26 AM
Screen Shot 2020-09-13 at 10 54 47 AM

PR #8744, opened 2020-04-13
DolphinQt: Saving and restoring NetPlay Session Browser dialog's settings.

cristian64:
The **NetPlay Session Browser** dialog's settings are now saved and restored when the widget is created/destroyed. This includes the widget's geometry, as well as the filter settings.

### DIFF
--- a/Source/Core/DolphinQt/NetPlay/NetPlayBrowser.cpp
+++ b/Source/Core/DolphinQt/NetPlay/NetPlayBrowser.cpp
@@ -27,6 +27,7 @@
 
 #include "DolphinQt/QtUtils/ModalMessageBox.h"
 #include "DolphinQt/QtUtils/RunOnObject.h"
+#include "DolphinQt/Settings.h"
 
 NetPlayBrowser::NetPlayBrowser(QWidget* parent) : QDialog(parent)
 {
@@ -44,6 +45,8 @@ NetPlayBrowser::NetPlayBrowser(QWidget* parent) : QDialog(parent)
   m_refresh_run.Set(true);
   m_refresh_thread = std::thread([this] { RefreshLoop(); });
 
+  RestoreSettings();
+
   UpdateList();
   Refresh();
 }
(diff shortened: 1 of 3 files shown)


JosJuice:
@dolphin-emu-bot rebuild

Source/Core/DolphinQt/NetPlay/NetPlayBrowser.cpp
```diff
@@ -323,3 +328,51 @@ void NetPlayBrowser::accept()
...
+{
+  const auto& settings = Settings::Instance().GetQSettings();
+
+  const QByteArray geometry =
+      settings.value(QStringLiteral("netplaybrowser/geometry")).toByteArray();
+  if (!geometry.isEmpty())
+    restoreGeometry(geometry);
+
+  const QString region = settings.value(QStringLiteral("netplaybrowser/region")).toString();
+  if (m_region_combo->findText(region) != -1)
```
leoetlino, line 362: Why not just compare the values?

Source/Core/DolphinQt/NetPlay/NetPlayBrowser.cpp
```diff
@@ -323,3 +328,51 @@ void NetPlayBrowser::accept()
...
+    m_region_combo->setCurrentText(region);
+  const QString name = settings.value(QStringLiteral("netplaybrowser/name")).toString();
```
leoetlino, line 364: Please add a new line before this line for readability (or eliminate the variable, I don't think it adds much here).

Source/Core/DolphinQt/NetPlay/NetPlayBrowser.cpp
cristian64: My intention was to only set the text if the text was a valid option of the combobox.

This is to avoid leaving the combobox in an invalid state, if the list of regions are ever renamed, but the user's settings still contain an old value.

I can possibly define a variable, with a relevant name, to make this more obvious:

```cpp
const bool valid_region = m_region_combo->findText(region) != -1;
if (valid_region)
  m_region_combo->setCurrentText(region);
```

Thoughts?

leoetlino: Sure, that sounds good to me.

PR #8745, opened 2020-04-14
Qt/Debugger: Refresh on savestate load.

AdmiralCurtiss:
Pretty self-explanatory. The debugger windows would show stale data after a savestate load when the emulator was in a paused state.

~~I'm not sure if this is a good place to put the `SavestateLoaded` signal, but the `EmulationStateChanged` is already there so... maybe?~~ See comments below.

### DIFF
--- a/Source/Core/DolphinQt/Host.cpp
+++ b/Source/Core/DolphinQt/Host.cpp
@@ -19,6 +19,7 @@
 #include "Core/Host.h"
 #include "Core/NetPlayProto.h"
 #include "Core/PowerPC/PowerPC.h"
+#include "Core/State.h"
 
 #include "DolphinQt/QtUtils/QueueOnObject.h"
 #include "DolphinQt/Settings.h"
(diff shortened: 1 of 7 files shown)


AdmiralCurtiss:
Looking at this a bit more, maybe we should just emit a `Host::UpdateDisasmDialog` instead when the savestate callback fires and then register that in all the debugger windows where it isn't yet? Unless we specifically need the savestate event for something else.

AdmiralCurtiss:
Yeah I think this is better. This also fixes various windows (most noticeably the Memory window) not updating when stepping code.

Source/Core/DolphinQt/Host.cpp
```diff
@@ -30,7 +31,15 @@
 #include "VideoCommon/RenderBase.h"
 #include "VideoCommon/VideoConfig.h"
 
-Host::Host() = default;
+Host::Host()
+{
+  State::SetOnAfterLoadCallback([this] { Host_UpdateDisasmDialog(); });
```
leoetlino, line 36: Looks like this was lost during the move to Qt, heh.

PR #8747, opened 2020-04-17
Support controlling Free Look via input bindings (motion controls, gamepad, etc!)

iwubcode:
I've always thought Free Look was such a cool and underutilized feature.  In an effort to make it more mainstream, I've removed the shackles of requiring to use the mouse.  Now you can assign motion controls (or gamepad) to the Free Look rotation movement.

_Rotation Free Look settings (mouse, gamepad, DSU or bluetooth emulated wiimote!)_
image

### DIFF
--- a/Source/Core/Core/FreeLookManager.h
+++ b/Source/Core/Core/FreeLookManager.h
@@ -4,6 +4,9 @@
 
 #pragma once
 
+#include <chrono>
+#include <optional>
+
 #include "Common/CommonTypes.h"
 #include "InputCommon/ControllerEmu/ControllerEmu.h"
 
@@ -13,6 +16,7 @@ namespace ControllerEmu
 {
 class ControlGroup;
 class Buttons;
+class IMUGyroscope;
 }  // namespace ControllerEmu
 
 enum class FreeLookGroup
@@ -20,7 +24,8 @@ enum class FreeLookGroup
   Move,
   Speed,
   FieldOfView,
-  Other
+  Other,
+  Rotation,
 };
 
 namespace FreeLook
@@ -52,6 +57,8 @@ private:
   ControllerEmu::Buttons* m_speed_buttons;
   ControllerEmu::Buttons* m_fov_buttons;
   ControllerEmu::Buttons* m_other_buttons;
+  ControllerEmu::IMUGyroscope* m_rotation_gyro;
 
   const unsigned int m_index;
+  std::optional<std::chrono::steady_clock::time_point> m_last_free_look_rotate_time;
 };
--- a/Source/Core/DolphinQt/HotkeyScheduler.cpp
+++ b/Source/Core/DolphinQt/HotkeyScheduler.cpp
@@ -141,6 +141,10 @@ void HotkeyScheduler::Run()
   {
     Common::SleepCurrentThread(1000 / 60);
 
+    g_controller_interface.SetCurrentInputChannel(ciface::InputChannel::FreeLook);
+    g_controller_interface.UpdateInput();
+    FreeLook::UpdateInput();
+
     g_controller_interface.SetCurrentInputChannel(ciface::InputChannel::Host);
     g_controller_interface.UpdateInput();
 
@@ -545,8 +549,6 @@ void HotkeyScheduler::Run()
       Config::SetCurrent(Config::FREE_LOOK_ENABLED, new_value);
       OSD::AddMessage(StringFromFormat("Free Look: %s", new_value ? "Enabled" : "Disabled"));
     }
-
-    FreeLook::UpdateInput();
 
     // Savestates
     for (u32 i = 0; i < State::NUM_STATES; i++)
--- a/Source/Core/DolphinQt/HotkeyScheduler.cpp
+++ b/Source/Core/DolphinQt/HotkeyScheduler.cpp
@@ -141,8 +141,8 @@ void HotkeyScheduler::Run()
   {
     Common::SleepCurrentThread(1000 / 60);
 
-    if (Core::GetState() == Core::State::Uninitialized || Core::GetState() == Core::State::Paused)
-      g_controller_interface.UpdateInput();
+    g_controller_interface.SetCurrentInputChannel(ciface::InputChannel::Host);
+    g_controller_interface.UpdateInput();
 
     if (!HotkeyManagerEmu::IsEnabled())
       continue;
--- a/Source/Core/DolphinQt/HotkeyScheduler.cpp
+++ b/Source/Core/DolphinQt/HotkeyScheduler.cpp
@@ -139,7 +139,7 @@ void HotkeyScheduler::Run()
 
   while (!m_stop_requested.IsSet())
   {
-    Common::SleepCurrentThread(1000 / 60);
+    Common::SleepCurrentThread(5);
 
     g_controller_interface.SetCurrentInputChannel(ciface::InputChannel::FreeLook);
     g_controller_interface.UpdateInput();
--- a/Source/Core/Common/Matrix.cpp
+++ b/Source/Core/Common/Matrix.cpp
@@ -4,6 +4,8 @@
 
 #include "Common/Matrix.h"
 
+#include "Common/MathUtil.h"
+
 #include <algorithm>
 #include <cmath>
 
@@ -119,6 +121,32 @@ Vec3 operator*(const Quaternion& lhs, const Vec3& rhs)
 {
   const auto result = lhs * Quaternion(0, rhs.x, rhs.y, rhs.z) * lhs.Conjugate();
   return Vec3(result.data.x, result.data.y, result.data.z);
+}
+
+Vec3 FromQuaternionToEuler(const Quaternion& q)
+{
+  Vec3 result;
+
+  const float qx = q.data.x;
+  const float qy = q.data.y;
+  const float qz = q.data.z;
+  const float qw = q.data.w;
+
+  const float sinr_cosp = 2 * (qw * qx + qy * qz);
+  const float cosr_cosp = 1 - 2 * (qx * qx + qy * qy);
+  result.x = std::atan2(sinr_cosp, cosr_cosp);
+
+  const float sinp = 2 * (qw * qy - qz * qx);
+  if (std::abs(sinp) >= 1)
+    result.y = std::copysign(MathUtil::PI / 2, sinp);  // use 90 degrees if out of range
+  else
+    result.y = std::asin(sinp);
+
+  const float siny_cosp = 2 * (qw * qz + qx * qy);
+  const float cosy_cosp = 1 - 2 * (qy * qy + qz * qz);
+  result.z = std::atan2(siny_cosp, cosy_cosp);
+
+  return result;
 }
 
 Matrix33 Matrix33::Identity()
--- a/Source/Core/Common/Matrix.h
+++ b/Source/Core/Common/Matrix.h
@@ -359,6 +359,8 @@ public:
 Quaternion operator*(Quaternion lhs, const Quaternion& rhs);
 Vec3 operator*(const Quaternion& lhs, const Vec3& rhs);
 
+Vec3 FromQuaternionToEuler(const Quaternion& q);
+
 class Matrix33
 {
 public:
--- a/Source/Core/Core/FreeLookManager.cpp
+++ b/Source/Core/Core/FreeLookManager.cpp
@@ -222,40 +222,40 @@ void FreeLookController::Update()
 
   g_freelook_camera.Rotate(gyro_motion_quat);
   if (m_move_buttons->controls[MoveButtons::Up]->GetState<bool>())
-    g_freelook_camera.MoveVertical(-g_freelook_camera.GetSpeed());
+    g_freelook_camera.MoveVertical(-g_freelook_camera.GetSpeed() * dt);
 
   if (m_move_buttons->controls[MoveButtons::Down]->GetState<bool>())
-    g_freelook_camera.MoveVertical(g_freelook_camera.GetSpeed());
+    g_freelook_camera.MoveVertical(g_freelook_camera.GetSpeed() * dt);
 
   if (m_move_buttons->controls[MoveButtons::Left]->GetState<bool>())
-    g_freelook_camera.MoveHorizontal(g_freelook_camera.GetSpeed());
+    g_freelook_camera.MoveHorizontal(g_freelook_camera.GetSpeed() * dt);
 
   if (m_move_buttons->controls[MoveButtons::Right]->GetState<bool>())
-    g_freelook_camera.MoveHorizontal(-g_freelook_camera.GetSpeed());
+    g_freelook_camera.MoveHorizontal(-g_freelook_camera.GetSpeed() * dt);
 
   if (m_move_buttons->controls[MoveButtons::Forward]->GetState<bool>())
-    g_freelook_camera.MoveForward(g_freelook_camera.GetSpeed());
+    g_freelook_camera.MoveForward(g_freelook_camera.GetSpeed() * dt);
 
   if (m_move_buttons->controls[MoveButtons::Backward]->GetState<bool>())
-    g_freelook_camera.MoveForward(-g_freelook_camera.GetSpeed());
+    g_freelook_camera.MoveForward(-g_freelook_camera.GetSpeed() * dt);
 
   if (m_fov_buttons->controls[FieldOfViewButtons::IncreaseX]->GetState<bool>())
-    g_freelook_camera.IncreaseFovX(g_freelook_camera.GetFovStepSize());
+    g_freelook_camera.IncreaseFovX(g_freelook_camera.GetFovStepSize() * dt);
 
   if (m_fov_buttons->controls[FieldOfViewButtons::DecreaseX]->GetState<bool>())
-    g_freelook_camera.IncreaseFovX(-1.0f * g_freelook_camera.GetFovStepSize());
+    g_freelook_camera.IncreaseFovX(-1.0f * g_freelook_camera.GetFovStepSize() * dt);
 
   if (m_fov_buttons->controls[FieldOfViewButtons::IncreaseY]->GetState<bool>())
-    g_freelook_camera.IncreaseFovY(g_freelook_camera.GetFovStepSize());
+    g_freelook_camera.IncreaseFovY(g_freelook_camera.GetFovStepSize() * dt);
 
   if (m_fov_buttons->controls[FieldOfViewButtons::DecreaseY]->GetState<bool>())
-    g_freelook_camera.IncreaseFovY(-1.0f * g_freelook_camera.GetFovStepSize());
+    g_freelook_camera.IncreaseFovY(-1.0f * g_freelook_camera.GetFovStepSize() * dt);
 
   if (m_speed_buttons->controls[SpeedButtons::Decrease]->GetState<bool>())
-    g_freelook_camera.ModifySpeed(1.0f / 1.1f);
+    g_freelook_camera.ModifySpeed(g_freelook_camera.GetSpeed() * -0.9 * dt);
 
   if (m_speed_buttons->controls[SpeedButtons::Increase]->GetState<bool>())
-    g_freelook_camera.ModifySpeed(1.1f);
+    g_freelook_camera.ModifySpeed(g_freelook_camera.GetSpeed() * 1.1 * dt);
 
   if (m_speed_buttons->controls[SpeedButtons::Reset]->GetState<bool>())
     g_freelook_camera.ResetSpeed();
--- a/Source/Core/Core/FreeLookManager.cpp
+++ b/Source/Core/Core/FreeLookManager.cpp
@@ -12,6 +12,7 @@
 #include "Core/FreeLookConfig.h"
 
 #include "InputCommon/ControllerEmu/ControlGroup/Buttons.h"
+#include "InputCommon/ControllerEmu/ControlGroup/IMUGyroscope.h"
 #include "InputCommon/InputConfig.h"
 
 #include "VideoCommon/FreeLookCamera.h"
@@ -60,6 +61,19 @@ enum FieldOfViewButtons
   DecreaseY,
 };
 }
+
+namespace GyroButtons
+{
+enum GyroButtons
+{
+  PitchUp,
+  PitchDown,
+  RollLeft,
+  RollRight,
+  YawLeft,
+  YawRight,
+};
+}
 }  // namespace
 
 FreeLookController::FreeLookController(const unsigned int index) : m_index(index)
@@ -89,6 +103,9 @@ FreeLookController::FreeLookController(const unsigned int index) : m_index(index
   m_fov_buttons->AddInput(ControllerEmu::Translate, _trans("Decrease X"));
   m_fov_buttons->AddInput(ControllerEmu::Translate, _trans("Increase Y"));
   m_fov_buttons->AddInput(ControllerEmu::Translate, _trans("Decrease Y"));
+
+  groups.emplace_back(m_rotation_gyro = new ControllerEmu::IMUGyroscope(
+                          _trans("Incremental Rotation"), _trans("Incremental Rotation")));
 }
 
 std::string FreeLookController::GetName() const
@@ -125,6 +142,35 @@ void FreeLookController::LoadDefaults(const ControllerInterface& ciface)
                                       hotkey_string({"Shift", "`Axis Z+`"}));
   m_fov_buttons->SetControlExpression(FieldOfViewButtons::DecreaseY,
                                       hotkey_string({"Shift", "`Axis Z-`"}));
+
+#if defined HAVE_X11 && HAVE_X11
+  m_rotation_gyro->SetControlExpression(GyroButtons::PitchUp,
+                                        "if(`Click 3`,`RelativeMouse Y-` * 0.10, 0)");
+  m_rotation_gyro->SetControlExpression(GyroButtons::PitchDown,
+                                        "if(`Click 3`,`RelativeMouse Y+` * 0.10, 0)");
+#else
+  m_rotation_gyro->SetControlExpression(GyroButtons::PitchUp,
+                                        "if(`Click 1`,`RelativeMouse Y-` * 0.10, 0)");
+  m_rotation_gyro->SetControlExpression(GyroButtons::PitchDown,
+                                        "if(`Click 1`,`RelativeMouse Y+` * 0.10, 0)");
+#endif
+
+  m_rotation_gyro->SetControlExpression(GyroButtons::RollLeft,
+                                        "if(`Click 2`,`RelativeMouse X-` * 0.10, 0)");
+  m_rotation_gyro->SetControlExpression(GyroButtons::RollRight,
+                                        "if(`Click 2`,`RelativeMouse X+` * 0.10, 0)");
+
+#if defined HAVE_X11 && HAVE_X11
+  m_rotation_gyro->SetControlExpression(GyroButtons::YawLeft,
+                                        "if(`Click 3`,`RelativeMouse X-` * 0.10, 0)");
+  m_rotation_gyro->SetControlExpression(GyroButtons::YawRight,
+                                        "if(`Click 3`,`RelativeMouse X+` * 0.10, 0)");
+#else
+  m_rotation_gyro->SetControlExpression(GyroButtons::YawLeft,
+                                        "if(`Click 1`,`RelativeMouse X-` * 0.10, 0)");
+  m_rotation_gyro->SetControlExpression(GyroButtons::YawRight,
+                                        "if(`Click 1`,`RelativeMouse X+` * 0.10, 0)");
+#endif
 }
 
 ControllerEmu::ControlGroup* FreeLookController::GetGroup(FreeLookGroup group) const
@@ -139,6 +185,8 @@ ControllerEmu::ControlGroup* FreeLookController::GetGroup(FreeLookGroup group) c
     return m_fov_buttons;
   case FreeLookGroup::Other:
     return m_other_buttons;
+  case FreeLookGroup::Rotation:
+    return m_rotation_gyro;
   default:
     return nullptr;
   }
@@ -151,6 +199,28 @@ void FreeLookController::Update()
 
   const auto lock = GetStateLock();
 
+  float dt = 1.0;
+  if (m_last_free_look_rotate_time)
+  {
+    using seconds = std::chrono::duration<float, std::ratio<1>>;
+    dt = std::chrono::duration_cast<seconds>(std::chrono::steady_clock::now() -
+                                             *m_last_free_look_rotate_time)
+             .count();
+  }
+  m_last_free_look_rotate_time = std::chrono::steady_clock::now();
+
+  const auto gyro_motion_rad_velocity =
+      m_rotation_gyro->GetState() ? *m_rotation_gyro->GetState() : Common::Vec3{};
+
+  // Due to gyroscope implementation we need to swap the yaw and roll values
+  // and because of the different axis used for Wii and the PS3 motion directions,
+  // we need to invert the yaw and roll as well
+  const Common::Vec3 gyro_motion_rad_velocity_converted{
+      gyro_motion_rad_velocity.x, gyro_motion_rad_velocity.z * -1, gyro_motion_rad_velocity.y * -1};
+  const auto gyro_motion_quat =
+      Common::Quaternion::RotateXYZ(gyro_motion_rad_velocity_converted * dt);
+
+  g_freelook_camera.Rotate(gyro_motion_quat);
   if (m_move_buttons->controls[MoveButtons::Up]->GetState<bool>())
     g_freelook_camera.MoveVertical(-g_freelook_camera.GetSpeed());
 
--- a/Source/Core/Core/HW/SI/SI.cpp
+++ b/Source/Core/Core/HW/SI/SI.cpp
@@ -655,6 +655,7 @@ void UpdateDevices()
 
   // Update inputs at the rate of SI
   // Typically 120hz but is variable
+  g_controller_interface.SetCurrentInputChannel(ciface::InputChannel::SerialInterface);
   g_controller_interface.UpdateInput();
 
   // Update channels and set the status bit if there's new data
--- a/Source/Core/Core/IOS/USB/Bluetooth/BTEmu.cpp
+++ b/Source/Core/Core/IOS/USB/Bluetooth/BTEmu.cpp
@@ -339,6 +339,7 @@ void BluetoothEmuDevice::Update()
 
   if (now - m_last_ticks > interval)
   {
+    g_controller_interface.SetCurrentInputChannel(ciface::InputChannel::Bluetooth);
     g_controller_interface.UpdateInput();
     for (auto& wiimote : m_wiimotes)
       wiimote->UpdateInput();
--- a/Source/Core/DolphinQt/CMakeLists.txt
+++ b/Source/Core/DolphinQt/CMakeLists.txt
@@ -122,6 +122,8 @@ add_executable(dolphin-emu
   Config/LogWidget.h
   Config/Mapping/FreeLookGeneral.cpp
   Config/Mapping/FreeLookGeneral.h
+  Config/Mapping/FreeLookRotation.cpp
+  Config/Mapping/FreeLookRotation.h
   Config/Mapping/GCKeyboardEmu.cpp
   Config/Mapping/GCKeyboardEmu.h
   Config/Mapping/GCMicrophone.cpp
--- a/Source/Core/DolphinQt/Config/Mapping/FreeLookRotation.cpp
+++ b/Source/Core/DolphinQt/Config/Mapping/FreeLookRotation.cpp
@@ -1,0 +1,62 @@
+// Copyright 2021 Dolphin Emulator Project
+// Licensed under GPLv2+
+// Refer to the license.txt file included.
+
+#include "DolphinQt/Config/Mapping/FreeLookRotation.h"
+
+#include <QGridLayout>
+#include <QGroupBox>
+#include <QHBoxLayout>
+#include <QLabel>
+#include <QPushButton>
+
+#include "Core/FreeLookManager.h"
+#include "DolphinQt/Config/ControllerInterface/ControllerInterfaceWindow.h"
+#include "InputCommon/InputConfig.h"
+
+FreeLookRotation::FreeLookRotation(MappingWindow* window) : MappingWidget(window)
+{
+  CreateMainLayout();
+}
+
+void FreeLookRotation::CreateMainLayout()
+{
+  m_main_layout = new QGridLayout;
+
+  auto* alternate_input_layout = new QHBoxLayout();
+  auto* note_label = new QLabel(
+      tr("Note: motion input may require configuring alternate input sources before use."));
+  note_label->setWordWrap(true);
+  auto* alternate_input_sources_button = new QPushButton(tr("Alternate Input Sources"));
+  alternate_input_layout->addWidget(note_label, 1);
+  alternate_input_layout->addWidget(alternate_input_sources_button, 0, Qt::AlignRight);
+  connect(alternate_input_sources_button, &QPushButton::clicked, this, [this] {
+    ControllerInterfaceWindow* window = new ControllerInterfaceWindow(this);
+    window->setAttribute(Qt::WA_DeleteOnClose, true);
+    window->setWindowModality(Qt::WindowModality::WindowModal);
+    window->show();
+  });
+  m_main_layout->addLayout(alternate_input_layout, 0, 0, 1, -1);
+
+  m_main_layout->addWidget(
+      CreateGroupBox(tr("Incremental Rotation (rad/sec)"),
+                     FreeLook::GetInputGroup(GetPort(), FreeLookGroup::Rotation)),
+      1, 0);
+
+  setLayout(m_main_layout);
+}
+
+InputConfig* FreeLookRotation::GetConfig()
+{
+  return FreeLook::GetInputConfig();
+}
+
+void FreeLookRotation::LoadSettings()
+{
+  FreeLook::LoadInputConfig();
+}
+
+void FreeLookRotation::SaveSettings()
+{
+  FreeLook::GetInputConfig()->SaveConfig();
+}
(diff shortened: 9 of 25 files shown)


iwubcode:
I thought of calling the groups "Modifiers" and "Rotate".  Then removing "Freelook" from the key names in the UI, ex: 

`"Freeloook Decrease Speed" -> "Decrease Speed"`

Would love to hear @MayImilae 's thoughts / suggestions on the UI.

Techjar:
Yeah may as well change the names while we're at it, I doubt many people change the freelook bindings so them getting reset shouldn't be a big deal.

However, the default bindings are not right. Right click should control pitch and yaw, while middle click should control roll.

Source/Core/DolphinQt/HotkeyScheduler.cpp
```diff
@@ -525,45 +530,51 @@ void HotkeyScheduler::Run()
...
+      if (IsHotkey(HK_FREELOOK_ZOOM_OUT, true))
+        VertexShaderManager::TranslateView(0.0, -fl_speed);
 
-    if (IsHotkey(HK_FREELOOK_RESET, true))
-      VertexShaderManager::ResetView();
+      if (IsHotkey(HK_FREELOOK_RESET, true))
+        VertexShaderManager::ResetView();
+
+      const auto gyro_amt = GetGyro(HGGP_FREELOOK_ROTATE);
+      VertexShaderManager::RotateView(gyro_amt.x, gyro_amt.y, gyro_amt.z);
```
Techjar, line 576: Yaw and roll are apparently swapped in `IMUGyroscope`, so Y and Z need to be swapped here.

iwubcode: Oh gosh, that explains so much!  I wondered why Dolphin wasn't behaving as expected...

Source/Core/Core/HotkeyManager.cpp
```diff
@@ -467,6 +507,13 @@ void HotkeyManager::LoadDefaults(const ControllerInterface& ciface)
   set_key_expression(HK_FREELOOK_ZOOM_OUT, SHIFT + " & S");
   set_key_expression(HK_FREELOOK_RESET, SHIFT + " & R");
 
+  set_imugyro_expression(HGGP_FREELOOK_ROTATE, 0, "`Click 1` & `Axis Y-`");  // Pitch left
+  set_imugyro_expression(HGGP_FREELOOK_ROTATE, 1, "`Click 1` & `Axis Y+`");  // Pitch right
+  set_imugyro_expression(HGGP_FREELOOK_ROTATE, 2, "`Click 1` & `Axis X+`");  // Roll left
+  set_imugyro_expression(HGGP_FREELOOK_ROTATE, 3, "`Click 1` & `Axis X-`");  // Roll right
+  set_imugyro_expression(HGGP_FREELOOK_ROTATE, 4, "`Click 2` & `Axis X-`");  // Yaw left
+  set_imugyro_expression(HGGP_FREELOOK_ROTATE, 5, "`Click 2` & `Axis X+`");  // Yaw right
```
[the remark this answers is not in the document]
iwubcode: Oh, did not know that.  Do you have any examples you can point me to?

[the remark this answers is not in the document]
iwubcode: Found it in `WiimoteEmu.cpp`

iwubcode:
I took another stab at the freelook hotkey buttons in the UI.  However, if that is too radical of a change, I can switch them back to a single group.

Thank you for reviewing @Techjar .  About the mouse sensitivity, were you able to get an expression you liked?  I need to retry the behavior on master but thought it was just as sensitive.

jordan-woyak:
The poorly named (my fault) "Axis X/Y" (relative mouse movement) mappings all had os-specific smoothing worked into them to make them semi-usable for analog sticks.
We should probably eliminate this now that input expressions support a "smooth" function. Perhaps we can just add new "Relative Mouse X/Y" inputs and leave the "Axis" inputs as non-detectable just for backwards compatibility?

Techjar:
To be clear I also tried binding my Wiimote gyro to the freelook rotation and it went absolutely nuts.

iwubcode:
Yes, I think I see my issue.  I'm binding angular velocity as the degrees for the rotate.  That's wrong.  Just not sure offhand what to do to fix it.

Techjar:
Well `RotateView` is cumulative right? So velocity actually is an acceptable input for it, I think it just needs to be scaled since the input is in radians.

iwubcode:
If someone is able to test the latest version with a real gyro, that'd be helpful.  With [OpenTrackDSUServer](OpenTrackDSUServer) and a circular motion, I'm getting multiple arcs for in game rotations, despite PadTest showing a 180 turn.

Uncertain if this is normal behavior for Dolphin (I see the same thing in Citra with OOT's first-person view) or a bug in my DSU server.  I was expecting 1:1 mapping.

At least in my mind, this is the only thing keeping this PR from being WIP still.

mbc07:
I can give a try with the DualShock 4 motion sensors. Just to be clear, what is the expected behavior?

iwubcode:
@mbc07 - I'm actually trying to figure out the expected behavior.  I don't own / haven't setup any gyro based controllers so I don't know what the normal behavior is.  If you move a DS4, do you get a 1:1 mapping in Dolphin?

For this freelook feature, when using my new DSU server, I see a 1:1 mapping in "PadTest" but not in Dolphin or Citra.  I would assume "PadTest" is correct but I have no other controllers to confirm..

mbc07:
I gave it a test and apart from some inverted axes (more on that later) it seems to work as it should

I used a DualShock 4 v2 (the one where the light bar is visible from the touchpad) with DS4Windows 2.0.16. I booted Zelda Wind Waker and started messing with the free look controls as soon as the player has control of Link (right after him waking up).

Up and down movement is working fine, however, roll and yaw movements are inverted. For yaw, assigning Yaw Right to Yaw Left and Yaw Left to Yaw Right on the hotkey settings was enough to correct that. 

Roll, however, I couldn't figure out. With Roll Left assigned to Roll Left and Roll Right assigned to Roll Right, the free-look camera was inverted (it actually rotated left when you rolled the controller right and vice versa). Then I tried to invert the bindings, just like I did with Yaw, and while that works, some time later the Roll movements goes inverted again (I couldn't figure out what triggers that).

TL;DR it works as it should if you invert the Yaw assignments, except for Roll, which is prone to do the inverted movement even after trying to account for that by inverting the bindings...

iwubcode:
Thanks for testing @mbc07 !  I actually forgot you can  use a Wiimote and luckily had one to test and I'm seeing similar behavior.

I will need to look and see why that is..

iwubcode:
@mbc07 - I've updated this to fix some things, drift should be gone and 1:1 motion is now as expected.  It should also fix the axis inversion you were experiencing.

One thing I'm seeing is some odd roll rotation with the Wiimote.  I don't think I'm adding it with my movement.  Do you see any surprising behavior with your DS4?

---

As an aside, I need to tweak the mouse movements some more.  If anyone gets some working defaults, let me know!

mbc07:
Sorry for the delay. Tested again with the DualShock 4 and everything is working as it should, without needing to invert any axes. Hasn't tested mouse movement, though...

iwubcode:
I followed @jordan-woyak 's recommendation and added a "RelativeMouse" input that doesn't do smoothing and deprecates the "Axis" for mouse movement.

Still trying to get mouse movement to match master.  That's the only blocker at this point.

iwubcode:
@mbc07 - if you don't mind, at your earliest convenience, please test this again.  I rebased with master which required some changes due to the new freelook control types!  From what I can tell, everything works appropriately but it'd be nice to get a second look.

mbc07:
@iwubcode just gave a try and noticed the UI has some oddities:
* "Freelook Motion Simulation" tab has a "Gyro" visual indicator that doesn't seem to do anything? (but the default mouse bindings still works in-game)

* "Freelook Motion Input" tab has the Gyro bindings assigned by default (yay!), but they don't do anything unless you select the correct controller on the "Device" drop-down menu. But then, by doing that, all other hotkeys will stop working unless you reassign every one of them with the proper "DInput/Keyboard Mouse/" prefix (that's on Windows, it's likely something else on Linux/macOS). 

    So, yeah, that's not trivial to resolve and I can't think of any reasonable compromise other than, perhaps, shipping with the Gyro bindings blank by default and needing to have the user configure them manually? At least it would be way easier than needing to reassign every single hotkey binding from all other tabs.

Other than that, everything is working as intended, just like the last time I tried this PR (tested with a DualShock 4 on Wind Waker)...

iwubcode:
@mbc07 

> "Freelook Motion Simulation" tab has a "Gyro" visual indicator that doesn't seem to do anything? (but the default mouse bindings still works in-game)

Not sure, when I do the appropriate mouse buttons, I do see movement in the gyro?

> "Freelook Motion Input" tab has the Gyro bindings assigned by default (yay!), but they don't do anything unless you select the correct controller on the "Device" drop-down menu. But then, by doing that, all other hotkeys will stop working unless you reassign every one of them with the proper "DInput/Keyboard Mouse/" prefix (that's on Windows, it's likely something else on Linux/macOS).
> 
> So, yeah, that's not trivial to resolve and I can't think of any reasonable compromise other than, perhaps, shipping with the Gyro bindings blank by default and requiring the user to configure them manually? At least it would be way easier than needing to reassign every single hotkey binding from all other tabs.

I ran into the same issue.  I ended up changing just the gyro mapping itself so that only those had "DSUClient/0/...".  But the previous default mappings made that confusing.  So I took your advice and cleared them.

----

In addition, I updated this to also add a "RelativeMouse" for linux.  I do not have a linux box to test though, so if someone else could give it a whirl that'd be helpful.

I think this PR is ready for review

mbc07:
I thought I edited my comment but seems I forgot? Anyway, the visual indicator works if you leave the hotkey window open and do the mouse movements on the game window. Enabling background input also makes it works regardless of what window has the focus. So, yeah, you can disregard what I said about the visual indicator earlier...

iwubcode:
Last night, @MayImilae commented on the UI for the general settings, saying that the four quadrant UI layout was not attractive and wanted to try stacking them instead, I pushed that up, with the result being:

image

In my local copy, I went ahead and went back to the original solution which I would also be fine with:

image

Please let me know which one you prefer!

Miksel12:
I did some testing and found that the relative mouse input is way more sensitive than the mouse controls in master. It is very hard to control the camera, setting a multplier of 0.1 helps this but shows that the controls are more jittery compared to master. For me, this is a blocker as mouse controls are for me the most usefull and with this pr, they are very unwieldy.

Also, the deadzone and calibration setting doesn't really belong in motion simulation.

I tested the gyro controls with my DS4 which works great. The only problem I had was that the roll direction was flipped. The indicator shows correct behaviour. @mbc07 said he didn't have any issues so I'm not sure what is causing this.

Btw, I like the horizontal version the most.

iwubcode:
> I did some testing and found that the relative mouse input is way more sensitive than the mouse controls in master. It is very hard to control the camera, setting a multplier of 0.1 helps this but shows that the controls are more jittery compared to master. For me, this is a blocker as mouse controls are for me the most usefull and with this pr, they are very unwieldy.

Thanks for the feedback.  I did not bother setting defaults yet for mouse, as I figured @MayImilae or others will likely want to tweak it.  Personally, I used `RelativeMouse/3` and had no issues controlling the camera.  Testing it again now, it seems `RelativeMouse/4` is roughly equivalent to the same turn in master for Mario Kart Wii.  The mouse in master is using Qt, while the mouse in this PR is using the inputs system.  Other than remapping, the other advantage of using this over the master version, is here we don't have to re-center the mouse each time.

I *personally* think mouse usage is limited.  If you're using a controller, you have to stop and switch to the mouse control, which feels very unnatural.  It's fine for the occasional screenshot or if you're just wanting to look around and see what the developers did to create the world.  I'm trying to break this trend and get freelook used more generally but I also understand everyone who uses it, uses it like that today.

That's not to say that I want to break current users.  However, now that the input system is used, the non-gyro controls and sensitivity are completely customizable.

If there are technical errors, that's understandable (and there may well be) but I don't think we should block this PR based off not matching the feeling of master 100%.

>Also, the deadzone and calibration setting doesn't really belong in motion simulation.

Good point, I can likely add some additional variables to not handle those cases.

>I tested the gyro controls with my DS4 which works great. The only problem I had was that the roll direction was flipped. The indicator shows correct behaviour. @mbc07 said he didn't have any issues so I'm not sure what is causing this.

Hmm good to know.  I don't own a DS4 but will try my Wiimote setup again.

Miksel12:
> I think mouse usage is limited. If you're using a controller, you have to stop and switch to the mouse control, which feels very unnatural. It's fine for the occasional screenshot or if you're just wanting to look around and see what the developers did to create the world. I'm trying to break this trend and get freelook used more generally but I also understand everyone who uses it, uses it like that today.

Yeah, I mainly use free look to fly around so I'm using free look when I'm not touching the controls.  
By 'using a controller', do you mean using the axes or by using the gyro? I just tried using axes and with no multiplier they are also very sensitive. Besides, I have the axes mapped to pointing and nunchuk so I would have to add an expression to avoid triggering an action while using free look. And while gyro works great, I don't think a lot of people can/are using it.  And for the stuff I do with free look I do prefer mouse over gyro.

> That's not to say that I want to break current users. However, now that the input system is used, the non-gyro controls and sensitivity are completely customizable.
> 
> If there are technical errors, that's understandable (and there may well be) but I don't think we should block this PR based off not matching the feeling of master 100% (as long as we can get close).
> 
> Curious though, I tried `0.1` and found that to be a lot slower than master. Why do you find this PR unwieldy?

I meant to say that the default values feel unwieldy, with a multiplier it isn't a problem. But I think the default values should be usable so that users shouldn't have to change the values when they want to use a mouse. 

The free look description still contains the default mapping, which should be removed. 
It would be nice if all the free look settings were grouped together, I think  #8823  could fix this ;) .

iwubcode:
> Yeah, I mainly use free look to fly around so I'm using free look when I'm not touching the controls.
By 'using a controller', do you mean using the axes or by using the gyro? I just tried using axes and with no multiplier they are also very sensitive. Besides, I have the axes mapped to pointing and nunchuk so I would have to add an expression to avoid triggering an action while using free look. And while gyro works great, I don't think a lot of people can/are using it. And for the stuff I do with free look I do prefer mouse over gyro.

Great points here.  Using freelook to fly-around is likely going to remain the norm for many.  For that situation, mouse is king.

I do hope to introduce other usages for freelook though.

Right now I have this PR which allows for head-tracking or just gyro look.  I actually don't  use any gyro-controllers with Dolphin right now but if I did, I'd be using this feature with a very large divisor.  Just so i could peek a little at the rest of the game world.

Also, hooking this up to a normal controller, I hope that can be used in conjunction with my multi-screen PR for couch co-op play.

>I meant to say that the default values feel unwieldy, with a multiplier it isn't a problem. But I think the default values should be usable so that users shouldn't have to change the values when they want to use a mouse.

Agree completely about setting up sensible defaults.  Actually @Techjar and I were just talking about this.  I think we will have to use the original solution for now, unless we are ok on breaking existing users presets...

>It would be nice if all the free look settings were grouped together, I think #8823 could fix this ;) .

Agreed.  I want to split freelook out at some point.  It's a bit confusing having it in two places.

>The free look description still contains the default mapping, which should be removed.

Thanks for reminding me about this.  Should I make note that they can be configured or just remove completely?

iwubcode:
> Also, the deadzone and calibration setting doesn't really belong in motion simulation.

I was in the middle of removing them...but then I wondered: don't they make sense for a normal controller?

MayImilae:
> The free look description still contains the default mapping, which should be removed.

Please don't remove the default controls from the description. Most users won't remap it, so it's been kept deliberately even after it was possible to remap it. It's there for a reason!

iwubcode:
Thanks @MayImilae .  I'll leave that untouched then.

iwubcode:
It pains me to do so because I want this merged but marking as a draft again.  As I was trying to implement something else, I found needing similar UI functionality.  However, I don't want to implement this again.  So I'm going to do something a little more grand (that I was originally going to wait on).

It may replace this PR or I might just update it.

iwubcode:
@mbc07 / @Miksel12 - if either of you can re-test this at your earliest convenience with DS4, that'd be really helpful!

Bearborg:
Just tested this with a DS4. Pitch and yaw work perfectly, but roll seems to have no effect.

iwubcode:
@Bearborg - that's odd, were you testing with the "First Person" or "Six Axis" camera option?  Appreciate the testing!

Bearborg:
Ah whoops, you're right, I had it set to First Person. In Six Axis mode all axes work correctly.

iwubcode:
Converted to draft (again), since #8867 has some changes that were brought up

mbc07:
Just retested this PR with my DS4 and all camera modes seems to be working as intended, although Roll only works in Six Axis mode (I'm citing this just to confirm if that's the correct behavior). BTW, this PR still has the same issues that PR #8867 had initially: clicking on the controller button does nothing and the device dropdown list on the Freelook config window doesn't appear unless you manually resize the window.

iwubcode:
@mbc07 - Thanks for testing.  About your camera mode comment, yes that is correct behavior.

Also, yes, this is now based off of #8867 but I'm not keeping it perfectly in sync as I try to prepare #8867 to be merge-able.  I should have an update later today

iwubcode:
Hoping this is ready for review.  I made my best attempt at matching Dolphin master's mouse settings but it's not an exact match.  I'm open to suggestions.

I tested a xbox 360 controller, mouse and keyboard and a Wiimote using DSU client.

Miksel12:
> @Miksel12 / @mbc07 / @Techjar - happy holidays to you. If you have time in the next few days and can retest, that'd be very helpful to get some more data points!
Thank you, and happy holidays to you as well! 
I unfortunately can only test with keyboard and mouse right now but since that is the main way I use free look, I still can give some good feedback.

It seems to be improved since I last tested but I still is a bit too jittery for my liking. I tried using cursor instead of axis and this works much smoother and nicer in my experience. I tried using the relative function with cursor to simulate the axis behaviour but I think relative doesn't work the way I thought it did.

iwubcode:
@Miksel12 - thanks for testing!  Can you post one of the input expressions you're using for reference?

The relative mouse logic was something I added based on a suggestion from @jordan-woyak .  I might have implemented it incorrectly or possibly there's a better expression we can  use that doesn't require new logic.

It's been the biggest pain point of this PR :/.

iwubcode:
@jordan-woyak asked for direct quaternion usage, so I prepped that in #9366 .  He also is planning on implementing true relative mouse input, so I guess I'll wait a little longer for this to be completed.  Marking as draft for now!

Miksel12:
> @Miksel12 - thanks for testing! Can you post one of the input expressions you're using for reference?

I used if(`Click 2`,`Cursor X-`,0), which has different behaviour than master but I think it works quite well.

iwubcode:
@Miksel12 - sorry to bother but I'd be curious what you think of the default controls for mouse now.  Does it feel better?

Miksel12:
> @Miksel12 - sorry to bother but I'd be curious what you think of the default controls for mouse now. Does it feel better?

Yes! Much better, it feels like master now. I did change the multiplier to 0.1.  
I did have some other feedback, in the Wii controller tabs, the motion input tab is on the right side of the motion simulation tab which isn't the case in the free look controller. 

About the deadzone setting, you said:
> > Also, the deadzone and calibration setting doesn't really belong in motion simulation.
> 
> I was in the middle of removing them...but then I wondered: don't they make sense for a normal controller? (well deadzone at least)

And this is the case for normal deadzone, but this deadzone combats drift which mouses and sticks don't have. Only gyro's.
I'm not sure how calibration could be used with a controller as I can't test with controller.

iwubcode:
@Miksel12 - thanks for testing!  Good catch on the motion input being on the wrong side.  I updated it.

It may make sense to make a new control for motion simulation.  Not sure whether this should be similar to the gyro control or be something entirely new.  Open to suggestions.

I'm going to re-open this PR.  The relative input code (for the mouse) is thanks to @jordan-woyak !  He did mention a potential issue around deadlocking but I see that as an edge case and believe the only concern would be a larger than expected delta would be experienced on rare occasions.  I don't imagine it is enough to block this PR.

Lastly, while I'd be happy to get rid of the euler-to-quaternion function in favor of #9366 , I don't think that's something that needs to block this PR either and can be fixed up later.

iwubcode:
@JMC47 gave his seal of approval on IRC.

MichaelJW:
@iwubcode I tested this branch out myself and it's really cool! But I'd like to make a suggestion. As I understand it, this allows changing the pitch, roll, and yaw by inputting gyroscopic rate (degrees/sec) data, either directly from a motion controller or via simulating a motion controller; also, displacement x/y/z can be incremented or decremented via binary inputs.

I propose adding another tab that allows setting these six values directly, i.e. allowing six binds: three for pitch, roll, and yaw (in degrees from neutral) and three more for displacement x, y, and z (in metres from centre). No easing or motion simulation, just jump the camera to the specified position and rotation.

This would allow mapping to an HMD with no drift, and might make it viable to implement free look using TrackIR, FaceTrackNoIR, and the like.

Granted, DSU/Cemuhook protocol doesn't have fields for displacement and rotation, but that data could always be passed through via the accelerometer and gyroscope fields anyway, or via the analog button fields for that matter - an HMD is unlikely to need binds for all those face buttons. 

What do you think? (I'm happy to have a go at implementing this myself if you like, but I wanted to run it by you rather than just diving in.)

iwubcode:
@MichaelJW - thank you for your comments.

>As I understand it (and please correct me if I'm wrong), this allows changing the pitch, roll, and yaw by inputting gyroscopic rate (degrees/sec) data, either directly from a motion controller or via simulating a motion controller; also, displacement x/y/z can be incremented or decremented via binary inputs.

Yes, the gyroscope is angular velocity (in radians)

>  This would allow mapping to an HMD with no drift, and might make it viable to implement free look using TrackIR, FaceTrackNoIR, and the like.

Yes, I can say that OpenTrack gives absolute values.  So I had to create a delta solution and it's kinda klunky.  So I do see some advantages to providing such a solution.

However, I'm surprised the HMD provides a way to get the exact orientation.  I'm not necessarily opposed to such a solution but my main concern is that if gyros are drifting in general, it feels like we need a general purpose solution for that.  Looking at discussion online, there are solutions to [combat](https://stackoverflow.com/questions/14210206/gyroscope-drift-on-mobile-phones) [drift](https://www.gamasutra.com/view/feature/1494/wheres_the_wiimote_using_kalman_.php) in [gyros](https://stackoverflow.com/questions/1586658/combine-gyroscope-and-accelerometer-data).  I'm not sure if any are the right way to tackle this problem.  I need to talk to @jordan-woyak about this and get his thoughts.

> I propose adding another tab that allows setting these six values directly, i.e. allowing six binds: three for pitch, roll, and yaw (in degrees from neutral) and three more for displacement x, y, and z (in metres from centre). No easing or motion simulation, just jump the camera to the specified position and rotation.

Just to note, I actually started on a similar idea last year for a _different_ reason.  I'm looking to implement [FlickStick](https://www.youtube.com/watch?v=C5L_Px3dFtE) in Dolphin (I didn't actually know that was a thing but found out later what I implemented had already had a coined term!).

MayImilae:
Tested! I have a few comments, but overall it absolutely works. Though it does immediately drift, but I understand that is a known issue and is being looked into. Anyway, comments.

The main one for me is that when mapping it one to one (roll left to roll left, pitch up to pitch up, etc), it's like I'm controlling the world and not a camera in the world. So like, if I lay down so left side of my head is facing down, the floor is now to my left. But with this setup currently (and 1:1 mapping), if I place the left side of the controller facing down, the floor is on the right. If I was controlling a camera in a space, I'd expect it to replicate the way it works with my head and eyes - lay a phone onto its left side, the floor is on the left in the recorded footage. However, in the PR as it is (with 1:1 mapping), it is instead locking the floor to the bottom of the controller, giving the impression that the world is rotating around the camera, rather than the camera rotating in the world. It feels WEIRD.

What I would prefer is for the controller to be a camera in the game world, and the IRL floor to be the "floor" of the game. That way, rotating the controller IRL should feel the same as rotating a phone IRL. That's the most intuitive control method for this, imo.

So I have another comment, but it's kind of related both to this PR and the last one. But hear me out. Motion Input and Motion Simulation tabs should be merged into one, and just called Rotation. Like, I know you kept it consistent with Wii Remotes but, this is not Wii Remotes. With Wiimotes, we need that separation because Motion Simulation is providing a number of abstract input methods (tilt, swing), while Motion Input is providing raw axis. But with Free Look, we have only the pure axis to deal with. No abstraction is necessary. So yea, we can just merge those two and simplify it tremendously! Also we aren't simulating motion at all so.

iwubcode:
@MayImilae - thanks for your feedback.  I'm going to merge the two tabs into one, i agree there's no real reason to have both.  I'll just call it "Rotation", then add a similar one for "Position" later.

@MichaelJW - with a single "Rotation" tab, we can have a "Incremental rotate" and an "Exact rotate" set of input (need to think of better names...).  This should allow for a "FlickStick" solution and also your HMD approach / my OpenTrack approach.

Miksel12:
> EDIT: With some further testing, it turns out it was way easier to create that "light bar is a camera" feel. All I had to do was invert Roll. Well, that's easy! As an advanced feature I don't feel too badly about not holding the users hand with this, but we could maybe provide invert options to make it easier? Just as a convenience. But it's fine as is though. That being said, I'm not sure if the inverted roll is a bug or a consequence of directly patching the controller input in. Might be worth looking into?

That is weird, last time I tested I didn't need to invert anything. @mbc07 also didn't need to invert: https://github.com/dolphin-emu/dolphin/pull/8747#issuecomment-616245536
Are you using a dualshock 4?

iwubcode:
@MayImilae / @mbc07 / @MichaelJW / @Miksel12 - would one or more of you be able to re-test this?  I just want to confirm I didn't break anything.  I attempted to fix the inverted roll (the keyboard was also inverted, so if you use that you need to click "default") and also I updated to a single tab "Rotation".  Some internal changes but everything worked for keyboard / and Wiimote testing (both DSU and bluetooth).

Thank you all for your help!

MayImilae:
Tested! Roll is no longer backwards! Also double checked the mouse controls once again, and everything is fine even after I pressed Default.

We covered this in IRC, but with some trickery with the parser to enable the gyro when a button is held, this is quite good.

ds4gyrohold

iwubcode:
Thank you @MayImilae !  I updated the text and the screenshot in the first post

Miksel12:
I believe evdev can use motion input directly so maybe: "motion input **may** require configuring..." which would be more correct.

iwubcode:
Thanks @Miksel12 .  I updated the text to reflect that and the image on the first post

MichaelJW:
> @MayImilae / @mbc07 / @MichaelJW / @Miksel12 - would one or more of you be able to re-test this? I just want to confirm I didn't break anything. I attempted to fix the inverted roll (the keyboard was also inverted, so if you use that you need to click "default" to update the bindings) and also I updated to a single tab "Rotation". Some internal changes but everything worked for me when testing with a mouse&keyboard / Wiimote (both DSU and bluetooth).
> 
> Thank you all for your help!

Just tested it out here too with an HMD and it works well when passing the gyro pitch/roll/yaw rates. Looking forward to that Exact Rotation panel! Let me know if I can do anything to help.

Source/Core/DolphinQt/HotkeyScheduler.cpp
```diff
@@ -139,10 +139,22 @@ void HotkeyScheduler::Run()
 
   while (!m_stop_requested.IsSet())
   {
-    Common::SleepCurrentThread(1000 / 60);
+    const u64 freelook_update_freq = 200;
+    const u64 interval = SystemTimers::GetTicksPerSecond() / freelook_update_freq;
+    const u64 now = CoreTiming::GetTicks();
```
Filoppi, line 144: CoreTiming::GetTicks() read might not be atomic/safe on all architectures? Why don't you hook this inside the emulation thread given it depends on emulation timings? It's weird it's in the hotkeys thread if then it needs to depend on the current game speed. I guess it's good for performance, but if this only update inputs, I doubt it would make a relevant difference.

Source/Core/DolphinQt/HotkeyScheduler.cpp
```diff
@@ -139,10 +139,22 @@ void HotkeyScheduler::Run()
 
   while (!m_stop_requested.IsSet())
   {
-    Common::SleepCurrentThread(1000 / 60);
+    const u64 freelook_update_freq = 200;
```
Filoppi, line 142: I'd move this to a constexpr above, together with the hotkeys 60Hz update rate (I've also done the same change, it's nicer and highlights that there shouldn't be any negative consequence in changing them).

Source/Core/DolphinQt/HotkeyScheduler.cpp
```diff
@@ -139,10 +139,22 @@ void HotkeyScheduler::Run()
...
+    const u64 interval = SystemTimers::GetTicksPerSecond() / freelook_update_freq;
+    const u64 now = CoreTiming::GetTicks();
 
-    if (Core::GetState() == Core::State::Uninitialized || Core::GetState() == Core::State::Paused)
+    if (now - m_last_freelook_ticked > interval)
```
Filoppi, line 146: Guessing it should be >= interval? e.g. if interval is 0, it should run every time (I know it's an impossible case).

Source/Core/DolphinQt/HotkeyScheduler.cpp
```diff
@@ -139,10 +139,22 @@ void HotkeyScheduler::Run()
...
+    {
+      m_last_freelook_ticked = now;
+      g_controller_interface.SetCurrentInputChannel(ciface::InputChannel::FreeLook);
       g_controller_interface.UpdateInput();
+      FreeLook::UpdateInput();
+    }
+
+    Common::SleepCurrentThread(1000 / 60);
```
Filoppi, line 154: If this sleeps the thread for ~16ms, how is freelook going to update at 200Hz? if (now - m_last_freelook_ticked > interval) is always gonna be true.

Source/Core/DolphinQt/HotkeyScheduler.cpp
Filoppi: Why is the refresh rate 200Hz? Maybe it should be hinted. Is it because you want the input polling to be as up to date as possible when a new render starts? Or can freelook actually output more frames that the game naturally would (I don't know how it works)?

Filoppi (commented):
Comment

Source/Core/DolphinQt/HotkeyScheduler.cpp
iwubcode: FreeLook used to be entirely managed via hotkeys and therefore was on the hotkey thread.  In order to keep things simple in the review where I made FreeLook a proper controller, I kept it on the hotkey thread.  I plan on reworking this in a future PR.

Source/Core/DolphinQt/HotkeyScheduler.cpp
iwubcode: This is a good point.  I thought I had tested that and this way felt closer to master but it does seem odd now that you mention it.

Source/Core/DolphinQt/HotkeyScheduler.cpp
iwubcode: I spent a very long time getting it to match master (you can see all the comments about it not feeling right by users).  Therefore I'm hesitant to change much related to the timing.  The previous timings were from Qt events.  I can't recall where I found the value.  But your other point makes me think it's not really working as I intended.  I will investigate.

Source/Core/DolphinQt/HotkeyScheduler.cpp
Filoppi: I guess you had not noticed because games don't run above 60 unless the speed is unlocked, but with a sleep in the middle, this is unlikely to actually update at 60, it's gonna lose time and you are going to miss input updates on frames quite often.

Source/Core/DolphinQt/HotkeyScheduler.cpp
Filoppi: The only advantage in updating at a higher rate than the game would be to have more up to date values, which is really important if you are wearing a headset (if that is possible), otherwise this is likely useless and was only done to increase the accuracy of fake relative inputs.

Source/Core/DolphinQt/HotkeyScheduler.cpp
Filoppi: If it works like I'm guessing, freelook input should be updated just before rendering starts, or as late as possible within a frame anyway, and should be in the cpu thread. Otherwise it's never gonna be accurate.

Source/Core/DolphinQt/HotkeyScheduler.cpp
iwubcode: I still need to test but I believe I picked a value of 200 likely to match Wiimote which I believe is [also updated at 200hz](https://github.com/dolphin-emu/dolphin/blob/0029ca84b07f44e3ce8c80a44fa0285a966b34d3/Source/Core/Core/HW/Wiimote.h#L78) and also because less frequent  updates yielded less accurate motion results.

Filoppi: Wiimote BT inputs are at 200 (I guess fixed, independently of the emulation speed).
Why would less frequent updates have less accurate motion results if you are just reading the current pitch/yaw/roll value?

Source/Core/DolphinQt/HotkeyScheduler.cpp
iwubcode: Well, from that standpoint it should exist on the graphics thread (which may or may not be the cpu thread if dual core is enabled).  However, I'm not sure if tying inputs to framerate is a great idea.

Filoppi: Sure, why not? If I run the game at 200%, freelook should update inputs at twice the speed.
And you tick timing checks seem to be exactly for that reason.

Filoppi: After the edit to your comment: you seem to want to link Freelook updates to the game speed, as that's what you'd get if you check the emulation ticks.
If you are increasing values by a fixed amount every frame, that is wrong and you should simply use a delta time there.

Source/Core/DolphinQt/HotkeyScheduler.cpp
iwubcode: DSU is angular velocity.  The values passed are relative motion, not absolute, so I believe the input rate is important to get that smooth feel.

Filoppi: Sure, if you are accumulating these values to move the camera within that update in the hotkey thread, then it makes sense.
Just add a comment explaining that's why it's 200 then.

jordan-woyak: Yeah. Ideally angular velocity values would be processed as their source device produces them. Googling implies DS4 controllers support 1000Hz over Bluetooth. It's a shame to only process gyro data at 200hz and not at all in sync with device updates.

The least intrusive solution I can think of would be to add gyro/accel processing at the input backend level. Absolute rotation angle inputs can then be provided.

Filoppi: Exactly. Little code involved, cleaner implementation, more accurate.

iwubcode: I'm fine with something like that but that is a problem beyond this PR imo.

Also, I worry a bit about your assumptions.  The DSU is provider specific.  For instance, my OpenTrack DSU implementation does not happen at the rate of the DS4.  In particular, it listens over UDP to OpenTrack, so it's whatever rate that program provides.  But that is just one provider and anyone is free to create their own.  I suppose we could make it configurable but eh.

Filoppi: Just use the delta time between packets. Or try to deduce the frequency at which they are sent and then use that number. It won't be any less accurate than it is now.

iwubcode: If there's work to be improved on the input side, that's fine but that's not really the kind of thing I'm interested in working on personally.  And if I'm being honest, I wouldn't have any idea where to begin.  Multiple people, with different devices, have stated that the current implementation works fine.  This may be sub-optimal but I don't feel like such a solution needs to hold up _this_ review.

iwubcode: (I'll just add, the input object is used for multiple devices: keyboard/mouse, gamepad, dsu, etc..so one frequency is not going to cut it.  It'd have to be the highest frequency from all the inputs because each input could hypothetically have its own frequency rate)

iwubcode:
Thanks to @Filoppi for noticing that the `GetTicks` function shouldn't be ran on a separate thread.

I went back to the basics and now update the FreeLook input every 1/60th of a second.  It seemed to work from my testing with Wiimote DSU.

I also made a change to have the movement consider the delta time.  That will help if we ever move this to a separate thread that runs faster or slower (may also help with the accelerometer?  I haven't tested yet).  Because of this change, I also re-tested gamepad and keyboard/mouse.  I believe I have things working very close to master.

@MayImilae / @mbc07 / @MichaelJW / @Miksel12 - sorry to nag you all again.  But if some of you can do a sanity check, it would be very helpful.  I really hope I didn't break anything!

Filoppi:
> I also made a change to have the movement consider the delta time. That will help if we ever move this to a separate thread that runs faster or slower (may also help with the accelerometer? I haven't tested yet). Because of this change, I also re-tested gamepad and keyboard/mouse. I believe I have things working very close to master.
Ideally you'd want to divide a fixed value by the emulation speed IMO, or the possibile huge differences in timings between frames (e.g. one frame could take 16ms, the next one 3ms, while you still run at 60 overall) would add jerk to the camera movement. But of course it's fine as it is now and that change would require Freelook to update once per video frame independently of the emulation speed.

Great changes overall.

Miksel12:
@iwubcode There seems to be quite a lot of drift now on my DS4 and it seems like it is harder to control. I did a sanity check on master which showed the correct behaviour.

On a somewhat related note, when I press on default with my DS4 as DSU selected input, it switches back to keyboard and mouse as selected input and configures the default mappings for keyboard. I would think it would be better if Dolphin didn't configure anything if the selected input doesn't have default mappings. I was half way into mapping my DS4 when I remembered I could just click on default, only to see my mappings replaced with keyboard mappings.

Filoppi:
> @iwubcode There seems to be quite a lot of drift now on my DS4 and it seems like it is harder to control. I did a sanity check on master which showed the correct behaviour.
> 
> On a somewhat related note, when I press on default with my DS4 as DSU selected input, it switches back to keyboard and mouse as selected input and configures the default mappings for keyboard. I would think it would be better if Dolphin didn't configure anything if the selected input doesn't have default mappings. I was half way into mapping my DS4 when I remembered I could just click on default, only to see my mappings replaced with keyboard mappings.

Off topic but "Default" restores the default profile, a profile also includes a specific default device, and in the case of the "Default" profile, the  default device is the first one in the list of devices, which is always the keyboard and mouse device.

Miksel12:
> Off topic but "Default" restores the default profile, a profile also includes a specific default device, and in the case of the "Default" profile, the default device is the first one in the list of devices, which is always the keyboard and mouse device.
> If we had a default mapping for a DSU device, that would be listed on the stock profiles on the right.

I see, misconception on my part.

Filoppi:
> Drats, ok. Thank you for testing. It's gotta be the DS4 poll rate is higher than the Wiimote. I wasn't able to find a solution that I was happy with keeping FreeLook at 200hz without introducing another thread or spinning. Maybe I'll have to add another thread for now with a `condition_variable` to keep from spinning, still rather ugly..

If you are going to add to this PR, I think it would be much better to just calculate the absolute rotation from within DSU, and add it as an input, then move FreeLook to the CPU thread and multiply the delta time you just added by the game speed.
That sounds better than a separate thread IMO, a separate thread won't ever be 100% aligned to the video frame rate, unless you implement some form of thread locking, which wouldn't great.

But really, as a simple solution you can just run the hotkeys at 200Hz, it's not gonna break anything.

iwubcode:
> If you are going to add to this PR, I think it would be much better to just calculate the absolute rotation from within DSU, and add it as an input, then move FreeLook to the CPU thread and multiply the delta time you just added by the game speed.
That sounds better than a separate thread IMO, a separate thread won't ever be 100% aligned to the video frame rate, unless you implement some form of thread locking, which wouldn't great.

Well, the problem seems that the input is only updated in `UpdateInput` and that happens at some rate.  I imagine all devices have this problem (not just DSU).

However, assuming we could get around that something like what you said would be the ideal solution.

> But really, as a simple solution you can just run the hotkeys at 200Hz, it's not gonna break anything.

I was thinking the thread would eat a lot more cycles but I guess I could try this.

Miksel12:
> What do you mean checked with master? You mean there's an issue with the mouse controls too?

No, I haven't played around with gyro for some time so I wasn't sure if the behaviour I saw was purely from your update or maybe a faulty DS4.

Filoppi:
> Well, the problem seems that the input is only updated in `UpdateInput` and that happens at some rate. I imagine all devices have this problem (not just DSU).
Not exactly sure what you mean. DSU is the only device backend that has got a "velocity" input which you need to multiply by delta time and add to some total sum of a rotation.
When DSU updates, you can retrieve the whole list of packages that were sent over UDP from the previous update, so if you sum all all the velocity inputs there and multiply them by the delta time between packages, you are going to have the absolute angular rotation in the most accurate way you could. There is no other reason to update input more often that the game frame rate otherwise AFAIK.

iwubcode:
@Miksel12 - at your earliest convenience, I updated the poll rate to 5ms.  Please try again and see if it's any better?

Miksel12:
Controls great again and no drift.

Source/Core/Core/FreeLookManager.h
```diff
@@ -13,14 +16,16 @@ namespace ControllerEmu
...
 }  // namespace ControllerEmu
 
 enum class FreeLookGroup
 {
   Move,
   Speed,
   FieldOfView,
-  Other
+  Other,
+  Rotation,
```
Rumi-Larry, line 28: Shouldn't "Other" be always the last one?

leoetlino (approved):

MayImilae (approved):
Gave it another try, and the drift problems are MUCH improved and the UI all still works and looks fine from the previous changes.

LGTM!

JMC47:
Multiple Approvals and Tests.  I also tested it and it was an improvement.

PR #8748, opened 2020-04-17
Android: Add Install WAD to menu_game_grid

Ebola16:
Implements https://bugs.dolphin-emu.org/issues/10788
Useful for setting up mod loaders.

### DIFF
--- a/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/ui/main/MainPresenter.java
+++ b/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/ui/main/MainPresenter.java
@@ -1,11 +1,13 @@
 package org.dolphinemu.dolphinemu.ui.main;
 
+import android.app.Activity;
 import android.content.BroadcastReceiver;
 import android.content.Context;
 import android.content.Intent;
 import android.content.IntentFilter;
 import android.widget.Toast;
 
+import androidx.appcompat.app.AlertDialog;
 import androidx.localbroadcastmanager.content.LocalBroadcastManager;
 
 import org.dolphinemu.dolphinemu.BuildConfig;
@@ -121,13 +123,30 @@ public final class MainPresenter
 
   public void installWAD(String file)
   {
-    if (NativeLibrary.InstallWAD(file))
+    final Activity mainPresenterActivity = (Activity) mContext;
+
+    AlertDialog dialog = new AlertDialog.Builder(mContext, R.style.DolphinDialogBase).create();
+    dialog.setTitle("Installing WAD");
+    dialog.setMessage("Installing...");
+    dialog.setCancelable(false);
+    dialog.show();
+
+    Thread installWADThread = new Thread(() ->
     {
-      Toast.makeText(mContext, R.string.wad_install_success, Toast.LENGTH_SHORT).show();
-    }
-    else
-    {
-      Toast.makeText(mContext, R.string.wad_install_failure, Toast.LENGTH_SHORT).show();
-    }
+      if (NativeLibrary.InstallWAD(file))
+      {
+        mainPresenterActivity.runOnUiThread(
+                () -> Toast.makeText(mContext, R.string.wad_install_success, Toast.LENGTH_SHORT)
+                        .show());
+      }
+      else
+      {
+        mainPresenterActivity.runOnUiThread(
+                () -> Toast.makeText(mContext, R.string.wad_install_failure, Toast.LENGTH_SHORT)
+                        .show());
+      }
+      mainPresenterActivity.runOnUiThread(dialog::dismiss);
+    }, "InstallWAD");
+    installWADThread.start();
   }
 }
--- a/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/ui/main/MainPresenter.java
+++ b/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/ui/main/MainPresenter.java
@@ -4,10 +4,12 @@ import android.content.BroadcastReceiver;
 import android.content.Context;
 import android.content.Intent;
 import android.content.IntentFilter;
+import android.widget.Toast;
 
 import androidx.localbroadcastmanager.content.LocalBroadcastManager;
 
 import org.dolphinemu.dolphinemu.BuildConfig;
+import org.dolphinemu.dolphinemu.NativeLibrary;
 import org.dolphinemu.dolphinemu.R;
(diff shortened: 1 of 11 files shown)


JosJuice:
Haven't tested it, but the code seems fine. One comment, though:

I know that one of the Android contributors we've had before has been pickier than me about making sure that long-running operations run on a separate thread so that the UI remains responsive. Importing a WAD likely counts as such an action, since it involves reading and writing tens of megabytes to storage. You might want to consider moving it to a different thread, but I personally don't consider this to be a blocker.

Ebola16:
I've never written multi-threaded code so please correct me if I'm misunderstanding something here. Wouldn't we want WAD installation to block on the UI thread here? We don't want users to begin emulation if a WAD is being installed. A toast is also given for success or failure so users will be notified when WAD installation finishes.

JosJuice:
> We don't want users to begin emulation if a WAD is being installed.

I guess that is a good idea, but that can be done without blocking the UI thread. (For instance, it's possible to show an "installing" dialog that prevents the user from interacting with the UI behind... Though I guess that might be unnecessary if WAD installation is quick enough.) The reason why blocking the UI thread is bad is because it can make users think that the app has frozen, and Android can even show the "application not responding" message if it goes on for a few seconds.

But again, I don't consider this a blocker, since WAD installs don't take all that long.

Ebola16:
@JosJuice let's avoid the (very unlikely) chance for ANRs. Please check to make sure my latest commit is the best way to accomplish this.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/ui/main/MainPresenter.java
```diff
@@ -121,13 +123,27 @@ public void onDirectorySelected(String dir)
...
-    if (NativeLibrary.InstallWAD(file))
+    Thread installWADThread = new Thread(() ->
     {
-      Toast.makeText(mContext, R.string.wad_install_success, Toast.LENGTH_SHORT).show();
-    }
-    else
-    {
-      Toast.makeText(mContext, R.string.wad_install_failure, Toast.LENGTH_SHORT).show();
-    }
+      Looper.prepare();
```
JosJuice, line 128: I haven't seen this solution of using a Looper on a separate thread in order to do UI stuff directly on that thread. I'm assuming it works if you've tested it, but I can't find any documentation on whether it's actually thread-safe... The more common solution would be to use `runOnUiThread`.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/ui/main/MainPresenter.java
```diff
@@ -111,4 +120,30 @@ public void onDirectorySelected(String dir)
   {
     mDirToAdd = dir;
   }
+
+  public void installWAD(String file)
+  {
+    Thread installWADThread = new Thread(() ->
+    {
+      Looper.prepare();
+      AlertDialog dialog = new AlertDialog.Builder(mContext).create();
```
JosJuice, line 129: To match the newly merged PR #8718, please use `R.style.DolphinDialogBase` here. (You will need to rebase on latest master first.)

Ebola16:
@JosJuice Ok, this implementation should be a little more standard.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/ui/main/MainPresenter.java
```diff
@@ -121,13 +123,31 @@ public void onDirectorySelected(String dir)
...
+      else
+      {
+        mainPresenterActivity.runOnUiThread(
+                () -> Toast.makeText(mContext, R.string.wad_install_failure, Toast.LENGTH_SHORT)
+                        .show());
+      }
+    }, "InstallWAD");
+    installWADThread.start();
+
+    dialog.dismiss();
```
JosJuice, line 151: Won't this dismiss the dialog as soon as the thread starts executing, rather than when it's done? It would make more sense to me to do the dismissing from the WAD installation thread.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/ui/main/MainPresenter.java
```diff
@@ -121,13 +123,31 @@ public void onDirectorySelected(String dir)
 
   public void installWAD(String file)
   {
-    if (NativeLibrary.InstallWAD(file))
-    {
-      Toast.makeText(mContext, R.string.wad_install_success, Toast.LENGTH_SHORT).show();
-    }
-    else
+    final Activity mainPresenterActivity = (Activity) mContext;
```
JosJuice, line 126: This cast is a bit ugly, but meh, I can't think of a specific way to improve this...

Ebola16: I agree and that's why I originally went with calling `Looper` directly. I remember reading somewhere that doing so was a possible solution to my problem. I looked at the documentation for Looper itself and it seemed viable but I found nothing that said it was truly thread safe. We might as well go with what I currently have since `runOnUIThread` is used elsewhere in Dolphin.

Ebola16:
The `dialog.dismiss()` issue has been resolved.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/ui/main/MainPresenter.java
```diff
@@ -111,4 +120,33 @@ public void onDirectorySelected(String dir)
...
+                () -> Toast.makeText(mContext, R.string.wad_install_success, Toast.LENGTH_SHORT)
+                        .show());
+      }
+      else
+      {
+        mainPresenterActivity.runOnUiThread(
+                () -> Toast.makeText(mContext, R.string.wad_install_failure, Toast.LENGTH_SHORT)
+                        .show());
+      }
+      dialog.dismiss();
```
JosJuice, line 148: I'm not sure if this is thread-safe... Probably best to wrap this in `runOnUiThread` too.

Ebola16:
@JosJuice Done

Ebola16:
I think this is good to merge.

Sanmorin:
Good afternoon. This option is not available in the Android TV Dolphin menu. Can you add it? Tested on Dolphin 5.0-12031

Ebola16:
I'll look into that.

PR #8750, opened 2020-04-18
IOS/FS: Fix FST write failure on some platforms

leoetlino:
On some platforms (like Windows), the temporary file must be closed
before it can be renamed.

I guess nobody noticed this for so long because (1) the FS code has a
failsafe for missing FST entries (because existing users do not have
a FST), and most games do not care about file metadata;
(2) the write failures can only be seen in the logs.

Because we don't want this to break, I have turned the ERROR_LOGs into
PanicAlerts.


JMC47:
Fixes Medal of Honor Vanguard not finding save files.

leoetlino:
Done

PR #8751, opened 2020-04-18
WiimoteEmu: Apply "Tilt" and "Point" rotations separately for proper tilted pointing.

jordan-woyak:
This should fix: https://bugs.dolphin-emu.org/issues/12058
Made the same changes to emulated Nunchuk to fix a probably rarely used tilted swinging.


Miksel12:
Pointing and tilting at the same time works perfectly now.

PR #8752, opened 2020-04-19
Remove unnecessary EFB to Texture disabling for Spider-Man 2

TheRealPSV:
Removed unnecessary EFBToTextureEnable=False for Spider-Man 2 from GK2.ini, as the game no longer requires it
Also removed empty config blocks

I've been testing with EFB to Texture enabled extensively for some time, and the game works just fine.

### DIFF
--- a/Data/Sys/GameSettings/GK2.ini
+++ b/Data/Sys/GameSettings/GK2.ini
@@ -2,10 +2,3 @@
 
 [Core]
 MMU = 1
-
-[OnFrame]
-
-[ActionReplay]
-
-[Video_Hacks]
-EFBToTextureEnable = False


JosJuice:
@dolphin-emu-bot rebuild

JMC47:
In older builds, the game had a lot of graphical errors at night-time with EFB Copies to Texture Only enabled.  I'm unsure of what could have fixed it, though.

TheRealPSV:
I'm not sure either, that's strange. I've actually been forcing `EFBToTextureEnabled=True` in my user-specific ini for it for over a year, and I've never seen any issues come out of it. I wonder if it could be related to the move to VideoCommon, or something related to the old depth issue.

JMC47:
I can confirm the graphical glitches are gone.  If someone else is okay with the empty blocks being removed, then this is fine.  It's a pretty big performance boost.

JosJuice:
Could it be paletted EFB copies? Those didn't work with EFB to texture back when this game INI was created.

Removing the empty blocks is fine.

TheRealPSV:
Great! Are you comfortable merging it in then?

PR #8755, opened 2020-04-21
Jit64: More addx and subfx optimizations

Sintendo:
Similar to #8551. We start by deduplicating some of the existing logic so the optimizations don't have to be repeated. We then add special handling for constant 0 in several places, carefully swap operands around if it can get us shorter instructions, and use LEA for subfx too.

---
**addx - Emit nothing when possible**

Before:
```
83 C7 00             add         edi,0
```
After:
```
```
---
**addx - Emit MOV when possible**

Before:
```
41 8D 7D 00          lea         edi,[r13]
```
After:
```
41 8B FD             mov         edi,r13d
```
---
**addx - Prefer smaller MOV+ADD sequence**

Before:
```
41 BE 40 00 00 00    mov         r14d,40h
44 03 75 A8          add         r14d,dword ptr [rbp-58h]
```
After:
```
44 8B 75 A8          mov         r14d,dword ptr [rbp-58h]
41 83 C6 40          add         r14d,40h
```

Before:
```
44 8B 7D F8          mov         r15d,dword ptr [rbp-8]
41 81 C7 00 68 00 CC add         r15d,0CC006800h
```
After:
```
41 BF 00 68 00 CC    mov         r15d,0CC006800h
44 03 7D F8          add         r15d,dword ptr [rbp-8]
```
---
**addx - Skip ADD after MOV when possible**

Before:
```
8B 7D F8             mov         edi,dword ptr [rbp-8]
83 C7 00             add         edi,0
```
After:
```
8B 7D F8             mov         edi,dword ptr [rbp-8]
```
---
**subfx - Use LEA when possible**

Before:
```
45 8B EE             mov         r13d,r14d
41 83 ED 08          sub         r13d,8
```
After:
```
45 8D 6E F8          lea         r13d,[r14-8]
```

### DIFF
--- a/Source/Core/Core/PowerPC/Jit64/Jit_Integer.cpp
+++ b/Source/Core/Core/PowerPC/Jit64/Jit_Integer.cpp
@@ -1346,7 +1346,15 @@ void Jit64::addx(UGeckoInstruction inst)
     {
       RCOpArg& Rimm = Ra.IsImm() ? Ra : Rb;
       RCOpArg& Rreg = Ra.IsImm() ? Rb : Ra;
-      LEA(32, Rd, MDisp(Rreg.GetSimpleReg(), Rimm.SImm32()));
+
+      if (Rimm.IsZero())
+      {
+        MOV(32, Rd, Rreg);
+      }
+      else
+      {
+        LEA(32, Rd, MDisp(Rreg.GetSimpleReg(), Rimm.SImm32()));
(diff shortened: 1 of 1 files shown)


BhaaLseN:
Const-propagation perhaps? The code seems fine, but I can't really think of any reasons why they would use an immediate valuee of zero either.

PR #8756, opened 2020-04-22
[WIP] Fix network hanging issues

sepalani:
This PR should truly fix: https://dolp.in/i11867

What happened (assuming there is no timing issues):
  1. The receive thread is on a blocking `RecvFrom` call
  2. The game calls `SOCleanup` which:
       * fires `IOCTL_NWC24_CLEANUP_SOCKET`
       * calls `IOS_Close` on the network handle
       * sets `soState = 1` (`soState = 2` before this)
  3. The game calls `SOClose` but `ioctlv` isn't fired because `soState = 1`

In other words, `RecvFrom` won't get a response and will wait forever (since Dolphin doesn't receive `IOCTL_SO_CLOSE`), again... 

It might require a hardware test to confirm that's the intended behaviour.

Ready to be reviewed.

### DIFF
--- a/Source/Core/Core/ConfigManager.cpp
+++ b/Source/Core/Core/ConfigManager.cpp
@@ -301,6 +301,7 @@ void SConfig::SaveNetworkSettings(IniFile& ini)
   network->Set("SSLVerifyCertificates", m_SSLVerifyCert);
   network->Set("SSLDumpRootCA", m_SSLDumpRootCA);
   network->Set("SSLDumpPeerCert", m_SSLDumpPeerCert);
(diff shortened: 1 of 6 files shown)


sepalani:
I added a timeout on blocking connect call based on this hardware test. Beware that it tries to connect to an ubisoft server (which timeout after ~3 minutes on hardware).

[wii-connect-timeout.zip](https://github.com/dolphin-emu/dolphin/files/4516755/wii-connect-timeout.zip)

I'll need to reverse IOS to confirm that behaviour, however.

sepalani:
Since all these commits will require appropriate hwtest, I'll split them in different PRs.

PR #8759, opened 2020-04-22
IOS/NET: Add timeout on blocking connect

sepalani:
This PR adds a timeout on connect for blocking sockets based on this hardware test. Beware that it tries to connect to an ubisoft server (which timeout after ~3 minutes on hardware).

[wii-connect-timeout.zip](https://github.com/dolphin-emu/dolphin/files/4516755/wii-connect-timeout.zip)

I'll need to reverse IOS to confirm that behaviour, however. The timeout value (in second) can also be changed in the config file under the `[Network]` section.

### DIFF
--- a/Source/Core/Core/Config/MainSettings.cpp
+++ b/Source/Core/Core/Config/MainSettings.cpp
@@ -157,6 +157,10 @@ const Info<bool> MAIN_NETWORK_SSL_DUMP_ROOT_CA{{System::Main, "Network", "SSLDum
 const Info<bool> MAIN_NETWORK_SSL_DUMP_PEER_CERT{{System::Main, "Network", "SSLDumpPeerCert"},
                                                  false};
 const Info<bool> MAIN_NETWORK_DUMP_AS_PCAP{{System::Main, "Network", "DumpAsPCAP"}, false};
+// Default value based on:
+//  - [RFC 1122] 4.2.3.5 TCP Connection Failures (at least 3 minutes)
+//  - https://dolp.in/pr8759 hwtest (3 minutes and 10 seconds)
+const Info<int> MAIN_NETWORK_TIMEOUT{{System::Main, "Network", "NetworkTimeout"}, 190};
 
 // Main.Interface
 
--- a/Source/Core/Core/IOS/Network/Socket.cpp
+++ b/Source/Core/Core/IOS/Network/Socket.cpp
@@ -331,11 +331,24 @@ void WiiSocket::Update(bool read, bool write, bool except)
       }
 
       // Fix blocking error codes
-      if (!nonBlock)
+      if (!nonBlock && it->net_type == IOCTL_SO_CONNECT)
       {
-        if (it->net_type == IOCTL_SO_CONNECT && ReturnValue == -SO_EISCONN)
+        switch (ReturnValue)
         {
+        case -SO_EAGAIN:
+        case -SO_EALREADY:
+        case -SO_EINPROGRESS:
+          if (std::chrono::steady_clock::now() > GetTimeout())
+          {
+            ReturnValue = -SO_ENETUNREACH;
+            ResetTimeout();
+          }
+          break;
+        case -SO_EISCONN:
           ReturnValue = SO_SUCCESS;
+          [[fallthrough]];
+        default:
+          ResetTimeout();
         }
       }
     }
@@ -658,6 +671,21 @@ void WiiSocket::Update(bool read, bool write, bool except)
       ++it;
     }
   }
+}
+
+const WiiSocket::Timeout& WiiSocket::GetTimeout()
+{
+  if (!timeout.has_value())
+  {
+    timeout = std::chrono::steady_clock::now() +
+              std::chrono::seconds(Config::Get(Config::MAIN_NETWORK_TIMEOUT));
+  }
+  return *timeout;
+}
+
+void WiiSocket::ResetTimeout()
+{
+  timeout.reset();
 }
 
 void WiiSocket::DoSock(Request request, NET_IOCTL type)
--- a/Source/Core/Core/IOS/Network/Socket.h
+++ b/Source/Core/Core/IOS/Network/Socket.h
@@ -43,8 +43,10 @@ typedef struct pollfd pollfd_t;
 #endif
 
 #include <algorithm>
+#include <chrono>
 #include <cstdio>
 #include <list>
+#include <optional>
 #include <string>
 #include <string_view>
 #include <unordered_map>
@@ -184,6 +186,7 @@ public:
   WiiSocket& operator=(WiiSocket&&) = default;
 
 private:
+  using Timeout = std::chrono::time_point<std::chrono::steady_clock>;
   struct sockop
   {
     Request request;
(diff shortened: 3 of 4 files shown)


Source/Core/Core/ConfigManager.cpp
```diff
@@ -301,6 +301,7 @@ void SConfig::SaveNetworkSettings(IniFile& ini)
   network->Set("SSLVerifyCertificates", m_SSLVerifyCert);
   network->Set("SSLDumpRootCA", m_SSLDumpRootCA);
   network->Set("SSLDumpPeerCert", m_SSLDumpPeerCert);
+  network->Set("NetworkTimeout", m_NetworkTimeout);
```
leoetlino, line 304: Could you use the new config system for this please? Migration has been slow, and adding new settings to SConfig won't help...

(Add a ConfigInfo to Core/Config/MainSettings.h/cpp, then update Core/ConfigLoaders/IsSettingSaveable.cpp)

Source/Core/Core/ConfigManager.cpp
```diff
@@ -592,6 +593,7 @@ void SConfig::LoadNetworkSettings(IniFile& ini)
   network->Get("SSLVerifyCertificates", &m_SSLVerifyCert, true);
   network->Get("SSLDumpRootCA", &m_SSLDumpRootCA, false);
   network->Get("SSLDumpPeerCert", &m_SSLDumpPeerCert, false);
+  network->Get("NetworkTimeout", &m_NetworkTimeout, 190);
```
leoetlino, line 596: It might be worth documenting that this value comes from a hwtest? Otherwise 190 just looks like a random timeout value.

Source/Core/Core/IOS/Network/Socket.cpp
```diff
@@ -601,6 +614,22 @@ void WiiSocket::Update(bool read, bool write, bool except)
   }
 }
 
+bool WiiSocket::HasTimeout()
```
leoetlino, line 617: The naming of this function bothers me a little bit, as the function doesn't just check whether something has timed out, but also starts a timer.

Source/Core/Core/IOS/Network/Socket.h
```diff
@@ -197,14 +199,21 @@ class WiiSocket
...
   void DoSock(Request request, SSL_IOCTL type);
   void Update(bool read, bool write, bool except);
   bool IsValid() const { return fd >= 0; }
+
   s32 fd = -1;
   s32 wii_fd = -1;
   bool nonBlock = false;
   std::list<sockop> pending_sockops;
+
+  using timeout_t = std::chrono::time_point<std::chrono::system_clock>;
```
leoetlino, line 215: I don't think introducing this type alias is necessary as it doesn't seem to be used. Also the _t suffix is reserved for POSIX stuff so I would avoid using it.

sepalani:
@leoetlino 
Done. I'm still open for a better name regarding the `DoTimeout()` method.

leoetlino:
StartOrCheckTimeout maybe? Naming is hard...

iwubcode:
I'm not great at naming either.  I can only offer my thoughts.  For boolean functions, I usually use 'is' or 'has', so maybe `HasTimedOut()` ?

Source/Core/Core/Config/MainSettings.cpp
```diff
@@ -143,4 +143,9 @@ const ConfigInfo<std::string> MAIN_RESOURCEPACK_PATH{{System::Main, "General", "
                                                      ""};
 const ConfigInfo<std::string> MAIN_FS_PATH{{System::Main, "General", "NANDRootPath"}, ""};
 const ConfigInfo<std::string> MAIN_SD_PATH{{System::Main, "General", "WiiSDCardPath"}, ""};
+
+// Main.Network
+
+const ConfigInfo<int> MAIN_NETWORK_TIMEOUT{{System::Main, "Network", "NetworkTimeout"},
+                                           190};  // Based on https://dolp.in/pr8759 hwtest
```
lioncash, line 150: You may wish to relocate this comment on top of the definition so that the wonky formatting can be avoided.

The comment can just be formatted to indicate it means the default value.

```cpp
// Default value based on https://dolp.in/pr8759 hwtest
```

sepalani:
@lioncash 
Done.

@iwubcode 
As @leoetlino said, the initial name was `HasTimeout` but the function doesn't only check for a boolean value but start the timer on-the-fly if needed and then check if a timeout occurred.

sepalani:
Rebased on latest master, I also added a reference to an RFC regarding the TCP connection failures.

Source/Core/Core/Config/MainSettings.cpp
```diff
@@ -156,6 +156,10 @@ const Info<bool> MAIN_NETWORK_SSL_VERIFY_CERTIFICATES{
 const Info<bool> MAIN_NETWORK_SSL_DUMP_ROOT_CA{{System::Main, "Network", "SSLDumpRootCA"}, false};
 const Info<bool> MAIN_NETWORK_SSL_DUMP_PEER_CERT{{System::Main, "Network", "SSLDumpPeerCert"},
                                                  false};
+// Default value based on:
+//  - [RFC 1122] 4.2.3.5 TCP Connection Failures (3 minutes)
+//  - https://dolp.in/pr8759 hwtest
```
leoetlino, line 161: Maybe add "(3 minutes and 10 seconds)"? Right now the 190 looks like a typo :D

Source/Core/Core/IOS/Network/Socket.cpp
```diff
@@ -331,11 +331,24 @@ void WiiSocket::Update(bool read, bool write, bool except)
...
-      if (!nonBlock)
+      if (!nonBlock && it->net_type == IOCTL_SO_CONNECT)
       {
-        if (it->net_type == IOCTL_SO_CONNECT && ReturnValue == -SO_EISCONN)
+        switch (ReturnValue)
         {
+        case -SO_EAGAIN:
+        case -SO_EALREADY:
+        case -SO_EINPROGRESS:
+          if (DoTimeout())
```
leoetlino, line 341: To me, DoTimeout makes it sound like you're performing the action of timing out something... Maybe this should just be split into two functions:

```cpp
void StartTimeoutIfNeeded()
{
  if (!timeout.has_value())
  {
    timeout = std::chrono::system_clock::now() +
              std::chrono::seconds(Config::Get(Config::MAIN_NETWORK_TIMEOUT));
  }
}

bool HasTimedOut() const { return timeout && std::chrono::system_clock::now() > *timeout; }
```

sepalani:
@leoetlino 
Done. Rather than splitting it into 2 functions, I decided to create simple `GetTimeout` function and perform the check inside the `if` statement.

Source/Core/Core/IOS/Network/Socket.h
```diff
@@ -184,6 +186,7 @@ class WiiSocket
   WiiSocket& operator=(WiiSocket&&) = default;
 
 private:
+  using Timeout = std::chrono::time_point<std::chrono::system_clock>;
```
leoetlino, line 189: Sorry, I missed something important in my previous review: this should be using a steady_clock, not a system_clock. The latter isn't monotonic and can be changed at any time by the user.

sepalani: Good catch, I completely forgot about that.

I also updated the RFC comment since it's recommending this timeout to be **at least** 3 minutes.

leoetlino (approved):

PR #8760, opened 2020-04-23
DiscIO: Implement multithreaded compression

JosJuice:
Split out from PR #8538. This PR makes the deflate compression of GCZ happen on multiple threads. Multithreading is more important for WIA than for GCZ because LZMA and bzip2 are much slower than deflate, but I thought this would make sense as its own PR anyway.

### DIFF
--- a/Source/Core/DiscIO/CompressedBlob.cpp
+++ b/Source/Core/DiscIO/CompressedBlob.cpp
@@ -28,6 +28,7 @@
 #include "DiscIO/Blob.h"
 #include "DiscIO/CompressedBlob.h"
 #include "DiscIO/DiscScrubber.h"
+#include "DiscIO/MultithreadedCompressor.h"
 #include "DiscIO/Volume.h"
 
 namespace DiscIO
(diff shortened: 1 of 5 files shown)


Source/Core/DiscIO/CompressedBlob.cpp
```diff
@@ -154,6 +155,119 @@ bool CompressedBlobReader::GetBlock(u64 block_num, u8* out_ptr)
...
+static ConversionResult<OutputParameters> Compress(CompressThreadState* state,
+                                                   CompressParameters parameters, int block_size,
+                                                   std::vector<u32>* hashes, int* num_stored,
+                                                   int* num_compressed)
+{
+  state->compressed_buffer.resize(block_size);
+
+  int retval = deflateReset(&state->z);
+  state->z.next_in = parameters.data.data();
+  state->z.avail_in = block_size;
```
degasus, line 203: What do you think about using parameters.data.size() instead of block_size?

JosJuice: It results in a warning (converting `size_t` to `unsigned int`).

PR #8765, opened 2020-04-26
Jit64 codegen space reuse.

AdmiralCurtiss:
This makes some minor alterations to the Jit64 codegen so it keeps track of free memory in the near and far code cache and reuses it when possible. This is to avoid full cache clears, which are costly due to having to recompile everything after the clear.

To do this, I've done the following things:
- The x64Emitter now knows the end of the memory region it's allowed to write into, and checks this. If a write is requested past the end, it does not actually write and instead sets a failure flag.
- `Jit64::Jit()` checks this flag before finalizing a Jit block. If it is set, it invokes a cache clear and retries. This replaces the 'early' cache clear when we're almost out of space, since we now detect when we actually are.
- Every JIT block now knows its own code area in the near and far code. When a JIT block is invalidated, these areas are marked as free-to-use for the code generation again.
- Before every code generation, we find the largest free near and far code blocks, and generate at that place.

And that's about it, fairly simple really. I'm somewhat surprised this worked out as well as it did, to be honest. That said,

Known issues:
- ~~Performance in the code emitter is probably slightly worse due to the bounds checking. We should test if this is actually noticeable in practice. If so, we can try to write a variant of this without the bounds checking, which I think is possible but definitely feels sketchier.~~ Although no large-scale testing has been done, this seems to not actually be a problem in practice according to JMC47's and my testing.
- It's entirely possible, and probably even likely, that this runs into severe memory fragmentation over time. I don't think this is fixable without making code relocatable. As such, this will still eventually invoke a cache clear due to lack of a singular large block of code space.

### DIFF
--- a/Externals/rangeset/include/rangeset/rangeset.h
+++ b/Externals/rangeset/include/rangeset/rangeset.h
@@ -1,0 +1,368 @@
+#pragma once
+
+#include <cassert>
+#include <map>
+
+namespace HyoutaUtilities {
+template <typename T> class RangeSet {
+private:
+  using MapT = std::map<T, T>;
+
+public:
+  struct const_iterator {
+  public:
+    const T& from() const {
+      return It->first;
+    }
+
+    const T& to() const {
+      return It->second;
+    }
+
+    const_iterator& operator++() {
+      ++It;
+      return *this;
+    }
+
+    const_iterator operator++(int) {
+      const_iterator old = *this;
+      ++It;
+      return old;
+    }
+
+    const_iterator& operator--() {
+      --It;
+      return *this;
+    }
+
+    const_iterator operator--(int) {
+      const_iterator old = *this;
+      --It;
+      return old;
+    }
+
+    bool operator==(const const_iterator& rhs) const {
+      return this->It == rhs.It;
+    }
+
+    bool operator!=(const const_iterator& rhs) const {
+      return !operator==(rhs);
+    }
+
+  private:
+    typename MapT::const_iterator It;
+    const_iterator(typename MapT::const_iterator it) : It(it) {}
+    friend class RangeSet;
+  };
+
+  void insert(T from, T to) {
+    if (from >= to)
+      return;
+
+    // Start by finding the closest range.
+    // upper_bound() returns the closest range whose starting position
+    // is greater than 'from'.
+    auto bound = Map.upper_bound(from);
+    if (bound == Map.end()) {
+      // There is no range that starts greater than the given one.
+      // This means we have three options:
+      // - 1. No range exists yet, this is the first range.
+      if (Map.empty()) {
+        insert_range(from, to);
+        return;
+      }
+
+      // - 2. The given range does not overlap the last range.
+      --bound;
+      if (from > get_to(bound)) {
+        insert_range(from, to);
+        return;
+      }
+
+      // - 3. The given range does overlap the last range.
+      maybe_expand_to(bound, to);
+      return;
+    }
+
+    if (bound == Map.begin()) {
+      // The given range starts before any of the existing ones.
+      // We must insert this as a new range even if we potentially overlap
+      // an existing one as we can't modify a key in a std::map.
+      auto inserted = insert_range(from, to);
+      merge_from_iterator_to_value(inserted, bound, to);
+      return;
+    }
+
+    auto abound = bound--;
+
+    // 'bound' now points at the first range in the map that
+    // could possibly be affected.
+
+    // If 'bound' overlaps with given range, update bounds object.
+    if (get_to(bound) >= from) {
+      maybe_expand_to(bound, to);
+      auto inserted = bound;
+      ++bound;
+      merge_from_iterator_to_value(inserted, bound, to);
+      return;
+    }
+
+    // 'bound' *doesn't* overlap with given range, check next range.
+
+    // If this range overlaps with given range,
+    if (get_from(abound) <= to) {
+      // insert new range
+      auto inserted = insert_range(from, to >= get_to(abound) ? to : get_to(abound));
+      // and delete overlaps
+      abound = erase_range(abound);
+      merge_from_iterator_to_value(inserted, abound, to);
+      return;
+    }
+
+    // Otherwise, if we come here, then this new range overlaps nothing
+    // and must be inserted as a new range.
+    insert_range(from, to);
+  }
+
+  void erase(T from, T to) {
+    if (from >= to)
+      return;
+
+    // Like insert(), we use upper_bound to find the closest range.
+    auto bound = Map.upper_bound(from);
+    if (bound == Map.end()) {
+      // There is no range that starts greater than the given one.
+      if (Map.empty()) {
+        // nothing to do
+        return;
+      }
+      --bound;
+      // 'bound' now points at the last range.
+      if (from >= get_to(bound)) {
+        // Given range is larger than any range that exists, nothing to do.
+        return;
+      }
+
+      if (to >= get_to(bound)) {
+        if (from == get_from(bound)) {
+          // Given range fully overlaps last range, erase it.
+          erase_range(bound);
+          return;
+        } else {
+          // Given range overlaps end of last range, reduce it.
+          reduce_to(bound, from);
+          return;
+        }
+      }
+
+      if (from == get_from(bound)) {
+        // Given range overlaps begin of last range, reduce it.
+        reduce_from(bound, to);
+        return;
+      } else {
+        // Given range overlaps middle of last range, bisect it.
+        bisect_range(bound, from, to);
+        return;
+      }
+    }
+
+    if (bound == Map.begin()) {
+      // If we found the first range that means 'from' is before any stored range.
+      // This means we can just erase from start until 'to' and be done with it.
+      erase_from_iterator_to_value(bound, to);
+      return;
+    }
+
+    // check previous range
+    auto abound = bound--;
+
+    if (from == get_from(bound)) {
+      // Similarly, if the previous range starts with the given one, just erase until 'to'.
+      erase_from_iterator_to_value(bound, to);
+      return;
+    }
+
+    // If we come here, the given range may or may not overlap part of the current 'bound'
+    // (but never the full range), which means we may need to update the end position of it,
+    // or possibly even split it into two.
+    if (from < get_to(bound)) {
+      if (to < get_to(bound)) {
+        // need to split in two
+        bisect_range(bound, from, to);
+        return;
+      } else {
+        // just update end
+        reduce_to(bound, from);
+      }
+    }
+
+    // and then just erase until 'to'
+    erase_from_iterator_to_value(abound, to);
+    return;
+  }
+
+  const_iterator erase(const_iterator it) {
+    return const_iterator(erase_range(it.It));
+  }
+
+  void clear() {
+    Map.clear();
+  }
+
+  bool contains(T value) const {
+    auto it = Map.upper_bound(value);
+    if (it == Map.begin())
+      return false;
+    --it;
+    return get_from(it) <= value && value < get_to(it);
+  }
+
+  size_t size() const {
+    return Map.size();
+  }
+
+  bool empty() const {
+    return Map.empty();
+  }
+
+  void swap(RangeSet<T>& other) {
+    Map.swap(other.Map);
+  }
+
+  const_iterator begin() const {
+    return const_iterator(Map.begin());
+  }
+
+  const_iterator end() const {
+    return const_iterator(Map.end());
+  }
+
+  const_iterator cbegin() const {
+    return const_iterator(Map.cbegin());
+  }
+
+  const_iterator cend() const {
+    return const_iterator(Map.cend());
+  }
+
+  bool operator==(const RangeSet<T>& other) const {
+    return this->Map == other.Map;
+  }
+
+  bool operator!=(const RangeSet<T>& other) const {
+    return !(*this == other);
+  }
+
+private:
+  // Assumptions that can be made about the data:
+  // - Range are stored in the form [from, to[
+  //   That is, the starting value is inclusive, and the end value is exclusive.
+  // - 'from' is the map key, 'to' is the map value
+  // - 'from' is always smaller than 'to'
+  // - Stored ranges never touch.
+  // - Stored ranges never overlap.
+  MapT Map;
+
+  T get_from(typename MapT::iterator it) const {
+    return it->first;
+  }
+
+  T get_to(typename MapT::iterator it) const {
+    return it->second;
+  }
+
+  T get_from(typename MapT::const_iterator it) const {
+    return it->first;
+  }
+
+  T get_to(typename MapT::const_iterator it) const {
+    return it->second;
+  }
+
+  typename MapT::iterator insert_range(T from, T to) {
+    return Map.emplace(from, to).first;
+  }
+
+  typename MapT::iterator erase_range(typename MapT::iterator it) {
+    return Map.erase(it);
+  }
+
+  typename MapT::const_iterator erase_range(typename MapT::const_iterator it) {
+    return Map.erase(it);
+  }
+
+  void bisect_range(typename MapT::iterator it, T from, T to) {
+    assert(get_from(it) < from);
+    assert(get_from(it) < to);
+    assert(get_to(it) > from);
+    assert(get_to(it) > to);
+    assert(from < to);
+    T itto = get_to(it);
+    reduce_to(it, from);
+    insert_range(to, itto);
+  }
+
+  typename MapT::iterator reduce_from(typename MapT::iterator it, T from) {
+    assert(get_from(it) < from);
+    T itto = get_to(it);
+    erase_range(it);
+    return insert_range(from, itto);
+  }
+
+  void maybe_expand_to(typename MapT::iterator it, T to) {
+    if (to <= get_to(it))
+      return;
+
+    expand_to(it, to);
+  }
+
+  void expand_to(typename MapT::iterator it, T to) {
+    assert(get_to(it) < to);
+    it->second = to;
+  }
+
+  void reduce_to(typename MapT::iterator it, T to) {
+    assert(get_to(it) > to);
+    it->second = to;
+  }
+
+  void merge_from_iterator_to_value(typename MapT::iterator inserted, typename MapT::iterator bound, T to) {
+    // Erase all ranges that overlap the inserted while updating the upper end.
+    while (bound != Map.end() && get_from(bound) <= to) {
+      maybe_expand_to(inserted, get_to(bound));
+      bound = erase_range(bound);
+    }
+  }
+
+  void erase_from_iterator_to_value(typename MapT::iterator bound, T to) {
+    // Assumption: Given bound starts at or after the 'from' value of the range to erase.
+    while (true) {
+      // Given range starts before stored range.
+      if (to <= get_from(bound)) {
+        // Range ends before this range too, nothing to do.
+        return;
+      }
+
+      if (to < get_to(bound)) {
+        // Range ends in the middle of current range, reduce current.
+        reduce_from(bound, to);
+        return;
+      }
+
+      if (to == get_to(bound)) {
+        // Range ends exactly with current range, erase current.
+        erase_range(bound);
+        return;
+      }
+
+      // Range ends later than current range.
+      // First erase current, then loop to check the range(s) after this one too.
+      bound = erase_range(bound);
+      if (bound == Map.end()) {
+        // Unless that was the last range, in which case there's nothing else to do.
+        return;
+      }
+    }
+  }
+};
+} // namespace HyoutaUtilities
--- a/Source/Core/Common/x64Emitter.h
+++ b/Source/Core/Common/x64Emitter.h
@@ -329,8 +329,18 @@ class XEmitter
 {
   friend struct OpArg;  // for Write8 etc
 private:
+  // Pointer to memory where code will be emitted to.
   u8* code = nullptr;
+
+  // Pointer past the end of the memory region we're allowed to emit to.
+  // Writes that would reach this memory are refused and will set the m_write_failed flag instead.
+  u8* m_code_end = nullptr;
+
   bool flags_locked = false;
+
+  // Set to true when a write request happens that would write past m_code_end.
+  // Must be cleared with SetCodePtr() afterwards.
+  bool m_write_failed = false;
 
   void CheckFlags();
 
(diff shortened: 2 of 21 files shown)


JMC47:
So, after some more exclusive testing in a few games, this does improve the situation a bit.  It isn't a complete fix, but it does make code flushing a lot rarer in games like True Crime: New York and N64 VC games.

However, I haven't done a great deal of general performance testing (this is probably slower?) and I think it'd be worth getting more people to do tests than just me, so we can see how it impacts more games.

AdmiralCurtiss:
Fully agreed. I'd recommend waiting on performance tests though, since at the very least the current `rangeset` needs to be changed, which could and probably will impact that.

Source/Core/Core/PowerPC/JitCommon/JitBase.h
```diff
@@ -17,6 +17,8 @@
 #include "Core/PowerPC/JitCommon/JitCache.h"
 #include "Core/PowerPC/PPCAnalyst.h"
 
+#include "../Jit64/rangeset.h"
```
lioncash, line 20: 64-bit specifics shouldn't be included in the architecture common areas. This needs to be modified. Just leaving a comment here to make sure it's not forgotten about.

Source/Core/Core/PowerPC/Jit64/rangeset.h
```diff
@@ -0,0 +1,986 @@
+ /*
```
lioncash, line 1: This should likely be moved into externals. If the header is to remain within our main codebase, then it should be complying with our coding style (which it currently does not)

Source/Core/Core/PowerPC/Jit64/rangeset.h
```diff
@@ -0,0 +1,986 @@
...
+  *  \author Martin Reinecke
+  */
+ 
+ #ifndef PLANCK_RANGESET_H
+ #define PLANCK_RANGESET_H
+ 
+ #include <algorithm>
+ #include <vector>
+ #include <utility>
+ #include <iostream>
```
lioncash, line 38: This should be swapped out with `<ostream>` or any other stream header. `<iostream>` injects a static constructor into every translation unit that includes it, so this has a runtime cost that can be avoided.

Source/Core/Core/PowerPC/JitCommon/JitBase.h
```diff
@@ -39,6 +41,48 @@
 #define JITDISABLE(setting)                                                                        \
   FALLBACK_IF(SConfig::GetInstance().bJITOff || SConfig::GetInstance().setting)
 
+struct RangeTracker
+{
+  rangeset<size_t> ranges;
+
+  void add(const u8* begin, const u8* end) { ranges.add((size_t)(begin), (size_t)(end)); }
```
lioncash, line 48: These function names don't obey our codebase coding style.

Also, C-style casts should be avoided and replaced with static_cast or a functional style cast.

Source/Core/Core/PowerPC/JitCommon/JitBase.h
```diff
@@ -39,6 +41,48 @@
...
+    ranges.clear();
+    add(begin, end);
+  }
+  std::optional<std::pair<u8*, u8*>> find_at_least(size_t size)
+  {
+    for (size_t i = 0; i < ranges.size(); ++i)
+    {
+      if (ranges.ivlen(i) >= size)
+      {
+        return std::make_pair((u8*)ranges.ivbegin(i), (u8*)ranges.ivend(i));
```
lioncash, line 61: These should be either explicit static_casts or reinterpret_casts since they're pointers.

Source/Core/Core/PowerPC/JitCommon/JitBase.h
```diff
@@ -39,6 +41,48 @@
...
+      size_t sz = ranges.ivlen(i);
+      if (sz > biggest_size)
+      {
+        biggest_size = sz;
+        biggest_index = i;
+      }
+    }
+    if (biggest_index == std::numeric_limits<size_t>::max())
+      return std::nullopt;
+    return std::make_pair((u8*)ranges.ivbegin(biggest_index), (u8*)ranges.ivend(biggest_index));
```
lioncash, line 82: Ditto

lioncash (commented):
Left a very cursory review

AdmiralCurtiss:
Alright, so I've replaced the rangeset with a custom one that keeps track of the largest free range to minimize the time spent searching for the largest free block -- see https://github.com/AdmiralCurtiss/rangeset/blob/master/rangesizeset.h

This still needs a bit of refactoring: Should probably put the rangeset into externals, need to make sure the ARM Jit still builds, also I'm not really happy with `m_free_ranges_near` being in the JitBase when only the 64bit one uses it. But in general I think this is now about where I want it to be.

If someone wants to do some performance testing, they can do so now.

autofire372:
Tested a few N64 VC titles, all NTSC-U.

Super Mario 64 - 120 FPS ingame in master/no change in PR

Kirby 64 - 137 FPS ingame in master/no change in PR

Super Smash Bros. - crashes Dolphin with this PR.

Source/Core/Common/RangeSizeSet.h
```diff
@@ -0,0 +1,458 @@
...
+// Like RangeSet, but additionally stores a map of the ranges sorted by their size, for quickly
+// finding the largest or smallest range.
+template <typename T>
+class RangeSizeSet
+{
+private:
+  struct Value;
+
+public:
+  // TODO: this should be an unsigned type when T is signed
```
lioncash, line 21: Just leaving this here so it's not forgotten about before merging, but this should be resolved.

Source/Core/Common/RangeSizeSet.h
```diff
@@ -0,0 +1,458 @@
...
+  // can't copy this class cause we store iterators
+  RangeSizeSet() = default;
+  RangeSizeSet(const RangeSizeSet<T>&) = delete;
+  RangeSizeSet(RangeSizeSet<T>&&) = default;
+  RangeSizeSet<T>& operator=(const RangeSizeSet<T>&) = delete;
+  RangeSizeSet<T>& operator=(RangeSizeSet<T>&&) = default;
+
+  void insert(T from, T to)
+  {
+    if (!(from < to))
```
lioncash, line 124:
```suggestion
    if (from >= to)
```

Source/Core/Common/RangeSizeSet.h
```diff
@@ -0,0 +1,458 @@
...
+    }
+
+    // if we come here then this new range overlaps nothing and must be inserted
+    // as a new range
+    insert_range(from, to);
+  }
+
+  void erase(T from, T to)
+  {
+    if (!(from < to))
```
lioncash, line 200: Ditto

Source/Core/Common/RangeSizeSet.h
```diff
@@ -0,0 +1,458 @@
...
+  // Key is the size of the range.
+  // Value is a pointer to the range in the regular range map.
+  // We use std::greater so that m_sizes.begin() gives us the largest range.
+  std::multimap<UT, typename std::map<T, Value>::iterator, std::greater<UT>> m_sizes;
+
+  T get_from(typename std::map<T, Value>::iterator it) { return it->first; }
+
+  T get_to(typename std::map<T, Value>::iterator it) { return it->second.m_to; }
+
+  typename std::map<T, Value>::iterator insert_range(T from, T to)
```
lioncash, line 354: May want to make a local using alias in the private section of the class to prevent needing to continually repeat this typename, etc.

Source/Core/Common/RangeSizeSet.h
```diff
@@ -0,0 +1,458 @@
...
+
+  void clear()
+  {
+    m_map.clear();
+    m_sizes.clear();
+  }
+
+  size_t size() const { return m_map.size(); }
+
+  const_iterator begin() const { return const_iterator(m_map.begin()); }
```
lioncash, line 299: this should also provide `cbegin()` and `cend()` overloads.

Source/Core/Common/RangeSizeSet.h
```diff
@@ -0,0 +1,458 @@
...
+  by_size_const_iterator by_size_begin() const { return by_size_const_iterator(m_sizes.begin()); }
+
+  by_size_const_iterator by_size_end() const { return by_size_const_iterator(m_sizes.end()); }
+
+  bool operator==(const RangeSizeSet<T>& other) const { return this->m_map == other.m_map; }
+
+  bool operator!=(const RangeSizeSet<T>& other) const { return !(*this == other); }
+
+private:
+  UT calc_size(T from, T to)
```
lioncash, line 312: This can be a static function

Source/Core/Common/RangeSizeSet.h
```diff
@@ -0,0 +1,458 @@
...
+  {
+    assert(get_from(it) < from);
+    T itto = get_to(it);
+    erase_range(it);
+    return insert_range(from, itto);
+  }
+
+  void maybe_expand_to(typename std::map<T, Value>::iterator it, T to)
+  {
+    if (to > get_to(it))
```
lioncash, line 390:
```suggestion
    if (to <= get_to(it))
      return;
```

Source/Core/Common/RangeSizeSet.h
```diff
@@ -0,0 +1,458 @@
...
+  }
+
+  // Value type stored in the regular range map.
+  struct Value
+  {
+    // End point of the range.
+    T m_to;
+
+    // Pointer to the same range in the by-size multimap.
+    typename std::multimap<UT, typename std::map<T, Value>::iterator, std::greater<UT>>::iterator
```
lioncash, line 325: May want to separate this out into local using aliases, so this is less jarring to read.

Source/Core/Common/RangeSizeSet.h
```diff
@@ -0,0 +1,458 @@
...
+
+  const_iterator end() const { return const_iterator(m_map.end()); }
+
+  by_size_const_iterator by_size_begin() const { return by_size_const_iterator(m_sizes.begin()); }
+
+  by_size_const_iterator by_size_end() const { return by_size_const_iterator(m_sizes.end()); }
+
+  bool operator==(const RangeSizeSet<T>& other) const { return this->m_map == other.m_map; }
+
+  bool operator!=(const RangeSizeSet<T>& other) const { return !(*this == other); }
```
lioncash, line 309:
```suggestion
  bool operator!=(const RangeSizeSet<T>& other) const { return !operator==(other); }
```

Source/Core/Common/RangeSizeSet.h
```diff
@@ -0,0 +1,458 @@
...
+    by_size_const_iterator operator--(int)
+    {
+      by_size_const_iterator old = *this;
+      --m_it;
+      return old;
+    }
+
+    bool operator==(const by_size_const_iterator& rhs) const { return this->m_it == rhs.m_it; }
+
+    bool operator!=(const by_size_const_iterator& rhs) const { return this->m_it != rhs.m_it; }
```
lioncash, line 102:
```suggestion
    bool operator!=(const by_size_const_iterator& rhs) const { return !operator==(rhs); }
```

Source/Core/Common/RangeSizeSet.h
```diff
@@ -0,0 +1,458 @@
...
+    const_iterator operator--(int)
+    {
+      const_iterator old = *this;
+      --m_it;
+      return old;
+    }
+
+    bool operator==(const const_iterator& rhs) const { return this->m_it == rhs.m_it; }
+
+    bool operator!=(const const_iterator& rhs) const { return this->m_it != rhs.m_it; }
```
lioncash, line 59:
```suggestion
    bool operator!=(const const_iterator& rhs) const { return !operator==(rhs); }
```

lioncash (commented):
What necessitates the x64 stuff leaking into the base class of the JIT, out of curiosity?

I also left some review comments on the range set code, though it may be moved over to the externals later.

AdmiralCurtiss:
re JitBase leak: It's been a week but I believe it was the additions in `JitBaseBlockCache::DestroyBlock()`. I guess we could make that virtual and override it in the Jit64's `JitBlockCache` instead?

AdmiralCurtiss:
Alright, I've moved the rangeset to externals (along with fixes, appreciate the comments!) and this now builds on Android/ARM, although I have no good way to actually test if it still runs there. I've also moved the code around a bit to remove the dependency in the JitBaseBlockCache. I think I'll open this for general review.

JMC47:
Super Smash Bros VC. works fine here, it may have the texture cache crash (at higher IRs, happens in master)

I tested: 
Rogue Squadron II: no flushes, but crashed due to fifo issues
True Crime New York: Less Flushes, but still runs like shit.
Metal Gear Solid: Twin Snakes: No Flushes but MMU still can't be re-enabled due to ISI exceptions causing slowdowns that aren't there when MMU is off.
Mario Kart 64: No flushes
The Legend of Zelda: Ocarina of Time (VC): Less Flushes
The Legend of Zelda: Ocarina of Time MQ:  Less Flushes
Metroid Prime: No Flushes, but I don't know how long I'm supposed to play.  I played for a half hour and got bored of just going through doors.

Performance wise, I did not notice any significant drop in overall performance.

Source/Core/Core/PowerPC/Jit64Common/BlockCache.h
```diff
@@ -13,6 +15,11 @@ class JitBlockCache : public JitBaseBlockCache
 public:
   explicit JitBlockCache(JitBase& jit);
 
+  void DestroyBlock(JitBlock& block) override;
+
+  std::vector<std::pair<u8*, u8*>> m_ranges_to_free_on_next_codegen_near;
```
lioncash, line 20: These should be private and instead have a class interface built around these.

Source/Core/Core/PowerPC/Jit64/Jit.h
```diff
@@ -18,6 +18,8 @@
 // ----------
 #pragma once
 
+#include <rangesizeset.h>
```
lioncash, line 21: The externals project should likely be structured so that the include is within an include folder with a CMake file defining a header only target in the externals folder.

e.g.

`include/rangesizeset/rangesizeset.h` (feel free to rename the first `rangesizeset` to a name of preferred choosing if you want).

```cmake
# Other pre-reqs before this

add_library(RangeSizeSet INTERFACE)
target_include_directories(RangeSizeSet INTERFACE include)
```

Then link it into core. This way the include would be:

`<rangesizeset/rangesizeset.h>`

Which significantly reduces the chances of this include ever clashing with anything that ends up being potentially added to externals in the future.

Source/Core/Common/x64Emitter.h
```diff
@@ -378,19 +380,24 @@ class XEmitter
...
   u8* AlignCodePage();
   const u8* GetCodePtr() const;
   u8* GetWritableCodePtr();
+  const u8* GetCodeEnd() const;
+  u8* GetWritableCodeEnd();
 
   void LockFlags() { flags_locked = true; }
   void UnlockFlags() { flags_locked = false; }
+
+  bool HasFailed() const { return write_failed; }
```
lioncash, line 399:
```suggestion
  bool HasWriteFailed() const { return write_failed; }
```

Source/Core/Common/x64Emitter.h
```diff
@@ -378,19 +380,24 @@ class XEmitter
 
 public:
   XEmitter() = default;
-  explicit XEmitter(u8* code_ptr) : code{code_ptr} {}
+  explicit XEmitter(u8* code_ptr, u8* code_end_ptr) : code(code_ptr), code_end(code_end_ptr) {}
   virtual ~XEmitter() = default;
-  void SetCodePtr(u8* ptr);
+  void SetCodePtr(u8* ptr, u8* end, bool failed = false);
```
lioncash, line 385:
```suggestion
  void SetCodePtr(u8* ptr, u8* end, bool has_write_failed = false);
```

So that it's explicit what's being set.

Source/Core/Core/PowerPC/Jit64/Jit.h
```diff
@@ -56,7 +58,8 @@ class Jit64 : public JitBase, public QuantizedMemoryRoutines
   // Jit!
 
   void Jit(u32 em_address) override;
-  u8* DoJit(u32 em_address, JitBlock* b, u32 nextPC);
+  bool SetEmitterState();
```
lioncash, line 61: This should likely have a comment describing what this actually does, since the name is quite vague.

Source/Core/Core/PowerPC/Jit64Common/BlockCache.cpp
```diff
@@ -21,7 +21,7 @@ void JitBlockCache::WriteLinkBlock(const JitBlock::LinkData& source, const JitBl
 
   u8* location = source.exitPtrs;
   const u8* address = dest ? dest->checkedEntry : dispatcher;
-  Gen::XEmitter emit(location);
+  Gen::XEmitter emit(location, location + 32 /* FIXME */);
```
lioncash, line 24: This still needs to be addressed. Especially given there's no explanation to others about what actually needs to be fixed.

AdmiralCurtiss: Completely forgot about that one, thanks.

Source/Core/Core/PowerPC/Jit64/Jit.h
AdmiralCurtiss: I assume this is fine now?

Source/Core/Core/PowerPC/Jit64Common/BlockCache.h
AdmiralCurtiss: Before I change this, I'm looking at this again I'm not 100% sure when to exactly clear those vectors. I would have just put the clear into `JitBlockCache::Clear()` but it looks like that can be called without the Jit64's `Clear()` being called, which could desync what the Jit64 thinks is free versus what actually is. Am I missing something here? Does it ever make sense to clear the block cache without clearing the entire Jit?

AdmiralCurtiss: Wait, I think I'm thinking about this in the wrong direction. A block cache clear should actually *add* all of the destroyed blocks to the ranges-to-free vectors, and then the next Jit() will pick that up correctly. I think that works, let's see...

Rumi-Larry:
The comments in the code lack capitalization and some have confusing grammar.

Rumi-Larry:
The comments in the code lack capitalization and some have confusing grammar.

AdmiralCurtiss:
While I'm not sure I agree about the comments being problematic as-is, I suppose it wouldn't hurt to do another pass on them, and maybe add a few more for details.

Externals/rangeset/include/rangeset/rangeset.h
```diff
@@ -0,0 +1,366 @@
...
+    friend class RangeSet;
+  };
+
+  void insert(T from, T to) {
+    if (from >= to)
+      return;
+
+    // find the closest range
+    // this will return the closest range whose starting position is greater
+    // than 'from'
```
Rumi-Larry, line 64:
```suggestion
    // Find the closest range:
    // This will return the closest range whose starting position is greater than 'from'
```

Externals/rangeset/include/rangeset/rangeset.h
```diff
@@ -0,0 +1,366 @@
...
+    auto bound = Map.upper_bound(from);
+    if (bound == Map.end()) {
+      // there is no range that starts greater than the given one
+      // this means we have three options:
+      // - no range exists yet, this is the first range
```
Rumi-Larry, line 69:
```suggestion
      // There is no range that starts greater than the given one,
      // this implies one of three cases:
      // 1. No range exists yet, this is the first range
```

Rumi-Larry (commented):
I just submitted two suggestions, so it's clearer what I mean.

delroth:
Please update Externals/licenses.md when adding a new external library.

Source/Core/Core/PowerPC/Jit64/Jit.cpp
```diff
@@ -785,12 +804,67 @@ void Jit64::Jit(u32 em_address)
...
+      blocks.FinalizeBlock(*b, jo.enableBlocklink, code_block.m_physical_addresses);
+      return;
+    }
+  }
+
+  // Code generation failed due to not enough free space in either the near or far code regions.
+  // Clear the entire JIT cache and retry.
+  WARN_LOG(POWERPC, "flushing code caches, please report if this happens a lot");
+  ClearCache();
+  Jit(em_address);
```
BhaaLseN, line 841: Nothing keeps us from running into a potential stack overflow here (by calling `Jit` over and over again after flushing); can this be a potential issue in practise? Or is it unlikely that we ever emit so much code for a block that it ever fills up everything available?

AdmiralCurtiss: Considering the JIT previously would just start writing into arbitrary memory if the JITed code was larger than the remaining buffer length and we've never -- at least knowingly -- ran into it, I'd say this is impossible in practice unless you designed PowerPC code to specifically exploit Dolphin's JIT. I guess a check is easy enough to add though, so might as well.

AdmiralCurtiss: I added a PanicAlert and exit(), is that okay?

BhaaLseN: Seems ok to me, thanks!

BhaaLseN (approved):
Code seems good for the most part, and the changes make sense to me. Untested though.

Source/Core/Core/PowerPC/Jit64Common/BlockCache.cpp
```diff
@@ -53,3 +53,35 @@ void JitBlockCache::WriteDestroyBlock(const JitBlock& block)
...
+}
+
+void JitBlockCache::DestroyBlock(JitBlock& block)
+{
+  JitBaseBlockCache::DestroyBlock(block);
+
+  if (block.near_begin != block.near_end)
+    m_ranges_to_free_on_next_codegen_near.emplace_back(block.near_begin, block.near_end);
+  if (block.far_begin != block.far_end)
+    m_ranges_to_free_on_next_codegen_far.emplace_back(block.far_begin, block.far_end);
```
SuperSamus, line 70: What's the reason why the rangeset modifications are delayed to `Jit64::Jit`, instead of done right now? Is it just because `JitBlockCache` doesn't have a reference to `Jit64`, or is it something else?

AdmiralCurtiss: It's been six years so who knows, but I assume because the invalidation can happen in the middle of a block and it didn't seem safe to change the block data while in the middle of one.

PR #8766, opened 2020-04-26
Don't defer EFB Copies to RAM - PMTTYD Issues

teconmoon:
This resolves a few issues with bounding box animations and others. Most noticably, it greatly reduces the bounding box slowdown seen on some NVIDIA cards and also fixes the odd overlay glitches when moving between rooms on the Excess Express.

### DIFF
--- a/Data/Sys/GameSettings/G8M.ini
+++ b/Data/Sys/GameSettings/G8M.ini
@@ -16,6 +16,7 @@
 EFBToTextureEnable = False
 BBoxEnable = True
 ImmediateXFBEnable = False
+DeferEFBCopies = False
 
 [Video_Stereoscopy]
 StereoConvergence = 545


JMC47:
Sounds reasonable as a conclusion.  This game does weird things with EFB2RAM, so the deferred copies could break things.

mbc07:
Any chance of this also affecting Super Paper Mario?

JMC47:
As both games use similar effects, I'd say probably.

teconmoon:
I can test SPM and post my results here, I don't recall having visual issues in my last Dolphin playthrough but not recent enough that I trust my memory.

JMC47:
It's not reason not to merge this though.

JMC47:
@dolphin-emu-bot rebuild

mbc07:
Definitely not a blocker, was just curious...

PR #8767, opened 2020-04-26
Expand freelook camera with control options

iwubcode:
This makes the freelook camera into a first class citizen (in both UI and code) and exposes a new option to change the freelook camera type to one of three options.

image

**Six Axis** is the type that Dolphin master uses.  It's essentially designed around those flight simulator games.  It makes it perfect for oddball screenshots because it allows roll but has the nasty side-effect of incurring unintended roll when pitching/yawing:

flight-sim-roll

On the other hand **First Person** is what I would have expected a freelook to work like originally.  It's similar in control to your stereotypical first person games, no roll will happen:

fps-no-roll

Finally, the **Orbital** type allows for orbiting around the original game camera point.  There's no movement controls, just rotation and zoom.  It is often referred to as "arcball":

orbital

### DIFF
--- a/Source/Core/Core/Config/GraphicsSettings.cpp
+++ b/Source/Core/Core/Config/GraphicsSettings.cpp
@@ -41,6 +41,8 @@ const Info<bool> GFX_DUMP_EFB_TARGET{{System::GFX, "Settings", "DumpEFBTarget"},
 const Info<bool> GFX_DUMP_XFB_TARGET{{System::GFX, "Settings", "DumpXFBTarget"}, false};
 const Info<bool> GFX_DUMP_FRAMES_AS_IMAGES{{System::GFX, "Settings", "DumpFramesAsImages"}, false};
 const Info<bool> GFX_FREE_LOOK{{System::GFX, "Settings", "FreeLook"}, false};
+const Info<FreelookControlType> GFX_FREE_LOOK_CONTROL_TYPE{
+    {System::GFX, "Settings", "FreeLookControlType"}, FreelookControlType::SixAxis};
 const Info<bool> GFX_USE_FFV1{{System::GFX, "Settings", "UseFFV1"}, false};
 const Info<std::string> GFX_DUMP_FORMAT{{System::GFX, "Settings", "DumpFormat"}, "avi"};
 const Info<std::string> GFX_DUMP_CODEC{{System::GFX, "Settings", "DumpCodec"}, ""};
--- a/Source/Core/Core/Config/GraphicsSettings.h
+++ b/Source/Core/Core/Config/GraphicsSettings.h
@@ -11,6 +11,7 @@
 enum class AspectMode : int;
 enum class ShaderCompilationMode : int;
 enum class StereoMode : int;
+enum class FreelookControlType : int;
 
 namespace Config
 {
@@ -41,6 +42,7 @@ extern const Info<bool> GFX_DUMP_EFB_TARGET;
 extern const Info<bool> GFX_DUMP_XFB_TARGET;
 extern const Info<bool> GFX_DUMP_FRAMES_AS_IMAGES;
 extern const Info<bool> GFX_FREE_LOOK;
+extern const Info<FreelookControlType> GFX_FREE_LOOK_CONTROL_TYPE;
 extern const Info<bool> GFX_USE_FFV1;
 extern const Info<std::string> GFX_DUMP_FORMAT;
 extern const Info<std::string> GFX_DUMP_CODEC;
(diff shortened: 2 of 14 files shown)


iwubcode:
Converted to draft as I still need to test the save state logic

iwubcode:
Thank you @MayImilae  for your feedback and testing!  Some of your issues have been resolved:

* UI descriptions have been updated as requested
* Orbit distance is now clamped, so you can't zoom through the point of origin!
* First person mode properly resets now
* Orbital camera has a default distance of 0, therefore mimicking the game's main view

Really appreciate you catching these!

Now onto the more interesting points:

> Orbital camera can also not roll the camera and thus cannot handle the games with a highly angled starting camera. This is very easy to fix: just add roll! For orbital I think that is a great solution.

I still need to find a game in my collection that has the extreme viewpoints you mention.  So far I think the closest I have is Little King's Story which is similar to Pikmin.  But there's gotta be some other Wii games that have the issues you speak of!

What would you expect when we add "roll"?  In my testing from adding this, everything seems off.  Either I'm rolling the orbiting axis which makes the orbit go all weird or I'm rolling the camera which seems counter-intuitive when my camera is facing a different way.  My matrix math days are long gone, so there might be a solution I'm missing.  Getting a clearer thought of your idea would be helpful though.

> First person view mode has no way to handle games with an angled camera. I mentioned this issue on the progress report back when we fixed it for the main option. That issue will be especially troublesome since the "just add roll" technique we did earlier won't apply here! It may have to be something smarter that is able to use the horizon as a reference point in the matrix, in addition to the original camera angle. I don't know how to do that but that is the only way I can think of to fix it. So maybe when you move the freelook camera it snaps alignment to the horizon or something, akin to how blender's game camera mode works?

Thanks for bringing this up.  Sadly, I don't have a good answer for this.  Your suggestion is interesting, I will have to look into blender to see if there's anything I can use.  The only solution I thought of is to add an option to remove the roll from the original game view but I'm not sure how feasible that is.  If anyone else has ideas, please let me know!

That being said, I would prefer not to hold up this PR from the possibility of being merged, especially when there's not a clear solution in sight, when 'SixAxis' does handle those specific cases.  I think as is, this PR has some value in being pushed to master.

Thank you again for the review.  Please let me know if you see anything else off!  I'll continue to investigate these last two issues.

DerKoun:
Sorry if this feels like a feature request in a place where it does not belong.

When talking about "zoom" here, I assume you mean moving the camera forward or backward, as it is in the current free look feature. While this works most of the time, it has limitations, e.g. "zooming out" i.e. moving backwards can move the camera through a wall, so the entire view can end up blocked.

The alternative "zoom" would be like an actual optical lens zoom, which does not have the above issue (though at large values you end up with a noticeable fish-eye effect).

The implementation is rather simple (I only tried it out hard-coded and that was some time ago): Multiply both fAspectRatioHackW and fAspectRatioHackH with the same dynamically changeable factor at https://github.com/dolphin-emu/dolphin/blob/533f6bfeb9a0c624d7837d0ce8daaac586e21233/Source/Core/VideoCommon/VertexShaderManager.cpp#L356-L364

So you'd have the equivalent to the zoom buttons on a camera. For example, I really like zooming out slightly to get a bit more image, especially when using stereoscopic 3D, which costs a bit of image on both sides.

I'd love to see this added, in addition to the various camera movement options already there (it's of course not a replacement) and currently being discussed here. Great to see that free-look is being improved upon anyway.

iwubcode:
Thanks @DerKoun !  This is PR already pretty beefy.  However, it's interesting you mentioned modifying the projection matrix, just the other day I wondered if anyone would want any control over the FOV.

Definitely something I can look at in the future.  Or feel free to open your own PR!

DerKoun:
@iwubcode: I'm not good with C++, so actually connecting the factor to a setting and/or key-binding would take me quite long. So if you could look into it at some point I'd much appreciate it.

psennermann:
Please note that this commit greatly decreases Dolphin performance (see https://bugs.dolphin-emu.org/issues/12157) 
Thanks

windwakr:
Hey, sorry if this isn't the best place for this. Regarding the comments about angled cameras and unwanted rotation, why not let the user set a "horizon" with a button press? That's how I handled this in a [freelook script I made](https://forums.desmume.org/viewtopic.php?id=11217)(beware, it's not pretty) for the DS emulator DeSmuME.
Basically, the user points at the horizon then presses a button. Their current pitch gets saved as the offset. Now you just apply the pitch offset before other calculations. Or something like that. Going back and looking at it I can't even tell what the code's doing, so I wouldn't use it as reference material. :P

There are some webms in that link that show it off.

iwubcode:
Thanks @windwakr .  Great idea!  I'll definitely keep that in mind for a future update.

PR #8768, opened 2020-04-27
Update Emulated Wii Remote defaults

TaoSc:
Fixes this issue when using the default Wii Remote preset:
Emulated Wii Remote Expression Parser

The expression parser throws a "Expected end of expression." warning when attempting to read expressions containing a whitespace.

I'm not sure if there are plans to change the parser's behavior but since this is a trivial fix, it shouldn't be problematic in the future

### DIFF
--- a/Source/Core/Core/HW/WiimoteEmu/WiimoteEmu.cpp
+++ b/Source/Core/Core/HW/WiimoteEmu/WiimoteEmu.cpp
@@ -683,14 +683,14 @@ void Wiimote::LoadDefaults(const ControllerInterface& ciface)
 // Buttons
 #if defined HAVE_X11 && HAVE_X11
   // A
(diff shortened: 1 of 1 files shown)


JMC47:
I can confirm that some of the defaults show up as invalid expressions in master.

BhaaLseN:
It _seems_ consistent with other expressions in that file, but at the same time not _all_ of them are consistent, so I'm not sure if that fixes all of it.

jordan-woyak:
Dolphin parses every expression in two ways. The expected way (supporting operators) and a fallback "bareword" mode which interprets the entire string as the name of an input.
Alternatively to changing the defaults we could make the status line aware of "bareword" expressions.
That might be a better solution.

Dentomologist:
This was fixed by #9416.

PR #8770, opened 2020-04-28
Android: Add DSP Emulation Engine

Ebola16:
Has multiple uses, including a temporary workaround for https://bugs.dolphin-emu.org/issues/10791

DSP Emulation Engine is actually two settings, so this is a `SingleChoiceSetting` special case.

### DIFF
--- a/Source/Android/app/src/main/res/values/arrays.xml
+++ b/Source/Android/app/src/main/res/values/arrays.xml
@@ -33,6 +33,26 @@
         <item>0</item>
     </integer-array>
 
+    <!-- DSP Emulation Engine -->
+    <string-array name="dspEngineEntriesX86_64" translatable="false">
+        <item>DSP HLE Emulation (fast)</item>
+        <item>DSP LLE Recompiler</item>
+        <item>DSP LLE Interpreter (slow)</item>
+    </string-array>
+    <integer-array name="dspEngineValuesX86_64" translatable="false">
+        <item>0</item>
+        <item>1</item>
+        <item>2</item>
+    </integer-array>
+    <string-array name="dspEngineEntriesGeneric" translatable="false">
+        <item>DSP HLE Emulation (fast)</item>
+        <item>DSP LLE Interpreter (slow)</item>
+    </string-array>
+    <integer-array name="dspEngineValuesGeneric" translatable="false">
+        <item>0</item>
+        <item>2</item>
+    </integer-array>
+
     <!-- GameCube System Languages -->
     <string-array name="gameCubeSystemLanguageEntries" translatable="false">
         <item>English</item>
--- a/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/model/Settings.java
+++ b/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/model/Settings.java
@@ -33,6 +33,10 @@ public class Settings
   public static final String SECTION_CONTROLS = "Controls";
   public static final String SECTION_PROFILE = "Profile";
 
+  private static final String DSP_HLE = "0";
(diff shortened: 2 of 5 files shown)


delroth:
LLE Recompiler should probably not be selectable on architectures where we don't have an LLE JIT.

Ebola16:
Do you know where I can find which architectures we support without LLE JIT?

delroth:
LLE JIT is only implemented on x86_64 right now.

JosJuice:
You can check for x86-64 using `NativeLibrary.DefaultCPUCore() == 1`.

delroth:
A better design would probably be to have some kind of "DefaultLLECore()". But that's slightly more involved.

JosJuice:
Unlike the LLE recompiler, I don't see any reason to include the LLE interpreter on x86-64 but not other architectures. It's equally slow on any architecture.

Ebola16:
As long as LLE interpreter works on all platforms I'd like to include it, even if it's really slow. There are cases where HLE doesn't work at all (like my linked issue). Also, the x86-64 detection I used is used elsewhere in Dolphin so if that gets changed, I'd like to do it in a separate PR. Is all of this acceptable?

JosJuice:
Seems okay to me.

Ebola16:
Just to make sure I understand, does this mean "EnableJIT" for the DSP should not be enabled on non-x86-64?

JosJuice:
I think it makes the most sense that way. However, I believe it doesn't actually matter what you set EnableJIT to when you're not on x86-64 or are using DSP HLE.

Ebola16:
Ok, if non-x86-64 ignores the setting then I'll leave this as is. Having platform-specific implementations for DSP HLE would be ugly. Thanks!

Source/Android/app/src/main/res/values/arrays.xml
```diff
@@ -33,6 +33,26 @@
         <item>0</item>
     </integer-array>
 
+    <!-- DSP Emulation Engine -->
+    <string-array name="dspEngineEntriesX86_64" translatable="false">
+        <item>DSP HLE Emulation (fast)</item>
```
leoetlino, line 38: Nitpick, but this string suffers from RAS syndrome. I suggest removing Emulation:

```suggestion
        <item>DSP HLE (fast)</item>
```

Source/Android/app/src/main/res/values/arrays.xml
```diff
@@ -33,6 +33,26 @@
...
+        <item>DSP LLE Recompiler</item>
+        <item>DSP LLE Interpreter (slow)</item>
+    </string-array>
+    <integer-array name="dspEngineValuesX86_64" translatable="false">
+        <item>0</item>
+        <item>1</item>
+        <item>2</item>
+    </integer-array>
+    <string-array name="dspEngineEntriesGeneric" translatable="false">
+        <item>DSP HLE Emulation (fast)</item>
```
leoetlino, line 48: ditto

Source/Android/app/src/main/res/values/arrays.xml
JosJuice: This is what we use in DolphinQt, though... I am planning to add translations for Android at some point, and when we get that, it's good if the strings are identical between Android and DolphinQt so that the translators don't get any extra work.

leoetlino: Oh, DolphinQt has the same problem :thinking: 

Do you think we should update the DolphinQt string too, then? Or leave it as is?

JosJuice: It would make sense to open a new PR with that change, I think.

PR #8771, opened 2020-04-28
Add /dev/dolphin interface to allow game to read the REAL product code

Leseratte10:
Related to issue 11930. 

Right now, when you import a NAND backup into Dolphin, the serial number from the NAND is used by Dolphin. As soon as you play an out-of-region game, the serial number prefix in the settings.txt is overwritten with one that matches the game region, unless you have the "Allow mismatched region settings" setting enabled. The original prefix is then permanently lost. 

However, as stated by JosJuice in the above issue report, it is possible that games don't like being booted with an out-of-region serial number, so this kind of prefix replacement is necessary. 

This pull request adds an ioctl value to the /dev/dolphin device that allows the game to access the REAL serial number prefix, meaning, the one that was imported from the actual NAND, no matter how many out-of-region games you are playing. 

This setting is stored in the Dolphin.ini config file, with a default value of "DOL" for Dolphin when the user didn't import a NAND backup yet. This will not cause any incompatibilities with games, as - same for all the other functions of the /dev/dolphin device - they won't be accessing that. 

Dolphin-aware software however can, by directly accessing this device, find out the real serial number prefix. 

This change is needed to make using Wiimmfi easier for Dolphin users - the serial number including the prefix is one of the things that Wiimmfi uses to identify a console, so when that changes, you effectively have a new console and a new 7-day wait. This means that currently users have to play games from the same region as their Wii NAND, or they have to remember to always keep the "allow non-matching settings" enabled - as soon as they disable it once and run an out-of-region game, the proper prefix is lost forever. 

With this change the proper prefix is stored in the config file, and the Dolphin-aware game (we can patch Mario Kart Wii from the server to include that code, same as we did for #8445) can access the actual prefix and let the server know, so it doesn't generate a new console identity with a new wait time. 

I couldn't fully test this change yet as I first need to write the game update code to actually read and send that prefix to the server, so this is not ready to be merged yet. Just wanted to create the PR so other people can already look at that.

### DIFF
--- a/Source/Core/Core/IOS/DolphinDevice.cpp
+++ b/Source/Core/Core/IOS/DolphinDevice.cpp
@@ -5,7 +5,12 @@
 #include <algorithm>
 #include <cstring>
 
+#include "Common/CommonPaths.h"
+#include "Common/File.h"
+#include "Common/FileUtil.h"
 #include "Common/Logging/Log.h"
+#include "Common/NandPaths.h"
+#include "Common/SettingsHandler.h"
 #include "Common/Timer.h"
 #include "Common/scmrev.h"
 #include "Core/BootManager.h"
@@ -25,6 +30,7 @@ enum
   IOCTL_DOLPHIN_GET_SPEED_LIMIT = 0x03,
   IOCTL_DOLPHIN_SET_SPEED_LIMIT = 0x04,
   IOCTL_DOLPHIN_GET_CPU_SPEED = 0x05,
+  IOCTL_DOLPHIN_GET_REAL_PRODUCTCODE = 0x06,
 
 };
 
@@ -123,6 +129,37 @@ IPCCommandResult SetSpeedLimit(const IOCtlVRequest& request)
   return DolphinDevice::GetDefaultReply(IPC_SUCCESS);
 }
 
+IPCCommandResult GetRealProductCode(const IOCtlVRequest& request)
+{
+  if (!request.HasNumberOfValidVectors(0, 1))
+  {
+    return DolphinDevice::GetDefaultReply(IPC_EINVAL);
+  }
+
+  const std::string backup_file_path = File::GetUserPath(D_BACKUP_IDX) + DIR_SEP + WII_SETTING;
+
+  File::IOFile file(backup_file_path, "rb");
+  if (!file)
+    return DolphinDevice::GetDefaultReply(IPC_ENOENT);
+
+  Common::SettingsHandler::Buffer data;
+
+  if (!file.ReadBytes(data.data(), data.size()))
+    return DolphinDevice::GetDefaultReply(IPC_ENOENT);
+
+  Common::SettingsHandler gen;
+  gen.SetBytes(std::move(data));
+  const std::string code = gen.GetValue("CODE");
+
+  const size_t length = std::min<size_t>(request.io_vectors[0].size, code.length());
+  if (length == 0)
+    return DolphinDevice::GetDefaultReply(IPC_ENOENT);
+
+  Memory::Memset(request.io_vectors[0].address, 0, request.io_vectors[0].size);
+  Memory::CopyToEmu(request.io_vectors[0].address, code.c_str(), length);
+  return DolphinDevice::GetDefaultReply(IPC_SUCCESS);
+}
+
 }  // namespace
 
 IPCCommandResult DolphinDevice::IOCtlV(const IOCtlVRequest& request)
@@ -144,6 +181,8 @@ IPCCommandResult DolphinDevice::IOCtlV(const IOCtlVRequest& request)
     return SetSpeedLimit(request);
   case IOCTL_DOLPHIN_GET_CPU_SPEED:
     return GetCPUSpeed(request);
+  case IOCTL_DOLPHIN_GET_REAL_PRODUCTCODE:
+    return GetRealProductCode(request);
   default:
     return GetDefaultReply(IPC_EINVAL);
   }


leoetlino:
What about simply having Dolphin restore the product code from Dolphin.ini? Some other titles such as the shopping channel (and possibly more things that use EC?) are also affected by the same issue.

JosJuice:
> Right now, when you import a NAND backup into Dolphin, the serial number from the NAND is used by Dolphin. As soon as you play an out-of-region game, the serial number prefix in the settings.txt is overwritten with one that matches the game region, unless you have the "Allow mismatched region settings" setting enabled. The original prefix is then permanently lost.

Due to PR #8672, it doesn't get permanently lost anymore. So you should be able to skip adding a new setting.

leoetlino:
Ah yes, I forgot we back up the whole setting.txt and SYSCONF now.

Leseratte10:
Damn, I should have confirmed that before I started working on this. I completely forgot about the PR that saves that. 

So, that does mean that when I import a PAL NAND into Dolphin, any PAL game I boot, independantly from that region setting and from the fact that I might have booted out-of-region games before, will see the original serial number prefix from the NAND and not an autogenerated one?

Okay, so that would mean I would just need to read the prefix from the backup SYSCONF when the game asks for it, so that I can get it when it is out-of-region, too.

JosJuice:
> So, that does mean that when I import a PAL NAND into Dolphin, any PAL game I boot, independantly from that region setting and from the fact that I might have booted out-of-region games before, will see the original serial number prefix from the NAND and not an autogenerated one?

Yes (unless you have booted out-of-region games using versions of Dolphin prior to that PR).

> Okay, so that would mean I would just need to read the prefix from the backup SYSCONF when the game asks for it, so that I can get it when it is out-of-region, too.

Yes.

Leseratte10:
That backup behaviour seems to be buggy. 
I tried finding the backup of the setting.txt, but didn't find it anywhere. Then I looked into the logs and saw that it should be written to the "Backups" folder. That didn't exist for me. Once I created that folder, the settings.txt was correctly backed-up and restored. Is it possible that the code to create the Backup folder is missing? That should probably result in a warning. 

Now that I have the backup working, I'll try making my code just read that backup file.

JosJuice:
Thanks for catching that. Should be fixed by PR #8772.

Leseratte10:
Thanks. When that is merged, can I assume that the Backup is always created (and can I return an error when it's missing) or do I need to fall back to reading the non-backup file?

JosJuice:
You can assume that it always exists while Wii emulation is running.

Leseratte10:
Okay, with the new commit that should be fixed. The prefix is no longer added to the config file, the ioctlv request now just reads the serial number prefix from the backup file. 

And a quick patch to Mario Kart Wii later, Dolphin now connects to Wiimmfi with the same identity as my real console, independantly from the game region or the "allow non-matching settings" setting, so that seems to work great. 

Could someone trigger the buildbot please so I can get some more people to test this with the updated MKWii patch before this gets merged.

JosJuice:
@dolphin-emu-bot rebuild

Leseratte10:
Huh, is there no longer a Windows buildbot or did that fail to start because I forgot `clang-format`?

JosJuice:
It just takes a little while to start building if other PRs also were pushed recently.

JosJuice:
@dolphin-emu-bot rebuild

JosJuice:
@dolphin-emu-bot rebuild

Source/Core/Core/IOS/DolphinDevice.cpp
```diff
@@ -123,6 +129,37 @@ IPCCommandResult SetSpeedLimit(const IOCtlVRequest& request)
...
+  {
+    return DolphinDevice::GetDefaultReply(IPC_EINVAL);
+  }
+
+  const std::string backup_file_path = File::GetUserPath(D_BACKUP_IDX) + DIR_SEP + WII_SETTING;
+
+  File::IOFile file(backup_file_path, "rb");
+  Common::SettingsHandler::Buffer data;
+
+  if (file && file.ReadBytes(data.data(), data.size()))
```
leoetlino, line 144: I'd invert the condition here and return early on failure so that the rest of the function can be unindented.

Source/Core/Core/IOS/DolphinDevice.cpp
```diff
@@ -25,6 +30,7 @@ enum
   IOCTL_DOLPHIN_GET_SPEED_LIMIT = 0x03,
   IOCTL_DOLPHIN_SET_SPEED_LIMIT = 0x04,
   IOCTL_DOLPHIN_GET_CPU_SPEED = 0x05,
+  IOCTL_DOLPHIN_GET_REAL_PRODUCTCODE = 0x06
```
leoetlino, line 33:
```suggestion
  IOCTL_DOLPHIN_GET_REAL_PRODUCTCODE = 0x06,
```

Source/Core/Core/IOS/DolphinDevice.cpp
```diff
@@ -123,6 +129,37 @@ IPCCommandResult SetSpeedLimit(const IOCtlVRequest& request)
...
+    gen.SetBytes(std::move(data));
+    const std::string code = gen.GetValue("CODE");
+
+    const auto length = std::min(size_t(request.io_vectors[0].size), code.length());
+    if (length == 0)
+      return DolphinDevice::GetDefaultReply(IPC_ENOENT);
+
+    Memory::Memset(request.io_vectors[0].address, 0, request.io_vectors[0].size);
+    Memory::CopyToEmu(request.io_vectors[0].address, code.c_str(), length);
+    gen.Reset();
```
leoetlino, line 156: Is this necessary?

Source/Core/Core/IOS/DolphinDevice.cpp
```diff
@@ -123,6 +129,37 @@ IPCCommandResult SetSpeedLimit(const IOCtlVRequest& request)
   return DolphinDevice::GetDefaultReply(IPC_SUCCESS);
 }
 
+IPCCommandResult GetRealProductcode(const IOCtlVRequest& request)
```
leoetlino, line 132:
```suggestion
IPCCommandResult GetRealProductCode(const IOCtlVRequest& request)
```

Source/Core/Core/IOS/DolphinDevice.cpp
Leseratte10: The Reset call? I'm not sure. Probably not, given that we aren't using the `gen` object afterwards so I can probably get rid of that.

Leseratte10:
No, not exactly. "Product code" in Nintendo-speak means just the letters in front of the serial number, not the whole serial number. 
And yes, this is related to Wiimmfi. Currently when you play an out-of-region game, Dolphin changes the product code to match the region (in case any game verifies that). This patch gives Mario Kart Wii the ability to get the actual code (unchanged) and use that when logging on to Wiimmfi.

Source/Core/Core/IOS/DolphinDevice.cpp
JosJuice: Yes, there is no reason to call `Reset` if we aren't going to use the object again.

Source/Core/Core/IOS/DolphinDevice.cpp
```diff
@@ -123,6 +129,38 @@ IPCCommandResult SetSpeedLimit(const IOCtlVRequest& request)
...
+  Common::SettingsHandler::Buffer data;
+
+  if (!file.ReadBytes(data.data(), data.size()))
+    return DolphinDevice::GetDefaultReply(IPC_ENOENT);
+
+  Common::SettingsHandler gen;
+  gen.SetBytes(std::move(data));
+  const std::string code = gen.GetValue("CODE");
+
+  const auto length = std::min(size_t(request.io_vectors[0].size), code.length());
```
JosJuice, line 154: I would prefer:

```suggestion
  const size_t length = std::min<size_t>(request.io_vectors[0].size, code.length());
```

But it isn't too important.

Source/Core/Core/IOS/DolphinDevice.cpp
Leseratte10: Okay, I've removed it.

Source/Core/Core/IOS/DolphinDevice.cpp
Leseratte10: Done.

JosJuice:
@dolphin-emu-bot rebuild

leoetlino:
@Leseratte10 Has this received enough testing from Wiimmfi users?

Leseratte10:
Not yet, but the Wiimmfi / game side for this patch is now ready so we will merge that to Wiimmfi soon, then I'll get some people with the builds from the buildbot to test. I'll let you know when this has been confirmed working by a couple people so this can be merged.

Leseratte10:
Sorry that took a while. The Wiimmfi side of this is now ready and available for everyone, and I've tested it with NANDs from multiple regions. Seems to work fine, so from my side this should be ready to be merged now.

PR #8775, opened 2020-04-28
MathUtil: Fix Rectangle::GetWidth/Height for unsigned types

leoetlino:
`std::abs(x - y)` where x and y are unsigned integers fails to compile
with an "call of overloaded 'abs(unsigned int)' is ambiguous" error
on GCC, and even if it did compile, that expression still wouldn't
give the correct result since `x - y` is unsigned.

### DIFF
--- a/Source/Core/Common/MathUtil.h
+++ b/Source/Core/Common/MathUtil.h
@@ -6,6 +6,7 @@
 
 #include <algorithm>
 #include <cmath>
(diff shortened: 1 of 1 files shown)


leoetlino:
Hmm, I just realised this gives incorrect results if both left/right are negative.

leoetlino:
Should be fixed now.

PR #8776, opened 2020-04-29
Require minizip 2.0.0 or newer

Techjar:
Minizip 1.x is the old broken version, while 2.x is the newer fork we use. Many distros ship the old version which is causing stack smashing in RedumpVerifier, so we'll only use the shared lib if it's the newer version.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -610,7 +610,7 @@ else()
   add_subdirectory(Externals/zlib)
 endif()
 
-pkg_check_modules(MINIZIP minizip)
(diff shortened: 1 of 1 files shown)


degasus:
The stack smashing is because of a header to lib mismatch.
I still don't know why, but we include the header from externals but link to the system library. This works fine for the function calls, but the struct unz_file_info does not match.

So indeed, the fork might be better. However the crash is because of a wrong include directory. GIT_ROOT/Externals must not be within the include directories.

Techjar:
Ah, yeah I did not know about that.

PR #8778, opened 2020-04-29
GC IPL Hack for Configurable MEM1 Size.

Minty-Meeo:
A follow up to my prior Pull Request (Configurable MEM1 and MEM2 sizes at runtime via Dolphin.ini #8722).  This is a hack to assembly edit retail IPLs to support different MEM1 sizes before they are copied into Emulator Memory.  As with any other Emulated Memory Size Overrides, this change does nothing if the MAIN_RAM_OVERRIDE_ENABLE config setting is false.
I also updated a comment in Memmap.cpp to reflect new behavior from this commit and my prior one as well.

### DIFF
--- a/Source/Core/Core/Boot/Boot.cpp
+++ b/Source/Core/Core/Boot/Boot.cpp
@@ -41,6 +41,9 @@ namespace fs = std::filesystem;
 #include "Core/FifoPlayer/FifoPlayer.h"
 #include "Core/HLE/HLE.h"
 #include "Core/HW/DVD/DVDInterface.h"
+#include "Core/HW/EXI/EXI.h"
+#include "Core/HW/EXI/EXI_Channel.h"
+#include "Core/HW/EXI/EXI_Device.h"
 #include "Core/HW/EXI/EXI_DeviceIPL.h"
 #include "Core/HW/Memmap.h"
 #include "Core/HW/VideoInterface.h"
@@ -311,8 +314,6 @@ std::unique_ptr<BootParameters> BootParameters::GenerateFromFile(std::vector<std
 
 BootParameters::IPL::IPL(DiscIO::Region region_) : region(region_)
 {
-  const std::string directory = SConfig::GetInstance().GetDirectoryForRegion(region);
-  path = SConfig::GetInstance().GetBootROMPath(directory);
 }
 
 BootParameters::IPL::IPL(DiscIO::Region region_, Disc&& disc_) : IPL(region_)
@@ -389,81 +390,6 @@ bool CBoot::LoadMapFromFilename()
   }
 
   return false;
-}
-
-// If ipl.bin is not found, this function does *some* of what BS1 does:
-// loading IPL(BS2) and jumping to it.
-// It does not initialize the hardware or anything else like BS1 does.
-bool CBoot::Load_BS2(const std::string& boot_rom_filename)
-{
-  // CRC32 hashes of the IPL file, obtained from Redump
-  constexpr u32 NTSC_v1_0 = 0x6DAC1F2A;
-  constexpr u32 NTSC_v1_1 = 0xD5E6FEEA;
-  constexpr u32 NTSC_v1_2 = 0x86573808;
-  constexpr u32 MPAL_v1_1 = 0x667D0B64;  // Brazil
-  constexpr u32 PAL_v1_0 = 0x4F319F43;
-  constexpr u32 PAL_v1_2 = 0xAD1B7F16;
-
-  // Load the whole ROM dump
-  std::string data;
-  if (!File::ReadFileToString(boot_rom_filename, data))
-    return false;
-
-  const u32 ipl_hash = Common::ComputeCRC32(data);
-  bool known_ipl = false;
-  bool pal_ipl = false;
-  switch (ipl_hash)
-  {
-  case NTSC_v1_0:
-  case NTSC_v1_1:
-  case NTSC_v1_2:
-  case MPAL_v1_1:
-    known_ipl = true;
-    break;
-  case PAL_v1_0:
-  case PAL_v1_2:
-    pal_ipl = true;
-    known_ipl = true;
-    break;
-  default:
-    PanicAlertFmtT("The IPL file is not a known good dump. (CRC32: {0:x})", ipl_hash);
-    break;
-  }
-
-  const DiscIO::Region boot_region = SConfig::GetInstance().m_region;
-  if (known_ipl && pal_ipl != (boot_region == DiscIO::Region::PAL))
-  {
-    PanicAlertFmtT("{0} IPL found in {1} directory. The disc might not be recognized",
-                   pal_ipl ? "PAL" : "NTSC", SConfig::GetDirectoryForRegion(boot_region));
-  }
-
-  // Run the descrambler over the encrypted section containing BS1/BS2
-  ExpansionInterface::CEXIIPL::Descrambler((u8*)data.data() + 0x100, 0x1AFE00);
-
-  // TODO: Execution is supposed to start at 0xFFF00000, not 0x81200000;
-  // copying the initial boot code to 0x81200000 is a hack.
-  // For now, HLE the first few instructions and start at 0x81200150
-  // to work around this.
-  Memory::CopyToEmu(0x01200000, data.data() + 0x100, 0x700);
-  Memory::CopyToEmu(0x01300000, data.data() + 0x820, 0x1AFE00);
-
-  PowerPC::ppcState.gpr[3] = 0xfff0001f;
-  PowerPC::ppcState.gpr[4] = 0x00002030;
-  PowerPC::ppcState.gpr[5] = 0x0000009c;
-
-  MSR.FP = 1;
-  MSR.DR = 1;
-  MSR.IR = 1;
-
-  PowerPC::ppcState.spr[SPR_HID0] = 0x0011c464;
-  PowerPC::ppcState.spr[SPR_IBAT3U] = 0xfff0001f;
-  PowerPC::ppcState.spr[SPR_IBAT3L] = 0xfff00001;
-  PowerPC::ppcState.spr[SPR_DBAT3U] = 0xfff0001f;
-  PowerPC::ppcState.spr[SPR_DBAT3L] = 0xfff00001;
-  SetupBAT(/*is_wii*/ false);
-
-  PC = 0x81200150;
-  return true;
 }
 
 static void SetDefaultDisc()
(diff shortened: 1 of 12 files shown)


Source/Core/Core/Boot/Boot.cpp
```diff
@@ -351,6 +351,44 @@ bool CBoot::Load_BS2(const std::string& boot_rom_filename)
   // Run the descrambler over the encrypted section containing BS1/BS2
   ExpansionInterface::CEXIIPL::Descrambler((u8*)data.data() + 0x100, 0x1AFE00);
 
+  // Perform a surgical modification on known IPLs for Emulated Memory Size Override.
+  if (Config::Get(Config::MAIN_RAM_OVERRIDE_ENABLE))
```
lioncash, line 355: This block should go in its own function so all of the memory override stuff is self contained.

Source/Core/Core/Boot/Boot.cpp
```diff
@@ -351,6 +351,44 @@ bool CBoot::Load_BS2(const std::string& boot_rom_filename)
...
+    u32* ori_that_ruins_things;
+
+    switch (ipl_hash)
+    {
+    case NTSC_v1_0:
+    case NTSC_v1_1:
+    case PAL_v1_0:
+    case MPAL_v1_1:
+      lis_simm_field = (s16*)(data.data() + 0xCB6);         // 0x820 + 0x494 + 2
+      ori_that_ruins_things = (u32*)(data.data() + 0x5FC);  // 0x100 + 0x4fc
```
lioncash, line 374: Dereferencing any of these will result in undefined behavior, as this violates C++ strict aliasing rules. If you want to make these modifications, then you need to make use of `std::memcpy` to do it in a well-defined manner e.g.

```cpp
const u32 nop = Common::swap16(0x60000000);
std::memcpy(data.data() + 0x5FC, &nop, sizeof(nop));
std::memcpy(data.data() + 0xCB6, &mem1_size, sizeof(mem1_size));
```

Source/Core/Core/Boot/Boot.h
```diff
@@ -162,3 +170,5 @@ void UpdateStateFlags(std::function<void(StateFlags*)> update_function);
 /// Normally, this is automatically done by ES when the System Menu is installed,
 /// but we cannot rely on this because we don't require any system titles to be installed.
 void CreateSystemMenuTitleDirs();
+
+void RAMOverrideForIPLMemoryValues(std::string& data, const u32 ipl_hash);
```
lioncash, line 174: This doesn't need to be exposed. This function can be an internal static function. It just needs to be defined above its usage spot.

Source/Core/Core/Boot/Boot.h
```diff
@@ -32,6 +32,14 @@ struct RegionSetting
   std::string code;
 };
 
+// CRC32 hashes of the IPL file, obtained from Redump
+constexpr u32 NTSC_v1_0 = 0x6DAC1F2A;
```
lioncash, line 36: Why were these moved out here? These can be kept within the cpp file.

Source/Core/Core/Boot/Boot.cpp
```diff
@@ -593,3 +588,38 @@ void CreateSystemMenuTitleDirs()
   const auto es = IOS::HLE::GetIOS()->GetES();
   es->CreateTitleDirectories(Titles::SYSTEM_MENU, IOS::SYSMENU_GID);
 }
+
+void RAMOverrideForIPLMemoryValues(std::string& data, const u32 ipl_hash)
+{
+  if (Config::Get(Config::MAIN_RAM_OVERRIDE_ENABLE))
```
lioncash, line 594: Now that this is localized within its own function, this can be inverted like so to allow unindenting all of the code within it.

```cpp
if (!Config::Get(Config::MAIN_RAM_OVERRIDE_ENABLE))
  return;
```

Source/Core/Core/Boot/Boot.h
Minty-Meeo: Good point

Source/Core/Core/Boot/Boot.cpp
```diff
@@ -300,19 +308,46 @@ bool CBoot::LoadMapFromFilename()
   return false;
 }
 
+void RAMOverrideForIPLMemoryValues(std::string& data, const u32 ipl_hash)
```
lioncash, line 311:
```suggestion
static void RAMOverrideForIPLMemoryValues(std::string& data, const u32 ipl_hash)
```

Otherwise this results in a -Wmissing-declaration warning.

Source/Core/Core/Boot/Boot.cpp
```diff
@@ -300,19 +308,46 @@ bool CBoot::LoadMapFromFilename()
...
+
+  // Pre-calculate the new values.
+  const s16 mem1_s_size = Common::swap16(Config::Get(Config::MAIN_MEM1_SIZE) >> 16);
+  const u32 nop = Common::swap32(0x60000000);
+
+  // [lis	r0, 0x0180] <-- This instruction is what sets the Simulated MEM1 Size in retail IPLs.
+  // The following code replaces the SIMM field and nothing else.
+  // [ori	r9, r9, 0x0003] <-- This instruction putting the value "3" into r9 causes a stw later
+  // down the line to revert the instruction we're trying to change.  I don't understand why, but
+  // I do know that NOPing this away doesn't break anything!
```
Pokechu22, line 324: It appears that `r9` is then written to [EXI0CR](http://hitmen.c02.at/files/yagcd/yagcd/chap5.html#sec5) (`0xCC00680C`), where 3 is start (1) DMA (2) read.  Setting this to 0 disables it (and since the code then polls EXI0CR for bit 0 to be cleared, it immediately finishes).

But perhaps it would be better to modify the data it's reading from EXI instead of patching it out?  I'm not sure where it'd be getting it from even (maybe it's something in `EXI_DeviceIPL`, but I don't see anything that explicitly implements it).

Minty-Meeo: Finally some closure on what the heck was going on.  I was so confused watching as a stw instruction writing to some bizarre address was causing instructions to revert of all things.

Pokechu22: ... in fact, the whole `Memory::CopyToEmu(0x01300000, data.data() + 0x820, 0x1AFE00);` line is unneeded, since BS1 does that itself ([see YAGCD](http://hitmen.c02.at/files/yagcd/yagcd/chap18.html#sec18.2), though when they say "clearing bit 17 of 0xcc006800" it looks like it should be bit 13 (`ROMDIS`)).

But I don't think you need to address that in this PR; just change the comment here to something like

> // [ori	r9, r9, 0x0003] <-- The value 3 in r9 later starts an EXI DMA read to copy BS2, which clobbers the patched version.  Leaving it at 0 makes it think the EXI transfer finished immediately, skipping it.

Source/Core/Core/Boot/Boot.cpp
```diff
@@ -300,19 +308,46 @@ bool CBoot::LoadMapFromFilename()
...
+static void RAMOverrideForIPLMemoryValues(std::string& data, const u32 ipl_hash)
+{
+  if (!Config::Get(Config::MAIN_RAM_OVERRIDE_ENABLE))
+    return;
+
+  // Pre-calculate the new values.
+  const s16 mem1_s_size = Common::swap16(Config::Get(Config::MAIN_MEM1_SIZE) >> 16);
+  const u32 nop = Common::swap32(0x60000000);
+
+  // [lis	r0, 0x0180] <-- This instruction is what sets the Simulated MEM1 Size in retail IPLs.
```
Pokechu22, line 320: As far as I can tell, it's writing to `80000028`, which [wiibrew calls](https://wiibrew.org/wiki/Memory_Map) is "Memory Size (Physical)", not Simulated MEM1 size.  Simulated Memory Size is 800000F0 instead (which doesn't appear to be written by the IPL).

Minty-Meeo: My mistake, I just assumed it was simulated.

Minty-Meeo:
lol I thought it got merged already I'm a dummy

RenaKunisaki:
Just curious why this doesn't affect the simulated size? And MEM2 size seems to do nothing, couldn't it change ARAM size?

Minty-Meeo:
> 
> 
> Just curious why this doesn't affect the simulated size? And MEM2 size seems to do nothing, couldn't it change ARAM size?

Changing MEM2 for GameCube does nothing since it doesn't have MEM2.  Changing ARAM should be possible but I didn't target that with my PR that added configurable RAM sizes since I wasn't aware that could be changed at the time.
I've been away from Dolphin devving for a bit since I'm busy with a romhack, but I will come back to this eventually.

Minty-Meeo:
It's time to bring this PR back to life.

Minty-Meeo:
I am exploring potentially emulating the IPL's presence in physical memory accurately and I might roll back re-ordering HW::Init() in the way I did.

Source/Core/Core/HW/EXI/EXI_DeviceIPL.cpp
```diff
@@ -98,31 +100,90 @@ void CEXIIPL::Descrambler(u8* data, u32 size)
...
+      break;
+    case CRC32_NTSC_v1_2:
+    case CRC32_PAL_v1_2:
+      std::memcpy(data + 0x820 + 0x490 + 2, &simm_p_mem1_size, sizeof(simm_p_mem1_size));
+      std::memcpy(data + 0x100 + 0x31C + 2, &simm_p_mem1_size, sizeof(simm_p_mem1_size));
+      std::memcpy(data + 0x100 + 0x36C + 2, &simm_p_mem1_size, sizeof(simm_p_mem1_size));
+      break;
+    default:
+      PanicAlertFmtT(
+          "Emulated Memory Size Override is not supported for unrecognized IPL dumps.\n");
```
BhaaLseN, line 128: Does this intentionally add the newline at the end? There's nothing behind it, and the dialog ends; so it is either a left-over or you forgot to add something _else_ there.

Source/Core/Core/Boot/Boot.cpp
```diff
@@ -594,18 +520,46 @@ bool CBoot::BootUp(std::unique_ptr<BootParameters> boot)
...
+          ExpansionInterface::GetChannel(0)->GetDevice(2));
+      DEBUG_ASSERT(exi_ipl->m_device_type == ExpansionInterface::EXIDeviceType::MaskROM);
+
+      if (!exi_ipl->IsROMLoaded())
       {
         if (ipl.disc)
           PanicAlertFmtT("Cannot start the game, because the GC IPL could not be found.");
         else
-          PanicAlertFmtT("Cannot find the GC IPL.");
+          PanicAlertFmtT("Could not find the GC IPL.");
```
BhaaLseN, line 535: Seems a bit odd that you only changed one branch of the check, and with a very minor change ("Cannot" vs. "Could not").

Any practical reason why past tense is more appropriate here than the previous present tense?

Source/Core/Core/Boot/Boot.cpp
```diff
@@ -594,18 +520,46 @@ bool CBoot::BootUp(std::unique_ptr<BootParameters> boot)
...
         return false;
       }
 
-      if (!Load_BS2(ipl.path))
-        return false;
+      // MSR.IP = 1;
```
BhaaLseN, line 539: Should this be further down with the other `MSR` initializations, and uncommented?

Source/Core/Core/HW/EXI/EXI_DeviceIPL.cpp
Minty-Meeo: Oh right.  This is a shortened copy of a message I had in a previous version of this PR.  There used to be a second line of dialogue, yes.  I'll fix that.

Source/Core/Core/Boot/Boot.cpp
Minty-Meeo: I found it appropriate to use past tense here, since loading the IPL now takes place earlier, in ExpansionInterface::Init().

Source/Core/Core/Boot/Boot.cpp
Minty-Meeo: According to yagcd, MSR.IP is set to 1 when the GameCube is powered on.  It's not part of the HLE hack that is being used, but I thought I wanted to document it somehow.  I think I'll remove this, now, since I want to try putting the IPL at physical memory 0xFFF00000 for a future PR.

Adamillo:
I don't quite understand this as a casual Dolphin user that likes to do dumb experiments. What is this feature about?

JosJuice:
Adding more memory does not improve the performance.

Minty-Meeo:
> I don't quite understand this as a casual Dolphin user that likes to do dumb experiments. What is this feature about?

Emulated Memory Override was made possible because CBoot::EmulatedBS2_GC can set the physical MEM1 size in Dolphin-OS globals to arbitrary values. It is incompatible with retail IPLs since they hardcode the physical MEM1 size to be 24MiB.  This PR tackles the niche use case of combining Emulated Memory Override with running the real IPL.  By assembly editing the SIMM fields containing the value 24MiB, now retail IPLs can join the fun.

Emulated Memory Override is only really useful for a leaked Metroid Prime 3 GCN build that ran on dev hardware (they had 48MiB), maybe one Fire Emblem game that has random crashes in Dolphin, and any romhacks willing to give up console compatibility for additional RAM.

Minty-Meeo:
I was reminded that the TDEV IPL exists and is configured for 48MiB of physical MEM1.  This got me thinking that maybe this feature should be a seperate toggle from Emulated Memory Override.  Thoughts?  Where in the GUI should this even go?  Also, how can I get verified hashes for these obscure IPLs?

Pokechu22:
>  Also, how can I get verified hashes for this obscure IPL and others?

I believe Redump's BIOS datfile from [here](http://redump.org/downloads/) contains the hashes for those now (though it didn't in the past)

Minty-Meeo:
I have found something interesting in the NPDP GDEV IPL.  It checks bit 3 of the DVD Interface Configuration Register (0xCC006024) which YAGCD says is unused.  If it is 0, the p mem1 size is set to 48MiB.  If it is 1, the p mem1 size is set to 24MiB.  Dolphin emulates it such that the 48MiB path is chosen.  I'm not sure how I should patch this IPL?

Minty-Meeo:
While everything in this PR is functional, I know it edges on being outside of Dolphin Emulator's scope.  If it gets accepted, awesome.  If it gets rejected, I'll bring the practical changes to a future PR.  I don't think there is anything more to change.

Minty-Meeo:
I have finished writing a command line tool in Python that does everything this Pull Request can.  Unless there are any objections, I think I will close this pull request tonight and bring the practical improvements to a future PR like I said before.  While this PR is neat, I see now that it is textbook feature-creep.  Thank you all for your help and guidance.

PR #8779, opened 2020-04-29
NetworkWidget: Reorganise SSL options group box

sepalani:
This PR splits SSL options group box into 2 group boxes. One dedicated to dumping options, the other to the network security options. On top of that a combo box was added to select the network dump format and a button to open the dump folder.

image
image

### DIFF
--- a/Source/Core/DolphinQt/Debugger/NetworkWidget.cpp
+++ b/Source/Core/DolphinQt/Debugger/NetworkWidget.cpp
@@ -5,10 +5,15 @@
 #include "DolphinQt/Debugger/NetworkWidget.h"
 
 #include <QCheckBox>
+#include <QComboBox>
+#include <QDesktopServices>
 #include <QGridLayout>
 #include <QGroupBox>
 #include <QHeaderView>
+#include <QLabel>
+#include <QPushButton>
 #include <QTableWidget>
+#include <QUrl>
 #include <QVBoxLayout>
 
 #ifdef _WIN32
@@ -18,6 +23,7 @@
 #include <sys/types.h>
 #endif
 
+#include "Common/FileUtil.h"
 #include "Core/Config/MainSettings.h"
 #include "Core/IOS/Network/SSL.h"
 #include "Core/IOS/Network/Socket.h"
@@ -191,7 +197,8 @@ void NetworkWidget::CreateWidgets()
   widget->setLayout(layout);
   layout->addWidget(CreateSocketTableGroup());
   layout->addWidget(CreateSSLContextGroup());
-  layout->addWidget(CreateSSLOptionsGroup());
+  layout->addWidget(CreateDumpOptionsGroup());
+  layout->addWidget(CreateSecurityOptionsGroup());
   layout->addItem(new QSpacerItem(0, 0, QSizePolicy::Expanding, QSizePolicy::Expanding));
   setWidget(widget);
 
@@ -200,6 +207,8 @@ void NetworkWidget::CreateWidgets()
 
 void NetworkWidget::ConnectWidgets()
 {
+  connect(m_dump_format_combo, QOverload<int>::of(&QComboBox::currentIndexChanged), this,
+          &NetworkWidget::OnDumpFormatComboChanged);
   connect(m_dump_ssl_read_checkbox, &QCheckBox::stateChanged, [](int state) {
     Config::SetBaseOrCurrent(Config::MAIN_NETWORK_SSL_DUMP_READ, state == Qt::Checked);
   });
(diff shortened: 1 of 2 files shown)


lioncash:
Seems quite largely sized relative to everything else in the SSL options to the point that it makes them look a little off kilter/unbalanced visually.

sepalani:
I added a size policy to prevent the button to fill the whole space:
image

leoetlino:
This still looks a bit odd IMO... also buttons are likely to be taller on other platforms, which would possibly result in inconsistent line height?

leoetlino:
So I tested this on Linux and the layout looks the same as on Windows, so no issues with line height. I'm still not a big fan of the UI but I can't think of any better alternative when you only have *one* button.

@lioncash any suggestions, or do you think the current layout is fine (now that the button width is fixed)?

sepalani:
@leoetlino @lioncash 
I changed this PR a bit since the PCAP PRs were merged. I added a combo box to select the network dump format (and greyed options when not available depending on the dump format).

image

leoetlino:
Why do we need separate options for SSL read/write, SSL read and SSL write? Wouldn't a simple SSL option suffice, considering the user can already choose to dump reads or writes or both?

sepalani:
The main issue I have, is that the SSL dump format is a legacy format that can only dump decrypted IOS SSL and was used to dump decrypted HTTPS messages in 2 text files. It was coded as 2 define which got replaced with 2 config options instead. So I have no easy way from a UI/UX point of view to make the differences between the None option (dump disabled) and an hypothetical "Binary SSL" option with both decrypted read and write disabled. It also can't dump raw SSL like PCAP.

In sum, if I have a to create a single Binary SSL option I would have to:
 - Make the distinction between the `None` option and `Binary SSL` with both decrypted read/write disabled
 - Or alternatively, Make sure both read/write can't be disabled or switch to None option when it's the case

leoetlino:
Okay, that makes sense to me.

Source/Core/DolphinQt/Debugger/NetworkWidget.cpp
```diff
@@ -309,28 +337,98 @@ QGroupBox* NetworkWidget::CreateSSLContextGroup()
...
-QGroupBox* NetworkWidget::CreateSSLOptionsGroup()
+QGroupBox* NetworkWidget::CreateDumpOptionsGroup()
 {
-  QGroupBox* ssl_options_group = new QGroupBox(tr("SSL options"));
-  QGridLayout* ssl_options_layout = new QGridLayout;
-  ssl_options_group->setLayout(ssl_options_layout);
-
-  m_dump_ssl_read_checkbox = new QCheckBox(tr("Dump SSL read"));
-  m_dump_ssl_write_checkbox = new QCheckBox(tr("Dump SSL write"));
+  QGroupBox* dump_options_group = new QGroupBox(tr("Dump options"));
```
leoetlino, line 342:
```suggestion
  auto* dump_options_group = new QGroupBox(tr("Dump options"));
```

to avoid repeating yourself when the type name is already written out on the right hand side

Source/Core/DolphinQt/Debugger/NetworkWidget.cpp
```diff
@@ -309,28 +337,98 @@ QGroupBox* NetworkWidget::CreateSSLContextGroup()
...
+  QVBoxLayout* dump_options_layout = new QVBoxLayout;
```
leoetlino, line 343:
```suggestion
  auto* dump_options_layout = new QVBoxLayout;
```

Source/Core/DolphinQt/Debugger/NetworkWidget.cpp
```diff
@@ -309,28 +337,98 @@ QGroupBox* NetworkWidget::CreateSSLContextGroup()
...
+  dump_options_layout->addWidget(m_dump_peer_cert_checkbox);
+  dump_options_layout->addWidget(m_open_dump_folder);
+
+  dump_options_layout->setSpacing(1);
+  return dump_options_group;
+}
+
+QGroupBox* NetworkWidget::CreateSecurityOptionsGroup()
+{
+  QGroupBox* security_options_group = new QGroupBox(tr("Security options"));
```
leoetlino, line 379:
```suggestion
  auto* security_options_group = new QGroupBox(tr("Security options"));
```

Source/Core/DolphinQt/Debugger/NetworkWidget.cpp
```diff
@@ -309,28 +337,98 @@ QGroupBox* NetworkWidget::CreateSSLContextGroup()
...
+  QVBoxLayout* security_options_layout = new QVBoxLayout;
```
leoetlino, line 380:
```suggestion
  auto* security_options_layout = new QVBoxLayout;
```

Source/Core/DolphinQt/Debugger/NetworkWidget.cpp
```diff
@@ -309,28 +337,98 @@ QGroupBox* NetworkWidget::CreateSSLContextGroup()
...
   m_verify_certificates_checkbox = new QCheckBox(tr("Verify certificates"));
+  security_options_layout->addWidget(m_verify_certificates_checkbox);
+
+  security_options_layout->setSpacing(1);
+  return security_options_group;
+}
+
+QComboBox* NetworkWidget::CreateDumpFormatCombo()
+{
+  QComboBox* combo = new QComboBox();
```
leoetlino, line 392:
```suggestion
  auto* combo = new QComboBox();
```

sepalani:
@leoetlino Done.

leoetlino (approved):

PR #8780, opened 2020-04-29
Externals: Do not expose the externals directory itself as include path.

degasus:
We must not provide the /Externals directory as global include directory.
Here, this yield a crash because of external minizip header and system library mismatch.
The wrong usage was written for soundtouch, and many libraries afterwards have been linked wrong.

Soundtouch itself recormends to include it with <SoundTouch.h> and -I/usr/include/soundtouch, so this should fit better.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -755,6 +755,7 @@ endif()
 if(USE_DISCORD_PRESENCE)
   message(STATUS "Using static DiscordRPC from Externals")
   add_subdirectory(Externals/discord-rpc)
+  include_directories(Externals/discord-rpc/include)
 endif()
 
 find_package(Libsystemd)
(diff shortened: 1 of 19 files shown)


CMakeLists.txt
```diff
@@ -649,12 +649,13 @@ add_subdirectory(Externals/FreeSurround)
 if (APPLE OR WIN32)
   message(STATUS "Using ed25519 from Externals")
   add_subdirectory(Externals/ed25519)
+  include_directories(Externals/ed25519)
 endif()
 
 # Using static soundtouch from Externals
 # Unable to use system soundtouch library: We require shorts, not floats.
 add_subdirectory(Externals/soundtouch)
-include_directories(Externals)
```
degasus, line 657: This one line was the source of the evil.

Source/VSProps/Base.props
```diff
@@ -33,13 +33,14 @@
         Note: Directory containing source file being compiled is always searched first.
       -->
       <AdditionalIncludeDirectories>$(CoreDir);%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
-      <AdditionalIncludeDirectories>$(ExternalsDir);%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
```
degasus, line 36: This line is as eval as doing so on linux. However without any system library, it likely won't matter.

PR #8782, opened 2020-04-30
Explain how to fix "Serial and/or version data is missing"

Pokechu22:
The downloads page on redump doesn't link to the version that has serials and versions.  Since the Wii data needs to be manually downloaded, it's easy to download the version without them (if you have access to them, of course), and it's not immediately obvious how to get the right version.  I only found out by looking at the code that does the downloading for GC data.  This PR amends the message to explain how to include that data in the download.

### DIFF
--- a/Source/Core/DiscIO/VolumeVerifier.cpp
+++ b/Source/Core/DiscIO/VolumeVerifier.cpp
@@ -313,7 +313,11 @@ std::vector<RedumpVerifier::PotentialMatch> RedumpVerifier::ScanDatfile(const st
     // so show a panic alert rather than just using ERROR_LOG
 
     // i18n: "Serial" refers to serial numbers, e.g. RVL-RSBE-USA
-    PanicAlertT("Serial and/or version data is missing from %s", GetPathForSystem(system).c_str());
+    PanicAlertT("Serial and/or version data is missing from %s\n"
+                "Please append \"%s\" (without the quotes) to the datfile URL when downloading\n"
+                "Example: %s",
+                GetPathForSystem(system).c_str(), "serial,version",
+                "http://redump.org/datfile/gc/serial,version");
     m_result = {Status::Error, Common::GetStringT("Failed to parse Redump.org data")};
     return {};
   }


JosJuice:
The intent was that Redump would make the Wii data public relatively soon and nobody would need to download the datfile manually and put it in the Cache folder in an undocumented way, but... Well, apparently Redump isn't going to be making that public any time soon. So I guess doing this type of clarification is fine. However, I'm not sure how clear users will find the current wording. Also, you should add an i18n comment telling translators to leave "serial,version" exactly as-is, not even modifying punctuation or whitespace.

mbc07:
> The intent was that Redump would make the Wii data public relatively soon [...]

Redump waits at least one full year after the release of the last official game of a given platform before making the DAT file public. Thanks to Ubisoft not letting the Wii die with yearly Just Dance releases (Just Dance 2020 last year was the last, however), that means the Wii DAT file won't go public until at least late 2020 or early 2021...

JosJuice:
Do you know where you get the one year number from? What I recall is ten years, though I can't actually find where this was originally stated.

mbc07:
I won't find the exact source now but I remember stumbling over this "one year" rationale a few times when I was searching for the Wii DAT file in the past...

Pokechu22:
Even when the data is public, it'll be useful to have additional information present in the event that someone manually downloads it.

I'm not sure exactly how to make the message clearer; one possibility would be to add something like `(for instance, http://redump.org/datfile/gc/serial,version)` but that seems a bit awkward.

Source/Core/DiscIO/VolumeVerifier.cpp
```diff
@@ -313,7 +313,9 @@ std::vector<RedumpVerifier::PotentialMatch> RedumpVerifier::ScanDatfile(const st
     // so show a panic alert rather than just using ERROR_LOG
 
     // i18n: "Serial" refers to serial numbers, e.g. RVL-RSBE-USA
-    PanicAlertT("Serial and/or version data is missing from %s", GetPathForSystem(system).c_str());
+    PanicAlertT("Serial and/or version data is missing from %s\n"
+                "Please append serial,version to the datfile URL when downloading",
```
leoetlino, line 317: Maybe:

```suggestion
                "Please append "%s" (without the quotes) to the datfile URL when downloading",
```

passing `serial,version` as a format argument to make sure translators cannot mess up the string? I think adding an example might be a good idea as well.

Pokechu22: Done.

Source/Core/DiscIO/VolumeVerifier.cpp
```diff
@@ -313,7 +313,11 @@ std::vector<RedumpVerifier::PotentialMatch> RedumpVerifier::ScanDatfile(const st
     // so show a panic alert rather than just using ERROR_LOG
 
     // i18n: "Serial" refers to serial numbers, e.g. RVL-RSBE-USA
-    PanicAlertT("Serial and/or version data is missing from %s", GetPathForSystem(system).c_str());
+    PanicAlertT("Serial and/or version data is missing from %s\n"
+                "Please append \"%s\" (without the quotes) to the datfile URL when downloading\n"
+                "(For instance, %s)",
```
leoetlino, line 318: Nitpicking, but maybe just `Example: %s` (without the parentheses) would be sufficient, since it's on its own line?

leoetlino (approved):

PR #8783, opened 2020-05-01
DolphinQt: Add some i18n comments for recently added strings

JosJuice:


Source/Core/DolphinQt/Settings/AdvancedPane.cpp
```diff
@@ -127,6 +127,10 @@ void AdvancedPane::CreateLayout()
   mem2_override_slider_layout->addWidget(m_mem2_override_slider_label);
 
   auto* ram_override_description =
+      // i18n: "Retail titles should be optimized for the retail memory limitations" is a
```
leoetlino, line 130: Hmm, maybe it's worth updating the string and clarifying the message then? I think this new wording is clearer.

JosJuice: Done.

PR #8785, opened 2020-05-01
DolphinQt: Split Config > Advanced into Misc and Advanced

JosJuice:
It has gotten too tall.

Before:

image

After:

image

image

Suggestions for alternative names for the two tabs are welcome. I was considering "Hardware" or "Specs" for the one with the CPU clock override and the memory override, since the tab contains settings that change the specifications of the emulated hardware, but maybe naming the tab like that could make people think that the tab is about the hardware/specs of the host computer...

### DIFF
--- a/Source/Core/DolphinQt/CMakeLists.txt
+++ b/Source/Core/DolphinQt/CMakeLists.txt
@@ -343,6 +343,8 @@ add_executable(dolphin-emu
   Settings/GeneralPane.h
   Settings/InterfacePane.cpp
   Settings/InterfacePane.h
+  Settings/MiscPane.cpp
+  Settings/MiscPane.h
   Settings/PathPane.cpp
   Settings/PathPane.h
   Settings/USBDeviceAddToWhitelistDialog.cpp
--- a/Source/Core/DolphinQt/Config/SettingsWindow.cpp
+++ b/Source/Core/DolphinQt/Config/SettingsWindow.cpp
@@ -16,6 +16,7 @@
 #include "DolphinQt/Settings/GameCubePane.h"
 #include "DolphinQt/Settings/GeneralPane.h"
 #include "DolphinQt/Settings/InterfacePane.h"
+#include "DolphinQt/Settings/MiscPane.h"
 #include "DolphinQt/Settings/PathPane.h"
 #include "DolphinQt/Settings/WiiPane.h"
 
@@ -40,6 +41,7 @@ SettingsWindow::SettingsWindow(QWidget* parent) : QDialog(parent)
   m_tab_widget->addTab(GetWrappedWidget(new PathPane, this, 125, 100), tr("Paths"));
   m_tab_widget->addTab(GetWrappedWidget(new GameCubePane, this, 125, 100), tr("GameCube"));
   m_tab_widget->addTab(GetWrappedWidget(new WiiPane, this, 125, 100), tr("Wii"));
+  m_tab_widget->addTab(GetWrappedWidget(new MiscPane, this, 125, 200), tr("Misc"));
   m_tab_widget->addTab(GetWrappedWidget(new AdvancedPane, this, 125, 200), tr("Advanced"));
 
   // Dialog box buttons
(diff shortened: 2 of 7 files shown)


iwubcode:
Do we ever expect more Memory changes?

We could do:

* CPU (with both CPU options)
* Memory
* Time (or "Misc")?

Just brainstorming.  I don't have any strong opinions

JosJuice:
I'm not expecting any particular changes to memory, but we might get GPU clock and disc drive speed overrides at some point. I think it would make sense to keep those two together with the CPU clock.

iwubcode:
Maybe it is just me but it seems a little arbitrary how those tabs are named.

One option is instead of adding new tabs, we need a new container that can be expanded?  Not sure if Qt has them.  But basically a little arrow with a title that when clicked expands to the full content.

Then we could have CPU, Memory, GPU, Time, categories?

Or we could adopt a double "tree" kinda configuration like Citra has (list box on left, tabs on right).  Though this is the only one with multiple tabs but maybe we could shuffle things around..

Or we could just keep the large tab.

Sorry, don't have any genius ideas!

mbc07:
The current proposal looks good for me. About the "list box on the left, tabs on the right" approach, I think it's doable and would even let we integrate other separate config windows like the graphics settings on the main window, but it would require far more work.

For grouping, we could do the following options on the list box on the left, that would display the tabs under the parenthesis on the right when selected: Emulator (General, Interface, Paths), Console (GameCube, Wii), Audio (just Audio*), Graphics (all tabs from the Graphics window), Input (all content from the Controller window**) Advanced (Advanced and the new Misc Tab from this PR).

As this would merge the graphics tab on the main window, it would also be good to inherit the "description on hover" box for all categories, which would allow explaining better what every option does, not just the Graphics options. Again, just brainstorming a possible redesign, it surely wouldn't be feasible to do all of that from one day to another...

---
_* perhaps we should disable the tab header on this category, as it would have only one tab._

_** perhaps split the current Controller window into some tabs, as it's also pretty tall nowadays, and put a button to open Hotkey settings somewhere. Also, the pop-up windows that opens when configuring emulated GC/Wii controllers or hotkeys should remain as standalone windows. They're already pretty crowded with several tabs each and I can't think of any way for shoving all of that on the main window without becoming a huge and confusing mess._

delroth:
As a reminder for those who might not be aware: the reason why Graphics/Audio/Input settings are in a separate window isn't necessarily because people thought it was good UI design. It dates back to the plugin system (pre-3.0), when plugins would render their own settings window and thus couldn't easily have all their settings in one unified location.

Though I'm sure the UI design was a consideration at the time, the tradeoff would likely look extremely different now.

JosJuice:
@dolphin-emu-bot rebuild

JMC47:
I'm in favor of splitting things up more or maybe going to @iwubcode's solution because things are getting a ridiculous in all senses.  On my old laptop, I can't fit any of the windows.

MayImilae:
I still think the better approach to this is just let advanced scroll. It's advanced, it can scroll, it's fine!

mbc07:
The problem is that the window will try to open with the maximum possible height before it adds the scroll bar, and it's not uncommon for the buttons at the bottom to get cropped out as a result of that behavior. Sure, it can be resized, but it's a bit annoying to do when you can't see all borders and it doesn't remember the previous set size, so you'll have to resize it every time.

Introducing the Misc tab would also provide a new place for settings that aren't exactly advanced but that get thrown there anyway. IMHO the Advanced tab currently feels like a catch-all place and while that works, it's not ideal...

MayImilae:
Make the automatic height ignore the advanced tab, or set it to a fixed starting height that accommodates the other windows and ignores the advanced tab. Easy~

Plus horizontal tab scrolling is in a tiny window like this is **awful.**

Honestly what I really want to do is figure out a better way to handle the text in the window. But that will require a rethink. I'll get to it eventually. I'm not against a bodge in the mean time but I don't think splitting it up is the best option.

jordan-woyak:
I don't necessarily like everything being thrown under "Advanced", but I don't know if I like the name "Misc" either.
It's not very specific so it effectively just becomes "Advanced 1" and "Advanced 2".

I think the "Advanced" scrolling is okay, and better than this tab scrolling :P.
image

PR #8786, opened 2020-05-01
DolphinQt: Store name of user style, not path

JosJuice:
This makes user styles not break when for instance moving a portable.txt install between different folders.

### DIFF
--- a/Source/Core/DolphinQt/Settings.cpp
+++ b/Source/Core/DolphinQt/Settings.cpp
@@ -7,6 +7,7 @@
 #include <QApplication>
 #include <QDir>
 #include <QFile>
+#include <QFileInfo>
 #include <QSize>
 
 #include "AudioCommon/AudioCommon.h"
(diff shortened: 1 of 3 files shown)


Source/Core/DolphinQt/Settings/InterfacePane.cpp
```diff
@@ -216,7 +214,7 @@ void InterfacePane::LoadConfig()
       m_combobox_theme->findText(QString::fromStdString(SConfig::GetInstance().theme_name)));
 
   const QString userstyle = Settings::Instance().GetCurrentUserStyle();
-  const int index = m_combobox_userstyle->findText(QFileInfo(userstyle).baseName());
+  const int index = m_combobox_userstyle->findText(QFileInfo(userstyle).completeBaseName());
```
iwubcode, line 217: I know little about this but is your thought that `baseName()` might be truncating the style a little too much, ex:

`mycoolstyle.123.css`

`baseName()` = `mycoolstyle`
`completeBaseName()` = `mycoolstyle.123`

JosJuice: Yes, exactly. Though I guess this `baseName` -> `completeBaseName` change isn't entirely related to the main point of this PR.

JosJuice: Actually, it seems like I forgot to push a change I made to this line... I'm going to do that right now.

iwubcode: Thanks.  I was also questioning it for the other line that uses `completeBaseName` but that explanation makes sense I think.

JMC47:
LGTM.  Tested

Source/Core/DolphinQt/Settings/InterfacePane.cpp
JosJuice: For what it's worth, for that other line, `completeBaseName` matches the behavior of the previously used `SplitPath` method.

Source/Core/DolphinQt/Settings.cpp
```diff
@@ -72,25 +73,34 @@ void Settings::SetThemeName(const QString& theme_name)
 
 QString Settings::GetCurrentUserStyle() const
 {
-  return GetQSettings().value(QStringLiteral("userstyle/path"), false).toString();
+  if (GetQSettings().contains(QStringLiteral("userstyle/name")))
+  {
+    return GetQSettings().value(QStringLiteral("userstyle/name")).toString();
+  }
+  else
```
leoetlino, line 80: else isn't necessary after a return

PR #8788, opened 2020-05-02
NetPlay: Reuse WiiSave code for save syncing

leoetlino:
It's simpler to just reuse the existing WiiSave serialization
and deserialization code and treat the saves as binary blobs,
especially when netplay doesn't actually look at any of the information
in the WiiSave structures.

To ensure save blobs can be compressed well, this adds an option to
WiiSave to disable encryption.

Bonus: we don't need to hold every single save file in memory at the
same time, and compressing entire save blobs will probably result in
smaller sizes compared to compressing each file separately.

### DIFF
--- a/Source/Core/Common/FileUtil.cpp
+++ b/Source/Core/Common/FileUtil.cpp
@@ -613,6 +613,19 @@ bool SetCurrentDir(const std::string& directory)
   return __chdir(directory.c_str()) == 0;
 }
 
+ScopedTemporaryDirectory::ScopedTemporaryDirectory()
+{
+  m_path = CreateTempDir();
+  ASSERT(!m_path.empty());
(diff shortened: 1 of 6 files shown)


JMC47:
I can confirm that Sync All Wii Saves functions as *currently* expected with this build.

Techjar:
Doesn't this break writing the Mii data to the temporary FS?

leoetlino:
I don't think so -- the Mii data syncing code is separate and this PR doesn't touch it at all

Techjar:
Oh yeah, up where it says `// Read the Mii data`, forgot about that. All good then.

leoetlino:
Forget this PR, I'm going to use another approach for save syncing.

PR #8789, opened 2020-05-02
Fix Windows CMake build errors

JosJuice:
Lambda expressions with uncaptured constants were leading to errors, and there were also some warnings about deprecated functions (QFontMetrics::width and inet_ntoa).

### DIFF
--- a/Source/Core/DolphinQt/Debugger/CodeViewWidget.cpp
+++ b/Source/Core/DolphinQt/Debugger/CodeViewWidget.cpp
@@ -194,11 +194,20 @@ void CodeViewWidget::FontBasedSizing()
   constexpr int extra_text_width = 8;
 
   const QFontMetrics fm(Settings::Instance().GetDebugFont());
+
+  const auto width = [&fm](QString text) {
+#if (QT_VERSION >= QT_VERSION_CHECK(5, 11, 0))
+    return fm.horizontalAdvance(text);
+#else
+    return fm.width(text);
+#endif
+  };
+
   const int rowh = fm.height() + 1;
   verticalHeader()->setMaximumSectionSize(rowh);
   horizontalHeader()->setMinimumSectionSize(rowh + 5);
   setColumnWidth(CODE_VIEW_COLUMN_BREAKPOINT, rowh + 5);
-  setColumnWidth(CODE_VIEW_COLUMN_ADDRESS, fm.width(QStringLiteral("80000000")) + extra_text_width);
+  setColumnWidth(CODE_VIEW_COLUMN_ADDRESS, width(QStringLiteral("80000000")) + extra_text_width);
 
   // The longest instruction is technically 'ps_merge00' (0x10000420u), but those instructions are
   // very rare and would needlessly increase the column size, so let's go with 'rlwinm.' instead.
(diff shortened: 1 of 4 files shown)


Source/Core/DolphinQt/Debugger/NetworkWidget.cpp
```diff
@@ -90,6 +90,16 @@ QTableWidgetItem* GetSocketState(s32 host_fd)
   return new QTableWidgetItem(QTableWidget::tr("Unbound"));
 }
 
+static QString GetAddressAndPort(const sockaddr_in& addr)
+{
+  char buffer[16];
+  const char* addr_str = inet_ntop(AF_INET, &addr.sin_addr, buffer, sizeof(buffer));
```
iwubcode, line 96: What's the advantage of this vs the `inet_ntoa` that you were using before?  Any concerns with the buffer not being large enough?

JosJuice: Visual Studio gives me a warning when I use `inet_ntoa`, and that warning gets turned into an error by our build settings. I don't know much beyond that, though I did find this comment when searching for more info about it: https://github.com/SFML/SFML/issues/1023#issuecomment-379424721

Source/Core/DolphinQt/Debugger/CodeViewWidget.cpp
```diff
@@ -194,11 +194,20 @@ void CodeViewWidget::FontBasedSizing()
   constexpr int extra_text_width = 8;
 
   const QFontMetrics fm(Settings::Instance().GetDebugFont());
+
+  const auto width = [&fm](QString text) {
```
leoetlino, line 198: Should this go into QtUtils? It looks like this function could be useful in other cases. (Also, any reason for passing text by value?)

JosJuice: > Should this go into QtUtils? It looks like this function could be useful in other cases.

It doesn't seem like we are using this anywhere else. Maybe we should move it if we decide to use this in more places.

> (Also, any reason for passing text by value?)

No particular reason. I suppose it's neither less efficient nor more efficient (at least by any measurable amount).

Source/Core/DolphinQt/Debugger/NetworkWidget.cpp
```diff
@@ -100,18 +110,19 @@ QTableWidgetItem* GetSocketName(s32 host_fd)
...
   sockaddr_in peer_addr;
   socklen_t peer_addr_len = sizeof(sockaddr_in);
   if (getpeername(host_fd, reinterpret_cast<sockaddr*>(&peer_addr), &peer_addr_len) != 0)
     return new QTableWidgetItem(sock_name);
 
-  const QString peer_name = QStringLiteral("%1:%2")
-                                .arg(QString::fromLatin1(inet_ntoa(peer_addr.sin_addr)))
-                                .arg(ntohs(peer_addr.sin_port));
+  const QString peer_name = GetAddressAndPort(peer_addr);
+  if (peer_name.isNull())
```
leoetlino, line 123: isEmpty

PR #8791, opened 2020-05-03
VideoCommon: Break build dependency cycle

phire:
In cmake, VideoCommon was depending on it's video
backends by depending on all of core.
Not only was this a cycle, with the video backends
depending on VideoCommon again, but VideoCommon
really shouldn't be depending on core.

So we move the initialization of the video backends into UICommon


phire:
On second thoughts, looks like VideoCommon does actually have quite a large dependency on Core.

Weird that it builds locally.

PR #8794, opened 2020-05-05
[DO NOT MERGE] FifoPlayer: Optimization

iwubcode:
This is currently in a broken state.

Pushing it up for @phire to give feedback after our discussion.

------------------

phire - is this somewhat what you were thinking?  Currently doesn't seem to have much performance benefit from looking at the fifolog dump.

This also currently crashes on load.  I think that's because the data-ptr in the opcode decoding isn't actually the base address but something else.  Any thoughts here?

### DIFF
--- a/Source/Core/Core/FifoPlayer/FifoAnalyzer.cpp
+++ b/Source/Core/Core/FifoPlayer/FifoAnalyzer.cpp
@@ -4,9 +4,13 @@
 
 #include "Core/FifoPlayer/FifoAnalyzer.h"
 
+#include <map>
+#include <memory>
 #include <numeric>
+#include <set>
 
 #include "Common/Assert.h"
+#include "Common/Hash.h"
 #include "Common/MsgHandler.h"
 #include "Common/Swap.h"
 
@@ -20,6 +24,9 @@
 
 namespace FifoAnalyzer
 {
+static std::map<u64, std::unique_ptr<FifoDrawCacheEntry>> hash_to_entry = {};
+static std::map<const u8*, FifoDrawCacheEntry*> data_to_entry = {};
+
 namespace
 {
 u8 ReadFifo8(const u8*& data)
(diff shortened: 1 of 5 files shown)


Source/Core/Core/FifoPlayer/FifoAnalyzer.cpp
```diff
@@ -300,4 +332,15 @@ void LoadCPReg(u32 subCmd, u32 value, CPMemory& cpMem)
...
   }
 }
+
+void ClearDrawCache()
+{
+  hash_to_entry.clear();
+  data_to_entry.clear();
+}
+
+FifoDrawCacheEntry* GetDrawCacheEntry(const u8* data)
```
iwubcode, line 342: This klunkyness is because the 'Analyze' function is used by both the Recorder and the Player.  I was trying not to impact the player if possible.

Source/Core/Core/FifoPlayer/FifoAnalyzer.cpp
```diff
@@ -219,6 +233,14 @@ u32 AnalyzeCommand(const u8* data, DecodeMode mode)
     break;
   }
 
+  case OpcodeDecoder::DOLPHIN_REPLAY_DATA:
```
iwubcode, line 236: To be honest when initially coding this up, I wasn't expecting this to be hit, so I wasn't sure what to do when it was.

Source/Core/Core/FifoPlayer/FifoAnalyzer.cpp
```diff
@@ -247,9 +270,18 @@ u32 AnalyzeCommand(const u8* data, DecodeMode mode)
           FifoRecordAnalyzer::WriteVertexArray(static_cast<int>(i), data + offsets[i], vertexSize,
                                                numVertices);
         }
+
+        auto [iter, inserted] = hash_to_entry.try_emplace(Common::GetHash64(data, datasize, 0),
```
phire, line 274: The VAT format (and anything else `CalculateVertexElementSizes` reads) really should be incorporated into this hash.

PR #8795, opened 2020-05-06
InputCommon: Use the Window handle when initializing DirectInput.

WamWooWam:
Previously, when initializing the Win32 input layer (and as a result DirectInput), Dolphin would use the handle of the render widget, which isn't updated when `ControllerInterface::ChangeWindow` is called. This had the annoying side effect of making all mouse input to Dolphin on Windows relative to the main Dolphin window, not the actual render target or render window. 

This PR makes sure the Win32 controller interface is initialized using the actual window handle, which in turn makes mouse input relative to the render target itself, not the main window.

### DIFF
--- a/Source/Core/InputCommon/ControllerInterface/ControllerInterface.cpp
+++ b/Source/Core/InputCommon/ControllerInterface/ControllerInterface.cpp
@@ -105,18 +105,17 @@ void ControllerInterface::RefreshDevices()
   InvokeDevicesChangedCallbacks();
 
 #ifdef CIFACE_USE_WIN32
-  // Should this be changed for other platforms? I don't have the means to test...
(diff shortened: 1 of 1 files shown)


JMC47:
@dolphin-emu-bot rebuild
@jordan-woyak please review.

BhaaLseN:
Maybe @stenzek could also take a look, since #8666 changed this.

WamWooWam:
Should be all done 👍

Filoppi:
This seems to have fixed a bug where about 1 times out of 4 the game windows would not receive any input (mouse/keyboard).

pizuz:
...and it broke MoltenVK on macOS... again...

https://bugs.dolphin-emu.org/issues/12107

PR #8796, opened 2020-05-06
Socket: Fix ENOTCONN error code

sepalani:
For some reason, Dolphin returns `EAGAIN` rather than `ENOTCONN`.

In some circumstances, it creates an infinite loop. For instance, using `send` with a socket which is not connected.

Is there any reason to keep this legacy value? As far as I know, `ENOTCONN` is only returned by the following functions:
- `recv`, `recvfrom`
- `send`, `sendto`
- `shutdown`

Ready to be reviewed.

### DIFF
--- a/Source/Core/Core/IOS/Network/Socket.cpp
+++ b/Source/Core/Core/IOS/Network/Socket.cpp
@@ -51,7 +51,12 @@ char* WiiSockMan::DecodeError(s32 ErrorCode)
 #endif
 }
 
-static s32 TranslateErrorCode(s32 native_error, bool isRW)
+// The following functions can return
+//  - EAGAIN / EWOULDBLOCK: send(to), recv(from), accept
+//  - EINPROGRESS: connect, bind
+//  - WSAEWOULDBLOCK: send(to), recv(from), accept, connect
+// On Windows is_rw is used to correct the return value for connect
+static s32 TranslateErrorCode(s32 native_error, bool is_rw)
 {
   switch (native_error)
   {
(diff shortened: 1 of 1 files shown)


Source/Core/Core/IOS/Network/Socket.cpp
```diff
@@ -50,7 +50,7 @@ char* WiiSockMan::DecodeError(s32 ErrorCode)
 #endif
 }
 
-static s32 TranslateErrorCode(s32 native_error, bool isRW)
+static s32 TranslateErrorCode(s32 native_error, bool is_rw)
```
iwubcode, line 53: I know nothing about this code.  Is `is_rw` mean "read"/"write"?

sepalani: It refers to read/write or rather `send(to)` and `recv(from)`.

The following functions can return
 - **EAGAIN**/**EWOULDBLOCK**:  `send(to)`, `recv(from)`, `accept`
 - **EINPROGRESS**: `connect`, `bind`
 - **WSAEWOULDBLOCK**: `send(to)`, `recv(from)`, `accept`, `connect`

It is used below to correct windows error codes **WSAEWOULDBLOCK** as it can be used in functions which don't return **EAGAIN**/**EWOULDBLOCK** but **EINPROGRESS** on Linux. Basically, the Linux code path will always return `-SO_EAGAIN`


I can add a comment about it if you want.

BhaaLseN: I think it might be more the naming than the usage. Not too sure if spelling out the words would help tho...

sepalani:
Rebased on latest master, I also added a comment regarding `is_rw`.

PR #8799, opened 2020-05-08
Android: Consistent TODO formatting

Ebola16:
- Removed an apparently outdated TODO in PlatformPagerAdapter
- Formatted the TODOs to fully display in Android Studio's TODO panel.

### DIFF
--- a/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/activities/AppLinkActivity.java
+++ b/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/activities/AppLinkActivity.java
@@ -130,8 +130,8 @@ public class AppLinkActivity extends FragmentActivity
 
   private void tryPlay(AppLinkHelper.PlayAction action)
   {
-    // TODO: This approach of getting the game from the game file cache without rescanning
-    // the library means that we can fail to launch games if the cache file has been deleted.
+    // TODO: This approach of getting the game from the game file cache without rescanning the
+    //       library means that we can fail to launch games if the cache file has been deleted.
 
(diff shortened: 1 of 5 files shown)


Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/activities/AppLinkActivity.java
```diff
@@ -131,7 +131,7 @@ private void browse()
   private void tryPlay(AppLinkHelper.PlayAction action)
   {
     // TODO: This approach of getting the game from the game file cache without rescanning
-    // the library means that we can fail to launch games if the cache file has been deleted.
+    //  the library means that we can fail to launch games if the cache file has been deleted.
```
BhaaLseN, line 134: Does it require _exactly_ two spaces, or _at least_ two?
I can see it guessing two comments next to each other if it were just one space; but having two spaces (compared to zero, one, three, ten, ...) feels inconsistent and weird for humans.

This is my personal opinion (and might not match what others think), but if it has to be indented _some way_ it should line up with the text that comes after `"TODO:"` - so in this case 7 spaces:
```java
    // TODO: This approach of getting the game from the game file cache without rescanning
    //       the library means that we can fail to launch games if the cache file has been deleted.
```

Thoughts?
_Note this shouldn't block a merge if nobody else chimes in or it ends up shedding the bikes in the wrong place._

Ebola16: Multi-line comments require at least two spaces. I just wanted the comments to properly display in Android Studio's TODO panel but I've implemented your suggested change. Let me know if there's anything else!

PR #8800, opened 2020-05-09
IOS/KD: Implement IOCTL_NWC24_REQUEST_SHUTDOWN

sepalani:
This PR implements IOCTL_NWC24_REQUEST_SHUTDOWN and renames some IOCTL names.

As far as I know, it's called during shutdown (when the stop button is pressed).

Based on IOS59 v9249.

Regarding the naming:
 - **IOCTL_SO_GETLASTERROR** comes from the fact it grabs the IOS SO thread error value
 - **IOCTLV_SO_STARTUP**/**IOCTLV_SO_CLEANUP** come from IOS KD strings
 - **IOCTL_SO_INITINTERFACE** educated guess since it's not SOStartup and initialises global variables only used by **SO INTERFACE** ioctl(v)s.

### DIFF
--- a/Source/Core/Core/IOS/Network/KD/NetKDRequest.cpp
+++ b/Source/Core/Core/IOS/Network/KD/NetKDRequest.cpp
@@ -141,9 +141,29 @@ IPCCommandResult NetKDRequest::IOCtl(const IOCtlRequest& request)
     break;
 
   case IOCTL_NWC24_REQUEST_SHUTDOWN:
-    // if ya set the IOS version to a very high value this happens ...
-    INFO_LOG(IOS_WC24, "NET_KD_REQ: IOCTL_NWC24_REQUEST_SHUTDOWN - NI");
+  {
+    if (request.buffer_in == 0 || request.buffer_in % 4 != 0 || request.buffer_in_size < 8 ||
+        request.buffer_out == 0 || request.buffer_out % 4 != 0 || request.buffer_out_size < 4)
+    {
+      return_value = IPC_EINVAL;
+      ERROR_LOG(IOS_WC24, "NET_KD_REQ: IOCTL_NWC24_REQUEST_SHUTDOWN = IPC_EINVAL");
+      break;
+    }
+
+    INFO_LOG(IOS_WC24, "NET_KD_REQ: IOCTL_NWC24_REQUEST_SHUTDOWN");
+    [[maybe_unused]] const u32 event = Memory::Read_U32(request.buffer_in);
+    // TODO: Advertise shutdown event
+    // TODO: Shutdown USB keyboard LEDs if event == 3
+    // TODO: IOCTLV_NCD_SETCONFIG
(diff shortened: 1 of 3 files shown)


Source/Core/Core/IOS/Network/KD/NetKDRequest.cpp
```diff
@@ -141,9 +141,32 @@ IPCCommandResult NetKDRequest::IOCtl(const IOCtlRequest& request)
...
+    if (request.buffer_in == 0 || (request.buffer_in & 3) != 0 || request.buffer_in_size < 8 ||
+        request.buffer_out == 0 || (request.buffer_out & 3) != 0 || request.buffer_out_size < 4)
+    {
+      return_value = IPC_EINVAL;
+      ERROR_LOG(IOS_WC24, "NET_KD_REQ: IOCTL_NWC24_REQUEST_SHUTDOWN = IPC_EINVAL");
+      break;
+    }
+
+    INFO_LOG(IOS_WC24, "NET_KD_REQ: IOCTL_NWC24_REQUEST_SHUTDOWN");
+    u32 event = Memory::Read_U32(request.buffer_in);
```
lioncash, line 154:
```suggestion
    [[maybe_unused]] const u32 event = Memory::Read_U32(request.buffer_in);
```

Source/Core/Core/IOS/Network/KD/NetKDRequest.cpp
```diff
@@ -141,9 +141,32 @@ IPCCommandResult NetKDRequest::IOCtl(const IOCtlRequest& request)
...
+    // TODO: Advertise shutdown event
+    if (event == 3)
```
lioncash, line 156: You may wish to omit this and instead expand the comment that the LEDs should be shut down if `event == 3`. Static analysis tools will flag empty conditional bodies like these as false positives otherwise.

sepalani:
@lioncash 
Done.

Source/Core/Core/IOS/Network/KD/NetKDRequest.cpp
```diff
@@ -141,9 +141,29 @@ IPCCommandResult NetKDRequest::IOCtl(const IOCtlRequest& request)
     break;
 
   case IOCTL_NWC24_REQUEST_SHUTDOWN:
-    // if ya set the IOS version to a very high value this happens ...
-    INFO_LOG(IOS_WC24, "NET_KD_REQ: IOCTL_NWC24_REQUEST_SHUTDOWN - NI");
+  {
+    if (request.buffer_in == 0 || (request.buffer_in & 3) != 0 || request.buffer_in_size < 8 ||
```
leoetlino, line 145: While it's fairly clear what this is checking for, I'd suggest using `x % 4 != 0` rather than `(x & 3) != 0` to make it even more obvious (bonus: it removes the need for the parentheses)

sepalani:
@leoetlino 
Done.

PR #8802, opened 2020-05-09
WiiRoot: Fix empty files being created when source is missing

leoetlino:
CopyNandFile must not create empty files on the destination filesystem
if the source file doesn't exist.

Otherwise, this can lead to an empty Mii database being created in the
session Wii root if there's no database in the configured Wii root and
netplay or Movie is used -- that database would then be copied back to
the configured root, which causes games like MKW to complain about
corrupted Mii data even when the player has stopped using netplay.

### DIFF
--- a/Source/Core/Core/WiiRoot.cpp
+++ b/Source/Core/Core/WiiRoot.cpp
@@ -80,21 +80,21 @@ static void CopySave(FS::FileSystem* source, FS::FileSystem* dest, const u64 tit
 static bool CopyNandFile(FS::FileSystem* source_fs, const std::string& source_file,
                          FS::FileSystem* dest_fs, const std::string& dest_file)
 {
-  const auto last_slash = dest_file.find_last_of('/');
-  if (last_slash != std::string::npos && last_slash > 0)
-  {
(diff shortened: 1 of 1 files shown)


Source/Core/Core/WiiRoot.cpp
```diff
@@ -85,11 +85,15 @@ static bool CopyNandFile(FS::FileSystem* source_fs, const std::string& source_fi
 
   auto source_handle =
       source_fs->OpenFile(IOS::PID_KERNEL, IOS::PID_KERNEL, source_file, IOS::HLE::FS::Mode::Read);
+  // Do not create an empty file on the destination filesystem if the source file doesn't exist
+  if (!source_handle)
+    return false;
```
JosJuice, line 90: Returning false results in a warning being logged, even though this is a normal scenario. Do we want that?

JosJuice: Also, the `CreateFullPath` call should probably be moved to just before the `CreateAndOpenFile` call.

leoetlino: Good points, I've made it return true and moved the CreateFullPath call

PR #8803, opened 2020-05-09
Android: Android INI section and make platform tab selection an INI setting

Ebola16:
Implements https://bugs.dolphin-emu.org/issues/10793
Prerequisite for https://bugs.dolphin-emu.org/issues/10957

Android-specific settings should be moved away from SharedPreferences and use INI settings. This allows these settings to be retained across multiple builds of Dolphin. I added an Android section to Dolphin.ini to make these settings easy to find. Most of the setting changes will happen in future PRs.

Android: Make last platform tab selection an INI setting was a test case. Now Dolphin will remember which platform tab was selected when (re)installing or starting Dolphin.

### DIFF
--- a/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/ui/main/MainActivity.java
+++ b/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/ui/main/MainActivity.java
@@ -4,6 +4,7 @@ import android.content.Intent;
 import android.content.pm.PackageManager;
 import android.os.Bundle;
 
+import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 
 import com.google.android.material.floatingactionbutton.FloatingActionButton;
@@ -16,14 +17,17 @@ import androidx.appcompat.widget.Toolbar;
 import android.view.Menu;
 import android.view.MenuInflater;
 import android.view.MenuItem;
-import android.view.View;
 import android.widget.Toast;
 
+import org.dolphinemu.dolphinemu.NativeLibrary;
 import org.dolphinemu.dolphinemu.R;
 import org.dolphinemu.dolphinemu.activities.EmulationActivity;
 import org.dolphinemu.dolphinemu.adapters.PlatformPagerAdapter;
+import org.dolphinemu.dolphinemu.features.settings.model.Settings;
 import org.dolphinemu.dolphinemu.features.settings.ui.MenuTag;
 import org.dolphinemu.dolphinemu.features.settings.ui.SettingsActivity;
+import org.dolphinemu.dolphinemu.features.settings.utils.SettingsFile;
+import org.dolphinemu.dolphinemu.utils.AfterDirectoryInitializationRunner;
 import org.dolphinemu.dolphinemu.utils.DirectoryInitialization;
 import org.dolphinemu.dolphinemu.services.GameFileCacheService;
 import org.dolphinemu.dolphinemu.ui.platform.Platform;
@@ -55,8 +59,6 @@ public final class MainActivity extends AppCompatActivity implements MainView
 
     setSupportActionBar(mToolbar);
 
-    mTabLayout.setupWithViewPager(mViewPager);
-
     // Set up the FAB.
     mFab.setOnClickListener(view -> mPresenter.onFabClick());
 
@@ -68,16 +70,8 @@ public final class MainActivity extends AppCompatActivity implements MainView
 
     if (PermissionsHandler.hasWriteAccess(this))
     {
-      PlatformPagerAdapter platformPagerAdapter = new PlatformPagerAdapter(
-              getSupportFragmentManager(), this);
-      mViewPager.setAdapter(platformPagerAdapter);
-      mViewPager.setOffscreenPageLimit(platformPagerAdapter.getCount());
-      showGames();
-      GameFileCacheService.startLoad(this);
-    }
-    else
-    {
-      mViewPager.setVisibility(View.INVISIBLE);
+      new AfterDirectoryInitializationRunner()
+              .run(this, this::setPlatformTabsAndStartGameFileCacheService);
     }
   }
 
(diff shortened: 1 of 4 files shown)


JosJuice:
Adding a new INI section and putting the current tab selection in it makes sense, but why are you moving those two existing settings to the new INI section? For the savestates enabled setting, it is true that that setting currently only exists on Android, but moving it resets the setting, and it might make sense to make that setting available on non-Android in the future. For the DSP setting, I don't see how it would work correctly after the move.

Ebola16:
I can see the reasoning to leave the savestate setting alone.
I moved the DSP setting since the "master" setting is Android-specific. The logic based on how that "master" setting changes the universal DSP settings still works from my testing. Is this still a bad idea?

JosJuice:
> I moved the DSP setting since the "master" setting is Android-specific. The logic based on how that "master" setting changes the universal DSP settings still works from my testing. Is this a bad idea?

Oh, sorry, I thought it was the normal DSP setting and not a special Android one. I suppose it makes sense then (though maybe this setting actually shouldn't exist, if possible...)

Ebola16:
Reverted the savestate change and kept the DSP Android-specific changes. I didn't want to make changes to the way Dolphin handles DSP settings in a PR that made Android UI changes so that's for a future PR.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/ui/main/MainActivity.java
```diff
@@ -259,4 +243,33 @@ private PlatformGamesView getPlatformGamesView(Platform platform)
...
+      public void onTabSelected(@NonNull TabLayout.Tab tab)
+      {
+        super.onTabSelected(tab);
+        NativeLibrary
+                .SetConfig(SettingsFile.FILE_NAME_DOLPHIN + ".ini", Settings.SECTION_INI_ANDROID,
+                        SettingsFile.KEY_LAST_PLATFORM_TAB, Integer.toString(tab.getPosition()));
+      }
+    });
+
+    // Don't call NativeLibrary before DirectoryInitialization.
```
JosJuice, line 266: I don't understand why there is a comment like this in this specific position. It would probably make more sense to put a comment above this function that says this function shouldn't be called before directory initialization, so as to provide documentation for callers of this code.

And isn't this code in fact potentially running `NativeConfig.GetConfig` before directory initialization is done? `DirectoryInitialization.start` is asynchronous, so all you're guaranteeing by calling that first is that directory initialization has started, not that is has finished.

Ebola16: I was aware that my code was suspicious but from my testing I didn't run into any problems with DirectioryInitialization. I wanted to leave a comment to draw attention to that. But I just remembered that `AfterDirectoryInitializationRunner` exists so that should solve the concerns!

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/ui/main/MainActivity.java
```diff
@@ -68,16 +70,8 @@ protected void onCreate(Bundle savedInstanceState)
...
-      PlatformPagerAdapter platformPagerAdapter = new PlatformPagerAdapter(
-              getSupportFragmentManager(), this);
-      mViewPager.setAdapter(platformPagerAdapter);
-      mViewPager.setOffscreenPageLimit(platformPagerAdapter.getCount());
-      showGames();
-      GameFileCacheService.startLoad(this);
-    }
-    else
-    {
-      mViewPager.setVisibility(View.INVISIBLE);
```
JosJuice, line 80: Was removing this intentional?

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/ui/main/MainActivity.java
```diff
@@ -259,4 +246,33 @@ private PlatformGamesView getPlatformGamesView(Platform platform)
...
+                .SetConfig(SettingsFile.FILE_NAME_DOLPHIN + ".ini", Settings.SECTION_INI_ANDROID,
+                        SettingsFile.KEY_LAST_PLATFORM_TAB, Integer.toString(tab.getPosition()));
+      }
+    });
+
+    String platformTab = NativeLibrary
+            .GetConfig(SettingsFile.FILE_NAME_DOLPHIN + ".ini", Settings.SECTION_INI_ANDROID,
+                    SettingsFile.KEY_LAST_PLATFORM_TAB, "0");
+
+    mViewPager.setCurrentItem(Integer.parseInt(platformTab));
```
JosJuice, line 274: You should probably have a fallback here (e.g. selecting the first tab) if `parseInt` fails. You never know what kinds of weird things users might end up with in their INIs.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/ui/main/MainActivity.java
Ebola16: Yes, the original code would call `mTabLayout.setupWithViewPager(mViewPager)` before write permission was obtained and change visibility accordingly. It seemed simpler to call mTabLayout.setupWithViewPager(mViewPager) after write permission was obtained, which removes the need to change visibility.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/ui/main/MainActivity.java
Ebola16: Would a try-catch be best here? I know one should avoid using them if possible but it seems like that is a good choice here.

JosJuice: That sounds fine to me. But please catch `NumberFormatException` instead of `Exception`, so that it's clearer what the purpose of the try-catch is.

PR #8804, opened 2020-05-09
DolphinQt / InputCommon - Support multiple DSU servers

iwubcode:
This review allows support for multiple DSU servers, needed if you want to use a motion controller and a separate head tracking source (#8747)

image
image

Also, master made some assumptions about the name of the controller.  It basically copied what DS4Windows used, where the "Mode" dictated the controller name.  However, [this documentation](https://v1993.github.io/cemuhook-protocol/) is more flexible and is what is used in this PR.  This means mode is now just how much gyro functionality the server supports.  So where does the "name" of the device come from?

It is now based on the server's description for the server added in the "Add New DSU Server" configuration window.

image

### DIFF
--- a/Source/Core/DolphinQt/Config/ControllerInterface/DualShockUDPClientWidget.cpp
+++ b/Source/Core/DolphinQt/Config/ControllerInterface/DualShockUDPClientWidget.cpp
@@ -4,14 +4,16 @@
 
 #include "DolphinQt/Config/ControllerInterface/DualShockUDPClientWidget.h"
 
+#include <fmt/format.h>
+
+#include <QBoxLayout>
 #include <QCheckBox>
-#include <QGridLayout>
-#include <QGroupBox>
 #include <QLabel>
-#include <QLineEdit>
-#include <QSpinBox>
+#include <QListWidget>
+#include <QPushButton>
 
 #include "Common/Config/Config.h"
+#include "DolphinQt/Config/ControllerInterface/DualShockUDPClientAddServerDialog.h"
 #include "InputCommon/ControllerInterface/DualShockUDPClient/DualShockUDPClient.h"
 
 DualShockUDPClientWidget::DualShockUDPClientWidget()
@@ -22,19 +24,27 @@ DualShockUDPClientWidget::DualShockUDPClientWidget()
 
 void DualShockUDPClientWidget::CreateWidgets()
 {
-  auto* main_layout = new QGridLayout;
+  auto* main_layout = new QVBoxLayout;
 
-  m_server_enabled = new QCheckBox(tr("Enable"));
-  m_server_enabled->setChecked(Config::Get(ciface::DualShockUDPClient::Settings::SERVER_ENABLED));
+  m_servers_enabled = new QCheckBox(tr("Enable"));
+  m_servers_enabled->setChecked(Config::Get(ciface::DualShockUDPClient::Settings::SERVERS_ENABLED));
+  main_layout->addWidget(m_servers_enabled, 0, 0);
 
-  m_server_address = new QLineEdit(
-      QString::fromStdString(Config::Get(ciface::DualShockUDPClient::Settings::SERVER_ADDRESS)));
-  m_server_address->setEnabled(m_server_enabled->isChecked());
+  m_server_list = new QListWidget();
+  main_layout->addWidget(m_server_list);
 
-  m_server_port = new QSpinBox();
-  m_server_port->setMaximum(65535);
-  m_server_port->setValue(Config::Get(ciface::DualShockUDPClient::Settings::SERVER_PORT));
-  m_server_port->setEnabled(m_server_enabled->isChecked());
+  m_add_server = new QPushButton(tr("Add..."));
+  m_add_server->setEnabled(m_servers_enabled->isChecked());
+
+  m_remove_server = new QPushButton(tr("Remove"));
+  m_remove_server->setEnabled(m_servers_enabled->isChecked());
+
+  QHBoxLayout* hlayout = new QHBoxLayout;
+  hlayout->addStretch();
+  hlayout->addWidget(m_add_server);
+  hlayout->addWidget(m_remove_server);
+
+  main_layout->addItem(hlayout);
 
   auto* description =
       new QLabel(tr("DSU protocol enables the use of input and motion data from compatible "
@@ -46,34 +56,75 @@ void DualShockUDPClientWidget::CreateWidgets()
   description->setWordWrap(true);
   description->setTextInteractionFlags(Qt::TextBrowserInteraction);
   description->setOpenExternalLinks(true);
-
-  main_layout->addWidget(m_server_enabled, 1, 1);
-  main_layout->addWidget(new QLabel(tr("Server IP Address")), 2, 1);
-  main_layout->addWidget(m_server_address, 2, 2);
-  main_layout->addWidget(new QLabel(tr("Server Port")), 3, 1);
-  main_layout->addWidget(m_server_port, 3, 2);
-  main_layout->addWidget(description, 4, 1, 1, 2);
+  main_layout->addWidget(description);
 
   setLayout(main_layout);
+
+  RefreshServerList();
 }
 
 void DualShockUDPClientWidget::ConnectWidgets()
 {
-  connect(m_server_enabled, &QCheckBox::toggled, this, [this] {
-    bool checked = m_server_enabled->isChecked();
-    Config::SetBaseOrCurrent(ciface::DualShockUDPClient::Settings::SERVER_ENABLED, checked);
-    m_server_address->setEnabled(checked);
-    m_server_port->setEnabled(checked);
-  });
+  connect(m_servers_enabled, &QCheckBox::clicked, this,
+          &DualShockUDPClientWidget::OnServersToggled);
+  connect(m_add_server, &QPushButton::clicked, this, &DualShockUDPClientWidget::OnServerAdded);
+  connect(m_remove_server, &QPushButton::clicked, this, &DualShockUDPClientWidget::OnServerRemoved);
+}
 
-  connect(m_server_address, &QLineEdit::editingFinished, this, [this] {
-    Config::SetBaseOrCurrent(ciface::DualShockUDPClient::Settings::SERVER_ADDRESS,
-                             m_server_address->text().toStdString());
-  });
+void DualShockUDPClientWidget::RefreshServerList()
+{
+  m_server_list->clear();
 
-  connect(m_server_port, static_cast<void (QSpinBox::*)(int)>(&QSpinBox::valueChanged), this,
-          [this] {
-            Config::SetBaseOrCurrent(ciface::DualShockUDPClient::Settings::SERVER_PORT,
-                                     static_cast<u16>(m_server_port->value()));
-          });
+  const auto& servers_setting = Config::Get(ciface::DualShockUDPClient::Settings::SERVERS);
+  const auto server_details = SplitString(servers_setting, ';');
+  for (const std::string& server_detail : server_details)
+  {
+    const auto server_info = SplitString(server_detail, ':');
+    if (server_info.size() < 3)
+      continue;
+
+    QListWidgetItem* list_item = new QListWidgetItem(QString::fromStdString(
+        fmt::format("{}:{} - {}", server_info[1], server_info[2], server_info[0])));
+    m_server_list->addItem(list_item);
+  }
+  emit ConfigChanged();
+}
+
+void DualShockUDPClientWidget::OnServerAdded()
+{
+  DualShockUDPClientAddServerDialog add_server_dialog(this);
+  connect(&add_server_dialog, &DualShockUDPClientAddServerDialog::accepted, this,
+          &DualShockUDPClientWidget::RefreshServerList);
+  add_server_dialog.exec();
+}
+
+void DualShockUDPClientWidget::OnServerRemoved()
+{
+  const int row_to_remove = m_server_list->currentRow();
+
+  const auto& servers_setting = Config::Get(ciface::DualShockUDPClient::Settings::SERVERS);
+  const auto server_details = SplitString(servers_setting, ';');
+
+  std::string new_server_setting;
+  for (int i = 0; i < m_server_list->count(); i++)
+  {
+    if (i == row_to_remove)
+    {
+      continue;
+    }
+
+    new_server_setting += server_details[i] + ';';
+  }
+
+  Config::SetBaseOrCurrent(ciface::DualShockUDPClient::Settings::SERVERS, new_server_setting);
+
+  RefreshServerList();
+}
+
+void DualShockUDPClientWidget::OnServersToggled()
+{
+  bool checked = m_servers_enabled->isChecked();
+  Config::SetBaseOrCurrent(ciface::DualShockUDPClient::Settings::SERVERS_ENABLED, checked);
+  m_add_server->setEnabled(checked);
+  m_remove_server->setEnabled(checked);
 }
--- a/Source/Core/DolphinQt/Config/ControllerInterface/DualShockUDPClientWidget.cpp
+++ b/Source/Core/DolphinQt/Config/ControllerInterface/DualShockUDPClientWidget.cpp
@@ -75,6 +75,22 @@ void DualShockUDPClientWidget::RefreshServerList()
 {
   m_server_list->clear();
 
+  const auto server_address_setting =
+      Config::Get(ciface::DualShockUDPClient::Settings::SERVER_ADDRESS);
+  const auto server_port_setting = Config::Get(ciface::DualShockUDPClient::Settings::SERVER_PORT);
+
+  // Update our servers setting if the user is using old configuration
+  if (!server_address_setting.empty() && server_port_setting != 0)
+  {
+    const auto& servers_setting = Config::Get(ciface::DualShockUDPClient::Settings::SERVERS);
+    Config::SetBaseOrCurrent(ciface::DualShockUDPClient::Settings::SERVERS,
+                             servers_setting + fmt::format("{}:{}:{};", "DS4",
+                                                           server_address_setting,
+                                                           server_port_setting));
+    Config::SetBase(ciface::DualShockUDPClient::Settings::SERVER_ADDRESS, "");
+    Config::SetBase(ciface::DualShockUDPClient::Settings::SERVER_PORT, 0);
+  }
+
   const auto& servers_setting = Config::Get(ciface::DualShockUDPClient::Settings::SERVERS);
   const auto server_details = SplitString(servers_setting, ';');
   for (const std::string& server_detail : server_details)
--- a/Source/Core/DolphinQt/Config/ControllerInterface/ServerStringValidator.cpp
+++ b/Source/Core/DolphinQt/Config/ControllerInterface/ServerStringValidator.cpp
@@ -1,0 +1,23 @@
+// Copyright 2020 Dolphin Emulator Project
+// Licensed under GPLv2+
+// Refer to the license.txt file included.
+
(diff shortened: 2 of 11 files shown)


mbc07:
Testing this ASAP!

Source/Core/InputCommon/ControllerInterface/DualShockUDPClient/DualShockUDPClient.cpp
```diff
@@ -484,6 +484,11 @@ void Device::UpdateInput()
   {
     if (auto pad_data = msg.CheckAndCastTo<Proto::MessageType::PadDataResponse>())
     {
+      if (pad_data->pad_id != m_index)
+      {
+        continue;
```
jordan-woyak, line 489: While this probably fixes the issue, having each device instance read from the one socket and dropping irrelevant messages seems very wrong. You're going to have a much less responsive controller if each device is reading and dropping the other devices' data. There probably needs to be a single reader of the socket or messages need to be peeked at before they are processed.

iwubcode: Hmm, never thought of it that way.  Having a server specific reading socket makes a lot of sense.  I'll remove this and we can fix it proper later on.  I'm thinking you could have a socket per server that fans out to the devices.

mbc07:
So, I threw every compatible controller I have, simultaneously, and everything seems to work as expected! Good work! The controllers were a DualShock 4 (through DS4Windows), a DualShock 3 (through ScpToolkit), a Wii Remote Plus (through WiimoteHook) and an old Wii Remote with a separate Motion Plus attachment (also through WiimoteHook).

Every controller entry was correctly shown on Dolphin's device list after adding the DSU server addresses on the "Alternate Input Sources" window. Strangely enough, I couldn't reproduce [issue 12052](https://dolp.in/i12052), even through this PR shouldn't address it (yes, I checked the PR build I used was compiled after you reverted the commit that was supposed to address this issue from this PR). Every controller reported its own motion data, that reflected the movement of the matching physical controller. Haven't tried on master, though. Perhaps it could be an OS-specific issue?

About the UI, I have some nitpicks. The actual design doesn't seem consistent with anything else we currently have on Dolphin and looks a little bit crowded, there's no padding between the elements too. Perhaps a better approach could be adopting a style similar to the existing USB Passthrough and Game Directory boxes from the Main Settings. I did a mockup of this suggestion below.

Starting from the current design on master, get rid of the current "Server Address" and "Server Port" text inputs but keep the "Enable" checkbox. Once the checkbox is marked, change the state from a list box that shows the currently added servers, with the "Add" and "Remove" buttons aligned on the right, from disabled to enabled and allow the user to interact with it:

multi_dsu_main_mock 

Upon clicking on "Add...", open a new window, with the actual Name, IP and Port text fields, and the common Add and Cancel buttons (almost identical to the USB Passthrough whitelist window, minus the device list part):

multi_dsu_add_mock

-----
-----
These footnotes aren't directly related to this PR, but I think are worth mentioning anyway (I noticed them while testing this PR but they probably apply to master too):

* When connected to a controller that is reported as partial gyro (e.g. DualShock 3, Wiimote without Motion Plus), the Advanced Input Window glitches a little. The "Battery" reading which was supposed to be on the input list for DSU Clients since PR #8474 seems gone and appears only for XInput devices on this case:

    

    Also noticed that some DSU Servers (like WiimoteHook) seems to change between reporting a Full Gyro device (Wiimote with Motion Plus) and a Partial Gyro device (Wiimote without Motion Plus) on the fly if you attach/detach a standalone Motion Plus accessory, but Dolphin's DSU implementation seems to check the controller type only once and assumes it will never change at runtime. Only relaunching Dolphin seems to correct this.

* Although this PR enables labeling different servers with different names, the actual button bindings will still be referenced with DualShock specific names, like "Cross", "Triangle", "L1", "L2", and so on. Given there's a wide variety of DSU Servers nowadays, perhaps we should switch to generic names, like Button 1, Button 2, Left Trigger, Right Trigger and so on? This will definitely break existing profiles, but I think it's more doable to do this now, while the feature isn't widely known/used than later, I think. 

    Another possible approach I thought (but that I could see getting messy code-wise very quickly) would be to filter the label the user gave to that DSU Server and react accordingly. E.g. if the word "DualShock" was found on the server name the user entered, somehow display PlayStation button names on the GUI when interacting with that server but still treat them internally (on the INI and on the expression parser) as Button 1, Button 2, Left Trigger, Right Trigger and so on...

iwubcode:
@mbc07 - thank you very much for your detailed feedback.  For your UI comments, I have updated this PR to follow your mockups.  Good ideas!

As for your other comments, the battery issue seems like something that needs to be fixed.  I agree with your comment about the button names but not sure what the best solution would be.  At the moment I'd probably lean toward generic button names.  Unless the user could specify some translation when adding a server.

mbc07:
Switching the bindings to generic button names (probably not in the scope of this PR), in conjunction with this PR, should be sufficient for now. We can think of solutions for having button labels without hardcoding PlayStation-specific names on the bindings later, I think. 

Anyway, LGTM apart from the code comment I made...

_(I should probably open an issue report to avoid burying these notes under the PR discussion, especially the Advanced Input Window glitches)_

Source/Core/DolphinQt/Config/ControllerInterface/DualShockUDPClientWidget.cpp
```diff
@@ -24,17 +26,20 @@ void DualShockUDPClientWidget::CreateWidgets()
...
+
+  m_server_list = new QListWidget();
+  main_layout->addWidget(m_server_list, 1, 0);
 
-  m_server_address = new QLineEdit(
-      QString::fromStdString(Config::Get(ciface::DualShockUDPClient::Settings::SERVER_ADDRESS)));
-  m_server_address->setEnabled(m_server_enabled->isChecked());
+  m_add_server = new QPushButton(tr("Add..."));
+  m_add_server->setEnabled(m_servers_enabled->isChecked());
+  main_layout->addWidget(m_add_server, 2, 0);
```
mbc07, line 38: I haven't checked if this compiles, but adding the buttons to an QHBoxLayout and then adding that QHBoxLayout to the main layout instead of adding the buttons directly to the main layout should fix the weird placement of the "Remove" button and produce similar results to the existing Game Config and USB Whitelist panels. Something along those lines: 
```suggestion
  QHBoxLayout* hlayout = new QHBoxLayout;
  hlayout->addStretch();

  hlayout->addWidget(m_add_server);
  hlayout->addWidget(m_remove_server);
  
  main_layout->addLayout(hlayout);
```

iwubcode: Updated, better?

mbc07: Yes, much better. LGTM

Source/Core/DolphinQt/Config/ControllerInterface/DualShockUDPClientWidget.cpp
```diff
@@ -46,34 +56,75 @@ void DualShockUDPClientWidget::CreateWidgets()
...
+  connect(m_add_server, &QPushButton::clicked, this, &DualShockUDPClientWidget::OnServerAdded);
+  connect(m_remove_server, &QPushButton::clicked, this, &DualShockUDPClientWidget::OnServerRemoved);
+}
+
+void DualShockUDPClientWidget::RefreshServerList()
+{
+  m_server_list->clear();
+
+  const auto& servers_setting = Config::Get(ciface::DualShockUDPClient::Settings::SERVERS);
+  const auto& server_details = SplitString(servers_setting, ';');
```
lioncash, line 79:
```suggestion
  const auto server_details = SplitString(servers_setting, ';');
```

Since SplitString returns by value

Source/Core/DolphinQt/Config/ControllerInterface/DualShockUDPClientWidget.cpp
```diff
@@ -46,34 +56,75 @@ void DualShockUDPClientWidget::CreateWidgets()
...
+  for (const std::string& server_detail : server_details)
+  {
+    const auto& server_info = SplitString(server_detail, ':');
```
lioncash, line 82: Ditto

Source/Core/DolphinQt/Config/ControllerInterface/DualShockUDPClientWidget.cpp
```diff
@@ -46,34 +56,75 @@ void DualShockUDPClientWidget::CreateWidgets()
...
+    if (server_info.size() < 3)
+      continue;
+
+    QListWidgetItem* list_item = new QListWidgetItem(QString::fromStdString(
+        fmt::format("{}:{} - {}", server_info[1], server_info[2], server_info[0])));
```
lioncash, line 87:
```cpp
QStringLiteral("%1:%2 - %3").arg(server_info[1], server_info[2], server_info[0])
```

Will allow this to be done without a need to convert from std::string.

Source/Core/DolphinQt/Config/ControllerInterface/DualShockUDPClientWidget.cpp
```diff
@@ -46,34 +56,75 @@ void DualShockUDPClientWidget::CreateWidgets()
...
+}
+
+void DualShockUDPClientWidget::OnServerRemoved()
+{
+  const int row_to_remove = m_server_list->currentRow();
+
+  const auto& servers_setting = Config::Get(ciface::DualShockUDPClient::Settings::SERVERS);
+  const auto& server_details = SplitString(servers_setting, ';');
+
+  std::string new_server_setting = "";
```
lioncash, line 108:
```suggestion
  std::string new_server_setting;
```

Source/Core/DolphinQt/Config/ControllerInterface/DualShockUDPClientWidget.cpp
```diff
@@ -46,34 +56,75 @@ void DualShockUDPClientWidget::CreateWidgets()
...
+          &DualShockUDPClientWidget::RefreshServerList);
+  add_server_dialog.exec();
+}
+
+void DualShockUDPClientWidget::OnServerRemoved()
+{
+  const int row_to_remove = m_server_list->currentRow();
+
+  const auto& servers_setting = Config::Get(ciface::DualShockUDPClient::Settings::SERVERS);
+  const auto& server_details = SplitString(servers_setting, ';');
```
lioncash, line 106:
```suggestion
  const auto server_details = SplitString(servers_setting, ';');
```

Source/Core/DolphinQt/Config/ControllerInterface/DualShockUDPClientWidget.cpp
```diff
@@ -46,34 +56,75 @@ void DualShockUDPClientWidget::CreateWidgets()
...
+
+  std::string new_server_setting = "";
+  for (int i = 0; i < m_server_list->count(); i++)
+  {
+    if (i == row_to_remove)
+    {
+      continue;
+    }
+
+    new_server_setting += server_details[i] + ";";
```
lioncash, line 116:
```suggestion
    new_server_setting += server_details[i] + ';';
```

Source/Core/InputCommon/ControllerInterface/DualShockUDPClient/DualShockUDPClient.cpp
```diff
@@ -155,16 +154,40 @@ constexpr auto SERVER_LISTPORTS_INTERVAL = std::chrono::seconds{1};
...
-static std::string s_server_address;
-static u16 s_server_port;
+struct Server
+{
+  Server(std::string description, std::string address, u16 port)
+      : m_description{description}, m_address{address}, m_port{port}
+  {
+  }
+  Server(const Server&) = delete;
+  Server(Server&& other)
```
lioncash, line 164: Any reason this isn't just a defaulted move constructor?

If it absolutely needs to be user-provided, this should also be marked noexcept.

Source/Core/InputCommon/ControllerInterface/DualShockUDPClient/DualShockUDPClient.cpp
```diff
@@ -155,16 +154,40 @@ constexpr auto SERVER_LISTPORTS_INTERVAL = std::chrono::seconds{1};
 constexpr int TOUCH_X_AXIS_MAX = 1000;
 constexpr int TOUCH_Y_AXIS_MAX = 500;
 
-static bool s_server_enabled;
-static std::string s_server_address;
-static u16 s_server_port;
+struct Server
+{
+  Server(std::string description, std::string address, u16 port)
+      : m_description{description}, m_address{address}, m_port{port}
```
lioncash, line 160:
```suggestion
      : m_description{std::move(description)}, m_address{std::move(address)}, m_port{port}
```

Source/Core/InputCommon/ControllerInterface/DualShockUDPClient/DualShockUDPClient.cpp
```diff
@@ -198,39 +221,46 @@ static void HotplugThreadFunc()
...
+      {
+        // Request info on the four controller ports
+        Proto::Message<Proto::MessageType::ListPorts> msg(s_client_uid);
+        auto& list_ports = msg.m_message;
+        list_ports.pad_request_count = 4;
+        list_ports.pad_id = {0, 1, 2, 3};
+        msg.Finish();
+        if (server.m_socket.send(&list_ports, sizeof list_ports, server.m_address, server.m_port) !=
+            sf::Socket::Status::Done)
+          ERROR_LOG(SERIALINTERFACE, "DualShockUDPClient HotplugThreadFunc send failed");
```
lioncash, line 234: This should be braced, since the condition travels over more than one line.

Source/Core/InputCommon/ControllerInterface/DualShockUDPClient/DualShockUDPClient.cpp
```diff
@@ -270,29 +303,56 @@ static void Restart()
...
+
+  if (servers_enabled != s_servers_enabled || servers_setting != new_servers_setting)
   {
-    s_server_enabled = server_enabled;
-    s_server_address = server_address;
-    s_server_port = server_port;
+    s_servers_enabled = servers_enabled;
+    s_servers.clear();
+
+    const auto& server_details = SplitString(servers_setting, ';');
```
lioncash, line 338:
```suggestion
    const auto server_details = SplitString(servers_setting, ';');
```

Source/Core/InputCommon/ControllerInterface/DualShockUDPClient/DualShockUDPClient.cpp
```diff
@@ -270,29 +303,56 @@ static void Restart()
...
+    for (const auto& server_detail : server_details)
+    {
+      const auto& server_info = SplitString(server_detail, ':');
```
lioncash, line 341:
```suggestion
      const auto server_info = SplitString(server_detail, ':');
```

Source/Core/InputCommon/ControllerInterface/DualShockUDPClient/DualShockUDPClient.cpp
```diff
@@ -326,7 +389,9 @@ void DeInit()
   StopHotplugThread();
 }
 
-Device::Device(Proto::DsModel model, int index) : m_model{model}, m_index{index}
+Device::Device(const std::string& name, int index, const std::string& server_address,
+               u16 server_port)
+    : m_name{name}, m_index{index}, m_server_address{server_address}, m_server_port{server_port}
```
lioncash, line 394: `name` and `server_address` can be taken by value and `std::move`d into their respective members.

Source/Core/InputCommon/ControllerInterface/DualShockUDPClient/DualShockUDPClient.cpp
iwubcode: Due to there being a mutex, it couldn't be implicitly created.  The others are there because of rule of 5.  I'll mark it `noexcept` thanks!

Source/Core/DolphinQt/Config/ControllerInterface/DualShockUDPClientWidget.cpp
iwubcode: Tried this but wasn't working.  Seems like `arg` can't accept `std::string` arguments.  I could make this work if I use `fromStdString` 3 times but seems like that'd be worse?  Maybe I misunderstood you.  I'm not very familiar with the Qt library.

iwubcode:
Thank you for reviewing @lioncash !  I have made most of the changes you have requested and left a couple of comments.  Please let me know if you see anything else

Source/Core/InputCommon/ControllerInterface/DualShockUDPClient/DualShockUDPClient.cpp
rlnilsen: > ... having each device instance read from the one socket and dropping irrelevant messages seems very wrong. You're going to have a much less responsive controller if each device is reading and dropping the other devices' data.

Each device has its own socket connection to the server.  My UDP knowledge is extremely shaky, but I think each socket is bound to a separate local port. So I don't think reading one socket drops data on another socket. See my [comment](https://bugs.dolphin-emu.org/issues/12052#note-6) on the bug tracker for related info.

jordan-woyak: @rlnilsen Ah. If each device has its own connection then it should work. But this seems like a sloppy way of doing things, duplicating the bandwidth for each device for no reason? Or is the protocol smart enough to send data for specific devices?

rlnilsen: > Or is the protocol smart enough to send data for specific devices?

Exactly. Except the DSU Server used in the [bug report](https://bugs.dolphin-emu.org/issues/12052) that prompted this code change was not. It sent data from all connected controllers even if data from only one was requested.

iwubcode:
>Entering a : into the description box breaks things of course.
Entering a : at the end of the "IP Address" box actually crashes dolphin in a stoi call.
Maybe you should remove any colons from the user input?

Good points.  Let me play with that idea

>You make no attempt to load the old single-server configs?
I guess I don't really care, but someone might?

I don't.  I'm really bad about this.  Part of me wonders if we should have a global solution to handling this because of how often it comes up but I don't know what that would look like.

Let me see how hard it would be to support the old config.  I could probably give it some dummy description.

> An "Edit" button would be nice, but I'm nitpicking.

I did think of an edit button but just didn't get to it

rlnilsen:
It could be handy to have enable/disable per DSU server.

iwubcode:
@jordan-woyak - I added a change to address your concerns.  More could be done to validate the ip address input but I couldn't find anything I was happy with.  Decided to leave it with just the changes we discussed.  Let me know what you think.

@rlnilsen - thought of that but didn't have a UI that I was very happy with and did not want to complicate the review.  I was assuming add/remove was reasonable enough for most users.

Source/Core/DolphinQt/Config/ControllerInterface/ServerStringValidator.cpp
```diff
@@ -0,0 +1,23 @@
+// Copyright 2020 Dolphin Emulator Pro
```
rlnilsen, line 1: Is this the new Patreon supporters version? :-D

iwubcode: LOL.  I copied that from some other file... I'll fix it, thanks!

JMC47:
While this PR works for me, I'm not able to test it in a satisfactory manner that would let me merge it, nor do I understand the code well enough.

PR #8806, opened 2020-05-10
cmake: add an option to control dependencies vendoring

delroth:
Disable by default, to be used by distributions who care to try and
prefer system-wide libraries when available. It makes sense for us to
keep using vendored libs by default when possible to make it easier for
users to compile, but we should provide appropriate tools for distro to
figure out which dependencies they can share with the rest of the
system.

cc @ashkitten (I used the nixpkgs derivation to test this, and it's missing a few trivial deps that are already packaged)


ashkitten:
this is great, do you think you could make a corresponding pr to nixpkgs to fix the missing deps? if not i can do it when i have time

delroth:
I'll probably wait until the next beta when this PR ships, and then we can both fix deps + add the vendoring control flag.

Or we could just fix deps now and add the flag later, but I'm lazy :-)

delroth:
Tagging a few more maintainers as an FYI (and eventual feedback):

- @ThaChillera (for https://aur.archlinux.org/packages/dolphin-emu-beta-git)
- @Mystro256 (for https://src.fedoraproject.org/rpms/dolphin-emu)

Unrelated aside: if you aren't already aware, we recently started publicly advertising our "beta" versions on https://dolphin-emu.org/download/ -- so they might be a good target to package if you think "stable" is too old. Better than randomly picked git revisions at least :)

delroth:
(FYI this was reviewed by Orphis and he let me know about CMake's IN_LIST feature, which this PR is now using. He's not able to log in to GH right now, so dropping the "requested review" bit.)

degasus:
Shall we also check FFMPEG?

delroth:
I don't think we ever build ffmpeg from externals?

Mystro256:
Thanks @delroth ! I'll check it out and let you know if I have an feedback.

PR #8807, opened 2020-05-10
GCAdapter: add libusb error logging on reads/writes

delroth:


delroth:
It builds and according to JMC it doesn't spam logs.

JMC47:
Doesn't spam logs when I was playing a GameCube games or just letting it sit with an adapter plugged in.

PR #8808, opened 2020-05-10
NetworkWidget: Add Blocking column

sepalani:
This PR adds a "Blocking" column to the Network widget telling if the socket is blocking.

### DIFF
--- a/Source/Core/Core/IOS/Network/Socket.cpp
+++ b/Source/Core/Core/IOS/Network/Socket.cpp
@@ -652,6 +652,12 @@ s32 WiiSockMan::AddSocket(s32 fd, bool is_rw)
   return wii_fd;
 }
 
+bool WiiSockMan::IsSocketBlocking(s32 wii_fd) const
+{
+  const auto it = WiiSockets.find(wii_fd);
+  return it != WiiSockets.end() && !it->second.nonBlock;
+}
+
 s32 WiiSockMan::NewSocket(s32 af, s32 type, s32 protocol)
 {
   if (af != 2 && af != 23)  // AF_INET && AF_INET6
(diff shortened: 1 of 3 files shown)


Source/Core/Core/IOS/Network/Socket.cpp
```diff
@@ -652,6 +652,11 @@ s32 WiiSockMan::AddSocket(s32 fd, bool is_rw)
   return wii_fd;
 }
 
+bool WiiSockMan::IsSocketBlocking(s32 wii_fd) const
+{
+  return WiiSockets.count(wii_fd) != 0 && !WiiSockets.at(wii_fd).nonBlock;
```
iwubcode, line 657: Would a `find` be more performant?  I think having both of these calls does two lookups.

sepalani:
@iwubcode 
You're right. I fixed it.

Source/Core/DolphinQt/Debugger/NetworkWidget.cpp
```diff
@@ -90,6 +90,15 @@ QTableWidgetItem* GetSocketState(s32 host_fd)
   return new QTableWidgetItem(QTableWidget::tr("Unbound"));
 }
 
+QTableWidgetItem* GetSocketBlocking(s32 wii_fd)
+{
+  const auto& socket_manager = IOS::HLE::WiiSockMan::GetInstance();
+  if (socket_manager.GetHostSocket(wii_fd) < 0)
+    return new QTableWidgetItem();
```
Tilka, line 97: Not saying it should be removed, but why are you checking for a host socket?

sepalani: The function is called for each possible socket fd. Some are used by the Wii and some aren't. The one which aren't will return `-EBADF`, that's why I'm checking the GetHostSocket return value. The same check is done in many get functions above: `if (host_fd < 0)`.

PR #8809, opened 2020-05-11
Replace some acronyms

Ebola16:
I replaced some acronyms to help users figure out (or search for) what those settings change.

### DIFF
--- a/Source/Core/Common/Logging/LogManager.cpp
+++ b/Source/Core/Common/Logging/LogManager.cpp
@@ -93,11 +93,11 @@ static size_t DeterminePathCutOffPoint()
 LogManager::LogManager()
 {
   // create log containers
-  m_log[ACTIONREPLAY] = {"ActionReplay", "ActionReplay"};
+  m_log[ACTIONREPLAY] = {"ActionReplay", "Action Replay"};
   m_log[AUDIO] = {"Audio", "Audio Emulator"};
   m_log[AUDIO_INTERFACE] = {"AI", "Audio Interface"};
   m_log[BOOT] = {"BOOT", "Boot"};
-  m_log[COMMANDPROCESSOR] = {"CP", "CommandProc"};
+  m_log[COMMANDPROCESSOR] = {"CP", "Command Processor"};
   m_log[COMMON] = {"COMMON", "Common"};
   m_log[CONSOLE] = {"CONSOLE", "Dolphin Console"};
   m_log[CORE] = {"CORE", "Core"};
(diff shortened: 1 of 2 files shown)


delroth:
I'm personally -1 on this. I don't think the expansion of the acronyms helps for the most part in figuring out what they mean (e.g. "Digital Signal Processor" does not indicate any more than "DSP" that it's for audio).

For some acronyms the expanded form is completely unfamiliar (TIL what "FPRF" expands to, and I wouldn't recognize it written like that, even though I perfectly know what FPRF is). DSP/MMU are super common computing terms, and just like we wouldn't expand "CPU" or "GPU" I'm not sure why you would expand "DSP" or "MMU" which are really just as familiar in the context where they matter.

Ebola16:
> I don't think the expansion of the acronyms helps for the most part in figuring out what they mean

From the messages I have received from the Smash Bros. modding community and personal friends, it seems there is a sizable portion of users that have minimal emulation knowledge but want to know what Dolphin's settings can do. I also frequently tell people to look at Dolphin logs while troubleshooting problems and I often get asked what certain acronyms mean. DSP and MMU's meanings have been topics of past questions (spoiler: DSP doesn't mean Dolby Sound Protocol). As long as expanding the acronym doesn't interfere with its understanding or UI readability, I think replacing empty space with meaningful information can only help users.

> For some acronyms the expanded form is completely unfamiliar

I am trying to avoid this situation. I expanded DSP in the title of the setting but left the acronym in the options so users can know that both are used. For FPRF and MMU, would changing the text to something like "Enable Floating-point Result Flag (FPRF)" be acceptable?

The whole point of this PR is provide the easiest way for users to figure out what Dolphin settings do. If a user doesn't know what a setting means, expanding the acronym in Dolphin will make their research easier.

sepalani:
IMHO, I'm fine with expanding names that are poorly abbreviated such as: ActionReplay, CommandProc, GPFifo, PixelEngine, etc. They don't seem abbreviated nor spelled properly to me and having proper spelling won't hurt here.

But as @delroth said, expanding names like DSP, IO, SSL, WFS, register names and others is just overkill. Your example is also not that good, AFAIK, almost each time you search on google "wii <keyword>" you will find a relevant answer pointing to a Dolphin blog or a wiibrew page.

Plus, by expanding names like WFS and OSReport you will do the opposite and make searching details for them harder.

Ebola16:
After hearing all of your feedback I made some changes to this PR. I definitely don't want to make the information about these settings harder to find so some of the changes have been reverted.

However, I have received questions from multiple people about FPRF and MMU so I am still convinced that expanding these acronyms will be beneficial to the user. AFAIK, there is no downside to expanding these acronyms and doing so will make curious users certain they have found information on the correct topic.

Are these changes acceptable?

Tilka:
I'll merge this if you remove the following changes: JIT, IPC, MMU, FPRF (MMU and FPRF already have tooltips).

Ebola16:
I can't build on Windows at the moment, so let this PR be built by the buildbot since I added a tooltip for MMU on the Advanced Settings tab.

Ebola16:
Should be good for merge now

PR #8811, opened 2020-05-13
Externals: Add zstd

JosJuice:
Split out from PR #8538.

I had to rename Source/Common/Compiler.h because the VS build confuses it with Externals/zstd/lib/common/compiler.h otherwise.


delroth:
Can you also update Externals/README.md?

JosJuice:
Done. I also added info about bzip2 and liblzma because I didn't know about that file when adding those two libraries in an earlier PR.

PR #3316, opened 2015-12-05
WSAPoll: Fixed - invalid argument was supplied

sepalani:
According to the [WSAPoll documentation](https://msdn.microsoft.com/en-us/library/windows/desktop/ms741669(v=vs.85).aspx):
> POLLPRI - Priority data may be read without blocking. This flag is not supported by the Microsoft Winsock provider.

Then the **POLLWRBAND** flag is not even documented but is at least defined. However, if used:
> The events member of the WSAPOLLFD structure must only contain a combination of the above flags that are supported by the Winsock provider. Any other values are considered errors and WSAPoll will return SOCKET_ERROR.

It was also discussed [here](https://social.msdn.microsoft.com/Forums/sqlserver/en-US/18769abd-fca0-4d3c-9884-1a38ce27ae90/wsapoll-and-nonblocking-connects-to-nonexistent-ports?forum=wsk).

This PR fixes those WSAPoll() issues.

PR #3143, opened 2015-10-05
WSAGetLastError: fix return value

sepalani:
This function was returning 0 rather than the proper error code in VS2015 debug builds preventing network functions to be handled properly. See [Issue #8997](https://bugs.dolphin-emu.org/issues/8997).

PR #8813, opened 2020-05-16
Make SO_POLL complete asynchronously in IOS_NET SO

nbouteme:
This is to make the behavior more consistent with the behavior on console.

Currently, SO_POLL blocks the PPC until poll() completes, so this prevents PPC threads on DolphinOS and libogc's LWP from running while waiting for the call to complete, while on a real console, IOS lets the PPC run, as with most other calls, and notifies the PPC through the usual mechanism.

I don't know if how I implemented it is the preferred way in regard to Dolphin's internals: Basically each SO_POLL operation gets queued into a poll table with the associated request, and when an event happens on a socket that is referenced in a poll table in WiiSockMan::Update, an IPC reply is generated for the request.

### DIFF
--- a/Source/Core/Core/IOS/Network/IP/Top.cpp
+++ b/Source/Core/Core/IOS/Network/IP/Top.cpp
@@ -50,13 +50,6 @@
 #include <unistd.h>
 #endif
 
-// WSAPoll doesn't support POLLPRI and POLLWRBAND flags
-#ifdef _WIN32
-#define UNSUPPORTED_WSAPOLL POLLPRI | POLLWRBAND
-#else
-#define UNSUPPORTED_WSAPOLL 0
-#endif
-
 namespace IOS::HLE::Device
 {
 enum SOResultCode : s32
@@ -78,6 +71,12 @@ NetIPTop::~NetIPTop()
 #ifdef _WIN32
   WSACleanup();
 #endif
+}
+
+void NetIPTop::DoState(PointerWrap& p)
+{
+  DoStateShared(p);
+  WiiSockMan::GetInstance().DoState(p);
 }
 
 static constexpr u32 inet_addr(u8 a, u8 b, u8 c, u8 d)
@@ -599,77 +598,45 @@ IPCCommandResult NetIPTop::HandleInetNToPRequest(const IOCtlRequest& request)
 
 IPCCommandResult NetIPTop::HandlePollRequest(const IOCtlRequest& request)
 {
-  // Map Wii/native poll events types
-  struct
+  WiiSockMan& sm = WiiSockMan::GetInstance();
+
+  if (!request.buffer_in || !request.buffer_out)
+    return GetDefaultReply(-SO_EINVAL);
+
+  // Negative timeout indicates wait forever
+  const s64 timeout = static_cast<s64>(Memory::Read_U64(request.buffer_in));
+
+  const u32 nfds = request.buffer_out_size / 0xc;
+  if (nfds == 0 || nfds > WII_SOCKET_FD_MAX)
   {
-    int native;
-    int wii;
-  } mapping[] = {
-      {POLLRDNORM, 0x0001}, {POLLRDBAND, 0x0002}, {POLLPRI, 0x0004}, {POLLWRNORM, 0x0008},
-      {POLLWRBAND, 0x0010}, {POLLERR, 0x0020},    {POLLHUP, 0x0040}, {POLLNVAL, 0x0080},
-  };
-
-  u32 unknown = Memory::Read_U32(request.buffer_in);
-  u32 timeout = Memory::Read_U32(request.buffer_in + 4);
-
-  int nfds = request.buffer_out_size / 0xc;
-  if (nfds == 0)
-    ERROR_LOG(IOS_NET, "Hidden POLL");
+    ERROR_LOG(IOS_NET, "IOCTL_SO_POLL failed: Invalid array size %d, ret=%d", nfds, -SO_EINVAL);
+    return GetDefaultReply(-SO_EINVAL);
+  }
 
   std::vector<pollfd_t> ufds(nfds);
 
-  for (int i = 0; i < nfds; ++i)
+  for (u32 i = 0; i < nfds; ++i)
   {
-    s32 wii_fd = Memory::Read_U32(request.buffer_out + 0xc * i);
-    ufds[i].fd = WiiSockMan::GetInstance().GetHostSocket(wii_fd);          // fd
-    int events = Memory::Read_U32(request.buffer_out + 0xc * i + 4);       // events
-    ufds[i].revents = Memory::Read_U32(request.buffer_out + 0xc * i + 8);  // revents
+    const s32 wii_fd = Memory::Read_U32(request.buffer_out + 0xc * i);
+    ufds[i].fd = sm.GetHostSocket(wii_fd);                                  // fd
+    const int events = Memory::Read_U32(request.buffer_out + 0xc * i + 4);  // events
+    ufds[i].revents = 0;
 
     // Translate Wii to native events
-    int unhandled_events = events;
-    ufds[i].events = 0;
-    for (auto& map : mapping)
-    {
-      if (events & map.wii)
-        ufds[i].events |= map.native;
-      unhandled_events &= ~map.wii;
-    }
+    ufds[i].events = WiiSockMan::ConvertEvents(events, WiiSockMan::ConvertDirection::WiiToNative);
     DEBUG_LOG(IOS_NET,
               "IOCTL_SO_POLL(%d) "
-              "Sock: %08x, Unknown: %08x, Events: %08x, "
+              "Sock: %08x, Events: %08x, "
               "NativeEvents: %08x",
-              i, wii_fd, unknown, events, ufds[i].events);
+              i, wii_fd, events, ufds[i].events);
 
     // Do not pass return-only events to the native poll
     ufds[i].events &= ~(POLLERR | POLLHUP | POLLNVAL | UNSUPPORTED_WSAPOLL);
-
-    if (unhandled_events)
-      ERROR_LOG(IOS_NET, "SO_POLL: unhandled Wii event types: %04x", unhandled_events);
   }
 
-  int ret = poll(ufds.data(), nfds, timeout);
-  ret = WiiSockMan::GetNetErrorCode(ret, "SO_POLL", false);
-
-  for (int i = 0; i < nfds; ++i)
-  {
-    // Translate native to Wii events
-    int revents = 0;
-    for (auto& map : mapping)
-    {
-      if (ufds[i].revents & map.native)
-        revents |= map.wii;
-    }
-
-    // No need to change fd or events as they are input only.
-    // Memory::Write_U32(ufds[i].fd, request.buffer_out + 0xc*i); //fd
-    // Memory::Write_U32(events, request.buffer_out + 0xc*i + 4); //events
-    Memory::Write_U32(revents, request.buffer_out + 0xc * i + 8);  // revents
-
-    DEBUG_LOG(IOS_NET, "IOCTL_SO_POLL socket %d wevents %08X events %08X revents %08X", i, revents,
-              ufds[i].events, ufds[i].revents);
-  }
-
-  return GetDefaultReply(ret);
+  // Prevents blocking emulation on a blocking poll
+  sm.AddPollCommand({request.address, request.buffer_out, std::move(ufds), timeout});
+  return GetNoReply();
 }
 
 IPCCommandResult NetIPTop::HandleGetHostByNameRequest(const IOCtlRequest& request)
--- a/Source/Core/Core/IOS/Network/IP/Top.h
+++ b/Source/Core/Core/IOS/Network/IP/Top.h
@@ -13,6 +13,13 @@
 #include <ws2tcpip.h>
 #endif
 
+// WSAPoll doesn't support POLLPRI and POLLWRBAND flags
+#ifdef _WIN32
+constexpr int UNSUPPORTED_WSAPOLL = POLLPRI | POLLWRBAND;
+#else
+constexpr int UNSUPPORTED_WSAPOLL = 0;
+#endif
+
 namespace IOS::HLE
 {
 enum NET_IOCTL
@@ -62,6 +69,7 @@ public:
   NetIPTop(Kernel& ios, const std::string& device_name);
   virtual ~NetIPTop();
 
+  void DoState(PointerWrap& p) override;
   IPCCommandResult IOCtl(const IOCtlRequest& request) override;
   IPCCommandResult IOCtlV(const IOCtlVRequest& request) override;
 
--- a/Source/Core/Core/IOS/Network/Socket.cpp
+++ b/Source/Core/Core/IOS/Network/Socket.cpp
@@ -6,6 +6,8 @@
 #include "Core/IOS/Network/Socket.h"
 
 #include <algorithm>
+#include <numeric>
+
 #include <mbedtls/error.h>
 #ifndef _WIN32
 #include <arpa/inet.h>
@@ -697,7 +699,7 @@ void WiiSockMan::Update()
 
   while (socket_iter != end_socks)
   {
-    WiiSocket& sock = socket_iter->second;
+    const WiiSocket& sock = socket_iter->second;
     if (sock.IsValid())
     {
       FD_SET(sock.fd, &read_fds);
@@ -712,7 +714,8 @@ void WiiSockMan::Update()
       socket_iter = WiiSockets.erase(socket_iter);
     }
   }
-  s32 ret = select(nfds, &read_fds, &write_fds, &except_fds, &t);
+
+  const s32 ret = select(nfds, &read_fds, &write_fds, &except_fds, &t);
 
   if (ret >= 0)
   {
@@ -730,6 +733,87 @@ void WiiSockMan::Update()
       elem.second.Update(false, false, false);
     }
   }
+  UpdatePollCommands();
+}
+
+void WiiSockMan::UpdatePollCommands()
+{
+  static constexpr int error_event = (POLLHUP | POLLERR);
+
+  if (pending_polls.empty())
+    return;
+
+  const auto now = std::chrono::high_resolution_clock::now();
+  const auto elapsed_d = std::chrono::duration_cast<std::chrono::milliseconds>(now - last_time);
+  const auto elapsed = elapsed_d.count();
+  last_time = now;
+
+  for (auto& pcmd : pending_polls)
+  {
+    // Don't touch negative timeouts
+    if (pcmd.timeout > 0)
+      pcmd.timeout = std::max<s64>(0, pcmd.timeout - elapsed);
+  }
+
+  pending_polls.erase(
+      std::remove_if(
+          pending_polls.begin(), pending_polls.end(),
+          [this](auto& pcmd) {
+            const auto request = Request(pcmd.request_addr);
+            auto& pfds = pcmd.wii_fds;
+            int ret = 0;
+
+            // Happens only on savestate load
+            if (pfds[0].revents & error_event)
+            {
+              ret = static_cast<int>(pfds.size());
+            }
+            else
+            {
+              // Make the behavior of poll consistent across platforms by not passing:
+              //  - Set with invalid fds, revents is set to 0 (Linux) or POLLNVAL (Windows)
+              //  - Set without a valid socket, raises an error on Windows
+              std::vector<int> original_order(pfds.size());
+              std::iota(original_order.begin(), original_order.end(), 0);
+              // Select indices with valid fds
+              auto mid = std::partition(original_order.begin(), original_order.end(), [&](auto i) {
+                return GetHostSocket(Memory::Read_U32(pcmd.buffer_out + 0xc * i)) >= 0;
+              });
+              const auto n_valid = std::distance(original_order.begin(), mid);
+
+              // Move all the valid pollfds to the front of the vector
+              for (auto i = 0; i < n_valid; ++i)
+                std::swap(pfds[i], pfds[original_order[i]]);
+
+              if (n_valid > 0)
+                ret = poll(pfds.data(), n_valid, 0);
+              if (ret < 0)
+                ret = GetNetErrorCode(ret, "UpdatePollCommands", false);
+
+              // Move everything back to where they were
+              for (auto i = 0; i < n_valid; ++i)
+                std::swap(pfds[i], pfds[original_order[i]]);
+            }
+
+            if (ret == 0 && pcmd.timeout)
+              return false;
+
+            // Translate native to Wii events,
+            for (u32 i = 0; i < pfds.size(); ++i)
+            {
+              const int revents = ConvertEvents(pfds[i].revents, ConvertDirection::NativeToWii);
+
+              // No need to change fd or events as they are input only.
+              // Memory::Write_U32(ufds[i].fd, request.buffer_out + 0xc*i); //fd
+              // Memory::Write_U32(events, request.buffer_out + 0xc*i + 4); //events
+              Memory::Write_U32(revents, pcmd.buffer_out + 0xc * i + 8);  // revents
+              DEBUG_LOG(IOS_NET, "IOCTL_SO_POLL socket %d wevents %08X events %08X revents %08X", i,
+                        revents, pfds[i].events, pfds[i].revents);
+            }
+            GetIOS()->EnqueueIPCReply(request, ret);
+            return true;
+          }),
+      pending_polls.end());
 }
 
 void WiiSockMan::Convert(WiiSockAddrIn const& from, sockaddr_in& to)
@@ -737,6 +821,43 @@ void WiiSockMan::Convert(WiiSockAddrIn const& from, sockaddr_in& to)
   to.sin_addr.s_addr = from.addr.addr;
   to.sin_family = from.family;
   to.sin_port = from.port;
+}
+
+s32 WiiSockMan::ConvertEvents(s32 events, ConvertDirection dir)
+{
+  constexpr struct
+  {
+    int native;
+    int wii;
+  } mapping[] = {
+      {POLLRDNORM, 0x0001}, {POLLRDBAND, 0x0002}, {POLLPRI, 0x0004}, {POLLWRNORM, 0x0008},
+      {POLLWRBAND, 0x0010}, {POLLERR, 0x0020},    {POLLHUP, 0x0040}, {POLLNVAL, 0x0080},
+  };
+
+  s32 converted_events = 0;
+  s32 unhandled_events = 0;
+
+  if (dir == ConvertDirection::NativeToWii)
+  {
+    for (const auto& map : mapping)
+    {
+      if (events & map.native)
+        converted_events |= map.wii;
+    }
+  }
+  else
+  {
+    unhandled_events = events;
+    for (const auto& map : mapping)
+    {
+      if (events & map.wii)
+        converted_events |= map.native;
+      unhandled_events &= ~map.wii;
+    }
+  }
+  if (unhandled_events)
+    ERROR_LOG(IOS_NET, "SO_POLL: unhandled Wii event types: %04x", unhandled_events);
+  return converted_events;
 }
 
 void WiiSockMan::Convert(sockaddr_in const& from, WiiSockAddrIn& to, s32 addrlen)
@@ -748,6 +869,35 @@ void WiiSockMan::Convert(sockaddr_in const& from, WiiSockAddrIn& to, s32 addrlen
     to.len = sizeof(WiiSockAddrIn);
   else
     to.len = addrlen;
+}
+
+void WiiSockMan::DoState(PointerWrap& p)
+{
+  bool saving =
+      p.mode == PointerWrap::Mode::MODE_WRITE || p.mode == PointerWrap::Mode::MODE_MEASURE;
+  auto size = pending_polls.size();
+  p.Do(size);
+  if (!saving)
+    pending_polls.resize(size);
+  for (auto& pcmd : pending_polls)
+  {
+    p.Do(pcmd.request_addr);
+    p.Do(pcmd.buffer_out);
+    p.Do(pcmd.wii_fds);
+  }
+
+  if (saving)
+    return;
+  for (auto& pcmd : pending_polls)
+  {
+    for (auto& wfd : pcmd.wii_fds)
+      wfd.revents = (POLLHUP | POLLERR);
+  }
+}
+
+void WiiSockMan::AddPollCommand(const PollCommand& cmd)
+{
+  pending_polls.push_back(cmd);
 }
 
 void WiiSockMan::UpdateWantDeterminism(bool want)
--- a/Source/Core/Core/IOS/Network/Socket.h
+++ b/Source/Core/Core/IOS/Network/Socket.h
@@ -212,6 +212,20 @@ private:
 class WiiSockMan
 {
 public:
+  enum class ConvertDirection
+  {
+    WiiToNative,
+    NativeToWii
+  };
+
+  struct PollCommand
+  {
+    u32 request_addr;
+    u32 buffer_out;
+    std::vector<pollfd_t> wii_fds;
+    s64 timeout;
+  };
+
   static s32 GetNetErrorCode(s32 ret, const char* caller, bool isRW);
   static char* DecodeError(s32 ErrorCode);
 
@@ -223,6 +237,10 @@ public:
   void Update();
   static void Convert(WiiSockAddrIn const& from, sockaddr_in& to);
   static void Convert(sockaddr_in const& from, WiiSockAddrIn& to, s32 addrlen = -1);
+  static s32 ConvertEvents(s32 events, ConvertDirection dir);
+
+  void DoState(PointerWrap& p);
+  void AddPollCommand(const PollCommand& cmd);
   // NON-BLOCKING FUNCTIONS
   s32 NewSocket(s32 af, s32 type, s32 protocol);
   s32 AddSocket(s32 fd, bool is_rw);
@@ -256,7 +274,12 @@ private:
   WiiSockMan(WiiSockMan&&) = delete;
   WiiSockMan& operator=(WiiSockMan&&) = delete;
 
+  void UpdatePollCommands();
+
   std::unordered_map<s32, WiiSocket> WiiSockets;
   s32 errno_last;
+  std::vector<PollCommand> pending_polls;
+  std::chrono::time_point<std::chrono::high_resolution_clock> last_time =
+      std::chrono::high_resolution_clock::now();
 };
 }  // namespace IOS::HLE
--- a/Source/Core/Core/State.cpp
+++ b/Source/Core/Core/State.cpp
@@ -74,7 +74,7 @@ static Common::Event g_compressAndDumpStateSyncEvent;
 static std::thread g_save_thread;
 
 // Don't forget to increase this after doing changes on the savestate system
-constexpr u32 STATE_VERSION = 117;  // Last changed in PR 7396
+constexpr u32 STATE_VERSION = 118;  // Last changed in PR 8813
 
 // Maps savestate versions to Dolphin versions.
 // Versions after 42 don't need to be added to this list,


JosJuice:
@dolphin-emu-bot rebuild

Source/Core/Core/IOS/Network/Socket.cpp
```diff
@@ -686,24 +688,19 @@ s32 WiiSockMan::DeleteSocket(s32 s)
...
     if (sock.IsValid())
     {
-      FD_SET(sock.fd, &read_fds);
-      FD_SET(sock.fd, &write_fds);
-      FD_SET(sock.fd, &except_fds);
-      nfds = std::max(nfds, sock.fd + 1);
+      const int write_event = (POLLOUT | POLLWRBAND);
+      const int read_event = (POLLIN | POLLRDNORM | POLLRDBAND | POLLPRI);
+      ufds.push_back({sock.fd, read_event | write_event, 0});
+      nfds = std::max(nfds, sock.wii_fd + 1);
```
sepalani, line 703: Good catch, the `wii_fd` should be used here.

Source/Core/Core/IOS/Network/Socket.cpp
```diff
@@ -712,24 +709,99 @@ void WiiSockMan::Update()
...
+  struct
+  {
+    int native;
+    int wii;
+  } mapping[] = {
+      {POLLRDNORM, 0x0001}, {POLLRDBAND, 0x0002}, {POLLPRI, 0x0004}, {POLLWRNORM, 0x0008},
+      {POLLWRBAND, 0x0010}, {POLLERR, 0x0020},    {POLLHUP, 0x0040}, {POLLNVAL, 0x0080},
+  };
+
+  auto mid =
```
sepalani, line 751: Can't this logic be simplified and have its own functions?

Source/Core/Core/IOS/Network/Socket.cpp
```diff
@@ -712,24 +709,99 @@ void WiiSockMan::Update()
...
+						std::make_move_iterator(mid));
+  pending_polls.erase(pending_polls.begin(), mid); 
+  std::move(pending_polls.begin(),
+            mid,
+            std::back_inserter(finished_polls));
+  std::for_each(finished_polls.begin(),
+                finished_polls.end(),
+                [&](auto &pcmd) {
+                  auto r = pcmd.timeout == 0 ? 0 : pcmd.trigger_count;
+                  r = WiiSockMan::GetNetErrorCode(r, "SO_POLL", false);
```
sepalani, line 782: This function should only be used against network functions and right after they returned otherwise the behaviour is undefined. Especially, since it sets the last error value and on top of that calls `WSAGetLastError` on Windows.

It seems unnecessary here as it's not an error value.

Source/Core/Core/IOS/Network/Socket.cpp
```diff
@@ -712,24 +709,99 @@ void WiiSockMan::Update()
       socket_iter = WiiSockets.erase(socket_iter);
     }
   }
-  s32 ret = select(nfds, &read_fds, &write_fds, &except_fds, &t);
 
-  if (ret >= 0)
+  // ret is unused as every socket gets updated anyway
+  [[maybe_unused]]
+	  s32 ret = poll(ufds.data(), nfds, 0);
```
sepalani, line 715: The return value should be checked and at least an `ERROR_LOG` should be printed when it fails. `WiiSockMan::GetNetErrorCode` might be used here.

delroth:
This will leave an unanswered/hanging IPC when saving/loading states.

nbouteme:
Thanks for the comments. I'll fix these ASAP.

> This will leave an unanswered/hanging IPC when saving/loading states.
This means I should add a DoState method to WiiSockMan, and answer IPC calls with errors when loading from a save state, right?

delroth:
I think that would make the most sense, yes.

Source/Core/Core/IOS/Network/Socket.cpp
```diff
@@ -739,6 +801,41 @@ void WiiSockMan::Convert(WiiSockAddrIn const& from, sockaddr_in& to)
...
+      {POLLRDNORM, 0x0001}, {POLLRDBAND, 0x0002}, {POLLPRI, 0x0004}, {POLLWRNORM, 0x0008},
+      {POLLWRBAND, 0x0010}, {POLLERR, 0x0020},    {POLLHUP, 0x0040}, {POLLNVAL, 0x0080},
+  };
+
+  s32 converted_events = 0;
+  s32 unhandled_events = 0;
+
+  if (dir == ConvertDirection::NATIVE_TO_WII)
+  {
+    for (auto& map : mapping)
```
BhaaLseN, line 820: This one needs curlies since the contents span over more than a single line.

Source/Core/Core/IOS/Network/IP/Top.h
```diff
@@ -62,6 +62,7 @@ class NetIPTop : public Device
   NetIPTop(Kernel& ios, const std::string& device_name);
   virtual ~NetIPTop();
 
+  void DoState(PointerWrap& p) override;
```
BhaaLseN, line 65: Since you added this, you must not forget to increment the save state version over here:
https://github.com/dolphin-emu/dolphin/blob/099197b480bcc35fd84f059359ab739a30b0bac0/Source/Core/Core/State.cpp#L76-L77

Source/Core/Core/IOS/Network/Socket.h
```diff
@@ -258,5 +279,11 @@ class WiiSockMan
 
   std::unordered_map<s32, WiiSocket> WiiSockets;
   s32 errno_last;
+  std::vector<poll_command> pending_polls;
+  std::vector<poll_command> failed_polls;
+  std::vector<poll_command> finished_polls;
+  std::vector<pollfd> ufds;
```
BhaaLseN, line 285: You should probably be consistent here and use the typedef for `pollfd_t` instead.

Besides, I don't see this being used outside the `WiiSockMan::Update` method. Why don't you move it in there as a local?

Source/Core/Core/IOS/Network/Socket.cpp
```diff
@@ -685,25 +687,27 @@ s32 WiiSockMan::DeleteSocket(s32 s)
...
 {
-  s32 nfds = 0;
-  fd_set read_fds, write_fds, except_fds;
-  struct timeval t = {0, 0};
-  FD_ZERO(&read_fds);
-  FD_ZERO(&write_fds);
-  FD_ZERO(&except_fds);
+  const int write_event = (POLLOUT | POLLWRBAND);
+  const int read_event = (POLLIN | POLLRDNORM | POLLRDBAND | POLLPRI);
+  const int error_event = (POLLHUP | POLLERR);
```
BhaaLseN, line 692: Any reason you don't make those `constexpr`?

Source/Core/Core/IOS/Network/Socket.h
nbouteme: Update is called frequently, so I wanted to avoid doing frequent allocation/deallocations.
But this made me realize I forgot to clear the container.

Source/Core/Core/IOS/Network/Socket.cpp
nbouteme: I forgot this was a thing. Fixed.

Source/Core/Core/IOS/Network/Socket.cpp
```diff
@@ -750,6 +854,29 @@ void WiiSockMan::Convert(sockaddr_in const& from, WiiSockAddrIn& to, s32 addrlen
...
+void WiiSockMan::DoState(PointerWrap& p)
+{
+  bool saving =
+      p.mode == PointerWrap::Mode::MODE_WRITE || p.mode == PointerWrap::Mode::MODE_MEASURE;
+  auto& target = saving ? pending_polls : failed_polls;
+  auto size = target.size();
+  p.Do(size);
+  if (!saving)
+    target.resize(size);
+  std::for_each(target.begin(), target.end(), [&](auto& pcmd) {
```
sepalani, line 866: - Is it useful to save the `timeout` and `trigger_count` value when they will error out anyway?
 - Can't on a savestate load `POLLHUP | POLLERR` be assigned to all `pending_polls.wii_fds.revents` rather than creating a new vector?

Source/Core/Core/IOS/Network/Socket.cpp
```diff
@@ -739,6 +802,47 @@ void WiiSockMan::Convert(WiiSockAddrIn const& from, sockaddr_in& to)
...
+
+  s32 converted_events = 0;
+  s32 unhandled_events = 0;
+
+  if (dir == ConvertDirection::NATIVE_TO_WII)
+  {
+    for (auto& map : mapping)
+    {
+      if (events & map.native)
+      {
```
sepalani, line 824: As the `if` body's one line, braces can be omitted.

Source/Core/Core/IOS/Network/Socket.cpp
```diff
@@ -739,6 +802,47 @@ void WiiSockMan::Convert(WiiSockAddrIn const& from, sockaddr_in& to)
...
+      }
+    }
+  }
+  else
+  {
+    unhandled_events = events;
+    for (auto& map : mapping)
+    {
+      if (events & map.wii)
+      {
```
sepalani, line 835: Ditto

Source/Core/Core/IOS/Network/Socket.cpp
```diff
@@ -712,24 +717,82 @@ void WiiSockMan::Update()
...
+  std::move(pending_polls.begin(), mid, std::back_inserter(finished_polls));
+  std::for_each(finished_polls.begin(), finished_polls.end(), [&](auto& pcmd) {
+    auto r = pcmd.timeout == 0 ? 0 : pcmd.trigger_count;
+
+    // Translate native to Wii events
+    auto& pfds = pcmd.wii_fds;
+    for (auto i = 0u; i < pfds.size(); ++i)
     {
-      elem.second.Update(false, false, false);
+      int revents = ConvertEvents(pfds[i].revents, ConvertDirection::NATIVE_TO_WII);
```
sepalani, line 773: `const int`

Source/Core/Core/IOS/Network/Socket.cpp
```diff
@@ -685,25 +687,28 @@ s32 WiiSockMan::DeleteSocket(s32 s)
...
-  FD_ZERO(&write_fds);
-  FD_ZERO(&except_fds);
+  constexpr int write_event = (POLLOUT | POLLWRBAND);
+  constexpr int read_event = (POLLIN | POLLRDNORM | POLLRDBAND | POLLPRI);
+  constexpr int error_event = (POLLHUP | POLLERR);
 
+  s32 nfds = 0;
   auto socket_iter = WiiSockets.begin();
   auto end_socks = WiiSockets.end();
+  ufds.clear();
```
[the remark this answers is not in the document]
nbouteme: Is it really appropriate to allocate and deallocate a vector in an update function? Clear preserves the allocated capacity.

Source/Core/Core/IOS/Network/Socket.cpp
```diff
@@ -712,24 +717,82 @@ void WiiSockMan::Update()
...
+  int n = 0;
+  for (auto& elem : WiiSockets)
   {
-    for (auto& elem : WiiSockets)
+    elem.second.Update(!!(ufds[n].revents & read_event), !!(ufds[n].revents & write_event),
+                       !!(ufds[n].revents & error_event));
+    n++;
+  }
+
+  auto mid = std::partition(pending_polls.begin(), pending_polls.end(), [&](auto& pcmd) {
```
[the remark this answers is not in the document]
nbouteme: I just replaced that with an equivalent [erase-remove idiom](https://en.wikipedia.org/wiki/Erase%E2%80%93remove_idiom).
It still allowed me to remove these two member variables. The only performance benefit to doing it this way instead of iterating is to do only one erase operation, which doesn't really matter in typical cases (IOS doesn't allow polling more than a couple tens of file descriptors). The other benefits in using the stdlib for container manipulations are purely theoretical, but are (maybe) still there. But mostly being less error prone.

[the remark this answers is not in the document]
sepalani: My only concern about doing it this way, is that it will call `begin()`/`end()`/`remove_if()`/`erase()` every time regardless of the `pending_polls` size. I also think that the whole update poll logic should belong to its own method: "UpdatePollCommands" or something like that.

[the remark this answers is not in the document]
nbouteme: Most (if not all) of these calls are inlined, the only external reference (on g++) is a realloc_insert call, probably for the push_back on ufds, which will probably never get called since its size is preallocated before the loop. I'll move all that logic into a separate function, though, it's true that it's too big to leech off another method.

Screenshot from 2020-05-19 19-28-38

Source/Core/Core/IOS/Network/Socket.cpp
```diff
@@ -712,24 +715,84 @@ void WiiSockMan::Update()
...
+              if (poll_it == ufds.end())
+                return false;
+
+              return !!((e.events | error_event) & poll_it->revents);
+            });
+
+            if (count == 0 && pcmd.timeout)
+              return false;
+
+            const auto r = pcmd.timeout == 0 ? 0 : count;
```
sepalani, line 777: IIRC, even if poll is non-blocking, it shouldn't prevent `count` to be reported so this ternary and the `r` variable are unnecessary.

Source/Core/Core/IOS/Network/Socket.cpp
```diff
@@ -739,6 +802,43 @@ void WiiSockMan::Convert(WiiSockAddrIn const& from, sockaddr_in& to)
...
+    for (auto& map : mapping)
+    {
+      if (events & map.wii)
+        converted_events |= map.native;
+      unhandled_events &= ~map.wii;
+    }
+  }
+  if (unhandled_events)
+    ERROR_LOG(IOS_NET, "SO_POLL: unhandled Wii event types: %04x", unhandled_events);
+  return events;
```
sepalani, line 839: Shouldn't this be `converted_events`?

Source/Core/Core/IOS/Network/Socket.cpp
```diff
@@ -712,24 +715,84 @@ void WiiSockMan::Update()
...
+  const std::chrono::time_point<std::chrono::high_resolution_clock> now =
+      std::chrono::high_resolution_clock::now();
+  const auto elapsed_d = std::chrono::duration_cast<std::chrono::milliseconds>(now - last_time);
+  const auto elapsed = elapsed_d.count();
+  last_time = now;
+
+  for (auto& pcmd : pending_polls)
+  {
+    if (pcmd.timeout > 0)  // don't touch negative timeouts
+      pcmd.timeout = std::max(static_cast<s64>(0), pcmd.timeout - elapsed);
```
sepalani, line 753: Does `std::max<s64>` work?

Source/Core/Core/IOS/Network/Socket.cpp
nbouteme: No, this is supposed to catch events that where requested from Wii mode that don't have a native equivalent and can't be converted.

nbouteme: My bad, good catch.

sepalani:
@dolphin-emu-bot rebuild

nbouteme:
Hm, this type error on the poll() call didn't happen on the previous windows builds. And MSVC doesn't seem to like the implicit constexpr capture from outside the scope of a lambda, even though the standard allows it. Or maybe I'm reading the error wrong.

JosJuice:
> And MSVC doesn't seem to like the implicit constexpr capture from outside the scope of a lambda, even though the standard allows it.

I've had that problem as well (though only when building with CMake, for some reason). I'd suggest just capturing it to make MSVC happy.

Source/Core/Core/IOS/Network/Socket.cpp
```diff
@@ -739,6 +799,43 @@ void WiiSockMan::Convert(WiiSockAddrIn const& from, sockaddr_in& to)
   to.sin_port = from.port;
 }
 
+s32 WiiSockMan::ConvertEvents(s32 events, ConvertDirection dir)
+{
+  struct
+  {
+    int native;
+    int wii;
+  } mapping[] = {
```
BhaaLseN, line 808: I wonder if this could (or even should) be `static` (and maybe `const`), perhaps even `constexpr`?

nbouteme: `constexpr` seems to work

nbouteme:
I set error_event as a `constexpr static` since on MSVC, explicitly capturing a constexpr seem to make the compiler disregard the constexpr qualifier? But this make GCC 10 print a warning about the variable being set but unused despite being used...

And about poll, I had to cast to an u32 since Windows doesn't provide the posix nfds_t type...

Source/Core/Core/IOS/Network/Socket.cpp
```diff
@@ -712,24 +715,81 @@ void WiiSockMan::Update()
...
+    n++;
   }
-  else
+
+  UpdatePollCommands();
+}
+
+void WiiSockMan::UpdatePollCommands()
+{
+  constexpr static int error_event = (POLLHUP | POLLERR);
```
lioncash, line 738: The static isn't necessary here. Given it's a trivial primitive type, this will be constant folded into expressions anyway (especially given it's constexpr).

Our convention is also to place the lifetime specifiers before mutability ones, just to keep in mind for the future.

Source/Core/Core/IOS/Network/Socket.cpp
```diff
@@ -712,24 +715,81 @@ void WiiSockMan::Update()
...
+
+  if (pending_polls.empty())
+    return;
+
+  const std::chrono::time_point<std::chrono::high_resolution_clock> now =
```
lioncash, line 743: Auto is particularly appropriate here.

```suggestion
  const auto now =
```

Source/Core/Core/IOS/Network/Socket.cpp
```diff
@@ -712,24 +715,81 @@ void WiiSockMan::Update()
...
+  const auto elapsed = elapsed_d.count();
+  last_time = now;
+
+  for (auto& pcmd : pending_polls)
   {
-    for (auto& elem : WiiSockets)
-    {
-      elem.second.Update(false, false, false);
-    }
+    if (pcmd.timeout > 0)  // don't touch negative timeouts
```
lioncash, line 751: Prefer placing comments above the conditional, rather than off to the side.

Source/Core/Core/IOS/Network/Socket.cpp
```diff
@@ -712,24 +715,81 @@ void WiiSockMan::Update()
...
+      pcmd.timeout = std::max<s64>(0, pcmd.timeout - elapsed);
   }
+
+  pending_polls.erase(
+      std::remove_if(
+          pending_polls.begin(), pending_polls.end(),
+          [this](auto& pcmd) {
+            const int count = std::count_if(pcmd.wii_fds.begin(), pcmd.wii_fds.end(), [&](auto& e) {
+              if (e.revents & error_event)  // Only happens if loaded from a save state
```
lioncash, line 760: Ditto

Source/Core/Core/IOS/Network/Socket.cpp
```diff
@@ -712,24 +715,81 @@ void WiiSockMan::Update()
...
+
+              return !!((e.events | error_event) & poll_it->revents);
+            });
+
+            if (count == 0 && pcmd.timeout)
+              return false;
+
+            // Translate native to Wii events
+            const auto& pfds = pcmd.wii_fds;
+            for (auto i = 0u; i < pfds.size(); ++i)
```
lioncash, line 778:
```suggestion
            for (u32 i = 0; i < pfds.size(); ++i)
```

Source/Core/Core/IOS/Network/Socket.cpp
```diff
@@ -739,6 +799,43 @@ void WiiSockMan::Convert(WiiSockAddrIn const& from, sockaddr_in& to)
...
+    for (auto& map : mapping)
+    {
+      if (events & map.native)
+        converted_events |= map.wii;
+    }
+  }
+  else
+  {
+    unhandled_events = events;
+    for (auto& map : mapping)
```
lioncash, line 827:
```suggestion
    for (const auto& map : mapping)
```

Source/Core/Core/IOS/Network/Socket.cpp
```diff
@@ -739,6 +799,43 @@ void WiiSockMan::Convert(WiiSockAddrIn const& from, sockaddr_in& to)
...
+      {POLLRDNORM, 0x0001}, {POLLRDBAND, 0x0002}, {POLLPRI, 0x0004}, {POLLWRNORM, 0x0008},
+      {POLLWRBAND, 0x0010}, {POLLERR, 0x0020},    {POLLHUP, 0x0040}, {POLLNVAL, 0x0080},
+  };
+
+  s32 converted_events = 0;
+  s32 unhandled_events = 0;
+
+  if (dir == ConvertDirection::NATIVE_TO_WII)
+  {
+    for (auto& map : mapping)
```
lioncash, line 818:
```suggestion
    for (const auto& map : mapping)
```

Source/Core/Core/IOS/Network/Socket.cpp
```diff
@@ -750,6 +847,42 @@ void WiiSockMan::Convert(sockaddr_in const& from, WiiSockAddrIn& to, s32 addrlen
...
+  if (!saving)
+    pending_polls.resize(size);
+  for (auto& pcmd : pending_polls)
+  {
+    p.Do(pcmd.request_addr);
+    p.Do(pcmd.buffer_out);
+    p.Do(pcmd.wii_fds);
+  }
+
+  if (!saving)
```
lioncash, line 865:
```suggestion
  if (saving)
    return;
```

Source/Core/Core/IOS/Network/Socket.h
```diff
@@ -211,7 +211,15 @@ class WiiSocket
 
 class WiiSockMan
 {
+  struct poll_command;
+
 public:
+  enum ConvertDirection
```
lioncash, line 217:
```suggestion
  enum class ConvertDirection
```

Source/Core/Core/IOS/Network/Socket.h
```diff
@@ -211,7 +211,15 @@ class WiiSocket
...
+  {
+    WII_TO_NATIVE,
```
lioncash, line 219:
```suggestion
    WiiToNative,
```

Source/Core/Core/IOS/Network/Socket.h
```diff
@@ -211,7 +211,15 @@ class WiiSocket
...
+    NATIVE_TO_WII
```
lioncash, line 220:
```suggestion
    NativeToWii
```

Source/Core/Core/IOS/Network/Socket.h
```diff
@@ -250,13 +262,27 @@ class WiiSockMan
   void UpdateWantDeterminism(bool want);
 
 private:
+  struct poll_command
```
lioncash, line 265: It seems like this should be part of the public interface, particularly given `AddPollCommand()` is a public function that makes use of this.

This should also be cased like `PollCommand` to be consistent with the naming convention.

Source/Core/Core/IOS/Network/Socket.cpp
nbouteme: I know, but it won't [build on Windows](https://dolphin.ci/#/builders/21/builds/948/steps/3/logs/stdio) without it due to a compiler bug.

Source/Core/Core/IOS/Network/IP/Top.cpp
```diff
@@ -621,55 +619,24 @@ IPCCommandResult NetIPTop::HandlePollRequest(const IOCtlRequest& request)
...
-    // Memory::Write_U32(events, request.buffer_out + 0xc*i + 4); //events
-    Memory::Write_U32(revents, request.buffer_out + 0xc * i + 8);  // revents
-
-    DEBUG_LOG(IOS_NET, "IOCTL_SO_POLL socket %d wevents %08X events %08X revents %08X", i, revents,
-              ufds[i].events, ufds[i].revents);
-  }
-
-  return GetDefaultReply(ret);
+  sm.AddPollCommand({request.address, request.buffer_out, std::move(ufds), timeout});
+  return {};
```
sepalani, line 639:
```C++
return GetNoReply();
```

Source/Core/Core/IOS/Network/Socket.cpp
```diff
@@ -712,24 +715,82 @@ void WiiSockMan::Update()
...
+              // Memory::Write_U32(ufds[i].fd, request.buffer_out + 0xc*i); //fd
+              // Memory::Write_U32(events, request.buffer_out + 0xc*i + 4); //events
+              Memory::Write_U32(revents, pcmd.buffer_out + 0xc * i + 8);  // revents
+              DEBUG_LOG(IOS_NET, "IOCTL_SO_POLL socket %d wevents %08X events %08X revents %08X", i,
+                        revents, pfds[i].events, pfds[i].revents);
+            }
+            GetIOS()->EnqueueIPCReply(Request(pcmd.request_addr), count);
+            return true;
+          }),
+      std::end(pending_polls));
```
sepalani, line 793: Is there a reason not to use `pending_polls.end()`? It will get rid of the `iterator` header if I'm not mistaken.

Source/Core/Core/IOS/Network/IP/Top.cpp
```diff
@@ -621,55 +619,24 @@ IPCCommandResult NetIPTop::HandlePollRequest(const IOCtlRequest& request)
...
-    // Memory::Write_U32(ufds[i].fd, request.buffer_out + 0xc*i); //fd
-    // Memory::Write_U32(events, request.buffer_out + 0xc*i + 4); //events
-    Memory::Write_U32(revents, request.buffer_out + 0xc * i + 8);  // revents
-
-    DEBUG_LOG(IOS_NET, "IOCTL_SO_POLL socket %d wevents %08X events %08X revents %08X", i, revents,
-              ufds[i].events, ufds[i].revents);
-  }
-
-  return GetDefaultReply(ret);
+  sm.AddPollCommand({request.address, request.buffer_out, std::move(ufds), timeout});
```
sepalani, line 638: Maybe you should leave a comment to explain the reason behind this as SO_POLL isn't async and returns a reply. Something along those lines:
```C++
// Prevent to block the emulation on blocking poll
```

Source/Core/Core/IOS/Network/Socket.cpp
```diff
@@ -712,24 +715,82 @@ void WiiSockMan::Update()
       socket_iter = WiiSockets.erase(socket_iter);
     }
   }
-  s32 ret = select(nfds, &read_fds, &write_fds, &except_fds, &t);
+  s32 ret = poll(ufds.data(), static_cast<u32>(ufds.size()), 0);
+  if (ret < 0)
+  {
+    ERROR_LOG(IOS_NET, "SO_POLL poll failed (%d)", GetNetErrorCode(ret, "Update", true));
```
sepalani, line 721: Could you move the GetNetErrorCode just above the ERROR_LOG, I don't want to take the risk of #3143 happenning again...
```C++
if (ret < 0)
{
  const s32 error_code = GetNetErrorCode(ret, "Update", true);
```

sepalani:
@dolphin-emu-bot rebuild

Source/Core/Core/IOS/Network/IP/Top.cpp
```diff
@@ -599,77 +605,47 @@ IPCCommandResult NetIPTop::HandleInetNToPRequest(const IOCtlRequest& request)
 
 IPCCommandResult NetIPTop::HandlePollRequest(const IOCtlRequest& request)
 {
-  // Map Wii/native poll events types
-  struct
+  WiiSockMan& sm = WiiSockMan::GetInstance();
+
+  if (!request.buffer_in || !request.buffer_out)
   {
```
sepalani, line 611: Braces can be removed if the body's one line.

Source/Core/Core/IOS/Network/Socket.cpp
```diff
@@ -685,25 +686,26 @@ s32 WiiSockMan::DeleteSocket(s32 s)
...
     WiiSocket& sock = socket_iter->second;
     if (sock.IsValid())
     {
-      FD_SET(sock.fd, &read_fds);
-      FD_SET(sock.fd, &write_fds);
-      FD_SET(sock.fd, &except_fds);
-      nfds = std::max(nfds, sock.fd + 1);
+#ifdef _WIN32
+      // On Win32, WSAPoll only accepts SOCKET types as file descriptors
+      ufds.push_back({static_cast<SOCKET>(sock.fd), read_event | write_event, 0});
```
sepalani, line 705: You will need to remove Windows' unsupported events here with `UNSUPPORTED_WSAPOLL`. Feel free to move it to `Socket.h`.

sepalani:
@dolphin-emu-bot rebuild

Source/Core/Core/IOS/Network/Socket.cpp
```diff
@@ -685,25 +686,23 @@ s32 WiiSockMan::DeleteSocket(s32 s)
 
 void WiiSockMan::Update()
 {
-  s32 nfds = 0;
```
sepalani, line 688: You shouldn't call `poll` if WiiSockets is empty. Otherwise, it spams error logs on Windows. Regardless, you still need to clear previous ufds values and update poll commands.

sepalani:
@dolphin-emu-bot rebuild

sepalani:
It seems Windows is still having issues, I'll have look to see what is going on.

Source/Core/Core/IOS/Network/IP/Top.h
```diff
@@ -13,6 +13,13 @@
 #include <ws2tcpip.h>
 #endif
 
+// WSAPoll doesn't support POLLPRI and POLLWRBAND flags
+#ifdef _WIN32
+#define UNSUPPORTED_WSAPOLL POLLPRI | POLLWRBAND
```
sepalani, line 18: Sorry, that was an oversight of #3316. It was missing parentheses.

Can you please change it to a `constexpr` constant instead
```C++
constexpr int UNSUPPORTED_WSAPOLL = POLLPRI | POLLWRBAND;
```

sepalani:
@dolphin-emu-bot rebuild

sepalani:
This PR introduces a regression in DJ Hero. I'll try to see where it's from.

### On master
```
Network - DemonWare::SetAuthenticationAddress AUTH IP = 185.34.107.68   LSG IP = 185.34.107.72
BeginAccountStage - Success, ENETWORK_MANAGER_ACCOUNT.
InternalAccountUpdate - Success, ENETWORK_MANAGER_LOBBY.
InternalLobbyConnectUpdate - Success, ENETWORK_MANAGER_READY.
**** Lobby connected successfully ****
```

### On this PR
```
Network - DemonWare::SetAuthenticationAddress AUTH IP = 185.34.107.68   LSG IP = 185.34.107.72
BeginAccountStage - Success, ENETWORK_MANAGER_ACCOUNT.
CNetworkAuthenticateAccountState - Authorization server failed in an unexpected manner.
Failed to connect to lobby service.
CNetworkManager::Fail_InternalAccount()
****** Global Network Error: 12.
```

nbouteme:
Does it require real nand certificates to reproduce?
What are the required steps?

sepalani:
You'll need a real NAND and find the logging function (OSReport, ___blank, etc.) which should be found using "Generate symbols from > Signature database" if you're lucky. After, that these messages should appear on the log console if OSReport logs are enabled.

At first glance, it seems to be a timing-issue since I can reproduce the issue on master if I put a breakpoint on the SOConnect function and wait about 3~5 seconds before resuming.

 - On master, the game says that the DJ Hero servers are unavailable.
 - On your PR, it says that the game servers are unavailable.

nbouteme:
Unfortunately I don't have access to a real nand at the moment.
Does IOCTL_SO_POLL appear in the log at debug level?

sepalani:
No, it doesn't. I just hope, that's not yet another issue with `WSAPoll` implementation, otherwise, it will be a blocker for Windows.

nbouteme:
I managed to reproduce the error message without certificates under Linux. OSReport is detected without generating symbols from the signature database, but I don't see these messages.

sepalani:
There are multiple logging functions, so you might need to generate symbols or do a bit of reverse-engineering to find this one.

nbouteme:
It actually does a SO_POLL call, see attached log.
[dolphin.log](https://github.com/dolphin-emu/dolphin/files/4697606/dolphin.log)

nbouteme:
Ok, I've figured it out.

It seems like when you do a poll with both POLLOUT and POLLWRNORM, the OS won't set POLLWRNORM even though they are equivalent. I think the most obvious solution is to re poll with a 0 timeout with the correct events.

sepalani:
@dolphin-emu-bot rebuild

sepalani:
IMHO, that is not the best way to address this issue since:
 - Some FDs might become invalid (after a `close` for instance) on the second `poll` call and it might fail.
 - Some `poll` implementations might clear or check `revents` before assigning new values to it.
 - `WSAPoll` seems to shadow some error codes.

Unfortunately, the best way I can see to address the issue would be to revert back to `select` on `Update` and call `poll` for each `pending_polls` with the appropriate events set on `UpdatePendingPolls`.

I expect small performance impacts on games using `poll` but at the very list it won't be blocking anymore.

nbouteme:
Shouldn't most of these issues be solved simply by lazily pruning closed sockets from each `pending_commands` ? I don't see what reverting back to `select` will accomplish. The remaining issue is that `poll` may get called on `pending_commands` after a savestate load, so I probably should just clear the revents fields before ending the update, and not poll commands with non-zero revents.

sepalani:
Reverting back to `select` will fix `Update` regressions on Windows since `WSAPoll` introduces issues due to its poll implementation being different.

On top of that you pointed out an issue forcing Dolphin to use poll twice while monitoring all events. In other word, I'm pretty sure that relying on such behaviours will cause other issues since not all poll implementations behave the same way.

In sum, it's safer and less error-prone (and probably more accurate) to store poll events in the PollCommand structure and call each requested poll in an async manner in `UpdatePollCommands`.

sepalani:
@dolphin-emu-bot rebuild

sepalani:
@dolphin-emu-bot rebuild

Source/Core/Core/IOS/Network/Socket.cpp
```diff
@@ -730,6 +732,78 @@ void WiiSockMan::Update()
...
+            auto& pfds = pcmd.wii_fds;
+            // Happens only on savestate load
+            int count;
+            if (pcmd.wii_fds[0].revents & error_event)
+            {
+              count = static_cast<int>(pcmd.wii_fds.size());
+            }
+            else
+            {
+              // Update FDs to a negative FD in case they were closed:
```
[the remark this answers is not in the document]
nbouteme: I see, I didn't know IOS also ignored invalid FDs. What you suggested seems to be the only way to guarantee consistent cross-platform behavior. Thanks for the advices.

Source/Core/Core/IOS/Network/IP/Top.cpp
```diff
@@ -599,77 +598,45 @@ IPCCommandResult NetIPTop::HandleInetNToPRequest(const IOCtlRequest& request)
...
-  u32 timeout = Memory::Read_U32(request.buffer_in + 4);
+  if (!request.buffer_in || !request.buffer_out)
+    return GetDefaultReply(-SO_EINVAL);
 
-  int nfds = request.buffer_out_size / 0xc;
-  if (nfds == 0)
+  // Negative timeout indicates wait forever
+  const s64 timeout = static_cast<s64>(Memory::Read_U64(request.buffer_in));
+
+  const int nfds = request.buffer_out_size / 0xc;
```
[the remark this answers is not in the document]
nbouteme: Unfortunately, Windows doesn't define nfds_t. We could get away with some hairy meta programming, but I'll settle for unsigned instead.

Source/Core/Core/IOS/Network/Socket.cpp
```diff
@@ -730,6 +732,78 @@ void WiiSockMan::Update()
...
+
+  pending_polls.erase(
+      std::remove_if(
+          pending_polls.begin(), pending_polls.end(),
+          [this](auto& pcmd) {
+            const auto request = Request(pcmd.request_addr);
+            auto& pfds = pcmd.wii_fds;
+            // Happens only on savestate load
+            int count;
+            if (pcmd.wii_fds[0].revents & error_event)
```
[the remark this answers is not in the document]
nbouteme: Wouldn't games get confused from receiving SO_ENETRESET after they initialized the SO module? I think the worst they are expected to handle is every socket hanging up.

sepalani:
@dolphin-emu-bot rebuild

sepalani:
Here is an [hardware test](https://github.com/dolphin-emu/dolphin/files/4707877/wii-sopoll.zip):
https://gist.github.com/sepalani/21db2c0d703f0350590cab76062360b5

Regarding the results:
 - The valid `nfds` are `1 <= nfds <= 24`.
 - Negative and invalid fds are "ignored" *but ...*
 - `revents` is **always** clear/set to zero.
 - The pollfd order doesn't seem to matter since they are all set properly if an event occurs.

Run the commented tests at your own risk. The result is the one expected, `poll` will work when the fds become valid **BUT** it seems to break most network features afterward. Basically, you'll have IOS_IOctl(v) returning successful return values, however, you can't interact with the socket properly. Listening on a port appears to work (based on the return values from IOS) but doesn't when you try to connect to it.

Source/Core/Core/IOS/Network/Socket.cpp
```diff
@@ -730,6 +733,92 @@ void WiiSockMan::Update()
...
+            // Happens only on savestate load
+            if (pcmd.wii_fds[0].revents & error_event)
+            {
+              count = static_cast<int>(pcmd.wii_fds.size());
+            }
+            else
+            {
+              // Update FDs to a negative FD in case they were closed:
+              // sockets that were closed during SO_POLL on IOS are ignored
+              for (u32 i = 0; i < pfds.size(); ++i)
```
sepalani, line 774: According to the hardware test, you don't need to take care of preserving the original order. All entries in the set will be set properly. So you can use ```std::copy_if``` or ```std::partition``` directly on the vector (without changing the fd to a negative value) and use ```std::distance``` or similar to get the size that will be used by ```poll```.

Source/Core/Core/IOS/Network/Socket.cpp
```diff
@@ -730,6 +733,92 @@ void WiiSockMan::Update()
...
+              // move everything back to where they were
+              for (auto i = 0; i < mid - original_order.begin(); ++i)
+                std::swap(pfds[i], pfds[original_order[i]]);
+            }
+
+            if (count == 0 && pcmd.timeout)
+              return false;
+
+            // Translate native to Wii events,
+            // The IOS SO_POLL implementation doesn't overwrite the revents field
```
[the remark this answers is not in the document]
nbouteme: Yeah my bad. But does it clear revents before or after poll is completed?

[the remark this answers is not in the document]
sepalani: It clears `revents` before the poll completion, right after the `nfds` sanity check. See https://github.com/dolphin-emu/dolphin/pull/8813#discussion_r432963316

In sum, you just need to write back `revents` from all entries like the original code:
```c++
for (std::size_t i = 0; i < pfds.size(); ++i)
// Remove the if (...) continue check
```

Source/Core/Core/IOS/Network/Socket.cpp
nbouteme: I'm not sure what this means. I need to somehow preserve the order of the pollfd vector to set the correct revents fields when SO_POLL completes, and I'm changing fds to a negative value because I can't pass invalid fds to poll.
I also need to restore the ordering after each poll because SO_POLL may not actually complete (count == 0). 
How do I properly set the revents field when SO_POLL completes if `pfds` doesn't have the same order as what was originally passed to SO_POLL?

sepalani: I thought there might be a simpler solution but since we have to write back `revents` based on the index there might not be. 

However, regarding the negative fds, you can perform the check in `std::partition` rather than assigning a negative fd. Since you move valid fds to the beginning and pass the appropriate size to `poll`, invalid ones aren't passed to poll if I understand the code correctly.

sepalani:
@dolphin-emu-bot rebuild

Source/Core/Core/IOS/Network/IP/Top.cpp
```diff
@@ -599,77 +598,45 @@ IPCCommandResult NetIPTop::HandleInetNToPRequest(const IOCtlRequest& request)
...
+  if (!request.buffer_in || !request.buffer_out)
+    return GetDefaultReply(-SO_EINVAL);
 
-  int nfds = request.buffer_out_size / 0xc;
-  if (nfds == 0)
-    ERROR_LOG(IOS_NET, "Hidden POLL");
+  // Negative timeout indicates wait forever
+  const s64 timeout = static_cast<s64>(Memory::Read_U64(request.buffer_in));
+
+  const unsigned nfds = request.buffer_out_size / 0xc;
```
sepalani, line 609:
```C++
const u32 nfds
```

Source/Core/Core/IOS/Network/IP/Top.cpp
```diff
@@ -599,77 +598,45 @@ IPCCommandResult NetIPTop::HandleInetNToPRequest(const IOCtlRequest& request)
...
+  if (nfds <= 0 || nfds > WII_SOCKET_FD_MAX)
+  {
+    ERROR_LOG(IOS_NET, "IOCTL_SO_POLL failed: Invalid array size %d, ret= %d", nfds, -SO_EINVAL);
+    return GetDefaultReply(-SO_EINVAL);
+  }
 
   std::vector<pollfd_t> ufds(nfds);
 
-  for (int i = 0; i < nfds; ++i)
+  for (unsigned i = 0; i < nfds; ++i)
```
sepalani, line 618:
```C++
u32 i = 0;
```

Source/Core/Core/IOS/Network/Socket.cpp
```diff
@@ -730,6 +733,84 @@ void WiiSockMan::Update()
...
+            auto& pfds = pcmd.wii_fds;
+            int count = 0;
+            // Happens only on savestate load
+            if (pcmd.wii_fds[0].revents & error_event)
+            {
+              count = static_cast<int>(pcmd.wii_fds.size());
+            }
+            else
+            {
+              // Update FDs to a negative FD in case they were closed:
```
sepalani, line 772: This comment needs to be changed to something more fitting like:
```C++
// Make the behavior of poll consistent across platforms by not passing:
//  - Set with invalid fds, revents is set to 0 (Linux) or POLLNVAL (Windows)
//  - Set without a valid socket, raises an error on Windows
```

Source/Core/Core/IOS/Network/Socket.cpp
```diff
@@ -730,6 +733,84 @@ void WiiSockMan::Update()
...
+              // sockets that were closed during SO_POLL on IOS are ignored
+              std::vector<int> original_order(pfds.size());
+              std::iota(original_order.begin(), original_order.end(), 0);
+              // Select indices with valid fds
+              auto mid = std::partition(original_order.begin(), original_order.end(), [&](auto i) {
+                return GetHostSocket(Memory::Read_U32(pcmd.buffer_out + 0xc * i)) > 0;
+              });
+              // move all the valid pollfds to the front of the vector
+              for (auto i = 0; i < mid - original_order.begin(); ++i)
```
sepalani, line 781: Can't `mid - original_order.begin()` be replaced with `n_valid`?

Source/Core/Core/IOS/Network/Socket.cpp
```diff
@@ -730,6 +733,84 @@ void WiiSockMan::Update()
...
+            {
+              // Update FDs to a negative FD in case they were closed:
+              // sockets that were closed during SO_POLL on IOS are ignored
+              std::vector<int> original_order(pfds.size());
+              std::iota(original_order.begin(), original_order.end(), 0);
+              // Select indices with valid fds
+              auto mid = std::partition(original_order.begin(), original_order.end(), [&](auto i) {
+                return GetHostSocket(Memory::Read_U32(pcmd.buffer_out + 0xc * i)) > 0;
+              });
+              // move all the valid pollfds to the front of the vector
```
sepalani, line 780:
```C++
// Move
```

Source/Core/Core/IOS/Network/Socket.cpp
```diff
@@ -730,6 +733,84 @@ void WiiSockMan::Update()
...
+              for (auto i = 0; i < mid - original_order.begin(); ++i)
+                std::swap(pfds[i], pfds[original_order[i]]);
+
+              const auto n_valid = std::distance(original_order.begin(), mid);
+              if (n_valid > 0)
+                count = poll(pfds.data(), n_valid, 0);
+              if (count < 0)
+                count = GetNetErrorCode(count, "UpdatePollCommands", false);
+
+              // move everything back to where they were
```
sepalani, line 790:
```C++
// Move
```

Source/Core/Core/IOS/Network/Socket.cpp
```diff
@@ -730,6 +733,84 @@ void WiiSockMan::Update()
...
+      pcmd.timeout = std::max<s64>(0, pcmd.timeout - elapsed);
+  }
+
+  pending_polls.erase(
+      std::remove_if(
+          pending_polls.begin(), pending_polls.end(),
+          [this](auto& pcmd) {
+            const auto request = Request(pcmd.request_addr);
+            auto& pfds = pcmd.wii_fds;
+            int count = 0;
```
sepalani, line 764: This should probably be renamed `ret` as it might not return the count but a network error code.

sepalani:
@dolphin-emu-bot rebuild

sepalani:
Could you please change/remove the commit description:
 > This also replaces the select() call in WiiSockMan::Update by a poll call ...

Do you have any games or homebrew that use SOPoll so we can test this PR against?

nbouteme:
Well besides DJ Hero and your test suite, no, I don't. I'll try to look around if I find some larger homebrew.

nbouteme:
On the homebrew side:
[WiiRadio](https://github.com/Metras/wiiradio/)
[WiiFlow Lite](https://github.com/Fledge68/WiiFlow_Lite)
[Quake GX](https://github.com/Stealthii/QuakeGX)

Source/Core/Core/IOS/Network/Socket.cpp
```diff
@@ -730,6 +733,87 @@ void WiiSockMan::Update()
...
+  pending_polls.erase(
+      std::remove_if(
+          pending_polls.begin(), pending_polls.end(),
+          [this](auto& pcmd) {
+            const auto request = Request(pcmd.request_addr);
+            auto& pfds = pcmd.wii_fds;
+            int ret = 0;
+
+            // Happens only on savestate load
+            if (pcmd.wii_fds[0].revents & error_event)
```
sepalani, line 767:
```C++
if (pfds[0].revents & error_event)
```

Source/Core/Core/IOS/Network/Socket.cpp
```diff
@@ -730,6 +733,87 @@ void WiiSockMan::Update()
...
+            {
+              ret = static_cast<int>(pcmd.wii_fds.size());
```
sepalani, line 769:
```c++
ret = static_cast<int>(pfds.size());
```

nbouteme:
I'm sorry, I'm not much proficient with git. Isn't amending just git commit --amend and force push, or am I forgetting something here? Or is the description still not what you wanted?

sepalani:
@nbouteme No worries. Feel free to ask or go over IRC if you're having troubles with git.

In fact, you're amending your commit but not the commit message/description.
https://help.github.com/en/github/committing-changes-to-your-project/changing-a-commit-message

During the amend process, you need to change/remove the commit description (i.e. the sentences under the commit message) with you text editor. Alternatively, you can do it directly from command-line:
```bash
git commit --amend -m "IOS/Network: Make poll complete asynchronously"
```

If you want to change the commit description you'll need to pass an extra `-m` flag:
```bash
git commit --amend -m "commit message" -m "commit description"
```

You can check your commit history using `git log` and see if amending worked. After that, you can force push.

sepalani:
@dolphin-emu-bot rebuild

Source/Core/Core/IOS/Network/IP/Top.cpp
```diff
@@ -599,77 +598,45 @@ IPCCommandResult NetIPTop::HandleInetNToPRequest(const IOCtlRequest& request)
...
+    return GetDefaultReply(-SO_EINVAL);
 
-  int nfds = request.buffer_out_size / 0xc;
-  if (nfds == 0)
-    ERROR_LOG(IOS_NET, "Hidden POLL");
+  // Negative timeout indicates wait forever
+  const s64 timeout = static_cast<s64>(Memory::Read_U64(request.buffer_in));
+
+  const u32 nfds = request.buffer_out_size / 0xc;
+  if (nfds <= 0 || nfds > WII_SOCKET_FD_MAX)
```
sepalani, line 610:
```C++
if (nfds == 0 || nfds > WII_SOCKET_FD_MAX)
```

Source/Core/Core/IOS/Network/Socket.cpp
```diff
@@ -730,6 +733,87 @@ void WiiSockMan::Update()
...
+            else
+            {
+              // Make the behavior of poll consistent across platforms by not passing:
+              //  - Set with invalid fds, revents is set to 0 (Linux) or POLLNVAL (Windows)
+              //  - Set without a valid socket, raises an error on Windows
+              std::vector<int> original_order(pfds.size());
+              std::iota(original_order.begin(), original_order.end(), 0);
+              // Select indices with valid fds
+              auto mid = std::partition(original_order.begin(), original_order.end(), [&](auto i) {
+                return GetHostSocket(Memory::Read_U32(pcmd.buffer_out + 0xc * i)) > 0;
```
sepalani, line 780: AFAIK, 0 is a valid socket fd, so it should be `>= 0`.

Source/Core/Core/IOS/Network/IP/Top.cpp
```diff
@@ -599,77 +598,45 @@ IPCCommandResult NetIPTop::HandleInetNToPRequest(const IOCtlRequest& request)
...
+  {
+    ERROR_LOG(IOS_NET, "IOCTL_SO_POLL failed: Invalid array size %d, ret= %d", nfds, -SO_EINVAL);
```
sepalani, line 612: `ret=%d` (without the space)

sepalani:
@dolphin-emu-bot rebuild

sepalani:
@nbouteme Could you please rebase your PR to the latest master and update the `STATE_VERSION`?

sepalani:
@dolphin-emu-bot rebuild

JMC47:
Is there any particular game or case that this helps so I could properly make a demonstration of how it affects things?  If it's all internally changing things, that's fine too, I just wanted to know if I could make some media for the Progress Report.

Thanks!

nbouteme:
Thanks for merging, and also thanks a lot to sepalani who helped a lot to make this a proper PR.

Unfortunately I don't think this change would be really visible to end users : it theoretically eliminates stutters, or possible deadlocks causing games to hang, when a game uses SO_POLL with a non-zero timeout, but I'm don't really know whether wii games do that often. I don't think they do, or else the stutter would have probably broken some games and this would have been fixed way earlier. On the other hand, it may also help with some homebrews but I don't think this is most users use cases. My use case was a mod to add some online functionnality to a game that lacked it, but using poll with a timeout of 0 didn't sit right with me... It's not as efficient as I'd like it to be because in the end, poll and select still need to be called with a 0 timeout, since IOS HLE runs on the emulation thread. This still allows people to write mods and have a behavior that's more in line with what would happen on a real wii. A thread waiting on a SO_POLL call to return will remain dormant instead of being immediately woken up then immediately sleep again, so it reduces CPU usage on PPC.

If IOS HLE ran on a separate thread, then the socket polling mechanism wouldn't be executed as often and would reduce CPU usage in HLE code by a bit, but this poses a lot of harder challenges. But if/when this will eventually be done, then anything that uses non-zero timeouts on SO_POLL will use a bit less CPU. So yeah, it's not that big of a deal in the end, it's more like a convenience for mod developers.

JMC47:
Yeah, this mixed with other async PRs seem to have eliminated some of the stuttering when connecting to servers online.

If there's no great demonstration yet, once more of it is done I'll definitely be able to show off a bigger change.  Mario Kart Wii in particular used to be a mess of 3 - 4 second stutters while connecting, and it's slowly been getting better.

PR #8815, opened 2020-05-17
InputCommon/XInput2: Added a state buffer to button and key presses to prevent missed inputs

TheConfuZzledDude:
This adds a buffer that is set on a key or button press, and is cleared when the state is next polled. Previously, there was no buffer at all, so events like the mouse wheel events (Buttons 4 and 5), where the press and release events are received at the same time, would almost always not be registered, unless Dolphin happened to poll it in between processing the two

### DIFF
--- a/Source/Core/InputCommon/ControllerInterface/Xlib/XInput2.cpp
+++ b/Source/Core/InputCommon/ControllerInterface/Xlib/XInput2.cpp
@@ -175,7 +175,7 @@ KeyboardMouse::KeyboardMouse(Window window, int opcode, int pointer, int keyboar
   XDisplayKeycodes(m_display, &min_keycode, &max_keycode);
   for (int i = min_keycode; i <= max_keycode; ++i)
   {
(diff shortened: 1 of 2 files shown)


JosJuice:
@dolphin-emu-bot rebuild

iwubcode:
I haven't really experienced this but would it be worthwhile to do this for Windows DInput as well?

TheConfuZzledDude:
I think it would probably be more correct to do something similar, though I'm not sure. This was mostly necessary here because of how X11 treats scroll wheel inputs as button presses, whereas DInput treats them as another mouse axis (X11 actually does too but that isn't implemented here yet). For this to actually be an issue you need to literally press the button for less than a frame, which I guess could happen, if only with auto-clickers, etc.

TheConfuZzledDude:
I'm closing this (a bit late lol) since it seems quite difficult to implement this is in a completely correct way, and there's some other possibilities for fixing scroll wheel inputs on Linux (i.e. treating it as an axis)

PR #8817, opened 2020-05-17
CMake: support using system-wide zstd library

Tilka:
CMake does not have native support for zstd yet, so find_package() doesn't work out of the box.

@JosJuice: Note that #8538 will need to link against `zstd` instead of `zstd::zstd` since CMake's FindPkgConfig module doesn't add an alias like Externals/zstd/CMakeLists.txt does.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -608,7 +608,7 @@ else()
   add_subdirectory(Externals/liblzma)
 endif()
 
-find_package(zstd)
+pkg_search_module(ZSTD QUIET libzstd)
 if(ZSTD_FOUND)
   message(STATUS "Using shared zstd")
 else()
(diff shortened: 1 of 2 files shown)


Externals/zstd/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-project(bzip2 C)
+project(zstd C)
```
JosJuice, line 1: Oops...

CMakeLists.txt
```diff
@@ -608,7 +608,7 @@ else()
   add_subdirectory(Externals/liblzma)
 endif()
 
-find_package(zstd)
+pkg_search_module(ZSTD QUIET libzstd)
```
degasus, line 611: I'd prefer if we provide a Findzstd.cmake helper instead.

Tilka: Why?

PR #8819, opened 2020-05-18
DolphinQt: Fix the panic alert deadlock, Pause on Focus Loss edition

JosJuice:
why are there so many ways to trigger this issue


dreamsyntax:
Just wanted to mention these fixes are very much appreciated. It's annoying how often they would occur.

PR #8820, opened 2020-05-19
Core / VideoCommon / DolphinQt - allow tweaking freelook camera's field of view

iwubcode:
In one of my other PRs a user wanted to be able to modify the freelook field of view.  I figured...why  not?

The uses are likely rather niche but:

- Tweak the view to show a bit more of the world
- You can use this to "zoom" the camera without colliding into objects
- Or you might want some funky pictures!

I didn't expose the near/far planes.

image

image


iwubcode:
@DerKoun

Source/Core/VideoCommon/VertexShaderManager.cpp
```diff
@@ -355,14 +355,17 @@ void VertexShaderManager::SetConstants()
     switch (xfmem.projection.type)
     {
     case GX_PERSPECTIVE:
-      g_fProjectionMatrix[0] = rawProjection[0] * g_ActiveConfig.fAspectRatioHackW;
+    {
+      Common::Vec2 fov =
```
lioncash, line 359:
```suggestion
      const Common::Vec2 fov =
```

DerKoun:
> @DerKoun - is this what you were wanting? Please let me know

Yes, it's exactly what I meant. Thanks for implementing it. (You spared a poor Java guy 2-3 circles of hell figuring it all out.)

If I may add some thoughts on the UI side:
- At first I thought that having separate X and Y settings was unnecessary, but as we are indeed absolutely niche already, why not. It would be quite convenient if there was a chekbox that disables the Y setting, using the X value for both.
- Maybe I missed it in the code-diff, but it would be great if the zoom (same for both dimensions) could be in/decreased via keys, like moving the free-look camera. Would ease tweaking the setting (and taking the perfect virtual photo or just playing around).

Use cases I can think of are:
- When stereoscopic 3D makes the left and right edges hard to focus on, you can zoom out a bit to not miss anything.
- When widescreen exposes just a bit to much you can zoom in slightly to compensate.
- It also gives another tool to people that make virtual "behind the scenes" videos.

MayImilae:
Gave this a try! First of all, it works as advertised. It changes the view FoV very nicely! But I have two notes.

Shift-R doesn't reset the FoV. That's REALLY needed! Everything should be reset too default when pressing reset!

There needs to be a set of hotkeys for FoV in and out. Personally, I'd suggest having both X and Y mapped to the same default buttons, so that it changes the FoV in and out without distorting the image. That is the most common use case that I can imagine, and it's less buttons to fuss with, so I think that's a sensible default. I certainly won't be using it in any other way! So with that in mind, I suggest shift+mousewheel as the default, as that would work really well, imo. And of course it needs to be in the hotkeys panel and what not, and to be able to separate them in hotkeys if someone wants that.

iwubcode:
@MayImilae / @DerKoun - thank you so much for your feedback.

I removed the configuration (that's why you couldn't reset FoV previously) and made the values entirely based around hotkeys.  This will help with something I'm planning on for later.

Please let me know if you see any other concerns.

MayImilae:
Gave this another look over! Here's a few notes.

First of all, I had to reset my hotkeys to default to get the shift+scroll wheel shortcut for FoV. I don't believe there is anything we can do about that but that is definitely something I'll need to mention in the Progress Report so this is kind of my reminder.

Now to a real note. :P The "Freelook Increase Speed" and "Freelook Decrease Speed" buttons have no effect on FoV. That's something that should be changed, as by default it is HELLA sensitive and would greatly benefit from being able to make it slow down a bit.

So the sensitivity by default is much too high on the hotkeys when enlarging the FoV ("zooming out"). Just three clicks on my scroll wheel is enough to _max it out_. That's way WAY too sensitive! Fortunately tightening FoV ("zooming in") isn't so sensitive, but yea, something needs to be adjusted there! 

Reset works! 

So yea, all in all, just sensitivity adjustment for the hotkeys and applying the speed increase/decrease effects so finer control is easier. With that it should be good to go! 

Note: Changing the mouse scroll speed does not have any effect on the FoV adjustment speed with the hotkeys. I didn't think it would but I would have felt really dumb if I didn't check it and that turned out to be a factor, so um, I checked it.

iwubcode:
@MayImilae - thank you for taking another look!

I did think about your suggestion during implementation (increasing via speed modifier) but I didn't really like it.  Personally, speed should be for movement, it makes no sense for speed to effect FoV.  Additionally, I could see users wanting a different movement rate compared to their field of view.

I see a couple of general solutions to this:

* Adding another increase / decrease / reset button for FoV
* Adding a UI that shows and allows to edit the current value (not really feasible with the current UI, need to wait for 8867)

My personal take would be to add a new tab for FoV.  Then add the increase/decrease/reset buttons under a "rate of change" section.

Once 8867 gets merged, I can add UI that allows for more direct control.

Of course I could see just waiting on this until 8867 gets merged, then things can be done in a more clean manner.

MayImilae:
While having the increase and decrease speed function for FoV would be nice, it's not really necessary, especially if you want to implement it in some other way in the future. My only real issue with this PR  right now is that enlarging the FoV ("zooming out") via hotkeys is so extreme it's basically useless. So maybe adjust it so that enlarging the FoV has finer control with the hotkey, and we can call it?

iwubcode:
@MayImilae - I agree with you, I modified the step size to be small,er let me know if you think it feels better?

Also, it's a bit silly but one of my new favorite things is playing Mario Kart "zoomed out" (or at least pretty far) and then move the camera in.  Things look far away initially but quickly jump into your view.  Kinda feels like I'm playing f-zero :P

Tilka:
I can't think of a reasonable use case for separate x/y values and having both impacts the ease of configuring the common case where the user wants them to be the same.

iwubcode:
@Tilka - I see what you mean but also don't think this solution is very bad either?  It's 4 buttons instead of 2 and we default to the same button (middle mouse wheel) and until #8747 is merged, I doubt useres would change it.

It's hard to say how users will use this.  One potential reason is ultrawide monitors.  I don't own one so I can't say how it'd look on that type of display.

PR #8822, opened 2020-05-20
DolphinQt: Fix freelook description typos

JosJuice:
cControls -> controls, axis -> axes, removed leading spaces.


iwubcode:
Thanks @JosJuice !  Legitimately didn't know that the plural of axis was axes, very neat.  Learn something new every day..

PR #8823, opened 2020-05-22
DolphinQt: Replace setting windows with unified settings window!

iwubcode:
This was loosely based off of @mbc07 's comment on #8785 .

I patterned it after Visual Studio's options window.  Ex:

image

I think the multi-tiered approach may be useful for the future, as we add more and more settings.

**Opening the window:**

image

**Search:**

image

### DIFF
--- a/Source/Core/DolphinQt/GrandSettingsWindow.cpp
+++ b/Source/Core/DolphinQt/GrandSettingsWindow.cpp
@@ -1,0 +1,233 @@
+// Copyright 2020 Dolphin Emulator Project
+// Licensed under GPLv2+
+// Refer to the license.txt file included.
+
+#include <QCompleter>
+#include <QDialogButtonBox>
+#include <QGridLayout>
+#include <QLineEdit>
+#include <QStackedWidget>
+#include <QString>
+#include <QTreeWidget>
+
+#include "Common/Config/Config.h"
+#include "Core/Config/MainSettings.h"
+#include "Core/ConfigManager.h"
+
+#include "DolphinQt/Config/CommonControllersWidget.h"
+#include "DolphinQt/Config/FreeLookWidget.h"
+#include "DolphinQt/Config/GamecubeControllersWidget.h"
+#include "DolphinQt/Config/Graphics/AdvancedWidget.h"
+#include "DolphinQt/Config/Graphics/EnhancementsWidget.h"
+#include "DolphinQt/Config/Graphics/GeneralWidget.h"
+#include "DolphinQt/Config/Graphics/HacksWidget.h"
+#include "DolphinQt/Config/Graphics/SoftwareRendererWidget.h"
+#include "DolphinQt/Config/WiimoteControllersWidget.h"
+#include "DolphinQt/GrandSettingsWindow.h"
+#include "DolphinQt/Settings/AdvancedPane.h"
+#include "DolphinQt/Settings/AudioPane.h"
+#include "DolphinQt/Settings/GameCubePane.h"
+#include "DolphinQt/Settings/GeneralPane.h"
+#include "DolphinQt/Settings/InterfacePane.h"
+#include "DolphinQt/Settings/PathPane.h"
+#include "DolphinQt/Settings/WiiPane.h"
+
+#include "VideoCommon/VideoBackendBase.h"
+#include "VideoCommon/VideoConfig.h"
+
+GrandSettingsWindow::GrandSettingsWindow(X11Utils::XRRConfiguration* xrr_config, QWidget* parent)
+    : GraphicsDialog(parent), m_xrr_config(xrr_config)
+{
+  // Set Window Properties
+  setWindowTitle(tr("Grand Settings"));
+  setWindowFlags(windowFlags() & ~Qt::WindowContextHelpButtonHint);
+
+  m_search_bar = new QLineEdit();
+  m_search_bar->setPlaceholderText(tr("Search settings"));
+  m_search_bar->setClearButtonEnabled(true);
+
+  m_widget_stack = new QStackedWidget();
+
+  m_tree = new QTreeWidget();
+  m_tree->setHeaderHidden(true);
+
+  PopulateWidgets();
+  OnBackendChanged(QString::fromStdString(Config::Get(Config::MAIN_GFX_BACKEND)));
+
+  connect(m_search_bar, &QLineEdit::textChanged, this, &GrandSettingsWindow::OnSearch);
+
+  // Main Layout
+  QGridLayout* layout = new QGridLayout;
+
+  // Add content to layout before dialog buttons.
+  layout->addWidget(m_search_bar, 0, 0, 1, 1);
+  layout->addWidget(m_tree, 1, 0, 1, 1);
+  layout->addWidget(m_widget_stack, 0, 1, 2, 1);
+
+  // Dialog box buttons
+  QDialogButtonBox* close_box = new QDialogButtonBox(QDialogButtonBox::Close);
+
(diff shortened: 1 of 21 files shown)


iwubcode:
Tonight I added the "Graphics" capability to the window.  Still missing the description box and a few other odds and ends but you can see how it looks.

Next I'll take a stab at inputs.

Techjar:
Close button position is a small improvement but it still needs to be aligned with the bottom of the tree view.

Source/Core/DolphinQt/GrandSettingsWindow.cpp
```diff
@@ -0,0 +1,180 @@
+// Copyright 2020 Dolphin Emulator Project
+// Licensed under GPLv2+
+// Refer to the license.txt file included.
+
+#include <QDialogButtonBox>
+#include <QGridLayout>
+#include <QLineEdit>
+#include <QSortFilterProxyModel>
+#include <QStackedWidget>
+#include <QStringLiteral>
```
Techjar, line 10: This doesn't exist, it should just be `#include <QString>`.

iwubcode:
Tonight I added "Inputs" capability to the window.  The code blows out a bit since those weren't split into separate pieces, so I've basically duplicated things.  At some point I'll replace the current windows code with the separated out ones (if the old windows aren't removed completely).

Inputs concludes the things I wanted to tackle for the settings window.

Next, I plan on adapting the game settings window to the same style

Source/Core/DolphinQt/GrandSettingsWindow.cpp
```diff
@@ -0,0 +1,180 @@
...
+
+GrandSettingsWindow::GrandSettingsWindow(X11Utils::XRRConfiguration* xrr_config, QWidget* parent)
+    : GraphicsDialog(parent), m_xrr_config(xrr_config)
+{
+  // Set Window Properties
+  setWindowTitle(tr("Grand Settings"));
+  setWindowFlags(windowFlags() & ~Qt::WindowContextHelpButtonHint);
+
+  m_search_bar = new QLineEdit();
+  m_search_bar->setPlaceholderText(QStringLiteral("Search settings"));
```
Techjar, line 45: All of these `QStringLiteral` usages should be replaced with `tr` so they get translated.

Source/Core/DolphinQt/GrandSettingsWindow.cpp
iwubcode: Thanks.  Yeah, it does seem to exist for some platforms, ex:  https://github.com/dolphin-emu/ext-win-qt/blob/a14e340be1ad66e0cb1d8180485896a9134800e0/Qt5.14.1/msvc2019_64/include/QtCore/qstringliteral.h

But a quick search shows that Qt mentions `QString` I'll move to that

Source/Core/DolphinQt/GrandSettingsWindow.cpp
iwubcode: I tend to forget about the `tr` thing.  I'll update this, thanks Techjar!

Rumi-Larry:
I would name the new window "Unified Settings" rather than "Grand Settings"

iwubcode:
With the tooltips / removal of the description box and the various PRs for moving the inputs logic to reusable widgets merged, I think this is pretty close to removing the 'draft' state.

Tonight I implemented a poor man's search.  I'll likely rework it to use some sort of tag based approach (right now it just searches the title of the page).  I also cleaned up a few bugs.

Miksel12:
Looks great and feels very streamlined! 
A few nitpicks: the free look settings aren't wraped in a 'window'. I can see how that is unnecessary with only one item on a page but the common page under input does have it and I think it looks more consistent. 

The Wii page has a small visual bug when resizing the window where the field of bluetooth devices gets stuck besides the buttons.

An easy way to get to hotkeys would also be a great improvement I think, maybe a button under common or a seperate tab under input? The same goes for the memory card manager. In general, with a cleaner UI some more option could be integrated into the main settings that are currently hidden in the drop down menu's. Though all that could be done in a follow up PR.

iwubcode:
Thanks for your comment @Miksel12 

> the free look settings aren't wraped in a 'window'. I can see how that is unnecessary with only one item on a page but the common page under input does have it and I think it looks more consistent.

Yeah, at the moment I'm using the exact same layout that are used in the other windows.  That is why FreeLook doesn't have a box around it.

> The Wii page has a small visual bug when resizing the window where the field of bluetooth devices gets stuck besides the buttons. This is actually a bug in master as well.

Huh, wasn't aware of that, I'll look into it.

> An easy way to get to hotkeys would also be a great improvement I think, maybe a button under common or a seperate tab under input?

I thought of that but I wasn't sure if a button "launch hotkeys" was a great solution.  Requiring two steps where it used to take one didn't seem like an improvement to me..

Miksel12:
> I thought of that but I wasn't sure if a button "launch hotkeys" was a great solution. Requiring two steps where it used to take one didn't seem like an improvement to me..
Maybe keeping both is an option? I think that having an option only in a drop down menu limits its 'reachability' for new users.

iwubcode:
Updated to add some tags or "keywords" to the search box.  Added auto-completion and a nice little x button.  Hoping @MayImilae can create a search icon if this ends up getting approved.

@Miksel12 - hmm, I think I see your point.  I don't have a strong opinion either way.. also I reproduced the issue you're talking about.  Will implement a fix

iwubcode:
@Miksel12 - I fixed the wiipane issue.

--

I went ahead and removed all the other settings windows and made this one behave accordingly with the buttons on the main Dolphin bar.

I don't have to do this *yet* if we want to slowly introduce the window but this gives you a good idea of what it would look like!

mbc07:
I didn't do extensive testing with this PR yet but from a quick look the current version looks good and very promising. Some preliminary points:

* Grand Settings window still is unusually tall. Could you try to backport PR #8785 changes to this PR? I think it would be sufficient to address that;
* Free look page looks odd, I'm not entirely sure why. Seems to be completely misaligned if compared to the other pages;
* Speaking of Free Look, I would probably change it from a top-level category to a subcategory of either Graphics or Input categories (the latter might fit better?). I recall you mentioning you plan to expand Free Look options even more in the future but as it stands right now, I don't think it's complex enough to warrant a top-level category;
* Similarly to the previous point, perhaps change "Alternate Input Sources" into its own sub-category of Input instead of having a button under "Common" that opens another window? The rationale was that this separate window would eventually get more tabs besides the existing "DSU Client" tab, but at least for now it remains alone and seems too simple to have its own window, especially now that it's part of the Grand Settings window;
* About Hotkey Settings: I also think it would be a nice addition to have a button for it on the "Common" input category, as it would help with discoverability. Keeping both the button and the existing drop-down menu option is fine too;
* This is a very minor and nitpicky point but it would be nice if the search bar had a magnifying glass icon on the right corner (similar to the Visual Studio settings window).

MayImilae:
@iwubcode Since I keep missing you, please talk with me about this PR on IRC sometime. I have a lot to go over, and it isn't that well suited for this format.

iwubcode:
@MayImilae - someone asked about this recently and I can't recall what the concerns were.  It looks like I forgot to take notes.  I vaguely recall you wanting to provide a mock-up of a design for something but can't remember if this was it.  Do you recall?

iwubcode:
No real updates, just noticed that the usb list was fixed in #10719 so I went ahead and dropped that commit from this PR.

cobalt2727:
What still needs to be done before this PR or something like it can get merged?

iwubcode:
This was a fun experiment but I don't expect this to be merged even if it is rebased.  I will close this.

PR #8824, opened 2020-05-23
Add option to disallow SD card writes

Techjar:
This is mainly useful when you have an SD card which you use for NetPlay and you don't want it accidentally written to when playing outside of NetPlay, as that results in a desync the next time it's used in NetPlay if you don't share your exact SD card with other players.

It's also fair to say that this kind of implements a hardware feature, as physical SD cards have a write lock switch. Although, this doesn't implement the actual IOS functionality that causes the write protection error message in Wii software. (Is this even a thing? Or does simply failing writes produce this behavior?)


Ebola16:
Implements https://bugs.dolphin-emu.org/issues/8965

Can you add a log entry on emulation start if this setting is disabled? I can foresee users disabling this setting for NetPlay and forgetting about it in the future.

Techjar:
Not sure of the appropriate place to put such a log call.

Ebola16:
Maybe after calling `BootManager::BootCore`?

Techjar:
There, that seems like a good place for it.

JMC47:
This has been tested by people within the Brawl modding community and does what is needed.

PR #8825, opened 2020-05-23
split rtc from sram

DacoTaco:
when testing out some stuff about the sram on my cube i noticed reading past rtc acted differently then dolphin did.
reading past the rtc address made the gamecube mirror the same data over and over untill a new read started, and dolphin just proceeded to read SRAM.

in this commit i have split them up, but im not sure about splitting the files up.
it breaks a few things (because of the change in size) and it sounds kinda nasty.
hence this is a draft PR for now until an opinion is given.

keeping them together however means that dumpers ( like swiss ) need to keep this in mind


Source/Core/Core/HW/Sram.cpp
```diff
@@ -13,7 +13,7 @@
 // English
 // This is just a template. Most/all fields are updated with sane(r) values at runtime.
 // clang-format off
-const Sram sram_dump = {Common::BigEndianValue<u32>{0},
+const Sram sram_dump = {//Common::BigEndianValue<u32>{0},
```
BhaaLseN, line 16: You should probably remove this instead of commenting it out.

Source/Core/Core/HW/EXI/EXI_DeviceIPL.h
```diff
@@ -42,8 +42,10 @@ class CEXIIPL : public IEXIDevice
   {
     ROM_BASE = 0,
     ROM_SIZE = 0x200000,
-    SRAM_BASE = 0x800000,
-    SRAM_SIZE = 0x44,
+    RTC_BASE = 0x800000,
+    RTC_SIZE = 0x3,
```
BhaaLseN, line 46: Is the RTC really just 3 bytes instead of 4? YAGCD says 32-bit value as well.

DacoTaco: it is indeed 4, starting at 0x800000 to 0x800003, 0x800004 is sram. if i make the lenght 4, it'll be 5 bytes inc 0x800004 will it not?

BhaaLseN: Uh, why? `0x800000` for `0x4` items is `0x800000`, `0x800001`, `0x800002` and `0x800003`. Or am I misunderstanding what you mean?

DacoTaco: ye no, youre right. its odd that the gamecube IPL could display the correct time then if it technically read a byte to little. 
then again, thats probably cause it does a 4 length read from 0x800000 and not 0x800000,0x800001,0x800002 etc etc , in which the address would get relatched

Source/Core/Core/HW/Sram.h
```diff
@@ -42,6 +42,12 @@ using CardFlashId = std::array<u8, 12>;
 
 #pragma pack(push, 1)
 
+struct Rtc
+{
+  Common::BigEndianValue<u32> rtc;
+  u8& operator[](size_t offset) { return reinterpret_cast<u8*>(&rtc)[offset]; }
```
lioncash, line 48: This should have a const variant

DacoTaco: what do you mean?

lioncash:
```cpp
const u8& operator[](size_t offset) const { return reinterpret_cast<const u8*>(&rtc)[offset]; }

DacoTaco: ah, i thought you were talking about the u32 or struct.
makes sense i guess.
shouldn't the sram struct be the same then? (it isn't/wasn't)

DacoTaco: the const also makes it not modifiable , does it not? :)

Techjar: I believe he means both const and non-const variants should exist.

lioncash: Yup

DacoTaco: mind me asking what the use is of having both the const and non-const variant? 
shouldn't the sram be the same?

DacoTaco:
i added the 60hz video mode flag to the sram structure as well. this should allow access to the flag if needed (which i might need later for settings & video mode with homebrew).

im basically done with this, but im not sure about the splitting up of files (changing the filesize and making existing sram.raw's invalid)...

PR #8827, opened 2020-05-24
FramebufferManager: Fix invalid readback of EFB D24S8 depth

stenzek:
Fixes assertion tripping when saving state on Adreno.

Should fix https://bugs.dolphin-emu.org/issues/12040 but I don't have easy access to an Adreno device to check.


svenkonings:
I've tested this on my device (Snapdragon 835) and this fixes the issue.

Miksel12:
This PR introduced some issues(https://bugs.dolphin-emu.org/issues/12265), removing `!force_r32f` fixes it but I guess that breaks readback of EFB D24S8. So I think the logic behind this is incorrect.

PR #8829, opened 2020-05-24
Core: Use window instead of surface for ControllerInterface updates

stenzek:
Fixes crashes with Vulkan on macOS. Regression from #8795.

Fixes https://bugs.dolphin-emu.org/issues/12107

I kinda wish we had a macOS maintainer or someone interested in doing it, it's not straightforward for me to set up/debug... :(


pizuz:
Working again. Thanks. Richard Bannister would come to my mind, but I guess he's pretty busy with all the other emulators he's maintaining.

tgcordell:
@stenzek Just a note of apprciation from those that do use MacOS and Dolphin. Thank you!

PR #8380, opened 2019-10-03
WIP - OpenXR

jordan-woyak:
Windows build uses OpenXR headers and a precompiled official loader lib/dll in Externals.
CMake build system enables OpenXR if a shared library is found.

OpenXR is enabled by selecting the relevant stereo mode in graphics settings.

Head tracking is working.
A matrix produced from each eye's view is pushed into the geometry shader where the game's projection is undone/replaced. (The old depth/convergence settings are then disabled)
A per-game setting will need to be added to specify the size of a meter.

Output appears to be working in D3D11.
Windows Mixed Reality simulator's OpenXR only supports D3D11, currently.
I've implemented D3D12 mirroring D3D11, but it is untested.
OpenGL/Vulkan need some work.

As far as I can tell unmatched game/HMD framerates should "just work".
Frames are pushed for the next "predicted display time" as they are produced.
The runtime should drop/reuse frames as needed.
There is currently zero attempt to sync up with the HMD. I don't know if this will cause avoidable studdering. I lack any hardware to test.

Things to save for another PR:
User input.
Optionally give ortho projections (and ImGui) some depth.
Per-game settings to specify the size of a meter.
"hiding" of game objects where VR causes issues.

PR #8832, opened 2020-05-25
DolphinQt / VideoCommon / Core - Add support for "split screen" using freelook

iwubcode:
Imagine you're hanging out at a friend's house and they pull up a single player game.  Wouldn't it be nice to do more than watch them?  Well now you can by using this PR (and in conjunction with this one for easier controls - #8747 ).

Introducing split screen freelook!

* Interact with your friend's game, while they are playing
* Get an advantage by looking from a different perspective while still playing the game normally
* More interesting reasons in the future :)

You can play in horizontal split screen:

image

or vertical split screen:

image

Here's the _current_ UI:

image

I felt like valuing screen real-estate made more sense (despite the weird aspect ratio) but open to suggestions.

-----

General testing and fixing the side-by-side / top-bottom 3d configurations are what is keeping this in draft.


Source/Core/Common/Matrix.cpp
```diff
@@ -234,6 +234,76 @@ void Matrix44::Multiply(const Matrix44& a, const Matrix44& b, Matrix44* result)
   result->data = MatrixMultiply<4, 4, 4>(a.data, b.data);
 }
 
+// TODO: Perhaps this should return optional<>?
```
lioncash, line 237: This wasn't mentioned in the PR description, but this should also be resolved before this is merged.

Source/Core/Common/Matrix.cpp
```diff
@@ -234,6 +234,76 @@ void Matrix44::Multiply(const Matrix44& a, const Matrix44& b, Matrix44* result)
...
+Matrix44 Matrix44::Inverted() const
+{
+  std::array<float, 16> inv;
+
+  auto& m = data;
```
lioncash, line 242:
```suggestion
  const auto& m = data;
```

Bearborg:
Rotating the camera seems to cause [extreme distortion of the geometry](https://i.imgur.com/XcBlD1A.png). Also, when using this in conjunction with stereoscopic rendering, the stereoscopic offset is applied incorrectly- both of the non-freelook screens render one eye's perspective, while both of the freelook screens render the other.

iwubcode:
Hi @Bearborg , thanks for looking.  This does not work with passive or anaglyph stereo views (since they use shaders as well).

I did see that top/bottom and side-by-side seemed to output correctly but I don't have any hardware to test with.  I will take a look.

As for the distorted geometry, I was seeing that too at one point during development.  However, I fixed that.  I will try to reproduce with a clean copy.

Source/Core/Common/Matrix.cpp
iwubcode: Thanks for taking a look lioncash!  This was copied from the OpenXR PR.  Looks like when that review was created there wasn't an `Inverted` function at all.  Now, in master there is a 3x3 `Inverted` function which just returns the value.  I'll go ahead and remove the comment since this matches that implementation.

iwubcode:
@Bearborg - please try again at your earliest convenience.  I'm sill working on the 3d but believe the distortion should be fixed.

Bearborg:
I'm no longer seeing any further distortion when I rotate the camera, but geometry seems to be [weird in general](https://imgur.com/a/gi9iguM) now. It seems like it might be depth-related?

iwubcode:
Thanks for the additional testing @Bearborg .  I will look into the fix and then go through more rigorous testing to ensure the solution works.  I only ran a couple of games last night.  I seem to have messed up something after I took the screenshots in the opening.

iwubcode:
So I went back to my old implementation @Bearborg .  But that once again flips the horizontal movement and zoom.  @stenzek suggested modifying the vertex shader to not apply the projection matrix in that case.  I guess that's what I'll try next.

Still not sure what is wrong with SBS/TAB stereo 3d logic.  Or rather, I feel like I fixed the problem where each eye wasn't rendering properly but the geometry shader isn't executing in the way I expect.  The "instancing" isn't happening and this if statement I have `eye = (layer_view < 2) ? 0 : 1` seems to always take the `true` path.  `layer_view` is essentially the `InstanceID` for geometry shader instancing.  I'm setting it to 4, so I'm not sure what is going wrong.  Anyone that might have an idea, please let me know!

Bearborg:
Geometry now looks correct and rotates correctly. However, I've noticed that geometry behind the camera plane is [completely absent](https://imgur.com/a/4iNjNts), even when that same geometry is normally visible in freelook.

iwubcode:
The issues @Bearborg reported should be fixed to a degree (thank you for the testing Bearborg!) as well as the movement being backward.  @stenzek 's change was the solution to not use the vertex shader but also ended up moving _a lot_ of stuff to the geometry shader and increased complexity.

There is a lot more work that needs to be done.  Vulkan and likely OpenGL have graphical errors on many games.  And the following issues are known:

* Per-pixel lighting fails because it needs the projected position in the vertex shader; it needs to be moved to the geometry shader
* "Emboss map" texture gen fails because it needs the projected position in the vertex shader; it needs to be moved to the geometry shader
* "Vertex rounding" needs to be moved to the geometry shader

I've also noticed some game errors even on D3D (the "best" backend):

* Last Story - for some reason, the freelook does no take affect at all
* Mario Kart Wii - some odd transparent images of the game that randomly gets put into the game world
* Spiderman Shattered Memories - the issue that @stenzek fixed, is back in splitscreen.  Not immediately obvious what needs to be changed

Ubershaders have not been tested yet and the stereo SBS/TAB are still busted.

Lots to be done, my biggest concern is this makes shader management more difficult by requiring us to maintain two pathes with very similar code.  I'm not sure what to do to fix that.

Additionally FWIW #8380 has these same issues.

Bearborg:
Can confirm that this fixes all the issues I was seeing with camera transformations.

iwubcode:
Maybe some day I will revisit but for now, this is dead.  Closing.

PR #8833, opened 2020-05-28
Android: Use DialogFragment for AlertMessage

Ebola16:
Fixes https://bugs.dolphin-emu.org/issues/7988
Fixes https://bugs.dolphin-emu.org/issues/10770

As far as I can tell, this works (even with multiple panic alerts in quick succession) but there are some things that I'm not sure I've done properly:

- ~~Did I properly handle AlertMessages while booting the core?~~
- ~~Since `NativeLibrary.displayAlertMsg` has a lock object, pausing emulation while waiting on the lock would freeze AlertMessages after rotating the screen. As far as I can tell, pausing emulation here seems unnecessary but correct me if I'm wrong.~~

Test case was https://bugs.dolphin-emu.org/issues/11424

### DIFF
--- a/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/NativeLibrary.java
+++ b/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/NativeLibrary.java
@@ -8,10 +8,10 @@ package org.dolphinemu.dolphinemu;
 
 import android.util.DisplayMetrics;
 import android.view.Surface;
-
-import androidx.appcompat.app.AlertDialog;
+import android.widget.Toast;
 
 import org.dolphinemu.dolphinemu.activities.EmulationActivity;
+import org.dolphinemu.dolphinemu.dialogs.AlertMessage;
 import org.dolphinemu.dolphinemu.utils.CompressCallback;
 import org.dolphinemu.dolphinemu.utils.Log;
 import org.dolphinemu.dolphinemu.utils.Rumble;
@@ -25,6 +25,9 @@ import java.util.LinkedHashMap;
  */
 public final class NativeLibrary
 {
+  private static final Object sAlertMessageLock = new Object();
+  private static boolean sIsShowingAlertMessage = false;
+
   private static WeakReference<EmulationActivity> sEmulationActivity = new WeakReference<>(null);
 
   /**
@@ -393,6 +396,8 @@ public final class NativeLibrary
    */
   public static native void StopEmulation();
 
+  public static native boolean IsBooting();
+
   public static native void WaitUntilDoneBooting();
 
   /**
@@ -440,8 +445,6 @@ public final class NativeLibrary
 
   public static native void SetObscuredPixelsTop(int height);
 
-  private static boolean alertResult = false;
-
   public static boolean displayAlertMsg(final String caption, final String text,
           final boolean yesNo)
   {
@@ -454,72 +457,55 @@ public final class NativeLibrary
     }
     else
     {
-      // Create object used for waiting.
-      final Object lock = new Object();
-      AlertDialog.Builder builder = new AlertDialog.Builder(emulationActivity,
-              R.style.DolphinDialogBase)
-              .setTitle(caption)
-              .setMessage(text);
-
-      // If not yes/no dialog just have one button that dismisses modal,
-      // otherwise have a yes and no button that sets alertResult accordingly.
-      if (!yesNo)
+      // AlertMessages while the core is booting will deadlock when WaitUntilDoneBooting is called.
+      // Report the AlertMessage text as a toast instead.
+      if (IsBooting())
       {
-        builder
-                .setCancelable(false)
-                .setPositiveButton("OK", (dialog, whichButton) ->
-                {
-                  dialog.dismiss();
-                  synchronized (lock)
-                  {
-                    lock.notify();
-                  }
-                });
+        emulationActivity.runOnUiThread(
+                () -> Toast.makeText(emulationActivity.getApplicationContext(), text,
+                        Toast.LENGTH_LONG)
+                        .show());
       }
       else
       {
-        alertResult = false;
+        sIsShowingAlertMessage = true;
 
-        builder
-                .setPositiveButton("Yes", (dialog, whichButton) ->
-                {
-                  alertResult = true;
-                  dialog.dismiss();
-                  synchronized (lock)
-                  {
-                    lock.notify();
-                  }
-                })
-                .setNegativeButton("No", (dialog, whichButton) ->
-                {
-                  alertResult = false;
-                  dialog.dismiss();
-                  synchronized (lock)
-                  {
-                    lock.notify();
-                  }
-                });
-      }
+        emulationActivity.runOnUiThread(() -> AlertMessage.newInstance(caption, text, yesNo)
+                .show(emulationActivity.getSupportFragmentManager(), "AlertMessage"));
 
-      // Show the AlertDialog on the main thread.
-      emulationActivity.runOnUiThread(builder::show);
-
-      // Wait for the lock to notify that it is complete.
-      synchronized (lock)
-      {
-        try
+        // Wait for the lock to notify that it is complete.
+        synchronized (sAlertMessageLock)
         {
-          lock.wait();
+          try
+          {
+            sAlertMessageLock.wait();
+          }
+          catch (Exception ignored)
+          {
+          }
         }
-        catch (Exception ignored)
+
+        if (yesNo)
         {
+          result = AlertMessage.getAlertResult();
         }
       }
-
-      if (yesNo)
-        result = alertResult;
     }
+    sIsShowingAlertMessage = false;
     return result;
+  }
+
+  public static boolean IsShowingAlertMessage()
+  {
+    return sIsShowingAlertMessage;
+  }
+
+  public static void NotifyAlertMessageLock()
+  {
+    synchronized (sAlertMessageLock)
+    {
+      sAlertMessageLock.notify();
+    }
   }
 
   public static void setEmulationActivity(EmulationActivity emulationActivity)
--- a/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/NativeLibrary.java
+++ b/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/NativeLibrary.java
@@ -446,7 +446,7 @@ public final class NativeLibrary
   public static native void SetObscuredPixelsTop(int height);
 
   public static boolean displayAlertMsg(final String caption, final String text,
-          final boolean yesNo)
+          final boolean yesNo, final boolean isWarning)
   {
     Log.error("[NativeLibrary] Alert: " + text);
     final EmulationActivity emulationActivity = sEmulationActivity.get();
@@ -454,6 +454,10 @@ public final class NativeLibrary
     if (emulationActivity == null)
     {
       Log.warning("[NativeLibrary] EmulationActivity is null, can't do panic alert.");
+    }
+    else if (emulationActivity.isIgnoringWarnings() && isWarning)
+    {
+      return true;
     }
     else
     {
@@ -470,8 +474,9 @@ public final class NativeLibrary
       {
         sIsShowingAlertMessage = true;
 
-        emulationActivity.runOnUiThread(() -> AlertMessage.newInstance(caption, text, yesNo)
-                .show(emulationActivity.getSupportFragmentManager(), "AlertMessage"));
+        emulationActivity.runOnUiThread(
+                () -> AlertMessage.newInstance(caption, text, yesNo, isWarning)
+                        .show(emulationActivity.getSupportFragmentManager(), "AlertMessage"));
 
         // Wait for the lock to notify that it is complete.
         synchronized (sAlertMessageLock)
--- a/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/dialogs/AlertMessage.java
+++ b/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/dialogs/AlertMessage.java
@@ -17,8 +17,10 @@ public final class AlertMessage extends DialogFragment
   private static final String ARG_TITLE = "title";
   private static final String ARG_MESSAGE = "message";
   private static final String ARG_YES_NO = "yesNo";
+  private static final String ARG_IS_WARNING = "isWarning";
 
-  public static AlertMessage newInstance(String title, String message, boolean yesNo)
+  public static AlertMessage newInstance(String title, String message, boolean yesNo,
+          boolean isWarning)
   {
     AlertMessage fragment = new AlertMessage();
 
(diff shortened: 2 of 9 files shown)


JosJuice:
> Since `NativeLibrary.displayAlertMsg` has a lock object, pausing emulation while waiting on the lock would freeze AlertMessages after rotating the screen. As far as I can tell, pausing emulation here seems unnecessary but correct me if I'm wrong.

Yes, not pausing emulation is probably the best workaround. This is the same issue as with PR #8819.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/dialogs/AlertMessage.java
```diff
@@ -0,0 +1,83 @@
...
+      setCancelable(false);
+      builder.setPositiveButton(android.R.string.ok, (dialog, which) ->
+      {
+        dialog.dismiss();
+        NativeLibrary.NotifyLock();
+      });
+    }
+    else
+    {
+      sAlertResult = false;
```
JosJuice, line 61: Is there a reason for this line?

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/NativeLibrary.java
```diff
@@ -463,74 +466,52 @@ public static boolean displayAlertMsg(final String caption, final String text,
...
+  {
+    return sAlertMessageText;
+  }
+
+  public static boolean IsAlertMessage()
+  {
+    return sIsAlertMessage;
+  }
+
+  public static void NotifyLock()
```
JosJuice, line 507: Maybe it would make sense to specify what this lock is for, like `NotifyAlertMessageLock`. (Same with the actual variable name.)

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/dialogs/AlertMessage.java
Ebola16: I think if a panic alert sets `sAlertResult` to true, a later panic alert will still return sAlertResult = true if a user backs out of the `yesNo` prompt.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/fragments/EmulationFragment.java
```diff
@@ -389,7 +392,22 @@ public synchronized void clearSurface()
           NativeLibrary.WaitUntilDoneBooting();
         }
 
-        NativeLibrary.SurfaceDestroyed();
+        // Handle AlertMessages while booting the core.
+        if (NativeLibrary.IsBooting())
+        {
+          if (NativeLibrary.GetAlertMessageText() != null)
+          {
+            throw new IllegalStateException(NativeLibrary.GetAlertMessageText());
```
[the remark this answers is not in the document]
Ebola16: Yes, I'm not sure the best way to handle this. The choices I thought of were deadlocking for an unknown reason or crashing for a known reason. I also assume panic alerts during core booting indicate major emulation problems anyway. This does fix `AlertMessages` outside of core booting so I think it's a worthwhile change that isn't worse than what we currently have. I'm open to other suggestions though.

Ebola16:
@dolphin-emu-bot rebuild

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/dialogs/AlertMessage.java
JosJuice: But doesn't every code path that calls `NotifyAlertMessageLock` set `sAlertResult`? So either there is some way to dismiss the alert without calling `NotifyAlertMessageLock`, or there shouldn't be any need for this line.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/fragments/EmulationFragment.java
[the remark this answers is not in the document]
JosJuice: I guess the other option would be to dismiss the dialog if you try to rotate the screen while showing an alert message while booting. While it does avoid crashing, it is a bit odd, so I'm not sure if it's actually better...

I guess either a crash or what I mentioned would be okay, but you should leave a comment about it in either case (either about this crash being intentional, or about why we dismiss the dialog).

Source/Core/Core/Core.cpp
```diff
@@ -170,6 +170,11 @@ void DisplayMessage(std::string message, int time_in_ms)
   OSD::AddMessage(std::move(message), time_in_ms);
 }
 
+bool IsBooting()
+{
+  return (s_is_booting.IsSet() || !s_hardware_initialized);
```
JosJuice, line 175:
```suggestion
  return s_is_booting.IsSet() || !s_hardware_initialized;
```

Also, `WaitUntilDoneBooting` should be made to call this function.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/fragments/EmulationFragment.java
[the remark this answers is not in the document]
JosJuice: Also, if you're going to have it crash, it probably makes more sense to always include "Attempted to destroy surface while booting the core" in the exception message (perhaps with the alert message text appended), since the error that causes the alert message to be shown is not in itself the reason why we are crashing.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/dialogs/AlertMessage.java
Ebola16: It turns out directly copying the old code into the DialogFragment didn't handle the case of backing out of a `yesNo` prompt, which never notified the lock. I removed the extra `sAlertResult` and made `yesNo` prompts not cancelable too. This makes sense since users can rapidly tap and accidentally dismiss the prompt.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/fragments/EmulationFragment.java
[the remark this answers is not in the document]
Ebola16: I tried another idea. Instead of deadlocking or crashing the emulator, we can simply end emulation. The AlertMessage has been replaced with a log error and toast. Hopefully this won't cause unforeseen problems.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/dialogs/AlertMessage.java
JosJuice: Sounds good to me.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/fragments/EmulationFragment.java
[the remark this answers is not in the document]
Ebola16: I also moved the `R.string.emulation_touch_button_help` toast to only run once during emulation. This will prevent it from reappearing when other toasts are displayed.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/fragments/EmulationFragment.java
```diff
@@ -389,7 +392,32 @@ public synchronized void clearSurface()
...
+            Toast.makeText(context, alertMessageError, Toast.LENGTH_LONG).show();
+          }
+          else
+          {
+            String generalError = "Attempted to destroy surface while booting the core.";
+
+            Log.error("[EmulationFragment] " + generalError);
+            Toast.makeText(context, generalError, Toast.LENGTH_LONG).show();
+          }
+          stop();
```
JosJuice, line 414: Are you sure that calling `StopEmulation` while booting works correctly? This seems very fishy to me.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/fragments/EmulationFragment.java
```diff
@@ -389,7 +392,32 @@ public synchronized void clearSurface()
           NativeLibrary.WaitUntilDoneBooting();
         }
 
-        NativeLibrary.SurfaceDestroyed();
+        // AlertMessages while booting the core will dereference nullptr. Silently dismissing
```
JosJuice, line 395: I don't understand how that would lead to a nullptr dereference. Could you clarify the comment with e.g. where we get the nullptr dereference?

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/fragments/EmulationFragment.java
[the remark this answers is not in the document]
JosJuice: Ending emulation is also kinda weird, but I guess it's no worse than crashing.

What about just turning the alert message into a toast and then calling `WaitUntilDoneBooting` as usual, without stopping emulation or anything?

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/fragments/EmulationFragment.java
JosJuice: If you're referring to the same nullptr dereference as in the comment above, you should write that we can't call `WaitUntilDoneBooting` if we get an alert message while booting rather than just writing that we will dereference nullptr.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/fragments/EmulationFragment.java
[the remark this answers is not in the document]
Ebola16: If `WaitUntilDoneBooting` is called instead of stopping emulation, we'll get the deadlock. I think that is the least appealing option of deadlock vs. crash vs. ending emulation. Still debating if I should go back to crashing the emulatior as I'm also not certain of the safety of stopping emulation either.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/fragments/EmulationFragment.java
Ebola16: Yes, I was referring to the same nullptr dereference.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/fragments/EmulationFragment.java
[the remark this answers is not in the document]
JosJuice: > If `WaitUntilDoneBooting` is called instead of stopping emulation, we'll get the deadlock.

Yes, unless we release the lock first. My suggestion is that we dismiss the dialog and release the lock, and then show a toast.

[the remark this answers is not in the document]
JosJuice: I just realized that regardless of what we do about this case, we're still going to get a deadlock in the case where we get an alert message after `WaitUntilDoneBooting` already has been called... I wonder if we maybe just should show all alert messages that are triggered when booting as toasts without even attempting to use a proper dialog.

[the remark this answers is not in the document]
Ebola16: I could use the new `NativeLibrary.IsBooting` check to handle alert messages differently during core boot. Just to clarify, this PR doesn't deadlock with alert messages after core boot because it skips calling `WaitUntilDoneBooting` for `AlertMessages`.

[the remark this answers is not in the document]
JosJuice: Yes, there is no problem after we're done booting.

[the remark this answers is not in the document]
Ebola16: Moving the implementation of `IsBooting` and making alert messages toasts while the core is booting seems to be a good solution here. Hopefully this version does the trick!

Ebola16:
Thanks for the help! This PR went in directions I didn't expect.

Ebola16:
Minor typo fix. I accidentally called the "AlertMessage" a "Panic Alert" in `NativeLibrary`.

Ebola16:
Can this be merged?

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/dialogs/AlertMessage.java
```diff
@@ -71,6 +75,17 @@ public Dialog onCreateDialog(Bundle savedInstanceState)
                 NativeLibrary.NotifyAlertMessageLock();
               });
     }
+
+    if (isWarning)
+    {
+      builder.setNeutralButton("IGNORE FOR THIS SESSION", (dialog, which) ->
```
JosJuice, line 81: This string should be in a resources file and should not be in allcaps.

Source/Android/jni/MainAndroid.cpp
```diff
@@ -146,14 +146,14 @@ void Host_TitleChanged()
...
-static bool MsgAlert(const char* caption, const char* text, bool yes_no, Common::MsgType /*style*/)
+static bool MsgAlert(const char* caption, const char* text, bool yes_no, Common::MsgType style)
 {
   JNIEnv* env = IDCache::GetEnvForThread();
 
   // Execute the Java method.
   jboolean result = env->CallStaticBooleanMethod(
       IDCache::GetNativeLibraryClass(), IDCache::GetDisplayAlertMsg(), ToJString(env, caption),
-      ToJString(env, text), yes_no ? JNI_TRUE : JNI_FALSE);
+      ToJString(env, text), yes_no ? JNI_TRUE : JNI_FALSE, style == Common::MsgType::Warning);
```
JosJuice, line 156: I don't know if having the style affect the behavior of the panic alert is a good idea or not. It would be better if this commit could be split into another PR so that the merge of the other fixes in this PR aren't blocked by this question being open.

JosJuice: Actually nevermind lol we already do this in Qt. So I guess it's fine

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/NativeLibrary.java
```diff
@@ -478,76 +481,64 @@ public static boolean displayAlertMsg(final String caption, final String text,
     {
       Log.warning("[NativeLibrary] EmulationActivity is null, can't do panic alert.");
     }
+    else if (emulationActivity.isIgnoringWarnings() && isWarning)
+    {
+      return false;
```
JosJuice, line 486: To be consistent with DolphinQt, this should return true.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/dialogs/AlertMessage.java
Ebola16: The string is converted to all caps when displayed but I now use a string resource that matches Qt.

JosJuice:
Well, it seems like nobody else is going to review this...

PR #8834, opened 2020-05-30
AudioStretcher: Tweak algorithm to be more forgiving

MerryMage:
Hopefully helps with https://bugs.dolphin-emu.org/issues/10254

* Allow for a more slop in backlog_fullness
* Snap to m_EmulationSpeed

With thanks to @Filoppi.

@Filoppi – I would be grateful if you could verify this PR solves the issue.


Filoppi:
Hey.
This is a good start but I don't think is enough. it will still lower quality than the non stretched sound.
I'm still working on my attempt daily and I got to a really good point, but I haven't reached a "perfect" solution that works in all the cases.
Also, this "if (std::abs(m_stretch_ratio - SConfig::GetInstance().m_EmulationSpeed) < 0.01)", from my test is definitely not enough, as it can easily fluctuate more than 0.01, though anything more than 0.01 is kind of too much and could introduce errors over time.
I guess this could be merged, though I would wait a few days for my larger, "better", fix.

MerryMage:
Noted!

PR #8835, opened 2020-05-31
NetPlay: Always use packet to fire power button event

Techjar:
This is a first pass on trying to fix the NetPlay Wii shutdown deadlock. Not entirely sure if it will fix the problem, more timing synchronization might be needed, but it's at least a necessary step towards fixing the issue.


JMC47:
Is there any further investigation being done on this issue?  If there isn't, maybe we should close this for now.  This didn't *entirely* fix the issue iirc.

Techjar:
I'm leaving it open for now as I do intend to come up with a proper fix at some point.

PR #8836, opened 2020-05-31
Android: Add recursive game paths to UI

Ebola16:
Implements https://bugs.dolphin-emu.org/issues/10806

### DIFF
--- a/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/ui/SettingsActivityPresenter.java
+++ b/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/ui/SettingsActivityPresenter.java
@@ -1,5 +1,6 @@
 package org.dolphinemu.dolphinemu.features.settings.ui;
 
+import android.content.Context;
 import android.content.IntentFilter;
 import android.os.Bundle;
 import android.text.TextUtils;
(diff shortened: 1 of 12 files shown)


JosJuice:
Your changes to `scanLibrary` are more complicated than needed. What you should do instead is make it so that you can pass a boolean parameter to `update` that's used instead of `false` on this line: https://github.com/dolphin-emu/dolphin/blob/master/Source/Android/jni/GameList/GameFileCache.cpp#L97

JosJuice:
Assuming there is no simple way for us to automatically refresh the library when changing the setting, LGTM after fixing the lint issue.

Ebola16:
Lint issue fixed. We could refresh the game library upon saving settings but I'd rather do that only if this setting is modified. I don't think there is a simple way to detect specific changed settings yet.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/ui/SettingsActivityPresenter.java
```diff
@@ -27,18 +31,22 @@
 
   private MenuTag menuTag;
   private String gameId;
+  private Context context;
+
+  private static final Set<String> modifiedSettings = new HashSet<>();
```
JosJuice, line 36: Would it be possible to make this non-static? You could pass it to `saveSettings` as an argument.

Ebola16: Yes, done.

PR #8839, opened 2020-06-01
INI changes

Miksel12:
Based on #8164 

-Adds a few games.
-Enables MMU for Crunchyroll Channel so it can boot and for Interactive Multi Game Demo Disc with demo's which require MMU.
-Enables FPRF for games that need it.
-Disables defered EFB for games that crashed or showed weird behaviour with defered EFB enabled.
-Sets TextureCacheAccuracy to safe for a few games that need it.
-Sets EFB to texture and ram for a few games that need it.
-Removes Wiimote.Shake and Wiimote.Swing as those are no longer functional.

I also removed TextureCacheAccuracy = Safe from NSMBW because accuracy on fast shows no issues. I think this was fixed with Partial Texture Updates.

### DIFF
--- a/Data/Sys/GameSettings/HC4.ini
+++ b/Data/Sys/GameSettings/HC4.ini
@@ -1,7 +1,8 @@
-# RDFE41, RDFP41 - Shaun White Snowboarding: Road Trip
+# HC4E9Z, HC4P9Z - Crunchyroll Channel
 
 [Core]
 # Values set here will override the main Dolphin settings.
+MMU = True
 
 [OnLoad]
 # Add memory patches to be loaded once on boot here.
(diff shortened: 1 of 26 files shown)


Data/Sys/GameSettings/HC4.ini
```diff
@@ -0,0 +1,14 @@
+# HC4P01, HC4E9Z, HC4P9Z - Crunchyroll Channel
```
Tilka, line 1: Where does "HC4P01" come from?

JosJuice: Probably from outdated wiki information. (Sometimes people seemingly guessed the publisher ID of downloadable titles because GameTDB doesn't provide that information.)

Miksel12: Weird, I have taken the GameID's from the wiki but I can't find "HC4P01" anywhere now. No idea how that happend. I will remove it.

PR #8840, opened 2020-06-01
ControllerInterface: Combine evdev devices with the same physical location in addition to unique ID

Techjar:
Some off-brand devices always report the same unique ID, resulting in them being wrongly combined. The physical location for these devices should always be different, making that a better comparison. I'm currently unsure whether the DS4's devices (the original target of the combining behavior in #8473) all have the same physical location, but theoretically they should.

### DIFF
--- a/Source/Core/InputCommon/ControllerInterface/evdev/evdev.cpp
+++ b/Source/Core/InputCommon/ControllerInterface/evdev/evdev.cpp
@@ -204,9 +204,10 @@ static int s_wakeup_eventfd;
 // sysfs is not stable, so this is probably the easiest way to get a name for a node.
 static std::map<std::string, std::weak_ptr<evdevDevice>> s_devnode_objects;
 
-static std::shared_ptr<evdevDevice> FindDeviceWithUniqueID(const char* unique_id)
+static std::shared_ptr<evdevDevice>
+FindDeviceWithUniqueIDAndPhysicalLocation(const char* unique_id, const char* physical_location)
 {
-  if (!unique_id)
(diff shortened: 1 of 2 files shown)


sclem:
Pulled and tested, confirmed this fixes https://bugs.dolphin-emu.org/issues/12126

Techjar:
Great, now I just need someone to test the DS4 to make sure that still works as intended.

sclem:
Sounds good. I would test that if I had one.

jordan-woyak:
Unfortunately this breaks my Mayflash 4-port GameCube controller adapter.
All four ports have the same "physical location" and are incorrectly combined.

Perhaps you can examine BOTH the physical location and the unique ID?

Techjar:
@jordan-woyak That does appear to stop it combing the Mayflash adapter on my system. Can you confirm?

jordan-woyak (approved):
LGTM

i30817:
Could this be done manually at the config/GUI level?

My usecase was attempting to use keyboard and touchpad as keys and pointer|motion-simulation to a emulated wiimote.

Apparently only 'XInput2/0/Virtual core pointer' allowed this naturally, all the others it's either one or the other for motion simulation. And i can't use X libraries in EGL/DRM.

Truth to be said, even this wouldn't solve the problem because the evdev/0/SynPS/2 Synaptics TouchPad entry does not behave like the mouse. The 0 origin starts in the left side instead of the center, etc, so i'm not sure even this would work.
I'm probably looking for 'Xinput but works in the nogui DRM' sigh.

PR #8841, opened 2020-06-01
Replace usage of std::result_of with std::invoke_result

Techjar:
`std::result_of` is deprecated in C++17, and removed in C++20. Microsoft has gone ahead with the removal as of Visual Studio 16.6.0, so before this change our code is broken there.

Externals/Qt also needs to be updated to the latest version to fully fix the build.

### DIFF
--- a/Source/Core/DolphinQt/QtUtils/RunOnObject.h
+++ b/Source/Core/DolphinQt/QtUtils/RunOnObject.h
@@ -26,7 +26,7 @@ class QObject;
 template <typename F>
 auto RunOnObject(QObject* object, F&& functor)
 {
-  using OptionalResultT = std::optional<std::result_of_t<F()>>;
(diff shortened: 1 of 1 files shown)


Filoppi:
You might want to update this issue if you fix all the cases:
https://bugs.dolphin-emu.org/issues/12086

Techjar:
Yeah we're just waiting on the Qt update.

chrisis123:
Just a heads up, if you update externals to QT 5.4.2 (currently it's still on 5.4.1) it should build fine again - I managed to build it with cmake on VS 2019 16.6.1 while using building against my local copy of QT 5.4.2 - likely the standard VS Project will work fine too if using 5.4.2 (didnt't try it though)

Techjar:
Yeah that's exactly what we're going to do. Changing your local VS project to C++17 also works as an easier temporary solution.

PR #8843, opened 2020-06-02
DolphinQt: Add fastmem option to debugger UI

Techjar:
I'm adding this as a convenience for people like me who always forget to turn off fastmem before debugging a segfault. Less effort required than having to edit the config file manually and restart Dolphin, which might be extra inconvenient if you've gone through a bunch of steps to get to the state needed to debug an issue.

### DIFF
--- a/Source/Core/DolphinQt/MenuBar.cpp
+++ b/Source/Core/DolphinQt/MenuBar.cpp
@@ -136,6 +136,7 @@ void MenuBar::OnEmulationStateChanged(Core::State state)
   m_jit_interpreter_core->setEnabled(running);
   m_jit_block_linking->setEnabled(!running);
   m_jit_disable_cache->setEnabled(!running);
+  m_jit_disable_fastmem->setEnabled(!running);
   m_jit_clear_cache->setEnabled(running);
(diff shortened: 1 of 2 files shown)


Techjar:
Hmm, true, I guess it could just be in the JIT menu. For the record, CMake remembers the flags you set for a given build cache. The assumption about translation was made under the premise that only a literal handful of users will actually see the string, since it would only be in special builds with the flag set, making it not worth the translators' time.

lioncash:
> For the record, CMake remembers the flags you set for a given build cache. 

Sure, and integrating it into the debugger UI gets rid of needing to remember to pass a flag entirely. It's just there in the UI ready to be used whenever the debugger is active.

> The assumption about translation was made under the premise that only a literal handful of users will actually see the string ...

We already translate the debugger UI which, by itself, isn't intended to be used by the common user. Not translating the option would be pretty out of sync with everything else.

Techjar:
Alright, I've now made it an option under the JIT menu in the debugger. This way it can be accessed in normal builds rather than requiring a compiler flag, but it's still hidden away enough that the average user won't touch it.

PR #8844, opened 2020-06-02
Android: Refresh game library when resuming (Tv)MainActivity

Ebola16:
Implements https://bugs.dolphin-emu.org/issues/10716


JosJuice:
LGTM other than one nitpick: I think it would be slightly cleaner to call `startRescan` after calling `addDirIfNeeded` instead of calling it inside `addDirIfNeeded`.

PR #8845, opened 2020-06-04
Android: Display disc number instead of company if mulipart disc

Ebola16:
Fixes https://bugs.dolphin-emu.org/issues/11413.

The company can be found in Game Details too so I'd rather use that space to display disc number for multipart discs. See Tales of Symphonia's cards in the image below.

Screenshot_20200604-013735_Dolphin Emulator

### DIFF
--- a/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/adapters/GameAdapter.java
+++ b/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/adapters/GameAdapter.java
@@ -1,5 +1,6 @@
 package org.dolphinemu.dolphinemu.adapters;
 
+import android.content.Context;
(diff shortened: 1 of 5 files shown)


Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/services/GameFileCacheService.java
```diff
@@ -86,6 +86,11 @@ public static GameFile findSecondDisc(GameFile game)
     return matchWithoutRevision;
   }
 
+  public static int GetDiscNumber(GameFile game)
```
JosJuice, line 89: There is no reason for this method to exist. Calls to it can be replaced with `game.getDiscNumber()`.

PR #8846, opened 2020-06-05
Fix Windows CMake builds with shared zlib

nyanpasu64:
On Linux, if shared zlib is present, zlib.h is always available and -lz
links to zlib, even if you don't run find_package(ZLIB).

For some reason I have zlib installed on Windows (possibly from vcpkg),
so find_package(ZLIB) succeeds and ZLIB_FOUND is true.
When Dolphin uses shared zlib on Windows, the problem is that zlib.h
is not in the default include path, and the CMake target is called
ZLIB::ZLIB and there's neither a target nor a library called z.

However, both find_package(ZLIB) and add_subdirectory(Externals/zlib)
create a target called ZLIB::ZLIB, so I'll switch to that instead.
Hopefully this change doesn't break anyone's build.

### DIFF
--- a/Externals/libpng/CMakeLists.txt
+++ b/Externals/libpng/CMakeLists.txt
@@ -23,7 +23,7 @@ PUBLIC
   ${CMAKE_CURRENT_SOURCE_DIR}
 )
 
-target_link_libraries(png PUBLIC z)
+target_link_libraries(png PUBLIC ZLIB::ZLIB)
 
 if(NOT MSVC)
   target_compile_options(png
--- a/Externals/minizip/CMakeLists.txt
+++ b/Externals/minizip/CMakeLists.txt
@@ -13,6 +13,6 @@ endif()
 add_library(minizip STATIC ${SRCS})
 target_include_directories(minizip PUBLIC .)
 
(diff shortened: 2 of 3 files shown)


JosJuice:
@dolphin-emu-bot rebuild

CMakeLists.txt
```diff
@@ -620,6 +620,7 @@ endif()
 find_package(ZLIB)
 if(ZLIB_FOUND)
   message(STATUS "Using shared zlib")
+  link_libraries(ZLIB::ZLIB)
```
Tilka, line 623: I would have expected that there are more target_link_libraries lines that would need to be changed rather adding this line here.

nyanpasu64: Oops... I'll remove this and test again at some point, but I'm on Linux at the moment.

nyanpasu64:
Just rebased to master and removed the unnecessary global ZLIB::ZLIB. Building Dolphin works on my branch and fail on master.

I haven't checked if there are any other libraries that can break this way.

JosJuice:
@dolphin-emu-bot rebuild

leoetlino (approved):
sgtm

PR #8847, opened 2020-06-06
Qt/GameListModel: Use absolutePath for file path column

Techjar:
canonicalPath is orders of magnitude slower as it has to perform actual disk I/O to resolve symlinks, which makes sorting by this column ridiculously slow for large game lists, especially if the games are on a NAS. We probably don't need that, simply resolving relative paths should be sufficient.

This should fix https://bugs.dolphin-emu.org/issues/12135


JMC47:
Sorting by the File Path list is faster in this PR.

PR #8849, opened 2020-06-06
Config/IsSettingSaveable: Use class template argument deduction

Techjar:
This eliminates the stupid array size that has to be manually updated and occasionally causes broken builds with PR race conditions.

libc++ currently seems to be missing several deduction guides so we cannot merge this for now.

### DIFF
--- a/Source/Core/Core/ConfigLoaders/IsSettingSaveable.cpp
+++ b/Source/Core/Core/ConfigLoaders/IsSettingSaveable.cpp
@@ -28,7 +28,7 @@ bool IsSettingSaveable(const Config::Location& config_location)
       return true;
   }
 
-  static constexpr std::array<const Config::Location*, 104> s_setting_saveable = {
+  static constexpr std::array s_setting_saveable = {
       // Main.Core
 
       &Config::MAIN_DEFAULT_ISO.location,


Techjar:
I have a workaround for the libc++ issue by adding this to the file:
```cpp
namespace std
{
// workaround for missing deduction guides in libc++
template <class... U>
array(const Config::Location*, U...)->array<const Config::Location*, 1 + sizeof...(U)>;
}  // namespace std
```
Seems kinda silly but let me know what you think.

Techjar:
Marking this as draft until a future date when all compilers handle it properly.

Source/Core/Core/ConfigLoaders/IsSettingSaveable.cpp
```diff
@@ -28,7 +28,7 @@ bool IsSettingSaveable(const Config::Location& config_location)
       return true;
   }
 
-  static constexpr std::array<const Config::Location*, 104> s_setting_saveable = {
+  static constexpr std::array s_setting_saveable = {
```
leoetlino, line 31: Alternatively, we could use auto for the variable itself (which will be deduced as a std::initializer_list). This avoids having to wait for C++20.

Techjar: Does that actually work?

leoetlino: It'll pretty much work the same as a std::array, except with a compiler-generated backing array (the symbol for the array is "reference temporary"). You'll need to use std::begin and std::end instead of the .begin() and .end() member functions, though.

jordan-woyak:
Closing in favor of #13164.

PR #8850, opened 2020-06-07
Show an OSD message when running a disc image with a large block size

JosJuice:
This is intended to catch WIA files which have been created using wit's default parameters (40 MiB block size), once PR #8538 is merged. The check does however also work for GCZ files – not that I think anyone has a GCZ file with a block size that large.

### DIFF
--- a/Source/Core/Core/HW/DVD/DVDInterface.cpp
+++ b/Source/Core/Core/HW/DVD/DVDInterface.cpp
@@ -36,6 +36,7 @@
 #include "Core/IOS/IOS.h"
 #include "Core/Movie.h"
 
+#include "DiscIO/Blob.h"
 #include "DiscIO/Enums.h"
 #include "DiscIO/Volume.h"
 #include "DiscIO/VolumeWii.h"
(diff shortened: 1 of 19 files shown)


Source/Core/Core/HW/DVD/DVDInterface.cpp
```diff
@@ -426,6 +427,17 @@ void SetDisc(std::unique_ptr<DiscIO::VolumeDisc> disc,
   bool had_disc = IsDiscInside();
   bool has_disc = static_cast<bool>(disc);
 
+  if (has_disc)
+  {
+    const DiscIO::BlobReader& blob = disc->GetBlobReader();
+    if (!blob.HasFastRandomAccessInBlock() && blob.GetBlockSize() > 0x200000)
```
iwubcode, line 433: Just curious, what is `0x200000`.  Is it possible to recommend a setting in the "convert feature" to rectify the issue?

JosJuice: > Just curious, what is `0x200000`?

2 MiB, the minimum block size of WIA. It seems to be fast enough in most cases.

> Also - is it possible to recommend a setting in the "convert feature" to rectify the issue?

The convert dialog doesn't allow you to pick any block size larger than 2 MiB.

PR #8853, opened 2020-06-08
Feature XLink Kai BBA

CrunchBite:
This PR adds support for GameCube LAN play via XLink Kai using a new BBA which communicates directly with the XLink client via UDP socket.

This relies on unreleased additions to the XLink Kai client but is functioning correctly on Windows, macOS, and Linux.

For more information see: https://bugs.dolphin-emu.org/issues/12108


Source/Core/Core/HW/EXI/BBA/XLINK_KAI_BBA.cpp
```diff
@@ -0,0 +1,281 @@
...
+
+  m_sf_recipient_ip = m_dest_ip.c_str();
+
+  // Send connect command with unique local name
+  // connect;locally_unique_name;emulator_name;optional_padding
+  uint8_t buffer[255] = {0x00};
+  std::string cmd =
+      "connect;" + m_client_identifier + ";dolphin;000000000000000000000000000000000000000000";
+
+  u32 size = (u32)cmd.length();
```
lioncash, line 46:
```suggestion
  const auto size = u32(cmd.length());
```

Source/Core/Core/HW/EXI/BBA/XLINK_KAI_BBA.cpp
```diff
@@ -0,0 +1,281 @@
...
+  {
+    ERROR_LOG(SP1, "Couldn't open XLink Kai UDP socket, unable to initialize BBA");
+    return false;
+  }
+
+  m_sf_recipient_ip = m_dest_ip.c_str();
+
+  // Send connect command with unique local name
+  // connect;locally_unique_name;emulator_name;optional_padding
+  uint8_t buffer[255] = {0x00};
```
lioncash, line 42:
```suggestion
  u8 buffer[255] = {};
```

Source/Core/Core/HW/EXI/BBA/XLINK_KAI_BBA.cpp
```diff
@@ -0,0 +1,281 @@
...
+  return RecvInit();
+}
+
+void CEXIETHERNET::XLinkNetworkInterface::Deactivate()
+{
+  // Send d; to tell XLink we want to disconnect cleanly
+  // disconnect;optional_locally_unique_name;optional_padding
+  std::string cmd =
+      "disconnect;" + m_client_identifier + ";0000000000000000000000000000000000000000000";
+  u32 size = (u32)cmd.length();
```
lioncash, line 67:
```suggestion
  const auto size = u32(cmd.length());
```

Source/Core/Core/HW/EXI/BBA/XLINK_KAI_BBA.cpp
```diff
@@ -0,0 +1,281 @@
...
+  uint8_t buffer[255] = {0x00};
```
lioncash, line 68:
```suggestion
  u8 buffer[255] = {};
```

Source/Core/Core/HW/EXI/BBA/XLINK_KAI_BBA.cpp
```diff
@@ -0,0 +1,281 @@
...
+    std::size_t bytes_read = 0;
+    if (self->m_sf_socket.receive(self->m_in_frame, 9004, bytes_read, sender, port) !=
+        sf::Socket::Done)
+    {
+      ERROR_LOG(SP1, "Failed to read from BBA, err=%lu", bytes_read);
+    }
+
+    // Did we get an ethernet frame? Copy the first 4 bytes to check the header
+    char tempCheck[4];
+    memmove(tempCheck, self->m_in_frame, 4);
```
lioncash, line 143:
```suggestion
    memmove(tempCheck, self->m_in_frame, std::size(temp_check));
```

Source/Core/Core/HW/EXI/BBA/XLINK_KAI_BBA.cpp
```diff
@@ -0,0 +1,281 @@
...
+    // *here* because XLink *could* send one
+    std::size_t bytes_read = 0;
+    if (self->m_sf_socket.receive(self->m_in_frame, 9004, bytes_read, sender, port) !=
+        sf::Socket::Done)
+    {
+      ERROR_LOG(SP1, "Failed to read from BBA, err=%lu", bytes_read);
+    }
+
+    // Did we get an ethernet frame? Copy the first 4 bytes to check the header
+    char tempCheck[4];
```
lioncash, line 142:
```suggestion
    char temp_check[4];
```

Source/Core/Core/HW/EXI/BBA/XLINK_KAI_BBA.cpp
```diff
@@ -0,0 +1,281 @@
...
+    m_eth_ref->SendComplete();
+    return true;
+  }
+}
+
+void CEXIETHERNET::XLinkNetworkInterface::ReadThreadHandler(
+    CEXIETHERNET::XLinkNetworkInterface* self)
+{
+  sf::IpAddress sender;
+  unsigned short port;
```
lioncash, line 125:
```suggestion
  u16 port;
```

Source/Core/Core/HW/EXI/BBA/XLINK_KAI_BBA.cpp
```diff
@@ -0,0 +1,281 @@
...
+          // (u32)(bytes_read - 4), 0x10).c_str());
+          self->m_eth_ref->mRecvBufferLength = (u32)(bytes_read - 4);
+          self->m_eth_ref->RecvHandlePacket();
+        }
+      }
+    }
+    // Otherwise we recieved control data or junk
+    else
+    {
+      char recv_str[9004] = {0x00};
```
lioncash, line 174:
```suggestion
      char recv_str[9004] = {};
```

Source/Core/Core/HW/EXI/BBA/XLINK_KAI_BBA.cpp
```diff
@@ -0,0 +1,281 @@
...
+      if (controlMsg.substr(0, 9) == "connected")
+      {
+        NOTICE_LOG(SP1, "XLink Kai BBA connected");
+        OSD::AddMessage("XLink Kai BBA connected", 4500);
+
+        // Send any extra settings now
+        // Enable XLink chat messages in OSD if set
+        if (self->m_chat_osd_enabled)
+        {
+          std::string cmd = "setting;chat;true";
```
lioncash, line 189:
```suggestion
          constexpr std::string_view cmd = "setting;chat;true";
```

Source/Core/Core/HW/EXI/BBA/XLINK_KAI_BBA.cpp
```diff
@@ -0,0 +1,281 @@
...
+          u32 size = (u32)cmd.length();
```
lioncash, line 190:
```suggestion
          const auto size = u32(cmd.length());
```

Source/Core/Core/HW/EXI/BBA/XLINK_KAI_BBA.cpp
```diff
@@ -0,0 +1,281 @@
...
+          uint8_t buffer[255] = {0x00};
```
lioncash, line 191:
```suggestion
          u8 buffer[255] = {};
```

Source/Core/Core/HW/EXI/EXI_DeviceEthernet.h
```diff
@@ -196,10 +198,16 @@ enum RecvStatus
 
 #define BBA_RECV_SIZE 0x800
 
+enum class BBADeviceType
+{
+  BBA_TAP,
```
lioncash, line 203:
```suggestion
  TAP,
```

Source/Core/Core/HW/EXI/EXI_DeviceEthernet.h
```diff
@@ -196,10 +198,16 @@ enum RecvStatus
...
+  BBA_XLINK,
```
lioncash, line 204:
```suggestion
  XLINK,
```

Source/Core/Core/HW/EXI/BBA/XLINK_KAI_BBA.cpp
```diff
@@ -0,0 +1,281 @@
...
+  memmove(m_out_frame, "e;e;", 4);
+  memmove(m_out_frame + 4, frame, size);
+  size += 4;
+
+  // Only uncomment for debugging, the performance hit is too big otherwise
+  // INFO_LOG(SP1, "SendFrame %x\n%s", size, ArrayToString(m_out_frame, size, 0x10).c_str());
+
+  if (m_sf_socket.send(m_out_frame, size, m_sf_recipient_ip, m_dest_port) != sf::Socket::Done)
+  {
+    ERROR_LOG(SP1, "SendFrame(): expected to write %d bytes, but failed, errno %d", size, errno);
```
lioncash, line 111:
```suggestion
    ERROR_LOG(SP1, "SendFrame(): expected to write %u bytes, but failed, errno %d", size, errno);
```

Source/Core/Core/HW/EXI/BBA/XLINK_KAI_BBA.cpp
```diff
@@ -0,0 +1,281 @@
...
+    if (!self->IsActivated())
+      break;
+
+    // XLink supports jumboframes but Gamecube BBA does not. We need to support jumbo frames
+    // *here* because XLink *could* send one
+    std::size_t bytes_read = 0;
+    if (self->m_sf_socket.receive(self->m_in_frame, 9004, bytes_read, sender, port) !=
+        sf::Socket::Done)
+    {
+      ERROR_LOG(SP1, "Failed to read from BBA, err=%lu", bytes_read);
```
lioncash, line 138:
```suggestion
      ERROR_LOG(SP1, "Failed to read from BBA, err=%zu", bytes_read);
```

Source/Core/Core/HW/EXI/BBA/XLINK_KAI_BBA.cpp
```diff
@@ -0,0 +1,281 @@
...
+      if ((bytes_read - 4) < BBA_RECV_SIZE)
+      {
+        DEBUG_LOG(SP1, "Received ethernet frame");  // TODO remove this logging for performance?
+
+        // Copy payload into BBA buffer as an ethernet frame
+        memmove(self->m_eth_ref->mRecvBuffer.get(), self->m_in_frame + 4, bytes_read - 4);
+
+        if (bytes_read < 1)
+        {
+          ERROR_LOG(SP1, "Failed to read from BBA, err=%lu", bytes_read - 4);
```
lioncash, line 159:
```suggestion
          ERROR_LOG(SP1, "Failed to read from BBA, err=%zu", bytes_read - 4);
```

Source/Core/Core/HW/EXI/BBA/XLINK_KAI_BBA.cpp
```diff
@@ -0,0 +1,281 @@
...
+        }
+        else if (self->readEnabled.IsSet())
+        {
+          // Only uncomment for debugging, the performance hit is too big otherwise
+          // DEBUG_LOG(SP1, "Read data: %s", ArrayToString(self->m_eth_ref->mRecvBuffer.get(),
+          // (u32)(bytes_read - 4), 0x10).c_str());
+          self->m_eth_ref->mRecvBufferLength = (u32)(bytes_read - 4);
```
lioncash, line 166:
```suggestion
          self->m_eth_ref->mRecvBufferLength = u32(bytes_read - 4);
```

Source/Core/Core/HW/EXI/BBA/XLINK_KAI_BBA.cpp
```diff
@@ -0,0 +1,281 @@
...
+        }
+      }
+    }
+    // Otherwise we recieved control data or junk
+    else
+    {
+      char recv_str[9004] = {0x00};
+      memmove(recv_str, self->m_in_frame, bytes_read);
+      INFO_LOG(SP1, "Received XLink Kai control data: %s", recv_str);
+      std::string controlMsg = std::string(recv_str);
```
lioncash, line 177: May want to use a std::string to begin with so you can initialize the data in place like:

```cpp
std::string recv_str(self->m_in_frame, self->m_in_frame + bytes_read);
```

instead of using both the stack and then converting the string over later.

Source/Core/Core/HW/EXI/BBA/XLINK_KAI_BBA.cpp
```diff
@@ -0,0 +1,281 @@
...
+
+  while (!self->readThreadShutdown.IsSet())
+  {
+    if (!self->IsActivated())
+      break;
+
+    // XLink supports jumboframes but Gamecube BBA does not. We need to support jumbo frames
+    // *here* because XLink *could* send one
+    std::size_t bytes_read = 0;
+    if (self->m_sf_socket.receive(self->m_in_frame, 9004, bytes_read, sender, port) !=
```
lioncash, line 135:
```suggestion
    if (self->m_sf_socket.receive(self->m_in_frame, std::size(self->m_in_frame), bytes_read, sender, port) !=
```

Source/Core/Core/HW/EXI/BBA/XLINK_KAI_BBA.cpp
```diff
@@ -0,0 +1,281 @@
...
+    memmove(tempCheck, self->m_in_frame, 4);
+
+    // Check for e;e; header; this indicates the received data is an ethernet frame
+    // e;e;raw_ethernet_frame_data
+    if (tempCheck[2] == 'e' && tempCheck[3] == ';')
+    {
+      // Is the frame larger than BBA_RECV_SIZE?
+      if ((bytes_read - 4) < BBA_RECV_SIZE)
+      {
+        DEBUG_LOG(SP1, "Received ethernet frame");  // TODO remove this logging for performance?
```
lioncash, line 152: This should be resolved before merging.

Source/Core/Core/HW/EXI/BBA/XLINK_KAI_BBA.cpp
```diff
@@ -0,0 +1,281 @@
...
+
+        if (bytes_read < 1)
+        {
+          ERROR_LOG(SP1, "Failed to read from BBA, err=%lu", bytes_read - 4);
+        }
+        else if (self->readEnabled.IsSet())
+        {
+          // Only uncomment for debugging, the performance hit is too big otherwise
+          // DEBUG_LOG(SP1, "Read data: %s", ArrayToString(self->m_eth_ref->mRecvBuffer.get(),
+          // (u32)(bytes_read - 4), 0x10).c_str());
```
lioncash, line 165:
```suggestion
          // u32(bytes_read - 4), 0x10).c_str());
```

Source/Core/Core/HW/EXI/BBA/XLINK_KAI_BBA.cpp
```diff
@@ -0,0 +1,281 @@
...
+    // Otherwise we recieved control data or junk
+    else
+    {
+      char recv_str[9004] = {0x00};
+      memmove(recv_str, self->m_in_frame, bytes_read);
+      INFO_LOG(SP1, "Received XLink Kai control data: %s", recv_str);
+      std::string controlMsg = std::string(recv_str);
+
+      // connected;identifier
+      if (controlMsg.substr(0, 9) == "connected")
```
lioncash, line 180: In [StringUtils](https://github.com/dolphin-emu/dolphin/blob/master/Source/Core/Common/StringUtil.h#L167) we have a `StringBeginsWith` function that can be used here instead of creating several different strings via `substr`, which prevents some string churn.

Ditto for other locations in this file where substr is used and this can apply.

JosJuice:
@dolphin-emu-bot rebuild

Source/Core/Core/HW/EXI/BBA/XLINK_KAI_BBA.cpp
```diff
@@ -0,0 +1,277 @@
...
+                                   self->m_dest_port) != sf::Socket::Done)
+        {
+          ERROR_LOG(SP1, "ReadThreadHandler(): failed to reply to XLink Kai client keepalive");
+        }
+      }
+      // message;message_text
+      else if (StringBeginsWith(control_msg, "message"))
+      {
+        NOTICE_LOG(SP1, "XLink Kai message: %s",
+                   (control_msg.substr(8, control_msg.length() - 1)).c_str());
```
lioncash, line 239: Missed this in the last review (my bad), but you may want to place this in a local like so:

```cpp
std::string msg = control_msg.substr(8, control_msg.length() - 1);
```

as this is duplicated in the line below, which can cause some unnecessary string churn.

Source/Core/Core/HW/EXI/BBA/XLINK_KAI_BBA.cpp
```diff
@@ -0,0 +1,277 @@
...
+        // Show OSD message for 15 seconds to make sure the user sees it
+        OSD::AddMessage(control_msg.substr(8, control_msg.length() - 1), 15000);
```
lioncash, line 241: with the review comment above, this line can be transformed to:

```cpp
OSD::AddMessage(std::move(msg), 15000);
```

Ditto for analogues in the other else if cases below this one.

Source/Core/Core/HW/EXI/EXI_DeviceEthernet.h
```diff
@@ -297,34 +305,97 @@ class CEXIETHERNET : public IEXIDevice
...
 #if defined(WIN32) || defined(__linux__) || defined(__APPLE__) || defined(__FreeBSD__) ||          \
     defined(__OpenBSD__)
-  std::thread readThread;
-  Common::Flag readEnabled;
-  Common::Flag readThreadShutdown;
+    sf::UdpSocket m_sf_socket;
+    sf::IpAddress m_sf_recipient_ip;
+    char m_in_frame[9004];
+    char m_out_frame[9004];
+    std::thread readThread;
```
lioncash, line 389: I know this was pre-existing, so feel free to ignore the following reviews, but it'd be cool if these were normalized to follow our naming convention, since the code is being changed.

```suggestion
    std::thread m_read_thread;
```

Source/Core/Core/HW/EXI/EXI_DeviceEthernet.h
```diff
@@ -297,34 +305,97 @@ class CEXIETHERNET : public IEXIDevice
...
+    Common::Flag readEnabled;
```
lioncash, line 390:
```suggestion
    Common::Flag m_read_enabled;
```

Source/Core/Core/HW/EXI/EXI_DeviceEthernet.h
```diff
@@ -297,34 +305,97 @@ class CEXIETHERNET : public IEXIDevice
...
+    Common::Flag readThreadShutdown;
```
lioncash, line 391:
```suggestion
    Common::Flag m_read_thread_shutdown;
```

Source/Core/Core/HW/EXI/EXI_DeviceEthernet.h
```diff
@@ -297,34 +305,97 @@ class CEXIETHERNET : public IEXIDevice
...
+    static void ReadThreadHandler(XLinkNetworkInterface* self);
 #endif
+  };
+
+  std::unique_ptr<NetworkInterface> network_interface;
```
lioncash, line 396:
```suggestion
  std::unique_ptr<NetworkInterface> m_network_interface;
```

Source/Core/Core/HW/EXI/EXI_Device.h
```diff
@@ -19,6 +19,7 @@ enum TEXIDevices : int
   EXIDEVICE_AD16,
   EXIDEVICE_MIC,
   EXIDEVICE_ETH,
+  EXIDEVICE_ETHXLINK,
   // Was used for Triforce in the past, but the implementation is no longer in Dolphin.
   // It's kept here so that values below will stay constant.
```
BhaaLseN, line 24: Since this comment indicates that values below should "stay constant", I wonder if we should move the entry lower _or_ rethink/reword the comment (which IMO is the more likely candidate). I don't think we store this value anywhere in a config (so that it actually _has to be constant_), but I might be wrong.
Maybe @JosJuice can shed some light on the comment?

Source/Core/Core/HW/EXI/EXI_DeviceEthernet.cpp
```diff
@@ -43,6 +37,38 @@ CEXIETHERNET::CEXIETHERNET()
...
+  {
+  case BBADeviceType::TAP:
+    m_network_interface = std::make_unique<TAPNetworkInterface>(this);
+    INFO_LOG(SP1, "Created TAP physical network interface.");
+    break;
+  case BBADeviceType::XLINK:
+    // Perform sanity check on BBA MAC address, XLink requires the vendor OUI to be correct.
+    // Don't actually stop the BBA from initializing though
+    if (mac_addr_setting.substr(0, 8) != "00:09:bf" && mac_addr_setting.substr(0, 8) != "00:17:ab")
+      NOTICE_LOG(SP1,
```
BhaaLseN, line 50: This should have curlies since it spans over multiple lines.

Source/Core/DolphinQt/Settings/GameCubePane.cpp
```diff
@@ -186,10 +189,23 @@ void GameCubePane::OnConfigPressed(int slot)
...
-      SConfig::GetInstance().m_bba_mac = new_mac.toStdString();
+      SConfig::GetInstance().m_bba_tap_mac = new_mac.toStdString();
+    return;
+  }
+  case ExpansionInterface::EXIDEVICE_ETHXLINK:
+  {
+    bool ok;
+    const auto new_dest = QInputDialog::getText(
+        this, tr("Broadband Adapter (XLink Kai) Destination Address"),
+        tr("Enter IP address of device running the XLink Kai Client. \nFor more information see"
```
BhaaLseN, line 204: This has some whitespace at the end (both after `Client.` before the `\n` as well as at the end after the URL), do those serve any particular purpose? If not, please remove them.

Source/Core/Core/HW/EXI/EXI_Device.h
JosJuice: We are storing it in the config: https://github.com/dolphin-emu/dolphin/blob/a660033e8c394634a22c02a031969685fcdb4fba/Source/Core/Core/ConfigManager.h#L227

So please put the new entry at the bottom.

CrunchBite:
Before this goes any further I'm going to see if I can change the ethernet link state on the fly so the link status in game is accurate

CrunchBite:
I've added a check for if the BBA actually connected to XLink. This will notify the user if they try to use LAN mode and the connection wasn't successful, or if the connection was dropped.

This is the last change I'll make.

Source/Core/Core/Core.vcxproj
```diff
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
```
Techjar, line 1: I figured out that this weird change is in fact *removing* the byte order mark that currently exists in the repo.

Source/Core/Core/HW/EXI/EXI_DeviceEthernet.cpp
```diff
@@ -43,6 +37,39 @@ CEXIETHERNET::CEXIETHERNET()
...
+  case BBADeviceType::TAP:
+    m_network_interface = std::make_unique<TAPNetworkInterface>(this);
+    INFO_LOG(SP1, "Created TAP physical network interface.");
+    break;
+  case BBADeviceType::XLINK:
+    // Perform sanity check on BBA MAC address, XLink requires the vendor OUI to be correct.
+    // Don't actually stop the BBA from initializing though
+    if (mac_addr_setting.substr(0, 8) != "00:09:bf" && mac_addr_setting.substr(0, 8) != "00:17:ab")
+    {
+      NOTICE_LOG(SP1,
```
Techjar, line 51: Maybe this should also be a PanicAlertT, since XLink will refuse the bad MAC address. Otherwise, the user might not understand what's going on.

CrunchBite: Agreed, no point in continuing if the BBA wont work

Source/Core/DolphinQt/Settings/GameCubePane.cpp
```diff
@@ -186,12 +189,25 @@ void GameCubePane::OnConfigPressed(int slot)
   {
     bool ok;
     const auto new_mac = QInputDialog::getText(
-        this, tr("Broadband Adapter MAC address"), tr("Enter new Broadband Adapter MAC address:"),
-        QLineEdit::Normal, QString::fromStdString(SConfig::GetInstance().m_bba_mac), &ok);
+        this, tr("Broadband Adapter (TAP) MAC address"),
```
Rumi-Larry, line 192: Probably a i18n comment explaining what the terms "TAP" and "MAC" would be wise.

Source/Core/DolphinQt/Settings/GameCubePane.cpp
```diff
@@ -186,13 +189,27 @@ void GameCubePane::OnConfigPressed(int slot)
   {
     bool ok;
     const auto new_mac = QInputDialog::getText(
+        // i18n: TAP is a type of virtual network interface, MAC stands for Media Access Control
+        // i18n: A MAC address uniquely identifies a network interface (physical) like a serial
+        // number.
+        // i18n: "TAP" and "MAC" should be kept in translations.
```
JosJuice, line 195: You only have to write `i18n:` once.

JMC47:
So now that the actual code comments have slowed down, I guess I'll comment on my experience diving into, installing, and using Xlink-Kai.  It's pretty funny, I remember hearing about things like Warppipe on TechTV back when I was a kid, never imagining that someday I'd be using something like that... to test an emulator no less.

Unlike with BBA-TAP and Netplay, XLink-Kai has a lot more infrastructure.  Not only can you use it locally with multiple Dolphin instances/GameCubes on a network, but you can also play online through their sort of IRC-like chat service.  I didn't test it personally, but from what I saw people with ~30 - 40ms of latency were able to play without incident.

I'm not a technical wizard, so I don't exactly know how different XLink-Kai is versus BBA Tap.  Instead of talking about the different ways they work, I'll stick mostly to the actual user experience.  The installation process of XLink-Kai is very different than trying to setup TAP devices.  One of the first potential issues is that Windows is **very** adamant about not letting you run XLink-Kai.   It warns me about them with smartscreen, refuses to install it, then I tell it I want to install it, it says are you sure, then I finally can install it.  This might spook some users from using it.  I don't know if there's a way this could be fixed, but it would be appreciated for our less tech enthusiastic users.

Unlike TAP, you have to register to XLink-Kai's service.  It's not a very invasive process, mostly requiring an email (and I haven't gotten any spam.  Just the initial confirmations.), username and password.  When you start XLink-Kai it connects your default web browser to a sort of web-irc interface.  This was all initially *very* overwhelming for just trying to use BBA.  After all, I was just trying to set it up locally...

However, once I figured out how everything worked it got simpler.  Basically, none of that was necessary for me if I'm just playing locally.  If I did want to play with friends, I can add them as contacts and then if I search for a console in Dolphin at the same time as them while using XLink-Kai, it'll automatically work.  You can also join rooms for playing with non-friends.  Unlike netplay, these games are very sensitive to ping so I don't know how feasible this is, but if you have friends around you it'll probably work okay.  As a bonus to Dolphin users, the chat interface for the room will get communicated over imgui, though you can't respond through Dolphin at this time like in Netplay's chat GUI.

A few comments on the overall interface.  I just kind of wish there was a "local" mode that didn't have all the registration requirements.  For long term preservation, I always worry about services disappearing and for all the flaws of BBA Tap, at least it'll always function.  I also worry about the lengthy process keeping away more casual users, but I don't know how much of that is even avoidable considering how BBA or XLink-Kai works.

However, as an enthusiast feature to use BBA, once you're past the initial setup everything worked very well.  Setting up rooms and such isn't hard and it's very "hands off."  You can get a BBA match going in less than a few minutes.

**Testing Results**

I tested 3 out of the 4 BBA titles.  I was too lazy to setup the Japanese exclusive Homeworld.

Mario Kart: Double Dash performed perfectly in a few short races.  I didn't get any desyncs once I got connected, however I did get one connection error.  All I had to do was retry and they connected next try so honestly that's probably something that can happen on console.

Kirby Air Ride:  This one is a bit more complicated but in the end I got it to work smoothly.  Kirby Air Ride is a lot more sensitive to latency... or so I thought.  I actually got it to work over Wi-Fi latency at least by upping Dolphin's Emulated CPU clock.  I don't know why this smoothed out the frame-rate but it did.  I need to try connecting this to a real console and seeing what's up sometime but that's for another day.

1080 Snowboarding:  I didn't do that much testing on this one, but it worked and had no desyncs over 10 minutes of testing.  Instantly connected too.  Not much else to say.

CrunchBite:
Heya, I'll address things one at a time. Full disclosure I'm the current head of Team XLink and the primary developer of the client and server architecture.

> Unlike with BBA-TAP and Netplay, XLink-Kai has a lot more infrastructure. Not only can you use it locally with multiple Dolphin instances/GameCubes on a network, but you can also play online through their sort of IRC-like chat service. I didn't test it personally, but from what I saw people with ~30 - 40ms of latency were able to play without incident.

These are the traffic patterns that work with XLink and this BBA PR:
- Local Dolphin and another local Dolphin (7 at once have been tested running MKDD)
- Local Dolphin and local Gamecube(s) (XLink handling the Ethernet bridging)
- Local Dolphin and remote Dolphin/Gamecube/Wii (nintendont)

As you mentioned most gamecube games are _extremely_ sensitive to latency so avoiding WiFi and being geographically close to other players is important.

> I'm not a technical wizard, so I don't exactly know how different XLink-Kai is versus BBA Tap. Instead of talking about the different ways they work, I'll stick mostly to the actual user experience. The installation process of XLink-Kai is very different than trying to setup TAP devices. One of the first potential issues is that Windows is **very** adamant about not letting you run XLink-Kai. It warns me about them with smartscreen, refuses to install it, then I tell it I want to install it, it says are you sure, then I finally can install it. This might spook some users from using it. I don't know if there's a way this could be fixed, but it would be appreciated for our less tech enthusiastic users.

From MS's point of view XLink checks most of the 'sketchiness' boxes despite being harmless; Unsigned exe, requires admin rights, has its own internal http client and httpd, self contained crypto functions, processes and injects raw ethernet frames etc etc... Windows 10 is especially picky about all this.

> Unlike TAP, you have to register to XLink-Kai's service. It's not a very invasive process, mostly requiring an email (and I haven't gotten any spam. Just the initial confirmations.), username and password. When you start XLink-Kai it connects your default web browser to a sort of web-irc interface. This was all initially _very_ overwhelming for just trying to use BBA. After all, I was just trying to set it up locally...

For local only use I agree its overkill but the main focus of the service is online play :)
You will **never** receive any sort of spam from using XLink Kai, we don't sell/share information with anyone else, and the service will always be completely free.

> However, once I figured out how everything worked it got simpler. Basically, none of that was necessary for me if I'm just playing locally. If I did want to play with friends, I can add them as contacts and then if I search for a console in Dolphin at the same time as them while using XLink-Kai, it'll automatically work. You can also join rooms for playing with non-friends. Unlike netplay, these games are very sensitive to ping so I don't know how feasible this is, but if you have friends around you it'll probably work okay. As a bonus to Dolphin users, the chat interface for the room will get communicated over imgui, though you can't respond through Dolphin at this time like in Netplay's chat GUI.

"these games" specifically being GameCube LAN games

> A few comments on the overall interface. I just kind of wish there was a "local" mode that didn't have all the registration requirements. For long term preservation, I always worry about services disappearing and for all the flaws of BBA Tap, at least it'll always function. I also worry about the lengthy process keeping away more casual users, but I don't know how much of that is even avoidable considering how BBA or XLink-Kai works.

When XLink eventually shuts down (wont be for years yet) we'll likely open-source everything. With that said; now that this PR exists it is fairly obvious how to add additional BBA types to Dolphin so I expect there will be alternatives in the future.

> However, as an enthusiast feature to use BBA, once you're past the initial setup everything worked very well. Setting up rooms and such isn't hard and it's very "hands off." You can get a BBA match going in less than a few minutes.

XLink itself is going through a bit of a revamp lately. A new website is on the way as well as a set of new setup guides. I've written a simple guide to using this BBA geared towards users who are already somewhat familiar with the service: https://www.teamxlink.co.uk/wiki/Dolphin_Gamecube_XLink_BBA_Tutorial

JMC47:
Any other thoughts on this PR?  I'm fine with the Dolphin side of it usage wise.  I'd like to know if I should prep a Progress Report entry for this.

JMC47:
Is there anyone else who wants to review this?  I obviously can't review the code, but most of it appears to be localized into the BBA/EXI sections and it doesn't seem to impact anything in emulation it shouldn't.

If anyone else doesn't have any qualms with this, then it's been sitting long enough and I'll merge it when the feature goes live into XLink.

Source/Core/Core/HW/EXI/BBA/XLINK_KAI_BBA.cpp
```diff
@@ -0,0 +1,298 @@
...
+
+  while (!self->m_read_thread_shutdown.IsSet())
+  {
+    if (!self->IsActivated())
+      break;
+
+    // XLink supports jumboframes but Gamecube BBA does not. We need to support jumbo frames
+    // *here* because XLink *could* send one
+    std::size_t bytes_read = 0;
+    if (self->m_sf_socket.receive(self->m_in_frame, std::size(self->m_in_frame), bytes_read, sender,
```
BhaaLseN, line 147: How (un)likely is it that this `receive` above returns garbage, or rather: less than 4 bytes read? Depending on the answer, it might be worthwhile/necessary to add more checks before we attempt to copy data into `temp_check` below.

Also, this prints an error to the log, but continues anyways. Should this `break`/ `continue` (or even `return`) in case reading fails?
(Similarly, if this needs to be changed in any way, the `send` below also does that)

Source/Core/DolphinQt/Settings/GameCubePane.cpp
```diff
@@ -186,12 +189,28 @@ void GameCubePane::OnConfigPressed(int slot)
   {
     bool ok;
     const auto new_mac = QInputDialog::getText(
-        this, tr("Broadband Adapter MAC address"), tr("Enter new Broadband Adapter MAC address:"),
-        QLineEdit::Normal, QString::fromStdString(SConfig::GetInstance().m_bba_mac), &ok);
+        // i18n: TAP is a type of virtual network interface, MAC stands for Media Access Control
+        // A MAC address uniquely identifies a network interface (physical) like a serial
+        // number. "TAP" and "MAC" should be kept in translations.
+        this, tr("Broadband Adapter (TAP) MAC address"),
```
BhaaLseN, line 195: The MAC is still used in both cases (TAP and XLink Kai), should this translation string really be updated to include the word "TAP"?

Also, since XLink Kai _also_ requires a MAC (not to mention from a specific vendor; as validated by the `CEXIETHERNET` ctor), it appears that currently you have to select TAP to generate (or update) the MAC and then switch to XLink Kai to update the Destination Address.
Is this something we want to address here (assuming I'm not mistaken by just reading the code; I didn't run or test anything)?

Source/Core/Core/HW/EXI/EXI_DeviceEthernet.cpp
```diff
@@ -43,6 +37,40 @@ CEXIETHERNET::CEXIETHERNET()
...
   }
 
+  switch (type)
+  {
+  case BBADeviceType::TAP:
+    m_network_interface = std::make_unique<TAPNetworkInterface>(this);
+    INFO_LOG(SP1, "Created TAP physical network interface.");
+    break;
+  case BBADeviceType::XLINK:
+    // TODO start BBA with network link down, bring it up after "connected" response from XLink
```
BhaaLseN, line 47: Is this `TODO` outdated, or does this still need to be addressed? If so, when and where?
(I don't necessarily want to force you to implement this here if you have plans for a follow-up PR, I just want to clarify if it is still up2date and not just left for whoever comes afterwards)

Source/Core/Core/HW/EXI/EXI_DeviceEthernet.cpp
```diff
@@ -23,15 +23,9 @@ namespace ExpansionInterface
...
+CEXIETHERNET::CEXIETHERNET(BBADeviceType type)
 {
-  tx_fifo = std::make_unique<u8[]>(BBA_TXFIFO_SIZE);
-  mBbaMem = std::make_unique<u8[]>(BBA_MEM_SIZE);
-  mRecvBuffer = std::make_unique<u8[]>(BBA_RECV_SIZE);
-
-  MXHardReset();
-
-  // Parse MAC address from config, and generate a new one if it doesn't
+  // Parse BBA (TAP) MAC address from config, and generate a new one if it doesn't
```
BhaaLseN, line 28: Same reasoning about TAP vs. XLink Kai for the MAC as in other places; it isn't TAP specific as far as I can see.

Source/Core/Core/HW/EXI/EXI_DeviceEthernet.cpp
```diff
@@ -43,6 +37,40 @@ CEXIETHERNET::CEXIETHERNET()
...
+    INFO_LOG(SP1, "Created TAP physical network interface.");
+    break;
+  case BBADeviceType::XLINK:
+    // TODO start BBA with network link down, bring it up after "connected" response from XLink
+
+    // Perform sanity check on BBA MAC address, XLink requires the vendor OUI to be correct.
+    // Don't actually stop the BBA from initializing though
+    if (mac_addr_setting.substr(0, 8) != "00:09:bf" && mac_addr_setting.substr(0, 8) != "00:17:ab")
+    {
+      PanicAlertT("BBA MAC address %s invalid for XLink Kai. MAC address must start with 00:09:bf "
```
BhaaLseN, line 53: Also, this was mentioned by @Techjar on IRC: those two MAC vendors are apparently the ones used by real GC hardware.

I feel we should be documenting this in some way; perhaps even in this message - and also include the fact that it is being checked for this specific reason (at least I assume thats why the check is here)

Source/Core/Core/HW/EXI/EXI_DeviceEthernet.cpp
CrunchBite: This TODO is still valid. Ideally it would be good to mimic a correct link state and only have the game think the link is up if the connection to XLink is up. I tried to get this working but was unsuccessful; I dont know enough about how the BBA works to set the correct flags for link down.

Source/Core/DolphinQt/Settings/GameCubePane.cpp
CrunchBite: I've updated the wording used

Source/Core/DolphinQt/Settings/GameCubePane.cpp
```diff
@@ -186,12 +189,27 @@ void GameCubePane::OnConfigPressed(int slot)
   {
     bool ok;
     const auto new_mac = QInputDialog::getText(
+        // i18n: TAP is a type of virtual network interface, MAC stands for Media Access Control
```
BhaaLseN, line 192: Since you removed the "TAP" from the string, it need not be in the translation comment I suppose.

JMC47:
I think this will be a helpful change for the BBA communities.  I hope we can continue to improve things but the change seems fairly isolated and shouldn't be too hard to maintain.  It's seen multiple reviews and I've seen the X-Link Kai community do many tests and I've verified it worked as well.  I'm essentially good with the user side of things as is and Bhaal approved the code.

I do think some of the GUI things can be made clearer in the future, but that can be handled in followups.

JMC47:
Did some quick and dirty performance testing with BBA Expansions turned off to make sure this doesn't impact performance otherwise, and it seems fine.

This is ready as far as I'm concerned.

PR #8855, opened 2020-06-08
DolReader: treat overly large data segment as bss

shuffle2:
some homebrew sets `dataOffset[i]+dataSize[i]` larger than file size

### DIFF
--- a/Source/Core/Core/Boot/DolReader.cpp
+++ b/Source/Core/Core/Boot/DolReader.cpp
@@ -32,7 +32,7 @@ DolReader::~DolReader() = default;
 
 bool DolReader::Initialize(const std::vector<u8>& buffer)
 {
-  if (buffer.size() < sizeof(SDolHeader))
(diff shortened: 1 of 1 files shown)


Helios747:
why is shuffle2 not auto trusted

@dolphin-emu-bot rebuild

lioncash:
Good question. Sent an invite to be added to the trusted devs

shuffle2:
tbh i don't really want to feel like i have responsibilities to the dolphin project. I've removed myself from everything related a while ago. considering the extremely low amount of patches I submit now, I think that's fine.

lioncash:
Trusted devs doesn't really have any responsibilities (other than, y'know, not being malicious with the build system). It's just for people that are trustworthy enough that the buildbot can automatically proceed for them.

Either way, I respect the decision

shuffle2:
However, someone could still merge it at some point. 😑

sepalani:
@dolphin-emu-bot rebuild

PR #8856, opened 2020-06-08
VSProps: Don't use /std:c++latest

JosJuice:
Using /std:c++latest can cause problems for people in the future who want to build old versions of Dolphin on new versions of Visual Studio (for bisecting and such). Let's set /std:c++17 instead, which is consistent with what we're doing in CMake.

This should also fix https://bugs.dolphin-emu.org/issues/12086 (untested), but it isn't a long-term fix, since it'll break again once we switch over to using C++20.

### DIFF
--- a/Source/VSProps/Base.props
+++ b/Source/VSProps/Base.props
@@ -80,7 +80,7 @@
       <FunctionLevelLinking>true</FunctionLevelLinking>
       <MultiProcessorCompilation>true</MultiProcessorCompilation>
       <!--Enable latest C++ standard-->
-      <LanguageStandard>stdcpplatest</LanguageStandard>
+      <LanguageStandard>stdcpp17</LanguageStandard>
       <!--Enable Standard Conformance-->
       <ConformanceMode>true</ConformanceMode>
       <!--Enforce some behaviors as standards-conformant when they don't default as such.-->


JosJuice:
We're already incompatible with those.

lioncash:
My disagreement with merging this is that it makes it more annoying to keep code future-proofed, and keeping it set to an older standard can prevent proactive fixes from being made as the compiler evolves and supports the newest standard. This can result in potentially larger breakages occurring all at once (if any breaking changes are made between standards, which has already happened) when the standard is eventually changed.

Particularly if anything were ever designed using a deprecated or removed feature in an upcoming standard, this would go completely undiagnosed until the standard version is incremented, potentially making it a pain to refactor the now-nonconforming code.

I don't think wanting to build dated versions of software is really worth getting rid of assuming the latest evolving standard as the default.

BhaaLseN:
It does however match up with CMake, where we specify C++17 as well.

lioncash:
And the CMake files can't be adjusted?

BhaaLseN:
No, i meant we use `set(CMAKE_CXX_STANDARD 17)` here:
https://github.com/dolphin-emu/dolphin/blob/4c64e18b244e8995f68d4166fc7d89afc0004a17/Source/CMakeLists.txt#L18

With this change here, both CMake and VSProps specify C++17 consistently, thats what I meant.

lioncash:
Right, but what I was saying is that the CMake file can likely be changed instead to track latest when using MSVC, as opposed to lowering it. The only reason we don't track "latest" with GCC and Clang is that they have no equivalent switch for doing so (as far as I know). There are reasons for having MSVC track latest, which I've already explained.

BhaaLseN:
I'm pretty sure the main reason was though that MSVC had no C++17 switch before, and changed the definition of "latest" between versions.
I don't mind either way though.

Techjar:
I'm siding with lioncash on this one. Having at least one of our builds always use the latest C++ standard, whatever that may be, forces us to ensure code is compliant with the latest standard. This way, when the time comes that we want to actually enable the full feature set of a newer standard, our code is already fully compliant and all we have to do is a flip a switch.

shuffle2:
It seems the current Qt release is 5.15.x , so maybe just update to that instead of 5.14.2?

Techjar:
Yeah, we usually just update to whatever is the latest at the time.

lioncash:
It's likely sufficient to use:

```cmake
# TODO: Remove once Qt is updated
if (MSVC)
  add_definitions(-D_HAS_DEPRECATED_RESULT_OF)
endif()
```

Which would allow the definition of result_of to be seen, while still allowing the standard to remain on latest.

Pokechu22:
I was able to build properly without these changes now that #9046 has been merged.

JosJuice:
Yes, this PR no longer has any use as a quick fix for the build being broken. But the main purpose of this PR isn't really that, but rather trying to avoid similar problems in the future.

leoetlino:
My two cents: Considering that we can't use C++20 on the Linux and other builders, I agree with @lioncash that it's nice to have one builder use the latest standard just so we can fix breaking changes as they come up rather than all at once 1-2 years later...

leoetlino:
Closing as the build issues are already fixed and we just changed CMake builds to be consistent with msbuild and use c++latest

PR #8858, opened 2020-06-08
Android: Add Pause / Unpause Emulation Toggle

Ebola16:
Implements https://bugs.dolphin-emu.org/issues/10602 and should survive orientation changes.

- [x] Confirm `fragment_ingame_menu.xml` changes for Android TV work.

Screenshot_20200608-120650_Dolphin Emulator

### DIFF
--- a/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/activities/EmulationActivity.java
+++ b/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/activities/EmulationActivity.java
@@ -70,6 +70,9 @@ public final class EmulationActivity extends AppCompatActivity
 
   private Settings mSettings;
 
+  private MenuItem mPauseEmulationButton;
+  private MenuItem mUnpauseEmulationButton;
+
   private boolean mDeviceHasTouchScreen;
   private boolean mMenuVisible;
 
@@ -81,12 +84,14 @@ public final class EmulationActivity extends AppCompatActivity
   private String mSelectedGameId;
   private int mPlatform;
   private String[] mPaths;
+  private static boolean sUserPausedEmulation;
   private boolean backPressedOnce = false;
 
   public static final String EXTRA_SELECTED_GAMES = "SelectedGames";
   public static final String EXTRA_SELECTED_TITLE = "SelectedTitle";
   public static final String EXTRA_SELECTED_GAMEID = "SelectedGameId";
   public static final String EXTRA_PLATFORM = "Platform";
+  public static final String EXTRA_USER_PAUSED_EMULATION = "sUserPausedEmulation";
 
   @Retention(SOURCE)
   @IntDef({MENU_ACTION_EDIT_CONTROLS_PLACEMENT, MENU_ACTION_TOGGLE_CONTROLS, MENU_ACTION_ADJUST_SCALE,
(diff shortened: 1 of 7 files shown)


Ebola16:
Reminder to update `NativeLibrary.UnPauseEmulation` conditions if #8833 is merged first (or vice versa).

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/activities/EmulationActivity.java
```diff
@@ -80,12 +85,14 @@
   private String mSelectedGameId;
   private int mPlatform;
   private String[] mPaths;
+  private static boolean sUserPausedEmulation;
```
JosJuice, line 88: This should not be static.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/activities/EmulationActivity.java
```diff
@@ -70,6 +70,11 @@
 
   private Settings mSettings;
 
+  private MenuItem mPauseEmulationButton;
+  private MenuItem mUnpauseEmulationButton;
+  private View mPauseEmulationTv;
+  private View mUnpauseEmulationTv;
```
JosJuice, line 76: To me, it would make more sense if these four didn't exist and we just called `findItem`/`findViewById` in the functions that need to access these. It's a bit hard to follow where and how these get initialized the way it is now.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/activities/EmulationActivity.java
Ebola16: `EmulationActivity.hasUserPausedEmulation()` uses `sUserPausedEmulation` and is called in EmulationFragment so don't we need a static variable for `EmulationActivity.hasUserPausedEmulation()`?

JosJuice: You should be able to use `activity.hasUserPausedEmulation()` in EmulationFragment.

Ebola16: But in EmulationFragment, class `EmulationState` is static and `hasUserPausedEmulation()` is called within that class. `activity` isn't static (and can't be static) so wouldn't my code be the best approach here?

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/activities/EmulationActivity.java
Ebola16: Just to be clear, calling `findItem/findViewById` in both `showPauseEmulationButton()` and `showUnpauseEmulationButton()` while having both of those methods take `Menu` as an argument is more readable? That would duplicate the `findItem/findViewById` for each method and make me add a variable to pass `Menu`, which seems worse to me.

I did find that I can group the variable's assignments together though.

JosJuice: Oh, I had missed the part that it was called on `menu` and not just the activity... I suppose it makes sense the way it currently is, then.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/activities/EmulationActivity.java
JosJuice: Ah, I had missed that. But can't we just make EmulationState non-static, or pass in `activity` when creating an EmulationState?

It is preferable to not use static state in a way that would cause things to break when creating two instances of a class at the same time, even if we never actually do that with EmulationActivity.

Ebola16: Let's be safe and pass `activity`

Ebola16:
Something went wrong while testing. Looking into it now.

Ebola16:
@JosJuice passing `activity` to `EmulationState` didn't seem to work after orientation changes and `EmulationState` has a static inner class so I think it would be best to go with the static variable here. Tested.

Ebola16:
@JosJuice Thanks! That helped and I think I got it working on Android TV but I still can't truly test it since I can't initialize OGL on the Android Emulator. Another check would be appreciated. Marking this as working on Android TV.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/fragments/MenuFragment.java
```diff
@@ -21,11 +22,17 @@
 
 public final class MenuFragment extends Fragment implements View.OnClickListener
 {
+  private View mPauseEmulationTv;
+  private View mUnpauseEmulationTv;
```
JosJuice, line 26: These two don't need to have `Tv` in the name anymore. (Feel free to ignore this if you want to.)

PR #8860, opened 2020-06-09
Android: Disallow multiple launcher activities

Ebola16:
Fixes https://bugs.dolphin-emu.org/issues/11935

### DIFF
--- a/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/activities/EmulationActivity.java
+++ b/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/activities/EmulationActivity.java
@@ -73,6 +73,7 @@ public final class EmulationActivity extends AppCompatActivity
   private boolean mDeviceHasTouchScreen;
   private boolean mMenuVisible;
 
+  private static boolean sIgnoreLaunchRequests = false;
   private static boolean sIsGameCubeGame;
 
   private boolean activityRecreated;
@@ -196,6 +197,11 @@ public final class EmulationActivity extends AppCompatActivity
 
   public static void launch(FragmentActivity activity, GameFile gameFile)
   {
+    if (sIgnoreLaunchRequests)
+      return;
+
+    sIgnoreLaunchRequests = true;
+
     Intent launcher = new Intent(activity, EmulationActivity.class);
 
     launcher.putExtra(EXTRA_SELECTED_GAMES, scanForSecondDisc(gameFile));
@@ -207,6 +213,11 @@ public final class EmulationActivity extends AppCompatActivity
 
   public static void launchFile(FragmentActivity activity, String[] filePaths)
   {
+    if (sIgnoreLaunchRequests)
+      return;
+
+    sIgnoreLaunchRequests = true;
+
     Intent launcher = new Intent(activity, EmulationActivity.class);
     launcher.putExtra(EXTRA_SELECTED_GAMES, filePaths);
 
@@ -235,6 +246,11 @@ public final class EmulationActivity extends AppCompatActivity
     }
 
     activity.startActivity(launcher);
+  }
+
+  public static void stopIgnoringLaunchRequests()
+  {
+    sIgnoreLaunchRequests = false;
   }
 
   public static void clearWiimoteNewIniLinkedPreferences(Context context)
--- a/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/fragments/EmulationFragment.java
+++ b/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/fragments/EmulationFragment.java
@@ -411,6 +411,7 @@ public final class EmulationFragment extends Fragment implements SurfaceHolder.C
             Log.debug("[EmulationFragment] Starting emulation thread.");
             NativeLibrary.Run(mGamePaths);
           }
+          EmulationActivity.stopIgnoringLaunchRequests();
         }, "NativeEmulation");
         emulationThread.start();
       }


Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/activities/EmulationActivity.java
```diff
@@ -196,45 +197,60 @@
...
   {
-    Intent launcher = new Intent(activity, EmulationActivity.class);
-
-    launcher.putExtra(EXTRA_SELECTED_GAMES, scanForSecondDisc(gameFile));
-    launcher.putExtra(EXTRA_SELECTED_TITLE, gameFile.getTitle());
-    launcher.putExtra(EXTRA_SELECTED_GAMEID, gameFile.getGameId());
-    launcher.putExtra(EXTRA_PLATFORM, gameFile.getPlatform());
-    activity.startActivity(launcher);
-  }
+    if (!sStartingEmulation)
```
sepalani, line 200: Can be replaced with:
```java
if (sStartingEmulation)
  return;
```

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/activities/EmulationActivity.java
```diff
@@ -196,45 +197,60 @@
...
       launcher.putExtra(EXTRA_SELECTED_GAMEID, gameFile.getGameId());
       launcher.putExtra(EXTRA_PLATFORM, gameFile.getPlatform());
+      activity.startActivity(launcher);
     }
-    else
+  }
+
+  public static void launchFile(FragmentActivity activity, String[] filePaths)
+  {
+    if (!sStartingEmulation)
```
sepalani, line 216: Ditto.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/activities/EmulationActivity.java
Ebola16: This is void so we're not returning anything. Isn't the current code cleaner than the following?
> ```java
> if (sStartingEmulation)
>   // do nothing;
> else
>   doSomething();
> ```

Ebola16:
@Tilka (I think you were flacs from IRC) After the IRC discussion, I looked at the code again. I think launch() and launchfile() are actually called from the same thread. Those methods eventually trigger `EmulationFragment.runWithValidSurface()`, which is where concurrent threads are made. At least I'm pretty sure that's what's going on here, still shaky with thread safety.

@sepalani I looked into a few options to see if I could detect if the NativeEmulation thread is already running. `Thread.isAlive()` was a possibility but I've read a few reports saying that it's unreliable during thread startup. Other options were more complicated than necessary from what I've read. The static boolean seems to work.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/activities/EmulationActivity.java
JosJuice: Yes, the current code is cleaner than what you just suggested, but we would prefer what sepalani suggested.

Ebola16: Inverted

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/activities/EmulationActivity.java
Ebola16: Inverted

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/activities/EmulationActivity.java
JosJuice: Your change doesn't implement sepalani's suggestion.

Just to be clear, the suggestion is that you put the if-return construct at the beginning of the function, followed by the main code of the function without any `else` statement.

Ebola16: Not sure what I was thinking. Went with @sepalani's original suggestions.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/activities/EmulationActivity.java
```diff
@@ -72,6 +72,7 @@
 
   private boolean mDeviceHasTouchScreen;
   private boolean mMenuVisible;
+  private static boolean sStartingEmulation = false;
```
sepalani, line 75: This variable should probably be renamed because its name is a bit misleading. It is set to `true` when the emulation is launching and/or running.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/activities/EmulationActivity.java
```diff
@@ -237,6 +248,11 @@ public static void launchFile(FragmentActivity activity, String[] filePaths)
     activity.startActivity(launcher);
   }
 
+  public static void finishedStartingEmulation()
+  {
+    sStartingEmulation = false;
```
sepalani, line 253: If the variable is renamed, this method should be renamed appropriately.

Ebola16:
@sepalani Good to hear we found the same thing on using a boolean here. Is the new variable naming scheme acceptable now?

sepalani:
@Ebola16 The naming is better, IMO. You just need to rename `finishedLaunchingEmulation` to something more fitting.

The other alternative I found regarding this issue is using a lock (mutex) but I'm not well versed in Android to know if it's the Android-way of handling this.

Ebola16:
@sepalani Hopefully the latest version of this PR solves the naming problems. I tried a lock object (like the one in #8833) but ended up getting an ANR since I was making the UI thread wait.

PR #8861, opened 2020-06-10
Make netplay's "same game" check more robust

JosJuice:
Instead of comparing the game ID, revision, disc number and name, we can compare a hash of important parts of the disc including all the aforementioned data but also additional data such as the FST. The primary reason why I'm making this change is to let us catch more desyncs before they happen, but this should also fix https://bugs.dolphin-emu.org/issues/12115. As a UX bonus, the UI can now distinguish the case where a client doesn't have the game at all from the case where a client has the wrong version of the game, and we can show translated/custom names in the netplay game list.

### DIFF
--- a/Source/Core/Core/CMakeLists.txt
+++ b/Source/Core/Core/CMakeLists.txt
@@ -37,6 +37,7 @@ add_library(core
   PatchEngine.h
   State.cpp
   State.h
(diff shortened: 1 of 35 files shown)


AdmiralCurtiss:
Something went wrong with your rebase here.

PR #8863, opened 2020-06-11
IOS/Network: Fix dual-core IPCEvent issue on shutdown

sepalani:
This PR is an attempt to fix the dual-core IPCEvent scheduling error from the following issues:
 - https://dolp.in/i11862
 - https://dolp.in/i11910

Ready to be reviewed.


BhaaLseN:
I wonder if this is the right place to do this (like, could it be inside those functions in case someone _else_ calls them?) or the right granularity (do we need other methods in `Shutdown` to be included as well?) to wrap them.

sepalani:
@BhaaLseN I'm not very familiar with the dual-core mode, so I don't know where would be the best place to handle this.

As far as I understand, the network code (I assume it also applies to IOS code) is executed from the CPU thread and isn't called from other threads. There should be no way to schedule IPC events from other threads. However, it does on Shutdown since the shutdown function destroy the SocketManager, releases the sockets and aborts socket pending ops. That's exactly where the schedule issue comes from.

The real question as you said is from which threads each one of these methods should be called and make sure it doesn't create a deadlock. Right now, I've only seen IOS/Network code raising this problem that's why I forced them to run from the CPU thread.

PR #8864, opened 2020-06-13
Update Readme.md

Alen-Alic:
Grammar fixes in line 37.


JosJuice:
This change isn't correct. The installation is what fails, not the download. And the error message is shown by Android, not by Dolphin.

PR #8867, opened 2020-06-13
Make Free Look a proper controller, move to separate UI

iwubcode:
This switches Free Look to be a proper controller in both the UI and internally.  I was always planning on doing this but assumed it would receive push back due to many users feeling comfortable with the current approach.

The biggest advantage to doing this is that `HotkeyManager` was only meant for Hotkeys.  So adding Gyro controls like I did in #8747 or other controls which I have planned, is really gross code-wise.

But there are a couple other reasons I'm doing this:

* You can now have proper Free Look profiles!  And per-game profiles are possible in the future
* #8832 was always meant to support multiple views, this change allows us to have multiple controllers

The freelook controllers don't really fit into any current category ("Inputs", "Graphics", etc).  So I've introduced another menu option "Free Look Settings".

image

The Free Look controller screen loses the description information it had previously.  I don't think this information should exist in Dolphin, we don't have _any_ other input specific descriptions in other places.  Therefore I'm hoping to add a new wiki page for Free Look.  I've added a link to this page to describe the different modes and various options supported.  This will be expanded as more of these PRs get merged.  @mbc07 hoping we can work together on this if possible or someone else who can help me with the wiki.

image

I took the liberty to make the change everyone was asking for in #8747.  The "General" tab now spans horizontally.  Additionally, I removed the "Zoom" since it isn't really a zoom and moved those controls to the "Move" section. 

Code specific details

- I moved speed under the camera.  This is needed if/when we make multiple Free Look cameras.  And could allow for game-specific speed settings in the future (maybe reusing the concept of game world units in meters like needed for #8380 )
- I am still reusing the HotkeyManager to call into the FreelookManager.  This is to avoid adding another thread.  Longer-term, maybe we need a thread to handle various pieces of input?  This was a quick fix


mbc07:
Will take a look at this very soon. Which wiki URL you're currently pointing to?

iwubcode:
@mbc07 - https://wiki.dolphin-emu.org/index.php?title=Free_Look  (Or maybe I should make it "one word"?)

Source/Core/DolphinQt/Config/Mapping/MappingWindow.h
```diff
@@ -42,7 +42,9 @@ class MappingWindow final : public QDialog
     // Wii
     MAPPING_WIIMOTE_EMU,
     // Hotkeys
-    MAPPING_HOTKEYS
+    MAPPING_HOTKEYS,
+    // Freelook
+    MAPPING_FREELOOK
```
BhaaLseN, line 47: This is really minor, but how about including a comma here anyways? Gets rid of a not-really-a-change (like with `MAPPING_HOTKEYS` above) on the next change.

Source/Core/VideoCommon/FreeLookCamera.cpp
```diff
@@ -234,10 +234,26 @@ void FreeLookCamera::Reset()
...
+float FreeLookCamera::GetSpeed() const
+{
+  return m_speed;
+}
+
 void FreeLookCamera::DoState(PointerWrap& p)
 {
   if (p.mode == PointerWrap::MODE_WRITE || p.mode == PointerWrap::MODE_MEASURE)
   {
+    p.Do(m_speed);
```
BhaaLseN, line 256: You should update the state version over here since you changed this:
https://github.com/dolphin-emu/dolphin/blob/0ef9b998cea75ee7025b2446f7c70f8efa296dd5/Source/Core/Core/State.cpp#L76-L77

Source/Core/Core/BootManager.cpp
```diff
@@ -319,6 +321,7 @@ bool BootCore(std::unique_ptr<BootParameters> boot, const WindowSystemInfo& wsi)
   }
 
   StartUp.m_GPUDeterminismMode = ParseGPUDeterminismMode(StartUp.m_strGPUDeterminismMode);
+  StartUp.m_freelook = Config::Get(Config::GFX_FREE_LOOK);
```
BhaaLseN, line 324: Not sure if I missed this somewhere, but since this is set once on boot; do we lock the setting so it cannot be changed while a game is running?
Also, on the other hand: is there a reason why we have to (or rather: want to) prevent this being enabled while in-game?

mbc07:
About "Freelook" vs "Free Look", I don't have a strong opinion on either, but I think the terminology should be consistent (e.g. if we end with "Free Look", it should be worded like that everywhere, on the UI and on the wiki URL as well). I would like to hear @MayImilae thoughts about this.

About the PR, it seems broken currently?
* Nothing happens when you click on Controllers button in the toolbar
* There's no device dropdown list on the Freelook Controller settings (Options => Other Settings => Configure), just the refresh button is there

Source/Core/Core/BootManager.cpp
iwubcode: This shouldn't impede changing while in game (at least I didn't think?) this is just so that the controllers can be initialized/setup properly (just like with Wii/Gamecube)...though maybe this needs to be handled differently since you can turn it off.

Source/Core/VideoCommon/FreeLookCamera.cpp
iwubcode: Ah good call.  Thanks for reminding me

iwubcode:
@mbc07 -  thanks for taking a look, appreciate your feedback.

> Nothing happens when you click on Controllers button in the toolbar
Hmm somehow I broke this, I'll take a look.

>There's no device dropdown list on the Freelook Controller settings (Options => Other Settings => Configure), just the refresh button is there
Looks like the window needs to be increased, not really sure why this is needed when other windows (Wiimote/Gamecube/Hotkey controller mapping windows) work fine.  They use the same base

iwubcode:
@mbc07 - updated to fix your concerns

Source/Core/Core/BootManager.cpp
iwubcode: Still thinking of how best to resolve this.  Appreciate you bringing it up

BhaaLseN: I mean enabling/disabling freelook (setting `GFX_FREE_LOOK`), which is then handled in `BootCore::EmuThread`. It is read there twice; once during the normal boot and once in the `ScopeGuard` to shut things down later. It _might_ be ok if you only set it here, since it should have the same value in both; but this also means enabling it while a game is running (and it was set to `false` previously) won't have it initialized.

iwubcode: Ah I see.  I will have to see how Wiimote/Gamecube controllers get dynamically connected then.  I don't want users to remember to set this during startup (that's not how it currently works!)

mbc07:
The UI issues indeed are fixed on the current version of this PR. Just don't forget to change the wiki URL to be consistent with the UI (currently referring as "freelook").

I'm a bit busy at the moment but ping me when this PR is about to be merged so I can work on the wiki guide in case someone else doesn't do that before. Having that description of the camera modes as a starting point would be great but I couldn't find where exactly that was submitted (I'm pretty sure it was on GitHub, but where?)...

Source/Core/Core/BootManager.cpp
iwubcode: I've fixed this by making the freelook controllers _always_ initialized and instead have their individual state controlled by the camera control type.  A value of "None" will turn off that camera.

iwubcode:
I'm still playing around with this but I have some ideas for the future.

Here's the rundown:

* Dropped the previous enable / disable checkbox.  Each controller/camera will have a "None" option just like the Gamecube controllers have.  This allows you to turn on/off individual cameras.  Important when dealing with multi-screen
* Changed "Configure" to "Configure Controller".  Admittedly a bit more verbose but I want the freedom to add an "Options" button later, for non-controller specific options.  And that's also why I...
* Removed the freelook settings from the graphics ini and instead made a specific freelook ini for non-controller settings.  Right now, one messily setting!  However, this will be more important in the future.  Here's what the setting looks like:

```
[Freelook.Controller1]
ControlType = 0
```
(that sets the control type to "None")

--

One thing I'm still trying to figure out is how to handle the "toggle".  Toggle makes less sense when the configuration it deals with is a dropdown!  Still, I have something that works for now.

iwubcode:
@mbc07 - sure, I'll ping you again when things are ready to go!

FYI I started on the wiki page - https://wiki.dolphin-emu.org/index.php?title=Freelook .  I wasn't sure if there's a way to hide it from the public and/or mark it as in development..

mbc07:
> I wasn't sure if there's a way to hide it from the public and/or mark it as in development..

Apart from some very specific administration-related special pages, everything on the wiki is public, so don't worry for now. I'll keep checking it from time to time and point you anything that might need attention in the meantime...

iwubcode:
With #8820 merged, I updated this review to include FoV.  Let me know what you think!

Source/Core/Core/FreelookManager.cpp
```diff
@@ -0,0 +1,223 @@
...
+  if (s_config.ControllersNeedToBeCreated())
+  {
+    s_config.CreateController<FreelookController>(0);
+  }
+
+  s_config.RegisterHotplugCallback();
+
+  Freelook::g_Config.Refresh();
+
+  // Load the saved controller config
```
BhaaLseN, line 199: This comment seems a bit pointless as it is right now. We can see it loads from config (and can probably guess that it is the controller config), but I don't see a particular reason why there needs to be a comment pointing out that "we load the saved config".

Is there anything else this comment should probably tells us? And if so, does it also apply to the one in `LoadConfig`?

Source/Core/Core/FreelookConfig.h
```diff
@@ -0,0 +1,42 @@
+// Copyright 2020 Dolphin Emulator Project
+// Licensed under GPLv2+
+// Refer to the license.txt file included.
+
+// IMPORTANT: UI etc should modify g_FreelookConfig. Freelook code should read
```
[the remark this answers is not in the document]
iwubcode: Thanks for catching this.  At one point I didn't have these in the freelook namespace.  I could swear I updated this, maybe it got lost when I was re-arranging / tidying up commits.  I'll add it back!

Source/Core/Core/FreelookManager.cpp
iwubcode: This structure was copied from Gamecube.  I'm pretty sure the comment existed there.  I will remove it

Source/Core/Core/FreelookConfig.cpp
```diff
@@ -0,0 +1,37 @@
...
+Config::Config()
+{
+  cameraConfig.iControlType = ControlType::None;
+}
+
+void Config::Refresh()
+{
+  if (!s_has_registered_callback)
+  {
+    ::Config::AddConfigChangedCallback(
```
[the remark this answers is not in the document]
iwubcode: This structure was copied from `VideoConfig`, I went back and forth on keeping the comment (see https://github.com/dolphin-emu/dolphin/blob/master/Source/Core/VideoCommon/VideoConfig.cpp#L64 ) or if it'd just be confusing.  Since `UpdateActiveConfig` is still called via the video thread, I imagine the race condition could still occur but wasn't 100% sure.  Thought it'd be better to be safe than sorry.

As for why this isn't just registered in the constructor?  I'm not sure

Source/Core/Core/ConfigLoaders/BaseConfigLoader.cpp
```diff
@@ -90,7 +90,7 @@ const std::map<Config::System, int> system_to_ini = {
     {Config::System::Logger, F_LOGGERCONFIG_IDX},
     {Config::System::Debugger, F_DEBUGGERCONFIG_IDX},
     {Config::System::DualShockUDPClient, F_DUALSHOCKUDPCLIENTCONFIG_IDX},
-};
+    {Config::System::Freelook, F_FREELOOKCONFIG_IDX}};
```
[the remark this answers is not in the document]
iwubcode: Ah good point, let me try adding a trailing comma (I can't recall if lint asked me or not)

Source/Core/Core/FreelookConfig.cpp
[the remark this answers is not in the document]
BhaaLseN: I can guess why `VideoConfig` does it (that is: keeping two copies; one for the UI update and another that is used by the video code; and swapping between them) - there is more than one member that might be updated during `Refresh`, and if it did that while a game was running (and unpaused) it might pick up one setting in the new state but another setting in the old state.

I suppose we can leave it at that, as long as we remember there are other places doing the same thing; for a later point if/when we change this.

MayImilae:
Gave this a look! Sorry for the delay, everything is just, well, you know. Everything is weird.

This resets all freelook hotkeys. It's to be expected due to the move, but I absolutely need to comment on this in the Progress Report! I really need to remember this time!

Ok so, first of all. It's written as "Free Look", so please use that in your settings and code. I know we commonly shorten it to the lazy "freelook", I'm even going to in this comment, but when properly writing or working on it we always use "Free Look". Please check those.

"Options > Other". Haaate it. Please Just make it "Options > Free Look" With the options for freelook expanding and it never really belonging in graphics anyway, I totally understand needing to move it to get more room. However, Options > Other is just terrible. Please don't make a dumping ground that devs in the future can throw crap into! You need the space, so that's fine, just own it! Make it a Freelook window, and all freelook related things can go there!

I do not like the Camera 1: None solution. It obfusgates turning Freelook on or off, and that's really bad. It's not uncommon for users to turn freelook on, play with it, then forget it's on, only to come back and be confused by it. As such, it's really important that there is a very accessible and extremely obvious way to turn Freelook on and off. Also, this will be very useful when you add multiple cameras, as it will allow users to turn Freelook off entirely without messing with their settings. Also this makes the freelook toggle hotkey situation a lot easier too, we can just have one hotkey that toggles freelook off globally and not worry about X number of freelook toggling hotkeys. In my head, this would look like "Enable Freelook: [toggle]" and with the camera options beneath it. As for where to put the Freelook global toggle hotkey, it can just live in the hotkeys panel. That would be perfectly fine.

-----

Now before I go, I want to comment on something. This isn't related to this but, I don't know how else to bring this up so I'll comment on it here. I'm still not feeling the First Person and Orbit camera modes. They just... they just aren't useful. Every scenario I've thrown at them, they've just been worse than the existing Six Axis. First Person doesn't slowly add roll, but most games have at least some angle to their camera, and without roll, there is no way to fix it. It really needs an overhaul to blender levels of awesome to be really useful, something completely different from the floating camera approach that freelook has had until now. Remind me to show you the blender FPS free look sometime, it's awesome, but it would probably be a hella lot of work to implement somethingon that level. Since Orbital lacks the context of selecting a point in space the way a 3D modeling software does, it needs to have a configurable rotation point that it orbits around to be useful, and it needs to have the option to roll the camera to fix the angled camera issues there. I won't press you to fix or change these any time soon, but just, this has been nagging at me for a while.

iwubcode:
Thank you for looking @MayImilae !

>Ok so, first of all, "Freelook" versus "Free Look". I'm absolutely prefer "Free Look"

Did not even consider this until @mbc07 mentioned it.  I updated all text I could find to this wording, let me know if you see places I might have missed.

>Options > Other". Haaate it. Please Just make it "Options > Free Look"

Yeah, I wasn't a huge fan either.  I almost added what you suggested originally but stopped because I didn't know if that'd be too much clutter for the Options menu if we added more sections.  Hence trying to condense things with a poorly named general window.  But if #8823 ever gets merged, it'll be much cleaner, so I guess it doesn't matter too much.  Regardless, I switched to "Free Look Settings" to match the other options.

>I do not like the Camera 1: None solution. It obfusgates turning Freelook on or off, and that's really bad.

It is certainly different but I don't know if it's as bad as you are making it out to be.  The "None" is the exact verbiage used for all the other controllers.

>Also, this will be very useful when you add multiple cameras, as it will allow users to turn Freelook off entirely without messing with their settings

Honestly, I don't like this because then users still need a way to opt out of multiple views.  Here's a mockup of what I'd eventually like to see:

image

I hate to get too far into the multiview discussion because it's still very much a WIP.  However, I wanted to account for the possibility of it in my design.  I didn't want users to get surprised with another drastic UI change later on to support the feature.

> Also this makes the freelook toggle hotkey situation a lot easier too, we can just have one hotkey that toggles freelook off globally and not worry about X number of freelook toggling hotkeys.

Free Look toggle is a bit odd but I don't mind it.  If multiview becomes a thing, it would allow for easy drop-in / drop-out behavior.  Almost like that "press start to join" text that would flash in arcade games.

> I'm still not feeling the First Person and Orbit camera modes. They just... they just aren't useful.

They were partially a bit selfish but I do think they are useful in some situations and I at least  hope they're of use to some users.

*First Person* - I don't take as many screenshots as you but as I have tested these Free Look features and as I've been playing some games, I have used this as my primary mode.  I rarely need rolling and found it to be more trouble when using Free Look casually as a supplement to gameplay.  I think it'll be much more useful when #8747 is merged.  Six Axis was completely unusable with that in my testing.

*Orbital* - I don't know if we'll get there but I'd love to have a Dolphin experience that promotes easy spectating.  Netplay is too cumbersome for this but it's the only solution at the moment.  Orbital is primarily meant for a casual spectator mode where you want to focus on the player playing.  I also see its use for a young child when a parent is playing a single-player game with multiview.  Completely agree that it's less useful than it could be since it focuses on the camera origin point.  It'd be nice if we could fix that in the future.  I'll be sure to credit you ;)

Orbital was also added just because a lot of applications / websites allow for this type of movement (my main inspiration was [noclip](https://noclip.website)).

I have a couple other ideas for control types but nothing concrete yet.

> It really needs an overhaul to blender levels of awesome to be really useful, something completely different from the floating camera paradigm that freelook has had until now. Remind me to show you the blender FPS free look sometime,

I'd love this.  I looked at blender documentation / videos and couldn't see anything specific to what you were mentioning.  I did not actually download it though.

mbc07:
So, we are settled on "Free Look", right? I'll move the WIP guide on the wiki to reflect the new naming, please update the URL on the PR before merging...

iwubcode:
@mbc07 - thank you!  I  updated the link to point to the new wiki location.

Source/Core/Core/FreelookManager.cpp
```diff
@@ -0,0 +1,222 @@
...
+}
+
+ControllerEmu::ControlGroup* GetGroup(int pad_num, FreelookGroup group)
+{
+  return static_cast<FreelookController*>(s_config.GetController(pad_num))->GetGroup(group);
+}
+
+void Update()
+{
+  static_cast<FreelookController*>(s_config.GetController(0))->Update();
```
BhaaLseN, line 219: Technically, shouldn't this update _all_ controllers?
Since we support multiple controllers (but only add _one_ in `Initialize`), I think we should either make this future-proof by just looping (if you can) or leaving a note that those two should be synchronized (if you can't).

Thoughts?

iwubcode: Good point.  My plan was to update this when I added more controller support but there's no reason I couldn't loop right now.

iwubcode:
There were a few things in the original PR that were to support the hypothetical multi-view concept.  While I'd like to see that eventually, I didn't want those changes to block this review from being merged.

Because of this, I added a checkbox to enable / disable free look.  I also removed the "toggle" binding in the free look controller window and instead kept the global toggle in the hotkey settings.

Source/Core/Core/Config/FreelookSettings.h
```diff
@@ -0,0 +1,25 @@
+// Copyright 2020 Dolphin Emulator Project
+// Licensed under GPLv2+
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <string>
```
leoetlino, line 7: This seems to be unused

Source/Core/Core/FreelookConfig.h
```diff
@@ -0,0 +1,42 @@
+// Copyright 2020 Dolphin Emulator Project
+// Licensed under GPLv2+
+// Refer to the license.txt file included.
+
+// IMPORTANT: UI etc should modify Freelook::g_Config. Freelook code should read
+// Freelook::g_ActiveConfig. The reason for this is to get rid of race conditions etc when the
+// configuration changes in the middle of a frame. This is done by copying Freelook::g_Config to
+// Freelook::g_ActiveConfig at the start of every frame. Noone should ever change members of
```
leoetlino, line 8: typo: "No one"

Source/Core/Core/FreelookManager.h
```diff
@@ -0,0 +1,57 @@
...
+
+InputConfig* GetConfig();
+ControllerEmu::ControlGroup* GetGroup(int pad_num, FreelookGroup group);
+
+}  // namespace Freelook
+
+class FreelookController final : public ControllerEmu::EmulatedController
+{
+public:
+  explicit FreelookController(const unsigned int index);
```
leoetlino, line 42: const is useless here

Source/Core/Core/FreelookManager.h
```diff
@@ -0,0 +1,57 @@
...
+
+  std::string GetName() const override;
+  void LoadDefaults(const ControllerInterface& ciface) override;
+
+  ControllerEmu::ControlGroup* GetGroup(FreelookGroup group);
```
leoetlino, line 47: This can be a const member function

Source/Core/DolphinQt/Config/FreelookWidget.h
```diff
@@ -0,0 +1,30 @@
+// Copyright 2020 Dolphin Emulator Project
+// Licensed under GPLv2+
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <QDialog>
```
leoetlino, line 7: This seems to be unused.

Source/Core/Core/FreelookManager.cpp
```diff
@@ -99,47 +99,50 @@ void FreelookController::Update()
 {
   g_freelook_camera.SetControlType(Freelook::g_ActiveConfig.cameraConfig.iControlType);
 
-  if (m_move_buttons->controls[0]->GetState<bool>())
-    g_freelook_camera.MoveVertical(-g_freelook_camera.GetSpeed());
+  if (g_freelook_camera.IsActive())
```
leoetlino, line 102: You can invert the conditional:
```
if (!g_freelook_camera.IsActive())
  return;
```
to avoid having to indent most of the lines in this function

leoetlino:
One small thing that bothers me a bit: the feature is called Free Look, yet most of the time it's referred to as "freelook" in the code... And it's inconsistent, since the camera code is FreeLookCamera. Could we make it consistent?

iwubcode:
@leoetlino - thank you so much for your review!  I try and use "Free Look" with user facing text but obviously I can't use spaces for variables.

When you say consistent, do you mean in terms of case or something else?

leoetlino:
Yes, FreeLook instead of Freelook (for the config file name as well).

iwubcode:
Thank you for the feedback @leoetlino !  All comments addressed~

iwubcode:
With the tooltips now in, I added tooltips back to the combo-box / checkbox.

Source/Core/Core/FreeLookConfig.cpp
```diff
@@ -0,0 +1,39 @@
...
+  cameraConfig.iControlType = ControlType::SixAxis;
+  bEnabled = false;
+}
+
+void Config::Refresh()
+{
+  if (!s_has_registered_callback)
+  {
+    ::Config::AddConfigChangedCallback(
+        []() { Core::RunAsCPUThread([]() { g_Config.Refresh(); }); });
```
leoetlino, line 32: `[] { Core::RunAsCPUThread([] { g_Config.Refresh(); }); });`

Source/Core/Core/FreeLookConfig.h
```diff
@@ -0,0 +1,42 @@
...
+enum class ControlType : int
+{
+  SixAxis,
+  FPS,
+  Orbital
+};
+
+struct CameraConfig
+{
+  ControlType iControlType;
```
leoetlino, line 24: Hm I missed this during my last review, but this should be `control_type` to match our naming conventions

Source/Core/Core/FreeLookConfig.h
```diff
@@ -0,0 +1,42 @@
...
+};
+
+// NEVER inherit from this class.
+struct Config final
+{
+  Config();
+  void Refresh();
+
+  CameraConfig cameraConfig;
```
leoetlino, line 33: camera_config

Source/Core/Core/FreeLookConfig.h
```diff
@@ -0,0 +1,42 @@
...
+  bool bEnabled;
```
leoetlino, line 34: enabled

Source/Core/Core/FreeLookConfig.h
```diff
@@ -0,0 +1,42 @@
...
+};
+
+extern Config g_Config;
```
leoetlino, line 37: Globals should be avoided in new code. Consider making `g_config` static rather than global (which is slightly better) and `GetConfig` / `GetActiveConfig`.

This would also let you enforce the rule that nothing should modify the active config directly (by returning a const-ref from GetActiveConfig).

Source/Core/Core/FreeLookManager.cpp
```diff
@@ -0,0 +1,195 @@
...
+
+void FreeLookController::LoadDefaults(const ControllerInterface& ciface)
+{
+  EmulatedController::LoadDefaults(ciface);
+
+  auto hotkey_string = [](std::vector<std::string> inputs) {
+    return "@(" + JoinStrings(inputs, "+") + ')';
+  };
+
+  m_move_buttons->SetControlExpression(0, hotkey_string({"Shift", "E"}));  // Up
```
leoetlino, line 62: Would it be possible to add enums to replace the numbers?

Source/Core/DolphinQt/Config/FreeLookWidget.cpp
```diff
@@ -0,0 +1,112 @@
...
+         "leave this unchecked.</dolphin_emphasis>"));
+  m_freelook_controller_configure_button = new QPushButton(tr("Configure Controller"));
+
+  m_freelook_control_type = new GraphicsChoice({tr("Six Axis"), tr("First Person"), tr("Orbital")},
+                                               Config::FL1_CONTROL_TYPE);
+  m_freelook_control_type->SetTitle(tr("Free Look Control Type"));
+  m_freelook_control_type->SetDescription(tr(
+      "Changes the in-game camera type during Free Look.<br><br>"
+      "Six Axis: Offers full camera control on all axes, akin to moving a spacecraft in zero "
+      "gravity. This is the most powerful Free Look option but is the most challenging to use.<br "
```
leoetlino, line 46: `<br>`

Source/Core/Core/FreeLookConfig.h
iwubcode: This is a great idea.  I also cleaned up the comment

iwubcode:
Thank you @leoetlino for your review.  I went ahead and addressed those changes!

Source/Core/Core/FreeLookManager.cpp
```diff
@@ -0,0 +1,237 @@
...
+  case FreeLookGroup::Other:
+    return m_other_buttons;
+  default:
+    return nullptr;
+  }
+}
+
+void FreeLookController::Update()
+{
+  g_freelook_camera.SetControlType(FreeLook::GetActiveConfig().camera_config.control_type);
```
BhaaLseN, line 137: Is this really necessary on `Update` (which I assume is either on every frame or every poll interval for controllers - so a LOT of times)? It feels like this could be done once during startup and then when configuration changes.

I mean, it _does_ have a check if the change is even necessary; but it still feels weird to be here.

Source/Core/Core/FreeLookManager.cpp
```diff
@@ -0,0 +1,232 @@
...
+#include "InputCommon/InputConfig.h"
+
+#include "VideoCommon/FreeLookCamera.h"
+#include "VideoCommon/OnScreenDisplay.h"
+
+namespace
+{
+enum MoveButtons
+{
+  Up,
```
leoetlino, line 24: These aren't enum classes, so each member should be prefixed with the name of the enum (`MoveButtons_Up` instead of `Up` for example). Otherwise this introduces `Up`, `Down`, etc. into file scope and it's not clear what they are :)

Alternatively, wrap them inside a namespace (e.g. namespace MoveButtons) so that referring to them requires typing MoveButtons:: like with enum classes.

Source/Core/Core/FreeLookManager.cpp
```diff
@@ -0,0 +1,232 @@
...
+  Down,
+  Left,
+  Right,
+  Forward,
+  Backward
```
leoetlino, line 29: This should have a trailing comma to reduce diff noise if more buttons are added in the future.

Source/Core/Core/FreeLookManager.cpp
```diff
@@ -0,0 +1,232 @@
...
+};
+
+enum SpeedButtons
+{
+  Decrease,
+  Increase,
+  Reset
```
leoetlino, line 36: ditto

Source/Core/Core/FreeLookManager.cpp
```diff
@@ -0,0 +1,232 @@
...
+};
+
+enum OtherButtons
+{
+  ResetView
```
leoetlino, line 41: ditto

Source/Core/Core/FreeLookManager.cpp
```diff
@@ -0,0 +1,232 @@
...
+};
+
+enum FieldOfViewButtons
+{
+  IncreaseX,
+  DecreaseX,
+  IncreaseY,
+  DecreaseY
```
leoetlino, line 49: ditto

Source/Core/Core/FreeLookManager.cpp
iwubcode: This was left here because the control type did effect whether the camera was active.  Additionally, in multiview each controller was tied to an individual camera.

Considering multiview is not a thing _yet_, I'll move this back to `RenderBase` until that PR.

Source/Core/Core/FreeLookManager.cpp
iwubcode: Makes sense, just ugly.  I guess I'll switch to the namespace approach, thanks!

Source/Core/Core/FreeLookManager.cpp
```diff
@@ -0,0 +1,242 @@
...
+void FreeLookController::Update()
+{
+  if (m_move_buttons->controls[MoveButtons::Up]->GetState<bool>())
+    g_freelook_camera.MoveVertical(-g_freelook_camera.GetSpeed());
+
+  if (m_move_buttons->controls[MoveButtons::Down]->GetState<bool>())
+    g_freelook_camera.MoveVertical(g_freelook_camera.GetSpeed());
+
+  if (m_move_buttons->controls[MoveButtons::Left]->GetState<bool>())
+    g_freelook_camera.MoveHorizontal(g_freelook_camera.GetSpeed());
```
leoetlino, line 159: Shouldn't this be `-speed` instead of `+speed`?

iwubcode: You would think but right is actually negative, not entirely sure why tbh.

leoetlino (approved):

JMC47 (approved):
I haven't been paying too much attention to this, but I only had to use it for like 2 minutes to realize that this is a huuuuuuuuuuuuge improvement.

BhaaLseN (approved):
Code looks good, untested.

lioncash (approved):

Source/Core/DolphinQt/HotkeyScheduler.cpp
```diff
@@ -537,57 +538,15 @@ void HotkeyScheduler::Run()
...
-
-    if (IsHotkey(HK_FREELOOK_INCREASE_FOV_Y, true))
-      g_freelook_camera.IncreaseFovY(g_freelook_camera.GetFovStepSize());
-
-    if (IsHotkey(HK_FREELOOK_DECREASE_FOV_Y, true))
-      g_freelook_camera.IncreaseFovY(-1.0f * g_freelook_camera.GetFovStepSize());
-
-    if (IsHotkey(HK_FREELOOK_RESET, true))
-      g_freelook_camera.Reset();
+    FreeLook::UpdateInput();
```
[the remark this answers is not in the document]
Filoppi: Also, not sure you've seen that FreeLook won't run if Hotkeys are disabled, there is a
`if (!HotkeyManagerEmu::IsEnabled()) return` condition above, I'm guessing it should run independently of that setting.

Source/Core/Core/FreeLookManager.cpp
```diff
@@ -0,0 +1,247 @@
...
+    return m_fov_buttons;
+  case FreeLookGroup::Other:
+    return m_other_buttons;
+  default:
+    return nullptr;
+  }
+}
+
+void FreeLookController::Update()
+{
```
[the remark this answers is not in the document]
iwubcode: Great catch!  This was not intentional.  I don't know much about all the inner-workings of inputs, so I didn't notice.  I'll open a PR to add it after doing the appropriate testing

Source/Core/DolphinQt/HotkeyScheduler.cpp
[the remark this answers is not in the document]
iwubcode: I did miss that hotkeys can be disabled.  But as I mention in this PR, this was a simple stop-gap to get things merged more quickly.  Long term I'd expect there to be a separate input thread that calls both hotkey and FreeLook systems.

I'd personally like to add a new "background input" setting specifically for FreeLook because while I'd assume users would want it to operate it similarly to the game, they might not.  I'd also add an "Input Sources" button like in the main controller window.  It's confusing to me if users have to go somewhere else to modify those settings for FreeLook.

PR #8870, opened 2020-06-14
JitArm64: no intermediate rounding for FMA

Tilka:
Completely untested atm.

### DIFF
--- a/Source/Core/Core/PowerPC/JitArm64/JitArm64_Paired.cpp
+++ b/Source/Core/Core/PowerPC/JitArm64/JitArm64_Paired.cpp
@@ -53,7 +53,7 @@ void JitArm64::ps_mergeXX(UGeckoInstruction inst)
       ARM64Reg V0 = fpr.GetReg();
       m_float_emit.INS(size, V0, 0, VA, 1);
       m_float_emit.INS(size, V0, 1, VB, 0);
-      m_float_emit.ORR(reg_encoder(VD), reg_encoder(V0), reg_encoder(V0));
+      m_float_emit.MOV(reg_encoder(VD), reg_encoder(V0));
       fpr.Unlock(V0);
     }
     break;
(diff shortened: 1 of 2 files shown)


Source/Core/Core/PowerPC/JitArm64/JitArm64_Paired.cpp
```diff
@@ -110,46 +110,136 @@ void JitArm64::ps_maddXX(UGeckoInstruction inst)
...
+      m_float_emit.MOV(VD, V0);
+    }
     break;
   case 28:  // ps_msub
-    m_float_emit.FMUL(size, V0, VA, VC);
-    m_float_emit.FSUB(size, VD, V0, VB);
+    // d = a * c - b
+    if (d != a && d != c)
+    {
+      m_float_emit.FNEG(size, VD, VB);
```
MerryMage, line 163: Consider instead: `FNMSUB`, which is equivalent to FMLA except the accumulator operand is negated beforehand.

Source/Core/Core/PowerPC/JitArm64/JitArm64_Paired.cpp
```diff
@@ -110,46 +110,136 @@ void JitArm64::ps_maddXX(UGeckoInstruction inst)
...
+      m_float_emit.FMLA(size, VD, VA, VC);
+    }
+    else
+    {
+      m_float_emit.FNEG(size, V0, VB);
+      m_float_emit.FMLA(size, V0, VA, VC);
+      m_float_emit.MOV(VD, V0);
```
MerryMage, line 170: Another thing to consider for the `d == b` case:```
      m_float_emit.FMLS(size, VD, VA, VC);
      m_float_emit.FNEG(size, VD, VD);
```
Note however this would not be accurate when the fp rounding mode is to +/- infinity.

iwubcode:
I tested four Wii games: Last Story, Little King's Story, Opoona, Rune Factory Frontier (opening areas, 5 minutes each) and did not see any of the issues @Tilka mentioned would happen if there were issues.  Not sure if more games need to be tested or more in-depth testing needs to be done but I didn't notice any regressions.

JosJuice:
> EDIT: I am assuming "JitArm64" is the Jit recompiler CPU option.

It is the JIT recompiler for ARM CPUs, which shows up as "JIT ARM64 Recompiler" on Android and "JIT Arm64 (experimental)" in DolphinQt.

Source/Core/Core/PowerPC/JitArm64/JitArm64_Paired.cpp
Tilka: Thanks, added. I'm already not handling +/- infinity rounding modes correctly, so I don't mind adding another case.

JosJuice:
Tested two games on my phone, with no problems observed.

This is probably just me not being used to ARM assembly (so I don't mind if someone merges without me getting an answer), but why do we need a special case for `d == a || d == c`?

degasus:
@JosJuice Thanks for testing. PPC has the 4 operant isntruction (d = a*c + b), aarch64 only the 3 operant instruction (d += a*c). So we need one copy d=b, and another one if d overlaps a or c. Else a or c will get overwritten before the MAD.

MANGOM1LK:
This PR is breaking stuff in Sonic Unleashed: https://youtu.be/CWRrhoeGArE

Notably, stick inputs are now incorrect. This should probably be reverted for the time being until more testing can happen :/

MerryMage:
@MANGOM1LK Spotte an obvious bug. Please try the following build: https://dolphin.ci/#/builders/15/builds/1251

PR #8871, opened 2020-06-14
UnitTests: clamp to avoid undefined conversions

Tilka:
Out-of-range conversions are undefined behavior.

### DIFF
--- a/Source/UnitTests/VideoCommon/VertexLoaderTest.cpp
+++ b/Source/UnitTests/VideoCommon/VertexLoaderTest.cpp
@@ -106,6 +106,15 @@ protected:
   std::unique_ptr<VertexLoaderBase> m_loader;
 };
 
+template <typename T>
+constexpr T clamp(float v)
+{
+  const T lo = std::numeric_limits<T>::lowest();
+  const T hi = std::numeric_limits<T>::max();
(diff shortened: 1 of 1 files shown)


Source/UnitTests/VideoCommon/VertexLoaderTest.cpp
```diff
@@ -106,6 +106,15 @@ class VertexLoaderTest : public testing::Test
   std::unique_ptr<VertexLoaderBase> m_loader;
 };
 
+template <typename T>
+T clamp(float v)
```
sepalani, line 110: Can't `lo` and `hi` be declared as `const` and `clamp` as a `constexpr` function?

Source/UnitTests/VideoCommon/VertexLoaderTest.cpp
```diff
@@ -106,6 +106,15 @@ class VertexLoaderTest : public testing::Test
   std::unique_ptr<VertexLoaderBase> m_loader;
 };
 
+template <typename T>
```
leoetlino, line 109: Would it make sense to move this function into MathUtil? (MathUtil::SafeFloatToInt?)

leoetlino:
Superseded by #9480

PR #8872, opened 2020-06-14
GCMemcard: Rename types to be more specific.

AdmiralCurtiss:
Mostly because it struck me as very scary that we have structs in the global namespace just called 'Header' and 'Directory' that are for the memory card implementation...

I also rewrote the static GCMemcard::Format() while I was here so it no longer relies on constructing objects into an `u8*`.


JosJuice:
Instead of prefixing GCMemcard to a bunch of names, wouldn't it be better to add a namespace?

AdmiralCurtiss:
...Yeah, probably.

PR #8874, opened 2020-06-15
gc-font-tool: use original glyph widths

Tilka:
This solves alignment issues with multiple GameCube games that print things like "Press (A) to continue" where "(A)" is a separately rendered button texture.

Font files regenerated from Droid Sans Fallback (without dithering).


JosJuice:
Could you post a screenshot of what this looks like in-game?

Tilka:
It's still not pretty.

old:
RB4E08_2020-06-15_10-13-58

new:
RB4E08_2020-06-15_10-11-16

IPL ROM:
RB4E08_2020-06-15_09-57-07

docs/gc-font-tool.cpp
```diff
@@ -123,6 +135,22 @@ const uint16_t windows_1252_font_table[] =
 	SEQUENCE_32(0xA0),
 	SEQUENCE_64(0xC0),
 };
+static_assert(std::size(windows_1252_font_table) == 224);
+
+// List of horizontal glyph widths
+const uint8_t windows_1252_widths_table[] =
```
iwubcode, line 141: Just curious, how did you get these arrays?

Should you use a `std::array` instead?

Tilka: It's not clear to me whether something as trivial as glyph widths falls under copyright, but I'd rather be safe than sorry, so I've closed this PR.

PR #8875, opened 2020-06-15
DolphinQt: Show a warning when launching an NKit disc image

JosJuice:
It is my opinion that nobody should use NKit disc images without being aware of the drawbacks of them. Since it seems like almost nobody who is using NKit disc images knows what NKit is (hmm, now how could that have happened...?), I am adding a warning to Dolphin so that you can't run NKit disc images without finding out about the drawbacks. In case someone really does want to use NKit disc images, the warning has a "Don't show this again" option. Unfortunately, I can't retroactively add the warning where it's most needed: in Dolphin 5.0, which does not support Wii NKit disc images.

image

### DIFF
--- a/Source/Core/Core/Config/MainSettings.cpp
+++ b/Source/Core/Core/Config/MainSettings.cpp
@@ -142,6 +142,7 @@ const Info<std::string> MAIN_FS_PATH{{System::Main, "General", "NANDRootPath"}, 
 const Info<std::string> MAIN_SD_PATH{{System::Main, "General", "WiiSDCardPath"}, ""};
 
 // Main.Network
+
 const Info<bool> MAIN_NETWORK_SSL_DUMP_READ{{System::Main, "Network", "SSLDumpRead"}, false};
 const Info<bool> MAIN_NETWORK_SSL_DUMP_WRITE{{System::Main, "Network", "SSLDumpWrite"}, false};
 const Info<bool> MAIN_NETWORK_SSL_VERIFY_CERTIFICATES{
@@ -149,4 +150,8 @@ const Info<bool> MAIN_NETWORK_SSL_VERIFY_CERTIFICATES{
 const Info<bool> MAIN_NETWORK_SSL_DUMP_ROOT_CA{{System::Main, "Network", "SSLDumpRootCA"}, false};
 const Info<bool> MAIN_NETWORK_SSL_DUMP_PEER_CERT{{System::Main, "Network", "SSLDumpPeerCert"},
                                                  false};
+
+// Main.Interface
+
+const Info<bool> MAIN_SKIP_NKIT_WARNING{{System::Main, "Interface", "SkipNKitWarning"}, false};
 }  // namespace Config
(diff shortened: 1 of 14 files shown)


mbc07:
While warning about NKIT shortcomings is a good idea, I personally think this is **way** too verbose and to be frank, most of the drawbacks don't even apply on normal usage (e.g. opening Dolphin and booting a game). 

With that said, I think a more reasonable approach would be unconditionally displaying a short OSD warning whenever you boot a NKIT image (similar to the recently added WIA block size warning) and only display that PanicAlert-like popup when trying to use a feature that is negatively affected by using NKIT (e.g. TAS tools and NetPlay) or when trying to boot a negatively affected game (like Super Paper Mario -- is any other game affected besides it?).

That's currently not accounted for but for NetPlay specifically, the ideal approach would be throwing the warning only when there's mixed NKIT/plain ISO/other compressed formats users on the same session, as a Netplay session where all users from the session are using NKIT images isn't any more prone to desync than a session where all users use regular ISOs AFAICT, but since that's way out of the scope of this PR, unconditionally showing the popup when starting Netplay should do.

Now, about the popup itself:
* Is the line saying that Wii games don't work in older versions and other software really relevant? After all this warning will end showing only in Dolphin versions that do support Wii NKIT images and old Dolphin versions that can't boot them will keep failing with cryptical error messages regardless;
* I personally would drop these checkboxes and just have "Don't show again" "OK" and "Cancel" buttons instead.

JosJuice:
I suppose a summary of what you said is that the warning feels too intimidating. I intentionally wanted it to feel intimidating, due to how many people use NKit while having *absolutely zero idea* what the drawbacks are and thinking it's a great format for use with Dolphin. I wanted to implement the warning message in a way such that it couldn't be ignored. Since I'm reluctant to change this, I suppose we should wait and see what others think about how intimidating the warning should be, and then I can change it if it turns out that more people are against my idea.

> (like Super Paper Mario -- is any other game affected besides it?)

One of the games in Namco Museum is affected unless you do some particular edits to the NKit config. It is possible that there are more games with problems that have not been found, but I'd guess that any such games would be of the more obscure kind.

> That's currently not accounted for but for NetPlay specifically, the ideal approach would be throwing the warning only when there's mixed NKIT/plain ISO/other compressed formats users on the same session, as a Netplay session where all users from the session are using NKIT images isn't any more prone to desync than a session where all users use regular ISOs AFAICT, but since that's way out of the scope of this PR, unconditionally showing the popup when starting Netplay should do.

For what it's worth, PR #8861 will tell you that players have the wrong version of the game if you try to mix NKit and non-NKit, just like if you try to mix regional variants of a game.

> Is the line saying that Wii games don't work in older versions and other software really relevant? After all this warning will end showing only in Dolphin versions that do support Wii NKIT images and old Dolphin versions that can't boot them will keep failing with cryptical error messages regardless;

My hope is that this warning will inform not only the specific user who reads it, but also that people who read it might stop encouraging others to use NKit. I believe mentioning it is relevant for the latter, and I also believe that the warning about other software is relevant even if the warning about older Dolphin versions is not.

JMC47:
NKIT games are causing loads of problems on netplay and Dolphin in general.  We need to dissuade people from using these if possible.

JosJuice:
Added a line to the warning about savestates not being compatible. Thanks to Filoppi for pointing this out to me.

degasus:
I think the alert should include a suggested action: Use NKit to convert this file back to ISO, and dolphin to convert the ISO to RVZ.

JosJuice:
> Use NKit to convert this file back to ISO

Anyone who knows how to convert to NKit should know how to convert from NKit. I don't want to include anything that only is useful for pirates.

> and dolphin to convert the ISO to RVZ.

I don't want this to be seen as a message where I'm trying to push my own format. I just want to make users aware of the facts about NKit, because these facts need to be more well-known than they currently are.

degasus:
Will this message also appears on an NKit image converted to RVZ? e.g. MyGame.NKit.rvz ?

JosJuice:
Yes.

PR #8877, opened 2020-06-16
DolphinQt: Make conversion error messages actually show up

JosJuice:
I don't know why QErrorMessage doesn't work, but ModalMessageBox works, so let's use ModalMessageBox.


Tilka:
the specificity of the second message could probably be improved^^

PR #7396, opened 2018-09-03
Include SRAM in savestates

Techjar:
This fixes an issue with saving after loading a savestate in Pokemon Colosseum. Also seems like a good idea anyways, as SRAM is part of system state after all. Should we add a flag to not write the SRAM back to disk if it was loaded from a savestate?

PR #8879, opened 2020-06-16
Keep memory card header and use it to reinitialize GCI folders on mismatch. (Fix savestate with GCI folders)

AdmiralCurtiss:
Note: The first two commits are also found in #8880. It doesn't really matter which gets merged first.

This is a safer but more complex implementation of #8476. Instead of hardcoding a card ID, we store the memory card header data in EXI_Channel and ensure the same game session (including savestates) always sees the same card header. This is necessary because some (apparently many!) games check this and will refuse to save to the memory card if they believe the card is a different card they initially loaded from.

Whenever we create a new GCI folder memory card, we pass along the header data to the device, which ensures that eg. switching from GCI folder to No Memory Card back to GCI folder while the game is running still results in the same header. Additionally, when loading a savestate, we compare the card header in the savestate to the card header of the current session and will reinitialize all GCI folder cards with the header from the savestate on mismatch.

It does feel a bit silly to store the card header data in EXI_Channel, but it seems to be the best place for it.

Oh, and I'm not 100% sure how many of my security precautions here are *actually* necessary, but I figured it's better to be safe when it comes to save files.

### DIFF
--- a/Source/Core/Core/HW/EXI/EXI_Channel.h
+++ b/Source/Core/Core/HW/EXI/EXI_Channel.h
@@ -6,7 +6,10 @@
 
 #include <array>
 #include <memory>
+
 #include "Common/CommonTypes.h"
+
+#include "Core/HW/GCMemcard/GCMemcard.h"
 
 class PointerWrap;
 
@@ -23,7 +26,7 @@ enum TEXIDevices : int;
 class CEXIChannel
 {
 public:
-  explicit CEXIChannel(u32 channel_id);
+  explicit CEXIChannel(u32 channel_id, const Memcard::HeaderData& memcard_header_data);
   ~CEXIChannel();
 
   // get device
@@ -105,6 +108,12 @@ private:
   u32 m_dma_length = 0;
   UEXI_CONTROL m_control;
   u32 m_imm_data = 0;
+
+  // This data is needed in order to reinitialize a GCI folder memory card when switching between
+  // GCI folder and other devices in the memory card slot or after loading a savestate. Even though
+  // this data is only vaguely related to the EXI_Channel, this seems to be the best place to store
+  // it, as this class creates the CEXIMemoryCard instances.
+  Memcard::HeaderData m_memcard_header_data;
 
   // Devices
   enum
--- a/Source/Core/Core/HW/GCMemcard/GCMemcard.cpp
+++ b/Source/Core/Core/HW/GCMemcard/GCMemcard.cpp
@@ -64,14 +64,16 @@ GCMemcard::GCMemcard()
 {
 }
 
-std::optional<GCMemcard> GCMemcard::Create(std::string filename, u16 size_mbits, bool shift_jis)
+std::optional<GCMemcard> GCMemcard::Create(std::string filename, const CardFlashId& flash_id,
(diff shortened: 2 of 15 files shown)


AdmiralCurtiss:
Oh, and as an extra note, #7396 may or may not be required as well for some games. Hard to say, but couldn't hurt.

Source/Core/Core/HW/EXI/EXI_Channel.h
```diff
@@ -106,6 +109,9 @@ class CEXIChannel
   UEXI_CONTROL m_control;
   u32 m_imm_data = 0;
 
+  // storing this here is a bit silly but we'll try this for now...
```
lioncash, line 112: This should probably have a more informative comment explaining why it's silly and what the ideal scenario would be for where it should be located.

Source/Core/Core/HW/GCMemcard/GCMemcard.cpp
```diff
@@ -1536,30 +1543,62 @@ void GCMBlock::Erase()
...
+{
+  std::memset(this, 0xFF, BLOCK_SIZE);
+  InitializeHeaderData(&m_data, flash_id, size_mbits, shift_jis, rtc_bias, sram_language,
+                       format_time);
+  FixChecksums();
+}
+
+Header::Header(const HeaderData& data)
+{
+  std::memset(this, 0xFF, BLOCK_SIZE);
```
lioncash, line 1600: This class should have a trivially copyable check on it as well.

Source/Core/Core/HW/GCMemcard/GCMemcardBase.h
```diff
@@ -0,0 +1,33 @@
...
+class PointerWrap;
+
+class MemoryCardBase
+{
+public:
+  explicit MemoryCardBase(int card_index = 0, int size_mbits = Memcard::MBIT_SIZE_MEMORY_CARD_2043)
+      : m_card_index(card_index), m_nintendo_card_id(size_mbits)
+  {
+  }
+  virtual ~MemoryCardBase() {}
```
lioncash, line 20: Given this is being migrated, this can be changed to:

```suggestion
  virtual ~MemoryCardBase() = default;
```

JMC47:
This fixes almost every game that complains about the memory card being different on loading savestates.

Stovehead:
This isn't compatible with save states made in the previous version, right?

JMC47:
That's generally how things work.  Savestates are meant to be temporary and the actual saves made more permanent.  This will make it much easier to keep using memory cards though, which is good.

Stovehead:
That sucks, since I'm using a save state instead of saves due to this very issue. Wouldn't there be any way to convert a save state to the next version?

AdmiralCurtiss:
The data required to fix your problem is literally not in your savestate, or at least not at a known location. You'd have to manually find the memory card serial ID in your savestate's game RAM.

Stovehead:
And how would I do that?

JosJuice:
Exactly how to do that is not something I believe anyone here has any significant experience with. All I can say is that you have to reverse engineer the game a bit.

If you want to ask about how to do it, please do so on the forums instead of here, since it's not relevant to the development of this pull request. (Not that I think people there know much about it either.)

PR #8880, opened 2020-06-16
Sketchy code cleanup in GCMemcard and GCMemcardDirectory.

AdmiralCurtiss:
Just some cleanup of undefined behavior and unclear code. Should not change program behavior.

(Yes, ClearBlock() really dir call a member function on the wrong class. I'm amazed this worked at all.)

### DIFF
--- a/Source/Core/Core/HW/GCMemcard/GCMemcard.cpp
+++ b/Source/Core/Core/HW/GCMemcard/GCMemcard.cpp
@@ -1373,16 +1373,17 @@ bool GCMemcard::Format(u8* card_data, const CardFlashId& flash_id, u16 size_mbit
 {
   if (!card_data)
     return false;
-  memset(card_data, 0xFF, BLOCK_SIZE * 3);
-  memset(card_data + BLOCK_SIZE * 3, 0, BLOCK_SIZE * 2);
 
-  *((Header*)card_data) =
-      Header(flash_id, size_mbits, shift_jis, rtc_bias, sram_language, format_time);
+  Header header(flash_id, size_mbits, shift_jis, rtc_bias, sram_language, format_time);
+  Directory dir;
+  BlockAlloc bat(size_mbits);
 
-  *((Directory*)(card_data + BLOCK_SIZE)) = Directory();
-  *((Directory*)(card_data + BLOCK_SIZE * 2)) = Directory();
-  *((BlockAlloc*)(card_data + BLOCK_SIZE * 3)) = BlockAlloc(size_mbits);
-  *((BlockAlloc*)(card_data + BLOCK_SIZE * 4)) = BlockAlloc(size_mbits);
+  std::memcpy(&card_data[BLOCK_SIZE * 0], &header, BLOCK_SIZE);
+  std::memcpy(&card_data[BLOCK_SIZE * 1], &dir, BLOCK_SIZE);
+  std::memcpy(&card_data[BLOCK_SIZE * 2], &dir, BLOCK_SIZE);
+  std::memcpy(&card_data[BLOCK_SIZE * 3], &bat, BLOCK_SIZE);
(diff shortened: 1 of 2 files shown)


Source/Core/Core/HW/GCMemcard/GCMemcard.cpp
```diff
@@ -1367,29 +1369,40 @@ std::optional<std::vector<GCMemcardAnimationFrameRGBA8>> GCMemcard::ReadAnimRGBA
...
     return false;
-  memset(card_data, 0xFF, BLOCK_SIZE * 3);
-  memset(card_data + BLOCK_SIZE * 3, 0, BLOCK_SIZE * 2);
 
-  *((Header*)card_data) = Header(SLOT_A, SizeMb, shift_jis);
+  Header header(flash_id, size_mbits, shift_jis, rtc_bias, sram_language, format_time);
+  Directory dir;
+  BlockAlloc bat(size_mbits);
+
+  static_assert(std::is_trivially_copyable_v<Header>);
```
lioncash, line 1382: These should be placed after the respective class/struct declarations in the header, so that the reader of the class knows that it's a requirement explicitly at the point of reading, rather than changes potentially being made and having it blow up on them in some other area of the code. Ditto for other occurrences.

AdmiralCurtiss: Yeah, I suppose that makes sense. Should I still leave them in the functions as well?

lioncash: Nah, just the one under the class declarations should be sufficient.

AdmiralCurtiss:
Anything else?

Source/Core/Core/HW/GCMemcard/GCMemcard.cpp
```diff
@@ -1374,16 +1374,17 @@ bool GCMemcard::Format(u8* card_data, const CardFlashId& flash_id, u16 size_mbit
...
-  memset(card_data, 0xFF, BLOCK_SIZE * 3);
-  memset(card_data + BLOCK_SIZE * 3, 0, BLOCK_SIZE * 2);
 
-  *((Header*)card_data) =
-      Header(flash_id, size_mbits, shift_jis, rtc_bias, sram_language, format_time);
+  Header header(flash_id, size_mbits, shift_jis, rtc_bias, sram_language, format_time);
+  Directory dir;
+  BlockAlloc bat(size_mbits);
+
+  std::memcpy(&card_data[BLOCK_SIZE * 0], &header, BLOCK_SIZE);
```
iwubcode, line 1382: Shouldn't this be `sizeof(Header)` instead of `BLOCK_SIZE` ?

AdmiralCurtiss: All blocks on memory cards are the same size so it doesn't really matter. I think it's clearer this way due to the multiplication to the left but ymmv.

https://github.com/dolphin-emu/dolphin/blob/master/Source/Core/Core/HW/GCMemcard/GCMemcard.h#L250
https://github.com/dolphin-emu/dolphin/blob/master/Source/Core/Core/HW/GCMemcard/GCMemcard.h#L364
https://github.com/dolphin-emu/dolphin/blob/master/Source/Core/Core/HW/GCMemcard/GCMemcard.h#L399

iwubcode: Ah, I was wondering about this but didn't look close enough.  Thank you!

jordan-woyak:
Can we get a rebase, please?

AdmiralCurtiss:
Oh, did this break? Sure, gimme a sec...

AdmiralCurtiss:
Rebased.

leoetlino (approved):
Looks fine to me too.

PR #8883, opened 2020-06-17
VideoCommon: Clean freelook camera dirty state when getting the new view

iwubcode:
Potentially fixes issue [12157](https://bugs.dolphin-emu.org/issues/12157)


Tilka:
Definitely necessary even if it doesn't fix the slowdown.

PR #8884, opened 2020-06-17
VertexShaderManager: Only look for freelook config changes if we're using freelook

iwubcode:
Another attempt to fix [12157](https://bugs.dolphin-emu.org/issues/12157).  I made the assumption that pulling from our onion config system was relatively fast but that doesn't seem to be the case from testing that fifolog (4fps difference)

### DIFF
--- a/Source/Core/VideoCommon/RenderBase.cpp
+++ b/Source/Core/VideoCommon/RenderBase.cpp
@@ -63,6 +63,7 @@
 #include "VideoCommon/FrameDump.h"
 #include "VideoCommon/FramebufferManager.h"
 #include "VideoCommon/FramebufferShaderGen.h"
+#include "VideoCommon/FreeLookCamera.h"
 #include "VideoCommon/ImageWrite.h"
 #include "VideoCommon/NetPlayChatUI.h"
 #include "VideoCommon/NetPlayGolfUI.h"
@@ -405,6 +406,11 @@ void Renderer::CheckForConfigChanges()
   const bool old_bbox = g_ActiveConfig.bBBoxEnable;
 
   UpdateActiveConfig();
+
+  if (g_ActiveConfig.bFreeLook)
+  {
+    g_freelook_camera.SetControlType(g_ActiveConfig.iFreelookControlType);
+  }
 
   // Update texture cache settings with any changed options.
   g_texture_cache->OnConfigChanged(g_ActiveConfig);
--- a/Source/Core/VideoCommon/VertexShaderManager.cpp
+++ b/Source/Core/VideoCommon/VertexShaderManager.cpp
@@ -449,9 +449,6 @@ void VertexShaderManager::SetConstants()
 
     dirty = true;
   }
-
-  // Handle a potential config change
-  g_freelook_camera.SetControlType(Config::Get(Config::GFX_FREE_LOOK_CONTROL_TYPE));
 }
 
 void VertexShaderManager::InvalidateXFRange(int start, int end)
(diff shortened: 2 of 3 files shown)


iwubcode:
I'd love to hear from other devs on how they handle config changes that can occur at any time?

delroth:
Snapshot at the beginning of the frame? It doesn't make sense to change this setting mid-rendering anyway.

delroth:
Also: `Config::AddConfigChangedCallback` means you don't even have to snapshot, you can update the local version when it gets set.

JosJuice:
> I'd love to hear from other devs on how they handle config changes that can occur at any time?

This function is called on config change and caches the current config values into `g_Config`, which is then copied to `g_ActiveConfig` when a new frame starts: https://github.com/dolphin-emu/dolphin/blob/a63510a55a25c7bd2777a18069de8497bef0d47b/Source/Core/VideoCommon/VideoConfig.cpp#L60

iwubcode:
Thank you both.  This is useful.

I'm going to be moving these settings out of the graphics/video config in #8867 so I was curious how that was working.  Despite having looked at `VideoConfig.cpp` so many times, I didn't even notice that `AddConfigChangedCallback`!

For now though, I will add this setting to `VideoConfig.cpp` so I don't need to query it every time when freelook is enabled.

iwubcode:
@delroth / @JosJuice - updated, let me know what you think

Source/Core/VideoCommon/VertexShaderManager.cpp
```diff
@@ -450,8 +450,11 @@ void VertexShaderManager::SetConstants()
     dirty = true;
   }
 
-  // Handle a potential config change
-  g_freelook_camera.SetControlType(Config::Get(Config::GFX_FREE_LOOK_CONTROL_TYPE));
+  if (g_ActiveConfig.bFreeLook)
```
stenzek, line 453: Any reason this can't go here? https://github.com/dolphin-emu/dolphin/blob/master/Source/Core/VideoCommon/RenderBase.cpp#L396

Would save checking every time constants are updated. Not that it'll make much difference.

iwubcode: Ah, didn't think of this.  I sort of view that function as just a way to recompile shaders when config changes.  Which is why I did make a similar change in #8832 when freelook specific shaders were added.

I see what you're saying.  This is only temporary though, as #8867 is my plan for freelook in the future.  Still, who knows when that'll get merged, so I'll go ahead and move this for now.

iwubcode:
Thanks @stenzek , updated per your suggestion.

Source/Core/VideoCommon/RenderBase.cpp
```diff
@@ -406,6 +407,11 @@ void Renderer::CheckForConfigChanges()
 
   UpdateActiveConfig();
 
+  if (g_ActiveConfig.bFreeLook)
+  {
+    g_freelook_camera.SetControlType(g_ActiveConfig.iFreelookControlType);
```
stenzek, line 412: Does this function _have_ to run every frame? Could you save the old freelook control type/freelook enable and only run it when that changes? e.g.
```cpp
const bool old_freelook = g_ActiveConfig.bFreeLook;
const int old_freelook_control_type = g_ActiveConfig.iFreelookControlType;

if (old_freelook != g_ActiveConfig.bFreeLook || old_freelook_control_type != g_ActiveConfig.iFreelookControlType)
{
  g_freelook_camera.SetControlType(g_ActiveConfig.iFreelookControlType);
}
```

iwubcode: It doesn't _have_ to but what's the point of all this checking?  It doesn't really matter that function is called so often because the actual function already bails if there are no changes.  The whole point is to just get the new value from configuration.

stenzek: Oh, I didn't see the check here: https://github.com/dolphin-emu/dolphin/blob/master/Source/Core/VideoCommon/FreeLookCamera.cpp#L179

I misremembered it unconditionally reallocating it. In that case, it's fine as-is.

PR #8885, opened 2020-06-18
PowerPC: partially implement thermal related SPRs

delroth:
Doesn't support triggering interrupts when the thermal threshold is
exceeded, but allows polling for temperature information.

The THRM[123] registers are documented in most PPC datasheets, see e.g.
this PPC750CX one: http://datasheets.chipdb.org/IBM/PowerPC/750/750cx_um3-17-05.pdf

Fixes https://bugs.dolphin-emu.org/issues/12160 . See the nice 42°C display on Swiss in this screenshot:

### DIFF
--- a/Source/Core/Core/PowerPC/Interpreter/Interpreter_SystemRegisters.cpp
+++ b/Source/Core/Core/PowerPC/Interpreter/Interpreter_SystemRegisters.cpp
@@ -459,6 +459,37 @@ void Interpreter::mtspr(UGeckoInstruction inst)
       PowerPC::IBATUpdated();
     }
     break;
+
+  case SPR_THRM1:
+  case SPR_THRM2:
+  case SPR_THRM3:
+  {
+    // We update both THRM1 and THRM2 when either of the 3 thermal control
+    // registers are updated. THRM1 and THRM2 are independent, but THRM3 has
+    // settings that impact both.
+    //
+    // TODO: Support thermal interrupts when enabled.
+    constexpr u32 SIMULATED_TEMP = 42;  // °C
+
+    auto UpdateThermalReg = [](UReg_THRM12* reg) {
+      if (!THRM3.E || !reg->V)
+      {
+        reg->TIV = 0;
+      }
+      else
+      {
+        reg->TIV = 1;
+        if (reg->TID)
+          reg->TIN = SIMULATED_TEMP < reg->THRESHOLD;
+        else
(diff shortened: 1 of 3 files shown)


JMC47:
👍

Source/Core/Core/PowerPC/Interpreter/Interpreter_SystemRegisters.cpp
```diff
@@ -459,6 +459,37 @@ void Interpreter::mtspr(UGeckoInstruction inst)
...
+    auto UpdateThermalReg = [](UReg_THRM12* reg) {
+      if (!THRM3.E || !reg->V)
+      {
+        reg->TIV = 0;
+      }
+      else
+      {
+        reg->TIV = 1;
+        if (reg->TID)
+          reg->TIN = SIMULATED_TEMP < reg->THRESHOLD;
```
hthh, line 483: I just tried building on Windows 10 for the first time today, so sorry if I'm doing something silly, but the only error I got was `'SIMULATED_TEMP' cannot be implicitly captured because no default capture mode has been specified`. Making it a `#define` or adding an `&` between the square brackets seems to fix things?

JosJuice: I'm assuming you're building with CMake? For some reason this kind of thing fails, but only when building with CMake on Windows, not when building with the VS solution on Windows or with CMake on any other platform.

hthh: Interesting, yeah I think so (although tbh I'm not 100% sure, I just followed https://github.com/dolphin-emu/dolphin/wiki/Building-for-Windows as closely as possible). Is dolphin-emu.sln preferred? Would it be better to look for a fix to whatever's the matter with CMake on Windows than fix this kind of thing as it shows up?

JosJuice: > Is dolphin-emu.sln preferred?

There's no real reason it should be better, but since the buildbot uses it rather than CMake for Windows, the Windows CMake build does break sometimes.

> Would it be better to look for a fix to whatever's the matter with CMake on Windows than fix this kind of thing as it shows up?

If we found a fix for it, that would be nice, but I have no idea where to start regarding that.

PR #8886, opened 2020-06-18
PatchEngine: Attempt to fix crash in IsStackSane

JosJuice:
HostIsInstructionRAMAddress uses XCheckTLBFlag::OpcodeNoException, so we should also use XCheckTLBFlag::OpcodeNoException when reading, to ensure that we use the IBAT (as opposed to the DBAT) for both.

I made this change in an attempt to fix a crash a user was experiencing at PatchEngine.cpp:234, but it seems like this change didn't actually fix that crash. I'm submitting it anyway because the mismatched behavior we have right now seems a little odd.


AdmiralCurtiss:
Makes sense to me, though I'm kinda surprised this hasn't been a problem before. I guess the two BATs are rarely different? And `HostRead_Instruction()` is not used in a lot of places either.

PR #8887, opened 2020-06-19
DolphinQt: Fix default game list sort order

JosJuice:
A first and Z last is more natural. (I thought ascending meant A first and Z last, but apparently not...)


BhaaLseN:
This seems wrong according to [the documentation](https://doc.qt.io/qt-5/qt.html#SortOrder-enum) (emphasis mine):

> | Constant | Value | Description |
> | -- | -- | -- |
> | `Qt::AscendingOrder` | `0` | **The items are sorted ascending e.g. starts with 'AAA' ends with 'ZZZ' in Latin-1 locales** |
> | `Qt::DescendingOrder` | `1` | The items are sorted descending e.g. starts with 'ZZZ' ends with 'AAA' in Latin-1 locales |

JosJuice:
That documentation does match up with what I was expecting... But it doesn't seem to match what actually happens in Dolphin. I'm not sure why.

jordan-woyak:
Closing in favor of #8958. :3

PR #8891, opened 2020-06-21
Show file format details in game properties

JosJuice:
image
Screenshot_20200621-203741


JMC47:
Threw a bunch of games in a bunch of formats at it and it worked fine, including a converted tgc file that shouldn't exist.

PR #8893, opened 2020-06-21
Android: Remove unnecessary JNI function declarations

JosJuice:
We generally have no reason to call these functions on our own, so there's not much reason to declare them, especially not in the cpp file where they're defined. In case we ever do get a reason to do it, we can add declarations for just the functions that need them.

### DIFF
--- a/Source/Android/jni/GameList/GameFile.cpp
+++ b/Source/Android/jni/GameList/GameFile.cpp
@@ -42,51 +42,6 @@ extern "C" {
 #endif
 
 JNIEXPORT void JNICALL Java_org_dolphinemu_dolphinemu_model_GameFile_finalize(JNIEnv* env,
-                                                                              jobject obj);
-JNIEXPORT jint JNICALL Java_org_dolphinemu_dolphinemu_model_GameFile_getPlatform(JNIEnv* env,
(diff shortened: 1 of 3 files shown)


lioncash:
Wouldn't this cause -Wmissing-declarations warnings?

JosJuice:
I don't get any such warnings in the log when building this change.

Ebola16:
I was wondering about this too. From what I've read, forward declarations help reduce building time if a method has the same name and different parameters. `NativeLibrary.run()` seems to fall in this category but I'm not sure if the involvement of the JNI makes a difference here. As long as both variants of `NativeLibrary.run()` work as expected, build time isn't significantly slower, and the relevant object files files aren't bloated, this PR seems fine. Unfortunately I don't know what would qualifly as "significantly slower".

JosJuice:
I don't see why adding a forward declaration would improve the build time. A common thing to do with forward declarations is to replace an include with a forward declaration, which does save build time by skipping the include, but includes are not relevant for what we are doing here.

`NativeLibrary.run` is only overloaded from Java's perspective. From C++'s perspective, the two variants of it have different names. (I believe JNI was designed this way for compatibility with C, which does not support function overloading.)

PR #8894, opened 2020-06-22
[Feedback Wanted] Android: Wii Overlay Controller Overhaul

Ebola16:
Fixes https://bugs.dolphin-emu.org/issues/10818 (along with #8739 and #8735).
Fixes https://bugs.dolphin-emu.org/issues/11371.

Notable changes:
- Added an Android-exclusive Interface setting "Disable Unselected Wii Overlay Controllers", disabled by default. When enabled, disables overlay GameCube controller when using Wii Remote and vice versa. This is inconvenient when switching between GameCube and Wii games, as it can potentially disable the GameCube controller. Unimplemented with custom game settings. However, this setting will handle cases when a user wants to switch between a GameCube and Wii controller during Wii emulation.
- Does not fix unrelated controller selection bugs with custom game settings. Headers were added to point this out and fixes should be made in a separate PR.
- Wii Remote orientation has been added as an option when selecting the "None" Wii Remote extension.
- Real controllers are now selectable via Wii Overlay controllers.
- GameCube Adapter scanning thread will be stopped if no controllers are using it during emulation.
- Since it's not possible to know if a user has disabled a controller to prevent it from occupying a controller slot, set the Wii overlay to None when disabling Wii Input controllers and see if another emulated controller is enabled when starting Wii emulation.
- Enable the Wii Remote scanning thread setting when choosing a real Wii Remote.
- Converted `wiiController` to an INI setting. This allows it to be retained across Dolphin builds and is part of the prep work for using Wii overlays on a per-game basis.
- Read the Wii overlay controller if WiimoteNew.ini is overwritten. This allows controllers to be retained across Dolphin builds.
- See TODOs for work that will be deferred to future PRs.

Potential bugs:
- [x] With "Disable Unselected Wii Overlay Controllers" enabled, switching the Wii Overlay controller from GameCube to any Wii Remote sets that controller as Player 2 instead of player 1 in Super Smash Bros. Brawl's character select screen. The previous player 1 GameCube controller is disabled as intended. Oddly, making this controller switch outside of Brawl's character select screen works as intended.

- [x] Fix detection of Wii Remote Orientation with custom GameSettings.

- [x] I can't test real Wii Remotes as I don't have a DolphinBar. We should probably verify that enabling real Wii Remotes works both during and outside of emulation. (Tested by @Sanfe)


Ebola16:
Real controllers are broken at the moment, working on a fix.

Ebola16:
Real controllers need to be added to the Wii overlay controller options in order to be affected by "Disable Unselected Wii Overlay Controllers". This will require reordering the Wii overlay controller array, which is used in may places in Dolphin. For now, it is best to update the description of "Disable Unselected Wii Overlay Controllers" and add a TODO for a future PR.

Also, deduplicated some of my code.

Ebola16:
Pretty sure I fixed the 1st mentioned issue by adding a half second delay when running `NativeLibrary::ReloadWiimoteConfig()` in `EmulationActivity.chooseController()`. Originally I thought this was a race condition between `NativeLibrary.SetConfig()` and `NativeLibrary::ReloadWiimoteConfig()` but making the latter wait if the former didn't complete had no effect.

Ebola16:
https://bugs.dolphin-emu.org/issues/11371 (Add Wii Remote orientation setting) depends on, and modifies, this PR so it's probably best to include it here. I converted the orientation setting to `int` so it can be better displayed in Wii Remote settings.

Ebola16:
This is an ocean of edge cases!

Ebola16:
The latest commit addresses another issue with per-game settings and modifies this PR. Changing the Wii overlay controller with custom game settings doesn't work but it didn't work on master either so I'll probably look into that in a future PR.

JMC47:
I'll be honest, I don't use Dolphin on Android as much as I should, so I'm not as familiar with some of these problems as much as I should be.  I tested out this PR on my SD835 and mostly used your Pull Request description as my testing bed.  I don't know if I tested everything correctly.

All I can say is that the Brawl issues make sense with how the game detects controllers.  For other games it seems to work as intended no matter what, like Fire Emblem: Radiant Dawn.

Everything else I don't have a huge opinion on, just running games, swapping controllers and such appeared to work.  Oddly enough, this changes Sonic Colors to work with Wii Remotes since it will **only** accept Wii Remote input if all GC controllers are unplugged.

Ebola16:
I appreciate the testing! Before this PR, changing the Wii overlay controller would not enable the controller that was selected, which was the initial problem I wanted to fix.

It also sounds like Sonic Colors is a good use case for the "Disable Unselected Wii Overlay Controllers" setting. I didn't realize that until your tests.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/activities/EmulationActivity.java
```diff
@@ -915,6 +923,63 @@ private void chooseController()
...
+      {
+        NativeLibrary
+                .SetConfig(SettingsFile.FILE_NAME_DOLPHIN + ".ini",
+                        Settings.SECTION_INI_CORE,
+                        SettingsFile.KEY_GCPAD_PLAYER_1,
+                        InputOverlay.EMULATED_GAMECUBE_CONTROLLER);
+
+        NativeLibrary
+                .SetOverlaySIDevice(
+                        Integer.parseInt(InputOverlay.EMULATED_GAMECUBE_CONTROLLER));
```
JosJuice, line 973: This needs error handling.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/activities/EmulationActivity.java
```diff
@@ -915,6 +923,63 @@ private void chooseController()
...
+                                InputOverlay.DISABLED_GAMECUBE_CONTROLLER);
+
+        NativeLibrary.SetOverlaySIDevice(selection == InputOverlay.OVERLAY_GAMECUBE ?
+                Integer.parseInt(InputOverlay.EMULATED_GAMECUBE_CONTROLLER) :
+                Integer.parseInt(InputOverlay.DISABLED_GAMECUBE_CONTROLLER));
+      }
+
+      if (!NativeLibrary
+              .GetConfig(SettingsFile.FILE_NAME_WIIMOTE + ".ini", SettingsFile.KEY_WIIMOTE_PLAYER_1,
+                      SettingsFile.KEY_WIIMOTE_TYPE, "1").equals("2"))
```
JosJuice, line 953: Any particular reason why "1" and "2" (and equivalents in the code below) aren't using constants?

Source/Android/jni/MainAndroid.cpp
```diff
@@ -29,6 +29,8 @@
 #include "Common/Version.h"
 #include "Common/WindowSystemInfo.h"
 
+#include <Core/HW/SI/SI.h>
+#include <Core/HW/SI/SI_Device.h>
```
JosJuice, line 33: Includes of Dolphin code need to use `""`, not `<>`.

Source/Android/jni/MainAndroid.cpp
```diff
@@ -604,6 +606,31 @@ JNIEXPORT void JNICALL Java_org_dolphinemu_dolphinemu_NativeLibrary_ReloadConfig
...
 }
 
+JNIEXPORT void JNICALL Java_org_dolphinemu_dolphinemu_NativeLibrary_SetOverlaySIDevice(
+    JNIEnv* env, jobject obj, jint gcController)
+{
+  SerialInterface::SIDevices SIDevice;
+
+  SConfig::GetInstance().LoadSIDevice0Setting();
+
+  switch (gcController)
```
JosJuice, line 616: You can skip this entire switch statement with `si_device = static_cast<SerialInterface::SIDevices>(gcController)`.

Source/Android/jni/MainAndroid.cpp
```diff
@@ -604,6 +606,31 @@ JNIEXPORT void JNICALL Java_org_dolphinemu_dolphinemu_NativeLibrary_ReloadConfig
   SConfig::GetInstance().LoadSettings();
 }
 
+JNIEXPORT void JNICALL Java_org_dolphinemu_dolphinemu_NativeLibrary_SetOverlaySIDevice(
+    JNIEnv* env, jobject obj, jint gcController)
+{
+  SerialInterface::SIDevices SIDevice;
```
JosJuice, line 612:
```suggestion
  SerialInterface::SIDevices si_device;
```

We don't use camelCase variable naming in C++. (Well, at least not in newly written code... There is a bunch of legacy code still.)

Source/Core/Core/ConfigManager.cpp
```diff
@@ -639,6 +639,17 @@ void SConfig::LoadJitDebugSettings(IniFile& ini)
...
 
+void SConfig::LoadSIDevice0Setting()
+{
+  Config::Load();
+  IniFile ini;
+  ini.Load(File::GetUserPath(F_DOLPHINCONFIG_IDX));
+  IniFile::Section* core = ini.GetOrCreateSection("Core");
+
+  core->Get("SIDevice0", &m_SIDevice[0],
+            0 ? SerialInterface::SIDEVICE_NONE : SerialInterface::SIDEVICE_GC_CONTROLLER);
```
JosJuice, line 650: Is there any good reason why this should load from the INI instead of using a value passed in as a parameter? What's especially strange is that this is called from a function which does receive the right value as a parameter as far as I can tell.

Also, `0 ? SerialInterface::SIDEVICE_NONE : SerialInterface::SIDEVICE_GC_CONTROLLER` is just a more complicated way of writing `SerialInterface::SIDEVICE_GC_CONTROLLER`.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/activities/EmulationActivity.java
```diff
@@ -906,7 +906,7 @@ private void chooseController()
                       getResources().getStringArray(R.array.controllersValues)[indexSelected]);
               NativeLibrary.SetConfig(SettingsFile.FILE_NAME_WIIMOTE + ".ini",
                       SettingsFile.KEY_WIIMOTE_PLAYER_1, SettingsFile.KEY_WIIMOTE_ORIENTATION,
-                      indexSelected == InputOverlay.OVERLAY_WIIMOTE_SIDEWAYS ? "True" : "False");
+                      indexSelected == InputOverlay.OVERLAY_WIIMOTE_HORIZONTAL ? "1" : "0");
```
JosJuice, line 909: Why is this using 1 and 0 instead of True and False now? I'm assuming there's some reason, but it isn't obvious to me.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/activities/EmulationActivity.java
Ebola16: Please check to make sure I updated this properly.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/activities/EmulationActivity.java
Ebola16: There are multiple instances of these values needing constants in code unrelated to this PR. If we're assigning constants to these values, I'd rather do it in a separate PR where all of those cases are handled.

Source/Core/Core/ConfigManager.cpp
Ebola16: INI changes to `SIDevice0` are made before `NativeLibrary.SetOverlaySIDevice` is called. Removing `SConfig::GetInstance().LoadSIDevice0Setting()` caused Dolphin to fail to disable the overlay GC controller with "Disable Unselected Wii Overlay Controllers" enabled. I suspect`SerialInterface::ChangeDevice` will eventually read `SIDevice0`'s value, which means it needs to be reloaded beforehand.

And I used your `SerialInterface::SIDEVICE_GC_CONTROLLER` suggestion.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/activities/EmulationActivity.java
Ebola16: The "Orientation" setting looks nicer as a `SingleChoiceSetting` (int values), rather than a `CheckBoxSetting` (boolean value) with a description to explain default orientation. It seems silly to convert `int` -> `boolean` -> `int`, so I think it's best jut to change the values of this setting to `int`.

Ebola16:
With the most recent commits, I added:
- Rename `KEY_WIIMOTE_ORIENTATION`'s value to "Options/Wiimote Orientation" since this setting is now `int`
- Additional error handling in `SingleChoiceSetting`
- Found a bug with detection of Wii Remote Orientation in custom GameSettings. I'll look into it.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/activities/EmulationActivity.java
JosJuice: But it's still a bool in the native code. If the native code re-saves the INI, it can replace the 1 and 0 with True and False, and the updated Android code can't parse that correctly.

JosJuice: Actually, now that you've renamed the setting, I suppose this doesn't correspond to any native setting at all anymore. So if you set the overlay controls to Sideways Wii Remote, the d-pad and motion controls will be oriented incorrectly.

Source/Core/Core/ConfigManager.cpp
JosJuice: Yes, I understand that you have to call `LoadSIDevice0Setting`, but I don't understand why `LoadSIDevice0Setting` has to load data from the INI file instead of receiving that data as a parameter.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/activities/EmulationActivity.java
Ebola16: Alright, do you think it would be best to keep `KEY_WIIMOTE_ORIENTATION`'s bool value and implement a `int` -> `boolean` -> `int` conversion so `SingleChoiceSetting` can be used over `CheckBoxSetting` for better UX design?

JosJuice: I would be fine with either that or making it a checkbox.

Ebola16: Reverted the `KEY_WIIMOTE_ORIENTATION` rename and made it 'boolean' again. I also moved "Catch getSetting ClassCastExceptions in ItemViews" to its own comment in case INI settings come across invalid values.

Still working on the rest of the review comments and the bug I found.

Ebola16:
With the most recent commits:
- "Catch getSetting Exceptions in ItemViews" now catches all exceptions.
- Fixed detection of Wii Remote Orientation with custom GameSettings (see "Android: Use custom Wii overlay controller with custom GameSettings" commit).

Source/Core/Core/ConfigManager.cpp
Ebola16: So I should pass in `gcController` and change `core->Get` to `core->Set`?

Ebola16: For some reason I didn't initially think of `core->Set`. Since we're setting one specific setting, let's just do that in this method.

Ebola16:
First round of review comments addressed. I also added a TODO to make "Disable Unselected Wii Overlay Controllers" setting work with custom GameSettings. This PR is already complicated enough so it's probably better addressed in a future PR.

Ebola16:
Now that I think about it there's no reason this won't work on Android TV so I'm removing the checkbox for that.

Source/Android/jni/MainAndroid.cpp
```diff
@@ -622,6 +623,18 @@ JNIEXPORT void JNICALL Java_org_dolphinemu_dolphinemu_NativeLibrary_ReloadConfig
...
 }
 
+JNIEXPORT void JNICALL Java_org_dolphinemu_dolphinemu_NativeLibrary_SetOverlaySIDevice(
+    JNIEnv* env, jobject obj, jint gcController)
+{
+  IniFile ini;
+  ini.Load(File::GetUserPath(F_DOLPHINCONFIG_IDX));  // load first to not kill unknown stuff
+  IniFile::Section* core = ini.GetOrCreateSection("Core");
+
+  core->Set("SIDevice0", gcController);
```
JosJuice, line 633: This doesn't actually do anything, since you're not saving the IniFile afterwards.

Ebola16: Fixed. Previously, I made the INI changes by calling `NativeLibrary.SetConfig` before `NativeLibrary.SetOverlaySIDevice`, making this dead code. Now I save the INI in `NativeLibrary.SetOverlaySIDevice` and removed the unnecessary `NativeLibrary.SetConfig` calls.

Ebola16:
Improved readability of `EmulationActivity.handleWiiOverlayDisablesControllersSetting`

Ebola16:
Made two updates to the last commit:
- Converted `hasCustomGameProfile` in `InputOverlay` to a local variable.
- Moved the initial `InputOverlay.refreshControls` to `EmulationFragment.runWithValidSurface`. Previously, GameCube controllers failed to be disabled by the "Disable Unselected Wii Overlay Controllers" setting via the initial `InputOverlay.refreshControls`. I'm not 100% sure that my fix is proper but it seems to work.

Ebola16:
Simplified `EmulationActivity.chooseController` changes. I also temporary disabled this feature with custom game settings. I didn't work properly on master and I'd rather implement it in a separate PR.

Ebola16:
Moved some of the cleanups to #9042 and improved `InputOverlay.refreshControls` again.

Ebola16:
Latest commit:
- Real controllers are now selectable via Wii Overlay controllers.
- Took care of a bug unrelated to this PR - GameCube Adapter selection will start the scanning thread. We also turn it off if no controllers are using it.

I can't test real Wii Remotes as I don't have a DolphinBar. We should probably verify that enabling real Wii Remotes works both during and outside of emulation.

Ebola16:
Update to latest commit should handle stopping the GameCube Adapter scanning thread when not in use.

Ebola16:
A lot of PRs that heavily modify this PR were / are about to be merged so this should be a draft.

Sanfe:
> Latest commit:
> 
> * Real controllers are now selectable via Wii Overlay controllers.
> * Split to #9052 - GameCube Adapter selection will now start the scanning thread.
> 
> I can't test real Wii Remotes as I don't have a DolphinBar. We should probably verify that enabling real Wii Remotes works both during and outside of emulation.

I could test because i have one and usually play in Android. One problem usually relationated with this, is that the real wiimotes have a 5 min connection limit. If you let the emulator open without touching anything and return in 5 minutes, the software disconnects the wiimotes and the only solution is save, return to menu and open again.

Ebola16:
Large parts of this PR were rewritten since the hacks I used are no longer necessary thanks to other PRs.
Notable changes:
- Does not fix unrelated controller selection bugs with custom game settings. Headers were added to point this out and fixes should be made in a separate PR.
- Took care of the magic I originally planned on deferring to another PR.

@JosJuice can you take a look to make sure I'm not duplicating logic from the C++ parser and new config system PRs?

@Sanfe Thank you for testing! Just to make sure, does the issue you describe also occur on the latest dev build? If so, that's something that should be submitted to https://bugs.dolphin-emu.org/projects/emulator/issues so it can be handled in a separate PR.

Ebola16:
The "Disable Unselected Wii Overlay Controllers" setting probably makes more sense as an Interface setting so I moved it there.

Sanfe:
> @Sanfe Thank you for testing! Can you reproduce the issue you described on the latest version of this PR? Also, does the issue you describe also occur on the latest dev build? If so, that's something that should be submitted to https://bugs.dolphin-emu.org/projects/emulator/issues so it can be handled in a separate PR.

Yes, I'm testing latest beta. This problem comes from large time. Thanks. i will open a new ticket. I need to add the wii remote log or the wii ipc log?

Ebola16:
> Yes, I'm testing latest beta. This problem comes from large time. Thanks. i will open a new ticket. I need to add the wii remote log or the wii ipc log?

I have a feeling the problem doesn't relate to Wii IPC but the logs you mention would be helpful when submitting the issue report.

Since you said you tested a real Wii Remote with Dolphin on Android, did you notice any problems that aren't present in the latest dev version?

JosJuice (commented):
Still not a complete review due to the complexity of the PR, but here are some additional comments.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/ui/SettingsAdapter.java
```diff
@@ -162,9 +166,32 @@ public void setSettings(ArrayList<SettingsItem> settings)
...
   public void onBooleanClick(CheckBoxSetting item, int position, boolean checked)
   {
+    SharedPreferences preferences = PreferenceManager.getDefaultSharedPreferences(mContext);
+    final SharedPreferences.Editor editor = preferences.edit();
+
+    boolean wiimoteOrientation =
+            item.getSetting().getSection().equals(SettingsFile.KEY_WIIMOTE_PLAYER_1) &&
+                    item.getSetting().getKey().equals(SettingsFile.KEY_WIIMOTE_ORIENTATION);
+    boolean customGameSettingsWiimoteOrientation =
+            item.getSetting().getKey().equals(SettingsFile.KEY_WIIMOTE_ORIENTATION + 0);
```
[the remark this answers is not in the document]
Ebola16: Should be fixed but worth another check since I haven't written much code for the new config system.

Source/Android/app/src/main/res/values/strings.xml
```diff
@@ -431,13 +432,18 @@ It can efficiently compress both junk data and encrypted Wii data.
 
     <string name="select_dir">Select This Directory</string>
 
+    <!-- Android Exclusive Settings-->
+    <string name="wii_overlay_disables_controllers">Disable Unselected Wii Overlay Controllers</string>
+    <string name="wii_overlay_disables_controllers_description">Disable overlay GameCube controller when using Wii Remote and vice versa. Inconvenient if switching between GameCube and Wii games. Unimplemented with custom game settings.</string>
```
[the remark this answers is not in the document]
Ebola16: I think it would be better to keep that sentence and one of the TODOs mentions modifying this string so we shouldn't forget. Users wouldn't understand why the setting was hidden and a hidden setting doesn't necessarily mean a non-functional setting. I'd rather add a sentence to prevent confusion.

Source/Android/app/src/main/res/values/strings.xml
```diff
@@ -431,13 +432,18 @@ It can efficiently compress both junk data and encrypted Wii data.
...
+
     <!-- Misc -->
     <string name="pitch">Total Pitch</string>
     <string name="yaw">Total Yaw</string>
     <string name="vertical_offset">Vertical Offset</string>
     <string name="slider_setting_value">%1$d%2$s</string>
     <string name="disc_number">Disc %1$d</string>
-    <string name="disabled_gc_overlay_notice">GameCube Controller 1 is set to \"None\"</string>
+    <string name="overlay_controller_disabled">Controller 1 is disabled</string>
+    <string name="custom_game_settings_wip">Values can be set but not properly read yet.</string>
```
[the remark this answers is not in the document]
Ebola16: I'd like to prevent this PR from becoming even more complex so I think this string should stay.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/model/BooleanSetting.java
```diff
@@ -102,7 +102,12 @@
           "JitSystemRegistersOff", false),
   MAIN_JIT_BRANCH_OFF(Settings.FILE_DOLPHIN, Settings.SECTION_DEBUG, "JitBranchOff", false),
   MAIN_JIT_REGISTER_CACHE_OFF(Settings.FILE_DOLPHIN, Settings.SECTION_DEBUG, "JitRegisterCacheOff",
-          false);
+          false),
+
+  // Android exclusive settings
```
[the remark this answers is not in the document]
Ebola16: Should be fixed but worth another check since I haven't written much code for the new config system.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/activities/EmulationActivity.java
```diff
@@ -911,6 +928,80 @@ private void chooseController()
...
+  // TODO: Make this setting work with custom GameSettings and update setting description.
+  public static void handleWiiOverlayDisablesControllersSetting(int selection, boolean IsFirstRun)
+  {
+    File dolphinFile = SettingsFile.getSettingsFile(Settings.FILE_DOLPHIN);
+    File wiimoteNewFile = SettingsFile.getSettingsFile(Settings.FILE_WIIMOTE);
+    IniFile dolphinIni = new IniFile(dolphinFile);
+    IniFile wiimoteNewIni = new IniFile(wiimoteNewFile);
+
+    boolean wiiOverlayDisablesControllers = dolphinIni.getBoolean(Settings.SECTION_INI_ANDROID,
+            SettingsFile.KEY_WII_OVERLAY_DISABLES_CONTROLLERS, false);
```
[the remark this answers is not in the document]
Ebola16: This is a `static` method so is the change I made acceptable?

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/ui/SettingsAdapter.java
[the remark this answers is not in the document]
JosJuice: Why is the orientation setting reading from `MAIN_WII_OVERLAY_DISABLES_CONTROLLERS` in the new code? I was assuming you would want to do something with `SettingsFile.KEY_WIIMOTE_ORIENTATION` instead.

Also, the implementation of `setBoolean` is only writing to the SharedPreferences setting and not to any INI file setting, unlike the old code.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/activities/EmulationActivity.java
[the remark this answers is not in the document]
JosJuice: It'll do for now. For what it's worth, the first commit of PR #9146 should make it easy to fix this cleanly. Let me know if you want me to split out that commit to a separate PR.

Also, since you've made this change, I think `MAIN_WII_OVERLAY_DISABLES_CONTROLLERS` works properly with per-game settings now? I'm not entirely sure, but it's worth testing.

Sanfe:
> > Yes, I'm testing latest beta. This problem comes from large time. Thanks. i will open a new ticket. I need to add the wii remote log or the wii ipc log?
> 
> I have a feeling the problem doesn't relate to Wii IPC but the logs you mention would be helpful when submitting the issue report.
> 
> Since you said you tested a real Wii Remote with Dolphin on Android, did you notice any problems that aren't present in the latest dev version?

No. The 5 min idlle disconnect are in all versions and flavours of dolphin.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/activities/EmulationActivity.java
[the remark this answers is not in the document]
Ebola16: `MAIN_WII_OVERLAY_DISABLES_CONTROLLERS` itself works with custom game settings but its usage does not yet. `EmulationActivity.handleWiiOverlayDisablesControllersSetting` doesn't modify custom game settings yet. That's a rather complex changed and it'd also be good to isolate Wii Overlay changes when using custom game settings, which should be done in another PR.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/ui/SettingsAdapter.java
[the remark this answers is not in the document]
Ebola16: Oops, I previously picked the wrong setting. This time I picked the right one but was this what you had in mind? It still uses hacky code but is at least contained in one place.

And I'll look into my changes more later but `setBoolean` changes are reflected in the INI with what I've currently written, possibly due to code that follows it. My changes to `SettingsFile.saveCustomGameSettings` no longer trigger but I appear to still get the result I want.

Ebola16:
> No. The 5 min idlle disconnect are in all versions and flavours of dolphin.

Alright, if anything went wrong with real Wii Remotes due to this PR, I assume it'd be very obvious (like failing to connect). I'll mark this as tested with real Wii Remotes. Thanks again!

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/ui/SettingsAdapter.java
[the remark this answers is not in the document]
JosJuice: This looks more like what I had in mind, yes. I'll leave further comments on the code directly instead of here.

> And I'll look into my changes more later but `setBoolean` changes are reflected in the INI with what I've currently written, possibly due to code that follows it.

I don't see how that is a problem. Did you miss a "not"?

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/ui/SettingsFragmentPresenter.java
```diff
@@ -871,7 +895,99 @@ private void addExtensionTypeSettings(ArrayList<SettingsItem> sl, int wiimoteNum
...
+                      .getSection(Settings.FILE_WIIMOTE,
+                              Settings.SECTION_WIIMOTE + wiimoteNumberAdjusted)
+                      .getBoolean(SettingsFile.KEY_WIIMOTE_ORIENTATION, false);
+            }
+          }
+
+          @Override
+          public boolean isRuntimeEditable()
+          {
+            return true;
```
JosJuice, line 972: I don't believe this actually works correctly if it's edited at runtime, as it's using the old settings system. So this should return `false`.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/ui/SettingsFragmentPresenter.java
```diff
@@ -871,7 +895,99 @@ private void addExtensionTypeSettings(ArrayList<SettingsItem> sl, int wiimoteNum
...
+          mSettings.loadWiimoteProfile(mGameID, wiimoteNumberAdjusted);
+        }
+        else
+        {
+          wiimoteNumberAdjusted = wiimoteNumber - 3;
+        }
+
+        AbstractBooleanSetting wiimoteOrientation = new AbstractBooleanSetting()
+        {
+          Boolean valueSelected;
```
JosJuice, line 913: I'm assuming this is simply some form of caching? This code is not performance critical, and fetching the value of a setting is relatively fast, so I don't really see the need for this. The code would be simpler if you removed this.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/ui/SettingsFragmentPresenter.java
```diff
@@ -871,7 +895,99 @@ private void addExtensionTypeSettings(ArrayList<SettingsItem> sl, int wiimoteNum
...
+            editor.putInt("wiiController", newValue ? InputOverlay.OVERLAY_WIIMOTE_HORIZONTAL :
+                    InputOverlay.OVERLAY_WIIMOTE_VERTICAL);
+            editor.commit();
+
+            getWiimoteSection(wiimoteNumberAdjusted, false).setBoolean(settings.isGameSpecific() ?
+                    SettingsFile.KEY_WIIMOTE_ORIENTATION + wiimoteNumberAdjusted :
+                    SettingsFile.KEY_WIIMOTE_ORIENTATION, newValue);
+
+            valueSelected = newValue;
+            mView.onSettingChanged();
```
JosJuice, line 946: There's no need to call `onSettingChanged` here and in `delete`. The calling code will call it.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/ui/SettingsFragmentPresenter.java
```diff
@@ -871,7 +895,99 @@ private void addExtensionTypeSettings(ArrayList<SettingsItem> sl, int wiimoteNum
...
+
+          @Override
+          public boolean isOverridden(Settings settings)
+          {
+            if (valueSelected != null)
+            {
+              return settings.isGameSpecific() && valueSelected != settings
+                      .getSection(Settings.FILE_WIIMOTE,
+                              Settings.SECTION_WIIMOTE + wiimoteNumberAdjusted)
+                      .getBoolean(SettingsFile.KEY_WIIMOTE_ORIENTATION, false);
```
JosJuice, line 957: This should use `exists` instead of `getBoolean`, for consistency with how other settings work. Reference: https://github.com/dolphin-emu/dolphin/blob/a7b9e6857b6c4e172a4ec333a6c83d624092cdf3/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/model/AbstractLegacySetting.java#L19

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/ui/SettingsAdapter.java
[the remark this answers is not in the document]
Ebola16: Oh good, because that was a pain to write lol.
And I was confused by
> Also, the implementation of setBoolean is only writing to the SharedPreferences setting and not to any INI file setting, unlike the old code.

...but I used `setBoolean` while properly saving the values to the INIs. There isn't a problem with the current code as far as I can tell.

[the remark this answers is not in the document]
JosJuice: I think I'm a little confused now... But if the current code is working, it's all in order.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/ui/SettingsFragmentPresenter.java
Ebola16: No. Initially we want to retrieve the boolean value from the INI for `getBoolean`. Once modified via `setBoolean`, the change is not stored in the INI until saving, so further calls to `getBoolean` wouldn't return the updated value. My workaround was to add a variable to make the up-to-date boolean value accessible to all overrides, and read from the INI again only if the setting is deleted.

JosJuice: Unless there's something I'm missing, `setBoolean` saves to the in-memory `IniFile` object and `getBoolean` reads from the in-memory `IniFile` object, so removing the variable should work.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/ui/SettingsFragmentPresenter.java
Ebola16: The problem with that is the custom controller profile has an orientation setting when created, unlike other custom game settings in the GameSettings folder.

JosJuice: But that means that the setting actually is overridden, no? So it should show up as overridden.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/ui/SettingsFragmentPresenter.java
Ebola16: I changed `getBoolean` to:
```
@Override
public boolean getBoolean(Settings settings)
{
  return getWiimoteSection(wiimoteNumberAdjusted, true)
          .getBoolean(SettingsFile.KEY_WIIMOTE_ORIENTATION, false);
}
```

When changing the orientation with custom game settings, this value doesn't update after changing the value via `setBoolean`. Exiting to the main screen and then returning to the orientation setting displays the updated value. I'm assuming that `getBoolean` is reading from the INI here given this series of events.

Am I misunderstanding something?

JosJuice: In `setBoolean`, you are writing to the key `SettingsFile.KEY_WIIMOTE_ORIENTATION + wiimoteNumberAdjusted` when `settings.isGameSpecific() == true`, but `getBoolean` is always reading from the key `SettingsFile.KEY_WIIMOTE_ORIENTATION`. I'm assuming this has something to do with it?

Ebola16: Oh ok but the initial retrieved key needs to be `SettingsFile.KEY_WIIMOTE_ORIENTATION` with custom game settings. `SettingsFile.KEY_WIIMOTE_ORIENTATION + wiimoteNumberAdjusted` is for subsequent calls before saving. I'd need to introduce some way to switch between those two so I used a variable.

JosJuice: Oh, I forgot that this is one of those garbage settings that are stored in multiple ways :(

Why does it need to be like this, though? Would it not be possible to load from and save to the same key? (Not that I expect you to necessarily know the answer, since the way it works in master is confusing enough that I don't really understand any part of it myself...)

Ebola16: Looking at `SettingsFile.saveCustomGameSettings` was the only way I could understand this questionable system. The key is used to determine save location. Unfortunately I think the real motivations for this lie in the C++ side and I'm less familiar with that part of the code.

I think we can agree that changing the way keys are saved is beyond the scope of this PR. Am I good to leave the variable as-is?

JosJuice: Well, I don't have a better suggestion, so I suppose you can leave it like this.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/ui/SettingsFragmentPresenter.java
Ebola16: I think I see what was causing me concern. "Extension" and "Orientation" are both Wii profile custom game settings created along with the Wii Profile. This means "Orientation" will always be considered overridden with your suggested change. "Extension" is not considered overridden unless the setting is modified (like the other settings). These settings shouldn't be behaving differently since they come from the same section of the same file. I think I should keep the behavior I introduced, as the `.exists` check seems to assume the setting doesn't exist by default, which isn't true for profile settings.

JosJuice:
I'm unfortunately not sure if I'll be able to fully review this PR. The current state of the Android controller config code is just too much of a mess for me to understand all of it.

Ebola16:
> I'm unfortunately not sure if I'll be able to fully review this PR. The current state of the Android controller config code is just too much of a mess for me to understand all of it.

@JosJuice Well this PR would be much worse without your reviews!

I feel fairly confident that these adaptations to our questionable Android controller config code are significant QOL improvements without making anything worse. Although I can't explain why the config system is the way it is, I'm capable of explaining any changes I've made to it.

Perhaps @sepalani @lioncash @jordan-woyak or @BhaaLseN would also like to review this PR since it is rather complex? I don't think I can request reviews directly so I'll use mentions.

And @JMC47 may want to test it again. A lot has changed under the hood but the only noticeable changes since your last test should be the new location of the "Disable Unselected Wii Overlay Controllers" setting and the removal of inadequate fixes for custom game setting issues unrelated to this PR.

Three playtesters (including myself) have already said this shows promise. Maybe one or two more people could look at this and then yolo merge it?

Ebola16:
- Converted `wiiController` to an INI setting. This allows it to be retained across Dolphin builds and is part of the prep work for using Wii Overlays on a per-game basis.
- Read the Wii Overlay controller if WiimoteNew.ini is overwritten. This allows controllers to be retained across Dolphin builds.

I love how a minor tweak to `handleWiiOverlayDisablesControllersSetting` solves an unrelated problem in `DirectoryInitialization`.

Ebola16:
@dolphin-emu-bot rebuild

Ebola16:
I changed `SetOverlaySIDevice` to `SetSIDevice`, which can be passed a channel. This will make expanding changes in this PR to all channels easier in future PRs.

Ebola16:
@dolphin-emu-bot rebuild

Ebola16:
Since real Wii remotes can be selected from the overlay now, enabling continuous scanning should also be handled in this PR, both inside and outside of emulation. Previously we told users to do this manually but we might as well do it for them. Continuous scanning can be manually disabled if a user really needs it disabled.

Ebola16:
I should probably mark this as a draft again since #9170 and #9193 will change some major parts of this PR.

Ebola16:
@dolphin-emu-bot rebuild

Ebola16:
Just an update, I've been resolving a bunch of problems after rebasing this PR. I still have some work to do but I'm getting close to completing it.

jkms:
I believe this PR is necessary for new super Mario bros Wii, as that game won't seem to accept input from a wiimote unless it is detected as being sideways.  Would cool to be able to play on my pixel 3.

JosJuice:
Sorry for the very long delay, but my input overhaul (PR #11385) has now been merged, and it fixes several of the problems you have tried to address in this PR. If you still are interested in getting fixes from this PR merged into Dolphin, please open new PRs for them, and try to split the changes into multiple PRs or multiple commits when possible. The fact that this PR contains many changes of various quality is why it has had trouble getting anywhere in the review process.

PR #8895, opened 2020-06-24
GCMemcard: Rename unknown field to m_dtv_status

sepalani:
This PR renames an unknown field based on the register value it uses (see [YAGCD](http://hitmen.c02.at/files/yagcd/yagcd/chap5.html#sec5.3)).

This value is set in the `__CARDFormatRegionAsync` function. I don't know if the old comments still apply.

Ready to reviewed.

### DIFF
--- a/Source/Core/Core/HW/GCMemcard/GCMemcard.cpp
+++ b/Source/Core/Core/HW/GCMemcard/GCMemcard.cpp
@@ -1568,11 +1568,9 @@ void InitializeHeaderData(HeaderData* data, const CardFlashId& flash_id, u16 siz
   }
   data->m_sram_bias = rtc_bias;
   data->m_sram_language = sram_language;
-  // TODO: determine the purpose of m_unknown_2
+  // TODO: determine the purpose of m_dtv_status
   // 1 works for slot A, 0 works for both slot A and slot B
-  std::memset(
-      data->m_unknown_2.data(), 0,
-      data->m_unknown_2.size());  // = _viReg[55];  static vu16* const _viReg = (u16*)0xCC002000;
+  data->m_dtv_status = 0;
   data->m_device_id = 0;
 }
 
--- a/Source/Core/Core/HW/GCMemcard/GCMemcard.h
+++ b/Source/Core/Core/HW/GCMemcard/GCMemcard.h
@@ -175,7 +175,7 @@ static_assert(std::is_trivially_copyable_v<GCMBlock>);
 struct HeaderData
 {
   // NOTE: libogc refers to 'Serial' as the first 0x20 bytes of the header,
-  // so the data from m_serial until m_unknown_2 (inclusive)
+  // so the data from m_serial until m_dtv_status (inclusive)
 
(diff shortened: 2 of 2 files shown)


Source/Core/Core/HW/GCMemcard/GCMemcard.h
```diff
@@ -189,8 +189,8 @@ struct HeaderData
   // 4 bytes at 0x0018: SRAM language
   Common::BigEndianValue<u32> m_sram_language;
 
-  // 4 bytes at 0x001c: ? almost always 0
-  std::array<u8, 4> m_unknown_2;
+  // 4 bytes at 0x001c: VI DTV status register value (u16 from 0xCC00206E)
```
iwubcode, line 192: Maybe add the documentation as a comment:

```
this register allows software to read the status of two i/o pins
```

Not the most helpful but might provide a hint to someone in the future?  I always wonder what might happen if those documentation websites die.  I feel like the source should be self documented.

sepalani: I don't think that's necesary for several reasons:
 - YAGCD is well known and available both as online and offline documentation, plus the description provided is `VISEL - don't care`.
 - This register is already documented in [libogc/video](https://github.com/devkitPro/libogc/blob/master/libogc/video.c).
 - It's also documented in Dolphin code itself [UVIDTVStatus](https://github.com/dolphin-emu/dolphin/blob/master/Source/Core/Core/HW/VideoInterface.h)

Source/Core/Core/HW/GCMemcard/GCMemcard.cpp
```diff
@@ -1568,11 +1568,9 @@ void InitializeHeaderData(HeaderData* data, const CardFlashId& flash_id, u16 siz
   }
   data->m_sram_bias = rtc_bias;
   data->m_sram_language = sram_language;
-  // TODO: determine the purpose of m_unknown_2
+  // TODO: determine the purpose of m_dtv_status
```
BhaaLseN, line 1571: Is this comment still necessary (or rather; something we could/should address here?) with the revalation that it appears to be the dtv status (guessing: component cable attached?)

Although it seems a bit odd that it would store that in the memory card; especially with the comment below that setting this flag only works for slot A.

sepalani: As I said I don't know if this old comment still apply. I'm almost certain that at least two of my GC games store the u16 value (promoted into u32) from 0xCC00206E (which happens to be VI DTV status register aka `_viReg[55]`). I can provide screenshots of the disassembled `__CARDFormatRegionAsync` function if needed.

The real question is rather how the testing was done, against which game/region. There is also a big chance that other region related settings might interfere with this one and is the cause of this misunderstanding.

leoetlino: Then we should probably still keep that comment :)

leoetlino (approved):

PR #8902, opened 2020-06-27
Android: Add disc image conversion

JosJuice:
Screenshot_20200627-125148

### DIFF
--- a/Source/Android/app/src/main/AndroidManifest.xml
+++ b/Source/Android/app/src/main/AndroidManifest.xml
@@ -96,6 +96,10 @@
             </intent-filter>
         </activity>
 
+        <activity
+            android:name=".activities.ConvertActivity"
+            android:theme="@style/DolphinBase" />
+
         <service android:name=".utils.DirectoryInitialization"/>
         <service android:name=".services.GameFileCacheService"/>
         <service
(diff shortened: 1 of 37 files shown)


JosJuice:
After coding this, I came to the conclusion that I don't like Android.

JosJuice:
JMC wanted to see how the progress dialog is supposed to look.

Screenshot_20200912-124959694

JMC47:
@Ebola16 Can you give this a quick look before I yolo merge it :)

JMC47:
I've tested this thoroughly and it appears that no one wants to look into the java code.  Space is at a premium on phones and having this feature would be fantastic.  It's been sitting here for 3 months now.

7aychu23:
Can i Request Android can play DTM TAS REPLAY like PC :) Or Add Cheat Generator = Memory Viewer Like Citra

Sanfe:
I tested this feature in last beta and have a bug. In RVZ format, the compression level only could be up to 9. If you put 10 or more crash. Also, a progress bar in notification bar is needed to could make the process in background. Batch multiple compression could be a good plus.

JMC47:
The crash will be fixed tomorrow.

JosJuice:
I only meant I would create the PR tomorrow, not that it will be merged tomorrow :P

But yeah, it should be a simple fix.

faroukabd000:
How about the method that have been used in nkit it's fast and the size really small

JosJuice:
It also breaks games in certain ways, while not really being faster or smaller than RVZ.

faroukabd000:
The converting process doesn't appear in mi a3

PR #8903, opened 2020-06-27
GCMemcard: Change behavior of TitlePresent() to more closely resemble how saves are actually identified.

AdmiralCurtiss:
This modifies `GCMemcard::TitlePresent()` to match my findings of how the GC BIOS and various games behave when you muck with the fields in the directory entry.

It looks like for a save to be recognized by a game, the following have to be true:
- Game code and maker code must exactly match what the game expects.
- Filename is only checked up to the first null byte. All bytes afterwards can be whatever.

The BIOS itself does a full compare of the filename when checking for whether it should allow copying a file from one card to another, but behaves oddly in some cases when there's non-null bytes after the first null. See the big comment in `HasSameIdentity()` for details.

The new `GCMemcardUtils.cpp` file may seem a bit excessive for now, but I have some plans for some improvements to the memory card manager that would fit well in there, and `GCMemcard.cpp` is pretty crowded already.

I'm also not sure if "Identity" is a good word to describe what we're comparing here; any other suggestions?

### DIFF
--- a/Source/Core/Core/HW/GCMemcard/GCMemcardUtils.cpp
+++ b/Source/Core/Core/HW/GCMemcard/GCMemcardUtils.cpp
@@ -1,0 +1,56 @@
+#include "Core/HW/GCMemcard/GCMemcardUtils.h"
+
+#include "Core/HW/GCMemcard/GCMemcard.h"
+
+namespace Memcard
+{
+bool HasSameIdentity(const DEntry& lhs, const DEntry& rhs)
+{
+  // The Gamecube BIOS identifies two files as being 'the same' (that is, disallows copying from one
+  // card to another when both contain a file like it) when the full array of all of m_gamecode,
+  // m_makercode, and m_filename match between them.
+
+  // However, despite that, it seems like the m_filename should be treated as a nullterminated
+  // string instead, because:
+  // - Games seem to identify their saves regardless of what bytes appear after the first null byte.
+  // - If you have two files that match except for bytes after the first null in m_filename, the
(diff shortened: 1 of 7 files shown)


lioncash:
Commit should probably have the extended description added to it, so that info is stored in the repository

AdmiralCurtiss:
Good point, done.

Source/Core/Core/HW/GCMemcard/GCMemcardUtils.cpp
```diff
@@ -0,0 +1,51 @@
...
+
+  // With all that in mind, I think we can safely assume that all bytes after the first null are to
+  // be ignored, and that m_filename is supposed to be a nullterminated string.
+
+  if (lhs.m_gamecode != rhs.m_gamecode)
+    return false;
+  if (lhs.m_makercode != rhs.m_makercode)
+    return false;
+
+  for (size_t i = 0; i < lhs.m_filename.size(); ++i)
```
BhaaLseN, line 39: Should this perhaps be `i < std::min(lhs.m_filename.size(), rhs.m_filename.size())`? Or do we make sure that the `lhs` filename is shorter than (or of equal length to) the `rhs` filename?

BhaaLseN: Scratch that, I just noticed this is an array and not an actual string.

AdmiralCurtiss:
Clarified the description a bit, code was not changed.

PR #8904, opened 2020-06-28
FreeLookCamera: Fix savestate branch inconsistency.

AdmiralCurtiss:
Fixes https://bugs.dolphin-emu.org/issues/12168


lioncash:
Bug being fixed should probably be included in the commit message itself

AdmiralCurtiss:
Like this?

lioncash:
Yup 👌

PR #8907, opened 2020-06-28
Android: Use octagonal stick gate in overlay

JosJuice:
Currently, the touch controller overlay uses a square gate for sticks. The first commit of this PR changes that so that it instead uses the stick gate configured in the INI, which ensures that the values sent to the core are appropriately scaled regardless of what is configured in the INI and makes the overlay look nicer if the INI is set to a stick gate that matches the graphics.

With that change out of the way, we can essentially revert PR #8503, fixing the current problems with being unable to run sideways when using certain gamepads in certain games, while not breaking the touch controls bug https://bugs.dolphin-emu.org/issues/11911 which was fixed by PR #8503.


jordan-woyak:
Won't removing the default calibration reintroduce "thumbstick down acts like up" issue for gamepads (not touchscreen)?

JosJuice:
It would, if the range of the gamepad stick is very large. I'm not sure if that is something that is common in practice. Of course, the proper fix for that problem is to make it configurable, which is easier said than done with the current state of the Android input code...

BreakPoints:
> It would, if the range of the gamepad stick is very large. I'm not sure if that is something that is common in practice. Of course, the proper fix for that problem is to make it configurable, which is easier said than done with the current state of the Android input code...

I can confirm that this reverts the fix for my cheap iPega PG-9025 Bluetooth controller. Down is now up again.

PR #8911, opened 2020-06-29
JitArm64: Don't assume fastmem arena is available

OatmealDome:
Should fix JitArm64 crashing when fastmem is off. The instruction ``lfd`` also is causing a crash for me, but that might just be an issue with my local build. Testers are appreciated...


JMC47:
@dolphin-emu-bot rebuild

degasus:
Sounds like a good fix. Thanks for the patch.

PR #8913, opened 2020-06-29
Support the new MSVC preprocessor

JosJuice:
Intends to fix https://bugs.dolphin-emu.org/issues/12170.

Untested, because I'm using VS 16.5 to avoid the std::result_of problem, and as far as I know the new preprocessor requires VS 16.6.

### DIFF
--- a/Source/Core/Common/MsgHandler.h
+++ b/Source/Core/Common/MsgHandler.h
@@ -32,7 +32,7 @@ bool MsgAlert(bool yes_no, MsgType style, const char* format, ...)
 void SetEnableAlert(bool enable);
 }  // namespace Common
 
-#ifdef _WIN32
+#if defined(_WIN32) && (!defined(_MSVC_TRADITIONAL) || _MSVC_TRADITIONAL == 1)
 #define SuccessAlert(format, ...)                                                                  \
   Common::MsgAlert(false, Common::MsgType::Information, format, __VA_ARGS__)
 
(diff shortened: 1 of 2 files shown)


BhaaLseN:
FYI, I tried building this branch on 16.6.2 (along with `/std:cpp17` and `/experimental:newLambdaProcessor` because it wouldn't build at all otherwise), and it works without `/Zc:preprocessor`. However, once I add that, it fails inside `ucrt_math.h` (included by `pch`) with Windows SDK 10.0.17763.0. filoppi confirmed on IRC that it does work with Windows SDK 10.0.19041.0 though, so we might also have to consider documenting this somewhere. We used to have the version in `README.md`, but since switched it to just say "Windows 10 SDK".

JosJuice:
> filoppi confirmed on IRC that it does work with Windows SDK 10.0.19041.0 though, so we might also have to consider documenting this somewhere.

Since we're not actually enabling `/Zc:preprocessor`, is there really a need to document it?

BhaaLseN:
I'm guessing it would become the default at some point, for compatibility with other compilers? I mean, good enough for me if we want to handle it if (or when) it comes up.

AdmiralCurtiss:
Is there anything stopping us from upgrading Qt, by the way?

JosJuice:
No, other than people not knowing how to or wanting to build the new version to put in Externals.

PR #8914, opened 2020-06-29
Jit64: Implement low DCBZ hack

JosJuice:
I was hoping this would improve the performance of Cars 2 by avoiding interpreter fallbacks, but it doesn't seem to have made any measurable impact.

### DIFF
--- a/Source/Core/Core/PowerPC/Jit64/Jit_LoadStore.cpp
+++ b/Source/Core/Core/PowerPC/Jit64/Jit_LoadStore.cpp
@@ -290,7 +290,6 @@ void Jit64::dcbz(UGeckoInstruction inst)
 {
   INSTRUCTION_START
   JITDISABLE(bJITLoadStoreOff);
-  FALLBACK_IF(SConfig::GetInstance().bLowDCBZHack);
 
   int a = inst.RA;
   int b = inst.RB;
@@ -302,6 +301,14 @@ void Jit64::dcbz(UGeckoInstruction inst)
 
     MOV_sum(32, RSCRATCH, Ra, Rb);
     AND(32, R(RSCRATCH), Imm32(~31));
+  }
+
+  FixupBranch end_dcbz_hack;
+  if (SConfig::GetInstance().bLowDCBZHack)
+  {
+    // HACK: Don't clear any memory in the [0x8000'0000, 0x8000'8000) region.
+    CMP(32, R(RSCRATCH), Imm32(0x8000'8000));
+    end_dcbz_hack = J_CC(CC_L);
   }
 
   bool emit_fast_path = MSR.DR && m_jit.jo.fastmem_arena;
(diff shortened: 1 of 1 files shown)


JMC47:
Because the DCBZ low hack is only used during the game's boot process, this shouldn't actually affect in-game performance.  May smooth out some hitches when loading through.

The actual game itself is slow for the same reason that Star Wars: Rogue Squadron 2/3 are slow.

JosJuice:
> Because the DCBZ low hack is only used during the game's boot process, this shouldn't actually affect in-game performance.

DCBZ is used after boot too. Maybe the hack doesn't matter at that point, but the hack remains enabled, so DCBZ has to fall back to interpreter.

JMC47:
Ah I see.  If we could figure out why Dynamic BATs slowed down MMU games, we could probably get a decent performance boost if we could figure out why.

Source/Core/Core/PowerPC/Jit64/Jit_LoadStore.cpp
```diff
@@ -304,6 +303,15 @@ void Jit64::dcbz(UGeckoInstruction inst)
     AND(32, R(RSCRATCH), Imm32(~31));
   }
 
+  FixupBranch end_dcbz_hack;
+  if (SConfig::GetInstance().bLowDCBZHack)
+  {
+    // HACK: Don't clear any memory in the 0x80000000 - 0x80008000 region
+    LEA(32, RSCRATCH2, MComplex(RSCRATCH, RSCRATCH, SCALE_ATREG, 0x80000000));
```
Sintendo, line 310: I'd prefer `MDisp` here:
```LEA(32, RSCRATCH2, MDisp(RSCRATCH, 0x80000000));```

File: Source/Core/Core/PowerPC/Jit64/Jit_LoadStore.cpp [hunk 2]
```diff
@@ -304,6 +303,15 @@ void Jit64::dcbz(UGeckoInstruction inst)
     AND(32, R(RSCRATCH), Imm32(~31));
   }
 
+  FixupBranch end_dcbz_hack;
+  if (SConfig::GetInstance().bLowDCBZHack)
+  {
+    // HACK: Don't clear any memory in the 0x80000000 - 0x80008000 region
+    LEA(32, RSCRATCH2, MDisp(RSCRATCH, 0x80000000));
+    CMP(32, R(RSCRATCH2), Imm32(0x7fff));
+    end_dcbz_hack = J_CC(CC_LE);
```
Tilka, line 312: CC_LE is a signed condition, I think you meant CC_BE. But since `0x8000'0000` is the smallest signed 32-bit integer, you can actually make use of that and get rid of the LEA (untested):

```suggestion
    // HACK: Don't clear any memory in the [0x8000'0000, 0x8000'8000) region.
    CMP(32, R(RSCRATCH), Imm32(0x8000'8000));
    end_dcbz_hack = J_CC(CC_L);
```
Neither GCC nor Clang seem to do this optimization so you might want to doublecheck...

(I also took the liberty of clarifying the interval boundaries.)

JosJuice: Makes sense, and I tested that it doesn't break Cars 2.

PR #8915, opened 2020-06-29
macOS: Untranslocate the app bundle's path for the updater

OatmealDome:
Starting in macOS Sierra, if an app is downloaded from the Internet, the system may "translocate" an app's bundle to a read-only container elsewhere on the filesystem and execute it from there. This is done to protect against a trusted application's resources from being hijacked to load untrusted code. More information is available [here](https://objective-see.com/blog/blog_0x15.html) (with many thanks to Patrick Wardle).

Unfortunately, this has the side-effect of breaking the updater, as it'll attempt to find the updater app bundle from the read-only container. So, ``GetBundleDirectory()`` has been changed to "untranslocate" the app bundle's path if necessary. The APIs used to do this were briefly public in Xcode 8 beta 1, but they were [made private in a later beta](https://lapcatsoftware.com/articles/mia-sectranslocate.html). There is no other way to untranslocate a path (as far as I know). The headers [are still open source](https://opensource.apple.com/source/Security/Security-57740.51.3/OSX/libsecurity_translocate/lib/SecTranslocate.h.auto.html), however.

Also, this PR includes a fix for https://bugs.dolphin-emu.org/issues/11784. The updater's working directory is now created in the OS's standard temp directory, and the log file is placed in ``User/Logs``.

### DIFF
--- a/Source/Core/Common/FileUtil.cpp
+++ b/Source/Core/Common/FileUtil.cpp
@@ -20,6 +20,9 @@
 #include "Common/CommonFuncs.h"
 #include "Common/CommonPaths.h"
 #include "Common/CommonTypes.h"
+#ifdef __APPLE__
+#include "Common/DynamicLibrary.h"
+#endif
 #include "Common/FileUtil.h"
 #include "Common/IOFile.h"
 #include "Common/Logging/Log.h"
@@ -64,6 +67,18 @@ namespace File
 {
 #ifdef ANDROID
 static std::string s_android_sys_directory;
+#endif
+
+#ifdef __APPLE__
+static Common::DynamicLibrary s_security_framework;
+
+using DolSecTranslocateIsTranslocatedURL = Boolean (*)(CFURLRef path, bool* isTranslocated,
+                                                       CFErrorRef* __nullable error);
+using DolSecTranslocateCreateOriginalPathForURL = CFURLRef
+__nullable (*)(CFURLRef translocatedPath, CFErrorRef* __nullable error);
+
+static DolSecTranslocateIsTranslocatedURL s_is_translocated_url;
+static DolSecTranslocateCreateOriginalPathForURL s_create_orig_path;
 #endif
 
 #ifdef _WIN32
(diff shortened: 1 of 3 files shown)


Source/Core/Common/FileUtil.cpp
```diff
@@ -59,6 +63,19 @@ namespace File
 static std::string s_android_sys_directory;
 #endif
 
+#ifdef __APPLE__
+static Common::DynamicLibrary s_security_framework;
+
+typedef Boolean (*dolSecTranslocateIsTranslocatedURL)(CFURLRef path, bool* isTranslocated,
```
lioncash, line 69:
```suggestion
using DolSecTranslocateIsTranslocatedURL = Boolean (*)(CFURLRef path, bool* isTranslocated,
```

Ditto for the other typedef

Source/Core/Common/FileUtil.cpp
```diff
@@ -659,16 +674,49 @@ std::string GetTempFilenameForAtomicWrite(std::string path)
...
+    // Security/OSX/libsecurity_translocate/lib/SecTranslocate.h
+    if (!s_security_framework.IsOpen())
+    {
+      s_security_framework.Open("/System/Library/Frameworks/Security.framework/Security");
+      s_security_framework.GetSymbol("SecTranslocateIsTranslocatedURL", &s_is_translocated_url);
+      s_security_framework.GetSymbol("SecTranslocateCreateOriginalPathForURL", &s_create_orig_path);
+    }
+
+    bool is_translocated = false;
+    s_is_translocated_url(bundle_ref, &is_translocated, NULL);
```
leoetlino, line 703:
```suggestion
    s_is_translocated_url(bundle_ref, &is_translocated, nullptr);
```

Source/Core/Common/FileUtil.cpp
```diff
@@ -659,16 +674,49 @@ std::string GetTempFilenameForAtomicWrite(std::string path)
...
+
+    if (is_translocated)
+    {
+      CFURLRef untranslocated_ref = s_create_orig_path(bundle_ref, NULL);
```
leoetlino, line 707:
```suggestion
      CFURLRef untranslocated_ref = s_create_orig_path(bundle_ref, nullptr);
```

leoetlino (changes requested):

OatmealDome:
Addressed comments.

Source/Core/UpdaterCommon/UpdaterCommon.cpp
```diff
@@ -747,10 +720,7 @@ bool RunUpdater(std::vector<std::string> args)
   TodoList todo = ComputeActionsToDo(this_manifest, next_manifest);
   todo.Log();
 
-  std::optional<std::string> maybe_temp_dir = FindOrCreateTempDir(opts.install_base_path);
-  if (!maybe_temp_dir)
-    return false;
-  std::string temp_dir = std::move(*maybe_temp_dir);
+  std::string temp_dir = File::CreateTempDir();
```
leoetlino, line 723: I'm sorry, there is a behavioural change I missed during my last review... CreateTempDir can fail, in which case it will return an empty string. The return value should be checked to avoid littering the current directory with update files when CreateTempDir fails.

OatmealDome:
Finally got around to rebasing. Should fix at least one of the issues plaguing the macOS updater right now (app not launching if downloaded from the Internet). Tested on an Intel MacBook Pro, but should still be OK on an M1.

leoetlino (approved):

PR #8917, opened 2020-06-29
Pushed required version of lib-zstd to 1.4.0 or higher

jwsharpe:


JosJuice:
As far as I know, 1.3.8 or higher should work fine, so enforcing 1.4.4 seems a little excessive.

jwsharpe:
Was just matching the `./Externals` package, can bump down to 1.3.8 if you want.

Tilka:
> Was just matching the `./Externals` package, can bump down to 1.3.8 if you want.

Please do.

cyrozap:
Per #9017, the minimum version should probably be kept at 1.4.0, since Dolphin fails to build with zstd 1.3.8 on Debian 10.

PR #8918, opened 2020-07-01
Revert "JitArm64: no intermediate rounding for paired FMA"

JosJuice:
This reverts commit a04ac23794aa09ae730ce7ac49143125f5b9f7c7.

Not sure what's wrong with the code, but a quite serious problem was reported in https://bugs.dolphin-emu.org/issues/12173, so I think it's best to revert it for now.


JosJuice:
Properly fixed in PR #8919.

PR #8920, opened 2020-07-01
Update cubeb to latest

Filoppi:
Based on https://github.com/kinetiknz/cubeb/commit/ea39471a989d35224cc5905f9467da5a36917827
-Lower latency on Windows 10 (in some cases) (not enabled by default)
-Fixes all around, possibly on other platforms as well
-Fixes a small memory leak
-Most notably fixed sound stopping after changing the properties of a device during playback on Windows
-Fixes inconsistent CoInitializeEx and CoUninitialize calls.

See here for information on the low latency:
https://docs.microsoft.com/en-us/windows-hardware/drivers/audio/low-latency-audio#windows-audio-session-api-wasapi

After the update commit, I've applied a few fixes separately which could be ignored on the next update, but they will likely have been fixed by then (I have made PRs for them):
-Make sure cubeb uses IAudioClient3 when using a rendering (output) device (actually enable low latency)
-Don't use AUDCLNT_STREAMFLAGS_NOPERSIST, which prevents volume and mute setting changes from being saves by the Windows Mixer

Needs testing on Mac OS, Linux, and Android, if it's used there as well. Needs testing on the GC controller emulated microphone. Though the changes in cubeb are actually quite small.


Filoppi:
BTW, I had tested this in its final version, I just split it up before submitting and messed up a copy and paste.

JMC47:
@dolphin-emu-bot rebuild

Filoppi:
@dolphin-emu-bot rebuild

Tilka:
Can you squash the update commits, please?

Filoppi:
This currently does not build on the Windows build machine, but it seems like it would build fine on my PC with c++ 17 and MSVC 14.25, what MSVC version is the Windows build machine using? Could someone else try this please?

Source/Core/AudioCommon/CubebUtils.cpp
```diff
@@ -62,6 +66,11 @@ std::shared_ptr<cubeb> CubebUtils::GetContext()
     ERROR_LOG(AUDIO, "Error setting cubeb log callback");
   }
 
+#ifdef _WIN32
+  // Cubeb requires us to do it ourselves
+  CoInitializeEx(nullptr, COINIT_MULTITHREADED);
```
ligfx, line 71: Note that this is a leak, of sorts (not sure what happens if you keep calling it without uninitializing, might just increment a counter).

It can also fail (probably won't, but still good practice to check the return value).

Microsoft documentation claims `CoInitializeEx must be called at least once, and is usually called only once, for each thread that uses the COM library`, which is *not* satisfied by just calling it in `GetContext()`. The returned context can and will be used across multiple threads. *However,* in practice this isn't an issue, as once the multithreaded apartment is initialized, all threads without COM initialized will act as though they in the multithreaded apartment. More information in kinetiknz/cubeb#534

Filoppi: Good find, I even went on to check every line of code in Cubeb to make sure it didn't have memory leaks and forgot to check on my own code. This would be a leak when the game/emulation thread is closed but Dolphin is not.

CoInitializeEx can be called as many times as you want on the same thread, calls after the first one have no effect, but it will need to be uninitialised yeah (if any of the calls succeeded).

I'm wondering if we need to call CoInitialize ourselves on the audio thread as well, given that cubeb_wasapi.cpp never calls it anywhere. Though it probably wouldn't work without it so I guess it's fine as it is.

ligfx: > Good find, I even went on to check every line of code in Cubeb to make sure it didn't have memory leaks and forgot to check on my own code. This would be a leak when the game/emulation thread is closed but Dolphin is not.
>
> CoInitializeEx can be called as many times as you want on the same thread, calls after the first one have no effect, but it will need to be uninitialised yeah (if any of the calls succeeded).

Right, it's not a leak per-se, it's just you won't be uninitializing it so I wanted to call it out. Probably worth putting comments in explaining what's happening.

This can affect multiple threads: whichever threads are creating the CubebStream (system thread? UI thread?), as well as whichever thread is creating the EXI_DeviceMic (system thread?). I remember seeing that it's very easy to trigger this. Again, not really a problem, but since it's counter to COM best practice I wanted to point it out. It would be needlessly convoluted to set it up correctly in Dolphin, its really cubeb's responsibility (I have an old cubeb PR open for this).

> I'm wondering if we need to call CoInitialize ourselves on the audio thread as well, given that cubeb_wasapi.cpp never calls it anywhere. Though it probably wouldn't work without it so I guess it's fine as it is.

This goes back to the default-multithreaded-apartment-once-the-multithreaded-apartment-is-created implementation: if any thread, anywhere, has called CoInitializeEx, then you can use COM from anywhere. Not documented, but works in practice.

Filoppi: Should be fixed, please take a look

Filoppi: I need to make sure only one thread can create and access the cubeb backend. Regarding the GC controller microphone, it should be fine. I actually forgot earlier but every succeeding call to CoInitializeEx need to have a CoUninitialize yeah, it probably works with a counter internally.

ligfx: Oh nope, that won't work. Since cubeb doesn't keep a reference to COM, uninitializing after get a cubeb context will break.

Like I said, there's not really a great way to handle this in Dolphin. I'd just call CoInitializeEx every time you need a cubeb context, punt on uninitializing, and put in a comment explaining the issue.

Filoppi: The CoUninitialize happens on the destruction of the cubeb context, which happens after the stream is closed, when the audio backend shuts down, so it's perfectly fine.

ligfx: You're right, my mistake. I didn't see the if-clause around the second CoUninitializeEx.

What happens if a cubeb context is created on one thread, and destroyed on another? Is COM okay with that?

Filoppi: No it wouldn't be ok, but it doesn't happen, the audio backend is created and destroyed from the same thread.
What is definitely called from multiple threads is SoundStream::SetRunning(bool running), whether Cubeb or WASAPI (I know because I had troubles with the wasapi CoInit there).
Also, not sure about the microphone code threading, I will ask around since I can't test it.

ligfx: When I tested this last time, it does get called from multiple threads, and causes segfaults.

Filoppi: https://github.com/dolphin-emu/dolphin/blob/92d24aec2e45fc65767444b220108595f8e74b94/Source/Core/InputCommon/ControllerInterface/Win32/Win32.cpp#L49 this line here is breaking WASAPI CoInit calls because some are COINIT_APARTMENTTHREADED and some are COINIT_MULTITHREADED. Are you fine with me changing that one to COINIT_MULTITHREADED?

Filoppi: No actually, that doesn't seem the problem. I'm getting RPC_E_CHANGED_MODE but I always use COINIT_MULTITHREADED. Investigating

Filoppi: It seems like QT or some other library is calling CoInitializeEx with COINIT_APARTMENTTHREADED in the main thread, then, my WASAPI calls use COINIT_MULTITHREADED so i get the RPC_E_CHANGED_MODE. I'm trying to understand which one we actually want to use then.

Filoppi: The microphone access seems to be safe and only done from one thread (I haven't tested at all but it seems that way)

ligfx: Microphone access from multiple threads can be reproduced by:
1. starting the game with a microphone configured in Slot A or B
2. unconfiguring the microphone while the game is running

Filoppi: True. I didn't know it was that easy to test. Not a problem, I will initialize and unitialize COM every time a cubeb function is called from a dolphin thread.

Filoppi: The fact that you can receive RPC_E_CHANGED_MODE is irrelevant as we can just keep going, unfortunately, there isn't much we can do, it's likely an external library that sets it up as COINIT_APARTMENTTHREADED .
Before this conversion, cubeb itself used to be COINIT_MULTITHREADED, and we likely want to keep it that way as all the code from our side thread safe.

ligfx: > Not a problem, I will initialize and unitialize COM every time a cubeb function is called from a dolphin thread.
 
This might work sometimes, but isn't how COM is supposed to be use. Any objects living in the apartment will be destroyed when the apartment is uninitialized. If you happen to get one of those, it will segfault the next time it's used.

Filoppi: If that's so, it would have done the same before this Cubeb update. The only change is that CoInit calls are moved before calling Cubeb instead than being inside of it.

Filoppi: But given that an apartment is already created even before lunching a game (when querying stuff for WASAPI in the UI and calling CoInit, I get RPC_E_CHANGED_MODE), then it's likely that whoever called the first CoInit, won't call CoUnit until the very end of the main thread. We can't guarantee that, but we can't do anything else, it's not in Dolphin or any externals (unless there are Windows calls that wrap CoInit inside of them).

Filoppi: And I doubt the apartment would be uninitialised until the very last CoUnit, so we have control over that, given that we have our main cubeb context CoInit ongoing until it is destroyed when the audio backend is shut down or the microphone removed.

Filoppi: @ligfx Do you know if accessing a COM object created in a STA thread from an MTA thread is fine? That's when you get the RPC_E_CHANGED_MODE.
Now, cubeb in Dolphin currently has this "problem" already so it's nothing new, again, all that changed is the place where CoInitializeEx is called, but the old cubeb ignores the RPC_E_CHANGED_MODE error, and just keeps going.
I'm still new to this stuff, all I can see is that to access a COM object created in a STA thread from another STA thread, you need marshalling, but I can't find if it's fine to access it from an MTA thread, and whether you need to marshal it yourself, or if it does it for you internally (I don't know what it means exactly). All I know is that all of this is thread safe and there are no concurrent accesses.
Also, note that the STA thread won't be destroyed nor call CoUninit (mirroring the first CoInit) until the very end of the life of the application, so that is not a problem.

ligfx: I believe all WASAPI COM objects live in the multithreaded apartment regardless of where they are created, so you can access them from STA or MTA threads fine (they already assume multithreaded access, unlike objects in the STA).

Filoppi: In WASAPI CoCreateInstance() of the IMMDeviceEnumerator is created on a MTA thread and accessed from both MTA and STA threads, but the IAudioClient can be created from an MTA and STA thread (main or emulation threads), and that is also a COM object.
In cubeb, it's very similar, the same can happen.

Filoppi: > I believe all WASAPI COM objects live in the multithreaded apartment regardless of where they are created, so you can access them from STA or MTA threads fine (they already assume multithreaded access, unlike objects in the STA).

By the way, not at all, the current implement of wasapi calls CoInitialize, not CoInitializeEx, which implies they are in an STA. But I have completely re-written the WASAPI stream, as it had loads of crashes and non thread safe accesses.

ligfx: I don't think you understood what I said. Let's step back a bit.

In COM, [objects live in certain apartments](https://docs.microsoft.com/en-us/previous-versions/ms809971(v=msdn.10)?redirectedfrom=MSDN), depending on their declared threading model. Threading models are "Apartment" aka STA-only, "Free" aka free-threaded aka MTA-only, or "Both", aka whichever apartment it was created from.

When I say that WASAPI COM objects live in the MTA, I mean that I think they have the "Free" threading model (you can check in the Registry).

So what happens when you create an MTA-only/free-threaded object from an STA? COM will create the MTA if it doesn't exist, and put the object there.

The point being, it doesn't matter which apartment you create or access free-threaded objects from, as long as COM has been initialized.

Filoppi: All right. No I had not understood that threading model concept at all, I had misinterpreted, I need to go back and read more. But, other than that, it seems like this PR is good to go then? (I still need to commit a CoInitialize inside CubebStream::SetRunning())

ligfx: What happens if you start a game with the microphone enabled, then disable it while the game is running? My worry is calling CoUninitialize on the wrong thread would break things.

Filoppi: Check again, it should be fine now. Again, remember that these problems are currently present already.

Filoppi: Yes there is that problem. void CEXIMic::StreamInit() and void CEXIMic::StreamTerminate() can be called by different threads for the same object. I'm not sure how to fix that (though it worked fine for this whole time)

Filoppi: A solution could be to wrap all the cubeb calls, like cubeb_init, cubeb_stream_init and cubeb_stream_start around CoInitializeEx and CoUninitialize, that way, no CoInitialize will ever be pending between threads. Will the COM objects be destroyed though?

ligfx: > Will the COM objects be destroyed though?

They might be. Note this is the reason why cubeb decided to move away from just wrapping calls, and towards requiring callers to handle COM: kinetiknz/cubeb#416

Filoppi: yes i had read that

Filoppi: all right then, what about this:
call the first CoInitialize by creating a thread_local ScopedThreadAccess (see it in CubebUtils.h), then, wrap every call to any cubeb method into another CoInitialize and CoUninitialize. When the thread is destroyed, the thread_local variable will be destroyed and call the final CoUninitialize.

ligfx: thread_local could work.

Not sure if you'd still need to wrap calls: COM seems to treat threads that haven't initialized COM as being in the MTA, if the MTA has already been initialized. Something to test.

Filoppi: Microsoft says: every thread that accesses COM objects needs to have called CoInit. If you don't call it, it's treated as implicit MTA: https://devblogs.microsoft.com/oldnewthing/20130419-00/?p=4613

ligfx: Ah, yep, there it is. So as long as at least one thread exists with an MTA, which is probably a reasonable assumption (e.g. emu thread starts before anything else, creates audio and EXI devices), you don't have to wrap calls.

Source/Core/AudioCommon/CubebUtils.cpp
```diff
@@ -16,6 +16,38 @@
...
+{
+#ifdef _WIN32
+  // COINIT_MULTITHREADED seems to be the better choice here given we handle
+  // multithread access safely ourself. It should also offer better performance.
+  // Note that calls to this from the main thread will return RPC_E_CHANGED_MODE
+  // as some external library is calling CoInitialize with COINIT_APARTMENTTHREADED
+  // without uninitializing it. We keep going.
+  // cubeb specifies we should use COINIT_MULTITHREADED:
+  // https://github.com/kinetiknz/cubeb/issues/416, but that can't be a problem
+  result = CoInitializeEx(NULL, COINIT_MULTITHREADED);
```
lioncash, line 29:
```suggestion
  result = CoInitializeEx(nullptr, COINIT_MULTITHREADED);
```

Source/Core/AudioCommon/CubebUtils.h
```diff
@@ -6,8 +6,26 @@
...
 #include <memory>
 
+#ifdef _WIN32
+#include <comdef.h>
+#endif
+
 struct cubeb;
 
+// Helper to initialize and uninitialize thread access for the libraries that cubeb uses
+struct ScopedThreadAccess
```
lioncash, line 16:
```suggestion
class ScopedThreadAccess
```

This should also be within the CubebUtils namespace below.

Source/Core/AudioCommon/CubebUtils.cpp
Filoppi: Check again, I've done what I mentioned

Source/Core/AudioCommon/CubebUtils.h
Filoppi: I moved it. I don't see how class or struct is relevant. I "copied" it from cubeb itself and it was a struct there.

Source/Core/AudioCommon/CubebUtils.cpp
Filoppi: Again, copied from cubeb, NULL is fine, unless there is a rule to not use it in Dolphin, not really worth changing.

Source/Core/AudioCommon/CubebUtils.h
```diff
@@ -4,6 +4,8 @@
 
 #pragma once
 
+#include "Common/Logging/Log.h"
```
Filoppi, line 7: Ignore this, I will fix it when I squash

Source/Core/AudioCommon/CubebUtils.cpp
JosJuice: There is a rule to not use it in Dolphin.

Source/Core/AudioCommon/CubebUtils.h
JosJuice: See https://github.com/dolphin-emu/dolphin/blob/master/Contributing.md#classes-and-structs for `class` vs `struct`. I agree with Lioncash here (not that it actually makes a huge difference which one you use).

JosJuice:
@dolphin-emu-bot rebuild

Source/Core/AudioCommon/CubebUtils.cpp
ligfx: Why avoid relying on the implicit MTA? It makes the code much cleaner, and doesn't expose the need to use COM to users of CubebUtils.

I'd suggest something like this:

```diff
diff --git a/Source/Core/AudioCommon/CubebUtils.cpp b/Source/Core/AudioCommon/CubebUtils.cpp
index b4af87115b..7e500ceb1b 100644
--- a/Source/Core/AudioCommon/CubebUtils.cpp
+++ b/Source/Core/AudioCommon/CubebUtils.cpp
@@ -14,6 +14,10 @@
 
 #include <cubeb/cubeb.h>
 
+#ifdef _WIN32
+#include <Objbase.h>
+#endif
+
 static ptrdiff_t s_path_cutoff_point = 0;
 
 static void LogCallback(const char* format, ...)
@@ -42,8 +46,47 @@ static void DestroyContext(cubeb* ctx)
   }
 }
 
+#ifdef _WIN32
+// cubeb's WASAPI requires the user to manage a valid COM context on all threads
+// accessing cubeb. There are two parts to this: 1) every thread that calls GetContext()
+// will have a COM context created if it doesn't exist 2) while a multi-threaded
+// COM context exists, any thread that doesn't initialize COM will automatically
+// be in the implicit MTA COM context.
+
+class auto_com {
+public:
+  auto_com() = default;
+  ~auto_com() {
+    if (m_should_uninitialize) {
+      CoUninitialize();
+    }
+  }
+  bool initialize() {
+    if (!m_initialized) {
+      HRESULT result = CoInitializeEx(nullptr, COINIT_MULTITHREADED);
+      m_initialized = SUCCEEDED(result) || result == RPC_E_CHANGED_MODE;
+      m_should_uninitialize = SUCCEEDED(result);
+    }
+    return m_initialized;
+  }
+private:
+  bool m_initialized = false;
+  bool m_should_uninitialize = false;
+};
+
+thread_local auto_com s_com_context;
+
+#endif // _WIN32
+
 std::shared_ptr<cubeb> CubebUtils::GetContext()
 {
+  #ifdef _WIN32
+    if (!s_com_context.initialize()) {
+       ERROR_LOG(AUDIO, "Error initializing COM for cubeb");
+       return nullptr;
+    }
+  #endif
+
   static std::weak_ptr<cubeb> weak;
 
   std::shared_ptr<cubeb> shared = weak.lock();
```

Filoppi: Did you manage to build PR this on Windows BTW?

JosJuice:
@dolphin-emu-bot rebuild

Source/Core/AudioCommon/CubebUtils.cpp
Filoppi: I think this the above code can work. GetContext() was called from the main thread, it would return RPC_E_CHANGED_MODE, meaning that it won't create any MTA. If there is no MTA, there is also no implicit MTA, right?

ligfx: I don't have a Windows computer at the moment so I can't build.

Here's the logic:
1) Emulation starts, emuthread is created, which starts audio (calls GetContext) and initializes EXI devices (might call GetContext), so initializes an MTA context and ensures the MTA exists.
2) If a microphone is hotplugged, the UI thread will call GetContext and get RPC_E_CHANGED_MODE. No problem, the MTA already exists, and it's fine to create WASAPI objects from the STA due to their threading model. 
3) If a microphone is hotunplugged, the UI thread will call cubeb functions directly. No problem, the MTA exists.
4) Any other thread that does stuff with cubeb: the emuthread should be running, which means the MTA exists.

Filoppi: Yes, I kind of understand that it would work, but this makes too many assumptions (some of which I don't have the knowledge to judge).
-How do you know the threading model of all WASAPI COM objects?
-cubeb calls CoInit with MTA on a lot of its functions just for satefy, and it calls it on every new thread, never relying on implicit MTA. Implicit MTA can be used yes, but if the thread that created the MTA closes it when another acesss is ongoing, it's not good. This doesn't happen now, but why risk it to avoid a few CoInit calls?
-What if some other code someone else implements ends up calling CoInit with STA on every relevant thread before we get the chance of opening an MTA? This code makes the assumption that won't happen.

ligfx: > How do you know the threading model of all WASAPI COM objects? Plus, we don't know which objects cubeb uses.

You can look it up in the registry. It's sufficient just to know that Firefox runs cubeb in MTA threads, though.

> Implicit MTA can be used yes, but if the thread that created the MTA closes it when another access is ongoing, it's not good. This doesn't happen now, but why risk it to avoid a few CoInit calls?

Because the scoped CoInit calls don't actually solve anything. In your scenario:
1. At least one other thread has a reference to the MTA
2. We call CoInitializeEx before using cubeb
3. All other threads drop reference to the MTA
4. We call CoUninitialize after using cubeb
5. *Now, the MTA is gone. All COM objects created in the MTA might be invalid and cause crashes.*

> What if some other code someone else implements ends up calling CoInit with STA on every relevant thread before we get the chance of opening an MTA? This code makes the assumption that won't happen.

Other code should not be creating STAs in threads in general. It's only needed for very specific UI code (e.g. ControllerInterface getting RawInput window messages) that ends up blocking the entire thread (GetMessage/TranslateMessage/DispatchMessage), so it's unlikely that it would be used in one of the main threads. If it is it's probably an error.

This code does make assumptions, but I think they hold up. In general, a thread that gets a cubeb context will exist at least until its finished with the context.

Other options could be making every thread in Dolphin initialize the MTA, or adding a thread to the cubeb context just to keep a reference to the MTA. I think those are overkill.

Filoppi: https://docs.microsoft.com/en-us/windows/win32/api/audioclient/nf-audioclient-iaudioclient-initialize
Note  In Windows 8, the first use of IAudioClient to access the audio device should be on the STA thread. Calls from an MTA thread may result in undefined behavior.

ligfx: > Note In Windows 8, the first use of IAudioClient to access the audio device should be on the STA thread. Calls from an MTA thread may result in undefined behavior.

[Firefox noted this a while back](https://bugzilla.mozilla.org/show_bug.cgi?id=1103824) and decided "I don't think we've ever seen evidence that this matters. It's possible it's only relevant in configurations we don't exercise (e.g. exclusive mode)."

Since the market share of Windows 8 is ~5% or less (not sure what it is for Dolphin), and Firefox hasn't seen issues, I don't think it's worth trying to handle that.

Source/Core/AudioCommon/CubebUtils.cpp
```diff
@@ -8,14 +8,52 @@
...
+}
+
+// Due to the way Dolphin is, we can't guarantee a cubeb context will be created and destroyed
+// from the same thread, this means that we can call CoInitialize on creation but not on destroy,
+// as we don't want to CoUninitialize a random thread, so before calling DestroyContext(),
+// we need to call CoInitialize ourselves. In case it was the same thread as the one it was created
+// on, the new init will just stack and a ScopedThreadAccess will be popped immediately,
+// if it's another thread, then the original ScopedThreadAccess will be called when the thread is
+// closed, avoiding memory leaks
+thread_local std::vector<CubebUtils::ScopedThreadAccess> scoped_thread_accesses;
```
lioncash, line 55: This should also be marked `static`, given it's only used within this translation unit. This will cause a `-Wmissing-declarations` warning otherwise. We also normally keep all static variables together near the top of the file to make for consistent reading

Filoppi:
> Can you squash the update commits, please?

I did, it's finished.
It builds on MSVC but these settings are needed: https://github.com/dolphin-emu/dolphin/pull/8856/files, I'm guessing that's why the build machine failed.

Source/Core/AudioCommon/CubebUtils.cpp
```diff
@@ -16,6 +16,45 @@
...
 static ptrdiff_t s_path_cutoff_point = 0;
 
+// Due to the way Dolphin is, we can't guarantee a cubeb context will be created and destroyed
+// from the same thread, this means that we can call CoInitialize on creation but not on destroy,
+// as we don't want to CoUninitialize a random thread, so before calling DestroyContext(),
+// we need to call CoInitialize ourselves. In case it was the same thread as the one it was created
+// on, the new init will just stack and a ScopedThreadAccess will be popped immediately,
+// if it's another thread, then the original ScopedThreadAccess will be called when the thread is
+// closed, avoiding memory leaks
+thread_local std::vector<CubebUtils::ScopedThreadAccess> scoped_thread_accesses;
```
lioncash, line 26:
```suggestion
static thread_local std::vector<CubebUtils::ScopedThreadAccess> scoped_thread_accesses;
```

Seems to have been missed when moving the code

ligfx:
@Filoppi Note that `CoUninitialize` will be called too many times, since in GetContext the ScopedAccess objects get copied into the vector and then the original gets destructed. Why are you using a vector anyways? The cubeb context is a singleton, so as written the code to push a ScopedAccess to a vector will only be called once before DestroyContext is called.

Also, did you see my note about how the ScopedAccess calls in the rest of the code don't solve anything? COM needs to exist for the entire lifetime of the objects, not just every time a function is called on them.

Filoppi:
> Also, did you see my note about how the ScopedAccess calls in the rest of the code don't solve anything? COM needs to exist for the entire lifetime of the objects, not just every time a function is called on them.

COM does exist for the entire lifetime, plus it's called around every function to make sure it's initialised on that thread. I'm pretty much mirroring what happens wit the current cubeb implementation. Microsoft says you should initialise com from any thread, so I am not comfortable with any implicit CoInit nor with making any other assumptions about the threads or other CoInit calls.
This works, if you want, you can change the code after the PR, I don't mind, but I don't see the point in avoid 5 CoInit calls that happen when pausing/unpausing the emulation, which could also fail so it would be useful to know.

Source/Core/AudioCommon/CubebUtils.cpp
Filoppi: Fixed, I had forgotten

Filoppi:
Forgot tls_ in front of scoped_thread_accesses

pidgey84:
Hello, firtsly im no sure were to put this but since you talk here about the builds of dolphin i think at least would be ok to post it here, i just test latest dolphin versions on windows and i would need to said that i notice a very high slowdown globally on speed comparing to the stable 5.0 old version. for example when i put donkey kong country returns i can put it on 2k on the old version but on the new one it get various slowdowns in game with the same resolution and options.

Thanks for read.

Filoppi:
> Hello, firtsly im no sure were to put this but since you talk here about the builds of dolphin i think at least would be ok to post it here, i just test latest dolphin versions on windows and i would need to said that i notice a very high slowdown globally on speed comparing to the stable 5.0 old version. for example when i put donkey kong country returns i can put it on 2k on the old version but on the new one it get various slowdowns in game with the same resolution and options.
> 
> Thanks for read.

I'm not sure what your point is.

pidgey84:
Well, my point is to share what I have experienced in going from an older version of the dolphin to a more modern one, and it turned out to be the more modern version much slower than the old one.
Does anyone know why that happens?

Filoppi:
> Well, my point is to share what I have experienced in going from an older version of the dolphin to a more modern one, and it turned out to be the more modern version much slower than the old one.
> Does anyone know why that happens?

Yeah but why on this pull request? You don't even seem to be blaming this PR.

pidgey84:
Ah, it's because I didn't know very well where to put it, I think I said it in my first post here, do you know where I should post on that subject?

Filoppi:
> Ah, it's because I didn't know very well where to put it, I think I said it in my first post here, do you know where I should post on that subject?

I'm pretty sure that the main authors are already aware of any performance downgrades if they are real, I'm not really working on any of that stuff so this is not the place. I don't think you need to do anything. Please don't answer this.

StLouisCPhT (commented):
I know this is an older PR, but I tested it on my M1 Mac for Filoppi. Everything seems to work fine.

Rumi-Larry:
> 
> 
> Screenshot (956)
> trying to build with this PR now throws these errors

Your specs are necessary to diagnose the issue.

Filoppi:
I recall having these errors depending on the version of MSVC and or C++ I used. This PR can in between dolphin upgrading to a new MSVC version, but I believe it built fine in the end, so I'm not sure. I will look into this at some point.

PR #8922, opened 2020-07-01
Timer fixes

Filoppi:
Fixes the timer returning 1 when it should return 0. The case mentioned by the comment does not exist anymore.

Also added a IsRunning function as it was impossible to know whether it had been started or not (I will use it in later PRs but it should be there anyway)

### DIFF
--- a/Source/Core/Common/Timer.cpp
+++ b/Source/Core/Common/Timer.cpp
@@ -118,10 +118,9 @@ void Timer::AddTimeDifference()
 // Get the time elapsed since the Start()
 u64 Timer::GetTimeElapsed()
 {
-  // If we have not started yet, return 1 (because then I don't
-  // have to change the FPS calculation in CoreRerecording.cpp .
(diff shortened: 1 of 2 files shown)


Source/Core/Common/Timer.cpp
```diff
@@ -118,10 +118,9 @@ void Timer::AddTimeDifference()
 // Get the time elapsed since the Start()
 u64 Timer::GetTimeElapsed()
 {
-  // If we have not started yet, return 1 (because then I don't
-  // have to change the FPS calculation in CoreRerecording.cpp .
```
Techjar, line 122: Any idea what this old comment was about?

Filoppi: I haven't looked at the history, but I've checked all the current uses of this function and none of them is affected by this change.

Tilka: I checked, the code that used to rely on this is gone.

PR #8923, opened 2020-07-01
fixes for "On Display Messages"

Filoppi:
Mainly fixed 2 things on "On Display Messages":
-They might have never drawn if DrawMessages wasn't called before they actually expired.
-Their fade was wrong if the duration of the message was less than the fade time.

This makes them much more useful for debugging, I know there might be other means of debugging like logs and imgui, but this was the simplest so that's what I used.
If you want to print the same message every frame, but with a slightly different value to see the changes, it now work.

To compensate for the fact that they are now always rendered once, so on start up a lot of old messages (printed while the emulation was off) could show up, I've added a "drop" time, which means if a msg isn't rendered for the first time within that time, it will be dropped and never rendered.

A previous solution was to clean all the messages before the emulation thread started, but that might have had side effects.

### DIFF
--- a/Source/Core/VideoCommon/OnScreenDisplay.cpp
+++ b/Source/Core/VideoCommon/OnScreenDisplay.cpp
@@ -21,9 +21,11 @@
 
 namespace OSD
 {
-constexpr float LEFT_MARGIN = 10.0f;    // Pixels to the left of OSD messages.
-constexpr float TOP_MARGIN = 10.0f;     // Pixels above the first OSD message.
-constexpr float WINDOW_PADDING = 4.0f;  // Pixels between subsequent OSD messages.
+constexpr float LEFT_MARGIN = 10.0f;         // Pixels to the left of OSD messages.
+constexpr float TOP_MARGIN = 10.0f;          // Pixels above the first OSD message.
+constexpr float WINDOW_PADDING = 4.0f;       // Pixels between subsequent OSD messages.
+constexpr float MESSAGE_FADE_TIME = 1000.f;  // Ms to fade OSD messages at the end of their life.
+constexpr float MESSAGE_DROP_TIME = 5000.f;  // Ms to drop OSD messages that has yet to ever render.
 
 static std::atomic<int> s_obscured_pixels_left = 0;
 static std::atomic<int> s_obscured_pixels_top = 0;
@@ -31,12 +33,14 @@ static std::atomic<int> s_obscured_pixels_top = 0;
 struct Message
 {
   Message() = default;
-  Message(std::string text_, u32 timestamp_, u32 color_)
-      : text(std::move(text_)), timestamp(timestamp_), color(color_)
+  Message(std::string text_, u32 timestamp_, u32 duration_, u32 color_)
+      : text(std::move(text_)), timestamp(timestamp_), duration(duration_), color(color_)
   {
   }
   std::string text;
   u32 timestamp = 0;
+  u32 duration = 0;
+  bool ever_drawn = false;
   u32 color = 0;
 };
 static std::multimap<MessageType, Message> s_messages;
@@ -50,7 +54,7 @@ static ImVec4 ARGBToImVec4(const u32 argb)
                 static_cast<float>((argb >> 24) & 0xFF) / 255.0f);
 }
 
-static float DrawMessage(int index, const Message& msg, const ImVec2& position, int time_left)
+static float DrawMessage(int index, Message& msg, const ImVec2& position, int time_left)
 {
   // We have to provide a window name, and these shouldn't be duplicated.
   // So instead, we generate a name based on the number of messages drawn.
@@ -60,9 +64,10 @@ static float DrawMessage(int index, const Message& msg, const ImVec2& position, 
   ImGui::SetNextWindowPos(position);
   ImGui::SetNextWindowSize(ImVec2(0.0f, 0.0f));
 
-  // Gradually fade old messages away.
-  const float alpha = std::min(1.0f, std::max(0.0f, time_left / 1024.0f));
-  ImGui::PushStyleVar(ImGuiStyleVar_Alpha, alpha);
+  // Gradually fade old messages away (except in their first frame)
+  const float fade_time = std::max(std::min(MESSAGE_FADE_TIME, (float)msg.duration), 1.f);
+  const float alpha = std::clamp(time_left / fade_time, 0.f, 1.f);
+  ImGui::PushStyleVar(ImGuiStyleVar_Alpha, msg.ever_drawn ? alpha : 1.0);
 
   float window_height = 0.0f;
   if (ImGui::Begin(window_name.c_str(), nullptr,
@@ -80,6 +85,8 @@ static float DrawMessage(int index, const Message& msg, const ImVec2& position, 
   ImGui::End();
   ImGui::PopStyleVar();
 
+  msg.ever_drawn = true;
+
   return window_height;
 }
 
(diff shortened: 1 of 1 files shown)


JosJuice:
@dolphin-emu-bot rebuild

Source/Core/Core/Core.cpp
```diff
@@ -414,6 +414,9 @@ static void FifoPlayerThread(const std::optional<std::string>& savestate_path,
 // See the BootManager.cpp file description for a complete call schedule.
 static void EmuThread(std::unique_ptr<BootParameters> boot, WindowSystemInfo wsi)
 {
+  // Clear messages printed before the window opened
```
lioncash, line 417: This should probably explain why, rather than what you're doing. It's easy enough to see that clearing messages is being performed because the name of the function is literally `ClearMessages()`

Filoppi: Because they printed before the window opened (and they are outdated)?

JMC47:
@dolphin-emu-bot rebuild

AdmiralCurtiss:
We should check if our various start-of-emulation-warning messages still work with this. Things like, GPU doesn't support important feature or GCI failed to load come to mind.

Filoppi:
> We should check if our various start-of-emulation-warning messages still work with this. Things like, GPU doesn't support important feature or GCI failed to load come to mind.

If you don't want to test, I can either get rid of the clean before the window open, at worse, you are going to see a few messages on screen for 1 frame, which is not a big deals, but from my tests, I don't remember any appearing.
Alternatively, we can put in a "force expire" timer, meaning that if more than, let's say, a second has elapsed from the moment that message was queued, then we discard (just like it would have been before) and not even render it once.

Filoppi:
I've implemented it in the alternative way that I had already described. The explanation is above.
This change should now have no possible negative consequences.

Source/Core/VideoCommon/OnScreenDisplay.cpp
```diff
@@ -110,10 +117,12 @@ void DrawMessages()
...
   for (auto it = s_messages.begin(); it != s_messages.end();)
   {
-    const Message& msg = it->second;
+    Message& msg = it->second;
     const int time_left = static_cast<int>(msg.timestamp - now);
 
-    if (time_left <= 0)
+    // Make sure we draw them at least once if they were printed with 0ms,
+    // unless enough time has expired, in that case, we drop them
+    if (time_left <= 0 && (msg.ever_drawn || -time_left >= MESSAGE_DROP_TIME))
```
iwubcode, line 125: Instead of this `ever_drawn` variable which means a couple locations have to be non-const, why not always draw the message first before seeing if it needs to be removed?

Could use a helper function to make it a little cleaner.

Ex:

```
void DrawMessages()
{
    const bool draw_messages = Config::Get(Config::MAIN_OSD_MESSAGES);
    if (!draw_messages)
      return;
    
  const u32 now = Common::Timer::GetTimeMs();
  const float current_x =
      LEFT_MARGIN * ImGui::GetIO().DisplayFramebufferScale.x + s_obscured_pixels_left;
  float current_y = TOP_MARGIN * ImGui::GetIO().DisplayFramebufferScale.y + s_obscured_pixels_top;
  int index = 0;
  std::lock_guard lock{s_messages_mutex};
  
    for (auto it = s_messages.begin(); it != s_messages.end();)
    {
      const Message& msg = it->second;
      const int time_left = static_cast<int>(msg.timestamp - now);
      DoDrawMessage(time_left, msg, index, current_x, current_y);
      
      if (time_left <= 0)
      {
        it = s_messages.erase(it);
      }
      else
      {
        ++it;
      }
    }
}

void DoDrawMessage(int time_left, const Message& msg, int& index, float current_x, float& current_y)
{
  if (-time_left >= MESSAGE_DROP_TIME)
    return;

  current_y += DrawMessage(index++, msg, ImVec2(current_x, current_y), time_left);
}
```

Filoppi: That would mean that all the expired messages draw one time more than necessary (which also mean they wouldn't follow the length of time you had specified). I think it's better to have a variable non const.

iwubcode: I agree with you about your concerns with my suggestion.  Still not a fan of that variable.  What do you think about this - add another timestamp `s_last_drawn_timestamp`.  Then your if check can be changed slightly:

```
if (time_left <= 0 && (msg.timestamp <= s_last_drawn_timestamp || -time_left >= MESSAGE_DROP_TIME))
```

At the end of the function:

```
s_last_drawn_timestamp = now;
```

Filoppi: Hypothetical case but that would break when going over ~1000fps though (time accuracy is 1ms):
Add a message: timestamp 53 (53ms current time + 0 duration).
Draw messages at time 53ms.
Add another message which is exactly the same: timestamp 53 (if we are still within the same ms).
Draw messages at time 53ms again: the previous message nor the new message will draw now.

iwubcode: That's true.  Sorry, my initial comment was not well thought out.  However, still not a fan of that variable so I have a different suggestion.

If I'm understanding things correctly, I think what you're trying to account for are messages that get missed because they expire sometime before `DrawMesages()` gets called (or in the case of a start/stop of the software expire sometime in between `DrawMessages()` calls).

I think a solution to that problem is to add another timestamp.

```
u32 s_last_drawn_timestamp = 0;

void DrawMessages()
{
  if (time_left <= 0 && (s_last_drawn_timestamp > msg.timestamp || -time_left >= MESSAGE_DROP_TIME))
  ...
  s_last_drawn_timestamp = now;
}
```

Filoppi: Yeah with > instead of >= it might work. Thinking about it.

Filoppi: changing >= to > has the opposite effect of the problem I mentioned above. If going to extremely high fps, they could now draw more than once. Which is not really a problem, but is theoretically wrong, and given that with the ever_drawn flag it works perfectly, I don't think it would be a worth changing.

iwubcode: > If edits are requested for such small things none of my other changes will ever go it, not within 3+ years 

It's just an unfortunate consequence of this being an open-source repo with very few developers doing constant reviews.  Because of that, code changes can take a while.  Only thing that gets merged fast are code-cleanup changes and even that can sometimes take a while depending on the size.

>  If going to extremely high fps, they could now draw more than once.

I personally think that this is a fine trade-off for removing that variable (that is included on every message) and returning `const` correctness.  When you get to that high framerate, timing accuracy isn't going to matter that much.  If we really want to fix this, we need a more granular clock.  I'm going to guess we aren't using `std::chrono` in `Timer` because it didn't exist back then.

Filoppi: Well on screen messages are not logs, there are barely any, what's the problem with adding a bool? And it's not like it's not const correct, ever_drawn is changed by the Draw() function. That kind of makes sense.

iwubcode: I'm definitely nitpicking.  Sorry to take up so much of your time.  It just felt counterintuitive for `DrawMessage` to be changing `Message` even though I see your reason for doing so.  My hope was to come up with a solution that would allow for the `Message` to be `const` again.

Anyway, I don't have merge powers, so someone else will need to review this regardless.  My apologies again.  These are good changes overall.

JMC47:
This has been sitting for a while and it seems like everyone here is in agreement that this is an improvement.  I've gone through and tested it and on-screen display messages at startup are fine.  I tested purposefully duplicate GCI files for something that would cause issues on boot.

PR #8925, opened 2020-07-02
DolphinQt: Allow customizing TAS input turbo interval

JosJuice:
As a side effect of PR #8347, Dolphin's frame counter was changed to run at 60/50 Hz even if the game is running at a lower framerate such as 30 fps. Since the TAS input turbo button functionality toggled the state of a button every other frame as reported by the frame counter, this change made the turbo button functionality not work with 30/25 fps games.

I believe it would be hard to change the frame counter back to how it used to work without undermining the point of PR #8347, and I'm not sure if doing so would be desireable or not anyway, so what I'm doing instead is letting the user determine how long turbo button presses should last. This lets users avoid the 30/25 fps game problem while also granting additional flexibility. Perhaps there is some game where it is useful to mash at a speed which is slower than frame perfect.


Blunderstab:
> Perhaps there is some game where it is useful to mash at a speed which is slower than frame perfect.

As a matter of fact, Donkey Kong: Jungle Beat has this quirk! I'll be excited to check out this improvement. <3

PR #8926, opened 2020-07-02
DolphinQt: Expose SYSCONF sound setting

JosJuice:


Source/Core/DolphinQt/Settings/WiiPane.cpp
```diff
@@ -126,6 +130,12 @@ void WiiPane::CreateMisc()
   m_system_language_choice->addItem(tr("Traditional Chinese"));
   m_system_language_choice->addItem(tr("Korean"));
 
+  m_sound_mode_choice_label = new QLabel(tr("Sound:"));
+  m_sound_mode_choice = new QComboBox();
```
Filoppi, line 134: Might need a tooltip explaining that with HLE there is no difference between Stereo and Surround

Filoppi: And also that this needs to be set to Surround to even enable DPLII in the Audio panel

JosJuice: Shouldn't that rather be explained where the DSP HLE and DPLII decoder settings are? This is not the only way to control whether games output surround sound, in particular not for GameCube games, for which this has no effect.

Filoppi: You definitely know better than me. As long as it's somewhere, it's fine. There are a lot of settings in Dolphin and for a new user it isn't easy.

PR #8927, opened 2020-07-03
Fixed volume of emulated wii mote speakers not being native.

Filoppi:
As in the code comment, the panning was denaturalizing the volume (when panning was set to 0).
Unless users actually want to use panning, their wii mote volume should not be reduced, it should come out at the same volume the samples were, without using the "Constant Power Pan Law" as we can't assume the original wii mote speakers loadness is comparable to the loudness of the audio device you are currently using (in fact, it's likely to be lower).
If users want to reduce the volume of the wii mote speakers, they can do so from the wii home menu.
I opted for this instead of adding another setting "Enable Panning" as it would have been confusing, and the changes in the panning formula are unlikely to have any negative effect, as it still works.

### DIFF
--- a/Source/Core/Core/HW/WiimoteEmu/Speaker.cpp
+++ b/Source/Core/Core/HW/WiimoteEmu/Speaker.cpp
@@ -128,16 +128,21 @@ void SpeakerLogic::SpeakerData(const u8* data, int length, float speaker_pan)
   }
 
   // SetWiimoteSpeakerVolume expects values from 0 to 255.
-  // Multiply by 256, cast to int, and clamp to 255 for a uniform conversion.
-  const double volume = float(reg_data.volume) / volume_divisor * 256;
+  // Multiply by 256, floor to int, and clamp to 255 for a uniformly mapped conversion.
+  const double volume = float(reg_data.volume) * 256.f / volume_divisor;
 
-  // Speaker pan using "Constant Power Pan Law"
-  const double pan_prime = MathUtil::PI * (speaker_pan + 1) / 4;
+  // This used the "Constant Power Pan Law", but it is undesirable
+  // if the pan is 0, and it implied that the loudness of a wiimote speaker
+  // is equal to the loudness of your device speakers, which isn't true at all.
+  // This way, if the pan is 0, it's like if it is not there.
+  // We should play the samples from the wiimote at the native volume they came with,
+  // because you can lower their volume from the Wii settings and because they are
+  // already extremely low quality, so any additional quality loss isn't welcome.
+  speaker_pan = std::clamp(speaker_pan, -1.f, 1.f);
+  const u32 l_volume = std::min(u32(std::min(1.f - speaker_pan, 1.f) * volume), 255u);
+  const u32 r_volume = std::min(u32(std::min(1.f + speaker_pan, 1.f) * volume), 255u);
 
-  const auto left_volume = std::min(int(std::cos(pan_prime) * volume), 255);
-  const auto right_volume = std::min(int(std::sin(pan_prime) * volume), 255);
-
-  g_sound_stream->GetMixer()->SetWiimoteSpeakerVolume(left_volume, right_volume);
+  g_sound_stream->GetMixer()->SetWiimoteSpeakerVolume(l_volume, r_volume);
 
   // ADPCM sample rate is thought to be x2.(3000 x2 = 6000).
   const unsigned int sample_rate = sample_rate_dividend / reg_data.sample_rate;


Filoppi:
This also fixes the max volume of the wii mote, which was always 1/255 lower than it could have been.

JosJuice:
I don't see how the "it implied that the loudness of a wiimote speaker is equal to the loudness of your device speakers" part makes sense. The constant power pan law is about making sure that the audio doesn't change in perceived loudness as you change the panning, not about making sure that the audio is the same in perceived loudness as the audio of a real Wii Remote speaker.

I think it's best to keep using the constant power pan law if we allow panning. But I do also think that the utility of letting the user pan the Wii Remote speaker audio is limited.

Filoppi:
Yes I understand that the pan law makes sense within panning, but it doesn't make sense if you don't want panning.
I don't want panning, and with panning at 0, Dolphin is assuming it needs to normalize the perceived volume of the wii mote speaker, so that I would hear it with the same volume as if it came from a single speaker. But that single speaker would have been a wii mote speaker, which isn't comparable to your device speakers/headphones at all, so not adjusting the volume at all is better than adjusting towards something potentially wrong.

Filoppi:
@JosJuice I discussed with multiple people yesterday, and they actually had one good use case for panning. For example, if multiple people were playing multplayer with 2 normal controllers on Dolphin, they could each pan their wii mote speaker sounds to the speaker on their side, so the direction of the wii mote sounds would be more clear.
This came out after I proposed to get rid of it altogether.

Panning still works fine without the pan law, so I don't think taking out the pan law is a problem. If a user wants panning, they mainly want it to have the sounds more or less loud on one side, and they can still achieve that, there is no need to adjust the perceived loudness towards a random direction which we can't know is right.

jordan-woyak:
The "pan law" is fun because it's cool to be mathematically correct but I think I might agree that it's not practical here.
95% of the time people will use pan `0` and basically everyone else will use a full `-1` or `+1`.
We might as well just push the samples at full volume.

That being said your math looks a bit wrong.
You have exactly the same calculation for `left_volume` and `right_volume`.
`std::min` shouldn't be needed after using `std::clamp`.
I think multiplying by 256 and rounding-down should be more uniform than multiplying by 255 and rounding-to-nearest.

Filoppi:
My CTRL key broke and CTRL+S doesn't work anymore, so I missed the last save where I fixed the math left/right

Filoppi:
> The "pan law" is fun because it's cool to be mathematically correct but I think I might agree that it's not practical here.
> 95% of the time people will use pan `0` and basically everyone else will use a full `-1` or `+1`.
> We might as well just push the samples at full volume.
> 
> That being said your math looks a bit wrong.
> You have exactly the same calculation for `left_volume` and `right_volume`.
> `std::min` shouldn't be needed after using `std::clamp`.
> I think multiplying by 256 and rounding-down should be more uniform than multiplying by 255 and rounding-to-nearest.

No, clamp is very much not needed, but I put it there to be safe in case someone passed a pan over 1 or below -1, in that case the volume result would loop around.

JosJuice:
If others agree on not wanting to keep the pan law, you don't have to treat my comment as a blocker.

JosJuice:
@dolphin-emu-bot rebuild

JosJuice:
@dolphin-emu-bot rebuild

Source/Core/Core/HW/WiimoteEmu/Speaker.cpp
```diff
@@ -128,16 +128,21 @@ void SpeakerLogic::SpeakerData(const u8* data, int length, float speaker_pan)
...
+  // This used the "Constant Power Pan Law", but it is undesiderable
+  // if the pan is 0, and it implied that the loudness of a wiimote speaker
+  // is equal to the loudness of your device speakers, which isn't true at all.
+  // This way, if the pan is 0, it's like if it is not there.
+  // We should play the samples from the wiimote at the native volume they came with,
+  // because you can lower their volume from the wii settings, and because they are
+  // already extremely low quality, so any additional quality loss isn't welcome.
+  speaker_pan = std::clamp(speaker_pan, -1.f, 1.f);
+  u32 l_volume = std::min(u32(std::min(1.f - speaker_pan, 1.f) * volume), 255u);
+  u32 r_volume = std::min(u32(std::min(1.f + speaker_pan, 1.f) * volume), 255u);
```
jordan-woyak, line 143: These can be `const`. Squash the two commits then this looks good to me.

Filoppi:
> @JosJuice I discussed with multiple people yesterday, and they actually had one good use case for panning. For example, if 2 people were playing multplayer with 2 normal controllers on Dolphin, they could each pan their wii mote speaker to the PC speaker on their side, so the direction of the wii mote sounds would be clear.
> This came out after I proposed to get rid of it altogether.
> 
> Panning still works fine without the pan law, so I don't think taking out the pan law is a problem. If a user wants panning, they mainly want it to have the sounds more or less loud on one side, and they can still achieve that, there is no need to adjust the perceived loudness towards a random direction which we can't know is right.

Thinking about it, when emulating 2 wii mote simultaneously, they won't be able able to play sound at the same time, it would cripple both sounds as it would play a few samples from the first, then a few samples from the second and so on. I don't see how it could work with the current code. Which make my above comment "wrong", so we don't really have any reason to keep the padding. Unless I add support to play 2 wii mote sounds simultaneously, which I'm looking to do.

Filoppi:
Worth mentioning here: if DPLII is enabled and panning is 0, the wii mote samples will only come out of the central channel because samples that are identical between the left and right channel are rerouted to the central speaker.

jordan-woyak (approved):

PR #8932, opened 2020-07-04
PPCSymbolDB: Improve symbol map column detection

sepalani:
This PR fixes https://dolp.in/i11995.

Ready to be review & merged.

### DIFF
--- a/Source/Core/Core/PowerPC/PPCSymbolDB.cpp
+++ b/Source/Core/Core/PowerPC/PPCSymbolDB.cpp
@@ -7,6 +7,7 @@
 #include <algorithm>
 #include <cstring>
 #include <map>
+#include <sstream>
 #include <string>
 #include <string_view>
 #include <utility>
@@ -243,12 +244,6 @@ bool PPCSymbolDB::LoadMap(const std::string& filename, bool bad)
     if (length < 4)
       continue;
 
-    if (length == 34 && strcmp(line, "  address  Size   address  offset\n") == 0)
-    {
-      column_count = 4;
-      continue;
-    }
-
     char temp[256]{};
     sscanf(line, "%255s", temp);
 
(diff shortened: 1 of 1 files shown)


Source/Core/Core/PowerPC/PPCSymbolDB.cpp
```diff
@@ -296,14 +291,37 @@ bool PPCSymbolDB::LoadMap(const std::string& filename, bool bad)
...
+      constexpr auto is_hex_str = [](const std::string& s) {
+        return !s.empty() && s.find_first_not_of("0123456789abcdefABCDEF") == std::string::npos;
+      };
+      const std::string stripped_line(StripSpaces(line));
+      std::istringstream oss(stripped_line);
+      std::string word;
+
+      // Two columns format:
+      // 80004000 zz_80004000_
+      if (!(oss >> word) || word.length() != 8 || !is_hex_str(word))
```
iwubcode, line 306: Did you consider `isxdigit` coupled with `std::all_of` ?  Not sure if there are any advantages to that.

Source/Core/Core/PowerPC/PPCSymbolDB.cpp
```diff
@@ -296,14 +291,37 @@ bool PPCSymbolDB::LoadMap(const std::string& filename, bool bad)
...
+        continue;
+      column_count = 2;
```
iwubcode, line 308: Would it be faster to tokenize the string by spaces, then later validate whether they are hex or not?

Source/Core/Core/PowerPC/PPCSymbolDB.cpp
```diff
@@ -296,14 +291,37 @@ bool PPCSymbolDB::LoadMap(const std::string& filename, bool bad)
...
+
+      // Three columns format:
+      //  Starting        Virtual
+      //  address  Size   address
+      //  -----------------------
+      if (oss && oss >> word && is_hex_str(word) && oss >> word && is_hex_str(word))
```
iwubcode, line 314: Sorry if a dumb question but I want to check my understanding.

The first word has been read, that's a hex value.  If the second value is read here, it's a size, and not a hex string.  The third value is a hex string.  Shouldn't the first `is_hex_str` in this check be a `!is_hex_str` ?

Source/Core/Core/PowerPC/PPCSymbolDB.cpp
sepalani: I did and I need to wrap `isxdigit(int)` around a lambda with an unsigned cast to avoid an assert error (`char` might be signed and negative) or use the locale version of it. However, I don't know if `isxdigit` is faster or not.

Source/Core/Core/PowerPC/PPCSymbolDB.cpp
sepalani: All columns are hexstring: address, size and vaddrress.

iwubcode: Ohhh.  Thank you for the clarification

Source/Core/Core/PowerPC/PPCSymbolDB.cpp
sepalani: I don't think it will be faster since we don't need to tokenize the whole line if the first word isn't a hexstring. If I understand correctly it will also require a container like a vector to store the string tokens and then process them which might be slower.

Source/Core/Core/PowerPC/PPCSymbolDB.cpp
```diff
@@ -296,14 +291,37 @@ bool PPCSymbolDB::LoadMap(const std::string& filename, bool bad)
...
     {
-      const std::string_view stripped_line = StripSpaces(line);
-      if (std::count(stripped_line.begin(), stripped_line.end(), ' ') == 1)
-        column_count = 2;
-      else
+      constexpr auto is_hex_str = [](const std::string& s) {
+        return !s.empty() && s.find_first_not_of("0123456789abcdefABCDEF") == std::string::npos;
+      };
+      const std::string stripped_line(StripSpaces(line));
+      std::istringstream oss(stripped_line);
```
jordan-woyak, line 301: You must `oss.imbue(std::locale::classic())` to prevent formatted reads from traversing whitespace with certain locales.
See #8542.

sepalani:
@jordan-woyak Good catch, done.

Rebased on latest master and I renamed `oss` to `iss`.

sepalani:
Just a simple rebase. I've tested it against the map files I have and didn't find any regression.

leoetlino (approved):

PR #8935, opened 2020-07-05
DVDInterface: Increase the latency for read commands

JosJuice:
Hardware tested using cached reads.

Fixes Ed, Edd n Eddy (untested). https://bugs.dolphin-emu.org/issues/10373

### DIFF
--- a/Source/Core/Core/HW/DVD/DVDInterface.cpp
+++ b/Source/Core/Core/HW/DVD/DVDInterface.cpp
@@ -43,9 +43,11 @@
 
 #include "VideoCommon/OnScreenDisplay.h"
 
-// The minimum time it takes for the DVD drive to process a command (in
-// microseconds)
-constexpr u64 COMMAND_LATENCY_US = 300;
+// The minimum time it takes for the DVD drive to process a command (in microseconds)
+constexpr u64 MINIMUM_COMMAND_LATENCY_US = 300;
+
+// The time it takes for a read command to start (in microseconds)
+constexpr u64 READ_COMMAND_LATENCY_US = 600;
 
 // The size of the streaming buffer.
 constexpr u64 STREAMING_BUFFER_SIZE = 1024 * 1024;
@@ -1182,10 +1184,10 @@ void ExecuteCommand(ReplyType reply_type)
 
   if (!command_handled_by_thread)
   {
-    // TODO: Needs testing to determine if COMMAND_LATENCY_US is accurate for this
-    CoreTiming::ScheduleEvent(COMMAND_LATENCY_US * (SystemTimers::GetTicksPerSecond() / 1000000),
-                              s_finish_executing_command,
-                              PackFinishExecutingCommandUserdata(reply_type, interrupt_type));
+    // TODO: Needs testing to determine if MINIMUM_COMMAND_LATENCY_US is accurate for this
+    CoreTiming::ScheduleEvent(
+        MINIMUM_COMMAND_LATENCY_US * (SystemTimers::GetTicksPerSecond() / 1000000),
+        s_finish_executing_command, PackFinishExecutingCommandUserdata(reply_type, interrupt_type));
   }
 }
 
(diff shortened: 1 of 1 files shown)


Source/Core/Core/HW/DVD/DVDInterface.cpp
```diff
@@ -1378,8 +1379,8 @@ void ScheduleReads(u64 offset, u32 length, const DiscIO::Partition& partition, u
             "Schedule reads: offset=0x%" PRIx64 " length=0x%" PRIx32 " address=0x%" PRIx32, offset,
             length, output_address);
 
-  // The DVD drive's minimum turnaround time on a command, based on a hardware test.
```
iwubcode, line 1381: This line seems to indicate there was a hardware test before.  Any idea what changed?

JosJuice: If I'm understanding it correctly, the earlier hardware test determined that 300 µs was the fastest that *any* command can complete. This is a more complex command, so it makes sense that its latency would be higher than the latency of the simplest command.

iwubcode: Hmm, I guess we can only ever approximate this.  What effect does this have on games that don't need this change? Longer load times?

Tilka:
Tested and confirmed working. Previously the loading screen for the first level would get stuck and now I get into the first level.

Source/Core/Core/HW/DVD/DVDInterface.cpp
JosJuice: Yes, but just a tiny bit.

Pokechu22:
I can confirm that this allows getting into the first level; however, after getting into the first level, if I then pause the game, select options (by pressing right), and then select "Exit to Cul-De-Sac", the game freezes.  The following is put in the log (I don't always get the invalid read part):

```
21:00:967 Core\PowerPC\PPCCache.cpp:192 I[PowerPC]: ICache read at 00da813c returned stale data: CACHED: 83c10020 vs. RAM: 9421ff68
21:00:967 Core\PowerPC\PPCCache.cpp:192 I[PowerPC]: ICache read at 00da8140 returned stale data: CACHED: 83e10024 vs. RAM: 7c0802a6
21:00:967 Core\PowerPC\PPCCache.cpp:192 I[PowerPC]: ICache read at 00da8144 returned stale data: CACHED: 38210028 vs. RAM: 91c10050
21:00:967 Core\PowerPC\PPCCache.cpp:192 I[PowerPC]: ICache read at 00da8148 returned stale data: CACHED: 4e800020 vs. RAM: 91e10054
21:00:968 Core\PowerPC\Jit64\Jit.cpp:784 W[PowerPC]: ISI exception at 0x6f693a70
21:00:978 Common\MsgHandler.cpp:115 E[MASTER]: Question: Invalid read from 0x64796e67, PC = 0x80144568 
21:32:390 Common\MsgHandler.cpp:115 E[MASTER]: Question: Invalid read from 0x64796e6b, PC = 0x80144570 
21:33:251 Common\MsgHandler.cpp:115 E[MASTER]: Question: Invalid read from 0x64796e67, PC = 0x80144580 
```

It also freezes upon selecting "Exit to main menu" and then loading the game again (though it does play, or at least start, the cutscene), with just this:

```
16:46:449 Core\PowerPC\PPCCache.cpp:192 I[PowerPC]: ICache read at 00dac280 returned stale data: CACHED: 4e800020 vs. RAM: 3d2080d7
```

This icache behavior is the same as with Scooby Doo! Mystery Mayhem ([Bug 9561](https://bugs.dolphin-emu.org/issues/9561), a false positive where the existing icache check is returning stale data, when it probably shouldn't).  I'll test.

Source/Core/Core/HW/DVD/DVDInterface.cpp
```diff
@@ -43,9 +43,11 @@
 
 #include "VideoCommon/OnScreenDisplay.h"
 
-// The minimum time it takes for the DVD drive to process a command (in
-// microseconds)
-constexpr u64 COMMAND_LATENCY_US = 300;
+// The minimum time it takes for the DVD drive to process a command (in microseconds)
+constexpr u64 MINIMUM_COMMAND_LATENCY_US = 300;
```
Rumi-Larry, line 47: I get that the `_US`, is supposed to stand for microseconds (µs), but this is redundant, since it's already explained in the comment. Like it is, it can be confused for "United States", as if this latency is specific to US consoles. I recommend that part of the name to be omitted, reading `MINIMUM_COMMAND_LATENCY`

JosJuice: It's only redundant if you're looking at this specific line. The suffix is useful when you see the variable name referenced in the code, as it explains why we're dividing by 1000000.

Rumi-Larry: Got it!

lioncash (approved):

PR #8936, opened 2020-07-05
ControllerInterface/Wiimote: Provide fallback values for extensions with bad calibration.

jordan-woyak:
Some Nunchuks and Classic Controllers have bad calibration data.
This was causing unusable inputs in ControllerInterface Wii Remotes (aka Hybrid Wiimote 2.0).

This PR adds fallback values to make things mostly usable if the calibration checksum is bad or values are not sane.
The sanity check is just that "min" and "max" values are on opposite sides of the "zero" value.

This fixes:
https://bugs.dolphin-emu.org/issues/12088
https://bugs.dolphin-emu.org/issues/12027

### DIFF
--- a/Source/Core/InputCommon/ControllerEmu/ControllerEmu.h
+++ b/Source/Core/InputCommon/ControllerEmu/ControllerEmu.h
@@ -14,6 +14,7 @@
 #include "Common/BitUtils.h"
 #include "Common/Common.h"
 #include "Common/IniFile.h"
+#include "Common/MathUtil.h"
 #include "InputCommon/ControlReference/ExpressionParser.h"
 #include "InputCommon/ControllerInterface/Device.h"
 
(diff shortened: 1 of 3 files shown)


CR96:
Just tested this locally against Build 12259 with Super Mario Galaxy, my Nunchuk works with Hybrid Wiimote now!

(I'm musicalmitten from the linked bug tracker issue)

Source/Core/InputCommon/ControllerEmu/ControllerEmu.h
```diff
@@ -37,6 +38,23 @@ struct TwoPointCalibration
...
 
+  constexpr bool IsSane() const
+  {
+    if constexpr (std::is_arithmetic_v<T>)
+    {
+      return max != zero;
+    }
+    else
+    {
+      for (size_t i = 0; i != std::size(zero.data); ++i)
```
lioncash, line 49: This can likely be a `std::all_of`

jordan-woyak: I'm indexing two arrays. I don't think it can be a clean `std::all_of` but I'll change it to an `std::equal`.

PR #8937, opened 2020-07-05
Disable ICache emulation for some games

Pokechu22:
Specifically, 'Scooby-Doo! Mystery Mayhem' ([Bug 9561](https://bugs.dolphin-emu.org/issues/9561)) and 'Ed, Edd n Eddy: The Mis-Edventures'.

For Scooby-Doo!, I tested by loading the credits under the extras menu.  Prior to this change, it would fail to load; now, it loads successfully (even on exiting and re-entering the credits).  I haven't tested much actual gameplay.  I was also unable to confirm that there was ever an issue with loading save files; saving and loading at the first camera works fine in 5.0-12247.

For Ed, Edd n Eddy, the fix is only meaningful when combined with #8935; see https://github.com/dolphin-emu/dolphin/pull/8935#issuecomment-653928322.  (**EDIT**: It also is only necessary when MMU is disabled; the icache issue doesn't come up with MMU enabled.)

There may be other games that need this fix (for instance, I see [Bug 10177](https://bugs.dolphin-emu.org/issues/10177) for Scooby Doo! Unmasked).  I don't own that game, so I don't know if it's a similar issue.

### DIFF
--- a/Source/Core/Core/ConfigManager.cpp
+++ b/Source/Core/Core/ConfigManager.cpp
@@ -512,6 +512,7 @@ void SConfig::LoadCoreSettings(IniFile& ini)
   core->Get("LowDCBZHack", &bLowDCBZHack, false);
   core->Get("FPRF", &bFPRF, false);
   core->Get("AccurateNaNs", &bAccurateNaNs, false);
+  core->Get("DisableICache", &bDisableICache, false);
   core->Get("EmulationSpeed", &m_EmulationSpeed, 1.0f);
   core->Get("Overclock", &m_OCFactor, 1.0f);
   core->Get("OverclockEnable", &m_OCEnable, false);
@@ -749,6 +750,7 @@ void SConfig::LoadDefaults()
   bFastmem = true;
   bFPRF = false;
   bAccurateNaNs = false;
+  bDisableICache = false;
   bMMU = false;
   bLowDCBZHack = false;
   iBBDumpPort = -1;
--- a/Data/Sys/GameSettings/G5D.ini
+++ b/Data/Sys/GameSettings/G5D.ini
@@ -1,4 +1,4 @@
-# GE9E5D - Ed, Edd n Eddy
+# G5DE78, G5DP78 - Scooby-Doo! Unmasked
 
 [Core]
 # Values set here will override the main Dolphin settings.
@@ -15,7 +15,3 @@ HLE_BS2 = True
 
 [ActionReplay]
 # Add action replay cheats here.
-
-[Video_Settings]
-SafeTextureCacheColorSamples = 512
-
--- a/Data/Sys/GameSettings/GC3.ini
+++ b/Data/Sys/GameSettings/GC3.ini
@@ -1,4 +1,4 @@
-# GE9E5D - Ed, Edd n Eddy
+# GC3F78, GC3P78, GC3D78, GC3E78 - Scooby-Doo! Mystery Mayhem
 
 [Core]
 # Values set here will override the main Dolphin settings.
@@ -15,7 +15,3 @@ HLE_BS2 = True
 
 [ActionReplay]
 # Add action replay cheats here.
-
-[Video_Settings]
-SafeTextureCacheColorSamples = 512
-
--- a/Data/Sys/GameSettings/RHF.ini
+++ b/Data/Sys/GameSettings/RHF.ini
@@ -2,3 +2,5 @@
 
 [Video_Settings]
 SuggestedAspectRatio = 2
+# The JIT cache causes problems with emulated icache invalidation in this game resulting in a crash on starting the game
+DisableICache = True
--- a/Source/Core/Core/BootManager.cpp
+++ b/Source/Core/Core/BootManager.cpp
@@ -81,6 +81,7 @@ private:
   bool bAccurateNaNs;
   bool bMMU;
   bool bLowDCBZHack;
+  bool bDisableICache;
   bool m_EnableJIT;
   bool bSyncGPU;
   int iSyncGpuMaxDistance;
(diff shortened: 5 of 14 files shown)


Pokechu22:
The new configuration value was modeled based on existing code for AccurateNaNs.  I didn't add an equivalent of the netplay-related line, though:

https://github.com/dolphin-emu/dolphin/blob/0dbe8fb2eaa608a6540df3d269648a596c29cf4b/Source/Core/Core/BootManager.cpp#L373

I don't know if I need to do something like that (for one, both of the affected games are singleplayer).

JosJuice:
It would be good to do so for completeness (if someone decides to use the setting with a multiplayer game in the future).

Pokechu22:
Done, though I'm not sure if there's some kind of protocol version constant I should update.

hthh:
Superficially this looks like a hack that decreases accuracy to get a game working, but reading the explanation in the issue (that "this is related to JIT caching causing the emulated icache to not get invalidated properly") it probably makes sense... Maybe that explanation could be added as a comment or to the commit message? Would it make sense to ignore this flag on (uncached) Interpreter so that that mode doesn't become less accurate?

Pokechu22:
It's definitely a hack, but I think it's overall more accurate for these games (icache definitely shouldn't be triggering here at least).  A more true fix would be to update icache when entering JIT blocks, but I looked into doing that in the past and didn't see a reasonable way of implementing it.

How about "The JIT cache causes problems with emulatied icache invalidation in this game resulting in areas failing to load" as a comment in the game config?

> Would it make sense to ignore this flag on (uncached) Interpreter so that that mode doesn't become less accurate?

That would probably be a good idea accuracy-wise, but I'm not sure if it's a good idea to add more checks here.  I feel like `ReadInstruction` is pretty hot code.

JosJuice:
> Done, though I'm not sure if there's some kind of protocol version constant I should update.

No. The Dolphin version fetched from git acts as the protocol version.

orbea:
> It's definitely a hack, but I think it's overall more accurate for these games (icache definitely shouldn't be triggering here at least). A more true fix would be to update icache when entering JIT blocks, but I looked into doing that in the past and didn't see a reasonable way of implementing it.

While I understand the desire to get games working, a hack likely could make it that a more correct fix is never implemented.

JMC47:
Even if proper icache emulation were ever finished, it'd likely be a big enough performance penalty that we'd need a hack like this anyway.

orbea:
At least proper icache emulation would exist then.

JMC47:
I'm saying in the case of icache emulation, they may be mutually exclusive.  This *hack* doesn't fix all cases, as the GameCube BIOS still needs proper icache emulation for disc swaps to work.

hthh:
Yeah, I'm not a fan of hacks, but I'm in favour of this change. It doesn't decrease accuracy, just adjusts behavior that's already inaccurate for performance reasons. Proper (or at least proper enough for these games) icache emulation already exists in the interpreter, so we're not losing out in that respect.

That comment sounds good to me.

(And although I'd imagine another well-predicted branch in `ReadInstruction` wouldn't have a measurable performance impact, I guess it's probably good to leave it disabled in interpreter anyway, just to keep the name simple and have it do what it says.)

Source/Core/Core/ConfigManager.cpp
```diff
@@ -214,6 +214,7 @@ void SConfig::SaveCoreSettings(IniFile& ini)
   core->Set("SyncGpuOverclock", fSyncGpuOverclock);
   core->Set("FPRF", bFPRF);
   core->Set("AccurateNaNs", bAccurateNaNs);
+  core->Set("DisableICache", bDisableICache);
```
Techjar, line 217: I don't think we need to save this value to the global config.

Pokechu22: I guess I'm a bit confused as to how the global config works.  Are there other options that are read from it (or game-specific configs only) but not saved?

Pokechu22: Looks like `BBDumpPort`, `FastDiscSpeed`, and `LowDCBZHack` are also read but not saved; at least `LowDCBZHack` seems equivalent so I'll remove the save.

JMC47:
I'd rather enable MMU emulation than disable icache emulation for Ed, Edd, and Eddy.  Unfortunately, I don't think that's an option for Android, so we may have to add the hack regardless.

JMC47:
Can this be rebased so that the disc timing fixes are in for testing further into Ed, Edd, and Eddy?

JMC47:
@SirMangler made a rebased test build for someone playing Ed, Edd, and Eddy and it got some further testing.

Apparently enabling MMU is a really large performance issue for this game even on non-Android and there are more crashes with the icache later in the game.  I'm now in favor of the icache hack for Ed, Edd and Eddy.

JMC47:
So, here's my thoughts on this hack in general.  I relate it most closely to the dcache hack we employ for the Disney Infinity and Disney's Cars 2 malicious anti-emulation code.  While these games aren't doing anything malicious, it appears they have the same bug that is manifesting in a similar way.

Similarly to the dcache lowmem hack, this is a *very* small footprint in the code and is able to handle multiple games.  It also is immediately descriptive of what it is doing when you look in the INI based on the option name.

While it's always a bit annoying to have to use a hack, it's likely dolphin never sees icache emulation accurate enough to fix all of these issues while maintaining decent performance.  And if it does, the fact this is one localized setting makes it extremely easy to see what games would be fixed and compare performance in them.

Note that this isn't a catch all for icache issues, and does not fix other games with more complicated icache issues, such as 1080 Snowboarding.

Miksel12:
Happy Feet on the Wii is also fixed by disabling ICache: https://bugs.dolphin-emu.org/issues/12415

Pokechu22:
It probably would be worth checking if Scooby Doo: Unmasked is also fixed by this (https://bugs.dolphin-emu.org/issues/10177), but I don't plan on buying a second Scooby Doo game I'll never play just to check.

JMC47:
I'll check on it.

JMC47:
I beat the cookie factory in the issue report without experiencing a crash.  Though, I kind of skipped some of the talking in the tutorial because this game has some interesting collision detection and I found it fun to see how far I could get it.  I do not think Scooby Doo Unmasked crashes anymore.

Pokechu22:
I've added this hack to the GameINIs for Scooby-Doo! Unmasked and the Wii version of Happy Feet.

PR #8940, opened 2020-07-06
add Break On Hit and Log On Hit for instruction breakpoints

RenaKunisaki:


Source/Core/Core/PowerPC/PowerPC.cpp
```diff
@@ -603,7 +604,15 @@ void CheckBreakPoints()
 {
   if (PowerPC::breakpoints.IsAddressBreakPoint(PC))
   {
-    CPU::Break();
+    if(PowerPC::breakpoints.IsBreakPointBreakOnHit(PC))
```
sepalani, line 607: `if (`

Source/Core/Core/PowerPC/PowerPC.cpp
```diff
@@ -603,7 +604,15 @@ void CheckBreakPoints()
...
+        CPU::Break();
+    if(PowerPC::breakpoints.IsBreakPointLogOnHit(PC))
```
sepalani, line 609: `if (`

BhaaLseN:
Feel free to combine those two commits; there is very little value in having that second commit stick around as an individual one.

RenaKunisaki:
How do I do that?

BhaaLseN:
Assuming command line (like git bash or similar), you run `git rebase -i upstream/master master` (assuming that `upstream` is the name of the Dolphin remote, use `git remote -v` to check and replace as necessary. It also helps to `git fetch upstream` first, just to be sure.) That should bring up a text editor that has two lines that start with the word `pick`. Change the second one (that should also say "run clang-format" next to a hash) to `fixup`; then save and quit the editor. Git should do a few things, and drop you back into a console when its done. Assuming it went without error, you have to force-push this to your repository (depending on how you set up your branch, `git push -f` might be enough; otherwise try `git push -f origin master` - again, check `git remote -v` for _your own_ repository here on GitHub and replace as necessary).
Let us know if you need any help (or drop by on IRC if thats easier)

RenaKunisaki:
Alright, that should do it.

sepalani:
@dolphin-emu-bot rebuild

lioncash:
Commit message seems to be missing an `A` at the beginning of it.

Source/Core/DolphinQt/Debugger/BreakpointWidget.cpp
```diff
@@ -188,7 +188,15 @@ void BreakpointWidget::Update()
     m_table->setItem(i, 3,
                      create_item(QStringLiteral("%1").arg(bp.address, 8, 16, QLatin1Char('0'))));
 
-    m_table->setItem(i, 4, create_item());
+    QString flags;
+
+    if (bp.break_on_hit)
+      flags.append(QLatin1Char{'p'});
```
sepalani, line 194: Why `p`? This should probably be replaced with `b` for `break` instead?

RenaKunisaki: I did this a while ago and don't remember why I chose `p`, so I'll change it to `b`.

RenaKunisaki:
Fixed commit message and changed 'p' to 'b'.

sepalani:
@dolphin-emu-bot rebuild

Source/Core/DolphinQt/Debugger/BreakpointWidget.cpp
```diff
@@ -308,7 +316,11 @@ void BreakpointWidget::OnSave()
 
 void BreakpointWidget::AddBP(u32 addr)
 {
-  PowerPC::breakpoints.Add(addr);
+  AddBP(addr, false, true, true);
```
sepalani, line 319: Is there a reason it's logging by default? Is this method still used? If not, it should be removed.

There is also a missing newline after the function. Not sure why the lint bot didn't catch this.

RenaKunisaki: I believe "write to log and break" was always the default? The log function does still work; I've been using it to monitor function calls.

sepalani:
@dolphin-emu-bot rebuild

PR #8942, opened 2020-07-08
Reword "Please avoid forcing Dolphin to use MSAA by the driver" message

JosJuice:
Some users who get this error don't seem to gain much understanding of how to resolve the problem from reading the error message.


MayImilae:
Much better.

PR #8944, opened 2020-07-08
DolphinQt: Replace QTBUG-10561 workaround with better workaround

JosJuice:
This workaround is not timing sensitive, unlike the old one.

Hopefully fixes https://bugs.dolphin-emu.org/issues/12158 and https://bugs.dolphin-emu.org/issues/12180.


13xforever:
Also fixes crash during iso checksum verification

PR #8949, opened 2020-07-12
Android: Use system cache directory as cache directory

JosJuice:
This lets Android automatically delete data in the cache directory when the device is running low on space or when Dolphin is uninstalled.

### DIFF
--- a/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/utils/DirectoryInitialization.java
+++ b/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/utils/DirectoryInitialization.java
@@ -64,7 +64,7 @@ public final class DirectoryInitialization
     {
       if (PermissionsHandler.hasWriteAccess(context))
       {
-        if (setDolphinUserDirectory())
+        if (setDolphinUserDirectory(context))
         {
           initializeInternalStorage(context);
           initializeExternalStorage(context);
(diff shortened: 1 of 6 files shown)


mbc07:
Ideally the entire user folder should be moved to the app's own external directory (Internal Storage/Android/data/com.dolphinemu.dolphinemu/), but regardless this is a nice first step.

Storage access is gonna be even more restrictive with Android 11 and we don't even support Scoped Storage yet, this is gonna bit our arses soon =/

JosJuice:
I will work on that eventually.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/utils/DirectoryInitialization.java
```diff
@@ -88,22 +88,27 @@ private static void init(Context context)
...
-        NativeLibrary.SetUserDirectory(userPath);
-        return true;
-      }
+    if (!Environment.MEDIA_MOUNTED.equals(Environment.getExternalStorageState()))
+      return false;
 
-    }
+    File externalPath = Environment.getExternalStorageDirectory();
+    if (externalPath == null)
+      return false;
```
Tilka, line 98: Is this error recoverable enough that it would still make sense to set the cache directory?

JosJuice: No. We don't have any code in Dolphin that tries to write to the cache directory if we don't have access to external storage, and besides, I don't see how the call to get the cache directory could succeed if this call fails.

PR #8952, opened 2020-07-12
Fix disc image conversion button not being clear

Filoppi:
It wasn't clear whether the conversion would create a new file or replace the old one

### DIFF
--- a/Source/Core/DolphinQt/ConvertDialog.cpp
+++ b/Source/Core/DolphinQt/ConvertDialog.cpp
@@ -84,7 +84,7 @@ ConvertDialog::ConvertDialog(QList<std::shared_ptr<const UICommon::GameFile>> fi
   grid_layout->addWidget(new QLabel(tr("Remove Junk Data (Irreversible):")), 4, 0);
   grid_layout->addWidget(m_scrub, 4, 1);
 
-  QPushButton* convert_button = new QPushButton(tr("Convert to New File..."));
+  QPushButton* convert_button = new QPushButton(tr("Convert..."));
 
   QVBoxLayout* options_layout = new QVBoxLayout;
   options_layout->addLayout(grid_layout);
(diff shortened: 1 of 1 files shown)


JosJuice:
@dolphin-emu-bot rebuild

jordan-woyak:
I agree that it's not clear but I don't like the suggested wording for some reason.
Perhaps the to-be-generated filename should be shown directly above the "Convert" button to make things even more clear?

Filoppi:
> I agree that it's not clear but I don't like the suggested wording for some reason.
> Perhaps the to-be-generated filename should be shown directly above the "Convert" button to make things even more clear?

@JosJuice gave me the exact words

JosJuice:
No, I gave you the exact capitalization and punctuation to use given the words that you suggested. I don't know if this is the best way to word it or if there's a better way.

@jordan-woyak The user will select the location and filename after pressing the convert button.

jordan-woyak:
Ah. If the path is chosen after clicking the button maybe it should just say "Convert...".
Like with "File->Open..." the ellipsis can show that an additional dialog will be shown on press.

Rumi-Larry:
IMO, there should be two options "save as new file" and "replace original file". with the later being the default, since that's what most users want to do, to save space.

jordan-woyak:
> IMO, there should be two options "save as new file" and "replace original file". with the later being the default, since that's what most users want to do, to save space.

I'd suggest a "Remove original file after successful conversion" checkbox for this feature.

Filoppi:
I don't mind, but that is different PR. Unless it will be implemented soon, just let me know if "Convert..." is final so I can amend the commit.

JosJuice:
"Convert..." is no doubt an improvement over "Convert", and I would be fine with merging that change. The question is whether there could be an even better way to word it, but if you're fine with changing it to "Convert...", then let's do that.

Filoppi:
https://github.com/dolphin-emu/dolphin/pull/8973

Filoppi:
This can be squashed on merge as all commits have the same text as the PR title.
Or close/merge https://github.com/dolphin-emu/dolphin/pull/8974

JosJuice:
We do not use squash on merge, as explained on IRC. Closing in favor of the other PR.

PR #8955, opened 2020-07-16
Android: Don't access /dev/ashmem on newer Android versions

JosJuice:
Fixes a critical regression where PR #8953 made us unable to start emulation on Android 10 and newer. Android is restricting direct access to `/dev/ashmem` starting with the new SDK version, but we can use the new (and simpler) `ASharedMemory` API instead.

While this change does fix the problem on Android 10, there is the problem that `ASharedMemory` was added in SDK version 26. We currently have `minSdkVersion` set to 21, and I do not want to increase it to as much as 26 unless it is absolutely necessary. My hope was that I could select between the two APIs at runtime, like we are able to do in Java, but it does not seem to be possible to link to `ASharedMemory` without increasing `minSdkVersion` (or its NDK equivalent, `ANDROID_PLATFORM`).

### DIFF
--- a/Source/Core/Common/MemArena.cpp
+++ b/Source/Core/Common/MemArena.cpp
@@ -23,6 +23,7 @@
 #include <sys/mman.h>
 #include <unistd.h>
 #ifdef ANDROID
+#include <dlfcn.h>
 #include <linux/ashmem.h>
 #include <sys/ioctl.h>
 #endif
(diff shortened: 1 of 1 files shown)


JosJuice:
Problem solved. Fortunately, [Firefox used to have the exact same problem](https://bugzilla.mozilla.org/show_bug.cgi?id=1598068), and I was able to use the approach they used. (You'd think Google would've documented anything about how to use NDK APIs that are from a newer SDK than the minimum required version, especially when they're requiring us to use an API that was added in Android 8, but no...)

PR #8956, opened 2020-07-16
Revert "Android: Increase targetSdkVersion to 29"

JosJuice:
This reverts commit 95945a000eaf168e19516739582efbdab2253990.

Quickfix for the same problem as PR #8955 (starting emulation is broken on Android 10), since I believe it might take a little while for that PR to get into a mergeable state.


JosJuice:
The other PR was merged.

PR #8961, opened 2020-07-17
Android: Add In-Game Speed Limit setting

Ebola16:
Implements https://bugs.dolphin-emu.org/issues/12174

@JMC47 Round two.
@JosJuice This PR won't bypass our config system as previously discussed.

### DIFF
--- a/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/NativeLibrary.java
+++ b/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/NativeLibrary.java
@@ -288,6 +288,8 @@ public final class NativeLibrary
    */
   public static native void SaveScreenShot();
 
+  public static native void UpdateSpeedLimit(float Value);
+
   /**
    * Saves a game state to the slot number.
    *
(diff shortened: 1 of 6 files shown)


Ebola16:
@dolphin-emu-bot rebuild

JosJuice:
While I suppose the way this interacts with the core is fine (since emulation speed is not in the new config system yet), this PR does contain a lot of one-off code for changing just one setting which will all need to be replaced if we ever get a proper system for accessing the settings in-game... But on the other hand, removing the code added by this PR should be pretty easy once we do get such a system. So I dunno. I'm a bit torn on whether the usefulness outweighs the code bloat, since I don't know how often this actually would be used.

Ebola16:
The speed limit cap has applications for user convenience and identifying performance regressions so I think it's worth the code complexity IMO. @JMC47 requested this feature so he may have something else to say.

JosJuice:
For what it's worth, PR #8975 adds the type of comprehensive system for accessing settings in-game that I was referring to. With that PR merged, the emulation speed will automatically become changeable in-game once the emulation speed setting has been ported to the new config system in C++.

Ebola16:
Alright, I'll look into this at a later time then. Thanks!

Ebola16:
Closing since porting this setting to the new config system is possible now.

PR #8962, opened 2020-07-18
Android: Use storage access framework for importing WADs

JosJuice:
This is part of my efforts to add support for scoped storage. I figured I would start with a relatively simple feature to make sure that everyone is fine with the approach I'm taking before I tackle more complicated features like the game list.

### DIFF
--- a/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/ui/SettingsAdapter.java
+++ b/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/ui/SettingsAdapter.java
@@ -293,9 +293,6 @@ public final class SettingsAdapter extends RecyclerView.Adapter<SettingViewHolde
       case MainPresenter.REQUEST_GAME_FILE:
         extensions = FileBrowserHelper.GAME_EXTENSIONS;
         break;
-      case MainPresenter.REQUEST_WAD_FILE:
-        extensions = FileBrowserHelper.WAD_EXTENSION;
-        break;
       default:
         throw new InvalidParameterException("Unhandled request code");
     }
--- a/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/ui/main/MainActivity.java
+++ b/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/ui/main/MainActivity.java
@@ -171,8 +171,10 @@ public final class MainActivity extends AppCompatActivity implements MainView
   @Override
   public void launchInstallWAD()
   {
-    FileBrowserHelper.openFilePicker(this, MainPresenter.REQUEST_WAD_FILE, false,
-            FileBrowserHelper.WAD_EXTENSION);
+    Intent intent = new Intent(Intent.ACTION_OPEN_DOCUMENT);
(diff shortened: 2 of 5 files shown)


mbc07:
Make sure this works on an actual Android TV device. I don't know if this was fixed but the last time I checked (~ 2 years ago), Android TV didn't ship the DocumentsUI system app, which made SAF do nothing when an app requested it. If that's still the case we'll need some sort of fallback...

JosJuice:
According to https://commonsware.com/blog/2017/12/27/storage-access-framework-missing-action.html, it works on the Shield TV, so it should be fine ;)

But more seriously, that's some really weird behavior, and I agree that it would be a good idea to test it before this is merged. Adding a fallback wouldn't be too hard... as long as there are no devices that have this problem and run Android 11.

mbc07:
Well, we should have plenty of time. This issue seems exclusive to some Android TV devices, I've never seen an Android phone missing the DocumentsUI app (not even on heavily skinned ones like Xiaomi), and Android TV releases seems to follow a different schedule: at the time of this comment, Android TV 9.0 still is the latest version available, with no ETA on when Android TV 10 will be actually released...

Ebola16 (changes requested):
- Currently all file types are visible in the new storage access framework. I haven't worked with `ACTION_OPEN_DOCUMENT` before but it may be possible to restrict the file types to the relevant ones. See https://stackoverflow.com/questions/33117592/restricting-selectable-file-types-via-intent-in-android.
- Can access to `ACTION_OPEN_DOCUMENT` be blocked if `WRITE_EXTERNAL_STORAGE` permission isn't granted? See #9061.

JosJuice:
> Currently all file types are visible in the new storage access framework. I haven't worked with `ACTION_OPEN_DOCUMENT` before but it may be possible to restrict the file types to the relevant ones. See https://stackoverflow.com/questions/33117592/restricting-selectable-file-types-via-intent-in-android.

This only works for MIME types and not file name extensions, and there is no MIME type we can use for WAD files as far as I know. An unfortunate limitation of the framework.

> Can access to  `ACTION_OPEN_DOCUMENT` be blocked if `WRITE_EXTERNAL_STORAGE` permission isn't granted? See #9061.

`ACTION_OPEN_DOCUMENT` doesn't require `WRITE_EXTERNAL_STORAGE`.

Ebola16:
> This only works for MIME types and not file name extensions, and there is no MIME type we can use for WAD files as far as I know. An unfortunate limitation of the framework.

Fair point, if this framework is needed for scoped storage we'll need to accept that.

> `ACTION_OPEN_DOCUMENT` doesn't require `WRITE_EXTERNAL_STORAGE`.

Although that's true, the process of installing the WAD itself would require that permission. Since the actions that follow `ACTION_OPEN_DOCUMENT` can require `WRITE_EXTERNAL_STORAGE`, we should probably display the toast message from #9061 instead of starting the storage access framework activity.

JosJuice:
I've made the Install WAD action check for directory initialization.

Ebola16 (approved):

JosJuice:
For the reference, the first commit of this PR ("Android: Add content provider support to File::IOFile") was merged as part of PR #8902.

PR #8966, opened 2020-07-21
Android: Clear game profiles and GameSettings

Ebola16:
Users may be surprised that "Clear Game Settings" doesn't also clear game profiles. I also specified when exceptions are ignored, as recommended by Android Studio.

### DIFF
--- a/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/dialogs/GamePropertiesDialog.java
+++ b/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/dialogs/GamePropertiesDialog.java
@@ -6,7 +6,6 @@ import android.os.Bundle;
 import androidx.annotation.NonNull;
 import androidx.appcompat.app.AlertDialog;
 import androidx.fragment.app.DialogFragment;
-import androidx.fragment.app.FragmentActivity;
 
 import android.widget.Toast;
 
@@ -18,6 +17,7 @@ import org.dolphinemu.dolphinemu.features.settings.ui.SettingsActivity;
 import org.dolphinemu.dolphinemu.features.settings.utils.SettingsFile;
 import org.dolphinemu.dolphinemu.ui.platform.Platform;
 import org.dolphinemu.dolphinemu.utils.DirectoryInitialization;
+import org.dolphinemu.dolphinemu.utils.Log;
 
 import java.io.File;
 
@@ -97,12 +97,16 @@ public class GamePropertiesDialog extends DialogFragment
 
   private void clearGameSettings(String gameId)
   {
-    String path =
+    String gameSettingsPath =
             DirectoryInitialization.getUserDirectory() + "/GameSettings/" + gameId + ".ini";
-    File gameSettingsFile = new File(path);
-    if (gameSettingsFile.exists())
+    String gameProfilesPath = DirectoryInitialization.getUserDirectory() + "/Config/Profiles/";
+    File gameSettingsFile = new File(gameSettingsPath);
+    File gameProfilesDirectory = new File(gameProfilesPath);
+    boolean hadGameProfiles = recursivelyDeleteGameProfiles(gameProfilesDirectory, gameId);
+
+    if (gameSettingsFile.exists() || hadGameProfiles)
     {
-      if (gameSettingsFile.delete())
+      if (gameSettingsFile.delete() || hadGameProfiles)
       {
         Toast.makeText(getContext(), "Cleared settings for " + gameId, Toast.LENGTH_SHORT)
                 .show();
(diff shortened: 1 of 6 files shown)


Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/dialogs/GamePropertiesDialog.java
```diff
@@ -23,6 +22,8 @@
 
 public class GamePropertiesDialog extends DialogFragment
 {
+  private boolean mHadGameProfiles = false;
```
JosJuice, line 25: This shouldn't need to be a member variable. You can store the return value of `recursivelyDeleteGameProfiles` as a local variable in `clearGameSettings`.

Ebola16: I think there is a problem with converting the member variable to a local variable in `recursivelyDeleteGameProfiles`. Initializing the local variable immediately will clobber value changes during the loop. Even if I initialize the local variable later on, I don't know if there's a clean way to check if a file is a game profile, set `hadGameProfiles = true` if game file, then evaluate a potentially uninitialized `hadGameProfiles` by setting it to `false` if `hadGameProfiles` isn't `true`.

JosJuice: You should be able to write `recursivelyDeleteGameProfiles` like this:

```
  private boolean recursivelyDeleteGameProfiles(File file, String gameId)
  {
    boolean hadGameProfiles = false;
    if (file.isDirectory())
    {
      try
      {
        for (File child : file.listFiles())
        {
          if (child.getName().startsWith(gameId) && child.isFile())
          {
            child.delete();
            hadGameProfiles = true;
          }
          hadGameProfiles |= recursivelyDeleteGameProfiles(child, gameId);
        }
      }
      catch (Exception ignored)
      {
      }
    }
    return hadGameProfiles;
  }
```

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/dialogs/GamePropertiesDialog.java
```diff
@@ -112,10 +116,34 @@ private void clearGameSettings(String gameId)
...
+        {
+          if (child.getName().startsWith(gameId) && child.isFile())
+          {
+            child.delete();
+            mHadGameProfiles = true;
+          }
+          recursivelyDeleteGameProfiles(child, gameId);
+        }
+      }
+      catch (Exception ignored)
```
JosJuice, line 143: I forgot to mention this earlier. This should catch the specific exception you need to catch instead of just `Exception`, to make it clearer what the purpose of this `catch` is. (I'm imagining it's something like `IOException`, but I'm not sure.)

Ebola16: Android studio complains that `IOException` can't be raised by this code block. I think I just need `NullPointerException` since `file.listFiles()` could potentially cause that.

I'll get better at recognizing these exceptions eventually.

JosJuice: I don't really see how `file.listFiles()` could cause a `NullPointerException`, though... You're only ever passing non-null `File`s into the function.

Why was this `catch` added to begin with? Because of an observed crash?

Ebola16: It was an Android Studio complaint. I also suspected that the exception wouldn't happen but I wasn't 100% sure. Should I just suppress the warning instead?

JosJuice: In that case, I would suggest using the `@NotNull` annotation if you care about getting rid of the warnings.

Ebola16: `@NonNull` alone wasn't enough to take care of the warnings but I think I found a way. I also applied the same change to `DirectoryInitialization.deleteDirectoryRecursively`.

Ebola16:
@dolphin-emu-bot rebuild

Ebola16:
@dolphin-emu-bot rebuild

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/dialogs/GamePropertiesDialog.java
```diff
@@ -118,4 +122,32 @@ private void clearGameSettings(String gameId)
       Toast.makeText(getContext(), "No game settings to delete", Toast.LENGTH_SHORT).show();
     }
   }
+
+  private boolean recursivelyDeleteGameProfiles(@NonNull final File file, String gameId)
+  {
+    boolean hadGameProfiles = false;
+    File[] files = file.listFiles();
```
JosJuice, line 129: `listFiles` should be called after checking `isDirectory`. The `isDirectory` check is a bit pointless otherwise.

PR #8968, opened 2020-07-22
Android: Update file picker

weihuoya:
`com.nononsenseapps:filepicker` is out of maintenance, so add it in dolphin repository, and update to use androidx.


JosJuice:
@dolphin-emu-bot rebuild

Tilka:
Is there no better way to fix this? Your description sounds like we should replace it with something else rather than maintain it ourselves. (I don't know anything about Android development.)

JosJuice:
We won't be able to use this file picker with scoped storage anyway, and Google Play will most likely force us to use scoped storage on Android 11 and newer by the end of 2021. I am planning to add support for scoped storage and replace this file picker with the SAF file picker before then for the features where it is feasible, with PR #8962 being the first step. I'm not sure what I'll do with features that would be rather annoying to get working with scoped storage, like setting the NAND path. My preliminary plan was to let those few features keep the old file picker for old Android versions and just not expose the features at all on Android 11 and up.

JosJuice:
To comment more directly on this PR, I think it would be fine to keep this library the way it is until we get support for scoped storage, and then we can consider whether we want to get rid of it or do something else with it.

leoetlino:
I agree with JosJuice that there's not much point in updating it if it's going to be removed at some point or only kept on legacy versions of Android.

PR #8971, opened 2020-07-22
Config: Simplify IsSettingSaveable

JosJuice:
Should make it both easier to edit and faster to run.


iwubcode:
This is really nice!

I'm assuming `s_setting_saveable` is now only necessary for groups of settings that are only partially merged?

JosJuice:
Yes, that was the intent.

iwubcode:
Great, coupled with #8849 and there would be hardly any upkeep here!

JosJuice:
Only one of them has settings left in ConfigManager, the General section. The General section was marked as saveable before this PR, and none of the General settings that are defined in the new config system are referenced in ConfigManager, so I figured it was okay to leave it as it is.

PR #7252, opened 2018-07-12
Android: Add Log Configuration to UI

Ebola16:
Implements https://bugs.dolphin-emu.org/issues/10709.

Screenshot_20200507-075228_Dolphin Emulator

PR #8972, opened 2020-07-22
Android: Some simple Clang-Tidy suggestions

Ebola16:
Split from #7252

### DIFF
--- a/Source/Android/jni/MainAndroid.cpp
+++ b/Source/Android/jni/MainAndroid.cpp
@@ -63,7 +63,7 @@
 
 namespace
 {
-static constexpr char DOLPHIN_TAG[] = "DolphinEmuNative";
+constexpr char DOLPHIN_TAG[] = "DolphinEmuNative";
 
 ANativeWindow* s_surf;
 IniFile s_ini;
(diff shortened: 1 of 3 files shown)


Source/Android/jni/MainAndroid.cpp
```diff
@@ -157,7 +157,7 @@ static bool MsgAlert(const char* caption, const char* text, bool yes_no, Common:
   return result != JNI_FALSE;
 }
 
-static void ReportSend(std::string endpoint, std::string report)
+static void ReportSend(const std::string& endpoint, std::string report)
```
JosJuice, line 160:
```suggestion
static void ReportSend(const std::string& endpoint, const std::string& report)
```

delroth:
I don't have any objection against using auto when the right side of the assignment already repeats the type (which I imagine is what clang-tidy is finding here).

Ebola16:
@dolphin-emu-bot rebuild

Ebola16:
I'm somewhat indifferent to the `auto` changes but I like less complaints from Clang-Tidy.

PR #8941, opened 2020-07-07
Android: Replace Java INI parser with C++ INI parser

JosJuice:
Fixes https://bugs.dolphin-emu.org/issues/12096 and opens up for future improvements.

I had to touch a lot of the settings code in the second commit, so testing is needed to make sure that all the settings still work correctly, in particular per-game Wii controller settings.

PR #7920, opened 2019-03-22
Android: Add Paths to UI

Ebola16:
Implements https://bugs.dolphin-emu.org/issues/10781
- Android: Add Set as Default ISO to UI
- Android: Add Paths to UI
- Very useful for loading game mods
- Removed some unnecessary import statements
- Renamed some things now that `FilePicker` is more versatile
- Added `ConfirmRunnable` to confirm running arbitrary runnables (so they're not accidentally clicked). It will also be useful for future PRs, especially when modifying multiple settings (i.e. toggling all log types).
- Fixed a bug where `launchOpenFileActivity()` could select multiple files. We don't want to be able to load multiple games at once!

- [X] Review requested: See the TODO added to `SettingsActivity`. Comment out the `finish()` to see the weirdness. My workaround works but it probably isn't the prettiest!

Screenshot_20200105-170115_Dolphin Emulator

Screenshot_20200322-054716_Dolphin Emulator

PR #8975, opened 2020-07-22
Android: Hook up the new config system

JosJuice:
This gives users the ability to edit settings while emulation is running, and improves the game settings editor by highlighting settings which are overridden and by showing the user's config instead of defaults for settings which are not overridden. This also makes it relatively easy for a future PR to add SYSCONF settings to the Android GUI.


Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/ui/SettingsActivity.java
```diff
@@ -172,22 +189,15 @@ protected void onActivityResult(int requestCode, int resultCode, Intent result)
...
-        Toast.makeText(this, "Saved settings to INI files", Toast.LENGTH_SHORT).show();
-      }
     }
 
+    // Save modified settings since finish() won't save them.
+    mPresenter.onStop(true);
+
     // TODO: After result of FilePicker, duplicate SettingsActivity appears.
     //       Finish to avoid this. Is there a better method?
     finish();
```
Ebola16, line 203: Now that I think about it, we'd both need to do less merge conflict resolution if turned #9014 into a review comment. Please remove `mPresenter.onStop(true);` and `finish();` after rebasing.

JosJuice: I'd rather keep PR #9014 open and solve the conflict once that PR is merged, since I don't know how long it's going to take for this PR to get reviewed and merged, but if you say so...

Done in the latest push.

Ebola16: I started pulling code from this PR to make `FilePickerViewHolder.bind` work the same way but those changes sprawled across multiple files. I think the change makes sense here since it relies on part of this PR.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/ui/viewholder/FilePickerViewHolder.java
```diff
@@ -45,15 +46,21 @@ public void bind(SettingsItem item)
     }
     else
     {
-      mTextSettingDescription.setText(NativeLibrary
-              .GetConfig(mFilePicker.getFile(), item.getSection(), item.getKey(),
-                      mFilePicker.getSelectedValue()));
+      mTextSettingDescription.setText(mFilePicker.getSelectedValue(getAdapter().getSettings()));
```
Ebola16, line 49: These descriptions are always blank.

JosJuice: I don't understand your comment. This is the line that makes the selected path show up in the UI.

Ebola16: When viewing the Paths submenu, all Paths have a blank description. It should display the value saved for that path instead.

JosJuice: It does display the saved value for me. However, it doesn't display anything when you haven't set a path, since the C++ config system's default for all paths is "".

Ebola16: We display default paths in QT as of #8710. I also added default paths in #7920 for Android. See `SettingsFragmentPresenter.getDefaultNANDRootPath` and the code after that. Hopefully I implemented them properly.

JosJuice: Ah, so DolphinQt doesn't just display the value straight from the config system... I suppose I could implement something similar in Android without actually changing what the config system considers to be the default value.

JosJuice: Should be fixed with the commit I just added.

Ebola16: It is. Apparently I can't close my conversations on other people's PRs. That's annoying.

JosJuice: Huh, I thought that was possible... I'll mark it as resolved myself, then.

lioncash:
After the conflicts are fixed, feel free to merge this once the buildbot checks pass. This has been sitting around for quite a while (sorry about that!)

JosJuice:
I'd like to leave some additional time before merging this in case @Ebola16 wants to review it. PR #8941 (which this PR was dependent on) was merged just two days ago, and I believe they said they would look in more detail at this PR after the merge of that PR.

Ebola16:
I'll probably be able to review this within a day or two. I still need to carefully rebase my Wii Overlay PR.

Ebola16:
I'll probably need another two days until I review this, sorry. Game-specific settings in the Wii Overlay PR are giving me trouble after a rebase so I should figure out what's going on there before I look at this PR.

JMC47:
Let's try to get everything with the settings/features setup before the next progress report so we can give Android Users some nice benefits and quality of life features.

Ebola16:
@JMC47 If two more days is too much time I can still review this after merge since @lioncash said this PR is good to go. I should definitely look into the custom game settings oddities though, although I highly suspect they're on my end and not caused by anything in Master.

JMC47:
Nope, I'm just saying we try to get this all squared away before the end of the month.

Ebola16 (commented):
- In-game graphics settings changes (tested Aspect Ratio and Post-processing) aren't immediately updated.
- In-game settings with custom game setting set are still modifiable but the change is ignored. Perhaps either make these non-modifiable and show a toast explaining why or make modifications save to the custom game settings?
- On Windows, "Insert SD Card" and the CPU Clock Override settings are modifiable in-game.

JosJuice:
> In-game graphics settings changes (tested Aspect Ratio and Post-processing) aren't immediately updated.

Weird, I'm pretty sure it worked before, but it doesn't seem to be working anymore... I'll take a closer look.

> In-game settings with custom game setting set are still modifiable but the change is ignored. Perhaps either make these non-modifiable and show a toast explaining why or make modifications save to the custom game settings?

I tried to match the DolphinQt behavior where changes to such settings persist until you end the emulation session, since I had no particular reason to want to go against the DolphinQt behavior.

> On Windows, "Insert SD Card" and the CPU Clock Override settings are modifiable in-game.

The reason they aren't modifiable in-game on Android is because they aren't using the new config system yet.

Ebola16:
> I tried to match the DolphinQt behavior where changes to such settings persist until you end the emulation session, since I had no particular reason to want to go against the DolphinQt behavior.

To clarify with an example, if the OSD display setting is overridden to false, modifying that setting in-game will not allow OSD messages to be displayed. If that's intended behavior I'd like a toast when a user tries to modify an overridden setting in-game explaining that those changes will do nothing and not be retained.

> The reason they aren't modifiable in-game on Android is because they aren't using the new config system yet.

That's fine, just making sure this was intentional.

JosJuice:
> To clarify with an example, if the OSD display setting is overridden to false, modifying that setting in-game will not allow OSD messages to be displayed. If that's intended behavior I'd like a toast when a user tries to modify an overridden setting in-game explaining that those changes will do nothing and not be retained.

The intended behavior in that case is that OSD messages will be displayed for the rest of the current emulation session. If that's not happening, it sounds like a similar issue to the first of the three issues.

Ebola16:
> The intended behavior in that case is that OSD messages will be displayed for the rest of the current emulation session. If that's not happening, it sounds like a similar issue to the first of your three issues.

Actually I can't reproduce that anymore. I'll assume I accidentally didn't clear my custom settings while testing unless I can reproduce it again. I think the "In-game graphics settings changes (tested Aspect Ratio and Post-processing) aren't immediately updated" is the only relevant issue now.

JosJuice:
The issue with the graphics settings is fixed now. Turns out that I have to call `Config::InvokeConfigChangedCallbacks` manually with the way I'm using the config API.

Ebola16 (approved):

PR #8976, opened 2020-07-22
Port some settings to the new config system

JosJuice:
Split out from the Android PR #8975.

Other than the controller settings and JIT debug settings, these are the only settings which were defined in Java code but not defined in the new config system in C++. (There are still a lot of settings that are defined in the new config system but not yet saveable in the new config system, though.)

### DIFF
--- a/Source/Core/VideoCommon/OnScreenDisplay.cpp
+++ b/Source/Core/VideoCommon/OnScreenDisplay.cpp
@@ -13,9 +13,10 @@
 #include <imgui.h>
 
 #include "Common/CommonTypes.h"
+#include "Common/Config/Config.h"
 #include "Common/Timer.h"
 
-#include "Core/ConfigManager.h"
+#include "Core/Config/MainSettings.h"
 
 namespace OSD
 {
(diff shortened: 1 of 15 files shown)


Source/Core/VideoCommon/OnScreenDisplay.cpp
```diff
@@ -94,7 +95,7 @@ void AddMessage(std::string message, u32 ms, u32 rgba)
 
 void DrawMessages()
 {
-  const bool draw_messages = SConfig::GetInstance().bOnScreenDisplayMessages;
+  const bool draw_messages = Config::Get(Config::MAIN_OSD_MESSAGES);
```
BhaaLseN, line 98: Is this a potentially time-sensitive location that could use some caching (and a hook into the config changed callback)? Feels like this might be called fairly often (like every frame or so).

JosJuice: I believe this is called exactly once a frame. I've assumed that running this 60 times a second wouldn't have any noticeable impact.

BhaaLseN: Thinking of #8884 somewhere in the back of my mind, which seemed similar. Didn't test (nor profile) it though, so I'm just guessing from staring at the code.

lioncash (approved):

PR #8977, opened 2020-07-23
GameSettings: SafeTextureCacheColorSamples for SEUPEY and SEVPEY

chungy:
Works around a rendering error in Retro City Rampage (disc) and
Shakedown: Hawaii.  Fixes redmine issues #12197 and #12198.

### DIFF
--- a/Data/Sys/GameSettings/SEU.ini
+++ b/Data/Sys/GameSettings/SEU.ini
@@ -1,0 +1,4 @@
+# SEUPEY - Retro City Rampage DX+
+
+[Video_Settings]
+SafeTextureCacheColorSamples = 0
--- a/Data/Sys/GameSettings/SEV.ini
+++ b/Data/Sys/GameSettings/SEV.ini
@@ -1,0 +1,4 @@
+# SEVPEY - Shakedown: Hawaii
+
+[Video_Settings]
+SafeTextureCacheColorSamples = 0


Data/Sys/GameSettings/SEVPEY.ini
```diff
@@ -0,0 +1,4 @@
+# SEUPEY - Shakedown: Hawaii
```
AdmiralCurtiss, line 1: Game ID is wrong here.

chungy: Right, fault of me using cp :)

JosJuice:
Game INIs which contain settings should use three-character names (SEU.ini/SEV.ini) instead of six-character names, unless it can be shown that there are versions of the game that do not require the setting (which of course is not the case here since the games only have one version each). LGTM other than that.

chungy:
> Game INIs which contain settings should use three-character names (SEU.ini/SEV.ini) instead of six-character names

Thanks, I was not aware -- I don't really know the intricacies of Wii serial numbers nor how dolphin-emu uses them.

JosJuice:
@dolphin-emu-bot rebuild

JosJuice:
No problem. This convention is essentially something I made up a number of years ago that hasn't been formally written down, so I don't blame anyone for not knowing about it.

MayImilae:
@chungy Do you have the titles in question? If so, could you give us before and after images for the progress report? We gotta cover a fix for the final wii game!

chungy:
Yup, that's why I made the PR.

I can do this soon.

PR #8980, opened 2020-07-24
Fix some GCC warnings

Tilka:
Still lots of warnings left.


Source/Core/DolphinQt/Debugger/ThreadWidget.cpp
```diff
@@ -468,7 +465,7 @@ void ThreadWidget::OnSelectionChanged(int row)
 {
   Common::Debug::PartialContext context;
 
-  if (row >= 0 && row < m_threads.size())
+  if (row >= 0 && (size_t)row < m_threads.size())
```
lioncash, line 468:
```suggestion
  if (row >= 0 && size_t(row < m_threads.size()))
```

PR #8981, opened 2020-07-24
LightingShaderGen/VertexShaderGen: Migrate over to fmt

lioncash:
Continues the fmt migration by migrating these two generators over. Both were done at the same time, since the vertex shader generator relied on some defines that the lighting shader generator used.

### DIFF
--- a/Source/Core/VideoCommon/LightingShaderGen.cpp
+++ b/Source/Core/VideoCommon/LightingShaderGen.cpp
@@ -17,60 +17,61 @@ static void GenerateLightShader(ShaderCode& object, const LightingUidData& uid_d
   const char* swizzle = alpha ? "a" : "rgb";
   const char* swizzle_components = (alpha) ? "" : "3";
 
-  int attnfunc = (uid_data.attnfunc >> (2 * litchan_index)) & 0x3;
-  int diffusefunc = (uid_data.diffusefunc >> (2 * litchan_index)) & 0x3;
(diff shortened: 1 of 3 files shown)


Tilka:
What's the advantage of using fmt here? Personally I find `%u` more obvious than `{}` because it shows that a number is being written.

lioncash:
Because it handles all types natively with one specifier. People also routinely keep getting %u and %d mixed up with types and it leads to compiler warnings. With fmt, you cannot get the specifiers wrong, because it properly handles their types. For types that cannot be printed, it will give you a compiler error instead of blindly trying to format the type anyway. It also handles C++ types out of the box like std::string and std::string_view in a straightforward manner (literally just passing the variable into the formatting call). Feel free to see the [standard paper](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p0645r7.html) on its drawbacks.

This isn't a particularly bespoke formatting specifier choice, the syntax was borrowed from Python, Rust, and several of the .NET languages.

fmt is also on average [faster](https://github.com/fmtlib/fmt#benchmarks) than printf. It's also standardized as `std::format` in C++20.

brujo5:
Hello sir, is there any chance to fix this? thx.
https://forums.dolphin-emu.org/Thread-dolphin-5-0-12764-crashes-every-time-i-try-to-start-a-game?action=lastpost

PR #8984, opened 2020-07-26
Curl: For windows cmake builds, build curl with schannel/winssl capability

r2dliu:
Added a few flags to the curl external static build so that it has ssl capabilities. otherwise, it doesn't support protocols like https and you get errors trying to send curl requests to those kind of endpoints

### DIFF
--- a/Externals/curl/lib/CMakeLists.txt
+++ b/Externals/curl/lib/CMakeLists.txt
@@ -1,5 +1,7 @@
-if(NOT CMAKE_SYSTEM_NAME MATCHES "Windows")
-  add_definitions(-DHAVE_CONFIG_H)
+if(CMAKE_SYSTEM_NAME MATCHES "Windows")
+    add_definitions(-DUSE_SCHANNEL -DUSE_WINDOWS_SSPI)
+else()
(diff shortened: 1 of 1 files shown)


JosJuice:
These are the same definitions as when building with the VS project, so the change makes sense to me.

Please squash the two commits into one.

r2dliu:
done

Externals/curl/lib/CMakeLists.txt
```diff
@@ -1,5 +1,7 @@
 if(NOT CMAKE_SYSTEM_NAME MATCHES "Windows")
```
lioncash, line 1: Given, this now has an else case, it'd likely be better to invert this as:

```cmake
if(CMAKE_SYSTEM_NAME MATCHES "Windows")
    add_definitions(-DUSE_SCHANNEL -DUSE_WINDOWS_SSPI)
else()
    add_definitions(-DHAVE_CONFIG_H)
endif()
```

so that the `NOT` can be dropped, making for less reading.

JosJuice:
@dolphin-emu-bot rebuild

PR #8987, opened 2020-07-28
NetPlay: Set Wiimotes at join

Tibladar:
This assigns a wiimote for every new user in NetPlay, like it's already done for GC pads.

There are probably better ways to do this (e.g. detect if the game is for Wii/GC and set the controllers accordingly) but it depends on what is wanted.

Should fix https://bugs.dolphin-emu.org/issues/12206

### DIFF
--- a/Source/Core/Core/NetPlayServer.cpp
+++ b/Source/Core/Core/NetPlayServer.cpp
@@ -399,6 +399,16 @@ unsigned int NetPlayServer::OnConnect(ENetPeer* socket, sf::Packet& rpac)
       break;
     }
   }
+  
+  // try to automatically assign new user a wiimote
+  for (PlayerId& mapping : m_wiimote_map)
+  {
+    if (mapping == 0)
+    {
+      mapping = player.pid;
+      break;
+    }
+  }
 
   // send join message to already connected clients
   sf::Packet spac;


JMC47:
Has this been tested to work with things like differing attachments and such?  That was one of the main hurdles that caused it to turn out the way it was now.

JosJuice:
Actually, wouldn't the problem with this PR rather be that it forces you to deal with Wii Remote sync (unless you manually turn off the Wii Remotes) when playing a Wii game that supports GameCube controllers?

JMC47:
Oops.  I misunderstood the purpose of this Pull Request.  Please ignore my previous comment.

Maybe we should auto-detect GC vs Wii?

Tibladar:
I'm not an expert, so I can't really say without testing.
The games I own run without problems with this PR but none supports both Wii and CameCube controllers.

As for attachments, I tested the Nunchuk a bit.

Maybe it's better to detect if the game is for Wii/GC and set the controllers accordingly but it really depends on what the user wants. Or this can be a NetPlay setting/switch (Mapping: GC only, Wii only, Both)

Tibladar:
Is there a way to detect from a NetPlay thread if the current game is for Wii or GC?
Calling `game->GetPlatform()` is only working in GUI threads.

jordan-woyak:
Like JosJuice said, I think this is going to be undesirable in Wii games that are typically played with GC controller (e.g. Brawl).

Tibladar:
Then the other option would be to add a setting where the user can switch between the controller presets 'GC only', 'Wii only' or 'Both'.

jordan-woyak:
An option to make it more clear about which ports are being assigned would probably make this better. I think the changes as-is it would make the situation worse for players of Wii games using GC gamepads. This PR has aged and needs a rebase so I'm going to close it for now, but we can re-open on new information.

PR #8988, opened 2020-07-28
VideoCommon: Don't round the refresh rate

JosJuice:
We now provide a double to the FPS counter and exact values to FIFO recording and frame dumping.

### DIFF
--- a/Source/Core/VideoCommon/FrameDump.cpp
+++ b/Source/Core/VideoCommon/FrameDump.cpp
@@ -186,11 +186,11 @@ bool FrameDump::CreateVideoFile()
     s_codec_context->codec_tag = MKTAG('X', 'V', 'I', 'D');
 
   s_codec_context->codec_type = AVMEDIA_TYPE_VIDEO;
-  s_codec_context->bit_rate = g_Config.iBitrateKbps * 1000;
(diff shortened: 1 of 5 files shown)


Source/Core/VideoCommon/FrameDump.cpp
```diff
@@ -186,11 +186,11 @@ bool FrameDump::CreateVideoFile()
...
 
   s_codec_context->codec_type = AVMEDIA_TYPE_VIDEO;
-  s_codec_context->bit_rate = g_Config.iBitrateKbps * 1000;
+  s_codec_context->bit_rate = static_cast<int64_t>(g_Config.iBitrateKbps) * 1000;
   s_codec_context->width = s_width;
   s_codec_context->height = s_height;
-  s_codec_context->time_base.num = 1;
-  s_codec_context->time_base.den = VideoInterface::GetTargetRefreshRate();
+  s_codec_context->time_base.num = VideoInterface::GetTargetRefreshRateDenominator();
+  s_codec_context->time_base.den = VideoInterface::GetTargetRefreshRateNumerator();
```
altimumdelta, line 193: Is .num and .den switched around intentional?

JosJuice: Yes. The time base is the length of a frame, not the number of frames per second.

umut-eng (commented):
Thanks

PR #8995, opened 2020-08-01
Fix some more GCC warnings

Tilka:
This works around some GCC bugs:
- [-Wshadow warns three times when a constructor parameter shadows a data member](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=78147)
- [std::optional can trigger bogus -Wmaybe-uninitialized warnings](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=80635)

### DIFF
--- a/Source/Core/DiscIO/Volume.h
+++ b/Source/Core/DiscIO/Volume.h
@@ -58,15 +58,20 @@ public:
   std::optional<u64> ReadSwappedAndShifted(u64 offset, const Partition& partition) const
   {
     const std::optional<u32> temp = ReadSwapped<u32>(offset, partition);
-    return temp ? static_cast<u64>(*temp) << GetOffsetShift() : std::optional<u64>();
+    if (!temp)
+      return std::nullopt;
+    return static_cast<u64>(*temp) << GetOffsetShift();
   }
 
   virtual bool IsEncryptedAndHashed() const { return false; }
   virtual std::vector<Partition> GetPartitions() const { return {}; }
   virtual Partition GetGamePartition() const { return PARTITION_NONE; }
-  virtual std::optional<u32> GetPartitionType(const Partition& partition) const { return {}; }
+  virtual std::optional<u32> GetPartitionType(const Partition& partition) const
+  {
+    return std::nullopt;
+  }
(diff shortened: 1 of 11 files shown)


Source/Core/DiscIO/Volume.h
```diff
@@ -66,7 +66,11 @@ class Volume
   virtual Partition GetGamePartition() const { return PARTITION_NONE; }
   virtual std::optional<u32> GetPartitionType(const Partition& partition) const { return {}; }
   std::optional<u64> GetTitleID() const { return GetTitleID(GetGamePartition()); }
+  // https://gcc.gnu.org/bugzilla/show_bug.cgi?id=80635
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wmaybe-uninitialized"
```
BhaaLseN, line 71: Sadly, those produce warnings on MSVC (which in turn gets promoted to an error). 

But what surprises me a bit more is that the warning isn't raised on `GetPartitionType` literally 3 lines before this (where the same happens, except using a `u32` instead of a `u64`).
Does it go away if you use `std::nullopt` instead?

Tilka: Thanks, didn't know about std::nullopt.

Source/Core/DiscIO/Volume.h
```diff
@@ -66,7 +66,9 @@ class Volume
   virtual Partition GetGamePartition() const { return PARTITION_NONE; }
   virtual std::optional<u32> GetPartitionType(const Partition& partition) const { return {}; }
   std::optional<u64> GetTitleID() const { return GetTitleID(GetGamePartition()); }
-  virtual std::optional<u64> GetTitleID(const Partition& partition) const { return {}; }
+  // TODO: change this back to "return {}" once this bug is fixed:
```
lioncash, line 69: These comments can probably be removed, given std::nullopt is more than acceptable here.

`std::nullopt` should be preferred anyways, since it allows implementations to avoid unnecessarily zeroing out the internal buffer of the optional ([example](https://godbolt.org/z/esrrbq))

`ReadSwappedAndShifted, and GetPartitionType() should probably be converted over to using std::nullopt as well.

PR #8996, opened 2020-08-01
Various improvements for the Memory Card Manager.

AdmiralCurtiss:
This rewrites a bunch of code regarding importing and exporting of external GC savefiles (GCI/GCS/SAV) in order to make the code easier to read and more correct, as well as improve the user experience of using the memory card manager.

Externally visible changes:
- Merged the Title and Comment columns, removed the fairly useless First Block column, added a new Filename column so files are identifiable in error messages. ([old](https://user-images.githubusercontent.com/4522237/89089592-5ace8980-d39e-11ea-940c-ee29a0cb1eb3.png), [new](https://user-images.githubusercontent.com/4522237/89824507-80ecea00-db53-11ea-9a65-d2fb490e589b.png))
- Importing saves now allows selecting multiple files at once.
- Exporting multiple saves now allows selecting the file type to export as (was always GCI before) and folder to export to.
  - This means you now need to select the file type to export as before selecting the filename/path, which I think is an okay trade-off.
  - I also removed the Export All button as you can just select all files and hit Export now and still get a target folder prompt.
- When importing or copying, it now checks in advance whether the import/copy even fits onto the target card. This means we can now give better error messages ([old](https://user-images.githubusercontent.com/4522237/89089593-5b672000-d39e-11ea-9cd5-daf1cdec3ac3.png) [new1](https://user-images.githubusercontent.com/4522237/89089589-5a35f300-d39e-11ea-9ea8-10a3d9c9795a.png) [new2](https://user-images.githubusercontent.com/4522237/89089590-5ace8980-d39e-11ea-89dd-f19edc681752.png) [new3](https://user-images.githubusercontent.com/4522237/89089591-5ace8980-d39e-11ea-8133-658550f4fe96.png)) and cancel the import before it even starts (so no selecting 2 files to copy and it only copies the 1st and fails the 2nd due lack of space or whatever, which could easily happen before).
- Columns are now resizable.
- Columns are now sortable.

Internal-only changes:
- In the actual memory card code:
  - Rewrote the entire file exporting and importing code. This now lives outside the memory card code in utility functions instead of being interwoven with the code that actually writes to or read from the memory card.
  - Removed a bunch of now-obsolete code in the memory card class.
- In the card manager:
  - Each table cell now knows its savefile rather than us detecting the selected files based on the highlighted rows. This allows the sorting.
  - Rewrote basically all of the import/export/copy/delete code to use the new import/export functions and give better error messages.
  - Copying and importing now share a large chunk of code instead of being separate.


Source/Core/DolphinQt/GCMemcardManager.cpp
```diff
@@ -189,66 +217,62 @@ void GCMemcardManager::UpdateSlotTable(int slot)
...
+        file_comments ? QString::fromStdString(file_comments->second).trimmed() : QString();
+    auto banner = GetBannerFromSaveFile(file_index, slot);
+    auto icon_data = GetIconFromSaveFile(file_index, slot);
 
-    QString blocks = QStringLiteral("%1").arg(memcard->DEntry_BlockCount(file_index));
-    QString block_count = QStringLiteral("%1").arg(memcard->DEntry_FirstBlock(file_index));
+    auto* item_banner = new QTableWidgetItem();
+    auto* item_text = new QTableWidgetItem(QStringLiteral("%1\n%2").arg(title).arg(comment));
+    auto* item_icon = new QTableWidgetItem();
+    auto* item_blocks = new QTableWidgetItem(QStringLiteral("%1").arg(block_count));
```
lioncash, line 253: Given there's no particular formatting other than number->string needing to be done here, this can be simplified to

```suggestion
    auto* item_blocks = new QTableWidgetItem(QString::number(block_count));
```

Source/Core/DolphinQt/GCMemcardManager.cpp
```diff
@@ -189,66 +217,62 @@ void GCMemcardManager::UpdateSlotTable(int slot)
...
+        file_comments ? QString::fromStdString(file_comments->first).trimmed() : QString();
+    const QString comment =
+        file_comments ? QString::fromStdString(file_comments->second).trimmed() : QString();
+    auto banner = GetBannerFromSaveFile(file_index, slot);
+    auto icon_data = GetIconFromSaveFile(file_index, slot);
 
-    QString blocks = QStringLiteral("%1").arg(memcard->DEntry_BlockCount(file_index));
-    QString block_count = QStringLiteral("%1").arg(memcard->DEntry_FirstBlock(file_index));
+    auto* item_banner = new QTableWidgetItem();
+    auto* item_text = new QTableWidgetItem(QStringLiteral("%1\n%2").arg(title).arg(comment));
```
lioncash, line 251: Multiple QStrings can be formatted in a single formatting pass:

```suggestion
    auto* item_text = new QTableWidgetItem(QStringLiteral("%1\n%2").arg(title, comment));
```

Source/Core/DolphinQt/GCMemcardManager.cpp
```diff
@@ -298,157 +321,309 @@ void GCMemcardManager::SetSlotFileInteractive(int slot)
...
+std::vector<u8> GCMemcardManager::GetSelectedFileIndices()
 {
-  auto selection = m_slot_table[m_active_slot]->selectedItems();
-  auto& memcard = m_slot_memcard[m_active_slot];
-
-  auto count = selection.count() / m_slot_table[m_active_slot]->columnCount();
-
-  for (int i = 0; i < count; i++)
+  const auto selection = m_slot_table[m_active_slot]->selectedItems();
+  std::vector<bool> lookup(Memcard::DIRLEN, false);
```
lioncash, line 327:
```suggestion
  std::vector<bool> lookup(Memcard::DIRLEN);
```

Source/Core/DolphinQt/GCMemcardManager.cpp
```diff
@@ -298,157 +321,309 @@ void GCMemcardManager::SetSlotFileInteractive(int slot)
...
-      if (path.isEmpty())
-        return;
-    }
-    else
+    const int index = item->data(Qt::UserRole).toInt();
+    if (index < 0 || index >= static_cast<int>(Memcard::DIRLEN))
     {
-      path = QString::fromStdString(File::GetUserPath(D_GCUSER_IDX)) +
-             QStringLiteral("/%1").arg(QString::fromStdString(gci_filename));
+      PanicAlert("Data inconsistency in GCMemcardManager, aborting action.");
```
lioncash, line 333:
```suggestion
      QMessageBox::warning(tr("Data inconsistency in GCMemcardManager, aborting action."));
```

or use `ModelMessageBox::warning`

Source/Core/DolphinQt/GCMemcardManager.cpp
```diff
@@ -298,157 +321,309 @@ void GCMemcardManager::SetSlotFileInteractive(int slot)
...
-  if (result != Memcard::GCMemcardImportFileRetVal::SUCCESS)
+  if (savefiles.size() == 1)
   {
-    ModalMessageBox::critical(this, tr("Import failed"), tr("Failed to import \"%1\".").arg(path));
+    // when exporting a single save file, let user specify exact path
+    const std::string basename = Memcard::GenerateFilename(savefiles[0].dir_entry);
+    const QString qformatdesc = GetFormatDescription(format);
+    const QString qfilename = QFileDialog::getSaveFileName(
+        this, tr("Export Save File"),
+        QString::fromStdString(File::GetUserPath(D_GCUSER_IDX)) +
```
lioncash, line 392: Perhaps this should be placred into a fmt::format call and then be converted to a QString at the end or something? This formatting has a lot going on churning QStrings from QStrings, and it's pretty wonky to follow.

Source/Core/DolphinQt/GCMemcardManager.cpp
```diff
@@ -298,157 +321,309 @@ void GCMemcardManager::SetSlotFileInteractive(int slot)
...
+  for (const Memcard::Savefile& savefile : savefiles)
+  {
+    const auto result = card->ImportFile(savefile);
+
+    // we've already checked everything that could realistically fail here, so this should only
+    // happen if the memory card data is corrupted in some way
+    if (result != Memcard::GCMemcardImportFileRetVal::SUCCESS)
+    {
+      const std::string filename = Memcard::GenerateFilename(savefile.dir_entry);
+      ModalMessageBox::critical(
```
lioncash, line 508: These should likely be warnings, given other failures related to importing are also warnings. Also this isn't really crititcal, given it doesn't threaten the stability of the application. It's a problem that can be dealt with.

Ditto for any other potential cases.

Source/Core/DolphinQt/GCMemcardManager.cpp
```diff
@@ -298,157 +321,309 @@ void GCMemcardManager::SetSlotFileInteractive(int slot)
...
+  if (!card)
+    return;
+
+  const auto selected_indices = GetSelectedFileIndices();
+  if (selected_indices.empty())
+    return;
+
+  const QString text =
+      selected_indices.size() == 1 ?
+          tr("Do you want to delete the selected save file?") :
```
lioncash, line 608: The way to handle plurals here is to do it in a single statement like:

```cpp
tr("Do you want to delete the %n selected save file(s)?", "", static_cast<int>(selected_indices.size()));
```

This allows languages with differing handling of quantities to be translated better. (see the handling plurals part of Qt localizing [docs](https://doc.qt.io/qt-5/i18n-source-translation.html#handling-plurals))

Source/Core/DolphinQt/GCMemcardManager.h
```diff
@@ -13,21 +13,26 @@
 
 #include "Common/CommonTypes.h"
 
+#include "Core/HW/GCMemcard/GCMemcardUtils.h"
```
lioncash, line 16: This can likely be replaced with forward declarations (the standard allows forward declaring types in std::vector).

Source/Core/DolphinQt/GCMemcardManager.cpp
AdmiralCurtiss: Ah, I didn't know you could do that! That probably applies in a few more places here too.

Source/Core/DolphinQt/GCMemcardManager.cpp
AdmiralCurtiss: Yeah the string concat is a bit messy in places... I'll see what I can do to make it nicer.

Source/Core/DolphinQt/GCMemcardManager.h
AdmiralCurtiss: This is for `void ExportFiles(Memcard::SavefileFormat format);` which I don't think you can forward declare.

Source/Core/DolphinQt/GCMemcardManager.cpp
AdmiralCurtiss: Alright, I think this is better now?

Source/Core/DolphinQt/GCMemcardManager.h
lioncash: It's permissible in the standard to forward declare enumerations that are either `enum class Name`, `enum class Name : <explicit type>`, or `enum Name : <explicit type>`

AdmiralCurtiss: Sure enough, that works. Good to know.

AdmiralCurtiss:
Did another minor pass on this.

- Columns are now resizable by the user.
- You can now sort by columns.
- Added a filename column so someone actually has a chance to figure out [what files are meant in the error messages](https://user-images.githubusercontent.com/4522237/89089591-5ace8980-d39e-11ea-8133-658550f4fe96.png). This makes the manager less compact again, which is a bit unfortunate, but oh well...
- Something about this massively reduced the CPU usage when playing icon animations. I'm guessing it now no longer tries to auto-resize the columns whenever icons update?

Source/Core/DolphinQt/GCMemcardManager.cpp
```diff
@@ -78,11 +98,20 @@ void GCMemcardManager::CreateWidgets()
...
+
+  m_export_button = new QToolButton(this);
+  m_export_menu = new QMenu(m_export_button);
+  m_export_gci_action = new QAction(tr("&Export as .gci..."), m_export_menu);
+  m_export_gcs_action = new QAction(tr("Export as .&gcs..."), m_export_menu);
+  m_export_sav_action = new QAction(tr("Export as .&sav..."), m_export_menu);
+  m_export_menu->addAction(m_export_gci_action);
+  m_export_menu->addAction(m_export_gcs_action);
+  m_export_menu->addAction(m_export_sav_action);
+  m_export_button->setText(tr("&Export as .gci..."));
```
BhaaLseN, line 111: Have you looked into using `m_export_button->setDefaultAction(m_export_gci_action)` instead? Usually with dropdown buttons, reusing actions from the menu should be preferred to avoid duplication. [According to documentation](https://doc.qt.io/qt-5/qtoolbutton.html#setDefaultAction), it sets the text amongst other properties.

AdmiralCurtiss: Ah, no I have not, I'll check if that works.

AdmiralCurtiss: Yup that works, did that.

Source/Core/DolphinQt/GCMemcardManager.cpp
```diff
@@ -15,25 +20,40 @@
...
 #include "Core/HW/GCMemcard/GCMemcard.h"
+#include "Core/HW/GCMemcard/GCMemcardUtils.h"
 
 #include "DolphinQt/GCMemcardCreateNewDialog.h"
 #include "DolphinQt/QtUtils/ModalMessageBox.h"
 
-constexpr float ROW_HEIGHT = 28;
+constexpr int ROW_HEIGHT = 36;
+constexpr int COLUMN_WIDTH_FILENAME = 100;
+constexpr int COLUMN_WIDTH_BANNER = 96 + 6;
```
Rumi-Larry, line 47: Any reason why this isn't just 102?

Source/Core/DolphinQt/GCMemcardManager.cpp
```diff
@@ -15,25 +20,40 @@
...
+constexpr int COLUMN_WIDTH_TEXT = 160;
+constexpr int COLUMN_WIDTH_ICON = 32 + 6;
```
Rumi-Larry, line 49: ditto, but 38

Source/Core/DolphinQt/GCMemcardManager.cpp
AdmiralCurtiss: 96 is the width of the banner and then the 6 is just to give it a bit of space in the table, IIRC. Probably not a bad idea to give that a name though.

Source/Core/DolphinQt/GCMemcardManager.cpp
AdmiralCurtiss: Same, but with the width of the icon.

Source/Core/DolphinQt/GCMemcardManager.cpp
```diff
@@ -293,162 +343,316 @@ void GCMemcardManager::SetSlotFileInteractive(int slot)
...
+    const auto* error_code = std::get_if<Memcard::ReadSavefileErrorCode>(&read_result);
+    if (error_code)
+    {
+      ModalMessageBox::warning(
+          this, tr("Import Failed"),
+          tr("Failed to import \"%1\".\n%2").arg(path, GetErrorMessageForErrorCode(*error_code)));
       return;
+    }
+
+    // should never get here but just in case...
```
iwubcode, line 594: I feel like a variant visitor would remove this code section and avoid the if checks above

AdmiralCurtiss: Honestly I'm not sure if that's really much better.

https://gist.github.com/AdmiralCurtiss/a363a3806c49de9c1b8a77e7b2b2cce6

Thoughts?

AdmiralCurtiss: Alright I massaged this a bit more and I think I ended up with something better. Also has the nice side-effect of showing all errors instead of just the first.

image

iwubcode: Yeah, that looks pretty good.  I do tend to prefer the class version because it allows for state and doesn't require that weird overloaded notation (though can use `constexpr` with a template) but I think it's a little more readable compared to what was there before

iwubcode: Whoops, was busy writing a message and edited multiple times, missed your new comment :)

iwubcode: LGTM

Source/Core/Core/HW/GCMemcard/GCMemcardUtils.cpp
```diff
@@ -53,4 +71,282 @@ bool HasSameIdentity(const DEntry& lhs, const DEntry& rhs)
...
+
+std::variant<ReadSavefileErrorCode, Savefile> ReadSavefile(const std::string& filename)
+{
+  File::IOFile file(filename, "rb");
+  if (!file)
+    return ReadSavefileErrorCode::OpenFileFail;
+
+  if (!file.Seek(0, SEEK_END))
+    return ReadSavefileErrorCode::IOError;
+  const u64 filesize = file.Tell();
```
iwubcode, line 198: Could this be moved down near its usage?

Source/Core/DolphinQt/GCMemcardManager.cpp
```diff
@@ -293,162 +344,314 @@ void GCMemcardManager::SetSlotFileInteractive(int slot)
...
-    const auto exportRetval = memcard->ExportGci(file_index, path.toStdString(), "");
-    if (exportRetval == Memcard::GCMemcardExportFileRetVal::UNUSED)
-    {
-      File::Delete(path.toStdString());
-    }
+  std::vector<u8> selected_indices;
+  for (u8 i = 0; i < Memcard::DIRLEN; ++i)
+  {
+    if (lookup[i])
+      selected_indices.push_back(i);
```
iwubcode, line 373: Is there a reason you need two loops?

Source/Core/DolphinQt/GCMemcardManager.cpp
```diff
@@ -293,162 +344,314 @@ void GCMemcardManager::SetSlotFileInteractive(int slot)
...
+  {
+    QString failure_string =
+        tr("Failed to export %n save file(s).", "", static_cast<int>(failures));
+    if (failures == savefiles.size())
+    {
+      ModalMessageBox::warning(this, tr("Export Failed"), failure_string);
+    }
+    else
+    {
+      QString success_string = tr("Successfully exported %n save file(s).", "",
```
iwubcode, line 481: Would it be worthwhile to put the total here too?

Source/Core/Core/HW/GCMemcard/GCMemcardUtils.cpp
AdmiralCurtiss: No, but what we can do is just use file.GetSize() instead.

Source/Core/DolphinQt/GCMemcardManager.cpp
AdmiralCurtiss: Wish you had asked me this a few months ago, but the code as-written ensures that the returned indices never contain any duplicates, which is important. We also get a consistent ascending order of the indices, though I honestly don't remember if that part is important (probably not?). I suppose if we don't mind the arbitrary order we could do easily do it with one loop too -- maybe even preferable if that respects the order the user selected the items and/or the order of the items in the list?

iwubcode: Ah yes, I didn't think about the ordering of the final result, that may matter.  No concerns with it as is, just was curious, thank you!

Source/Core/DolphinQt/GCMemcardManager.cpp
AdmiralCurtiss: Sure, why not.

JMC47:
It always takes me a minute to figure out how to use the Memory Card Manager, but once I get into it, it seems fine.  The UI improvements are massive but it seems to behave roughly the same outside of the exceptions mentioned in the original post.  LGTM.

PR #8998, opened 2020-08-02
Update warning log OpenGL 2.0 to 3.x

RedDevilus:
This will change the warning message for users when they run an outdated computer or one where it can't detect the GLextensions.

This is what displayed on a computer with OpenGL 2.1 and DirectX  10.1 version. Pressing either one of these options doesn't work to use Dolphin.
Either way minimum is 3.x for OpenGL.
image

### DIFF
--- a/Source/Core/VideoBackends/OGL/main.cpp
+++ b/Source/Core/VideoBackends/OGL/main.cpp
@@ -122,8 +122,8 @@ bool VideoBackend::InitializeGLExtensions(GLContext* context)
   if (!GLExtensions::Init(context))
   {
     // OpenGL 2.0 is required for all shader based drawings. There is no way to get this by
-    // extensions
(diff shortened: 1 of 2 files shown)


degasus:
Dolphin requires OpenGL 3.1 (or 3.0 + UBO), however this line of code should only fail with OpenGL < 2.0.

RedDevilus:
I know but for users they'll think they only need 2.0 and i made it 3.x instead of a number to make it obvious for that's the minimum while for developers know it's for OpenGL 2.0 . Not sure how it failed with OpenGL 2.1though.

jordan-woyak:
Like mentioned, this particular error message is based on a OpenGL 2.0 feature so it's not entirely accurate to make it say 3.x. I'm going to close this PR for now but we can re-open on new information.

PR #8999, opened 2020-08-02
Add a few input expression functions (trigonometry, sqrt, pow, min, max, clamp)

AdmiralCurtiss:
Someone [on the forums](https://forums.dolphin-emu.org/Thread-feature-request-thread?pid=510144#pid510144) requested trigonometric functions to be added, so I figured might as well, easy enough. Also added a few more functions that seemed useful.


jordan-woyak:
Just an FYI. min and max are already available using the `&` and `|` operators but it probably makes sense to have named versions too.

Filoppi:
I've added some of these here already:
https://github.com/dolphin-emu/dolphin/blob/85c26e1b018f331f55a5e2ef2e4d7468018119a6/Source/Core/InputCommon/ControlReference/FunctionExpression.cpp

Rumi-Larry:
Poke. Is this pr still alive?

AdmiralCurtiss:
This is good-to-go from my side.

PR #9001, opened 2020-08-03
Port JIT debug settings to the new config system

OatmealDome:
Title. I think this is how you do a "depends-on" PR?

Depends on #8976.


Techjar:
I'm not sure this is not acceptable in its current form. Lookups in onion config are very slow (relatively speaking) due to string comparisons, and JIT functions are pretty hot during codegen, so this is likely going to have a noticeable impact on how long it takes to JIT a block.

JosJuice:
Yep. That is why I didn't do the JIT debug settings in my PR, because I wasn't sure about what approach to use for caching the values.

OatmealDome:
Fair enough. I only really need part of this for my fork.

PR #9002, opened 2020-08-03
MainWindow: Attach quit emulation sheet to RenderWidget if shown on macOS

OatmealDome:
_Extremely_ minor UI tweak. Without this PR, the sheet would be attached to MainWindow and macOS would change the focused window to it. Always kinda annoyed me.

image

### DIFF
--- a/Source/Core/DolphinQt/MainWindow.cpp
+++ b/Source/Core/DolphinQt/MainWindow.cpp
@@ -824,7 +824,7 @@ bool MainWindow::RequestStop()
       Core::SetState(Core::State::Paused);
 
     auto confirm = ModalMessageBox::question(
-        this, tr("Confirm"),
+        m_rendering_to_main ? static_cast<QWidget*>(this) : m_render_widget, tr("Confirm"),
         m_stop_requested ? tr("A shutdown is already in progress. Unsaved data "
                               "may be lost if you stop the current emulation "
                               "before it completes. Force stop?") :


JosJuice:
@dolphin-emu-bot rebuild

Source/Core/DolphinQt/MainWindow.cpp
```diff
@@ -824,7 +824,7 @@ bool MainWindow::RequestStop()
       Core::SetState(Core::State::Paused);
 
     auto confirm = ModalMessageBox::question(
-        this, tr("Confirm"),
+        m_rendering_to_main ? static_cast<QWidget*>(this) : m_render_widget, tr("Confirm"),
```
JosJuice, line 827: Why is this cast needed? Does the ternary operator not compile otherwise?

OatmealDome: clang gives me the following error:
```
../Source/Core/DolphinQt/MainWindow.cpp:827:29: error: incompatible operand types ('MainWindow *' and 'RenderWidget *')
        m_rendering_to_main ? this : m_render_widget, tr("Confirm"),
```

JosJuice: Ah, right, it works when you have one pointer that's a QWidget* and one that derives from QWidget*, but not when you have two different types that derive from QWidget*. Makes sense then. I think I was assuming that `m_render_widget` already had the type QWidget*.

Tilka:
I think this actually affects all platforms (other than Android).

PR #9005, opened 2020-08-04
DolphinQt: Ignore "-psn" command line option on macOS

OatmealDome:
On an application's first launch on macOS, a command line option matching the format "-psn_X_XXXXXX" is passed. This is to set the ["ProcessSerialNumber", something used by the legacy Process Manager from Carbon](https://web.archive.org/web/20040705133028/developer.apple.com/documentation/Carbon/Reference/Process_Manager/prmref_main/data_type_5.html). optparse recognizes it as an unknown option and immediately quits, so the application appears to crash.

This PR makes DolphinQt ignore the "-psn" command line option if it's found.

### DIFF
--- a/Source/Core/DolphinQt/Main.cpp
+++ b/Source/Core/DolphinQt/Main.cpp
@@ -109,6 +109,18 @@ int main(int argc, char* argv[])
   }
 #endif
 
+#ifdef __APPLE__
+  // On macOS, a command line option matching the format "-psn_X_XXXXXX" is passed when
+  // the application is launched for the first time. This is to set the "ProcessSerialNumber",
+  // something used by the legacy Process Manager from Carbon. optparse will fail if it finds
+  // this as it isn't a valid Dolphin command line option, so pretend like it doesn't exist
+  // if found.
+  if (strncmp(argv[argc - 1], "-psn", 4) == 0)
+  {
+    argc--;
+  }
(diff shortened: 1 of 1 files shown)


JosJuice:
@dolphin-emu-bot rebuild

OatmealDome:
Fixed lint (I think).

JosJuice:
@dolphin-emu-bot rebuild

BhaaLseN:
Is this guaranteed to be the _last_ argument being passed? Like, does it also happen if you happen it for the first time using a command line argument like a DOL or so to run? Or does that only happen from Finder (being passed on first run)?

OatmealDome:
> Is this guaranteed to be the last argument being passed?

There's no documentation for this argument that I can find, so I can't definitively say that this is true. However, from what I can tell, whenever it does appear in the arguments list it is always the sole one. If you try and pass arguments on the first run (for example, ``open -a Dolphin.app --args file.dol``), it doesn't get included.

> Like, does it also happen if you happen it for the first time using a command line argument like a DOL or so to run? Or does that only happen from Finder (being passed on first run)?

It does not appear if you run the executable directly from the command line. Instead, it appears to be added when the application is opened via LaunchServices (via the Finder, ``open`` without ``--args``, AppleScript, etc).

PR #9006, opened 2020-08-04
Linux Buildbot

CrankySupertoon:


JosJuice:
While I'm assuming you didn't intend to submit this as a pull request here, there is one thing I must comment on regardless (since it is of relevance regardless of what you are intending to do with this branch): This PR contains copyrighted software stored on the Wii NAND, which is illegal to upload. Look in the list of added files for binary files larger than a megabyte.

PR #9007, opened 2020-08-05
GameSettings: Add patches for Pokémon Colosseum and Pokémon XD to circumvent the check that the save being overwritten matches the last known save.

AdmiralCurtiss:
I think the [last Progress Report](https://dolphin-emu.org/blog/2020/07/05/dolphin-progress-report-may-and-june-2020/) sums this up pretty well:

> Note: These fixes do not affect Pokemon Colosseum and Pokemon XD's save issues regarding savestates. They have purposeful memory card checks to prevent the duping of Pokemon. Unlike the previous problem, which was more of an emulation oversight, this one is something that should be solved through a game patch. Currently no one is investigating this issue and you should be wary of using savestates with these games. If someone does write a patch or a cheat for this issue, we will be happy to include it with the game's INI file for easy access to users.

These are the relevant patches.

I have not thoroughly tested this yet except for the simple 'make savestate, make regular save, load savestate, try to save again' action chain that didn't work before but works with these patches. It's technically possible that this has side-effects I am unaware of.

### DIFF
--- a/Data/Sys/GameSettings/GC6E01.ini
+++ b/Data/Sys/GameSettings/GC6E01.ini
@@ -1,0 +1,27 @@
+# GC6E01 - Pokémon Colosseum
+
+[OnFrame]
+# This game has extra memory card checks compared to most games,
+# presumably to prevent Pokémon duping. These can interfere with
+# saving your game when using savestates.
+#
+# 0x801cfc2c:
+# Originally a branch that checks if the save counter in memory
+# matches the one on the memory card. We patch this to instead
+# overwrite the counter in memory with the one from the memory
+# card. We need to do this because each memory card file actually
+# contains three save slots the game rotates when saving, and
+# when loading it loads the save with the biggest save counter.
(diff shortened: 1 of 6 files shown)


JMC47:
Currently testing this on each version of the game.

AdmiralCurtiss:
FYI one minor problem I'm aware of is that when you load a savestate from before the first time you've ever saved the game, it will not ask you to overwrite an existing save, it just does it. Not really a big deal in the grand scheme of things, but something one could potentially improve.

I've also not tried what happens in sections where the game autosaves (eg: before credits).

JMC47:
I grabbed a savefile at the end and am running through it now.  I tested the Japanese and PAL versions of the code near the intro, and everything seemed to work fine with savestates following the instructions provided.

JMC47:
I'm probably not going to mess with this too much more at this point.  It works, so let's merge it and see if users run into anything before the next monthly release.

PR #9014, opened 2020-08-07
Android: Remove finish from FilePicker

Ebola16:
Should probably be merged before #8975.
Thanks to the recent lifecycle changes this is possible.

`NativeLibrary.SetConfig` updates the FilePicker selection but won't retain that selection after saving.
`mView.putSetting` doesn't update the FilePicker selection but will allow the selection to be saved.

...so why not both? Still not perfect but it's better than prematurely calling `finish`.


JosJuice:
You can skip having to use `NativeLibrary.SetConfig` in `onFilePickerConfirmation` by not using `NativeLibrary.GetConfig` in `FilePickerViewHolder.bind`. Feel free to use c873026 (from PR #8975) as a reference.

PR #9015, opened 2020-08-07
Add DRM platform (through DRM+EGL+GBM backend) to dolphin-nogui

rtissera:
Tested on ODROID N2 (Mali G52 GPU).
This relies on libdrm, libgbm and EGL. Everything is properly detected through CMake modules.
There's room for improvements for sure, especially in EGLDRM.cpp code.
Feedback is welcome.


JosJuice:
@dolphin-emu-bot rebuild

Source/Core/Common/CMakeLists.txt
```diff
@@ -222,10 +222,21 @@ target_sources(common PRIVATE
 )
 
 if(ENABLE_EGL AND EGL_FOUND)
+  find_package(Libdrm)
+  find_package(Libgbm)
+  if (LIBDRM_FOUND AND LIBGBM_FOUND)
+    target_sources(common PRIVATE
```
lioncash, line 228: This command should be indented like so:

```cmake
  target_sources(common PRIVATE
    GL/GLInterface/EGL.cpp
    GL/GLInterface/EGL.h
    GL/GLInterface/EGLDRM.cpp
    GL/GLInterface/EGLDRM.h
  )
```

Source/Core/Common/CMakeLists.txt
```diff
@@ -222,10 +222,21 @@ target_sources(common PRIVATE
...
+    GL/GLInterface/EGL.cpp
+    GL/GLInterface/EGL.h
+    GL/GLInterface/EGLDRM.cpp
+    GL/GLInterface/EGLDRM.h
+    )
+  else()
   target_sources(common PRIVATE
```
lioncash, line 235: This command should be indented by two spaces like the command in the related if statement.

Source/Core/Common/GL/GLInterface/EGLDRM.h
```diff
@@ -0,0 +1,55 @@
...
+
+   unsigned major;
+   unsigned minor;
+
+} egl_ctx_data_t;
+
+class GLContextEGLDRM : public GLContext
+{
+public:
+  virtual ~GLContextEGLDRM() override;
```
lioncash, line 30:
```suggestion
  ~GLContextEGLDRM() override;
```

Source/Core/Common/GL/GLInterface/EGLDRM.h
```diff
@@ -0,0 +1,55 @@
...
+#pragma once
+
+#include <EGL/egl.h>
+#include <EGL/eglext.h>
+#include <string>
+#include <vector>
+
+#include "Common/GL/GLContext.h"
+
+typedef struct
```
lioncash, line 14: In C++ this can formatted more concisely like so (with naming altered to follow our naming conventions):

```cpp
struct EGLContextData
{
   EGLContext ctx;
   EGLSurface surf;
   EGLDisplay dpy;
   EGLConfig config;
   int interval;

   unsigned major;
   unsigned minor;
};
```

Source/Core/Common/GL/GLInterface/EGLDRM.cpp
```diff
@@ -0,0 +1,1287 @@
+// Copyright 2020 Dolphin Emulator Project
+// Licensed under GPLv2+
+// Refer to the license.txt file included.
+
```
lioncash, line 4: Top level comment for the file just to state that a lot of this code currently doesn't match our naming conventions and coding styles and should conform to that.

A lot of it is very C89 (variables at top of scopes, rather than where they're used, etc) and should be migrated to idiomatic C++ before being merged.

JosJuice:
@dolphin-emu-bot rebuild

Source/Core/Common/GL/GLInterface/EGLDRM.cpp
```diff
@@ -0,0 +1,1287 @@
...
+ *
+ *  RetroArch is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ *  without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR
+ *  PURPOSE.  See the GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License along with RetroArch.
+ *  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include <stdint.h>
```
lioncash, line 22: These includes should be formatted like so:

```cpp
#include <array>
#include <cerrno>
#include <cmath>
#include <cstddef>
#include <cstdint>
#include <cstdio>
#include <cstdlib>
#include <cstring>
#include <sstream>
#include <vector>

#include <fcntl.h>
#include <poll.h>
#include <sched.h>
#include <sys/stat.h>
#include <sys/time.h>
#include <sys/types.h>
#include <unistd.h>

#include <gbm.h>
#include <libdrm/drm.h>
#include <xf86drm.h>
#include <xf86drmMode.h>
```

JosJuice:
Please make sure you press "view all lines" on the lint log. It's consists of about 2000 lines, not just the 40 ones that are shown initially.

Source/Core/Common/GL/GLInterface/EGLDRM.cpp
```diff
@@ -0,0 +1,1287 @@
...
+#define EGL_CONTEXT_OPENGL_RESET_NOTIFICATION_STRATEGY_KHR 0x31BD
+#define EGL_NO_RESET_NOTIFICATION_KHR 0x31BE
+#define EGL_LOSE_CONTEXT_ON_RESET_KHR 0x31BF
+#define EGL_CONTEXT_OPENGL_FORWARD_COMPATIBLE_BIT_KHR 0x00000002
+#define EGL_CONTEXT_OPENGL_ROBUST_ACCESS_BIT_KHR 0x00000004
+#define EGL_CONTEXT_OPENGL_CORE_PROFILE_BIT_KHR 0x00000001
+#define EGL_CONTEXT_OPENGL_COMPATIBILITY_PROFILE_BIT_KHR 0x00000002
+#endif /* EGL_KHR_create_context */
+
+typedef bool (*egl_accept_config_cb_t)(void *display_data, EGLDisplay dpy, EGLConfig config);
```
lioncash, line 79:
```suggestion
using EGLAcceptConfigCB = bool (*)(void *display_data, EGLDisplay dpy, EGLConfig config);
```

Source/Core/Common/GL/GLInterface/EGLDRM.cpp
```diff
@@ -0,0 +1,1287 @@
...
+
+typedef struct gfx_ctx_drm_data
```
lioncash, line 81:
```cpp
struct GFXContextDRMData
{
  egl_ctx_data_t egl;
  int fd;
  int interval;
  unsigned fb_width;
  unsigned fb_height;

  bool core_hw_context_enable;
  bool waiting_for_flip;
  gbm_bo *bo;
  gbm_bo *next_bo;
  gbm_surface *gbm_surface;
  gbm_device  *gbm_dev;
};
```

Source/Core/Common/GL/GLInterface/EGLDRM.cpp
```diff
@@ -0,0 +1,1287 @@
...
+   uint32_t fb_id;
+};
+
+/* TODO/FIXME - globals */
+static drmEventContext g_drm_evctx;
+static struct pollfd g_drm_fds;
+static uint32_t g_connector_id           = 0;
+static int g_drm_fd                      = 0;
+static uint32_t g_crtc_id                = 0;
+static drmModeCrtc *g_orig_crtc          = NULL;
```
lioncash, line 109: nullptr instead of NULL. Ditto for all other occurrences.

Source/Core/DolphinNoGUI/PlatformDRM.cpp
```diff
@@ -0,0 +1,83 @@
...
+public:
+  ~PlatformDRM() override;
+
+  bool Init() override;
+  void SetTitle(const std::string& string) override;
+  void MainLoop() override;
+
+  WindowSystemInfo GetWindowSystemInfo() const override;
+
+private:
```
lioncash, line 40: This isn't necessary, given it's not used.

Source/Core/DolphinNoGUI/PlatformDRM.cpp
```diff
@@ -0,0 +1,83 @@
...
+};
+
+PlatformDRM::~PlatformDRM()
```
lioncash, line 43:
```suggestion
PlatformDRM::~PlatformDRM() = default;
```

rtissera:
Thanks for the productive comments and suggestions, I will tackle most of these as soon as possible.

JosJuice:
@dolphin-emu-bot rebuild

Source/Core/Common/CMakeLists.txt
rtissera: Fixed in 414cb0ec131cad6667f229c7b5b8b16969d42772

Source/Core/Common/CMakeLists.txt
rtissera: Fixed in 414cb0ec131cad6667f229c7b5b8b16969d42772

Source/Core/Common/GL/GLInterface/EGLDRM.h
rtissera: Fixed in c55dc590e461bc7e9d647ad0cb8e0d7ff50044ae

Source/Core/Common/GL/GLInterface/EGLDRM.cpp
rtissera: Fixed in c55dc590e461bc7e9d647ad0cb8e0d7ff50044ae

Source/Core/Common/GL/GLInterface/EGLDRM.cpp
rtissera: Fixed in 02575c2cbfa074043668d7d3b41b0bd122e9e664

Source/Core/DolphinNoGUI/PlatformDRM.cpp
rtissera: Fixed in 67cb42244209095b64022b3616000c7686859655

Source/Core/Common/GL/GLInterface/EGLDRM.cpp
rtissera: Fixed in dcabf7a330cfeb8db857ceced4e40034b3731877

Source/Core/Common/GL/GLInterface/EGLDRM.cpp
```diff
@@ -0,0 +1,1222 @@
...
+// Refer to the license.txt file included.
+
+// EGL DRM GBM code based on RetroArch, RetroArch licenses follows
+
+/*  RetroArch - A frontend for libretro.
+ *  Copyright (c) 2011-2017 - Daniel De Matteis
+ *
+ *  RetroArch is free software: you can redistribute it and/or modify it under the terms
+ *  of the GNU General Public License as published by the Free Software Found-
+ *  ation, either version 3 of the License, or (at your option) any later version.
```
linkmauve, line 12: It seems you copied a GPLv3+ file into a GPLv2+ project, you should first get the approval from all of the owners of the original file to relicense under a different license if you were to do so.

rtissera: Indeed this all started as an experiment but I will rebase the problematic code on another DRM implementation from some batocera folks to avoid licensing issues. I have started a big refactor to meet both the licensing and Dolphin coding standards.

rtissera:
Can some relaunch a buildbot try ? I tried to fix failing checks.

JosJuice:
@dolphin-emu-bot rebuild

rtissera:
@dolphin-emu-bot rebuild

rtissera:
sorry, I missed some #ifdef and lint-checks

JosJuice:
@dolphin-emu-bot rebuild

rtissera:
All green :) What are the next steps to move forward ?

RinMaru:
why are you adding DRM to Dolphin?

rtissera:
To be able to run it on ARM boards (ODROID N2 or Khadas VIM3 for instance)
using Linux distros like batocera

Tilka:
to avoid any confusion: in this context DRM stands for Direct Rendering Manager, not Digital Rights Management

RinMaru:
> to avoid any confusion: in this context DRM stands for Direct Rendering Manager, not Digital Rights Management

thats a load off lol

rtissera:
Latest commit refactors the whole code to C++, could you re-run a buildbot and reconsider ?

JosJuice:
@dolphin-emu-bot rebuild

i30817:
This means that the retroarch fork will be smaller?

MANGOM1LK:
retroarch fork is straight-up not currently being maintained at the moment as far as i am aware

i30817:
Yeah, i hoped this would fix that, much like i can compile ppsspp as a core (even with all of its problems), but lesser (since not completely upstream).

NielsGx:
I don't think a RPi 4 4Gb would be enough to run dolphin android ?
When you say you tested on ODROID N2, how was it like performance wise ? Thanks btw

orbea:
> I don't think a RPi 4 4Gb would be enough to run dolphin android ?

This would be potentially useful for more powerful *unix computers too to avoid requiring X11/wayland.

i30817:
Correct, i play dolphin in KMS in retroarch because it's a bit faster.

Having shed a lot of useless processes.

There are problems with this approach, though - they occur because of hardware and driver problems and are not especially relevant for dolphin but for example, for scummvm.

My portable has a 'quirky' touchpad hardware. Lots of hardware 'quirks' are not specified in the kernel drivers, but in higher level libraries like libinput, that are loaded by desktop environments.

In short, i can't play scummvm in KMS because the touchpad doesn't have a right click (the difference between 'left and right' is probably a question of coordinates in the quirk) and because the cursor projects absolutely into the screen, so if you lift your finger and move it , it jumps.

rtissera:
Another approach would be to revive a SDL2 build, command-line based, as SDL2 now features a really solid kmsdrm backend (and it's even going to be better in upcoming SDL 2.0.13 release).

orbea:
> Another approach would be to revive a SDL2

I can vouch this already works well in some cases, but I suspect a lot of distros do not ship SDL2 with their kmsdrm backend enabled. Maybe it will be enabled by default in future SDL2 releases?

i30817:
Hey, I tried this (with a slightly modified [retroarch dolphin launcher core](https://github.com/RobLoach/libretro-dolphin-launcher) to use a standalone path and receive the -p drm argument.

This mode has a major problem, both there and even with x11 and no-gui: the hotkeys simply don't work, so there is no way to exit the game (which is a bit of a problem if you happen to be on KMS/tty.

Not sure if it's a bug or a lack of window (and thus focus) preventing them from being delivered, but it makes it almost useless for the 'no window manager case'.

i30817:
Mh. Ok from reading the code, PlatformX11 has its own hardcode of some events (quit, savestates, fullscreen toggle, pause, screenshot, load last savestate.

But it's broken for x11 somehow. DRM doesn't even handle it.

i30817:
I have a patch that can get the keys running, but all the modifier keys don't activate and i have no idea why.

Also ideally, HotkeyScheduler could be abstracted away from QT and reused as a central point to have shortcuts with a thread. Would need to have all the hotkeys be redifined without the emit, maybe by making it subclassable. Without pulling in QT into the compilation unit preferably.

i30817:
Btw you should probably also disable the signal to stop (ctrl+s) if you don't expect lusers to complain about unexpected freezing if they don't know how to unfreeze (ctrl+q).

The above (that happened to me, i'm a luser does not mean i found a way to make ctrl work, these are handled by the kernel keyboard driver i think.

rtissera:
Can you share / submit your patch ?

i30817:
As a pr? it's up there in the previous post to the last. And ctrl not working is a pain in the ass considering the default dolphin shortcuts relating to savestates.

It's also not the best way to do this probably, reusing hotkeyscheduler would seem saner, after a refactoring moving it to common or something like that.

i30817:
I just tried some games and emulated wiimotes aren't working here, though the classic attachment is if the game supports it, so that's fun.

i30817:
Ah I think the above happens because only 'xinput' allowed the merging of input sources (in this case the keyboard and a touchpad / mouse pointer). edev doesn't and when you try to configure one of the other in the dolphin compilation, you get one of the other working and when you drop down to the actual DRM layer it's no different.

A shame.

silvarios:
> I just tried some games and emulated wiimotes aren't working here, though the classic attachment is if the game supports it, so that's fun.
> 
> Possibly because the wiimote tries to get the 'virtual pointer' (eg: the mouse) and evdev goes 'what mouse?'. But i'm not going to fearmonger too much until someone more competent than me confirms it. I don't have a real wiimote to test either.

@i30817 
I built this branch the way you suggested, with all your flags and I definitely have Wii Remote support, via my Dolphin Bar adapter. I did have to set the correct settings in Dolphin.ini for it to show up -- WiimoteContinuousScanning = True and you have to tweak the WiimoteNew.ini too. I think you can get away with just putting the following setting in for basic functionality (for each Wiimote):
[Wiimote1]
Source = 2

This is running in DRM mode -- `dolphin-emu-nogui -p drm`.  Initially I tested the various settings in an X11 supported build to figure out what settings were required to make the nogui versions function with sound, Wii Remotes, etc. I'm sure you have tried this already, but if you have not, go ahead and give the settings a tweak.

I also built the branch with no special flags and same behavior resulted, Wii Remote and nunchuk support both work fine in drm.

silvarios:
I am encountering an odd problem of my own.
My setup:
Platform: Linux EndeavourOS (Arch Linux distro)
Linux Kernel: 5.8.9-arch2-1
CPU: Haswell i7-4600U 2.1 GHZ
GPU: HD Graphics 4400
RAM: 8GB
Dolphin build: This branch's latest commits, trying no flags and also i30817's build with `-DENABLE_HEADLESS=ON -DENABLE_EGL=ON -DENABLE_EVDEV=ON -DLINUX_LOCAL_DEV=ON -DOpenGL_GL_PREFERENCE=GLVND -DENABLE_TESTS=OFF -DENABLE_LLVM=OFF -DENABLE_ANALYTICS=OFF -DENABLE_X11=OFF -DENCODE_FRAMEDUMPS=OFF`

My laptop's screen is failing, so I have grub set to disable the internal screen and only output over HDMI to my 4K TV. Grub is set for 3840x2160@30Hz. Give me a moment to describe how RetroArch behaves in KMS on the same system. My system boots into a systemd unit file that launches RetroArch in TTY3 at boot:

1. If I leave RetroArch with default configs, it sets itself to 1920x1080@60Hz.
2. If I set the RetroArch fullscreen setting to "True", then RetroArch will take the system's resolution and refresh, which is 3840x2160@30 Hz. Unfortunately, 30 Hz and gaming is not a good combination, but bear with me.
3. If I set the fullscreen resolution setting with RetroArch to something else, say 1360x768, RetroArch will run at that resolution and proper 60 Hz refresh.

So the problem I'm having is dolphin-emu-nogui is behaving the exact same way as RetroArch, except, the seemingly similar settings for Dolphin.ini are not being recognized. I think the correct settings for Dolphin.ini would be something like:
```
[Display]
Fullscreen = True
FullscreenDisplayRes = HDMI1: 1360x768
```

I have tried the same setting without "HDMI1:" but does not seem to make a difference. When I launched Dolphin in GUI mode, those particular settings were saved when running fullscreen (as I was running the display in 1360x768 to see if the settings would autosave) and so that's why I assumed they would work in nogui mode too. I have tried changing the kernel parameter in grub to 1360x768, but dolphin-emu-nogui still switches to 1920x1080 in TTY. I know dolphin-emu-nogui is reading the Dolphin.ini file for some settings, because if I switch audio backend away from ALSA to Pulse, my sound starts working. Similarly, if I change WiimoteContinuousScanning to True, my Wii Remotes start working. If I enable frame counters or framerates, then those show up too in nogui mode.

If I am adding the wrong settings to Dolphin.ini please advise so I can test the correct settings, but I am beginning to wonder if those particular Display settings simply are not processed when running in DRM instead of x11. I can attest that if I make those changes to Dolphin.ini, no matter my resolution on my desktop, dolphin-emu-nogui starts in the correct resolution when running in x11.  So if I'm in 4K on the desktop, Dolphin would start with 1360x768@60Hz, 640x480@60Hz, or any compatible resolution I have tested. I just need DRM to function the same way if possible.

Ps No, I do not really understand why apps launched via a TTY session behave differently than graphical X11 apps or the TTY terminal itself. But my ignorance in this setup is definitely showing. I apologize in advance. The reason I do not just use 1080p is there is definitely a performance hit in that resolution versus a lower resolution. My older laptop is fast enough to play many GameCube/Wii games but not all within Dolphin, turning the resolution down definitely makes the system perform better and makes more of the library usable.

i30817:
I'm talking about the emulated wiimote, i don't have the actual one. Wii remote is a unique controller so it isn't surprising that edev supports it, the problem was when you tried to configure keys and used the touchpad as a pointer/tilt axis at the same time.

silvarios:
@ i30817 
I get it now. Thanks for the clarification.

camdenorrb:
Any updates on this? Spent a few days on a similar thing

camdenorrb:
https://github.com/dolphin-emu/dolphin/pull/13222

PR #9016, opened 2020-08-07
JitArm64: Avoid double rounding in fctiwzx

JosJuice:
FCVT doesn't necessarily round to zero, so the result might be inaccurate if we use it. To ensure correct rounding, we use FCVTS from double FPR to 32-bit GPR. Unfortunately, FCVTS can't do double FPR to single FPR.

This is my first time making a JitArm64 PR, so apologies if the change doesn't make sense :)

### DIFF
--- a/Source/Core/Core/PowerPC/JitArm64/JitArm64_FloatingPoint.cpp
+++ b/Source/Core/Core/PowerPC/JitArm64/JitArm64_FloatingPoint.cpp
@@ -339,8 +339,12 @@ void JitArm64::fctiwzx(UGeckoInstruction inst)
   }
   else
   {
-    m_float_emit.FCVT(32, 64, EncodeRegToDouble(VD), EncodeRegToDouble(VB));
-    m_float_emit.FCVTS(EncodeRegToSingle(VD), EncodeRegToSingle(VD), ROUND_Z);
+    ARM64Reg V1 = gpr.GetReg();
+
+    m_float_emit.FCVTS(V1, EncodeRegToDouble(VB), ROUND_Z);
+    m_float_emit.FMOV(EncodeRegToSingle(VD), V1);
+
+    gpr.Unlock(V1);
   }
   m_float_emit.ORR(EncodeRegToDouble(VD), EncodeRegToDouble(VD), EncodeRegToDouble(V0));
   fpr.Unlock(V0);


Tilka:
(The reason we cannot use FCVT is not that it doesn't always round-towards-zero. The problem is it only converts between different floating-point formats but fctiwzx converts to integer format.)

LGTM. Please test though.

JosJuice:
That is the reason why we can't use just FCVT, forgoing FCVTS. What I am arguing for is why we shouldn't use FVCT in combination with FCVTS.

I have tested two games that break when intentionally messing up the emitted code for this instruction, and they seem to work fine.

Source/Core/Core/PowerPC/JitArm64/JitArm64_FloatingPoint.cpp
```diff
@@ -339,8 +339,12 @@ void JitArm64::fctiwzx(UGeckoInstruction inst)
   }
   else
   {
-    m_float_emit.FCVT(32, 64, EncodeRegToDouble(VD), EncodeRegToDouble(VB));
-    m_float_emit.FCVTS(EncodeRegToSingle(VD), EncodeRegToSingle(VD), ROUND_Z);
+    ARM64Reg V1 = gpr.GetReg();
+
+    m_float_emit.FCVTS(V1, EncodeRegToDouble(VB), ROUND_Z);
+    m_float_emit.FMOV(EncodeRegToSingle(VD), V1);
```
degasus, line 345: Why do you need a temp register here?
```
    m_float_emit.FCVTS(EncodeRegToSingle(VD), EncodeRegToDouble(VB), ROUND_Z);
    m_float_emit.FMOV(EncodeRegToSingle(VD), EncodeRegToSingle(VD));
```

Tilka: My understanding is that the FCVT*S instruction family supports:
- f64-to-s32 with the result in a general purpose register
- f64-to-s64 with the result in a SIMD&FP register
but we need f64-to-s32 and the result in a SIMD&FP register.

degasus: Oh, I have not noticed that this temporary register was a GPR one, my bad.

PR #9018, opened 2020-08-08
Wii TAS input: fix offset in IR coordinates

Felk:
The TAS input currently has some weird placement for its made-up IR lights. This causes an offset while the TAS UI suggests the cursor should be perfectly centered.

Here are before and after screenshots for visualization:

ir1
ir2

Also: Do you think adding a fixed offset to the y-coordinate is ok? This will probably make the offset _worse_ for people who changed their sensor bar's position settings to "below screen", but I don't know if people even do that.

### DIFF
--- a/Source/Core/Core/HW/WiimoteEmu/Camera.cpp
+++ b/Source/Core/Core/HW/WiimoteEmu/Camera.cpp
@@ -70,6 +70,11 @@ void CameraLogic::Update(const Common::Matrix44& transform)
   if (!IOS::g_gpio_out[IOS::GPIO::SENSOR_BAR])
     return;
 
+  WriteIRDataForTransform(data.data(), m_reg_data.mode, transform);
+}
+
+void CameraLogic::WriteIRDataForTransform(u8* data, u8 mode, const Common::Matrix44& transform)
+{
   using Common::Matrix33;
   using Common::Matrix44;
   using Common::Vec3;
@@ -127,7 +132,7 @@ void CameraLogic::Update(const Common::Matrix44& transform)
     return CameraPoint{{0xffff, 0xffff}, 0xff};
   });
 
-  switch (m_reg_data.mode)
+  switch (mode)
   {
   case IR_MODE_BASIC:
     for (std::size_t i = 0; i != camera_points.size() / 2; ++i)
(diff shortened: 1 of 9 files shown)


jordan-woyak:
Couldn't this be fixed properly by using the same routine that the emulated remote's `CameraLogic` class uses?

JosJuice:
The current value of the sensor bar position setting should ideally be taken into account for the y coordinate. This is how it's currently handled by the regular IR code: https://github.com/dolphin-emu/dolphin/blob/75e79ece73afaf86c45f83f382e97f7983288467/Source/Core/Core/HW/WiimoteEmu/Dynamics.cpp#L236

However, I'm not sure if that approach works correctly in the case where the value of the setting in the temporary TASing NAND differs from the value of the setting in the normal NAND. Since the SYSCONF in the temporary TAS has most values set to the default, it may be better to not take the setting into account than to use an approach which takes the wrong setting into account.

Felk:
> Couldn't this be fixed properly by using the same routine that the emulated remote's `CameraLogic` class uses?

Looks like it could, Ideally by just extracting the conversion from `Matrix44` to bytes to some InputCommon utility I suppose?

Felk:
If I change the code to use an emulated position of the wiimote, having the TAS window go from 0-1024 for x and 0-768 for y doesn't make much sense anymore. Here's what I came up with instead:
grafik

That's now just arbitrarily goes from -100 to 100, which is the wiimote position's offset in centimeters. The additional slider below  that is currently set to 500 is the distance from the IR sensor LEDs in centimeters (5 meters by default), and arbitrarily goes from 50 to 10000 (10 meters). I did not apply any offset in y-direction this time, so setting x and y to 0 causes the in-game cursor to be slighly above the center again.

Would this be acceptable?

jordan-woyak:
@Felk 
The "x" and "y" should probably apply rotations instead of centimeter offsets.
I would make the default distance the same as that of the emulated remote (2 meters).

Felk:
With x and y being rotation in arc minutes:
grafik
grafik

Dimon12321:
Thank you for posting this issue. I have the same trouble with Resident Evil 4 for Wii which sometimes goes out of sync on specific IRs. 1x and 2x works fine, but 3x desyncs.

Felk:
> Thank you for posting this issue. I have the same trouble with Resident Evil 4 for Wii which sometimes goes out of sync on specific IRs. 1x and 2x works fine, but 3x desyncs.

I don't think that this merge request will affect any TAS desyncing issues, as it's basically just a code refactor and gui change

JMC47:
changing IR to games that read the framebuffer is expected to cause desyncs.

jordan-woyak:
The "made-up IR lights" issue has since been resolved by TAS using the "input override" system. These changes are no longer needed. Thank you.

PR #9019, opened 2020-08-08
VolumeVerifier: Split TMD error from ticket error for WADs

JosJuice:
When I first made VolumeVerifier, I figured that the distinction between an unsigned ticket and an unsigned TMD was a technical detail that users would have no reason to care about. However, while this might be true for discs, it isn't equally true for WADs, due to the widespread practice of fakesigning tickets to set the console ID to 0. This practice does not require fakesigning the TMD (though apparently people do it anyway, at least sometimes...), and the presence of a correctly signed TMD is a useful indicator that the contents have not been tampered with, even if the ticket isn't correctly signed.

### DIFF
--- a/Source/Core/DiscIO/VolumeVerifier.cpp
+++ b/Source/Core/DiscIO/VolumeVerifier.cpp
@@ -391,8 +391,6 @@ void VolumeVerifier::Start()
       (m_volume.GetVolumeType() == Platform::WiiDisc && !m_volume.IsEncryptedAndHashed()) ||
       IsDebugSigned();
 
-  if (m_volume.GetVolumeType() == Platform::WiiWAD)
-    CheckCorrectlySigned(PARTITION_NONE, Common::GetStringT("This title is not correctly signed."));
   CheckDiscSize(CheckPartitions());
   CheckMisc();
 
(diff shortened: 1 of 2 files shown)


Source/Core/DiscIO/VolumeVerifier.cpp
```diff
@@ -525,10 +523,24 @@ bool VolumeVerifier::CheckPartition(const Partition& partition)
...
+            es->VerifyContainer(IOS::HLE::Device::ES::VerifyContainerType::Ticket,
+                                IOS::HLE::Device::ES::VerifyMode::DoNotUpdateCertStore,
+                                m_volume.GetTicket(partition), cert_chain) ||
+        IOS::HLE::IPC_SUCCESS !=
+            es->VerifyContainer(IOS::HLE::Device::ES::VerifyContainerType::TMD,
+                                IOS::HLE::Device::ES::VerifyMode::DoNotUpdateCertStore,
+                                m_volume.GetTMD(partition), cert_chain))
+    {
+      AddProblem(Severity::Low,
+                 std::move(StringFromFormat(
```
BhaaLseN, line 540: Do you need this `std::move` here? I realize the original `CheckCorrectlySigned` method had it, but the argument would be to either use it consistently in all places or not at all (if it is elided anyways).

The alignment `if` right before this doesn't have it for example, nor do the newly added checks for alignment; but then we have others like the integrity check `if` right afterwards that does.

JosJuice: This `std::move` is unnecessary. I just hadn't noticed it was there.

The integrity check uses `std::move` because the string is stored into a variable, which was done because the code otherwise would be formatted like this:

```
AddProblem(
    Severity::Low,
    StringFromFormat(
        Common::GetStringT("The H3 hash table for the %s partition is not correct.").c_str(),
        name.c_str()));
```

Tilka:
Tested, works.

PR #9020, opened 2020-08-08
IOS/USB: fix potential race condition

Tilka:


degasus:
Just to be sure: The StopThreads change swapps the order of child ressource destruction and thread finialization. Is there a chance that the thread might still be within some virtual methods?

PR #9025, opened 2020-08-11
Fix builds with system LZMA when using versions of CMake before 3.14

cyrozap:
The "FindLibLZMA.cmake" module in CMake versions prior to 3.14 do not
set an alias like how "Externals/liblzma/CMakeLists.txt" does, so builds
performed using one of those older CMake versions will fail if the
system LZMA library is detected. To fix this, we need to link against
"lzma" instead of "LibLZMA::LibLZMA".

This is an update of #9017, with the zstd build changes removed and the commit message altered to make it more clear what problem it's fixing.

This patch is needed to fix builds on Debian 10 (buster).

### DIFF
--- a/Source/Core/DiscIO/CMakeLists.txt
+++ b/Source/Core/DiscIO/CMakeLists.txt
@@ -59,7 +59,7 @@ add_library(discio
 target_link_libraries(discio
 PUBLIC
   BZip2::BZip2
-  LibLZMA::LibLZMA
+  lzma
   zstd
 
 PRIVATE


JosJuice:
@dolphin-emu-bot rebuild

orbea:
Seems to work for `cmake-3.18.2` too.
```
-- Looking for lzma_auto_decoder in /usr/lib64/liblzma.so
-- Looking for lzma_auto_decoder in /usr/lib64/liblzma.so - found
-- Looking for lzma_easy_encoder in /usr/lib64/liblzma.so
-- Looking for lzma_easy_encoder in /usr/lib64/liblzma.so - found
-- Looking for lzma_lzma_preset in /usr/lib64/liblzma.so
-- Looking for lzma_lzma_preset in /usr/lib64/liblzma.so - found
-- Found LibLZMA: /usr/lib64/liblzma.so (found version "5.2.5") 
-- Looking for lzma.h
-- Looking for lzma.h - found
-- Using shared lzma
```
```
$ ldd Binaries/dolphin-emu | grep -i lzma
	liblzma.so.5 => /lib64/liblzma.so.5 (0x00007f089f480000)
```

PR #9027, opened 2020-08-11
Improve behavior of savestates made while game is assuming the state of the inserted memory card.

AdmiralCurtiss:
Fixes https://bugs.dolphin-emu.org/issues/12220

This looks pretty convoluted but I hope the comments make it clear what's happening and why.


Source/Core/Core/HW/EXI/EXI.cpp
```diff
@@ -152,15 +152,15 @@ static void ChangeDeviceCallback(u64 userdata, s64 cyclesLate)
 }
 
 void ChangeDevice(const u8 channel, const TEXIDevices device_type, const u8 device_num,
-                  CoreTiming::FromThread from_thread)
+                  CoreTiming::FromThread from_thread, u32 delay)
```
lioncash, line 155: Perhaps `delay` should be a `std::chrono` type, so that the expected time unit is made evident in the interface?

AdmiralCurtiss: It's CPU cycles, actually, which we always just use an integer for. I've made it more clear.

AdmiralCurtiss: Though to be clear, I'm not opposed to adding a custom std::chrono type for CPU cycles, but that's probably for a different PR.

Source/Core/Core/HW/EXI/EXI_Channel.cpp
```diff
@@ -238,48 +240,87 @@ void CEXIChannel::DoState(PointerWrap& p)
...
+      // changed and it should not assume that the data (especially the file system blocks) it may
+      // or may not have read before is still valid.
+
+      // To accomplish this we do the following:
+
+      CEXIMemoryCard* card_device = static_cast<CEXIMemoryCard*>(m_devices[device_index].get());
+
+      if (p.GetMode() != PointerWrap::MODE_READ)
+      {
+        // In any of the non-modifying modes, we just read the file system blocks from the currently
```
BhaaLseN, line 274: You say "non-modifying modes" down here, but the `if` uses `mode != READ` - this seems a bit confusing. Any thoughts on reversing the condition and rewriting the comment to reflect the same intent?

AdmiralCurtiss: `MODE_READ` means 'read from state stream, write to passed data'... but yes, I see what you mean. I'll think about this.

AdmiralCurtiss: Better?

JMC47:
Considering this prevents save corruption, we should merge this once it is reviewed and good to go.

JMC47:
Is this still planning to be merged?

AdmiralCurtiss:
Considering I've yet to see anyone yell about corrupted savedata in the many years this issue has potentially existed, it's probably not particularly critical. If you want to merge it sure, if not eh, that's probably okay. I'll rebase it tomorrow just in case.

Rumi-Larry:
pOkE

AdmiralCurtiss:
There was a request for me to rebase this on Discord, so I did.

I will note, however, that looking back on this I'm not 100% sure this whole DisableWrites() -> wait for a few frames approach is even necessary at all. It's *probably* okay but it would probably also work just fine if we, immediately after loading the state is done but before handing back to the actual emulation logic, just run the

```
      m_status.EXTINT = 1;
      m_system.GetExpansionInterface().UpdateInterrupts();
```

code in the relevant `CEXIChannel` so that the game is notified of the 'unplugged' memory card before it has the chance to run any writes.


I will also note that we should probably fix folder memory cards not storing their card ID / header before considering merging this, as this will make the situtation where you suddenly have a 'different' memory card after loading a savestate much more likely.

PR #9028, opened 2020-08-14
DolphinQt: Prevent deadlock when exiting the NetPlay Session Browser dialog

cristian64:
It was possible to enter a deadlock when the **NetPlay Session Browser** dialog was closed while a refresh was still ongoing.

**Reproduction steps:**

- Launch Dolphin.
- Open the **NetPlay Session Browser** dialog via **Tools > Browse NetPlay Sessions...**.
- Press `ESC` as soon as the dialog pops up, so it is dismissed straight away.
- Dolphin will (potentially) deadlock.

Stacktrace (Release build):
```
Thread 1 (Thread 0x7f3f319e6080 (LWP 4361)):
#0  0x00007fa6be274d2d in __GI___pthread_timedjoin_ex (threadid=140352865023744, thread_return=0x0, abstime=0x0, block=<optimised out>) at pthread_join_common.c:89
#1  0x00007fa6bdfa0933 in std::thread::join() () at /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#2  0x000055c4f2f996ea in NetPlayBrowser::~NetPlayBrowser() ()
#3  0x000055c4f2f99799 in NetPlayBrowser::~NetPlayBrowser() ()
#4  0x00007fa6c50acf31 in QDialog::exec() () at /usr/lib/x86_64-linux-gnu/libQt5Widgets.so.5
#5  0x000055c4f2e993d2 in MainWindow::ShowNetPlayBrowser() ()
...

Thread 14 (Thread 0x7f8d41ffb700 (LWP 27323)):
#0  0x00007f8d7f8fb9f3 in futex_wait_cancelable (private=<optimised out>, expected=0, futex_word=0x7f8d41ff1430) at ../sysdeps/unix/sysv/linux/futex-internal.h:88
#1  0x00007f8d7f8fb9f3 in __pthread_cond_wait_common (abstime=0x0, mutex=0x7f8d41ff1438, cond=0x7f8d41ff1408) at pthread_cond_wait.c:502
#2  0x00007f8d7f8fb9f3 in __pthread_cond_wait (cond=0x7f8d41ff1408, mutex=0x7f8d41ff1438) at pthread_cond_wait.c:655
#3  0x00007f8d7f61c8bc in std::condition_variable::wait(std::unique_lock<std::mutex>&) () at /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#4  0x000055b844a9ea5f in Common::Event::Wait() [clone .part.75] ()
#5  0x000055b844aa2710 in NetPlayBrowser::RefreshLoop() ()
#6  0x00007f8d7f6226df in  () at /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#7  0x00007f8d7f8f56db in start_thread (arg=0x7f8d41ffb700) at pthread_create.c:463
#8  0x00007f8d7eef7a3f in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95
...
```

When the `NetPlayBrowser::~NetPlayBrowser()` is invoked, it attempts to cancel the network thread, and waits for it to return control. However, if a refresh request was still in progress, the network thread might still attempt to update some widgets via `RunOnObject()`. Therefore, Qt's event loop remains busy, waiting for the network thread to yield control, and the network thread is waiting for Qt to run a lambda in the object.

### DIFF
--- a/Source/Core/DolphinQt/NetPlay/NetPlayBrowser.cpp
+++ b/Source/Core/DolphinQt/NetPlay/NetPlayBrowser.cpp
@@ -26,7 +26,6 @@
 #include "Core/ConfigManager.h"
 
 #include "DolphinQt/QtUtils/ModalMessageBox.h"
-#include "DolphinQt/QtUtils/RunOnObject.h"
 #include "DolphinQt/Settings.h"
 
 NetPlayBrowser::NetPlayBrowser(QWidget* parent) : QDialog(parent)
@@ -148,6 +147,11 @@ void NetPlayBrowser::ConnectWidgets()
   connect(m_table_widget, &QTableWidget::itemSelectionChanged, this,
           &NetPlayBrowser::OnSelectionChanged);
   connect(m_table_widget, &QTableWidget::itemDoubleClicked, this, &NetPlayBrowser::accept);
+
+  connect(this, &NetPlayBrowser::UpdateStatusRequested, this,
+          &NetPlayBrowser::OnUpdateStatusRequested, Qt::QueuedConnection);
+  connect(this, &NetPlayBrowser::UpdateListRequested, this, &NetPlayBrowser::OnUpdateListRequested,
+          Qt::QueuedConnection);
 }
 
 void NetPlayBrowser::Refresh()
@@ -191,10 +195,7 @@ void NetPlayBrowser::RefreshLoop()
 
       lock.unlock();
 
-      RunOnObject(this, [this] {
-        m_status_label->setText(tr("Refreshing..."));
-        return nullptr;
-      });
+      emit UpdateStatusRequested(tr("Refreshing..."));
 
       NetPlayIndex client;
 
@@ -202,19 +203,12 @@ void NetPlayBrowser::RefreshLoop()
 
       if (entries)
       {
-        RunOnObject(this, [this, &entries] {
-          m_sessions = *entries;
-          UpdateList();
-          return nullptr;
-        });
+        emit UpdateListRequested(std::move(*entries));
       }
       else
       {
-        RunOnObject(this, [this, &client] {
-          m_status_label->setText(tr("Error obtaining session list: %1")
-                                      .arg(QString::fromStdString(client.GetLastError())));
-          return nullptr;
-        });
+        emit UpdateStatusRequested(tr("Error obtaining session list: %1")
+                                       .arg(QString::fromStdString(client.GetLastError())));
       }
     }
   }
@@ -275,6 +269,17 @@ void NetPlayBrowser::OnSelectionChanged()
 {
   m_button_box->button(QDialogButtonBox::Ok)
       ->setEnabled(!m_table_widget->selectedItems().isEmpty());
+}
+
+void NetPlayBrowser::OnUpdateStatusRequested(const QString& status)
+{
+  m_status_label->setText(status);
+}
+
+void NetPlayBrowser::OnUpdateListRequested(std::vector<NetPlaySession> sessions)
+{
+  m_sessions = std::move(sessions);
+  UpdateList();
 }
 
 void NetPlayBrowser::accept()
--- a/Source/Core/DolphinQt/NetPlay/NetPlayBrowser.cpp
+++ b/Source/Core/DolphinQt/NetPlay/NetPlayBrowser.cpp
@@ -35,6 +35,7 @@ NetPlayBrowser::NetPlayBrowser(QWidget* parent) : QDialog(parent)
   setWindowFlags(windowFlags() & ~Qt::WindowContextHelpButtonHint);
 
   CreateWidgets();
+  RestoreSettings();
   ConnectWidgets();
 
   resize(750, 500);
@@ -44,8 +45,6 @@ NetPlayBrowser::NetPlayBrowser(QWidget* parent) : QDialog(parent)
 
   m_refresh_run.Set(true);
   m_refresh_thread = std::thread([this] { RefreshLoop(); });
-
-  RestoreSettings();
 
   UpdateList();
   Refresh();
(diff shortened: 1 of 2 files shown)


Source/Core/DolphinQt/NetPlay/NetPlayBrowser.cpp
```diff
@@ -127,6 +124,8 @@ void NetPlayBrowser::CreateWidgets()
   m_button_box->button(QDialogButtonBox::Ok)->setEnabled(false);
 
   setLayout(layout);
+
+  RestoreSettings();
```
iwubcode, line 128: Just curious, why was this moved?  It seems odd for it to be in `CreateWidgets`

cristian64: It's a minor thing, but I guess I should have mentioned it, or at least added a comment.

`RestoreSettings()` set new values for the widgets, and those widgets trigger `Refresh()` when changed. The idea is that settings need to be restored *before* `ConnectWidgets()`, or else a bunch of `Refresh()` calls will be made (although only the last one will probably win).

I argued with myself whether the `RestoreSettings()` call should be placed after `CreateWidgets()`, or inside `CreateWidgets()`. I thought that, if `CreateWidgets()` was ever to be called again, then `RestoreSettings()` would need to be called again as well, so having it inside makes sense.

However, the same argument could be used for having `ConnectWidgets()` inside `CreateWidgets()`, which we don't have. So, for consistency, `RestoreSettings()` should be invoked directly from the constructor, before `ConnectWidgets()`.

I'll update the PR.

cristian64:
Updated to put the changes regarding `RestoreSettings()` in its own commit.

The function is now (still) being invoked from the constructor, *before* `ConnectWidgets()`, to avoid triggering `Refresh()` multiple times as widgets with change their values and emit their signals.

Source/Core/DolphinQt/NetPlay/NetPlayBrowser.cpp
```diff
@@ -192,30 +195,20 @@ void NetPlayBrowser::RefreshLoop()
...
       auto entries = client.List(filters);
 
       if (entries)
       {
-        RunOnObject(this, [this, &entries] {
-          m_sessions = *entries;
-          UpdateList();
-          return nullptr;
-        });
+        emit UpdateListRequested(*entries);
```
lioncash, line 206:
```suggestion
        emit UpdateListRequested(std::move(*entries));
```

Source/Core/DolphinQt/NetPlay/NetPlayBrowser.cpp
```diff
@@ -278,6 +271,17 @@ void NetPlayBrowser::OnSelectionChanged()
       ->setEnabled(!m_table_widget->selectedItems().isEmpty());
 }
 
+void NetPlayBrowser::OnUpdateStatusRequested(const QString& status)
+{
+  m_status_label->setText(status);
+}
+
+void NetPlayBrowser::OnUpdateListRequested(const std::vector<NetPlaySession>& sessions)
```
lioncash, line 279: This can take the vector by value and then std::move `sessions` into `m_sessions`

Source/Core/DolphinQt/NetPlay/NetPlayBrowser.cpp
cristian64: Done!

Source/Core/DolphinQt/NetPlay/NetPlayBrowser.cpp
cristian64: Done!

Source/Core/DolphinQt/NetPlay/NetPlayBrowser.cpp
```diff
@@ -278,6 +271,17 @@ void NetPlayBrowser::OnSelectionChanged()
       ->setEnabled(!m_table_widget->selectedItems().isEmpty());
 }
 
+void NetPlayBrowser::OnUpdateStatusRequested(QString status)
```
lioncash, line 274: This should  still be taken by const reference, as `setText` in Qt's API is defined to always take a `const QString&`, so moving in this function wouldn't have any benefit.

cristian64: I don't know why I assumed that `setText(QString&&)` would have been overloaded in the API already.

I'll change it back.

cristian64: Done.

cristian64: (To add to that, I guess Qt will not be able to provide `QLabel::setText(QString&&)` without annoyingly breaking users code. Signal connections that refer to that slot as `&QLabel::setText` would become ambiguous all of the sudden, if the slot was overloaded. Maybe they'll be able to do it in a major version.)

Source/Core/DolphinQt/NetPlay/NetPlayBrowser.h
```diff
@@ -35,6 +35,8 @@ class NetPlayBrowser : public QDialog
   void accept() override;
 signals:
   void Join();
+  void UpdateStatusRequested(const QString& status);
+  void UpdateListRequested(const std::vector<NetPlaySession>& sessions);
```
lioncash, line 39:
```suggestion
  void UpdateListRequested(std::vector<NetPlaySession> sessions);
```

or

```suggestion
  void UpdateListRequested(std::vector<NetPlaySession>&& sessions);
```

Otherwise moving into emissions of this signal won't actually perform a move.

cristian64: That makes sense. I should have added the `const&`.

I'll update it.

cristian64: Done.

PR #9029, opened 2020-08-16
vs/msbuild: fix the build generating log files in source dirs

shuffle2:
prevents the source dirs becoming dirty to git / causes log files to be placed correctly.


sepalani:
@dolphin-emu-bot rebuild

PR #9030, opened 2020-08-16
msvc: disable C4789 on specific files to avoid false-positive

shuffle2:
Prevents the following, which appears to be bogus:
```
Wiimote.cpp(343): error C2220: the following warning is treated as an error
Wiimote.cpp(343): warning C4789: buffer 'rpt' of size 1 bytes will be overrun; 2 bytes will be written starting at offset 0
```
After some experimenting, it appears to be a fairly hard to hit edge case (as also evidenced by the fact that no code in the rest of dolphin triggers it). It has to do with the `/permissive-` flag and the exact way the struct is declared, concerning the ordering of named/unnamed fields.

Afterwards (and with Qt 5.15.0), dolphin builds fine on current latest msvc: `Microsoft (R) C/C++ Optimizing Compiler Version 19.27.29111`

### DIFF
--- a/Source/Core/Core/Core.vcxproj
+++ b/Source/Core/Core/Core.vcxproj
@@ -202,7 +202,12 @@
     <ClCompile Include="HW\WiimoteEmu\Speaker.cpp" />
     <ClCompile Include="HW\WiimoteEmu\WiimoteEmu.cpp" />
     <ClCompile Include="HW\WiimoteReal\IOWin.cpp" />
-    <ClCompile Include="HW\WiimoteReal\WiimoteReal.cpp" />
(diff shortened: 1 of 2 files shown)


sepalani:
@dolphin-emu-bot rebuild

AdmiralCurtiss:
Wouldn't it be easier to just name the unnamed field `unused` or `unknown` or something?

shuffle2:
That’s true, I just wanted to do something that’s clearly a work around that should be reverted later. I’ll submit another pr that names the field with a comment and I’ll plan to revert it when msvc is fixed.

PR #9032, opened 2020-08-16
UnitTests: avoid using glob to find compilation units

shuffle2:
Using `*` was always a hack, and now VS issues a warning upon solution/project load, so let's avoid hitting bad behavior by explicitly listing files.

This change exposed some compiler warnings (apparently because some files were not being compiled on windows before?!), so the PR also fixes those.

Also fixes a bug in `IOS::HLE::FS::HostFileSystem::Rename` (comparing to invalidated vector iterator) introduced in a40f297d1d46255ac8e4655148738be82de672d2


JosJuice:
@dolphin-emu-bot rebuild

JosJuice:
@dolphin-emu-bot rebuild

Source/UnitTests/Core/IOS/ES/TestBinaryData.h
```diff
@@ -8,8 +8,170 @@
 #include <cstddef>
 #include "Common/CommonTypes.h"
 
-extern const std::array<u8, 520> soup01_tmd;
-extern const std::array<u8, 108> soup01_tmd_view;
+constexpr std::array<u8, 520> soup01_tmd = {{
```
leoetlino, line 11: Just wondering, what warning does this fix?

shuffle2: `extern const` is not equivalent to the actual storage type (which is `constexpr`) and apparently [extern constexpr is not possible](https://stackoverflow.com/questions/30208685/how-to-declare-constexpr-extern)

shuffle2: Perhaps the existing code could be used, msvc just requires another flag: https://docs.microsoft.com/en-us/cpp/build/reference/zc-externconstexpr?view=vs-2017

note: 
> The /permissive- option does not enable /Zc:externConstexpr.

Source/UnitTests/UnitTests.vcxproj
```diff
@@ -59,16 +59,51 @@
...
+    <ClCompile Include="Common\Crypto\EcTest.cpp" />
+    <ClCompile Include="Common\EventTest.cpp" />
+    <ClCompile Include="Common\FixedSizeQueueTest.cpp" />
+    <ClCompile Include="Common\FlagTest.cpp" />
+    <ClCompile Include="Common\FloatUtilsTest.cpp" />
+    <ClCompile Include="Common\MathUtilTest.cpp" />
+    <ClCompile Include="Common\NandPathsTest.cpp" />
+    <ClCompile Include="Common\SPSCQueueTest.cpp" />
+    <ClCompile Include="Common\StringUtilTest.cpp" />
+    <ClCompile Include="Common\SwapTest.cpp" />
```
Rumi-Larry, line 88: Extra line?

Source/UnitTests/UnitTests.vcxproj
```diff
@@ -59,16 +59,51 @@
...
+    <ClCompile Include="Core\CoreTimingTest.cpp" />
+    <ClCompile Include="Core\DSP\DSPAcceleratorTest.cpp" />
+    <ClCompile Include="Core\DSP\DSPAssemblyTest.cpp" />
+    <ClCompile Include="Core\DSP\DSPTestBinary.cpp" />
+    <ClCompile Include="Core\DSP\DSPTestText.cpp" />
+    <ClCompile Include="Core\DSP\HermesBinary.cpp" />
+    <ClCompile Include="Core\IOS\ES\FormatsTest.cpp" />
+    <ClCompile Include="Core\IOS\FS\FileSystemTest.cpp" />
+    <ClCompile Include="Core\MMIOTest.cpp" />
+    <ClCompile Include="Core\PageFaultTest.cpp" />
```
Rumi-Larry, line 98: ditto here

Source/UnitTests/UnitTests.vcxproj
shuffle2: what?

shuffle2: i think you're indicating i should insert a blank line to delineate the difference between root dirs here? I would prefer not to.

Source/UnitTests/Core/IOS/ES/TestBinaryData.h
leoetlino: I think this is fine, it's not like we're embedding huge amounts of data. I wish [std::embed](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p1040r0.html) was a thing though...

PR #9037, opened 2020-08-18
Code cleanup

shuffle2:
* fixing some stuff that VS' code analysis detected.
* make msvc settings more standard conformant
* re-enable msvc warning for deprecated items, and make build pass
* enables msvc to merge and collect unused functions + data
* makes msvc builds deterministic / reproducible

### DIFF
--- a/Source/Core/Common/CompatPatches.cpp
+++ b/Source/Core/Common/CompatPatches.cpp
@@ -183,7 +183,7 @@ static bool GetModuleVersion(const wchar_t* name, Version* version)
   if (!data_len)
     return false;
   std::vector<u8> block(data_len);
-  if (!GetFileVersionInfoW(path->c_str(), handle, data_len, block.data()))
+  if (!GetFileVersionInfoW(path->c_str(), 0, data_len, block.data()))
     return false;
   void* buf;
   UINT buf_len;
--- a/Source/Core/Common/Timer.cpp
+++ b/Source/Core/Common/Timer.cpp
@@ -254,7 +254,7 @@ std::string Timer::GetDateTimeFormatted(double time)
 
 #ifdef _WIN32
   wchar_t tmp[32] = {};
-  wcsftime(tmp, sizeof(tmp), L"%x %X", localTime);
+  wcsftime(tmp, std::size(tmp), L"%x %X", localTime);
   return WStringToUTF8(tmp);
 #else
   char tmp[32] = {};
--- a/Source/Core/Core/IOS/Network/IP/Top.cpp
+++ b/Source/Core/Core/IOS/Network/IP/Top.cpp
@@ -86,13 +86,12 @@ static constexpr u32 inet_addr(u8 a, u8 b, u8 c, u8 d)
 
 static int inet_pton(const char* src, unsigned char* dst)
 {
-  int saw_digit, octets;
+  int saw_digit = 0;
+  int octets = 0;
+  unsigned char tmp[4]{};
+  unsigned char* tp = tmp;
   char ch;
-  unsigned char tmp[4], *tp;
 
-  saw_digit = 0;
-  octets = 0;
-  *(tp = tmp) = 0;
   while ((ch = *src++) != '\0')
   {
     if (ch >= '0' && ch <= '9')
@@ -927,8 +926,9 @@ IPCCommandResult NetIPTop::HandleRecvFromRequest(const IOCtlVRequest& request)
 IPCCommandResult NetIPTop::HandleGetAddressInfoRequest(const IOCtlVRequest& request)
 {
   addrinfo hints;
+  const bool hints_valid = request.in_vectors.size() > 2 && request.in_vectors[2].size;
 
-  if (request.in_vectors.size() > 2 && request.in_vectors[2].size)
+  if (hints_valid)
   {
     hints.ai_flags = Memory::Read_U32(request.in_vectors[2].address);
     hints.ai_family = Memory::Read_U32(request.in_vectors[2].address + 0x4);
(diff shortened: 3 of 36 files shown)


Source/Core/DiscIO/WIABlob.cpp
```diff
@@ -1235,7 +1235,7 @@ static void RVZPack(const u8* in, OutputParametersEntry* out, u64 bytes_per_chun
       }
 
       const u64 junk_bytes = next_junk_end - current_offset;
-      if (junk_bytes > 0)
+      if (junk_bytes > 0 && seed)
```
shuffle2, line 1238: @JosJuice is this actually possible? tbh I didn't fully understand the function to reason if it's actually possible to reach here with `seed==nullptr`, and if it is, I'm not sure if it can be silently ignored or not?

Source/Core/Common/CompatPatches.cpp
```diff
@@ -183,7 +183,7 @@ static bool GetModuleVersion(const wchar_t* name, Version* version)
   if (!data_len)
     return false;
   std::vector<u8> block(data_len);
-  if (!GetFileVersionInfoW(path->c_str(), handle, data_len, block.data()))
+  if (!GetFileVersionInfoW(path->c_str(), 0, data_len, block.data()))
```
BhaaLseN, line 186: Is this really necessary? The [docs](https://docs.microsoft.com/en-us/windows/win32/api/winver/nf-winver-getfileversioninfow) say "`dwHandle` This parameter is ignored"; and immediately before the Call to [`GetFileVersionInfoSizeW`](https://docs.microsoft.com/en-us/windows/win32/api/winver/nf-winver-getfileversioninfosizew) indicates that "`lpdwHandle` A pointer to a variable that the function sets to zero."

So, same result, except without using a variable.

Where did you get the "must be zero" from?

Source/Core/Common/Timer.cpp
```diff
@@ -254,7 +254,7 @@ std::string Timer::GetDateTimeFormatted(double time)
 
 #ifdef _WIN32
   wchar_t tmp[32] = {};
-  wcsftime(tmp, sizeof(tmp), L"%x %X", localTime);
+  wcsftime(tmp, _countof(tmp), L"%x %X", localTime);
```
BhaaLseN, line 257: I realize this is Windows specific code, but is there a reason why we cannot use [`std::size`](https://en.cppreference.com/w/cpp/iterator/size) for this?

Source/Core/Core/IOS/Network/IP/Top.cpp
```diff
@@ -86,13 +86,11 @@ static constexpr u32 inet_addr(u8 a, u8 b, u8 c, u8 d)
 
 static int inet_pton(const char* src, unsigned char* dst)
 {
-  int saw_digit, octets;
+  int saw_digit = 0, octets = 0;
```
BhaaLseN, line 89: Thoughts on splitting this into two individual declarations? We rarely declare multiple variables like this in other places.
```suggestion
  int saw_digit = 0;
  int octets = 0;
```

Source/Core/Core/IOS/Network/IP/Top.cpp
```diff
@@ -927,8 +925,9 @@ IPCCommandResult NetIPTop::HandleRecvFromRequest(const IOCtlVRequest& request)
 IPCCommandResult NetIPTop::HandleGetAddressInfoRequest(const IOCtlVRequest& request)
 {
   addrinfo hints;
+  bool hints_valid = request.in_vectors.size() > 2 && request.in_vectors[2].size;
```
BhaaLseN, line 928: Could make this `const` while you're at it.
```suggestion
  const bool hints_valid = request.in_vectors.size() > 2 && request.in_vectors[2].size;
```

Source/Core/Common/Timer.cpp
shuffle2: Thanks, I tried converting the `_countof` instances in this commit to `std::size`, but it then triggers narrowing warnings, as the windows apis are using `int` and `DWORD`, whereas `std::size` returns `size_t`. Of course the warnings are harmless, but I'd rather just use the method traditional to the platform here).

Source/Core/Common/CompatPatches.cpp
shuffle2: The tool isn't smart enough to see that it will always be zero (maybe in some cases, it won't be?). In any case, I don't see a problem with just explicitly passing zero.

Source/Core/Common/Timer.cpp
shuffle2: I updated this instance (as param is `size_t` here), tho :)

shuffle2:
I've moved the nullderef thingy to it's own PR, so this PR should be OK to merge.

Source/VSProps/Base.props
```diff
@@ -111,6 +111,12 @@
                 Currently jits use some annoying code patterns which makes this common
       -->
       <DisableSpecificWarnings>4245;%(DisableSpecificWarnings)</DisableSpecificWarnings>
+      <!-- Disable warnings that get tripped when the optimizer gets too smart for it's own good.
+      For instance, it thinks a value can be zero, but actually the value is set from an API like OpenGL
+      which specifies bounds. Technically the compiler is correct, but...
+      4724  potential mod/div by 0
+      -->
+      <DisableSpecificWarnings Condition="'$(DolphinRelease)'=='true'">4724;%(DisableSpecificWarnings)</DisableSpecificWarnings>
```
shuffle2, line 119: fwiw, this complains about `s_ubo_align` in ProgramShaderCache.cpp , specifically when passed to `Common::AlignUp` and `StreamBuffer::Map`, which mod by `s_ubo_align`.

PR #9038, opened 2020-08-18
prevent nullptr deref in WIABlob

shuffle2:
@JosJuice is this actually possible? tbh I didn't fully understand the function to reason if it's actually possible to reach here with seed==nullptr, and if it is, I'm not sure if it can be silently ignored or not?

(PR broken off from #9037 )

### DIFF
--- a/Source/Core/DiscIO/WIABlob.cpp
+++ b/Source/Core/DiscIO/WIABlob.cpp
@@ -1235,7 +1235,7 @@ static void RVZPack(const u8* in, OutputParametersEntry* out, u64 bytes_per_chun
       }
 
       const u64 junk_bytes = next_junk_end - current_offset;
-      if (junk_bytes > 0)
+      if (junk_bytes > 0 && seed)
       {
         PushBack(&entry.main_data, Common::swap32(static_cast<u32>(junk_bytes) | 0x80000000));
         PushBack(&entry.main_data, *seed);


JosJuice:
It should be impossible for this to happen, but the function is complicated enough that I honestly can't explain why right away. I'll look more closely at the code and see if I can write an explanation in a way that makes sense.

JosJuice:
Assume that line 1209, the only line which assigns a non-null value to `seed`, is not executed. This must mean that the two preceding lines, which are the only lines that assign values other than `end_offset` to `next_junk_start` and `next_junk_end`, were not executed either. Because of this, the assignment `non_junk_bytes = next_junk_start - current_offset` is equivalent to `non_junk_bytes = end_offset - current_offset`, meaning that the line `current_offset += non_junk_bytes` sets `current_offset` to the same value as `end_offset`. With the knowledge that both `current_offset` and `next_junk_end` are equal to `end_offset`, we can see that `junk_bytes = next_junk_end - current_offset` is equivalent to `junk_bytes = 0`, meaning that `seed` cannot be dereferenced.

JosJuice:
In case we were to somehow have `junk_bytes > 0 && !seed` anyway, something is probably catastrophically wrong, and we might either output a nonsense file or get an infinite loop. Perhaps it would make sense to add an assertion.

shuffle2:
Thanks for the explanation, I don't think any change is actually required here.

PR #9039, opened 2020-08-19
windows: keep an extra reference to BluetoothApis.dll

shuffle2:
Prevents BluetoothApis.dll being continuously loaded/unloaded which is 1) very annoying during debugging 2) probably somewhat of a speed hit.

### DIFF
--- a/Source/Core/Core/HW/WiimoteReal/IOWin.cpp
+++ b/Source/Core/Core/HW/WiimoteReal/IOWin.cpp
@@ -24,6 +24,7 @@
 
 #include "Common/CommonFuncs.h"
 #include "Common/CommonTypes.h"
+#include "Common/DynamicLibrary.h"
 #include "Common/Logging/Log.h"
 #include "Common/ScopeGuard.h"
 #include "Common/Thread.h"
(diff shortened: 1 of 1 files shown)


AdmiralCurtiss:
Ah, thanks for that. That one definitely annoyed me too but I couldn't figure out where it came from.

Pokechu22:
This one annoyed me too.  There is actually a way to disable that message in VS, though, under [Debugging &rarr; Output Window](https://user-images.githubusercontent.com/8334194/90669800-084af500-e207-11ea-8168-91f59a2595af.png) (Exception messages and threads stopping can also be disabled there).

I'm not sure if it staying open forever is a good thing; theoretically it should be closed when Dolphin exists, but I'm guessing that's something that automatically happens when the process closes.

AdmiralCurtiss:
Static variables get destructed when the program exits, so that's fine.

(And really, it's program exit, the OS will take care of all our garbage anyway...)

shuffle2:
> This one annoyed me too. There is actually a way to disable that message in VS, though, under [Debugging → Output Window](https://user-images.githubusercontent.com/8334194/90669800-084af500-e207-11ea-8168-91f59a2595af.png) (Exception messages and threads stopping can also be disabled there).

True, but I think it's beneficial to just avoid actually running the dynamic linker all the time.

JMC47:
Works on my machine.

PR #9046, opened 2020-08-22
Update Windows build to use Qt 5.15.0.

AdmiralCurtiss:
Let's see if this works!

### DIFF
--- a/CMakeSettings.json
+++ b/CMakeSettings.json
@@ -11,7 +11,7 @@
       "variables": [
         {
           "name": "Qt5_DIR",
-          "value": "${workspaceRoot}\\Externals\\Qt\\Qt5.14.1\\msvc2019_64\\lib\\cmake\\Qt5"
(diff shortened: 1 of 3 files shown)


MayImilae:
@AdmiralCurtiss Did this bring any new features or fixes of note?

shuffle2:
I _think_ it improved DPI scaling stuff (see https://github.com/dolphin-emu/ext-win-qt/pull/14#issuecomment-674591211 ) , but the older version I was comparing to was a bit dated, so not sure if it was because of this Qt update or not.

MayImilae:
@shuffle2 The old version you showed is definitely from Qt 5.13.1. We've had users complain abouot 5.14.1 making the UI larger like that.

JosJuice:
@MayImilae I suppose the most notable thing is the fixing of https://bugs.dolphin-emu.org/issues/12014.

AdmiralCurtiss:
The big 'fix' I care about here is that Dolphin now actually builds again out-of-the-box when using an up-to-date Visual Studio.

PR #9047, opened 2020-08-22
DiscIO: Adjust WIA/RVZ header size heuristic

JosJuice:
The heuristic was not allocating enough space for Metroid: Other M, at least when using the default settings. (This didn't break the file, it just caused some headers to be placed at the end of the file instead of at the start and wasted a few hundred kilobytes.)

### DIFF
--- a/Source/Core/DiscIO/WIABlob.cpp
+++ b/Source/Core/DiscIO/WIABlob.cpp
@@ -1752,7 +1752,7 @@ WIARVZFileReader<RVZ>::Convert(BlobReader* infile, const VolumeDisc* infile_volu
 
     // RVZ's added data in GroupEntry usually compresses well
     if (RVZ && compression_type > WIARVZCompressionType::Purge)
-      upper_bound += group_entries_size / 2;
+      upper_bound += static_cast<u64>(group_entries_size) * 9 / 16;
     else
       upper_bound += group_entries_size;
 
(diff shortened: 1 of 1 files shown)


AdmiralCurtiss:
Other games I've seen that don't have enough space allocated with the previous heuristic:

- Arc Rise Fantasia (U)
- Kirby's Epic Yarn (E)
- Little King's Story (both U and E)
- Metroid Prime 3 (U)
- Metroid Prime Trilogy (U)
- Star Wars The Force Unleashed (E)
- Xenoblade (E)

All of them get enough space with this one.

Source/Core/DiscIO/WIABlob.cpp
```diff
@@ -1752,7 +1752,7 @@ WIARVZFileReader<RVZ>::Convert(BlobReader* infile, const VolumeDisc* infile_volu
 
     // RVZ's added data in GroupEntry usually compresses well
     if (RVZ && compression_type > WIARVZCompressionType::Purge)
-      upper_bound += group_entries_size / 2;
+      upper_bound += static_cast<u64>(group_entries_size) * 9 / 16;
```
jordan-woyak, line 1755: I think there should be some kind of explanation in a comment here.
Lots of magic going on here.

JosJuice: I've expanded the comments.

Rumi-Larry:
I think a named constant would be better for the long term. It could be something like RVZ_compression_ratio

JosJuice:
The problem is that if I put `9 / 16` in a constant, either I would need to unnecessarily make it a float or it would end up getting rounded to 0.

PR #9048, opened 2020-08-22
require msvc 19.27.29111 and Windows SDK 10.0.19041

shuffle2:
extracted from #9037 again, we'll see if it builds

### DIFF
--- a/Source/PCH/pch.h
+++ b/Source/PCH/pch.h
@@ -2,6 +2,27 @@
 // Licensed under GPLv2+
 // Refer to the license.txt file included.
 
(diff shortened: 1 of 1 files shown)


shuffle2:
Added some logging to display current version of whatever doesn't pass the checks.
Buildbot is updated now.

leoetlino (approved):
sgtm, requiring recent-ish versions reduces possible differences between CI builds and user builds

shuffle2:
This actually took so long to get merged that the versions should be updated... buildbot is already ahead

PR #9051, opened 2020-08-25
Corrected a number of embarassing grammar errors for Norwegian Bokmål

DandelionSprout:
Hello. Since my request to join the Transifex project still hasn't been accepted after 3 **YEARS**, I figured I should try to fix a number of pretty horrid and embarassing grammar and spelling errors in the Norwegian Bokmål translation here on GitHub instead.

Highlights that I felt I had to correct, include (but are not limited to):
* "addresse" → "adresse"
* Missing dashes.
* "tablett" (The term for medical pills) → "tegnebrett" (drawing-tablet)
* "liket" (The term for "The corpse") → "Likhet" (equality)
* Unusually frequent use of capitalisation.
* GameTDB being spelt wrong.
* "instilling" → "innstilling"
* "venligst" → "vennligst"
* The same term ("port") having been used for both controller ports and memory card slots.
* Wrong use of spaces in terms that should've been conjoined.

### DIFF
--- a/Languages/po/nb.po
+++ b/Languages/po/nb.po
@@ -6,6 +6,7 @@
 # Allan Nordhøy <epost@anotheragency.no>, 2016-2018
 # Christer Hansen <chriztr@gmail.com>, 2013
 # Specced, 2019
+# Imre Kristoffer Eilertsen, 2020
 # KHRZ <e-k-nut@hotmail.com>, 2013-2016,2019
 # KHRZ <e-k-nut@hotmail.com>, 2011,2013
 # KHRZ <e-k-nut@hotmail.com>, 2011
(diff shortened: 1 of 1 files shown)


JosJuice:
Sorry, but we can't accept any translation updates through git commits, since the changes would get overwritten as soon as the translations are synced with Transifex the next time. Please send me a private message on either Transifex or the Dolphin forums (I have the same username there as here) and let me know your Transifex username and I'll see what I can do about getting you into the Transifex project. If neither the forum nor Transifex allow you to send PMs because you're a new user (I forget what rules there are for that on Transifex), I suppose you can reply to this PR with your Transifex username instead.

Most of the changes in this PR that I was able to review looked fine (though my Norwegian really isn't that great, so I wouldn't say that I was able to review the majority of them), but one thing to keep in mind: Make sure to read the "i18n" comment if there is one. In particular, "Vanlig" is not a correct translation of the string "Common" in the context Dolphin uses it ;)

PR #9052, opened 2020-08-27
Android: Better GCAdapter scanning thread management

Ebola16:
Previously, the GCAdapter scanning thread would only be started if the GameCube adapter was enabled when starting the app. That's obviously a problem.

### DIFF
--- a/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/model/Settings.java
+++ b/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/model/Settings.java
@@ -232,6 +232,7 @@ public class Settings
       NativeLibrary.ReloadConfig();
       NativeLibrary.ReloadWiimoteConfig();
       NativeLibrary.ReloadLoggerConfig();
+      NativeLibrary.UpdateGCAdapterScanThread();
 
       if (modifiedSettings.contains(SettingsFile.KEY_RECURSIVE_ISO_PATHS))
       {
--- a/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/NativeLibrary.java
+++ b/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/NativeLibrary.java
@@ -376,6 +376,8 @@ public final class NativeLibrary
 
   public static native void ReloadConfig();
 
+  public static native void UpdateGCAdapterScanThread();
+
   /**
    * Initializes the native parts of the app.
    *
--- a/Source/Android/jni/MainAndroid.cpp
+++ b/Source/Android/jni/MainAndroid.cpp
@@ -50,6 +50,7 @@
 
 #include "InputCommon/ControllerInterface/Android/Android.h"
 #include "InputCommon/ControllerInterface/Touch/ButtonManager.h"
+#include "InputCommon/GCAdapter.h"
 
 #include "UICommon/UICommon.h"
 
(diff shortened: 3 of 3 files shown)


Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/model/Settings.java
```diff
@@ -249,6 +260,32 @@ public void saveSettings(SettingsActivityView view, Context context, Set<String>
...
+  public static void manageGCAdapterScanThread()
+  {
+    boolean useAdapter = false;
+
+    for (int i = 0; i < 4; i++)
+    {
+      if (NativeLibrary
+              .GetConfig(SettingsFile.FILE_NAME_DOLPHIN + ".ini", Settings.SECTION_INI_CORE,
+                      SettingsFile.KEY_GCPAD_TYPE + i, InputOverlay.EMULATED_GAMECUBE_CONTROLLER)
+              .equals(InputOverlay.GAMECUBE_ADAPTER))
```
JosJuice, line 272: This function should be written in C++. That way, you can check ConfigManager's m_SIDevice array instead of having to read from the INI file.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/activities/EmulationActivity.java
```diff
@@ -411,6 +411,7 @@ protected void onPause()
   protected void onStop()
   {
     super.onStop();
+    Settings.manageGCAdapterScanThread();
```
JosJuice, line 414: Why is this called here?

JosJuice (commented):
Oh, so *that's* why I couldn't get my GC adapter to work on the first try :)

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/activities/EmulationActivity.java
Ebola16: If the GCAdapter is used when emulation starts and then changed to a different controller, the GCAdapter scanning thread may no longer be needed. Selecting the GCAdapter outside of emulation will restart the scanning thread if needed. GCAdapter changes during emulation (which currently aren't an option) will be handled by #8894.

Ebola16: I should probably move this to after making a Wii Overlay controller change.

JosJuice: I don't get it, though. If GCAdapter changes during emulation aren't an option, how could it be changed into a different controller during emulation? And why should the thread stop in `onStop` instead of as soon as the controller changes?

Ebola16: You can't select the GameCube Adapter during emulation but you can change the controller to something else (like a Classic Controller), making the scanning thread unnecessary. Agreed on the last sentence.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/model/Settings.java
JosJuice: Now that I look, it seems like there already is a C++ function that does this checking: https://github.com/dolphin-emu/dolphin/blob/bed2d66bedca7e9537d67d3e7a5506aee8717c8f/Source/Core/InputCommon/GCAdapter_Android.cpp#L378-L385

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/activities/EmulationActivity.java
Ebola16: Currently, changing the controller doesn't update `m_SIDevice`. This is a more complicated change so I'd rather handle this in  #8894 now.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/model/Settings.java
Ebola16: Better?

JosJuice (approved):
Unsure whether "manage" is the best way to name the function, but that's a pretty minor remark. LGTM.

Ebola16:
Would "UpdateGCAdapterScanThread" be better?

PR #9055, opened 2020-08-29
AudioInterface: Use 32029/48043 Hz in more places

JosJuice:
In particular, I wanted to do change this in AudioInterface::Init so that dumped GC audio doesn't need to have a file split (changing from 32000 Hz to 32029 Hz) when the emulated software initializes the AI registers. I've also made the same change to DI's DTK code.

### DIFF
--- a/Source/Core/AudioCommon/AudioCommon.cpp
+++ b/Source/Core/AudioCommon/AudioCommon.cpp
@@ -15,6 +15,7 @@
 #include "Common/FileUtil.h"
 #include "Common/Logging/Log.h"
 #include "Core/ConfigManager.h"
+#include "Core/HW/AudioInterface.h"
 
 // This shouldn't be a global, at least not here.
 std::unique_ptr<SoundStream> g_sound_stream;
@@ -66,6 +67,12 @@ void InitSoundStream()
     g_sound_stream = std::make_unique<NullSound>();
     g_sound_stream->Init();
   }
+
+  // Ideally these two calls would be done in AudioInterface::Init so that we don't
+  // need to have a dependency on AudioInterface here, but this has to be done
+  // after creating g_sound_stream (above) and before starting audio dumping (below)
+  g_sound_stream->GetMixer()->SetDMAInputSampleRate(AudioInterface::GetAIDSampleRate());
+  g_sound_stream->GetMixer()->SetStreamInputSampleRate(AudioInterface::GetAISSampleRate());
 
   UpdateSoundStream();
   SetSoundStreamRunning(true);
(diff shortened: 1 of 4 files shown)


JosJuice:
@iwubcode I had to add another commit to actually get rid of the file split. Could you re-review?

Source/Core/Core/HW/DVD/DVDInterface.cpp
```diff
@@ -294,8 +294,11 @@ static u32 AdvanceDTK(u32 maximum_samples, u32* samples_to_process)
 static void DTKStreamingCallback(DIInterruptType interrupt_type, const std::vector<u8>& audio_data,
                                  s64 cycles_late)
 {
+  // TODO: Should we use GetAISSampleRate instead of Get48KHzSampleRate?
```
iwubcode, line 297: I'm assuming you intend to investigate this at a later date?

JosJuice: It would involve some kind of hardware testing that I'm not sure how to do. The issue (if it even is an issue) is present in master as well, so I added the comment just for clarity.

Source/Core/AudioCommon/AudioCommon.cpp
```diff
@@ -67,6 +68,12 @@ void InitSoundStream()
     g_sound_stream->Init();
   }
 
+  // Ideally these two calls would be done in AudioInterface::Init so that we don't
+  // need to have a dependency on AudioInterface here, but this has to be done
+  // after creating g_sound_stream (above) and starting audio dumping (below)
```
iwubcode, line 73: One thought is that `InitSoundStream` could take the rates?  And then `HW` could expose them via getters and those values could be passed in `Core.cpp`.  Not sure if that's worth it though?

JosJuice: Yeah, it would work but I'm not sure if it's worth it. Does anyone else have an opinion on this?

Source/Core/Core/HW/DVD/DVDInterface.cpp
lioncash: I think the comment should expand on why we might want to use `GetAISSampleRate` instead, so anyone else that wants to investigate in some form has all the reasoning behind the TODO.

JosJuice: I've reworded the comment in a way that hopefully should make the reason clearer.

BhaaLseN (approved):

lioncash (approved):

PR #9057, opened 2020-08-30
Android: Make DSP setting work with custom GameSettings

Ebola16:
This is a special case since DSP is actually two settings.


JosJuice:
This is also fixed by PR #8975, for the reference. Though I dunno, I guess adding a workaround in the meantime isn't *too* bad since it's relatively easy to remove...

Ebola16:
I plan on looking at #8975 once #8941 is merged. If you think it will be a while before #8975 is merged it's probably worth merging this.

Ebola16:
@dolphin-emu-bot rebuild

Ebola16:
Closing in favor of #8975.

PR #9058, opened 2020-08-31
Android: Remove unused files

Ebola16:
Let me know if I shouldn't delete the unused files.


JosJuice (approved):

Ebola16:
I was able to remove some more files after the first commit.

PR #9059, opened 2020-09-01
WiimoteReal: actually stop continuous scanning

Tilka:
Under certain conditions, continuous scanning would not actually stop if disabled.


Source/Core/Core/HW/WiimoteReal/WiimoteReal.cpp
```diff
@@ -669,15 +669,19 @@ void WiimoteScanner::ThreadFunc()
...
+
+    // Stop scanning if not in continous mode.
+    auto scan_mode = WiimoteScanMode::SCAN_ONCE;
+    m_scan_mode.compare_exchange_strong(scan_mode, WiimoteScanMode::DO_NOT_SCAN);
+
     // Does stuff needed to detect disconnects on Windows
     for (const auto& backend : m_backends)
       backend->Update();
 
     CheckForDisconnectedWiimotes();
```
jordan-woyak, line 683: I'm worried that not calling `Update` and `CheckForDisconnectedWiimotes` regularly will be a problem (especially on Windows).

Were these functions doing something undesirable on your system when "Continuous Scanning" was off?

shuffle2:
@dolphin-emu-bot rebuild

Source/Core/Core/HW/WiimoteReal/WiimoteReal.cpp
Trihy: Was this issue fixed at all?

My issues with this started with this commit https://github.com/dolphin-emu/dolphin/pull/8575

On previous builds I just could turn off continuous scan and desire effect will be fine.

But after that commit, or continuous scan is somehow always doing something, or dolphin started to reset bt adapter status even when having continuous scan disabled. Which didn't happen before that commit.

jordan-woyak:
After #13970 we no longer continuously invoke `BluetoothRemoveDevice` which I think was the main thing trying to be stopped so I think this PR is no longer needed?

PR #9064, opened 2020-09-06
Android: Make FilePicker act like a normal setting

JosJuice:
The reason why the finish() call was added no longer exists. (Also, there was never a duplicate SettingsActivity as far as I can tell, only a duplicate SettingsFragment.)

Split out from PR #8975.


Ebola16 (approved):

7aychu23:
Hi josjuice Can you Add Cheat Generator On Dolphin Like Citra MMj???

JosJuice:
Probably, once I get time. I don't know if I'll get enough time this year or not.

7aychu23:
Thanks Dude other people make 60fps of every game if we have that

lioncash (approved):

PR #9066, opened 2020-09-07
Android: Add "Default Value" button to SliderSetting AlertDialog

Ebola16:
Quickly returning to the default value and not making users remember that value is good.
Should also be applied to #8961 after merging.

### DIFF
--- a/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/model/view/FloatSliderSetting.java
+++ b/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/model/view/FloatSliderSetting.java
@@ -13,6 +13,12 @@ public class FloatSliderSetting extends SliderSetting
     mDefaultValue = defaultValue;
   }
 
+  @Override
+  public int getDefaultValue()
+  {
+    return Math.round(mDefaultValue);
+  }
+
   public int getSelectedValue(Settings settings)
   {
     float value = settings.getSection(getFile(), getSection()).getFloat(getKey(), mDefaultValue);
--- a/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/model/view/IntSliderSetting.java
+++ b/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/model/view/IntSliderSetting.java
@@ -13,6 +13,12 @@ public final class IntSliderSetting extends SliderSetting
     mDefaultValue = defaultValue;
   }
 
+  @Override
+  public int getDefaultValue()
+  {
+    return mDefaultValue;
+  }
+
   public int getSelectedValue(Settings settings)
   {
     return settings.getSection(getFile(), getSection()).getInt(getKey(), mDefaultValue);
--- a/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/ui/SettingsAdapter.java
+++ b/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/ui/SettingsAdapter.java
@@ -232,6 +232,11 @@ public final class SettingsAdapter extends RecyclerView.Adapter<SettingViewHolde
     builder.setTitle(item.getNameId());
     builder.setView(view);
     builder.setPositiveButton(R.string.ok, this);
+    builder.setNeutralButton(R.string.default_value, (dialogInterface, i) ->
+    {
+      // Overridden below to prevent button from closing the dialog.
+      // This can make manipulating multiple sliders easier.
+    });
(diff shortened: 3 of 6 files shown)


Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/ui/SettingsAdapter.java
```diff
@@ -232,6 +235,10 @@ public void onSliderClick(SliderSetting item, int position)
     builder.setTitle(item.getNameId());
     builder.setView(view);
     builder.setPositiveButton(R.string.ok, this);
+    builder.setNeutralButton(R.string.default_value, (dialogInterface, i) ->
+    {
+      // Overridden below
+    });
```
JosJuice, line 241: Does this really have to be done in two steps? It looks to me like you could do it in one step if you put the `setNeutralButton` call after the `SeekBar` code and the `builder.show()` call after the `setNeutralButton` call.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/model/view/IntSliderSetting.java
```diff
@@ -13,6 +13,11 @@ public IntSliderSetting(String file, String section, String key, int titleId, in
     mDefaultValue = defaultValue;
   }
 
+  public int getDefaultValue()
```
JosJuice, line 16: Since this method signature is identical to the one in `PercentSliderSetting`, I think it would make sense to add this as an abstract method in `SliderSetting`. That way you don't have do add any casts in this PR.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/ui/SettingsAdapter.java
Ebola16: Originally I wanted the neutral button to not close the dialog but now that I think about it it would probably be better if it did.

JosJuice: Oh, I didn't realize that there was that difference in behavior between the current code and what I suggested. Choose whichever way you want, but if you want to keep the current approach, please insert a comment about why you're doing it this way.

Ebola16: I kept my original method and updated the comment. My way may be more useful for AlertDialogs with multiple sliders. Although none of the settings currently use it, I plan on implementing something similar for our in-game AlertDialogs with multiple sliders so I may as well make this consistent.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/ui/SettingsAdapter.java
```diff
@@ -232,6 +232,11 @@ public void onSliderClick(SliderSetting item, int position)
     builder.setTitle(item.getNameId());
     builder.setView(view);
     builder.setPositiveButton(R.string.ok, this);
+    builder.setNeutralButton(R.string.default_value, (dialogInterface, i) ->
+    {
+      // Prevent button from closing the dialog to allow ease of manipulating multiple sliders.
+      // Overridden below.
```
JosJuice, line 238: The connection between the two sentences isn't especially clear. What do you think about: "Overridden below to prevent button from closing the dialog." (Feel free to re-add something about ease of manipulating multiple sliders to what I wrote.)

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/model/view/FloatSliderSetting.java
```diff
@@ -13,6 +13,11 @@ public FloatSliderSetting(String file, String section, String key, int titleId,
     mDefaultValue = defaultValue;
   }
 
+  public int getDefaultValue()
```
JosJuice, line 16: Please add the `@Override` specifier here and in `IntSliderSetting`.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/ui/SettingsAdapter.java
Ebola16: I like it.

JosJuice (approved):

Ebola16:
This will be unnecessary with #9095. I'll make a separate PR that implements this for in-game sliders.

PR #5169, opened 2017-03-26
SI_DeviceGBA: refactor and make GBASockServer a member rather than parent

ligfx:
Inspired by, and a replacement for, #5147.

PR #9069, opened 2020-09-08
SI_DeviceGBA: Wait for an answer from the GBA

Steelskin:
In the refactore done in #5169, the logic was slightly altered to no
longer wait in case the answer from the GBA is delayed. This broke
some games like Final Fantasy Crystal Chronicles.
This fixes Final Fantasy Crystal Chronicles GBA connection on trunk.

### DIFF
--- a/Source/Core/Core/HW/SI/SI_DeviceGBA.cpp
+++ b/Source/Core/Core/HW/SI/SI_DeviceGBA.cpp
@@ -272,7 +272,7 @@ int GBASockServer::Receive(u8* si_buffer)
   if (recv_stat == sf::Socket::Disconnected)
   {
     Disconnect();
-    return 0;
+    return -1;
   }
 
   if (recv_stat == sf::Socket::NotReady || num_received == 0)
@@ -330,13 +330,15 @@ int CSIDevice_GBA::RunBuffer(u8* buffer, int request_length)
   case NextAction::ReceiveResponse:
   {
     int num_data_received = m_sock_server.Receive(buffer);
-    m_next_action = NextAction::SendCommand;
-    if (num_data_received == 0)
+    if (num_data_received < 0)
     {
+      m_next_action = NextAction::SendCommand;
       u32 reply = Common::swap32(SI_ERROR_NO_RESPONSE);
       std::memcpy(buffer, &reply, sizeof(reply));
       return sizeof(reply);
     }
+    if (num_data_received == 0)
+      return 0;
 #ifdef _DEBUG
     const Common::Log::LOG_LEVELS log_level =
         (m_last_cmd == CMD_STATUS || m_last_cmd == CMD_RESET) ? Common::Log::LERROR :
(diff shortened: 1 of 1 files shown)


AdmiralCurtiss:
@dolphin-emu-bot rebuild

AdmiralCurtiss:
I assume this commit introduced the problem? becb1a744b762c5ee166b46ab8904e3236c7127b

Steelskin:
Yes, I believe it came from that specific commit. I went by Dolphin nightly builds to bisect the issue so I had it down to a single PR.

JMC47:
Requesting Review from endrift even though this is kind of a hotfix.

Source/Core/Core/HW/SI/SI_DeviceGBA.cpp
```diff
@@ -330,13 +330,15 @@ int CSIDevice_GBA::RunBuffer(u8* buffer, int request_length)
...
-    m_next_action = NextAction::SendCommand;
-    if (num_data_received == 0)
+    if (num_data_received < 0)
     {
+      m_next_action = NextAction::SendCommand;
       u32 reply = Common::swap32(SI_ERROR_NO_RESPONSE);
       std::memcpy(buffer, &reply, sizeof(reply));
       return sizeof(reply);
     }
+    if (num_data_received == 0)
```
endrift, line 340: Nit: If this `if` is done first, since it's a return, you don't need to move (or duplicate) the `m_next_action` assignment; it can be done between the first `if`'s return and subsequent checks.

endrift (commented):
I'm a little iffy on how this corresponds to hardware behavior; it's basically telling the SI bus to "try again later" for if it gets a reply, which is not how JOY operates. That said, if it improves link reliability I'm not sure it's necessarily a "bad" change. It will add more timing non-determinism though.

JMC47:
I'm okay with hacky fixes until (eventually) this is completely rewritten as most of this will probably get thrown out anyway?

endrift:
I should check to see if this resolves the issues I was having with Orca. Would be nice if this made it actually feasible to merge the ancient branch.

Source/Core/Core/HW/SI/SI_DeviceGBA.cpp
Steelskin: Thank you, I changed it and added a comment.

Steelskin:
> I'm a little iffy on how this corresponds to hardware behavior; it's basically telling the SI bus to "try again later" for if it gets a reply, which is not how JOY operates. That said, if it improves link reliability I'm not sure it's necessarily a "bad" change. It will add more timing non-determinism though.

I understand this is not hardware behavior but I think there are other issues with the GC/GBA connection anyway. For FFCC, we only seem to hit this case during the game's initialization routine for the GBA. There seems to be some other issues when the bus is under heavy load, like when the game displays the radar on the GBA, see this issue: https://bugs.dolphin-emu.org/issues/9528

I do not know enough about both consoles architecture to figure out what a proper fix would be, but I would be happy to dig more into the issue if I knew where and how to look.

AdmiralCurtiss:
@dolphin-emu-bot rebuild

shuffle2:
@dolphin-emu-bot rebuild

JMC47:
I've tested this with modern VBA-M and such, and I'm wondering the validity of it.  I was unable to determine any change in behavior.  There is also questions to whether this is actually correct behavior.

Steelskin:
I tried recent builds with mGBA and can confirm that this is indeed working. So it sounds like the issue is with VBA-M on Windows at least. I will see if I can get to the bottom of this issue but I think we can safely exclude any issue on Dolphin side. I will close this PR as a result.

JMC47:
Yeah, endrift said that this was a hack.

PR #9070, opened 2020-09-09
Add support for built-in input profiles

Techjar:
They exist in the sys directory with the same structure as the config directory. Also includes a profile for controller interface Wiimote with gyro-emulated IR.


iwubcode:
Great!  I know I've had a lot of requests for this.

Should the `IMU*` values be surrounded by tick marks?

Techjar:
The UI generated this config, so I trust that it's valid and correct.

iwubcode (approved):

BhaaLseN (approved):
Code seems good (untested)

Just one question: can users take a built-in profile, change something and then save it as their own modified profile? Disabling save/delete is one thing, but if this prevents saving modifications it might be something to look into.

Techjar:
Yes, changing the name to something other than the built-in one (it's an editable combo box) will re-enable the buttons.

MayImilae:
Built-in is frankly, not great. But I can't really figure out any alternatives that do that while being short. So, I'd prefer "Dolphin Profile" instead, but that's longer. So, hmm. Well, if that fits, I'd prefer it. But built-in is fine, imo.

Also only sort of related but I'm very glad you only included a profile for Wii Remote with Motion Pointing. That's a genuinely very useful addon and it doesn't have any of the mapping subjectivity that comes with eeevvverythiing else. Nice! If you (or anyone else) ever want to expand this to default profiles for traditional gamepads, it's going to get _messy._

Techjar:
There's no issue with length, since the combo box size is going to depend on the window size anyways. I'm just not sure how I feel about "Dolphin Profile". Seems a bit redundant doesn't it?

I think we could include default profiles for some common and non-messy stuff. XInput is a fixed set of buttons, so that's a good candidate. We can worry about that in a different PR though.

Techjar:
I have changed the term to "Stock" in the UI, and made it translatable.

MayImilae:
Alright, stock is fine. Another option would be "Included" or something, which is another route if you prefer that.

-----

Regarding including more default mapping (again, this is not all that relevant to this PR, it's for future stuff I guess), I wasn't even really talking about the Xinput or Dinput issue. Even if we limit ourselves to only supporting Xinput controllers, including default mapping for gamecube and wii games is still very complex. From what I've seen, the GameCube alone has 4 different common mappings (though one of those is kind of bad), varying by personal choice and the necessities of certain titles. And the Wii... that's just a nightmare. While there are some commonalities in later Wii titles, for the most part 360 to Wii profiles have to be made unique for each and every game. Nightmare. IMO, I think it is very viable for us to ship the three major profiles for 360 to GameCube controller mapping, and perhaps classic controller as well since that one is actually easy. Just, please, for the love of all that is good and wholesome in the world, don't try to make default wii remote mappings!

Techjar:
"Included" is what I thought of, but it sounded weird, so I dug through my thesaurus and came up with "Stock".

Yeah, like I said I'm not touching default wiimote mappings with a ten-foot pole. Though I wasn't aware there was more than one way people map 360 to GC. It has basically all the same buttons, save for a left bumper and clicky sticks, along with the face button layout being different. I can't think of more than one sensible way to bind it.

JosJuice:
I personally map the face buttons A-B B-A X-Y Y-X to match the layout of most of Nintendo's consoles, since I'm very used to that layout. Some others map them A-A B-B X-X Y-Y because they're used to the Xbox layout or just want the button labels to match, and some others map them A-A B-X X-B Y-Y because this supposedly is the most spatially similar to the GameCube...

Data/Sys/Profiles/Wiimote/Wii Remote With MotionPlus Pointing.ini
```diff
@@ -0,0 +1,86 @@
...
+Classic/Triggers/Dead Zone = 5.0000000000000000
+Classic/Triggers/L = `Classic L`
+Classic/Triggers/R = `Classic R`
+Classic/Triggers/L-Analog = `Classic L-Analog`
+Classic/Triggers/R-Analog = `Classic R-Analog`
+Classic/D-Pad/Up = `Classic Up`
+Classic/D-Pad/Down = `Classic Down`
+Classic/D-Pad/Left = `Classic Left`
+Classic/D-Pad/Right = `Classic Right`
+Guitar/Stick/Center = 0.00 0.00
```
jordan-woyak, line 76: These "Center" entries could be removed.
There's some above too.

Data/Sys/Profiles/Wiimote/Wii Remote With MotionPlus Pointing.ini
```diff
@@ -0,0 +1,86 @@
...
+IMUIR/Total Yaw = 20.000000000000000
+Extension/Attach MotionPlus = `Attached MotionPlus`
+Extension = `Attached Extension`
+Nunchuk/Buttons/C = `Nunchuk C`
+Nunchuk/Buttons/Z = `Nunchuk Z`
+Nunchuk/Stick/Up = `Nunchuk Y+`
+Nunchuk/Stick/Down = `Nunchuk Y-`
+Nunchuk/Stick/Left = `Nunchuk X-`
+Nunchuk/Stick/Right = `Nunchuk X+`
+Nunchuk/Stick/Calibration = 99.03 96.27 97.29 102.30 108.95 101.86 99.22 100.30 105.52 100.70 99.97 101.77 103.68 99.30 95.07 94.67 98.02 93.32 92.44 95.21 97.31 91.13 88.89 90.15 95.15 92.96 92.37 95.43 100.26 96.04 93.30 94.24
```
jordan-woyak, line 36: I think it would be cleaner if you set these three "Calibration" entries to "100 100 100 100 100 100 100 100".
That should produce a perfect octagon to match up with the calibration adjusted values from the Nunchuk/CC.

Data/Sys/Profiles/Wiimote/Wii Remote With MotionPlus Pointing.ini
Techjar: Why?

jordan-woyak: Because "0 0" is the default. And the "Guitar" and such are not relevant in this profile. :P
I thought I fixed the default "Center"s being saved but apparently not.

jordan-woyak:
I'm a `B-X X-B` person but I think when we do include an Xbox controller profile we should just match the letters.
Users that are smart enough to want X to be B should be smart enough to change it.
Users that are not smart enough to change it are probably not smart enough to know X means B.

Techjar:
Yeah, I was thinking about how to even name such profiles and I don't know how to make it not confusing, so I'm also in favor of only including a "matching labels" profile when we do that.

MayImilae:
I would really prefer to include all three. If we only ship one, then it's like we're saying it's the correct one, and I really don't like that. As for the naming, "Direct", "Nintendo" (or SNES if you prefer), and "Spatial" I think would work well.

mbc07:
Perhaps "GameCube Controller with Xbox button layout", "GameCube Controller with Nintendo button layout" and "GameCube Controller with spatial button layout"? That got a bit long but I think it works, however it still doesn't make clear it's intended only for XInput devices, and I can't think of a good way to inform that...

JMC47:
Is this more or less reviewed at this point?

devinprater:
Spacial layout sounds a bit vague, but I’m so glad this is being worked on. I hope gamepad rumble features are also considered. This would help a lot of blind users get more into Dolphin, as it’s not easy to everyone to configure all this correctly.
Devin Prater
d.prater@me.com
Https://devinprater.github.io

jordan-woyak (approved):

PR #9073, opened 2020-09-10
Adding AmdPowerXpressRequestHighPerformance

darkain:
Adding AmdPowerXpressRequestHighPerformance

This will allow AMD drivers to detect the request to use the dGPU instead of the iGPU on compatible hybrid graphics systems.

Reference: https://community.amd.com/thread/169965


JosJuice:
@dolphin-emu-bot rebuild

JMC47:
Does anyone have an afflicted machine they can test it on?  It'd be nice to have verification it helps before merging.

darkain:
It worked on my Dell G5-SE which has a Ryzen 4800u and Radeon 5600m. Dolphin would only use the iGPU (which was extremely slow). Even setting the application to "high performance" mode in Windows. Adding this small patch allowed the AMD GPU drivers to properly detect the request to use the dedicated GPU.

(side note: the laptop is being returned to Dell due to other issues with it, so I won't be able to do any further testing beyond what I've already done)

JosJuice (approved):

PR #9076, opened 2020-09-12
DolphinQt: properly remove slot connection for ConfigChange when object is destructed

iwubcode:
When destroying a Dolphin "graphics" control, we were still connected to the settings ConfigChange event.  That was causing a segfault.

This is not required by any current Dolphin code but is used in #8823


iwubcode:
Went ahead and fixed some other controls that had the same issue

jordan-woyak (approved):
LGTM

lioncash (approved):

PR #9078, opened 2020-09-13
Android: Get game metadata from core

JosJuice:
Trying to get it from a GameFile before emulation starts is unreliable.

### DIFF
--- a/Source/Android/jni/MainAndroid.cpp
+++ b/Source/Android/jni/MainAndroid.cpp
@@ -80,6 +80,7 @@ std::mutex s_host_identity_lock;
 Common::Event s_update_main_frame_event;
 Common::Event s_emulation_end_event;
 bool s_have_wm_user_stop = false;
+bool s_game_metadata_is_valid = false;
 }  // Anonymous namespace
 
 void UpdatePointer()
@@ -149,6 +150,10 @@ void Host_YieldToUI()
 
 void Host_TitleChanged()
 {
+  s_game_metadata_is_valid = true;
+
+  JNIEnv* env = IDCache::GetEnvForThread();
+  env->CallStaticVoidMethod(IDCache::GetNativeLibraryClass(), IDCache::GetOnTitleChanged());
 }
 
 static bool MsgAlert(const char* caption, const char* text, bool yes_no, Common::MsgType style)
@@ -608,6 +613,7 @@ static void Run(JNIEnv* env, const std::vector<std::string>& paths,
     }
   }
 
+  s_game_metadata_is_valid = false;
   Core::Shutdown();
   ButtonManager::Shutdown();
   guard.unlock();
@@ -750,6 +756,31 @@ JNIEXPORT void JNICALL Java_org_dolphinemu_dolphinemu_NativeLibrary_SetObscuredP
     JNIEnv* env, jobject obj, jint height)
 {
   OSD::SetObscuredPixelsTop(height);
+}
+
+JNIEXPORT jboolean JNICALL
+Java_org_dolphinemu_dolphinemu_NativeLibrary_IsGameMetadataValid(JNIEnv* env, jobject obj)
+{
+  return s_game_metadata_is_valid;
+}
+
+JNIEXPORT jboolean JNICALL
+Java_org_dolphinemu_dolphinemu_NativeLibrary_IsEmulatingWiiUnchecked(JNIEnv* env, jobject obj)
+{
+  return SConfig::GetInstance().bWii;
+}
+
+JNIEXPORT jstring JNICALL
+Java_org_dolphinemu_dolphinemu_NativeLibrary_GetCurrentGameIDUnchecked(JNIEnv* env, jobject obj)
+{
+  return ToJString(env, SConfig::GetInstance().GetGameID());
+}
+
+JNIEXPORT jstring JNICALL
+Java_org_dolphinemu_dolphinemu_NativeLibrary_GetCurrentTitleDescriptionUnchecked(JNIEnv* env,
+                                                                                 jobject obj)
+{
+  return ToJString(env, SConfig::GetInstance().GetTitleDescription());
 }
 
 #ifdef __cplusplus
--- a/Source/Android/jni/MainAndroid.cpp
+++ b/Source/Android/jni/MainAndroid.cpp
@@ -780,7 +780,12 @@ JNIEXPORT jstring JNICALL
 Java_org_dolphinemu_dolphinemu_NativeLibrary_GetCurrentTitleDescriptionUnchecked(JNIEnv* env,
                                                                                  jobject obj)
 {
-  return ToJString(env, SConfig::GetInstance().GetTitleDescription());
+  // Prefer showing just the name. If no name is available, show just the game ID.
+  std::string description = SConfig::GetInstance().GetTitleName();
+  if (description.empty())
+    description = SConfig::GetInstance().GetTitleDescription();
+
+  return ToJString(env, description);
 }
 
 #ifdef __cplusplus
--- a/Source/Core/Core/Core.cpp
+++ b/Source/Core/Core/Core.cpp
@@ -85,6 +85,10 @@
 #include "VideoCommon/VideoBackendBase.h"
 #include "VideoCommon/VideoConfig.h"
 
+#ifdef ANDROID
+#include "jni/AndroidCommon/IDCache.h"
+#endif
+
 namespace Core
 {
 static bool s_wants_determinism;
@@ -334,6 +338,12 @@ static void CpuThread(const std::optional<std::string>& savestate_path, bool del
 
   // This needs to be delayed until after the video backend is ready.
   DolphinAnalytics::Instance().ReportGameStart();
+
+#ifdef ANDROID
+  // For some reason, calling the JNI function AttachCurrentThread from the CPU thread after a
+  // certain point causes a crash if fastmem is enabled. Let's call it early to avoid that problem.
+  static_cast<void>(IDCache::GetEnvForThread());
+#endif
 
   if (_CoreParameter.bFastmem)
     EMM::InstallExceptionHandler();  // Let's run under memory watch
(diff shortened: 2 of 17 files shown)


Ebola16 (commented):
This will display the gameId in the emulation menu's game title. What do you think about removing the gameID there to maximize space for menu options?

JosJuice:
I guess that makes sense. But what do we do in the case where the title database doesn't recognize the running game? Do we should just the game ID (like how it works right now in this PR) or nothing?

Ebola16:
I think displaying the game ID as a fallback if no title is found is fine.

Ebola16 (changes requested):
Also, GeckoOS is giving me this exception on startup:
2020-09-13 09:42:12.622 8486-8486/org.dolphinemu.dolphinemu.debug E/AndroidRuntime: FATAL EXCEPTION: main
    Process: org.dolphinemu.dolphinemu.debug, PID: 8486
    java.lang.IllegalStateException: No game is running
        at org.dolphinemu.dolphinemu.NativeLibrary.CheckGameMetadataValid(NativeLibrary.java:456)
        at org.dolphinemu.dolphinemu.NativeLibrary.IsEmulatingWii(NativeLibrary.java:436)
        at org.dolphinemu.dolphinemu.overlay.InputOverlay.refreshControls(InputOverlay.java:706)
        at org.dolphinemu.dolphinemu.overlay.InputOverlay.initTouchPointer(InputOverlay.java:140)
        at org.dolphinemu.dolphinemu.fragments.EmulationFragment.initInputPointer(EmulationFragment.java:158)
        at org.dolphinemu.dolphinemu.activities.EmulationActivity.initInputPointer(EmulationActivity.java:1141)
        at org.dolphinemu.dolphinemu.-$$Lambda$VaQ0DnlfVw6B_598YlWOOA_cqe0.run(Unknown Source:2)
        at android.os.Handler.handleCallback(Handler.java:873)
        at android.os.Handler.dispatchMessage(Handler.java:99)
        at android.os.Looper.loop(Looper.java:214)
        at android.app.ActivityThread.main(ActivityThread.java:7073)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:494)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:965)

JosJuice:
Both problems should now be fixed. (Regarding Gecko OS, I hadn't taken into account that `Host_TitleChanged` doesn't get called on boot for DOL/ELF files.)

Ebola16:
Even with the most recent push, I'm getting a (unhelpful) `DeadObjectException` when loading GeckoOS. I double checked by launching Master and it works fine so the problem isn't with my setup.

Ebola16:
Nevermind, it's the combination of https://bugs.dolphin-emu.org/issues/11113 and Android Dolphin resetting my default controls upon restart.

JosJuice:
I don't really get how that could cause a `DeadObjectException`... But I'll ignore it then.

Ebola16:
- GeckoOS now launches properly but but I'm getting the same `DeadObjectException` when choosing "Launch Game", even with accounting for https://bugs.dolphin-emu.org/issues/11113.
- When the point where noting is displayed for the in-game emulation menu game title, can we instead display the game filename? This is much more useful info.

JosJuice:
I tried turning off Wii Remotes and launching Xenoblade Chronicles through Gecko OS (saying yes to when Gecko OS asked me to install IOS 53), and it worked perfectly for me. Do you have any details on what's triggering the exception? I would assume it has something to do with the call to Host_TitleChanged, but I'm not sure.

I will try to create a separate PR for displaying a name for DOL/ELF files. (I don't believe it's possible to reach the case where nothing, not even a game ID, is displayed when running a disc or WAD.)

JosJuice:
> It can let users know if they're currently using a mod launcher if they can't remember.

Perhaps I should mention this explicitly: With this PR, when you launch a Wii game through something like Gecko OS, the title will update to match the game you launched. This is one of the advantages of getting the title from the core.

shuffle2:
@dolphin-emu-bot rebuild

Source/Android/jni/MainAndroid.cpp
```diff
@@ -144,6 +145,10 @@ void Host_YieldToUI()
 
 void Host_TitleChanged()
 {
+  s_game_metadata_is_valid = true;
+
+  JNIEnv* env = IDCache::GetEnvForThread();
```
Ebola16, line 150: This line seems to be the cause of Gecko OS crashing. Removing it and the associated line below allows Gecko OS to load mods seemingly without problem. Only removing the associated line below didn't help.

Ebola16 (changes requested):
> Perhaps I should mention this explicitly: With this PR, when you launch a game through something like Gecko OS, the title will update to match the game you launched. This is one of the advantages of getting the title from the core.

Now that I've gotten to the point where mods were loaded in this PR, I was able to notice that change and it makes sense.

Source/Android/jni/MainAndroid.cpp
JosJuice: Huh. Guess there's a reason why `Host_RequestRenderWindowSize` is using a temporary thread, then... I'm still not quite sure what's wrong, but this is a good hint.

JosJuice: I've made this use the same workaround as `Host_RequestRenderWindowSize`. Please test it.

Ebola16 (approved):
That fixed the problem. I don't have much C++ review experience but as far as I can tell this PR LGTM.

JosJuice:
I've managed to reproduce the crash now (with the last commit reverted). The reason why I couldn't reproduce it before was because it only happens with fastmem on and I had accidentally left fastmem off (since I need it off when using the debugger). Unfortunately, this means that I can't use the debugger to figure out what's going on, and I would really like to know what's going on before this gets merged. I'm also not getting a `DeadObjectException` – my old phone says SIGSEGV and my new phone says nothing at all. I'm assuming the `DeadObjectException` is actually just a symptom of the underlying native crash.

Source/Android/jni/MainAndroid.cpp
```diff
@@ -152,8 +152,10 @@ void Host_TitleChanged()
 {
   s_game_metadata_is_valid = true;
 
-  JNIEnv* env = IDCache::GetEnvForThread();
-  env->CallStaticVoidMethod(IDCache::GetNativeLibraryClass(), IDCache::GetOnTitleChanged());
+  std::thread([] {
```
leoetlino, line 155: This should have a comment explaining why this is required.

JosJuice: The problem is that I have no idea why this is required. It's the one thing keeping this PR from being mergeable in my opinion.

sepalani: I had a similar issue and know where it crashes. The culprit is JVM's `AttachCurrentThread` in `GetEnvForThread`.

In IDCache.cpp:
```c++
if (status == JNI_EDETACHED)
  s_java_vm->AttachCurrentThread(&env, nullptr);
```

The things is, for me at least, it raises randomly many different type of exceptions/errors such as segfault, libc/stack corruption, memory access errors, etc. I tried to find any relevant pieces of documentation regarding this behaviour but couldn't find any. I also don't know (yet) how to debug native code so I couldn't go deeper.

I suspect there is something fishy in that function and that depending on how the thread was created it might produce an undefined behaviour.

JosJuice: Hmm, so `AttachCurrentThread` fails if we call it from the CPU thread, but only when fastmem is enabled... Maybe we could make sure to attach the CPU thread before enabling fastmem? I'll give that a try.

JosJuice: That seems to do the trick. I'm much happier with this than the earlier workaround.

My new commit calls `AttachCurrentThread` before `EMM::InstallExceptionHandler`, but calling it immediately afterwards also seems to work. Perhaps the problem only occurs after the first fastmem fault rather than after the exception handler is installed?

JosJuice:
This should be ready for merge now.

Source/Core/Core/Core.cpp
```diff
@@ -335,6 +339,12 @@ static void CpuThread(const std::optional<std::string>& savestate_path, bool del
   // This needs to be delayed until after the video backend is ready.
   DolphinAnalytics::Instance().ReportGameStart();
 
+#ifdef ANDROID
+  // For some reason, calling the JNI function AttachCurrentThread from the CPU thread after a
+  // certain point causes a crash if fastmem is enabled. Let's call it early to avoid that problem.
+  IDCache::GetEnvForThread();
```
leoetlino, line 345: What do you think of adding a function called AttachCurrentThread that just calls GetEnvForThread, but makes it the purpose/effect of this call more explicit?

JosJuice: It seems a bit odd to add a function that just calls another function with no extra processing, and I don't think creating a function named `AttachCurrentThread` is a great idea since there already is a JNI function called `AttachCurrentThread` (which in fact is freestanding if you use the C API). Not that it'll lead to compile errors, but it might be confusing for someone who isn't familiar with AttachCurrentThread and searches for the name in order to look it up. Given that it would only be used for this one workaround, I don't think it makes sense.

leoetlino: Hmm, it just looks a bit weird to me to call a "get [something]" function and then proceed to ignore the return value. I feel like the comment expects the reader to know that GetEnvForThread internally calls AttachCurrentThread, but someone who isn't extremely familiar with the JNI code would expect to see a call to "it" (AttachCurrentThread), not GetEnvForThread.

Though now that I know the JNI has a free function called `AttachCurrentThread`, I agree that it wouldn't be a good idea to use that name again. Maybe just do a cast to make it abundantly clear that GetEnvForThread is being called for its side effect (`static_cast<void>(IDCache::GetEnvForThread());`)?

JosJuice: Cast added.

leoetlino (approved):

PR #9079, opened 2020-09-13
Android: Check for granted permission when returning to MainActivity

JosJuice:


Ebola16 (approved):
Yeah, I like this approach better.

lioncash (approved):

PR #9080, opened 2020-09-13
MemoryWidget: Add 'Offset' and General Improvements

SirMangler:
Summary of changes:
- Added a new field for Search Address Offset
- Fixed a hard crash when trying set a value while the core is not running.
- Brushed up the UI and spacing

The primary change here is adding an `Offset` field next to `Search Address`. All this does is add the offset to the supplied search address. While this sounds mediocre on paper, this is incredibly convenient for hack and cheat developers (and especially reverse engineers) who frequently work with base addresses and classes. I have been working on [PrimeHack](https://github.com/shiiion/dolphin) for nearly a year, and I constantly need to work with addresses that are offset from dynamically allocated classes in memory. Adding this simple `Offset` field streamlines traversing through large classes in memory rather than the otherwise clunky process.

The `Offset` field is resizeable and can be totally collapsed if wanted, but it really doesn't get in the way of the regular `Search Address`. The `Find Next`/`Find Previous` buttons work as expected, they simply clear the offset field if it finds a result.

**Before/After**
image

### DIFF
--- a/Source/Core/DolphinQt/Debugger/MemoryWidget.cpp
+++ b/Source/Core/DolphinQt/Debugger/MemoryWidget.cpp
@@ -23,6 +23,7 @@
 #include "Common/File.h"
 #include "Common/FileUtil.h"
 #include "Core/ConfigManager.h"
+#include "Core/Core.h"
 #include "Core/HW/AddressSpace.h"
 #include "DolphinQt/Debugger/MemoryViewWidget.h"
 #include "DolphinQt/Host.h"
@@ -85,12 +86,22 @@ void MemoryWidget::CreateWidgets()
   //// Sidebar
 
   // Search
+  auto* m_address_splitter = new QSplitter(Qt::Horizontal);
+
   m_search_address = new QLineEdit;
+  m_search_offset = new QLineEdit;
   m_data_edit = new QLineEdit;
   m_set_value = new QPushButton(tr("Set &Value"));
 
   m_search_address->setPlaceholderText(tr("Search Address"));
+  m_search_offset->setPlaceholderText(tr("Offset"));
   m_data_edit->setPlaceholderText(tr("Value"));
+
+  m_address_splitter->addWidget(m_search_address);
+  m_address_splitter->addWidget(m_search_offset);
+  m_address_splitter->setHandleWidth(1);
+  m_address_splitter->setCollapsible(0, false);
+  m_address_splitter->setStretchFactor(1, 2);
 
   // Dump
   m_dump_mram = new QPushButton(tr("Dump &MRAM"));
@@ -183,16 +194,23 @@ void MemoryWidget::CreateWidgets()
 
   sidebar->setLayout(sidebar_layout);
 
-  sidebar_layout->addWidget(m_search_address);
+  sidebar_layout->addWidget(m_address_splitter);
   sidebar_layout->addWidget(m_data_edit);
-  sidebar_layout->addWidget(m_find_ascii);
-  sidebar_layout->addWidget(m_find_hex);
+
+  auto* types_layout = new QHBoxLayout;
+  types_layout->addWidget(m_find_ascii);
+  types_layout->addItem(new QSpacerItem(20, 1));
+  types_layout->addWidget(m_find_hex);
+  types_layout->setAlignment(Qt::AlignCenter);
+  sidebar_layout->addLayout(types_layout);
+
   sidebar_layout->addWidget(m_set_value);
-  sidebar_layout->addItem(new QSpacerItem(1, 32));
+  sidebar_layout->addItem(new QSpacerItem(1, 20));
   sidebar_layout->addWidget(m_dump_mram);
   sidebar_layout->addWidget(m_dump_exram);
   sidebar_layout->addWidget(m_dump_aram);
   sidebar_layout->addWidget(m_dump_fake_vmem);
+  sidebar_layout->addItem(new QSpacerItem(1, 15));
   sidebar_layout->addWidget(search_group);
   sidebar_layout->addWidget(address_space_group);
   sidebar_layout->addWidget(datatype_group);
@@ -222,6 +240,7 @@ void MemoryWidget::CreateWidgets()
 void MemoryWidget::ConnectWidgets()
 {
   connect(m_search_address, &QLineEdit::textChanged, this, &MemoryWidget::OnSearchAddress);
+  connect(m_search_offset, &QLineEdit::textChanged, this, &MemoryWidget::OnSearchAddress);
 
   connect(m_data_edit, &QLineEdit::textChanged, this, &MemoryWidget::ValidateSearchValue);
   connect(m_find_ascii, &QRadioButton::toggled, this, &MemoryWidget::ValidateSearchValue);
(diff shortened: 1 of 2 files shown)


sepalani:
@dolphin-emu-bot rebuild

JosJuice:
@dolphin-emu-bot rebuild

shuffle2:
fwiw ctrl+k,ctrl+d is "format document" in VS (also a named action in vscode / shift+alt+f)

shuffle2:
@dolphin-emu-bot rebuild

AdmiralCurtiss:
Not sure if I fully understand how the offset should be used, but otherwise looks fine to me. I could see extracting the 'calc address from GUI fields' code to a separate function so it's not there three times, maybe? But not a huge deal.

Rumi-Larry:
All groups have a header, except for the dumping buttons. I suggest adding one saying "Memory Dump"

Rumi-Larry:
Also, the "Search" group should be closer to the actual search bar. Like after the "Set Value" button, the "Find Next" and "Find Previous" buttons should be right after, making the "Search" header unnecessary.

SirMangler:
> All groups have a header, except for the dumping buttons. I suggest adding one saying "Memory Dump"

> the "Search" group should be closer to the actual search bar.

Not bad suggestions!

> Not sure if I fully understand how the offset should be used

Say hypothetically you have a class at address `8000000`. At `8000004` you have it's health. `8000008` you have ammo. `800000C` you have some random array. You can just do the `8000000` as the address, and in offset you put in `4` for health, `8` for ammo, `C` for the array. 

Now sure that's easy to do in your head and easily change the search address field. But what if we use more real numbers, you have a function loading the address `805C72AB` into `r4`, and an instruction loading `0x241(r4)` (which means 805C72AB + 0x241 if you didn't know) into r5, and another instruction loading `0x14c(r4)` into r6 (aka 801C72AB  + 0x14c). 

It does an operation, r5 + r6. You want to figure out the answer to that? You have to copy the register `r4` into a calculator, add `0x241` to that, and paste the result into the memory search address field. Then once you note that value, you would have to copy register r4 again (most likely from an open notepad somewhere), paste into your calculator and add `0x14c`. Paste that into the memory search address field and note that answer down.

Or with this new PR? Copy r4 into the memory search address field. Copy `0x241` into the offset and note the value. Copy `0x14c` into the offset and note the second value.

I have to do this process a lot for a variety of different scenarios, this feature while very simple, would be a pretty big time saver. No more needing to store the address somewhere and open a calculator every time I need to calculate an offset without doing it in my head. Especially useful if you're trying to map out a class and have to go between several offsets to see what is doing what.

In the end, you can ignore it if you want, you can even collapse it so its hidden.

JMC47:
Yeah, using offsets is something I see in a lot of other debugging GUIs and I believe it would be a nice feature to have in Dolphin's debugger too.

Source/Core/DolphinQt/Debugger/MemoryWidget.cpp
```diff
@@ -426,14 +445,15 @@ void MemoryWidget::OnSearchAddress()
 {
   bool good;
   u32 addr = m_search_address->text().toUInt(&good, 16);
+  u32 offset = m_search_offset->text().toUInt(nullptr, 16);  // Returns 0 if conversion fails
```
AdmiralCurtiss, line 448: Why is an invalid string in the address field an error but an invalid string in the offset field okay? This could probably be handled better.

Source/Core/DolphinQt/Debugger/MemoryWidget.cpp
```diff
@@ -468,8 +488,12 @@ void MemoryWidget::ValidateSearchValue()
 
 void MemoryWidget::OnSetValue()
 {
+  if (!Core::IsRunning())
+    return;
+
   bool good_address;
   u32 addr = m_search_address->text().toUInt(&good_address, 16);
+  addr += m_search_offset->text().toUInt(nullptr, 16);  // Returns 0 if conversion fails
```
AdmiralCurtiss, line 496: Same here.

AdmiralCurtiss:
So while I'm not 100% happy with some of this, this works fine for what it's intended to do, so I don't really have any objections to just merging it as-is even if those are not addressed. The debugger needs more work anyway.

Source/Core/DolphinQt/Debugger/MemoryWidget.cpp
```diff
@@ -85,13 +86,23 @@ void MemoryWidget::CreateWidgets()
   //// Sidebar
 
   // Search
+  m_address_splitter = new QSplitter(Qt::Horizontal);
```
jordan-woyak, line 89: `m_address_splitter` doesn't need to be a member variable. It's only used in this function. It can be just a local variable.

Source/Core/DolphinQt/Debugger/MemoryWidget.cpp
SirMangler: I omitted it simply because the offset is just a modifier and unable to supply an incorrect value to the address. If the offset input is invalid, the address is not effected. The offset is expected to be blank, so this is not an error state. If the offset is malformed, it simply does not modify the address. The address itself is checked to see if the number is true hexadecimal, but the only result of that error detection is making the address field bold and red.

I can see that being handled better by making an indication that the offset is malformed in the same way that the address is. I intend on making another commit for this debugger, so if this is merged before I make this edit I will include it in the next one (or it's own one with similar edits).

SirMangler: Pushed a change to add this error handling in. Even if it's not fully what you had in mind, it should be better than before.

Source/Core/DolphinQt/Debugger/MemoryWidget.cpp
SirMangler: Good catch!! Fixed.

AdmiralCurtiss:
@dolphin-emu-bot rebuild

Source/Core/DolphinQt/Debugger/MemoryWidget.cpp
AdmiralCurtiss: Yeah this is definitely better.

SirMangler:
Is this okay to be merged? It seems all specific issues and comments have been resolved.

Source/Core/DolphinQt/Debugger/MemoryWidget.cpp
```diff
@@ -468,15 +496,25 @@ void MemoryWidget::ValidateSearchValue()
 
 void MemoryWidget::OnSetValue()
 {
-  bool good_address;
+  if (!Core::IsRunning())
+    return;
+
+  bool good_address, good_off;
```
leoetlino, line 502: ditto

Source/Core/DolphinQt/Debugger/MemoryWidget.cpp
```diff
@@ -468,15 +496,25 @@ void MemoryWidget::ValidateSearchValue()
...
   u32 addr = m_search_address->text().toUInt(&good_address, 16);
+  addr += m_search_offset->text().toUInt(&good_off, 16);  // Returns 0 if conversion fails
```
leoetlino, line 504: The comment should be on its own line.

leoetlino (changes requested):

SirMangler:
@leoetlino With the exception of moving the comment to it's own line ( It's within the character limit but it is fair ) and using `auto`, all of your requested changes aren't necessarily relevant to this PR. 

You requested I do not use `off` as an abbreviation for offset because it is not a common abbreviation. I may be inclined to agree with you, but that is not why I chose that. I was following the naming scheme of the existing code.
```
  QFont addr_font, off_font;
  QPalette addr_palette, off_palette;
```
  
I did not write that second line. 
  
You have asked me to change some lines that aren't mine, adding `const` to some. Again this may be a fair change, but the point of this PR was not to refactor existing code.

PR #9083, opened 2020-09-14
Fix ini_file_section_class type specifier

Ebola16:


JosJuice:
If you try running this (just open the settings), I think you will find that it doesn't work.

Ebola16:
Yes, that's the first time I've encountered an invalid type specifier mismatch.

PR #9084, opened 2020-09-15
Qt/AdvancedPane: Force UTC time for custom RTC entry

Techjar:
Using the local time zone doesn't make any sense, as the RTC does not have a time zone field. It just results in the actual time being offset from what is entered in the settings.


JMC47:
Works for me.

lioncash (approved):

PR #9085, opened 2020-09-15
Qt: Use better method of getting default monospace font

Techjar:
This should give us a nicer font on Windows, while also not severely impacting the existing behavior on Linux.

At least on my Linux system, it actually results in using the monospace font I have configured (which is "Hack").

Before ("Monospace")

After ("Hack")

### DIFF
--- a/Source/Core/DolphinQt/Config/LogWidget.cpp
+++ b/Source/Core/DolphinQt/Config/LogWidget.cpp
@@ -119,8 +119,7 @@ void LogWidget::UpdateFont()
   case 0:  // Default font
     break;
   case 1:  // Monospace font
-    f = QFont(QStringLiteral("Monospace"));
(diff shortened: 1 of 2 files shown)


JosJuice:
We're using the same approach in the log widget. I assume it would make sense to change that as well? https://github.com/dolphin-emu/dolphin/blob/82012c4cd6778c3ebb1bc48c50a76e07667f359c/Source/Core/DolphinQt/Config/LogWidget.cpp#L122-L123

Techjar:
Ah, I wasn't aware it was separate.

JMC47:
Before:

image

After:

image

And the after runs ~10X smoother.

lioncash (approved):

pizuz:
This fixes the [startup delay experienced in macOS](https://bugs.dolphin-emu.org/issues/12035#change-738847) as well.

PR #9088, opened 2020-09-16
Android: Add Wii SYSCONF settings to GUI

JosJuice:
Now that PR #8975 is merged, we can finally add this without hacks.

https://bugs.dolphin-emu.org/issues/11605


Ebola16 (commented):
PAL60 mode is incorrectly reported as overridden in-game.

JosJuice:
It's actually correctly reported as overridden. Dolphin automatically disables the PAL60 setting when you run an NTSC game, since some games crash otherwise. I've amended the last commit so that PAL60 doesn't show up as overridden if you actually had it disabled originally, but it showing up as overridden when it originally was enabled is the logical end result of how Dolphin works.

Ebola16 (approved):
Ah, good to know. I haven't looked into SYSCONF in depth but the changes here look good.

PR #9090, opened 2020-09-16
Add nullptr check in VariableExpression parser

hyazinthh:
Added nullptr checks in `ExpressionParser::VariableExpression`, preventing access violations occurring when switching between input profiles such as the following:
> [Profile]
> Device = XInput/0/Gamepad
> IR/Up = $enable_point = toggle(\`Trigger R\`), \`Right Y+\` & $enable_point
> IR/Down = \`Right Y-\` & $enable_point
> IR/Left = \`Right X-\` & $enable_point
> IR/Right = \`Right X+\` & $enable_point

### DIFF
--- a/Source/Core/InputCommon/ControlReference/ExpressionParser.cpp
+++ b/Source/Core/InputCommon/ControlReference/ExpressionParser.cpp
@@ -358,9 +358,9 @@ class VariableExpression : public Expression
 public:
   VariableExpression(std::string name) : m_name(name) {}
 
-  ControlState GetValue() const override { return *m_value_ptr; }
+  ControlState GetValue() const override { return (m_value_ptr != nullptr) ? *m_value_ptr : 0; }
 
-  void SetValue(ControlState value) override { *m_value_ptr = value; }
+  void SetValue(ControlState value) override { if (m_value_ptr != nullptr) *m_value_ptr = value; }
 
   int CountNumControls() const override { return 1; }
 


jordan-woyak:
If `m_value_ptr` is causing access violations after a profile switch we should fix the real problem.
It sounds like `UpdateReferences` isn't being called.

hyazinthh:
> If `m_value_ptr` is causing access violations after a profile switch we should fix the real problem.
> It sounds like `UpdateReferences` isn't being called.

It seems like it's a race condition as the CPU thread is polling the input state, while the HotkeyScheduler thread loads the config and then calls `UpdateReferences` afterwards. I'm unsure how to effectively prevent this, adding a mutex there may have some performance implications.

In any case, I think having a nullptr check there is a good idea, as the `VariableExpression` may be in an invalid state if the client forgets to call `UpdateReferences`.

AdmiralCurtiss:
If it's a race then the nullpointer check isn't gonna help you, as you could be interrupted between the check and the access.

hyazinthh:
> If it's a race then the nullpointer check isn't gonna help you, as you could be interrupted between the check and the access.

Only if the other thread may reset the pointer to null, which doesn't seem to be the case here.

jordan-woyak:
Sorry, I think this is just hiding the real problem of our racey input code so I'm going to close this PR for now.
We should instead make sure `UpdateReferences` is called safely.
Hopefully soon input paths can be made more thread-safe so this issue you are having doesn't occur.

PR #9091, opened 2020-09-17
Add some files to pch

shuffle2:
Drops compile time by a couple seconds

### DIFF
--- a/Source/PCH/pch.h
+++ b/Source/PCH/pch.h
@@ -23,11 +23,16 @@
 #include <execinfo.h>
 #endif
 #include <fcntl.h>
+#include <filesystem>
 #include <float.h>
+#include <fmt/format.h>
 #include <fstream>
 #include <functional>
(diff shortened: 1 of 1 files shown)


Source/PCH/pch.h
```diff
@@ -54,6 +60,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <string_view>
```
leoetlino, line 63: Would it make sure to include `<string>` as well?

shuffle2: I was just going by profiler metrics so...probably not ? ( I guess it was not reported in the top hits/time)

shuffle2: You (or anyone) can easily profile the msvc build, I recommend starting here: https://devblogs.microsoft.com/cppblog/author/kevca/

PR #9096, opened 2020-09-19
ControllerEmu: Improve and simplify UpdateCalibrationData.

jordan-woyak:
Previously, the closest (by angle) calibration point was set to the distance of the provided input.
This produced slightly wrong calibration that would even draw outside of indicator square.
Clamping was done which hid the problem.

Now calibration is calculated from intersection with a line segment of the last two input values.
This is cleaner and more accurate.
Also the calibration shape is no longer restricted to a convex polygon.

### DIFF
--- a/Source/Core/InputCommon/ControllerEmu/StickGate.cpp
+++ b/Source/Core/InputCommon/ControllerEmu/StickGate.cpp
@@ -25,18 +25,24 @@ constexpr auto CALIBRATION_CONFIG_SCALE = 100;
 constexpr auto CENTER_CONFIG_NAME = "Center";
 constexpr auto CENTER_CONFIG_SCALE = 100;
 
-// Calculate distance to intersection of a ray with a line defined by two points.
-double GetRayLineIntersection(Common::DVec2 ray, Common::DVec2 point1, Common::DVec2 point2)
+// Calculate distance to intersection of a ray with a line segment defined by two points.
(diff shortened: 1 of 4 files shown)


Source/Core/InputCommon/ControllerEmu/StickGate.cpp
```diff
@@ -25,18 +25,24 @@ constexpr auto CALIBRATION_CONFIG_SCALE = 100;
 constexpr auto CENTER_CONFIG_NAME = "Center";
 constexpr auto CENTER_CONFIG_SCALE = 100;
 
-// Calculate distance to intersection of a ray with a line defined by two points.
-double GetRayLineIntersection(Common::DVec2 ray, Common::DVec2 point1, Common::DVec2 point2)
+// Calculate distance to intersection of a ray with a line segment defined by two points.
+std::optional<double> GetRayLineIntersection(Common::DVec2 ray, Common::DVec2 point1,
```
iwubcode, line 29: This might be useful to move to Math?

iwubcode (approved):

Source/Core/InputCommon/ControllerEmu/StickGate.cpp
jordan-woyak: I'd say move it to `Math` when it's needed in more than one place. Adding it to the math headers now would cause recompilation of basically everything for no benefit.

PR #9098, opened 2020-09-20
WiimoteReal: Only duplicate data reports when speaker data is enabled.

jordan-woyak:
Currently, data reports of real remotes are repeated to trigger sending of speaker data.

This causes some games to fail at detecting motion gestures.
https://bugs.dolphin-emu.org/issues/10111
https://bugs.dolphin-emu.org/issues/12030

With this change if speaker data is disabled so is repeating of data reports.
This fixes gestures in these games so long as speaker data is disabled.

Another potential fix might be to interpolate accelerometer data in repeated reports but that is getting pretty hacky.
I think this is cleaner.

### DIFF
--- a/Source/Core/Core/HW/WiimoteReal/WiimoteReal.cpp
+++ b/Source/Core/Core/HW/WiimoteReal/WiimoteReal.cpp
@@ -173,12 +173,6 @@ void Wiimote::WriteReport(Report rpt)
       m_speaker_mute = (rpt[2] & 0x4) != 0;
       break;
 
-    case OutputReportID::ReportMode:
-      // Force non-continuous reporting for less BT traffic.
-      // We duplicate reports to maintain 200hz anyways.
-      rpt[2] &= ~0x4;
-      break;
-
     default:
       break;
     }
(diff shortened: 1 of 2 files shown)


JMC47:
Everything in this makes sense to me.  I quickly tested Sonic and the Secret Rings and it seemed to be fine with Speaker Data off.

Question?  Does Bluetooth Passthrough get into sniff mode though?  If it does, why is audio data so iffy still?

jordan-woyak:
> 
> 
> Question? Does Bluetooth Passthrough get into sniff mode though? If it does, why is audio data so iffy still?

Yes, "sniff" mode _is_ set in Bluetooth Passthrough where Wii remote audio works flawlessly here.

leoetlino (approved):

cobalt2727:
I've got a report of this causing severe input delay for the pointer and nothing else, oddly enough.
https://discord.com/channels/521709831132807179/521711075721478145/787156878944894986
(you have to join the unofficial Dolphin Discord or the link won't work)

PR #9099, opened 2020-09-20
IP/Top: Add SO_LINGER optname

sepalani:
This PRs adds the SO_LINGER optname and silents the following message `SO_SETSOCKOPT: unknown optname 128`.

### DIFF
--- a/Source/Core/Core/IOS/Network/IP/Top.cpp
+++ b/Source/Core/Core/IOS/Network/IP/Top.cpp
@@ -143,6 +143,8 @@ static s32 MapWiiSockOptNameToNative(u32 optname)
   {
   case 0x4:
     return SO_REUSEADDR;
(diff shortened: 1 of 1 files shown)


BhaaLseN (approved):
According to `sys/socket.h`, `SO_LINGER` is `0x0080` - so this checks out. Untested tho.

I _did_ notice `SO_ERROR` is `0x1007` and **not** `0x1009`, but thats probably out of scope here.

sepalani:
@BhaaLseN It wouldn't surprised me that they didn't keep the original value for `SO_ERROR` or that for some system it's defined as `0x1009` rather than `0x1007`. I'll write an hardware test to confirm that and see what are the other missing values in another PR.

lioncash (approved):

PR #9101, opened 2020-09-20
IP/Top: Fix fallback IP address

sepalani:
I don't know if that will change anything but this PR should fix the IP fallback values.

### DIFF
--- a/Source/Core/Core/IOS/Network/IP/Top.cpp
+++ b/Source/Core/Core/IOS/Network/IP/Top.cpp
@@ -79,11 +79,6 @@ void NetIPTop::DoState(PointerWrap& p)
   WiiSockMan::GetInstance().DoState(p);
 }
 
-static constexpr u32 inet_addr(u8 a, u8 b, u8 c, u8 d)
-{
-  return (static_cast<u32>(a) << 24) | (static_cast<u32>(b) << 16) | (static_cast<u32>(c) << 8) | d;
-}
-
 static int inet_pton(const char* src, unsigned char* dst)
 {
   int saw_digit = 0;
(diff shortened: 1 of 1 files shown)


BhaaLseN (approved):
I guess we can take that as-is, at least until we're also considering big endian systems /shrug
(untested though)

Source/Core/Core/IOS/Network/IP/Top.cpp
```diff
@@ -258,8 +254,8 @@ static std::optional<DefaultInterface> GetSystemDefaultInterface()
 
 static DefaultInterface GetSystemDefaultInterfaceOrFallback()
 {
-  static constexpr DefaultInterface FALLBACK_VALUES{
-      inet_addr(10, 0, 1, 30), inet_addr(255, 255, 255, 0), inet_addr(10, 0, 255, 255)};
+  // Fallback values in little endian: {10.0.1.30, 255.255.255.0, 10.0.255.255}
+  static constexpr DefaultInterface FALLBACK_VALUES{0x0e01000a, 0x00ffffff, 0xffff000a};
```
leoetlino, line 258: Couldn't you just change our `inet_addr` to use a different endianness (or just add a call to Common::Swap32)? This is less readable.

leoetlino:
This needs a rebase

Source/Core/Core/IOS/Network/IP/Top.cpp
sepalani: I took a similar "swap" approach using the native `inet_addr` as it returns the value as expected by the DefaultInterface structure.

JosJuice:
@dolphin-emu-bot rebuild

PR #9102, opened 2020-09-20
Fix EFB access

Miksel12:
This reverts the readback changes of https://github.com/dolphin-emu/dolphin/pull/8827.
This fixes https://bugs.dolphin-emu.org/issues/12265 and unfixes https://bugs.dolphin-emu.org/issues/12040
I think fixing the first issue is more important as multisampling is used a lot. I think save states on broken Adreno drivers are less of a priority.

### DIFF
--- a/Source/Core/VideoCommon/FramebufferManager.cpp
+++ b/Source/Core/VideoCommon/FramebufferManager.cpp
@@ -255,14 +255,10 @@ AbstractTexture* FramebufferManager::ResolveEFBColorTexture(const MathUtil::Rect
   return m_efb_resolve_color_texture.get();
 }
 
-AbstractTexture* FramebufferManager::ResolveEFBDepthTexture(const MathUtil::Rectangle<int>& region,
-                                                            bool force_r32f)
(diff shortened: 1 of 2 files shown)


JMC47:
@dolphin-emu-bot rebuild

JMC47:
This doesn't break either case on Android

Vulkan Savestates still work
OpenGLES EFB Access to CPU still works.

Miksel12:
Did you test one of the games mentioned in the issue report? https://bugs.dolphin-emu.org/issues/12040

JMC47:
I tested Wind Waker, yes.

Source/Core/VideoCommon/FramebufferManager.cpp
```diff
@@ -258,8 +258,7 @@ AbstractTexture* FramebufferManager::ResolveEFBColorTexture(const MathUtil::Rect
 AbstractTexture* FramebufferManager::ResolveEFBDepthTexture(const MathUtil::Rectangle<int>& region,
                                                             bool force_r32f)
```
BhaaLseN, line 259: The `force_r32f` parameter is now unused; either it can go away as well or we're still missing something

stenzek:
See #9109.

PR #9103, opened 2020-09-20
Add caching to Config::Info

JosJuice:
The goal of this change is to make `Config::Get(const Info<T>&)` super fast so that we can use it in hot paths.

This probably isn't the only reasonable way to solve the performance problem. I tried to pick a way that would require little changes to the existing config code, but unfortunately this way requires us to do some slightly ugly stuff to work around the unpredictable initialization order of static objects.

### DIFF
--- a/Source/Core/Common/Config/ActiveInfos.h
+++ b/Source/Core/Common/Config/ActiveInfos.h
@@ -1,0 +1,88 @@
+// Copyright 2020 Dolphin Emulator Project
+// Licensed under GPLv2+
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <memory>
+#include <optional>
+#include <unordered_set>
+#include <vector>
+
+#include "Common/Config/Config.h"
+#include "Common/Config/Layer.h"
+#include "Common/Config/Location.h"
+
+namespace Config
+{
+template <typename T>
+class Info;
+
+class ActiveInfosBase
+{
+public:
+  static void InvalidateCache();
+
+protected:
+  // unique_ptr is used to avoid problems with initialization order.
+  // We should be able to stop using unique_ptr once we are using C++20 everywhere,
+  // since C++20 makes std::vector's default constructor constexpr.
+  static std::unique_ptr<std::vector<void (*)()>> s_invalidate_cache_functions;
+
+  static bool s_program_is_exiting;
+};
+
+template <typename T>
+class ActiveInfos : public ActiveInfosBase
+{
+public:
+  static void Add(Info<T>* info)
+  {
+    if (!s_active_infos)
+    {
+      if (!s_invalidate_cache_functions)
+        s_invalidate_cache_functions = std::make_unique<std::vector<void (*)()>>();
+      s_invalidate_cache_functions->push_back(&InvalidateCache);
+
+      s_active_infos = std::make_unique<InfoSet>();
+    }
+    s_active_infos->set.insert(info);
+  }
+
+  static void Remove(Info<T>* info)
+  {
+    if (!s_program_is_exiting)
+      s_active_infos->set.erase(info);
+  }
+
+  static void InvalidateCache()
+  {
+    for (Info<T>* info : s_active_infos->set)
+      Update(info);
+  }
+
+  static void Update(Info<T>* info)
+  {
+    info->m_cached_value = Get(info->GetLocation()).value_or(info->GetDefaultValue());
+  }
+
+private:
+  static std::optional<T> Get(const Location& location)
+  {
+    const std::shared_ptr<Layer> layer = GetLayer(GetActiveLayerForConfig(location));
+    return layer ? layer->Get<T>(location) : std::nullopt;
+  }
+
+  struct InfoSet
+  {
+    // This destructor exists to avoid a crash when the program exits.
+    ~InfoSet() { s_program_is_exiting = true; }
+
+    std::unordered_set<Info<T>*> set;
+  };
+
+  // unique_ptr is used to avoid problems with initialization order.
+  static inline std::unique_ptr<InfoSet> s_active_infos;
+};
+
+}  // namespace Config
--- a/Source/Core/Common/Config/ConfigInfo.h
+++ b/Source/Core/Common/Config/ConfigInfo.h
@@ -4,10 +4,11 @@
 
 #pragma once
 
-#include <string>
 #include <type_traits>
+#include <utility>
 
-#include "Common/Config/Enums.h"
+#include "Common/Config/ActiveInfos.h"
+#include "Common/Config/Location.h"
 
 namespace Config
 {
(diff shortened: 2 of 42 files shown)


Techjar:
Can you provide some before/after benchmarks for reference?

JosJuice:
There isn't really any difference in emulation performance right now since we aren't calling `Config::Get` in hot paths. I'll see later if I can port over some frequently accessed setting to the new config system to test.

Techjar:
I actually just meant raw benchmarks of the config calls, like put it in a loop of a million iterations or something, just to get an idea of how much of an impact this will have.

Source/Core/Common/Config/ConfigInfo.h
```diff
@@ -18,36 +18,43 @@ template <typename T>
...
+                                             other.GetDefaultValue())}
   {
+    ActiveInfos<T>::Update(this);
+    ActiveInfos<T>::Add(this);
   }
 
-  Location location;
-  T default_value;
+  Info(const Info<T>& other) = default;
+  Info(Info<T>&& other) = delete;
```
lioncash, line 47: The assignment operators should be visibly defaulted or deleted to make the assignment semantics explicit as well.

lioncash (commented):
What would the use case of querying the config in hot paths be used for, just to get an example on the record?

JosJuice:
@Techjar One million calls to `Config::Get(const Info<bool>&)` and adding each returned value to an int variable:

Before: 2856 ms
After: 0.001 ms

But this kind of tight loop is of course not representative of actual usage.

@lioncash The CPU clock setting in particular is accessed once for each call to `CoreTiming::Advance()`.

endrift:
How the heck did you manage to crash clang++?

JosJuice:
I'm wondering the same thing...

Techjar:
Of course it's not a fully representative sample, but a difference of over 6 orders of magnitude (from seconds to microseconds) for the tight loop is a pretty safe bet.

Source/Core/Common/Config/ActiveInfos.h
```diff
@@ -0,0 +1,88 @@
...
+public:
+  static void InvalidateCache();
+
+protected:
+  // unique_ptr is used to avoid problems with initialization order.
+  // We should be able to stop using unique_ptr once we are using C++20 everywhere,
+  // since C++20 makes std::vector's default constructor constexpr.
+  static std::unique_ptr<std::vector<void (*)()>> m_invalidate_cache_functions;
+
+  static bool m_program_is_exiting;
```
AdmiralCurtiss, line 32: Please don't prefix static variables with `m_`, that's very confusing. In fact, if you can, please avoid static variables at all...

What exactly does this `m_program_is_exiting` protect against, anyway?

JosJuice: Ah, I wasn't sure whether to use `s_` or `m_` for static class members. I'll change it to `s_` then.

`ActiveInfos::Remove` can get called from the destructor of a `ConfigInfo` after the destructor of `m_active_infos` runs, so I added `m_program_is_exiting` so that `ActiveInfos::Remove` can check whether it's okay to access `m_active_infos`.

endrift:
clang++ backtrace for the segfault:

```
* thread #1, name = 'c++', stop reason = signal SIGSEGV
  * frame #0: 0x0000000000a9a3eb c++`clang::Expr::getExprLoc(void) const + 27
    frame #1: 0x000000000093e96a c++`clang::CodeGen::CodeGenFunction::GenerateCXXGlobalVarDeclInitFunc(llvm::Function*, clang::VarDecl const*, clang::CodeGen::CodeGenFunction::GenerateCXXGlobalVarDeclInitFunc::GlobalVariable*, bool) + 314
    frame #2: 0x000000000093e07c c++`clang::CodeGen::CodeGenModule::EmitCXXGlobalVarDeclInitFunc(clang::VarDecl const*, llvm::GlobalVariable*, bool) + 572
    frame #3: 0x00000000007ec2d3 c++`clang::CodeGen::CodeGenModule::EmitGlobalVarDefinition(clang::VarDecl const*, bool) + 3219
    frame #4: 0x00000000007e6a97 c++`clang::CodeGen::CodeGenModule::EmitGlobalDefinition(clang::GlobalDecl, llvm::GlobalValue*) + 263
    frame #5: 0x00000000007ed29d c++`clang::CodeGen::CodeGenModule::EmitTopLevelDecl(clang::Decl*) + 173
    frame #6: 0x00000000014d0531 c++`clang::Sema::InstantiateVariableDefinition(clang::SourceLocation, clang::VarDecl*, bool, bool, bool) + 5121
    frame #7: 0x00000000014d0a54 c++`clang::Sema::PerformPendingInstantiations(bool) + 900
    frame #8: 0x00000000014ce3ec c++`clang::Sema::InstantiateFunctionDefinition(clang::SourceLocation, clang::FunctionDecl*, bool, bool, bool) + 3692
    frame #9: 0x00000000014d0850 c++`clang::Sema::PerformPendingInstantiations(bool) + 384
    frame #10: 0x00000000014ce3ec c++`clang::Sema::InstantiateFunctionDefinition(clang::SourceLocation, clang::FunctionDecl*, bool, bool, bool) + 3692
    frame #11: 0x00000000014d0850 c++`clang::Sema::PerformPendingInstantiations(bool) + 384
    frame #12: 0x00000000011fb18e c++`clang::Sema::ActOnEndOfTranslationUnit(void) + 382
    frame #13: 0x000000000124a73b c++`clang::Parser::ParseTopLevelDecl(clang::OpaquePtr<clang::DeclGroupRef>&) + 459
    frame #14: 0x00000000007bdc15 c++`clang::ParseAST(clang::Sema&, bool, bool) + 405
    frame #15: 0x00000000007b3af5 c++`clang::FrontendAction::Execute(void) + 69
    frame #16: 0x0000000000d03408 c++`clang::CompilerInstance::ExecuteAction(clang::FrontendAction&) + 1176
    frame #17: 0x000000000040b536 c++`clang::ExecuteCompilerInvocation(clang::CompilerInstance*) + 4326
    frame #18: 0x0000000000400999 c++`char const* cc1_main(Argv=<unavailable>, Argv0="/usr/bin/c++", MainAddr=0x0000000000406020) at cc1_main.cpp:221
    frame #19: 0x0000000000408bb3 c++`main [inlined] llvm::StringRef ExecuteCC1Tool(Tool=(Data = <no value available>, Length = 0)) at driver.cpp:309
    frame #20: 0x0000000000408b8d c++`main(argc_=<unavailable>, argv_=<unavailable>) at driver.cpp:388
    frame #21: 0x0000000000400275 c++`_start + 149
```

JosJuice:
I think I'm just going to give up on this PR. I don't know what to do about the clang segfault, and the code is kinda ugly anyway.

PR #9104, opened 2020-09-21
DolphinQt: Handle non-ASCII characters in Windows cmd arguments

JosJuice:
`CommandLineParse` expects UTF-8 strings. (`QApplication`, on the other hand, seems to be designed so that you can pass in the `char**` `argv` untouched on Windows and get proper Unicode handling.)

### DIFF
--- a/Source/Core/Common/StringUtil.cpp
+++ b/Source/Core/Common/StringUtil.cpp
@@ -31,6 +31,7 @@
 
 #ifdef _WIN32
 #include <Windows.h>
+#include <shellapi.h>
 constexpr u32 CODEPAGE_SHIFT_JIS = 932;
 constexpr u32 CODEPAGE_WINDOWS_1252 = 1252;
 #else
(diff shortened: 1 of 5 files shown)


shuffle2:
...where are you seeing a problem?
image
(this is already handled by qtmain)

shuffle2:
I'm guessing you used cmake, and dolphin's cmake seems to be lacking this currently: https://stackoverflow.com/questions/14115024/how-to-link-qtmain-in-cmake-with-qt5

JosJuice:
`WinMain` is calling `WideCharToMultiByte`, which converts to the system codepage, i.e. not UTF-8.

My test case is placing a file named "Pokémon Channel.rvz" in the same folder as Dolphin and running `Dolphin -e "Pokémon Channel.rvz"` in `cmd`. It works in this PR but not in master.

I was not using CMake.

shuffle2:
OK, it seems an improvement would be to not use qtmain and _only_ this PR, then?

JosJuice:
If qtmain doesn't have any purpose other than parsing command line arguments, then I suppose so. I'm not very familiar with what it does.

shuffle2:
Yea (link in my post above is the complete qtmain.lib sources).
To remove it from dolphin, just need to remove it from [being linked](https://github.com/dolphin-emu/dolphin/blob/master/Source/VSProps/QtCompile.props#L38) and make `#ifdef _WIN32` around the `main` definition to let windows build use a `wWinMain` instead.

JosJuice:
Does this look good?

shuffle2:
yea

MisakaMikotoDolphin:
Thank you! I had given up hope that this issue would be solved. Thanks again for fixing this niggling problem.

PR #9106, opened 2020-09-22
WiimoteEmu: Increase camera FOV to match that of a real Wii remote.

jordan-woyak:
I've measured the field of view of Nintendo Wii Remote IR cameras (original and "MotionPlus Inside") using the known distance of a sensor bar with LEDs of known separation and trigonometry!

It comes out at 42 degrees horizontally by 31.5 vertically.
This is an increase from the 33 by 23 guess we took from Wiibrew.

Increasing the FOV on the emulated camera makes `IMUPointer`'s simulated sensor bar pointing noticeably less sensitive.
Many users complained it was too sensitive. Gyro pointing should more closely match IR pointing now.

We could potentially expose the FOV as an adjustable setting.
Even though it shouldn't ever change it would allow adjusting the sensitivity of gyro pointing.

Point's "Total Yaw" and "Total Pitch" settings must be increased to compensate for the higher FOV to produce similar behavior in game.
I've adjusted the defaults a bit more than just FOV compensation so screen edges in the Wii Menu can now be reached by default.

I've made the "Total Yaw" default of `IMUPointer` match.
I've removed the line from our included Wii remote profile to take on the default value.

I've added an "IR Distance" input to ControllerInterface's Wii Remote device which performs a sensor bar distance calculation using math.

Merging this can close #8631.

### DIFF
--- a/Source/Core/Common/MathUtil.h
+++ b/Source/Core/Common/MathUtil.h
@@ -135,8 +135,12 @@ public:
 
   constexpr size_t Count() const { return m_running_mean.Count(); }
   constexpr T Mean() const { return m_running_mean.Mean(); }
+
   constexpr T Variance() const { return m_variance / (Count() - 1); }
-  constexpr T StandardDeviation() const { return std::sqrt(Variance()); }
+  T StandardDeviation() const { return std::sqrt(Variance()); }
+
+  constexpr T PopulationVariance() const { return m_variance / Count(); }
+  T PopulationStandardDeviation() const { return std::sqrt(PopulationVariance()); }
 
 private:
   RunningMean<T> m_running_mean;
--- a/Data/Sys/Profiles/Wiimote/Wii Remote with MotionPlus Pointing.ini
+++ b/Data/Sys/Profiles/Wiimote/Wii Remote with MotionPlus Pointing.ini
@@ -21,7 +21,6 @@ IMUGyroscope/Roll Right = `Gyro Roll Right`
 IMUGyroscope/Yaw Left = `Gyro Yaw Left`
 IMUGyroscope/Yaw Right = `Gyro Yaw Right`
 IMUIR/Enabled = True
-IMUIR/Total Yaw = 20
 Extension/Attach MotionPlus = `Attached MotionPlus`
 Extension = `Attached Extension`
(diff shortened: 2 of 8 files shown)


Source/Core/Common/MathUtil.h
```diff
@@ -135,9 +135,13 @@ class RunningVariance
 
   constexpr size_t Count() const { return m_running_mean.Count(); }
   constexpr T Mean() const { return m_running_mean.Mean(); }
+
   constexpr T Variance() const { return m_variance / (Count() - 1); }
   constexpr T StandardDeviation() const { return std::sqrt(Variance()); }
```
lioncash, line 140: I'm pretty sure both this and `PopulationStandardDeviation` cannot ever actually be used in a constexpr context, since `std::sqrt` isn't constexpr

iwubcode (approved):
LGTM.

I tested a number of Wii games with pointer controls using the built-in profile that was recently merged:

- Boom Blox
- Bully Scholarship Edition
- Epic Mickey
- Final Fantasy Crystal Bearers
- Link's Crossbow Training
- Lost Winds
- Mini Ninjas
- Red Steel
- Silent Hill Shattered Memories
- Trauma Center: Second Opinion

The only one that couldn't get to the edge of the screen was Mini Ninjas.

This was my first time testing Crossbow Training with an actual Wiimote.  I found it difficult to interact precisely.  I don't think this review changed that but wanted to mention it.

Source/Core/Common/MathUtil.h
jordan-woyak: True. I was hoping it would be `constexpr` in C++20, but apparently not.

JMC47:
My only worry is that people with existing controller profiles on the old FOV will need to update things, but that's not really a huge deal.  We'll just have to mention this in the progress report.

Techjar (approved):
Code looks sensible to me (besides the thing lioncash already pointed out). Tested it and gyro IR feels much better (less sensitive) with the larger FOV.

Source/Core/Common/MathUtil.h
jordan-woyak: Fixed

PR #9109, opened 2020-09-24
FramebufferManager: Fix EFB readbacks with MSAA on

stenzek:
Regression from #8827.


JMC47:
I tried to load a savestate from Vulkan to OpenGL and emulation suddenly crashed.

Trying to savestate in OpenGL crashes emulation.

EFB access appears to be working in both backends,  though.

iwubcode (approved):

JMC47:
Tested this and it appears to be working now.  Thanks!

PR #9095, opened 2020-09-18
Android: Long press a setting to reset it

JosJuice:
This is particularly important for game INIs, where a setting being unset is not the same as it being set to the default value.

PR #9110, opened 2020-09-25
Android: Add "Default Values" button for overlay seekbars

Ebola16:
Quickly returning to the default value and not making users remember that value is good.
Note: #9095 will handle this for other settings by clearing them.


JosJuice:
Instead of hardcoding the default value in the "set to default value" code, you should be able to use `IniFile.deleteKey` and `SharedPreferences.Editor.remove` to get rid of any non-default value.

Ebola16:
This PR now overlaps with #9152 so I'll combine them.

PR #9111, opened 2020-09-25
Android: Don't run emulation menu actions if emulation menu isn't running

Ebola16:
Should prevent https://bugs.dolphin-emu.org/issues/12278


JMC47:
@JosJuice what do you think of this?

JosJuice:
I'm skeptical that this is a proper fix, but I haven't had time to look into exactly what's causing the issue, so I'm not sure.

PR #9112, opened 2020-09-26
Android: seekbar.setMin requires API level 26

Ebola16:
Apparently there's no good alternative before `setMin` was implemented for dealing with minimums less than 0.

### DIFF
--- a/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/ui/SettingsAdapter.java
+++ b/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/ui/SettingsAdapter.java
@@ -2,6 +2,7 @@ package org.dolphinemu.dolphinemu.features.settings.ui;
 
 import android.content.Context;
 import android.content.DialogInterface;
+import android.os.Build;
 import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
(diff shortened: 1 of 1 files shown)


JosJuice:
Oh, indeed. Weird that it didn't exist before then, but good catch.

Could you add some clamping to `SliderSetting.getValue` (i.e. changing `return x` to `return Math.max(x, getMin())`) so that we don't write invalid values to the SYSCONF, preferably with a comment mentioning the API 26 problem? I can merge it after that.

JosJuice:
On second thought, it's probably better to clamp here, so that the workaround for the issue is in the same part of the code as the issue: https://github.com/dolphin-emu/dolphin/blob/86a37214b74f9a11ceea5ed18981eb3ba016497f/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/ui/SettingsAdapter.java#L417-L421

Ebola16:
Like this? Since we're now clamping, I'm assuming it's cleaner to remove `setMin` completely.

JosJuice:
Please keep calling `setMin` on API 26 and above. That way we can avoid subpar UX for users who are on API 26 or higher. Being able to set the slider to 0 but then having it revert to 1 isn't very clean from the user's perspective.

Alternatively... Maybe the solution which would be the cleanest for all users (but would require slightly more code) would be to not call `setMin` at all, and instead call `getMax` with `item.getMax() - item.getMin()`, and adjust for the discrepancy in `onSliderClick` and `onProgressChanged`.

Ebola16:
The problem with adjusting for `item.getMin` in `onProgressChanged` is that multiple presses of the slider dialog will apply the adjustment multiple times. Is the current state of this PR acceptable instead?

JosJuice:
I'll merge this as-is so that we'll have it fixed before tagging the next beta. I might make another PR later to refine the logic.

PR #9113, opened 2020-09-26
Add .ini configuration files for Phantasy Star Online Episode I&II Trial Edition and Episode III Trial Edition

ShiftaDeband:
This pull request adds two new .ini configuration files for game IDs DPSJ8P (Phantasy Star Online Episode I&II Trial Edition) and DPOJ8P (Phantasy Star Online Episode III C.A.R.D. Revolution Trial Edition).

To summarize, both games require the same `SafeTextureCacheColorSamples = 512` configuration to render text more properly. PSO Episode III appears to have or had issues with Immediate XFB, so it is set to `ImmediateXFBEnable = False`.

In addition, the PSO Ep. I&II Trial Edition requires a modem to be present. Since the modem was the only option at the time, the BBA will not be recognized. This requires either a 32-bit patch (80194F40 4182002C, thank you to evilhamwizard) or an Action Replay code (04194F40 4182002C, thank you Sappharad) to bypass the initial screen as outlined on [the wiki page](https://wiki.dolphin-emu.org/index.php?title=Phantasy_Star_Online_Episode_I_%26_II_Trial_Edition).

Thank you for your consideration.

### DIFF
--- a/Data/Sys/GameSettings/DPOJ8P.ini
+++ b/Data/Sys/GameSettings/DPOJ8P.ini
@@ -1,0 +1,16 @@
+# DPOJ8P - PHANTASY STAR ONLINE EPISODE I&II TRIAL EDITION
+
+[Core]
+# Values set here will override the main Dolphin settings.
+
+[OnLoad]
+# Add memory patches to be loaded once on boot here.
+
+[OnFrame]
+# Add memory patches to be applied every frame here.
+
+[ActionReplay]
+# Add action replay cheats here.
+
+[Video_Settings]
+SafeTextureCacheColorSamples = 512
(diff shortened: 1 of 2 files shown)


Rumi-Larry:
Since there has been fixes to EFB emulation recently, it would be better if someone could confirm if InmediateXFB was still broken on this game, so it isn't turned off unecessarilly.

JosJuice:
The fixes to EFB readback shouldn't affect the behavior of Immediately Present XFB.

ShiftaDeband:
You bet. I’ll test and get back to you all within the next 24 hours.

ShiftaDeband:
It appears that this game used to have issues rendering some of the FMVs without that option. After testing, it appears that it may not be needed for this game now.

Removed and requesting another review.

Thank you!

leoetlino:
[15:20:03] leoetlino: JMC47: does https://github.com/dolphin-emu/dolphin/pull/9113 look fine to you?
[15:32:48] JMC47: leoetlino: looks good to me

PR #9117, opened 2020-09-27
Add tapserver network interface for macOS since TunTap is now unmaintained

fuzziqersoftware:
The existing TAP_Apple implementation uses TunTap, which has recently become unmaintained. It seems Apple wants developers to move away from kexts in general; TunTap currently takes some finagling to work on Catalina, and it may not work at all on Big Sur. This necessitates a non-kext-based implementation.

Fortunately, fake Ethernet devices were introduced in Sierra and can be used similarly to tap adapters. This commit adds a new type of BBA interface implementation which uses fake Ethernet devices via tapserver (https://github.com/fuzziqersoftware/tapserver) to communicate with the host. This implementation was tested with PSO Episodes I & II, which can successfully connect to a private server running locally. We use tapserver instead of creating the interfaces within Dolphin since that has to be done as root, and we'd prefer not to require Dolphin itself to run as root to access the network.

This implementation is only available on macOS, since that's the only place it's needed - Windows/Linux/Unix are unaffected by TunTap being deprecated.


Source/Core/Core/HW/EXI/BBA/TAPServer_Apple.cpp
```diff
@@ -0,0 +1,127 @@
...
+// refuses to activate.
+
+static const std::string socket_path = "/tmp/dolphin-tap";
+
+bool CEXIETHERNET::TAPServerNetworkInterface::Activate()
+{
+  if (IsActivated())
+    return true;
+
+  struct sockaddr_un sun;
```
lioncash, line 29:
```suggestion
  struct sockaddr_un sun = {};
```

Is sufficient to zero the structure out. `memset` isn't necessary.

Source/Core/Core/HW/EXI/BBA/TAPServer_Apple.cpp
```diff
@@ -0,0 +1,127 @@
...
+#include "Core/HW/EXI/EXI_DeviceEthernet.h"
+
+namespace ExpansionInterface
+{
+// This interface is only implemented on macOS, since macOS needs a replacement
+// for TunTap when the kernel extension is no longer supported. This interface
+// only appears in the menu on macOS, so on other platforms, it does nothing and
+// refuses to activate.
+
+static const std::string socket_path = "/tmp/dolphin-tap";
```
lioncash, line 22:
```suggestion
constexpr char socket_path[] = "/tmp/dolphin-tap";
```

So that this doesn't require a runtime static constructor.

Source/Core/Core/HW/EXI/BBA/TAPServer_Apple.cpp
```diff
@@ -0,0 +1,127 @@
...
+}
+
+bool CEXIETHERNET::TAPServerNetworkInterface::SendFrame(const u8* frame, u32 size)
+{
+  {
+    const std::string s = ArrayToString(frame, size, 0x10);
+    INFO_LOG(SP1, "SendFrame %x\n%s\n", size, s.c_str());
+  }
+
+  uint16_t size16 = size;
```
lioncash, line 65:
```suggestion
  u16 size16 = size;
```

lioncash, line 65:
```suggestion
  auto size16 = u16(size);
```

otherwise this can cause a truncation warning.

Source/Core/Core/HW/EXI/BBA/TAPServer_Apple.cpp
```diff
@@ -0,0 +1,127 @@
...
+  if (write(fd, &size16, 2) != 2)
+  {
+    ERROR_LOG(SP1, "SendFrame(): could not write size field\n");
+    return false;
+  }
+  int writtenBytes = write(fd, frame, size);
```
lioncash, line 71:
```suggestion
  int written_bytes = write(fd, frame, size);
```

Source/Core/Core/HW/EXI/BBA/TAPServer_Apple.cpp
```diff
@@ -0,0 +1,127 @@
...
+  if ((u32)writtenBytes != size)
```
lioncash, line 72:
```suggestion
  if (u32(written_bytes) != size)
```

Source/Core/Core/HW/EXI/BBA/TAPServer_Apple.cpp
```diff
@@ -0,0 +1,127 @@
...
+    FD_ZERO(&rfds);
+    FD_SET(fd, &rfds);
+
+    struct timeval timeout;
+    timeout.tv_sec = 0;
+    timeout.tv_usec = 50000;
+    if (select(fd + 1, &rfds, nullptr, nullptr, &timeout) <= 0)
+      continue;
+
+    uint16_t size;
```
lioncash, line 98:
```suggestion
    u16 size;
```

Source/Core/Core/HW/EXI/BBA/TAPServer_Apple.cpp
```diff
@@ -0,0 +1,127 @@
...
+    if (read(fd, &size, 2) != 2)
+    {
+      ERROR_LOG(SP1, "Failed to read size field from BBA, err=%d", errno);
+    }
+    else
+    {
+      int readBytes = read(fd, m_eth_ref->mRecvBuffer.get(), size);
```
lioncash, line 105:
```suggestion
      int read_bytes = read(fd, m_eth_ref->mRecvBuffer.get(), size);
```

Source/Core/Core/HW/EXI/BBA/TAPServer_Apple.cpp
```diff
@@ -0,0 +1,127 @@
...
+      if (readBytes < 0)
+      {
+        ERROR_LOG(SP1, "Failed to read packet data from BBA, err=%d", errno);
+      }
+      else if (readEnabled.IsSet())
+      {
+        std::string dataString = ArrayToString(m_eth_ref->mRecvBuffer.get(), readBytes, 0x10);
```
lioncash, line 112:
```suggestion
        std::string data_string = ArrayToString(m_eth_ref->mRecvBuffer.get(), readBytes, 0x10);
```

fuzziqersoftware:
Thanks for the review - I've made all the suggested changes.

BhaaLseN:
If this is really macOS only, you might want to add a couple more platform `#ifdef`. This would probably _also_ mean that you don't need the `Stub` though.

In case we _do_ want to keep this for all platforms, you'll have to update the Visual Studio project files as well for Windows.

fuzziqersoftware:
Added some more #ifdefs and removed the stub, since this interface type probably will never be needed for non-macOS platforms.

PR #9118, opened 2020-09-29
Create KHP.ini

cobalt2727:
add GameSettings file for a modpack that uses its own game ID

### DIFF
--- a/Data/Sys/GameSettings/KHP.ini
+++ b/Data/Sys/GameSettings/KHP.ini
@@ -1,0 +1,17 @@
+# KHPE01 - Kirby Air Ride Hack Pack 
+# copy of GKY.ini
+
+[Core]
+# Values set here will override the main Dolphin settings.
+
+[OnLoad]
+# Add memory patches to be loaded once on boot here.
+
+[OnFrame]
+# Add memory patches to be applied every frame here.
+
+[ActionReplay]
+# Add action replay cheats here.
+
+[Video_Stereoscopy]
+StereoConvergence = 6


JosJuice:
I don't think we've added a default game INI for a hacked game before. Just setting `StereoConvergence` doesn't seem like a great reason to start doing it to me.

mbc07:
I don't think we should start carrying game INIs for modded/custom games either, it can get out of hand really fast. Perhaps you could distribute the custom game INI together with the mod?

JMC47:
I'm actually on the other side of this.  With the final Wii game being released now, it's mostly romhacks/patches.  We've already seen a lot of popular ones in Dolphin.

I think instead of trying to downplay their existence, we should embrace it as Dolphin is the easiest way to play these hacks.

mbc07:
It's not a matter of downplaying their existence, my main issue with that is that the IDs for these mods are totally inconsistent. Flacs and Kolano started documenting those custom game IDs on the wiki a while ago and there are cases of completely different game mods reusing the same custom ID, thus optimal settings for a mod won't be optimal for another, and there's no way to make that distinction when they reuse the same ID.

Also, those mods have to be distributed somewhere before people can get them and they generally don't come ready to play (e.g. mod is distributed as a diff patch for the base game), thus they ship with instructions, patch tools and so on. They could easily ship the modded game INI together with the mod and it would be an INI guaranteed to work with that specific mod version...

JMC47:
The first step to having things like this fixed it to acknowledge the problem though.  If we worked with mod creators and established standards/documented things, we could make sure that popular mods designed for release with Dolphin had proper game IDs.  We could be the ones to set the standard going forward.

I totally get that there are a lot of problems, but we've actually run into this with real games too!  Should we not include gameIDs for Datel stuff because it's unlicensed/unofficial?

cobalt2727:
Well I've apparently started a decent debate here. I'm inclined to back up JMC's reasoning, and while I definitely understand where @mbc07 is coming from, that's only really an issue if two mods for _different games_ share an ID, right? Like, I copied the contents of this file from GKY.ini since KHP is effectively just Kirby Air Ride with QoL mods.

> I totally get that there are a lot of problems, but we've actually run into this with real games too! Should we not include gameIDs for Datel stuff because it's unlicensed/unofficial?

@JMC47 wait, is there Datel stuff that shares IDs with real games or is that a separate issue?

mbc07:
> The first step to having things like this fixed it to acknowledge the problem though. If we worked with mod creators and established standards/documented things, we could make sure that popular mods designed for release with Dolphin had proper game IDs.

Fair enough, but in that case I would rather have those details properly sorted out before including custom game INIs on the codebase, than including potentially clashing game INIs as they are now and deal with the standardization later.
<br/>

> I totally get that there are a lot of problems, but we've actually run into this with real games too! Should we not include gameIDs for Datel stuff because it's unlicensed/unofficial?

That's exactly the case at the moment, there's no game INIs for Datel stuff despite them requiring non-default settings to boot properly, even the game IDs they used that doesn't conflict with officially released games (DTLE01, DTLX01, RFLPWK) also lacks game INIs. 

Apart from Datel, AFAICT the only other case of conflicting IDs are from TGC images extracted from official bonus/demo discs, but that only becomes an issue if you try to boot the TGC file directly, otherwise they inherit the same ID from the disc that contains them (an official ID, not reused)...

leoetlino:
Closing for the reasons that have been mentioned above.

I'm not opposed to the idea of adding GameINIs for mods (though I still think it'd be easier to just distribute INIs alongside the mod) but IMO this is currently not a good idea because of the ID issue and because any updates to GKY will also have to be manually copied to KHP and [insert all the GameINIs for other mods]. It's annoying for maintenance... It might be wiser to keep the original 3-letter game ID and only change the last three bytes instead.

PR #9119, opened 2020-09-29
Android: Correct SliderSetting minimum value behavior on API < 26

JosJuice:
See PR #9112.


Ebola16 (approved):

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/ui/SettingsAdapter.java
```diff
@@ -240,12 +242,11 @@ public void onSliderClick(SliderSetting item, int position)
 
     SeekBar seekbar = view.findViewById(R.id.seekbar);
 
-    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O)
```
Techjar, line 243: I believe the import for `Build` should be removed as well as this was the only reference.

PR #9120, opened 2020-09-29
Fix controller defaults being empty on fresh run

JosJuice:
A very early call to Config::Save is now creating empty controller INI files.

https://bugs.dolphin-emu.org/issues/12283


jordan-woyak (approved):
Code looks fine. Untested.

faroukabd000:
Tested ★÷★work fine for me

PR #4601, opened 2017-01-03
ShaderGen: Always calculate lighting for both color channels

stenzek:
Up until now, we assumed that the **XFMEM_SETNUMCHAN** register controlled how many channels the lighting pipeline processes. If this was set to zero, no lighting was calculated at all, and the colors were passed through as-is from the corresponding attribute.

Cel-Damage is an interesting case, as it doesn't module the vertex color with the texture result as you would expect. Instead, it uses two sets of texture coordinates. The first reference a diffuse texture, as usual. The second set of texture coordinates, however, are derived from the vertex color. This set indexes a LUT-like texture, which is blended with the first texture result in a second TEV stage, to produce the lit pixels.

As we were setting the vertex color to white before calculating texture coordinates, this resulted in the same texel (at normalized coordinates 1,1) being used from the LUT texture, and flat, brightly-colored polygons. The software rasterizer used black as a "default" color, resulting in flat, dark-colored polygons.

This PR changes the shader generation to always process the lighting pipeline, and only "clear" the colors based on **XFMEM_SETNUMCHAN** after texture coordinate generation, fixing the lighting in Cel-Damage.

I've tested this behavior on hardware by using the lighting pipeline's color as texture coordinates, as cel-damage does, and comparing results when **XFMEM_SETNUMCHAN** is set to zero and one. No change is visible, confirming that the lighting pipeline does indeed run regardless of this register's value.

This opens up the door to further questions though, see my comment below for more info. However, to keep things simple, this change is self-contained, fixes cel-damage, and has no other regressions in fifoci.

PR #3315, opened 2015-12-05
ShaderGen: Toggle value of uninitialized color.

degasus:
SMS seems to need 0, no regressions either.

PR #9122, opened 2020-09-30
Fix Super Mario Sunshine debug cubes (originally PR #4601 by stenzek)

Pokechu22:
Code copied from #4601 before it was changed to remove the Super Mario Sunshine debug cube fixes, specifically from commits 0ddcd28f16c4e57d7a3ad8b2665e947d242ddd81 and 8fa5b13afe25e3030be89c4482e3a74f66739136.  I haven't tested this code myself, and don't fully understand what it does; this is just those two commits cherry-picked with merge conflicts fixed (some parts were already merged).

FifoCI for these changes: ~~https://fifo.ci/version/07b69225a24138559aa36034104464b01ba85823/~~ https://fifo.ci/version/d6ce8eef36f5b0bfd0078c51701b052cc9a77005/
FifoCI for #4601's changes: https://fifo.ci/version/0ddcd28f16c4e57d7a3ad8b2665e947d242ddd81/

### DIFF
--- a/Source/Core/VideoBackends/Software/SWVertexLoader.cpp
+++ b/Source/Core/VideoBackends/Software/SWVertexLoader.cpp
@@ -18,6 +18,7 @@
 #include "VideoBackends/Software/Tev.h"
 #include "VideoBackends/Software/TransformUnit.h"
 
+#include "VideoCommon/CPMemory.h"
 #include "VideoCommon/DataReader.h"
 #include "VideoCommon/IndexGenerator.h"
 #include "VideoCommon/OpcodeDecoding.h"
@@ -68,9 +69,6 @@ void SWVertexLoader::DrawCurrentBatch(u32 base_index, u32 num_indices, u32 base_
   {
     const u16 index = m_cpu_index_buffer[i];
     memset(static_cast<void*>(&m_vertex), 0, sizeof(m_vertex));
-
-    // Super Mario Sunshine requires those to be zero for those debug boxes.
-    m_vertex.color = {};
 
     // parse the videocommon format to our own struct format (m_vertex)
     SetFormat(g_main_cp_state.last_id, primitiveType);
@@ -184,6 +182,39 @@ static void ReadVertexAttribute(T* dst, DataReader src, const AttributeFormat& f
   }
 }
 
+static void ParseColorAttributes(InputVertexData* dst, DataReader& src,
+                                 const PortableVertexDeclaration& vdec)
+{
+  const auto set_default_color = [](u8* color, int i) {
+    // The default alpha channel seems to depend on the number of components in the vertex format.
+    const auto& g0 = g_main_cp_state.vtx_attr[g_main_cp_state.last_id].g0;
+    const u32 color_elements = i == 0 ? g0.Color0Elements : g0.Color1Elements;
+    color[0] = color_elements == 0 ? 255 : 0;
+    color[1] = 255;
+    color[2] = 255;
+    color[3] = 255;
+  };
+
+  if (vdec.colors[0].enable)
+  {
+    // Use color0 for channel 0, and color1 for channel 1 if both colors 0 and 1 are present.
+    ReadVertexAttribute<u8>(dst->color[0].data(), src, vdec.colors[0], 0, 4, true);
+    if (vdec.colors[1].enable)
+      ReadVertexAttribute<u8>(dst->color[1].data(), src, vdec.colors[1], 0, 4, true);
+    else
+      set_default_color(dst->color[1].data(), 1);
+  }
+  else
+  {
+    // If only one of the color attributes is enabled, it is directed to color 0.
+    if (vdec.colors[1].enable)
+      ReadVertexAttribute<u8>(dst->color[0].data(), src, vdec.colors[1], 0, 4, true);
+    else
+      set_default_color(dst->color[0].data(), 0);
+    set_default_color(dst->color[1].data(), 1);
+  }
+}
+
 void SWVertexLoader::ParseVertex(const PortableVertexDeclaration& vdec, int index)
 {
   DataReader src(m_cpu_vertex_buffer.data(),
@@ -197,10 +228,7 @@ void SWVertexLoader::ParseVertex(const PortableVertexDeclaration& vdec, int inde
     ReadVertexAttribute<float>(&m_vertex.normal[i][0], src, vdec.normals[i], 0, 3, false);
   }
 
-  for (std::size_t i = 0; i < m_vertex.color.size(); i++)
-  {
-    ReadVertexAttribute<u8>(m_vertex.color[i].data(), src, vdec.colors[i], 0, 4, true);
-  }
+  ParseColorAttributes(&m_vertex, src, vdec);
 
   for (std::size_t i = 0; i < m_vertex.texCoords.size(); i++)
   {
--- a/Source/Core/VideoCommon/PixelShaderGen.cpp
+++ b/Source/Core/VideoCommon/PixelShaderGen.cpp
@@ -203,9 +203,6 @@ PixelShaderUid GetPixelShaderUid()
 
   if (g_ActiveConfig.bEnablePixelLighting)
   {
-    // The lighting shader only needs the two color bits of the 23bit component bit array.
-    uid_data->components =
-        (VertexLoaderManager::g_current_components & (VB_HAS_COL0 | VB_HAS_COL1)) >> VB_COL_SHIFT;
     uid_data->numColorChans = xfmem.numChan.numColorChans;
     GetLightingShaderUid(uid_data->lighting);
   }
@@ -768,8 +765,11 @@ ShaderCode GeneratePixelShaderCode(APIType api_type, const ShaderHostConfig& hos
     // out.SetConstantsUsed(C_PLIGHT_COLORS, C_PLIGHT_COLORS+7); // TODO: Can be optimized further
     // out.SetConstantsUsed(C_PLIGHTS, C_PLIGHTS+31); // TODO: Can be optimized further
     // out.SetConstantsUsed(C_PMATERIALS, C_PMATERIALS+3);
-    GenerateLightingShaderCode(out, uid_data->lighting, uid_data->components << VB_COL_SHIFT,
-                               "colors_", "col");
+    GenerateLightingShaderCode(out, uid_data->lighting, "colors_", "col");
+    if (uid_data->numColorChans == 0)
+      out.Write("col0 = float4(0.0, 0.0, 0.0, 0.0);\n");
+    if (uid_data->numColorChans <= 1)
+      out.Write("col1 = float4(0.0, 0.0, 0.0, 0.0);\n");
   }
 
   // HACK to handle cases where the tex gen is not enabled
--- a/Source/Core/VideoCommon/UberShaderVertex.cpp
+++ b/Source/Core/VideoCommon/UberShaderVertex.cpp
@@ -176,8 +176,43 @@ ShaderCode GenVertexShader(APIType api_type, const ShaderHostConfig& host_config
             "\n");
 
   // Hardware Lighting
-  WriteVertexLighting(out, api_type, "pos.xyz", "_norm0", "rawcolor0", "rawcolor1", "o.colors_0",
-                      "o.colors_1");
+  out.Write("// xfmem.numColorChans controls the number of color channels available to TEV,\n"
+            "// but we still need to generate all channels here, as it can be used in texgen.\n"
+            "// Cel-damage is an example of this.\n"
+            "float4 vertex_color_0, vertex_color_1;\n"
+            "\n");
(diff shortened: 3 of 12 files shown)


JMC47:
I can confirm this fixes the SMS gray cubes and Vegas Party didn't regress.  Do we remember the downside to this?

Source/Core/VideoCommon/VertexShaderGen.h
```diff
@@ -43,7 +43,7 @@ struct vertex_shader_uid_data
 
   u32 texMtxInfo_n_projection : 16;  // Stored separately to guarantee that the texMtxInfo struct is
                                      // 8 bits wide
-  u32 pad : 18;
+  u32 pad : 20;
```
Pokechu22, line 46: 23+4+2+1+16 is 46, so a pad of 18 gives 64 bits total while a pad of 20 gives 66 bits total.  I'm not sure why this was changed; does anyone know?  It might just be a left-over from some other changes that previously adjusted the padding but no longer are relevant.

Source/Core/VideoCommon/VertexShaderGen.cpp
```diff
@@ -17,7 +17,6 @@
 VertexShaderUid GetVertexShaderUid()
 {
   ASSERT(bpmem.genMode.numtexgens == xfmem.numTexGen.numTexGens);
-  ASSERT(bpmem.genMode.numcolchans == xfmem.numChan.numColorChans);
```
Pokechu22, line 20: I'm not sure if this assert should be removed; based on 940bccd2412fdec5d22f53d9a9e28fec8ef41207 it was targeted at [bug 10180](https://bugs.dolphin-emu.org/issues/10180), which was instead fixed by #8717.

Pokechu22: I can confirm that the assert isn't hit anymore since #8717 skips rendering in this case.  So, it probably shouldn't be removed.

phire:
Please check for invisible characters in THPS3.

see #3315

JMC47:
I tested for invisible characters in Tony Hawk Pro Skater 3, and it is fine.  In fact, this Pull Request **fixes** invisible characters in Tony Hawk Pro Skater 3 in the software renderer.

Source/Core/VideoCommon/VertexShaderGen.h
stenzek: Yeah, seems like it should be 18.

Source/Core/VideoCommon/VertexShaderManager.cpp
```diff
@@ -617,6 +616,17 @@ void VertexShaderManager::SetVertexFormat(u32 components)
     constants.components = components;
     dirty = true;
   }
+
+  // The default alpha channel seems to depend on the number of components in the vertex format.
+  // If the vertex attribute has an alpha channel, zero is used, otherwise one.
+  u32 color_chan_alpha =
```
stenzek, line 622: const

Source/Core/VideoCommon/VertexShaderGen.cpp
```diff
@@ -187,6 +186,34 @@ ShaderCode GenerateVertexShaderCode(APIType api_type, const ShaderHostConfig& ho
...
   out.Write("VS_OUTPUT o;\n");
 
+  // xfmem.numColorChans controls the number of color channels available to TEV, but we still need
+  // to generate all channels here, as it can be used in texgen. Cel-damage is an example of this.
+  out.Write("float4 vertex_color_0, vertex_color_1;\n");
+
+  // To use color 1, the vertex descriptor must have color 0 and 1.
+  // If color 1 is present but not color 0, it is used for lighting channel 0.
+  bool use_color_1 =
+      (uid_data->components & (VB_HAS_COL0 | VB_HAS_COL1)) == (VB_HAS_COL0 | VB_HAS_COL1);
```
stenzek, line 196: const

stenzek:
Code seems fine, but it's been a while since I messed with this so I can't remember the specifics of the hardware tests. But if there's no known regressions, I say go for it.

It may be worth bumping the UID version too (which will annoy some people), since the vertex struct is changed and you'll have a bunch of "dead" UIDs which don't match up.

JMC47:
I tried finding regressions and couldn't.  Merging it mid-month to get wider testing could be a way to find something...

Source/Core/VideoCommon/VertexShaderGen.cpp
Pokechu22: Assert re-added.

Pokechu22:
Resolved all of those other than bumping the UID version.

JMC47:
This needs to be rebased and we probably should bump the UID version if it does what I think it should.

Source/Core/VideoCommon/UberShaderVertex.cpp
```diff
@@ -176,8 +176,45 @@ ShaderCode GenVertexShader(APIType api_type, const ShaderHostConfig& host_config
...
-                      "o.colors_1");
+  out.WriteFmt("// xfmem.numColorChans controls the number of color channels available to TEV,\n"
+               "// but we still need to generate all channels here, as it can be used in texgen.\n"
+               "// Cel-damage is an example of this.\n"
+               "float4 vertex_color_0, vertex_color_1;\n"
+               "\n");
+  out.WriteFmt("// To use color 1, the vertex descriptor must have color 0 and 1.\n"
+               "// If color 1 is present but not color 0, it is used for lighting channel 0.\n"
+               "bool use_color_1 = ((components & {0}) == {0}); // VB_HAS_COL0 | VB_HAS_COL1\n",
+               VB_HAS_COL0 | VB_HAS_COL1);
```
Pokechu22, line 187: The code in this area previously wrote `static_cast<u32>(VB_HAS_COL0 | VB_HAS_COL1)`.  Based on https://github.com/dolphin-emu/dolphin/pull/4601#pullrequestreview-78152510 I don't think the cast is necessary, so I've removed it.  (This line specifically also had it duplicated, which I've replaced with `{0}`).

Source/Core/VideoCommon/PixelShaderGen.cpp
```diff
@@ -769,8 +766,11 @@ ShaderCode GeneratePixelShaderCode(APIType api_type, const ShaderHostConfig& hos
     // out.SetConstantsUsed(C_PLIGHT_COLORS, C_PLIGHT_COLORS+7); // TODO: Can be optimized further
     // out.SetConstantsUsed(C_PLIGHTS, C_PLIGHTS+31); // TODO: Can be optimized further
     // out.SetConstantsUsed(C_PMATERIALS, C_PMATERIALS+3);
-    GenerateLightingShaderCode(out, uid_data->lighting, uid_data->components << VB_COL_SHIFT,
-                               "colors_", "col");
+    GenerateLightingShaderCode(out, uid_data->lighting, "colors_", "col");
+    if (uid_data->numColorChans == 0)
+      out.Write("col0 = float4(0.0, 0.0, 0.0, 0.0);\n");
+    if (uid_data->numColorChans <= 1)
+      out.Write("col1 = float4(0.0, 0.0, 0.0, 0.0);\n");
```
Pokechu22, line 773: These lines still use `out.Write`.  I didn't change them because there are a few other lines in this file that also use it instead of `WriteFmt`, and because they don't have any format arguments.  If these should be changed, I'll also change the similar calls in the file.

Pokechu22 (commented):
Done, including bumping the shader UID version.

JMC47:
I think this is good to go from what I've checked.  I really hope I'm not missing something, as I did a lot of testing on it.

Source/Core/VideoCommon/PixelShaderGen.cpp
lioncash: They should be changed, given those two remaining write calls will be migrated in #9197. Tthey were accidentally overlooked

Source/Core/VideoCommon/UberShaderVertex.cpp
```diff
@@ -176,8 +176,45 @@ ShaderCode GenVertexShader(APIType api_type, const ShaderHostConfig& host_config
...
+               VB_HAS_COL0);
+  out.WriteFmt(
+      "    float4 color_value;\n"
+      "    // Use color0 for channel 0, and color1 for channel 1 if both colors 0 and 1 are "
+      "present.\n"
+      "    if (color == 0u)\n"
+      "      vertex_color_0 = rawcolor0;\n"
+      "    else\n"
+      "      vertex_color_1 = rawcolor1;\n"
+      "  } else if (color == 0 && (components & %uu) != 0) {\n",
```
lioncash, line 200: braces need to be doubled up in fmt contexts, like `{{`, which several other added lines don't seem to be doing. A format specifier is also using printf-like formatting, which fmt doesn't use (this, too, seems to be an issue on multiple lines of added code).

Pokechu22: Thanks for mentioning this.  I was careful about it for lines where there were merge conflicts (redoing the changes off of the fmt version), but I must have missed some in code that was completely new.

Source/Core/VideoCommon/VertexShaderGen.cpp
```diff
@@ -191,6 +191,35 @@ ShaderCode GenerateVertexShaderCode(APIType api_type, const ShaderHostConfig& ho
...
+  const bool use_color_1 =
+      (uid_data->components & (VB_HAS_COL0 | VB_HAS_COL1)) == (VB_HAS_COL0 | VB_HAS_COL1);
+  for (u32 color = 0; color < NUM_XF_COLOR_CHANNELS; color++)
+  {
+    if ((color == 0 || use_color_1) && uid_data->components & (VB_HAS_COL0 << color))
+    {
+      // Use color0 for channel 0, and color1 for channel 1 if both colors 0 and 1 are present.
+      out.WriteFmt("vertex_color_%u = rawcolor%u;\n", color, color);
+    }
+    else if (color == 0 && (uid_data->components & VB_HAS_COL1))
```
lioncash, line 209:
```suggestion
    else if (color == 0 && (uid_data->components & VB_HAS_COL1) != 0)
```

To be explicit about the comparison

Source/Core/VideoCommon/VertexShaderGen.cpp
```diff
@@ -191,6 +191,35 @@ ShaderCode GenerateVertexShaderCode(APIType api_type, const ShaderHostConfig& ho
...
+  // to generate all channels here, as it can be used in texgen. Cel-damage is an example of this.
+  out.WriteFmt("float4 vertex_color_0, vertex_color_1;\n");
+
+  // To use color 1, the vertex descriptor must have color 0 and 1.
+  // If color 1 is present but not color 0, it is used for lighting channel 0.
+  const bool use_color_1 =
+      (uid_data->components & (VB_HAS_COL0 | VB_HAS_COL1)) == (VB_HAS_COL0 | VB_HAS_COL1);
+  for (u32 color = 0; color < NUM_XF_COLOR_CHANNELS; color++)
+  {
+    if ((color == 0 || use_color_1) && uid_data->components & (VB_HAS_COL0 << color))
```
lioncash, line 204:
```suggestion
    if ((color == 0 || use_color_1) && (uid_data->components & (VB_HAS_COL0 << color)) != 0)
```

to be explicit about the comparison

Source/Core/VideoBackends/Software/SWVertexLoader.cpp
```diff
@@ -184,6 +182,39 @@ static void ReadVertexAttribute(T* dst, DataReader src, const AttributeFormat& f
   }
 }
 
+static void ParseColorAttributes(InputVertexData* dst, DataReader& src,
+                                 const PortableVertexDeclaration& vdec)
+{
+  auto SetDefaultColor = [](u8* color, int i) {
```
leoetlino, line 188:
```suggestion
  const auto set_default_color = [](u8* color, int i) {
```
maybe there should be an exception for lambdas that are used to structure code, but for now this should be using snake_case for consistency with the rest of the codebase

Source/Core/VideoBackends/Software/SWVertexLoader.cpp
```diff
@@ -184,6 +182,39 @@ static void ReadVertexAttribute(T* dst, DataReader src, const AttributeFormat& f
...
+    // The default alpha channel seems to depend on the number of components in the vertex format.
+    const auto& g0 = g_main_cp_state.vtx_attr[g_main_cp_state.last_id].g0;
+    u32 color_elements = i == 0 ? g0.Color0Elements : g0.Color1Elements;
```
leoetlino, line 191:
```suggestion
    const u32 color_elements = i == 0 ? g0.Color0Elements : g0.Color1Elements;
```

Pokechu22:
@dolphin-emu-bot rebuild

Pokechu22:
@dolphin-emu-bot rebuild

PR #9124, opened 2020-10-01
Qt: Use passthrough for hidpi rounding policy

stenzek:
Currently, Dolphin will round up the scaling factor when using a non-integer factor on Windows (e.g. 125%, 150%, 175%). This results in everything being way too large compared to the rest of the applications and desktop.

This patch uses setHighDpiScaleFactorRoundingPolicy() to set the rounding policy to passthrough, which will use the fractional scale. Needs testing on Linux, we might have to #ifdef it out.

Also changes the ImGui render function to use linear sampling rather than point sampling, which looks terrible at non-integer scales. We should really replace the bitmap font at some point, but apparently the ImGui font renderer has issues...

### DIFF
--- a/Source/Core/DolphinQt/Main.cpp
+++ b/Source/Core/DolphinQt/Main.cpp
@@ -128,6 +128,11 @@ int main(int argc, char* argv[])
 
   QCoreApplication::setAttribute(Qt::AA_EnableHighDpiScaling);
   QCoreApplication::setAttribute(Qt::AA_UseHighDpiPixmaps);
+#if QT_VERSION >= QT_VERSION_CHECK(5, 14, 0)
+  QGuiApplication::setHighDpiScaleFactorRoundingPolicy(
+      Qt::HighDpiScaleFactorRoundingPolicy::PassThrough);
+#endif
+
   QCoreApplication::setOrganizationName(QStringLiteral("Dolphin Emulator"));
   QCoreApplication::setOrganizationDomain(QStringLiteral("dolphin-emu.org"));
   QCoreApplication::setApplicationName(QStringLiteral("dolphin-emu"));
--- a/Source/Core/VideoCommon/RenderBase.cpp
+++ b/Source/Core/VideoCommon/RenderBase.cpp
@@ -1053,7 +1053,12 @@ void Renderer::DrawImGui()
 
   // Set up common state for drawing.
   SetPipeline(m_imgui_pipeline.get());
-  SetSamplerState(0, RenderState::GetPointSamplerState());
+
+  // Use linear sampling for fractional scaling, point for integer.
+  const float scale = ImGui::GetIO().DisplayFramebufferScale.x;
+  const bool is_integer_scale = (scale == std::floor(scale));
+  SetSamplerState(0, is_integer_scale ? RenderState::GetPointSamplerState() :
+                                        RenderState::GetLinearSamplerState());
   g_vertex_manager->UploadUtilityUniforms(&ubo, sizeof(ubo));
 
   for (int i = 0; i < draw_data->CmdListsCount; i++)


JosJuice:
This PR makes Dolphin look too large for me. The text is still the correct size, but at least the icons are larger. I'm using 125% DPI scaling on Windows 10.

stenzek:
Interesting, thanks for checking 125%. It might've been rounding down to 100%, whereas 150% was rounding up. Still strange that the text is unchanged but the icons are scaled, will look into it when I get a chance.

stenzek:
Yeah, from what I can tell it seems that things like spacing, icons, etc use 100% scale for 125% prior to this PR, but are scaled to 125% with these changes. A good example is the game list, look at the spacing between titles, current master with 125% has less padding than 100%, relatively speaking, and looks more squished.

It also explains why everything seemed to "large" to me at 150% - the spacing/margins were larger than they should be.

We also get the added bonus of enabling per-monitor DPI awareness. Which means it handles DPI changes _somewhat_ well (some things e.g. checkboxes get stuck, but for the most part it works). Can blame Qt for that one.

I guess it's up for debate as to what looks better. Personally I prefer it with this change, as it looks much closer to what it would at 100%. But I'm happy to be outnumbered here :)

JosJuice:
Ah, so you mean that the icon sizes that I thought were correct in master are actually smaller than they should be? Now that I try running master with DPI scaling handled by the system instead of the program (the mode where everything is blurry), it does look more like this PR than how master normally looks for me... So I suppose this PR is working correctly for me after all.

stenzek:
Seems like it from what I can tell, but I'm hardly an expert on the too-many-DPI-scaling-options available in Windows :)

Also for what it's worth, the old setting is still in the manifest. I considered changing (or removing) that too, since it's redundant and Qt will override it on startup, but if for some reason someone is still building on Windows with Qt <5.14, it would break scaling.

JosJuice:
Certain lines around certain boxes in the UI seem to be missing. But I don't know if there's anything good we can do about that in practice...

image

The icons for GameCube games also look pretty bad now. While I certainly wouldn't expect them to look great when upscaled by 25%, it looks as if they're being upscaled with nearest neighbor, which is a pretty bad choice for a non-integer ratio like this. Not the fault of this PR, but still.

stenzek:
Ah yeah, I forgot about the borders on some widgets.. I had that in DuckStation too, my "fix" was to use a custom theme.. :/ Will have to dig around and see if it's a known issue for Qt anywhere.

As for the icons, we might be able to use a resize filter?

shuffle2:
>if for some reason someone is still building on Windows with Qt <5.14, it would break scaling.

I think we should disregard such cases. Just consider Qt version on windows pinned to whatever is in repo.

Morph1984:
I had tried to solve this issue in yuzu where our application is running at 100% scale no matter the scaling factor set in the OS. Passthrough does destroy the resolution of game icons unfortunately, and Qt uses some sort of scaling method that's worse than nearest neighbor itself for icons and other UI elements.

There isn't really any good solution for this at all I fear, since any method I've tried either results in bilinear filtering by windows (OS scaling) or crap scaling from Qt resulting in artifacts in icons and UI elements. Only integer scales (200%/300% etc) manage to have these icons render properly.

I had also looked into what other projects such as RPCS3 did to solve this issue. However, it seems that the issues are universal and we have to settle on some kind of compromise.

shuffle2:
Will this still be a problem with Qt6? They make it sound like it "just works" ? https://doc.qt.io/qt-6/portingguide.html#high-dpi

stenzek:
From my very brief testing, the border issue appeared to still be there in qt6.

shuffle2:
Is dolphin doing something strange and/or wrong? What is the complication which makes dolphin need special handling?

stenzek:
Nope, it seems to be a qt bug when running at non integer scales. I have the same issue in duckstation. My fix was to just use a custom theme 👎

PR #9126, opened 2020-10-02
Externals: Update MoltenVK to v1.1

stenzek:
Needs testing as my mac environment is not usable at the moment.

### DIFF
--- a/Externals/MoltenVK/version.txt
+++ b/Externals/MoltenVK/version.txt
@@ -1,3 +1,1 @@
-MoltenVK/macOS/dynamic/libMoltenVK.dylib from vulkansdk-macos-1.2.131.2, renamed to libvulkan.dylib
-
-Downloaded from: https://sdk.lunarg.com/sdk/download/1.2.131.2/mac/vulkansdk-macos-1.2.131.2.tar.gz
+MoltenVK from https://github.com/KhronosGroup/MoltenVK, commit b9b78def172074872bfbb1015ccf75eeec554ae2


spotlightishere:
Is it possible that this could be built to target both x86_64 and arm64? Upstream supports both configurations as of v1.1, and it might be beneficial within one PR now.

Otherwise - works as it should under Big Sur!

stenzek:
Thanks for checking.

I don't see why not, but the main dolphin application would have to support ARM too. And that would require Qt from brew to support it, and I'm not sure if that does yet?

spotlightishere:
Sadly not, quite yet -  you can apply patches listed in [their upstream tracking bug](https://bugreports.qt.io/browse/QTBUG-85279) to have a single architecture build function and then `lipo` together x86_64 normal builds, but that is fairly manual at the moment. Dolphin itself works with some edits which I haven't quite yet finalized to draft a PR.

Mostly I figured I'd mention here so another update doesn't need to take place later down the line :)

stenzek:
Will keep it in mind, but to keep the download sizes smaller I'll probably just add a separate dylib later on rather than going for the merged library now. Unless of course dolphin itself is a fat binary, not sure what we're going to do there yet.

gilcel:
Tested with some GC & Wii-games and it works great on macOS Mojave 10.14.6 (Mac mini Late 2012 with Intel HD Graphics 4000, not the fastest setup). thanks!

pizuz:
I'm seeing a nice performance boost, here. Also GPU texture decoding, disabled in #8743 seems to be unbroken by the new release, from what I can tell.

KevinVitale:
Compiled Dolphin for the first time to try out Vulkan on my macOS. Seems stable overall, with some nice performance boost. Some specific performance dips on NHL HItz 20-03 are noticeable; mostly 60fps solid, but when players have the _"smoke"_ effect, performance drops by nearly half (with lots of audio stutter).

Playing on:
2017 MacBook Pro
10.15.7 (19H2)
3.1 GHz Quad-Core Intel Core i7
16GB 2133 MHz LPDDR3
Radeon Pro 560 4 GB

<hr/>

Screen Shot 2020-10-04 at 4 31 16 PM

JMC47:
Smoke effects result in a lot of blending, which is probably hard on MoltenVK.  It's normal for those to cause slowdown in non-optimal setups or weaker computers.

Rumi-Larry:
> I'm seeing a nice performance boost, here. Besides, GPU texture decoding, disabled in #8743 seems to be unbroken by the new release, from what I can tell. No more crashes, black textures or stuff loading on the second launch only in the games I tested (the screenshot was taken from the last build before #8743 with the dylib replaced, the vanilla build crashes as expected).

In that case, that change can be reverted can't it?

PR #9128, opened 2020-10-03
IOS/ES: Add Wii System Transfer to IOS62 SetUID whitelist

InvoxiPlayGames:
The Wii System Transfer tool (used on the vWii) requires SetUID, which the vWii's IOS62 has an exception to allow. This PR fixes the "An error has occurred" screen that you would get when launching the Wii System Transfer tool on Dolphin.

### DIFF
--- a/Source/Core/Core/IOS/ES/ES.cpp
+++ b/Source/Core/Core/IOS/ES/ES.cpp
@@ -182,7 +182,7 @@ static ReturnCode CheckIsAllowedToSetUID(Kernel& kernel, const u32 caller_uid,
   if (kernel.GetVersion() == 62)
   {
     const bool is_wiiu_transfer_tool =
-        active_tmd.IsValid() && (active_tmd.GetTitleId() | 0xFF) == 0x00010001'484353ff;
+        active_tmd.IsValid() && ((active_tmd.GetTitleId() | 0xFF) == 0x00010001'484353ff || (active_tmd.GetTitleId() | 0xFF) == 0x00010001'484354ff);
     if (is_wiiu_transfer_tool)
       return IPC_SUCCESS;
   }


leoetlino:
This is probably a harmless change but Wii-side IOS62 only has an exception for `HCS*` and I'm not sure if we want to implement vWii IOS changes. Written as is, the check matches neither Wii IOS62 nor vWii IOS62.

vWii IOS62:

```c
    if ( !is_system_menu
      && (*(_DWORD *)(dword_201126A4 + 396) != 0x10001 || (*(_DWORD *)(dword_201126A4 + 400) & 0xFFFFFF00) != 0x48435400)
      && (*(_DWORD *)(dword_201126A4 + 396) != 0x10008 || (*(_DWORD *)(dword_201126A4 + 400) & 0xFFFFFF00) != 0x48435A00) )
    {
      ret = -1017;
    }
```

leoetlino:
This should probably be locked behind a vWii setting so that Dolphin doesn't end up emulating an inaccurate mix of Wii and vWii IOS behavior.

To add such a setting:

* Add a `extern const Info<bool> MAIN_VWII_MODE;` to https://github.com/dolphin-emu/dolphin/blob/master/Source/Core/Core/Config/MainSettings.h
* Define it in MainSettings.cpp, with the default value set to false
* Add it to the "is setting saveable" list here: https://github.com/dolphin-emu/dolphin/blob/f265c412b449f7054468fae6e13a890bb0d3c098/Source/Core/Core/ConfigLoaders/IsSettingSaveable.cpp#L50
* Use Config::Get to check whether the Wii or the vWii version of the check should be used.

The title ID check should also be fixed to match the snippet I've posted.

leoetlino:
This PR cannot be merged in the current state and seems abandoned. Feel free to open another PR once you have made the requested changes.

PR #9130, opened 2020-10-04
Vulkan: fix validation error in bSupportsGeometryShaders=false case

flto:
In CreateDescriptorSetLayouts(), one less dynamic binding is created when
bSupportsGeometryShaders=false. Reduce the dynamicOffsetCount argument by
one in that case. Avoids this validation error:

Attempting to bind 3 descriptorSets with 2 dynamic descriptors, but
dynamicOffsetCount is 3. It should exactly match the number of dynamic
descriptors. The Vulkan spec states: dynamicOffsetCount must be equal to
the total number of dynamic descriptors in pDescriptorSets


stenzek (approved):
Nice catch, thanks!

stenzek:
@dolphin-emu-bot rebuild

PR #9132, opened 2020-10-05
Remove geckocodes.org downloading because the site shut down

JosJuice:


Miksel12:
I reckon `#include "Common/HttpRequest.h"` should also be removed.

iwubcode:
Do we have an alternative in mind?  I only used it a bit but this was really nice to quickly pull up codes.  Maybe it's time to add resource packs that contain a list of searchable codes...

cobalt2727:
http://web.archive.org/web/20200824101754/https://geckocodes.org
How feasible would it be to host a mirror on Dolphin's website?

SirMangler:
Geckocodes is now directing people to https://gamehacking.org/ which still hosts a lot of codes, both AR and Gecko. Perhaps that would be a decent alternative. It appears to be missing codes on various games, but some more popular games still have more than enough codes to make it worthwhile. As a plus the site also appears to offer pages for individual game versions, which I believe geckocodes didn't do.

Here's the page for Metroid Prime as an example: https://gamehacking.org/game/54210

cobalt2727:
https://codes.rc24.xyz/
Got something you guys might be interested in - the RiiConnect24 team has apparently mirrored the entire geckocodes.org site.

JosJuice:
@cobalt2727 Opened as PR #9133. (I was actually contacted about this by someone who does RC24 stuff at around the same time you posted this, as it happens.) It seems to work perfectly with just a simple URL replace.

jordan-woyak:
Closing in favor of #9133.

PR #9134, opened 2020-10-06
Shaders: Remove AsciiArt shader

SpaceXCheeseWheel:
This shader has been rendered nonfunctional due to the removal of the function SampleFontLocation, which was never implemented in VideoCommon. The last reference to SampleFontLocation was located in the OGL video backend. It was removed in this commit: https://github.com/dolphin-emu/dolphin/commit/f039149198657c1891e1c6462ed30c31ed4b8486, where many functions were moved to video common, among other things.

This shader should be removed so it does not confuse users with the error that it causes. Alternatively, it could be rewritten so it is compatible with modern versions of Dolphin. I am not familiar with GLSL, so this is currently beyond my skill set.


leoetlino:
Since it is currently totally broken, there's not much point to keeping it around. We can always add it back later when/if somebody fixes it or reimplements the missing shader utility function.

degasus:
If someone comes up to this PR and wants to reimplement this shader: Having an ASCII art shader would be nice, but DO NOT USE the same method as this old shader. It was fun to write, but the outcome was not really worth to ship to anyone. Sorry for bothering so many people....

PR #9135, opened 2020-10-06
Show NKitness in file format string

JosJuice:
To make people more aware that they're not using a normal disc image.

### DIFF
--- a/Source/Core/UICommon/GameFile.cpp
+++ b/Source/Core/UICommon/GameFile.cpp
@@ -31,6 +31,7 @@
 #include "Common/HttpRequest.h"
 #include "Common/Image.h"
 #include "Common/IniFile.h"
+#include "Common/MsgHandler.h"
 #include "Common/NandPaths.h"
 #include "Common/StringUtil.h"
 #include "Common/Swap.h"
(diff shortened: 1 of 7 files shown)


Source/Core/UICommon/GameFile.cpp
```diff
@@ -667,8 +671,14 @@ std::string GameFile::GetFileFormatName() const
     // substr removes the dot
     return extension.substr(std::min<size_t>(1, extension.size()));
   }
+
   default:
-    return DiscIO::GetName(m_blob_type, true);
+  {
+    std::string name = DiscIO::GetName(m_blob_type, true);
+    if (m_is_nkit)
+      name = StringFromFormat(Common::GetStringT("%s (NKit)").c_str(), name.c_str());
```
BhaaLseN, line 679: This feels weird, taking `std::string` to `.c_str()` them, so we can `%s` a `.c_str()` parameter that is later converted back into  a `std::string`. Don't we have anything else that doesn't requires this back and forth? Maybe some `fmt` stuff?

JosJuice: I have made it use `fmt` now. I was under the impression that this part of the code hadn't started using `fmt` yet, but that was actually incorrect.

Note that I am using a positional argument (`{0}` instead of `{}`). While it doesn't really matter for this string, positional arguments are an important feature to have for translatable strings with two or more arguments, and I figured it would be best to consistent and use it in all translatable strings to make things as easy as possible for the translators.

BhaaLseN (approved):
Code looks good, untested

leoetlino (approved):

PR #9136, opened 2020-10-07
VolumeVerifier: Make "no valid data in partition" hide some other errors

JosJuice:
Not much use in knowing that the H3 hashes are wrong when the partition consists of complete garbage.


leoetlino (approved):
sgtm

PR #9137, opened 2020-10-07
DiscIO: Decrease RAM usage during zstd compression

JosJuice:
By calling `ZSTD_CCtx_setPledgedSrcSize`, we can let zstd know how large a chunk is going to be before which start compressing it, which lets zstd avoid allocating more memory than needed for various internal buffers. This greatly reduces the RAM usage when using a high compression level with a small chunk size, and doesn't have much of an effect in other circumstances.

A side effect of calling `ZSTD_CCtx_setPledgedSrcSize` is that zstd by default will write the uncompressed size into the compressed data stream as metadata. In order to save space, and since the decompressed size can be figured out through the structure of the RVZ format anyway, we disable writing the uncompressed size by setting `ZSTD_c_contentSizeFlag` to 0.

### DIFF
--- a/Source/Core/DiscIO/WIABlob.cpp
+++ b/Source/Core/DiscIO/WIABlob.cpp
@@ -1040,7 +1040,7 @@ std::optional<std::vector<u8>> WIARVZFileReader<RVZ>::Compress(Compressor* compr
 {
   if (compressor)
   {
-    if (!compressor->Start() || !compressor->Compress(data, size) || !compressor->End())
+    if (!compressor->Start(size) || !compressor->Compress(data, size) || !compressor->End())
       return std::nullopt;
 
     data = compressor->GetData();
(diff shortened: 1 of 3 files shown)


Source/Core/DiscIO/WIABlob.cpp
```diff
@@ -1564,7 +1564,7 @@ WIARVZFileReader<RVZ>::ProcessAndCompress(CompressThreadState* state, CompressPa
 
     if (state->compressor)
     {
-      if (!state->compressor->Start())
+      if (!state->compressor->Start(entry.exception_lists.size() + entry.main_data.size()))
```
leoetlino, line 1567: I had to check that exceptions_list indeed stored bytes rather than exception structs (as the name would imply IMO). That's probably just because I'm unfamiliar with this code, though :)

Still, it might be worth renaming this member to exception_list_data for more clarity and for consistency with main_data in a followup PR?

leoetlino (approved):

PR #9139, opened 2020-10-07
Android: add reverse landscape to screen orientation settings

nickbeth00:
Until now, when screen orientation is set to Landscape, the emulation activity will be fixed on a 90 degrees rotated position. 
This pull request aims at allowing 90 and 270 degrees landscape rotations, based on sensor data (independently of user android settings).

Controllers like the Gamesir X2, which have a USB Type-C port on the left will force the phone to a 270 degrees rotated state, so this is useful for those users (like me).

### DIFF
--- a/Source/Android/app/src/main/res/values/arrays.xml
+++ b/Source/Android/app/src/main/res/values/arrays.xml
@@ -435,11 +435,13 @@
 
     <string-array name="orientationEntries">
         <item>Landscape</item>
+        <item>Landscape (reverse)</item>
         <item>Portrait</item>
         <item>Auto</item>
     </string-array>
     <integer-array name="orientationValues">
         <item>0</item>
+        <item>8</item>
         <item>1</item>
         <item>-1</item>
     </integer-array>


JosJuice:
I specifically made it default to a mode that does not take the sensor data into account in order to avoid accidentally triggering orientation changes when using motion controls. I would not mind adding `SCREEN_ORIENTATION_SENSOR_LANDSCAPE` as an additional option, but I feel strongly about keeping `SCREEN_ORIENTATION_LANDSCAPE` as an option. I'm less certain about which one of the two fits best as the default.

nickbeth00:
Uhm yeah I definitely missed that. Another thing we can do is add SCREEN_ORIENTATION_REVERSE_LANDSCAPE as an option and keep LANDSCAPE as default, so in both cases they don't interfere with motion controls.

JosJuice:
That sounds good to me.

nickbeth00:
Ok what I haven't considered now is if motion controls are expecting a reverse landscape setting. I will have to test that.

JosJuice:
@dolphin-emu-bot rebuild

JosJuice:
Motion controls should be working correctly regardless of orientation, but feel free to check to make sure.

nickbeth00:
Tested all combinations, working as expected.

JosJuice (approved):
LGTM if you squash the commits into one and update the pull request title.

nickbeth00:
Should be good to go.

JosJuice:
@dolphin-emu-bot rebuild

Ebola16 (approved):

PR #9140, opened 2020-10-07
Revert "VideoBackends: Disable GPU Texture Decoding under MoltenVK"

Miksel12:
Reverts dolphin-emu/dolphin#8743
GPU Texture Decoding seems to be fixed with the latest MoltenVK update: https://github.com/dolphin-emu/dolphin/pull/9126#issuecomment-703267843 though  some more test cases would be nice to have.


Techjar:
Let's not do a straight revert. This PR also had some cleanup in it.

PR #9141, opened 2020-10-08
Re-enable GPU Texture Decoding under MoltenVK

Techjar:
Apparently working again with the MoltenVK update in #9126. This supersedes #9140 by not undoing the moving of the closing bracket. @Stenzek should we remove the DriverDetails bug itself as well, or keep it around in case it's needed again?

Would like to see some test cases just to be sure it's working properly.

### DIFF
--- a/Source/Core/VideoBackends/Vulkan/VulkanContext.cpp
+++ b/Source/Core/VideoBackends/Vulkan/VulkanContext.cpp
@@ -346,10 +346,6 @@ void VulkanContext::PopulateBackendInfoFeatures(VideoConfig* config, VkPhysicalD
   // with depth clamping. Fall back to inverted depth range for these.
   if (DriverDetails::HasBug(DriverDetails::BUG_BROKEN_REVERSED_DEPTH_RANGE))
     config->backend_info.bSupportsReversedDepthRange = false;
-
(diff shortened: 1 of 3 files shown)


MayImilae:
Tested on my macbook. GPU Texture Decoding works, and this PR reveals it as expected. And MoltenVK is still glitchy (not GPU texture decoding's fault). Everything is as expected!

**Game Tested**
Sonic Colors (NTSC)

**System Specs**
MacBook Pro 13in mid-2018 with Touch Bar
macOS 10.15.6
Core i7-8559U @ 4.5ghz
Intel Iris Plus Graphics 655
16GB 2133MHz LPDDR3

stenzek:
Probably just drop the option. I think the only other place it would be relevant would be drivers without sufficient texel buffer space, and if I remember correctly we test the size there and disable it accordingly anyway.

pizuz:
Working without problems on all the games I tested so far (Wind Waker, Twilight Princess, Mario Kart Double Dash, Paper Mario TTYD and Super Mario Galaxy).

System: Late 2012 iMac with nVidia GeForce GT 750M

MayImilae:
Can we get this one merged? It's already tested and this isn't a very big change at all.

faroukabd000:
Run in android no problem happened yet
Mi a3

PR #9142, opened 2020-10-10
WiimoteEmu: Expose IR camera FOV to adjust IMU pointing sensitivity. 

jordan-woyak:
This exposes the previously hardcoded field of view of the camera.

I've located the settings under the Point group of the Motion Input tab as the most common use case will be to adjust the sensitivity of gyroscope-based pointing.

image

### DIFF
--- a/Source/Core/Core/HW/WiimoteEmu/Camera.cpp
+++ b/Source/Core/Core/HW/WiimoteEmu/Camera.cpp
@@ -53,7 +53,7 @@ int CameraLogic::BusWrite(u8 slave_addr, u8 addr, int count, const u8* data_in)
   return RawWrite(&m_reg_data, addr, count, data_in);
 }
 
-void CameraLogic::Update(const Common::Matrix44& transform)
+void CameraLogic::Update(const Common::Matrix44& transform, Common::Vec2 field_of_view)
 {
   // IR data is read from offset 0x37 on real hardware.
   auto& data = m_reg_data.camera_data;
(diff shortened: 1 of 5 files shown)


Source/Core/Core/HW/WiimoteEmu/Camera.cpp
```diff
@@ -53,7 +53,7 @@ int CameraLogic::BusWrite(u8 slave_addr, u8 addr, int count, const u8* data_in)
   return RawWrite(&m_reg_data, addr, count, data_in);
 }
 
-void CameraLogic::Update(const Common::Matrix44& transform)
+void CameraLogic::Update(const Common::Matrix44& transform, Common::Vec2 field_of_view)
```
iwubcode, line 56: Just curious, is there a reason this isn't `const&`?

jordan-woyak: Not really a reason. It's just two floats so passing by-value is maybe possibly more efficient than referencing them, but I can change it.

Source/Core/Core/HW/WiimoteEmu/Camera.cpp
```diff
@@ -88,9 +88,9 @@ void CameraLogic::Update(const Common::Matrix44& transform)
       Vec3{SENSOR_BAR_LED_SEPARATION / 2, 0, 0},
   };
 
-  const auto camera_view = Matrix44::Perspective(CAMERA_FOV_Y, CAMERA_AR, 0.001f, 1000) *
-                           Matrix44::FromMatrix33(Matrix33::RotateX(float(MathUtil::TAU / 4))) *
-                           transform;
+  const auto camera_view =
+      Matrix44::Perspective(field_of_view.x, field_of_view.x / field_of_view.y, 0.001f, 1000) *
```
leoetlino, line 92: I've probably missed something, but shouldn't the first argument be fov.y?

jordan-woyak: Yeah.. probably..Thanks. :P

leoetlino (approved):

PR #9143, opened 2020-10-11
Android: Remove finish from ConfirmRunnableViewHolder

Ebola16:
I'm not 100% sure of the implications of this change but it appears to fix the TODO in `ConfirmRunnableViewHolder`.


JosJuice:
I took a look at the problem, and I think this less intrusive solution would work: #9144

Ebola16:
Closed in favor of #9144

PR #9145, opened 2020-10-11
AudioStretcher: Keep default parameters.

jordan-woyak:
A user said Citra's sound stretching sounding much better than ours.

Looking at their code about the only difference was some explicit settings in Dolphin.
https://github.com/dolphin-emu/dolphin/blob/master/Source/Core/AudioCommon/AudioStretcher.cpp
https://github.com/citra-emu/citra/blob/master/src/audio_core/time_stretch.cpp

It looks like the default "sequence" and "seek window" adjust automatically with the tempo, maybe?

> By default, this setting value is calculated automatically according to tempo value.

https://www.surina.net/soundtouch/README.html

The user said this change improved audio stretching.
I don't know if I can tell a difference one way or the other.

Please test.

### DIFF
--- a/Source/Core/AudioCommon/AudioStretcher.cpp
+++ b/Source/Core/AudioCommon/AudioStretcher.cpp
@@ -18,10 +18,6 @@ AudioStretcher::AudioStretcher(unsigned int sample_rate) : m_sample_rate(sample_
   m_sound_touch.setSampleRate(sample_rate);
   m_sound_touch.setPitch(1.0);
   m_sound_touch.setTempo(1.0);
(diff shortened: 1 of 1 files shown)


JMC47:
Citra's audio stretching does sound a lot better than Dolphin's, if I'm being honest.  I'll test this when I get a chance.

MerryMage:
sgtm

Adamillo:
> Citra's audio stretching does sound a lot better than Dolphin's, if I'm being honest. I'll test this when I get a chance.
Did you test it today?

PR #9146, opened 2020-10-11
Android: Allow disabling cover downloading

JosJuice:
In case someone wants to be very careful with how much bandwidth they use or with what data GameTDB.com collects on you. This is already an option in DolphinQt (though in DolphinQt it will switch entirely from using covers to banners when turned off).

### DIFF
--- a/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/adapters/GameAdapter.java
+++ b/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/adapters/GameAdapter.java
@@ -121,6 +121,14 @@ public final class GameAdapter extends RecyclerView.Adapter<GameViewHolder> impl
   }
 
   /**
+   * Re-fetches game metadata from the game file cache.
+   */
+  public void refetchMetadata()
+  {
+    notifyItemRangeChanged(0, getItemCount());
+  }
+
+  /**
    * Launches the game that was clicked on.
    *
    * @param view The card representing the game the user wants to play.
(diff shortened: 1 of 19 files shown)


Ebola16 (approved):
There's probably no need to split off the first commit since this PR seems pretty straightforward.

Ebola16:
At some point `InputOverlay` will need access to `Settings` so settings can be saved there instead of using `SharedPreferences`. Do you think that should be accomplished by using the method you removed from `EmulationActivity` in this PR or something more global?

JosJuice:
I believe it would make most sense to re-add the method I removed from EmulationActivity. (Unless we put the Settings object directly in InputOverlay?)

Ebola16:
I think I'll go with the method removed in this PR then. I just wanted to make sure I wouldn't conflict with your planned changes.
One other question, does ` mSettings.close` need to be called in `onDestroy` if used in `InputOverlay`? I wouldn't expect `mSettings` to survive after `InputOverlay`'s destruction but I might be misunderstanding something.

Ebola16:
Oh `mSettings.close` is called in `EmulationActivity`, never mind.

Ebola16 (changes requested):
While working on the `SharedPreferences` to INI PRs, I found that `SharedPreferences` are set in `EmulationActivity.toggleControls`. When converted to INI settings, I'll need access to `Settings` to set those values. Please revert the change that removes `mSettings` in `EmulationActivity`.

JosJuice:
Done.

Ebola16 (approved):
Reminder to self to update #8894 after this PR is merged.

JosJuice:
I'll need to update PR #9148 too. (Or this PR, if that PR gets merged first.)

leoetlino (approved):

PR #9147, opened 2020-10-12
WiimoteEmu: Expose IMU pointing accelerometer weight setting.

jordan-woyak:
Reducing this value can produce a less jittery pointer with reduced drift correction.

A user reported they had better results with 1% vs the default 2%.

The best name for the setting that I could come up with is "Accelerometer Influence".
I think it's descriptive and accurate enough. Anything else I can think if is ambiguous or too long.

### DIFF
--- a/Source/Core/InputCommon/ControllerEmu/ControlGroup/IMUCursor.cpp
+++ b/Source/Core/InputCommon/ControllerEmu/ControlGroup/IMUCursor.cpp
@@ -38,11 +38,26 @@ IMUCursor::IMUCursor(std::string name_, std::string ui_name_)
               // i18n: Refers to emulated wii remote movements.
               _trans("Clamping of rotation about the yaw axis.")},
              25, 0, 360);
+
+  AddSetting(&m_accel_weight_setting,
+             {// i18n: Percentage value of accelerometer data (complementary filter coefficient).
+              _trans("Accelerometer Influence"),
+              // i18n: The symbol/abbreviation for percent.
+              _trans("%"),
+              // i18n: Refers to a setting controling the influence of accelerometer data.
+              _trans("Influence of accelerometer data on pitch and roll. Higher values will reduce "
+                     "drift at the cost of noise. Consider values between 1% and 3%.")},
+             2, 0, 100);
 }
 
 ControlState IMUCursor::GetTotalYaw() const
 {
   return m_yaw_setting.GetValue() * MathUtil::TAU / 360;
+}
+
+ControlState IMUCursor::GetAccelWeight() const
+{
+  return m_accel_weight_setting.GetValue() / 100;
 }
 
 }  // namespace ControllerEmu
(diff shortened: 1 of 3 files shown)


iwubcode:
Just based off your comment, I'm inclined to suggest "Accelerometer drift correction" or something to that effect.

And then maybe instead of a value, have a combo box with "high", "medium", "low" with some preset values.  Otherwise I imagine users are going to crank it up to 100%.

Source/Core/InputCommon/ControllerEmu/ControlGroup/IMUCursor.cpp
```diff
@@ -39,11 +39,25 @@ IMUCursor::IMUCursor(std::string name_, std::string ui_name_)
...
               _trans("Total rotation about the yaw axis.")},
              25, 0, 360);
+
+  AddSetting(&m_accel_weight_setting,
+             // i18n: Refers to a setting controling the influence of accelerometer data.
+             {_trans("Accelerometer Correction"),
+              // i18n: The symbol/abbreviation for percent.
+              _trans("%"),
+              // i18n: Refers to a setting controling the influence of accelerometer data.
+              _trans("Influence of accelerometer data on pitch and roll.")},
```
Rumi-Larry, line 49: Consider instead "Accelerometer induced pitch-and-roll bias"

rlnilsen:
This setting is the ratio of how much trust you put in accelerometer vs gyroscope data (at least in my original code, I don't know if the new quaternion stuff is exactly equivalent). The gyroscope is accurate but drifty, the accelerometer is inaccurate but does not drift. So we mostly trust the gyroscope (98% by default), but use the accelerometer to correct gyroscope drift. So maybe "Accelerometer to gyroscope trust ratio"? I'm bad with names.

> ... I'm inclined to suggest "Accelerometer drift correction" or something to that effect.

Unlike gyroscopes, accelerometers do not have drift.

rlnilsen (changes requested):
The code ~~is untested by me but~~ seems very straight forward and boiler plate, so once the wording issues are resolved this LGTM.

Edit: Gave it a spin and seems to be working well. The pointer gets pretty shaky already at 5%, so I don't know how useful it is to go above that though. This was with a DS4.

rlnilsen:
Perhaps just "Drift Correction" would be sufficient? That's really all one needs to draw the user's attention if they are experiencing drift. Then the tooltip can give a deeper explanation.

JMC47:
I like the idea of drift correction for the user facing option name.  If that's the only thing blocking this, we should probably change this and merge it.

jordan-woyak:
My issue with "Drift Correction" is it doesn't convey that accelerometer data is used.
In the future we might also have correction from IR data or whatever.
I'm leaning towards maybe "Accelerometer Coefficient" or "Accelerometer Weight".

AdmiralCurtiss:
Do you want to put this as an Advanced setting now that we have that?

jordan-woyak:
> Do you want to put this as an Advanced setting now that we have that?

I'm thinking not. It's already on an "Advanced" tab.

AdmiralCurtiss:
Well alright then, this seems good otherwise, so...

PR #9148, opened 2020-10-13
Android: Fix setting read during play with local game layer active

JosJuice:
During emulation, when `LocalGame` has a value but `CurrentRun` doesn't, we want to read from `LocalGame`, not `CurrentRun`. This change exposes a `LAYER_ACTIVE` option that handles this correctly.


JosJuice:
To clarify what the bug is, here are reproduction steps:

1. Turn on Show FPS in the game-specific settings for some game.
2. Start that game.
3. Open the graphics settings. Observe that Show FPS shows up as disabled, even though the FPS actually is being displayed.

Ebola16 (approved):
I still need to get more familiar with the new config system to understand how changes can potentially go wrong but this makes sense to me.

leoetlino (approved):

PR #9150, opened 2020-10-14
DiscIO: Make use of fmt where applicable

JosJuice:
Once nice benefit of fmt is that we can use positional arguments in localizable strings. This a feature which has been requested for the Korean translation of strings like "Errors were found in %zu blocks in the %s partition." and which will no doubt be useful for other languages too.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -308,6 +308,14 @@ if(CMAKE_SYSTEM_NAME MATCHES "Darwin")
   # Drop unreachable code and data.
   set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -Wl,-dead_strip,-dead_strip_dylibs")
 
+  # Set -fno-objc-exceptions, for consistency with -fno-exceptions earlier.
+  # If we set only -fno-exceptions, fmt fails to compile when included from
+  # Objective-C++ because fmt tries try to use throw because __EXCEPTIONS is defined.
+  #
+  # TODO: Only enable this for Objective-C(++).
+  # We get warnings if we enable this when building regular C(++) code.
+  check_and_add_flag(NO_OBJC_EXCEPTIONS -fno-objc-exceptions)
+
   find_library(APPKIT_LIBRARY AppKit)
   find_library(APPSERV_LIBRARY ApplicationServices)
   find_library(CARBON_LIBRARY Carbon)
--- a/Source/Core/DiscIO/NANDImporter.cpp
+++ b/Source/Core/DiscIO/NANDImporter.cpp
@@ -9,12 +9,13 @@
 #include <cinttypes>
 #include <cstring>
 
+#include <fmt/format.h>
+
 #include "Common/Crypto/AES.h"
 #include "Common/File.h"
 #include "Common/FileUtil.h"
 #include "Common/Logging/Log.h"
 #include "Common/MsgHandler.h"
-#include "Common/StringUtil.h"
 #include "Common/Swap.h"
 #include "Core/IOS/ES/Formats.h"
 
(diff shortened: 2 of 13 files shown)


Source/Core/DiscIO/NANDImporter.cpp
```diff
@@ -132,9 +133,10 @@ std::string NANDImporter::GetPath(const NANDFSTEntry& entry, const std::string&
 
 std::string NANDImporter::FormatDebugString(const NANDFSTEntry& entry)
 {
-  return StringFromFormat("%12.12s 0x%02x 0x%02x 0x%04x 0x%04x 0x%08x 0x%04x 0x%04x 0x%04x 0x%08x",
-                          entry.name, entry.mode, entry.attr, entry.sub, entry.sib, entry.size,
-                          entry.x1, entry.uid, entry.gid, entry.x3);
+  return fmt::format(
+      "{:12.12} {:#04x} {:#04x} {:#06x} {:#06x} {:#010x} {:#06x} {:#06x} {:#06x} {:#010x}",
```
BhaaLseN, line 137: This change in width seems a bit weird, at least I can't see any indication in the docs that the width used for the alternate form _includes_ the prefix.

Then again, it probably doesn't matter too much since this is a debug string.

BhaaLseN (approved):
Code seems good, and you can probably ignore that debug string remark (untested)

Source/Core/DiscIO/NANDImporter.cpp
JosJuice: I've tested this code, and it works correctly if I write it like this.

For what it's worth, https://fmt.dev/latest/syntax.html contains the following example (but other than that I indeed can't find anything relevant in the documentation):
```
fmt::format("{:#04x}", 0);
// Result: "0x00"
```

Source/Core/DiscIO/VolumeVerifier.cpp
```diff
@@ -519,10 +520,8 @@ bool VolumeVerifier::CheckPartition(const Partition& partition)
   if (partition.offset % VolumeWii::BLOCK_TOTAL_SIZE != 0 ||
       m_volume.PartitionOffsetToRawOffset(0, partition) % VolumeWii::BLOCK_TOTAL_SIZE != 0)
   {
-    AddProblem(
-        Severity::Medium,
-        StringFromFormat(Common::GetStringT("The %s partition is not properly aligned.").c_str(),
-                         name.c_str()));
+    AddProblem(Severity::Medium,
+               fmt::format(Common::GetStringT("The {0} partition is not properly aligned."), name));
```
lioncash, line 524: Given this is used quite a bit, maybe this should be its own helper function in common, like:

```cpp
typename <typename... Args>
std::string GetStringTFmt(const char* string, Args&&... args)
{
    return fmt::format(GetStringT(string), std::forward<Args>(args)...);
}
```

So that `fmt::format` doesn't need to be typed all the time in conjunction with another function immediately one after another in several spots.

JosJuice: Added. I wasn't sure whether it would fit best in StringUtil.h (because it deals with string formatting) or MsgHandler.h (because it deals with translation), but I put it in StringUtil.h.

JosJuice:
Does anyone understand why the CMake build fails to find the fmt/format.h header now? Common does have fmt::fmt in its target_link_libraries. (I had the same problem before when I used fmt::format from DiscIO without adding fmt::fmt to DiscIO's target_link_libraries.)

BhaaLseN:
Those errors come from `AudioCommon`, maybe because it is a `template` and the compiler inlines it or something like that?

JosJuice:
```
(19:47:39) leoetlino: looks like AudioCommon does not have Common as a dependency
```

Fixing that seems to have made it work.

JosJuice:
Okay, I don't understand what's wrong with macOS now...

leoetlino:
The issue is that fmt checks whether `__EXCEPTIONS` is defined to determine whether C++ exceptions are enabled, but that macro is still defined when C++ exceptions are disabled and Obj-C exceptions are enabled. Have you tried compiling with `-fno-objc-exceptions`?

CMakeLists.txt
```diff
@@ -308,6 +308,8 @@ if(CMAKE_SYSTEM_NAME MATCHES "Darwin")
   # Drop unreachable code and data.
   set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -Wl,-dead_strip,-dead_strip_dylibs")
 
+  check_and_add_flag(NO_OBJC_EXCEPTIONS -fno-objc-exceptions)
```
lioncash, line 311: This change probably warrants a comment, so it doesn't accidentally get removed in the future or anything.

Source/Core/DiscIO/VolumeVerifier.cpp
JosJuice: I moved it to MsgHandler.h because I figured we will need to include fmt/format.h in that header anyway if we migrate panic alerts over to using fmt.

lioncash: Sounds good!

lioncash (approved):

PR #9089, opened 2020-09-16
Android: Move orientation setting to main settings screen

JosJuice:
I moved it from the main settings screen to the in-game menu in PR #8439 so that it could be changed while a game is running, but now that the main settings can be accessed while a game is running, there's no reason to not put it in the main settings.

https://bugs.dolphin-emu.org/issues/12067

PR #9152, opened 2020-10-16
Android: Convert SharedPreferences to INI Settings (simple cases)

Ebola16:
Part of https://bugs.dolphin-emu.org/issues/10957

This will allow some settings to be retained across Dolphin builds. Other PRs will handle more involved `SharedPreferences` changes.

I'm aware that globally getting settings will come out soon but I figured I should submit this for review before I proceed further.

Note: Android: wiiController is implemented in #8894 and orientation is implemented in #9089.


Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/activities/EmulationActivity.java
```diff
@@ -934,37 +926,31 @@ private void showMotionControlsOptions()
...
 
     builder.show();
   }
 
   private void chooseOrientation()
   {
     final int[] orientationValues = getResources().getIntArray(R.array.orientationValues);
-    int initialChoice = mPreferences.getInt("emulationActivityOrientation",
-            ActivityInfo.SCREEN_ORIENTATION_LANDSCAPE);
+    int initialChoice = IntSetting.MAIN_EMULATION_ACTIVITY_ORIENTATION.getInt(mSettings);
```
JosJuice, line 937: This one has already been migrated in PR #9089.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/fragments/EmulationFragment.java
```diff
@@ -124,20 +120,11 @@ public void onDetach()
     super.onDetach();
   }
 
-  public void toggleInputOverlayVisibility()
+  public void toggleInputOverlayVisibility(Settings settings)
   {
-    SharedPreferences.Editor editor = mPreferences.edit();
-
     // If the overlay is currently set to INVISIBLE
```
JosJuice, line 125: This comment should be removed.

Source/Core/Core/Config/UISettings.cpp
```diff
@@ -15,6 +15,77 @@ const Info<bool> MAIN_RECURSIVE_ISO_PATHS{{System::Main, "General", "RecursiveIS
...
+const Info<bool> MAIN_BUTTON_TOGGLE_WII_6{
+    {System::Main, "Android", "ButtonToggleWiimoteButtonHome"}, true};
+const Info<bool> MAIN_BUTTON_TOGGLE_WII_7{{System::Main, "Android", "ButtonToggleWiimoteDPad"},
+                                          true};
+const Info<bool> MAIN_BUTTON_TOGGLE_WII_8{{System::Main, "Android", "ButtonToggleNunchukButtonC"},
+                                          true};
+const Info<bool> MAIN_BUTTON_TOGGLE_WII_9{{System::Main, "Android", "ButtonToggleNunchukButtonZ"},
+                                          true};
+const Info<bool> MAIN_BUTTON_TOGGLE_WII_10{
+    {System::Main, "Android", "ButtonToggleWiimoteNunchuckStick"}, true};
```
JosJuice, line 80: Okay, maybe the idea I had in the past of adding every Android setting to the C++ settings list was a bad one :) There's a lot of repetition here, and we don't actually use all these objects for anything... I would appreciate the input from someone who doesn't deal much with the Android side of things here.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/activities/EmulationActivity.java
```diff
@@ -748,71 +743,74 @@ public boolean dispatchKeyEvent(KeyEvent event)
...
     if (sIsGameCubeGame || mPreferences.getInt("wiiController", 3) == 0)
     {
-      for (int i = 0; i < enabledButtons.length; i++)
+      boolean[] gcEnabledButtons = new boolean[11];
+      String gcSettingBase = "MAIN_BUTTON_TOGGLE_GC_";
+
+      for (int i = 0; i < gcEnabledButtons.length; i++)
       {
-        enabledButtons[i] = mPreferences.getBoolean("buttonToggleGc" + i, true);
+        gcEnabledButtons[i] = BooleanSetting.valueOf(gcSettingBase + i).getBoolean(mSettings);
```
JosJuice, line 755: I didn't know that this `valueOf` method existed, but that's pretty handy. An alternative way to accomplish this would be to use the `AdHocBooleanSetting` class, which would let us skip having to declare every possible controller setting in `BooleanSetting`... but I don't know, maybe declaring every possible controller setting is a good thing? If you want to keep declaring all controller settings, then keep using `valueOf`.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/model/BooleanSetting.java
```diff
@@ -39,6 +39,84 @@
...
+  MAIN_BUTTON_TOGGLE_WII_6(Settings.FILE_DOLPHIN, Settings.SECTION_INI_ANDROID,
+          "ButtonToggleWiimoteButtonHome", true),
+  MAIN_BUTTON_TOGGLE_WII_7(Settings.FILE_DOLPHIN, Settings.SECTION_INI_ANDROID,
+          "ButtonToggleWiimoteDPad", true),
+  MAIN_BUTTON_TOGGLE_WII_8(Settings.FILE_DOLPHIN, Settings.SECTION_INI_ANDROID,
+          "ButtonToggleNunchukC", true),
+  MAIN_BUTTON_TOGGLE_WII_9(Settings.FILE_DOLPHIN, Settings.SECTION_INI_ANDROID,
+          "ButtonToggleNunchukZ", true),
+  MAIN_BUTTON_TOGGLE_WII_10(Settings.FILE_DOLPHIN, Settings.SECTION_INI_ANDROID,
+          "ButtonToggleNunchukStick", true),
```
JosJuice, line 113: What do you think about putting these button-specific settings in a separate section? Mostly just to make things less cluttered for people who want to edit the Android section of an INI file manually.

Source/Core/Core/Config/UISettings.cpp
Ebola16: Oh, I'm in favor of removing these Android-only settings if there's no plans to use them in the future from the C++ side. Regex can convert the settings between C++ and Java if needed. Should I just remove them?

JosJuice: Sounds fine to me. We can always re-add them if we need them.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/activities/EmulationActivity.java
JosJuice: You wouldn't need a map, but rather just passing `gcSettingBase + i` to the `AdHocBooleanSetting` constructor instead of to `BooleanSetting.valueOf`. But I'm okay with keeping this approach either way.

Ebola16: Passing `gcSettingBase + i` as a key to `AdHocBooleanSetting` makes less informative keys when read from the INI, which is why I only see `AdHocBooleanSetting` being used with key maps.

JosJuice: Oh, you're referring to how your current setup has integers in the enum names but the actual button names in the key names? Okay, I understand what you mean now.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/model/BooleanSetting.java
Ebola16: The Java-only implementation of adding new sections is gone so I'm not sure how to properly add "AndroidOverlayButtons" section. I tried adding the section string to `Settings` but I couldn't save values to that section without calling `mSettings.saveSettings`, whereas that's unneeded for the "Android" section.

"Android" section is set in `ButtonManager.cpp`. Putting `ini.GetOrCreateSection("AndroidOverlayButtons")` after that didn't work. Any idea what I'm missing for properly making a section? I have a feeling it's obvious.

JosJuice: Ah, right, I didn't think of mentioning that problem. You need to add it to this list: https://github.com/dolphin-emu/dolphin/blob/653f89701db560ffe2c9421c81f3b5347c8dc46f/Source/Core/Core/ConfigLoaders/IsSettingSaveable.cpp#L29

Ebola16: I just realized that an extra space at the end of "Android" while searching prevented me from finding this :-(

Ebola16:
Feel free to review / merge this at any time. I'll occasionally work on the more complicated INI conversions and if I figure them out I'll commit them here, like the latest commit.

JosJuice:
I'd much rather prefer to not have the number of settings migrated by this PR increase, otherwise this PR becomes harder for me to review the more time passes, which is kind of the opposite of how it usually works. I would suggest that you remove the ISOPaths commit from this PR so that I can approve it sooner.

Ebola16:
That's fair, most PR merges I've done are to avoid conflicts with my other PRs.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/model/Settings.java
```diff
@@ -24,6 +24,7 @@
   public static final String FILE_WIIMOTE = "WiimoteNew";
 
   public static final String SECTION_INI_ANDROID = "Android";
+  public static final String SECTION_INI_ANDROID_OVERLAY_BUTTONS = "AndroidOverlayButtons";
```
JosJuice, line 27: This section name feels to me like it would only include things like button positions and whether each button is enabled, but it also includes things like `MAIN_JOYSTICK_REL_CENTER`. If such other settings are also going to be included, what do you think about calling it "AndroidOverlay" instead?

JosJuice (commented):
We should probably call `mSettings.saveSettings` here, so that settings will persist the next time the user runs the app: https://github.com/dolphin-emu/dolphin/blob/5a939ccfaf79c4a3b451e25d7550f663f0845314/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/activities/EmulationActivity.java#L325-L328

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/model/Settings.java
Ebola16: Button position data will occupy a lot of space. I think I'd rather keep the `AndroidOverlayButtons` section and move out settings that don't have the word "button" in them. Some of these settings could fit in either section.

JosJuice (approved):

PR #8010, opened 2019-04-19
VideoCommon: Add imgui-based graphics settings window

stenzek:
This PR adds a graphics configuration window which can be used in-game. The main motivation is so that graphics options can be changed while the game is running on android, but more relevant to myself, nogui builds.

Note that the options won't be tweakable on android yet, as we haven't hooked up imgui input to touch events. This will come later.

imgui

Options overriden by gameini/netplay/etc will be shown in red (we can't do bold in imgui, since it's a bitmap font).
image

PR #8006, opened 2019-04-17
DolphinQt: Enable RTL layout

JosJuice:
Fixes https://bugs.dolphin-emu.org/issues/11508

PR #9153, opened 2020-10-17
Remove description box in graphics tabs and use custom tooltips instead

iwubcode:
Instead of a bulky description box for graphics descriptions, we now display a custom tooltip.

Based on @MayImilae 's mockup:  https://i.imgur.com/cAeAO7X.png

Could easily be used for other (non graphics) settings later.

Examples of how it looks on each color theme below.

Light theme (windows default):

image

Dark theme (qt theme "Skyline Nightfall"):

image

or if you prefer, you can use the low contrast version (enabled via option, off by default):

Light theme (windows default):

image

Dark theme (qt theme "Skyline Nightfall"):

image


iwubcode:
This is using Qt internal code as a base.  It changes the arrow / box direction if there's not enough screen space for it to fit (try moving the window around to edges of the screen).  More work would be needed to match some of the specifics of the mockup.

Source/Core/DolphinQt/Config/Graphics/BalloonTip.cpp
```diff
@@ -0,0 +1,299 @@
...
+#if QT_VERSION >= QT_VERSION_CHECK(5, 14, 0)
+#include <QScreen>
+#else
+#include <QApplication>
+#include <QDesktopWidget>
+#endif
+
+namespace
+{
+static BalloonTip* theSolitaryBalloonTip = nullptr;
```
lioncash, line 28: `snake_case`

This should be wrapped into a unique_ptr, so it doesn't get potentially flagged as a memory leak by tools like asan.

Source/Core/DolphinQt/Config/Graphics/BalloonTip.h
```diff
@@ -0,0 +1,38 @@
...
+  void balloon(const QPoint&, int, bool);
+
+protected:
+  void paintEvent(QPaintEvent*) override;
+  void resizeEvent(QResizeEvent*) override;
+  void mousePressEvent(QMouseEvent* e) override;
+  void timerEvent(QTimerEvent* e) override;
+
+private:
+  QPixmap pixmap;
```
lioncash, line 35: m_pixmap

Source/Core/DolphinQt/Config/Graphics/BalloonTip.h
```diff
@@ -0,0 +1,38 @@
...
+  int timerId;
```
lioncash, line 36: `int m_timer_id = -1;`

Would allow the initializer to be removed from the constructor initializer list

Source/Core/DolphinQt/Config/Graphics/BalloonTip.h
```diff
@@ -0,0 +1,38 @@
+// Copyright 2020 Dolphin Emulator Project
+// Licensed under GPLv2+
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <QWidget>
+
+#include <QPixmap>
```
lioncash, line 9: This should go above QWidget's include.

Source/Core/DolphinQt/Config/Graphics/BalloonTip.cpp
```diff
@@ -0,0 +1,299 @@
...
+  QPainter painter(this);
+  painter.drawPixmap(rect(), pixmap);
+}
+
+void BalloonTip::resizeEvent(QResizeEvent* ev)
+{
+  QWidget::resizeEvent(ev);
+}
+
+void BalloonTip::balloon(const QPoint& pos, int msecs, bool showTheArrow)
```
lioncash, line 173:
```suggestion
void BalloonTip::balloon(const QPoint& pos, int msecs, bool show_the_arrow)
```

msecs can likely also be a std::chrono type.

Source/Core/DolphinQt/Config/Graphics/GraphicsBool.cpp
```diff
@@ -28,6 +30,28 @@ GraphicsBool::GraphicsBool(const QString& label, const Config::Info<bool>& setti
...
 }
 
+void GraphicsBool::SetTitle(QString title)
+{
+  m_title = title;
+}
+
+void GraphicsBool::SetDescription(QString description)
+{
+  m_description = description;
```
lioncash, line 40:
```suggestion
  m_description = std::move(description);
```

Source/Core/DolphinQt/Config/Graphics/BalloonTip.h
```diff
@@ -0,0 +1,38 @@
...
+protected:
+  void paintEvent(QPaintEvent*) override;
+  void resizeEvent(QResizeEvent*) override;
+  void mousePressEvent(QMouseEvent* e) override;
+  void timerEvent(QTimerEvent* e) override;
+
+private:
+  QPixmap pixmap;
+  int timerId;
+  bool showArrow;
```
lioncash, line 37:
```suggestion
  bool m_show_arrow = true;
```

Would allow the initializer to be removed from the constructor initializer list

Source/Core/DolphinQt/Config/Graphics/BalloonTip.cpp
```diff
@@ -0,0 +1,299 @@
...
+{
+  this->showArrow = showTheArrow;
+
+#if QT_VERSION < QT_VERSION_CHECK(5, 14, 0)
+  const QRect screenRect = QApplication::desktop()->screenGeometry(pos);
```
lioncash, line 178: snake_case, ditto anywhere else in this file.

Source/Core/DolphinQt/Config/Graphics/BalloonTip.cpp
```diff
@@ -0,0 +1,299 @@
...
+                       QColor background, QColor text)
+    : QWidget(nullptr, Qt::ToolTip), timerId(-1), showArrow(true)
+{
+  setWindowFlags(Qt::FramelessWindowHint);
+  setAttribute(Qt::WA_NoSystemBackground);
+  setAttribute(Qt::WA_TranslucentBackground);
+  setAttribute(Qt::WA_TransparentForMouseEvents);
+  setAttribute(Qt::WA_DeleteOnClose);
+
+#if QT_CONFIG(label)
```
lioncash, line 86: This if isn't necessary (ditto elsewhere this is present). A large amount of our code would have broken already if QLabel wasn't present.

Source/Core/DolphinQt/Config/Graphics/BalloonTip.cpp
```diff
@@ -0,0 +1,299 @@
...
+#if QT_VERSION < QT_VERSION_CHECK(5, 14, 0)
+  const int limit = QApplication::desktop()->availableGeometry(msgLabel).width() / 3;
+#else
+  const int limit = msgLabel->screen()->availableGeometry().width() / 3;
+#endif
+  if (msgLabel->sizeHint().width() > limit)
+  {
+    msgLabel->setWordWrap(true);
+
+    // TODO: if we are still greater than the limit we need to set the
```
lioncash, line 114: This should be removed if this isn't going to be used.

Source/Core/DolphinQt/Config/Graphics/GraphicsBool.cpp
```diff
@@ -28,6 +30,28 @@ GraphicsBool::GraphicsBool(const QString& label, const Config::Info<bool>& setti
   });
 }
 
+void GraphicsBool::SetTitle(QString title)
+{
+  m_title = title;
```
lioncash, line 35:
```suggestion
  m_title = std::move(title);
```

iwubcode:
Updated to support most General/Enhancement tab options.  Will wait on @MayImilae 's feedback before doing the rest of the tabs and cleanup.

mbc07:
Just gave a try, and I'm neutral on this one. On the plus side, it'll allow getting rid of the description box, which made the Graphics config window too tall to fit on a 1080p screen at 125% scaling since the recent DPI fixes, but on the negative side:

* The tooltip "spawns" another window, visible on Alt+Tab menu and other places
* That spawned window drives away the focus from the main window, and that makes both the taskbar icon and the main window to flash, as they trigger the OS transitions/animations
* Tooltip style isn't consistent with anything else on current GUI, feels very out of place, the arrow from the tooltip doesn't seem to align properly either

With that said, isn't there any way to make Qt spawn "regular" tooltips, that follows the theming of the OS and that doesn't spawn a new window (like the tooltips that appear when hovering over a file/folder on Windows Explorer and on Finder)? That would address all of that points in a single "shot"...

iwubcode:
@mbc07 - these tooltips are completely custom code, so don't go thinking it's a "Qt" tooltip.  We do this so we can get theming and the arrow that was presented in the mockup.

The tooltips produced were not their own window but when I introduced the fade-in logic, the whole widget wasn't fading, just the contents and background color.  A black background appeared instead.

To avoid this, looking online I used the flag 'FramelessWindowHint'.  That fixed the animation but also inadvertently made it a window.  I saw the window was losing focus but didn't see your other concerns so I thought it'd be ok.

I guess I will take the animation out for now, until I can figure out a proper way to have my cake and eat it too.

mbc07:
> these tooltips are completely custom code. I did this so as to reproduce the theming and the arrow that was presented in the mockup.

Regular tooltips will follow the OS theme, that's the whole point. If we go custom and pick a different style, how can we assure whatever style we pick will display nicely on all OSes we support? That becomes a mess very quickly when you take Linux in consideration, as there's a HUGE variety of distros, each one with their own visual style. By using regular tooltips they will follow the OS theme and that's guarantee that it will "fit" regardless of which OS Dolphin is running.

Going completely custom just to have an arrow and different border seems counterintuitive. You can also use HTML-style formatting with the regular tooltips, which is probably enough to make it stand out more while retaining the OS theme, and it's probably less code to write too...

iwubcode:
@mbc07 - it's not that hard to make this tooltip follow an "OS" theme (my initial code had that) but again the mockup that was provided to me had custom colors.  The idea is to have a high contrast color so it stands out.  We might be able to request that from Qt.  I haven't looked.

I don't mind switching to the OS colors if that is what the consensus is but again, that wasn't what was suggested.

I do think the arrow and rounded corners are a nice touch.  I guess one could argue that we should save these sort of modern changes for a UI overhaul (which Dolphin desperately needs) but I see nothing wrong in slowly updating things.  Hence why I decided to do this and also implemented #8823 as well.

> You can also use HTML-style formatting with the text of regular tooltips, which is probably enough to make it stand out more while retaining the OS theme, and it's probably less code to write too...

I'm well aware of what they support.  It just wasn't desired.

mbc07:
Fair enough. For me the custom style (at least the current design) feels very out of place, at least on Windows, and I would rather go with the regular tooltips using the OS theme instead.

If the consensus ends going with fully customized tooltips, the window spawning and focus change problems should be definitely fixed, and I would rather rave those custom tooltips somehow inherit a suitable color scheme from the host OS than hardcoding a specific theming that there's no way to ensure it will display nicely on all OSes Dolphin can run on...

iwubcode:
@mbc07 - thanks for looking!  `QBalloonTip` is what this is based off of.  Just with a few tweaks and removals.

iwubcode:
@mbc07 - as far as I can tell, qt requires the use of a window in order to do transparent effects.  Looks like a qt bug was opened at one point but then closed due to age...  I was able to get rid of the focus loss but the hidden window remained.

So I removed the fade-in logic for now, the tooltip is no longer a window.  Please retest when you have time.

MayImilae:
So the whole point of this is that standard windows tooltips... suck. Like, they are the _worst possible_ tooltips. White box on black text over white box on black text, with only a tiny shadow and a faint touch of blue to distinguish it. It's awful. Tooltips need to be attention grabbing so users know they exist at all, and they need to be very clear what it is pointing at and what its contents are so users can quickly understand what the new information applies to and what that information is so they can move on. The Windows standard tooltip fails on both fronts. 

The standard windows tooltip is not good enough to replace our description area, which is why it was created in the first place. But a better tooltip just might. 

So, I said standard windows tooltip here because, well, windows HAS better tooltips! But no one uses them, go figure. None of them are as good as the mockup I presented though, and all of the unusual ones also haven't been updated since Windows XP so... they look like Windows XP. I suppose that's a good reason not to touch them for us right now.

mbc07:
@iwubcode - gave another try. It's a bummer the fade effect got axed, but not having the constant flashing from the focus changes and the windows spawning is a huge improvement. It won't replace the fade effect but perhaps adding a delay of one second before showing the tooltip might somehow help? 

About the tooltip arrow, have you tried making it point to where the mouse cursor is located instead of always pointing to the left corner of the checkbox/dropdown lists? Or perhaps add a little bit of padding on the left so it always point to the middle of the checkbox when hovered (although I have no idea how that padding would behave in right-to-left languages)? It would mimick the behavior of regular tooltips a little bit better...

---

@MayImilae - nearly black balloon tooltip with an arrow (current design, and hardcoded) on a mostly white window isn't any better, definitely looks out of place and worse than a regular tooltip TBH. Even the dated Windows Vista/7 style found in some Win32 apps looks better than that, and unlikely the current approach (and my main issue with it), they **will react** to OS theme changes. 

[MSI Afterburner is a perfect example of that*](https://user-images.githubusercontent.com/8005966/96360558-3e640280-10ec-11eb-9532-38b630fe8f83.png), the tooltip always point to where the mouse cursor is and it properly reacts to OS theme changes. It's also very comparable to Dolphin since Afterburner also has a tabbed config window with various checkboxes/dropdown lists and big descriptions on each one.

On this PR, however, If I enable dark mode or turn on high contrast mode, the custom tooltip will retain the same hardcoded style and the border and arrow will get essentially invisible as result, looking way worse. That wouldn't happen when using the regular tooltips nor with the balloon tooltips because they would inherit the OS theme **by design**, regardless if you're using Windows, macOS or one of the thousands of Linux distros available, which is where that hardcoded approach has the most potential for issues.

**TL;DR**, unless there's a way to ensure that same "inherit theme from OS" functionality on the proposed custom tooltips instead of a hardcoded style, I strongly think regular or balloon tooltips should be used instead (although I'm not sure if the latter is available on other OSes). Going with a custom solution that (currently) looks and behaves worse than the standard solution available across all OSes just for the sake of "regular tooltips sucks" seems very counterintuitive. It should at least be on the same level of functionality and the current approach isn't, TBH.

---

_*@iwubcode mentioned this PR was based on QBallonTip, which seems able to produce [this kind of tooltips](https://docs.microsoft.com/en-us/windows/win32/uxguide/ctrl-balloons) on Windows (and that's what MSI Afterburner uses), so let's hope it can be extended to other OSes_

MayImilae:
Hmm, we have a bit of a language issue here, so I'll try to clarify. By better, I'm specifically referring to functionality. The standard windows tooltip that dolphin has been using functions horrifically as a tooltip. It just doesn't do the job well at all. This one, however, is far FARA superior at doing the job a tooltip is meant to perform - letting users know about auxillary info and allowing them to quickly consume it.  

How it looks is a separate matter to that, and we need to keep the apart as we discuss it. It's totally fine if you don't like how it looks, I made the mockup to demonstrate as many function tooltip techniques as possible, after all! When talking about tooltips, it's very hard to talk about them if the only thing anyone has ever seen is the garbage windows standard ones. So yea, you may not like how this looks, and that's ok! We can talk about the aesthetics. But the point of this is that functionality improvement, and as we talk about this, we need to be sure to keep the aesthetic wants and functional separate clearly defined.

Speaking of functionality, here is all of the functionality that is in the original mockup. I outlined it in IRC but it hasn't been spoken here yet, so, I'll just outline the techniques it is demonstrating.

https://i.imgur.com/cAeAO7X.png

* Fading out of non-hovered items
* Tooltip is connected to the pointing it is pointing out (not bound to the mouse cursor, go away windows 95)
* Tooltip is above or to the right of the item in question, NOT below (we tend to read top to bottom, having the tooltip below can obstruct the next thing we want to read. bad design)
* Literally points to the item it is pointing out with a triangle
* High contrast tooltip makes it extremely obvious that it is there, making sure users notice the functionality, and improving clarity (white box with black text on white box with black text is horrible design, the whole point of a tooltip is to be seen!)
* Text is slightly enlarged compared to the text it is pointing out for clarity
* Glanceable information is in bold and a different colour, making it especially stand out

I never thought that we'd have ALL of these in a tooltip, and yea, most of these aren't in this PR. That's fine! ...though a couple of them I miss, I'll be making another comment later.

mbc07:
It still has a hardcoded style instead of inheriting the OS theme and reacting to changes made to it, that's what I meant when I said it's not on "the same level of functionality" and the entire point I'm arguing since the beginning. 

That single point alone can cause various readability issues because it's simply **not** possible to hardcode a theme or color scheme that will magically fit in all OSes we support and in all themes they might or might not have (again, Linux and its thousands of distros). Regular tooltips and even the current description boxes that are going to be replaced by this PR doesn't have this problem since they'll always have have a consistent and "readable" visual regardless of the OS or theme being used, because they inherit the OS theme and will react to changes made to it, by design. 

If this can't be achieved on the custom solution this PR is proposing, then it's not worth the trouble IMO, it simply won't matter how much better our custom tooltip implementation is if they have the potential of being unreadable depending of the combination of OS and theme our end-users might be using.

altimumdelta:
I honestly had no problem with the description box, infact I found it to be a great idea that I included in some of my own apps as I liked it in Dolphin.

I find the tooltips in general to be distracting and blocking the window/text behind it. I generally do not support removal of the description box, but I wouldn't mind it if this is an option one can choose to switch between. However if the tooltip or some later variation of it would be made to be as good as it can in preventing getting-in-the-way and popping up at wrong times, then I might change my mind, but if the vertical size of the config dialogs is really that of a problem I'm not sure yet.

> I guess one could argue that we should save these sort of modern changes for a UI overhaul (which Dolphin **desperately needs**)

On what OS/Disctro/Theme is that statement based upon anyway? Who or what is behind the term "modern"? I bet it's about the (IMO ugly) Google's Material design no? It is not fair for some specific trend that comes and goes to seize the context of a common english language term, and we are helping it if we do not recognize this mistake.

At each milestone of such propositions, let's pause and ask what we have is really art and design, is it really original and unique to a specific project like Dolphin, or it's just about submitting to the styles that tech giants have been heavily forcing upon us. Everwhere I look around FOSS, there's usually this peer-pressure and some sort of IMO made up need to follow and switch closer to the popular trends, just for the sake of being part of that trend almost, styles and liknessess of the biggest actors who are behind promoting and persuading us why we should all use their kind of GUIs and why theirs is the best, and most importantly, **why there is no need for anything else to coexist**.

I believe many such GUI redesigns are influenced by the tech giants remotely/passively, by their users, who have been heavily influenced, including from early age, to their styles by the use of their smartphones and other devices that are completely unrelated to a project the user is trying to project* their ideas onto. *(different context of the term "project")

So I would like that such proposers are fully informed about this key aspect of conscious and subconscious bias and that they have to put some more effort into their propositions to be more fair to the project it self when desiging a new GUI for a project that never had any relation to any of sources of these external influences. The external influences should have a lesser weight in decisions and ability to drive the development and design process.

That said, I am by no means defending the old and existing designs, anything could still improve or it could change, some things are never obsolete or worse or better, some things are more equal than they seem, and in the end it's just down to subjective taste. What I'm warning about is that this "personal taste" can be manipulated with, unknowingly. It also happens that these tech giants and the world of advertising are masters at this.

In the end the existing themes and styles that most desktop application had are also influences, external ones too, yes we users didn't really build them ourselfs at the births of various operating systems, I don't deny that ofcourse. The new solution usually has a high mountain to climb, not only it has to persuade why it's better, how it's better, why it's the only way to go, it needs user's trust so they get trained and used with it and only a long way later when the user got used to it, only then can the user evaluate whether it was worth it. Was the transition fast enough or slow? Was it annoying and hard? Did it pay off? Was the new solution really good enough over the old one on day one? How long did it take to be perfected and polished?
The problem is this usually is not true, new GUIs from tech giants always miss a lot when they're first introduced, they're rushed to the market, I remember the new AMD GPU drivers were missing Color/Brightness/Contrast controls for like a whole year for my GPU at least, I was furious. 

New solutions that replace old ones simply need more time, even when they are completed they have to linger and coexist, this gives the users vital real-time ability to compare and analyze how either one feels for them, and this is also the crucial beta-testing stage that discovers previously undetected issues with the either old or the new solution.

> The standard windows tooltip is not good enough to replace our description area, which is why it was created in the first place. But a better tooltip just might.

I agree with that, IMO if an existing solution +95% works and is tried and tested overtime, then something replacing it should be quite a bit better to justify, it even goes into philosophical debate, it has to be good enough to eliminate any feeling of loss from existing users who are quite attached to the existing solution. If a new solution is the same or even slightly worse it usually becomes a popularity contest and taste, which makes things unfair.

I'm not totally against this idea if the community fairly determines that vertical space would be a big issue warranting removal among other things, because I've seen rounded "?" icons appearing at the end of option or statement labels, even with Microsoft's GUIs (IIRC) to click on which shows the tooltip and that feels a lot better, it means only a conscious and intended manual action from the user triggers a tooltip.

Except ... the process of determining what are the existing problems with the description box and whether these problems are heavy enough to first warrant changes to it, which then proceeds into research and prototyping if those changes work out, and if they don't, only then the research and prototyping for a replacement should occur, all this before a replacement is proposed. Optimally in a proper development team, but ofcourse since this whole FOSS ecosystem is all free-for-all unless specifically managed and depends on whether project has such rules or not, everyone just brings in their ideas day-one, which is another subject and I'm not arguing about that now.

iwubcode:
> It still has a hardcoded style instead of inheriting the OS theme and reacting to changes made to it, that's what I meant when I said it's not on "the same level of functionality" and the entire point I'm arguing since the beginning.

@mbc07 - to address your concerns, the plan is to have two preset themes for the tooltip: dark and light.  The dark theme will be applied when the window color the users are using is determined "light" and the light theme will be applied when the window color the users are using is determined "dark".

We have to do something in code since there isn't a way in Qt to request a "contrast" color, at least not as far as I know.

I'm still looking into it.  But rest assured, I don't intend to have dark users have this dark tooltip.

> I honestly had no problem with the description box, infact I found it to be a great idea that I included in some of my own apps as I liked it in Dolphin.

@altimumdelta  - I'm glad to hear you like it but I've felt it needs to go a couple reasons:

* No other application presents its "help" information in this way.  Old Dolphin users are used to it but it's not really a good practice.  It gets back to my "modern" comment.
* It takes up a good chunk of the screen, this means it won't fit in certain designs.  Imagine putting this in the inputs window, where would it go?  Behind the scenes, I have other features with new UI, there's no way to put help information because to do so would require a "description box"

> On what OS/Disctro/Theme is that statement based upon anyway? 

I made this comment about the rounded edges but that was out of place and using the word "modern" didn't really help.  My problem with the current desisgn is that Dolphin is a mishmash of design ideas thrown into one application.

Issues have cropped up such as:

* Space issues like #8785 
* Klunky double tab views thrown in the game specific UI
* Not being able to edit the game specific settings in the same way as "global" UI
* Users wanting to set settings via their controller (in a say "tv mode").  The workaround at the moment is to use IMGUI to build a secondary UI.  See:  #7930 and #8010 

and now the description box for reasons which I mention above.

It's things users (myself included) are used to.  But the system born off of WxWidgets is just starting to show it's age.

I have in my head a system similar to PSP but that is just my own personal viewpoint.  It'd also be a lot of work, so I'm trying to make fixes where I can.

> I'm not totally against this idea if the community fairly determines that vertical space would be a big issue warranting removal among other things, because I've seen rounded "?" icons appearing at the end of option or statement labels, even with Microsoft's GUIs (IIRC) to click on which shows the tooltip and that feels a lot better, it means only a conscious and intended manual action from the user triggers a tooltip.

I suggested this but the thought was there would be so many "?" icons, it might be distracting.  I don't necessarily have a problem with making the tooltip be displayed on demand.

---

Again, my only goal of this PR to get rid of the description box so we can leverage a "help" system across our entire UI.  How we do it, I don't really mind.

iwubcode:
Updated with light/dark theme logic.  Tweaked the arrow point location on all controls.

mbc07:
Still not a fan of having dark tooltips on a light theme and vice-versa, but the dark/light detection works and properly switches the tooltip color scheme when necessary. A future improvement would be retrieving dark/light color values from the OS or the currently applied Qt theme instead of hardcoding two sets of different colors for the dark/light tooltip variants, but regardless of that the current approach pretty much addresses my concerns with readability.

The tooltip arrows, however, still behave oddly. If you just open the Graphics config window, they'll point to the middle-ish of checkboxes/radio buttons and somewhere between the left border and the center of drop-down lists (like on the screenshots from the OP) and that looks fine. However, it starts falling apart the moment you start messing with the window:

* Resizing the window in any way that enlarges or contracts the horizontal size will make the arrows be misplaced. They will end pointing to the middle of the text instead of the checkbox/radio buttons on a large window and on dropdown lists they might point outside of the graphics window if it's too small.

* Moving the window to a position where it won't have enough space to display the tooltip below the highlighted option will move the tooltip upwards (as expected), but the arrow will still be anchored downwards and that obstructs both the checkbox/radio button and part of the button label in that situation.

I wanted to test a right-to-left language to see how it would behave with the repositioned checkboxes, but just changing Dolphin's language doesn't seem enough to trigger that, as the checkboxes remained at the same place. Might give a try later by changing the OS settings to a right-to-left language instead, but I kinda can't do that at the moment...

iwubcode:
@mbc07 - thanks for your feedback.  I changed the positions again and I'm now a bit smarter about the location.  Please take a look and see what you think..

mbc07:
It improved, like, A LOT. The arrow won't obstruct checkboxes when it's drawn upwards due to lack of space anymore and will always point exactly at the center of the checkbox or dropdown list, even after resizing and moving the window around. 

The arrow might still point outside the window if it becomes too small and you hover over dropdown lists, but I guess that's not really fixable since even Qt seems to give up at that point, as it will stop resizing the controls and just slap a horizontal scrollbar there when the window becomes too small, so that's fine. Really good work!

iwubcode:
@MayImilae  asked if it could be made more like the mockup.  So I increased the thickness of the arrow and made it point upward by default.

I'll wait a bit longer for feedback then proceed to change the other options and remove the description box.

altimumdelta:
> I suggested this but the thought was there would be so many "?" icons, it might be distracting. I don't necessarily have a problem with making the tooltip be displayed on demand.

Oh, exactly. I had this thought in the back of my mind but forgot to mention it. Perhaps some kind of a rule could be introduced that follows several guidelines of how long or big the tooltip would end up being (based on the amount of text it holds) where such a button or some other way of manual trigger is introduced. Some tooltips I think would be so big the would block more than half the dialog.

And ofcourse you've experimented with the hover delay, how long one has to hover for it to appear. 

> I made the modern comment as I was referring to the rounded corners, soft shadows, etc suggested by the design but I feel this could extend out to Dolphin's UI design as a whole. As I said, I think it needs a refresh. Dolphin is still using its WxWidgets base and hasn't really improved on it since adopting Qt.

Yeah, I am not against a complete overhaul in the future or otherwise, but in the sense that it's a more involved process which starts with re-evauluation and research stage first, before going into the idea nd prototyping stage with some idea which would eventually be protested against and would morph into something totally different anyway. So that we have a GUI that you can proudly say that it feels like Dolphin and is unique in some ways to other projects, some mechanics that do make sense would obviously be similar or same, that's just the nature of things, there are a number of modern ideas and solution I myself jumped onboard very quickly, where my mind feels like it. If I was the project owner I would create a PR submitting readme file as part of the continuation of "how to contribute" page, just a short or as necessar list of recommended guidelines for what kind of PR's are and aren't welcome and there it would explain that GUI is a delicate  thing and people shouldn't waste too much effort into some specific idea without consultation and discussion first (the evaluation and research stage), currently this is all done informally and word-of-mouth through IRC, on any subject, AFAIK.

Excuse me if I haven't been more specific yet on commenting the actual tooltip experience as it's being developed. I will hopefully try to find more time to test this PR myself in a bit, I'm in a middle of a major PC maintenance and build right now with stuff not installed and so on, I do have an Kubuntu and Git that I just setup a few days ago on a tertiary machine and I'll use that OS for testing this for now.

JosJuice:
> I wanted to test a right-to-left language to see how it would behave with the repositioned checkboxes, but just changing Dolphin's language doesn't seem enough to trigger that, as the checkboxes remained at the same place. Might give a try later by changing the OS settings to a right-to-left language instead, but I kinda can't do that at the moment...

@mbc07 DolphinQt currently has no support for RTL layouts. You would need to edit the source code a little in order to get it. The proper fix is PR #8006, but simply downloading this PR won't give you RTL layout, because the translation files haven't been updated to accommodate it yet.

iwubcode:
Updated to add a timeout between creating the tooltip.  Right now set to 300ms (arbitrary).  That makes it so moving your mouse over options doesn't obstruct your view in the process.

I also put some infrastructure code in place that will help to convert the rest of the tabs.

MayImilae:
It's still crashing in macOS. :(. 

Here is what Dolphin logged into Terminal - https://hastebin.com/ruwitoziva.yaml

And here is the crash report from macOS - https://hastebin.com/liretixifi.yaml

And that is from -
Branch: refs/pull/9153/head:
Revision: ef69003a082f08440a20358ae5daae4a260b9c4b

altimumdelta:
> Right now set to 300ms (arbitrary).

Eventually, I wouldn't mind this being a proper adjustable option (slider) in general settings. Would anyone else?

MayImilae:
Tested on macOS. No more crashing! But it is a definite downgrade, even compared to the description area. Meh. It's macOS, can't be helped.

Screen Shot 2020-10-24 at 21 07 00-2

iwubcode:
We discussed and for all the power users out there, we added an option to disable the high contrast.  I updated the main post with images.

With the crashing fix and this change in, I'm marking this 'ready for review'.  Appreciate everyone's feedback and help!

mbc07:
Gave it another try and it works as expected both with high contrast tooltips enabled/disabled, they still react to theme changes too. However, there's still some points I think it's worth discussing:

* It seems there's two different light tooltip styles and two different dark tooltip styles, and they're interchanged based if high contrast tooltips are enabled or not. Is there any reason we couldn't reuse the same light and dark styles on both "high contrast tooltips enabled" and "high contrast tooltips disabled" modes?

* Dark tooltips on a dark theme got the readability issue I kept mentioning earlier: it's nearly impossible to distinguish the tooltip border from the actual config window. I have observed how the native Windows balloon tooltips behaves on these cases and it's rather simple: the border of the tooltip will just always use the same color of the text (both in [dark](https://user-images.githubusercontent.com/8005966/97102104-19860700-1679-11eb-80a1-0ccdb46fd6ad.png) and [light](https://user-images.githubusercontent.com/8005966/97102189-e98b3380-1679-11eb-8a27-2634bfa7879d.png) modes). This ensures the tooltips are always distinguishable from the background, regardless of the theme being used. Would it be possible to make Dolphin's custom tooltip borders do this too and reuse the same color from the actual text? It looks like it's hardcoded to gray right now.

* "Synchronous (Ubershaders)" tooltip seems truncated. Not sure if it's due to high DPI mode (which I use) but it abruptly ends in "Only recommended for high end". I checked all other tooltips and this happens only on this one. Also, "Free Look Control Type" tooltip seems very crowded. Could you try adding an empty line between each control type (six axis, orbital, etc.)? I think this would help a lot with the readability of that particular tooltip.

Other than that, I'm overall satisfied with the results. The graphics config window now fits entirely on my screen again with the removal of the description boxes 🙂

iwubcode:
@mbc07 - thanks for the further feedback.  Really appreciate it!

> Is there any reason we couldn't reuse the same light and dark styles on both "high contrast tooltips enabled" and "high contrast tooltips disabled" modes?

I'm not sure what you mean.  The whole reason we have two different modes is because "high contrast" has a different color scheme than "low contrast".

> Would it be possible to make Dolphin's custom tooltip borders do this too and reuse the same color from the actual text? It looks like it's hardcoded to gray right now.

Yep, I updated to make this change, thanks!

> "Synchronous (Ubershaders)" tooltip seems truncated. Not sure if it's due to high DPI mode (which I use) but it abruptly ends in "Only recommended for high end". I checked all other tooltips and this happens only on this one

I triple checked this one but I'm not sure what you mean.  This has the same text as what was in the description box.  What do you mean by "abruptly ends"?

> Also, "Free Look Control Type" tooltip seems very crowded. Could you try adding an empty line between each control type (six axis, orbital, etc.)? I think this would help a lot with the readability of that particular tooltip.

This also matched the previous description box but I went ahead and added an extra line to make it less crowded.

iwubcode:
For all the thumbs down users, I'm curious what it is you don't like.  If it's just the description box being replaced, there's not much I can do there but if it's something else, please leave a comment and maybe a change can be made.

mbc07:
> I'm not sure what you mean. The whole reason we have two different modes is because "high contrast" has a different color scheme than "low contrast".

What I meant is that the light tooltip used in dark mode when high contrast enabled seems different from the light tooltip used in light mode when high contrast disabled (the same with the dark tooltips when high contrast is enabled/disabled). Not a blocker nor anything, I was just a bit curious about that :)

> I triple checked this one but I'm not sure what you mean. This has the same text as what was in the description box. What do you mean by "abruptly ends"?

The last line of text in that tooltip, specifically, is truncated. It displays as "Only recommended for high end" instead of "Only recommended for high end systems." and now I'm suspecting this might be somehow caused by high DPI mode, although it's puzzling that "Synchronous (Ubershaders)" is the only tooltip affected, all other tooltips are normal.

I'm not at my PC right now but I can provide a screenshot later, if it helps...

iwubcode:
> What I meant is that the light tooltip used in dark mode when high contrast enabled seems different from the light tooltip used in light mode when high contrast disabled (the same with the dark tooltips when high contrast is enabled/disabled). Not a blocker nor anything, I was just a bit curious about that :)

Ah ok, that's what I thought.  The low-contrast version uses the window background color and window text color.  The high contrast doesn't allow us to pick from the theme's colors, so we have to hardcode something.

For the light theme, I used the mockup color.  For the dark theme, it was decided to use white.

That's why they are different unfortunately.

> The last line of text in that tooltip, specifically, is truncated. It displays as "Only recommended for high end" instead of "Only recommended for high end systems." and now I'm suspecting this might be somehow caused by high DPI mode, although it's puzzling that "Synchronous (Ubershaders)" is the only tooltip affected, all other tooltips are normal.
> I'm not at my PC right now but I can provide a screenshot later, if it helps...

I think I understand the issue but yeah, not sure why that would be happening.

In addition to a screenshot, I'd be interesting on your windows settings.  I use a 4k screen with 125% scale.

mbc07:
@iwubcode it looks like this: 
image

I rechecked all other tooltips, all of them display normally, only the "Synchronous (Ubershaders)" tooltip is truncated, regardless of using high contrast mode or not. My screen is 15.6" and it runs at 1920x1080 with 125% scaling on Windows settings...

iwubcode:
@mbc07 - thanks!  I'm still looking into what exactly the problem is but I pushed up a slight change, can you retest when  you have time?

mbc07:
The last commit fixed the truncated text, but it also made the text size inside the tooltips slightly smaller (they seem to be the same size as the button labels now) and tooltips with long descriptions got significantly wider, some examples:

image 
image

I particularly don't mind those, but one of the design goals pinpointed by MayImilae earlier was for the tooltips to have slightly larger text, which is gone with the latest commit =/

iwubcode:
@mbc07 - thanks!

Yes, I am still working on a more appropriate solution, I just wanted to know if this change works.  That was the solution qt used for their internal code.

The issue with many of these text blobs, is they have no line breaks.  For our description box, that didn't matter as it was a set size.  These tooltips can be any size, though qt's internal code limits it to a 3rd of the total size of the screen and sets the boxes as being a fixed size (still not sure how the height dimension is determined).  The issue you were seeing originally is likely due to the fact that at 125% and a 1920 screen, the width / height is set to some fixed size but the text overshoots it, so the text is clipped.  Still not sure why some of the longer text blobs were not showing this behavior.

Qt's internal code leverages some internal logic to handle this edge case, allowing the wrapping to wrap anywhere.  The public api only allows us to wrap on spaces..

It's also worth noting that the long text blocks were always there, your screen (+125% scale) was likely too small to see it.

The easiest solution is to remove the constraint on the tooltip size or to add line breaks but both of those are less than ideal solutions, especially because May's original concept wanted the text _even larger_.

iwubcode:
@mbc07  - please retest when you have time.  I believe I have fixed the issue!

mbc07:
The tooltip on "Synchronous (Ubershaders)" wraps properly now instead of being truncated, but the tooltip text size seems to remain the same size of the labels and not slightly bigger like before:
image

-----

Most tooltips remain the same size as before your attempts to fix the truncated text, but the following ones got slightly wider (still way less wider than the previous attempt, though):

image
image
image
image
image
image
image

I tried to find a pattern between the ones that got wider but I failed ¯\\\_(ツ)_/¯

iwubcode:
@mbc07 - thanks again for testing!

> The tooltip on "Synchronous (Ubershaders)" wraps properly now instead of being truncated, but the tooltip text size seems to remain the same size of the labels and not slightly bigger like before:
> image

Hmm strange, I didn't change the text size.  Are you sure it was bigger?  MayImilae's mockup was bigger but I never implemented that.  Looking at your [before](https://user-images.githubusercontent.com/8005966/97120423-33f9c800-16ed-11eb-88c5-bab911ba8754.png) and [after](https://user-images.githubusercontent.com/8005966/97131359-5bb35500-171a-11eb-9b47-dea87a4751d4.png) I think the size is the same?

mbc07:
Hmm, they do look the same. Maybe I confused May's mockup as an actual, implemented version of this PR? Don't know.

Anyway, this PR now LGTM, at least there's nothing else I feel discussing about with the actual implementation...

MayImilae:
Gave this a try on my desktop. No problems with mixed scaling (200% or 150%, the text was and the tooltip were the correct size) and the text was nice and sharp!

Sadly though, the tooltip itself is not HIDPI at all and is bad aliased. At 200%, it's pretty jarring (see below). Plus, to echo everyone else, the corners are rarely correctly round. And I know, Qt problems, I don't know how much you can fix. but just pointing them out.

I'm just glad we don't have mixed DPI problems.

dolphintooltip

altimumdelta:
Idea: Could you try if you can clamp the tooltips to be placed inside the borders of the dialog behind it (parent), and see how that looks overall, perhaps it may help with the sizing of those cases where it looks off.

JMC47:
While I think we can all admit things aren't completely perfect, after using it for a while this morning I'm more than happy enough with the performance to switch over to this.

Our GUI is simply too tall in spots and is a pain in the ass to use on my Linux Laptop that isn't fully 1080p, so I think this is a good first step toward fixing that problem.  The tooltips keep my eyes in one place instead of constantly having to look down toward the bottom of the page every time I hover over something.

I think the rest of the GUI will have to be ported over at some point, but for now just doing the Graphics Tabs will be a nice start.

Techjar:
Yeah, I think this is a great improvement. Hopefully we can sort out the corner, aliasing and hidpi issues in due time. As for macOS, it's unfortunate that we can't seem to fix the crashes and have to just use the OS tooltip, but I agree with MayImilae that we can't let that stall progress.

Definitely would like to see this used for all settings eventually, but the graphics settings is a good start seeing as we already have descriptions there. There's some settings in the NetPlay dialog that I think would really benefit from this as well.

altimumdelta:
However I wouldn't rush to conclusions just yet, secondly, if the reasoning behind it is space that doesn't look well on a "_monitor that isn't fully 1080p_", isn't what is taking place here that it's a change that is designed to fix a problem on low-end device that isn't even up to the mainstream standards for gaming if we care about those, while again not taking into account that on a proper device, according to my knowledge, wasn't a problem at all?

So is it about technicalities or subjectivities, or personal experience, because if one's monitor size is one of the deciding factors, I could just forget all the discussion and just say well I have a 1440p screen that has enough space and so I don't need any changes, that just seems low effort kind of a reasoning IMO that I try to avoid.

Techjar:
Unfortunately sub-1080p displays are still extremely common for things like laptops, or sometimes HiDPI reduces their effective size and creates the same problem.

altimumdelta:
Hmm, allright, I wonder if these laptops are even powerful enough to get that top Dolphin experience, but I'm not familiar.

As long as the behavior of these tooltips is fine tuned really good, to not be in the way, I think it'll work for me just great too in the end.

JMC47:
For what it's worth, I do use that laptop for debugging Linux issues sometimes, so having the GUI work on it would be rather nice.  As it is, I can use it yes, but it's unwieldy and for users who don't know what the setting mean, it can make things very hard to use.

Source/Core/DolphinQt/Config/Graphics/BalloonTip.h
```diff
@@ -0,0 +1,36 @@
...
+{
+  Q_OBJECT
+
+  struct PrivateTag
+  {
+  };
+
+public:
+  static void ShowBalloon(const QIcon& icon, const QString& title, const QString& msg,
+                          const QPoint& pos, QWidget* parent, bool show_arrow = true);
```
leoetlino, line 21: Consider replacing the boolean with an enum class (`ShowArrow` with members `Yes` and `No`) to improve readability at call sites.

If somebody passes false as the last argument to ShowBalloon, there is no way of figuring out what that boolean does without checking the function prototype and remembering what the last parameter is called.

Source/Core/DolphinQt/Config/ToolTipControls/ToolTipCheckBox.h
```diff
@@ -0,0 +1,28 @@
...
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <optional>
+
+#include <QCheckBox>
+#include <QString>
+
+class ToolTipCheckBox : public QCheckBox
```
leoetlino, line 12: All those classes seem to be extremely similar... You should probably turn this into a class template (`ToolTipWidget<T>?`) instead of having 5 classes with duplicated code.

Some of the classes have extra logic in timerEvent. I'd recommend adding a `onTimerEvent` protected member function that can be overridden by derived classes so you don't need to override timerEvent() and duplicate code. onTimerEvent would be called from ToolTipWidget's timerEvent and would do nothing by default.

Source/Core/DolphinQt/Config/Graphics/GeneralWidget.cpp
```diff
@@ -197,89 +198,123 @@ void GeneralWidget::AddDescriptions()
 // We need QObject::tr
 #if defined(_WIN32)
   static const char TR_BACKEND_DESCRIPTION[] = QT_TR_NOOP(
-      "Selects which graphics API to use internally.\n\nThe software renderer is extremely "
+      "Selects which graphics API to use internally.<br /><br />The software renderer is extremely "
```
leoetlino, line 201: Does `<br>` not work? There's no reason to use the XHTML compatible form here.

If `<br>` does work, please change all the other occurrences of `<br />` for readability reasons.

leoetlino (changes requested):

Source/Core/DolphinQt/Config/ToolTipControls/ToolTipCheckBox.h
iwubcode: I did start by looking into a template class but I can't remember why I chose not to do that, I vaguely recall switching designs toward the end so maybe that's why.  I'll give it another whirl, thanks!

iwubcode: I did it a little differently than suggested, let me know if you feel like there's a better solution

Source/Core/DolphinQt/Config/ToolTipControls/ToolTipCheckBox.h
```diff
@@ -0,0 +1,18 @@
+// Copyright 2020 Dolphin Emulator Project
+// Licensed under GPLv2+
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include "ToolTipWidget.h"
```
leoetlino, line 7: DolphinQt/Config/ToolTipControls/ToolTipWidget.h

Source/Core/DolphinQt/Config/ToolTipControls/ToolTipComboBox.h
```diff
@@ -0,0 +1,15 @@
+// Copyright 2020 Dolphin Emulator Project
+// Licensed under GPLv2+
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include "ToolTipWidget.h"
```
leoetlino, line 7: ditto

Source/Core/DolphinQt/Config/ToolTipControls/ToolTipRadioButton.h
```diff
@@ -0,0 +1,18 @@
+// Copyright 2020 Dolphin Emulator Project
+// Licensed under GPLv2+
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include "ToolTipWidget.h"
```
leoetlino, line 7: ditto

Source/Core/DolphinQt/Config/ToolTipControls/ToolTipSlider.h
```diff
@@ -0,0 +1,18 @@
+// Copyright 2020 Dolphin Emulator Project
+// Licensed under GPLv2+
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include "ToolTipWidget.h"
```
leoetlino, line 7: ditto

Source/Core/DolphinQt/Config/ToolTipControls/ToolTipSpinBox.h
```diff
@@ -0,0 +1,15 @@
+// Copyright 2020 Dolphin Emulator Project
+// Licensed under GPLv2+
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include "ToolTipWidget.h"
```
leoetlino, line 7: ditto

Source/Core/DolphinQt/Config/Graphics/GeneralWidget.cpp
```diff
@@ -197,89 +198,123 @@ void GeneralWidget::AddDescriptions()
...
-      QT_TR_NOOP("Uses the entire screen for rendering.\n\nIf disabled, a "
-                 "render window will be created instead.\n\nIf unsure, leave this unchecked.");
+      QT_TR_NOOP("Uses the entire screen for rendering.<br><br>If disabled, a "
+                 "render window will be created instead.<br><br><dolphin_emphasis>If "
+                 "unsure, leave this unchecked.</dolphin_emphasis>");
   static const char TR_AUTOSIZE_DESCRIPTION[] =
-      QT_TR_NOOP("Automatically adjusts the window size to the internal resolution.\n\nIf unsure, "
-                 "leave this unchecked.");
-
+      QT_TR_NOOP("Automatically adjusts the window size to the internal resolution.<br><br "
```
leoetlino, line 221: You missed one `<br />` ;)

Hopefully it will be wrapped in a better way with `<br>`.

leoetlino (approved):
LGTM codewise.

MayImilae:
For future reference - to disable high contrast tooltips, go to Dolphin.ini, and under [Interface] set "UseHighContrastTooltips" to False.

PR #9154, opened 2020-10-18
Android: ConfirmRunnable to RunRunnable with optional confirmation

Ebola16:
Some future runnable calls may not need confirmation.


Ebola16:
And I changed the dialog buttons to OK / Cancel to make them more applicable for future uses.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/model/view/RunRunnable.java
```diff
@@ -0,0 +1,53 @@
...
+import org.dolphinemu.dolphinemu.features.settings.model.AbstractSetting;
+
+public final class RunRunnable extends SettingsItem
+{
+  private final boolean mConfirmRunnable;
+  private final int mAlertText;
+  private final int mToastTextAfterRun;
+  private final Runnable mRunnable;
+
+  public RunRunnable(int titleId, int descriptionId, boolean confirmRunnable, int alertText,
```
JosJuice, line 12: What do you think about removing the `confirmRunnable` variable and instead making the code check `alertText != 0`, like you've done with `toastTextAfterRun`?

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/model/view/RunRunnable.java
```diff
@@ -0,0 +1,53 @@
...
+
+  public Runnable getRunnable()
+  {
+    return mRunnable;
+  }
+
+  @Override
+  public int getType()
+  {
+    return TYPE_CONFIRM_RUNNABLE;
```
JosJuice, line 45: The name of this constant should be changed as well.

JosJuice (commented):
Hmm, I wonder why git is showing the files as entirely replaced instead of renamed and slightly edited... Oh well.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/model/view/RunRunnable.java
Ebola16: Good idea. I used `alertTextID > 0` to stay consistent with the other ViewHolders.

JosJuice (approved):
Huh, now one of the two files shows up as renamed...? LGTM regardless.

Ebola16:
I'm guessing git doesn't consider a file renamed if methods are modified? That's the most striking difference between the files in question.

PR #9155, opened 2020-10-18
DolphinQt: Reset TAS input slider to default on right-click

JosJuice:
This is a feature which existed in DolphinWX. Seems like it got implemented in DolphinQt for IRWidget/StickWidget but not sliders.

### DIFF
--- a/Source/Core/DolphinQt/TAS/TASInputWindow.cpp
+++ b/Source/Core/DolphinQt/TAS/TASInputWindow.cpp
@@ -2,6 +2,8 @@
 // Licensed under GPLv2+
 // Refer to the license.txt file included.
 
+#include "DolphinQt/TAS/TASInputWindow.h"
+
 #include <cmath>
 
 #include <QCheckBox>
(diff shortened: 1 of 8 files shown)


Source/Core/DolphinQt/TAS/TASInputWindow.cpp
```diff
@@ -79,22 +81,27 @@ QGroupBox* TASInputWindow::CreateStickInputs(QString name, QSpinBox*& x_value, Q
                         .arg(name, x_shortcut_key_sequence.toString(QKeySequence::NativeText),
                              y_shortcut_key_sequence.toString(QKeySequence::NativeText)));
 
+  const int x_default = static_cast<int>(std::round(max_x / 2.));
```
jordan-woyak, line 84: You could use `std::lround` here.
But I realize you just moved this from somewhere else.
Otherwise LGTM.

JosJuice: But I would still have to cast, right? Since it returns a long and I want an int.

jordan-woyak: Yeah, I suppose it's good as is.

jordan-woyak (approved):

leoetlino (approved):

PR #9157, opened 2020-10-18
WiimoteEmu: Allow tilt to wrap around and simulate full 360 degree rotations.

jordan-woyak:
Emulated Wii remote's tilt simulation will now wrap around to reach the target angle in the shortest direction.

This can easily be triggered when the "Angle" setting is higher than 90 degrees (it's 85 by default).

A full 360 degree rotation required by some games is now possible.
Fixes: https://bugs.dolphin-emu.org/issues/12289

### DIFF
--- a/Source/Core/Core/HW/WiimoteEmu/Dynamics.cpp
+++ b/Source/Core/Core/HW/WiimoteEmu/Dynamics.cpp
@@ -120,9 +120,19 @@ void EmulateTilt(RotationalState* state, ControllerEmu::Tilt* const tilt_group, 
   const ControlState roll = target.x * MathUtil::PI;
   const ControlState pitch = target.y * MathUtil::PI;
 
+  const auto target_angle = Common::Vec3(pitch, -roll, 0);
+
+  // For each axis, wrap around current angle if target is farther than 180 degrees.
+  for (std::size_t i = 0; i != target_angle.data.size(); ++i)
+  {
+    auto& angle = state->angle.data[i];
+    if (std::abs(angle - target_angle.data[i]) > float(MathUtil::PI))
+      angle -= std::copysign(MathUtil::TAU, angle);
+  }
+
   const auto max_accel = std::pow(tilt_group->GetMaxRotationalVelocity(), 2) / MathUtil::TAU;
 
-  ApproachAngleWithAccel(state, Common::Vec3(pitch, -roll, 0), max_accel, time_elapsed);
+  ApproachAngleWithAccel(state, target_angle, max_accel, time_elapsed);
 }
 
 void EmulateSwing(MotionState* state, ControllerEmu::Force* swing_group, float time_elapsed)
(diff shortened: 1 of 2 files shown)


Source/Core/Core/HW/WiimoteEmu/Dynamics.cpp
```diff
@@ -120,9 +120,19 @@ void EmulateTilt(RotationalState* state, ControllerEmu::Tilt* const tilt_group,
   const ControlState roll = target.x * MathUtil::PI;
   const ControlState pitch = target.y * MathUtil::PI;
 
+  const auto target_angle = Common::Vec3(pitch, -roll, 0);
+
+  // For each axis, wrap around current angle if target is farther than ~180 degrees.
+  for (std::size_t i = 0; i != target_angle.data.size(); ++i)
+  {
+    auto& angle = state->angle.data[i];
+    if (std::abs(angle - target_angle.data[i]) > MathUtil::PI * 1.001)
```
iwubcode, line 129: Why are you doing the multiply of 1.001?  Why not just compare against `PI`?

jordan-woyak: When "Angle" was set to 180 tilting would occur in the wrong direction. I assume it's because the floats aren't holding the exact value. I can add a comment.

iwubcode:
I don't have any games that require full 360 degree rotations (that I can recall) so I just did some regression testing.  I didn't notice anything when playing a couple titles.  I set my angle to 95°

Source/Core/DolphinQt/Config/Mapping/MappingIndicator.cpp
```diff
@@ -356,8 +356,14 @@ void TiltIndicator::Draw()
...
 
-  const auto adj_coord =
-      Common::DVec2{-m_motion_state.angle.y, m_motion_state.angle.x} / MathUtil::PI;
+  auto adj_coord = Common::DVec2{-m_motion_state.angle.y, m_motion_state.angle.x} / MathUtil::PI;
+
+  // Angle may extend beyond 180 degrees when wrapping around.
+  // Apply modulo to draw within the indicator.
+  // Scale down the value a bit so +1 does not become -1.
+  adj_coord /= 1.0001f;
+  adj_coord.x = std::fmod(adj_coord.x + 3, 2) - 1;
```
leoetlino, line 365: Why +3 and not +1? Actually, could you explain what the constants are?

jordan-woyak: Values may be as low as -2 when the wrap-around logic runs. Only adding one wouldn't bring this into the (-1,+1) range for proper display.
I can try to add a comment or make it more clear somehow.

Rumi-Larry:
Could it be possible to add an option to clamp the angles (the previous behavior) with a value decided by the user? I can foresee that some 3rd party controllers would wrongly input a full rotation without that.

jordan-woyak:
> 
> 
> Could it be possible to add an option to clamp the angles (the previous behavior) with a value decided by the user? I can foresee that some 3rd party controllers would wrongly input a full rotation without that.

It won't wrap around unless the set angle is greater than 90 degrees. I'd like to not add an option if possible. It's rare for a game to even require more than 90 degrees. The only ones I can think of will want complete rotations.

leoetlino (approved):

PR #9160, opened 2020-10-19
Linux/XInput2: Fix keys being stuck pressed on focus loss.

jordan-woyak:
This fixes the issue of "Alt" being stuck after an Alt+Tab away from Dolphin.

Keyboard state is now cleared on `XI_FocusOut`.

At least on my system, cycling Alt-Tab and refocusing Dolphin does not capture `XI_FocusOut` or `XI_KeyRelease`.
I've added a workaround with `XQueryKeymap`.

Other minor cleanups.


leoetlino (approved):
Untested, but the logic looks sane to me.

PR #9161, opened 2020-10-19
VertexShaderGen: Fix unescaped { in D3D shader.

AdmiralCurtiss:
Fixes crash from #8981 when using D3D.


lioncash:
Thanks! Was in the middle of resolving this, but you beat me to it.

lioncash (approved):

PR #9162, opened 2020-10-19
Replace stateful rotational matrices with quaternions.

jordan-woyak:
This is an attempt to fix: https://bugs.dolphin-emu.org/issues/12077

Wii remote orientations were maintained as `Matrix33`.
We never made any attempt to normalize these and floating point inaccuracies will eventually make them invalid rotational matrices.

Orientations now use quaternions and are normalized.

### DIFF
--- a/Source/Core/Common/Matrix.h
+++ b/Source/Core/Common/Matrix.h
@@ -154,6 +154,8 @@ union TVec4
   TVec4(TVec3<T> _vec, T _w) : TVec4{_vec.x, _vec.y, _vec.z, _w} {}
   TVec4(T _x, T _y, T _z, T _w) : data{_x, _y, _z, _w} {}
 
+  T Dot(const TVec4& other) const { return x * other.x + y * other.y + z * other.z + w * other.w; }
+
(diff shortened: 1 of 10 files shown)


Source/Core/Common/Matrix.h
```diff
@@ -321,11 +323,40 @@ auto operator/(TVec2<T> lhs, T2 scalar)
...
+
+  Quaternion() = default;
+  Quaternion(float w, float x, float y, float z);
+
+  float Norm() const;
+  Quaternion Normalized() const;
+  Quaternion Conjugate() const;
+  Quaternion Inverted() const;
+
+  Quaternion& operator*=(const Quaternion& rhs);
```
lioncash, line 347: Shouldn't an `operator*` also be provided as part of the interface?

jordan-woyak: It's a free function below the class definition.

iwubcode:
Yay, a much needed addition.  Excited to see this.  I'll likely update the freelook camera controllers once this is merged to use normalized quaternions.

lioncash (approved):

PR #9164, opened 2020-10-19
GeometryShaderGen: Migrate over to fmt

lioncash:
Continues migration of the shader generators over to fmt.


leoetlino (approved):
Went through this very carefully (in particular to make sure no `{` was unescaped), hopefully I haven't missed anything.

PR #9167, opened 2020-10-20
Android: Convert ISOPaths to INI settings

Ebola16:
Part of https://bugs.dolphin-emu.org/issues/10957
Simple cases are done so time to start going after the more complicated ones.

### DIFF
--- a/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/model/GameFileCache.java
+++ b/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/model/GameFileCache.java
@@ -1,22 +1,18 @@
 package org.dolphinemu.dolphinemu.model;
 
-import android.content.Context;
-import android.content.SharedPreferences;
-import android.preference.PreferenceManager;
-
 import androidx.annotation.Keep;
 
+import org.dolphinemu.dolphinemu.NativeLibrary;
 import org.dolphinemu.dolphinemu.features.settings.model.BooleanSetting;
+import org.dolphinemu.dolphinemu.features.settings.model.Settings;
+import org.dolphinemu.dolphinemu.features.settings.utils.SettingsFile;
+import org.dolphinemu.dolphinemu.utils.IniFile;
 
 import java.io.File;
-import java.util.HashSet;
-import java.util.Set;
+import java.util.LinkedHashSet;
 
 public class GameFileCache
 {
-  private static final String GAME_FOLDER_PATHS_PREFERENCE = "gameFolderPaths";
-  private static final Set<String> EMPTY_SET = new HashSet<>();
-
   @Keep
   private long mPointer;
 
@@ -30,50 +26,72 @@ public class GameFileCache
   @Override
   public native void finalize();
 
-  public static void addGameFolder(String path, Context context)
+  public static void addGameFolder(String path)
   {
-    SharedPreferences preferences = PreferenceManager.getDefaultSharedPreferences(context);
-    Set<String> folderPaths = preferences.getStringSet(GAME_FOLDER_PATHS_PREFERENCE, EMPTY_SET);
+    File dolphinFile = SettingsFile.getSettingsFile(Settings.FILE_DOLPHIN);
+    IniFile dolphinIni = new IniFile(dolphinFile);
+    LinkedHashSet<String> pathSet = getPathSet(false);
+    int totalISOPaths =
+            dolphinIni.getInt(Settings.SECTION_INI_GENERAL, SettingsFile.KEY_ISO_PATHS, 0);
 
-    if (folderPaths == null)
+    if (!pathSet.contains(path))
     {
-      return;
+      dolphinIni.setInt(Settings.SECTION_INI_GENERAL, SettingsFile.KEY_ISO_PATHS,
+              totalISOPaths + 1);
+      dolphinIni.setString(Settings.SECTION_INI_GENERAL, SettingsFile.KEY_ISO_PATH_BASE +
+              totalISOPaths, path);
+      dolphinIni.save(dolphinFile);
+      NativeLibrary.ReloadConfig();
     }
-
-    Set<String> newFolderPaths = new HashSet<>(folderPaths);
-    newFolderPaths.add(path);
-    SharedPreferences.Editor editor = preferences.edit();
-    editor.putStringSet(GAME_FOLDER_PATHS_PREFERENCE, newFolderPaths);
-    editor.apply();
   }
 
-  private void removeNonExistentGameFolders(Context context)
+  private static LinkedHashSet<String> getPathSet(boolean removeNonExistentFolders)
   {
-    SharedPreferences preferences = PreferenceManager.getDefaultSharedPreferences(context);
-    Set<String> folderPaths = preferences.getStringSet(GAME_FOLDER_PATHS_PREFERENCE, EMPTY_SET);
+    File dolphinFile = SettingsFile.getSettingsFile(Settings.FILE_DOLPHIN);
+    IniFile dolphinIni = new IniFile(dolphinFile);
+    LinkedHashSet<String> pathSet = new LinkedHashSet<>();
+    int totalISOPaths =
+            dolphinIni.getInt(Settings.SECTION_INI_GENERAL, SettingsFile.KEY_ISO_PATHS, 0);
 
-    if (folderPaths == null)
+    for (int i = 0; i < totalISOPaths; i++)
     {
-      return;
-    }
+      String path = dolphinIni.getString(Settings.SECTION_INI_GENERAL,
+              SettingsFile.KEY_ISO_PATH_BASE + i, "");
 
-    Set<String> newFolderPaths = new HashSet<>();
-    for (String folderPath : folderPaths)
-    {
-      File folder = new File(folderPath);
+      File folder = new File(path);
       if (folder.exists())
       {
-        newFolderPaths.add(folderPath);
+        pathSet.add(path);
       }
     }
 
-    if (folderPaths.size() != newFolderPaths.size())
+    if (removeNonExistentFolders && totalISOPaths > pathSet.size())
     {
-      // One or more folders are being deleted
-      SharedPreferences.Editor editor = preferences.edit();
-      editor.putStringSet(GAME_FOLDER_PATHS_PREFERENCE, newFolderPaths);
-      editor.apply();
+      int setIndex = 0;
+
+      dolphinIni.setInt(Settings.SECTION_INI_GENERAL, SettingsFile.KEY_ISO_PATHS,
+              pathSet.size());
+
+      // One or more folders have been removed.
+      for (String entry : pathSet)
+      {
+        dolphinIni.setString(Settings.SECTION_INI_GENERAL, SettingsFile.KEY_ISO_PATH_BASE +
+                setIndex, entry);
+
+        setIndex++;
+      }
+
+      // Delete known unnecessary keys. Ignore i values beyond totalISOPaths.
+      for (int i = setIndex; i < totalISOPaths; i++)
+      {
+        dolphinIni.deleteKey(Settings.SECTION_INI_GENERAL, SettingsFile.KEY_ISO_PATH_BASE + i);
+      }
+
+      dolphinIni.save(dolphinFile);
+      NativeLibrary.ReloadConfig();
     }
+
+    return pathSet;
   }
 
   /**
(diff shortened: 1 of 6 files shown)


Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/model/GameFileCache.java
```diff
@@ -28,67 +22,89 @@ public GameFileCache(String path)
...
     boolean recursiveScan = BooleanSetting.MAIN_RECURSIVE_ISO_PATHS.getBooleanGlobal();
 
-    removeNonExistentGameFolders(context);
-
-    SharedPreferences preferences = PreferenceManager.getDefaultSharedPreferences(context);
-    Set<String> folderPathsSet = preferences.getStringSet(GAME_FOLDER_PATHS_PREFERENCE, EMPTY_SET);
+    LinkedHashSet<String> folderPathsSet = getPathSet(true);
 
-    if (folderPathsSet == null)
+    if (folderPathsSet.isEmpty())
```
JosJuice, line 107: This check can be removed, and I believe it in fact has to be removed in order to correctly handle the case where all folders have gotten deleted from the paths set.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/model/GameFileCache.java
```diff
@@ -28,67 +22,89 @@ public GameFileCache(String path)
...
 
-    if (folderPaths == null)
+    if (!pathSet.contains(path))
     {
-      return;
+      dolphinIni.setInt(Settings.SECTION_INI_GENERAL, SettingsFile.KEY_ISO_PATHS,
+              totalISOPaths + 1);
+      dolphinIni.setString(Settings.SECTION_INI_GENERAL, SettingsFile.KEY_ISO_PATH_BASE +
+              totalISOPaths, path);
+      dolphinIni.save(dolphinFile);
```
JosJuice, line 39: Please call `NativeLibrary.ReloadConfig()` after saving here, otherwise the core might steamroll over the changes you just did once these settings have been migrated to the new config system.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/model/GameFileCache.java
```diff
@@ -28,67 +22,89 @@ public GameFileCache(String path)
...
+        setIndex = setIndex + 1;
+      }
+
+      // Delete known unnecessary keys. Ignore i values beyond index-adjusted totalISOPaths.
+      for (int i = totalISOPaths - pathSet.size() - 1; i < totalISOPaths; i++)
+      {
+        dolphinIni.deleteKey(Settings.SECTION_INI_GENERAL, SettingsFile.KEY_ISO_PATH_BASE + i);
+      }
+
+      dolphinIni.save(dolphinFile);
```
JosJuice, line 90: Same here.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/model/GameFileCache.java
```diff
@@ -28,67 +22,89 @@ public GameFileCache(String path)
...
+      dolphinIni.setInt(Settings.SECTION_INI_GENERAL, SettingsFile.KEY_ISO_PATHS,
+              pathSet.size());
+
+      // One or more folders have been removed.
+      for (String entry : pathSet)
+      {
+        dolphinIni.setString(Settings.SECTION_INI_GENERAL, SettingsFile.KEY_ISO_PATH_BASE +
+                setIndex, entry);
+
+        setIndex = setIndex + 1;
```
JosJuice, line 81: `setIndex++` (or `++setIndex`)

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/model/GameFileCache.java
```diff
@@ -28,67 +22,89 @@ public GameFileCache(String path)
...
+      }
+
+      // Delete known unnecessary keys. Ignore i values beyond index-adjusted totalISOPaths.
+      for (int i = totalISOPaths - pathSet.size() - 1; i < totalISOPaths; i++)
```
JosJuice, line 85: The start value calculation seems incorrect to me. And even if it is correct, I believe it can be replaced with simply `int i = setIndex` (or you could use `setIndex` directly instead of creating a new variable `i`).

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/model/GameFileCache.java
JosJuice: Calling `ReloadConfig` would also make it possible for us to change `GameFileCache.update` to read from the in-memory config instead of needing to pass an array with paths. (I can do that later in a separate PR.)

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/model/GameFileCache.java
Ebola16: I guess the original intention of that check was to skip updating the cache if that update would not show games. I agree that is something we'd want to change.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/model/GameFileCache.java
Ebola16: Good idea. From what I tested (before the removal of the above empty set check), this is equivalent to `setIndex` but I didn't think to use that so I will.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/model/GameFileCache.java
JosJuice: I think the original intention was just to avoid a null pointer dereference. Note that it only checked for the null case and not for the empty case.

Ebola16: Oh right, I forgot I changed the initialization of the set.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/model/GameFileCache.java
```diff
@@ -28,70 +23,89 @@ public GameFileCache(String path)
...
+      // One or more folders have been removed.
+      for (String entry : pathSet)
+      {
+        dolphinIni.setString(Settings.SECTION_INI_GENERAL, SettingsFile.KEY_ISO_PATH_BASE +
+                setIndex, entry);
+
+        setIndex++;
+      }
+
+      // Delete known unnecessary keys. Ignore i values beyond totalISOPaths.
```
JosJuice, line 86: What is meant by "Ignore i values beyond totalISOPaths"?

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/model/GameFileCache.java
```diff
@@ -28,70 +23,89 @@ public GameFileCache(String path)
...
-    SharedPreferences preferences = PreferenceManager.getDefaultSharedPreferences(context);
-    Set<String> folderPaths = preferences.getStringSet(GAME_FOLDER_PATHS_PREFERENCE, EMPTY_SET);
+    File dolphinFile = SettingsFile.getSettingsFile(Settings.FILE_DOLPHIN);
+    IniFile dolphinIni = new IniFile(dolphinFile);
+    LinkedHashSet<String> pathSet = new LinkedHashSet<>();
+    int totalISOPaths =
+            dolphinIni.getInt(Settings.SECTION_INI_GENERAL, SettingsFile.KEY_ISO_PATHS, 0);
 
-    if (folderPaths == null)
+    if (totalISOPaths < 1)
```
JosJuice, line 53: This check is unnecessary.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/model/GameFileCache.java
```diff
@@ -28,70 +23,89 @@ public GameFileCache(String path)
...
-    Set<String> newFolderPaths = new HashSet<>();
-    for (String folderPath : folderPaths)
+    for (int i = 0; i < totalISOPaths; i++)
     {
-      File folder = new File(folderPath);
+      String path = dolphinIni.getString(Settings.SECTION_INI_GENERAL,
+              SettingsFile.KEY_ISO_PATH_BASE + i, "");
+
+      File folder = new File(path);
       if (folder.exists())
```
JosJuice, line 64: What do you think about changing this to `if (!removeNonExistentFolders || folder.exists())`? It seems like `addGameFolder` is expecting the `getPathSet(false)` call to return everything that's in the INI file.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/model/GameFileCache.java
Ebola16: I wanted to make it clear that we don't care about keys beyond the value of totalISOPaths. Since we don't read them, there's no reason to check for their existence. It's possible that unread `ISOPath#` keys exist if a user manually reduces the value of `ISOPaths`. I can remove that sentence if you think it causes confusion though.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/model/GameFileCache.java
Ebola16: It's in case a user manually places a negative number for the value of `ISOPaths`.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/model/GameFileCache.java
JosJuice: I'm fine with either keeping or removing it. Your choice.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/model/GameFileCache.java
JosJuice: Ah, I didn't think of that. It still seems unnecessary to me, though. The `i < totalISOPaths` and `totalISOPaths > pathSet.size()` checks will always be false if `totalISOPaths` is 0 or less.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/model/GameFileCache.java
Ebola16: I think I'll keep it.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/model/GameFileCache.java
Ebola16: Good point.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/model/GameFileCache.java
Ebola16: If folders that don't exist are added to the set when `addGameFolder` is called, won't that be unnecessary items for `!pathSet.contains(path)` to search through?

JosJuice: It depends on what we consider unnecessary. The only case where the distinction should matter is when we try to add a folder that the system considers not to exist. If that is a case we care about, then folders that don't exist need to be in `pathSet` so that we don't get duplicates. It would probably be pretty hard to trigger this case in practice with the ways `addGameFolder` currently are called, though...

Ebola16: `FilePicker` (in `MODE_DIR`) only allows folders to be passed to `addGameFolder` so I think we shouldn't doubt the existence of the to-be-added folder. Unless the system that will replace `FilePicker` for shared storage doesn't check a folder's existence before returning it, I still don't see a reason to not minimize the size of the set.

JosJuice: Yeah, something really unlikely would have to happen, like the folder getting deleted right after the file picker is done running. I was thinking more about if we decide to call `addGameFolder` in new ways in the future. Whoever makes such a change in some time from now probably won't think of fixing this problem, since finding out about it requires reading the implementation of `addGameFolder` fairly closely.

Ebola16: Even if that was the case, `scanLibrary` is called after `addGameFolder` (in addition to MainActivity's `onResume`). Non-existent files in `ISOPath#` won't survive very long. Correct me if I'm making a bad design decision here but this just seems like a scenario that's adequately handled elsewhere.

JosJuice: Fair enough then.

JosJuice:
Normally I would have approved this PR for merge now, but... The thing is, we're going to need to switch game list scanning over to the Storage Access Framework at some point soon, and it would be convenient if we cleared out the user's ISO paths when doing so, which this PR happens to accomplish. So I would like to wait with merging this and include it in the PR that switches game list scanning over to SAF.

Ebola16:
That's fine. After the Wii Overlay and `SharedPreferences` PRs (minus this one), I planned on looking into scoped storage. If my phone gets the ability to update to Android 10, I hope to remember to push it off until it isn't an issue. I only want to mess with scoped storage on emulator.

JosJuice:
You shouldn't lose out on anything by updating, because Android 10 lets you control whether to enable scoped storage using this application manifest attribute:

https://github.com/dolphin-emu/dolphin/blob/7e197186b97d35e90e2892e96044222ade6eed7b/Source/Android/app/src/main/AndroidManifest.xml#L35

I'm also planning to look into scoped storage, but I'm not sure when I'll have time.

Ebola16:
Oh that's good news!

Ebola16:
Combined with #9318

PR #9168, opened 2020-10-20
DataReader: Make use of if constexpr where applicable

lioncash:
We can make use of if constexpr for cases where booleans will always statically be known at compile-time.


leoetlino (approved):
LGTM, I believe this also fixes a possible MSVC warning.

PR #9170, opened 2020-10-20
Android: Remove hacks for Wii Remote extension setting, round 2

JosJuice:
It's still not exactly pretty, but now all the mess is contained in one place and has a proper interface.

Fixes https://bugs.dolphin-emu.org/issues/12218.


Ebola16 (approved):
This is better than what we had. Reminder to self to update https://github.com/dolphin-emu/dolphin/pull/8894 if this gets merged first.

PR #9174, opened 2020-10-20
Common: Move OSThreads into Core

lioncash:
Common shouldn't be depending on APIs in Core (in this case, depending on the PowerPC namespace). Because of the poor separation here, this moves OSThread functionality into core, so that it resolves the implicit dependency on core.


lioncash:
Windows buildbots don't seem to be clearing their CMake cache, so its assuming the moved files are still present in their original location.

JosJuice:
The Windows buildbots are not using CMake. You need to update the VS project files.

lioncash:
Right, forgot about that.

leoetlino (approved):
whoops, looks like we both missed that in our previous reviews

PR #9175, opened 2020-10-20
ARDecrypt: Resolve seedtable at compile-time

lioncash:
We can boot the formation of the table off to the compiler at compile-time, rather than building the table at run-time.

### DIFF
--- a/Source/Core/Core/ARDecrypt.cpp
+++ b/Source/Core/Core/ARDecrypt.cpp
@@ -9,6 +9,7 @@
 #include "Core/ARDecrypt.h"
 
 #include <algorithm>
+#include <array>
 #include <cstring>
 #include <string>
 #include <vector>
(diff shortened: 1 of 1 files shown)


jordan-woyak (approved):

Source/Core/Core/ARDecrypt.cpp
```diff
@@ -176,40 +179,35 @@ static void generateseeds(u32* seeds, const u8* seedtable, u8 doreverse)
...
-      tmp3 = seeds[i];
-      seeds[i] = seeds[j - 1];
-      seeds[j - 1] = tmp3;
-
-      tmp3 = seeds[i + 1];
-      seeds[i + 1] = seeds[j];
-      seeds[j] = tmp3;
-      j -= 2;
-    }
+    u32 tmp3 = seeds[i];
```
leoetlino, line 199: Possible followup PR: use std::swap to get rid of `tmp3`?

leoetlino (approved):
LGTM

Source/Core/Core/ARDecrypt.cpp
lioncash: Unfortunately std::swap isn't constexpr in C++17, so this can't be done. This would be a candidate whenever we move to C++20 though.

PR #9178, opened 2020-10-21
DiscIO: Migrate logging over to fmt

lioncash:
Eliminates quite a bit of the PRI* macros used for handling 64-bit values.

### DIFF
--- a/Source/Core/DiscIO/DiscScrubber.cpp
+++ b/Source/Core/DiscIO/DiscScrubber.cpp
@@ -5,9 +5,7 @@
 #include "DiscIO/DiscScrubber.h"
 
 #include <algorithm>
-#include <cinttypes>
 #include <cstddef>
-#include <cstdio>
 #include <memory>
 #include <optional>
 #include <string>
@@ -16,7 +14,6 @@
 #include "Common/Align.h"
 #include "Common/Assert.h"
 #include "Common/CommonTypes.h"
-#include "Common/File.h"
 #include "Common/Logging/Log.h"
 
 #include "DiscIO/DiscExtractor.h"
@@ -59,7 +56,7 @@ void DiscScrubber::MarkAsUsed(u64 offset, u64 size)
   u64 current_offset = Common::AlignDown(offset, CLUSTER_SIZE);
   const u64 end_offset = offset + size;
 
-  DEBUG_LOG(DISCIO, "Marking 0x%016" PRIx64 " - 0x%016" PRIx64 " as used", offset, end_offset);
+  DEBUG_LOG_FMT(DISCIO, "Marking {:#018x} - {:#018x} as used", offset, end_offset);
 
   while (current_offset < end_offset && current_offset < m_file_size)
   {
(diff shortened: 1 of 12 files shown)


Source/Core/DiscIO/DiscScrubber.cpp
```diff
@@ -59,7 +56,7 @@ void DiscScrubber::MarkAsUsed(u64 offset, u64 size)
   u64 current_offset = Common::AlignDown(offset, CLUSTER_SIZE);
   const u64 end_offset = offset + size;
 
-  DEBUG_LOG(DISCIO, "Marking 0x%016" PRIx64 " - 0x%016" PRIx64 " as used", offset, end_offset);
+  DEBUG_LOG_FMT(DISCIO, "Marking 0x{:016x} - 0x{:016x} as used", offset, end_offset);
```
JosJuice, line 59: Any thoughts about whether `{:#018x}` is better than `0x{:016x}`?

Source/Core/DiscIO/NANDImporter.cpp
```diff
@@ -105,8 +105,8 @@ void NANDImporter::FindSuperblock()
   {
     if (!memcmp(m_nand.data() + pos, "SFFS", 4))
     {
-      u32 version = Common::swap32(&m_nand[pos + 4]);
-      INFO_LOG(DISCIO, "Found superblock at 0x%zx with version 0x%x", pos, version);
+      const u32 version = Common::swap32(&m_nand[pos + 4]);
+      INFO_LOG_FMT(DISCIO, "Found superblock at 0x{:x} with version 0x{:x}", pos, version);
```
Starsam80, line 109: Can't `0x{:x}` be replaced with `{:#x}`? Minor nitpick but it appears a lot.

Source/Core/DiscIO/DiscScrubber.cpp
BhaaLseN, line 59: I've seen this being used in other places, so I'll just ask: would it make sense to use the extended format modifier here instead of putting the `0x` prefix in the message?
```suggestion
  DEBUG_LOG_FMT(DISCIO, "Marking {:#018x} - {:#018x} as used", offset, end_offset);
```

lioncash: I have no opinion either way. I'm fine with whichever variant is comfortable for most.

JosJuice: It's basically the same for me... Though I did use `#` in my PR, for what it's worth.

lioncash: Alright, I'll go with `#` then.

JosJuice: You should change the `0x{:x}` ones too in that case.

lioncash: Done

JosJuice (approved):

Source/Core/DiscIO/VolumeGC.cpp
```diff
@@ -176,7 +176,7 @@ VolumeGC::ConvertedGCBanner VolumeGC::LoadBannerFile() const
   }
   else
   {
-    WARN_LOG(DISCIO, "Invalid opening.bnr. Type: %0x Size: %0" PRIx64, banner_file.id, file_size);
+    WARN_LOG_FMT(DISCIO, "Invalid opening.bnr. Type: {:#x} Size: {:#x}", banner_file.id, file_size);
```
leoetlino, line 179:
```suggestion
    WARN_LOG_FMT(DISCIO, "Invalid opening.bnr. Type: {:#0x} Size: {:#0x}", banner_file.id, file_size);
```

BhaaLseN (approved):

leoetlino (approved):

PR #7395, opened 2018-09-03
Cmake option to enable/disable vulkan video backend

mazes-80:
Allows build system to skip vulkan video backend at compile time.
Useful for "weak" hardware were vulkan is unsupported

PR #9179, opened 2020-10-21
VideoCommon: Get rid of the global g_available_video_backends

leoetlino:
Replace it with a function-local static that is initialized on first
use. This gets rid of a global variable and removes the need for
manual initialization in UICommon.

This commit also replaces the weird find_if that looks for a non-null
unique_ptr with a simple "is vector empty" check considering that
none of the pointers can be null by construction.

---

Also simplifies GetDefaultBackendName: now we can just call
GetDefaultVideoBackend to get the default backend
and get its name by calling GetName.

### DIFF
--- a/Source/Core/DolphinQt/Config/Graphics/GeneralWidget.cpp
+++ b/Source/Core/DolphinQt/Config/Graphics/GeneralWidget.cpp
@@ -64,9 +64,11 @@ void GeneralWidget::CreateWidgets()
 
   m_video_box->setLayout(m_video_layout);
 
-  for (auto& backend : g_available_video_backends)
+  for (auto& backend : VideoBackendBase::GetAvailableBackends())
+  {
     m_backend_combo->addItem(tr(backend->GetDisplayName().c_str()),
                              QVariant(QString::fromStdString(backend->GetName())));
+  }
 
   m_video_layout->addWidget(new QLabel(tr("Backend:")), 0, 0);
   m_video_layout->addWidget(m_backend_combo, 0, 1);
@@ -159,8 +161,8 @@ void GeneralWidget::SaveSettings()
   const auto current_backend = m_backend_combo->currentData().toString().toStdString();
   if (Config::Get(Config::MAIN_GFX_BACKEND) != current_backend)
   {
-    auto warningMessage =
-        g_available_video_backends[m_backend_combo->currentIndex()]->GetWarningMessage();
+    auto warningMessage = VideoBackendBase::GetAvailableBackends()[m_backend_combo->currentIndex()]
+                              ->GetWarningMessage();
     if (warningMessage)
     {
       ModalMessageBox confirm_sw(this);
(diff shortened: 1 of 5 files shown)


Source/Core/DolphinQt/Config/Graphics/GeneralWidget.cpp
```diff
@@ -64,7 +64,7 @@ void GeneralWidget::CreateWidgets()
 
   m_video_box->setLayout(m_video_layout);
 
-  for (auto& backend : g_available_video_backends)
+  for (auto& backend : VideoBackendBase::GetAvailableBackends())
     m_backend_combo->addItem(tr(backend->GetDisplayName().c_str()),
                              QVariant(QString::fromStdString(backend->GetName())));
```
BhaaLseN, line 69: Do we want to add some curlies here while we're at it, since the contents of the loop span over multiple lines?

Source/Core/VideoCommon/VideoBackendBase.cpp
```diff
@@ -200,61 +198,58 @@ u16 VideoBackendBase::Video_GetBoundingBox(int index)
...
 std::string VideoBackendBase::GetDefaultBackendName()
 {
-#ifdef HAS_OPENGL
-  return OGL::VideoBackend::NAME;
-#elif defined(_WIN32)
-  return DX11::VideoBackend::NAME;
-#else
-  return Vulkan::VideoBackend::NAME;
-#endif
+  return GetDefaultVideoBackend()->GetName();
```
BhaaLseN, line 211: Is this a good idea when `GetDefaultVideoBackend` can _theoretically_ return `nullptr`?

leoetlino: Whoops, good catch

Source/Core/VideoCommon/VideoBackendBase.cpp
```diff
@@ -200,61 +198,59 @@ u16 VideoBackendBase::Video_GetBoundingBox(int index)
...
-#endif
+  auto* default_backend = GetDefaultVideoBackend();
+  return default_backend ? default_backend->GetName() : "";
 }
 
-void VideoBackendBase::PopulateList()
+const std::vector<std::unique_ptr<VideoBackendBase>>& VideoBackendBase::GetAvailableBackends()
 {
-  // OGL > D3D11 > D3D12 > Vulkan > SW > Null
+  static auto s_available_backends = [] {
```
iwubcode, line 217: Why do you need the lambda?

leoetlino: The lambda is just a convenient way to make sure the initialization code only runs once.

iwubcode: Ah right, it's static, that makes sense.  Thanks!

iwubcode (approved):

JosJuice (approved):

RinMaru:
Vulkan has vanished from the graphics menu after this

leoetlino:
Oh, that must be a regression from #7395... I'll investigate it

PR #9181, opened 2020-10-22
Android: Replace GetJString with fmt

Ebola16:
I'll eventually need a clean way to convert `jint` to `std::string` for properly accessing all SIDevice channels in #8894. This also makes `GetJString` unneeded.

### DIFF
--- a/Source/Android/jni/AndroidCommon/AndroidCommon.cpp
+++ b/Source/Android/jni/AndroidCommon/AndroidCommon.cpp
@@ -5,23 +5,14 @@
 #include "jni/AndroidCommon/AndroidCommon.h"
 
 #include <string>
-#include <string_view>
 #include <vector>
+
+#include <fmt/format.h>
 
 #include <jni.h>
 
 #include "Common/StringUtil.h"
 #include "jni/AndroidCommon/IDCache.h"
-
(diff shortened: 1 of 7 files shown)


JosJuice:
Are you sure `fmt` actually supports this?

Ebola16:
According to the buildbot, I'm not now. I'll look into this further.

Ebola16:
This isn't a replacement for `GetJString`, but `fmt::to_string(jint)` seems to be a concise way to convert `jint` to `std::string`. Since this is only needed for `SetSIDevice` in #8894, I put it in that PR.

JosJuice:
That sounds in line with what I expected. Just so you're aware, it is also possible to use `std::to_string` instead of `fmt::to_string` for converting integers to strings.

Ebola16:
I initially though converting `jint` to `std::string` would be similar to `jstring` to `std::string` in `GetJString`'s approach. Turns out I over-complicated things.

JosJuice:
Yeah, `jint` is just a typedef for `int`, so you can treat it like any normal C++ integer. Actually, all the primitive types in Java are mapped to corresponding primitive types in C++ (except for `jboolean`, which is mapped to some kind of `char` instead of `bool` IIRC).

PR #8121, opened 2019-05-26
WIP: FrameDump: Multiple Upgrades, Reimplement API, Proper VFR, CFR and TAS Support

altimumdelta:
The FFmpeg external has to be updated for any of this to actually work on Windows. Currently it is not certain yet what kind of FFmpeg integration will be chosen, it's up to the community, I have initially provided the static method here #8123 while the other delayed-dynamic method is here #7110 

There is also a testing branch with all the dependent commits combined: https://github.com/altimumdelta/dolphin/tree/FrameDumpUpgradeCOMB
Which I'll keep up to date as development continues.

---
### Merged PRs:
- [x] #8128 - Code Refactor: Rename AVIDump to FrameDump
- [x] #9182 - FrameDump Fixes & Cleanup

### Merged Improvements:
- #9182 - Changes to frame dump timing code ("PTS code")
- #9182 - Output filename format standardized with the one used by screenshots (GAMEID + Timestamp), while the original timestamp from when the frame dump was manually started will be preserved,during automatic splits.
- #9182 - Framedump will now create a new file (restart) on VI refresh-rate change and savestate load
 

### PRs Pending Review:
- [ ] #9188 - FrameDump: Remove code for older versions of libavcodec

### PRs Ready To Merge:
- nothing yet

### Work In Progress (altimumdelta):
- [ ] Error messaging standardization

### Long-term ToDo Overview:
- [ ] Support different codec/container combinations separately, with their own defaults/configs.
- [ ] more to be added as development continues

----

## Variable frame rate (VFR) and video editor compatability support - MPEG-TS / Matroska

Similarly to the recent FFV1 BGRA->BRG0 patch tackling inoptimal output, the AVI container format it self is not optimal since it doesn't really support variable frame rate (VFR) and may lead to compatability and playback issues if there are any frame rate changes during dumping. It is possible only with hacks and those create a non-standard AVI file. The initial purpose of this PR was to support it if it does happen during the existing method of correcting for game-speed, not that the dump output has to be the same as original game-speed (variable FPS) according to performance, but with such support the original behavior can be dumped optionally for some other purpose, possibly testing and troubleshooting. 

If there were any hacks done within Dolphin to compensate for AVI limitations then those would need to be looked at, and probably removed. I'm merely speculating right now as I'm not sure if any of them exist at this moment, it does seem to work fine as is.

[...]

PR #9182, opened 2020-10-22
VideoCommon: FrameDump fixes/cleanups.

jordan-woyak:
Fixed calculation of PTS values.
This should fix: https://bugs.dolphin-emu.org/issues/12228
Warnings are issued on dropped/skipped frames.

Filename of dumped video now includes the GameID and timestamp like our screenshots.
e.g. "G9SE8P_2020-10-22_11-51-30_0.avi"

Dumping previously restarted on resolution change.
It now also restarts on VI refresh-rate change and savestate load.
Note these restarts maintain the original timestamp to help with file organization/splicing.

Eliminated static variables and general cleanup in FrameDump.
Variable renames and minor cleanup of related code in RenderBase.

### DIFF
--- a/Source/Core/Core/State.cpp
+++ b/Source/Core/Core/State.cpp
@@ -213,10 +213,6 @@ static void DoState(PointerWrap& p)
   p.DoMarker("Wiimote");
   Gecko::DoState(p);
   p.DoMarker("Gecko");
-
-#if defined(HAVE_FFMPEG)
-  FrameDump::DoState();
-#endif
 }
 
 void LoadFromBuffer(std::vector<u8>& buffer)
--- a/Source/Core/VideoCommon/FrameDump.cpp
+++ b/Source/Core/VideoCommon/FrameDump.cpp
@@ -2,6 +2,8 @@
 // Licensed under GPLv2+
 // Refer to the license.txt file included.
 
+#include "VideoCommon/FrameDump.h"
+
 #if defined(__FreeBSD__)
 #define __STDC_CONSTANT_MACROS 1
 #endif
@@ -9,6 +11,7 @@
 #include <sstream>
 #include <string>
 
+#include <fmt/chrono.h>
 #include <fmt/format.h>
 
 extern "C" {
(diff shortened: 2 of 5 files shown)


lioncash (approved):

altimumdelta:
@vadosnaprimer Make a test with this and tell us what do you think, it happened to get merged rather quick.

altimumdelta:
@vadosnaprimer 
@thecoreyburton
Make a test with this and tell us what do you think, it happened to get merged rather quick. However this is not the whole thing, but it's the kinds of stuff I had on my mind as well, namely the filename creation format. 
I'll update the progress here #8121 as usual.

vadosnaprimer:
Maybe it fixes something, but I can't notice any improvements since https://github.com/dolphin-emu/dolphin/pull/8347#issuecomment-557795933 and https://github.com/dolphin-emu/dolphin/pull/8347#issuecomment-557795933 Due to this, I can't test if segmented AV desync is still there.

Also having filename base constantly changing (since timestamp is now a part of it) makes it impossible to import through avisynth automation, that would rely only on _segment number_ changing. Now it has to *guess* the next filename somehow, and I don't even know how to make it guess properly.

altimumdelta:
Yeah, but it wasn't expected anyway, this is nonetheless a step in the right direction, eventually, and it's long overdue, but I still have a couple of months of other stuff I have to finish, I'd be back at this right about now, if I wouldn't start a new PC build :)

Although I don't exactly understand why you can't test? Surely you didn't figure it all out in less than an hour?

vadosnaprimer:
I mean all the initially duplicate frames at frame 0 are missing from the dump, so there's still huge initial AV desync, which is why I can't check if it still desyncs with every new segment dumper (it was desyncing before frame 0 was broken).

altimumdelta:
Yar, I wasn't as active/focused during those times when the frame duplication and AV desync PR was going around, I forgot some of it a bit too, so I can't speak specifically right now, just that this PR I think wasn't even meaning to fix any of that, it started as a basic code cleanup with some stuff added. I think you should discuss the details of that overall issue in the #8121 ticket.

JosJuice:
#8121 (or any other PR for that matter) is not appropriate to use as a centralized ticket for everything related to frame dumping. If you want something like that, the issue tracker would be much better place, or perhaps the forums.

altimumdelta:
@JosJuice Right about now is when I started thinking the same thing, you're right, although it would be quite a chore to fill out all these issues into the bug tracker one by one, there has to be like 30 of them, give or take, IMO.

PR #9184, opened 2020-10-22
InputCommon: Migrate logging over to fmt

lioncash:
Continues the migration of the logging calls over to the fmt capable ones.

### DIFF
--- a/Source/Core/InputCommon/ControllerInterface/DualShockUDPClient/DualShockUDPProto.h
+++ b/Source/Core/InputCommon/ControllerInterface/DualShockUDPClient/DualShockUDPProto.h
@@ -242,15 +242,16 @@ struct Message
   template <class ToMsgType>
   std::optional<ToMsgType> CheckAndCastTo()
   {
-    u32 crc32_in_header = m_message.header.crc32;
+    const u32 crc32_in_header = m_message.header.crc32;
     // zero out the crc32 in the packet once we got it since that's whats needed for calculation
     m_message.header.crc32 = 0;
-    u32 crc32_calculated = CRC32(&m_message, sizeof(ToMsgType));
+    const u32 crc32_calculated = CRC32(&m_message, sizeof(ToMsgType));
     if (crc32_in_header != crc32_calculated)
     {
-      NOTICE_LOG(SERIALINTERFACE,
-                 "DualShockUDPClient Received message with bad CRC in header: got %u, expected %u",
-                 crc32_in_header, crc32_calculated);
+      NOTICE_LOG_FMT(
+          SERIALINTERFACE,
+          "DualShockUDPClient Received message with bad CRC in header: got {:08x}, expected {:08x}",
+          crc32_in_header, crc32_calculated);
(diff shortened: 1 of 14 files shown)


iwubcode:
I haven't kept up with this.  Just curious but why not just change the base log function and then search/replace the ones that broke or are non-conforming?  Not super easy I guess?

Do you plan on removing the non FMT version after all the upgrades are finished?

lioncash:
Because that would be a huge PR that touches everything all at once, which sucks to review. Having both logging paths exist at once and migrating them over to the new one allows individual PRs to be kept compact and easily reviewed quickly and more thoroughly. I'd rather people not get mentally checked out halfway through 1000+ line PR that almost entirely consists of text formatting.

Yes the `_FMT` version will eventually have the suffix dropped once everything is moved over and the old one will be removed. There's no reason to keep two logging systems in place.

jordan-woyak (approved):

Source/Core/InputCommon/ControllerInterface/DualShockUDPClient/DualShockUDPProto.h
```diff
@@ -242,15 +242,16 @@ struct Message
...
-    u32 crc32_calculated = CRC32(&m_message, sizeof(ToMsgType));
+    const u32 crc32_calculated = CRC32(&m_message, sizeof(ToMsgType));
     if (crc32_in_header != crc32_calculated)
     {
-      NOTICE_LOG(SERIALINTERFACE,
-                 "DualShockUDPClient Received message with bad CRC in header: got %u, expected %u",
-                 crc32_in_header, crc32_calculated);
+      NOTICE_LOG_FMT(
+          SERIALINTERFACE,
+          "DualShockUDPClient Received message with bad CRC in header: got {}, expected {}",
```
BhaaLseN, line 253: Just wondering, would those CRCs be better suited for output as 8-character hexadecimal? Since thats what a CRC is usually presented as in many other places (rather than the 32-bit integer that it is right now)

Source/Core/InputCommon/ControllerInterface/OSX/OSXJoystick.mm
```diff
@@ -109,8 +109,8 @@
         break;
       }
 
-      NOTICE_LOG(SERIALINTERFACE, "Unknown IOHIDElement, ignoring (Usage: %x, Type: %x)\n", usage,
-                 IOHIDElementGetType(e));
+      NOTICE_LOG_FMT(SERIALINTERFACE, "Unknown IOHIDElement, ignoring (Usage: {:x}, Type: {:x})\n",
```
BhaaLseN, line 112: Do we need that `\n` at the end? There's nothing else in that message beyond it, and every message is its own line by default anyways.

BhaaLseN (approved):
Just two tiny nits, up to you if you want to address them or not. LGTM.

lioncash:
@BhaaLseN resolved them both

leoetlino (approved):

PR #9189, opened 2020-10-23
VideoCommon: Fix Vulkan not being enabled on Windows builds

leoetlino:
Ideally this should probably be added into the VS solution files, but
it feels less confusing to keep the HAS_OPENGL and the HAS_VULKAN together.

Fixes a regression from #7395


Source/Core/VideoCommon/VideoBackendBase.cpp
```diff
@@ -26,6 +26,9 @@
 #if !defined(_WIN32) || !defined(_M_ARM64)
 #define HAS_OPENGL 1
 #endif
+#ifdef _WIN32
+#define HAS_VULKAN 1
```
iwubcode, line 30: Should be `ENABLE_VULKAN`

PR #9191, opened 2020-10-24
IP/Top: Add Android network interface

sepalani:
This PR adds for (not so old) Android devices network interface support allowing to use network functions such as `bind`. This PRs should work for Ethernet, Wi-Fi and maybe other connection.

Ready to be reviewed.

### DIFF
--- a/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/utils/NetworkHelper.java
+++ b/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/utils/NetworkHelper.java
@@ -1,0 +1,126 @@
+package org.dolphinemu.dolphinemu.utils;
+
+import android.app.Service;
+import android.content.Context;
+import android.net.ConnectivityManager;
+import android.net.LinkAddress;
+import android.net.LinkProperties;
+import android.net.Network;
+import android.net.RouteInfo;
+import android.os.Build;
+
+import androidx.annotation.Keep;
+import androidx.annotation.RequiresApi;
+
+import org.dolphinemu.dolphinemu.DolphinApplication;
+
+import java.net.Inet4Address;
+import java.net.InetAddress;
+import java.net.UnknownHostException;
+import java.util.List;
+
+public class NetworkHelper
+{
+  private static ConnectivityManager GetConnectivityManager()
+  {
+    Context context = DolphinApplication.getAppContext();
+    ConnectivityManager manager =
+            (ConnectivityManager) context.getSystemService(Service.CONNECTIVITY_SERVICE);
+    if (manager == null)
+      Log.warning("Cannot get Network link as ConnectivityManager is null.");
+    return manager;
+  }
+
+  @RequiresApi(api = Build.VERSION_CODES.M)
+  private static LinkAddress GetIPv4Link()
+  {
+    ConnectivityManager manager = GetConnectivityManager();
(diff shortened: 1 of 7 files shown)


Source/Core/Core/IOS/Network/IP/Top.cpp
```diff
@@ -254,8 +259,21 @@ static std::optional<DefaultInterface> GetSystemDefaultInterface()
                               get_addr(iface->ifa_broadaddr)};
     }
   }
+#else
+  u32 addr = 0;
+  u32 netmask = 0;
+  u32 gateway = 0;
+  // The thread prevent a stack corruption from JVM->AttachCurrentThread
+  std::thread([&] {
```
JosJuice, line 267: You should be able to remove this since the last commit of PR #9078 added a different workaround for the problem.

Ebola16:
I've never worked with anything network related so I'm probably not the best person to review this PR. Feel free to mention me if you need help with testing something Android network-related in the future though.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/utils/NetworkHelper.java
```diff
@@ -0,0 +1,109 @@
...
+import java.net.Inet4Address;
+import java.net.InetAddress;
+import java.net.UnknownHostException;
+import java.util.List;
+
+public class NetworkHelper
+{
+  private static ConnectivityManager GetConnectivityManager()
+  {
+    Context context = NativeLibrary.getEmulationActivity();
```
JosJuice, line 23: Is there a reason why you need the `EmulationActivity` context specifically? You should be able to use `DolphinApplication.getAppContext()` otherwise.

Source/Core/Core/IOS/Network/IP/Top.cpp
```diff
@@ -254,8 +258,15 @@ static std::optional<DefaultInterface> GetSystemDefaultInterface()
                               get_addr(iface->ifa_broadaddr)};
     }
   }
+#else
```
JosJuice, line 261: It would be tidier to invert the `!defined(__ANDROID__)` condition above and put the Android code below the Windows code.

JosJuice (commented):
Can't comment on the actual network stuff, but the JNI bindings look fine.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/utils/NetworkHelper.java
```diff
@@ -0,0 +1,109 @@
...
+import java.net.InetAddress;
+import java.net.UnknownHostException;
+import java.util.List;
+
+public class NetworkHelper
+{
+  private static ConnectivityManager GetConnectivityManager()
+  {
+    Context context = DolphinApplication.getAppContext();
+    if (context == null)
```
JosJuice, line 24: It's no longer possible for this to be null. That's the advantage of not using EmulationActivity here.

sepalani:
@JosJuice Done. I also added some missing null checks which prevents a crash when there is no active network.

leoetlino:
Could you add the Keep annotations like https://github.com/dolphin-emu/dolphin/pull/9256 did?

sepalani:
@leoetlino Done.

leoetlino (approved):

PR #9192, opened 2020-10-24
Android: Convert SharedPreferences to INI Settings Part 2

Ebola16:
Not all PRs are merged yet, but this is the final part of https://bugs.dolphin-emu.org/issues/10957
Prerequisite for https://bugs.dolphin-emu.org/issues/10720

From now on, please use INI settings to retain settings across Dolphin builds instead of `SharedPreferences`.


Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/overlay/InputOverlay.java
```diff
@@ -807,61 +812,48 @@ else if (mPreferences.getInt("wiiController", 3) == 4)
...
+  private void saveControlPosition(int buttonId, int x, int y, int controller,
           String orientation)
   {
-    final SharedPreferences sPrefs = PreferenceManager.getDefaultSharedPreferences(getContext());
-    SharedPreferences.Editor sPrefsEditor = sPrefs.edit();
-    sPrefsEditor.putFloat(getXKey(sharedPrefsId, controller, orientation), x);
-    sPrefsEditor.putFloat(getYKey(sharedPrefsId, controller, orientation), y);
-    sPrefsEditor.apply();
+    FloatSetting.valueOf(getXKey(buttonId, controller, orientation)).setFloat(mSettings, x);
+    FloatSetting.valueOf(getYKey(buttonId, controller, orientation)).setFloat(mSettings, y);
```
JosJuice, line 819: It seems like we're only ever storing and loading ints with these settings even though they're floats. If we're going to do it like that, the settings should probably be IntSettings instead, but... What do you think about taking the opportunity to change the ways these settings are stored? If we store a float from 0 to 1 (where 0 is one edge of the screen and 1 is the other) instead of storing a number of pixels, you won't have the problem of controls ending up outside of the screen if you move settings between different devices or if you switch between split-screen mode and regular mode. (Though, getting it working properly in split-screen mode would also require us to stop using `DisplayMetrics`, which should probably be done in a separate PR so as to not have too much stuff in one PR.)

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/model/view/RumbleBindingSetting.java
```diff
@@ -46,8 +47,8 @@ private void saveRumble(Settings settings, InputDevice device)
     }
     else
     {
-      setValue(settings, "",
-              DolphinApplication.getAppContext().getString(R.string.rumble_not_found));
+      Toast.makeText(DolphinApplication.getAppContext(), R.string.rumble_not_found,
+              Toast.LENGTH_SHORT).show();
```
JosJuice, line 51: Was changing this from saving a setting to showing a toast intentional?

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/ui/viewholder/InputBindingSettingViewHolder.java
```diff
@@ -37,13 +32,12 @@ protected void findViews(View root)
...
-    SharedPreferences sharedPreferences = PreferenceManager.getDefaultSharedPreferences(mContext);
-
     mItem = (InputBindingSetting) item;
 
     mTextSettingName.setText(mItem.getNameId());
-    mTextSettingDescription
-            .setText(sharedPreferences.getString(mItem.getKey() + mItem.getGameId(), ""));
+    mTextSettingDescription.setText(AdHocStringSetting
+            .getStringGlobal(Settings.FILE_DOLPHIN, Settings.SECTION_BINDINGS,
+                    mItem.getKey() + "Description", ""));
```
JosJuice, line 40: I think it would make sense to create a method in `InputBindingSetting` that does the call to `getString`, so that the code here doesn't need to hardcode any section names or such.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/overlay/InputOverlay.java
JosJuice: Also, please call `saveSettings` after the user has changed something.

JosJuice:
I'm unsure if I want to merge the second commit. The thing is, if the Android input overhaul is going to happen soon (putting the controller mappings directly in GCPadNew.ini/WiimoteNew.ini like on PC), there's not much reason to switch from one half-assed way of storing the mappings to a slightly less half-assed way of storing the mappings now (since each change in how the mappings are stored requires users to re-map everything)... But the problem is that I don't know when I'll get around to doing the overhaul :(

Either way, I think it would be best to split these two commits into two PRs.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/model/view/RumbleBindingSetting.java
Ebola16: Yes. I don't see a reason to save this string since it would be formatted differently than the rest of the binding settings. I think it makes more sense to just notify the user if rumble isn't found and not modify any settings.

Ebola16:
In regards to the second commit, I was going off of this TODO.
https://github.com/dolphin-emu/dolphin/blob/3cd4c56645541640f61a761dfde16464b1ef09df/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/utils/DirectoryInitialization.java#L153
I assume that the part about not extracting the INIs should be changed if we want to persistently modify them. Also, would the input overhaul still need to use two settings for native code and user-readable code on Android? This PR handles that and prevents loss of binding settings from cleared `SharedPreferences`. Ideally those settings should be able to be copied into their new location if further changes are made. I think the second commit is still worth merging at his point.

JosJuice:
> I assume that the part about not extracting the INIs should be changed if we want to persistently modify them. Perhaps replace "extract" with "overwrite"?

I'm not entirely sure what you mean here, but that TODO is referring to the same overhaul I was referring to in my previous commit. If we do that overhaul, we will have no reason to include a pre-made GCPadNew.ini and WiimoteNew.ini with the app, just like on PC.

> Also, would the input overhaul still need to use two settings for native code and user-readable code on Android?

Why are these two separate anyway? There's barely any difference. Either way, if we would want to store these two separately after the overhaul, we'd need to add some hacks on top of how the input system works on PC, which I wouldn't be super happy about.

Ebola16:
> I'm not entirely sure what you mean here, but that TODO is referring to the same overhaul I was referring to in my previous commit. If we do that overhaul, we will have no reason to include a pre-made GCPadNew.ini and WiimoteNew.ini with the app, just like on PC.

Oh ok, I'm just making sure I understand the intention of that overhaul.

> Why are these two separate anyway? There's barely any difference. Either way, if we would want to store these two separately after the overhaul, we'd need to add some hacks on top of how the input system works on PC, which I wouldn't be super happy about.

On my phone, a sample binding description key is saved as "qpnp_pon: Button 25" but the binding key is saved as "Device '[40 character long ID]`-Button 25". The description is much more meaningful in this case. I also don't know if I should trust all buttons from all devices returning an identical length button ID or naming scheme. The old code used two strings, one a `SharedPreference` and the other an INI setting, to accomplish this so I made them both INI settings.

JosJuice:
Ah, I didn't know the descriptor was something as cumbersome as a 40 character long ID. Makes sense that we wouldn't want to show it to users then. It would be nice to only have one way of representing devices after the overhaul, though... Maybe it would be possible to use something like name plus controller number, like on PC? Otherwise, perhaps we could store the string like "name (descriptor)" and filter out the descriptor from being displayed to the user.

Ebola16:
> Ah, I didn't know the descriptor was something as cumbersome as a 40 character long ID. Makes sense that we wouldn't want to show it to users then. It would be nice to only have one way of representing devices after the overhaul, though... Maybe it would be possible to use something like name plus controller number, like on PC? Otherwise, perhaps we could store the string like "name (descriptor)" and filter out the descriptor from being displayed to the user.

Originally I just wanted to convert these settings to INI settings while minimally modifying how those settings worked. I could look into further modifications later today or tomorrow though.

> I also don't know if I should trust all buttons from all devices returning an identical length button ID or naming scheme when retrieving the key.

Since combining the settings was beyond the scope of what I originally planned, I didn't look up the button descriptor terminology. I'll need to do that so I know how I should filter these settings in a way that works for all devices.

JosJuice:
Just to be clear, I was talking about making these changes in the future overhaul, not in the second commit of this PR. I suppose that if you want to, you could make the changes in the second commit of this PR in order to make things carry across cleanly once the overhaul is done, but if you're going to do that, please make sure you exactly match the syntax used by the C++ input parsing code.

Ebola16:
Yes, I'll look into making this as seamless as possible to transition into the input overhaul. The problem with `SharedPreferences` being lost upon reinstall is something that I think should be solved sooner rather than later.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/overlay/InputOverlay.java
Ebola16: I like the idea of changing the range from 0 to 1 but not in this PR. I made that a TODO. Bounds checking will be required for a change like that anyway so this still should be good to merge.

And I'm still confused about when `saveSettings` needs to be called. The modified value gets saved to the INI without calling `saveSettings`. I assume it's being saved somewhere in the C++ side. Why is calling `saveSettings` needed here?

JosJuice: > I like the idea of changing the range from 0 to 1 but not in this PR. I made that a TODO. Bounds checking will be required for a change like that anyway so this still should be good to merge.

If we merge this as-is, we will need to add migration code for ignoring or rescaling old values once we change it to go from 0 to 1. That's the reason I brought this up now. I don't believe we would need any kind of bounds checking if we make the change at the same time as we switch the positions over from SharedPreferences to the INI file.

> And I'm still confused about when `saveSettings` needs to be called. The modified value gets saved to the INI without calling `saveSettings`. I assume it's being saved somewhere in the C++ side. Why is calling `saveSettings` needed here?

I think I forgot that we added a call to `saveSettings` in `EmulationActivity.onSaveInstanceState`. With that call there, we don't need an extra `saveSettings` call here. But otherwise we would've needed one.

(`saveSettings` needs to be called at least once before the app process gets killed, but it doesn't have to be called on the same `Settings` instance as the one that was used for actually making the changes. Well, as long as legacy settings aren't involved, that is...)

Ebola16: > If we merge this as-is, we will need to add migration code for ignoring or rescaling old values once we change it to go from 0 to 1. That's the reason I brought this up now. I don't believe we would need any kind of bounds checking if we make the change at the same time as we switch the positions over from SharedPreferences to the INI file.

Shouldn't we have bounds checking anyway to prevent users from manually setting invalid values after switching to a 0 to 1 range? These could simply be treated as invalid values when that time comes.

> I think I forgot that we added a call to saveSettings in EmulationActivity.onSaveInstanceState. With that call there, we don't need an extra saveSettings call here. But otherwise we would've needed one.

I assumed the same thing too but I put a breakpoint in `saveSettings` and it wasn't triggered when starting emulation, changing a button position, then exiting emulation. The setting still was still saved to the INI though.

JosJuice: > Shouldn't we have bounds checking anyway to prevent users from manually setting invalid values after switching to a 0 to 1 range? These could simply be treated as invalid values when that time comes.

Well, if the user manually puts buttons outside of the screen, it's their own fault that the buttons are now outside of the screen :) This isn't one of those situations where invalid values cause crashes, and it's always possible to solve the problem by resetting the button positions. But if you want to add bounds checks anyway, I'm not going to object to it.

> I assumed the same thing too but I put a breakpoint in `saveSettings` and it wasn't triggered when starting emulation, changing a button position, then exiting emulation. The setting still was still saved to the INI though.

I actually realized after leaving my comment that the Java `saveSettings` actually isn't getting called on emulation exit the way things are right now - see PR #9230. I'm assuming that what happened in your case is that there was no call to the Java `saveSettings` but there was a call to the C++ function which `saveSettings` calls. I'm not actually sure what C++ code would've done that, though. Probably something that runs when exiting emulation...

Ebola16: > Well, if the user manually puts buttons outside of the screen, it's their own fault that the buttons are now outside of the screen :) This isn't one of those situations where invalid values cause crashes, and it's always possible to solve the problem by resetting the button positions. But if you want to add bounds checks anyway, I'm not going to object to it.

I was thinking of bounds checking similar to the latest commit. It doesn't cover every case but it should be good enough. Is this acceptable?

PR #9193, opened 2020-10-24
Android: Re-add motionControlsEnabled to clearWiimoteNewIniLinkedPreferences

JosJuice:
This was removed in PR #9152 when it should have been converted to the new setting instead, and it seems like I didn't notice when reviewing it.

I'm changing the `DirectoryInitialization` logic a little so that `clearWiimoteNewIniLinkedPreferences` also is called when there was no `WiimoteNew.ini` previously, in case the user deleted `WiimoteNew.ini` but not `Dolphin.ini`.

### DIFF
--- a/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/utils/DirectoryInitialization.java
+++ b/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/utils/DirectoryInitialization.java
@@ -78,7 +78,7 @@ public final class DirectoryInitialization
           {
             // This has to be done after calling NativeLibrary.Initialize(),
             // as it relies on the config system
-            EmulationActivity.clearWiimoteNewIniLinkedPreferences(context);
+            EmulationActivity.updateWiimoteNewIniPreferences(context);
           }
 
           directoryState = DirectoryInitializationState.DOLPHIN_DIRECTORIES_INITIALIZED;
--- a/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/utils/DirectoryInitialization.java
+++ b/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/utils/DirectoryInitialization.java
@@ -70,9 +70,16 @@ public final class DirectoryInitialization
         if (setDolphinUserDirectory(context))
         {
           initializeInternalStorage(context);
-          initializeExternalStorage(context);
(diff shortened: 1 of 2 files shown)


Ebola16:
As far as I can tell, `MAIN_MOTION_CONTROLS` doesn't modify anything in `WiimoteNew.ini`. Why does this key need to be deleted in `clearWiimoteNewIniLinkedPreferences`?

JosJuice:
This line is executed when we modify `MAIN_MOTION_CONTROLS`: https://github.com/dolphin-emu/dolphin/blob/18b2553538acf647b9e965136b035adaeef87b88/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/activities/EmulationActivity.java#L881

Ebola16 (changes requested):
Oh ok but could the value of `IMUIR/Enabled` be changed based upon `MAIN_MOTION_CONTROLS` after (re)creating WiimoteNew.ini?

JosJuice:
That's not a bad idea. Please take a look at the newly added second commit.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/utils/DirectoryInitialization.java
```diff
@@ -70,10 +70,17 @@ private static void init(Context context)
...
-          initializeExternalStorage(context);
+          boolean wiimoteIniWritten = initializeExternalStorage(context);
           NativeLibrary.Initialize();
           NativeLibrary.ReportStartToAnalytics();
 
+          if (wiimoteIniWritten)
+          {
+            // This has to be done after calling NativeLibrary.Initialize(),
+            // as it relies on the config system
+            EmulationActivity.updateWiimoteNewIniPreferences(context);
```
Ebola16, line 81: Does this need to be passed to `AfterDirectoryInitializationRunner.run`? I believe that is how we make sure this starts after `NativeLibrary.Initialize` completes.

JosJuice: That would make sense if we were outside of `DirectoryInitialization.init`, but since we're inside it, we just have to make sure that this is called after `NativeLibrary.Initialize`. (The call to it is just a few lines above this line.)

Ebola16 (approved):

PR #9195, opened 2020-10-25
DolphinQt: Add a warning to the MEM1/MEM2 override setting description.

jordan-woyak:
I've run into a few users that think more memory is just better then they can't figure out why games stop working.
Maybe a warning will help at least a little.


JosJuice:
What do you think about rewording the whole message to something like this to keep it simple and short?

```
Adjusts the amount of RAM in the emulated console.

WARNING: Enabling this will completely break many games. Only a small number of games can benefit from this.
```

(This is in part because I find the word "retail" really hard to handle as a translator...)

jordan-woyak:
@JosJuice Sounds good to me. I'll update PR.

leoetlino (approved):

PR #9187, opened 2020-10-23
Common: Migrate logging to fmt

lioncash:
Continues the migration of our code over to the fmt logger.

PR #9198, opened 2020-10-26
FileUtil: Fix format string

leoetlino:
Fixes a recent regression: https://github.com/dolphin-emu/dolphin/pull/9187#issuecomment-716835091


lioncash (approved):

brujo5:
Unfortunately the crash continues. As soon as I open the emulator, it closes. At least in my system since 5.0-12919. thx

PR #9199, opened 2020-10-26
Externals: Update fmt to 7.1.0

lioncash:
Updates fmt to the latest release from 6.1.2, keeping it up to date.

Involves interface improvements, alongside codegen improvements: See the release [changelogs](https://github.com/fmtlib/fmt/releases) for more info.


iwubcode:
Nice!  Have you tried enabling `FMT_USE_FULL_CACHE_DRAGONBOX` to see if it gives us any noticeable improvements and what kind of impact it has on the binary size?

lioncash:
Nope. I don't want to experiment with settings when upgrading the library. I'd like to update it first and keep it working in the predictable manner it always has, then start fiddling around with things.

iwubcode:
Oh, I wasn't suggesting you include that in this PR,  I just noticed it in the release notes and was curious if you had tried it.

lioncash:
Oh yeah, no, I haven't played around with it at all yet. I'd probably look into how it gets enabled before doing that. Depending on how it gets built, it might be preferable to move off of the header-only configuration.

leoetlino:
https://github.com/dolphin-emu/dolphin/blob/23c9179fb61782999ce0cce867cd9813fd525b51/CMakeLists.txt#L548

This should probably be modified to require version 7 so that the build doesn't fail when we start using any of the new features -- CI doesn't use the system fmt so such a build issue would be easily missed.

lioncash:
Done

PR #9203, opened 2020-10-27
Logic & UI for "Virtual Notches"

nick-michael:
This is a feature that I wanted in Dolphin for quite some time now. It should make playing games like Super Monkey Ball feel a bit closer to using a GC controller when using a controller without an Octagonal gate (e.g. Xbox, PS4, Steam e.t.c).

Adds an option in the controller config to set the size of the virtual notches. These virtual notches add a sticky behaviour that will snap the controller to the different axis.

ezgif com-gif-maker (3)

Disclaimer:
I've never written C++ before (besides Hello World about 10 years ago) so I'm completely open to suggestions for improvements. I was unable to get the linter running locally but I think Visual Studio was enforcing code styling while developing so I'm hoping it's already formatted correctly.

### DIFF
--- a/Source/Core/DolphinQt/Config/Mapping/MappingIndicator.cpp
+++ b/Source/Core/DolphinQt/Config/Mapping/MappingIndicator.cpp
@@ -169,6 +169,45 @@ QPolygonF GetPolygonFromRadiusGetter(F&& radius_getter)
   return shape;
 }
 
+// Constructs a polygon by querying a radius at varying angles:
+template <typename F>
+QPolygonF GetPolygonSegmentFromRadiusGetter(F&& radius_getter, double direction,
+                                            double segment_size, double segment_depth)
+{
+  constexpr int shape_point_count = 6;
+  QPolygonF shape{shape_point_count};
+
+  // We subtract from the provided direction angle so it's better
+  // to add Tau here to prevent a negative value instead of
+  // expecting the function call to be aware of this internal logic
+  const double center_angle = direction + MathUtil::TAU;
+  const double center_radius_outer = radius_getter(center_angle);
+  const double center_radius_inner = center_radius_outer - segment_depth;
+
+  const double lower_angle = center_angle - segment_size / 2;
+  const double lower_radius_outer = radius_getter(lower_angle);
+  const double lower_radius_inner = lower_radius_outer - segment_depth;
+
+  const double upper_angle = center_angle + segment_size / 2;
+  const double upper_radius_outer = radius_getter(upper_angle);
+  const double upper_radius_inner = upper_radius_outer - segment_depth;
+
+  shape[0] = {std::cos(lower_angle) * (lower_radius_inner),
+              std::sin(lower_angle) * (lower_radius_inner)};
+  shape[1] = {std::cos(center_angle) * (center_radius_inner),
+              std::sin(center_angle) * (center_radius_inner)};
+  shape[2] = {std::cos(upper_angle) * (upper_radius_inner),
+              std::sin(upper_angle) * (upper_radius_inner)};
+  shape[3] = {std::cos(upper_angle) * upper_radius_outer,
+              std::sin(upper_angle) * upper_radius_outer};
+  shape[4] = {std::cos(center_angle) * center_radius_outer,
+              std::sin(center_angle) * center_radius_outer};
+  shape[5] = {std::cos(lower_angle) * lower_radius_outer,
+              std::sin(lower_angle) * lower_radius_outer};
+
+  return shape;
+}
+
 // Used to check if the user seems to have attempted proper calibration.
 bool IsCalibrationDataSensible(const ControllerEmu::ReshapableInput::CalibrationData& data)
 {
@@ -208,6 +247,24 @@ bool IsPointOutsideCalibration(Common::DVec2 point, ControllerEmu::ReshapableInp
   constexpr double ALLOWED_ERROR = 1.3;
 
   return current_radius > input_radius * ALLOWED_ERROR;
+}
+
+void DrawVirtualNotches(QPainter& p, ControllerEmu::ReshapableInput& stick, QColor notch_color)
+{
+  const double segment_size = stick.GetVirtualNotchSize();
+  if (segment_size <= 0.0)
+    return;
+
+  p.setBrush(notch_color);
+  for (int i = 0; i < 8; ++i)
+  {
+    const double segment_depth = 1.0 - ControllerEmu::MINIMUM_NOTCH_DISTANCE;
+    const double segment_gap = MathUtil::TAU / 8.0;
+    const double direction = segment_gap * i;
+    p.drawPolygon(GetPolygonSegmentFromRadiusGetter(
+        [&stick](double ang) { return stick.GetGateRadiusAtAngle(ang); }, direction, segment_size,
+        segment_depth));
+  }
 }
 
 template <typename F>
@@ -300,6 +357,8 @@ void ReshapableInputIndicator::DrawReshapableInput(
   p.setBrush(gate_brush_color);
   p.drawPolygon(
       GetPolygonFromRadiusGetter([&stick](double ang) { return stick.GetGateRadiusAtAngle(ang); }));
+
+  DrawVirtualNotches(p, stick, gate_pen_color);
 
   const auto center = stick.GetCenter();
 
--- a/Source/Core/InputCommon/ControllerEmu/ControlGroup/ControlGroup.cpp
+++ b/Source/Core/InputCommon/ControllerEmu/ControlGroup/ControlGroup.cpp
@@ -28,6 +28,17 @@ ControlGroup::ControlGroup(std::string name_, std::string ui_name_, const GroupT
 {
 }
 
+void ControlGroup::AddVirtualNotchSetting(SettingValue<double>* value, double max_virtual_notch_deg)
+{
+  AddSetting(value,
+             {_trans("Virtual Notches"),
+              // i18n: The degrees symbol.
+              _trans("°"),
+              // i18n: Snap the thumbstick position to the nearest octagonal axis.
+              _trans("Snap the thumbstick position to the nearest octagonal axis.")},
+             0, 0, max_virtual_notch_deg);
+}
+
 void ControlGroup::AddDeadzoneSetting(SettingValue<double>* value, double maximum_deadzone)
 {
   AddSetting(value,
--- a/Source/Core/InputCommon/ControllerEmu/StickGate.cpp
+++ b/Source/Core/InputCommon/ControllerEmu/StickGate.cpp
@@ -48,6 +48,16 @@ std::optional<double> GetRayLineIntersection(Common::DVec2 ray, Common::DVec2 po
   return diff.Cross(-point1) / dot;
 }
 
+double GetNearestNotch(double angle, double virtual_notch_angle)
+{
+  constexpr auto sides = 8;
+  constexpr auto rounding = MathUtil::TAU / sides;
+  const auto closest_notch = std::round(angle / rounding) * rounding;
+  const auto angle_diff =
+      std::fmod(angle - closest_notch + MathUtil::PI, MathUtil::TAU) - MathUtil::PI;
+  return std::abs(angle_diff) < virtual_notch_angle / 2 ? closest_notch : angle;
+}
+
 Common::DVec2 GetPointFromAngleAndLength(double angle, double length)
 {
   return Common::DVec2{std::cos(angle), std::sin(angle)} * length;
@@ -276,15 +286,23 @@ ReshapableInput::ReshapeData ReshapableInput::Reshape(ControlState x, ControlSta
   y -= m_center.y;
 
   // TODO: make the AtAngle functions work with negative angles:
-  const ControlState angle = std::atan2(y, x) + MathUtil::TAU;
+  ControlState angle = std::atan2(y, x) + MathUtil::TAU;
 
-  const ControlState gate_max_dist = GetGateRadiusAtAngle(angle);
   const ControlState input_max_dist = GetInputRadiusAtAngle(angle);
+  ControlState gate_max_dist = GetGateRadiusAtAngle(angle);
 
   // If input radius (from calibration) is zero apply no scaling to prevent division by zero.
   const ControlState max_dist = input_max_dist ? input_max_dist : gate_max_dist;
 
   ControlState dist = Common::DVec2{x, y}.Length() / max_dist;
+
+  const double virtual_notch_size = GetVirtualNotchSize();
+
+  if (virtual_notch_size > 0.0 && dist >= MINIMUM_NOTCH_DISTANCE)
+  {
+    angle = GetNearestNotch(angle, virtual_notch_size);
+    gate_max_dist = GetGateRadiusAtAngle(angle);
+  }
 
   // If the modifier is pressed, scale the distance by the modifier's value.
   // This is affected by the modifier's "range" setting which defaults to 50%.
--- a/Source/Core/InputCommon/ControllerEmu/StickGate.h
+++ b/Source/Core/InputCommon/ControllerEmu/StickGate.h
@@ -15,6 +15,9 @@
 
 namespace ControllerEmu
 {
+// Minimum stick distance from the center before virtual notches are applied.
+constexpr ControlState MINIMUM_NOTCH_DISTANCE = 0.9;
+
 // An abstract class representing the plastic shell that limits an analog stick's movement.
 class StickGate
 {
@@ -84,6 +87,8 @@ public:
   ControlState GetInputRadiusAtAngle(double angle) const;
 
   ControlState GetDeadzonePercentage() const;
+
+  virtual ControlState GetVirtualNotchSize() const { return 0.0; };
 
   virtual ControlState GetGateRadiusAtAngle(double angle) const = 0;
   virtual ReshapeData GetReshapableState(bool adjusted) = 0;
--- a/Source/Core/InputCommon/ControllerEmu/ControlGroup/AnalogStick.cpp
+++ b/Source/Core/InputCommon/ControllerEmu/ControlGroup/AnalogStick.cpp
@@ -65,6 +65,12 @@ OctagonAnalogStick::OctagonAnalogStick(const char* name_, const char* ui_name_,
                                        ControlState gate_radius)
     : AnalogStick(name_, ui_name_, std::make_unique<ControllerEmu::OctagonStickGate>(gate_radius))
 {
+  AddVirtualNotchSetting(&m_virtual_notch_setting, 45);
+}
+
+ControlState OctagonAnalogStick::GetVirtualNotchSize() const
+{
+  return m_virtual_notch_setting.GetValue() * MathUtil::TAU / 360;
 }
 
 }  // namespace ControllerEmu
--- a/Source/Core/InputCommon/ControllerEmu/ControlGroup/AnalogStick.h
+++ b/Source/Core/InputCommon/ControllerEmu/ControlGroup/AnalogStick.h
@@ -33,6 +33,11 @@ class OctagonAnalogStick : public AnalogStick
 public:
   OctagonAnalogStick(const char* name, ControlState gate_radius);
   OctagonAnalogStick(const char* name, const char* ui_name, ControlState gate_radius);
+
+  ControlState GetVirtualNotchSize() const override;
+
+private:
+  SettingValue<double> m_virtual_notch_setting;
 };
 
 }  // namespace ControllerEmu
--- a/Source/Core/InputCommon/ControllerEmu/ControlGroup/ControlGroup.h
+++ b/Source/Core/InputCommon/ControllerEmu/ControlGroup/ControlGroup.h
@@ -82,6 +82,8 @@ public:
         std::make_unique<NumericSetting<T>>(value, details, default_value_, min_value, max_value));
   }
 
+  void AddVirtualNotchSetting(SettingValue<double>* value, double max_virtual_notch_deg);
+
   void AddDeadzoneSetting(SettingValue<double>* value, double maximum_deadzone);
 
   template <typename T>


Source/Core/DolphinQt/Config/Mapping/MappingIndicator.cpp
```diff
@@ -301,6 +326,19 @@ void ReshapableInputIndicator::DrawReshapableInput(
   p.drawPolygon(
       GetPolygonFromRadiusGetter([&stick](double ang) { return stick.GetGateRadiusAtAngle(ang); }));
 
+  const double size = stick.GetVirtualNotchDegrees() * (MathUtil::TAU / 360);
```
lioncash, line 329: This added block should likely be placed into its own function and be called here.

Source/Core/InputCommon/ControllerEmu/StickGate.cpp
```diff
@@ -275,6 +281,41 @@ ReshapableInput::ReshapeData ReshapableInput::Reshape(ControlState x, ControlSta
   x -= m_center.x;
   y -= m_center.y;
 
+  // Let's just convert everything to positive then convert back afterwards
+  const int xNegMultiplier = x >= 0 ? 1 : -1;
```
lioncash, line 285:
```suggestion
  const int x_neg_multiplier = x >= 0 ? 1 : -1;
```

ditto for `yNegMultiplier` and `absX` and `absY`

Source/Core/InputCommon/ControllerEmu/StickGate.cpp
```diff
@@ -275,6 +281,41 @@ ReshapableInput::ReshapeData ReshapableInput::Reshape(ControlState x, ControlSta
   x -= m_center.x;
   y -= m_center.y;
 
+  // Let's just convert everything to positive then convert back afterwards
```
lioncash, line 284: This code, likewise, should probably go into its own function, that returns a pair, which can then be called like so:

```cpp
std::tie(x, y) = YourFunction();
```

Source/Core/InputCommon/ControllerEmu/StickGate.cpp
```diff
@@ -48,6 +48,42 @@ std::optional<double> GetRayLineIntersection(Common::DVec2 ray, Common::DVec2 po
...
+  }
+
+  const double diff = absX - absY;
+  if (std::abs(diff) <= diagonal_notch_threshold)
+  {
+    absX -= diff / 2;
+    absY += diff / 2;
+  }
+
+  return std::tuple(absX * x_neg_multiplier, absY * y_neg_multiplier);
```
lioncash, line 84:
```suggestion
  return std::make_pair(absX * x_neg_multiplier, absY * y_neg_multiplier);
```

Source/Core/DolphinQt/Config/Mapping/MappingIndicator.cpp
```diff
@@ -210,6 +234,22 @@ bool IsPointOutsideCalibration(Common::DVec2 point, ControllerEmu::ReshapableInp
...
 
+void DrawVirtualNotches(QPainter& p, ControllerEmu::ReshapableInput& stick, QColor gate_brush_color)
+{
+  const double size = stick.GetVirtualNotchDegrees() * (MathUtil::TAU / 360);
+  if (size > 0)
+  {
+    p.setBrush(gate_brush_color.darker(125));
+    for (int i = 1; i <= 8; ++i)
+    {
+      const double segment_gap = MathUtil::TAU / 8;
```
lioncash, line 245:
```suggestion
      const double segment_gap = MathUtil::TAU / 8.0;
```

Source/Core/DolphinQt/Config/Mapping/MappingIndicator.cpp
```diff
@@ -210,6 +234,22 @@ bool IsPointOutsideCalibration(Common::DVec2 point, ControllerEmu::ReshapableInp
   return current_radius > input_radius * ALLOWED_ERROR;
 }
 
+void DrawVirtualNotches(QPainter& p, ControllerEmu::ReshapableInput& stick, QColor gate_brush_color)
```
lioncash, line 237:
```suggestion
void DrawVirtualNotches(QPainter& p, ControllerEmu::ReshapableInput& stick, QColor gate_brush_color)
```

Source/Core/InputCommon/ControllerEmu/StickGate.cpp
```diff
@@ -48,6 +48,42 @@ std::optional<double> GetRayLineIntersection(Common::DVec2 ray, Common::DVec2 po
   return diff.Cross(-point1) / dot;
 }
 
+std::tuple<double, double> ApplyVirtualNotches(double x, double y, double virtual_notch_degrees)
+{
+  // Let's just convert everything to positive then convert back afterwards
+  const int x_neg_multiplier = x >= 0 ? 1 : -1;
+  const int y_neg_multiplier = y >= 0 ? 1 : -1;
+
+  double absX = std::abs(x);
```
lioncash, line 57:
```suggestion
  double abs_x = std::abs(x);
```

ditto for abs_y

lioncash (commented):
Only a few minor nits, otherwise the changes in response to the suggestions look pretty good!

JMC47:
I do believe something like this would be very helpful in games that use notches.  It's one of the reasons I use original GameCube controllers via adapter for most games.

Source/Core/InputCommon/ControllerEmu/StickGate.cpp
```diff
@@ -48,6 +48,42 @@ std::optional<double> GetRayLineIntersection(Common::DVec2 ray, Common::DVec2 po
   return diff.Cross(-point1) / dot;
 }
 
+std::tuple<double, double> ApplyVirtualNotches(double x, double y, double virtual_notch_degrees)
```
jordan-woyak, line 51: This function would be incredibly simpler if you passed the angle and adjusted it instead of the x and y.
e.g. https://pastebin.com/4N9Frueh

You should adjust the `angle` value in `Reshape` after calculating `input_max_dist` and before calculating `gate_max_dist` to not break reshaping.
Reshaping is slightly broken in your PR (noticeable when very near the notch tip).

Rather than jumping to the exact notch angle, it might make sense to `MathUtil::Lerp` it towards the notch angle based on distance from center. (Pass the distance into `ApplyVirtualNotches`).
The notches wouldn't affect anything on a real controller until you hit the edge anyways. Also this would maintain smooth stick input.

Source/Core/InputCommon/ControllerEmu/StickGate.cpp
```diff
@@ -138,6 +175,11 @@ ControlState ReshapableInput::GetInputRadiusAtAngle(double angle) const
   return GetCalibrationDataRadiusAtAngle(m_calibration, angle);
 }
 
+ControlState ReshapableInput::GetVirtualNotchDegrees() const
```
jordan-woyak, line 178: Please make this function return radians. Any use of this value in code will need to be in radians.
Only the UI need be in degrees for mortals/users.

Source/Core/InputCommon/ControllerEmu/StickGate.cpp
```diff
@@ -118,6 +154,7 @@ std::optional<u32> SquareStickGate::GetIdealCalibrationSampleCount() const
 ReshapableInput::ReshapableInput(std::string name_, std::string ui_name_, GroupType type_)
     : ControlGroup(std::move(name_), std::move(ui_name_), type_)
 {
+  AddVirtualNotchSetting(&m_virtual_notch_setting, 45);
```
jordan-woyak, line 157: Unfortunately this will give us a "Notches" setting for things that shouldn't have it. (emulated Wii Remote's Swing, Tilt, etc.).
The solution might be to add an `OctagonAnalogStick` class inheriting `AnalogStick` and replace relevant occurrences.

Source/Core/InputCommon/ControllerEmu/ControlGroup/ControlGroup.cpp
```diff
@@ -28,6 +28,17 @@ ControlGroup::ControlGroup(std::string name_, std::string ui_name_, const GroupT
...
 }
 
+void ControlGroup::AddVirtualNotchSetting(SettingValue<double>* value, double max_virtual_notch_deg)
+{
+  AddSetting(value,
+             {_trans("Virtual Notches"),
+              // i18n: The degrees symbol.
+              _trans("°"),
+              // i18n: Virtual Notch snap for circular controller gates.
+              _trans("Virtual Notch snap for circular controller gates.")},
```
jordan-woyak, line 38: I don't really like the "for circular controller gates" bit. This feature applies to any shape of stick.

Source/Core/DolphinQt/Config/Mapping/MappingIndicator.cpp
```diff
@@ -169,6 +169,30 @@ QPolygonF GetPolygonFromRadiusGetter(F&& radius_getter)
   return shape;
 }
 
+// Constructs a polygon by querying a radius at varying angles:
+template <typename F>
+QPolygonF GetPolygonSegmentFromRadiusGetter(F&& radius_getter, double direction, double size)
```
jordan-woyak, line 174: I'm being picky, but I don't like the look this adds to the indicators. It's very.. invasive.
I think I'd rather have something more subtle, maybe a little arc drawn in the 8 corners? (this might make more sense with my comment about using `MathUtil::Lerp` to adjust the angle).
I'd be fine merging it appearance-wise as-is though to fix it later.

jordan-woyak (changes requested):
Super neat feature. I just had a few suggestions.
Your code was not too bad for being new to C++.
The main thing was `ApplyVirtualNotches` being overly complicated.

Source/Core/InputCommon/ControllerEmu/StickGate.cpp
nick-michael: So I don't fully understand exactly what you mean here but what I can gather is that Lerp would kind of "nudge" the angle towards the notch using the midpoint? I don't quite like this concept since the point of the notches would be to force an exact angle on the joystick.

Having said that I was thinking since yesterday if we want this to actually behave a bit closer to notches it should really be inactive for most of the inner section of the stick. Maybe the virtual notch "zones" should look something more like the image below (thanks to my awesome photoshop skills):
image

Essentially using the virtual notch size as a diameter for the circular sticky zones. Thoughts?

nick-michael: So I added a new commit with a version of the above. Some of the logic feels a little bit hacky and there's a couple of other small things I don't like (e.g. should I include the `MINIMUM_NOTCH_DISTANCE` inside MappingIndicator when calculating the polygon points?) but I don't feel like I have enough of an understanding to improve it correctly. Take a look at eb8443234 and let me know what you think. Here's a GIF of the new UI and zones:

ezgif com-gif-maker (2)

Source/Core/InputCommon/ControllerEmu/StickGate.cpp
```diff
@@ -276,16 +294,28 @@ ReshapableInput::ReshapeData ReshapableInput::Reshape(ControlState x, ControlSta
...
 
   // If input radius (from calibration) is zero apply no scaling to prevent division by zero.
   const ControlState max_dist = input_max_dist ? input_max_dist : gate_max_dist;
-
+  
   ControlState dist = Common::DVec2{x, y}.Length() / max_dist;
 
+  const double virtual_notch_size = GetVirtualNotchSize();
+
+  if (virtual_notch_size > 0.0 && dist >= MINIMUM_NOTCH_DISTANCE)
```
jordan-woyak, line 309: Move this whole if-block to before where `gate_max_dist` is first defined (line 300). Re-assigning it here won't fix all the resulting distance calculations.

Source/Core/DolphinQt/Config/Mapping/MappingIndicator.cpp
jordan-woyak: What?

Source/Core/InputCommon/ControllerEmu/StickGate.cpp
jordan-woyak: The indicators look slick with the chevrons. And forget what I said about `Lerp`. You've accomplished it with `MINIMUM_NOTCH_DISTANCE`. And yes, please use this constant for calculating polygon points instead of having hardcoded numbers. Without doing that the connections between this value and the indicator aren't clear and it requires intimate knowledge to update this code in the future.

Source/Core/InputCommon/ControllerEmu/StickGate.cpp
nick-michael: The problem I have is that `gate_max_dist` is used to calculate the value of `dist` which means I can't just put this block beforehand since `dist` is now used to check against `MINIMUM_NOTCH_DISTANCE`.

nick-michael:
> `NULL` isn't a special value and isn't doing what you think it is. You're basically just writing `0.0` here.
> I don't know why you changed this. Always assigning `angle` in `Rehape` was fine.

Ahh I left that in from when I was making additional changes to the `dist` value that should only be applied when inside a virtual notch. I removed that code so this can also be reverted to your original suggestion

Source/Core/InputCommon/ControllerEmu/StickGate.cpp
jordan-woyak: Ah sorry. I missed the `dist` bit. :P

Source/Core/DolphinQt/Config/Mapping/MappingIndicator.cpp
lioncash: This was likely a suggestion I meant to delete during review but accidentally left in. This can be disregarded.

Source/Core/InputCommon/ControllerEmu/StickGate.cpp
nick-michael: I've spent the morning trying to work out the best way to go about doing this but I'm really struggling here.

My idea was to move the `AddVirtualNotchSetting` and  `GetVirtualNotchSize` implementation into the `OctagonalAnalogStick`  implementation. I changed the `GetVirtualNotchSize` to be a virtual function on `StickGate` and tried to add a default implementation on the base `ReshapableInput` class but it seems like the compiler is expecting it to be implemented across the different `ControlGroup` classes. Is there no way to implement a default behaviour that is specifically overrided by the Octagonal stick? Maybe I'm going about this in the wrong way I really just don't know!

Here is a diff of my current changes:
https://diffy.org/diff/wadmi1m9haj6oqawoisk0529

And here is the error output:
```
Severity	Code	Description	Project	File	Line	Suppression State
Error	C2259	'ControllerEmu::Force': cannot instantiate abstract class	E:\Users\Nick\Documents\Code\dolphin\Build\Release\dolphin	E:\Users\Nick\Documents\Code\dolphin\Source\Core\Core\HW\WiimoteEmu\Extension\Nunchuk.cpp	49	
Error	C2259	'ControllerEmu::Force': cannot instantiate abstract class	E:\Users\Nick\Documents\Code\dolphin\Build\Release\dolphin	E:\Users\Nick\Documents\Code\dolphin\Source\Core\Core\HW\WiimoteEmu\Extension\Nunchuk.cpp	49	
Error	C2259	'ControllerEmu::Tilt': cannot instantiate abstract class	E:\Users\Nick\Documents\Code\dolphin\Build\Release\dolphin	E:\Users\Nick\Documents\Code\dolphin\Source\Core\Core\HW\WiimoteEmu\Extension\Nunchuk.cpp	52	
Error	C2259	'ControllerEmu::Tilt': cannot instantiate abstract class	E:\Users\Nick\Documents\Code\dolphin\Build\Release\dolphin	E:\Users\Nick\Documents\Code\dolphin\Source\Core\Core\HW\WiimoteEmu\Extension\Nunchuk.cpp	52	
Error	C2259	'ControllerEmu::Cursor': cannot instantiate abstract class	E:\Users\Nick\Documents\Code\dolphin\Build\Release\dolphin	E:\Users\Nick\Documents\Code\dolphin\Source\Core\Core\HW\WiimoteEmu\WiimoteEmu.cpp	210	
Error	C2259	'ControllerEmu::Cursor': cannot instantiate abstract class	E:\Users\Nick\Documents\Code\dolphin\Build\Release\dolphin	E:\Users\Nick\Documents\Code\dolphin\Source\Core\Core\HW\WiimoteEmu\WiimoteEmu.cpp	210	
Error	C2259	'ControllerEmu::Force': cannot instantiate abstract class	E:\Users\Nick\Documents\Code\dolphin\Build\Release\dolphin	E:\Users\Nick\Documents\Code\dolphin\Source\Core\Core\HW\WiimoteEmu\WiimoteEmu.cpp	211	
Error	C2259	'ControllerEmu::Force': cannot instantiate abstract class	E:\Users\Nick\Documents\Code\dolphin\Build\Release\dolphin	E:\Users\Nick\Documents\Code\dolphin\Source\Core\Core\HW\WiimoteEmu\WiimoteEmu.cpp	211	
Error	C2259	'ControllerEmu::Tilt': cannot instantiate abstract class	E:\Users\Nick\Documents\Code\dolphin\Build\Release\dolphin	E:\Users\Nick\Documents\Code\dolphin\Source\Core\Core\HW\WiimoteEmu\WiimoteEmu.cpp	212	
Error	C2259	'ControllerEmu::Tilt': cannot instantiate abstract class	E:\Users\Nick\Documents\Code\dolphin\Build\Release\dolphin	E:\Users\Nick\Documents\Code\dolphin\Source\Core\Core\HW\WiimoteEmu\WiimoteEmu.cpp	212	
```

nick-michael: I think I got it figured out (b556c798)

Source/Core/InputCommon/ControllerEmu/ControlGroup/ControlGroup.cpp
nick-michael: Changed it to:
> Snap the thumbstick position to the nearest octagonal axis.

Finding it a bit tricky to properly define this in such a short space but I think this somewhat does the job

jordan-woyak: Good enough for me.

Source/Core/InputCommon/ControllerEmu/StickGate.cpp
jordan-woyak: Oh. I didn't realize we already had an `OctagonAnalogStick` class. :P

jordan-woyak (commented):
After fixing the lint issues and squashing the commits this looks good to me.

jordan-woyak (approved):

Source/Core/InputCommon/ControllerEmu/StickGate.h
```diff
@@ -15,6 +15,9 @@
 
 namespace ControllerEmu
 {
+// Minimum stick distance from the center before virtual notches are applied.
+constexpr auto MINIMUM_NOTCH_DISTANCE = 0.9;
```
lioncash, line 19: This should likely be:

```suggestion
constexpr ControlState MINIMUM_NOTCH_DISTANCE = 0.9;
```

Given it's used with values of that type.

Source/Core/InputCommon/ControllerEmu/StickGate.h
```diff
@@ -85,6 +88,8 @@ class ReshapableInput : public ControlGroup
 
   ControlState GetDeadzonePercentage() const;
 
+  virtual ControlState GetVirtualNotchSize() { return 0.0; };
```
lioncash, line 91: Shouldn't this be:

```suggestion
  virtual ControlState GetVirtualNotchSize() const { return 0.0; };
```

Source/Core/DolphinQt/Config/Mapping/MappingIndicator.cpp
```diff
@@ -210,6 +244,24 @@ bool IsPointOutsideCalibration(Common::DVec2 point, ControllerEmu::ReshapableInp
...
 }
 
+void DrawVirtualNotches(QPainter& p, ControllerEmu::ReshapableInput& stick, QColor gate_brush_color)
+{
+  const double segment_size = stick.GetVirtualNotchSize();
+  if (segment_size <= 0.0)
+    return;
+
+  p.setBrush(gate_brush_color.darker(125));
+  for (int i = 1; i <= 8; ++i)
```
jordan-woyak, line 254: Looping from 1 to 8 is kinda unusual. I'd change this to start at 0 and stop at 8.

Source/Core/DolphinQt/Config/Mapping/MappingIndicator.cpp
```diff
@@ -210,6 +244,24 @@ bool IsPointOutsideCalibration(Common::DVec2 point, ControllerEmu::ReshapableInp
   return current_radius > input_radius * ALLOWED_ERROR;
 }
 
+void DrawVirtualNotches(QPainter& p, ControllerEmu::ReshapableInput& stick, QColor gate_brush_color)
+{
+  const double segment_size = stick.GetVirtualNotchSize();
+  if (segment_size <= 0.0)
+    return;
+
+  p.setBrush(gate_brush_color.darker(125));
```
jordan-woyak, line 253: This looks bad in dark themes. You could `p.setBrush(gate_pen_color);` in `DrawReshapableInput` right before the `DrawVirtualNotches` call instead.

Source/Core/DolphinQt/Config/Mapping/MappingIndicator.cpp
nick-michael: I originally implemented it to start at 0 but it doesn't render properly (more than likely due to the `lower_angle` being negative) - I can either add Tau to all the angles and start at 0, or start from 1 (and perhaps leave a comment there for clarity?)

jordan-woyak: Ah. Either adding tau or a comment works. Maybe I'll fix the `Radius` getters with negative angles at some point.

nick-michael: Added some code to change the angle from within the segment logic and added a comment to explain as well:
```c++
  // We subtract from the provided direction angle so it's better
  // to add Tau here to prevent a negative value instead of
  // expecting the function call to be aware of this internal logic
  const double center_angle = direction + MathUtil::TAU;
```

It may be a bit verbose but at least it gets the point across I guess 🤷‍♂️

PR #9204, opened 2020-10-27
fix emulator issue #12139

fbastos1:
modify file export to create folder in OS when explicitly extracting a directory

### DIFF
--- a/Source/Core/DiscIO/DiscExtractor.cpp
+++ b/Source/Core/DiscIO/DiscExtractor.cpp
@@ -130,13 +130,15 @@ void ExportDirectory(const Volume& volume, const Partition& partition, const Fil
                      const std::string& export_folder,
                      const std::function<bool(const std::string& path)>& update_progress)
 {
-  File::CreateFullPath(export_folder + '/');
+  const std::string export_root =
+      export_folder + (directory.IsDirectory() ? "/" + directory.GetName() + "/" : "/");
+  File::CreateFullPath(export_root);
 
   for (const FileInfo& file_info : directory)
   {
     const std::string name = file_info.GetName() + (file_info.IsDirectory() ? "/" : "");
     const std::string path = filesystem_path + name;
-    const std::string export_path = export_folder + '/' + name;
+    const std::string export_path = export_root + name;
 
     if (update_progress(path))
       return;


JosJuice:
@dolphin-emu-bot rebuild

JosJuice:
@dolphin-emu-bot rebuild

JosJuice:
@dolphin-emu-bot rebuild

Source/Core/DiscIO/DiscExtractor.cpp
```diff
@@ -130,13 +130,15 @@ void ExportDirectory(const Volume& volume, const Partition& partition, const Fil
                      const std::string& export_folder,
                      const std::function<bool(const std::string& path)>& update_progress)
 {
-  File::CreateFullPath(export_folder + '/');
+  File::CreateFullPath(export_folder +
+                       (directory.IsDirectory() ? "/" + directory.GetName() + "/" : "/"));
```
BhaaLseN, line 134: I would probably make sense to put this ternary in a local variable (perhaps even including the `export_folder` itself), rather than doing it both here and once for every iteration in the loop below.

fbastos1: Changed. Generally speaking, do we want dolphin to be more CPU-bound or have a smaller memory footprint? I realize this is probably a non-issue here since this operation isn't time-sensitive and not particularly slow to begin with, but I can imagine similar situations elsewhere in the codebase.

JosJuice: When it comes to the emulation core, being constrained by the CPU is way more common than being constrained by the RAM size, to the point where I can't remember a single time the latter has happened for anyone other than when custom texture prefetching is involved. When it comes to things other than the emulation core... Well, the RAM usage is even lower when not emulated, so using a bit of extra RAM should be no problem. Not that I believe keeping this string around actually increases the memory usage, as the alternative of reconstructing it every time still has to allocate the string for the duration of the lines where it's used.

Source/Core/DiscIO/DiscExtractor.cpp
```diff
@@ -130,13 +130,15 @@ void ExportDirectory(const Volume& volume, const Partition& partition, const Fil
                      const std::string& export_folder,
                      const std::function<bool(const std::string& path)>& update_progress)
 {
-  File::CreateFullPath(export_folder + '/');
+  std::string export_root =
```
BhaaLseN, line 133: You should be able to make this `const` as well.

fbastos1: good catch. done

BhaaLseN (approved):
Changes seem good so far, untested tho.

You do want to squash those commits into one before continuing though; you can use the following to do this (replace `upstream` with the name of the Dolphin remote and `origin` with the name of your Fork remote):
```
git fetch upstream
git reset --soft upstream/master
git commit -m "modify file export to create folder in OS when explicitly extracting a directory"
git push -f origin HEAD
```
The first command makes sure you get the latest commits from `master` (from the Dolphin repository) to work on; the second picks all your changes, switches to that `master` state and puts the changes in the staging area. The third command then commits it so you have one commit total that has all the changes; and the forth command pushes your current branch into your own remote (which should be your Fork)

JosJuice:
@dolphin-emu-bot rebuild

JosJuice (approved):

leoetlino (approved):

7aychu23:
Can i Request upper Block size on GCZ convert???

PR #7064, opened 2018-06-03
[WIP] Add Python Scripting support

Felk:
Disclaimer: This is still far away from being complete, but I figured it's better to get some early feedback on this, as for me it's still very unclear what direction this should go to be considered for a merge somewhen.

I am currently in the process of adding scripting support in the form of embedded python scripts, which will work in a simple event-callback-based fashion. A script can basially register callbacks on events at startup, and also call back into functions defined by dolphin, e.g. reading memory etc., similar to how it's done in other emulators like [Bizhawk](http://tasvideos.org/Bizhawk/LuaFunctions.html) or [vba](http://tasvideos.org/EmulatorResources/VBA/LuaScriptingFunctions.html). Some basic infrastructure to get Python running is done now (albeit it needs cleaning at some places). I'm also adding some `async/await` magic, but that's just convenience specific to python I like.

Minimal scope goals:
- [x] put the python external into a submodule and add a readme on how to collect the required files from scratch for maintenance.
- [ ] add basic memory reading/writing support
- [ ] add controller input support
- [ ] add basic support for on-screen drawing
- [ ] add emulation control (start/stop/pause/step/save/load etc.)
- [ ] add basic events
  - [ ] frameadvance
  - [ ] interrupts. VI interrupts might be especially useful for timing purposes.
- [ ] add additional features that turn out to be crucial for a minimal scope. Should come up in development and testing
- [ ] command line flag for automatic start of a script
- [ ] Reimplement [this thing](https://github.com/TwitchPlaysPokemon/pbrEngine) as an embedded python script to prove it works and is usable.

Polishing goals:
- [ ] Linux + MacOS support. right now I only got it to work on 2 Windows 10 machines. I have to figure out CMake somewhen.
- [ ] add GUI for running scripts.
- [ ] get people to try to productively use the minimal scope, and improve off of the feedback
- [ ] no global state in python implementation (use of sub-interpreters)

Open for discussion:
- [ ] Is Python a good choice? Should default scripting support be in LUA or something else instead? If Python is OK, then:
  - [ ] what about the additional file size? In its current form it would add ~4.7 MB to the zipped file.
  - [ ] should it be disabled (as in, not compiled) by default?
- [ ] I used a few C++17 features which aren't supported by the minimum requirements yet (e.g. constexpr if). Remove that, or depend on it?

[...]

PR #6117, opened 2017-10-14
Lua scripting (Gamecube only)

NickDriscoll:
This branch adds a Lua console and Lua scripting capabilities to the Dolphin Emulator.

See the documentation on how to write scripts for Dolphin [Here](https://github.com/NickDriscoll/dolphin/blob/Lua_scripting/Lua.md)

It's GameCube only for now simply because I've had to sift through the source code myself to discern how Dolphin works, and I couldn't figure out Wiimotes.

PR #3521, opened 2016-01-17
IPC_HLE: Separate declaration from implementation

lioncash:
So, yeah, the original IPC HLE code did this pretty epic thing where it stuffed whole implementations into the header file, pulling in a pile of headers. This moves the implementations to their appropriate file.

Now, I'd like to take a moment to talk about the IPC HLE network code. Nothing, can adequately describe how baffled or monumentally impressed I was to see that much code stuffed into a single header/implementation pair. I've done pretty big cleanups in the past, but goddamn.

So this puts the respective network IPC HLE devices into their own header and implementation file. A pretty cool benefit comes from it – which must have been overlooked in the original implementation of this – is that you actually know which code does what upon first glance; also, you don't expose a (in this case) metric ton of unrelated internals into other device types.

[Review on Reviewable](https://reviewable.io/reviews/dolphin-emu/dolphin/3521)

PR #9205, opened 2020-10-28
Lua scripting

terorie:
## Summary

This PR/RFC aims to add debug scripting support to the PowerPC using Lua 5.3.

My motivationg was to fix the lack of a scriptable breakpoint/debugging interface.
Dolphin's inbuilt break/logging features can be quite tedious when debugging lots of events.
Any complex runtime state modifications would require Gecko codes, and frankly that's a bit overwhelming to me.

The existing GDBStub-integration seemed incomplete and would probably require patches to binutils-gdb to add the missing Broadway registers. Therefore, I think it's best to extend the powerful Dolphin debug system. 

The idea of adding Lua to Dolphin isn't new and has been attempted in at least [fork "Dolphin-Lua-Core"](https://github.com/SwareJonge/Dolphin-Lua-Core), #3521, #6117 and #7064 so I recognize the chance of this getting merged is fairly low. However, unlike these previous attempts, this PR aims to improve debugging instead of focusing on UI / TAS / usability.

Initially, I was looking at CPython/pybind11 because LLDB uses Python too, but I didn't like having to introduce a heavy, global interpreter and fighting with the C++ template system.
Lua is a lean and performant alternative.

[Documentation here](https://github.com/terorie/dolphin/blob/luajit/docs/Scripts.md)

## Progress

This is draft code, and I hope to clean it up in the next few weeks.

- [x] Proof-of-concept Lua function hooks
- [x] 64-bit memory access and register support
- [x] ~~Lua global error handler with PanicErrorMessage~~ (`pcall` already provides this)
- [x] Lua: Investigate userdata/metatables instead of get/set functions
- [x] Support multiple scripts
- [ ] Support changing hooks within a hook
- [ ] Allow hooks to unregister themselves
- [x] Working UI
- [ ] Redirect Lua stdout to Dolphin log system
- [x] Make compile optional behind CMake flag (`-DLUA_SCRIPTS=true`)
- [ ] Unit tests
- [x] Documentation

### High-level system changes

When compiling with the "LUA_SCRIPTS" CMake flag, the following changes get introduced:
* The UI exposes Lua scripting functionality
* The ini file gets new properties regarding Lua scripting
* Dolphin dynamically links against Lua 5.3
* PowerPC interpreter + Jit code pieces check for any script-hook breakpoints in addition to the existing simple breakpoints
* Core gets a new "ScriptEngine" that maintains a set of Lua states (one for each script)

The script system will only work in debug mode. When used without debugging, it is idle and is going to appear as disabled.

### PUC Lua vs LuaJIT

An interesting debate is which flavor of Lua to use. PUC Lua 5.3 is the reference, and LuaJIT 2.0 is a performance-oriented Lua 5.1 runtime.

Initially, this PR aimed to integrate LuaJIT. However, @rlnilsen pointed out some interesting points:
* LuaJIT allocates only in the lower 2GB range which might cause portability issues.
* LuaJIT does not natively support the i64 or u64 integer types, but it ships a userdatum wrapper in the `ffi` package.
* LuaJIT is more complex than PUC Lua, and for a scripting system this additional complexity might not be warranted.
* PUC Lua 5.3 has language-native bitwise operators, LuaJIT needs functions.

### UI changes

The game properties dialog will get a new "Scripts" tab that allows adding scripts by file path, and disable/enable them individually like in the patches tab. If debugging is disabled, the tab should display a warning.

image

### Lua API

A `dolphin` Lua C module is introduced, supporting at least:
* access to `PowerPC::ppcState` where reasonable
* memory access through MMU
* logging to Dolphin
* registering "hooks" that execute at specified PC register values (script breakpoint)
* Lua "standard library" functions (is hacking by social engineering a risk when enabling those? e.g. "run my cool script `os.execute("/bin/sh")`")

Lua scripts get executed when loading the emulator before the first PPC instruction.

The attached small proof-of-concept is a script that hooks a function for loading layouts and prints out resource names.

image

The final API could make use of table indexing overloads for a syntax like so:

```lua
dolphin.mem_u8[0x80000000] = 123
dolphin.mem_u16[0x82300000] = 123
dolphin.gpr[0] = dolphin.gpr[0] + 8
```

### Script/Hook state

To keep things simple, any Lua state is ephemeral and will be gone after reloading the emulator or using save states. If Lua scripts need persistence, they can implement it themselves.

Each Lua script will get an incrementally-assigned u32 "script ID", used to identify which hook belongs to which script.

Data structures:
* Lua references stored in registry table
* `Core::ScriptEngine` with a vector of scripts with IDs and Lua states
* `PowerPC::ScriptBreakPoints` mapping addresses => list(script ID)

Loading a script
1. Increment and get new script ID
1. Register script to `Core::ScriptEngine`
1. Init Lua state and execute script

Adding a hook (removing is inverse):
1. Register hook in hooks table
1. Add breakpoint to `PowerPC::ScriptBreakPoints`

Check for script breakpoint existence (used by JIT)
1. Query `PowerPC::ScriptBreakPoints` multimap by address and count values

Execute script breakpoint
1. Query `PowerPC::ScriptBreakPoints` for script ID values
1. Execute each script

Unloading a script:
1. Removing all `PowerPC::ScriptBreakPoints` map values with matching script ID
1. Destroy Lua state
1. Unload script entry from `Core::ScriptEngine`

### PowerPC "script" breakpoints

Script hooks require changes in the PowerPC module. I chose to modify add a new `PowerPC::ScriptBreakPoints` class instead of extending the `PowerPC::BreakPoints` class. This new class saves its checkpoints in a `std::multimap<u32, ScriptFuncHandle>`, with address as key, and a handle for the Lua function as value.

The handle is compromised of a pointer to the Lua state and the Lua function reference.

As with the simple breakpoints, (I think) the JIT compiler has a performance-critical section where it has to check for each instructions whether it overlaps with a breakpoint. So the choice of a `multimap` is controversial and might be slower than using a `vector`.

The `PowerPC` directly calls out to `Core::ScriptEngine`. I'm not sure if this is right, maybe an indirection through `PPCDebugInterface` makes sense.

----

Thanks for reading through, and looking forward to your feedback!


lioncash:
Hi. There is already an attempt at this using Python that exists over at #7064. Most developers/maintainers of the project at the time were OK with scripting being added, provided Python was used instead of Lua.

lioncash:
Note: In the future, its best to ask in the developer IRC channel if certain features (that are nontrivial) are desirable before devoting time to them, so that potentially conflicting features don't occur.

terorie:
@lioncash Thanks for pointing this out. I don't actively follow Dolphin development, and since I just needed a small feature, it was faster for me to just implement it than to wait for feedback. As said above, I recognize the chance of this getting merged is fairly low.

terorie:
The basic integration works for me, and I think most code is ready for review.

Changes:
* Added local ini section for scripts
* Added script tab on game properties
* Added basic script documentation [link](https://github.com/terorie/dolphin/blob/luajit/docs/Scripts.md)
* Scripting: PowerPC state access
* Scripting: Basic memory access
* Scripting: Support instruction hooks
* Scripting: Support frame hook

Source/Core/Core/ScriptEngine.cpp
```diff
@@ -21,12 +22,17 @@
 #include "Core/PowerPC/MMU.h"
 #include "Core/PowerPC/PowerPC.h"
 
+// clang-format off
+#define DOLPHIN_LUA_METHOD(x) {.name = #x, .func = (x)}
```
BhaaLseN, line 26: Would it make sense to move this further down to the table that uses it? Plus, at least something I like to do for local macros like this one; `#undef` it after use?

rlnilsen:
I'm just curious: Why LuaJIT over PUC (regular) Lua? If I remember correctly, the speed increase isn't significant when calling small Lua scripts. PUC Lua 5.3 has bitwise operators which i guess would be useful in this context. And also has much less complex source code which probably means fewer undiscovered bugs.

terorie:
> Why LuaJIT over PUC (regular) Lua? 

@rlnilsen There wasn't much thought behind this and I'm open to changing it to PUC Lua 5.3 if that's what preferred. :+1: 
* LuaJIT has the interesting `ffi` package which ships support for u64 via userdata: https://luajit.org/ext_ffi_semantics.html This would allow `dolphin.mem_u64` (u64 memory access) and `dolphin.ppc.ps_u64` (Paired Single u64 mode)
* LuaJIT also has built-in bitwise functions, but of course, first-class support in Lua 5.3 is preferable: https://luajit.org/extensions.html
* Scripts can call out to other C libraries, if we expose `ffi` to the script running (currently this is on). But at the same time this just invites phishing / trojan scripts.
* Complexity / Bloat: On my system, `liblua5.3.so` is 236 KB and `libluajit-5.1.so` is 498 KB. Will have to be statically linked if we choose to compile LuaJIT with custom flags such as extended Lua 5.2 support (`-DLUAJIT_ENABLE_LUA52COMPAT`)
* Complexity / Bugs: I'm confident that LuaJIT is stable enough where we can assume it's bug free. LuaJIT powers a significant part of the web through Kubernetes NGINX Ingress and OpenResty (https://github.com/openresty/lua-nginx-module)
* Performance: I agree that in typical use-cases the difference is tiny. In edge cases like frequently executed instruction hooks or expensive calculations, maybe it can provide a speed up.
* Contributes to the JITception

Unrelated, something weird I noticed while debugging, is that the `lua_State *` is allocated in a low address on Jit64: `0x0000000041d9e378`. Isn't the lower 32-bit address range reserved for PowerPC? Maybe the Lua integration needs some changes to the allocator.

JosJuice:
It used to be the case that Dolphin's x86-64 JIT required certain things to be in the lower 2 GiB of memory, but this is no longer the case.

Source/Core/Core/ScriptEngine.cpp
```diff
@@ -0,0 +1,945 @@
...
+// clang-format on
+
+namespace ScriptEngine
+{
+// Namespace Lua hosts static binding code.
+namespace Lua
+{
+static const char* s_context_registry_key = "c";
```
lioncash, line 34:
```suggestion
static constexpr char s_context_registry_key[] = "c";
```

Will only save the string data into the binary rather than a pointer as well.

Source/Core/Core/ScriptEngine.cpp
```diff
@@ -0,0 +1,945 @@
...
+    return 0;
+  if (!lua_isfunction(L, 2))
+    return 0;
+  u32 addr = *addr_arg;
+
+  // Get pointer to Lua function.
+  const void* lua_ptr = lua_topointer(L, 2);
+  // Skip if function is already registered.
+  auto range = PowerPC::script_hooks.m_hooks.equal_range(addr);
+  for (auto it = range.first; it != range.second; it++)
```
lioncash, line 75: This should be braced

Source/Core/Core/ScriptEngine.cpp
```diff
@@ -0,0 +1,945 @@
...
+// Namespace Lua hosts static binding code.
+namespace Lua
+{
+static const char* s_context_registry_key = "c";
+static std::unordered_map<const void*, LuaFuncHandle> s_frame_hooks;
+
+// Casts a value on the Lua stack to an u32 and raises a Dolphin alert on dubious integer values.
+static std::optional<u32> luaL_checkaddress(lua_State* L, int numIdx)
+{
+  lua_Integer addr_arg = luaL_checkinteger(L, numIdx);
```
lioncash, line 40:
```suggestion
  const lua_Integer addr_arg = luaL_checkinteger(L, numIdx);
```

etc in other locations for local variables.

Source/Core/Core/ScriptEngine.cpp
```diff
@@ -0,0 +1,945 @@
...
+  if (addr_arg)
+  {
+    size_t len = 0;
+    if (len_arg > 0)
+      len = static_cast<size_t>(len_arg);
+    std::string cpp_str = PowerPC::HostGetString(*addr_arg, len);
+    lua_pushlstring(L, cpp_str.data(), len);
+  }
+  else
+    lua_pushlstring(L, "", 0);
```
lioncash, line 177: We brace both the if and else branch if one contains braces for consistency

Source/Core/Core/ScriptEngine.cpp
```diff
@@ -0,0 +1,945 @@
...
+      if (c == '\0')
+        break;
+    }
+    if (!chars.empty())
+      lua_pushlstring(L, chars.data(), chars.size());
+    else
+      lua_pushlstring(L, "", 0);
+  }
+  else
+    lua_pushlstring(L, "", 0);
```
lioncash, line 226: Needs braces

Source/Core/Core/ScriptEngine.cpp
```diff
@@ -0,0 +1,945 @@
...
+  if (e.length() >= 8)
+    res |= static_cast<u64>(e[7]);
+  if (e.length() >= 9)
+    res = 0;
+  return res;
+}
+
+// Lua: ppc_get(table ppc, string field) -> (any)
+// Gets a field or register on the special "ppc" table.
+static inline int ppc_get(lua_State* L)
```
lioncash, line 534:
```suggestion
static int ppc_get(lua_State* L)
```

inline does nothing here, given the internal linkage

Source/Core/Core/ScriptEngine.cpp
```diff
@@ -0,0 +1,945 @@
...
+{
+  const char* field = luaL_checkstring(L, 2);
+  std::string_view field_str(field);
+  u64 field_case = string_to_u64(field_str);
+  u32 value = 0;
+
+#define PPC_GET_REGISTER(name, v)                                                                  \
```
lioncash, line 541: This should be undef'd after all usages.

Source/Core/Core/ScriptEngine.cpp
```diff
@@ -0,0 +1,945 @@
...
+    lua_rawget(L, 1);
+    return 1 - lua_isnil(L, 1);
+  }
+  lua_pushinteger(L, static_cast<lua_Integer>(value));
+  return 1;
+}
+
+// Lua: ppc_set(table ppc, string field, number value) -> ()
+// Sets a register on the special "ppc" table.
+static inline int ppc_set(lua_State* L)
```
lioncash, line 634:
```suggestion
static int ppc_set(lua_State* L)
```

Source/Core/Core/ScriptEngine.cpp
```diff
@@ -0,0 +1,945 @@
...
+{
+  const char* field = luaL_checkstring(L, 2);
+  lua_Integer value = luaL_checkinteger(L, 3);
+
+#define PPC_SET_REGISTER(name, v)                                                                  \
```
lioncash, line 639: Undef after usages

Source/Core/Core/ScriptEngine.cpp
```diff
@@ -0,0 +1,945 @@
...
+    DOLPHIN_LUA_METHOD(mem_read),
+    DOLPHIN_LUA_METHOD(mem_write),
+    DOLPHIN_LUA_METHOD(str_read),
+    DOLPHIN_LUA_METHOD(str_write),
+    {nullptr, nullptr}
+};
+// clang-format on
+
+// Sets a custom getter and setter to the table or userdata on the top of the stack.
+static inline void new_getset_metatable(lua_State* L, lua_CFunction get, lua_CFunction set)
```
lioncash, line 747:
```suggestion
static void new_getset_metatable(lua_State* L, lua_CFunction get, lua_CFunction set)
```

Source/Core/Core/ScriptEngine.cpp
```diff
@@ -0,0 +1,945 @@
...
+  lua_pushcfunction(L, get);
+  lua_setfield(L, -2, "__index");
+  // metatable[__newindex] = set
+  lua_pushcfunction(L, set);
+  lua_setfield(L, -2, "__newindex");
+}
+
+// Registers a new array-like object with a custom get, set and ipairs
+// to the table on the top of the stack.
+static inline void register_getset_object(lua_State* L, const char* name, lua_CFunction get,
```
lioncash, line 760:
```suggestion
static void register_getset_object(lua_State* L, const char* name, lua_CFunction get,
```

Source/Core/Core/ScriptEngine.cpp
```diff
@@ -0,0 +1,945 @@
...
+                                          lua_CFunction set)
+{
+  lua_newtable(L);
+  new_getset_metatable(L, get, set);
+  lua_setmetatable(L, -2);
+  lua_setfield(L, -2, name);
+}
+
+// Registers a table with 16 userdatas representing the PS (paired single) registers.
+static inline void register_ps_registers(lua_State* L, const char* name, lua_CFunction get,
```
lioncash, line 770:
```suggestion
static void register_ps_registers(lua_State* L, const char* name, lua_CFunction get,
```

Source/Core/Core/ScriptEngine.cpp
```diff
@@ -0,0 +1,945 @@
...
+{
+  fflush(stdout);
+  delete m_ctx;
+  m_ctx = nullptr;
+}
+
+Script::Context::~Context()
+{
+  auto& hooks = PowerPC::script_hooks.m_hooks;
+  for (auto it = hooks.cbegin(); it != hooks.cend();)
```
lioncash, line 874: This should be braced

Source/Core/Core/ScriptEngine.cpp
```diff
@@ -0,0 +1,945 @@
...
+  std::set<std::string> pathsEnabled;
+  std::set<std::string> pathsDisabled;
+  localIni.GetLines(section, &lines);
+  for (const std::string& line : lines)
+  {
+    if (line.length() < 2 || line[1] != ' ')
+      continue;
+    std::string file_path = line.substr(2);
+    if (line[0] == 'y')
+      pathsEnabled.insert(file_path);
```
lioncash, line 911:
```suggestion
      pathsEnabled.insert(std::move(file_path));
```

Source/Core/Core/ScriptEngine.cpp
```diff
@@ -0,0 +1,945 @@
...
+    else if (line[0] == 'n')
+      pathsDisabled.insert(file_path);
```
lioncash, line 913:
```suggestion
      pathsDisabled.insert(std::move(file_path));
```

Source/Core/Core/ScriptEngine.cpp
```diff
@@ -0,0 +1,945 @@
...
+    const char* error_str = lua_tostring(L, -1);
+    PanicAlert("Error running Lua hook: %s", error_str);
+  }
+  lua_settop(L, frame);
+}
+
+static std::vector<Script> s_scripts;
+
+void LoadScriptSection(const std::string& section, std::vector<ScriptTarget>& scripts,
+                       IniFile& localIni)
```
lioncash, line 898:
```suggestion
                       IniFile& local_ini)
```

We use `snake_case` for variables and this file contains quite a bit that don't follow this.

Source/Core/Core/ScriptEngine.cpp
```diff
@@ -0,0 +1,945 @@
...
+}
+
+void LoadScripts()
+{
+  IniFile localIni = SConfig::GetInstance().LoadLocalGameIni();
+  std::vector<ScriptTarget> scriptTargets;
+  LoadScriptSection("Scripts", scriptTargets, localIni);
+
+  s_scripts.clear();
+  for (auto& target : scriptTargets)
```
lioncash, line 928: This should be braced

Source/Core/Core/ScriptEngine.h
```diff
@@ -0,0 +1,91 @@
...
+// External pointer to a Lua function, for use in Core/PowerPC.
+// Must not be used after unloading the script.
+class LuaFuncHandle
+{
+public:
+  LuaFuncHandle(Script::Context* ctx, const void* lua_ptr, int lua_ref)
+      : m_ctx(ctx), m_lua_ptr(lua_ptr), m_lua_ref(lua_ref)
+  {
+  }
+  inline void Execute() { m_ctx->ExecuteHook(m_lua_ref); };
```
lioncash, line 74: inline isn't necessary here, along with the three other functions.

Member functions declared and defined inside a class body are already inline by default.

Source/Core/DolphinQt/Config/ScriptsWidget.cpp
```diff
@@ -0,0 +1,122 @@
...
+{
+  m_scripts[m_list->row(item)].active = (item->checkState() == Qt::Checked);
+  SaveScripts();
+}
+
+void ScriptsWidget::OnAdd()
+{
+  QString path = QFileDialog::getOpenFileName(this, tr("Select a script"), QDir::currentPath(),
+                                              tr("Lua script (*.lua);; All Files (*)"));
+  if (!path.isEmpty())
```
lioncash, line 68:
```suggestion
  if (path.isEmpty())
      return;
```

CMake/FindLuaJit.cmake
```diff
@@ -0,0 +1,43 @@
+# from https://github.com/neovim/neovim/blob/master/cmake/FindLuaJit.cmake
+# - Try to find luajit
+# Once done this will define
+#  LUAJIT_FOUND - System has luajit
+#  LUAJIT_INCLUDE_DIRS - The luajit include directories
+#  LUAJIT_LIBRARIES - The libraries needed to use luajit
+#
+# Copyright neovim, Apache 2.0 license.
```
lioncash, line 8: Apache 2.0 is not compatible with GPLv2+ (which this project is). We cannot use this until you contact the author and get an explicit exception for using this.

CMakeLists.txt
```diff
@@ -816,6 +816,10 @@ else()
   message(STATUS "libsystemd not found, disabling traversal server watchdog support")
 endif()
 
+find_package(LuaJit)
+include_directories(${LUAJIT_INCLUDE_DIRS})
+
+
```
lioncash, line 822: unnecessary newline.

Source/Core/Core/PatchEngine.cpp
```diff
@@ -28,6 +28,12 @@
 #include "Core/PowerPC/MMU.h"
 #include "Core/PowerPC/PowerPC.h"
 
+// TODO Hacky forward decl to ScriptEngine
```
lioncash, line 31: Just include the header?

Source/Core/Core/ScriptEngine.cpp
terorie: Thanks, good point. Fixed

CMake/FindLuaJit.cmake
terorie: Hm, this is unfortunate, I'll reach out to them. In any case is it fine linking LuaJIT dynamically through pkg-config like this? Or should it be moved to `Externals` like some of the other deps?

Source/Core/Core/ScriptEngine.cpp
terorie: I believe I market most local variables as const now including pointers, I hope I didn't overdo it.

Source/Core/Core/ScriptEngine.cpp
terorie: Renamed all `camelCase` variables I could find in ScriptEngine and ScriptWidget to `snake_case`.

terorie:
@lioncash Thanks for the detailed review. Some of the `camelCase` variables came from bits of `PatchEngine.cpp` that I used as a template. Can I submit a PR for fixing the code style there?

rlnilsen:
@terorie I'm just a peripheral contributor to Dolphin and have no say in what gets merged. These are just further thoughts I hope may help in your endeavours.

>LuaJIT has the interesting `ffi` package which ships support for u64 via userdata: https://luajit.org/ext_ffi_semantics.html 

LuaJIT's 64 bit integers are boxed and are not useful as table keys (search for "cdata objects as table keys" in the link you provided). PUC Lua 5.3 has full native support for 64 bit integers.

>I'm confident that LuaJIT is stable enough where we can assume it's bug free.

You have more faith in human infallability than I.

> Unrelated, something weird I noticed while debugging, is that the `lua_State *` is allocated in a low address on Jit64: `0x0000000041d9e378`. Isn't the lower 32-bit address range reserved for PowerPC? Maybe the Lua integration needs some changes to the allocator.

Not sure if related, but stable LuaJIT is limited to 1 or 2GB of memory per process. That is a consequence of the extreme focus on speed in LuaJIT. 2.1 beta can support more memory by enabling GC64 (64 bit garbage collector) mode at compile time.

terorie:
> LuaJIT's 64 bit integers are boxed and are not useful as table keys (search for "cdata objects as table keys" in the link you provided). PUC Lua 5.3 has full native support for 64 bit integers.
> PUC Lua 5.3 has full native support for 64 bit integers.

@rlnilsen I don't think this is entirely accurate. LuaJIT's native `lua_Integer` is an i64 (`ptrdiff_t`) and looking at the Lua 5.3 manual, it seems to be the same case for PUC Lua 5.3. LuaJIT additionally supports u64 through userdatum though.

> > I'm confident that LuaJIT is stable enough where we can assume it's bug free.
>
> You have more faith in human infallability than I.

My words were a bit too strong, admittedly. :wink: 

> Not sure if related, but stable LuaJIT is limited to 1 or 2GB (depending on kernel version) of memory per process. That is a consequence of the extreme focus on speed in LuaJIT. 2.1 beta can support more memory by enabling GC64 (64 bit garbage collector) mode at compile time.

Good to know, thanks for pointing out! I hope this doesn't bite with Dolphin's Jit64.

In the end I'm torn between PUC Lua 5.3 vs. LuaJIT 2.0. Let's see if there are strong opinions about this decision and the maintainers will have their final say.

terorie:
@rlnilsen Sorry, I take it back! LuaJIT surprisingly does not support i64. Lua 5.3 it is then!

image

rlnilsen:
> > LuaJIT's 64 bit integers are boxed and are not useful as table keys (search for "cdata objects as table keys" in the link you provided). PUC Lua 5.3 has full native support for 64 bit integers.
> 
> @rlnilsen I don't think this is entirely accurate. LuaJIT's native `lua_Integer` is an i64 (`ptrdiff_t`) and looking at the Lua 5.3 manual, it seems to be the same case for PUC Lua 5.3. LuaJIT additionally supports u64 through userdatum though.

@terorie `lua_Integer` is just the type used by the C API. Before Lua 5.3 (including LuaJIT), internally there was only one native number type: the 64 bit double precision float. That limits LuaJIT integers to 53 bits.

terorie:
@rlnilsen One more question, is PUC Lua 5.3 able to handle u64 natively, or maybe via wrappers around i64? IMHO having 64-bit integer support is only really useful when they can fully represent unsigned 64-bit integers.

rlnilsen:
@terorie Lua 5.3 integers are stored and accessible as full 64 bits. Semantics are however biased towards signed, meaning comparison operators treat the numbers as signed. But there is a function math.ult() for unsigned comparison.

Add/sub/bitwise operators are signed/unsigned agnostic. Not sure about mul/div/mod/exp.

From the manual:
> In case of overflows in integer arithmetic, all operations wrap around, according to the usual rules of two-complement arithmetic. (In other words, they return the unique representable integer that is equal modulo 2^64 to the mathematical result.)

jordan-woyak:
I'd rather have python #7064 than lua, Just throwing that out there.

CMake/FindLuaJit.cmake
terorie: FindLuaJit.cmake was removed so this is resolved.

Source/Core/DolphinQt/CMakeLists.txt
```diff
@@ -334,35 +341,35 @@ if(WIN32)
...
     COMMAND ${CMAKE_COMMAND} -E copy_if_different "${CMAKE_CURRENT_SOURCE_DIR}/qt.conf.win" "${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/qt.conf"
   )
 
-  # Delegate to Qt's official deployment binary on Windows to copy over the necessary Qt-specific libraries, etc.
+#Delegate to Qt's official deployment binary on Windows to copy over the necessary Qt-specific libraries, etc.
   get_target_property(MOC_EXECUTABLE_LOCATION Qt5::moc IMPORTED_LOCATION)
   get_filename_component(QT_BINARY_DIRECTORY "${MOC_EXECUTABLE_LOCATION}" DIRECTORY)
   find_program(WINDEPLOYQT_EXE windeployqt HINTS "${QT_BINARY_DIRECTORY}")
 
-  # Note: We set the PATH for the duration of this command so that the
```
terorie, line 352: Oops, did not mean to modify this

terorie:
@Felk The PowerPC/MMU interface changes and events engine at #7064 looks super useful. Would you mind if I cherry-pick your changes to here? As long as nothing is squash merged this should preserve git blame / authors. In return, I can try porting breakpoints support to Python :)

Felk:
> Would you mind if I cherry-pick your changes to here?

@terorie Cherry-pick everything you want. You only need to worry about preserving the author if it's in your own interest, e.g. to point people at me if they have questions about some specific code

> In return, I can try porting breakpoints support to Python :)

Sure, contributions are welcome!

docs/Scripts.md
```diff
@@ -0,0 +1,122 @@
...
+
+Access Wii memory from the main processor's view.
+
+#### Memory arrays
+
+Lua tables taking an address as index. Supports unaligned access.
+
+| identifier            | type     | description               |
+| --------------------- | -------- | ------------------------- |
+| `dolphin.ppc.mem_u8`  | `u8[*]`  | Read/write u8             |
```
smurf3tte, line 53: These aren't part of ppc in the implementation, which seems correct. Copy/paste bug in the docs?

smurf3tte:
I find this personally useful. However, I also understand that it is, as you say, unlikely to be merged. If you have more time to invest and would like to discuss potential debugger improvements - which may or may not involve embedding a scripting language in Dolphin - you should drop by the developer IRC ([details](https://forums.dolphin-emu.org/Thread-possible-development-help)). Disclaimer: I am not a member of the dev team.

Source/Core/Core/ScriptEngine.cpp
```diff
@@ -0,0 +1,1010 @@
...
+  return 1;
+}
+
+// Lua: set_ps_f64(userdata ps_reg, number idx, number value) -> ()
+// Sets one of the two elements of a PS (paired single) register in f64 mode.
+static int set_ps_f64(lua_State* L)
+{
+  const u8* idx = reinterpret_cast<const u8*>(lua_touserdata(L, 1));
+  const lua_Integer idx_arg = luaL_checkinteger(L, 2);
+  const lua_Number value = luaL_checkinteger(L, 3);
```
smurf3tte, line 555:
```suggestion
  const lua_Number value = luaL_checknumber(L, 3);
```

Source/Core/Core/ScriptEngine.cpp
```diff
@@ -0,0 +1,1010 @@
...
+{
+  const std::optional<u32> addr_arg = luaL_checkaddress(L, 1);
+  const lua_Integer len_arg = luaL_checkinteger(L, 2);
+
+  if (addr_arg)
+  {
+    size_t len = 0;
+    if (len_arg > 0)
+      len = static_cast<size_t>(len_arg);
+    const std::string cpp_str = PowerPC::HostGetString(*addr_arg, len);
```
smurf3tte, line 171: HostGetString() stops when it reaches a null terminator, making it unsuitable for mem_read(). It would be good for implementing str_read(), though.

terorie:
Forgot to close this. :^)

PR #9208, opened 2020-10-29
[Feedback Wanted] Android: IR pointer improvements

Ebola16:
Fixes https://bugs.dolphin-emu.org/issues/12307 and should be the final IR pointer PR.
This PR combined with #9209 should significantly improve the IR pointer.


Ebola16:
Updated TODO

Ebola16:
I think this properly fixes the IR pointer now but I'm still not 100% sure it's the best solution. When testing, be sure to use the updated default values in #9209.

JosJuice:
This change doesn't make much sense to me. I believe #9215 is the proper fix.

Ebola16:
Yeah, the values I used worked on my phone and manipulating the buffer should be device agnostic but I had a feeling there was a better way to accomplish what I wanted. #9215 is less hacky so let's go with that.

PR #9209, opened 2020-10-29
Android: Update IR sensitivity default values

Ebola16:
Fixes https://bugs.dolphin-emu.org/issues/12308
Note that there will still be some inaccuracies but this PR allows me to reach the edge of the screen on the games I tested (Wii Sports, Wii Music, Sukeban Shachou Rena). See #9208 for more information.


Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/activities/EmulationActivity.java
```diff
@@ -950,15 +950,15 @@ private void setIRSensitivity()
...
+    int ir_yaw = ini.getInt(Settings.SECTION_CONTROLS, SettingsFile.KEY_WIIBIND_IR_YAW, 25);
 
     TextView text_slider_value_yaw = view.findViewById(R.id.text_ir_yaw);
     TextView units_yaw = view.findViewById(R.id.text_ir_yaw_units);
     SeekBar seekbar_yaw = view.findViewById(R.id.seekbar_width);
 
     text_slider_value_yaw.setText(String.valueOf(ir_yaw));
     units_yaw.setText(getString(R.string.yaw));
-    seekbar_yaw.setMax(100);
+    seekbar_yaw.setMax(360);
```
JosJuice, line 961: This isn't necessarily a good idea. While the value *can* go this high, making it go this high is basically useless, and any increase of the max value makes the setting harder to edit since we're using sliders.

Ebola16: I haven't messed around with these settings too much. Should I just leave this value at 100?

JosJuice: Yeah, I think that would make the most sense for now.

Ebola16: Done

JosJuice (approved):
This won't overwrite values in existing INIs, but that's not necessarily a bad thing. LGTM.

Ebola16:
Yes, I figured that user modified values are still valid and shouldn't be changed in this case.

PR #9210, opened 2020-10-30
DolphinQt: Add Show/Hide All options to gamelist region menu

Dentomologist:
If you wish to only display games from only one region you currently have to open the View -> Regions menu 13 times to manually deselect every other region.  This PR adds "Show All" and "Hide all" options to simplify the process.

image

### DIFF
--- a/Source/Core/DolphinQt/MenuBar.cpp
+++ b/Source/Core/DolphinQt/MenuBar.cpp
@@ -695,20 +695,29 @@ void MenuBar::AddShowRegionsMenu(QMenu* view_menu)
       {tr("Show World"), &SConfig::GetInstance().m_ListWorld},
       {tr("Show Unknown"), &SConfig::GetInstance().m_ListUnknown}};
 
-  QActionGroup* region_group = new QActionGroup(this);
-  QMenu* region_menu = view_menu->addMenu(tr("Show Regions"));
-  region_group->setExclusive(false);
+  QMenu* const region_menu = view_menu->addMenu(tr("Show Regions"));
+  const QAction* const show_all_regions = region_menu->addAction(tr("Show All"));
+  const QAction* const hide_all_regions = region_menu->addAction(tr("Hide All"));
+  region_menu->addSeparator();
 
   for (const auto& key : region_map.keys())
   {
-    bool* config = region_map[key];
-    QAction* action = region_group->addAction(region_menu->addAction(key));
-    action->setCheckable(true);
-    action->setChecked(*config);
-    connect(action, &QAction::toggled, [this, config, key](bool value) {
-      *config = value;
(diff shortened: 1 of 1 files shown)


JosJuice:
@dolphin-emu-bot rebuild

Source/Core/DolphinQt/MenuBar.cpp
```diff
@@ -695,20 +695,29 @@ void MenuBar::AddShowRegionsMenu(QMenu* view_menu)
...
-    connect(action, &QAction::toggled, [this, config, key](bool value) {
-      *config = value;
-      emit GameListRegionVisibilityToggled(key, value);
-    });
+    bool* const config = region_map[key];
+    QAction* const menu_item = region_menu->addAction(key);
+    menu_item->setCheckable(true);
+    menu_item->setChecked(*config);
+
+    const auto SetVisibility = [this, config, key, menu_item](bool visibility) {
```
leoetlino, line 710:
```suggestion
    const auto set_visibility = [this, config, key, menu_item](bool visibility) {
```

> [2020-10-29 16:44:05] Lioncash: I generally write the lambda name as `snake_case` like a variable, as opposed to `CamelCase` like we do for functions, so it's visually obvious that a lambda is likely being invoked. (and also doesn't add another special case to variable naming to remember).

Source/Core/DolphinQt/MenuBar.cpp
```diff
@@ -695,20 +695,29 @@ void MenuBar::AddShowRegionsMenu(QMenu* view_menu)
...
+      menu_item->setChecked(visibility);
+      *config = visibility;
+      emit GameListRegionVisibilityToggled(key, visibility);
+    };
+    const auto SetVisible = std::bind(SetVisibility, true);
```
leoetlino, line 715:
```suggestion
    const auto set_visible = std::bind(SetVisibility, true);
```

Source/Core/DolphinQt/MenuBar.cpp
```diff
@@ -695,20 +695,29 @@ void MenuBar::AddShowRegionsMenu(QMenu* view_menu)
...
+    const auto SetHidden = std::bind(SetVisibility, false);
```
leoetlino, line 716:
```suggestion
    const auto set_hidden = std::bind(SetVisibility, false);
```

leoetlino (changes requested):

Source/Core/DolphinQt/MenuBar.cpp
Dentomologist: I needed to change the SetVisibility on this line and set_hidden, too.

leoetlino (approved):

PR #9211, opened 2020-10-31
CMake: unset cached variable used by try_compile

brianlheim:
On Arch Linux, using CMake 3.18.4, this fixed a bug in the build system caused by FindMBEDTLS.cmake

Steps to reproduce:
1. Without mbedtls installed, run `cmake ..` to configure the project in a clean build directory
2. Install mbedtls
3. Reconfigure the project by running `cmake ..`

Expected: mbedtls is found and used
Actual:

```
-- Could NOT find MBEDTLS (missing: MBEDTLS_VERSION_OK)
-- Using static mbed TLS from Externals
```

This commit fixes the issue by unset-ing this cache variable, which I think is created by the
`find_package_handle_standard_args` call? In any case `try_compile` seems to refuse to set it if it's a cache variable.


leoetlino:
Thanks for the PR. However, I believe the standard way to deal with system environment changes like these is to clear the CMake cache manually by deleting CMakeCache.txt. Otherwise, you'll probably still have issues with other libraries and I don't think we should be unsetting random variables for every library we use for maintenance reasons.

PR #9212, opened 2020-11-01
Game INI Changes

nf6429:
This adds the INI changes to the SNES games which needed EFB Copies to Texture Only turned off to not display a black screen, the Dualcore issues with some SNES and Sega Genesis games in terms of stuttering and darkening the screen, and also the XFB issues for Sega Genesis and Sega Master System games in terms of purple screens

### DIFF
--- a/Data/Sys/GameSettings/JAP.ini
+++ b/Data/Sys/GameSettings/JAP.ini
@@ -1,0 +1,5 @@
+# JAPJ01 - Fire Emblem: Seisen no Keifu
+
+[Video_Hacks]
+# Fixes black screen.
+EFBToTextureEnable = False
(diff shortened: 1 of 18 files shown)


JMC47:
Changes seem to make sense to me.

nf6429:
Just added the Dragon Quest 25th Collection, since I just realised that the game need EFB Texture Only turned off not to black screen, and Texture Cache Accuracy on Safe for the cursor to update it's position, so you can navigate the menu's

mbc07:
Is the Virtual Console changes from this PR applicable to all SNES, Genesis and Master System games or it's just required on select games? If it's the former, it would probably be wise to set them on the global VC INI file instead (J.ini for SNES, L.ini for Master System and M.ini for Genesis)...

nf6429:
> Is the Virtual Console changes from this PR applicable to all SNES, Genesis and Master System games or it's just required on select games? If it's the former, it would probably be wise to set them on the global VC INI file instead (J.ini for SNES, L.ini for Master System and M.ini for Genesis)...

The changes In this PR are only for select SNES, genesis, and master system games, so I can't do a S.ini, etc. I tried every other game from all of these systems to confirm this.

PR #9213, opened 2020-11-01
Simplify Arm64GPRCache::Flush

JosJuice:
`flush = cond ? flush : true` makes little sense when `flush` is always true at that point.


degasus (approved):
I've tried to analyze why there is actually this code. I think this was from the very old days (2014) when we tried to speed up interpreter fallbacks. So this was missed on a cleanup. And indeed, as we mostly use Flush for branching, there is no point in keeping any registers dirty.

lioncash (approved):

PR #8758, opened 2020-04-22
Socket: Abort pending operations on shutdown

sepalani:
~~This PR will need a hardware test (similar #8389) to confirm this behaviour.~~

## PR goal
This PR tries to fix inconsistencies regarding `shutdown` implementation (on a socket) and be as close as possible to what the Wii does *(not sure about the GC behaviour)*.

### What it fixes
It fixes crashes/hangs in some Just Dance games : https://dolp.in/i11910

### What it doesn't fix
 * Dolphin not getting the right network interface on :
   - Android (https://github.com/dolphin-emu/dolphin/pull/9191)
   - MacOS (https://github.com/dolphin-emu/dolphin/pull/9214)
 * Return values of other network functions

## Hardware test
I created 2 hardware tests:
1. one that fires `shutdown` on blocking operations
2. the second one that checks return values of network functions after `shutdown` was called

### How-to run
1. Get the latest version of Dolphin (at least after https://github.com/dolphin-emu/dolphin/pull/9099)
2. Enable Dolphin debugging interface under `Options > Config > Interface` then tick `Show Debugging UI`
3. Enable the Dolphin Log window
4. In the Log configuration tab
   * Tick "Write to File" in Logger Outputs
   * Tick "OSREPORT" and "HLE"
   * Set the verbosity level to `Info` or `Debug`
5. Make sure you run the ELF (not the DOL) hardware tests
*(On top of that, you can also grab the result using netcat on your local IP address and the result port displayed at the end of the test)*

### Hardware tests results
Dolphin results are based on ~~this build #8796~~ latest master and https://github.com/dolphin-emu/dolphin/pull/9191 for Android.

#### Results of the [first hwtest](https://github.com/dolphin-emu/dolphin/files/4605323/wii-shutdown-v2.zip)
 - [x] [Wii](https://gist.github.com/sepalani/d1a9514cf410a139f280681bf76f8e99)
 - [x] [Dolphin (Windows 10)](https://gist.github.com/sepalani/820928366b670c20a843186cc55cba5b)
 - [x] Dolphin (Linux) | [Parrot](https://gist.github.com/sepalani/8566d82614c36ae38ec402e6e2be166b)
 - [x] Dolphin (Mac) | [macOS 11](https://gist.github.com/sepalani/427c866bd0febd514ae12dc2cd808e59)
 - [x] Dolphin (Android) | [ASUS RoG Phone II](https://gist.github.com/sepalani/e42880547e96d66864d1c69d0c8f7af3)

[...]

PR #9214, opened 2020-11-01
Top: Set port when connecting a socket to find the default interface

OatmealDome:
This fixes an issue that was found in #8758. Previously, the call to ``connect`` in ``GetSystemDefaultInterface`` would fail with ``EADDRNOTAVAIL`` on macOS.  If we set ``sin_port``, the call will succeed.

Tested on macOS 11 Big Sur beta 10.

cc @sepalani

### DIFF
--- a/Source/Core/Core/IOS/Network/IP/Top.cpp
+++ b/Source/Core/Core/IOS/Network/IP/Top.cpp
@@ -223,8 +223,10 @@ static std::optional<DefaultInterface> GetSystemDefaultInterface()
     sockaddr_in addr{};
     socklen_t length = sizeof(addr);
     addr.sin_family = AF_INET;
-    // The address is irrelevant -- no packet is actually sent. This just needs to be a public IP.
+    // The address and port are irrelevant -- no packet is actually sent. These just need to be set
+    // to a valid IP and port.
     addr.sin_addr.s_addr = inet_addr(8, 8, 8, 8);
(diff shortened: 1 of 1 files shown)


BhaaLseN (approved):
Looks like a sensible change; untested though.

Source/Core/Core/IOS/Network/IP/Top.cpp
```diff
@@ -223,8 +223,10 @@ static std::optional<DefaultInterface> GetSystemDefaultInterface()
     sockaddr_in addr{};
     socklen_t length = sizeof(addr);
     addr.sin_family = AF_INET;
-    // The address is irrelevant -- no packet is actually sent. This just needs to be a public IP.
+    // The address and port are irrelevant -- no packet is actually sent. These just need to be set
+    // to a valid IP and port.
     addr.sin_addr.s_addr = inet_addr(8, 8, 8, 8);
+    addr.sin_port = 53;
```
sepalani, line 229: This should probably be `htons(53)` instead.

sepalani (commented):
Otherwise, looks good to me.

Source/Core/Core/IOS/Network/IP/Top.cpp
OatmealDome: fixed, thanks!

leoetlino (approved):
yeah, sounds sane.

PR #9215, opened 2020-11-02
Android/InputOverlayPointer: Don't assume surface covers whole screen

JosJuice:
This assumption is false both in portrait mode (where it only covers the top half of the screen) and when using two apps at once.

Fixes https://bugs.dolphin-emu.org/issues/12307. Alternative to PR #9208.


Ebola16 (approved):
Along with the changes in #9209, this makes the IR pointer much better.

lioncash (approved):

PR #9218, opened 2020-11-02
JitArm64: Implement HLE function hooking

JosJuice:
Closely based on the x86-64 implementation. Fixes https://bugs.dolphin-emu.org/issues/10965.

### DIFF
--- a/Source/Core/Core/PowerPC/JitArm64/Jit.cpp
+++ b/Source/Core/Core/PowerPC/JitArm64/Jit.cpp
@@ -209,20 +209,16 @@ void JitArm64::FallBackToInterpreter(UGeckoInstruction inst)
   }
 }
 
-void JitArm64::HLEFunction(UGeckoInstruction inst)
+void JitArm64::HLEFunction(u32 hook_index)
 {
+  FlushCarry();
   gpr.Flush(FlushMode::FLUSH_ALL);
   fpr.Flush(FlushMode::FLUSH_ALL);
 
   MOVI2R(W0, js.compilerPC);
-  MOVI2R(W1, inst.hex);
+  MOVI2R(W1, hook_index);
   MOVP2R(X30, &HLE::Execute);
   BLR(X30);
-
-  ARM64Reg WA = gpr.GetReg();
-  LDR(INDEX_UNSIGNED, WA, PPC_REG, PPCSTATE_OFF(npc));
-  WriteExit(WA);
-  gpr.Unlock(WA);
 }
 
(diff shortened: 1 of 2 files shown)


Source/Core/Core/PowerPC/JitArm64/Jit.cpp
```diff
@@ -209,20 +209,15 @@ void JitArm64::FallBackToInterpreter(UGeckoInstruction inst)
   }
 }
 
-void JitArm64::HLEFunction(UGeckoInstruction inst)
+void JitArm64::HLEFunction(u32 hook_index)
 {
   gpr.Flush(FlushMode::FLUSH_ALL);
   fpr.Flush(FlushMode::FLUSH_ALL);
```
degasus, line 215: Is there a chance that the first instruction is an integer instruction? If so, "FlushCarry();" might need to be called as well.

Source/Core/Core/PowerPC/JitArm64/Jit.cpp
```diff
@@ -490,6 +485,21 @@ void JitArm64::WriteExceptionExit(ARM64Reg dest, bool only_external)
...
+bool JitArm64::HandleFunctionHooking(u32 address)
+{
+  return HLE::ReplaceFunctionIfPossible(address, [&](u32 hook_index, HLE::HookType type) {
+    HLEFunction(hook_index);
+
+    if (type != HLE::HookType::Replace)
+      return false;
+
+    LDR(INDEX_UNSIGNED, DISPATCHER_PC, PPC_REG, PPCSTATE_OFF(npc));
+    js.downcountAmount += js.st.numCycles;
```
degasus, line 497: Honestly, I do not get the point why we add the cycles of this single instruction. I assume te HLE functions shall decrease downCount on their own, so this single instruction likely has no effect. However it is the same in the x64 JIT, so I'm fine with it.

degasus (approved):
With the FlushCarry, it looks good to me.

Source/Core/Core/PowerPC/JitArm64/Jit.cpp
JosJuice: `js.st.numCycles` is for the whole block, not the current instruction. Though, implementing it that way also seems weird, considering that we might hit the HLE case halfway through a block after we've already added some cycles to `js.downcountAmount` the normal way...

Source/Core/Core/PowerPC/JitArm64/Jit.cpp
JosJuice: FlushCarry added.

PR #9220, opened 2020-11-04
WiimoteEmu: Implement extension encryption edge case.

jordan-woyak:
Attempting find more elegant behavior behind extension encryption I seem to have discovered it is even more complicated.

When encryption key generation fails to match with one of the 7 indices, extensions use alternate sboxes (which differ between 1st and 3rd party extensions).

Previously, tables were 0x17-filled (to simulate the outcome of a 0x00-filled key).
Behavior should now match that of real extensions for arbitrary key data.

Reorganized/renamed some of the constants to make logic more clear.

Fixed some missing table data for 3rd party extensions. They shouldn't randomly fail anymore.


Source/Core/Core/HW/WiimoteEmu/Encryption.cpp
```diff
@@ -491,15 +565,49 @@ EncryptionKey KeyGen1stParty::GenerateTables(const EncryptionKey::RandData& rand
...
+  result.GenerateTables(rand, key, sboxes_3rd_party[idx], sboxes_3rd_party[idx + 1]);
+  return result;
+}
+
+EncryptionKey KeyGen1stParty::GenerateFallbackTables(const EncryptionKey::RandData& rand,
+                                                     const EncryptionKey::KeyData& key) const
+{
+  // Reverse engineering did not produce exact tables, only the XOR effect of each.
+  // This is the XOR difference between the tables.
+  // This also works with 0x97.
```
iwubcode, line 585: I'm not the target of this kind of information but what does this mean?  That using `0x97` also produces the same result logically?  Why did you choose `0x17` over `0x97`?

iwubcode:
One general comment, I couldn't reverse engineer anything to save my life, so maybe to someone who does this on a daily basis this will see this as a dumb comment.  However, there is not any information on how these tables were acquired.  It's just my personal opinion that I wish there was more documentation in Dolphin itself about _how_ logic was decided instead of just the end result.

I don't own any 3rd party extensions myself but I'm sure this PR will make many people happy.  Nice job @jordan-woyak !

Source/Core/Core/HW/WiimoteEmu/Encryption.cpp
jordan-woyak: Yes. It's the same result logically.
The encryption function is: `enc = (val - ft) ^ sb`
`ft` and `sb` are produced from an XOR of values from each of the reverse engineered tables.
Subtracting, Adding, or XORing 0x80 to a u8 all produce the same result so if the table XOR difference is 0x80 more the end result is the same.
I chose 0x17 because it's a lower number. :P
Some more info here: https://wiibrew.org/wiki/Wiimote/Extension_Controllers

PR #9224, opened 2020-11-05
Android: BindingSettings to INI settings

Ebola16:
Split from #9192. Part of https://bugs.dolphin-emu.org/issues/10957.

This method currently works but differs from how C++ handles bindings. This PR can be improved by handling the binding settings in the same way as the settings from #9170. I'll look into ways to use regex to combine the binding and binding description keys after that PR is merged.


JosJuice:
Superseded by PR #11385. We're now storing the bindings just like on PC.

PR #9228, opened 2020-11-06
Android: Fix jobject/jclass warnings

JosJuice:
Static methods use jclass, and non-static methods use jobject.

### DIFF
--- a/Source/Android/jni/AndroidCommon/IDCache.h
+++ b/Source/Android/jni/AndroidCommon/IDCache.h
@@ -8,8 +8,6 @@
 
 namespace IDCache
 {
-static constexpr jint JNI_VERSION = JNI_VERSION_1_6;
-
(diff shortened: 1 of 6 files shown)


Ebola16 (commented):
- Android Studio wants `jclass obj` for `onGamePadMoveEvent`
- Should we suppress the unused local variable warnings for `MainAndroid.cpp`? There's a lot of them.

JosJuice:
I've fixed `onGamePadMoveEvent`. Not sure about the unused local variable warnings.

Ebola16:
I'd like something to be done about the large amounts of unused `obj` warnings if not suppression.

JosJuice:
The latest commit should address it.

Ebola16 (approved):
That works too.

lioncash (approved):

PR #9230, opened 2020-11-06
Android: Call saveSettings in EmulationActivity.onStop

JosJuice:
I didn't realize that onSaveInstanceState doesn't get called when finishing the activity.


Ebola16 (approved):
I also made this change in a WIP PR so LGTM.

lioncash (approved):

PR #9231, opened 2020-11-07
Android: Remove unnecessary WrongConstant warning suppression

Ebola16:
Android Studio doesn't raise any warnings when these are removed.


JosJuice (approved):
Not sure why these existed, so... LGTM I guess?

lioncash (approved):

PR #9232, opened 2020-11-08
Qt/IOWindow: Show result value in expression editor.

AdmiralCurtiss:
This displays the current result of an input expression right there in the editor, instead of the parser error box.

success fail

Updates automatically just like the available inputs table above and immediately reacts to any edits in the expression.

### DIFF
--- a/Source/Core/DolphinQt/Config/Mapping/IOWindow.cpp
+++ b/Source/Core/DolphinQt/Config/Mapping/IOWindow.cpp
@@ -93,9 +93,8 @@ QTextCharFormat GetCommentCharFormat()
 }
 }  // namespace
 
-ControlExpressionSyntaxHighlighter::ControlExpressionSyntaxHighlighter(QTextDocument* parent,
-                                                                       QLineEdit* result)
-    : QSyntaxHighlighter(parent), m_result_text(result)
+ControlExpressionSyntaxHighlighter::ControlExpressionSyntaxHighlighter(QTextDocument* parent)
+    : QSyntaxHighlighter(parent)
 {
 }
 
@@ -168,17 +167,10 @@ void ControlExpressionSyntaxHighlighter::highlightBlock(const QString&)
   }
 
   // This doesn't need to be run for every "block", but it works.
-  if (ciface::ExpressionParser::ParseStatus::Successful != tokenize_status)
-  {
-    m_result_text->setText(tr("Invalid Token."));
-  }
-  else
+  if (ciface::ExpressionParser::ParseStatus::Successful == tokenize_status)
   {
     ciface::ExpressionParser::RemoveInertTokens(&tokens);
     const auto parse_status = ciface::ExpressionParser::ParseTokens(tokens);
-
-    m_result_text->setText(
-        QString::fromStdString(parse_status.description.value_or(_trans("Success."))));
 
     if (ciface::ExpressionParser::ParseStatus::Successful != parse_status.status)
     {
@@ -203,7 +195,8 @@ private:
 
 IOWindow::IOWindow(MappingWidget* parent, ControllerEmu::EmulatedController* controller,
                    ControlReference* ref, IOWindow::Type type)
-    : QDialog(parent), m_reference(ref), m_controller(controller), m_type(type)
+    : QDialog(parent), m_reference(ref), m_original_expression(ref->GetExpression()),
+      m_controller(controller), m_type(type)
 {
   CreateMainLayout();
 
@@ -242,7 +235,7 @@ void IOWindow::CreateMainLayout()
 
   m_expression_text = new QPlainTextEdit();
   m_expression_text->setFont(QFontDatabase::systemFont(QFontDatabase::FixedFont));
-  new ControlExpressionSyntaxHighlighter(m_expression_text->document(), m_parse_text);
+  new ControlExpressionSyntaxHighlighter(m_expression_text->document());
 
   m_operators_combo = new QComboBox();
   m_operators_combo->addItem(tr("Operators"));
@@ -404,6 +397,7 @@ void IOWindow::ConnectWidgets()
   connect(m_expression_text, &QPlainTextEdit::textChanged, [this] {
     m_apply_button->setText(m_apply_button->text().remove(QStringLiteral("*")));
     m_apply_button->setText(m_apply_button->text() + QStringLiteral("*"));
+    UpdateExpression(m_expression_text->toPlainText().toStdString());
   });
 
   connect(m_operators_combo, qOverload<int>(&QComboBox::activated), [this](int index) {
@@ -423,6 +417,9 @@ void IOWindow::ConnectWidgets()
 
     m_functions_combo->setCurrentIndex(0);
   });
+
+  // revert the expression when the window closes without using the OK button
+  connect(this, &IOWindow::finished, [this] { UpdateExpression(m_original_expression); });
 }
 
 void IOWindow::AppendSelectedOption()
@@ -448,8 +445,10 @@ void IOWindow::OnDialogButtonPressed(QAbstractButton* button)
     return;
   }
 
-  m_reference->SetExpression(m_expression_text->toPlainText().toStdString());
-  m_controller->UpdateSingleControlReference(g_controller_interface, m_reference);
+  const auto lock = m_controller->GetStateLock();
+
+  UpdateExpression(m_expression_text->toPlainText().toStdString());
+  m_original_expression = m_reference->GetExpression();
 
   m_apply_button->setText(m_apply_button->text().remove(QStringLiteral("*")));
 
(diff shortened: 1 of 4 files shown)


Source/Core/DolphinQt/Config/Mapping/IOWindow.cpp
```diff
@@ -315,7 +311,10 @@ void IOWindow::CreateMainLayout()
     m_option_list->setColumnWidth(1, 64);
     m_option_list->horizontalHeader()->setSectionResizeMode(1, QHeaderView::Fixed);
 
-    m_option_list->setItemDelegate(new InputStateDelegate(this));
+    m_option_list->setItemDelegate(new InputStateDelegate(this, 1, [&](int row) {
+      // Clamp off negative values but allow greater than one in the text display.
```
BhaaLseN, line 315: I can see that from the code, but..._why_?
I realize you probably just moved this from `InputStateDelegate::paint` but comments that only reflect what the code does without adding any additional value seem pointless to me.

Source/Core/DolphinQt/Config/Mapping/IOWindow.cpp
```diff
@@ -532,7 +562,29 @@ void IOWindow::UpdateDeviceList()
       QString::fromStdString(m_controller->GetDefaultDevice().ToString()));
 }
 
-InputStateDelegate::InputStateDelegate(IOWindow* parent) : QItemDelegate(parent), m_parent(parent)
+void IOWindow::UpdateExpression(std::string new_expression)
+{
+  const auto lock = m_controller->GetStateLock();
+  if (new_expression != m_reference->GetExpression())
```
BhaaLseN, line 568: Would this be a candidate for invert-condition/early-exit, since the remainder of the method is just inside this `if`?

Source/Core/DolphinQt/Config/Mapping/IOWindow.cpp
AdmiralCurtiss: You're not wrong, I was wondering that too. Apparently this was written by @jordan-woyak in #8396, do you happen to remember?

Source/Core/DolphinQt/Config/Mapping/IOWindow.cpp
AdmiralCurtiss: Yup.

Source/Core/DolphinQt/Config/Mapping/IOWindow.cpp
jordan-woyak: Input states may return negative values when opposing inputs are activated ("Axis X-" is negative when "Axis X+" is positive)
https://github.com/dolphin-emu/dolphin/blob/master/Source/Core/InputCommon/ControllerInterface/Device.h#L76

This helps input detection determine the difference between full axis and half axis mappings in a generic manner.
Otherwise it can't tell the difference between "Axis X+" and "Full Axis X+".

Normally this is clamped off by ControlExpression: https://github.com/dolphin-emu/dolphin/blob/master/Source/Core/InputCommon/ControlReference/ExpressionParser.cpp#L267

But because `ControlExpression` is bypassed in this display they have to be manually clamped to not confuse users with weird negative values.

Greater than one values are allowed for accelerometer/gyroscope input.

Source/Core/DolphinQt/Config/Mapping/IOWindow.cpp
```diff
@@ -358,7 +357,31 @@ void IOWindow::CreateMainLayout()
...
+  m_result_value->horizontalHeader()->setSectionResizeMode(0, QHeaderView::Fixed);
+  m_result_value->verticalHeader()->hide();
+  m_result_value->verticalHeader()->setDefaultSectionSize(
+      m_result_value->verticalHeader()->minimumSectionSize());
+  m_result_value->setEditTriggers(QAbstractItemView::NoEditTriggers);
+  m_result_value->setSelectionBehavior(QAbstractItemView::SelectRows);
+  m_result_value->setSelectionMode(QAbstractItemView::SingleSelection);
+  m_result_value->setColumnWidth(0, 64);
+  m_result_value->setRowHeight(0, 22);
+  m_result_value->setFixedSize(65, 23);
```
jordan-woyak, line 378: Creating another `QTableWidget` just for this is pretty ugly. Maybe you can create a text box (or other simple widget) and break out the paint function of `InputStateDelegate`?

Source/Core/DolphinQt/Config/Mapping/IOWindow.cpp
```diff
@@ -532,7 +562,29 @@ void IOWindow::UpdateDeviceList()
...
+  m_controller->UpdateSingleControlReference(g_controller_interface, m_reference);
+
+  if (error)
+    m_parse_text->setText(QString::fromStdString(*error));
+  else if (status == ciface::ExpressionParser::ParseStatus::EmptyExpression)
+    m_parse_text->setText(QString());
+  else if (status != ciface::ExpressionParser::ParseStatus::Successful)
+    m_parse_text->setText(tr("Invalid Expression."));
+  else
+    m_parse_text->setText(tr("Success."));
```
jordan-woyak, line 582: Maybe if the parse is successful the result text can be hidden and the new `m_result_value` can take up the whole row?
Similarly, if the parse fails you could hide `m_result_value`?

Source/Core/DolphinQt/Config/Mapping/IOWindow.cpp
```diff
@@ -448,8 +476,10 @@ void IOWindow::OnDialogButtonPressed(QAbstractButton* button)
     return;
   }
 
-  m_reference->SetExpression(m_expression_text->toPlainText().toStdString());
-  m_controller->UpdateSingleControlReference(g_controller_interface, m_reference);
+  const auto lock = m_controller->GetStateLock();
+
+  UpdateExpression(m_expression_text->toPlainText().toStdString());
+  m_original_expression = m_reference->GetExpression();
```
jordan-woyak, line 482: With this new `m_original_expression` state maybe we can just kill the "Apply" button entirely now that it applies in real time?

AdmiralCurtiss:
@jordan-woyak Okay, I did another pass on this and am *much* happier with the result now. The indicator now paints over the error textbox in the successful case. See if you agree!

Source/Core/DolphinQt/Config/Mapping/IOWindow.cpp
AdmiralCurtiss: I added a commit that removes the apply button, but I don't really care whether it stays or not. If we decide to keep it we can just drop the commit.

Source/Core/DolphinQt/Config/Mapping/IOWindow.cpp
AdmiralCurtiss: So... should I extend that comment to include the reasoning?

Source/Core/DolphinQt/Config/Mapping/IOWindow.cpp
```diff
@@ -192,18 +184,32 @@ void ControlExpressionSyntaxHighlighter::highlightBlock(const QString&)
...
+};
+
+class InputStateLineEdit : public QLineEdit
+{
+public:
+  InputStateLineEdit(std::function<ControlState()> state_evaluator);
+
+  void paintEvent(QPaintEvent* event) override;
+
+  std::function<ControlState()> m_state_evaluator;
```
jordan-woyak, line 205: I think throughout dolphin we only use the "m_" prefix for private members?

Source/Core/DolphinQt/Config/Mapping/IOWindow.h
```diff
@@ -34,13 +37,10 @@ class ControlExpressionSyntaxHighlighter final : public QSyntaxHighlighter
 {
   Q_OBJECT
 public:
-  ControlExpressionSyntaxHighlighter(QTextDocument* parent, QLineEdit* result);
+  ControlExpressionSyntaxHighlighter(QTextDocument* parent);
```
leoetlino, line 42: This constructor should be explicit to avoid unintended implicit conversions.

Source/Core/DolphinQt/Config/Mapping/IOWindow.h
```diff
@@ -4,6 +4,9 @@
 
 #pragma once
 
+#include <memory>
```
leoetlino, line 7: note to self: this is for std::shared_ptr, which is outside of the diff

Source/Core/DolphinQt/Config/Mapping/IOWindow.cpp
```diff
@@ -184,13 +184,27 @@ void ControlExpressionSyntaxHighlighter::highlightBlock(const QString&)
...
 private:
-  IOWindow* m_parent;
+  std::function<ControlState(int row)> m_state_evaluator;
+  int m_column;
+};
+
+class InputStateLineEdit : public QLineEdit
+{
+public:
+  InputStateLineEdit(std::function<ControlState()> state_evaluator);
```
leoetlino, line 201: explicit

Source/Core/DolphinQt/Config/Mapping/IOWindow.cpp
```diff
@@ -230,7 +244,10 @@ void IOWindow::CreateMainLayout()
   m_range_slider = new QSlider(Qt::Horizontal);
   m_range_spinbox = new QSpinBox();
 
-  m_parse_text = new QLineEdit();
+  m_parse_text = new InputStateLineEdit([&]() {
```
leoetlino, line 246: `[&]()` can just be `[&]`, or even `[this]`

Source/Core/DolphinQt/Config/Mapping/IOWindow.h
```diff
@@ -74,7 +76,7 @@ class IOWindow final : public QDialog
   void UpdateOptionList();
   void UpdateDeviceList();
 
-  void UpdateExpression(std::string new_expression);
+  void UpdateExpression(std::string new_expression, bool force);
```
leoetlino, line 79: Consider taking an enum class (e.g. `UpdateMode` with two values, Normal and Force) because having boolean parameters hurts readability at call sites. When I see a call to UpdateExpression, I have no idea what the boolean does and I have to check the header to know and remember the name. (https://abseil.io/tips/94 explains the reasoning a bit more)

leoetlino (changes requested):

Source/Core/DolphinQt/Config/Mapping/IOWindow.h
AdmiralCurtiss: Fair point. I assume this is fine now?

leoetlino (approved):
Ignore the FifoCI changes, it's because this PR is based on an older commit that was broken

PR #9237, opened 2020-11-09
Common/MsgHandler: Add fmt-capable variants of the alert macros

lioncash:
Adds an interface that uses fmt under the hood, which is much more flexible than printf, particularly for localization purposes, given fmt supports positional formatters in a cross-platform manner out of the box with no configuration necessary.

Now alerts can be progressively migrated over as well, like with the logging macros.


JosJuice:
Please add the new translatable macros as `--keyword` parameters in `Languages/update-source-strings.sh`.

lioncash:
Done

BhaaLseN (approved):

PR #9238, opened 2020-11-09
CMakeLists: Make messages about FFMPEG / FRAMEDUMP more generic, deassociate with AVI file format

altimumdelta:
It's because after the Framedump updates, it will *support* more than just the AVI format, infact it already does, if we take user custom ini configs into account, except not on Windows, where it's not supporting anything other than AVI and x264 right now, still.

Secondly, PNG image dumping is still possible even if FFmpeg is found and enabled here, again with custom ini configs, so the wording was further changed to indicate that it's an additional option to PNG, not a switch, but ffmpeg encoding does become the default tho.


leoetlino:
This would require a buildbot config update for very little benefit.

altimumdelta:
Perhaps add it to a todo/queue list when buildbot configs would get updated for another bigger reason.

PR #9239, opened 2020-11-09
FrameDump Logging: Separate log type and migrate to fmt

altimumdelta:
In anticipation for framedump improvements, more logging may be added later, this would prevent log cluttering, as well as for framedump testing during development, to hide other things under _VIDEO_


altimumdelta:
Lint issue should be fixed.

BhaaLseN:
The fmt changes seem fine, altho it feels a bit weird that a few calls were already `xxx_LOG_FMT` while most others were just `xxx_LOG`.
I don't know if those changes are completed though; I didn't check all the messages that might be FrameDump related nor did I run an Android build to see if the logging configuration there works (or any other builds, for that matter).

JosJuice:
> The fmt changes seem fine, altho it feels a bit weird that a few calls were already `xxx_LOG_FMT` while most others were just `xxx_LOG`.

Billiard used the fmt variant when adding some new log calls in a recent PR if I recall correctly.

altimumdelta:
I was surprised as well, thought I was seeing things, then it clicked, those inconsistencies are due to changes of #9182 where the new method was used in cases where logging was touched.

PR #9241, opened 2020-11-10
WIP - DolphinQt/MenuBar: Decouple Frame and Audio Dump from Input Recorder

altimumdelta:
Further framedump improvement preparation. Separation from TAS Movies because of many reasons:

First it literally has nothing to do with TAS Movies (input recording) in code or otherwise and can be used on it's own.

Second, if things go according to my plan, FrameDump will become bigger in scope with more options and dialogs that would go right here into this new menu. It would eventually deserve it's own menu IMO, if not, perhaps a place in Tools menu or something.

Third, the whole "movies" (TAS Movies) moniker has been in name clash with what FrameDump is about for a long time, at least in my view, not that it's a problem, but if I'm here why not improve this as well. FrameDump encodes dumps and makes movies or videos, even if we take the strict meaning of those into account, the output of FrameDump is closer to a movie than not, whereas TAS Movies does not actually make anything comparable to a video let alone a movie, we all know it really creates input recording files. This PR is just the first part of solving that for good.

Warning: Due to sizable code rearrangement there is a lot of change indications in diff, however litterally nothing else has been changed(removed) except what is described above.

### DIFF
--- a/Source/Core/DolphinQt/GameList/GameList.cpp
+++ b/Source/Core/DolphinQt/GameList/GameList.cpp
@@ -115,7 +115,7 @@ void GameList::MakeListView()
   connect(m_list, &QTableView::customContextMenuRequested, this, &GameList::ShowContextMenu);
   connect(m_list->selectionModel(), &QItemSelectionModel::selectionChanged,
           [this](const QItemSelection&, const QItemSelection&) {
-            emit SelectionChanged(GetSelectedGame());
+            emit GameSelectionChanged(GetSelectedGame());
           });
 
   QHeaderView* hor_header = m_list->horizontalHeader();
@@ -255,7 +255,7 @@ void GameList::MakeGridView()
   connect(m_grid, &QTableView::customContextMenuRequested, this, &GameList::ShowContextMenu);
   connect(m_grid->selectionModel(), &QItemSelectionModel::selectionChanged,
           [this](const QItemSelection&, const QItemSelection&) {
-            emit SelectionChanged(GetSelectedGame());
+            emit GameSelectionChanged(GetSelectedGame());
           });
 }
 
@@ -265,7 +265,7 @@ void GameList::ShowHeaderContextMenu(const QPoint& pos)
   if (!menu_bar)
     return;
 
-  QMenu* const list_columns_menu = menu_bar->GetListColumnsMenu();
+  QMenu* const list_columns_menu = menu_bar->GetGameListColumnsSubMenu();
   if (!list_columns_menu)
     return;
 
--- a/Source/Core/DolphinQt/GameList/GameList.h
+++ b/Source/Core/DolphinQt/GameList/GameList.h
@@ -52,7 +52,7 @@ public:
 signals:
   void GameSelected();
   void NetPlayHost(const UICommon::GameFile& game);
-  void SelectionChanged(std::shared_ptr<const UICommon::GameFile> game_file);
+  void GameSelectionChanged(std::shared_ptr<const UICommon::GameFile> game_file);
   void OpenGeneralSettings();
 
 private:
--- a/Source/Core/DolphinQt/MainWindow.cpp
+++ b/Source/Core/DolphinQt/MainWindow.cpp
@@ -234,7 +234,7 @@ MainWindow::MainWindow(std::unique_ptr<BootParameters> boot_parameters,
     if (!movie_path.empty())
     {
       if (Movie::PlayInput(movie_path, &m_pending_boot->savestate_path))
-        emit RecordingStatusChanged(true);
+        emit INRECSStatusChanged(true);
     }
   }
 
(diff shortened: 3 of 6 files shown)


MayImilae:
I've always found it odd that Framedump was connected with TASing in that menu. So, I'm all for this. Of course I don't use TASing at all and use framedump very little, so, you know.

altimumdelta:
Initial preview: 

Screenshot_20201110_020558_Dolphin-Emu_Preview_MenuBar_Decouple_FFDUMP_TASMovies

MayImilae:
Hmm.... to be honest, I'd prefer both Movie and Framedump to be submenus of Tools. They are very niche to be in such high placement, imo.

altimumdelta:
Mnyah ... let's see, but the _Movie_ menu has some of what seems to be common switches that I think could be deassociated as well, since it pertains to what happens in emulation, they wouldn't fit in _View_ menu, would it be wise moving _Show Lag Counter_, _Show Frame Counter_, _Show Input Display_, _Show System Clock_ into the _Emulation_ menu?

Now, _Tools_ menu is quite populated as it is IMO, so _Movies_ should literally be moved in there so the items become nested further, right, while FrameDump would eventually be a dialog on it's own needing only a single button, but since dumping audio is not part of FrameDump, that would require a different approach, probably "merging" into _AV Dump ..._ button, which kinda brings us creepingly back close to "AVI", ... but thankfully the context and meaning is quite different, one such exception and only the GUI front-end.

Since none of that around FrameDump is ready yet, I think it's fine to just move _Dump Frames_ and _Dump Audio_ items directly into Tools menu without nesting for the time being, fine?

altimumdelta:
Allright, I did like this:

Screenshot_20201110_020558_Dolphin-Emu_Preview_MenuBar_Decouple_FFDUMP_TASMovies_V2-1

Screenshot_20201110_020558_Dolphin-Emu_Preview_MenuBar_Decouple_FFDUMP_TASMovies_V2-2

Screenshot_20201110_020558_Dolphin-Emu_Preview_MenuBar_Decouple_FFDUMP_TASMovies_V2-3

Except it makes these features even more burried, right when they should kinda need a kick to enocurage discovery/usage, but we might find some other remedy for that in the future.

MayImilae:
Oh that's clean! I really like that. As for discoverability... it will be reduced a bit, but, this is a niche feature for very high level users. And they won't really care what you put it, they use or don't use FrameDump for other reasons. For example, I don't use FrameDump is because, well, it sucks atm. It fails to do what I need it to do every time I try it, so now I just use shadowplay and ignore Framedump. But If you make it better, I'll revisit it. I think a lot of people are in a position like mine! So don't wory, people will find it! Also we'll be totally talking about this move on the Progress Report for sure!

altimumdelta:
> For example, I don't use FrameDump is because, well, it sucks atm.
Soon(TM)

> It fails to do what I need it to do every time I try it,
Oh please do tell as much as you can so it's made sure to be indcluded in further FrameDump improvements.
We know the alternatives drove over FrameDump, but I decided back then that eventually we still have to finish this because we've had so much detailed discussions and I made so much, I was clear, not exactly promises, but on that fence of being so close. I would feel crappy if this saga would end up with a wimper forgotten forever. Besides, Billiard's FrameDump fixes/cleanup may only come to fully shine with the API calls (the file saving mechanism in FrameDump.cpp) updated.


Lint issues fixed, sorry about that. Getting my first development days on quite a different environment (Kubuntu) than I was used to with MSVS. It'll take another couple of months* (to be safe) before I finish setting up my new PC with Win10 and VS2019 and all. But I have to thank Billiard for the great timing of the FrameDump PR raising the morale/motivation forcing me to explore another dev environment, gaining much needed experience and flexibility.

altimumdelta:
> For example, I don't use FrameDump is because, well, it sucks atm.
Soon(TM)

> It fails to do what I need it to do every time I try it,

Oh please do tell as much as you can so it's made sure to be indcluded in further FrameDump improvements.
We know the alternatives drove over FrameDump, but I decided back then that eventually we still have to finish this because we've had so much detailed discussions and I made so much, I was clear, not exactly promises, but on that fence of being so close. I would feel crappy if this saga would end up with a wimper forgotten forever. Besides, Billiard's FrameDump fixes/cleanup may only come to fully shine with the API calls (the file saving mechanism in FrameDump.cpp) updated.

Sorry for lint issues. Getting my first development days on quite a different environment (Kubuntu) than I was used to with MSVS. It'll take another couple of months* (to be safe) before I finish setting up my new PC with Win10 and VS2019 and all. But I have to thank Billiard for the great timing of the FrameDump PR raising the morale/motivation forcing me to explore another dev environment, gaining much needed experience and flexibility. 

I have installed "clang-format" package and it picked version 10 automatically, and used that with KDevelop to "Reformat Source", it claims to detect the clang-format file automatically, but it appears to put 4 spaces, whereas it was 2 before.

So I guess I would have to try using version 9 then if that's what the buildbot uses.

altimumdelta:
Lint issues fixed.

altimumdelta:
While at it, I was so tempted to do a cleanup first, but ended up just doing it all at once otherwise I'd be shuffling code that would very soon be removed/moved/renamed again. Detailed changelog may follow if necessary, expand the commit message for more info. 
Also refactor of Movies terminology is subject to change, it's not necessairly final, I have still gone through many variations and yes this is one of those tricky things, don't want something inoptimal to stick so I rather put in the effort and I did.

All the screenshots are still valid, except with some names such as "TAS Movies" is not longer "TAS Movies". It's funny how I realized that TAS isn't what this is about either, it's neither a movie* nor "TAS", input recording first and foremost, with the ability to feed tool-assisted input into the recorder. Furthermore, before any of this could be used for a movie*, it would first be a video after dumping frames or recording ehm a **video** of the recorded input track replaying in dolphin.

Hopefully that person who signaled motivation to upgrade "TAS Movies" Input Recorder, didn't went too deep into it yet, now this PR is just about the GUI stuff, PT2 will be based on this.

JosJuice:
This is way too many changes in one commit, and most of the changes aren't even an improvement.

altimumdelta:
Yeah now that I looked at the diff, I meant to say I expected MenuBar.cpp to be a diff mess and it is, better looking at the new version alone.

I suppose I got impatient and put TAS Movies refactor in here, I could despin that, but it won't make the MenuBar diff mess any different.

I don't know if I'll have the patience to spend first renaming, resorting, moving all separately, what I consider cleanup stuff, this already was tedious enough.

PR #9245, opened 2020-11-12
Wiimote: Use dynamic_cast where applicable

Ebola16:
Suggested by clang while looking around the Wiimote code. And removes a random semicolon.

### DIFF
--- a/Source/Core/Core/HW/Wiimote.cpp
+++ b/Source/Core/Core/HW/Wiimote.cpp
@@ -68,7 +68,7 @@ HIDWiimote* GetHIDWiimoteSource(unsigned int index)
   switch (GetSource(index))
   {
   case WiimoteSource::Emulated:
-    hid_source = static_cast<WiimoteEmu::Wiimote*>(::Wiimote::GetConfig()->GetController(index));
+    hid_source = dynamic_cast<WiimoteEmu::Wiimote*>(::Wiimote::GetConfig()->GetController(index));
     break;
 
   case WiimoteSource::Real:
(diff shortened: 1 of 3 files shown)


JosJuice:
We don't have RTTI, do we?

Ebola16:
Another change that fine with Android but Windows dislikes.
"'dynamic_cast' used on polymorphic type 'ControllerEmu::EmulatedController' with /GR-". I'm not sure what the "with /GR-" refers to here.
But I just read that RTTI should be avoided if possible. I'll look into alternatives.

JosJuice:
`/GR-` is a compiler flag for MSVC: https://docs.microsoft.com/en-us/cpp/build/reference/gr-enable-run-time-type-information

Ebola16:
Oh that explains things. We don't have RTTI. And `typeid(a) == typeid(b)` probably isn't a good enough replacement for `dyanmic_cast`. Double or multiple dispatch are possible alternatives here. Regardless, this isn't the correct solution and it may be some time before I look into those alternative methods so I'll close this for now.

JosJuice:
It feels a bit to me like you're trying to solve a problem that doesn't exist. I don't see how multiple dispatch would improve this code.

Keep in mind that just because Android Studio makes a suggestion doesn't necessarily mean that it's a change you need to make – whether the change makes sense depends on what the program looks like as a whole.

Ebola16:
Sometimes I need to just go down the rabbit hole if I don't full understand what's going on. With our `static_cast` uses, the "object referred to doesn't contain the type casted to as a base class" seems possible so I tried the `dynamic_cast` suggestion.

But if you still think this is a non-issue, I'll drop it. That'll save me some time and prevent some unnecessary email notifications for you. I haven't looked into multiple dispatch yet, I just saw that it was an alternative to RTTI.

PR #9246, opened 2020-11-12
DolphinQt: Don't return 0x04 bytes from tr

JosJuice:
`tr` calls with more than one argument would have a 0x04 byte in the returned string when no translation was found (which always is the case when using Dolphin in English).


CookiePLMonster (approved):
Looks good.

PR #9247, opened 2020-11-13
Fix updater not always cleaning up temp directory

Dentomologist:
Updater was only deleting TempUpdate folder when an error occured.

### DIFF
--- a/Source/Core/UpdaterCommon/UpdaterCommon.cpp
+++ b/Source/Core/UpdaterCommon/UpdaterCommon.cpp
@@ -755,10 +755,10 @@ bool RunUpdater(std::vector<std::string> args)
   UI::SetDescription("Performing Update...");
 
   bool ok = PerformUpdate(todo, opts.install_base_path, opts.content_store_url, temp_dir);
+  CleanUpTempDir(temp_dir, todo);
   if (!ok)
   {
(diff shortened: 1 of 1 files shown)


leoetlino (approved):

leoetlino:
Fwiw, the bug wasn't present in the initial implementation (07c160310ec67b30116a5ed817282bdd124fe927):

```cpp
  bool ok = PerformUpdate(todo, opts.install_base_path, opts.content_store_url, temp_dir);
  if (!ok)
    fprintf(log_fp, "Failed to apply the update.\n");

  CleanUpTempDir(temp_dir, todo);
  return !ok;
}
```

StLouisCPhT:
Thanks for the fix! It was getting annoying having to delete that manually from the macOS Application folder. Any chance that the updater log could be changed to be saved with the rest of the logs instead of off by itself?

Dentomologist:
> Thanks for the fix! It was getting annoying having to delete that manually from the macOS Application folder. Any chance that the updater log could be changed to be saved with the rest of the logs instead of off by itself?

As it happens, I'm working on a separate PR that will include that.

StLouisCPhT:
I look forward to it :) Even if it is next decade before it is ready.

PR #9249, opened 2020-11-14
add a working fxaa + scanlines shader

Nerboruto:
first contribution
sorry if i did something wrong..

### DIFF
--- a/Data/Sys/Shaders/FXAA-SharpScanlines.glsl
+++ b/Data/Sys/Shaders/FXAA-SharpScanlines.glsl
@@ -1,0 +1,150 @@
+// scanlines shader by Nerboruto 2020
+//
+// fxaa Copyright (C) 2013 mudlord
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy 
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights 
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell 
+// copies of the Software, and to permit persons to whom the Software is furnished
+// to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in all 
+// copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR 
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, 
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER 
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+/*
+[configuration]
+
+[OptionBool]
+GUIName = Sharp filter Active
+OptionName = SHARP_ON
+DefaultValue = true
+
+[OptionRangeFloat]
+GUIName = Sharp Amount
+OptionName = SHARP_C
+MinValue = 0.50
+MaxValue = 1.50
+StepAmount = 0.10
+DefaultValue = 0.50
+DependentOption = SHARP_ON
+
+[OptionBool]
+GUIName = Scanlines filter Active
+OptionName = SCAN_ON
+DefaultValue = true
+
+[OptionBool]
+GUIName = DoubleScanline
+OptionName = SCAN_D
+DefaultValue = false
+DependentOption = SCAN_ON
+
+[OptionRangeFloat]
+GUIName = Scanlines Amount
+OptionName = SCAN_LINES
+MinValue = 0.1
+MaxValue = 0.3
+StepAmount = 0.01
+DefaultValue = 0.2
+DependentOption = SCAN_ON
+
+[OptionRangeFloat]
+GUIName = White Amount
+OptionName = SCAN_W
+MinValue = 0.5
+MaxValue = 1.0
+StepAmount = 0.01
+DefaultValue = 0.85
+DependentOption = SCAN_ON
+
+[/configuration]
+*/
+
+#define FXAA_REDUCE_MIN		(1.0/ 128.0)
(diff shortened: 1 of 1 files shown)


Data/Sys/Shaders/FXAA Scanlines.glsl
```diff
@@ -0,0 +1,82 @@
...
+// Copyright (C) 2013 mudlord
+
+// Everyone is permitted to copy and distribute verbatim or modified
+// copies of this license document, and changing it is allowed as long
+// as the name is changed.
+
+//			DO WHAT THE FUCK YOU WANT TO PUBLIC LICENSE
+//	TERMS AND CONDITIONS FOR COPYING, DISTRIBUTION AND MODIFICATION
+
+// 0. You just DO WHAT THE FUCK YOU WANT TO.
```
Rumi-Larry, line 13: I'm sure this is against the code style

sonic2kk: Amazingly, this seems to be a valid GPL license:

https://en.m.wikipedia.org/wiki/WTFPL
http://www.wtfpl.net

BhaaLseN:
You should pick one of the two files and put it in the right location. I don't think the root directory is the correct one (but I'm not sure about the other one either)

Nerboruto:
> You should pick one of the two files and put it in the right location. I don't think the root directory is the correct one (but I'm not sure about the other one either)

sorry...
I wanted to replace the first version of the file with the second but I mistakenly put it in the root...

Nerboruto:
feedback are welcome...

iwubcode (changes requested):
My request would be to have one shader that does scanlines and _optionally_ can be defined as "sharp" or not.

Nerboruto:
> La mia richiesta sarebbe quella di avere 
config
uno shader che faccia scanline e _opzionalmente_ possa essere definito come "sharp" o meno.

Data/Sys/Shaders/FXAA Scanlines.glsl
delroth: WTFPL is a bad license. It's not widely recognized as something valid worldwide, it does not provide a warranty waiver, etc.

If you don't care at all about usage of what you release, I would suggest using either:
* The CC0 license: https://creativecommons.org/publicdomain/zero/1.0/
* The MIT license (which requires attribution -- but given that this ships as source it makes very little difference): https://opensource.org/licenses/MIT

FWIW I would strongly oppose adding WTFPL licensed components to Dolphin.

Nerboruto: I used the header already present in the fxaa file

iwubcode:
At the very least, please squash your commits.  I do not have any major concerns with the current implementation.  Thanks for adding the options!

Nerboruto:
> At the very least, please squash your commits. I do not have any major concerns with the current implementation. Thanks for adding the options!

I don't know how to do it..
 checks say "PR not built because Nerboruto is not auto-trusted"

Data/Sys/Shaders/FXAA Scanlines.glsl
leoetlino: Couldn't you relicence it under MIT?

Nerboruto: > Couldn't you relicence it under MIT?

I don't know how to behave for the part dell'fxaa that was already present since 2013 by mudlord
I don't know if it is appropriate to remove it and see some other type of filtering

Nerboruto: it is OK?

PatrickFerry:
Please squash the commits with

`git rebase -i HEAD^X `

Replace X with the number of commits you have. This is in interactive rebase. An editor will appear and you have to edit every commit but the top one to say squash, so it should look something like this

```
pick aaaaaaa commit 1
pick bbbbbbb commit 2
pick ccccccc commit 3
....
```
And you have to make it edit it to something like this
```
pick aaaaaaa commit 1
squash bbbbbbb commit 2
squash ccccccc commit 3
....
```
It will likely open in VIM if that's set as your preferred editor, look up how to edit, save and quit in VIM if you don't know how or change your git preferences for the editor you normally use.

Nerboruto:
I tried to make a squash but it doesn't work if someone can help me ...

AdmiralCurtiss:
@iwubcode Can you re-check this?

iwubcode:
If we want to temporarily merge this, that's fine but I would be deleting it in #10362 .  As I mentioned in #9875 we shouldn't be creating our own implementation but rather picking one of the agreed on [standard shaders](http://filthypants.blogspot.com/2015/04/more-crt-shaders.html) chosen by the community.

If you look at #10362 you'll see that a user was actually porting mame's crt-geom which I think is the more appropriate way to go.

Nerboruto:
ok no problem was born as an experiment but I liked result a lot even at native resolution, anyway lately I've been having fun with color reduction dithering end aperture simulation... I don't like rounding of the geometry.

iwubcode:
Yeah, I understand.  The geom one apparently has the option to turn off the curvature.  See [example1](https://user-images.githubusercontent.com/1188012/152717257-0198b6b6-872e-4fb0-a1a5-b3779ac0663b.png) and [example2](https://user-images.githubusercontent.com/1188012/152717263-654eefc1-07bf-41db-9c0a-119073eae89a.png).  (the curvature isn't working properly atm anyway but that's another thing).

I'm not really a scanline person but from what I've seen online it seems like crt-royale is the most popular.  Hopefully that gets implemented in Dolphin one day.

Nerboruto:
Yes I know...
personally I don't like it at all the result is blurry ...
try mine with native resolution + msaa 2x + with 1080p desktop resolution on 1080p monitor... run fine on shit gt710.
if the monitor has more than 1080p you have to use double scanline... on some monitors with many ppi you have to change the brightness of the scanlines ..

iwubcode:
Again, fine for other devs to chime in but my take is that we should have a limited set of built-in post processing shaders and that those that are included are  A) high quality.  B) useful.  C) ideally standard implementations  (if one exists).

It's because of this I've done a small purge in #10362 and I plan on going through on a second pass to remove more.

I encourage users to distribute their own post processing shaders.  In the same way that users can grab reshade shaders.  However, my take is the base shipped set of shaders with Dolphin should be rather small.

Nerboruto:
ok thank you...

EpilodieResearch:
Request on the 'chrismas' shader -- several are candidates for removal as they are not useful, but I would contend that 'chrismas' is both funny and worth keeping in for sentimental value. Very similarly to the green bell pepper within the GIMP imaging program.

Many thanks if so, please. Hope you all are having a very fine and well day today.

PR #9250, opened 2020-11-15
Fix file rename errors on Windows

Dentomologist:
On Windows, when the Rename function fails to replace an existing file
it will now retry the operation multiple times with increasingly long
delays between attempts.  This fixes transient rename failures.

I've been getting sporadic yet annoyingly frequent errors saying:
'IOS_FS: Failed to rename temporary FST file'
These typically appear on startup but I've also gotten them randomly.

Investigation shows this happens when the Windows ReplaceFile function
returns the error ERROR_UNABLE_TO_REMOVE_REPLACED.  That happens in the
context of using ReplaceFile to perform an atomic file overwrite, which
is required when saving updates to a file to avoid corruption.  The
error mainly happens with the /Wii/fst.bin file but I've seen it
happen with multiple other files as well.

I haven't been able to definitively pin down why the error occurs,
though online discussions suggest antivirus scanning may be a major
culprit.  That said, I've excluded the Dolphin folder from Windows
Defender scans to no avail and don't have any other antivirus running,
so this is likely to be a problem others are experiencing as well.

The number and duration of retry delays is arbitrary but I feel like a
combined second or so in the worst case is an acceptable tradeoff for
the reduction (actually elimination in my experience) of those errors.
This is even more true when you consider the time it takes to read and
dismiss the error dialogs.

### DIFF
--- a/Source/Core/Common/FileUtil.cpp
+++ b/Source/Core/Common/FileUtil.cpp
@@ -3,6 +3,7 @@
 // Refer to the license.txt file included.
 
 #include <algorithm>
+#include <chrono>
 #include <cstddef>
 #include <cstdio>
 #include <cstring>
@@ -11,6 +12,7 @@
 #include <limits.h>
 #include <string>
 #include <sys/stat.h>
+#include <thread>
 #include <vector>
 
 #include "Common/Assert.h"
@@ -276,33 +278,65 @@ bool DeleteDir(const std::string& filename)
   return false;
 }
 
+// Repeatedly invokes func until it returns true or max_attempts failures.
+// Waits after each failure, with each delay doubling in length.
+template <typename FuncType>
+static bool AttemptMaxTimesWithExponentialDelay(int max_attempts, std::chrono::milliseconds delay,
+                                                std::string_view func_name, const FuncType& func)
+{
+  for (int failed_attempts = 0; failed_attempts < max_attempts; ++failed_attempts)
+  {
+    if (func())
+    {
+      return true;
+    }
+    if (failed_attempts + 1 < max_attempts)
+    {
+      INFO_LOG_FMT(COMMON, "{} attempt failed, delaying for {} milliseconds", func_name,
+                   delay.count());
+      std::this_thread::sleep_for(delay);
+      delay *= 2;
+    }
+  }
+  return false;
+}
+
 // renames file srcFilename to destFilename, returns true on success
 bool Rename(const std::string& srcFilename, const std::string& destFilename)
 {
   INFO_LOG_FMT(COMMON, "Rename: {} --> {}", srcFilename, destFilename);
 #ifdef _WIN32
-  auto sf = UTF8ToTStr(srcFilename);
-  auto df = UTF8ToTStr(destFilename);
-  // The Internet seems torn about whether ReplaceFile is atomic or not.
-  // Hopefully it's atomic enough...
-  if (ReplaceFile(df.c_str(), sf.c_str(), nullptr, REPLACEFILE_IGNORE_MERGE_ERRORS, nullptr,
-                  nullptr))
-    return true;
-  // Might have failed because the destination doesn't exist.
-  if (GetLastError() == ERROR_FILE_NOT_FOUND)
-  {
-    if (MoveFile(sf.c_str(), df.c_str()))
-      return true;
-  }
-  ERROR_LOG_FMT(COMMON, "Rename: MoveFile failed on {} --> {}: {}", srcFilename, destFilename,
-                GetLastErrorString());
+  const std::wstring source_wstring = UTF8ToTStr(srcFilename);
+  const std::wstring destination_wstring = UTF8ToTStr(destFilename);
+
+  // On Windows ReplaceFile can fail spuriously due to antivirus checking or other noise.
+  // Retry the operation with increasing delays, and if none of them work there's probably a
+  // persistent problem.
+  const bool success = AttemptMaxTimesWithExponentialDelay(
+      3, std::chrono::milliseconds(5), "Rename", [&source_wstring, &destination_wstring] {
+        if (ReplaceFile(destination_wstring.c_str(), source_wstring.c_str(), nullptr,
+                        REPLACEFILE_IGNORE_MERGE_ERRORS, nullptr, nullptr))
+        {
+          return true;
+        }
+        // Might have failed because the destination doesn't exist.
+        if (GetLastError() == ERROR_FILE_NOT_FOUND)
+        {
+          return MoveFile(source_wstring.c_str(), destination_wstring.c_str()) != 0;
+        }
+        return false;
+      });
+  constexpr auto error_string_func = GetLastErrorString;
 #else
-  if (rename(srcFilename.c_str(), destFilename.c_str()) == 0)
-    return true;
-  ERROR_LOG_FMT(COMMON, "Rename: rename failed on {} --> {}: {}", srcFilename, destFilename,
-                LastStrerrorString());
+  const bool success = rename(srcFilename.c_str(), destFilename.c_str()) == 0;
+  constexpr auto error_string_func = LastStrerrorString;
 #endif
-  return false;
+  if (!success)
+  {
+    ERROR_LOG_FMT(COMMON, "Rename: rename failed on {} --> {}: {}", srcFilename, destFilename,
+                  error_string_func());
+  }
+  return success;
 }
 
 #ifndef _WIN32


AdmiralCurtiss:
see also #9023

AdmiralCurtiss:
That said, retrying a rename on failure is not the _worst_ idea in the world, though 8 attempts with a total of ~1 second delay seems pretty excessive. Imagine if you had a case where it legitimately can't rename files due to eg. missing rights and something attempted to rename even just 10 files in succession.

Dentomologist:
I'm totally open to adjusting the number and timing of delays.  The current set were chosen with an eye to minimizing the likelihood of a transient error still popping up a dialog, but your point about multiple files is well taken.

I'll read up on #9023 but I notice it's from back in August.  If finishing it would take a while we could use this PR (with adjusted delays; maybe 5/10/25 milliseconds) as a stopgap until #9023 is done.

Source/Core/Common/FileUtil.cpp
```diff
@@ -276,33 +279,64 @@ bool DeleteDir(const std::string& filename)
...
+    }
+    // Might have failed because the destination doesn't exist.
+    if (GetLastError() == ERROR_FILE_NOT_FOUND)
+    {
+      return MoveFile(source_ptr, destination_ptr) != 0;
+    }
+    return false;
+  };
+  using namespace std::chrono_literals;
+  constexpr std::array delays_after_failures = {5ms, 10ms, 25ms, 50ms, 100ms, 250ms, 500ms};
```
leoetlino, line 327: This seems a bit weird... I think some kind of exponential backoff would make more sense than having a manual list of delays?

Source/Core/Common/FileUtil.cpp
```diff
@@ -276,33 +279,68 @@ bool DeleteDir(const std::string& filename)
   return false;
 }
 
+// Repeatedly invokes func until it returns true or max_attempts failures.
+// Waits after each failure, with each delay doubling in length.
+template <typename FuncType>
+bool AttemptMaxTimesWithExponentialDelay(const FuncType& func, int max_attempts,
```
lioncash, line 285:
```suggestion
static bool AttemptMaxTimesWithExponentialDelay(const FuncType& func, int max_attempts,
```

since this is only used in this translation unit

Source/Core/Common/FileUtil.cpp
```diff
@@ -276,33 +279,68 @@ bool DeleteDir(const std::string& filename)
...
+{
+  for (int failed_attempts = 0; failed_attempts < max_attempts; ++failed_attempts)
+  {
+    if (func())
+    {
+      return true;
+    }
+    if (failed_attempts + 1 < max_attempts)
+    {
+      INFO_LOG_FMT(COMMON, "Rename attempt failed, delaying for {} milliseconds", delay.count());
```
leoetlino, line 296: Since this is a generic utility function, I think the name of the operation should be passed as a parameter to this function:

```suggestion
      INFO_LOG_FMT(COMMON, "{} attempt failed, delaying for {} milliseconds", operation_name, delay.count());
```

Source/Core/Common/FileUtil.cpp
```diff
@@ -276,33 +279,68 @@ bool DeleteDir(const std::string& filename)
...
-  // The Internet seems torn about whether ReplaceFile is atomic or not.
-  // Hopefully it's atomic enough...
-  if (ReplaceFile(df.c_str(), sf.c_str(), nullptr, REPLACEFILE_IGNORE_MERGE_ERRORS, nullptr,
-                  nullptr))
-    return true;
-  // Might have failed because the destination doesn't exist.
-  if (GetLastError() == ERROR_FILE_NOT_FOUND)
-  {
-    if (MoveFile(sf.c_str(), df.c_str()))
+  // Store converted strings in local variables so cstrings don't point to temporary objects
```
leoetlino, line 309: This shouldn't be required. source_wstring and destination_wstring won't go out of scope and won't get destructed until `attempt_func` completes, so there is no lifetime issue here.

That said, the original naming is pretty terrible so I guess we can keep the new names but get rid of source_ptr and destination_ptr.

Source/Core/Common/FileUtil.cpp
```diff
@@ -276,33 +279,68 @@ bool DeleteDir(const std::string& filename)
...
+  const std::wstring source_wstring = UTF8ToTStr(srcFilename);
+  const std::wstring destination_wstring = UTF8ToTStr(destFilename);
+  const wchar_t* const source_ptr = source_wstring.c_str();
+  const wchar_t* const destination_ptr = destination_wstring.c_str();
+
+  // On Windows ReplaceFile can fail spuriously due to antivirus checking or other noise.
+  // Retry the operation with increasing delays, and if none of them work there's probably a
+  // persistent problem.
+  const auto attempt_func = [&, source_ptr, destination_ptr]() {
```
leoetlino, line 318:
```suggestion
  const auto attempt_func = [&source_wstring, &destination_ptr] {
```

Source/Core/Common/FileUtil.cpp
```diff
@@ -276,33 +279,68 @@ bool DeleteDir(const std::string& filename)
...
+// Repeatedly invokes func until it returns true or max_attempts failures.
+// Waits after each failure, with each delay doubling in length.
+template <typename FuncType>
+static bool AttemptMaxTimesWithExponentialDelay(const FuncType& func, int max_attempts,
+                                                std::chrono::milliseconds delay)
+{
+  for (int failed_attempts = 0; failed_attempts < max_attempts; ++failed_attempts)
+  {
+    if (func())
+    {
```
leoetlino, line 291: To make this less noisy, I'd get rid of the braces here.

Source/Core/Common/FileUtil.cpp
```diff
@@ -276,33 +279,68 @@ bool DeleteDir(const std::string& filename)
   return false;
 }
 
+// Repeatedly invokes func until it returns true or max_attempts failures.
+// Waits after each failure, with each delay doubling in length.
+template <typename FuncType>
+static bool AttemptMaxTimesWithExponentialDelay(const FuncType& func, int max_attempts,
```
leoetlino, line 285: I'd recommend having the function to retry be the last parameter to make caller sites look a bit better once formatted with clang-format.

Source/Core/Common/FileUtil.cpp
```diff
@@ -276,33 +279,68 @@ bool DeleteDir(const std::string& filename)
...
+  const wchar_t* const source_ptr = source_wstring.c_str();
+  const wchar_t* const destination_ptr = destination_wstring.c_str();
+
+  // On Windows ReplaceFile can fail spuriously due to antivirus checking or other noise.
+  // Retry the operation with increasing delays, and if none of them work there's probably a
+  // persistent problem.
+  const auto attempt_func = [&, source_ptr, destination_ptr]() {
+    if (ReplaceFile(destination_ptr, source_ptr, nullptr, REPLACEFILE_IGNORE_MERGE_ERRORS, nullptr,
+                    nullptr))
+    {
```
leoetlino, line 321: Same regarding the braces here -- this would be more readable if it were more compact.

Source/Core/Common/FileUtil.cpp
```diff
@@ -276,33 +279,68 @@ bool DeleteDir(const std::string& filename)
...
       return true;
-  }
-  ERROR_LOG_FMT(COMMON, "Rename: MoveFile failed on {} --> {}: {}", srcFilename, destFilename,
-                GetLastErrorString());
+    }
+    // Might have failed because the destination doesn't exist.
+    if (GetLastError() == ERROR_FILE_NOT_FOUND)
+    {
```
leoetlino, line 326: braces

Source/Core/Common/FileUtil.cpp
```diff
@@ -276,33 +279,68 @@ bool DeleteDir(const std::string& filename)
...
+      return MoveFile(source_ptr, destination_ptr) != 0;
+    }
+    return false;
+  };
+  const bool success =
+      AttemptMaxTimesWithExponentialDelay(attempt_func, 3, std::chrono::milliseconds(5));
```
leoetlino, line 332: Since the lambda isn't reused, it would be clearer to just not assign it to a variable and pass it directly to this function call.

leoetlino (changes requested):

Source/Core/Common/FileUtil.cpp
```diff
@@ -3,6 +3,8 @@
 // Refer to the license.txt file included.
 
 #include <algorithm>
+#include <array>
```
leoetlino, line 6: This change can be reverted since std::array isn't used anymore

Source/Core/Common/FileUtil.cpp
Dentomologist: I know Dolphin has a lot of them but I'm not a fan of braceless conditionals/loops; they reduce vertical spacing but allow for things like the Apple SSL bug. Unless it's a deal-breaker I'd rather leave them as is.

This would be a nice place for a Ruby-style "return if [condition]" statement, if only c++ had those.

Source/Core/Common/FileUtil.cpp
Dentomologist: The comment was referring to why I created source_wstring instead of assigning source_ptr = UTF8ToTStr(srcFilename).c_str(), but given the spacing I see how that's ambiguous.  Getting rid of source_ptr makes the comment redundant anyway so I'll get rid of it.

leoetlino (approved):

PR #9253, opened 2020-11-16
Use positional arguments in all translatable fmt strings

JosJuice:
We want to use positional arguments in translatable strings that have more than one argument so that translators can change the order of them, but the question is: Should we also use positional arguments in translatable strings with only one argument? I think it makes most sense that way, partially so that translators don't even have to be aware of the non-positional syntax and partially because "translatable strings use positional arguments" is an easier rule for us to remember than "transitional strings which have more than one argument use positional arguments". But let me know if you have a different opinion.

### DIFF
--- a/Source/Core/Core/IOS/USB/Bluetooth/BTReal.cpp
+++ b/Source/Core/Core/IOS/USB/Bluetooth/BTReal.cpp
@@ -560,8 +560,8 @@ bool BluetoothReal::OpenDevice(libusb_device* device)
   const int ret = libusb_open(m_device, &m_handle);
   if (ret != 0)
   {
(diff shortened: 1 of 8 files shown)


BhaaLseN:
Coming mostly from a C# background, I _only_ know the positional variant; and as such would be in favor of using it in general (at least for translatable strings). At least from my experience there, translators are more likely to understand what it means if a string contains `{0}`, `{1}` etc. rather than just a bunch of `{}` with no apparent content.
For other stuff, I have no real preference; the thing that gets me more than the absence of position is the way certain format specifiers work (mainly the combination of the alternate form specifier with a total width, which includes the specifier in it).

leoetlino (approved):

BhaaLseN (approved):

PR #9254, opened 2020-11-17
Fallback Region Option

flagrama:
A user-selected fallback to use instead of the default PAL

This can be used for unknown region or region free titles to give them the ability to force a different region than the current fallback region of PAL. This can be useful if a user is trying to play a region free title that is originally NTSC and expects to be run at NTSC speeds. This may be done when a user attempts to dump a WAD of their own without understanding the settings they have chosen, or could be an intentional decision by a developer of a ROM hack that can be injected into a Virtual Console WAD.

This is primarily a proof-of-concept in its current stage.

### DIFF
--- a/Source/Core/Core/Config/MainSettings.h
+++ b/Source/Core/Core/Config/MainSettings.h
@@ -7,6 +7,7 @@
 #include <string>
 
 #include "Common/Config/Config.h"
+#include "DiscIO/Enums.h"
 
 namespace PowerPC
 {
@@ -88,6 +89,7 @@ extern const Info<u32> MAIN_CUSTOM_RTC_VALUE;
 extern const Info<bool> MAIN_AUTO_DISC_CHANGE;
 extern const Info<bool> MAIN_ALLOW_SD_WRITES;
 extern const Info<bool> MAIN_ENABLE_SAVESTATES;
+extern const Info<DiscIO::Region> MAIN_FALLBACK_REGION;
 
 // Main.DSP
 
--- a/Source/Core/Core/ConfigManager.cpp
+++ b/Source/Core/Core/ConfigManager.cpp
@@ -31,6 +31,7 @@
 #include "Core/Analytics.h"
 #include "Core/Boot/Boot.h"
 #include "Core/CommonTitles.h"
+#include "Core/Config/MainSettings.h"
 #include "Core/Config/SYSCONFSettings.h"
 #include "Core/ConfigLoaders/GameConfigLoader.h"
 #include "Core/Core.h"
@@ -929,18 +930,7 @@ bool SConfig::SetPathsAndGameMetadata(const BootParameters& boot)
 
 DiscIO::Region SConfig::GetFallbackRegion()
 {
-  // Fall back to the system menu region, if possible.
-  IOS::HLE::Kernel ios;
-  const IOS::ES::TMDReader system_menu_tmd = ios.GetES()->FindInstalledTMD(Titles::SYSTEM_MENU);
-  if (system_menu_tmd.IsValid())
-  {
-    const DiscIO::Region region = system_menu_tmd.GetRegion();
-    if (region != DiscIO::Region::Unknown)
-      return region;
-  }
-
-  // Fall back to PAL.
-  return DiscIO::Region::PAL;
+  return Config::Get(Config::MAIN_FALLBACK_REGION);
 }
 
 DiscIO::Language SConfig::GetCurrentLanguage(bool wii) const
--- a/Source/Core/Core/Config/MainSettings.cpp
+++ b/Source/Core/Core/Config/MainSettings.cpp
@@ -12,6 +12,7 @@
 #include "Core/HW/Memmap.h"
 #include "Core/HW/SI/SI_Device.h"
 #include "Core/PowerPC/PowerPC.h"
+#include "DiscIO/Enums.h"
 #include "VideoCommon/VideoBackendBase.h"
 
 namespace Config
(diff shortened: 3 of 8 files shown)


JosJuice:
If we're going to have a setting for this, the setting should be checked before checking the region of the system menu, as the region of the system menu is rather annoying to change.

It would be simpler if we used the `Region` enum as the type of the setting instead of a string.

I personally don't think it makes sense to have both "override" and "fallback" in the name. It's either an override or a fallback, not both. The way it's implemented, I would not call it an override, as it has no effect for games with a known region.

flagrama:
> 
> 
> If we're going to have a setting for this, the setting should be checked before checking the region of the system menu, as the region of the system menu is rather annoying to change.

I can certainly do that.
> It would be simpler if we used the `Region` enum as the type of the setting instead of a string.

I tried that at first, didn't work immediately, went with strings for the POC. I definitely plan to look into this more.
> I personally don't think it makes sense to have both "override" and "fallback" in the name. It's either an override or a fallback, not both. The way it's implemented, I would not call it an override, as it has no effect for games with a known region.

That's fair I suppose. I named it as such because it overrides the default fallback of forcing PAL if no other method exists to find one. (not identifiable from the running title, no System Menu installed)

I'll try to get some of this done soon.

JosJuice:
Hm, maybe "fallback override" does actually make sense if it overrides the system menu check... I would appreciate if someone else could weigh in on this.

flagrama:
The fallback override should now take precedence over the current installed Wii System Menu. It also uses the `DiscIO::Region` enum instead of strings. Converting between the Combo Box selection and the region still uses separate index values because I personally find Disabled at the top of the list much more user-friendly than having it fourth in the list, not even as the last one, that using the Region enum directly would require.

leoetlino (changes requested):
> Hm, maybe "fallback override" does actually make sense if it overrides the system menu check... I would appreciate if someone else could weigh in on this.

I agree with this reasoning. However, I'm not sure most users know what *fallback* it's overriding, or even that the system menu region is used as a fallback... Removing the "override" wouldn't be a good solution because it *is* technically an override and you can disable the override, so I think what we should do is add an explanation in the UI.

Source/Core/Core/ConfigManager.cpp
```diff
@@ -929,6 +931,13 @@ bool SConfig::SetPathsAndGameMetadata(const BootParameters& boot)
 
 DiscIO::Region SConfig::GetFallbackRegion()
 {
+  // Fall back to the override option if set
+  const DiscIO::Region fallback = SConfig::GetInstance().m_fallback_region_override;
```
leoetlino, line 935: Missed this one, but this should use Config::Get as well, of course.

flagrama:
> [...] so I think what we should do is add an explanation in the UI.

How does this work for an explanation as a QLabel under the form row?
> Overrides the fallback region for titles that do not have one.
> 
> Titles without a region that can be determined automatically will fallback to this option. If this option is disabled, instead the region will fallback to the installed Wii System Menu region, or if one cannot be found, the PAL region.

I'm certainly not a wordsmith, so am up for changing it any way it is needed.

leoetlino:
I'd remove the first sentence as vertical space is at a premium in the setting windows:
 
> Dolphin will use this for titles whose region cannot be determined automatically. If this option is disabled, the region of the installed Wii System Menu will be used, or if one cannot be found, the PAL region.

flagrama:
I've added that explanation text and it appears like so:
image

flagrama:
Appearance after latest commit:
image

Source/Core/Core/ConfigManager.cpp
```diff
@@ -929,18 +930,8 @@ bool SConfig::SetPathsAndGameMetadata(const BootParameters& boot)
 
 DiscIO::Region SConfig::GetFallbackRegion()
```
JosJuice, line 931: You could probably get rid of this function entirely and just call `Config::Get(Config::MAIN_FALLBACK_REGION)` instead.

flagrama: It's used for GameCube and Wii region fallbacks, so I'm not sure I want to change it from one function being used twice to using a configuration var in two different places. If it is indeed preferred, I can make that change though.

flagrama: > flagrama: JosJuice, do you definitely want me to make the change mentioned in your comment?
JosJuice: It's not too important
JosJuice: Though if you don't want to make the change, then I would suggest making the function's implementation one line instead of two lines
JosJuice: You can return the result of Config::Get directly without storing it to a variable

Went with the alternative given.

Source/Core/Core/Config/MainSettings.h
```diff
@@ -4,6 +4,7 @@
 
 #pragma once
 
+#include <DiscIO/Enums.h>
 #include <string>
 
 #include "Common/Config/Config.h"
```
JosJuice, line 10:
```suggestion
#include <string>

#include "Common/Config/Config.h"
#include "DiscIO/Enums.h"
```

Source/Core/Core/ConfigManager.cpp
```diff
@@ -11,6 +11,7 @@
 #include <sstream>
 #include <variant>
 
+#include <Core/Config/MainSettings.h>
```
JosJuice, line 14: Same here. (Place the include below `#include "Common/Config/Config.h"`.)

leoetlino: Also, `#include "name"` for Dolphin headers, not `#include <name>`.

leoetlino (approved):

PR #9255, opened 2020-11-17
Android: Gradle and dependency updates with cleanup

Ebola16:
Changes should be rather simple except:
- Id references will no longer be final in Gradle. `MainPresenter` can't use them as switch cases.
- Android Studio wanted me to extend `Service` in `DirectoryInitialization` due to its usage in `AndroidManifest`.

### DIFF
--- a/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/activities/AppLinkActivity.java
+++ b/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/activities/AppLinkActivity.java
@@ -32,16 +32,22 @@ public class AppLinkActivity extends FragmentActivity
   protected void onCreate(Bundle savedInstanceState)
   {
     super.onCreate(savedInstanceState);
-    Intent intent = getIntent();
-    Uri uri = intent.getData();
+    Uri uri = getIntent().getData();
 
-    Log.v(TAG, uri.toString());
-
-    if (uri.getPathSegments().isEmpty())
+    if (uri != null)
     {
-      Log.e(TAG, "Invalid uri " + uri);
-      finish();
-      return;
+      Log.v(TAG, uri.toString());
+
+      if (uri.getPathSegments().isEmpty())
+      {
+        Log.e(TAG, "Invalid uri " + uri);
+        finish();
+        return;
+      }
+    }
+    else
+    {
+      throw new IllegalStateException("Could not get data uri.");
     }
 
     AppLinkHelper.AppLinkAction action = AppLinkHelper.extractAction(uri);
--- a/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/fragments/ConvertFragment.java
+++ b/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/fragments/ConvertFragment.java
@@ -34,7 +34,7 @@ public class ConvertFragment extends Fragment implements View.OnClickListener
   {
     private int mValuesId = -1;
     private int mCurrentPosition = -1;
-    private ArrayList<Runnable> mCallbacks = new ArrayList<>();
+    private final ArrayList<Runnable> mCallbacks = new ArrayList<>();
 
     @Override
     public void onItemSelected(AdapterView<?> adapterView, View view, int position, long id)
@@ -109,10 +109,10 @@ public class ConvertFragment extends Fragment implements View.OnClickListener
   private static final int COMPRESSION_LZMA2 = 4;
   private static final int COMPRESSION_ZSTD = 5;
 
-  private SpinnerValue mFormat = new SpinnerValue();
-  private SpinnerValue mBlockSize = new SpinnerValue();
-  private SpinnerValue mCompression = new SpinnerValue();
-  private SpinnerValue mCompressionLevel = new SpinnerValue();
+  private final SpinnerValue mFormat = new SpinnerValue();
+  private final SpinnerValue mBlockSize = new SpinnerValue();
+  private final SpinnerValue mCompression = new SpinnerValue();
+  private final SpinnerValue mCompressionLevel = new SpinnerValue();
 
   private GameFile gameFile;
 
@@ -395,6 +395,7 @@ public class ConvertFragment extends Fragment implements View.OnClickListener
   {
     if (requestCode == REQUEST_CODE_SAVE_FILE && resultCode == Activity.RESULT_OK)
     {
+      //noinspection ConstantConditions
       convert(data.getData().toString());
     }
   }
--- a/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/model/GameFile.java
+++ b/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/model/GameFile.java
@@ -4,6 +4,8 @@ import android.content.Context;
 
 import androidx.annotation.Keep;
 
+import java.io.File;
+
 public class GameFile
 {
   @Keep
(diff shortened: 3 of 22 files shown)


Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/model/GameFile.java
```diff
@@ -62,7 +66,16 @@ private GameFile(long pointer)
...
-    return context.getExternalCacheDir().getPath() + "/GameCovers/" + getGameTdbId() + ".png";
+    File externalCacheDir = context.getExternalCacheDir();
+
+    if (externalCacheDir != null)
+    {
+      return externalCacheDir.getPath() + "/GameCovers/" + getGameTdbId() + ".png";
+    }
+    else
+    {
+      throw new IllegalStateException("[GameFile] externalCacheDir is null");
```
JosJuice, line 77: If this is null, we can't expect Dolphin to function at all. I don't think adding an explicit check for it here adds much.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/utils/TvUtil.java
```diff
@@ -63,7 +66,16 @@ public static int getNumberOfChannels(Context context)
                             null,
                             null,
                             null);
-    return cursor != null ? cursor.getCount() : 0;
+    if (cursor != null)
+    {
+      channels = cursor.getCount();
+      cursor.close();
```
JosJuice, line 72: Please use try-with-resources instead.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/utils/TvUtil.java
```diff
@@ -175,8 +187,12 @@ else if ((cover = new File(game.getCoverPath(context))).exists())
     }
     catch (Exception e)
     {
+      String message = e.getMessage();
+
       Log.e(TAG, "Failed to create banner");
-      Log.e(TAG, e.getMessage());
+
+      if (message != null)
```
JosJuice, line 194: Can this actually happen?

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/fragments/ConvertFragment.java
```diff
@@ -395,7 +397,16 @@ public void onActivityResult(int requestCode, int resultCode, Intent data)
...
-      convert(data.getData().toString());
+      Uri uri = data.getData();
+
+      if (uri != null)
+      {
+        convert(uri.toString());
+      }
+      else
+      {
+        Log.error("[ConvertFragment] Could not get data uri.");
```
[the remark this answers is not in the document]
Ebola16: `getData` is `Nullable` so I assume the answer is yes.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/model/GameFile.java
```diff
@@ -2,9 +2,13 @@
...
 import android.content.Context;
 
+import java.io.File;
+
 public class GameFile
 {
-  private long mPointer;  // Do not rename or move without editing the native code
+  // Do not rename or move without editing the native code
+  @SuppressWarnings({"FieldCanBeLocal", "unused"})
+  private final long mPointer;
```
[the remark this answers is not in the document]
Ebola16: Ah, that's good information to know. I'll revert these changes if you want to take care of that.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/utils/TvUtil.java
Ebola16: Also `RecentlyNullable`

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/model/GameFile.java
Ebola16: Is it possible that another app modifies `externalCacheDir`, which is `Nullable`? I know there are some questionable cleanup apps out there... I don't think the overhead of an if statement here is too atrocious but correct me if I'm misunderstanding something.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/utils/TvUtil.java
Ebola16: I think I've done this properly now but worth a second look.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/activities/AppLinkActivity.java
```diff
@@ -35,13 +35,16 @@ protected void onCreate(Bundle savedInstanceState)
     Intent intent = getIntent();
     Uri uri = intent.getData();
 
-    Log.v(TAG, uri.toString());
-
-    if (uri.getPathSegments().isEmpty())
+    if (uri != null)
```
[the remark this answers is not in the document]
Ebola16: It would probably be less code if I just throw an exception upon failing this check.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/model/GameFile.java
JosJuice: > Is it possible that another app modifies `externalCacheDir`, which is `Nullable`? I know there are some questionable cleanup apps out there...

The case where this would be `null` is not when the directory has been deleted, it's when external storage is not mounted.

> I don't think the overhead of an if statement here is too atrocious but correct me if I'm misunderstanding something.

There's no problem performance-wise. I was thinking more about the maintenance side of things, if we should add null checks everywhere where Android Studio warns us even if the added checks don't actually solve anything.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/fragments/ConvertFragment.java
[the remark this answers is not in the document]
JosJuice: That `Nullable` attribute has to take into account every possible type of intent, but we have a very specific type of intent here. Is it really not possible to assume that when a file has been successfully selected, the URI is not null?

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/model/GameFile.java
Ebola16: > The case where this would be null is not when the directory has been deleted, it's when external storage is not mounted.

It is possible for another app to unmount external storage. I believe the SAF can also do this, as there's an "eject" symbol next to the SD card on my phone.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/fragments/ConvertFragment.java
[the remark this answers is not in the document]
Ebola16: After looking into this further, I no longer think this case can have a null URI so I just suppressed the warning instead. However, I think the other review comments mention code that can return null though.

Ebola16:
Is there anything holding this PR back? I think the remaining checks are useful for preventing unlikely exceptions.

JosJuice:
I don't recall why I didn't do another review round before, but by now this is very out of date. If you still want to do the cleanup, I would suggest putting it in a PR that's separate from dependency updates.

PR #9257, opened 2020-11-17
Remove a redundant i18n comment

JosJuice:


BhaaLseN (approved):
The comment removal seems ok, but I wonder: does that preceding `i18n` comment only apply to the next _translation_ that follows (which is the degree symbol) or to the next _line_ (which would be both of them)?

JosJuice:
Just the first one. I ran `update-source-strings.sh` to make sure.

leoetlino (approved):

PR #9259, opened 2020-11-18
Fixed non-UTF8 chars in PeakFinder.cpp

Strum355:
Encoding of this file was iso-8859-1 instead of utf-8, this PR addresses that to bring it inline with the .editorconfig file : ) discovered while working on indexing dolphin repo with https://github.com/sourcegraph/lsif-clang, so thanks for surfacing this! :grin:

### DIFF
--- a/Externals/soundtouch/PeakFinder.cpp
+++ b/Externals/soundtouch/PeakFinder.cpp
@@ -64,7 +64,7 @@ int PeakFinder::findTop(const float *data, int peakpos) const
 
     refvalue = data[peakpos];
 
-    // seek within ±10 points
+    // seek within Â±10 points
     start = peakpos - 10;
     if (start < minPos) start = minPos;
     end = peakpos + 10;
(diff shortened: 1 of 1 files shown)


BhaaLseN (approved):

JosJuice:
We normally avoid making changes in Externals. Has this problem been fixed upstream?

Strum355:
Ah did not know this was 3rd party code. It appears that it was https://gitlab.com/soundtouch/soundtouch/-/commit/3e74d1d18f35ab127c9702b7ccda5bfb093d9219

BhaaLseN:
Oops, I didn't even notice that :D

leoetlino:
> Ah did not know this was 3rd party code. It appears that it was https://gitlab.com/soundtouch/soundtouch/-/commit/3e74d1d18f35ab127c9702b7ccda5bfb093d9219

I guess merging this should be okay then as the change won't be reverted next time soundtouch is updated. The other option would be to update it right now but the last change to our Externals copy was in 2013 so let's keep that for another PR...

PR #9260, opened 2020-11-19
Common/Log: Add compile-time format string checks

leoetlino:
As this is a common source of bugs since we started migrating to {fmt}, I think it's worth enabling and adding more compile-time format string checks. This should also make the migration PRs easier to review ^^

As an added bonus, the static_assert will cause an invalid format call to be identified as an error in IDEs immediately, rather than at compilation time.

### DIFF
--- a/Source/Core/Common/FormatUtil.h
+++ b/Source/Core/Common/FormatUtil.h
@@ -1,0 +1,41 @@
+// Copyright 2020 Dolphin Emulator Project
+// Licensed under GPLv2+
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <cstddef>
+#include <string_view>
+
+namespace Common
+{
+constexpr std::size_t CountFmtReplacementFields(std::string_view s)
+{
+  std::size_t count = 0;
+  for (std::size_t i = 0; i < s.size(); ++i)
+  {
+    if (s[i] != '{')
+      continue;
+
+    // If the opening brace is followed by another brace, what we have is
+    // an escaped brace, not a replacement field.
+    if (i + 1 < s.size() && s[i + 1] == '{')
+    {
+      // Skip the second brace.
+      // This ensures that e.g. {{{}}} is counted correctly: when the first brace character
+      // is read and detected as being part of an '{{' escape sequence, the second character
+      // is skipped so the most inner brace (the third character) is not detected
+      // as the end of an '{{' pair.
+      ++i;
+      continue;
+    }
+
+    ++count;
+  }
+  return count;
+}
+
+static_assert(CountFmtReplacementFields("") == 0);
+static_assert(CountFmtReplacementFields("{} test {:x}") == 2);
+static_assert(CountFmtReplacementFields("{} {{}} test {{{}}}") == 2);
+}  // namespace Common
--- a/Source/Core/Common/Logging/Log.h
+++ b/Source/Core/Common/Logging/Log.h
@@ -76,13 +76,14 @@ enum LOG_LEVELS
 static const char LOG_LEVEL_TO_CHAR[7] = "-NEWID";
 
 void GenericLogFmtImpl(LOG_LEVELS level, LOG_TYPE type, const char* file, int line,
-                       std::string_view format, const fmt::format_args& args);
+                       fmt::string_view format, const fmt::format_args& args);
 
-template <typename... Args>
-void GenericLogFmt(LOG_LEVELS level, LOG_TYPE type, const char* file, int line,
-                   std::string_view format, const Args&... args)
+template <typename S, typename... Args>
+void GenericLogFmt(LOG_LEVELS level, LOG_TYPE type, const char* file, int line, const S& format,
+                   const Args&... args)
 {
-  GenericLogFmtImpl(level, type, file, line, format, fmt::make_format_args(args...));
+  GenericLogFmtImpl(level, type, file, line, format,
+                    fmt::make_args_checked<Args...>(format, args...));
 }
 
 void GenericLog(LOG_LEVELS level, LOG_TYPE type, const char* file, int line, const char* fmt, ...)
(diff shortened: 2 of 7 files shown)


Source/Core/Common/FormatUtil.h
```diff
@@ -0,0 +1,33 @@
...
+{
+  std::size_t count = 0;
+  for (std::size_t i = 0; i < s.size(); ++i)
+  {
+    if (s[i] != '{')
+      continue;
+
+    // If the opening brace is followed by another brace, what we have is
+    // an escaped brace, not a replacement field.
+    if (i + 1 < s.size() && s[i + 1] == '{')
```
BhaaLseN, line 22: I might be mixing up projects, but don't we have a string somewhere that attempts to output `{something}` and looks like `{{{}}}` in the string? The first two `{{` for a literal `{` and the third `{` as the inner `fmt` placeholder?

This might not work correctly with this validation.

leoetlino: I think you're right, I also seem to remember that we have such a string so I made it a test case when I was testing this on godbolt: https://godbolt.org/z/GnKvWz

It should work fine because when an escaped brace is detected, the second brace character is skipped, so the third brace character isn't detected as being part of an `{{` escape sequence for the second character. 

As you can see in the first and third examples, `{{{}}}` actually works fine

leoetlino: I added this explanation to the comment and some static_assert tests

JosJuice:
Would it be possible to also add this for `MsgAlertFmt` and `FmtFormatT`?

leoetlino:
For FmtFormatT, I think it's unfortunately possible without adding a macro because the format string wouldn't be a compile-time constant. Adding this to MsgAlertFmt should be easy though

Source/Core/Common/MsgHandler.h
```diff
@@ -88,33 +94,41 @@ std::string FmtFormatT(const char* string, Args&&... args)
...
+  [&] {                                                                                            \
+    /* Use a macro-like name to avoid shadowing warnings */                                        \
+    constexpr auto GENERIC_ALERT_FMT_N = Common::CountFmtReplacementFields(format);                \
+    return Common::MsgAlertFmt<GENERIC_ALERT_FMT_N>(yes_no, style, FMT_STRING(format),             \
+                                                    ##__VA_ARGS__);                                \
+  }()
+
 #define SuccessAlertFmt(format, ...)                                                               \
-  Common::MsgAlertFmt(false, Common::MsgType::Information, FMT_STRING(format), ##__VA_ARGS__)
+  GenericAlertFmt(false, Common::MsgType::Information, format, ##__VA_ARGS__);
```
lioncash, line 106:
```suggestion
  GenericAlertFmt(false, Common::MsgType::Information, format, ##__VA_ARGS__)
```

leoetlino: Whoops, good catch!

Source/Core/Common/Logging/Log.h
```diff
@@ -76,13 +78,17 @@ enum LOG_LEVELS
 static const char LOG_LEVEL_TO_CHAR[7] = "-NEWID";
 
 void GenericLogFmtImpl(LOG_LEVELS level, LOG_TYPE type, const char* file, int line,
-                       std::string_view format, const fmt::format_args& args);
+                       fmt::string_view format, const fmt::format_args& args);
```
BhaaLseN, line 81: Whats the difference between `std::string_view` and `fmt::string_view`? AFAICT `fmt::string_view` is from a time where `std::string_view` wasn't part of the standard yet, and using `std::string_view` should be fine.

Source/Core/Common/Logging/LogManager.cpp
```diff
@@ -81,7 +81,7 @@ void GenericLog(LOG_LEVELS level, LOG_TYPE type, const char* file, int line, con
 }
 
 void GenericLogFmtImpl(LOG_LEVELS level, LOG_TYPE type, const char* file, int line,
-                       std::string_view format, const fmt::format_args& args)
+                       fmt::string_view format, const fmt::format_args& args)
```
BhaaLseN, line 84: Same over here.

Source/Core/Common/Logging/Log.h
```diff
@@ -76,13 +78,17 @@ enum LOG_LEVELS
...
-void GenericLogFmt(LOG_LEVELS level, LOG_TYPE type, const char* file, int line,
-                   std::string_view format, const Args&... args)
+template <std::size_t NumFields, typename S, typename... Args>
+void GenericLogFmt(LOG_LEVELS level, LOG_TYPE type, const char* file, int line, const S& format,
+                   const Args&... args)
 {
-  GenericLogFmtImpl(level, type, file, line, format, fmt::make_format_args(args...));
+  static_assert(
+      NumFields == sizeof...(args),
+      "Unexpected number of replacement fields in format string; did you pass too many arguments?");
```
BhaaLseN, line 89: Should this message perhaps go both ways, like "[...] did you pass too few or too many arguments"?

Source/Core/Common/MsgHandler.h
```diff
@@ -37,9 +40,12 @@ bool MsgAlert(bool yes_no, MsgType style, const char* format, ...)
...
                      const fmt::format_args& args);
 
-template <typename... Args>
-bool MsgAlertFmt(bool yes_no, MsgType style, fmt::string_view format, const Args&... args)
+template <std::size_t NumFields, typename S, typename... Args>
+bool MsgAlertFmt(bool yes_no, MsgType style, const S& format, const Args&... args)
 {
+  static_assert(
+      NumFields == sizeof...(args),
+      "Unexpected number of replacement fields in format string; did you pass too many arguments?");
```
BhaaLseN, line 48: Same down here.

Source/Core/Common/Logging/Log.h
leoetlino: This doesn't work with FMT_STRING, which returns custom objects that only have a conversion operator for fmt::string_view.

leoetlino:
Apparently adding compile-time checks to FmtFormatT *without adding macros* will become possible with C++20, but for now we're stuck with C++17...

BhaaLseN (approved):
I can live with this for now, since it is an improvement overall.

PR #9266, opened 2020-11-20
Android: Fix IllegalStateException on startup

JosJuice:
See https://github.com/dolphin-emu/dolphin/pull/9095#pullrequestreview-535576465.

Marked as a draft because I haven't been able to build it yet. Latest master fails to compile on my Android Studio install.


JosJuice:
Tested and working.

leoetlino (approved):

PR #9268, opened 2020-11-20
EXI_DeviceMemoryCard: Medium cleanup

leoetlino:


Source/Core/Core/HW/EXI/EXI_DeviceMemoryCard.cpp
```diff
@@ -178,52 +178,52 @@ CEXIMemoryCard::GetGCIFolderPath(int card_index, AllowMovieFolder allow_movie_fo
...
   {
-    CurrentGameId = Common::swap32(reinterpret_cast<const u8*>(game_id.c_str()));
+    current_game_id = Common::swap32(reinterpret_cast<const u8*>(game_id.c_str()));
   }
 
   // TODO(C++20): Use structured bindings when we can use C++20 and refer to structured bindings
   // in lambda captures
   const auto folder_path = GetGCIFolderPath(card_index, AllowMovieFolder::Yes);
-  const auto& strDirectoryName = folder_path.first;
+  const auto& dir_name = folder_path.first;
```
JosJuice, line 191: This is actually a path, so I would prefer `dir_path`. (Maybe also `std::string` instead of `auto`, just to make things as clear as possible?)

leoetlino: Addressed

Source/Core/Core/HW/EXI/EXI_DeviceMemoryCard.cpp
```diff
@@ -276,7 +276,7 @@ void CEXIMemoryCard::TransferComplete()
 void CEXIMemoryCard::CmdDoneLater(u64 cycles)
 {
   CoreTiming::RemoveEvent(s_et_cmd_done[card_index]);
```
iwubcode, line 278: Maybe not worthwhile to do this here but I was thrown off by the fact this was actually a member variable.  Looks like none of the class members is prefixed with `m_`

leoetlino: Oh yeah, that should probably be fixed... I'll fix the enum in the header as well.

Source/Core/Core/HW/EXI/EXI_DeviceMemoryCard.cpp
```diff
@@ -509,56 +512,56 @@ void CEXIMemoryCard::DoState(PointerWrap& p)
...
   {
-    INFO_LOG_FMT(EXPANSIONINTERFACE, "writing to block: {:x}", address / Memcard::BLOCK_SIZE);
+    INFO_LOG_FMT(EXPANSIONINTERFACE, "writing to block: {:x}", m_address / Memcard::BLOCK_SIZE);
   }
 
   // Schedule transfer complete later based on write speed
-  CoreTiming::ScheduleEvent(_uSize * (SystemTimers::GetTicksPerSecond() / MC_TRANSFER_RATE_WRITE),
-                            s_et_transfer_complete[card_index], (u64)card_index);
+  CoreTiming::ScheduleEvent(size * (SystemTimers::GetTicksPerSecond() / MC_TRANSFER_RATE_WRITE),
+                            s_et_transfer_complete[m_card_index], (u64)m_card_index);
```
iwubcode, line 565: Was this c-style cast intentional?

Source/Core/Core/HW/EXI/EXI_DeviceMemoryCard.h
```diff
@@ -71,40 +71,40 @@ class CEXIMemoryCard : public IEXIDevice
...
-  int command;
-  int status;
-  u32 m_uPosition;
-  u8 programming_buffer[128];
+  int m_interrupt_switch;
+  bool m_interrupt_set;
+  Command m_command;
+  int m_status;
+  u32 m_position;
+  u8 m_programming_buffer[128];
```
iwubcode, line 102: Any benefit to having this be a `std::array` instead?

Source/Core/Core/HW/EXI/EXI_DeviceMemoryCard.cpp
```diff
@@ -106,22 +106,22 @@ void CEXIMemoryCard::Shutdown()
...
-  command = 0;
-  status = MC_STATUS_BUSY | MC_STATUS_UNLOCKED | MC_STATUS_READY;
-  m_uPosition = 0;
-  memset(programming_buffer, 0, sizeof(programming_buffer));
+  m_interrupt_switch = 0;
+  m_interrupt_set = false;
+  m_command = Command::NintendoID;
+  m_status = MC_STATUS_BUSY | MC_STATUS_UNLOCKED | MC_STATUS_READY;
+  m_position = 0;
+  memset(m_programming_buffer, 0, sizeof(m_programming_buffer));
```
iwubcode, line 124: `std::file(m_programming_buffer.begin(), m_programming_buffer.end(), 0);`

leoetlino: or even m_programming_buffer.fill(0)

Source/Core/Core/HW/EXI/EXI_DeviceMemoryCard.cpp
```diff
@@ -260,231 +260,234 @@ bool CEXIMemoryCard::IsPresent() const
...
+    case Command::ReadID:
+    case Command::ReadErrorBuffer:
+    case Command::WakeUp:
+    case Command::Sleep:
+    case Command::ClearStatus:
+    case Command::SectorErase:
+    case Command::PageProgram:
+    case Command::ExtraByteProgram:
+    case Command::ChipErase:
       DEBUG_LOG_FMT(EXPANSIONINTERFACE, "EXI MEMCARD: command {:02x} at position 0. seems normal.",
```
leoetlino, line 371: Note: recent versions of {fmt} format enum classes as integers so a cast is not required.

iwubcode (approved):

PR #9269, opened 2020-11-20
Common: Assert that translatable strings use positional arguments

JosJuice:
Like PR #9260, but for a different requirement (see PR #9253).


leoetlino:
Could the commit order be swapped, so that "Fix translatable strings which contain non-positional arguments" is the first commit, to ensure each commit can be built?

JosJuice:
Done.

leoetlino (approved):

BhaaLseN (approved):

PR #9272, opened 2020-11-21
Android: Expand WiimoteProfileSetting to more setting types

Ebola16:
This PR would be used for exposing `wiimoteOrientaiton` settings in the UI, a feature requested here https://bugs.dolphin-emu.org/issues/11371. An alternative (and probably worse) solution is already present in #8894, see `AbstractBooleanSetting wiimoteOrientation`.

- ~Would saving settings immediately after setting them in `WiimoteProfileSetting` help prevent the warning comment? I'll need to access wii profile settings other than strings, like the boolean orientation setting in my Wii Overlay PR.~
- What do you think about renaming `WiimoteProfileSetting` to `WiimoteProfileStringSetting` and then creating other `WiimoteProfile[SettingType]Setting` files?

Just checking before I proceed. Currently untested.


JosJuice:
> Would saving settings immediately after setting them in `WiimoteProfileSetting` help prevent the warning comment? I'll need to access wii profile settings other than strings, like the boolean orientation setting in my Wii Overlay PR.

Actually, the profile settings are already being saved to disk immediately after setting them. The problem is that the other `WiimoteProfileSetting`s are not reloading the profile from disk afterwards.

If we want to use this system for more than one setting (which I'm not fond of), then I think the best solution would be to keep the in-memory copy of the profile in `Settings` instead of `WiimoteProfileSetting` so that there only is one in-memory copy of each profile.

> What do you think about renaming `WiimoteProfileSetting` to `WiimoteProfileStringSetting` and then creating other `WiimoteProfile[SettingType]Setting` files?

If we want to use this system for more than one setting, then I think that makes perfect sense.

Ebola16:
Using the system for more than one setting type sounds like an acceptable option until Wiimote Profile settings are converted to the new config system.

I'll look into moving the in-memory copy of the profile to `Settings` and working on the new files within the next day or so.

JosJuice:
Just a minor remark: For controller settings, what I'm planning to do is to get rid of the indirection we currently have on Android, not to start using the new layered config system. The C++ code isn't using the layered config system for controller settings. But your point is valid anyway.

Ebola16:
@JosJuice 
- Are the added static variables acceptable in `Settings`? `getWiimoteProfile` needs a `padID` and the INI file needs to be reused in other places.
- I think I should only implement the setting types I need for now so there aren't unused files.

JosJuice:
Static should be avoided in this case, and I don't see why it would be necessary to use. You already have access to a `Settings` instance in all `WiimoteProfileSetting` methods where you're accessing your new static methods in `Settings`.

Also, please keep in mind that if the user decides to edit the settings for different Wii Remotes (different `padID`s) in one session, the profile INI for each `padID` must be kept track of separately in `Settings`. You can't have the profile of one `padID` overwrite another.

Ebola16:
Oops, this is still wrong.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/model/Settings.java
```diff
@@ -59,6 +61,15 @@
...
   private HashMap<String, IniFile> mIniFiles = new HashMap<>();
 
+  private File mWiimoteProfileFile0;
+  private File mWiimoteProfileFile1;
+  private File mWiimoteProfileFile2;
+  private File mWiimoteProfileFile3;
+  private IniFile mWiimoteProfileIni0 = new IniFile();
+  private IniFile mWiimoteProfileIni1 = new IniFile();
+  private IniFile mWiimoteProfileIni2 = new IniFile();
+  private IniFile mWiimoteProfileIni3 = new IniFile();
```
JosJuice, line 71: You should use arrays (or some other kind of collection, like a HashMap) for these so you don't need to duplicate code.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/model/Settings.java
```diff
@@ -87,6 +98,70 @@ public boolean isGameSpecific()
     return !TextUtils.isEmpty(mGameId);
   }
 
+  public IniFile getWiimoteProfile(String profile, int padID)
+  {
+    File wiimoteProfileFile = SettingsFile.getWiiProfile(profile);
+    IniFile wiimoteProfileIni = new IniFile();
+
+    if (!wiimoteProfileIni.load(wiimoteProfileFile, false))
```
JosJuice, line 106: This should only load from disk if we don't already have the INI in memory.

Ebola16:
I overthought `getWiimoteProfile` and I think a previous attempt was the best choice. Fixed (hopefully).

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/model/Settings.java
```diff
@@ -87,6 +98,70 @@ public boolean isGameSpecific()
...
+    return getWiimoteControlsSection(settings).delete(profileKey);
+  }
+
+  public boolean isWiimoteProfileEnabled(Settings settings, String profile,
+          String profileKey)
+  {
+    return profile.equals(getWiimoteControlsSection(settings).getString(profileKey, ""));
+  }
+
+  public IniFile.Section getWiimoteControlsSection(Settings s)
```
JosJuice, line 160: This is a helper function that can be private.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/model/Settings.java
```diff
@@ -87,6 +98,70 @@ public boolean isGameSpecific()
...
+  {
+    return s.getSection(GAME_SETTINGS_PLACEHOLDER_FILE_NAME, Settings.SECTION_CONTROLS);
```
JosJuice, line 162: If you're going to drop the `Settings` prefix from `GAME_SETTINGS_PLACEHOLDER_FILE_NAME`, you should drop it from `SECTION_CONTROLS` too. (And you could also rename `s` to `settings`, because I only did that to make this fit on one line,

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/model/Settings.java
```diff
@@ -184,6 +259,31 @@ public void saveSettings(SettingsActivityView view, Context context)
 
       NativeConfig.save(NativeConfig.LAYER_LOCAL_GAME);
     }
+
+    saveWiimoteProifleSettings();
+  }
+
+  public void saveWiimoteProifleSettings()
```
JosJuice, line 266:
```suggestion
  public void saveWiimoteProfileSettings()
```

Ebola16: Yeah, I might be a little too sleep deprived right now. I'll revisit this later today.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/model/WiimoteProfileStringSetting.java
```diff
@@ -0,0 +1,67 @@
...
+    else
+      return mDefaultValue;
+  }
+
+  @Override
+  public void setString(Settings settings, String newValue)
+  {
+    settings.getWiimoteProfile(mProfile, mPadID).setString(mSection, mKey, newValue);
+
+    settings.saveWiimoteProifleSettings();
```
JosJuice, line 64: There isn't really a need to call this here, now that we're calling it in `Settings.saveSettings`. Calling it in `Settings.saveSettings` makes the most sense to me since it better matches how normal settings work.

Ebola16: I thought the same thing too but removing that line of code prevents the extension setting from updating when backing out of an extension submenu. At least I'm pretty sure that was the problem, I'll verify when I cleanup this PR.

JosJuice: Ah, right, that's because `getWiimoteProfile` is reloading from disk every time it's called. If you change that, you can remove this.

JosJuice (commented):
While I don't really like the idea of adding more profile settings, if we are going to do it, then the way this PR is doing it now is the way I would want it to be done. So I think I'll do it like this: I'm not going to be the one who will merge this PR, but if someone else thinks adding more profile settings is worth it, then I don't have anything against someone else merging it. But I have one request: The commit that adds the second profile setting should be in this PR.

Ebola16:
To the 2nd reviewer: This PR would be used for exposing `wiimoteOrientaiton` in the UI, a feature requested here https://bugs.dolphin-emu.org/issues/11371. An alternative (and probably worse) solution is already present in #8894, see `AbstractBooleanSetting wiimoteOrientation`.

@JosJuice I'd rather not decouple orientation from #8894 so I'd like to leave that PR and this PR as-is until this is merged or rejected.

I already have most of the #8894 rewrite completed. The changes from this PR will modify `AbstractBooleanSetting wiimoteOrientation` in the following way:
```
        AbstractBooleanSetting wiimoteOrientation;
        if (mGameID.isEmpty())
        {
          wiimoteOrientation = new LegacyBooleanSetting(Settings.FILE_WIIMOTE,
                  Settings.SECTION_WIIMOTE + (wiimoteNumber - 3), SettingsFile.KEY_WIIMOTE_EXTENSION,
                  false);
        }
        else
        {
          wiimoteOrientation =
                  new WiimoteProfileBooleanSetting(mGameID, wiimoteNumber - 4, Settings.SECTION_PROFILE,
                          SettingsFile.KEY_WIIMOTE_EXTENSION, false);
        }
```

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/model/Settings.java
```diff
@@ -58,6 +60,7 @@
           FILE_WIIMOTE};
 
   private HashMap<String, IniFile> mIniFiles = new HashMap<>();
+  private final HashMap<File, IniFile> mWiimoteProfileFiles = new HashMap<>();
```
Pokechu22, line 63: IMO it'd be better for this to be a `Map<String, IniFile>`.  It seems a bit odd to use the `File` as a key here (since only loading and saving care about the `File`, and it already assumes that `SettingsFile.getWiiProfile` returns the same `File` for the same input).  Also generally it's better to use the highest interface for the field instead of the implementation (`mIniFiles` doesn't do this, even though `java.util.Map` is already imported (and unused...)).

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/model/Settings.java
```diff
@@ -87,6 +90,55 @@ public boolean isGameSpecific()
...
+
+      wiimoteProfileIni.load(defaultWiiProfilePath, false);
+
+      wiimoteProfileIni
+              .setString(SECTION_PROFILE, "Device", "Android/" + (padID + 4) + "/Touchscreen");
+
+      mWiimoteProfileFiles.put(wiimoteProfileFile, wiimoteProfileIni);
+    }
+
+    return mWiimoteProfileFiles.get(wiimoteProfileFile);
```
Pokechu22, line 118: With the above change, this could be something like this:

```java
if (!mWiimoteProfileFiles.containsKey(profile))
{
  IniFile newIni = new IniFile();
  newIni.load(SettingsFile.getWiiProfile(profile), false);
  mWiimoteProfileFiles.put(profile, wiimoteProfileIni);
}

IniFile wiimoteProfileIni = mWiimoteProfileFiles.get(profile);

if (!wiimoteProfileIni.exists(SECTION_PROFILE))
{
  String defaultWiiProfilePath = DirectoryInitialization.getUserDirectory() +
          "/Config/Profiles/Wiimote/WiimoteProfile.ini";

  wiimoteProfileIni.load(defaultWiiProfilePath, false);

  wiimoteProfileIni
          .setString(SECTION_PROFILE, "Device", "Android/" + (padID + 4) + "/Touchscreen");
}

return wiimoteProfileIni;
```

or (if Java 8 is available):

```java
IniFile wiimoteProfileIni = mWiimoteProfileFiles.computeIfAbsent(profile, profile -> {
  IniFile newIni = new IniFile();
  newIni.load(SettingsFile.getWiiProfile(profile), false);
  return newIni;
});

if (!wiimoteProfileIni.exists(SECTION_PROFILE))
{
  String defaultWiiProfilePath = DirectoryInitialization.getUserDirectory() +
          "/Config/Profiles/Wiimote/WiimoteProfile.ini";

  wiimoteProfileIni.load(defaultWiiProfilePath, false);

  wiimoteProfileIni
          .setString(SECTION_PROFILE, "Device", "Android/" + (padID + 4) + "/Touchscreen");
}

return wiimoteProfileIni;
```

Note that with this set-up, `wiimoteProfileIni` is mutated in place if `SECTION_PROFILE` does not exist, so it doesn't need to be put into the map again (we're mutating the one already in the map).

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/model/Settings.java
```diff
@@ -184,6 +236,11 @@ public void saveSettings(SettingsActivityView view, Context context)
 
       NativeConfig.save(NativeConfig.LAYER_LOCAL_GAME);
     }
+
+    for (Map.Entry<File, IniFile> entry : mWiimoteProfileFiles.entrySet())
+    {
+      entry.getValue().save(entry.getKey());
```
Pokechu22, line 242: This would become `entry.getValue().save(SettingsFile.getWiiProfile(entry.keyKey()))`.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/ui/SettingsFragmentPresenter.java
```diff
@@ -775,8 +775,8 @@ private void addWiimoteSubSettings(ArrayList<SettingsItem> sl, int wiimoteNumber
     }
     else
     {
-      extension = new WiimoteProfileSetting(mGameID, wiimoteNumber - 4, Settings.SECTION_PROFILE,
-              SettingsFile.KEY_WIIMOTE_EXTENSION, defaultExtension);
+      extension = new WiimoteProfileStringSetting(mGameID, wiimoteNumber - 4,
```
Pokechu22, line 778: What's the deal with `wiimoteNumber - 4`?  `getWiimoteProfile` just adds 4 to it again, so why the subtraction?  (I do see `mProfileKey = SettingsFile.KEY_WIIMOTE_PROFILE + (padID + 1)` in the settings, but I'm still not sure why it's like that)

Ebola16: It's due to different mappings from the native code.

https://github.com/dolphin-emu/dolphin/blob/5abae61a8c7e0e4ac03bb8d3031793037e87fdd3/Source/Android/app/src/main/res/values/strings.xml#L25-L27

It could probably be improved but that's outside of the scope of this PR.

Pokechu22: Ah, I also missed the comment a few lines up:

https://github.com/dolphin-emu/dolphin/blob/de353551475808a7c403ca97df196331c1c628cb/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/ui/SettingsFragmentPresenter.java#L764-L767

The naming seems counterintuitive to me (it'd make more sense if what's in `WiimoteProfile*Setting` was called `wiimoteNumber` and what's in `addWiimoteSubSettings` was called `padID` since then `padID` would be 4-7 for wiimotes and the wiimote number would be 0-3), but it's probably something that would be handled by a larger rework outside of this PR.

Ebola16:
Other than some variable renaming to avoid conflicts, I've used @Pokechu22 's suggestions. Do you agree with https://github.com/dolphin-emu/dolphin/pull/9272#issuecomment-733130255 in that this PR is a better solution than `AbstractBooleanSetting wiimoteOrientation` in #8894?

Pokechu22:
It seems reasonable to me, though I haven't looked over what #8894 is doing in detail.

Ebola16:
Then is there anything left of concern in this PR @JosJuice? Keep in mind that the orientation setting is loaded after choosing the "None" Wiimote extension so there would already be an INI file in memory anyway.

JosJuice:
I still think that the second profile setting should be added in this PR, so that we don't end up adding code that we aren't going to be using. But I stand by what I said before: I'm not going to merge this myself, but if someone else wants to merge it (either in its current state or after adding the second profile setting), then I don't mind.

Ebola16:
Another alternative would be moving this commit to the Wii Overlay PR. Let me know if you'd prefer that.

JosJuice:
While I don't have anything against it, please keep in mind that this will make it less likely to get merged any time soon, with how many things the Wii overlay PR already is doing.

Ebola16:
Alternative alternative idea, would this be approved by dropping `WiimoteProfileStringSetting` from this PR and after this gets merged, add just that file to the Wii Overlay PR?

JosJuice:
That doesn't make any meaningful difference to the current state of the PR in my opinion.

Ebola16:
Alright, since @Pokechu22 can't merge either, I guess I should request a review from @lioncash? Feel free to pass this on to someone else if needed.

JMC47:
I think the general consensus on this is that the system isn't all that great.  This improves the system as it currently is, and doesn't make it harder to rip out and replace with something more robust later.  As such, I'm okay with merging it for usability's sake.

PR #9275, opened 2020-11-22
FrameDump: Start timing at 0 ticks when starting from boot

JosJuice:
@vadosnaprimer Does this fix your "frame 0" A/V sync problem?

### DIFF
--- a/Source/Core/VideoCommon/FrameDump.cpp
+++ b/Source/Core/VideoCommon/FrameDump.cpp
@@ -47,7 +47,7 @@ struct FrameDumpContext
   int width = 0;
   int height = 0;
 
-  u64 first_frame_ticks = 0;
+  u64 start_ticks = 0;
   u32 savestate_index = 0;
 
   bool gave_vfr_warning = false;
@@ -111,7 +111,7 @@ std::string GetDumpPath(const std::string& extension, std::time_t time, u32 inde
 
 }  // namespace
 
-bool FrameDump::Start(int w, int h)
+bool FrameDump::Start(int w, int h, u64 start_ticks)
 {
   if (IsStarted())
     return true;
@@ -120,15 +120,18 @@ bool FrameDump::Start(int w, int h)
   m_start_time = std::time(nullptr);
   m_file_index = 0;
 
-  return PrepareEncoding(w, h);
+  return PrepareEncoding(w, h, start_ticks, m_savestate_index);
 }
 
-bool FrameDump::PrepareEncoding(int w, int h)
+bool FrameDump::PrepareEncoding(int w, int h, u64 start_ticks, u32 savestate_index)
 {
   m_context = std::make_unique<FrameDumpContext>();
 
   m_context->width = w;
   m_context->height = h;
+
+  m_context->start_ticks = start_ticks;
+  m_context->savestate_index = savestate_index;
 
   InitAVCodec();
   const bool success = CreateVideoFile();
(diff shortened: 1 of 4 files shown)


Source/Core/VideoCommon/FrameDump.cpp
```diff
@@ -447,14 +444,15 @@ void FrameDump::CheckForConfigChange(const FrameData& frame)
   {
     Stop();
     ++m_file_index;
-    PrepareEncoding(frame.width, frame.height);
+    PrepareEncoding(frame.width, frame.height, frame.state.ticks, frame.state.savestate_index);
```
JosJuice, line 447: I'm using `frame.state.ticks` here to match master's behavior, but master's behavior seems fishy to me. Doesn't this discard the information about how much time passed between the first frame of this file and the last frame of the previous file? And if so, would the correct thing to use here be the time of the previous frame, or the time of the previous frame plus one "timebase unit"...?

vadosnaprimer:
Can't comment on the code but this patch helped!

Source/Core/VideoCommon/FrameDump.cpp
```diff
@@ -300,7 +297,7 @@ void FrameDump::AddFrame(const FrameData& frame)
     else if (pts > m_context->last_pts + 1 && !m_context->gave_vfr_warning)
     {
       WARN_LOG_FMT(FRAMEDUMP, "PTS delta > 1. Resulting file will have variable frame rate. "
-                              "Subsequent occurances will not be reported.");
+                              "Subsequent occurences will not be reported.");
```
leoetlino, line 300:
```suggestion
                              "Subsequent occurrences will not be reported.");
```

Source/Core/VideoCommon/FrameDump.cpp
JosJuice: After looking into the ffmpeg API some more, I think the right thing to do would be to explicitly set the duration of the old stream and to use `frame.state.ticks` here. I will leave setting the duration for another PR.

leoetlino (approved):
The rest of the PR LGTM.

Source/Core/VideoCommon/FrameDump.cpp
JosJuice: After looking into this some more, I'm not sure if there's any solution that works great... other than implementing writing audio data to the AVI file, which is a desirable feature regardless of this problem. I think we should move forward with this PR as it is.

Pokechu22:
This PR seems to have [changed FifoCI](https://fifo.ci/version/27712c63218ce4bc225298ce4e97be591147950d/), apparently just due to the introduction of one additional frame.  Is that to be expected?

JosJuice:
No, that's not expected. I'll have to look into this more closely.

JosJuice:
My latest commit fixes the problem with FifoCI reporting duplicate frames, but... I'm not sure if my change makes sense, and the rendering results are different in a way that makes me unsure whether the new behavior or the old behavior is right. If the old behavior is right, I suppose my latest commit is writing to the FIFO too quickly?

JosJuice:
These ones looks a lot better after the change: https://fifo.ci/compare/5726986-5724632/ https://fifo.ci/compare/5727030-5724655/ https://fifo.ci/compare/5727032-5724676/
But this one is just white: https://fifo.ci/compare/5726978-5724709/
This one is missing the shadow: https://fifo.ci/compare/5727093-5724838/
And these ones still have duplicated frames: https://fifo.ci/compare/5727075-5724683/ https://fifo.ci/compare/5727035-5724706/

JosJuice:
I've now removed that commit and placed it in PR #9287 instead.

What I'm thinking is that it would make sense to just set the FifoCI runners to use PNG dumping instead of ffmpeg dumping so that they will completely ignore frame timings. Would this be a reasonable solution? (Of course, we should do it in the config files for the FifoCI runners instead of by changing the default value of `GFX_DUMP_FRAMES_AS_IMAGES` like I've done now for testing.)

JosJuice:
Okay, it seems like FifoCI wasn't set up to handle PNG dumps like I thought...

JosJuice:
@dolphin-emu-bot rebuild

Pokechu22:
@dolphin-emu-bot rebuild

JosJuice:
Now that https://github.com/dolphin-emu/fifoci/pull/34 (and the two PRs before it) is merged, FifoCI is ignoring frame timings and this PR is not causing FifoCI diffs anymore.

PR #9276, opened 2020-11-23
Core: Convert logging over to fmt pt.4

lioncash:
Continues the migration of logging in the Core library. This part finishes up the remaining log calls within the IOS code.

### DIFF
--- a/Source/Core/Core/IOS/Network/IP/Top.cpp
+++ b/Source/Core/Core/IOS/Network/IP/Top.cpp
@@ -61,8 +61,8 @@ enum SOResultCode : s32
 NetIPTop::NetIPTop(Kernel& ios, const std::string& device_name) : Device(ios, device_name)
 {
 #ifdef _WIN32
-  int ret = WSAStartup(MAKEWORD(2, 2), &InitData);
-  INFO_LOG(IOS_NET, "WSAStartup: %d", ret);
+  const int ret = WSAStartup(MAKEWORD(2, 2), &InitData);
+  INFO_LOG_FMT(IOS_NET, "WSAStartup: {}", ret);
 #endif
 }
 
@@ -132,7 +132,7 @@ static s32 MapWiiSockOptLevelToNative(u32 level)
   if (level == 0xFFFF)
     return SOL_SOCKET;
 
-  INFO_LOG(IOS_NET, "SO_SETSOCKOPT: unknown level %u", level);
+  INFO_LOG_FMT(IOS_NET, "SO_SETSOCKOPT: unknown level {}", level);
   return level;
 }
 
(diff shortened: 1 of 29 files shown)


Source/Core/Core/IOS/Network/IP/Top.cpp
```diff
@@ -554,48 +556,52 @@ IPCCommandResult NetIPTop::HandleGetPeerNameRequest(const IOCtlRequest& request)
...
     return GetDefaultReply(0);
   }
 
-  Memory::Write_U32(Common::swap32(*(u32*)remoteHost->h_addr_list[0]), request.buffer_out);
-  INFO_LOG(IOS_NET,
-           "IOCTL_SO_INETATON = 0 "
-           "%s, BufferIn: (%08x, %i), BufferOut: (%08x, %i), IP Found: %08X",
-           hostname.c_str(), request.buffer_in, request.buffer_in_size, request.buffer_out,
-           request.buffer_out_size, Common::swap32(*(u32*)remoteHost->h_addr_list[0]));
+  const auto ip = Common::swap32(*(u32*)remoteHost->h_addr_list[0]);
```
BhaaLseN, line 589: You changed other places to C++ style casts, but not this one - was that on purpose or did you simply miss it?

BhaaLseN (approved):
LGTM otherwise (and that remark isn't even what this PR is about)

Source/Core/Core/IOS/Network/IP/Top.cpp
lioncash: Changed it over. Also we can use the swap32 variant that takes a `u8*` to avoid invoking undefined behavior wrt pointer casting and dereferencing.

Source/Core/Core/IOS/Network/Socket.cpp
```diff
@@ -100,12 +100,12 @@ static s32 TranslateErrorCode(s32 native_error, bool is_rw)
 }
 
 // Don't use string! (see https://github.com/dolphin-emu/dolphin/pull/3143)
-s32 WiiSockMan::GetNetErrorCode(s32 ret, const char* caller, bool isRW)
+s32 WiiSockMan::GetNetErrorCode(s32 ret, std::string_view caller, bool is_rw)
```
sepalani, line 103: The `caller` type shouldn't be changed as suggested by the above comment.

lioncash: this was for std::string. std::string_view is just a wrapper around a pointer and size variable.

sepalani: I couldn't find anything suspicious in Windows' `string_view` constructor beside debug macros, so it should be fine as you said.

leoetlino (approved):

PR #9278, opened 2020-11-23
Core: Convert missed log calls over to fmt

lioncash:
Converts over some lingering log calls that I missed in previously converted files within Core.


BhaaLseN (approved):
I usually only check the diffs (for changes), but don't check the files themselves (for missed ones) D:

But we'll get there eventually, once we try to get rid of the old methods/macros we'll see the places we missed :D

leoetlino (approved):

PR #9280, opened 2020-11-24
Fix bounding box incorrectly disabled on OpenGL ES 3.1, 3.2

blaahaj:
Fixes bounding box games like Paper Mario on OpenGL ES Android phones.


degasus (approved):
I think this fixes a regression of f0391491986. Before it, bSupportsBBox was always true, and the implementation just switched between the SSBO and stencil approach. Afterwards, this bit was wrongly set for GLES.

faroukabd000:
Something happened to the emulator in android  since this update
Amm a black screen when launching a game
And when clickin' the exit emulator button the emulator freeze
(I tried legend of zelda twilight princess) 
Xiomi Mi a3
Sd 665 / adreno 605/ 3 ram

JosJuice:
Are you sure it's since this update and not since 5.0-13127? If it is since this update, please check whether it happens with both OpenGL and Vulkan. If it's since 5.0-13127, test 5.0-13148 or later.

faroukabd000:
Since 5.0-13127
After that i tried 5.0-13163 bith in vulkan and opl and the problem is gone

PR #9281, opened 2020-11-25
IOFile: Replace fprintf with WriteString/fmt.

AdmiralCurtiss:
This actually started as an attempt to remove the `GetHandle()` method, but I figured this makes as a decent PR on its own.

### DIFF
--- a/Source/Core/Common/File.h
+++ b/Source/Core/Common/File.h
@@ -7,6 +7,7 @@
 #include <cstddef>
 #include <cstdio>
 #include <string>
+#include <string_view>
 
 #include "Common/CommonTypes.h"
 
@@ -66,6 +67,8 @@ public:
   {
     return WriteArray(reinterpret_cast<const char*>(data), length);
   }
+
+  bool WriteString(std::string_view str) { return WriteBytes(str.data(), str.size()); }
 
   bool IsOpen() const { return nullptr != m_file; }
   // m_good is set to false when a read, write or other function fails
--- a/Source/Core/Core/PowerPC/Jit64/Jit.cpp
+++ b/Source/Core/Core/PowerPC/Jit64/Jit.cpp
@@ -473,7 +473,7 @@ static void ImHere()
     if (!f)
       f.Open("log64.txt", "w");
 
-    fprintf(f.GetHandle(), "%08x\n", PC);
+    f.WriteString(fmt::format("{0:08x}\n", PC));
   }
   if (been_here.find(PC) != been_here.end())
   {
--- a/Source/Core/Core/PowerPC/JitInterface.cpp
+++ b/Source/Core/Core/PowerPC/JitInterface.cpp
@@ -16,6 +16,8 @@
 #include "Common/PerformanceCounter.h"
 #endif
 
+#include <fmt/format.h>
+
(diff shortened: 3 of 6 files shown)


Source/Core/Common/File.h
```diff
@@ -67,6 +68,10 @@ class IOFile
     return WriteArray(reinterpret_cast<const char*>(data), length);
   }
 
+  bool WriteString(const char* str) { return WriteString(std::string_view(str)); }
+  bool WriteString(const std::string_view& str) { return WriteBytes(str.data(), str.size()); }
```
lioncash, line 72:
```suggestion
  bool WriteString(std::string_view str) { return WriteBytes(str.data(), str.size()); }
```

This overload alone should be sufficient to make the const char* and std::string overloads unnecessary as well.

AdmiralCurtiss: I was under the impression that that had weird edge cases where the string could be deallocated before the string_view is used, but trying to produce such a case was fruitless, so yeah I guess that works!

JosJuice: That is only a problem if we keep the string_view around after the function finishes executing. (For instance, if we return it from the function.)

Source/Core/Core/PowerPC/Jit64/Jit.cpp
```diff
@@ -473,7 +473,7 @@ static void ImHere()
     if (!f)
       f.Open("log64.txt", "w");
 
-    fprintf(f.GetHandle(), "%08x\n", PC);
+    f.WriteString(fmt::format("{0:08x}", PC));
```
BhaaLseN, line 476: Did you intentionally get rid of the trailing `\n`?

Source/Core/Core/PowerPC/JitInterface.cpp
```diff
@@ -103,17 +105,17 @@ void WriteProfileResults(const std::string& filename)
...
   {
     std::string name = g_symbolDB.GetDescription(stat.addr);
     double percent = 100.0 * (double)stat.cost / (double)prof_stats.cost_sum;
     double timePercent = 100.0 * (double)stat.tick_counter / (double)prof_stats.timecost_sum;
-    fprintf(f.GetHandle(),
-            "%08x\t%s\t%" PRIu64 "\t%" PRIu64 "\t%" PRIu64 "\t%.2f\t%.2f\t%.2f\t%i\n", stat.addr,
-            name.c_str(), stat.run_count, stat.cost, stat.tick_counter, percent, timePercent,
-            (double)stat.tick_counter * 1000.0 / (double)prof_stats.countsPerSec, stat.block_size);
+    f.WriteString(fmt::format(
+        "{0:08x}\t{1}\t{2}\t{3}\t{4}\t{5:.2f}\t{6:.2f}\t{7:.2f}\t{8}\n", stat.addr, name.c_str(),
```
BhaaLseN, line 116: You don't need the `.c_str()` here anymore.

Source/Core/Core/PowerPC/Jit64/Jit.cpp
AdmiralCurtiss: Nope, thanks!

leoetlino (approved):

PR #9282, opened 2020-11-25
Core: Convert logging over to fmt pt.5

lioncash:
Converts the remaining PowerPC code over to fmt-capable logging.

Now, all that's left to convert over are the lingering remnants within the frontend code.

### DIFF
--- a/Source/Core/Core/PowerPC/Interpreter/Interpreter.cpp
+++ b/Source/Core/Core/PowerPC/Interpreter/Interpreter.cpp
@@ -132,11 +132,11 @@ static void Trace(UGeckoInstruction& inst)
   }
 
   const std::string ppc_inst = Common::GekkoDisassembler::Disassemble(inst.hex, PC);
-  DEBUG_LOG(POWERPC,
-            "INTER PC: %08x SRR0: %08x SRR1: %08x CRval: %016" PRIx64 " FPSCR: %08x MSR: %08x LR: "
-            "%08x %s %08x %s",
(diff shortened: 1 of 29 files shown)


Source/Core/Core/PowerPC/Interpreter/Interpreter.cpp
```diff
@@ -285,25 +285,25 @@ void Interpreter::Run()
...
               if (j > 0)
               {
-                if (PCVec.at(j) != PCVec.at(j - 1) + 4)
-                  NOTICE_LOG(POWERPC, "");
+                if (PCVec[j] != PCVec[(j - 1) + 4]
+                  NOTICE_LOG_FMT(POWERPC, "");
               }
 
-              NOTICE_LOG(POWERPC, "PC: 0x%08x", PCVec.at(j));
+              NOTICE_LOG_FMT(POWERPC, "PC: {:#010x]", PCVec[j]);
```
BhaaLseN, line 303: Shouldn't the compile time checking hit this one?
```suggestion
              NOTICE_LOG_FMT(POWERPC, "PC: {:#010x}", PCVec[j]);
```

lioncash: `SHOW_HISTORY` isn't defined on the buildbot, so this code isn't checked.

Arguably, this code should be removed entirely since I don't think it's been used for ages, but that'd be something for a different PR.

leoetlino (approved):

PR #9283, opened 2020-11-25
Common: Optimize Config::Get

JosJuice:
The way `Config::Get` works in master, it first calls `Config::GetActiveLayerForConfig` which searches for the setting in all layers, and then calls `Config::Layer::Get` which searches for the same setting again within the given layer. We can remove this second search by combining the logic of `Config::GetActiveLayerForConfig` and `Config::Layer::Get` into one function.


leoetlino (approved):
The logic looks sane.

PR #9285, opened 2020-11-26
Update bundled libpng to version 1.6.37 (from 1.6.29)

AdmiralCurtiss:
Apparently fixes a use-after-free bug in `png_image_free()`, see http://www.libpng.org/pub/png/libpng.html.

(though really I just noticed that a new version existed when I was looking up the documentation and figured, might as well update)


AdmiralCurtiss:
FYI, I put in the architecture-specific hardware optimization files this time but left them turned off in cmake. Might be interesting to see if it's worthwhile turning those on.

PR #9286, opened 2020-11-26
Frontends: Migrate logs over to fmt

lioncash:
Migrates the final remaining logs over to fmt-equivalents

After this PR, renaming of the new fmt functions to the old one can be undertaken. The printf logging can't be removed until all alerts and assertions are converted over however.

### DIFF
--- a/Source/Core/VideoCommon/VertexLoader_TextCoord.cpp
+++ b/Source/Core/VideoCommon/VertexLoader_TextCoord.cpp
@@ -16,24 +16,6 @@
 
 namespace
 {
-template <int N>
-void LOG_TEX();
-
-template <>
-void LOG_TEX<1>()
-{
-  // warning: mapping buffer should be disabled to use this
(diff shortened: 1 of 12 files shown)


Source/Core/VideoCommon/VertexLoader_TextCoord.cpp
```diff
@@ -23,14 +23,14 @@ template <>
 void LOG_TEX<1>()
 {
   // warning: mapping buffer should be disabled to use this
-  // PRIM_LOG("tex: %f, ", ((float*)g_vertex_manager_write_ptr)[-1]);
+  // PRIM_LOG("tex: {}, ", ((float*)g_vertex_manager_write_ptr)[-1]);
```
iwubcode, line 26: Curious what these are used for when they are commented out?

BhaaLseN (approved):

leoetlino (approved):

PR #9289, opened 2020-11-26
Use Simplified libpng API for writing PNGs.

AdmiralCurtiss:
This makes the code way simpler and also actually writes RGB images when alpha discard is requested (instead of writing RGBA files that just have full alpha). Also deduplicates the PNG writing code.

One slight caveat here is that writing using this API automatically adds either an `sRGB` chunk (with a value of 0) or, if the `PNG_IMAGE_FLAG_COLORSPACE_NOT_sRGB` flag is set, a `gAMA` chunk (with a value of 45455). We previously wrote neither of those. It's unclear to me what this means in practice, if anything, though? I also have no idea whether things like screenshots or textures should be considered sRGB or not.


iwubcode (commented):
Code wise this looks great.  I can't say what impact the changes in colorspace you mention would bring.

I'll try to find time in the next day or two to test dynamic input textures to make sure everything still works properly.

leoetlino (approved):
LGTM codewise

Adamillo:
Idk if this pr has to something with the textures not dumping but after updating to the latest dev version (namely 5.0-13215), the textures wouldnt dump anymore. my theory states that the previous dolphin version that merged this pr (namely 5.0-13211) caused the issue not to dump the textures anymore

AdmiralCurtiss:
I'm certain I tested texture dumping but I'll take a look.

AdmiralCurtiss:
Yes, works totally fine here on 5.0-13215. (using the Vulkan backend)

Adamillo:
I think finally found out whats going on. On the latest PR the textures wont dump if the Windows User has russian letters in their names, because after I moved Dolphin to the C: folder instead of Downloads and used portable.txt, the textures finally dumped. After that, I made Dolphin create a new Documents folder with the default settings and the textures don't dump! I think the issue I have is the russian name of the Windows User!

JosJuice:
Ah, yeah, it doesn't surprise me that that would happen on Windows... We probably need to replace the call to `png_image_write_to_file` with `png_image_write_to_memory` plus `FileIO::File` or similar.

AdmiralCurtiss:
Aaah yeah I could see how that could happen if libpng doesn't assume UTF8 strings. Well that's annoying.

PR #9290, opened 2020-11-27
Update .editorconfig

aungkokk:
root = true

[*]
charset = utf-8
indent_size = 2
trim_trailing_whitespace = true

[*.{c,cpp,h,java}]
indent_style = space
insert_final_newline = true
max_line_length = 100

[{CMakeLists.txt,*.cmake}]
indent_style = tab

[.editorconfig]
insert_final_newline = true

[*.gradle]
indent_size = 4
indent_style = space
insert_final_newline = false

[*.xml]
indent_size = 4
indent_style = space
insert_final_newline = true

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -25,3 +25,4 @@ insert_final_newline = false
 indent_size = 4
 indent_style = space
 insert_final_newline = true
+.


JosJuice:
Why was this change made?

BhaaLseN:
Aside from the fact that it isn't even a valid change. It looks like they pasted the Contents into the PR instead of the file (which looks like a Web-Edit anyways).
And even that has no difference to our current `.editorconfig`.

PR #9291, opened 2020-11-27
AudioCommon: Convert alerts over to fmt-based variants

lioncash:
Continues the migration over to fmt

### DIFF
--- a/Source/Core/AudioCommon/WaveFile.cpp
+++ b/Source/Core/AudioCommon/WaveFile.cpp
@@ -9,6 +9,7 @@
 #include "Common/CommonTypes.h"
 #include "Common/File.h"
 #include "Common/FileUtil.h"
+#include "Common/Logging/Log.h"
 #include "Common/MsgHandler.h"
 #include "Common/StringUtil.h"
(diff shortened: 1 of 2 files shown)


Source/Core/AudioCommon/WaveFile.cpp
```diff
@@ -117,10 +117,10 @@ void WaveFileWriter::Write4(const char* ptr)
 void WaveFileWriter::AddStereoSamplesBE(const short* sample_data, u32 count, int sample_rate)
 {
   if (!file)
-    PanicAlertT("WaveFileWriter - file not open.");
+    PanicAlertFmtT("WaveFileWriter - file not open.");
```
BhaaLseN, line 120: It feels a bit weird that this one is translated but the other one is not. Both messages seem like they don't make too much sense being either translated or shown to the user (since they likely can't do much about it anyways)

BhaaLseN (approved):
Changes LGTM, but I'm not sure if that one even makes sense being translated.

leoetlino (approved):

PR #9292, opened 2020-11-28
Fix liblzma config on macOS

Dentomologist:
/Externals/liblzma/config.h #defines HAVE_PTHREAD_CONDATTR_SETCLOCK
which results in /Externals/liblzma/tuklib/mythread.h calling the
function pthread_condattr_setclock.  This function doesn't exist in
macOS 10.12+, which are the only macOS versions Dolphin supports, but
inserting an include guard makes macOS use an alternate function.

### DIFF
--- a/Externals/liblzma/config.h
+++ b/Externals/liblzma/config.h
@@ -190,7 +190,9 @@
 #define HAVE_POSIX_FADVISE 1
 
 /* Define to 1 if you have the `pthread_condattr_setclock' function. */
+#ifndef __APPLE__
 #define HAVE_PTHREAD_CONDATTR_SETCLOCK 1
+#endif
 
 /* Have PTHREAD_PRIO_INHERIT. */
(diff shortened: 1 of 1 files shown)


Externals/liblzma/config.h
```diff
@@ -190,7 +190,9 @@
 #define HAVE_POSIX_FADVISE 1
 
 /* Define to 1 if you have the `pthread_condattr_setclock' function. */
-#define HAVE_PTHREAD_CONDATTR_SETCLOCK 1
+#ifndef __APPLE__
+  #define HAVE_PTHREAD_CONDATTR_SETCLOCK 1
```
JosJuice, line 194: To match the rest of the file, there should be no indentation here.

JosJuice (commented):
LGTM other than that.

JosJuice:
@dolphin-emu-bot rebuild

PR #9293, opened 2020-11-28
JitArm64: Properly set m_stack_pointer

JosJuice:
In order to reach the middle guard (at `m_stack_base + GUARD_OFFSET`) before the bottom guard (at `m_stack_base`), the stack pointer must start at an address which is higher than the middle guard. It also didn't make sense that we were allocating memory and then not using the top part of it.


smurf3tte:
This just makes me wonder if JitArm64::HandleStackFault() works, given that it would have been impossible to hit... 🤔

degasus (approved):
Yes, I think this makes by far more sense. Seems like this is really a very rare edge case, which just never happens. Do you also want to update the definition of `GUARD_SIZE = 64 * 1024;` to match the number schemas of the other constants?

JosJuice:
Done. And yes, "normal" PPC software should never hit this case.

PR #9295, opened 2020-11-28
Android: Cheat GUI

idcavili:
I've started adding a cheat GUI to the Android version.  I intend for it to be accessed from the in-game menu.  It will use the Settings Activity.


leoetlino:
Reopen when this is actually ready for review.

PR #9297, opened 2020-11-28
Movie: Fix 83b9fef regressions

JosJuice:
1. Comparing `std::string_view`s does not behave the same as `strncmp` in the case where `SConfig`'s game ID is longer than 6 chars.
2. `DTMHeader::GetGameID` wasn't excluding null bytes for game IDs shorter than 6 chars.
3. `==` was accidentally used instead of `!=`.

### DIFF
--- a/Source/Core/Core/Movie.cpp
+++ b/Source/Core/Core/Movie.cpp
@@ -227,7 +227,7 @@ void Init(const BootParameters& boot)
     ReadHeader();
     std::thread md5thread(CheckMD5);
     md5thread.detach();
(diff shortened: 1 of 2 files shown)


AdmiralCurtiss:
This is all kinds of messy, there's gotta be a better way to handle this...

JosJuice:
The idea of making a new format to replace DTM has been floating around for a while, and that would (along with the main benefits) give us the ability to store variable-length strings and get rid of this awkward code. But who knows when it will happen...

leoetlino (approved):

PR #9299, opened 2020-11-29
JitArm64: Do downcount immediately before jumping to dispatcher

JosJuice:
Fixes https://bugs.dolphin-emu.org/issues/12327.

When we started using fmt in CheckExternalExceptions, JitArm64 mysteriously stopped working even though the code path where fmt was used never was reached. This is because the compiler added a function prologue and epilogue to set up the stack, since the code path that used fmt required the use of the stack.

However, the breakage didn't actually have anything to do with the usage of the stack in itself, but rather with the compiler's insertion of a stack canary. In the function epilogue, a cmp instruction was inserted to check that the stack canary had not been overwritten during the execution of the function. This cmp instruction overwriting the status flags ended up having a disastrous side effect once execution returned to code emitted by JitArm64::WriteExceptionExit.

JitArm64's dispatcher contains a branch to the "do_timing" code which is intended to be taken if the PPC downcount is negative. However, the dispatcher doesn't update the status flags on its own before this conditional branch, but rather expects the calling code to have set them as a side effect of DoDownCount. The root cause of our bug was that JitArm64::WriteExceptionExit was calling DoDownCount before Check(External)Exceptions instead of after.

### DIFF
--- a/Source/Core/Core/PowerPC/JitArm64/Jit.cpp
+++ b/Source/Core/Core/PowerPC/JitArm64/Jit.cpp
@@ -658,7 +658,7 @@ void JitArm64::DoJit(u32 em_address, JitBlock* b, u32 nextPC)
       MOVI2R(W0, static_cast<u32>(JitInterface::ExceptionType::PairedQuantize));
       MOVP2R(X1, &JitInterface::CompileExceptionCheck);
       BLR(X1);
-      B(dispatcher);
+      B(dispatcher_no_check);
       SwitchToNearCode();
       SetJumpTarget(no_fail);
       js.assumeNoPairedQuantize = true;
--- a/Source/Core/Core/PowerPC/JitArm64/Jit.cpp
+++ b/Source/Core/Core/PowerPC/JitArm64/Jit.cpp
@@ -299,8 +299,8 @@ void JitArm64::FreeStack()
 void JitArm64::WriteExit(u32 destination, bool LK, u32 exit_address_after_return)
 {
   Cleanup();
-  DoDownCount();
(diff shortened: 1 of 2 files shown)


smurf3tte:
JitArm64::DoJit() emits a branch to the dispatcher without setting up the status flags. Should this be replaced with dispatcher_no_check?

https://github.com/dolphin-emu/dolphin/blob/2863b3ff5b54388a94d26520d2ed19afd1a9c898/Source/Core/Core/PowerPC/JitArm64/Jit.cpp#L661

smurf3tte:
A more open-ended question: JitArm64BlockCache::WriteLinkBlock() can also emit branches to the dispatcher. Are the status flags guaranteed to be in the right state when it does?

JosJuice:
> JitArm64::DoJit() emits a branch to the dispatcher without setting up the status flags. Should this be replaced with dispatcher_no_check?

Yes, I believe so. I've added a commit that does this.

> A more open-ended question: JitArm64BlockCache::WriteLinkBlock() can also emit branches to the dispatcher. Are the status flags guaranteed to be in the right state when it does?

I don't think I understand how block linking works well enough to answer this.

smurf3tte:
Yeah, I started looking at it and quickly realized I'd need to set aside more time if I wanted to really understand it. :D

That said, this change seems like a solid improvement as-is.

degasus:
re `WriteLinkBlock()`: This method overwrites old WriteExit blocks, so as long as all WriteExit* calls are fine, this method should not be an issue.

degasus (approved):
Nice catch.

PR #9300, opened 2020-11-29
IOS: WD and NCD fixes

leoetlino:
This PR fixes the implementation of three ioctlvs: Lock, Unlock (in NCD) and GetInfo (in WD), based on the implementation in IOS59.

Without a correct implementation for those three ioctlvs, Nintendo's MPDL code (no clue what it stands for) will fail at initialization time and return an error to the caller. If a title does not check for errors correctly (e.g. Driver San Francisco or Tales of Graces), this can result in game crashes.

DS communication is still unimplemented but now it actually has a chance of working if it is implemented in the future.

Fixes [issue 11977](https://bugs.dolphin-emu.org/issues/11977) **but now the game hangs when cancelling**, so this is a partial fix. I am submitting this now to keep the diff small and avoid merge conflicts.

Also, the Mii Channel doesn't instantly abort DS communication anymore:
 
HACA01_2020-11-29_17-40-27

### DIFF
--- a/Source/Core/Core/IOS/Network/NCD/Manage.cpp
+++ b/Source/Core/Core/IOS/Network/NCD/Manage.cpp
@@ -20,6 +20,12 @@ NetNCDManage::NetNCDManage(Kernel& ios, const std::string& device_name) : Device
   config.ReadConfig(ios.GetFS().get());
 }
 
+void NetNCDManage::DoState(PointerWrap& p)
+{
+  Device::DoState(p);
+  p.Do(m_ipc_fd);
+}
+
 IPCCommandResult NetNCDManage::IOCtlV(const IOCtlVRequest& request)
 {
   s32 return_value = IPC_SUCCESS;
@@ -29,11 +35,51 @@ IPCCommandResult NetNCDManage::IOCtlV(const IOCtlVRequest& request)
   switch (request.request)
   {
   case IOCTLV_NCD_LOCKWIRELESSDRIVER:
+    if (!request.HasNumberOfValidVectors(0, 1))
+      return GetDefaultReply(IPC_EINVAL);
+
+    if (request.io_vectors[0].size < 2 * sizeof(u32))
+      return GetDefaultReply(IPC_EINVAL);
+
+    if (m_ipc_fd != 0)
+    {
+      // It is an error to lock the driver again when it is already locked.
+      common_result = IPC_EINVAL;
+    }
+    else
+    {
+      // NCD writes the internal address of the request's file descriptor.
+      // We will just write the value of the file descriptor.
+      // The value will be positive so this will work fine.
+      m_ipc_fd = request.fd;
+      Memory::Write_U32(request.fd, request.io_vectors[0].address + 4);
+    }
     break;
 
   case IOCTLV_NCD_UNLOCKWIRELESSDRIVER:
-    // Memory::Read_U32(request.in_vectors.at(0).address);
+  {
+    if (!request.HasNumberOfValidVectors(1, 1))
+      return GetDefaultReply(IPC_EINVAL);
+
+    if (request.in_vectors[0].size < sizeof(u32))
+      return GetDefaultReply(IPC_EINVAL);
+
+    if (request.io_vectors[0].size < sizeof(u32))
+      return GetDefaultReply(IPC_EINVAL);
+
+    const u32 request_handle = Memory::Read_U32(request.in_vectors[0].address);
+    if (m_ipc_fd == request_handle)
+    {
+      m_ipc_fd = 0;
+      common_result = 0;
+    }
+    else
+    {
+      common_result = -3;
+    }
+
     break;
+  }
 
   case IOCTLV_NCD_GETCONFIG:
     INFO_LOG_FMT(IOS_NET, "NET_NCD_MANAGE: IOCTLV_NCD_GETCONFIG");
--- a/Source/Core/Core/IOS/Network/NCD/Manage.h
+++ b/Source/Core/Core/IOS/Network/NCD/Manage.h
@@ -20,6 +20,8 @@ public:
 
   IPCCommandResult IOCtlV(const IOCtlVRequest& request) override;
 
+  void DoState(PointerWrap& p) override;
+
 private:
   enum
   {
@@ -34,5 +36,6 @@ private:
   };
 
   Net::WiiNetConfig config;
+  u32 m_ipc_fd = 0;
 };
 }  // namespace IOS::HLE::Device
--- a/Source/Core/Core/IOS/Network/WD/Command.cpp
+++ b/Source/Core/Core/IOS/Network/WD/Command.cpp
@@ -4,33 +4,330 @@
 
 #include "Core/IOS/Network/WD/Command.h"
 
-#include <algorithm>
 #include <cstring>
 #include <string>
 
+#include "Common/BitSet.h"
 #include "Common/CommonTypes.h"
 #include "Common/Logging/Log.h"
 #include "Common/Network.h"
 #include "Common/Swap.h"
-
+#include "Core/Analytics.h"
 #include "Core/HW/Memmap.h"
 #include "Core/IOS/Network/MACUtils.h"
 
 namespace IOS::HLE::Device
 {
-NetWDCommand::NetWDCommand(Kernel& ios, const std::string& device_name) : Device(ios, device_name)
+namespace
 {
+// clang-format off
+//                           Channel:   FEDC BA98 7654 3210
+constexpr u16 LegalChannelMask =      0b0111'1111'1111'1110u;
+constexpr u16 LegalNitroChannelMask = 0b0011'1111'1111'1110u;
+// clang-format on
+
+u16 SelectWifiChannel(u16 enabled_channels_mask, u16 current_channel)
+{
+  const Common::BitSet<u16> enabled_channels{enabled_channels_mask & LegalChannelMask};
+  u16 next_channel = current_channel;
+  for (int i = 0; i < 16; ++i)
+  {
+    next_channel = (next_channel + 3) % 16;
+    if (enabled_channels[next_channel])
+      return next_channel;
+  }
+  // This does not make a lot of sense, but it is what WD does.
+  return u16(enabled_channels[next_channel]);
 }
 
-// This is just for debugging / playing around.
-// There really is no reason to implement wd unless we can bend it such that
-// we can talk to the DS.
+u16 MakeNitroAllowedChannelMask(u16 enabled_channels_mask, u16 nitro_mask)
+{
+  nitro_mask &= LegalNitroChannelMask;
+  // TODO: WD's version of this function has some complicated logic to determine the actual mask.
+  return enabled_channels_mask & nitro_mask;
+}
+}  // namespace
+
+NetWDCommand::Status NetWDCommand::GetTargetStatusForMode(WD::Mode mode)
+{
+  switch (mode)
+  {
+  case WD::Mode::DSCommunications:
+    return Status::ScanningForDS;
+  case WD::Mode::AOSSAccessPointScan:
+    return Status::ScanningForAOSSAccessPoint;
+  default:
+    return Status::Idle;
+  }
+}
+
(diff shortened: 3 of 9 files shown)


Source/Core/Core/IOS/Network/NCD/Manage.cpp
```diff
@@ -29,11 +35,51 @@ IPCCommandResult NetNCDManage::IOCtlV(const IOCtlVRequest& request)
...
   case IOCTLV_NCD_UNLOCKWIRELESSDRIVER:
-    // Memory::Read_U32(request.in_vectors.at(0).address);
+  {
+    if (!request.HasNumberOfValidVectors(1, 1))
+      return GetDefaultReply(IPC_EINVAL);
+
+    if (request.in_vectors[0].size < sizeof(u32))
+      return GetDefaultReply(IPC_EINVAL);
+
+    if (request.io_vectors[0].size < sizeof(u32))
```
iwubcode, line 67: Did you mean to check the same statement twice?

leoetlino: No, that's a good catch!

Source/Core/Core/IOS/Network/WD/Command.cpp
```diff
@@ -60,14 +120,15 @@ IPCCommandResult NetWDCommand::IOCtlV(const IOCtlVRequest& request)
...
-    Info* info = (Info*)Memory::GetPointer(request.io_vectors.at(0).address);
-    memset(info, 0, sizeof(Info));
-    // Probably used to disallow certain channels?
-    memcpy(info->country, "US", 2);
-    info->ntr_allowed_channels = Common::swap16(0xfffe);
-
-    const Common::MACAddress address = IOS::Net::GetMACAddress();
-    std::copy(address.begin(), address.end(), info->mac);
+    // Nintendo does not actually check this...
+    if (request.in_vectors.size() + request.io_vectors.size() < 1)
```
iwubcode, line 124: Would it be clearer to do `request.in_vectors.empty()` ?

leoetlino: I think writing it like this makes it clearer that the code is checking for the total number of vectors, and that it requires the request to have at least one vector (in or in-out). This could be turned into !in.empty() && !io.empty() but then it would be inconsistent with the other ioctl handlers which don't necessarily only receive one vector.

iwubcode: Gotcha.  I missed that they were separate variables

Source/Core/Core/IOS/Network/NCD/Manage.cpp
iwubcode: They are separate variables.  Sorry, `in_vectors` and `io_vectors` looks very similar...

Source/Core/Core/IOS/Network/NCD/Manage.cpp
```diff
@@ -29,11 +35,51 @@ IPCCommandResult NetNCDManage::IOCtlV(const IOCtlVRequest& request)
...
+
+    const u32 request_handle = Memory::Read_U32(request.in_vectors[0].address);
+    if (m_context.ipc_fd == request_handle)
+    {
+      m_context.ipc_fd = 0;
+      common_result = 0;
+    }
+    else
+    {
+      common_result = -3;
```
BhaaLseN, line 78: Do we know what `-3` is?

leoetlino: No, it is an unknown error code. I don't remember it being used elsewhere so I'm not sure how to name it...

Pokechu22: Are the error codes in [this table](https://github.com/devkitPro/libogc/blob/b4ad94705b8d9d68fb51014615582b3b9f9ae836/libogc/network_wii.c#L157-L235) applicable?  If so, `-3` would be `EADDRINUSE`.

leoetlino: I don't think they're related, as this module also uses -4 which is the standard IPC_EINVAL. -3 is probably another standard IOS error code.

Source/Core/Core/IOS/Network/WD/Command.cpp
```diff
@@ -8,29 +8,415 @@
...
-
 #include "Core/HW/Memmap.h"
 #include "Core/IOS/Network/MACUtils.h"
 
 namespace IOS::HLE::Device
 {
+namespace
+{
+constexpr u16 LegalChannelMask = 0b111111111111110u;
+constexpr u16 LegalNitroChannelMask = 0b11111111111110u;
```
BhaaLseN, line 24: I didnt see those differed by a one, could you please fill them up to the next multiple of 8, and/or use digit separators?

leoetlino: Sure

Source/Core/Core/IOS/Network/NCD/Manage.h
```diff
@@ -33,6 +35,12 @@ class NetNCDManage : public Device
     IOCTLV_NCD_GETWIRELESSMACADDRESS = 0x8,  // NCDGetWirelessMacAddress
   };
 
+  struct Context
```
iwubcode, line 38: Maybe a dumb question - do you expect more values to be added here eventually?  Is that why there's a context for a single value?

Source/Core/Core/IOS/Network/WD/Command.cpp
```diff
@@ -8,29 +8,418 @@
...
+  {
+    const auto flags = u32(request.flags);
+    const auto mode = WD::Mode(flags & 0xFFFF);
+    const auto buffer_flags = flags & 0x7FFF0000;
+    INFO_LOG_FMT(IOS_NET, "Opening with mode={} buffer_flags={:08x}", mode, buffer_flags);
+    bool is_valid_mode = false;
+    if (m_link_state.target_status == LinkState::Status::Ready)
+    {
+      // TODO: check that state.status is not Ready
+      if (mode <= WD::Mode::_6)
```
iwubcode, line 278: Worth using `WD::IsValidMode()` for consistency?

Source/Core/Core/IOS/Network/WD/Command.cpp
```diff
@@ -8,29 +8,418 @@
...
+      {
+        m_mode = mode;
+        is_valid_mode = true;
+      }
+    }
+
+    if (is_valid_mode)
+    {
+      m_ipc_flags = flags + 1;
```
iwubcode, line 287: Any idea why `+1`?

sepalani: Probably a bit/flag used to indicate it's open or ready I guess.

Otherwise, its not uncommon to have state flags to be incremented in a setup/negotiation process when it enters the next step of the process.

Source/Core/Core/IOS/Network/NCD/Manage.h
leoetlino: Yes, there will be more values in this struct, just like in NCD.

Source/Core/Core/IOS/Network/WD/Command.cpp
leoetlino: IsValidMode would also check whether mode is greater than or equal to 1, which WD doesn't do here.

Source/Core/Core/IOS/Network/WD/Command.cpp
leoetlino: Yeah. The weird thing is that this performs an addition and not a bitwise OR though...

Source/Core/Core/IOS/Network/WD/Command.cpp
```diff
@@ -8,29 +8,418 @@
...
-
 #include "Core/HW/Memmap.h"
 #include "Core/IOS/Network/MACUtils.h"
 
 namespace IOS::HLE::Device
 {
+namespace
+{
+// clang-format off
+//                           Channel:   FEDC BA98 7654 3210
```
sepalani, line 24: Where does this "Channel" value comes from? It reminds me an RFC IPv6 address: https://tools.ietf.org/html/rfc2732

AdmiralCurtiss: I believe this is just an easy reference for which bit is which channel in the two values below.

sepalani: I see, I guess that's not related to IPv6 address representation and prefix if that's the case:
https://tools.ietf.org/html/rfc1884
https://tools.ietf.org/html/rfc3513

(The newer versions of these drafts don't have this IPv6 placeholder)

leoetlino: Yes, this is just to make it easier to see which channels are enabled.

Source/Core/Core/IOS/Network/NCD/Manage.cpp
leoetlino: Do you think it's worth adding it to the ReturnCode enum anyway (named something like `IPC_ERROR_3`), or are you fine with this as is?

JMC47:
Pokemon Battle Revolution and My Pokemon Ranch do not freeze.  So far it's just Tales of Graces.

iwubcode:
Anyone able to test any of the other games??  Google tells me:

- Batman: the Brave and Bold
- Geometry Wars Galaxies
- FFCC: Echoes of Time
- Castlevania Judgment

All had some form of DS connectivity.  Unfortunately I don't own any of those..

AdmiralCurtiss:
No crashes on Geometry Wars (RGLP7D), both the Game Share and the Connect to DS options. Can also repeatedly start and cancel to no apparent ill effect.

Kolano:
For testers a more complete list of titles with DS connectivity can be found on the Wiki here...
https://wiki.dolphin-emu.org/index.php?title=Category:Nintendo_DS_(Input_supported)

JMC47:
You can add Driver: San Francisco to the list.  It has an invisible DS multiplayer thing it broadcasts at all times.

Source/Core/Core/IOS/Network/WD/Command.cpp
```diff
@@ -4,33 +4,333 @@
...
+  return GetDefaultReply(IPC_SUCCESS);
+}
+
+IPCCommandResult NetWDCommand::GetLinkState(const IOCtlVRequest& request) const
+{
+  INFO_LOG_FMT(IOS_NET, "WD_GetLinkState called (status={}, mode={})", m_status, m_mode);
+  if (!WD::IsValidMode(m_mode))
+    return GetDefaultReply(u32(ResultCode::UnavailableCommand));
+
+  return GetDefaultReply(m_status == GetTargetStatusForMode(m_mode));
```
smurf3tte, line 276: What kind of return values are expected here? The error path returns a result code, but this success path returns a boolean.

Source/Core/Core/IOS/Network/WD/Command.cpp
```diff
@@ -4,33 +4,333 @@
...
 
+  m_ipc_owner_fd = -1;
+  m_wake_up_proxy_request_thread.Set();
+  return Device::Close(fd);
+}
+
+IPCCommandResult NetWDCommand::SetLinkState(const IOCtlVRequest& request)
+{
+  const auto* vector = request.GetVector(0);
+  if (!vector)
```
smurf3tte, line 237: Should this check that vector->address != 0 as well?

Source/Core/Core/IOS/Network/WD/Command.cpp
leoetlino: Yes, it's supposed to return a boolean here (converted to an integer). I'll make this more explicit with an int cast.

smurf3tte (approved):

BhaaLseN (approved):
I have to say, it looks a lot nicer without the (yet) unknown parts of the state machine; and thats what delayed my review so far (since I wanted to take a closer look myself on what IOS does there).
I still didn't find the time to do so, but the code looks sound. Untested though.

Pokechu22 (approved):

PR #9302, opened 2020-11-30
Qt: Add missing tr calls for tooltip descriptions

leoetlino:


JosJuice (approved):

lioncash (approved):

iwubcode:
Ah, sorry all.  I had assumed `QT_TR_NOOP` was doing some sort of magic translation.  Looks like you still need a `tr` even with it..

PR #9304, opened 2020-12-02
General: Convert PanicAlerts over to fmt equivalent

lioncash:
Converts lingering panic alert calls over to the fmt-capable ones.

### DIFF
--- a/Source/Core/VideoBackends/OGL/main.cpp
+++ b/Source/Core/VideoBackends/OGL/main.cpp
@@ -123,15 +123,15 @@ bool VideoBackend::InitializeGLExtensions(GLContext* context)
   {
     // OpenGL 2.0 is required for all shader based drawings. There is no way to get this by
     // extensions
-    PanicAlert("GPU: OGL ERROR: Does your video card support OpenGL 2.0?");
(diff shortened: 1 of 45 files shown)


Source/Core/VideoBackends/OGL/main.cpp
```diff
@@ -159,7 +159,8 @@ bool VideoBackend::FillBackendInfo()
   g_Config.backend_info.MaxTextureSize = static_cast<u32>(max_texture_size);
   if (max_texture_size < 1024)
   {
-    PanicAlert("GL_MAX_TEXTURE_SIZE too small at %i - must be at least 1024.", max_texture_size);
+    PanicAlertFmtT("GL_MAX_TEXTURE_SIZE too small at {0} - must be at least 1024.",
+                   max_texture_size);
```
JosJuice, line 163: I think translators are likely to misinterpret the word "at" in this string. What do you think about something like "GL_MAX_TEXTURE_SIZE is {0} - must be at least 1024"

BhaaLseN (approved):

leoetlino (approved):

PR #9307, opened 2020-12-03
Add File::Delete and File::DeleteDir warning flags

Dentomologist:
Adds a flag to File::Delete and File::DeleteDir dictating whether they emit console warnings when the specified path doesn't exist.  The flags are true by default to maintain existing behavior.

Also modifies the return value of DeleteDir when attempting to delete a nonexistent path to better reflect the intention of the function (nothing's there after the call).  This change doesn't break anything because none of the existing callers actually check the return value.

Adds tests for Delete and DeleteDir.

### DIFF
--- a/Source/Core/Common/FileUtil.h
+++ b/Source/Core/Common/FileUtil.h
@@ -136,12 +136,20 @@ bool CreateDir(const std::string& filename);
 // Creates the full path of fullPath returns true on success
 bool CreateFullPath(const std::string& fullPath);
 
+enum class IfAbsentBehavior
+{
+  ConsoleWarning,
+  NoConsoleWarning
+};
+
 // Deletes a given filename, return true on success
 // Doesn't supports deleting a directory
-bool Delete(const std::string& filename);
+bool Delete(const std::string& filename,
+            IfAbsentBehavior behavior = IfAbsentBehavior::ConsoleWarning);
 
 // Deletes a directory filename, returns true on success
-bool DeleteDir(const std::string& filename);
+bool DeleteDir(const std::string& filename,
+               IfAbsentBehavior behavior = IfAbsentBehavior::ConsoleWarning);
 
 // renames file srcFilename to destFilename, returns true on success
 bool Rename(const std::string& srcFilename, const std::string& destFilename);
--- a/Source/UnitTests/FileUtil.cpp
+++ b/Source/UnitTests/FileUtil.cpp
@@ -1,0 +1,103 @@
+// Copyright 2020 Dolphin Emulator Project
+// Licensed under GPLv2+
+// Refer to the license.txt file included.
+
+#include <array>
+#include <gtest/gtest.h>
+
+#include "Common/FileUtil.h"
+
+class FileUtilTest : public testing::Test
+{
+protected:
+  FileUtilTest()
+      : m_parent_directory(File::CreateTempDir()), m_file_path(m_parent_directory + "/file.txt"),
+        m_directory_path(m_parent_directory + "/dir"),
+        m_invalid_path(m_parent_directory + "/invalid.txt")
+  {
+  }
+
+  ~FileUtilTest() override
+  {
+    if (!m_parent_directory.empty())
+    {
+      File::DeleteDirRecursively(m_parent_directory);
+    }
+  }
+
+  void SetUp()
+  {
+    if (m_parent_directory.empty())
+    {
+      FAIL();
+    }
+  }
+
+  constexpr static std::array<File::IfAbsentBehavior, 2> s_warning_behaviors = {
+      File::IfAbsentBehavior::ConsoleWarning,
+      File::IfAbsentBehavior::NoConsoleWarning,
+  };
+
+  const std::string m_parent_directory;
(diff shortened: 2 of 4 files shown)


JosJuice:
@dolphin-emu-bot rebuild

lioncash:
I don't think a bool is particularly useful visually here, because

`File::Delete(path, false);` doesn't really communicate intent particularly well. Maybe consider using an enum class instead.

Dentomologist:
> I don't think a bool is particularly useful visually here, because
> 
> `File::Delete(path, false);` doesn't really communicate intent particularly well. Maybe consider using an enum class instead.

I considered that but was reluctant to add an enum that was only used by two function parameters.  If you feel that's a better route I can go ahead and do that.

lioncash:
Please do. It's much better to be noisily informed about behavior/intent, than not being informed about it at all.

Dentomologist:
Noted!  I've added the enum class as requested.

Source/Core/Common/FileUtil.h
```diff
@@ -136,12 +136,18 @@ bool CreateDir(const std::string& filename);
 // Creates the full path of fullPath returns true on success
 bool CreateFullPath(const std::string& fullPath);
 
+enum class IfAbsent
```
leoetlino, line 139: IfAbsentBehavior perhaps? That's a bit more verbose but I don't think IfAbsent is clear enough; it doesn't sound like the name of an enum class.

Dentomologist: Done.

Source/UnitTests/FileUtil.cpp
```diff
@@ -0,0 +1,57 @@
...
+#include <gtest/gtest.h>
+
+#include "Common/FileUtil.h"
+
+class FileUtilTest : public testing::Test
+{
+protected:
+  FileUtilTest()
+      : m_parent_directory{File::CreateTempDir()}, file_path(m_parent_directory + "/file.txt"),
+        directory_path(m_parent_directory + "/dir"),
```
leoetlino, line 14: This seems dangerous when m_parent_directory can be the empty string (if a temporary directory fails to be created for whatever reason).

Dentomologist: Fixed, thank you for pointing that out.  

There are a few other places in the codebase with a similar problem, but that's an issue for another PR.

Dentomologist:
I refactored a few things in the unit tests so that could use another look.  In addition to renaming the path variables to use the m_ prefix I also separated assertions into separate functions.  This is intended make it clearer exactly what is being tested, and also that the results of the delete shouldn't depend on the value of the IfAbsentBehavior.

Source/UnitTests/FileUtil.cpp
```diff
@@ -0,0 +1,102 @@
...
+{
+protected:
+  FileUtilTest()
+      : m_parent_directory(File::CreateTempDir()), m_file_path(m_parent_directory + "/file.txt"),
+        m_directory_path(m_parent_directory + "/dir"),
+        m_invalid_path(m_parent_directory + "/invalid.txt")
+  {
+  }
+
+  virtual ~FileUtilTest()
```
leoetlino, line 20:
```suggestion
  ~FileUtilTest() override
```

Source/UnitTests/FileUtil.cpp
```diff
@@ -0,0 +1,102 @@
...
+      File::IfAbsentBehavior::NoConsoleWarning,
+  };
+
+  const std::string m_parent_directory;
+  const std::string m_file_path;
+  const std::string m_directory_path;
+  const std::string m_invalid_path;
+};
+
+void DeleteShouldNotRemoveDirectory(const std::string& path, File::IfAbsentBehavior behavior)
```
leoetlino, line 47: This function (and the ones below) should be marked static.

Source/UnitTests/FileUtil.cpp
```diff
@@ -0,0 +1,102 @@
...
+  FileUtilTest()
+      : m_parent_directory(File::CreateTempDir()), m_file_path(m_parent_directory + "/file.txt"),
+        m_directory_path(m_parent_directory + "/dir"),
+        m_invalid_path(m_parent_directory + "/invalid.txt")
+  {
+  }
+
+  virtual ~FileUtilTest()
+  {
+    if (m_parent_directory != "")
```
leoetlino, line 22:
```suggestion
    if (!m_parent_directory.empty())
```

Source/UnitTests/FileUtil.cpp
```diff
@@ -0,0 +1,102 @@
...
+    {
+      File::DeleteDirRecursively(m_parent_directory);
+    }
+  }
+
+  void SetUp()
+  {
+    if (m_parent_directory == "")
```
leoetlino, line 30:
```suggestion
    if (m_parent_directory.empty())
```

leoetlino (changes requested):

Dentomologist:
Added suggested changes by @leoetlino

leoetlino (approved):

PR #9308, opened 2020-12-04
Patches for Resident Evil 2/3 audio issues

smurf3tte:
These games are erroneously zeroing buffers before they can be fully copied to ARAM by DMA. The responsible memset() calls are followed by a call to DVDRead() which issues dcbi instructions that effectively cancel the memset() on real hardware. Because Dolphin lacks dcache emulation, the effects of the memset() calls are observed, which causes missing audio.

In a comment on the original bug, phire noted that the issue can be corrected by simply nop'ing out the offending memset() calls. Because the games dynamically load different .rel executables based on the character and/or language, the addresses of these calls can vary.

To deal generally with the problem of dynamic code being loaded to fixed, known addresses, the patch engine is extended to support conditional patches which require a match against a known value. This sort of thing is already achievable with Action Replay codes, but their use depends on enabling cheats globally in Dolphin, which is not a prerequisite shared by patches.

Patches are included for every region, character, and language combination.

The end result is an approximation of the games' behavior on real hardware without the associated complexity of proper dcache emulation.

https://bugs.dolphin-emu.org/issues/9840

### DIFF
--- a/Data/Sys/GameSettings/GHAE08.ini
+++ b/Data/Sys/GameSettings/GHAE08.ini
@@ -1,0 +1,16 @@
+# GHAE08 - Resident Evil 2
+
+[OnFrame]
+# Work around a game bug that causes background sounds to be zeroed during load.
+# The bug was masked on real hardware by dcache. This patch fully fixes the bug
+# and should also work on real hardware. Dolphin doesn't emulate dcache because
+# the performance hit would be huge.
+$Fix audio issues
+# main.dol
+0x800339E4:dword:0x60000000
+# leon.rel
+0x8055ACBC:dword:0x60000000:0x4BAA8445
+# claire.rel
+0x8055AB54:dword:0x60000000:0x4BAA85AD
+[OnFrame_Enabled]
+$Fix audio issues
--- a/Source/Core/Core/PatchEngine.cpp
+++ b/Source/Core/Core/PatchEngine.cpp
@@ -90,6 +90,11 @@ void LoadPatchSection(const std::string& section, std::vector<Patch>& patches, I
           bool success = true;
           success &= TryParse(items[0], &pE.address);
           success &= TryParse(items[2], &pE.value);
+          if (items.size() >= 4)
+          {
+            success &= TryParse(items[3], &pE.comparand);
+            pE.conditional = true;
+          }
 
           const auto iter =
               std::find(s_patch_type_strings.begin(), s_patch_type_strings.end(), items[1]);
@@ -184,16 +189,20 @@ static void ApplyPatches(const std::vector<Patch>& patches)
       {
         u32 addr = entry.address;
         u32 value = entry.value;
+        u32 comparand = entry.comparand;
         switch (entry.type)
         {
         case PatchType::Patch8Bit:
-          PowerPC::HostWrite_U8(static_cast<u8>(value), addr);
+          if (!entry.conditional || PowerPC::HostRead_U8(addr) == static_cast<u8>(comparand))
+            PowerPC::HostWrite_U8(static_cast<u8>(value), addr);
           break;
         case PatchType::Patch16Bit:
-          PowerPC::HostWrite_U16(static_cast<u16>(value), addr);
+          if (!entry.conditional || PowerPC::HostRead_U16(addr) == static_cast<u16>(comparand))
+            PowerPC::HostWrite_U16(static_cast<u16>(value), addr);
           break;
         case PatchType::Patch32Bit:
-          PowerPC::HostWrite_U32(value, addr);
+          if (!entry.conditional || PowerPC::HostRead_U32(addr) == comparand)
+            PowerPC::HostWrite_U32(value, addr);
           break;
         default:
           // unknown patchtype
--- a/Source/Core/DolphinQt/Config/NewPatchDialog.cpp
+++ b/Source/Core/DolphinQt/Config/NewPatchDialog.cpp
@@ -4,6 +4,7 @@
 
 #include "DolphinQt/Config/NewPatchDialog.h"
 
+#include <QCheckBox>
 #include <QDialogButtonBox>
 #include <QGridLayout>
 #include <QGroupBox>
@@ -84,6 +85,25 @@ void NewPatchDialog::AddEntry()
   m_entry_layout->addWidget(CreateEntry(m_patch.entries[m_patch.entries.size() - 1]));
 }
 
+static u32 OnTextEdited(QLineEdit* edit, const QString& text)
+{
+  bool okay = false;
+  u32 value = text.toUInt(&okay, 16);
+
+  QFont font;
+  QPalette palette;
+
+  font.setBold(!okay);
+
+  if (!okay)
+    palette.setColor(QPalette::Text, Qt::red);
+
+  edit->setFont(font);
+  edit->setPalette(palette);
+
+  return value;
+}
+
 static bool PatchEq(const PatchEngine::PatchEntry& a, const PatchEngine::PatchEntry& b)
 {
   if (a.address != b.address)
@@ -93,6 +113,12 @@ static bool PatchEq(const PatchEngine::PatchEntry& a, const PatchEngine::PatchEn
     return false;
 
   if (a.value != b.value)
+    return false;
+
+  if (a.comparand != b.comparand)
+    return false;
+
+  if (a.conditional != b.conditional)
     return false;
 
   return true;
@@ -115,56 +141,41 @@ QGroupBox* NewPatchDialog::CreateEntry(PatchEngine::PatchEntry& entry)
   type_layout->addWidget(dword);
   type->setLayout(type_layout);
 
-  auto* offset = new QLineEdit;
+  auto* address = new QLineEdit;
   auto* value = new QLineEdit;
+  auto* comparand = new QLineEdit;
 
-  m_edits.push_back(offset);
+  m_edits.push_back(address);
   m_edits.push_back(value);
+  m_edits.push_back(comparand);
+
+  auto* conditional = new QCheckBox(tr("Conditional"));
+  auto* comparand_label = new QLabel(tr("Comparand:"));
 
   auto* layout = new QGridLayout;
   layout->addWidget(type, 0, 0, 1, -1);
-  layout->addWidget(new QLabel(tr("Offset:")), 1, 0);
-  layout->addWidget(offset, 1, 1);
+  layout->addWidget(new QLabel(tr("Address:")), 1, 0);
+  layout->addWidget(address, 1, 1);
   layout->addWidget(new QLabel(tr("Value:")), 2, 0);
   layout->addWidget(value, 2, 1);
-  layout->addWidget(remove, 3, 0, 1, -1);
+  layout->addWidget(conditional, 3, 0, 1, -1);
+  layout->addWidget(comparand_label, 4, 0);
+  layout->addWidget(comparand, 4, 1);
+  layout->addWidget(remove, 5, 0, 1, -1);
   box->setLayout(layout);
 
-  connect(offset, qOverload<const QString&>(&QLineEdit::textEdited),
-          [&entry, offset](const QString& text) {
-            bool okay = true;
-            entry.address = text.toUInt(&okay, 16);
-
-            QFont font;
-            QPalette palette;
-
-            font.setBold(!okay);
-
-            if (!okay)
-              palette.setColor(QPalette::Text, Qt::red);
-
-            offset->setFont(font);
-            offset->setPalette(palette);
-          });
+  connect(address, qOverload<const QString&>(&QLineEdit::textEdited),
+          [&entry, address](const QString& text) { entry.address = OnTextEdited(address, text); });
 
   connect(value, qOverload<const QString&>(&QLineEdit::textEdited),
-          [&entry, value](const QString& text) {
-            bool okay;
-            entry.value = text.toUInt(&okay, 16);
+          [&entry, value](const QString& text) { entry.value = OnTextEdited(value, text); });
 
-            QFont font;
-            QPalette palette;
-
-            font.setBold(!okay);
-
-            if (!okay)
-              palette.setColor(QPalette::Text, Qt::red);
-
-            value->setFont(font);
-            value->setPalette(palette);
+  connect(comparand, qOverload<const QString&>(&QLineEdit::textEdited),
+          [&entry, comparand](const QString& text) {
+            entry.comparand = OnTextEdited(comparand, text);
           });
 
-  connect(remove, &QPushButton::clicked, [this, box, offset, value, entry] {
+  connect(remove, &QPushButton::clicked, [this, box, address, value, comparand, entry] {
     if (m_patch.entries.size() > 1)
     {
       box->setVisible(false);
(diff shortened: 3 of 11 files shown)


JosJuice:
@dolphin-emu-bot rebuild

smurf3tte:
I added two new strings when updating the patch editor UI. Let me know if anything else is required to make then properly localizable.
```
  auto* conditional = new QCheckBox(tr("Conditional"));
  auto* comparand_label = new QLabel(tr("Comparand:"));
```

JosJuice:
What you've already done (adding a call to `tr`) is all you should do.

Source/Core/Core/PatchEngine.cpp
```diff
@@ -191,15 +196,37 @@ static void ApplyPatches(const std::vector<Patch>& patches)
...
         {
         case PatchType::Patch8Bit:
+          if (entry.conditional)
+          {
+            if (PowerPC::HostRead_U8(addr) != static_cast<u8>(comparand))
+            {
+              break;
+            }
+          }
           PowerPC::HostWrite_U8(static_cast<u8>(value), addr);
```
JosJuice, line 210: My personal preference would be to write this like:

```suggestion
          if (!entry.conditional || PowerPC::HostRead_U8(addr) == static_cast<u8>(comparand))
            PowerPC::HostWrite_U8(static_cast<u8>(value), addr);
```

But I'm okay with the current implementation if others prefer that.

JosJuice (commented):
Looks good to me other than one comment, though I haven't looked at the DolphinQt part in detail.

Source/Core/Core/PatchEngine.cpp
BhaaLseN: It seems unnecessarily verbose to me right now. At the very least, the two conditions should be joined into a single `if` (unless there are plans to have other conditionals in the future.)
I'm impartial to leaving the `break` vs. inverting the condition for the write.

smurf3tte: I opted for JosJuice's suggestion.

BhaaLseN:
You're welcome to combine those commits down into one, since the 2nd commit doesn't make much sense  in the resulting history. We try to do the commits right during the PR so we can refer to them later if necessary (instead of squashing during the merge).
Also, for future reference, you might be easier off using `git commit --amend` (or a similarly named switch in your GUI tool of choice) instead. Do note that this action will require a force push afterwards.

-----
To squash those commits into one before continuing; you can use the following commands (replace `upstream` with the name of the Dolphin remote and `origin` with the name of your Fork remote):
```
git fetch upstream
git reset --soft upstream/master
git commit -m "your commit message here"
git push -f origin HEAD
```
The first command makes sure you get the latest commits from `master` (from the Dolphin repository) to work on; the second picks all your changes, switches to that `master` state and puts the changes in the staging area. The third command then commits it so you have one commit total that has all the changes (which you can also replace with your commit method of choice, such as `git gui` or any other GUI tool); and the forth command pushes your current branch into your own remote (which should be your Fork) while overwriting all previous changes (to update this PR).

smurf3tte:
I can give it a shot, but I'm a bit perplexed... how is that preferable to a squash merge on PR completion?

BhaaLseN:
It makes no difference for your PR, but other (larger) ones would make it more difficult in finding problems after the fact when all we have is a single squashed commit (without the detailed info of every single commit).
Feel free to drop by on IRC if you need any help.

JosJuice:
> I can give it a shot, but I'm a bit perplexed... how is that preferable to a squash merge on PR completion?

The squash merge doesn't result in a merge commit getting added.

smurf3tte:
Now enabled by default.

Source/Core/DolphinQt/Config/NewPatchDialog.cpp
```diff
@@ -164,6 +179,23 @@ QGroupBox* NewPatchDialog::CreateEntry(PatchEngine::PatchEntry& entry)
             value->setPalette(palette);
           });
 
+  connect(comparand, qOverload<const QString&>(&QLineEdit::textEdited),
```
leoetlino, line 182: I missed this at first, but can you deduplicate the signal handler? It's the same code as lines 148-180, just with different widget variables...

Source/Core/DolphinQt/Config/PatchesWidget.cpp
```diff
@@ -145,8 +145,18 @@ void PatchesWidget::SavePatches()
 
     for (const auto& entry : patch.entries)
     {
-      lines.emplace_back(StringFromFormat("0x%08X:%s:0x%08X", entry.address,
-                                          PatchEngine::PatchTypeAsString(entry.type), entry.value));
+      if (!entry.conditional)
+      {
+        lines.emplace_back(StringFromFormat("0x%08X:%s:0x%08X", entry.address,
```
leoetlino, line 150: We should probably use [fmt](https://fmt.dev/latest/index.html) for new format calls.
```suggestion
        lines.emplace_back(fmt::format("{:#10X}:{}:{:#10X}", entry.address,
```
(`#include <fmt/format.h>` first)

Source/Core/DolphinQt/Config/PatchesWidget.cpp
```diff
@@ -145,8 +145,18 @@ void PatchesWidget::SavePatches()
...
+                                            PatchEngine::PatchTypeAsString(entry.type),
+                                            entry.value));
+      }
+      else
+      {
+        lines.emplace_back(StringFromFormat("0x%08X:%s:0x%08X:0x%08X", entry.address,
```
leoetlino, line 156: Ditto.

leoetlino (changes requested):

Source/Core/DolphinQt/Config/NewPatchDialog.cpp
```diff
@@ -84,6 +85,23 @@ void NewPatchDialog::AddEntry()
   m_entry_layout->addWidget(CreateEntry(m_patch.entries[m_patch.entries.size() - 1]));
 }
 
+void NewPatchDialog::OnTextEdited(QLineEdit* edit, const QString& text, u32* field)
```
leoetlino, line 88: Apologies, I should have been more precise. This doesn't use `this` at all, so I would turn this into either a static free function (`static void OnTextEdited(...)`) or into a lambda next to the connect calls.

Source/Core/DolphinQt/Config/NewPatchDialog.cpp
```diff
@@ -117,54 +141,40 @@ QGroupBox* NewPatchDialog::CreateEntry(PatchEngine::PatchEntry& entry)
...
   layout->addWidget(value, 2, 1);
-  layout->addWidget(remove, 3, 0, 1, -1);
+  layout->addWidget(conditional, 3, 0, 1, -1);
+  layout->addWidget(comparand_label, 4, 0);
+  layout->addWidget(comparand, 4, 1);
+  layout->addWidget(remove, 5, 0, 1, -1);
   box->setLayout(layout);
 
-  connect(offset, qOverload<const QString&>(&QLineEdit::textEdited),
-          [&entry, offset](const QString& text) {
```
smurf3tte, line 134: It just occurred to me; storing references to entries is very unsafe, since they are elements of a vector that can be allocated! I haven't made this problem worse, but I think I'll have a go at fixing it.

Data/Sys/GameSettings/GHAE08.ini
```diff
@@ -0,0 +1,13 @@
+# GHAE08 - Resident Evil 2
+
+[OnFrame]
+# Skip memset call that is followed by dcbi, simulating dcache behavior
```
phire, line 4: Maybe I'm splitting hairs here, but it doesn't really "simulate" the dcache behaviour. 

It fixes a bug that causes the original game to rely on a correctly implemented dcache,

smurf3tte: I used the word simulate because it produces the same observable side effect as a functional data cache: the memset is canceled, so to speak. In that sense, it makes Dolphin deviate less from the behavior of the game on real hardware.

It can also be fairly described as a bug fix for the game, but whereas some bug fixes would cause a greater deviation from the game's behavior on real hardware, this one does the opposite.

phire: I don't like "simulate" because it insinuates the patched behaviour will be identical to unpatched code on real hardware. 

The unpatched code has theoretical edge cases where it might produce glitchy audio even on real hardware. There is quite a bit of code between the memset and the dcbi (they are in different functions). 
  
An interrupt might come along at the wrong moment and execute more code causing extra code to be executed, which might replace and commit some cachelines. Or an intermittent disk read error might cause a thread switch before retrying.

smurf3tte: How "roughly approximating dcache behavior"? Or would you prefer if the comment went into extra detail?

phire: "approximating" is a better word.

Though more detail is always better. Imagine you stumble across this .ini file knowing nothing about RE2 or the history of this bug and the pull requests. What details would such a reader want to know in a quick summary? 

For me, the key points are:

 * It was a bug in the original game
 * The bug causes background sounds to be zeroed during load
 * The bug was masked on real hardware by dcache
 * This fully fixes the bug. This patch should also work on real hardware
 * dolphin doesn't emulate dcache because the performance hit would be huge,.

smurf3tte: I believe most of these points are touched upon by the commit message, but I can put a summary in the .ini as well.

phire: Yeah, summary in .ini. 

Sometimes the whole git repo gets rearranged and it might be hard to track back those changes to a git commit or a PR

smurf3tte: Mind if I just use your summary, paraphrased slightly?

> Work around for a game bug that causes background sounds to be zeroed during load. The bug was masked on real hardware by dcache. This patch fully fixes the bug and should also work on real hardware. Dolphin doesn't emulate dcache because the performance hit would be huge.

smurf3tte:
The last update to this PR fixed a pretty nasty heap corruption bug in the existing code. All it takes to reproduce is to add patch entries until the internal std::vector<> is reallocated, then go back and edit any of the entries that were already there. I can put that fix in a separate PR if there is still any hesitation about this change.

Source/Core/DolphinQt/Config/NewPatchDialog.cpp
```diff
@@ -18,6 +18,14 @@
 #include "Core/PatchEngine.h"
 #include "DolphinQt/QtUtils/ModalMessageBox.h"
 
+struct NewPatchEntry
+{
+  QLineEdit* address = nullptr;
+  QLineEdit* value = nullptr;
+  QLineEdit* comparand = nullptr;
+  PatchEngine::PatchEntry entry;
+};
```
JosJuice, line 27: To avoid someone re-creating the same mistake in the future, please make this struct non-copyable and non-movable, with a comment saying why.

Source/Core/DolphinQt/Config/NewPatchDialog.cpp
```diff
@@ -145,9 +135,12 @@ QGroupBox* NewPatchDialog::CreateEntry(PatchEngine::PatchEntry& entry)
   auto* value = new QLineEdit;
   auto* comparand = new QLineEdit;
 
-  m_edits.push_back(address);
-  m_edits.push_back(value);
-  m_edits.push_back(comparand);
+  auto new_entry = new NewPatchEntry;
```
JosJuice, line 138: Try to avoid using raw `new` here.

JosJuice (approved):
Code and approach LGTM, but I haven't tested the patches since I don't have any of these games.

@JMC47 If you haven't tested the patches, please do so. In particular, if Resident Evil 2 has some way to return to the main menu (even if it's only when you beat the game), please test that the patch continues working correctly after that.

Source/Core/DolphinQt/Config/NewPatchDialog.cpp
```diff
@@ -38,6 +56,10 @@ NewPatchDialog::NewPatchDialog(QWidget* parent, PatchEngine::Patch& patch)
   }
 }
 
+NewPatchDialog::~NewPatchDialog()
```
leoetlino, line 59:
```suggestion
NewPatchDialog::~NewPatchDialog() = default;
```

Source/Core/DolphinQt/Config/NewPatchDialog.cpp
```diff
@@ -115,95 +140,91 @@ QGroupBox* NewPatchDialog::CreateEntry(PatchEngine::PatchEntry& entry)
   type_layout->addWidget(dword);
   type->setLayout(type_layout);
 
-  auto* offset = new QLineEdit;
+  auto* address = new QLineEdit;
   auto* value = new QLineEdit;
+  auto* comparand = new QLineEdit;
+
+  m_entries.emplace_back(std::make_unique<NewPatchEntry>());
```
leoetlino, line 147: emplace_back returns a reference to the item so you don't have to get the last element manually:

```cpp
auto* new_entry = m_entries.emplace_back(std::make_unique<NewPatchEntry>()).get();
```

leoetlino (approved):

JMC47:
Tested Resident Evil 2 NTSC/PAL/NTSC-J and swapped modes multiple times.
Tested Resident Evil 3 NTSC/PAL.

Note that in single core, Resident Evil 3 hangs after the first area, but that is unrelated to these codes.  LGTM.

JMC47:
Oh yeah, forgot to note - if these codes are used in older versions of Dolphins, they will make the game crash.  Not sure if there's anything that can be done about it but is worth mentioning.

IceLancerSR:
I just tested 5.0-13963 and sound cut off is there again. RE2 didn't test 3, should be same issue

smurf3tte:
> I just tested 5.0-13963 win10 x64 and sound cut off is there again. RE2 didn't test 3, should be same

I'm not hearing any issues on that build. A few questions:
1. In the game list in Dolphin, if you right click the game, then Properties, then the Patches tab, is "Fix audio issues" checked?
2. Are you hearing issues at the very start of the game, or do they only occur later?
3. Is it possible you are experiencing a different audio issue? Can you compare how it sounds now vs. how it did on a version of Dolphin from before the fix (e.g. Dolphin 5.0)?

JosJuice:
Dolphin 5.0 wouldn't be appropriate to use for a comparison, since it contains the ARAM DMA timing hack which avoids the audio issue in these two games.

smurf3tte:
Good point. I noticed the progress report actually included a sample of the original bug, so I just linked that for comparison.

IceLancerSR:
> Good point. I noticed the progress report actually included a sample of the original bug, so I just linked that for comparison.

its a hickup, hear it for 1 sec then nothing for 2 sec then hear again and loop

IceLancerSR:
> Good point. I noticed the progress report actually included a sample of the original bug, so I just linked that for comparison.

My bad. I updated dolphin but somehow managed to take old patches instead of new ones. And thus missed the "patches" for this.
Sorry

smurf3tte:
No worries! Glad that's all it was.

PR #9310, opened 2020-12-04
Turn format string issues into compile-time errors

leoetlino:
If the compiler can detect an issue with a format string at compile
time, then we should take advantage of that and turn the issue into a
hard compile-time error as such problems almost always lead to UB.

This helps with catching logging or assertion messages that have been
converted over to fmt but are still using the old, non-fmt variants
of the logging macros.

This commit also fixes all incorrect usages that I could find.

(Rebased version of #9284 which GitHub stopped updating for some reason)

### DIFF
--- a/Source/CMakeLists.txt
+++ b/Source/CMakeLists.txt
@@ -93,6 +93,12 @@ else()
   if (HAS_STRINGOP_TRUNCATION_WARNING)
     check_and_add_flag(NO_STRINGOP_TRUNCATION -Wno-stringop-truncation)
   endif()
+
+  # Format string issues that the compiler can detect should be compile time errors.
+  check_cxx_compiler_flag(-Wformat HAS_FORMAT_WARNING)
+  if (HAS_FORMAT_WARNING)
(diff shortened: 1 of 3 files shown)


Source/CMakeLists.txt
```diff
@@ -93,6 +93,12 @@ else()
   if (HAS_STRINGOP_TRUNCATION_WARNING)
     check_and_add_flag(NO_STRINGOP_TRUNCATION -Wno-stringop-truncation)
   endif()
+
+  # Format string issues that the compiler can detect should be compile time errors.
+  check_cxx_compiler_flag(-Wformat HAS_FORMAT_WARNING)
+  if (HAS_FORMAT_WARNING)
+    check_and_add_flag(FORMAT_WARNING_TO_ERROR -Werror=format)
```
iwubcode, line 100: I haven't been able to find an equivalent for VS unfortunately

leoetlino: Yeah, MSVC seems to have format checks too, but only for some standard functions such as printf... :/

iwubcode (approved):
LGTM

PR #9311, opened 2020-12-05
Add caching to Config::Info

JosJuice:
The goal of this change is to make `Config::Get(const Info<T>&)` fast so that we can use it in hot paths.

This is the second iteration of PR #9103. The core idea is the same, but the way cache invalidation is handled is completely different. Compared to PR #9103, reading settings is slower (but still much faster than master) and changing settings is faster (very close to the speed of master).

A benchmark of one million calls to `Config::Get(const Info<bool>&)`, adding each returned value to an int variable:

5.0-13178: 930 ms
This PR: 27 ms

### DIFF
--- a/Source/Core/Common/Config/Config.cpp
+++ b/Source/Core/Common/Config/Config.cpp
@@ -3,6 +3,7 @@
 // Refer to the license.txt file included.
 
 #include <algorithm>
+#include <atomic>
 #include <list>
 #include <map>
 #include <mutex>
@@ -17,6 +18,7 @@ using Layers = std::map<LayerType, std::shared_ptr<Layer>>;
 static Layers s_layers;
 static std::list<ConfigChangedCallback> s_callbacks;
 static u32 s_callback_guards = 0;
+static std::atomic<u64> s_config_version = 0;
 
 static std::shared_mutex s_layers_rw_lock;
 
@@ -31,7 +33,7 @@ static void AddLayerInternal(std::shared_ptr<Layer> layer)
     const Config::LayerType layer_type = layer->GetLayer();
     s_layers.insert_or_assign(layer_type, std::move(layer));
   }
-  InvokeConfigChangedCallbacks();
+  OnConfigChanged();
 }
 
 void AddLayer(std::unique_ptr<ConfigLayerLoader> loader)
@@ -59,7 +61,7 @@ void RemoveLayer(LayerType layer)
 
     s_layers.erase(layer);
   }
-  InvokeConfigChangedCallbacks();
+  OnConfigChanged();
 }
 
 void AddConfigChangedCallback(ConfigChangedCallback func)
@@ -67,13 +69,20 @@ void AddConfigChangedCallback(ConfigChangedCallback func)
   s_callbacks.emplace_back(std::move(func));
 }
 
-void InvokeConfigChangedCallbacks()
+void OnConfigChanged()
 {
   if (s_callback_guards)
     return;
 
+  s_config_version.fetch_add(1, std::memory_order_relaxed);
+
   for (const auto& callback : s_callbacks)
     callback();
+}
+
+u64 GetConfigVersion()
+{
+  return s_config_version.load(std::memory_order_relaxed);
 }
 
 // Explicit load and save of layers
@@ -85,7 +94,7 @@ void Load()
     for (auto& layer : s_layers)
       layer.second->Load();
   }
-  InvokeConfigChangedCallbacks();
+  OnConfigChanged();
 }
 
 void Save()
@@ -96,7 +105,7 @@ void Save()
     for (auto& layer : s_layers)
       layer.second->Save();
   }
-  InvokeConfigChangedCallbacks();
+  OnConfigChanged();
 }
 
 void Init()
@@ -207,7 +216,7 @@ ConfigChangeCallbackGuard::~ConfigChangeCallbackGuard()
   if (--s_callback_guards)
     return;
 
-  InvokeConfigChangedCallbacks();
+  OnConfigChanged();
 }
 
 }  // namespace Config
--- a/Source/Core/Common/Config/Config.h
+++ b/Source/Core/Common/Config/Config.h
@@ -24,7 +24,10 @@ std::shared_ptr<Layer> GetLayer(LayerType layer);
 void RemoveLayer(LayerType layer);
 
 void AddConfigChangedCallback(ConfigChangedCallback func);
-void InvokeConfigChangedCallbacks();
+void OnConfigChanged();
+
+// Returns the number of times the config has changed in the current execution of the program
+u64 GetConfigVersion();
 
 // Explicit load and save of layers
 void Load();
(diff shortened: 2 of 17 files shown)


Source/Core/Common/Config/ConfigInfo.h
```diff
@@ -42,12 +65,64 @@ struct Info
...
+    if (m_cached_value.config_version < cached_value.config_version)
+      m_cached_value = cached_value;
+  }
+
+private:
+  Location m_location;
+  T m_default_value;
+
+  mutable CachedValue<T> m_cached_value;
+  mutable std::mutex m_cached_value_mutex;
```
smurf3tte, line 126: A good opportunity to use std::shared_mutex?

JosJuice: I wasn't aware that that class existed. It does sound like a good fit. (Not that I think we access the same setting from multiple threads often at all.)

Source/Core/Common/Config/Config.cpp
```diff
@@ -17,6 +18,7 @@ using Layers = std::map<LayerType, std::shared_ptr<Layer>>;
 static Layers s_layers;
 static std::list<ConfigChangedCallback> s_callbacks;
 static u32 s_callback_guards = 0;
+static std::atomic<u32> s_config_version = 0;
```
leoetlino, line 21: I'm not sure this will ever be an issue... but maybe use u64 just so we never have to worry about overflowing?

Source/Core/Common/Config/Config.h
```diff
@@ -24,7 +24,10 @@ std::shared_ptr<Layer> GetLayer(LayerType layer);
 void RemoveLayer(LayerType layer);
 
 void AddConfigChangedCallback(ConfigChangedCallback func);
-void InvokeConfigChangedCallbacks();
+void OnConfigChanged();
+
+// Returns the number of times the config has changed in the current execution of the program
+u32 GetConfigVersion();
```
leoetlino, line 30:
```suggestion
u64 GetConfigVersion();
```

Source/Core/Common/Config/Config.cpp
```diff
@@ -59,23 +61,30 @@ void RemoveLayer(LayerType layer)
...
   if (s_callback_guards)
     return;
 
+  s_config_version.fetch_add(1, std::memory_order_relaxed);
+
   for (const auto& callback : s_callbacks)
     callback();
 }
 
+u32 GetConfigVersion()
```
leoetlino, line 83:
```suggestion
u64 GetConfigVersion()
```

Source/Core/Common/Config/ConfigInfo.h
```diff
@@ -30,10 +34,30 @@ struct Location
...
+template <typename T>
+class Info
+{
+public:
+  constexpr Info(const Location& location, const T& default_value)
+      : m_location{location}, m_default_value{default_value}, m_cached_value{default_value, 0}
+  {
+  }
+
+  Info(const Info<T>& other)
```
leoetlino, line 52: This can be implemented using the copy assignment operator to avoid duplicating code:

```cpp
Info(const Info<T>& other) { *this = other; }
```

(or with the copy-and-swap idiom, but I prefer this one for symmetry with the move constructor and for simplicity reasons)

Source/Core/Common/Config/ConfigInfo.h
```diff
@@ -30,10 +34,30 @@ struct Location
...
+      : m_location{other.GetLocation()}, m_default_value{other.GetDefaultValue()},
+        m_cached_value{other.GetCachedValue()}
+  {
+  }
+
+  Info(Info<T>&& other)
```
leoetlino, line 58: Ditto

```cpp
Info(Info<T>&& other) { *this = std::move(other); }
```

Source/Core/Common/Config/ConfigInfo.h
```diff
@@ -30,10 +34,30 @@ struct Location
...
+      : m_location{std::move(other.m_location)}, m_default_value{std::move(other.m_default_value)},
+        m_cached_value{std::move(other.m_cached_value)}
```
leoetlino, line 60: Why is it safe to not lock the mutex here?

Source/Core/Common/Config/ConfigInfo.h
```diff
@@ -42,12 +66,64 @@ struct Info
...
+        m_cached_value{other.template GetCachedValueCasted<T>()}
+  {
+  }
+
+  Info<T>& operator=(const Info<T>& other)
+  {
+    m_location = other.GetLocation();
+    m_default_value = other.GetDefaultValue();
+    m_cached_value = other.GetCachedValue();
+  }
```
leoetlino, line 79: `return *this;` -- Clang will happily see it's UB and assume codepath leading to this function is unreachable :)

Source/Core/Common/Config/ConfigInfo.h
```diff
@@ -42,12 +66,64 @@ struct Info
...
+
+  Info<T>& operator=(Info<T>&& other)
+  {
+    m_location = std::move(other.m_location);
+    m_default_value = std::move(other.m_default_value);
+    m_cached_value = std::move(other.m_cached_value);
```
leoetlino, line 85: `return *this;`

Source/Core/Common/Config/ConfigInfo.h
```diff
@@ -42,12 +66,64 @@ struct Info
   template <typename Enum,
             std::enable_if_t<std::is_same<T, detail::UnderlyingType<Enum>>::value>* = nullptr>
   Info(const Info<Enum>& other)
-      : location{other.location}, default_value{static_cast<detail::UnderlyingType<Enum>>(
-                                      other.default_value)}
+      : m_location{other.GetLocation()}, m_default_value{static_cast<T>(other.GetDefaultValue())},
```
leoetlino, line 69: Same here, this can be implemented using the assignment operator

Source/Core/Common/Config/ConfigInfo.h
```diff
@@ -42,12 +66,64 @@ struct Info
...
+  // Make it easy to convert Info<Enum> into Info<UnderlyingType<Enum>>
+  // so that enum settings can still easily work with code that doesn't care about the enum values.
+  template <typename Enum,
+            std::enable_if_t<std::is_same<T, detail::UnderlyingType<Enum>>::value>* = nullptr>
+  Info<T>& operator=(const Info<Enum>& other)
   {
+    m_location = other.GetLocation();
+    m_default_value = static_cast<T>(other.GetDefaultValue());
+    m_cached_value = other.template GetCachedValueCasted<T>();
   }
```
leoetlino, line 97: `return *this`

leoetlino (changes requested):

Source/Core/Common/Config/ConfigInfo.h
JosJuice: Why doesn't C++ give you an error when you forget to return :(

Source/Core/Common/Config/ConfigInfo.h
JosJuice: Can't I leave out the `std::move` since `other` already is an xvalue?

Source/Core/Common/Config/ConfigInfo.h
BhaaLseN: Google suggests `-Wreturn-type` does that.

Source/Core/Common/Config/ConfigInfo.h
JosJuice: It's not that it's safe to do this without locking the mutex, it's that moving the object while another thread wants to use it doesn't make any sense even if we lock the mutex. The other thread will either read a valid setting or read some useless data depending on timing, and I can't think of any scenario where that would be acceptable.

leoetlino (approved):

PR #9312, opened 2020-12-05
InputCommon: fix dynamic input textures when host key isn't mapped

iwubcode:
This fixes dynamic input textures not being generated when a host key isn't mapped.

I broke this when I switched the logic to use regions instead of color keys...

Master:

xenoblade-master
(mapped A -> B at the end just to show that we are using dynamic input textures)

This PR:

xenoblade-pr

### DIFF
--- a/Source/Core/InputCommon/DynamicInputTextureConfiguration.cpp
+++ b/Source/Core/InputCommon/DynamicInputTextureConfiguration.cpp
@@ -300,24 +300,44 @@ bool DynamicInputTextureConfiguration::GenerateTexture(
     }
   }
 
-  // Load image copy
-  auto base_image = LoadImage(m_base_path + texture_data.m_image_name);
-  bool dirty = false;
+  // Two copies of the loaded texture
+  // The first one is used as a fallback if a key or device isn't mapped
+  // the second one is used as the final image to write to the textures directory
+  const auto original_image = LoadImage(m_base_path + texture_data.m_image_name);
+  auto image_to_write = original_image;
 
+  bool dirty = false;
   for (auto& [emulated_key, rects] : emulated_controls_iter->second)
   {
-    std::string host_key = "";
-    sec->Get(emulated_key, &host_key);
+    // TODO: Remove this line when we move to C++20
+    auto& rects_ref = rects;
+    auto apply_original = [&] {
+      for (const auto& rect : rects_ref)
+      {
+        CopyImageRegion(*original_image, *image_to_write, rect, rect);
+        dirty = true;
+      }
+    };
 
     if (!device_found)
     {
       // If we get here, that means the controller is set to a
       // device not exposed to the pack
+      // We still apply the original image, in case the user
+      // switched devices and wants to see the changes
+      apply_original();
       continue;
     }
 
+    std::string host_key;
+    sec->Get(emulated_key, &host_key);
+
     const auto input_image_iter = host_devices_iter->second.find(host_key);
-    if (input_image_iter != host_devices_iter->second.end())
+    if (input_image_iter == host_devices_iter->second.end())
+    {
+      apply_original();
+    }
+    else
     {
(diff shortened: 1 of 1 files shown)


Source/Core/InputCommon/DynamicInputTextureConfiguration.cpp
```diff
@@ -300,24 +300,42 @@ bool DynamicInputTextureConfiguration::GenerateTexture(
...
+  // the second is the image to write to the textures directory
+  auto original_image = LoadImage(m_base_path + texture_data.m_image_name);
+  auto image_to_write = LoadImage(m_base_path + texture_data.m_image_name);
 
+  bool dirty = false;
   for (auto& [emulated_key, rects] : emulated_controls_iter->second)
   {
-    std::string host_key = "";
-    sec->Get(emulated_key, &host_key);
+    auto apply_original = [&, rects = std::ref(rects)] {
```
iwubcode, line 312: In case anyone is curious why this is needed, see:  https://stackoverflow.com/questions/46114214/lambda-implicit-capture-fails-with-variable-declared-from-structured-binding

Source/Core/InputCommon/DynamicInputTextureConfiguration.cpp
```diff
@@ -300,24 +300,42 @@ bool DynamicInputTextureConfiguration::GenerateTexture(
...
     {
       // If we get here, that means the controller is set to a
       // device not exposed to the pack
+      // We still apply the original image, in case the user
+      // switched devices and wants to see the changes
+      apply_original();
       continue;
     }
 
+    std::string host_key = "";
```
leoetlino, line 330:
```suggestion
    std::string host_key;
```

Source/Core/InputCommon/DynamicInputTextureConfiguration.cpp
BhaaLseN: I wonder if this is good enough of a reason to leave a comment in source 🤔 
At least I didn't know this was a thing (nor have i seen `std::ref` before); but I've never ran into the issue either.

Alternatively...declaring a local ref inside the `for` loop to get around this? At least that sounds like a way to fix this with non-obscure syntax.

leoetlino: Yeah, maybe consider adding a `// TODO: Remove the initializer once we can use C++20` comment like we did in the GC memcard code when we hit this issue last time.

Source/Core/InputCommon/DynamicInputTextureConfiguration.cpp
```diff
@@ -300,24 +300,44 @@ bool DynamicInputTextureConfiguration::GenerateTexture(
     }
   }
 
-  // Load image copy
-  auto base_image = LoadImage(m_base_path + texture_data.m_image_name);
-  bool dirty = false;
+  // Load source texture twice
+  // once as an original image to fallback to if a key or device isn't mapped
+  // the second is the image to write to the textures directory
+  auto original_image = LoadImage(m_base_path + texture_data.m_image_name);
```
leoetlino, line 306: Can this be made const?

Source/Core/InputCommon/DynamicInputTextureConfiguration.cpp
```diff
@@ -300,24 +300,44 @@ bool DynamicInputTextureConfiguration::GenerateTexture(
...
+  auto image_to_write = LoadImage(m_base_path + texture_data.m_image_name);
```
leoetlino, line 307: `auto image_to_write = original_image;`? I don't think the image needs to be loaded twice -- you just want a copy of the source texture, right?

leoetlino (approved):

Source/Core/InputCommon/DynamicInputTextureConfiguration.cpp
iwubcode: Good catch

PR #9313, opened 2020-12-05
WiiUtils: Check hashes to determine if a title is installed and up-to-date

leoetlino:
Nintendo's official title installation code and ES both only look at
content IDs but we should probably check for content hashes in addition
to checking for IDs for at least two reasons:

1. Some of the installed contents could be corrupted -- this cannot be
   easily detected without checking hashes.

2. Some mod distributors do not bother to update content IDs, which
   means that installing updated WADs from the UI would not actually
   update the installed game. This is confusing for users.

To keep the existing semantic (for IOS especially), the new content
hash checks are opt-in for callers of GetStoredContentsFromTMD.

This commit changes WiiUtils's WAD installation logic to enable
the content hash checks.

### DIFF
--- a/Source/Core/Core/IOS/ES/ES.h
+++ b/Source/Core/Core/IOS/ES/ES.h
@@ -81,6 +81,12 @@ public:
     s32 ipc_fd = -1;
   };
 
+  enum class CheckContentHashes : bool
+  {
+    Yes = true,
+    No = false,
+  };
+
   IOS::ES::TMDReader FindImportTMD(u64 title_id) const;
   IOS::ES::TMDReader FindInstalledTMD(u64 title_id) const;
   IOS::ES::TicketReader FindSignedTicket(u64 title_id) const;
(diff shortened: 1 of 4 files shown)


Source/Core/Core/IOS/ES/ES.h
```diff
@@ -81,6 +81,12 @@ class ES final : public Device
     s32 ipc_fd = -1;
   };
 
+  enum class CheckContentHashes : bool
```
iwubcode, line 84: Silly question - what's the point of this deriving from bool?

leoetlino: It's not strictly necessary, but I think this makes it very obvious that this enum class is just a glorified boolean.

Source/Core/Core/IOS/ES/NandUtils.cpp
```diff
@@ -174,10 +177,32 @@ std::vector<IOS::ES::Content> ES::GetStoredContentsFromTMD(const IOS::ES::TMDRea
...
-                        m_ios.GetFS()->GetMetadata(PID_KERNEL, PID_KERNEL, path).Succeeded();
+                 if (path.empty())
+                   return false;
+
+                 // Check whether the content file exists.
+                 const auto file = fs->OpenFile(PID_KERNEL, PID_KERNEL, path, FS::Mode::Read);
+                 if (!file.Succeeded())
+                   return false;
+                 const auto status = file->GetStatus();
+                 if (!status.Succeeded())
```
iwubcode, line 192: Just curious, why would a status be not succeeded but the file would succeed?

dragonbane0:
This would be amazing. I got truly stuck on this at least twice during Homebrew dev because I thought a change I made was just not working.

Source/Core/Core/IOS/ES/NandUtils.cpp
leoetlino: You're right, this shouldn't be possible. I've added a comment to FileSystem::GetStatus to mention it will always succeed for valid file descriptors.

iwubcode (approved):
Untested

PR #2663, opened 2015-06-26
Jit64: dcbf + dcbi

degasus:
Both dcbx instructions are called quite often and usually NOPs. So inline to first check within the JIT itself to skip over them much faster.

PR #9314, opened 2020-12-06
Jit: Fix correctness issue in dcbf/dcbi/dcbst

smurf3tte:
PR #2663 added a Jit64 implementation of dcbX and a fast path to skip JIT cache invalidation. Unfortunately, there is a mismatch between address spaces in this optimization. It tests the effective address (with the top 3 bits cleared) against the valid block bitset which is intended to be indexed by physical address. While this works in the common case, it fails (for example) when the effective address is in the 7E... region (a.k.a. "fake VMEM"). This may also fall apart under more complex memory mapping scenarios requiring full MMU emulation.

The good news is that even without this fast path, the underlying call to JitInterface::InvalidateICache() still does very little work in the common case. It correctly translates the effective address to a physical address which it tests against the valid block bitset, skipping invalidation if it is not necessary. As such, the cost of removing the fast path should not be too high.

The Jit64 implementation is retained, though all it does now is emit a call. This is marginally more efficient than simple interpreter fallback, which involves an extra call. The JitArm64 implementation has also been fixed.

The game Happy Feet is fixed by this change, as it loads code in the 7E... address region and depends upon JIT cache invalidation in reponse to dcbf.

https://bugs.dolphin-emu.org/issues/12133

### DIFF
--- a/Source/Core/Core/PowerPC/Jit64/Jit_LoadStore.cpp
+++ b/Source/Core/Core/PowerPC/Jit64/Jit_LoadStore.cpp
@@ -233,37 +233,21 @@ void Jit64::dcbx(UGeckoInstruction inst)
   JITDISABLE(bJITLoadStoreOff);
 
   X64Reg addr = RSCRATCH;
-  X64Reg value = RSCRATCH2;
   RCOpArg Ra = inst.RA ? gpr.Use(inst.RA, RCMode::Read) : RCOpArg::Imm32(0);
   RCOpArg Rb = gpr.Use(inst.RB, RCMode::Read);
-  RCX64Reg tmp = gpr.Scratch();
-  RegCache::Realize(Ra, Rb, tmp);
+  RegCache::Realize(Ra, Rb);
 
   MOV_sum(32, addr, Ra, Rb);
+  AND(32, R(addr), Imm8(~31));
 
-  // Check whether a JIT cache line needs to be invalidated.
-  LEA(32, value, MScaled(addr, SCALE_8, 0));  // addr << 3 (masks the first 3 bits)
-  SHR(32, R(value), Imm8(3 + 5 + 5));         // >> 5 for cache line size, >> 5 for width of bitset
-  MOV(64, R(tmp), ImmPtr(GetBlockCache()->GetBlockBitSet()));
-  MOV(32, R(value), MComplex(tmp, value, SCALE_4, 0));
-  SHR(32, R(addr), Imm8(5));
-  BT(32, R(value), R(addr));
-
-  FixupBranch c = J_CC(CC_C, true);
-  SwitchToFarCode();
-  SetJumpTarget(c);
   BitSet32 registersInUse = CallerSavedRegistersInUse();
   ABI_PushRegistersAndAdjustStack(registersInUse, 0);
   MOV(32, R(ABI_PARAM1), R(addr));
-  SHL(32, R(ABI_PARAM1), Imm8(5));
   MOV(32, R(ABI_PARAM2), Imm32(32));
   XOR(32, R(ABI_PARAM3), R(ABI_PARAM3));
   ABI_CallFunction(JitInterface::InvalidateICache);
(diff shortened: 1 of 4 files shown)


Source/Core/Core/PowerPC/Jit64/Jit_LoadStore.cpp
```diff
@@ -233,37 +233,22 @@ void Jit64::dcbx(UGeckoInstruction inst)
...
-  MOV(64, R(tmp), ImmPtr(GetBlockCache()->GetBlockBitSet()));
-  MOV(32, R(value), MComplex(tmp, value, SCALE_4, 0));
-  SHR(32, R(addr), Imm8(5));
-  BT(32, R(value), R(addr));
-
-  FixupBranch c = J_CC(CC_C, true);
-  SwitchToFarCode();
-  SetJumpTarget(c);
   BitSet32 registersInUse = CallerSavedRegistersInUse();
   ABI_PushRegistersAndAdjustStack(registersInUse, 0);
```
degasus, line 245: The main performance enhancement of the assembly version instead of calling the C++ handler is to avoid flushing the register. However the push/pop combination is often even worse then just flushing the register cache. So I'd suggest to use a `gpr.Flush(); fpr.Flush();` instead for unconditional C++ calls.

JosJuice: Would there be any performance advantage to doing that compared to falling back to the interpreter (which also does a `gpr.Flush(); fpr.Flush();`)?

degasus: Not really. Once you flush the register cache (or use push/pop), the speedup is vanished.

smurf3tte: From limited inspection, it looks like these instructions are generally used in tight loops in small functions. In such cases, only a small number of guest registers need to be flushed (or alternatively, host registers saved), so the exact mechanism does not make a big difference to the generated code.

Interpreter fallback involves an extra call to Interpreter::dcb*(), which in turn calls JitInterface::InvalidateICache(), and the effective address calculation will take a few more instructions than the JIT. That said, it does have the benefit of being less code, so if any change is to be made, I am leaning in this direction.

JosJuice: > Interpreter fallback involves an extra call to Interpreter::dcb*(), which in turn calls JitInterface::InvalidateICache(), and the effective address calculation will take a few more instructions than the JIT.

Ah, I hadn't considered that. I suppose it is better to go with something like the current solution (either with push/pop or with flush) instead of the interpreter fallback then.

smurf3tte: In IRC, degasus suggested to leave this as-is.

Source/Core/Core/PowerPC/Jit64/Jit_LoadStore.cpp
```diff
@@ -233,37 +233,22 @@ void Jit64::dcbx(UGeckoInstruction inst)
   JITDISABLE(bJITLoadStoreOff);
 
   X64Reg addr = RSCRATCH;
-  X64Reg value = RSCRATCH2;
   RCOpArg Ra = inst.RA ? gpr.Use(inst.RA, RCMode::Read) : RCOpArg::Imm32(0);
   RCOpArg Rb = gpr.Use(inst.RB, RCMode::Read);
   RCX64Reg tmp = gpr.Scratch();
```
MerryMage, line 238: `tmp` is no longer required and can be removed

MerryMage (approved):

smurf3tte:
Now fixed in JitArm64 without interpreter fallback.

degasus (approved):

JMC47:
This is causing some significant slowdown in certain games.  Noticed in Super Mario Sunshine on boot and Arc Rise Fantasia between transitions.

PR #3743, opened 2016-03-23
[WIP] Software renderer's specialCase breaks sms-water

OrN:
If the current texture's projection is XF_TEXPROJ_STQ the specialCase still triggers, and messes up rendering in the post matrix stage.

---
This change is [Reviewable](https://reviewable.io/reviews/dolphin-emu/dolphin/3743)

PR #9315, opened 2020-12-06
Software: Fix failed special case assertion

Pokechu22:
This case is hit with Super Mario Sunshine's water.

<details>
<summary>Old/incorrect information</summary>

Draft PR to test fifoci.  I don't fully understand what's going on here, only that [there's a fifoci case](https://fifo.ci/dff/sms-water/) for it that fails with an assertion on the software renderer.

The special case in question is this:

https://github.com/dolphin-emu/dolphin/blob/2952f99f691e69e450c5ac52dca9516483ade2ee/Source/Core/VideoBackends/Software/SWVertexLoader.cpp#L124-L127

which honestly seems a bit odd to me since `m_tex_gen_special_case` is used for multiple vertices (but I don't know much about rendering).  The [linked libogc documentation](https://libogc.devkitpro.org/gx_8h.html#a55a426a3ff796db584302bddd829f002) doesn't really clear things up.  In any case, it looks like it's only appropriate for `XF_TEXPROJ_ST`, which seems to be what the assert indicates.

I think the same special case is implemented for the hardware backends here, but I don't really understand it:

https://github.com/dolphin-emu/dolphin/blob/2952f99f691e69e450c5ac52dca9516483ade2ee/Source/Core/VideoCommon/VertexShaderGen.cpp#L426-L436
</details>

(See also: #3743)

### DIFF
--- a/Source/Core/VideoBackends/Software/SWVertexLoader.cpp
+++ b/Source/Core/VideoBackends/Software/SWVertexLoader.cpp
@@ -84,7 +84,12 @@ void SWVertexLoader::DrawCurrentBatch(u32 base_index, u32 num_indices, u32 base_
           &m_vertex, (VertexLoaderManager::g_current_components & VB_HAS_NRM2) != 0, outVertex);
     }
     TransformUnit::TransformColor(&m_vertex, outVertex);
-    TransformUnit::TransformTexCoord(&m_vertex, outVertex, m_tex_gen_special_case);
+    // Special case for verticies if they only contain position and tex coord 0.
+    // There is no VB_HAS_POS constant as all verticies have positions, so only UV0 must be
+    // checked. The special case also only applies to XF_TEXPROJ_ST, but that is checked on a
+    // per-vertex basis.  https://libogc.devkitpro.org/gx_8h.html#a55a426a3ff796db584302bddd829f002
+    const bool specialCaseIfST = VertexLoaderManager::g_current_components == VB_HAS_UV0;
+    TransformUnit::TransformTexCoord(&m_vertex, outVertex, specialCaseIfST);
 
     // assemble and rasterize the primitive
     m_setup_unit.SetupVertex();
@@ -120,11 +125,6 @@ void SWVertexLoader::SetFormat(u8 attributeIndex, u8 primitiveType)
   m_vertex.texMtx[5] = xfmem.MatrixIndexB.Tex5MtxIdx;
   m_vertex.texMtx[6] = xfmem.MatrixIndexB.Tex6MtxIdx;
   m_vertex.texMtx[7] = xfmem.MatrixIndexB.Tex7MtxIdx;
-
-  // special case if only pos and tex coord 0 and tex coord input is AB11
-  // http://libogc.devkitpro.org/gx_8h.html#a55a426a3ff796db584302bddd829f002
-  m_tex_gen_special_case = VertexLoaderManager::g_current_components == VB_HAS_UV0 &&
-                           xfmem.texMtxInfo[0].projection == XF_TEXPROJ_ST;
 }
 
 template <typename T, typename I>
--- a/Source/Core/VideoBackends/Software/SWVertexLoader.h
+++ b/Source/Core/VideoBackends/Software/SWVertexLoader.h
@@ -28,6 +28,4 @@ protected:
 
   InputVertexData m_vertex;
   SetupUnit m_setup_unit;
-
-  bool m_tex_gen_special_case;
 };
--- a/Source/Core/VideoBackends/Software/TransformUnit.cpp
+++ b/Source/Core/VideoBackends/Software/TransformUnit.cpp
@@ -109,7 +109,7 @@ void TransformNormal(const InputVertexData* src, bool nbt, OutputVertexData* dst
   }
 }
 
-static void TransformTexCoordRegular(const TexMtxInfo& texinfo, int coordNum, bool specialCase,
+static void TransformTexCoordRegular(const TexMtxInfo& texinfo, int coordNum, bool specialCaseIfST,
                                      const InputVertexData* srcVertex, OutputVertexData* dstVertex)
 {
   Vec3 src;
@@ -140,7 +140,7 @@ static void TransformTexCoordRegular(const TexMtxInfo& texinfo, int coordNum, bo
 
   if (texinfo.projection == XF_TEXPROJ_ST)
   {
-    if (texinfo.inputform == XF_TEXINPUT_AB11 || specialCase)
+    if (texinfo.inputform == XF_TEXINPUT_AB11)
       MultiplyVec2Mat24(src, mat, *dst);
     else
       MultiplyVec3Mat24(src, mat, *dst);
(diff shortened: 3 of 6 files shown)


Miksel12:
Fixes: https://bugs.dolphin-emu.org/issues/7476

Pokechu22:
The tests that are still failing [on fifoci](https://fifo.ci/version/c0be1b6ca37901dbfec64e85e9ca131bc130d517/) are [lloyd](https://fifo.ci/dff/ssbb-mod-lloyd/) (related to https://bugs.dolphin-emu.org/issues/9268, fails `ASSERT(texinfo.sourcerow == XF_SRCCOLORS_INROW)` on line 432 of TransformUnit.cpp; `sourcerow` is `XF_SRCTEX1_INROW` (6)) and all of the Rogue Squadron 2 ones (which segfualt per [the buildbot log](https://dolphin.ci/#/builders/17/builds/2416/steps/7/logs/stdio) (lines 309, 1624, 2939, and 4509)).  The RS2 issue started in 609a17a0cd41a12c0fbf9a7c47c8889232594778 (hybrid XFB) and also happens on Windows (where it fails with `Critical error detected c0000374` (heap corruption)), and seems to be https://bugs.dolphin-emu.org/issues/11393.  rs3-bumpmapping also looks incorrect, but no longer hits an assertion; the incorrect appearance does not happen on the hardware renderers.

OrN:
If I'm not mistaking, rs3-bumpmap looks that way because of another thing that was fixed only in HW renderers, can't quite remember what it was. It was never added to the SW renderer if I remember correctly.

OrN:
The code you mention in the HW backend isn't the same thing. This is the software renderer implementation of it https://github.com/dolphin-emu/dolphin/blob/c0be1b6ca37901dbfec64e85e9ca131bc130d517/Source/Core/VideoBackends/Software/TransformUnit.cpp#L190

Pokechu22:
Hmm, I can't actually find the special case in the hardware renderer at all... nothing else references `XF_TEXPROJ_ST`, and the only place that references `XF_TEXPROJ_STQ` is this:

https://github.com/dolphin-emu/dolphin/blob/608dd6a37b1f0208340063b5ffe34193eea7ebc3/Source/Core/VideoCommon/VertexShaderGen.cpp#L370-L402

and the only reference to `XF_TEXINPUT_AB11` outside of SW is this:

https://github.com/dolphin-emu/dolphin/blob/608dd6a37b1f0208340063b5ffe34193eea7ebc3/Source/Core/VideoCommon/VertexShaderGen.cpp#L334-L337

(along with the ubershader version [here](https://github.com/dolphin-emu/dolphin/blob/12a215c232b5279d2e554212ae5e29906b146a19/Source/Core/VideoCommon/UberShaderVertex.cpp#L483-L504) and [here](https://github.com/dolphin-emu/dolphin/blob/12a215c232b5279d2e554212ae5e29906b146a19/Source/Core/VideoCommon/UberShaderVertex.cpp#L436-L440)).  I don't see anything that looks like the checks for the special case at all.

Pokechu22:
Further testing indicates that nothing significant changes [in fifoci](https://github.com/Pokechu22/ghidra-mn102-lang/blob/master/README.md) if the special case is removed completely (the above is a diff from [an earlier commit in this PR](https://fifo.ci/version/c0be1b6ca37901dbfec64e85e9ca131bc130d517/), not from master; thus it is able to reveal that there are no changes for sms-water and negligible changes for rs3-bumpmapping).

[This code](https://github.com/dolphin-emu/dolphin/blob/608dd6a37b1f0208340063b5ffe34193eea7ebc3/Source/Core/VideoCommon/VertexShaderGen.cpp#L370-L402) in VertexShaderGen is definitely the equivalent to the changed SWVertexLoader code; the part where it [sets z to 1 in AB11](https://github.com/dolphin-emu/dolphin/blob/608dd6a37b1f0208340063b5ffe34193eea7ebc3/Source/Core/VideoCommon/VertexShaderGen.cpp#L370-L402) is equivalent to the two different branches for AB11 in SW (`MultiplyVec2Mat24` and `MultiplyVec2Mat34` act as if `z` is 1 and add both columns 3 and 4 of the matrix instead of multiplying column 3 by z first).  The special case version does not exist in the hardware renderer.  The only thing I'm not certain about is why the special case was added in the first place, then, and if it _ever_ existed in the hardware renderer.

Pokechu22:
Looks like the special case was first introduced in a19be73be32b3117719b201fbfd9f909df1a65ed, checking for position, UV0, and AB11.  It was tweaked slightly in cc7c6cd35f0cccb19a4e3681259b55cebeeb80ba ([these lines](https://github.com/dolphin-emu/dolphin/commit/cc7c6cd35f0cccb19a4e3681259b55cebeeb80ba#diff-36fc84d958c6cdd6b4588cd48620fbb4b1729ca4c6cb3b182d2ba92065b0eb81L163-R164)) (changing the `z` (`q`?) value from 0 to 1).  It was changed to checking for projection of ST and `texinfo.inputform == XF_TEXINPUT_AB11 || specialCase` was added in 1f660de7e53eb3675bc258dc57dfff99d779fc55, though there doesn't seem to be a super clear explanation as to why either of those were changed.  The most likely explanation is that it's an optimization to not compute the `z` value since it's ignored in the special case, but that doesn't quite explain the assertion for `XF_TEXPROJ_STQ`.  If the whole matrix multiplication special case was just an optimization in dolphin that doesn't have any side-effects, I'll be a bit annoyed.

The initial introduction also had the special case in the hardware renderer.  A reverse blame shows that it was disabled in 896850b1767ea041211d1335af748970da3d5fe9 because it "has caused problems in some games", and was eventually dropped in 4c5c33c5145ee33052b28625f24f2624b0b2997c (#3196).  The hardware renderer version only looked at AB11, and only touched multiplication by the post-matrix, not the other condition (where the assert is).

[Libogc's documentation](https://libogc.devkitpro.org/gx_8h.html#a55a426a3ff796db584302bddd829f002) says that the special case only triggers "[w]hen a vertex contains only position and one texture coordinate and the texgen type is `GX_TG_MTX2x4`".  I'm not quite sure what `GX_TG_MTX2x4` corresponds to, but I am more inclined to believe that it's `XF_TEXPROJ_ST` than that it's `AB11` (please do correct me if I'm wrong here).  It also only mentions normalization, not the matrix multiplication bit.

I'll try to reimplement the special case in both SW and HW, though first I'll push a commit that _just_ removes the assertion without fixing anything so that I can abuse fifoci diffing again.

Pokechu22:
Here's what [the changes now look like](https://fifo.ci/version/6cbcac52291f5db1e5baee741fc7fd114a66e187/) (again a diff from [an earlier change](https://fifo.ci/version/0dfa573ead245efb6ac53397e5d23652139c9af8/), this time one that just ignores the assert).  The only changes are [sms-water](https://fifo.ci/compare/5788137-5787815/) (fixed water texture, [bug 7476](https://bugs.dolphin-emu.org/issues/7476)) and [rs3-bumpmapping](https://fifo.ci/compare/5788109-5787852/) (a few pixels), both on the software renderer; nothing is different on the hardware renderer.  Similarly, [sms-water](https://fifo.ci/compare/5788137-5785129/) and [rs3-bumpmapping](https://fifo.ci/compare/5788109-5785121/) render identically to if the special case is removed.  So the only thing those tests are picking up on is the incorrect implementation of the special case.

As I mention in 6cbcac52291f5db1e5baee741fc7fd114a66e187, the problem with the existing implementation was that it only checked if the first texinfo's projection was `XF_TEXPROJ_ST`, and then used that to determine whether the special case applied to all of the vertices, even if they were `XF_TEXPROJ_STQ`.  The fix is to check the projection for each vertex.

What's left with this PR is to create an actual hardware test for this special case; something that actually detects the difference in normalization.  Unfortunately, I'm not quite sure how to do that, as I really have no idea how to actually do graphics programming, and the test would need to produce something that's obviously different for it to be useful (not the slightly differently white pixels from rs3-bumpmapping).  I do think the code in this PR is ready for review (apart from a "verticies" typo) though.

Pokechu22:
I wrote a [test case](https://gist.github.com/Pokechu22/0644659572c92d247ea7c703889015d5) for it.  On hardware, normalization was performed if and only if normalization was enabled; the special case didn't happen.  On dolphin (with this change/as it stands in SW), the special case did trigger for `XF_TEXPROJ_ST`, producing a visually different image.  As a test, if I changed dolphin's special case to AB11 (as it was at introduction in a19be73be32b3117719b201fbfd9f909df1a65ed), the image was visually different with both `XF_TEXPROJ_ST` and `XF_TEXPROJ_STQ`.  Both of those are wrong, so I'm not sure if the special case actually exists; if it does exist, my test case doesn't hit it.

I did also confirm that `XF_TEXPROJ_ST` is equivalent to libogc's `GX_TG_MTX2x4`, and `XF_TEXPROJ_STQ` is equivalent to `GX_TG_MTX3x4`.  The relevant code is [here](https://github.com/devkitPro/libogc/blob/8fc51464ddaf40e97fc6c96ebcb2587e796b63c3/libogc/gx.c#L2717-L2722) and [here](https://github.com/dolphin-emu/dolphin/blob/eb6fd56c1d820924a2b455084e689515b5e2bc66/Source/Core/VideoCommon/XFMemory.h#L170-L171) (bit 1, projection, is set to 1 &rArr; `XF_TEXPROJ_STQ` if `GX_TG_MTX3x4` is used).  `AB11` is a bit less obvious to me, but occurs in any case other than `GX_TG_POS` and `GX_TG_NRM` (where `stq = 1`), going into bit 2, inputform.  I haven't done any concrete tests where it was not `AB11`, because it's not obvious to me how those other modes are used.

AdmiralCurtiss:
So if the conclusion here is that the special case doesn't exist, where did the idea of it even come from?

Pokechu22:
Closing this in favor of #9424 at least for now; I still am not sure where the special case came from and it's probably better to just get rid of it and reintroduce it if it's eventually found to be actually significant (which seems unlikely at this point since it only exists in the software renderer).

PR #9316, opened 2020-12-10
GameSettings: Add patch for Monster High: Ghoul Spirit (SAOE78/SAOEVZ)

smurf3tte:
The first call to GXCopyDisp() corrupts game data, but on real hardware it isn't observed thanks to the data cache. Skipping the call works too, preventing a crash on boot.

This patch is enabled default.

### DIFF
--- a/Data/Sys/GameSettings/SAOE78.ini
+++ b/Data/Sys/GameSettings/SAOE78.ini
@@ -1,0 +1,9 @@
+# SAOE78 - Monster High: Ghoul Spirit
+
+[OnFrame]
+# The first call to GXCopyDisp() corrupts game data, but on real hardware
+# it isn't observed thanks to the data cache. Skipping the call works too.
+$Fix crash on boot
+0x803A5F20:dword:0x60000000
+[OnFrame_Enabled]
+$Fix crash on boot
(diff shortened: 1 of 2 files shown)


leoetlino:
Marking this as "do not merge" until #9317 is merged.

Data/Sys/GameSettings/SAOE78.ini
```diff
@@ -0,0 +1,7 @@
+# SAOE78 - Monster High: Ghoul Spirit
+
+[OnFrame]
+# The first call to GXCopyDisp() corrupts game data, but on real hardware
+# it isn't observed thanks to the data cache. Skipping the call works too.
+$Fix crash on boot
+0x803A5F20:dword:0x60000000
```
leoetlino, line 7: Please enable this by default, with

```ini
[OnFrame_Enabled]
$Fix crash on boot
```

leoetlino (changes requested):

Data/Sys/GameSettings/SAOE78.ini
smurf3tte: Done!

smurf3tte:
With help from @JMC47 I was able to add support for the version of the game published by Little Orbit (SAOEVZ). They also confirmed that the PAL release does not need this fix. Thanks JMC47!

JMC47:
PAL doesn't need a patch.  Tested the patch for the v1.01 version of the game.

leoetlino (approved):

ytrezq:
@smurf3tte would it be possible to produce a patch for an Icache access when going in game menus, for this game I own https://bugs.dolphin-emu.org/issues/8037?

PR #9319, opened 2020-12-11
Core: Don't copy default _Enabled sections to user INIs

JosJuice:


smurf3tte (approved):
Works on my machine.™

PR #9321, opened 2020-12-12
Software: Fix out of bounds accesses in CopyRegion

Pokechu22:
Fixes [issue 11393](https://bugs.dolphin-emu.org/issues/11393) (I haven't actually tested the game, only fifologs).

All of the Rogue Squadron 2 fifologs segfualt (per [the buildbot log](https://dolphin.ci/#/builders/17/builds/2416/steps/7/logs/stdio) (lines 309, 1624, 2939, and 4509)), starting in 609a17a0cd41a12c0fbf9a7c47c8889232594778 (hybrid XFB).  This also happens on Windows (where it fails with `Critical error detected c0000374` (heap corruption))

The problem is that left and top make no sense for a width by height array; they only make sense in a larger array where from which a smaller part is extracted.  EncodeXFB already handles the extraction, so CopyRegion's only use is to resize the image.  The rects in ScaleTexture have left and top as 0 (from [TextureConfig::GetRect](https://github.com/Pokechu22/dolphin/blob/b4b2a12974da96a684755ba473a67c496f28ab16/Source/Core/VideoCommon/TextureConfig.cpp#L21-L24) as called by [TextureCacheBase::ScaleTextureCacheEntryTo](https://github.com/Pokechu22/dolphin/blob/b4b2a12974da96a684755ba473a67c496f28ab16/Source/Core/VideoCommon/TextureCacheBase.cpp#L404-L407)), so this changes nothing for them.  I'm not entirely sure why `Rectangle` is used here at all; a `Size` or something like that (which I don't think currently exists) would be better.

This is a draft to check for regressions with FifoCI (like #9315).

### DIFF
--- a/Source/Core/VideoBackends/Software/CopyRegion.h
+++ b/Source/Core/VideoBackends/Software/CopyRegion.h
@@ -1,33 +1,53 @@
 #pragma once
 
+#include "Common/Assert.h"
 #include "Common/MathUtil.h"
 
 #include <cmath>
 
 namespace SW
 {
-// Modified from
-// http://tech-algorithm.com/articles/nearest-neighbor-image-scaling/
+// Copies a region of source to a region of destination, performing nearest-neighbor rescaling
 template <typename T>
-void CopyRegion(const T* const source, const MathUtil::Rectangle<int>& srcrect, T* destination,
-                const MathUtil::Rectangle<int>& dstrect)
+void CopyRegion(const T* const source, const MathUtil::Rectangle<int>& srcrect, const int src_width,
+                const int src_height, T* destination, const MathUtil::Rectangle<int>& dstrect,
+                const int dst_width, const int dst_height)
 {
+  ASSERT(srcrect.top >= 0 && srcrect.bottom <= src_height);
+  ASSERT(srcrect.left >= 0 && srcrect.right <= src_width);
+  ASSERT(dstrect.top >= 0 && dstrect.bottom <= dst_height);
+  ASSERT(dstrect.left >= 0 && dstrect.right <= dst_width);
+
+  int copy_width = dstrect.GetWidth();
+  int copy_height = dstrect.GetHeight();
+
   double x_ratio = srcrect.GetWidth() / static_cast<double>(dstrect.GetWidth());
   double y_ratio = srcrect.GetHeight() / static_cast<double>(dstrect.GetHeight());
-  for (int i = 0; i < dstrect.GetHeight(); i++)
+
+  for (int y_off = 0; y_off < copy_height; y_off++)
   {
-    for (int j = 0; j < dstrect.GetWidth(); j++)
+    for (int x_off = 0; x_off < copy_width; x_off++)
     {
-      int destination_x = j + dstrect.left;
-      int destination_y = i + dstrect.top;
-      int destination_offset = (destination_y * dstrect.GetWidth()) + destination_x;
+      int dst_x = dstrect.left + x_off;
+      int dst_y = dstrect.top + y_off;
+      int dst_offset = (dst_y * dst_width) + dst_x;
 
-      double src_x = std::round(destination_x * x_ratio) + srcrect.left;
-      double src_y = std::round(destination_y * y_ratio) + srcrect.top;
-      int src_offset = static_cast<int>((src_y * srcrect.GetWidth()) + src_x);
+      int src_x = srcrect.left + static_cast<int>(std::round(x_off * x_ratio));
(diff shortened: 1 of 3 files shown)


Pokechu22:
Hmm, per [fifoci](https://fifo.ci/version/b4b2a12974da96a684755ba473a67c496f28ab16/), my changes have shifted a lot of things in addition to fixing Rogue Squadron 2.  Some changed horizontally (e.g. [simpsons-game](https://fifo.ci/compare/5768756-5767361/)) and some changed vertically ([sms-gc](https://fifo.ci/compare/5768741-5767317/)).  As for RS2, [rs2-glass](https://fifo.ci/result/5768714/) looks incorrectly centered compared to [on OGL](https://fifo.ci/result/5768686/) (probably the same issue), and the HUD looks incorrectly colored and is missing blue dots in the minimap for [rs2-skybox](https://fifo.ci/result/5768707/) ([OGL](https://fifo.ci/result/5768699/)) and [rs2-bumpmapping](https://fifo.ci/result/5768769/) ([OGL](https://fifo.ci/result/5768618/)).  [rs2-zfreeze](https://fifo.ci/compare/5768740-5767323/) shows missing triangles, which is probably the same issue as rs3-bumpmapping (https://github.com/dolphin-emu/dolphin/pull/9315#issuecomment-740304787).  I'm also not sure what's up with the messed up first frames (not a new issue, and an software renderer only thing).

I did notice that there is a case for textures that uses a nonzero top and left, too, which might explain some of the shifts (I'm not entirely sure what all this code does or how it's supposed to work, though):

https://github.com/dolphin-emu/dolphin/blob/b4b2a12974da96a684755ba473a67c496f28ab16/Source/Core/VideoCommon/TextureCacheBase.cpp#L1942-L1956

Source/Core/VideoBackends/Software/CopyRegion.h
```diff
@@ -1,32 +1,29 @@
 #pragma once
 
-#include "Common/MathUtil.h"
-
 #include <cmath>
+#include <vector>
```
lioncash, line 4: This include doesn't seem necessary

Pokechu22: Yep, it's a leftover from testing.  Though it's worth noting that both uses of `CopyRegion` start with vectors and use `data()` for it, so there isn't any strong reason to use raw pointers here (apart from performance maybe, but it's the software renderer...)

lioncash: It was likely to keep it from needing to depend on a single container type, making it more flexible, since it can copy to any contiguous region of memory.

Pokechu22: Similarly... why does `ScaleTexture` make copies into and out of vectors?  Wouldn't something like

```cpp
  const SWTexture* software_source_texture = static_cast<const SWTexture*>(src_texture);
  SWTexture* software_dest_texture = static_cast<SWTexture*>(dst_framebuffer->GetColorAttachment());

  CopyRegion(software_source_texture->GetData(), src_rect, software_dest_texture->GetData(), dst_rect);
```

work?  (Ignoring issues with the rectangles themselves).

Pokechu22: Oh, it's because `GetData()` returns a pointer to `u8` while this is supposed to operate on a pointer to `Pixel` (or something else) with a larger size.  It still seems awkward to make a copy; is that *actually* required?

lioncash: degasus, Stenzek, or some other graphic dev would be way more qualified to answer the *why* here than me. But at a glance, I agree.

Pokechu22:
Hmm, interesting.  The shifting issue I had before was due to previous use of `std::round` in `CopyRegion`, which I had removed.  I guess it makes sense (you'd want to round to the nearest pixel, not the previous pixel).  With that, a modified version of `CopyRegion` to not go out of bounds [works fine](https://fifo.ci/version/5126b2ec39d9bded2640e2e9122ffa974909860e/) and only has changes for the (previously broken) RS2 tests.  Oddly, that happens regardless of whether `CopyRegion` respects `top` and `left` or not; ignoring them [produces no changes](https://fifo.ci/version/a2133bb078970249b0cd89ec89f8e96fd5c9730d/) from not ignoring them.  Perhaps none of the test cases actually hit `dst_x`/`dst_y` being nonzero.  It's probably better to still support regions other than the full buffer in `CopyRegion` though, so I'll keep it.

iwubcode (approved):
Whoops, sorry about that.  LGTM!

PR #9309, opened 2020-12-04
NetBSD support proposal.

devnexen:

PR #9323, opened 2020-12-12
Rehabilitate Haiku support.

waddlesplash:
Much less invasive than the equivalent changes that were removed; it looks like things are a lot simpler than they were a few years ago when I last attempted this.

Unlike that attempt, this one is indeed able to boot games (after fixing a bug in our Qt port): http://0x0.st/iFif.png

### DIFF
--- a/Source/Core/Common/GL/GLInterface/BGL.cpp
+++ b/Source/Core/Common/GL/GLInterface/BGL.cpp
@@ -1,0 +1,96 @@
+// Copyright 2017 Dolphin Emulator Project
+// Licensed under GPLv2+
+// Refer to the license.txt file included.
+
+#include "Common/GL/GLInterface/BGL.h"
+
+#include <GLView.h>
+#include <Size.h>
+#include <Window.h>
+
+#include "Common/Assert.h"
+
+BGLView* GLContextBGL::s_current = nullptr;
+
+GLContextBGL::~GLContextBGL()
+{
+  if (!m_window)
+    delete m_gl;
+}
+
+bool GLContextBGL::Initialize(const WindowSystemInfo& wsi, bool stereo, bool core)
+{
+  m_window = static_cast<BWindow*>(wsi.render_window);
+
+  m_gl = new BGLView(m_window ? m_window->Bounds() : BRect(), "GLContextBGL", B_FOLLOW_ALL_SIDES, 0,
+                     BGL_RGB | BGL_DOUBLE | BGL_ALPHA);
+  if (m_window)
+    m_window->AddChild(m_gl);
+
+  m_opengl_mode = Mode::OpenGL;
+
+  m_gl->LockLooper();
+  BRect size = m_gl->Frame();
+  m_gl->UnlockLooper();
+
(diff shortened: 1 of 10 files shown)


Source/Core/Common/GL/GLInterface/BGL.cpp
```diff
@@ -0,0 +1,93 @@
...
+  m_current = m_gl;
+  return true;
+}
+
+bool GLContextBGL::ClearCurrent()
+{
+  if (!m_current)
+    return true;
+
+  //assert(m_gl == m_current);
```
leoetlino, line 61: This should either be uncommented or removed entirely.

leoetlino (changes requested):

Source/Core/Common/GL/GLInterface/BGL.cpp
waddlesplash: I commented it because I was not sure what assert facilities you generally use, i.e. if I should use something besides `<cassert>` (I did not immediately see anything.) Should I just include the header and uncomment it, then?

leoetlino: `#include "Common/Assert.h"` and ASSERT(). (`ASSERT(m_gl == m_current);` for example)

waddlesplash:
Ha, seems I now have conflicts to resolve, thanks to #9309.

waddlesplash:
Should be all fixed.

Source/Core/Common/GL/GLInterface/BGL.h
```diff
@@ -0,0 +1,36 @@
...
+  void* GetFuncAddress(const std::string& name) override;
+
+protected:
+  bool Initialize(const WindowSystemInfo& wsi, bool stereo, bool core) override;
+
+private:
+  BWindow* m_window;
+  BGLView* m_gl;
+
+  static BGLView* m_current;
```
leoetlino, line 35: Missed this last time but s_ should be used for static variables (including static member variables):
```suggestion
  static BGLView* s_current;
```
Also, it should be ordered before the non-static members as per the [style guide](https://github.com/dolphin-emu/dolphin/blob/master/Contributing.md).

Source/Core/Common/GL/GLInterface/BGL.cpp
```diff
@@ -0,0 +1,96 @@
...
+  m_backbuffer_height = size.IntegerHeight();
+
+  MakeCurrent();
+
+  return true;
+}
+
+bool GLContextBGL::IsHeadless() const
+{
+  return !m_window;
```
leoetlino, line 46: In this particular case, it might be clearer to write the check like this
```suggestion
  return m_window == nullptr;
```
The combination of the ! and the "is ..." function name makes it look like you're inverting a boolean.

leoetlino (commented):
Don't forget to squash the commits :)

waddlesplash:
Done.

leoetlino (approved):
Changes look fine and this doesn't require that many changes, so why not :man_shrugging:

waddlesplash:
Thanks!

PR #9324, opened 2020-12-13
DolphinQt: Assign subsequent devices on profile load for ports 2,3,4.

jordan-woyak:
A user suggested this to me.

Now when you load a profile (e.g. our stock "Wii Remote with MotionPlus Pointing") on Wii Remote 2, if a 2nd Wii Remote is connected it will automatically be used instead of requiring the user manually select it after loading the profile.

I can see how this might make things weird in certain situations, though.
If a user has GCPad 1 configured with keyboard controls then loads an xbox controller profile on GCPad 2 while 2 controllers are attached it will automatically select the second controller.

This could potentially be smarter by looking at the previous ports for the same device name, but I think this situation is unusual and users can still manually select different devices when it happens.

Let me know what you think.

### DIFF
--- a/Source/Core/DolphinQt/Config/Mapping/MappingWindow.cpp
+++ b/Source/Core/DolphinQt/Config/Mapping/MappingWindow.cpp
@@ -273,6 +273,17 @@ void MappingWindow::OnLoadProfilePressed()
   ini.Load(profile_path.toStdString());
 
   m_controller->LoadConfig(ini.GetOrCreateSection("Profile"));
+
+  // If additional devices with the same name are connected, use them for ports 2,3,4.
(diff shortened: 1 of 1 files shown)


Techjar:
I think I have a better idea. What if we make some special "internal" device names to be used by the stock profiles that will resolve to the corresponding port when loaded? This way, we can accomplish what we're trying to do here, while not interfering with normal user-created profiles.

jordan-woyak:
@Techjar I think user defined profiles would benefit from this too though. Creating an Xbox controller profile and loading it for multiple controllers is probably a common situation.

jordan-woyak:
This PR is dead.

PR #9326, opened 2020-12-13
InputCommon: Fix callback dispatch deadlock

oldmud0:
Fix for [issue 12253](https://bugs.dolphin-emu.org/issues/12253).

Make sure `m_is_populating_devices` is true when a `WM_INPUT_DEVICE_CHANGE` event is received directly on the ciface thread, so that callbacks aren't (improperly) dispatched while in the process of removing devices. This fixes a race condition between the ciface thread and the CPU thread that results in a deadlock when disconnecting an emulated Wii remote.

### DIFF
--- a/Source/Core/InputCommon/ControllerInterface/ControllerInterface.cpp
+++ b/Source/Core/InputCommon/ControllerInterface/ControllerInterface.cpp
@@ -140,6 +140,19 @@ void ControllerInterface::RefreshDevices()
   InvokeDevicesChangedCallbacks();
 }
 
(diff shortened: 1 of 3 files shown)


jordan-woyak:
Nice. I wonder if this affects https://bugs.dolphin-emu.org/issues/11702 at all.

jordan-woyak (approved):

Filoppi:
> Nice. I wonder if this affects https://bugs.dolphin-emu.org/issues/11702 at all.

No it didn't fix bug 11702. Still happening.

PR #9327, opened 2020-12-13
VideoBackend: Prefer Vulkan on supported platforms

oldmud0:
PR mostly to spur some discussion on [issue 12083](https://bugs.dolphin-emu.org/issues/12083).

There's definitely a tradeoff to be made here between performance and stability. Vulkan is an API and makes no assumptions about the capabilities of the underlying hardware. If we can query whether or not the device has the capabilities to reasonably support the Vulkan backend, then it's possible to be confident about giving a significant performance boost by default on machines that are capable of it, with similar accuracy and little compatibility downside, and then fall back to OpenGL for those devices that do not support it.

(On Windows, there might be a strong argument made for Direct3D 12 being the default backend - though this of course is not supported on Linux.)

### DIFF
--- a/Source/Core/VideoBackends/D3D/VideoBackend.h
+++ b/Source/Core/VideoBackends/D3D/VideoBackend.h
@@ -19,7 +19,7 @@ public:
   std::string GetDisplayName() const override;
   std::optional<std::string> GetWarningMessage() const override;
 
-  void InitBackendInfo() override;
+  bool InitBackendInfo() override;
 
   static constexpr const char* NAME = "D3D";
 
--- a/Source/Core/VideoBackends/D3D/main.cpp
+++ b/Source/Core/VideoBackends/D3D/main.cpp
@@ -62,13 +62,14 @@ std::optional<std::string> VideoBackend::GetWarningMessage() const
   return result;
 }
 
-void VideoBackend::InitBackendInfo()
+bool VideoBackend::InitBackendInfo()
 {
   if (!D3DCommon::LoadLibraries())
-    return;
+    return false;
 
   FillBackendInfo();
   D3DCommon::UnloadLibraries();
+  return true;
 }
 
 void VideoBackend::FillBackendInfo()
--- a/Source/Core/VideoBackends/D3D12/VideoBackend.cpp
+++ b/Source/Core/VideoBackends/D3D12/VideoBackend.cpp
@@ -35,13 +35,14 @@ std::string VideoBackend::GetDisplayName() const
   return "Direct3D 12";
 }
 
-void VideoBackend::InitBackendInfo()
+bool VideoBackend::InitBackendInfo()
 {
   if (!D3DCommon::LoadLibraries())
-    return;
+    return false;
 
   FillBackendInfo();
   D3DCommon::UnloadLibraries();
+  return true;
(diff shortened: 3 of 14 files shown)


Degerz:
It's probably not a bad idea to make Vulkan the default backend since most hardware vendors don't plan on maintaining their OpenGL drivers in the long-term ... 

On Android, it might make more sense to keep OpenGL ES as the default. Windows on ARM only supports OpenGL 3.3 currently which is well below the required version Dolphin needs and this leaves out Intel users too on Windows who are on Broadwell and older so making D3D12 the default would be helpful in those cases ...

WamWooWam:
Is there any major advantage to D3D12 over D3D11? For Windows it's probably more compatible (i.e. versions pre-10, older GPUs, etc.)

RinMaru:
> Is there any major advantage to D3D12 over D3D11? For Windows it's probably more compatible to go with 11 (i.e. versions pre-10, older GPUs, etc.)

not really because 12 is still broken in some games

Miksel12:
I have seen many reports on issues which are fixed by not using OpenGL (https://www.reddit.com/r/DolphinEmulator/comments/m9gu6k/glitched_text_and_random_flickering_across_all/, https://www.reddit.com/r/DolphinEmulator/comments/m8g23q/dolphin_emulator_lag_on_2020_macbook_pro_can/ to name a few) so I think having it not be the default will be more user friendly.
Though  the order should probably be  D3D11 > D3D12 > Vulkan > OGL > SW > Null

Techjar:
On NVIDIA drivers, Vulkan seems to have stuttering issues that are out of our control, and bounding box is ridiculously slow on D3D11/D3D12. On the other hand, OpenGL is quite slow on AMD drivers. In the mobile device department, compatibility is all over the place (though gradually improving). So really I don't think there is a best choice of default backend, it very much depends on what hardware you have and what games you're playing.

All that said, we should probably switch the macOS default to Vulkan because OpenGL is on the chopping block.

Miksel12:
Are you sure those issues still apply? I use Vulkan most of the time with my GTX 1060 and I have never encountered any stuttering except for shader stuttering which were gone after a second playthrough. And I just tested Super Paper Mario with D3D11 and Vulkan and in bounding box heavy moments D3D11 actually seems to outperform Vulkan at 4K 2xSSAA.

Techjar:
Try running Metroid Prime Trilogy and you'll see it. In that game, I experience occasional brief stutters on Vulkan that do not occur in any other backend. My understanding is it has something to do with shader pipelines being generated by the driver that we can't detect or cache ourselves.

Your experience with bounding box in D3D seems contradictory to both my experience and the information on the wiki, so that I can't explain.

Miksel12:
Hmm, yeah I don't have any very shader intensive games so that could explain that.

stenzek:
-1 to this, some devices (phones and hybrid GPU setups) have broken Vulkan drivers which crash on startup, which would make Dolphin completely unusable without editing the ini. GL is a reasonable default IMHO.

leoetlino:
We can revisit this later if the situation on mobile devices changes.

(Worth noting that some time after this PR was opened, #9962 made Vulkan the default backend on macOS.)

PR #9331, opened 2020-12-16
Qt: Fix deprecated use of MidButton

leoetlino:
MidButton has been deprecated since Qt 4.7. The replacement is
MiddleButton.


BhaaLseN (approved):
Seems like a no-brainer. I'm assuming it is just a drop-in and should work the same way (untested).

leoetlino:
Yeah, MidButton is defined as

```cpp
MidButton Q_DECL_ENUMERATOR_DEPRECATED_X("MidButton is deprecated. Use MiddleButton instead") = MiddleButton
```

lioncash (approved):

PR #9332, opened 2020-12-16
Core: Fix various warnings

leoetlino:


Source/Core/Common/Image.h
```diff
@@ -22,4 +22,9 @@ enum class ImageByteFormat
 
 bool SavePNG(const std::string& path, const u8* input, ImageByteFormat format, u32 width,
              u32 height, int stride = 0);
+bool ConvertRGBAToRGBAndSavePNG(const std::string& path, const u8* input, u32 width, u32 height,
+                                int stride = 0);
+
+std::vector<u8> RGBAToRGB(const u8* input, u32 width, u32 height, int row_stride = 0);
```
BhaaLseN, line 28: I don't see this being called outside of `Image.cpp` (at least not in this diff), but it probably doesn't matter if we leave it for others (future) use.

BhaaLseN (approved):
Changes seem reasonable (and the remark is ignorable), untested tho.

Source/Core/Common/Image.h
leoetlino: Yep, the idea was that it could come in handy in the future.

lioncash (approved):

PR #6728, opened 2018-05-02
DebugInterface: Breakpoints methods added

sepalani:
This PR depends on PR #6341 and adds breakpoints methods to the debug interface. It also allows to disable breakpoints on DolphinWx by right-clicking on them and through the Breakpoint view. ~*I'm also porting it to Qt, just haven't finished it yet*.~

Ready to reviewed & tested.

PR #9334, opened 2020-12-17
Debugger: Conditional breakpoints

smurf3tte:
Conditional breakpoints! Like breakpoints, but they only work some of the time! Er...

**Note:** I am seeking feedback on the design and implementation of this feature. Don't be shy!

### Who is this for
- People investigating emulation bugs
- People reverse engineering games
- ...
- Me 😀

### What does it do
Allow a C-style expression to be attached to breakpoints. The breakpoint will only trigger a break in execution if the expression evaluates to true.

### Why is this useful
There is often code that executes many times but is only interesting under specific conditions. This gives users a way to find the signal in the noise. It's a pretty common feature in debuggers.

### How does it work
When creating a breakpoint, there is now a text box for an optional condition expression. As an example, in Final Fantasy Crystal Chronicles (GCCE01), setting a breakpoint at address `80017B18` with the following condition will trigger a break right before the game performs a texture copy that will overrun the destination buffer: `r26 <= r0 && ((r5+3)&-4) * ((r6+3)&-4) * 4 > r0` (see #9329  for details).

### Design considerations
- Basis for expression engine
  Currently using [expr](https://github.com/zserge/expr) with a slight modification (replacing float with double, which can hold 32-bit integers at full precision), which is < 1k LOC in a single header and implements most of the standard C operators and supports variable assignment. Also considered [TinyExpr](https://github.com/codeplea/tinyexpr), but it doesn't support bitwise, logical, comparison, or assignment operators. There's also [cparse](https://github.com/cparse/cparse), which seems to handle a superset of what expr does but is ~3x larger.
- Allow mutation of guest state?
  Allowing expressions to modify guest state, such as registers or memory, would be only marginal additional work. Expressions with side effects require extra care from the user, but they seem useful nonetheless.
- Allow user-defined variables?
  Expression local variables e.g. ("x = r3+r4, x > 42") would permit some expressions to be simplified considerably. The bigger questions are whether variables should persist across evaluations of a given expression, and whether they should be shared between expressions. There is a lot of power in a persistent global scope for variables, but they could become frustrating to use without an additional means for inspection. We may want to persist them to save states as well.
- Future debugger enhancements?
  An obvious next step would be to add an immediate window, or REPL, where the user can enter any expression they like for immediate evaluation. This could be extended to support debugger commands, as well. There are other PRs that go further and wire the emulator to a full scripting language (see #9205, #7064). What I am suggesting is something a bit simpler: exposing the existing debugger functionality - currently accessible only to the GUI - through a text interface.

### TODO
- ✔️ UI for condition expression on new breakpoints
- ✔️ Syntax validation at breakpoint creation time
- ✔️ Show condition in breakpoint list UI
- ✔️ Predicate breakpoint hits on result of expression evaluation
- ✔️ Register access (read/write) including GPRs, FPRs, LR, CTR, PC (read-only)
- ✔️ Guest memory read/write functions (read_u32(), write_u32(), etc.)
- ✔️ Convenience casts (u16(), s32(), etc.)
- ✔️ Integrate with breakpoint list save/load
- ✔️ Cache compiled expression for performance
- ✔️ Accept hexadecimal literals

### Cut/postponed
- Support for conditional memory watches
- Edit button for breakpoints
- Watch widget integration
- Immediate widget for expression evaluation
- Persistent/global expression variables

### DIFF
--- a/Source/Core/Core/PowerPC/BreakPoints.cpp
+++ b/Source/Core/Core/PowerPC/BreakPoints.cpp
@@ -126,9 +126,16 @@ BreakPoints::TBreakPointsStr BreakPoints::GetStrings() const
     if (!bp.is_temporary)
     {
       std::ostringstream ss;
-      ss << std::hex << bp.address << " " << (bp.is_enabled ? "n" : "")
-         << (bp.log_on_hit ? "l" : "") << (bp.break_on_hit ? "b" : "");
-      bp_strings.push_back(ss.str());
+      ss << "$" << fmt::format("{:08x}", bp.address) << " ";
+      if (bp.is_enabled)
+        ss << "n";
+      if (bp.log_on_hit)
+        ss << "l";
+      if (bp.break_on_hit)
+        ss << "b";
+      if (!bp.condition.empty())
+        ss << "c " << bp.condition;
+      bp_strings.emplace_back(ss.str());
     }
   }
 
@@ -142,10 +149,19 @@ void BreakPoints::AddFromStrings(const TBreakPointsStr& bp_strings)
     TBreakPoint bp;
     std::stringstream ss;
     ss << std::hex << bp_string;
+    if (ss.peek() == '$')
+      ss.ignore();
     ss >> bp.address;
-    bp.is_enabled = bp_string.find('n') != bp_string.npos;
-    bp.log_on_hit = bp_string.find('l') != bp_string.npos;
-    bp.break_on_hit = bp_string.find('b') != bp_string.npos;
+    std::string flags;
+    ss >> flags;
+    bp.is_enabled = flags.find('n') != std::string::npos;
+    bp.log_on_hit = flags.find('l') != std::string::npos;
+    bp.break_on_hit = flags.find('b') != std::string::npos;
+    if (flags.find('c') != std::string::npos)
+    {
+      ss >> std::ws;
+      std::getline(ss, bp.condition);
+    }
     bp.is_temporary = false;
     Add(bp);
   }
--- a/Source/Core/Core/PowerPC/BreakPoints.cpp
+++ b/Source/Core/Core/PowerPC/BreakPoints.cpp
@@ -5,182 +5,18 @@
 #include "Core/PowerPC/BreakPoints.h"
 
 #include <algorithm>
-#include <cctype>
 #include <cstddef>
-#include <cstdlib>
-#include <cstring>
-#include <memory>
-#include <optional>
 #include <sstream>
 #include <string>
-#include <utility>
 #include <vector>
 
-#include <expr.h>
-
-#include "Common/BitUtils.h"
 #include "Common/CommonTypes.h"
 #include "Common/DebugInterface.h"
 #include "Common/Logging/Log.h"
 #include "Core/Core.h"
+#include "Core/PowerPC/Expression.h"
 #include "Core/PowerPC/JitInterface.h"
 #include "Core/PowerPC/MMU.h"
-#include "Core/PowerPC/PowerPC.h"
-
-struct ExprDeleter
-{
-  void operator()(expr* expression) const { expr_destroy(expression, nullptr); }
-};
-
-using ExprPointer = std::unique_ptr<expr, ExprDeleter>;
-
-class ExprVarList
-{
-public:
-  ExprVarList() = default;
-  ExprVarList(const ExprVarList&) = delete;
-  ExprVarList& operator=(const ExprVarList&) = delete;
-
-  ExprVarList(ExprVarList&& other) noexcept : m_vars{std::exchange(other.m_vars, {})} {}
-
-  ExprVarList& operator=(ExprVarList&& other) noexcept
-  {
-    std::swap(m_vars, other.m_vars);
-    return *this;
-  }
-
-  ~ExprVarList() { expr_destroy(nullptr, &m_vars); }
-
-  expr_var* GetHead() { return m_vars.head; }
-  expr_var_list* GetAddress() { return &m_vars; }
-
-private:
-  expr_var_list m_vars = {};
-};
-
-static std::optional<int> ParseGPR(const char* name)
-{
-  if (std::strlen(name) >= 2 && name[0] == 'r' && std::isdigit(name[1]))
-  {
-    char* end = nullptr;
-    int index = std::strtol(&name[1], &end, 10);
-    if (index < 32 && *end == '\0')
-      return index;
-  }
-  return {};
-}
-
-template <typename T>
-static T HostRead(u32 address);
-
-template <typename T>
-static void HostWrite(T var, u32 address);
-
-template <>
-u8 HostRead(u32 address)
-{
-  return PowerPC::HostRead_U8(address);
-}
-
-template <>
-u16 HostRead(u32 address)
-{
-  return PowerPC::HostRead_U16(address);
-}
-
-template <>
-u32 HostRead(u32 address)
-{
-  return PowerPC::HostRead_U32(address);
-}
-
-template <>
-u64 HostRead(u32 address)
-{
-  return PowerPC::HostRead_U64(address);
-}
-
-template <>
-void HostWrite(u8 var, u32 address)
-{
-  PowerPC::HostWrite_U8(var, address);
-}
-
-template <>
-void HostWrite(u16 var, u32 address)
-{
-  PowerPC::HostWrite_U16(var, address);
-}
-
-template <>
-void HostWrite(u32 var, u32 address)
-{
-  PowerPC::HostWrite_U32(var, address);
-}
-
-template <>
-void HostWrite(u64 var, u32 address)
-{
-  PowerPC::HostWrite_U64(var, address);
-}
-
-template <typename T, typename U = T>
-double HostReadFunc(expr_func* f, vec_expr_t* args, void* c)
-{
-  if (vec_len(args) != 1)
-    return 0;
-  u32 address = static_cast<u32>(expr_eval(&vec_nth(args, 0)));
-  return Common::BitCast<T>(HostRead<U>(address));
-}
-
-template <typename T, typename U = T>
-double HostWriteFunc(expr_func* f, vec_expr_t* args, void* c)
-{
-  if (vec_len(args) != 2)
-    return 0;
-  T var = static_cast<T>(expr_eval(&vec_nth(args, 0)));
-  u32 address = static_cast<u32>(expr_eval(&vec_nth(args, 1)));
-  HostWrite<U>(Common::BitCast<U>(var), address);
-  return var;
-}
-
-static expr_func g_expr_funcs[] = {{"read_u8", HostReadFunc<u8>},
-                                   {"read_s8", HostReadFunc<s8, u8>},
-                                   {"read_u16", HostReadFunc<u16>},
-                                   {"read_s16", HostReadFunc<s16, u16>},
-                                   {"read_u32", HostReadFunc<u32>},
-                                   {"read_s32", HostReadFunc<s32, u32>},
-                                   {"read_f32", HostReadFunc<float, u32>},
-                                   {"read_f64", HostReadFunc<double, u64>},
-                                   {"write_u8", HostWriteFunc<u8>},
-                                   {"write_u16", HostWriteFunc<u16>},
-                                   {"write_u32", HostWriteFunc<u32>},
-                                   {"write_f32", HostWriteFunc<float, u32>},
-                                   {"write_f64", HostWriteFunc<double, u64>},
-                                   {}};
-
-static double EvaluateExpression(const std::string& expression_string)
-{
-  ExprVarList vars;
-  ExprPointer expression{expr_create(expression_string.c_str(), expression_string.length(),
-                                     vars.GetAddress(), g_expr_funcs)};
-  if (!expression)
-    return false;
-
-  for (auto* v = vars.GetHead(); v != nullptr; v = v->next)
-  {
-    auto index = ParseGPR(v->name);
-    if (index)
-      v->value = GPR(*index);
-  }
-
-  return expr_eval(expression.get());
-}
-
-static bool EvaluateCondition(const std::string& condition)
-{
-  return condition.empty() || EvaluateExpression(condition) != 0;
-}
 
 bool BreakPoints::IsAddressBreakPoint(u32 address) const
 {
@@ -223,8 +59,8 @@ BreakPoints::TBreakPointsStr BreakPoints::GetStrings() const
         ss << "l";
       if (bp.break_on_hit)
         ss << "b";
-      if (!bp.condition.empty())
-        ss << "c " << bp.condition;
+      if (bp.condition)
+        ss << "c " << bp.condition->GetText();
       bp_strings.emplace_back(ss.str());
     }
   }
@@ -250,21 +86,23 @@ void BreakPoints::AddFromStrings(const TBreakPointsStr& bp_strings)
     if (flags.find('c') != std::string::npos)
     {
       ss >> std::ws;
-      std::getline(ss, bp.condition);
+      std::string condition;
+      std::getline(ss, condition);
+      bp.condition = Expression::TryParse(condition);
     }
     bp.is_temporary = false;
-    Add(bp);
+    Add(std::move(bp));
   }
 }
 
-void BreakPoints::Add(const TBreakPoint& bp)
+void BreakPoints::Add(TBreakPoint bp)
 {
   if (IsAddressBreakPoint(bp.address))
     return;
 
-  m_breakpoints.push_back(bp);
-
   JitInterface::InvalidateICache(bp.address, 4, true);
+
+  m_breakpoints.emplace_back(std::move(bp));
 }
 
 void BreakPoints::Add(u32 address, bool temp)
@@ -273,7 +111,7 @@ void BreakPoints::Add(u32 address, bool temp)
 }
 
 void BreakPoints::Add(u32 address, bool temp, bool break_on_hit, bool log_on_hit,
-                      std::string condition)
+                      std::optional<Expression> condition)
 {
   // Only add new addresses
   if (IsAddressBreakPoint(address))
@@ -287,7 +125,7 @@ void BreakPoints::Add(u32 address, bool temp, bool break_on_hit, bool log_on_hit
   bp.address = address;
   bp.condition = std::move(condition);
 
-  m_breakpoints.push_back(bp);
+  m_breakpoints.emplace_back(std::move(bp));
 
   JitInterface::InvalidateICache(address, 4, true);
 }
--- a/Source/Core/Core/PowerPC/BreakPoints.cpp
+++ b/Source/Core/Core/PowerPC/BreakPoints.cpp
@@ -52,7 +52,7 @@ BreakPoints::TBreakPointsStr BreakPoints::GetStrings() const
     if (!bp.is_temporary)
     {
       std::ostringstream ss;
-      ss << "$" << fmt::format("{:08x}", bp.address) << " ";
+      ss << fmt::format("${:08x} ", bp.address);
       if (bp.is_enabled)
         ss << "n";
       if (bp.log_on_hit)
--- a/Source/Core/Core/PowerPC/BreakPoints.cpp
+++ b/Source/Core/Core/PowerPC/BreakPoints.cpp
@@ -5,10 +5,18 @@
 #include "Core/PowerPC/BreakPoints.h"
 
 #include <algorithm>
+#include <cctype>
 #include <cstddef>
+#include <cstdlib>
+#include <cstring>
+#include <memory>
+#include <optional>
 #include <sstream>
 #include <string>
+#include <utility>
 #include <vector>
+
+#include <expr.h>
 
 #include "Common/CommonTypes.h"
 #include "Common/DebugInterface.h"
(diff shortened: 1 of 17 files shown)


Source/Core/Core/PowerPC/BreakPoints.cpp
```diff
@@ -10,12 +10,57 @@
...
+  static expr_var_list vars = {0};
+  expr* e = expr_create(expression.c_str(), expression.length(), &vars, nullptr);
+  if (!e)
+    return false;
+
+  for (auto* v = vars.head; v != nullptr; v = v->next)
+  {
+    int index = 0;
+    if (ParseGPR(v->name, &index))
+      v->value = GPR(index);
```
phire, line 55: Should probably return false if ParseGPR fails

smurf3tte: This is actually intentional (for now). This allows expressions to introduce their own variables. If that feature is cut, then yes, we would want to fail here.

Source/Core/Core/PowerPC/BreakPoints.cpp
```diff
@@ -10,12 +10,57 @@
...
 #include "Common/Logging/Log.h"
 #include "Core/Core.h"
 #include "Core/PowerPC/JitInterface.h"
 #include "Core/PowerPC/MMU.h"
+#include "Core/PowerPC/PowerPC.h"
+
+static bool ParseGPR(const char* name, int* index)
+{
+  if (name[0] == 'r' && isdigit(name[1]) &&
+      (name[2] == '\0' || isdigit(name[2]) && (name[3] == '\0')))
```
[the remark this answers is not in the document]
smurf3tte: Speed was my principal concern here. Expression evaluation should be fast, especially once compiled expression caching is in place. That said, the variable binding code is likely to change anyway, and in the meantime atoi() would be a win for clarity. I haven't benchmarked it anyway, so this was premature optimization on my part. 😅

Source/Core/Core/PowerPC/BreakPoints.cpp
```diff
@@ -10,12 +10,57 @@
...
+
 #include "Common/CommonTypes.h"
 #include "Common/DebugInterface.h"
 #include "Common/Logging/Log.h"
 #include "Core/Core.h"
 #include "Core/PowerPC/JitInterface.h"
 #include "Core/PowerPC/MMU.h"
+#include "Core/PowerPC/PowerPC.h"
+
+static bool ParseGPR(const char* name, int* index)
```
lioncash, line 23: you may wish to return a `std::optional<int>` here, rather than an out pointer.

Source/Core/Core/PowerPC/BreakPoints.cpp
```diff
@@ -10,12 +10,57 @@
...
+    }
+  }
+
+  *index = 0;
+  return false;
+}
+
+static bool EvaluateExpression(const std::string& expression)
+{
+  static expr_var_list vars = {0};
```
lioncash, line 46: Maybe organize this such that a local static isn't required. This inherently makes the function thread-unsafe.

Source/Core/Core/PowerPC/BreakPoints.cpp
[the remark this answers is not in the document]
smurf3tte: Unfortunately atoi() is unsuitable for other reasons. It returns 0 on error (indistinguishable from success) and it ignores extra characters (so "r0z" would parse the same as "r0"). That said, strtol() should do the trick.

Source/Core/Core/PowerPC/BreakPoints.cpp
```diff
@@ -10,12 +10,57 @@
...
+  expr* e = expr_create(expression.c_str(), expression.length(), &vars, nullptr);
```
lioncash, line 47:
```cpp
struct Deleter {
  void operator()(expr* expression) const {
    expr_destroy(expression, nullptr);
  }
};
using ExprPointer = std::unique_ptr<expr, Deleter>;

ExprPointer e{expr_create(expression.c_str(), expression.length(), &vars, nullptr)};
```

allows wrapping the c-style destruction in a manner that prevents any leaks from occurring if any other exits to the function are added in the future (through refactoring or feature additions otherwise).

Source/Core/Core/PowerPC/BreakPoints.cpp
```diff
@@ -88,10 +134,11 @@ void BreakPoints::Add(const TBreakPoint& bp)
 
 void BreakPoints::Add(u32 address, bool temp)
 {
-  BreakPoints::Add(address, temp, true, false);
+  BreakPoints::Add(address, temp, true, false, {});
 }
 
-void BreakPoints::Add(u32 address, bool temp, bool break_on_hit, bool log_on_hit)
+void BreakPoints::Add(u32 address, bool temp, bool break_on_hit, bool log_on_hit,
+                      const std::string& condition)
```
lioncash, line 141:
```suggestion
                      std::string condition)
```

Source/Core/Core/PowerPC/BreakPoints.cpp
```diff
@@ -103,6 +150,7 @@ void BreakPoints::Add(u32 address, bool temp, bool break_on_hit, bool log_on_hit
   bp.break_on_hit = break_on_hit;
   bp.log_on_hit = log_on_hit;
   bp.address = address;
+  bp.condition = condition;
```
lioncash, line 153:
```suggestion
  bp.condition = std::move(condition);
```

Source/Core/Core/PowerPC/BreakPoints.cpp
smurf3tte: I don't intend to leave it this way. That said, code like this has been thread safe on conforming compilers since C++11, thanks to [magic statics](https://blog.mbedded.ninja/programming/languages/c-plus-plus/magic-statics/).

Source/Core/Core/PowerPC/BreakPoints.cpp
smurf3tte: I don't think that will do anything different for a const reference. That said, this function could accept a string by value.

smurf3tte: I clearly read your comments out of order.

Source/Core/Core/PowerPC/BreakPoints.cpp
lioncash: That only applies for *initialization* of the local static. It does not mean data races do not matter in relation to the local static. If you call this in two different threads, you will clobber the data within the variable list (which is a race, and by extension, undefined behavior).

smurf3tte: Ah, you are correct, of course. I assumed your comment was about initialization since that has historically been an issue with statics in C++.

The scope and lifetime of expression variables is still up in the air. In the current implementation they all go in persistent global scope that is shared between all expressions. I believe they are only accessed from the CPU thread at the moment, but if they start to become exposed in various places in the UI, we'll definitely need a synchronization primitive to ensure thread safety.

smurf3tte: That said, I don't see any much of anything in the way of synchronization in this area, today. PowerPC::breakpoints is accessed freely from the UI thread while the game is running.... so that's not great. 😅

Source/Core/Core/PowerPC/BreakPoints.cpp
smurf3tte: I was going to look up unique_ptr deleters (again) but you spared me the trouble. Thanks!

iwubcode:
> Basis for expression engine

It would likely be a bigger task but inputs already uses an expression parser.  It'd be nice in general if we could split that out and leverage it in other places (such as this).  I know several times I've thought about a feature that required some complex expression parsing and shied away from it.

Rumi-Larry:
I'm guessing that this pr makes https://github.com/dolphin-emu/dolphin/pull/6728/files completely obsolete, doesn't it?

Source/Core/Core/PowerPC/BreakPoints.cpp
leoetlino: Any reason this even needs to be static? The expression is destructed at the end of this function anyway...

Source/Core/Core/PowerPC/BreakPoints.cpp
```diff
@@ -6,16 +6,60 @@
...
 #include "Core/PowerPC/JitInterface.h"
 #include "Core/PowerPC/MMU.h"
+#include "Core/PowerPC/PowerPC.h"
+
+struct ExprDeleter
+{
+  void operator()(expr* expression) const { expr_destroy(expression, nullptr); }
+};
+
+using ExprPointer = std::unique_ptr<expr, ExprDeleter>;
```
leoetlino, line 29: `<memory>` needs to be included as well.

Source/Core/Core/PowerPC/BreakPoints.cpp
```diff
@@ -6,16 +6,60 @@
...
+  ExprPointer expression{
+      expr_create(expression_string.c_str(), expression_string.length(), &vars, nullptr)};
+  if (!expression)
+    return false;
+
+  for (auto* v = vars.head; v != nullptr; v = v->next)
+  {
+    auto index = ParseGPR(v->name);
+    if (index)
+      v->value = GPR(index.value());
```
leoetlino, line 56: We typically use `std::optional<T>::operator*` if the optional has already been checked to see if it contains a value:
 
```suggestion
      v->value = GPR(*index);
```

Source/Core/Core/PowerPC/BreakPoints.cpp
```diff
@@ -6,16 +6,60 @@
...
+struct ExprDeleter
+{
+  void operator()(expr* expression) const { expr_destroy(expression, nullptr); }
+};
+
+using ExprPointer = std::unique_ptr<expr, ExprDeleter>;
+
+static std::optional<int> ParseGPR(const char* name)
+{
+  if (name[0] == 'r' && isdigit(name[1]))
```
BhaaLseN, line 33: Should this perhaps also perform sanity checks, such as "at least two characters long" etc?

Source/Core/Core/PowerPC/BreakPoints.cpp
```diff
@@ -33,7 +77,8 @@ bool BreakPoints::IsTempBreakPoint(u32 address) const
 bool BreakPoints::IsBreakPointBreakOnHit(u32 address) const
 {
   return std::any_of(m_breakpoints.begin(), m_breakpoints.end(), [address](const auto& bp) {
-    return bp.address == address && bp.break_on_hit;
+    return bp.address == address && bp.break_on_hit &&
+           (bp.condition.empty() || EvaluateExpression(bp.condition));
```
BhaaLseN, line 81: I wonder if UX would be better if `EvaluateExpression` specifically handled the `.empty()` case. That way you cannot forget about it if a different place also wants to consider `bp.condition` the same way.
It would mean that an empty expression is always `true` though.

Thoughts?

Source/Core/Core/PowerPC/BreakPoints.cpp
leoetlino: If you add a string length check, it might be worth changing ParseGPR to take a std::string_view so you don't have to call strlen manually.

Source/Core/Core/PowerPC/BreakPoints.cpp
Dentomologist: You could add an EmptyExpressionValue enum class and have EvaluateExpression take one defaulted to true.

Source/Core/Core/PowerPC/BreakPoints.cpp
smurf3tte: Slapping static on the variable list was a hack to create a global scope for expression variables. I believe this idea has utility, but it was not meant to be the final implementation. I should have left a comment in the code communicating this intent.

smurf3tte:
> > Basis for expression engine
> 
> It would likely be a bigger task but inputs already uses an expression parser. It'd be nice in general if we could split that out and leverage it in other places (such as this). I know several times I've thought about a feature that required some complex expression parsing and shied away from it.
> 
> EDIT: we don't have to use the inputs one, I just think it'd be nice to standardize on a single expression parser.

I did notice there was a parser for input mappings, but it seemed fairly purpose-built. Trying to generalize it appeared to be more effort (and would risk breaking the input code) than introducing a general purpose (but small) expression parser. I am open to feedback on this point, though I would prefer not to write a parser from scratch or make sweeping changes to an existing one. 😀

smurf3tte:
> I'm guessing that this pr makes https://github.com/dolphin-emu/dolphin/pull/6728/files completely obsolete, doesn't it?

That PR actually looks like it would still be very useful. It would be nice to be able to disable breakpoints without deleting them. I may dust that one off if the original author has moved on.

Source/Core/Core/PowerPC/BreakPoints.cpp
smurf3tte: A length check is technically redundant, but probably not a bad idea.

smurf3tte: Unfortunately string_view doesn't require null termination, but strtol does. It would be unsafe to accept a string_view and pass the underlying pointer straight to strtol, so I'll leave the parameter as C string for now.

leoetlino: Oh right, strtol... You can use [std::from_chars](https://en.cppreference.com/w/cpp/utility/from_chars) instead, which doesn't require a null terminated string.

Source/Core/Core/PowerPC/BreakPoints.cpp
smurf3tte: I'd like to expose expression evaluation directly to the user at some point, so I think I'll split EvaluateExpression in two, and the second function (EvaluateCondition) will handle the empty string case.

smurf3tte:
I believe I've addressed everything in the first round of feedback. I removed my strawman global variable implementation for now; I may add them back later. In the meantime I'll be starting down the TODO list.

Source/Core/Core/PowerPC/BreakPoints.cpp
```diff
@@ -51,9 +52,16 @@ BreakPoints::TBreakPointsStr BreakPoints::GetStrings() const
     if (!bp.is_temporary)
     {
       std::ostringstream ss;
-      ss << std::hex << bp.address << " " << (bp.is_enabled ? "n" : "")
-         << (bp.log_on_hit ? "l" : "") << (bp.break_on_hit ? "b" : "");
-      bp_strings.push_back(ss.str());
+      ss << "$" << fmt::format("{:08x}", bp.address) << " ";
```
lioncash, line 55:
```suggestion
      ss << fmt::format("${:08x} ", bp.address);
```

Condenses this to a single string insertion

Source/Core/Core/PowerPC/Expression.cpp
```diff
@@ -0,0 +1,229 @@
...
+
+template <typename T, typename U = T>
+static double CastFunc(expr_func* f, vec_expr_t* args, void* c)
+{
+  if (vec_len(args) != 1)
+    return 0;
+  return Common::BitCast<T>(static_cast<U>(expr_eval(&vec_nth(args, 0))));
+}
+
+static expr_func g_expr_funcs[] = {{"read_u8", HostReadFunc<u8>},
```
lioncash, line 102: std::array

Source/Core/Core/PowerPC/Expression.cpp
```diff
@@ -0,0 +1,229 @@
...
+                                   {"write_u32", HostWriteFunc<u32>},
+                                   {"write_f32", HostWriteFunc<float, u32>},
+                                   {"write_f64", HostWriteFunc<double, u64>},
+                                   {"u8", CastFunc<u8>},
+                                   {"s8", CastFunc<s8, u8>},
+                                   {"u16", CastFunc<u16>},
+                                   {"s16", CastFunc<s16, u16>},
+                                   {"u32", CastFunc<u32>},
+                                   {"s32", CastFunc<s32, u32>},
+                                   {}};
```
lioncash, line 121: Place a comma after the `{}` here, so that this formats a little more nicely.

Source/Core/Core/PowerPC/Expression.cpp
```diff
@@ -0,0 +1,229 @@
...
+
+const std::unordered_map<std::string_view, u32&> g_register_vars = {
```
lioncash, line 123: Please consider using a std::array here instead of unordered map to avoid static allocating constructors. This is a small enough data set that linear lookup is more or less fine.

Ditto with g_float_register_vars

Source/Core/Core/PowerPC/Expression.cpp
```diff
@@ -0,0 +1,229 @@
...
+        v->value = static_cast<double>(r->second);
+      }
+      else
+      {
+        r->second = static_cast<u32>(static_cast<s64>(v->value));
+      }
+      continue;
+    }
+
+    auto fr = g_float_register_vars.find(v->name);
```
lioncash, line 168:
```suggestion
    const auto fr = g_float_register_vars.find(v->name);
```

Source/Core/Core/PowerPC/Expression.cpp
```diff
@@ -0,0 +1,229 @@
...
+{
+  FromRegister,
+  ToRegister,
+};
+
+static void SynchronizeRegisters(const expr_var_list& vars, SynchronizeDirection dir)
+{
+  for (auto* v = vars.head; v != nullptr; v = v->next)
+  {
+    auto r = g_register_vars.find(v->name);
```
lioncash, line 154:
```suggestion
    const auto r = g_register_vars.find(v->name);
```

Source/Core/Core/PowerPC/Expression.cpp
```diff
@@ -0,0 +1,229 @@
...
+}
+
+void ExprDeleter::operator()(expr* expression) const
+{
+  expr_destroy(expression, nullptr);
+}
+
+void ExprVarListDeleter::operator()(expr_var_list* vars) const
+{
+  expr_destroy(nullptr, vars);  // free list elements
```
lioncash, line 196: We place documentation comments above the line they apply to rather than off to the side where applicable.

Source/Core/DolphinQt/Debugger/NewBreakpointDialog.cpp
```diff
@@ -165,7 +169,15 @@ void NewBreakpointDialog::accept()
       return;
     }
 
-    m_parent->AddBP(address, false, do_break, do_log);
+    QString condition = m_instruction_condition->text().trimmed();
```
lioncash, line 172:
```suggestion
    const QString condition = m_instruction_condition->text().trimmed();
```

Source/Core/Core/PowerPC/Expression.cpp
smurf3tte: Would population of the map on first use be an acceptable alternative? Is your principal concern app startup time?

lioncash: No.

We generally try to avoid using allocating structures at file scope when they're avoidable. Given this is a small set of entries (< 100) it is fine to use a std::array here and make use of `std::find_if`

smurf3tte: They're gone. Variable binding now occurs at expression compile time, so no string comparisons are necessary at evaluation time.

smurf3tte: I opted to remove them entirely. Variable binding is now done at expression compile time (as opposed to evaluation time) so fast lookup is not such a concern.

sepalani:
> 
> 
> > I'm guessing that this pr makes https://github.com/dolphin-emu/dolphin/pull/6728/files completely obsolete, doesn't it?
> 
> That PR actually looks like it would still be very useful. It would be nice to be able to disable breakpoints without deleting them. I may dust that one off if the original author has moved on.

The initial purpose of this PR is unrelated to what my PR is doing. The goal is to expose debugging features (in that case BPs) though the debugger interface. That interface could be used later for scripting languages such as python/lua. Feel free take over my PR or create your own as this one is really low on my TODO list.

smurf3tte:
For reference, the last force push was to sneak "pc" back in as a built-in variable.

Pokechu22:
This doesn't seem to work correctly with "write to log"; if I create a conditional breakpoint in Super Mario Sunshine at 802f3630 (`J3DGDSetChanCtrl`) with the condition `r3 == 5`, the following gets logged (heavily abbreviated):

```
43:27:735 Core\PowerPC\PowerPC.cpp:615 N[MI]: BP 802f3630  --- (00000000 00000000 00000000 00000001 00000000 00000002 00000001 0000100d 00000000 802d7aa8) LR=802d7b68
43:27:736 Core\PowerPC\PowerPC.cpp:615 N[MI]: BP 802f3630  --- (00000002 00000000 00000000 00000001 00000000 00000002 00000001 00000001 00000001 802d7aa8) LR=802d7b68
43:27:736 Core\PowerPC\PowerPC.cpp:615 N[MI]: BP 802f3630  --- (00000001 00000001 00000000 00000000 00000000 00000000 00000000 00000001 00000001 802d7aa8) LR=802d7b68
43:27:736 Core\PowerPC\PowerPC.cpp:615 N[MI]: BP 802f3630  --- (00000003 00000000 00000000 00000000 00000000 00000000 00000002 00000001 00000000 802d7aa8) LR=802d7b68
```

The code that does this logging is this, and the first value in parentheses is r3, but it's getting logged even though it's clearly not equal to 5:

https://github.com/dolphin-emu/dolphin/blob/c21581a91219b50919b21ea1938d3420fe61dc57/Source/Core/Core/PowerPC/PowerPC.cpp#L609-L616

It doesn't actually break in those cases, but this does make conditional breakpoints unusable for e.g. checking how many times a specific condition is hit per frame (which is something I wanted to do).  Checking the condition in `IsBreakPointLogOnHit` would solve this (though that does mean that the condition is evaluated twice for "Write to Log and Break" breakpoints).

TryTwo:
Will any more work be done on this?

I was thinking BPs with a specific call stack conditional would be really helpful for me.  I don't think doing it as an individual PR would be right, but I imagine it could be added onto this after it gets merged.

JoshuaMKW:
I would love to see this completed and merged. One of my biggest time wasters when debugging with breakpoints is pressing play 100+ times until I stumble across the correct instance of the breakpoint for what I need.

This would completely eliminate that and also make debugging with Dolphin more enticing.

smurf3tte:
Completed with #11015. Thanks to @TryTwo for getting it across the finish line.

PR #9337, opened 2020-12-17
GameSettings: Set Safe Texture Cache to Safe for Pitfall: The Lost Ex…

Avasam:
…pedition and Pitfall: The Big Adventure

1024 is not enough. 2048 works just fine.
Fixes stuttering Atari 2600 emulation https://bugs.dolphin-emu.org/issues/12347

### DIFF
--- a/Data/Sys/GameSettings/RPF.ini
+++ b/Data/Sys/GameSettings/RPF.ini
@@ -1,4 +1,4 @@
-# GPHD52, GPHE52, GPHF52, GPHP52 - Pitfall: The Lost Expedition
+# RPFP52, RPFE52, RPFU52 - Pitfall: The Big Adventure
 
 [Core]
 # Values set here will override the main Dolphin settings.


JMC47:
Because this is more than just a VC game, would you mind trying some other values?

2048 usually will work.  That's between medium and safe.  If that doesn't work, then using 0 means hashing the whole texture and is safer, but is more performance intensive.

Avasam:
@JMC47 2048 works just fine. Re-tested for both TLE and TBA, in-game and using the cheat. No noticeable stutters when moving around and standing still (minimal changes) I can still see the timer tick down.

I went ahead and also tested at 1024. Is not enough and still slightly stutters.

JosJuice:
@dolphin-emu-bot rebuild

JosJuice (approved):

lioncash (approved):

PR #9340, opened 2020-12-19
Common/LinearDiskCache: Handle truncated shadercache files.

AdmiralCurtiss:
Without this you run into two issues:

- Since it breaks out after reading the `value_size` but before reading the actual value, the file position is not aligned on a boundary between values, so it would have no chance to find any newly written values the next time the cachefile is opened.
- ...But it probably doesn't come to that because of a [quirk in files opened in r+ mode](https://stackoverflow.com/questions/3206913/c-standard-library-corner-case). Thankfully Visual Studio's debug mode [catches this](https://i.imgur.com/NgqVNC8.png), I would have never found this otherwise.

I also took the opportunity to replace the manual new/delete with an unique_ptr.

### DIFF
--- a/Source/Core/Common/LinearDiskCache.h
+++ b/Source/Core/Common/LinearDiskCache.h
@@ -76,6 +76,7 @@ public:
       std::unique_ptr<V[]> value = nullptr;
       u32 value_size = 0;
       u32 entry_number = 0;
+      u64 last_valid_value_start = m_file.Tell();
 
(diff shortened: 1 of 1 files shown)


iwubcode (approved):
Untested but this LGTM.  I've been bitten by this issue multiple times now but wasn't sure what the issue was.  I just would delete my shader cache which was a pain.

smurf3tte (approved):

lioncash (approved):

PR #9341, opened 2020-12-20
Common: Write PNGs with png_image_write_to_stdio to allow Unicode target paths.

AdmiralCurtiss:
As an alternative to #9339, as discussed there. Both options have advantages and disadvantages.


AdmiralCurtiss:
Ah fair enough, not needed then.

PR #9343, opened 2020-12-20
Major Refactor: Rename "Movies" to "InputRecorder"

altimumdelta:
Half-way through I realized that this isn't primairly about TAS either, it's really just input recording and replaying, with the added ability feed tool-assisted inputs without having to rely on the capabilities of the original controller and human involvement. What eventually becomes "TAS Videos" is just one of the possible end products made with the use of this Dolphin component.  And ofcourse it has been obvious that this component/feature isn't a "movie" either, before any of this could be used for a movie*, it would first be a video after dumping frames or recording ehm a video of the recorded input track data with OBS or similar.

In it self, perhaps it's not the most important thing, but since there was some hints of some people suggesting upgrading TAS, and the other temptations to upgrade FrameDump, I thought to clear the way of possible confusions/conflicts before that starts. 
Decoupling of Frame and Audio Dump from Input Recorder will come after this.

- Input Recorder is now located as a submenu in Tools menu
- `Export Recording` was renamed to `Save Recording` because the meaning behind "export" revolves around producing a type of file mainly meant for other programs, while the `.dtm` format to my (limited) knowledge is mainly meant for use by Dolphin. There is a comment that does mention other programs are likely to use the format too, but judging by the fact that it is a dolphin-original format that never existed before, I think the reasons outweight in favor of saving rather than exporting.
- Where applicable and makes sense, in the code or GUI, "input recording data" is referred to as a `track`, I found that to be fitting because first it avoids the repetitive use of the long `recording` term, second it avoids conflicting with the actions such as `IsRecording` or `StopRecording` and third it kinda auto-implies that it is a file and a previous recording of the inputs (IMO). PC games have largely used the term "replay" but i've seen `.track` too (happens to be something I'm involved in a lot in the recent years).

Initial preview:
Screenshot_20201220_195609_Dolphin-Emu_Preview_Refactor-Rename-Movies-InputRecorder_MenuBar_V1

- [  ] Todo: Negotiate what happens to `.dtm` file extension
- [  ] Todo: Negotiate what happens with config/INI keys/names with `movies` as part of their string.
- [  ] Todo: Negotiate what happens with "movie" entry in Analytics.cpp line 396, whether it affects https://dolphin.ci

Above mentioned unresolved changes are currently temporairly suffixed with a `// INREC-TODO` comment at the end of the line for easier search/grep.

Opinion on file format and extension name: While I expect that `.dtm` file extension would be kept for obvious compatability purposes, but whether or not a new acceptable extension type would be added now or later, I think we should specifically delay introducting a new one (such as `.dit` (Dolphin Input Track, which I thought it would be okay to use this as the new description for the extension already)) until there is an upgrade the format it self, and then the new format with the new magic pattern would use the new extension type `.dit`, with possible backward compatability for `.dtm`, however I would refer to it as the "old Dolphin Input Track format" and avoid any mention of "TAS" or "Movies".

(I think in cases like this it's actually preferrable to be a single commit, no? While it's sizable, the nature of the changes shouldn't produce a diff-move-mess and should therefore still be digestable)

Update 1:
I think I misunderstood the "TAS Input", I guess it's not about config/adjustment but more like a control station that is continiously used while recording, I will remove the "Configuration" part out of the name then, or propose another suffix.


JosJuice:
For the reference, there was a discussion about what to call the Movie menu back when it was added: #1134

Personally I like the term "input recording" a bit more than "movie" for describing a specific input recording (i.e. a specific DTM file). I'm however not sure if it's a great fit for the name of a top-level menu since it's a bit long. (I think moving it to a submenu of Tools like you have done would make things rather annoying for TASers.) I think "input recorder" is worse as a menu name as "input recording".

"Track" feels very confusing and unfamiliar to me. It isn't something I've ever heard a TASer use. Also, I definitely don't think you should change the DTM extension or any config entries, analytics entries, command-line flags or anything similar as it's a very unnecessary break of compatibility. Renaming everything in the code also seems unnecessary to me, but at least there aren't any serious compatibility consequences for doing that.

In short: I'm not in favor of these changes. I guess "Save Recording" is an OK change, though.

altimumdelta:
I was unfortunately not involved around here that far back.

I specifically made this to be first and foremost a description of the component, because that is most optimal in code, with the GUI strings having more  not the primary use of the component, as in that 



Projects such as FFmpeg change command line syntax all the time, users just learn how to deal with it, I had to as well.

altimumdelta:
I have finished reading through #1134 and I was quite interesting (I actually recall seening it 2 years or so back.) which gave me an interesting perspective ... which lead me to mention (if I forgot?) the **biggest point** of this PR:

It's about the code it self, not the GUI labels, the fact that "**Movie**" is used for things like `Movie.cpp` and `Movie::StartRecording` :trollface: and so on, but I kinda understand the historical reasons and the thinking how it became a Movie, this is just normal process, I come back to my own programs a year later and suddenly see so many points of improvement, so I hope that wall of text wasn't inimidating hehe, I don't have any ill will against ... poor Movie, thank you for serving us well for all these years :)

Heh, in a way I wouldn't want to mention right now that I eventually wouldn't insist on the GUI labeling one way or another, but for now I would like to see if preferring to keep GUI consistent will stick.

leoetlino:
I also disagree with these changes for all the reasons that @JosJuice has already mentioned, and in my opinion renaming every occurrence of Movie in the code is kinda pointless. I also strongly disagree with moving the menu to the Tools menu for usability reasons.

For all those reasons, and considering that you intend to keep those changes, this PR is unlikely to be merged so I am closing it.

PR #9347, opened 2020-12-21
Jit64: Fix FPURegCache::GetRegUtilization

JosJuice:
This performance bug was probably a simple copy-paste error. (The function was identical to `GPRRegCache::GetRegUtilization`.)


lioncash (approved):

degasus (approved):
I assume this implementation is still as broken as fprInXmm is not updated correctly. However spilling based on gpr utilization is even worse...
However this is just a hint for which to register to spill, I'm fine with both ways.

JosJuice:
@degasus It took me until now to realize, but the existing `fprInXmm` logic does make sense: `strncmp` returns 0 when the strings *match*, not the other way around.

PR #9348, opened 2020-12-21
DSP: Eliminate most global state

lioncash:
An unfortunately large single commit that deglobalizes the DSP code. (which I'm very sorry about).

This would have otherwise been extremely difficult to separate due to extensive use of the globals in very coupling ways that would result in more scaffolding to work around than is worth it.

Aside from the video code, I believe only the DSP code is the hairiest to deal with in terms of globals, so I guess it's best to get this dealt with right off the bat.

A summary of what this commit does:

  - Turns the DSPInterpreter into its own class
    This is the most involved portion of this change. The bulk of the changes are turning non-member functions into member
    functions that would be situated into the Interpreter class.

  - Eliminates all usages of globals within DSPCore.
    This generally involves turning a lot of non-member functions into member functions that are either situated within SDSP or DSPCore.

  - Discards DSPDebugInterface (it wasn't hooked up to anything, and for the sake of eliminating global state, I'd rather get rid of  it than think up ways for this class to be integrated with everything else.
     This hasn't been used since the migration to Qt.

  - Readjusts the DSP JIT to handle calling out to member functions.
    In most cases, this just means wrapping respective member function calls into thunk functions.

Surprisingly, this doesn't even make use of the introduced System class. It was possible all along to do this without it. We can house everything within the DSPLLE class, which is quite nice =)


Source/Core/Core/State.cpp
```diff
@@ -74,7 +74,7 @@ static Common::Event g_compressAndDumpStateSyncEvent;
 static std::thread g_save_thread;
 
 // Don't forget to increase this after doing changes on the savestate system
-constexpr u32 STATE_VERSION = 124;  // Last changed in PR 9097
+constexpr u32 STATE_VERSION = 125;  // Last changed in PR 9348
```
smurf3tte, line 77: This looks like a result of removing the debug interface. Are there any other functional changes in this refactor?

Source/Core/Core/DSP/Interpreter/DSPIntTables.cpp
```diff
@@ -266,7 +265,7 @@ void InitInstructionTables()
   // ext op table
   for (size_t i = 0; i < s_ext_op_table.size(); i++)
   {
-    s_ext_op_table[i] = nop;
+    s_ext_op_table[i] = &Interpreter::nop_ext;
```
smurf3tte, line 268: It looks like the nop implementation is serving double duty: an actual handler for nop and an unrecognized opcode handler that reports an error message. Interpreter::nop() does both, but nop_ext() only does the former. By switching implementations to nop_ext(), this change eliminates the reporting of unrecognized opcodes.

Defining a dedicated unrecognized opcode handler would retain the intent and behavior of the original code.

Source/Core/Core/DSP/Jit/x64/DSPEmitter.cpp
```diff
@@ -128,6 +133,11 @@ bool DSPEmitter::FlagsNeeded() const
   return !(flags & Analyzer::CODE_START_OF_INST) || (flags & Analyzer::CODE_UPDATE_SR);
 }
 
+static void FallbackThunk(Interpreter::Interpreter& interpreter, UDSPInstruction inst)
+{
+  (interpreter.*Interpreter::GetOp(inst))(inst);
```
smurf3tte, line 138: I assume that deferring this lookup to execution time is a tradeoff for not having to worry about the compiler implementation of pointers to member functions...?

lioncash: Yes

Source/Core/Core/DSP/DSPCore.h
```diff
@@ -338,35 +412,118 @@ struct DSPInitOptions
...
-
 enum class State
 {
   Stopped,
   Running,
   Stepping,
 };
 
-int DSPCore_RunCycles(int cycles);
+class DSPCore
```
iwubcode, line 422: I know a lot of our classes / structs don't do this but since this is a new class, should this follow the rule of 5?

iwubcode:
Still need another pass but this looks really good!

Source/Core/Core/DSP/DSPCore.h
lioncash: Done

Source/Core/Core/DSP/DSPHWInterface.cpp
```diff
@@ -23,349 +21,346 @@
...
 {
   ERROR_LOG_FMT(DSPLLE, "*** idma_out IRAM_DSP ({:#06x}) -> RAM ({:#010x}) : size ({:#010x})",
                 dsp_addr / 2, addr, size);
-
   return nullptr;
 }
 
 // TODO: These should eat clock cycles.
-static const u8* gdsp_ddma_in(u16 dsp_addr, u32 addr, u32 size)
+const u8* SDSP::DDMAIn(u16 dsp_addr, u32 addr, u32 size)
```
iwubcode, line 304: Could some of these in / out functions be `const`?

Source/Core/Core/DSP/DSPCore.h
```diff
@@ -263,56 +278,115 @@ struct DSP_Regs
 // should be moved here.
 struct SDSP
```
iwubcode, line 279: Sorry, missed this originally.  But this could also follow rule of 5...

Source/Core/Core/DSP/DSPHWInterface.cpp
lioncash: Yes, but I'd rather they be left non-const, given if the data DMA ones are made more accurate in the future (by eating cycles), they'd need to become non-const anyways. That and the instruction DMA ones alter the write protections on the iram pointer. I could mark IDMAOut as const, but this would be an inconsistent interface.

Source/Core/Core/DSP/Jit/x64/DSPEmitter.cpp
smurf3tte: Oh, I know. The rest of the thunks should be free, or nearly so. I called this one out because of the added runtime cost from the lookup and the subsequent indirect branch. I understand it would take some extra work to avoid, though. It may not be worth the effort for the DSP JIT, but I was thinking it might be for the PPC JIT (assuming you plan to refactor the rest of Dolphin in this way).

Source/Core/Common/x64Emitter.h
```diff
@@ -1122,6 +1129,18 @@ class XEmitter
...
   }
 
+  // Pass a pointer and register as a parameter.
+  template <typename FunctionPointer>
+  void ABI_CallFunctionPR(FunctionPointer func, const void* ptr, X64Reg reg1)
+  {
+    MOV(64, R(ABI_PARAM1), Imm64(reinterpret_cast<u64>(ptr)));
+
+    if (reg1 != ABI_PARAM2)
+      MOV(32, R(ABI_PARAM2), R(reg1));
```
AdmiralCurtiss, line 1139: What happens if `reg1` == `ABI_PARAM1`? This would overwrite the data in reg1 by the pointer before it gets copied into parameter 2. Can that happen?

Source/Core/Common/x64Emitter.h
```diff
@@ -1130,6 +1149,15 @@ class XEmitter
     ABI_CallFunction(func);
   }
 
+  // Pass a pointer and two registers as parameters.
+  template <typename FunctionPointer>
+  void ABI_CallFunctionPRR(FunctionPointer func, const void* ptr, X64Reg reg1, X64Reg reg2)
+  {
+    MOV(64, R(ABI_PARAM1), Imm64(reinterpret_cast<u64>(ptr)));
+    MOVTwo(64, ABI_PARAM2, reg1, 0, ABI_PARAM3, reg2);
```
AdmiralCurtiss, line 1157: Same thing here.

Source/Core/Common/x64Emitter.h
```diff
@@ -1122,6 +1129,15 @@ class XEmitter
     ABI_CallFunction(func);
   }
 
+  // Pass a pointer and register as a parameter.
+  template <typename FunctionPointer>
+  void ABI_CallFunctionPR(FunctionPointer func, const void* ptr, X64Reg reg1)
+  {
+    MOV(64, R(ABI_PARAM1), Imm64(reinterpret_cast<u64>(ptr)));
+    MOV(32, R(ABI_PARAM2), R(reg1));
```
AdmiralCurtiss, line 1137: I'm still not sure this is right. Consider the following:

ABI_PARAM1 = rcx
ABI_PARAM2 = rdx
reg1 = rcx

which would result in

(outside this function)
mov ecx, (parameter value)
(inside this function)
mov rcx, (function address)
mov edx, ecx

Clearly not what we want! Unless I'm grossly misunderstanding something here? If you just reverse the order of the two instructions it would work, though.

lioncash:
Are there any more reviews to be done? Last review was 6 days ago

BhaaLseN (approved):
Some files were difficult to review because of method reordering (which individually are just renames/relocations to match the new structure), so I kinda avoided this PR in the past days, but here we go. Code LGTM, untested. But I think we'll get some quick feedback once this is out in the wild in case something is really off.

leoetlino (approved):
lgtm, couldn't spot anything wrong

PR #9349, opened 2020-12-21
GameSettings: Patch for freezing in Gladius (GLSE64)

smurf3tte:
This game can deadlock the CPU and GPU by setting FIFO breakpoints too infrequently, provided the CPU gets too far ahead, as can happen under Dolphin due to timing inaccuracies. The game never clears breakpoints, and it will skip setting them if the previous one has not been hit by the GPU. If the CPU gets far enough ahead it will reach the FIFO high water mark and trigger an overflow interrupt, causing the render thread to be suspended. The GPU will make forward progress until it hits the last set breakpoint. However, if the distance between that breakpoint and the FIFO write pointer is greater than the low water mark, then the GPU will never generate an underflow interrupt and the render thread will never be resumed. This patch forces the game to update the breakpoint unconditionally and has been tested on real hardware with no apparent ill effect.

This patch covers the US release (GLSE64), but there are three more (GLSD64, GLSF64, GLSP64) that will likely require similar patches.

https://bugs.dolphin-emu.org/issues/10669

### DIFF
--- a/Data/Sys/GameSettings/GLSE64.ini
+++ b/Data/Sys/GameSettings/GLSE64.ini
@@ -1,0 +1,20 @@
+# GLSE64 - Gladius
+
+[OnFrame]
+# This game can deadlock the CPU and GPU by setting FIFO breakpoints too
+# infrequently, provided the CPU gets too far ahead, as can happen under Dolphin
+# due to timing inaccuracies. The game never clears breakpoints, and it will
+# skip setting them if the previous one has not been hit by the GPU. If the CPU
+# gets far enough ahead it will reach the FIFO high water mark and trigger an
+# overflow interrupt, causing the render thread to be suspended. The GPU will
+# make forward progress until it hits the last set breakpoint. However, if the
+# distance between that breakpoint and the FIFO write pointer is greater than
+# the low water mark, then the GPU will never generate an underflow interrupt
+# and the render thread will never be resumed. This patch forces the game to
+# update the breakpoint unconditionally and has been tested on real hardware
+# with no apparent ill effect.
+$Fix freeze in opening cutscene
+0x8010B4B0:dword:0x60000000
+
+[OnFrame_Enabled]
+$Fix freeze in opening cutscene


iwubcode:
@JMC47 - do you have this game and is the freezing resolved by this patch?  Can it be merged?  Or are we waiting for the other 3 patches?

JosJuice:
Unlike some of the other patches that have been merged recently, it *should* be possible to fix this problem without a patch (even if it's hard). So I'm not sure if we should merge it or not.

JMC47:
My thoughts on this patch was that it shouldn't be merged. at first.

However, I think it's actually going to be more like the Final Fantasy Crystal Chronicles patch long term.  This patch gets the game working in dualcore, where as once we have proper GPU timings and whatnot, those we'll likely only be for single core.  While this does fix a game that should work in Dolphin anyway, I'm fine with this being some kind of performance patch as we still need to fix our timings for a bunch of other games.  The patch also gives us a lot of information on what's going wrong with these games that are hanging.

Rumi-Larry:
Is it possible to add a TODO reminder in an INI or is it invisible anyway?

JMC47:
We'll document this on the wiki and issue tracker.  This only fixes GLSE64, when there are also GLSP64, GLSF64, GLSD64.  I'm guessing the last one is a demo.

This will likely never get fixed in dualcore, and there is no roadmap to fix it in single core.  This fixes what is essentially a timing issue that is hard to emulate on the game side.  I've played ~4 hours in with the patch with no adverse side-effects.

Even if this is a temporary solution, I still feel a bit at fault as I was the one who originally broke the game. So today, let's make it work again.  I do hope that someday this patch is unnecessary for single core, though.

JosJuice:
For reference, I said on IRC that I wasn't opposing the merge.

AdmiralCurtiss:
> This only fixes GLSE64, when there are also GLSP64, GLSF64, GLSD64. I'm guessing the last one is a demo.

According to redump GSLP is the generic EU release, while GLSF and GLSD are French and German specific variants. Shouldn't be too hard to find second-hand if you live in France or Germany, if someone wants to port this patch.

PR #9351, opened 2020-12-22
Jits: Discard registers which we know will be overwritten

JosJuice:
This pull request adds a new "discarded" state for registers. Discarding a register is like flushing it, but without actually writing its value back to memory. We can discard a register only when it is guaranteed that no instruction will read from the register before it is next written to.

Discarding reduces the register pressure a little, and can also let us skip a few flushes on interpreter fallbacks.

### DIFF
--- a/Source/Core/Core/PowerPC/Jit64/RegCache/CachedReg.h
+++ b/Source/Core/Core/PowerPC/Jit64/RegCache/CachedReg.h
@@ -5,6 +5,7 @@
 #pragma once
 
 #include <cstddef>
+#include <optional>
 
 #include "Common/Assert.h"
 #include "Common/CommonTypes.h"
@@ -20,6 +21,8 @@ public:
   {
     /// Value is currently at its default location
     Default,
+    /// Value is not stored anywhere because we know it won't be read before the next write
+    Discarded,
     /// Value is currently bound to a x64 register
     Bound,
     /// Value is known as an immediate and has not been written back to its default location
@@ -35,32 +38,43 @@ public:
   {
   }
 
-  const Gen::OpArg& Location() const { return location; }
+  const std::optional<Gen::OpArg>& Location() const { return location; }
 
   LocationType GetLocationType() const
   {
+    if (!location.has_value())
+      return LocationType::Discarded;
+
     if (!away)
     {
       ASSERT(!revertable);
 
-      if (location.IsImm())
+      if (location->IsImm())
         return LocationType::SpeculativeImmediate;
 
       ASSERT(location == default_location);
       return LocationType::Default;
     }
 
-    ASSERT(location.IsImm() || location.IsSimpleReg());
-    return location.IsImm() ? LocationType::Immediate : LocationType::Bound;
+    ASSERT(location->IsImm() || location->IsSimpleReg());
+    return location->IsImm() ? LocationType::Immediate : LocationType::Bound;
   }
 
   bool IsAway() const { return away; }
+  bool IsDiscarded() const { return !location.has_value(); }
   bool IsBound() const { return GetLocationType() == LocationType::Bound; }
 
   void SetBoundTo(Gen::X64Reg xreg)
   {
     away = true;
     location = Gen::R(xreg);
+  }
+
+  void SetDiscarded()
+  {
+    ASSERT(!revertable);
+    away = false;
+    location = std::nullopt;
   }
 
   void SetFlushed()
@@ -104,7 +118,7 @@ public:
 
 private:
   Gen::OpArg default_location{};
-  Gen::OpArg location{};
+  std::optional<Gen::OpArg> location{};
   bool away = false;  // value not in source register
   bool revertable = false;
   size_t locked = 0;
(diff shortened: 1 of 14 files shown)


Source/Core/Core/PowerPC/Jit64/RegCache/CachedReg.h
```diff
@@ -106,6 +123,7 @@ class PPCCachedReg
   Gen::OpArg default_location{};
   Gen::OpArg location{};
   bool away = false;  // value not in source register
+  bool discarded = false;
```
MerryMage, line 126: Since a discarded vallue can never have a valid location, perhaps consider implementing this as `location == std::nullopt` for a `location` of type `std::optional<Gen::OpArg>`.

JosJuice: Done.

MerryMage:
x64 changes lgtm

Source/Core/Core/PowerPC/JitArm64/JitArm64_RegCache.cpp
```diff
@@ -480,6 +507,10 @@ ARM64Reg Arm64FPRCache::R(size_t preg, RegType type)
     }
     return host_reg;
   }
+  case RegType::Discarded:
+    host_reg = GetReg();
+    reg.Load(host_reg, type);
```
degasus, line 512: I assume assert(false) or a fallthrough to NotLoaded would fit better here.

Source/Core/Core/PowerPC/JitArm64/JitArm64_RegCache.cpp
```diff
@@ -391,7 +417,8 @@ void Arm64FPRCache::Flush(FlushMode mode, PPCAnalyst::CodeOp* op)
   {
     const RegType reg_type = m_guest_registers[i].GetType();
 
-    if (reg_type != RegType::NotLoaded && reg_type != RegType::Immediate)
+    if (reg_type != RegType::NotLoaded && reg_type != RegType::Discarded &&
+        reg_type != RegType::Immediate)
```
degasus, line 421: I assume FlushRegister shall assert if the reg_type is Discarded, so we should call it here and assert there. This will tell us if there is a bug here in this code.
However this will also trigger on interpreter fallbacks, so I'm not sure about this signaling...

anyways, the next comment is very outdated.

Source/Core/Core/PowerPC/PPCAnalyst.cpp
```diff
@@ -551,6 +551,10 @@ void PPCAnalyzer::SetInstructionStats(CodeBlock* block, CodeOp* code, const Gekk
   code->outputFPRF = (opinfo->flags & FL_SET_FPRF) != 0;
   code->canEndBlock = (opinfo->flags & FL_ENDBLOCK) != 0;
 
+  // TODO: Is it possible to determine that some FPU instructions never cause exceptions?
+  code->canCauseException =
+      (opinfo->flags & (FL_LOADSTORE | FL_USE_FPU | FL_PROGRAMEXCEPTION)) != 0;
```
degasus, line 556: The first FPU instruction will throw an exception, so all others are fine.

Source/Core/Core/PowerPC/PPCAnalyst.cpp
```diff
@@ -939,21 +943,26 @@ u32 PPCAnalyzer::Analyze(u32 address, CodeBlock* block, CodeBuffer* buffer, std:
...
     op.fprInXmm = fprInXmm;
-    // TODO: if there's no possible endblocks or exceptions in between, tell the regcache
-    // we can throw away a register if it's going to be overwritten later.
     gprInUse |= op.regsIn;
-    gprInReg |= op.regsIn;
     fprInUse |= op.fregsIn;
+    if (op.canEndBlock || op.canCauseException)
+    {
+      gprDiscardable = BitSet32{};
+      fprDiscardable = BitSet32{};
```
degasus, line 954: I wonder, is there a reason why wantsCR, wantsFPRF, ..., is only set for op.canEndBlock?

Source/Core/Core/PowerPC/PPCAnalyst.cpp
JosJuice: Can't FPU instructions throw exceptions in other situations as well (that may not have been implemented in the JITs yet), like division by zero? (See the discussion about True Crime on IRC earlier this month)

Source/Core/Core/PowerPC/PPCAnalyst.cpp
JosJuice: I'm not sure exactly how the logic works, and I'm also not sure if it has to do with this PR.

Source/Core/Core/PowerPC/JitArm64/JitArm64_RegCache.cpp
JosJuice: With interpreter fallbacks in mind, I think I would like to keep it the way it is.

> anyways, the next comment is very outdated.

What are you referring to by this?

degasus: I was talking about `XXX: Determine if we can keep a register in the lower 64bits \n Which will allow it to be callee saved.` However I don't care if you drop it or not.

degasus (approved):

PR #9352, opened 2020-12-22
Software: Implement line-width and point-width

Pokechu22:
Somewhat equivalent to https://bugs.dolphin-emu.org/issues/7749 / #1706.  For whatever reason, textures aren't working (they already were broken for lines, but this change doesn't fix it even though I think it should).

### DIFF
--- a/Source/Core/VideoBackends/Software/Clipper.cpp
+++ b/Source/Core/VideoBackends/Software/Clipper.cpp
@@ -331,18 +331,36 @@ void ProcessTriangle(OutputVertexData* v0, OutputVertexData* v1, OutputVertexDat
   }
 }
 
-static void CopyVertex(OutputVertexData* dst, const OutputVertexData* src, float dx, float dy,
-                       unsigned int sOffset)
+constexpr std::array<float, 8> LINE_PT_TEX_OFFSETS = {
+    0, 1 / 16.f, 1 / 8.f, 1 / 4.f, 1 / 2.f, 1, 1, 1,
+};
+
+static void CopyLineVertex(OutputVertexData* dst, const OutputVertexData* src, int px, int py,
+                           bool apply_line_offset)
 {
-  dst->screenPosition.x = src->screenPosition.x + dx;
-  dst->screenPosition.y = src->screenPosition.y + dy;
+  const float line_half_width = bpmem.lineptwidth.linesize / 12.0f;
+
+  dst->projectedPosition = src->projectedPosition;
+  dst->screenPosition.x = src->screenPosition.x + px * line_half_width;
+  dst->screenPosition.y = src->screenPosition.y + py * line_half_width;
   dst->screenPosition.z = src->screenPosition.z;
 
   dst->normal = src->normal;
   dst->color = src->color;
 
-  // todo - s offset
   dst->texCoords = src->texCoords;
+
+  if (apply_line_offset && LINE_PT_TEX_OFFSETS[bpmem.lineptwidth.lineoff] != 0)
+  {
+    for (u32 coord_num = 0; coord_num < xfmem.numTexGen.numTexGens; coord_num++)
+    {
+      if (bpmem.texcoords[coord_num].s.line_offset)
+      {
+        dst->texCoords[coord_num].x += (bpmem.texcoords[coord_num].s.scale_minus_1 + 1) *
+                                       LINE_PT_TEX_OFFSETS[bpmem.lineptwidth.lineoff];
+      }
+    }
+  }
 }
 
 void ProcessLine(OutputVertexData* lineV0, OutputVertexData* lineV1)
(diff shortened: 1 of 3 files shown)


Source/Core/VideoBackends/Software/Clipper.cpp
```diff
@@ -331,7 +331,9 @@ void ProcessTriangle(OutputVertexData* v0, OutputVertexData* v1, OutputVertexDat
   }
 }
 
-static void CopyVertex(OutputVertexData* dst, const OutputVertexData* src, float dx, float dy,
+static const int LINE_PT_TEX_OFFSETS[8] = {0, 16, 8, 4, 2, 1, 1, 1};
```
lioncash, line 334:
```suggestion
constexpr std::array<int, 8> LINE_PT_TEX_OFFSETS{0, 16, 8, 4, 2, 1, 1, 1};
```

Pokechu22: Note that this was copied from here:

https://github.com/dolphin-emu/dolphin/blob/0029ca84b07f44e3ce8c80a44fa0285a966b34d3/Source/Core/VideoCommon/GeometryShaderManager.cpp#L14

It also was originally an array of floats until 0ac710339100dc6dc2f1f4c094b372846decdce5; I don't think the concerns there apply to the software renderer.

lioncash: Sure, but this should still be migrated to `std::array`, since library implementations allow tuning/enabling bounds checks both in debug and release builds, allowing OOB accesses to be caught (and also forces decays to pointer to be explicit, but that doesn't apply here).

Pokechu22:
After adding `dst->projectedPosition = src->projectedPosition`, textures show up; it's used in `Rasterizer`:

https://github.com/dolphin-emu/dolphin/blob/1d489b3fd5cdcb9f3da2fd07123d6664ef9fa9a6/Source/Core/VideoBackends/Software/Rasterizer.cpp#L348-L349

I also am currently adding 32 to tex coords regardless of the value of `sOffset` (as long as it's nonzero), which seems to work for Mega Man Network Transmission but almost certainly isn't correct (though the only test case that seems to be broken by it is [burnout2-vehicles](https://fifo.ci/compare/5802859-5796946/), where the sparkles are repeated, and _maybe_ metroid-visor).  Using `128*LINE_PT_TEX_OFFSETS[sOffset]` approximates it, but gives incorrect results for part of the image.  Hardware code uses something that basically matches `LINE_PT_TEX_OFFSETS` (it divides by an integer instead of multiplying by a fraction):

https://github.com/dolphin-emu/dolphin/blob/1d489b3fd5cdcb9f3da2fd07123d6664ef9fa9a6/Source/Core/VideoCommon/GeometryShaderGen.cpp#L263-L270

Are the units the same (0 to the width/height of the texture)?  Or do the texture corods in GeometryShaderGen have a range of 0 to 1?  If it is the texture size, then I'll need an explanation of how to actually access that; I have no idea how to do it.

-----

When combined with #9321, this also fixes the RS2 hud coloration and missing minimap dot problems mentioned in https://github.com/dolphin-emu/dolphin/pull/9321#issuecomment-743721889.  The differences are basically the same as seen in [rs3-skybox](https://fifo.ci/compare/5802856-5796910/).

Pokechu22:
Using `(bpmem.texcoords[coordNum].s.scale_minus_1 + 1) * LINE_PT_TEX_OFFSETS[sOffset]` gives what appears to be the correct result.

Source/Core/VideoBackends/Software/Clipper.cpp
```diff
@@ -388,19 +404,67 @@ void ProcessLine(OutputVertexData* lineV0, OutputVertexData* lineV1)
...
+
+  dst->texCoords = src->texCoords;
+
+  if (LINE_PT_TEX_OFFSETS[sOffset] != 0)
+  {
+    for (u32 coordNum = 0; coordNum < xfmem.numTexGen.numTexGens; coordNum++)
+    {
+      if (bpmem.texcoords[coordNum].s.point_offset)
+      {
+        if (px)
```
lioncash, line 439: We brace if statements where the body travels more than one line , ditto for the `if (py)` if statement below this one.

Source/Core/VideoBackends/Software/Clipper.cpp
```diff
@@ -331,18 +331,34 @@ void ProcessTriangle(OutputVertexData* v0, OutputVertexData* v1, OutputVertexDat
...
   dst->screenPosition.y = src->screenPosition.y + dy;
   dst->screenPosition.z = src->screenPosition.z;
 
   dst->normal = src->normal;
   dst->color = src->color;
 
-  // todo - s offset
   dst->texCoords = src->texCoords;
+
+  if (LINE_PT_TEX_OFFSETS[sOffset] != 0)
```
BhaaLseN, line 351: Is `sOffset` guaranteed to be in the range 0..7 here? Or do we want this to blow up if it isn't? (Or do we want to add a range check?)
Same for the other one in `CopyPointVertex`.

Pokechu22: Yes, it's guaranteed because it comes from a 3-bit field:

https://github.com/dolphin-emu/dolphin/blob/0029ca84b07f44e3ce8c80a44fa0285a966b34d3/Source/Core/VideoCommon/BPMemory.h#L577-L578

Thoughts on removing the parameter and directly using `bpmem.lineptwidth.pointoff`/`lineoff` in the function though?

Pokechu22:
I've tidied up the relevant functions a little bit more (`CopyLineVertex` and `CopyPointVertex` directly get BP values, with the parameters being used to indicate whether an offset should be performed), and also copied a note from [the hardware implementation](https://github.com/dolphin-emu/dolphin/blob/e62c33c413c7c6f86bf6a2808a3c43c0a1a47f7f/Source/Core/VideoCommon/GeometryShaderGen.cpp#L168-L174).

One thing that that comment claims is that the hardware implementation creates lines that have the wrong width, mentioning the Twilight Princess map, but it doesn't say whether it's too thin or thick.  I think the issue it's referring to is [5325](https://bugs.dolphin-emu.org/issues/5325), which is long fixed, but the image links are broken so I'm not sure.  Looking at [a longplay](https://youtu.be/cLBN-r9YeEw?t=2083) the thicker/darker version this change and HW have in Rogue Squadron seems correct.

JMC47:
The fifologs look sane, it's been reviewed and these are some valuable improvements to the software renderer.  LGTM.

PR #9354, opened 2020-12-23
Enable Rumble with Original GameCube Controller connected with 8Bitdo GBros Adapter xinput Firmware v2.24

Titanius:
How can I activate Rumble in the app under Controller -> Rumble?

Use Original GameCube Controller connected with 8Bitdo GBros Adapter xinput Firmware v2.24

add the 8Bitdo GBros Adapter in iOS works fine and the buttons on the controller working too.


leoetlino:
This is not a support forum.

PR #9355, opened 2020-12-23
Call UpdatePerformanceMonitor when needed

JosJuice:
I find the performance monitor functionality useful for timing hardware tests, but unfortunately it hasn't worked correctly in Dolphin except when using the x86-64 JIT with no idle skipping. This PR makes it work (mostly) correctly on all CPU emulation cores.

### DIFF
--- a/Source/Core/Core/CoreTiming.cpp
+++ b/Source/Core/Core/CoreTiming.cpp
@@ -380,6 +380,7 @@ void Idle()
     Fifo::FlushGpu();
   }
 
+  PowerPC::UpdatePerformanceMonitor(PowerPC::ppcState.downcount, 0, 0);
   s_idled_cycles += DowncountToCycles(PowerPC::ppcState.downcount);
   PowerPC::ppcState.downcount = 0;
 }
--- a/Source/Core/Core/PowerPC/CachedInterpreter/CachedInterpreter.cpp
+++ b/Source/Core/Core/PowerPC/CachedInterpreter/CachedInterpreter.cpp
@@ -134,6 +134,17 @@ static void EndBlock(UGeckoInstruction data)
 {
   PC = NPC;
   PowerPC::ppcState.downcount -= data.hex;
+  PowerPC::UpdatePerformanceMonitor(data.hex, 0, 0);
+}
+
+static void UpdateNumLoadStoreInstructions(UGeckoInstruction data)
+{
+  PowerPC::UpdatePerformanceMonitor(0, data.hex, 0);
+}
+
+static void UpdateNumFloatingPointInstructions(UGeckoInstruction data)
+{
+  PowerPC::UpdatePerformanceMonitor(0, 0, data.hex);
 }
 
 static void WritePC(UGeckoInstruction data)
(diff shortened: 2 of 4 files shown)


Source/Core/Core/PowerPC/CachedInterpreter/CachedInterpreter.cpp
```diff
@@ -274,13 +291,19 @@ void CachedInterpreter::Jit(u32 address)
       if (idle_loop)
         m_code.emplace_back(CheckIdle, js.blockStart);
       if (endblock)
+      {
         m_code.emplace_back(EndBlock, js.downcountAmount);
+        m_code.emplace_back(UpdateNumLoadStoreInstructions, js.numLoadStoreInst);
```
smurf3tte, line 296: Are any games known to use these counters? If not, two calls at the end of every block is steep price to pay. Have you considered replicating the MCR check that is in the JIT? It's definitely a hack, but it seems justifiable if the feature is only exercised by test code.

JosJuice: No games use them as far as I know. (We should've gotten reports about things breaking on ARM by now otherwise.)

I was a bit unsure if I should add that hack here or not when I was working on the code. But I'd like to have input from one more person before switching it over.

leoetlino: Have you measured the performance impact?

JosJuice: I tried it now, and it actually seems to be running one FPS faster with this change than without. I'm going to assume it's just a random variation. (The scene I used was the file select screen in Anubis II, which runs at about 33 FPS for me. The scene has 3D graphics but is otherwise relatively light.)

leoetlino: IMO it's not worth porting the speed hack to the interpreter, then.

Source/Core/Core/CoreTiming.cpp
```diff
@@ -380,6 +380,7 @@ void Idle()
     Fifo::FlushGpu();
   }
 
+  PowerPC::UpdatePerformanceMonitor(PowerPC::ppcState.downcount, 0, 0);
```
leoetlino, line 383: Is it fine to use 0 as the number of load/store instructions? Idle loops typically have at least one load, right? (Not sure if this actually matters in practice.)

JosJuice: Technically it's wrong, but it's not trivial to fix, and the behavior of the load/store counter is the same as before this line was added.

leoetlino: Yeah, I guess counting the number of load/store instructions in PPCAnalyzer would only give an approximation anyway...

leoetlino (approved):

Source/Core/Core/PowerPC/CachedInterpreter/CachedInterpreter.cpp
JosJuice: Actually, never mind about no games using them. Harry Potter and the Prisoner of Azkaban uses them according to cdace9d, and that game has indeed been known to be broken in JitArm64.

PR #9357, opened 2020-12-24
Android: Only have one settings entrypoint per activity/dialog

JosJuice:
Basically, instead of having one button for config, one button for graphics settings and so on, we now have just one settings button which takes you to a screen where you pick between config/graphics/GameCube controllers/Wii Remotes.

The main reason I want to do this is because people still have trouble finding Overlay Controls in the "new" in-game menu. Typically (depending on the screen size and the length of the game name), the scrollable part of the menu can fit 4 items, and merging Config and Graphics Settings into one item would move Overlay Controls from 5th place to 4th place (assuming the user doesn't have savestates enabled), which makes it findable even for users who don't realize the menu can be scrolled.

The dialog that's shown when long pressing a game in the game list is also shortened. While not a pressing matter, I think it was getting a bit long.

An additional reason to do this is because we probably will want to make it possible to edit the controller settings from the in-game menu at some point in the future. With the old approach, this would require us to dedicate a whopping 4 menu items just for settings (not including Overlay Controls), which I think is excessive.


lioncash (approved):

Bankaimaster999:
awesome

PR #9358, opened 2020-12-24
Prevent hotkey suppressions from sticking when disabling freelook.

jordan-woyak:
If hotkeys were being held while disabling freelook they would continue to block regular inputs.
This fixes that minor issue.

### DIFF
--- a/Source/Core/Core/FreeLookManager.cpp
+++ b/Source/Core/Core/FreeLookManager.cpp
@@ -11,6 +11,7 @@
 #include "Core/ConfigManager.h"
 #include "Core/FreeLookConfig.h"
 
+#include "InputCommon/ControlReference/ControlReference.h"
 #include "InputCommon/ControllerEmu/ControlGroup/Buttons.h"
 #include "InputCommon/InputConfig.h"
 
@@ -146,8 +147,15 @@ ControllerEmu::ControlGroup* FreeLookController::GetGroup(FreeLookGroup group) c
 
 void FreeLookController::Update()
 {
+  const bool prev_input_gate = ControlReference::GetInputGate();
+
   if (!g_freelook_camera.IsActive())
-    return;
+  {
+    // We must GetState even when inactive in case hotkey @() expressions were pressed during
+    // deactivation, otherwise they will never release their button suppressions.
+    // Turning the input gate off will block all input and release any hotkeys.
+    ControlReference::SetInputGate(false);
+  }
 
   if (m_move_buttons->controls[MoveButtons::Up]->GetState<bool>())
     g_freelook_camera.MoveVertical(-g_freelook_camera.GetSpeed());
@@ -190,6 +198,8 @@ void FreeLookController::Update()
 
   if (m_other_buttons->controls[OtherButtons::ResetView]->GetState<bool>())
     g_freelook_camera.Reset();
+
+  ControlReference::SetInputGate(prev_input_gate);
 }
 
 namespace FreeLook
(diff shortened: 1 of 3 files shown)


iwubcode:
I'd honestly prefer this wait until after #8867 is merged since that completely revamps how Free Look is handled (it's been signed off code wise and usability wise...).

However, the changes here seem fine to me (not sure if the active functor will be needed outside of Free Look though?)

SirMangler:
> I'd honestly prefer this wait until after #8867 is merged

While this typically would be ideal, this scenario differs. As stated, the @() syntax breaks any controller profiles that make use of SHIFT combined with all the primary alphabetical keys (including WASD, F, Q, R as well as Scroll Wheel). They simply do not activate for no apparent reason (unless you're aware of the full behaviour of @() and the new free look defaults).

Because it this effects so many mouse+keyboard profiles and the cause isn't clear to most users, I strongly feel this should be merged as soon as feasible to minimise the problem's lifespan.

iwubcode:
If I'm understanding the issue, with #8667 now merged, this should be resolved.

Source/Core/Core/FreeLookManager.cpp
```diff
@@ -146,8 +147,15 @@ ControllerEmu::ControlGroup* FreeLookController::GetGroup(FreeLookGroup group) c
...
 {
+  const bool prev_input_gate = ControlReference::GetInputGate();
+
   if (!g_freelook_camera.IsActive())
-    return;
+  {
+    // We must GetState even when inactive in case hotkey @() expressions were pressed during
+    // deactivation, otherwise they will never release their button suppressions.
+    // Turning the input gate off will block all input and release any hotkeys.
+    ControlReference::SetInputGate(false);
```
iwubcode, line 157: I think we still need a return after this?  The idea is that we don't want to do any current or future logic if we are inactive.

iwubcode: Could we do:

```
if (g_freelook_camera.IsActive())
{
  ControlReference::SetInputGate(true);
}
else
{
  ControlReference::SetInputGate(false);
  return;
}
```

That would preserve the early out behavior

jordan-woyak: This won't work. It can't return. The current fix is closing the input gate AND reading the inputs so they release their suppressions.

iwubcode: Ahhhhh, I missed that.

iwubcode: Not sure if this makes sense:

```
if (g_freelook_camera.IsActive())
{
  ControlReference::SetInputGate(true);
}
else
{
  const bool prev_input_gate = ControlReference::GetInputGate();
  ControlReference::SetInputGate(false);
  if (prev_input_gate)
    DoUpdate();  // Flush suppressions...
  return;
}

DoUpdate();  // This function is essentially 'Update' now without the active check
```

jordan-woyak: That's roughly my logic, but this suggestion has potential to get stuck hotkeys on focus loss situations if you don't call `DoUpdate` when the input gate was closed.

iwubcode: I think the above snippet should always close it because we will always transition between an input gate of true to false.  The catch is we only call `DoUpdate()` once to do the flush.  Your current code is calling it every single time.

The problem with this is when there are non-hotkey buttons mapped.  An example today would be assigning Free Look speed increase to a gamepad button.  With your code, if I hit that gamepad button, it'll trigger even if Free Look is inactive.  The current code only triggers if Free Look is active.

iwubcode: Maybe I'm misunderstanding the input gate.  Does turning it to false block _all_ input or just hotkeys?

jordan-woyak:
A lot of input code has changed. I'll have to re-figure out if this is even still relevant.

PR #9360, opened 2020-12-25
Split OSREPORT logging type

Minty-Meeo:
The enumerated LOG_TYPE "OSREPORT" is currently used in both EXI_DeviceIPL.cpp and HLE_OS.cpp.  In many games, the multitude of game functions detected by HLE_OS.cpp for OSREPORT logging results in poor log readability.  This Pull Request remedies that by adding a new enumerated LOG_TYPE "OSREPORT_HLE" for log usage in HLE_OS.cpp.

In the future, further changing how logging in HLE_OS.cpp works may be desirable.  As it is, game functions are detected that send a single character to the log.  This is a major source of poor readability.


Pokechu22:
IMO it might be better to create a new category for EXI_DeviceIPL and use that for everything that's logged there, rather than leaving OSREPORT for just the EXI logs.  Though, that could make things a bit confusing if you specifically want the EXI OSReport logs.

Minty-Meeo:
I left the original enumerated LOG_TYPE "OSREPORT" as is to minimally effect people's existing log configuration and existing code.  In the GUI, however, the toggle is now named OSReport EXI.

sepalani (approved):
This PR LGTM.

Some remarks to keep in mind:
 - The apploader requires an HLE OSReport, so the apploader messages will go through `OSReport HLE`
 - This point is wrong (`that send a single character to the log`) and just means you patched the wrong function which might be a signature false-positive, a stubbed function or the game literally sends one character at a time through the log

leoetlino:
Yeah I think it makes more sense to keep EXI OSReport in OSREPORT...

leoetlino (approved):

PR #9362, opened 2020-12-26
VideoCommon: Fix crash that occurs on loading a fifo log when Free Look is enabled

iwubcode:
During the Free Look refactor (#8867 ) I apparently forgot to test loading a fifo log.  A crash occurs there because the camera control type is not set and therefore the controller is null.

This fixes the crash and adds some safety checks to check for null pointers.

Sorry about that!

EDIT: This crash is not specific to fifo per-se.  It's just the fifo logs I tested had an initial frame of projection-type perspective where we call [GetView()](https://github.com/dolphin-emu/dolphin/blob/master/Source/Core/VideoCommon/VertexShaderManager.cpp#L423).  Most games I tested startup with a projection type of orthographic, giving time for `Renderer::CheckForConfigChanges()` to [update the control type](https://github.com/dolphin-emu/dolphin/blob/master/Source/Core/VideoCommon/RenderBase.cpp#L409).

### DIFF
--- a/Source/Core/VideoCommon/FreeLookCamera.cpp
+++ b/Source/Core/VideoCommon/FreeLookCamera.cpp
@@ -175,6 +175,11 @@ private:
 };
 }  // namespace
 
+FreeLookCamera::FreeLookCamera()
+{
+  SetControlType(FreeLook::ControlType::SixAxis);
+}
+
 void FreeLookCamera::SetControlType(FreeLook::ControlType type)
 {
   if (m_current_type && *m_current_type == type)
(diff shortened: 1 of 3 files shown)


leoetlino:
Alternative approach, which I think would be more robust: call SetControlType(SixAxis) (for instance) in the FreeLookCamera constructor so the object is always in a valid state and we don't have to do those null checks elsewhere.

iwubcode:
@leoetlino - nice idea!  Appreciate the suggestion.  I updated the code to make that change.

leoetlino (approved):

PR #9365, opened 2020-12-27
DolphinQt: Add shortcut to desktop

Aminoa:
Adds option to context menu when selecting a game to add a shortcut of that game to the desktop. It creates a shortcut to the desktop when selected. Uses native windows API since Qt doesn't support shortcut creation.

### DIFF
--- a/Source/Core/DolphinQt/GameList/GameList.cpp
+++ b/Source/Core/DolphinQt/GameList/GameList.cpp
@@ -53,6 +53,17 @@
 
 #include "UICommon/GameFile.h"
 
+#include "windows.h"
+#include "winnls.h"
+#include "shobjidl.h"
+#include "objbase.h"
+#include "objidl.h"
+#include "shlguid.h"
+#include <iostream>
+#include "shlobj.h"
+#include <qcoreapplication.h>
+using namespace std;
+
 GameList::GameList(QWidget* parent) : QStackedWidget(parent), m_model(this)
 {
   m_list_proxy = new ListProxyModel(this);
@@ -382,6 +393,7 @@ void GameList::ShowContextMenu(const QPoint&)
 
     menu->addAction(tr("Open &Containing Folder"), this, &GameList::OpenContainingFolder);
     menu->addAction(tr("Delete File..."), this, &GameList::DeleteFile);
+    menu->addAction(tr("Add Shortcut to Desktop"), this, &GameList::AddShortcutToDesktop);
 
     menu->addSeparator();
 
@@ -626,6 +638,60 @@ void GameList::OpenGCSaveFolder()
 
   if (!found)
     ModalMessageBox::information(this, tr("Information"), tr("No save data found."));
+}
+
+void GameList::AddShortcutToDesktop()
+{
+  const auto game = GetSelectedGame();
+  string game_name = game->GetLongName();
+  string file_path = game->GetFilePath();
+  string file_name = game->GetFileName();
+  wstring dolphin_path = QCoreApplication::applicationFilePath().toStdWString();
+  LPCWSTR dolphin_path_convert = dolphin_path.c_str();
+
+  // Arguments
+  string args = "-e \"" + file_path + "\"";
+  std::wstring args_convert = std::wstring(args.begin(), args.end());
+
+  // Gets the desktop position with shorcut
+  WCHAR* desktop = 0;
+  SHGetKnownFolderPath(FOLDERID_Desktop, KF_FLAG_NO_ALIAS, NULL, &desktop);
+  CoTaskMemFree(desktop);
+  // Converts desktop path to string
+  wstring ws = wstring(desktop);
+  string desktop_path = string(ws.begin(), ws.end());
+  desktop_path = desktop_path + "\\" + game_name + ".lnk";
+  LPCSTR pathLink = desktop_path.c_str();
+
+  HRESULT hres;
+  IShellLink* psl;
+
+  CoInitialize(NULL);
+  hres =
+      CoCreateInstance(CLSID_ShellLink, NULL, CLSCTX_INPROC_SERVER, IID_IShellLink, (LPVOID*)&psl);
+
+  if (SUCCEEDED(hres))
+  {
+    IPersistFile* ppf;
+    LPCWSTR pszArgs = args_convert.c_str();
+
+    psl->SetPath(dolphin_path_convert);
+    psl->SetArguments(pszArgs);
+
+    hres = psl->QueryInterface(IID_IPersistFile, (LPVOID*)&ppf);
+
+    if (SUCCEEDED(hres))
+    {
+      WCHAR wsz[MAX_PATH];
+      MultiByteToWideChar(CP_ACP, 0, pathLink, -1, wsz, MAX_PATH);
+      hres = ppf->Save(wsz, TRUE);
+      ppf->Release();
+    }
+    psl->Release();
+  }
+
+  ModalMessageBox::information(this, tr("Add Shortcut to Desktop"),
+                               tr("Shortcut successfully added to the desktop"));
 }
 
 void GameList::DeleteFile()
--- a/Source/Core/DolphinQt/GameList/GameList.cpp
+++ b/Source/Core/DolphinQt/GameList/GameList.cpp
@@ -62,7 +62,7 @@
 #include <iostream>
 #include "shlobj.h"
 #include <qcoreapplication.h>
-using namespace std;
+//using namespace std;
 
 GameList::GameList(QWidget* parent) : QStackedWidget(parent), m_model(this)
 {
@@ -643,14 +643,14 @@ void GameList::OpenGCSaveFolder()
 void GameList::AddShortcutToDesktop()
 {
   const auto game = GetSelectedGame();
-  string game_name = game->GetLongName();
-  string file_path = game->GetFilePath();
-  string file_name = game->GetFileName();
-  wstring dolphin_path = QCoreApplication::applicationFilePath().toStdWString();
+  std::string game_name = game->GetLongName();
+  std::string file_path = game->GetFilePath();
+  std::string file_name = game->GetFileName();
+  std::wstring dolphin_path = QCoreApplication::applicationFilePath().toStdWString();
   LPCWSTR dolphin_path_convert = dolphin_path.c_str();
 
   // Arguments
-  string args = "-e \"" + file_path + "\"";
+  std::string args = "-e \"" + file_path + "\"";
   std::wstring args_convert = std::wstring(args.begin(), args.end());
 
   // Gets the desktop position with shorcut
@@ -658,8 +658,8 @@ void GameList::AddShortcutToDesktop()
   SHGetKnownFolderPath(FOLDERID_Desktop, KF_FLAG_NO_ALIAS, NULL, &desktop);
   CoTaskMemFree(desktop);
   // Converts desktop path to string
-  wstring ws = wstring(desktop);
-  string desktop_path = string(ws.begin(), ws.end());
+  std::wstring ws = std::wstring(desktop);
+  std::string desktop_path = std::string(ws.begin(), ws.end());
   desktop_path = desktop_path + "\\" + game_name + ".lnk";
   LPCSTR pathLink = desktop_path.c_str();
 
--- a/Source/Core/DolphinQt/GameList/GameList.cpp
+++ b/Source/Core/DolphinQt/GameList/GameList.cpp
@@ -53,16 +53,11 @@
 
 #include "UICommon/GameFile.h"
 
-#include "windows.h"
-#include "winnls.h"
-#include "shobjidl.h"
-#include "objbase.h"
-#include "objidl.h"
-#include "shlguid.h"
-#include <iostream>
+#ifdef _WIN32
 #include "shlobj.h"
 #include <qcoreapplication.h>
-//using namespace std;
+#include <wrl/client.h>
+#endif 
 
 GameList::GameList(QWidget* parent) : QStackedWidget(parent), m_model(this)
 {
@@ -393,7 +388,9 @@ void GameList::ShowContextMenu(const QPoint&)
 
     menu->addAction(tr("Open &Containing Folder"), this, &GameList::OpenContainingFolder);
     menu->addAction(tr("Delete File..."), this, &GameList::DeleteFile);
+    #ifdef _WIN32
     menu->addAction(tr("Add Shortcut to Desktop"), this, &GameList::AddShortcutToDesktop);
+    #endif
 
     menu->addSeparator();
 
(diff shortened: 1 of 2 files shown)


JosJuice:
@dolphin-emu-bot rebuild

Source/Core/DolphinQt/GameList/GameList.cpp
```diff
@@ -53,6 +53,17 @@
...
+#include "windows.h"
+#include "winnls.h"
+#include "shobjidl.h"
+#include "objbase.h"
+#include "objidl.h"
+#include "shlguid.h"
+#include <iostream>
+#include "shlobj.h"
+#include <qcoreapplication.h>
+//using namespace std;
```
JosJuice, line 65: Please remove this line entirely.

Source/Core/DolphinQt/GameList/GameList.cpp
```diff
@@ -628,6 +640,60 @@ void GameList::OpenGCSaveFolder()
...
+  std::string args = "-e \"" + file_path + "\"";
+  std::wstring args_convert = std::wstring(args.begin(), args.end());
+
+  // Gets the desktop position with shorcut
+  WCHAR* desktop = 0;
+  SHGetKnownFolderPath(FOLDERID_Desktop, KF_FLAG_NO_ALIAS, NULL, &desktop);
+  CoTaskMemFree(desktop);
+  // Converts desktop path to string
+  std::wstring ws = std::wstring(desktop);
+  std::string desktop_path = std::string(ws.begin(), ws.end());
```
JosJuice, line 662: This is not a correct way of converting from UTF-16. I would suggest using `TStrToUTF8` from StringUtil.h.

Source/Core/DolphinQt/GameList/GameList.cpp
```diff
@@ -628,6 +640,60 @@ void GameList::OpenGCSaveFolder()
...
+
+    psl->SetPath(dolphin_path_convert);
+    psl->SetArguments(pszArgs);
+
+    hres = psl->QueryInterface(IID_IPersistFile, (LPVOID*)&ppf);
+
+    if (SUCCEEDED(hres))
+    {
+      WCHAR wsz[MAX_PATH];
+      MultiByteToWideChar(CP_ACP, 0, pathLink, -1, wsz, MAX_PATH);
```
JosJuice, line 686: Using "multibyte" strings will cause problems in the case where we have a path that contains characters which are foreign from the system codepage's point of view. If you switch to `TStrToUTF8` above, you can solve this by using `UTF8ToTStr` here.

JosJuice (commented):
If you're going to add code that only works on Windows, you need to wrap the code in `#ifdef _WIN32`.

Source/Core/DolphinQt/GameList/GameList.cpp
```diff
@@ -53,6 +53,17 @@
 
 #include "UICommon/GameFile.h"
 
+#include "windows.h"
```
lioncash, line 56: I am pretty sure *all* of these includes aren't necessary. I'd pare it down to the ones that are necessary.

Source/Core/DolphinQt/GameList/GameList.cpp
```diff
@@ -628,6 +640,60 @@ void GameList::OpenGCSaveFolder()
...
+  // Arguments
+  std::string args = "-e \"" + file_path + "\"";
+  std::wstring args_convert = std::wstring(args.begin(), args.end());
+
+  // Gets the desktop position with shorcut
+  WCHAR* desktop = 0;
+  SHGetKnownFolderPath(FOLDERID_Desktop, KF_FLAG_NO_ALIAS, NULL, &desktop);
+  CoTaskMemFree(desktop);
+  // Converts desktop path to string
+  std::wstring ws = std::wstring(desktop);
```
lioncash, line 661: Given desktop is passed to CoTaskMemFree above, is this even valid to do?

Source/Core/DolphinQt/GameList/GameList.cpp
```diff
@@ -628,6 +640,60 @@ void GameList::OpenGCSaveFolder()
...
+  std::wstring dolphin_path = QCoreApplication::applicationFilePath().toStdWString();
+  LPCWSTR dolphin_path_convert = dolphin_path.c_str();
+
+  // Arguments
+  std::string args = "-e \"" + file_path + "\"";
+  std::wstring args_convert = std::wstring(args.begin(), args.end());
+
+  // Gets the desktop position with shorcut
+  WCHAR* desktop = 0;
+  SHGetKnownFolderPath(FOLDERID_Desktop, KF_FLAG_NO_ALIAS, NULL, &desktop);
```
lioncash, line 658: We use nullptr instead of NULL

Source/Core/DolphinQt/GameList/GameList.cpp
```diff
@@ -628,6 +640,60 @@ void GameList::OpenGCSaveFolder()
...
+  std::string file_name = game->GetFileName();
+  std::wstring dolphin_path = QCoreApplication::applicationFilePath().toStdWString();
+  LPCWSTR dolphin_path_convert = dolphin_path.c_str();
+
+  // Arguments
+  std::string args = "-e \"" + file_path + "\"";
+  std::wstring args_convert = std::wstring(args.begin(), args.end());
+
+  // Gets the desktop position with shorcut
+  WCHAR* desktop = 0;
```
lioncash, line 657:
```suggestion
  WCHAR* desktop = nullptr;
```

Source/Core/DolphinQt/GameList/GameList.cpp
```diff
@@ -628,6 +640,60 @@ void GameList::OpenGCSaveFolder()
     ModalMessageBox::information(this, tr("Information"), tr("No save data found."));
 }
 
+void GameList::AddShortcutToDesktop()
+{
+  const auto game = GetSelectedGame();
+  std::string game_name = game->GetLongName();
```
lioncash, line 646: These variables should be marked as const where applicable.

Source/Core/DolphinQt/GameList/GameList.cpp
```diff
@@ -628,6 +640,60 @@ void GameList::OpenGCSaveFolder()
...
+  SHGetKnownFolderPath(FOLDERID_Desktop, KF_FLAG_NO_ALIAS, NULL, &desktop);
+  CoTaskMemFree(desktop);
+  // Converts desktop path to string
+  std::wstring ws = std::wstring(desktop);
+  std::string desktop_path = std::string(ws.begin(), ws.end());
+  desktop_path = desktop_path + "\\" + game_name + ".lnk";
+  LPCSTR pathLink = desktop_path.c_str();
+
+  HRESULT hres;
+  IShellLink* psl;
```
lioncash, line 667: You may wish to use [ComPtr](https://docs.microsoft.com/en-us/cpp/cppcx/wrl/comptr-class?view=msvc-160) here, so that the release calls further down below aren't necessary.

Source/Core/DolphinQt/GameList/GameList.cpp
```diff
@@ -628,6 +640,60 @@ void GameList::OpenGCSaveFolder()
...
+  WCHAR* desktop = 0;
+  SHGetKnownFolderPath(FOLDERID_Desktop, KF_FLAG_NO_ALIAS, NULL, &desktop);
+  CoTaskMemFree(desktop);
+  // Converts desktop path to string
+  std::wstring ws = std::wstring(desktop);
+  std::string desktop_path = std::string(ws.begin(), ws.end());
+  desktop_path = desktop_path + "\\" + game_name + ".lnk";
+  LPCSTR pathLink = desktop_path.c_str();
+
+  HRESULT hres;
```
lioncash, line 666: Please join this with its assignment.

Source/Core/DolphinQt/GameList/GameList.cpp
```diff
@@ -628,6 +640,60 @@ void GameList::OpenGCSaveFolder()
...
+  IShellLink* psl;
+
+  CoInitialize(NULL);
+  hres =
+      CoCreateInstance(CLSID_ShellLink, NULL, CLSCTX_INPROC_SERVER, IID_IShellLink, (LPVOID*)&psl);
+
+  if (SUCCEEDED(hres))
+  {
+    IPersistFile* ppf;
```
lioncash, line 675: Ditto wrt ComPtr

Source/Core/DolphinQt/GameList/GameList.cpp
```diff
@@ -628,6 +640,60 @@ void GameList::OpenGCSaveFolder()
...
+    LPCWSTR pszArgs = args_convert.c_str();
+
+    psl->SetPath(dolphin_path_convert);
+    psl->SetArguments(pszArgs);
+
+    hres = psl->QueryInterface(IID_IPersistFile, (LPVOID*)&ppf);
```
lioncash, line 681: Consider using a different variable here instead of reassigning to the same variable.

Also consider using a C++ style cast here rather than a C-style one

Source/Core/DolphinQt/GameList/GameList.cpp
```diff
@@ -628,6 +640,60 @@ void GameList::OpenGCSaveFolder()
...
+  std::string desktop_path = std::string(ws.begin(), ws.end());
+  desktop_path = desktop_path + "\\" + game_name + ".lnk";
+  LPCSTR pathLink = desktop_path.c_str();
+
+  HRESULT hres;
+  IShellLink* psl;
+
+  CoInitialize(NULL);
+  hres =
+      CoCreateInstance(CLSID_ShellLink, NULL, CLSCTX_INPROC_SERVER, IID_IShellLink, (LPVOID*)&psl);
```
lioncash, line 671: nullptr, also consider using a C++ style cast here rather than a C-style one

Source/Core/DolphinQt/GameList/GameList.cpp
```diff
@@ -628,6 +640,60 @@ void GameList::OpenGCSaveFolder()
...
+    psl->SetPath(dolphin_path_convert);
+    psl->SetArguments(pszArgs);
+
+    hres = psl->QueryInterface(IID_IPersistFile, (LPVOID*)&ppf);
+
+    if (SUCCEEDED(hres))
+    {
+      WCHAR wsz[MAX_PATH];
+      MultiByteToWideChar(CP_ACP, 0, pathLink, -1, wsz, MAX_PATH);
+      hres = ppf->Save(wsz, TRUE);
```
lioncash, line 687: Consider using a different variable instead of assigning to a different one. This return value also isn't checked against--is this intentional?

Source/Core/DolphinQt/GameList/GameList.cpp
Aminoa: Would the args conversion also need to be altered (going from a string to a wstring)?

JosJuice: Yes. (I think I missed that line.)

Aminoa: I have changed the conversion method for the args as well

Source/Core/DolphinQt/GameList/GameList.cpp
Aminoa: If I attempt to use a different variable for QueryInterface, the compiler states that hres_check (a new variable made for QueryInterface) was nullptr.

JosJuice:
@dolphin-emu-bot rebuild

Source/Core/DolphinQt/GameList/GameList.cpp
```diff
@@ -53,6 +53,12 @@
 
 #include "UICommon/GameFile.h"
 
+#ifdef _WIN32
+#include "shlobj.h"
```
leoetlino, line 57:
```suggestion
#include <shlobj.h>
```

Source/Core/DolphinQt/GameList/GameList.cpp
```diff
@@ -53,6 +53,12 @@
...
+#include <qcoreapplication.h>
```
leoetlino, line 58:
```suggestion
#include <QCoreApplication>
```

Source/Core/DolphinQt/GameList/GameList.cpp
```diff
@@ -382,6 +388,9 @@ void GameList::ShowContextMenu(const QPoint&)
 
     menu->addAction(tr("Open &Containing Folder"), this, &GameList::OpenContainingFolder);
     menu->addAction(tr("Delete File..."), this, &GameList::DeleteFile);
+    #ifdef _WIN32
```
leoetlino, line 391: This file needs to be reformatted with clang-format

Source/Core/DolphinQt/GameList/GameList.h
```diff
@@ -65,6 +65,9 @@ class GameList final : public QStackedWidget
   void OpenWiki();
   void SetDefaultISO();
   void DeleteFile();
+   #ifdef _WIN32
```
leoetlino, line 68: clang-format

Source/Core/DolphinQt/GameList/GameList.cpp
```diff
@@ -628,6 +637,48 @@ void GameList::OpenGCSaveFolder()
     ModalMessageBox::information(this, tr("Information"), tr("No save data found."));
 }
 
+#ifdef _WIN32
+void GameList::AddShortcutToDesktop()
+{
+  const auto game = GetSelectedGame();
+  const auto game_name = game->GetLongName();
```
leoetlino, line 644:
```suggestion
  const auto& game_name = game->GetLongName();
```
Otherwise, you're copying the name for no reason. The next two lines need to be fixed too.

Source/Core/DolphinQt/GameList/GameList.cpp
```diff
@@ -628,6 +637,48 @@ void GameList::OpenGCSaveFolder()
...
+  const auto file_path = game->GetFilePath();
+  const auto file_name = game->GetFileName();
+
+  std::wstring args = UTF8ToTStr("-e \"" + file_path + "\"");
+  std::wstring dolphin_path = QCoreApplication::applicationFilePath().toStdWString();
+
+  // Gets the desktop position with shorcut
```
leoetlino, line 651:
```suggestion
  // Gets the desktop position with shortcut
```

Source/Core/DolphinQt/GameList/GameList.cpp
```diff
@@ -628,6 +637,48 @@ void GameList::OpenGCSaveFolder()
...
+
+    hres = psl->QueryInterface(IID_IPersistFile, static_cast<LPVOID*>(&ppf));
+
+    if (SUCCEEDED(hres))
+    {
+      hres = ppf->Save(desktop_path.c_str(), TRUE);
+    }
+  }
+
+  ModalMessageBox::information(this, tr("Add Shortcut to Desktop"),
```
leoetlino, line 677: Shouldn't this go inside the `if (SUCCEEDED(hres))` block?

Source/Core/DolphinQt/GameList/GameList.cpp
```diff
@@ -628,6 +637,48 @@ void GameList::OpenGCSaveFolder()
...
+  SHGetKnownFolderPath(FOLDERID_Desktop, KF_FLAG_NO_ALIAS, NULL, &desktop);
+  std::wstring desktop_path = std::wstring(desktop) + UTF8ToTStr("\\" + game_name + ".lnk");
+
+  HRESULT hres;
+  Microsoft::WRL::ComPtr<IShellLink> psl;
+
+  CoInitialize(NULL);
+  hres = CoCreateInstance(CLSID_ShellLink, nullptr, CLSCTX_INPROC_SERVER, IID_PPV_ARGS(&psl));
+
+  if (SUCCEEDED(hres))
```
leoetlino, line 662: Consider inverting the conditional to return early on failure to avoid unneeded indentation.

Source/Core/DolphinQt/GameList/GameList.cpp
```diff
@@ -628,6 +637,54 @@ void GameList::OpenGCSaveFolder()
     ModalMessageBox::information(this, tr("Information"), tr("No save data found."));
 }
 
+#ifdef _WIN32
+void GameList::AddShortcutToDesktop()
```
leoetlino, line 641: Consider returning a boolean and showing the message box in the caller instead to deduplicate the error message code.

The addAction call above would look like this:

```cpp
menu->addAction(tr("Add Shortcut to Desktop"), this, [this] {
  if (AddShortcutToDesktop())
  {
    ModalMessageBox::information(...);
  }
  else
  {
    ModalMessageBox::critical(...);
  }
});
```

Note that the error message box should use `critical` instead of `information` since it's an error.

leoetlino (commented):
You forgot two of lioncash's comments ^^

Source/Core/DolphinQt/GameList/GameList.cpp
```diff
@@ -628,6 +637,54 @@ void GameList::OpenGCSaveFolder()
...
+  SHGetKnownFolderPath(FOLDERID_Desktop, KF_FLAG_NO_ALIAS, NULL, &desktop);
+  std::wstring desktop_path = std::wstring(desktop) + UTF8ToTStr("\\" + game_name + ".lnk");
+
+  HRESULT hres;
+  Microsoft::WRL::ComPtr<IShellLink> psl;
+
+  CoInitialize(NULL);
+  hres = CoCreateInstance(CLSID_ShellLink, nullptr, CLSCTX_INPROC_SERVER, IID_PPV_ARGS(&psl));
+
+  if (!SUCCEEDED(hres))
```
leoetlino, line 661: I believe it's more idiomatic to use `FAILED`.

leoetlino:
A modified version of this was merged (#9438). Thanks for your contribution!

PR #9366, opened 2020-12-27
VideoCommon: allow Free Look to be manipulated by a quaternion

iwubcode:


Source/Core/Common/Matrix.cpp
```diff
@@ -56,6 +56,13 @@ Quaternion Quaternion::RotateZ(float rad)
   return Rotate(rad, Vec3(0, 0, 1));
 }
 
+Quaternion Quaternion::RotateXYZ(const Vec3& rads)
```
jordan-woyak, line 59: There should maybe be a comment that explains this produces a quaternion of simultaneous rotations about the 3 axes. e.g. processing gyroscope input.
I don't know if I like the name `RotateXYZ`, but I don't know what a better name is either. :P

iwubcode: Yeah, I'm horrible at naming things :D .  Please let me know if you think of something better!

Rumi-Larry: Here are some suggestions "Rotate_all_axes" "3D_rotation" "Quaternion_rotation"

leoetlino (changes requested):
Changes look sane, but the savestate version needs to be updated.

PR #9367, opened 2020-12-27
Collision

Goatman012019:
allow the rendering of invisible collision


leoetlino:
This is not a support forum or an issue tracker.

PR #9369, opened 2020-12-27
Android : Option to ignore Android command from controllers and/or better support for XInput controllers

Sosios:
I've noticed that on the Android version of Dolphin, pressing the B button on my XInput controllers brings up the onscreen menu and none of the buttons work even after being configured in the menu. It may be a compatibility issue with some devices though.


JosJuice:
This is not an issue tracker.

PR #9372, opened 2020-12-27
CMake: Update cmake_minimum_required calls using versions less than 2.8.12.

jordan-woyak:
My IDE was showing me CMake warnings and it was annoying.

"CMake Deprecation Warning at Externals/enet/CMakeLists.txt:1 (cmake_minimum_required):Compatibility with CMake < 2.8.12 will be removed from a future version of\nCMake.\n\nUpdate the VERSION argument <min> value or use a ...<max> suffix to tell\nCMake that the project does not need compatibility with older versions."

I've specified version 2.8.12 to eliminate the warnings.

### DIFF
--- a/Externals/enet/CMakeLists.txt
+++ b/Externals/enet/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.6)
+cmake_minimum_required(VERSION 2.8.12)
 
 project(enet)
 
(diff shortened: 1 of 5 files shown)


BhaaLseN:
Considering that CMake 3.0.0 is over 6 years old at this point, is there any particular reason to not simply say "you need at least 3.0"? Limiting this to a _maximum_ version just to get rid of a deprecation warning feels weird.

jordan-woyak:
> 
> 
> Considering that CMake 3.0.0 is over 6 years old at this point, is there any particular reason to not simply say "you need at least 3.0"? Limiting this to a _maximum_ version just to get rid of a deprecation warning feels weird.

I have no preference. I can do that instead.

PR #9375, opened 2020-12-28
BitUtils: Fix uint64_t gcc compile (Linux)

iWeaker4you:


Source/Core/Common/BitUtils.h
```diff
@@ -8,6 +8,7 @@
 #include <climits>
 #include <cstddef>
 #include <cstring>
+#include <cstdint> // include this header for uint64_t
```
lioncash, line 11: This should go above `<cstring>`, we also don't use comments to indicate what we use from the standard header

iWeaker4you: @lioncash Thanks, it's already corrected.

leoetlino (approved):

PR #9376, opened 2020-12-28
Jit_Integer: Optimize rlwimix

merryhime:
Last PR for a while.

The main purpose of this PR was the final commit, wherein we implement cases where mask is 0xFF or 0xFFFF with partial register writes.

A few additional commits for other things noticed along the way, these can be split into separate PRs if desired.

Sidenotes for future optimziations: 
* Using AH/BH/CH/DH is unlikely to be sufficiently profitable due to encoding limitations.
* Using pinsrb is unlikely to be profitable due to the cost of register moves.

### DIFF
--- a/Source/Core/Core/PowerPC/Jit64/Jit_Integer.cpp
+++ b/Source/Core/Core/PowerPC/Jit64/Jit_Integer.cpp
@@ -1771,11 +1771,7 @@ void Jit64::rlwnmx(UGeckoInstruction inst)
     RCOpArg Rs = gpr.Use(s, RCMode::Read);
     RegCache::Realize(Ra, Rs);
 
-    if (a != s)
-      MOV(32, Ra, Rs);
-
-    if (amount)
-      ROL(32, Ra, Imm8(amount));
+    RotateLeft(32, Ra, Rs, amount);
 
     // we need flags if we're merging the branch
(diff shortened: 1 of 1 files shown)


Source/Core/Core/PowerPC/Jit64/Jit_Integer.cpp
```diff
@@ -1735,32 +1735,47 @@ void Jit64::rlwimix(UGeckoInstruction inst)
...
       {
         RotateLeft(32, RSCRATCH, Rs, inst.SH);
-        XOR(32, R(RSCRATCH), Ra);
-        AndWithMask(RSCRATCH, mask);
-        XOR(32, Ra, R(RSCRATCH));
+      }
+
+      if (mask == 0xFF || mask == 0xFFFF)
+      {
+        MOV(mask == 0xFF ? 8 : 16, Ra, R(RSCRATCH));
```
Tilka, line 1751: Does this really work? I would have expected MOVZX.

MerryMage: Intended behaviour is `Ra &= 0xFFFF0000; Ra |= RSCRATCH & 0xFFFF`, movzx is not appropriate here.

JosJuice (approved):

PR #9377, opened 2020-12-28
DSPAnalyzer: Migrate off file-scope state

lioncash:
Turns the DSP analyzer into a class and situates it within the SDSP struct. With this, I believe all relevant state within the DSP code has been made non-global.

These changes were significantly easier to chunk out into parts for review, as opposed to the deglobalization efforts in DSPCore, as the file-scope state itself was already hidden behind an interface. This more-or-less allowed for "sliding" the class into place in an easier manner.

Each commit should explain what it does and any rationale for it if necessary.

### DIFF
--- a/Source/Core/Core/DSP/DSPAnalyzer.h
+++ b/Source/Core/Core/DSP/DSPAnalyzer.h
@@ -18,17 +18,6 @@ namespace DSP::Analyzer
 // Useful things to detect:
 // * Loop endpoints - so that we can avoid checking for loops every cycle.
 
-enum CodeFlags : u8
-{
-  CODE_NONE = 0,
-  CODE_START_OF_INST = 1,
-  CODE_IDLE_SKIP = 2,
-  CODE_LOOP_START = 4,
-  CODE_LOOP_END = 8,
-  CODE_UPDATE_SR = 16,
-  CODE_CHECK_EXC = 32,
-};
-
 class Analyzer
 {
 public:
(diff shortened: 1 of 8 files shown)


Source/Core/Core/DSP/Interpreter/DSPInterpreter.cpp
```diff
@@ -114,7 +116,7 @@ int Interpreter::RunCyclesDebug(int cycles)
       }
 
       // Idle skipping.
```
BhaaLseN, line 118: Do we still need this (and the other) comment now that it is a fairly descriptive method?

BhaaLseN (approved):
Code LGTM, untested; and that one remark is ignorable if we want to keep it.

Source/Core/Core/DSP/Interpreter/DSPInterpreter.cpp
lioncash: Nope, it's not necessary anymore. Removed it.

Source/Core/Core/DSP/DSPAnalyzer.h
```diff
@@ -20,22 +20,22 @@ namespace DSP
 class Analyzer
 {
 public:
-  explicit Analyzer(const SDSP& dsp);
+  explicit Analyzer();
   ~Analyzer();
 
   Analyzer(const Analyzer&) = default;
```
leoetlino, line 26: Since none of the {copy,move} {constructors,assignment operators} do anything special anymore, should we get rid of them entirely? I feel like defaulting them explicitly doesn't add much when the compiler already does the same thing with less code.

lioncash: These operators won't actually get generated by the compiler if a user-provided destructor is present. (see the conditions under "Implicitly-declared move constructor" [here](https://en.cppreference.com/w/cpp/language/move_constructor#Implicitly-declared_move_constructor). This also applies to the copy equivalents as well.

leoetlino (approved):

PR #9378, opened 2020-12-28
[RFC] Jit_FloatingPoint: Branchless FloatCompare

MerryMage:
Remove branches from FloatCompare.

We do this by using ZF:CF as an index into our required lookup tables.
We also notice that we can overlap our entries in our lookup tables as long as we properly mask the results afterwards.

### DIFF
--- a/Source/Core/Core/PowerPC/Jit64/Jit_FloatingPoint.cpp
+++ b/Source/Core/Core/PowerPC/Jit64/Jit_FloatingPoint.cpp
@@ -6,6 +6,7 @@
 #include <vector>
 
 #include "Common/Assert.h"
+#include "Common/BitUtils.h"
 #include "Common/CPUDetect.h"
 #include "Common/CommonTypes.h"
 #include "Common/x64Emitter.h"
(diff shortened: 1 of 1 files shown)


Source/Core/Core/PowerPC/Jit64/Jit_FloatingPoint.cpp
```diff
@@ -557,64 +559,56 @@ void Jit64::FloatCompare(UGeckoInstruction inst, bool upper)
...
+  // Please note that our arguments are REVERSED to normal. (See above UCOMISD instruction.)
+  //
+  //               x64 flags
+  //               ZF  PF  CF
+  // Unordered      1   1   1
+  // Less than      0   0   0
+  // Greater than   0   0   1
+  // Equal          1   0   0
+  //
+  // Thus we can take use ZF:CF as an index into an array.
```
Sintendo, line 571: I think you meant to write either 'take' or 'use', not both.

MerryMage:
3% performance impact on benchmarks, will revisit.

PR #9379, opened 2020-12-29
Fix latency field being initially enabled on audio backends not supporting it

Pokechu22:
On platforms where there is at least one audio backend supporting latency (e.g. Windows), the latency field was enabled on all backends, even if they don't support it, though it would become disabled after changing the backend.  This also happened after emulation is stopped.  The cause was `OnEmulationStateChanged` not checking if the backend supported latency; `OnEmulationStateChanged` is called when the audio pane is first constructed.


Filoppi:
True. There were quite a lot of bugs like that in the AudioPane, all I could found have been fixed here (including this), https://github.com/Filoppi/dolphin/blob/audio-input-fixes/Source/Core/DolphinQt/Settings/AudioPane.cpp, still being worked on. But this fix is nice to have now.

JosJuice (approved):

leoetlino (approved):

PR #9383, opened 2020-12-29
DSP: Eliminate some magic values related to the CR register

lioncash:
Makes some values more immediately readable.

### DIFF
--- a/Source/Core/Core/DSP/Jit/x64/DSPJitBranch.cpp
+++ b/Source/Core/Core/DSP/Jit/x64/DSPJitBranch.cpp
@@ -273,9 +273,9 @@ void DSPEmitter::rti(const UDSPInstruction opc)
 // HALT
 // 0000 0000 0020 0001
 // Stops execution of DSP code. Sets bit DSP_CR_HALT in register DREG_CR.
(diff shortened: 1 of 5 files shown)


Source/Core/Core/DSP/Jit/x64/DSPJitBranch.cpp
```diff
@@ -273,9 +273,9 @@ void DSPEmitter::rti(const UDSPInstruction opc)
 // HALT
 // 0000 0000 0020 0001
 // Stops execution of DSP code. Sets bit DSP_CR_HALT in register DREG_CR.
-void DSPEmitter::halt(const UDSPInstruction opc)
+void DSPEmitter::halt(const UDSPInstruction)
```
leoetlino, line 276: Since the argument isn't used at all, is there a point in keeping the const qualifier?

lioncash: I kept it for consistency with the rest of the instructions

leoetlino: makes sense

leoetlino (approved):

PR #9385, opened 2020-12-29
Jit_Integer: Use SHLX, SHRX, SARX

MerryMage:
Avoid unnecessary `MOV` instructions and `RCX` use when BMI2 is available.

### DIFF
--- a/Source/Core/Core/PowerPC/Jit64/Jit_Integer.cpp
+++ b/Source/Core/Core/PowerPC/Jit64/Jit_Integer.cpp
@@ -1870,6 +1870,16 @@ void Jit64::srwx(UGeckoInstruction inst)
         SHR(32, Ra, Imm8(amount));
     }
   }
+  else if (cpu_info.bBMI2)
+  {
+    RCX64Reg Ra = gpr.Bind(a, RCMode::Write);
+    RCX64Reg Rb = gpr.Bind(b, RCMode::Read);
+    RCX64Reg Rs = gpr.Bind(s, RCMode::Read);
+    RegCache::Realize(Ra, Rb, Rs);
+
+    // Rs must be in register: This is a 64-bit operation, using an OpArg will have invalid results
+    SHRX(64, Ra, Rs, Rb);
+  }
   else
   {
     RCX64Reg ecx = gpr.Scratch(ECX);  // no register choice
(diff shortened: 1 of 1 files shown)


MerryMage:
Implementing this exposed a bug where upper 32 bits of GPR registers were not zero (from existing JIT code it seems intentional for these bits to always be zero).

This bug exists in the existing implementation but is less commonly hit in the existing code because in order for it to be triggered it required a == s for a srwx instruction as well as a previous instruction sequence resulting in non-zero bits in the upper half of a GPR.

There are two ways to fix this:
(a) Never assume the upper half of a GPR is all zeros.
(b) Fix all code that violates the assumption that upper half of a GPR is all zeros.

MerryMage:
Fixed.

In srwx, allowing Rs to reference memory had resulted in a 64-bit load rather than the desired 32-bit load, resulting in invalid data in the upper half of the register which was then shifted to the right.

<0.5% speed improvement in povray.

Source/Core/Core/PowerPC/Jit64/Jit_Integer.cpp
```diff
@@ -2023,6 +2052,29 @@ void Jit64::srawx(UGeckoInstruction inst)
...
+      MOV(32, R(RSCRATCH), Ra);
+      SHR(64, Ra, Imm8(32));
+      TEST(32, Ra, R(RSCRATCH));
+    }
+    else
+    {
+      SHR(64, Ra, Imm8(32));
+    }
+    FinalizeCarry(CC_NZ);
+  }
```
Sintendo, line 2077: I think we can drop the `a != b` requirement and allow Rs to be an immediate if we do something like this:

```suggestion
  else if (cpu_info.bBMI2)
  {
    RCX64Reg Ra = gpr.Bind(a, RCMode::Write);
    RCX64Reg Rb = gpr.Bind(b, RCMode::Read);
    RCOpArg Rs = gpr.Use(s, RCMode::Read);
    RegCache::Realize(Ra, Rb, Rs);

    X64Reg tmp = RSCRATCH;
    if (a == s && a != b)
      tmp = Ra;
    else
      MOV(32, R(tmp), Rs);
    
    SHL(64, R(tmp), Imm8(32));
    SARX(64, Ra, R(tmp), Rb);
    if (js.op->wantsCA)
    {
      MOV(32, R(RSCRATCH), Ra);
      SHR(64, Ra, Imm8(32));
      TEST(32, Ra, R(RSCRATCH));
    }
    else
    {
      SHR(64, Ra, Imm8(32));
    }
    FinalizeCarry(CC_NZ);
  }
```

MerryMage:
@Sintendo Thanks, makes sense to me, applied.

Sintendo (approved):

lioncash (approved):

PR #9388, opened 2020-12-29
Set console's default language/country/region based on computer settings

JosJuice:
Marking as a draft until PR #9571 is merged, to ensure that this doesn't cause any nasty desyncs for netplay and movies. Ready for review otherwise.

### DIFF
--- a/Source/Core/Core/Config/DefaultLocale.cpp
+++ b/Source/Core/Core/Config/DefaultLocale.cpp
@@ -1,0 +1,199 @@
+// Copyright 2020 Dolphin Emulator Project
+// Licensed under GPLv2+
+// Refer to the license.txt file included.
+
+#include "Core/Config/DefaultLocale.h"
+
+#include <algorithm>
+#include <array>
+#include <locale>
+#include <optional>
+#include <string>
+#include <string_view>
+#include <vector>
+
+#ifdef _WIN32
+#include <Windows.h>
+#endif
+
+#include "Common/Assert.h"
+#include "Common/CommonTypes.h"
+#include "Common/StringUtil.h"
+#include "Core/Host.h"
+#include "DiscIO/Enums.h"
+
+namespace Config
+{
+static std::vector<std::string> GetPreferredLocales()
+{
+  static const std::vector<std::string> locales = Host_GetPreferredLocales();
+  return locales;
+}
+
+static std::optional<DiscIO::Language> TryParseLanguage(const std::string& locale)
+{
+  const std::vector<std::string> split_locale = SplitString(locale, '-');
+  if (split_locale.empty())
+    return std::nullopt;
+
+  // Special handling of Chinese due to its two writing systems
+  if (split_locale[0] == "zh")
+  {
+    const auto locale_contains = [&split_locale](std::string_view str) {
+      return std::find(split_locale.cbegin(), split_locale.cend(), str) != split_locale.cend();
+    };
+
+    if (locale_contains("Hans"))
+      return DiscIO::Language::SimplifiedChinese;
+    if (locale_contains("Hant"))
+      return DiscIO::Language::TraditionalChinese;
+
+    // Mainland China and Singapore use simplified characters
+    if (locale_contains("CN") || locale_contains("SG"))
+      return DiscIO::Language::SimplifiedChinese;
+    else
+      return DiscIO::Language::TraditionalChinese;
+  }
+
+  // Same order as in Wii SYSCONF
+  constexpr std::array<std::string_view, 10> LANGUAGES = {
+      "ja", "en", "de", "fr", "es", "it", "nl", "zh", "zh", "ko",
+  };
+
+  const auto it = std::find(LANGUAGES.cbegin(), LANGUAGES.cend(), split_locale[0]);
+  if (it == LANGUAGES.cend())
+    return std::nullopt;
+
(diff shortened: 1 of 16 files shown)


Source/Core/Core/Config/DefaultLocale.cpp
```diff
@@ -0,0 +1,201 @@
...
+{
+static std::vector<std::string> GetPreferredLocales()
+{
+  static const std::vector<std::string> locales = Host_GetPreferredLocales();
+  return locales;
+}
+
+static std::optional<DiscIO::Language> TryParseLanguage(const std::string& locale)
+{
+  const std::vector<std::string> split_locale = SplitString(locale, '-');
```
BhaaLseN, line 35: Are we sure we get something like `en-US` (with a dash) and not `en_US` (with an underscore) as `locale`?

JosJuice: Based on Qt's documentation and the tests I've done, it seems like we're only getting `-`. But I admittedly haven't tested macOS or Linux.

Source/Core/Core/Config/DefaultLocale.cpp
```diff
@@ -0,0 +1,201 @@
...
+{
+  static const std::optional<u8> country = ComputeDefaultCountry();
+  return country;
+}
+
+u8 GetDefaultCountry()
+{
+  constexpr u8 FALLBACK_COUNTRY = 0x6c;  // Switzerland, as per Dolphin tradition
+
+  const std::optional<u8> country = GetOptionalDefaultCountry();
```
leoetlino, line 191:
```suggestion
  return GetOptionalDefaultCountry().value_or(FALLBACK_COUNTRY);
```

Source/Core/Core/Config/DefaultLocale.cpp
```diff
@@ -0,0 +1,201 @@
...
+{
+  const std::optional<u8> country = GetDefaultCountry();
+
+  if (country)
+  {
+    const DiscIO::Region region = DiscIO::SysConfCountryToRegion(GetDefaultCountry());
+    ASSERT(region != DiscIO::Region::Unknown);
+    return region;
+  }
+  else
```
leoetlino, line 161: Unless the policy has changed recently, I think we typically omit else after a return

Source/Core/Core/Config/DefaultLocale.cpp
```diff
@@ -0,0 +1,201 @@
...
+static std::optional<DiscIO::Language> TryParseLanguage(const std::string& locale)
+{
+  const std::vector<std::string> split_locale = SplitString(locale, '-');
+  if (split_locale.empty())
+    return std::nullopt;
+
+  // Special handling of Chinese due to its two writing systems
+  if (split_locale[0] == "zh")
+  {
+    const auto locale_contains = [&split_locale](const std::string& str) {
```
leoetlino, line 42:
```suggestion
    const auto locale_contains = [&split_locale](std::string_view str) {
```

Source/Core/Core/Config/DefaultLocale.cpp
```diff
@@ -0,0 +1,201 @@
...
+      return std::find(split_locale.cbegin(), split_locale.cend(), str) != split_locale.cend();
+    };
+
+    if (locale_contains("Hans"))
+      return DiscIO::Language::SimplifiedChinese;
+    else if (locale_contains("Hant"))
```
leoetlino, line 48:
```suggestion
    if (locale_contains("Hant"))
```

Source/Core/Core/Config/DefaultLocale.cpp
```diff
@@ -0,0 +1,201 @@
...
+
+    // Mainline China and Singapore use simplified characters
+    if (locale_contains("CN") || locale_contains("SG"))
+      return DiscIO::Language::SimplifiedChinese;
+    else
+      return DiscIO::Language::TraditionalChinese;
+  }
+
+  constexpr std::array<std::string_view, 10> LANGUAGES = {
+      "ja", "en", "de", "fr", "es", "it", "nl", "zh", "zh", "ko",
```
leoetlino, line 59: Might be worth mentioning that this must match the Language enum. Though it might be obvious...

Source/Core/Core/Config/DefaultLocale.cpp
```diff
@@ -0,0 +1,201 @@
...
+{
+#ifdef _WIN32
+  // Windows codepath: Check the regional information.
+  // More likely to match the user's physical location than locales are.
+  // TODO: Can we use GetUserDefaultGeoName? (It was added in a Windows 10 update)
+  GEOID geo = GetUserGeoID(GEOCLASS_NATION);
+  const int buffer_size = GetGeoInfoW(geo, GEO_ISO2, nullptr, 0, 0);
+  std::vector<wchar_t> buffer(buffer_size);
+  const int result = GetGeoInfoW(geo, GEO_ISO2, buffer.data(), buffer_size, 0);
+  if (result)
```
leoetlino, line 103:
```suggestion
  if (result != 0)
```

leoetlino (approved):
Code changes look fine, and this works correctly on Linux.

iwubcode (approved):
Booted up and it seemed to pick the correct settings on Windows.  Code seems fine.

leoetlino:
Looks like a rebase is required

PR #9390, opened 2020-12-29
INI Changes

nf6429:
This PR deals with more Texture Cache Accuracy cases, and some Dualcore ones too. Might add a commit with bounding box fixes for demo discs, but I've only been able to test US versions, not PAL versions, so I can only safely do discs 20-22 and 35, not 17-19. For the PAL version I do have, the games don't match, so probably best to leave them alone for another PR.

### DIFF
--- a/Data/Sys/GameSettings/SAQ.ini
+++ b/Data/Sys/GameSettings/SAQ.ini
@@ -1,0 +1,5 @@
+# SAQE5G - Harley Pasternak's Hollywood Workout
+
+[Video_Settings]
+# Fixes game text.
+SafeTextureCacheColorSamples - 0
--- a/Data/Sys/GameSettings/SAQ.ini
+++ b/Data/Sys/GameSettings/SAQ.ini
@@ -2,4 +2,4 @@
 
 [Video_Settings]
 # Fixes game text.
-SafeTextureCacheColorSamples - 0
+SafeTextureCacheColorSamples = 0
--- a/Data/Sys/GameSettings/SW6.ini
+++ b/Data/Sys/GameSettings/SW6.ini
@@ -1,0 +1,5 @@
+# SW6P78, SW6E78 - WWE'12
+
+[Video_Settings]
+# Fixes Create a Superstar invisible items.
+SafeTextureCacheColorSamples = 0
--- a/Data/Sys/GameSettings/G6W.ini
+++ b/Data/Sys/GameSettings/G6W.ini
@@ -13,4 +13,5 @@
 # Add action replay cheats here.
 
 [Video_Settings]
-SafeTextureCacheColorSamples = 512
+# Fixes game text.
+SafeTextureCacheColorSamples = 0
--- a/Data/Sys/GameSettings/GEZ.ini
+++ b/Data/Sys/GameSettings/GEZ.ini
@@ -1,4 +1,4 @@
-# GEZE8P, GEZP8P - BillyHatcher
+# GEZE8P, GEZP8P - Billy Hatcher and the Giant Egg
 
 [Core]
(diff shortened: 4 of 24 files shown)


Data/Sys/GameSettings/SAQ.ini
```diff
@@ -0,0 +1,5 @@
+# SAQE5G - Harley Pasternak's Hollywood Workout
+
+[Video_Settings]
+# Fixes game text.
+SafeTextureCacheColorSamples - 0
```
Rumi-Larry, line 5: This should be a "=" not a "-"

nf6429: Well spotted.

JMC47:
Can you try using texture cache value 2048 instead of 0 in some of the games?  It's less performance intensive than 0 but more accurate than 512, so it may result in a net gain in performance while also being accurate enough to produce correct emulation.

The Dualcore changes are in titles that make sense to me.

nf6429:
> Can you try using texture cache value 2048 instead of 0 in some of the games? It's less performance intensive than 0 but more accurate than 512, so it may result in a net gain in performance while also being accurate enough to produce correct emulation.
> 
> The Dualcore changes are in titles that make sense to me.

I could, but main concern is how do I test which titles work with 2048, since 0= safe, 512 = middle, 128 = fast, how do i test it in dolphin?

JosJuice:
You have to type 2048 into the INI file and then test.

nf6429:
Had some issues with previous computer I was doing this on, and I'm not particularly good enough at git bash to do this PR in one commit all squashed (normally use git desktop), so I'll likely close this PR and create a new one when the changes are ready on another computer. Note that I tested the 2048 changes, 11 titles work with it.

Data/Sys/GameSettings/SW6.ini
```diff
@@ -0,0 +1,5 @@
+# SW6P78, SW6E78 - WWE'12
+
+[Video_Settings]
+# Fixes Create a Superstar invisible items.
```
Rumi-Larry, line 4: I would place the "Create a Superstar" phrase between quotes so it's more readable. Ditto on the othher WWE games.

nf6429: That makes sense, thanks.

Avasam:
@nf6429 
> [...] I'm not particularly good enough at git bash to do this PR in one commit all squashed (normally use git desktop),[...]

`git rebase -i HEAD^X` where X is the number of commits you want to rebase. The `-i` flag is for an interactive rebase. It will use your prefered editor configured in git (defaults to Vim iirc). You can configure it to use whatever tool you prefer (like Notepad++, or VSCode).

Speaking of VSCode, since you seem to prefer using Git with a GUI, it has a nice looking rebase window. Here's how to configure git to use VSCode: https://code.visualstudio.com/docs/editor/versioncontrol#_vs-code-as-git-editor . VSCode + GitLens extension has completely replace any other git tools for me.
And here's for Notepad++: `git config --global core.editor "'C:/Program Files/Notepad++/notepad++.exe' -multiInst -notabbar -nosession -noPlugin"`

JosJuice:
HEAD means the commit you are currently at, and HEAD^X means the commit that's X commits before the current commit. The X is just a placeholder, you have to replace it with a number.

nf6429:
@JosJuice Don't want to bother you, but this is my first time rebasing in git, and as you can see commit wise, its very weird. All I did in the interactive rebase was squash fix spelling error into INI changes and its made 4 commits. I have no idea what I have done

nf6429:
I've got the branch ready for a new PR, I'll close this one since the history is messed up from me figuring out git.

PR #9391, opened 2020-12-29
IOS: Allow for heterogenous name lookup

lioncash:
Allows lookups to be done with std::string_view or any other string type. This allows for non-allocating strings to be used with the name lookup without needing to construct a std::string.

While we're at it, we can make use of `insert_or_assign` in AddDevice to elide unnecessary default constructions (not like it particularly matters, however it's essentially a "free" change, so why not).

### DIFF
--- a/Source/Core/Core/IOS/IOS.h
+++ b/Source/Core/Core/IOS/IOS.h
@@ -10,6 +10,7 @@
 #include <memory>
 #include <mutex>
 #include <string>
+#include <string_view>
 #include <vector>
 
 #include "Common/CommonTypes.h"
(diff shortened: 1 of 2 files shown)


Source/Core/Core/IOS/IOS.h
```diff
@@ -110,14 +110,14 @@ class Kernel
   void AddDevice(std::unique_ptr<Device::Device> device);
   void AddCoreDevices();
   void AddStaticDevices();
-  std::shared_ptr<Device::Device> GetDeviceByName(const std::string& device_name);
+  std::shared_ptr<Device::Device> GetDeviceByName(std::string_view device_name);
```
leoetlino, line 113: This is missing a `<string_view>` include too :p

leoetlino (approved):

PR #9392, opened 2020-12-29
DSP: Fix write masks on AUDIO_*/AR_* MMIO registers

smurf3tte:
https://bugs.dolphin-emu.org/issues/6749

This change fixes the scratchy audio in Teenage Mutant Ninja Turtles (SX7E52/SX7P52). The game starts an audio interface DMA with an unaligned address, and because Dolphin was not masking off the low 5 bits of AUDIO_DMA_START_LO, all future AI DMAs were misaligned. To understand why, it is instructive to refer to AUDIO_InitDMA() in libogc, which behaves the same as the official SDK:

`_dspReg[25] = (_dspReg[25]&~0xffe0)|(startaddr&0xffff);`

The implementation does not mask off the low bits of the passed in value before it ORs them with low bits of the current register value. Therefore, if they are not masked off by the hardware itself, they become permanently stuck once set.

Adding a write mask for AUDIO_DMA_START_LO is enough to fix the bug in TMNT, but I decided to run some tests on GC and Wii to find the correct write masks for the surrounding registers, as only a couple were already being masked. Dolphin has gotten away with not masking the rest because many are already A) masked on read (or never read) by the SDK and/or B) masked on use (or never used) in Dolphin.

This leaves just three registers where the difference may be observable: AR_DMA_CNT_H and AUDIO_DMA_START_HI/LO.

### DIFF
--- a/Source/Core/Core/HW/DSP.cpp
+++ b/Source/Core/Core/HW/DSP.cpp
@@ -231,30 +231,51 @@ void Shutdown()
 
 void RegisterMMIO(MMIO::Mapping* mmio, u32 base)
 {
+  static constexpr u16 WMASK_NONE = 0x0000;
+  static constexpr u16 WMASK_AR_INFO = 0x007f;
+  static constexpr u16 WMASK_AR_REFRESH = 0x07ff;
+  static constexpr u16 WMASK_AR_HI_RESTRICT = 0x03ff;
+  static constexpr u16 WMASK_AR_CNT_DIR_BIT = 0x8000;
+  static constexpr u16 WMASK_AUDIO_HI_RESTRICT_GCN = 0x03ff;
+  static constexpr u16 WMASK_AUDIO_HI_RESTRICT_WII = 0x1fff;
+  static constexpr u16 WMASK_LO_ALIGN_32BIT = 0xffe0;
+
   // Declare all the boilerplate direct MMIOs.
   struct
   {
     u32 addr;
     u16* ptr;
-    bool align_writes_on_32_bytes;
+    u16 wmask;
   } directly_mapped_vars[] = {
-      {AR_INFO, &s_ARAM_Info.Hex},
-      {AR_MODE, &s_AR_MODE},
-      {AR_REFRESH, &s_AR_REFRESH},
-      {AR_DMA_MMADDR_H, MMIO::Utils::HighPart(&s_arDMA.MMAddr)},
-      {AR_DMA_MMADDR_L, MMIO::Utils::LowPart(&s_arDMA.MMAddr), true},
-      {AR_DMA_ARADDR_H, MMIO::Utils::HighPart(&s_arDMA.ARAddr)},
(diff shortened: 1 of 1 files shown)


Source/Core/Core/HW/DSP.cpp
```diff
@@ -231,30 +231,51 @@ void Shutdown()
 
 void RegisterMMIO(MMIO::Mapping* mmio, u32 base)
 {
+  constexpr u16 WMASK_NONE = 0x0000;
+  constexpr u16 WMASK_AR_INFO = 0X007f;
+  constexpr u16 WMASK_AR_REFRESH = 0X07ff;
+  constexpr u16 WMASK_AR_HI_RESTRICT = 0x03ff;
+  constexpr u16 WMASK_AR_CNT_DIR_BIT = 0x8000;
+  constexpr u16 WMASK_LO_ALIGN_32BIT = 0xffe0;
+  const u16 WMASK_AUDIO_HI_RESTRICT = SConfig::GetInstance().bWii ? 0x1fff : 0x03ff;
```
JosJuice, line 240: Does this work as intended if you launch a GameCube game from the Wii menu? (I'm not sure which behavior is actually intended)

Source/Core/Core/HW/DSP.cpp
```diff
@@ -231,30 +231,51 @@ void Shutdown()
 
 void RegisterMMIO(MMIO::Mapping* mmio, u32 base)
 {
+  constexpr u16 WMASK_NONE = 0x0000;
+  constexpr u16 WMASK_AR_INFO = 0X007f;
+  constexpr u16 WMASK_AR_REFRESH = 0X07ff;
```
JosJuice, line 236: These two should use a lowercase x to match the rest.

Source/Core/Core/HW/DSP.cpp
smurf3tte: The MMIO handlers are only registered once, so the Wii behavior will be inherited in that case. If that's wrong, then CommandProcessor::RegisterMMIO() is likely wrong as well.

JMC47:
Fixes The Bachelor for Wii as well.

Source/Core/Core/HW/DSP.cpp
smurf3tte: Good catch. I have now verified that a Wii in GC mode behaves like a real GC. I've deferred the mask selection to time of use to properly support this scenario.

Someone should audit the other uses of bWii throughout Dolphin. The usage in CommandProcessor, at least, is very likely incorrect for this case...

BhaaLseN (approved):
Code looks good, untested

Silanda:
It looks like this fixes Dragon's Lair Trilogy too: https://bugs.dolphin-emu.org/issues/12349

delroth:
LGTM, configmed with @smurf3tte that this was hardware tested and that the register masking does happen when writing and not in the DMA engine itself.

PR #9393, opened 2020-12-29
SysConf: Make use of std::string_view 

lioncash:
Allows strings to be used in potentially non-allocating manners.


leoetlino (approved):

leoetlino (commented):
Whoops, I actually missed the fact that `<string_view>` needs to be included

leoetlino (approved):
the new commits LGTM too

PR #9380, opened 2020-12-29
Most Wanted turn on the sun

ivan89el:
Restores the sun as on the original GameCube: https://sun9-61.userapi.com/impg/idHG7PQSWoBMtpkVO424oIbQIdQb2YJskX2djA/xgS4EFRc8ts.jpg?size=1920x1080&quality=96&proxy=1&sign=eefdd4db144526388784a849e42fee73&type=album

PR #9395, opened 2020-12-30
GameSettings: Enable perf queries to fix sun in Need for Speed: Most Wanted.

AdmiralCurtiss:
Rebase of #9380 with proper commit message. Mostly because I genuinely smiled at the PR name of 'turn on the sun'.

### DIFF
--- a/Data/Sys/GameSettings/GOW.ini
+++ b/Data/Sys/GameSettings/GOW.ini
@@ -2,3 +2,6 @@
 
 [Core]
 JITFollowBranch = False
+
+[Video]
+PerfQueriesEnable = True


AdmiralCurtiss:
Note that I haven't actually tested this, but since this exact fix is suggested on the wiki I assume it's correct: https://wiki.dolphin-emu.org/index.php?title=Need_for_Speed%3A_Most_Wanted#Sun_Halo

iwubcode:
@JMC47 do you have this game to confirm working behavior?

JMC47:
I can confirm that Perf Queries produces the correct effect, however I am worried about the potential performance impact on weaker devices.

JMC47:
Let us turn on the sun.

PR #9396, opened 2020-12-30
JitArm64_RegCache: Interface cleanup

lioncash:
Tidies up the interface in various minor ways, such as:

- Marking member funcs as const when they don't modify internal state
- Converting enums to enum classes
- In-class initializing member variables where applicable.

Converting the RegType enum over to an enum class in particular uncovered an odd quirk in two places. RegType enum members were being passed as the second argument to `BindToRegister` [here](https://github.com/dolphin-emu/dolphin/commit/8659a4d6d02dc352d2c2e41f39d485d4efbf1ca1#diff-bc54c6351f0ce20b932a4aa98992e19a69cdc5c55d5be79f839b25c8bf4b6d7cR60) and [here](https://github.com/dolphin-emu/dolphin/compare/master...lioncash:arm?expand=1#diff-bc54c6351f0ce20b932a4aa98992e19a69cdc5c55d5be79f839b25c8bf4b6d7cR152) which is very unusual, as the second member of `BindToRegister` is a `bool`, not a `RegType`, which seems pretty sketchy. I don't know if this is an actual bug rather than identifiers being abused however, so I'd appreciate input on that.

### DIFF
--- a/Source/Core/Core/PowerPC/JitArm64/Jit.cpp
+++ b/Source/Core/Core/PowerPC/JitArm64/Jit.cpp
@@ -144,8 +144,8 @@ void JitArm64::Shutdown()
 void JitArm64::FallBackToInterpreter(UGeckoInstruction inst)
 {
   FlushCarry();
-  gpr.Flush(FlushMode::FLUSH_ALL, js.op);
-  fpr.Flush(FlushMode::FLUSH_ALL, js.op);
+  gpr.Flush(FlushMode::All, js.op);
+  fpr.Flush(FlushMode::All, js.op);
 
   if (js.op->opinfo->flags & FL_ENDBLOCK)
   {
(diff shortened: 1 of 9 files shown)


BhaaLseN (approved):
Enum classes feel a lot nicer for those. Untested.

leoetlino (approved):
The REG_REG thing seems to be an oversight... It was added as part of https://github.com/dolphin-emu/dolphin/commit/bcde1aa8ff934278c004b59722d09ea855daf295#diff-bc54c6351f0ce20b932a4aa98992e19a69cdc5c55d5be79f839b25c8bf4b6d7cL132 which changed a lot of `fpr.R(..., false);` to `fpr.R(..., REG_REG);`. I bet it's a find-and-replace error :P

PR #9398, opened 2020-12-30
Add safety checking and features to FixedSizeQueue

Filoppi:
-Added safety checking.
-Added optimize copy from and to.
-Changed from int to size_t (more in line with std and it was giving annoying warnings).
-Added [] index access operator.
-Added optimized erase of multiple items.
-Added a few helper functions.

Note that there is a test for this class already (FixedSizeQueueTest.cpp). These changes are needed for the new DPLII implementation and serve as optimizations there.
FixedSizeQueue is only used for logs in the whole project so the changes don't affect anything.


delroth:
As you mentioned, there are tests for this, could you update them to also test the new APIs you are adding?

Also squash your typo fix commit.

Filoppi:
> As you mentioned, there are tests for this, could you update them to also test the new APIs you are adding?
> 
> Also squash your typo fix commit.

Done both things.

Source/Core/Common/FixedSizeQueue.h
```diff
@@ -45,6 +42,46 @@ class FixedSizeQueue
     tail = (tail + 1) % N;
   }
 
+  // Copies over an array, loops over if num is greater than max_size
+  void push_array(const T* t, size_t num)
+  {
+    size_t back_writable_num = std::min(tail_to_end(), num);
+    memcpy(&back(), t, sizeof(T) * back_writable_num);
```
leoetlino, line 49: This is undefined behavior is T is not trivially copyable. (Also, this should be std::memcpy, not just memcpy)

Source/Core/Common/FixedSizeQueue.h
```diff
@@ -45,6 +42,46 @@ class FixedSizeQueue
...
+    t += back_writable_num;
+    size_t readable_num = num - back_writable_num;
+    while (readable_num > 0)
+    {
+      size_t beginning_writable_num = std::min(readable_num, N);
+      memcpy(&beginning(), t, sizeof(T) * beginning_writable_num);
```
leoetlino, line 55: Also possibly UB.

Source/Core/Common/FixedSizeQueue.h
```diff
@@ -45,6 +42,46 @@ class FixedSizeQueue
...
+
+  // Takes a max num to copy (from head to tail), returns the actual copied number.
+  // Doesn't loop over if num is greater than our size
+  size_t copy_to_array(T* t, size_t num)
+  {
+    num = std::min(num, count);
+    const size_t front_readable_num = std::min(head_to_end(), num);
+    const size_t beginning_readable_num = num - front_readable_num;
+
+    memcpy(&t[0], &front(), sizeof(T) * front_readable_num);
```
leoetlino, line 79: Here too.

Source/Core/Common/FixedSizeQueue.h
```diff
@@ -45,6 +42,46 @@ class FixedSizeQueue
...
+    if (beginning_readable_num > 0)
+      memcpy(&t[front_readable_num], &beginning(), sizeof(T) * beginning_readable_num);
```
leoetlino, line 81: And here.

Source/Core/Common/FixedSizeQueue.h
```diff
@@ -57,8 +94,10 @@ class FixedSizeQueue
     tail = (tail + 1) % N;
   }
 
+  // From front
   void pop()
   {
+    assert(count > 0);
```
leoetlino, line 100: Dolphin code should use `ASSERT` (from Common/Assert.h).

Source/Core/Common/FixedSizeQueue.h
```diff
@@ -73,15 +112,58 @@ class FixedSizeQueue
     return temp;
   }
 
+  // From front
+  void erase(size_t num)
+  {
+    assert(count >= num);
```
leoetlino, line 118: ASSERT

Source/Core/Common/FixedSizeQueue.h
```diff
@@ -73,15 +112,58 @@ class FixedSizeQueue
...
+  {
+    assert(index < count);
+    return storage[(head + index) % N];
+  }
+
   T& front() noexcept { return storage[head]; }
   const T& front() const noexcept { return storage[head]; }
+  T& back() noexcept { return storage[tail]; }
+  const T& back() const noexcept { return storage[tail]; }
+  // Returns the first index of the internal array. Only use if head_to_end() > size()
```
leoetlino, line 152: Do we really want this to be part of the public API? Seems easy to misuse and I can't think of any situation where external code would need to use this.

Source/UnitTests/Common/FixedSizeQueueTest.cpp
```diff
@@ -30,6 +30,11 @@ TEST(FixedSizeQueue, Simple)
   EXPECT_EQ(1004, q.pop_front());
 
   EXPECT_EQ(0u, q.size());
+
+  q.push(0);
```
leoetlino, line 34: IMO, this should be part of another test ("Clear"?)

Source/UnitTests/Common/FixedSizeQueueTest.cpp
```diff
@@ -38,18 +43,48 @@ TEST(FixedSizeQueue, RingBuffer)
...
+  EXPECT_EQ(5u, q.max_size());
 
+  EXPECT_EQ(1, q.pop_front());
   EXPECT_EQ(4u, q.size());
+  q.erase(2);
+  EXPECT_EQ(2u, q.size());
+  q.erase(2);
+  EXPECT_EQ(0u, q.size());
+
+  // Copy over an array bigger than us, which means the copy will loop over
```
leoetlino, line 70: This should be in another test ("Copy"?).

Source/UnitTests/Common/FixedSizeQueueTest.cpp
```diff
@@ -38,18 +43,48 @@ TEST(FixedSizeQueue, RingBuffer)
...
+  q.clear();
+  int other_array[6] = {1, 2, 3, 4, 5, 6};
```
leoetlino, line 72: std::array<int, 6>

Source/UnitTests/Common/FixedSizeQueueTest.cpp
```diff
@@ -38,18 +43,48 @@ TEST(FixedSizeQueue, RingBuffer)
...
+  q.push_array(other_array, 6);
```
leoetlino, line 73:
```suggestion
  q.push_array(other_array.data(), other_array.size());
```

leoetlino (changes requested):

Source/Core/Common/FixedSizeQueue.h
```diff
@@ -45,6 +42,46 @@ class FixedSizeQueue
     tail = (tail + 1) % N;
   }
 
+  // Copies over an array, loops over if num is greater than max_size
+  void push_array(const T* t, size_t num)
```
leoetlino, line 46: To be honest, I'm not convinced this complicated logic is required. Do you have profiling data that shows that calling `emplace()` in loop is a bottleneck?

Source/Core/Common/FixedSizeQueue.h
Filoppi: we can just put a static assert for that then?

Source/Core/Common/FixedSizeQueue.h
Filoppi: I think I made these before adding the push_array and copy_to_array methods, as I was doing the same thing but in some external code. They aren't strictly needed sure, but I think they would be nice to have if someone wanted to do some direct memory copying which isn't natively implemented.

Source/Core/Common/FixedSizeQueue.h
leoetlino: No, this class is used with non-trivially copyable types (like a pair of std::string and [whatever], and std::string is already not trivially copyable).

Source/Core/Common/FixedSizeQueue.h
Filoppi: I don't have a profiling no, and this is definitely not mandatory to have, but once implemented it's there, so, why not?
I made this for the DPLII which given the new sample rates implementations in my other draft audio PR, could end up copying 2688000 floats per second, so I though a memcopy would have been better.

Source/UnitTests/Common/FixedSizeQueueTest.cpp
Filoppi: ok

Source/Core/Common/FixedSizeQueue.h
leoetlino: Because it's possibly a premature optimisation and as the other review comments show it's easy to miss edge cases like T not being trivially copyable.

If this is not even a bottleneck, it's a lot of code that does not have to exist and be maintained.

Source/Core/Common/FixedSizeQueue.h
Filoppi: Never mind, what I said made no sense.
Are you saying that it's undefined behaviour because the destructor on overwritten items is not being called?
Or because I'm not calling the copy constructor? Or both?

leoetlino: Both. The memcpy itself is not UB but you'll almost certainly run into UB later, when you try to use those incorrectly copied strings, or when they get destructed and you get a double free, etc.

Filoppi: There are 3 solutions I can think of:
-Make a different version which just falls back on "push()".
-Make a child class which only takes trivial types, where these methods are implemented.
-Put an assert if the type is non trivial.

Filoppi: I opted for std::enable_if. Works great.

Source/Core/Common/FixedSizeQueue.h
Filoppi: Over the max number of floats DPLII can copy per audio frame, this is from ~17% to ~40% faster in release on MSVC v14.28.
I've tested a lot of times with microseconds accuracy. The difference goes bigger as the copy num grows of course.

PR #9399, opened 2020-12-30
JitArm64_LoadStore: Explicitly annotate switch fallthrough cases

lioncash:
Makes it explicit that these are intentional. Also prevents compiler warnings.


JosJuice:
You missed one for `lhzux`.

JosJuice (approved):

PR #9262, opened 2020-11-19
Jit64: More constant propagation optimizations

Sintendo:
It is possible to generate more optimal instruction sequences for various instructions if some or all of the input registers are known to hold a specific value.

---
**subfic**
<details><summary>Example</summary>

Before:
```
BE 03 00 00 00       mov         esi,3
83 EE 08             sub         esi,8
0F 93 45 58          setae       byte ptr [rbp+58h]
```
After:
```
C6 45 58 00          mov         byte ptr [rbp+58h],0
```
</details>

---

**srwx**
<details><summary>Example 1</summary>

Before:
```
B9 18 00 00 00       mov         ecx,18h
45 8B C1             mov         r8d,r9d
49 D3 E8             shr         r8,cl
```
After:
```
45 8B C1             mov         r8d,r9d
41 C1 E8 18          shr         r8d,18h
```
</details>
<details><summary>Example 2</summary>

Before:
```
B9 00 00 00 00       mov         ecx,0
45 8B C1             mov         r8d,r9d
49 D3 E8             shr         r8,cl
```
After:
```
45 8B C1             mov         r8d,r9d
```
</details>
<details><summary>Example 3</summary>

Before:
```
B9 F8 FF FF FF       mov         ecx,0FFFFFFF8h
45 8B C1             mov         r8d,r9d
49 D3 E8             shr         r8,cl
```
After:
Nothing, register is set to constant zero.
</details>

---
**slwx**

<details><summary>Example 1</summary>

Before:
```
B9 18 00 00 00       mov         ecx,18h
41 8B F7             mov         esi,r15d
48 D3 E6             shl         rsi,cl
8B F6                mov         esi,esi
```
After:
```
41 8B CF             mov         ecx,r15d
C1 E1 18             shl         ecx,18h
```
</details>
<details><summary>Example 2</summary>

Before:
```
BE 00 00 00 00       mov         esi,0
8B CE                mov         ecx,esi
41 8B F3             mov         esi,r11d
48 D3 E6             shl         rsi,cl
8B F6                mov         esi,esi
```
After:
```
41 8B FB             mov         edi,r11d
```
</details>
<details><summary>Example 3</summary>

Before:
```
BE F8 FF FF FF       mov         esi,0FFFFFFF8h
8B CE                mov         ecx,esi
41 8B F4             mov         esi,r12d
48 D3 E6             shl         rsi,cl
8B F6                mov         esi,esi
```
After:
Nothing, register is set to constant zero.
</details>

---

**srawx**
[...]

PR #9400, opened 2020-12-30
JitArm64: More constant propagation optimizations

JosJuice:
PR #9262 added a bunch of Jit64 optimizations, some of which were already in JitArm64 and some which weren't. This PR ports the latter ones to JitArm64.

### DIFF
--- a/Source/Core/Core/PowerPC/JitArm64/JitArm64_Integer.cpp
+++ b/Source/Core/Core/PowerPC/JitArm64/JitArm64_Integer.cpp
@@ -1494,7 +1494,7 @@ void JitArm64::srawx(UGeckoInstruction inst)
 
     MOVI2R(WA, 32);
     SUB(WC, WA, WC);
-    LSL(WC, RS, WC);
+    LSLV(WC, RS, WC);
     CMP(WC, 0);
     CSET(WA, CC_NEQ);
     FixupBranch end = B();
--- a/Source/Core/Core/PowerPC/JitArm64/JitArm64_Integer.cpp
+++ b/Source/Core/Core/PowerPC/JitArm64/JitArm64_Integer.cpp
@@ -30,6 +30,24 @@ void JitArm64::ComputeRC0(u64 imm)
     SXTW(gpr.CR(0), DecodeReg(gpr.CR(0)));
 }
 
+void JitArm64::ComputeCarry(ARM64Reg reg)
+{
+  js.carryFlagSet = false;
+
+  if (!js.op->wantsCA)
+    return;
+
+  if (CanMergeNextInstructions(1) && js.op[1].wantsCAInFlags)
+  {
+    CMP(reg, 1);
+    js.carryFlagSet = true;
+  }
+  else
+  {
+    STRB(IndexType::Unsigned, reg, PPC_REG, PPCSTATE_OFF(xer_ca));
+  }
+}
+
 void JitArm64::ComputeCarry(bool Carry)
 {
   js.carryFlagSet = false;
@@ -1252,6 +1270,12 @@ void JitArm64::slwx(UGeckoInstruction inst)
     if (inst.Rc)
       ComputeRC0(gpr.GetImm(a));
   }
+  else if (gpr.IsImm(s) && gpr.GetImm(s) == 0)
+  {
+    gpr.SetImmediate(a, 0);
+    if (inst.Rc)
+      ComputeRC0(0);
+  }
   else if (gpr.IsImm(b))
   {
     u32 i = gpr.GetImm(b);
@@ -1338,7 +1362,6 @@ void JitArm64::srawx(UGeckoInstruction inst)
   JITDISABLE(bJITIntegerOff);
 
   int a = inst.RA, b = inst.RB, s = inst.RS;
-  bool inplace_carry = CanMergeNextInstructions(1) && js.op[1].wantsCAInFlags;
 
   if (gpr.IsImm(b) && gpr.IsImm(s))
   {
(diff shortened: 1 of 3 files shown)


Source/Core/Core/PowerPC/JitArm64/JitArm64_Integer.cpp
```diff
@@ -1353,16 +1376,57 @@ void JitArm64::srawx(UGeckoInstruction inst)
...
+  }
+  else if (gpr.IsImm(b))
   {
     int amount = gpr.GetImm(b);
-    if (amount & 0x20)
-      amount = 0x1F;
+
+    if (amount == 0)
+    {
+      ComputeCarry(false);
```
Sintendo, line 1393: 1. By checking for zero before applying any masking, we may miss out on some cases where the entire register may not be zero, but the bits that matter for the shift are. Probably doesn't happen too often in practice, though.
2. Don't we at least need to do a MOV from gpr.R(s) to gpr.R(a) in the shift-by-zero case? (Disclaimer, not terribly familiar with ARM64...)

JosJuice: > Don't we at least need to do a MOV from gpr.R(s) to gpr.R(a) in the shift-by-zero case? (Disclaimer, not terribly familiar with ARM64...)

Oops, yes :D

JosJuice: > By checking for zero before applying any masking, we may miss out on some cases where the entire register may not be zero, but the bits that matter for the shift are. Probably doesn't happen too often in practice, though.

By masking, you mean masking out the top bit of `amount`, right? If so, it doesn't seem correct to me to do this check after masking.

Sintendo: > By masking, you mean masking out the top bit of `amount`, right? If so, it doesn't seem correct to me to do this check after masking.

Yes.

https://github.com/dolphin-emu/dolphin/blob/master/Source/Core/Core/PowerPC/Interpreter/Interpreter_Integer.cpp#L310

Only the 6 lowest bits of the register matter, the rest can be ignored.

JosJuice: Ah, right, this isn't about doing the `amount & 0x1f` masking, but rather the `amount & 0x3f` masking. I see now. Thanks!

Source/Core/Core/PowerPC/JitArm64/JitArm64_Integer.cpp
```diff
@@ -1359,16 +1382,68 @@ void JitArm64::srawx(UGeckoInstruction inst)
...
+      }
+    }
+    else if (amount == 0)
+    {
+      if (a != s)
+      {
+        gpr.BindToRegister(a, false);
+
+        ADDS(gpr.R(a), gpr.R(s), 0);
+        ComputeCarry();
```
degasus, line 1419: Carry is always false here. So you can use MOV.

JosJuice: I figured this was an efficient way to set the carry flag in case the next instruction wants it in the host carry flag. But I should probably make it use MOV instead when `!js.op->wantsCA`.

Source/Core/Core/PowerPC/JitArm64/JitArm64_Integer.cpp
```diff
@@ -1359,16 +1382,68 @@ void JitArm64::srawx(UGeckoInstruction inst)
...
-    ASR(gpr.R(a), gpr.R(s), amount);
+    {
+      gpr.BindToRegister(a, a == s);
+
+      ARM64Reg WA = gpr.GetReg();
+
+      LSL(WA, gpr.R(s), 32 - amount);
+      ASR(gpr.R(a), gpr.R(s), amount);
+      TST(WA, gpr.R(a));
+      CSET(WA, CC_NEQ);
```
degasus, line 1441: After thinking 10 minutes about this algorithm, I think we should have a small comment why this actually is the valid result:
LSL shifts the hidden bits to the MSB (like a rotation) with zero in the LSBs
ASR duplicates the sign bit
bitwise AND of both yields if any high bit was shifted out of the range && the sign bit was set.

IMO a very clever approach, and I assume we should use so in the lower code as well (the lower code was generated with gcc....)

However this code here can be optimized quite a bit:
TST (alias for ANDS) is a shifted register instruction. higher cost, however it includes a LSL on the second operand for free. Use it.

It hurts quite a bit to emit CSET/CMP just to move the carry from Z to C, but I have not found a better approach :(

degasus: Untested, but feel free to pick it: https://gist.github.com/degasus/d837423ad814ff741802243e9520ecd2

Source/Core/Core/PowerPC/JitArm64/JitArm64_Integer.cpp
```diff
@@ -1359,16 +1382,68 @@ void JitArm64::srawx(UGeckoInstruction inst)
...
+
+      ComputeCarry(WA);
+
+      gpr.Unlock(WA);
+    }
```
degasus, line 1446: Might you also fix the `LSL(WC, RS, WC);` below?

Source/Core/Core/PowerPC/JitArm64/JitArm64_Integer.cpp
JosJuice: We discussed this briefly on IRC the day the comment was made, and the best solution long-term is probably to implement a way for the following instruction to read the carry flag at compile-time. So I've made this use `MOV` plus `ComputeCarry(false)` in all cases.

Source/Core/Core/PowerPC/JitArm64/JitArm64_Integer.cpp
JosJuice: The shifted variant of TST is one cycle slower than the normal variant (at least on the CPUs I checked), so I don't think this approach is worth it if we have to use a MOV. But I've made us use it when `a != s`.

PR #9402, opened 2020-12-31
Arm64Emitter: Interface cleanup

lioncash:
Tidies up the interface a little by annotating switch fallthrough to avoid warnings. While we're at it we can convert quite a few enums into enum classes to prevent polluting namespaces and making their usage strongly typed.


BhaaLseN (approved):
Code LGTM

leoetlino (approved):

PR #9405, opened 2020-12-31
Rename "Use Fullscreen" setting to "Start in Fullscreen"

Filoppi:
The name was confusing as changing it at runtime would not change the window to fullscreen, as it effectively only affects the start of the emulation.
Also blocked the ability to change it when the emulation is running, to be more inline with other similar settings, like "Render to main Window".

### DIFF
--- a/Source/Core/DolphinQt/Config/Graphics/GeneralWidget.cpp
+++ b/Source/Core/DolphinQt/Config/Graphics/GeneralWidget.cpp
@@ -61,7 +61,7 @@ void GeneralWidget::CreateWidgets()
                          Config::GFX_ASPECT_RATIO);
   m_adapter_combo = new ToolTipComboBox;
   m_enable_vsync = new GraphicsBool(tr("V-Sync"), Config::GFX_VSYNC);
-  m_enable_fullscreen = new GraphicsBool(tr("Use Fullscreen"), Config::MAIN_FULLSCREEN);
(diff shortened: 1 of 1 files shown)


Rumi-Larry:
Hmm, why not make it so the setting changes the window. I assume it's not technically viable, given that one can always use the key combo.

Filoppi:
> Hmm, why not make it so the setting changes the window? I assume it's not technically viable, given that one can always use the key combo.

I don't know, I don't like the idea of making a window that isn't the active/focused one fullscreen when you press a button in another window. Also it just seems unnecessary, we already have the hotkey, users should use that (if they don't know how to enter fullscreen, they wouldn't know how to exit it).

MayImilae:
I'm pretty happy with this change.

So, we got the "Use Fullscreen" language from games, cause well, that's how they put it. However, it isn't that great of a fit for us. In games, the button is just, check mark it to make it fullscreen, uncheck it to make it windowed. That's it. But for us, it's way more complex, since we have a separate interface and main window and the render window can go in and out of fullscreen at any time with hotkeys or the button in the interface.

"Start in Fullscreen" is a much better way to phrase it for us, since well, that's how it works in Dolphin! Also, greying it out after emulation has started is good too, especially thanks to the increase clarity of the naming.

So yea, this is great!

MayImilae (approved):

PR #9408, opened 2020-12-31
SSL: Workaround to remove SNI from ClientHello

sepalani:
This PR is a **dirty** workaround to remove the SNI extension from ClientHello. This PR is an alternative to https://github.com/dolphin-emu/dolphin/pull/9406.

It does that by using the only hook that I could find that can be used between handshake steps (see `mbedtls_ssl_handshake_step`). My approach uses the send callback called in `mbedtls_ssl_flush_output` at the beginning of each handshake step. I delayed `mbedtls_ssl_set_hostname` after the client hello as the hostname is needed during the common name verification when connecting to an IP address.

This PR needs testing to be sure there isn't any regression introduced and that there isn't a better way to achieve this goal

### DIFF
--- a/Source/Core/Core/IOS/Network/SSL.cpp
+++ b/Source/Core/Core/IOS/Network/SSL.cpp
@@ -35,6 +35,39 @@ static constexpr mbedtls_x509_crt_profile mbedtls_x509_crt_profile_wii = {
     0,         /* No RSA min key size */
 };
 
+namespace
+{
+// Dirty workaround to disable SNI which isn't supported by the Wii.
+//
+// This SSL extension can ONLY be disabled by undefining
+// MBEDTLS_SSL_SERVER_NAME_INDICATION and recompiling the library. When
+// enabled and if the hostname is set, it uses the SNI extension which is sent
+// with the Client Hello message.
+//
+// This workaround doesn't require recompiling the library. It does so by
+// deferring mbedtls_ssl_set_hostname after the Client Hello message. The send
+// callback is used as it's the (only?) hook called at the beginning of
+// each step of the handshake by the mbedtls_ssl_flush_output function.
+//
+// The hostname still needs to be set as it is checked against the Common Name
+// field during the certificate verification process.
+int SSLSendWithoutSNI(void* ctx, const unsigned char* buf, size_t len)
+{
+  auto* ssl = static_cast<IOS::HLE::WII_SSL*>(ctx);
+
+  if (ssl->ctx.state == MBEDTLS_SSL_SERVER_HELLO)
+    mbedtls_ssl_set_hostname(&ssl->ctx, ssl->hostname.c_str());
+  return mbedtls_net_send(&ssl->hostfd, buf, len);
+}
+
+int SSLRecv(void* ctx, unsigned char* buf, size_t len)
+{
+  auto* ssl = static_cast<IOS::HLE::WII_SSL*>(ctx);
+
+  return mbedtls_net_recv(&ssl->hostfd, buf, len);
+}
+}  // namespace
+
 NetSSL::NetSSL(Kernel& ios, const std::string& device_name) : Device(ios, device_name)
 {
   for (WII_SSL& ssl : _SSL)
@@ -223,8 +256,6 @@ IPCCommandResult NetSSL::IOCtlV(const IOCtlVRequest& request)
       mbedtls_ssl_conf_renegotiation(&ssl->config, MBEDTLS_SSL_RENEGOTIATION_ENABLED);
 
       ssl->hostname = hostname;
-      mbedtls_ssl_set_hostname(&ssl->ctx, ssl->hostname.c_str());
-
       ssl->active = true;
       WriteReturnValue(freeSSL, BufferIn);
     }
@@ -445,7 +476,7 @@ IPCCommandResult NetSSL::IOCtlV(const IOCtlVRequest& request)
       WiiSockMan& sm = WiiSockMan::GetInstance();
       ssl->hostfd = sm.GetHostSocket(ssl->sockfd);
       INFO_LOG_FMT(IOS_SSL, "IOCTLV_NET_SSL_CONNECT socket = {}", ssl->sockfd);
-      mbedtls_ssl_set_bio(&ssl->ctx, &ssl->hostfd, mbedtls_net_send, mbedtls_net_recv, nullptr);
+      mbedtls_ssl_set_bio(&ssl->ctx, ssl, SSLSendWithoutSNI, SSLRecv, nullptr);
       WriteReturnValue(SSL_OK, BufferIn);
     }
     else
--- a/Source/Core/Core/IOS/Network/SSL.h
+++ b/Source/Core/Core/IOS/Network/SSL.h
@@ -74,8 +74,8 @@ struct WII_SSL
   mbedtls_x509_crt cacert;
   mbedtls_x509_crt clicert;
   mbedtls_pk_context pk;
-  int sockfd;
-  int hostfd;
+  int sockfd = -1;
+  int hostfd = -1;
   std::string hostname;
   bool active;
 };
--- a/Source/Core/Core/IOS/Network/Socket.cpp
+++ b/Source/Core/Core/IOS/Network/Socket.cpp
@@ -457,14 +457,14 @@ void WiiSocket::Update(bool read, bool write, bool except)
           }
           case IOCTLV_NET_SSL_WRITE:
           {
-            int ret = mbedtls_ssl_write(&Device::NetSSL::_SSL[sslID].ctx,
-                                        Memory::GetPointer(BufferOut2), BufferOutSize2);
+            WII_SSL* ssl = &Device::NetSSL::_SSL[sslID];
+            const int ret =
+                mbedtls_ssl_write(&ssl->ctx, Memory::GetPointer(BufferOut2), BufferOutSize2);
 
             if (ret >= 0)
             {
-              PowerPC::debug_interface.NetworkLogger()->LogSSLWrite(
-                  Memory::GetPointer(BufferOut2), ret,
-                  static_cast<mbedtls_net_context*>(Device::NetSSL::_SSL[sslID].ctx.p_bio)->fd);
+              PowerPC::debug_interface.NetworkLogger()->LogSSLWrite(Memory::GetPointer(BufferOut2),
+                                                                    ret, ssl->hostfd);
               // Return bytes written or SSL_ERR_ZERO if none
               WriteReturnValue((ret == 0) ? SSL_ERR_ZERO : ret, BufferIn);
             }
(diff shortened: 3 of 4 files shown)


Source/Core/Core/IOS/Network/SSL.cpp
```diff
@@ -21,6 +21,12 @@
 #include "Core/HW/Memmap.h"
 #include "Core/IOS/Network/Socket.h"
 
+namespace
+{
+// Dirty workaround to remove SNI from Client Hello
+int mbedtls_net_send_without_sni(void* ctx, const unsigned char* buf, size_t len);
```
lioncash, line 27: Any reason that this anonymous namespace cant be within the device namespace? (and also have the implementation there as well)

Maybe also expand the comment to explain why the workaround needs to exist.

Source/Core/Core/IOS/Network/SSL.cpp
```diff
@@ -572,3 +577,23 @@ IPCCommandResult NetSSL::IOCtlV(const IOCtlVRequest& request)
   return GetDefaultReply(IPC_SUCCESS);
 }
 }  // namespace IOS::HLE::Device
+
+namespace
+{
+int mbedtls_net_send_without_sni(void* ctx, const unsigned char* buf, size_t len)
+{
+  mbedtls_ssl_context* mbed_ssl_ctx = reinterpret_cast<mbedtls_ssl_context*>(ctx);
```
lioncash, line 585:
```suggestion
  auto* mbed_ssl_ctx = static_cast<mbedtls_ssl_context*>(ctx);
```

Source/Core/Core/IOS/Network/SSL.cpp
```diff
@@ -572,3 +577,23 @@ IPCCommandResult NetSSL::IOCtlV(const IOCtlVRequest& request)
...
+
+  if (mbed_ssl_ctx->state == MBEDTLS_SSL_SERVER_HELLO)
+  {
+    const auto& ssl_table = IOS::HLE::Device::NetSSL::_SSL;
+    auto it = std::find_if(std::begin(ssl_table), std::end(ssl_table),
+                           [=](const IOS::HLE::WII_SSL& ssl) { return &ssl.ctx == ctx; });
```
lioncash, line 591:
```suggestion
                           [ctx](const IOS::HLE::WII_SSL& ssl) { return &ssl.ctx == ctx; });
```

sepalani:
@lioncash Done.

Source/Core/Core/IOS/Network/SSL.cpp
```diff
@@ -35,6 +35,40 @@ static constexpr mbedtls_x509_crt_profile mbedtls_x509_crt_profile_wii = {
     0,         /* No RSA min key size */
 };
 
+namespace
+{
+// Dirty workaround to disable SNI.
```
leoetlino, line 40: This comment should probably explain why SNI is being disabled.

sepalani: Done.

Source/Core/Core/IOS/Network/SSL.cpp
```diff
@@ -35,6 +35,40 @@ static constexpr mbedtls_x509_crt_profile mbedtls_x509_crt_profile_wii = {
...
+// enabled, it uses the SNI extension which is sent with the Client Hello
+// message.
+//
+// This workaround defers mbedtls_ssl_set_hostname after the Client Hello. It
+// uses the send callback as it's the (only?) hook used at the beginning of
+// each step during the handshake. If the hostname is NULL, SNI isn't sent
+// with the Client Hello message.
+//
+// However, the hostname needs to be set as it is checked against the Common
+// Name field during the certificate verification process.
```
Rumi-Larry, line 53: I recommend the following wording: 
`// The Wii doesn't support SNI but the only way to disable it is to disable SSL. 
// We do this by undefining MBEDTLS_SSL_SERVER_NAME_INDICATION and recompiling the library `
This is followed by the third and fourth paragraphs unedited.
In my opinion this is a more concise and clear explanation.

Leseratte10: But it would be wrong, wouldn't it? The point of this PR instead of #9406 is that we *don't* recompile the library with that flag disabled.

Rumi-Larry: Oof, shout I delete that comment? I feel embarrased 😣

sepalani: You don't have to. The wording wasn't clear enough, so I'll rewrite it to something better (hopefully). I'm open to suggestions/better phrasing.

Source/Core/Core/IOS/Network/SSL.cpp
```diff
@@ -35,6 +35,40 @@ static constexpr mbedtls_x509_crt_profile mbedtls_x509_crt_profile_wii = {
...
+namespace
+{
+// Dirty workaround to disable SNI which isn't supported by the Wii.
+//
+// This SSL extension can ONLY be disabled by undefining
+// MBEDTLS_SSL_SERVER_NAME_INDICATION and recompiling the library. When
+// enabled and if the hostname is set, it uses the SNI extension which is sent
+// with the Client Hello message.
+//
+// This workaround doesn't require to recompile the library. It does so by
```
leoetlino, line 47:
```suggestion
// This workaround doesn't require recompiling the library. It does so by
```

Source/Core/Core/IOS/Network/SSL.cpp
```diff
@@ -35,6 +35,40 @@ static constexpr mbedtls_x509_crt_profile mbedtls_x509_crt_profile_wii = {
...
+// defering mbedtls_ssl_set_hostname after the Client Hello message. The send
```
leoetlino, line 48:
```suggestion
// deferring mbedtls_ssl_set_hostname after the Client Hello message. The send
```

Source/Core/Core/IOS/Network/SSL.cpp
```diff
@@ -35,6 +35,40 @@ static constexpr mbedtls_x509_crt_profile mbedtls_x509_crt_profile_wii = {
...
+// callback is used as it's the (only?) hook called at the beginning of
+// each step of the handshake by the mbedtls_ssl_flush_output function.
+//
+// The hostname needs to be set as it is checked against the Common Name field
```
leoetlino, line 52:
```suggestion
// The hostname still needs to be set as it is checked against the Common Name field
```

Source/Core/Core/IOS/Network/SSL.cpp
```diff
@@ -35,6 +35,40 @@ static constexpr mbedtls_x509_crt_profile mbedtls_x509_crt_profile_wii = {
...
+// during the certificate verification process.
+int mbedtls_net_send_without_sni(void* ctx, const unsigned char* buf, size_t len)
```
leoetlino, line 54: The function name makes this look like it's an mbedtls function. Can we rename this to avoid any possible confusion? (Maybe `SendWithoutSNI`?)

leoetlino (changes requested):

Source/Core/Core/IOS/Network/SSL.cpp
```diff
@@ -445,7 +477,7 @@ IPCCommandResult NetSSL::IOCtlV(const IOCtlVRequest& request)
       WiiSockMan& sm = WiiSockMan::GetInstance();
       ssl->hostfd = sm.GetHostSocket(ssl->sockfd);
       INFO_LOG_FMT(IOS_SSL, "IOCTLV_NET_SSL_CONNECT socket = {}", ssl->sockfd);
-      mbedtls_ssl_set_bio(&ssl->ctx, &ssl->hostfd, mbedtls_net_send, mbedtls_net_recv, nullptr);
+      mbedtls_ssl_set_bio(&ssl->ctx, &ssl->hostfd, SSLSendWithoutSNI, mbedtls_net_recv, nullptr);
```
leoetlino, line 480: The context/userdata pointer (second argument) should be `&ssl->ctx`, not `&ssl->hostfd`.

Better yet, I'd recommend passing `ssl` directly so you don't have to look up the mbedtls SSL context manually.

Source/Core/Core/IOS/Network/SSL.cpp
```diff
@@ -35,6 +35,40 @@ static constexpr mbedtls_x509_crt_profile mbedtls_x509_crt_profile_wii = {
...
+  {
+    const auto& ssl_table = IOS::HLE::Device::NetSSL::_SSL;
+    auto it = std::find_if(std::begin(ssl_table), std::end(ssl_table),
+                           [ctx](const IOS::HLE::WII_SSL& ssl) { return &ssl.ctx == ctx; });
+    if (it != std::end(ssl_table))
+      mbedtls_ssl_set_hostname(mbed_ssl_ctx, it->hostname.c_str());
+    else
+      ERROR_LOG(IOS_SSL, "Failed to apply SNI hostname workaround!");
+  }
+  return mbedtls_net_send(ctx, buf, len);
```
leoetlino, line 68: This is incorrect. mbedtls_net_send should be passed `&ssl->hostfd` (the socket), not the SSL context.

leoetlino (changes requested):

leoetlino:
Is it guaranteed that mbedtls will honour the hostname setting and still verify the common name if the hostname is set along the way?

sepalani:
@leoetlino Done, it's way cleaner that way, indeed.

I don't think this is guaranteed since it's a dirty workaround. As long as the used library code is similar to one in Externals it should be safe because :
 - `mbedtls_ssl_set_hostname` only set the hostname in the SSL context
 - the hostname doesn't seem to be inspected/copied/used prior to this (except for SNI)

leoetlino (approved):
Looks more acceptable now I guess, and the workaround is clearly documented.

leoetlino:
@Leseratte10 Could you verify if this PR fixes your issue?

Leseratte10:
I can confirm that using the version in this PR gets rid of the server_name extensions as planned, yeah. I did not verify whether mbedtls still correctly validates the Common Name with these changes. If that's needed as well I would need to make a certificate with a wrong name and see what happens. 

When I compiled the changes from sepalani's 'sni' branch, Dolphin for some reason displayed / used a version number of 4.0-22756 instead of a 5.0-xxxxx version, that doesn't seem like it's related to the changes in this PR - no idea why that happened.

JosJuice:
Seems like the 5.0 tag hasn't been pushed to sepalani's repo. It should have no ill effects other than the version number being wrong.

leoetlino (changes requested):
Sorry, missed this but the UI code needs to be updated as it currently assumes ctx.p_bio is a mbedtls_next_context. Which is true on master, but this PR breaks the assumption since p_bio is a pointer to a WII_SSL now.

There should probably be a function that takes a WII_SSL and returns the associated mbedtls_net_context.

sepalani:
@leoetlino I opted for a simpler approach. I don't need to grab `p_bio` from `mbedtls_net_context` since I already have access to the `WII_SSL` context and that's what is passed to `p_bio`.

leoetlino (approved):

PR #9409, opened 2020-12-31
Make WiiSave::Import() behave closer to the Wii System Menu's SD Card save copying.

AdmiralCurtiss:
In practice, this means:

- ~~Saves are now only allowed to be imported if the game was previously 'seen' by ES. This includes launching it or even just inserting the game into the disc drive while the Wii Menu is running.~~ Turns out this was previously the case too, but in a rather implicit fashion. This is now more explicit and works like the Wii Menu where it reinstalls a TMD if it's in the `tmds.sys` backup file. (which can happen if you had a save once and then delete it)
- Saves are now deleted before a new one is imported. This prevents a few potential oddities like ending up with mixed old and new files, or [completely failing the import on saves with subdirectories](https://forums.dolphin-emu.org/Thread-problem-with-importing-game-save-file).

Please scrutinize this, I'm not entirely sure if everything here is correct and makes sense.

### DIFF
--- a/Source/Core/Core/HW/WiiSave.cpp
+++ b/Source/Core/Core/HW/WiiSave.cpp
@@ -477,28 +477,28 @@ StoragePointer MakeDataBinStorage(IOS::HLE::IOSC* iosc, const std::string& path,
   return StoragePointer{new DataBinStorage{iosc, path, mode}};
 }
 
-bool Copy(Storage* source, Storage* dest)
+CopyResult Copy(Storage* source, Storage* dest)
 {
   // first make sure we can read all the data from the source
   const auto header = source->ReadHeader();
   if (!header)
   {
     ERROR_LOG_FMT(CORE, "WiiSave::Copy: Failed to read header");
-    return false;
+    return CopyResult::CorruptedSource;
   }
 
   const auto bk_header = source->ReadBkHeader();
   if (!bk_header)
   {
     ERROR_LOG_FMT(CORE, "WiiSave::Copy: Failed to read bk header");
-    return false;
+    return CopyResult::CorruptedSource;
   }
 
   const auto files = source->ReadFiles();
   if (!files)
   {
     ERROR_LOG_FMT(CORE, "WiiSave::Copy: Failed to read files");
-    return false;
+    return CopyResult::CorruptedSource;
   }
 
   // once we have confirmed we can read the source, erase corresponding save in the destination
@@ -507,7 +507,7 @@ bool Copy(Storage* source, Storage* dest)
     if (!dest->EraseSave())
     {
       ERROR_LOG_FMT(CORE, "WiiSave::Copy: Failed to erase existing save");
-      return false;
+      return CopyResult::Error;
     }
   }
 
@@ -515,25 +515,25 @@ bool Copy(Storage* source, Storage* dest)
   if (!dest->WriteHeader(*header))
   {
     ERROR_LOG_FMT(CORE, "WiiSave::Copy: Failed to write header");
-    return false;
+    return CopyResult::Error;
   }
 
   if (!dest->WriteBkHeader(*bk_header))
   {
     ERROR_LOG_FMT(CORE, "WiiSave::Copy: Failed to write bk header");
-    return false;
+    return CopyResult::Error;
   }
 
   if (!dest->WriteFiles(*files))
   {
     ERROR_LOG_FMT(CORE, "WiiSave::Copy: Failed to write files");
-    return false;
+    return CopyResult::Error;
   }
 
-  return true;
+  return CopyResult::Success;
 }
 
-bool Import(const std::string& data_bin_path, std::function<bool()> can_overwrite)
+CopyResult Import(const std::string& data_bin_path, std::function<bool()> can_overwrite)
 {
   IOS::HLE::Kernel ios;
   const auto data_bin = MakeDataBinStorage(&ios.GetIOSC(), data_bin_path, "rb");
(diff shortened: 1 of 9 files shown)


Source/Core/Core/WiiUtils.cpp
```diff
@@ -223,6 +224,70 @@ bool IsTitleInstalled(u64 title_id)
...
+    return std::nullopt;
+
+  // structure of this file is as follows:
+  // - 32 bytes descriptor of a TMD, which contains a title ID and a length
+  // - the TMD, with padding aligning to 32 bytes
+  // - repeat for as many TMDs as stored
+  while (true)
+  {
+    std::array<u8, 32> descriptor;
+    if (!file->Read(descriptor.data(), 32))
```
lioncash, line 248:
```suggestion
    if (!file->Read(descriptor.data(), descriptor.size()))
```

Source/Core/Core/WiiUtils.cpp
```diff
@@ -223,6 +224,70 @@ bool IsTitleInstalled(u64 title_id)
...
+      return std::nullopt;
+
+    u64 tid = Common::swap64(descriptor.data());
```
lioncash, line 251: This and `tmd_length` can be const

Source/Core/Core/HW/WiiSaveStructs.h
```diff
@@ -101,7 +101,8 @@ class Storage
   };
 
   virtual ~Storage() = default;
-  virtual bool SaveExists() { return true; }
+  virtual bool SaveExists() = 0;
```
lioncash, line 104: Given this code is being modified, would it make sense to make this const qualified?

Source/Core/Core/HW/WiiSave.cpp
```diff
@@ -70,7 +71,37 @@ class NandStorage final : public Storage
 
   bool SaveExists() override
   {
-    return m_uid && m_gid && m_fs->GetMetadata(*m_uid, *m_gid, m_data_dir + "/banner.bin");
+    return !m_files_list.empty() ||
+           (m_uid && m_gid && m_fs->GetMetadata(*m_uid, *m_gid, m_data_dir + "/banner.bin"));
+    ;
```
BhaaLseN, line 76: Did you mean to leave this stray semicolon in here?

iwubcode:
> Saves are now only allowed to be imported if the game was previously 'seen' by ES. This includes launching it or even just inserting the game into the disc drive while the Wii Menu is running.

I might not be understanding this PR or maybe I'm misremembering but I thought that was already enforced?

I'll add this here, since I mentioned it on IRC a couple months ago but we really shouldn't require titles to have been launched / inserted.  And here's my story as to why:

A couple months ago, I thought that my NAND was corrupted.  The system menu wasn't able to be launched but games could play normally.  To fix this, I backed up my saves, intending to import them after creating a new NAND.  However, I was blocked from doing so because of the requirement to have used the title first.  With over 100 games, this wasn't a great option.

So again, we really need a way to solve this requirement.  Doesn't have to be fixed here but I wanted to mention it..  on IRC we had brainstormed some ways to resolve this but I can't recall the solutions that were mentioned.

AdmiralCurtiss:
If it was already enforced than it either wasn't explicitly or it was somewhere other than the importing logic itself. I'll see if I can find it.

I agree that our UX of Wii NAND handling is very poor -- I would personally like to have a GCI folder-like approach where a temp NAND is build from scratch with the correct data when launching games from the game list, mostly to get rid of years of built-up cruft in many users' NANDs ranging from older Dolphin versions not handling stuff correctly to just plain user error that leads to very hard to diagnose (for users) issues. However, that is far outside the scope of this issue with how the Wii handles its saves.

AdmiralCurtiss:
re: already-seen requirement, I checked and what happens in the old version is that it this query here fails:

https://github.com/dolphin-emu/dolphin/blob/f45a4a59168b571a573cce66169eae3b51dc4357/Source/Core/Core/HW/WiiSave.cpp#L205

Which then leads to the rest of the functions returning immediately cause they have no UID/GID. Yes, I suppose that is an enforcement, if not a very obvious one.

Source/Core/DolphinQt/MenuBar.cpp
```diff
@@ -1066,19 +1066,35 @@ void MenuBar::ImportWiiSave()
...
+  const auto result = WiiSave::Import(file.toStdString(), can_overwrite);
+  switch (result)
+  {
+  case WiiSave::CopyResult::Success:
+    ModalMessageBox::information(this, tr("Save Import"), tr("Successfully imported save file."));
+    break;
+  case WiiSave::CopyResult::CorruptedSource:
+    ModalMessageBox::critical(this, tr("Save Import"),
+                              tr("Failed to import save file. The given file appears to be "
+                                 "corrupted or not a valid Wii save."));
```
leoetlino, line 1085: "or is not"

Source/Core/Core/WiiUtils.cpp
```diff
@@ -223,6 +224,67 @@ bool IsTitleInstalled(u64 title_id)
...
+  return entries && std::any_of(entries->begin(), entries->end(),
+                                [](const std::string& file) { return file == "title.tmd"; });
+}
+
+IOS::ES::TMDReader FindBackupTMD(IOS::HLE::FS::FileSystem& fs, u64 title_id)
+{
+  auto file = fs.OpenFile(IOS::PID_KERNEL, IOS::PID_KERNEL,
+                          "/title/00000001/00000002/data/tmds.sys", IOS::HLE::FS::Mode::Read);
+  if (!file)
+    return IOS::ES::TMDReader();
```
leoetlino, line 239: or `{}` for brevity -- I think we tend to `return {};` for failure paths rather than write the return type explicitly (which is a bit verbose when there are several returns...)

leoetlino (approved):

Adamillo:
I don't know if this update caused this to happen or not but here's my theory: After I downloaded this update yesterday all the save data for my Wii games got WIPED! Maybe it's because of this update, or maybe not, but the fact is: Wii save data gone after I updated!

JMC47:
This just changes the save import function.  If your saves somehow got wiped, it would have to be something else.

Adamillo:
Still sad that my save data is gone :(

JMC47:
Do you have any idea what happened to it?  Obviously any time savedata gets removed is a problem.  Do you know what builds you jumped between?  Do you have the NANDs to see what's left in them?

PR #9410, opened 2021-01-01
Arm64Emitter: Convert ArithOption enums into enum classes

lioncash:
Makes the enums strongly typed. While we're at it, we can also make these enums private.

### DIFF
--- a/Source/Core/Common/Arm64Emitter.h
+++ b/Source/Core/Common/Arm64Emitter.h
@@ -382,34 +382,33 @@ enum class BarrierType
 
 class ArithOption
 {
-public:
-  enum WidthSpecifier
-  {
-    WIDTH_DEFAULT,
-    WIDTH_32BIT,
-    WIDTH_64BIT,
-  };
-
-  enum ExtendSpecifier
-  {
-    EXTEND_UXTB = 0x0,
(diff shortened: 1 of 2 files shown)


iwubcode (approved):

Source/Core/Common/Arm64Emitter.h
```diff
@@ -474,18 +473,18 @@ class ArithOption
...
       return (m_shifttype << 22) | (m_shift << 10);
-      break;
     default:
       DEBUG_ASSERT_MSG(DYNA_REC, false, "Invalid type in GetData");
       break;
     }
     return 0;
   }
+
+  bool IsExtended() const { return GetType() == TypeSpecifier::ExtendedReg; }
```
MerryMage, line 487: Perhaps remove ArithOption::GetType since TypeSpecifier is private.

BhaaLseN (approved):

Source/Core/Common/Arm64Emitter.h
lioncash: Ah, good catch. Removed it entirely, since it was just a simple getter

leoetlino (approved):

PR #9412, opened 2021-01-01
Arm64Emitter: Improve MOVI2R

JosJuice:
More or less a complete rewrite of the function which aims to be equally good or better for each given input, without relying on special cases like the old implementation did.

In particular, we now have more extensive support for MOVN, as mentioned in a TODO comment.

### DIFF
--- a/Source/Core/Common/Arm64Emitter.cpp
+++ b/Source/Core/Common/Arm64Emitter.cpp
@@ -2040,6 +2040,7 @@ void ARM64XEmitter::MOVI2RImpl(ARM64Reg Rd, T imm)
     MOVN,
     ADR,
     ADRP,
+    ORR,
   };
 
   struct Part
@@ -2055,6 +2056,12 @@ void ARM64XEmitter::MOVI2RImpl(ARM64Reg Rd, T imm)
 
   SmallVector<Part, max_parts> best_parts;
   Approach best_approach;
+  u64 best_base;
+
+  const auto instructions_required = [](const SmallVector<Part, max_parts>& parts,
+                                        Approach approach) {
+    return parts.size() + (approach > Approach::MOVN);
+  };
 
   const auto try_base = [&](T base, Approach approach, bool first_time) {
     SmallVector<Part, max_parts> parts;
@@ -2068,32 +2075,52 @@ void ARM64XEmitter::MOVI2RImpl(ARM64Reg Rd, T imm)
         parts.emplace_back(imm_shifted, static_cast<ShiftAmount>(i));
     }
 
-    if (first_time || parts.size() < best_parts.size())
+    if (first_time ||
+        instructions_required(parts, approach) < instructions_required(best_parts, best_approach))
     {
       best_parts = std::move(parts);
       best_approach = approach;
+      best_base = base;
     }
   };
 
+  // Try MOVZ/MOVN
+  try_base(T(0), Approach::MOVZ, true);
+  try_base(~T(0), Approach::MOVN, false);
+
+  // Try PC-relative approaches
   const auto sext_21_bit = [](u64 x) {
     return static_cast<s64>((x & 0x1FFFFF) | (x & 0x100000 ? ~0x1FFFFF : 0));
   };
-
   const u64 pc = reinterpret_cast<u64>(GetCodePtr());
   const s64 adrp_offset = sext_21_bit((imm >> 12) - (pc >> 12)) << 12;
   const s64 adr_offset = sext_21_bit(imm - pc);
   const u64 adrp_base = (pc & ~0xFFF) + adrp_offset;
   const u64 adr_base = pc + adr_offset;
-
-  // First: Try approaches for which instruction_count = max(parts.size(), 1)
-  try_base(T(0), Approach::MOVZ, true);
-  try_base(~T(0), Approach::MOVN, false);
-
-  // Second: Try approaches for which instruction_count = parts.size() + 1
   if constexpr (sizeof(T) == 8)
   {
     try_base(adrp_base, Approach::ADRP, false);
     try_base(adr_base, Approach::ADR, false);
+  }
+
+  // Try ORR (or skip it if we already have a 1-instruction encoding - these tests are non-trivial)
+  if (instructions_required(best_parts, best_approach) > 1)
+  {
+    if constexpr (sizeof(T) == 8)
+    {
+      for (u64 orr_imm : {(imm << 32) | (imm & 0x0000'0000'FFFF'FFFF),
+                          (imm & 0xFFFF'FFFF'0000'0000) | (imm >> 32),
+                          (imm << 48) | (imm & 0x0000'FFFF'FFFF'0000) | (imm >> 48)})
+      {
+        if (IsImmLogical(orr_imm, 64))
+          try_base(orr_imm, Approach::ORR, false);
+      }
+    }
+    else
+    {
+      if (IsImmLogical(imm, 32))
+        try_base(imm, Approach::ORR, false);
+    }
   }
 
   size_t parts_uploaded = 0;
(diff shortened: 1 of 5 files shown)


Source/Core/Common/Arm64Emitter.cpp
```diff
@@ -1992,97 +1993,114 @@ void ARM64XEmitter::ADRP(ARM64Reg Rd, s32 imm)
...
-  u64 aligned_pc = (u64)GetCodePtr() & ~0xFFF;
-  s64 aligned_offset = (s64)imm - (s64)aligned_pc;
-  // The offset for ADR/ADRP is an s32, so make sure it can be represented in that
-  if (upload_part.Count() > 1 && std::abs(aligned_offset) < 0x7FFFFFFFLL)
-  {
-    // Immediate we are loading is within 4GB of our aligned range
-    // Most likely a address that we can load in one or two instructions
-    if (!(std::abs(aligned_offset) & 0xFFF))
+  Approach best_approach;
+  std::vector<Part> best_parts(max_parts + 1);
```
MerryMage, line 2019: Consider a comment here as this may not be obvious. Maybe (? rewording required): 
```suggestion
  // Have best_parts.size() initially be big to ensure first try_base call replaces it.
  std::vector<Part> best_parts(max_parts + 1);
```

MerryMage (approved):

JosJuice:
Now that PR #9423 exists, ideally this PR should be merged after that one.

JosJuice:
Added ORR support, made us prefer ADRP+ADD over ADR+MOVK and ADRP+MOVK, and fixed a bug in my ADRP code.

degasus:
If I calculated correctly, this code will call malloc 8 times per MOVI2R. I fear this might have a significat time hit in code compilation.

JosJuice:
What do you think about replacing `std::vector<Part>` with `std::array<Part, 4>` plus a `size_t` to keep track of the size?

degasus:
I assume the std::array will make this a lot faster. I'm also fine with the vector if you tell me that it has no large impact in your profile ;)

JosJuice:
Done.

Source/Core/Common/Arm64Emitter.cpp
```diff
@@ -2003,96 +2004,152 @@ void ARM64XEmitter::ADRP(ARM64Reg Rd, s64 imm)
...
 
-// Wrapper around MOVZ+MOVK (and later MOVN)
-void ARM64XEmitter::MOVI2R(ARM64Reg Rd, u64 imm)
+template <typename T, size_t MaxSize>
+class SmallVector final
 {
-  unsigned int parts = Is64Bit(Rd) ? 4 : 2;
-  BitSet32 upload_part(0);
+public:
+  SmallVector() {}
```
leoetlino, line 2011: = default;

Source/Core/Common/Arm64Emitter.cpp
```diff
@@ -2003,96 +2004,152 @@ void ARM64XEmitter::ADRP(ARM64Reg Rd, s64 imm)
...
+  SmallVector(size_t size) : m_size(size) {}
```
leoetlino, line 2012: explicit

leoetlino: This constructor should probably initialize the array member, otherwise uninitialized values can be returned from operator[]... Alternative: `std::array<T, MaxSize> m_array{};`

Source/Core/Common/Arm64Emitter.cpp
```diff
@@ -2003,96 +2004,152 @@ void ARM64XEmitter::ADRP(ARM64Reg Rd, s64 imm)
...
+    Part(u16 imm_, ShiftAmount shift_) : imm(imm_), shift(shift_) {}
+
+    u16 imm;
+    ShiftAmount shift;
+  };
+
+  constexpr size_t max_parts = sizeof(T) / 2;
+
+  // Have best_parts initially be big enough to ensure try_base overwrites it
+  SmallVector<Part, max_parts> best_parts(max_parts + 1);
```
leoetlino, line 2052: Why max_parts + 1? Wouldn't that be larger than the vector's capacity (max_parts)?

Source/Core/Common/Arm64Emitter.cpp
```diff
@@ -2003,96 +2004,152 @@ void ARM64XEmitter::ADRP(ARM64Reg Rd, s64 imm)
...
-  unsigned int parts = Is64Bit(Rd) ? 4 : 2;
-  BitSet32 upload_part(0);
+public:
+  SmallVector() {}
+  SmallVector(size_t size) : m_size(size) {}
 
-  // Always start with a movz! Kills the dependency on the register.
-  bool use_movz = true;
+  void push_back(const T& x) { m_array[m_size++] = x; }
+  void push_back(T&& x) { m_array[m_size++] = std::move(x); }
```
leoetlino, line 2015: Might be worth adding emplace_back too:

```cpp
template <typename... Args>
T& emplace_back(Args&&... args)
{
    return m_array[m_size++] = T{std::forward<Args>(args)...};
}
```

Source/Core/Common/Arm64Emitter.cpp
```diff
@@ -2003,96 +2004,152 @@ void ARM64XEmitter::ADRP(ARM64Reg Rd, s64 imm)
...
+    for (size_t i = 0; i < max_parts; ++i)
     {
-      // Aligned ADR
-      ADRP(Rd, (s32)aligned_offset);
-      return;
+      const size_t shift = i * 16;
+      const u16 imm_shifted = static_cast<u16>(imm >> shift);
+      const u16 base_shifted = static_cast<u16>(base >> shift);
+      if (imm_shifted != base_shifted)
+        parts.push_back(Part{imm_shifted, static_cast<ShiftAmount>(i)});
```
leoetlino, line 2070: With emplace_back, this could just be `parts.emplace_back(imm_shifted, ...);`

Source/Core/Common/Arm64Emitter.cpp
leoetlino: Oh I see, it's for the instructions_required check... This smells a bit fishy though -- could we avoid out of bounds indices/sizes? :)

JosJuice: This is the most efficient way to implement it as far as I can tell, but yeah, it does look rather fishy even though it does work correctly. I suppose the options would be to either make the compile-time size one bigger or to have some kind of variable `bool must_overwrite_best_parts` which is set to true only the first time. Which would you prefer?

leoetlino: must_overwrite_best_parts sounds good!

leoetlino (approved):
Changes LGTM. The unit tests still pass, I hope?

JosJuice:
Yes, the PowerPC tests pass. (The VertexLoader tests still aren't passing, like in master, but that has no relation to this PR.)

dovart:
@JosJuice not building on Arm64

```
../Source/Core/Common/Arm64Emitter.cpp: In member function ‘void Arm64Gen::ARM64XEmitter::MOVI2RImpl(Arm64Gen::ARM64Reg, T)’:
../Source/Core/Common/Arm64Emitter.cpp:2039:5: internal compiler error: Segmentation fault
     MOVZ,
     ^~~~
0x7f99269d23 __libc_start_main
	../csu/libc-start.c:308
```

degasus:
@dovart This is an `internal compiler error`. So which compiler are you using? Which version of the compiler is it? Is this for android, windows or linux?

dovart:
@degasus  gcc version 8.3.0 (Debian 8.3.0-6)  
platform: pi

dovart:
@degasus was building fine before the commits from 2 days ago.

JosJuice:
I wonder if it works fine on newer versions of GCC... If so, maybe it's time to simply drop support for GCC 8. (See PR #9294)

MerryMage:
I can reproduce the above ICE on `gcc (Debian 8.3.0-6) 8.3.0` on a desktop AArch64 Debian buster system.
Installed gcc-10 (most recent package available via `testing`).
Unfortunately this ICE remains on the latest version of GCC.
Will submit a GCC bugreport once I get an account on their bugzilla.

PR #9413, opened 2021-01-02
GraphicsSettings: Parallelize Shader Pre-Compilation

DevJPM:
While running Dolphin with asynchronous UberShader's on a friends computer (on a 8 core CPU) I noticed that the shader pre-compilation only used what looked like 1-2 threads. 

So I investigated Dolphin's source code and found out that the full infrastructure for parallel shader precompilation is already there. Further investigation then revealed that the default value is actually set to 1 with no GUI option to change that.

However, this precompilation is an actual loading time which can take anywhere from 10s to several minutes on most machines and not having it will drastically lower the FPS in some titles.

So this PR suggests to set the default thread count for this pre-compilation to the amount of logical threads supported by the CPU.

Potential Questions:

1. Is it ok for a GraphicSettings default value to be a runtime constant? (i.e. `std::thread::hardware_concurrency()` in this case)
2. Should I add a comment to explain why the max reduction with 1 is there? (`hardware_concurrency()` returns 0 on error)


DevJPM:
Closed for now because there is a better strategy to handling this: Defaulting to the alternative automatic selection and tuning that.

PR #9416, opened 2021-01-02
InputCommon: Wrap remaining invalid default input expressions around ``

Filoppi:
Otherwise they give an error when clicking ok if trying to view or edit the input expressions.
These should be the last ones left that had not been converted yet.

### DIFF
--- a/Source/Core/Core/HW/GCKeyboardEmu.cpp
+++ b/Source/Core/Core/HW/GCKeyboardEmu.cpp
@@ -225,43 +225,43 @@ void GCKeyboard::LoadDefaults(const ControllerInterface& ciface)
 #elif __APPLE__
   m_keys0x->SetControlExpression(0, "Home");
   m_keys0x->SetControlExpression(1, "End");
-  m_keys0x->SetControlExpression(2, "Page Up");
-  m_keys0x->SetControlExpression(3, "Page Down");
+  m_keys0x->SetControlExpression(2, "`Page Up`");
+  m_keys0x->SetControlExpression(3, "`Page Down`");
   m_keys0x->SetControlExpression(4, "");  // Scroll lock
 
-  m_keys2x->SetControlExpression(9, "-");
+  m_keys2x->SetControlExpression(9, "`-`");
   m_keys2x->SetControlExpression(10, "Paragraph");
   m_keys2x->SetControlExpression(11, "");  // Print Scr
-  m_keys2x->SetControlExpression(12, "'");
-  m_keys2x->SetControlExpression(13, "[");
-  m_keys2x->SetControlExpression(14, "=");
-  m_keys2x->SetControlExpression(15, "Keypad *");
-  m_keys3x->SetControlExpression(0, "]");
-  m_keys3x->SetControlExpression(1, ",");
-  m_keys3x->SetControlExpression(2, ".");
-  m_keys3x->SetControlExpression(3, "/");
-  m_keys3x->SetControlExpression(4, "\\");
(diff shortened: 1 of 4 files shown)


Source/Core/Core/HW/GCKeyboardEmu.cpp
```diff
@@ -248,27 +248,27 @@ void GCKeyboard::LoadDefaults(const ControllerInterface& ciface)
...
   m_keys4x->SetControlExpression(13, "Space");
-  m_keys4x->SetControlExpression(14, "Right Command");
+  m_keys4x->SetControlExpression(14, "`Right Command`");
   m_keys4x->SetControlExpression(15, "");  // Menu
 
-  m_keys5x->SetControlExpression(0, "Left Arrow");
-  m_keys5x->SetControlExpression(1, "Down Arrow");
+  m_keys5x->SetControlExpression(0, "`Left Arrow`");
+  m_keys5x->SetControlExpression(1, "`Down Arrow`");
   m_keys5x->SetControlExpression(2, "Up Arrow");
```
BhaaLseN, line 263: Did you leave that one un-ticked intentionally?

Filoppi: No, I just missed it :) 👍

Source/Core/Core/HW/GCKeyboardEmu.cpp
```diff
@@ -225,21 +225,21 @@ void GCKeyboard::LoadDefaults(const ControllerInterface& ciface)
...
+  m_keys0x->SetControlExpression(2, "`Page Up`");
+  m_keys0x->SetControlExpression(3, "`Page Down`");
   m_keys0x->SetControlExpression(4, "");  // Scroll lock
 
-  m_keys2x->SetControlExpression(9, "-");
+  m_keys2x->SetControlExpression(9, "`-`");
   m_keys2x->SetControlExpression(10, "Paragraph");
   m_keys2x->SetControlExpression(11, "");  // Print Scr
   m_keys2x->SetControlExpression(12, "'");
   m_keys2x->SetControlExpression(13, "[");
```
jordan-woyak, line 236: All of these symbols will produce the error message too. I think you should wrap those with back ticks too.
Anything that isn't alphanumeric (or underscore) (and also starting with a letter) Is going to display the "syntax error" message.

Filoppi: I've fixed the remaining ones I think. If there are any left, they can be done later on by someone else I guess.

Filoppi: But yeah when adding these inputs to the expression, they should automatically be wrapped around `` unless they are fully alphanumeric I guess. I will have a look into that but it would be a different PR anyway.

Filoppi: Never mind that's already the way it works.

Filoppi:
Fixed lint check on code I had not touched in the first place.

Filoppi:
The lint check makes no sense IMO and it's not on failing on lines I changed.

JosJuice:
You did change those lines. (Either that or your local clang-format changed the lines, but I think it's been a while since I last saw the buildbot's clang-format disagree with someone's local clang-format.)

Filoppi:
Yes I changed the lines trying to fix them after the first lint check failed, and then got the same error again because the fix it was asking for wasn't the one I expected.
Feel like the lint check is broken for this one and it's better the way it is now but I really don't know what I'm talking about.

jordan-woyak (approved):

PR #9417, opened 2021-01-02
Fix FPS counter and Game Window speed % breaking on pause/unpause

Filoppi:
…by adding the pause state to FPSCounter.
-Add ability to have more than one "OnStateChanged" callback in core.
-Add GetActualEmulationSpeed() in Core, which will be used by input functions added by future commits (requires the new FPS counter features). It returns 1 if the emulation is not running.

Changes are actually really minor. I suggest just testing against master and seeing the problem being fixed for yourself.

### DIFF
--- a/Source/Core/Core/Core.cpp
+++ b/Source/Core/Core/Core.cpp
@@ -103,11 +103,12 @@ static bool s_hardware_initialized = false;
 static bool s_is_started = false;
 static Common::Flag s_is_booting;
 static std::thread s_emu_thread;
-static StateChangedCallbackFunc s_on_state_changed_callback;
+static std::vector<StateChangedCallbackFunc> s_on_state_changed_callbacks;
 
 static std::thread s_cpu_thread;
 static bool s_request_refresh_info = false;
 static bool s_is_throttler_temp_disabled = false;
+static std::atomic<double> s_last_actual_emulation_speed{1.0};
 static bool s_frame_step = false;
 static std::atomic<bool> s_stop_frame_step;
 
@@ -136,6 +137,11 @@ bool GetIsThrottlerTempDisabled()
 void SetIsThrottlerTempDisabled(bool disable)
 {
   s_is_throttler_temp_disabled = disable;
+}
+
+double GetActualEmulationSpeed()
+{
+  return s_last_actual_emulation_speed;
 }
 
 void FrameUpdateOnCPUThread()
@@ -269,9 +275,9 @@ void Stop()  // - Hammertime!
 
   s_is_stopping = true;
 
-  // Notify state changed callback
-  if (s_on_state_changed_callback)
-    s_on_state_changed_callback(State::Stopping);
+  s_timer.Stop();
+
+  CallOnStateChangedCallbacks(State::Stopping);
 
   // Dump left over jobs
   HostDispatchJobs();
@@ -293,6 +299,8 @@ void Stop()  // - Hammertime!
 
     g_video_backend->Video_ExitLoop();
   }
+
+  s_last_actual_emulation_speed = 1.0;
 }
 
 void DeclareAsCPUThread()
@@ -424,16 +432,14 @@ static void FifoPlayerThread(const std::optional<std::string>& savestate_path,
 static void EmuThread(std::unique_ptr<BootParameters> boot, WindowSystemInfo wsi)
 {
   const SConfig& core_parameter = SConfig::GetInstance();
-  if (s_on_state_changed_callback)
-    s_on_state_changed_callback(State::Starting);
+  CallOnStateChangedCallbacks(State::Starting);
   Common::ScopeGuard flag_guard{[] {
     s_is_booting.Clear();
     s_is_started = false;
     s_is_stopping = false;
     s_wants_determinism = false;
 
-    if (s_on_state_changed_callback)
-      s_on_state_changed_callback(State::Uninitialized);
+    CallOnStateChangedCallbacks(State::Uninitialized);
 
     INFO_LOG_FMT(CONSOLE, "Stop\t\t---- Shutdown complete ----");
   }};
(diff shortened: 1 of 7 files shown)


Source/Core/Core/Core.cpp
```diff
@@ -993,9 +1008,38 @@ void Shutdown()
   HostDispatchJobs();
 }
 
-void SetOnStateChangedCallback(StateChangedCallbackFunc callback)
+int AddOnStateChangedCallback(StateChangedCallbackFunc callback)
 {
-  s_on_state_changed_callback = std::move(callback);
+  for (int i = 0; i < s_on_state_changed_callbacks.size(); ++i)
```
lioncash, line 1013:
```suggestion
  for (size_t i = 0; i < s_on_state_changed_callbacks.size(); ++i)
```

Otherwise this will cause a sign-compare warning

Source/Core/Core/Core.cpp
```diff
@@ -993,9 +1008,38 @@ void Shutdown()
...
-void SetOnStateChangedCallback(StateChangedCallbackFunc callback)
+int AddOnStateChangedCallback(StateChangedCallbackFunc callback)
 {
-  s_on_state_changed_callback = std::move(callback);
+  for (size_t i = 0; i < s_on_state_changed_callbacks.size(); ++i)
+  {
+    if (!s_on_state_changed_callbacks[i])
+    {
+      s_on_state_changed_callbacks[i] = std::move(callback);
+      return i;
```
lioncash, line 1018:
```suggestion
      return int(i);
```

Source/Core/Core/Core.cpp
Filoppi: Good catch, thanks.

Source/Core/VideoCommon/FPSCounter.h
```diff
@@ -11,20 +11,26 @@
 class FPSCounter
 {
 public:
-  // Initializes the FPS counter.
   FPSCounter();
+  ~FPSCounter();
```
iwubcode, line 15: If you add a destructor, you need to add a copy-constructor, move-assignment, etc.  See [rule of 5](https://en.cppreference.com/w/cpp/language/rule_of_three)

Source/Core/VideoCommon/FPSCounter.h
```diff
@@ -11,20 +11,26 @@
...
 
   // Called when a frame is rendered (updated every second).
   void Update();
 
   float GetFPS() const { return m_fps; }
+  double GetDeltaTime() { return m_time_diff_secs; }
```
iwubcode, line 21: `double GetDeltaTime() const { return m_time_diff_secs; }`

Filoppi: OMG not sure how I missed that.

Source/Core/VideoCommon/FPSCounter.h
Filoppi: Sure but do we really need to if it's only ever created like this "FPSCounter m_fps_counter;"?

iwubcode: > if it's only ever created like this "FPSCounter m_fps_counter;"

It's just a good way to bullet proof code, even if the possibility is very slim.

Filoppi: Seems like a small nitpick to me, I have loads of PRs like this to submit yet, can't satisfy everyone :).

JosJuice: If you don't actually need the copy constructor etc, feel free to mark them as ` = delete` instead of implementing them. That way nobody will accidentally use the auto-generated ones.

Filoppi: Is it fine now? I marked it as solved just in case no one answered.

JosJuice: Yep, that looks fine.

Source/Core/Core/Core.h
```diff
@@ -140,7 +143,11 @@ void RunOnCPUThread(std::function<void()> function, bool wait_for_completion);
 
 // for calling back into UI code without introducing a dependency on it in core
 using StateChangedCallbackFunc = std::function<void(Core::State)>;
-void SetOnStateChangedCallback(StateChangedCallbackFunc callback);
+// Returns a handle
+int AddOnStateChangedCallback(StateChangedCallbackFunc callback);
+// Also invalidates the handle
+bool RemoveOnStateChangedCallback(int& handle);
```
jordan-woyak, line 149: Our coding conventions recommend passing a pointer rather than a reference if the function modifies the input.
This makes things more clear at the call site.

Filoppi: Making a typedef for the handle would force me to include Core.h in FPSCounter.h, which I'd like to avoid (the reason why I made the handle an int in the first place).
About the reference/pointer, I've followed the same way Config::AddConfigChangedCallback() and ControllerInterface::RegisterDevicesChangedCallback() do, so it should be fine because the functions they are registered to also modify the input, if I understood what you mean correctly.

jordan-woyak: The functions you mention don't modify their input.
See the "Functions" section. https://github.com/dolphin-emu/dolphin/blob/master/Contributing.md

Filoppi: I thought you were talking about something else. Fixed.

JMC47:
@dolphin-emu-bot rebuild

JMC47:
Not sure about the FIFOCI stuff; just looks like noise.  Seems to work fine.

Filoppi:
Just made the static game speed variable atomic, as in my input branch multiple threads could end up reading it.

Source/Core/VideoCommon/RenderBase.cpp
```diff
@@ -1329,7 +1329,9 @@ void Renderer::Swap(u32 xfb_addr, u32 fb_width, u32 fb_stride, u32 fb_height, u6
       {
         // Remove stale EFB/XFB copies.
         g_texture_cache->Cleanup(m_frame_count);
-        Core::Callback_FramePresented();
+        double LastSpeed =
```
leoetlino, line 1332: last_speed

Source/Core/VideoCommon/RenderBase.cpp
```diff
@@ -1329,7 +1329,9 @@ void Renderer::Swap(u32 xfb_addr, u32 fb_width, u32 fb_stride, u32 fb_height, u6
...
+            1.0 / (m_fps_counter.GetDeltaTime() * VideoInterface::GetTargetRefreshRate());
```
leoetlino, line 1333: The denominator should probably be checked to ensure it is non-zero before dividing to make this bulletproof (in case there are timer precision issues for example).

Filoppi: That is something I thought about but deemed extremely unlikely (I think I checked both functions and was quite sure they couldn't return 0), but the main problem is, what do we set LastSpeed to if there is a division by 0?

leoetlino (approved):

PR #9418, opened 2021-01-02
Fix DualShockUDP not adding/removing devices correctly

Filoppi:
-If adding 2 devices with the same name, their unique identifier wouldn't be increased, causing a conflict.
-Removing a device wouldn't actually remove it from the internal devices list because the local list of devices had already been updated when iterating it.
-It was possible to remove devices belonging to other input sources by adding a device with the same name and then removing it.

### DIFF
--- a/Source/Core/InputCommon/ControllerInterface/DualShockUDPClient/DualShockUDPClient.cpp
+++ b/Source/Core/InputCommon/ControllerInterface/DualShockUDPClient/DualShockUDPClient.cpp
@@ -28,6 +28,8 @@
 
 namespace ciface::DualShockUDPClient
 {
+constexpr std::string_view DUALSHOCKUDP_SOURCE_NAME = "DSUClient";
+
 namespace Settings
 {
 const Config::Info<std::string> SERVER_ADDRESS{
@@ -317,7 +319,7 @@ static void Restart()
     }
   }
 
-  PopulateDevices();  // remove devices
+  PopulateDevices();  // Only removes devices
 
   if (s_servers_enabled && !s_servers.empty())
     StartHotplugThread();
@@ -386,11 +388,14 @@ void PopulateDevices()
 {
   INFO_LOG_FMT(SERIALINTERFACE, "DualShockUDPClient PopulateDevices");
 
+  // s_servers has already been updated so we can't use it to know which devices we removed,
+  // also it's good to remove all of them before adding new ones so that their id will be set
+  // correctly if they have the same name
+  g_controller_interface.RemoveDevice(
+      [](const auto* dev) { return dev->GetSource() == DUALSHOCKUDP_SOURCE_NAME; });
+
   for (auto& server : s_servers)
   {
-    g_controller_interface.RemoveDevice(
-        [&server](const auto* dev) { return dev->GetName() == server.m_description; });
-
     std::lock_guard lock{server.m_port_info_mutex};
     for (size_t port_index = 0; port_index < server.m_port_info.size(); port_index++)
     {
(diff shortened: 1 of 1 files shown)


Source/Core/InputCommon/ControllerInterface/DualShockUDPClient/DualShockUDPClient.cpp
```diff
@@ -26,6 +26,8 @@
 #include "InputCommon/ControllerInterface/ControllerInterface.h"
 #include "InputCommon/ControllerInterface/DualShockUDPClient/DualShockUDPProto.h"
 
+#define DUALSHOCKUDP_SOURCE_NAME "DSUClient"
```
lioncash, line 29:
```suggestion
constexpr std::string_view DUALSHOCKUDP_SOURCE_NAME = "DSUClient"
```

This should also be moved down into the namespace.

Filoppi: True, though I copied the way from DInput.h, which declared its name outside of the namespace and as a define.

iwubcode:
Huh, didn't ever imagine someone would use a duplicate device name.  That's certainly not intended and probably should be enforced.  But these changes look good.

Filoppi:
> Huh, didn't ever imagine someone would use a duplicate device name. That's certainly not intended and probably should be enforced. But these changes look good.

I thought about enforcing it but that's not necessary given that two devices can have the exact same name under the same source, they will be distinguished by ID.

iwubcode:
> I thought about enforcing it but that's not necessary given that two devices can have the exact same name under the same source, they will be distinguished by ID.

Yeah, not a problem from a software standpoint, more from a user standpoint.  Imagine you have two devices named the same thing and come back later.  How do you distinguish them?  I guess you have to look at the order.

More importantly, if I'm remembering right, profiles are based on device name.  I wonder what would happen if it was applied.  Maybe both would get it?

But like I said, it was never intended in my design, just an oversight :(

Filoppi:
> > I thought about enforcing it but that's not necessary given that two devices can have the exact same name under the same source, they will be distinguished by ID.
> 
> Yeah, not a problem from a software standpoint, more from a user standpoint. Imagine you have two devices named the same thing and come back later. How do you distinguish them? I guess you have to look at the order.
> 
> More importantly, if I'm remembering right, profiles are based on device name. I wonder what would happen if it was applied. Maybe both would get it?
> 
> But like I said, it was never intended in my design, just an oversight :(

Well of course the first thing I tried to do was adding two devices with the same name to break it :).
Profiles seem to save the whole device path (e.g. DInput/0/Wireless Controller), when restoring it, if the device is not present, it will just add it back to the list and mark it as disconnected, all you got to do now is select a new default device (of the same type) and it would work, so not really a problem.
If I want to have two devices with the same name, it's my problem really, especially because if you attach two Xbox controllers, they will also have the same name, but a different ID.

Source/Core/InputCommon/ControllerInterface/DualShockUDPClient/DualShockUDPClient.cpp
Filoppi: fixed

iwubcode:
@Filoppi - sorry, I had just woken up when I looked at this initially and was thinking of the path not the actual device name.  You're right, there's no reason we can't allow the same name.

This LGTM

iwubcode (approved):

AdmiralCurtiss:
Commits still need to be squashed.

Filoppi:
> Commits still need to be squashed.

Fixed

Filoppi:
Fixed lint code check

jordan-woyak (approved):

PR #9419, opened 2021-01-02
Input improvements (UI, settings and input functions)

Filoppi:
Stuff is nicely describe in commits but main features are:
-Split the state of the WiiMote Cursor between emulation and game (they were polluting each other).
-Improvements to all the input widgets, making it clearer to users how to use them, by adding tooltips, specifying which one is the "default" device, and making the usage of function expressions much easier.
-Added a lot of new function expressions: mostly to remap a mouse to an analog stick.
-Restored custom range support for modifiers and input settings.
-Increased frequency of UI updates of mapping widgets from 30 to 60.

With the new input functions, an Analog Stick can be mapped to a Mouse Axis much better than before, this is especially useful for GC FPS games.
This mapping has been tested with _TimeSplitters: Future Perfect_ and it works great:
```antiAcceleration(bezierCurve(antiDeadzone(`Mouse Axis X+`, 0.25), 0.25, 0.75), 0.2, 0.2, 0.22, 0.4, 1)```
Similar value should also work with other FPS titles, and I'm sure numbers could be tweaked even further.
It undoes the 3 processings that games apply to the stick axes to get the resulting camera rotation, which are:
Deadzone->Bezier Curve or Pow or Linear Mapping->Acceleration (the more your keep it pressed, the faster the camera moves).
The main remaining problem is the maximum speed which is clamped, but it's not that annoying once you get used to it; it's still easier than to use a controller (without aim assist), it just feels like a PC FPS from the 90s more or less. Contrasting the acceleration also can't be perfect.
Unfortunately this won't work as nicely on master as on my branch because relative input devices aren't implemented so half of the mouse movements are lost (PR incoming for that).

I can't be 100% certain of course but I've tested these changes A LOT, and most of them were trying to fix bugs I was hunting for in the first place.

### DIFF
--- a/Source/Core/DolphinQt/Config/Mapping/IOWindow.cpp
+++ b/Source/Core/DolphinQt/Config/Mapping/IOWindow.cpp
@@ -23,6 +23,7 @@
 #include <QTableWidget>
 #include <QVBoxLayout>
 
+#include "Core/ConfigManager.h"
 #include "Core/Core.h"
 
 #include "DolphinQt/Config/Mapping/MappingCommon.h"
@@ -34,6 +35,7 @@
 
 #include "InputCommon/ControlReference/ControlReference.h"
 #include "InputCommon/ControlReference/ExpressionParser.h"
+#include "InputCommon/ControlReference/FunctionExpression.h"
 #include "InputCommon/ControllerEmu/ControllerEmu.h"
 #include "InputCommon/ControllerInterface/ControllerInterface.h"
 
@@ -195,6 +197,7 @@ private:
   int m_column;
 };
 
+// Also used for output
 class InputStateLineEdit : public QLineEdit
 {
 public:
@@ -216,7 +219,13 @@ IOWindow::IOWindow(MappingWidget* parent, ControllerEmu::EmulatedController* con
 
   connect(parent, &MappingWidget::Update, this, &IOWindow::Update);
 
-  setWindowTitle(type == IOWindow::Type::Input ? tr("Configure Input") : tr("Configure Output"));
+  // TODO: it would be nice to have the name of the input in the window
+  // (control->ui_name from MappingWidget.cpp)
+  QString title =
+      m_type == IOWindow::Type::Output ?
+          tr("Configure Output") :
+          (m_type == IOWindow::Type::Input ? tr("Configure Input") : tr("Configure Input Setting"));
+  setWindowTitle(title);
   setWindowFlags(windowFlags() & ~Qt::WindowContextHelpButtonHint);
 
   ConfigChanged();
@@ -256,7 +265,7 @@ void IOWindow::CreateMainLayout()
   m_operators_combo = new QComboBox();
   m_operators_combo->addItem(tr("Operators"));
   m_operators_combo->insertSeparator(1);
-  if (m_type == Type::Input)
+  if (m_type == Type::Input || m_type == Type::InputSetting)
   {
     m_operators_combo->addItem(tr("! Not"));
     m_operators_combo->addItem(tr("* Multiply"));
@@ -270,55 +279,79 @@ void IOWindow::CreateMainLayout()
     m_operators_combo->addItem(tr("^ Xor"));
   }
   m_operators_combo->addItem(tr("| Or"));
-  if (m_type == Type::Input)
+  if (m_type == Type::Input || m_type == Type::InputSetting)
   {
     m_operators_combo->addItem(tr(", Comma"));
   }
 
   m_functions_combo = new QComboBox(this);
   m_functions_combo->addItem(tr("Functions"));
-  m_functions_combo->insertSeparator(1);
-  m_functions_combo->addItem(QStringLiteral("if"));
-  m_functions_combo->addItem(QStringLiteral("timer"));
-  m_functions_combo->addItem(QStringLiteral("toggle"));
-  m_functions_combo->addItem(QStringLiteral("deadzone"));
-  m_functions_combo->addItem(QStringLiteral("smooth"));
-  m_functions_combo->addItem(QStringLiteral("hold"));
-  m_functions_combo->addItem(QStringLiteral("tap"));
-  m_functions_combo->addItem(QStringLiteral("relative"));
-  m_functions_combo->addItem(QStringLiteral("pulse"));
-  m_functions_combo->addItem(QStringLiteral("sin"));
-  m_functions_combo->addItem(QStringLiteral("cos"));
-  m_functions_combo->addItem(QStringLiteral("tan"));
-  m_functions_combo->addItem(QStringLiteral("asin"));
-  m_functions_combo->addItem(QStringLiteral("acos"));
-  m_functions_combo->addItem(QStringLiteral("atan"));
-  m_functions_combo->addItem(QStringLiteral("atan2"));
-  m_functions_combo->addItem(QStringLiteral("sqrt"));
-  m_functions_combo->addItem(QStringLiteral("pow"));
-  m_functions_combo->addItem(QStringLiteral("min"));
-  m_functions_combo->addItem(QStringLiteral("max"));
-  m_functions_combo->addItem(QStringLiteral("clamp"));
+  // This might an empty unselectable whitespace at the end of the ComboBox selection
+  m_functions_combo->insertSeparator(m_functions_combo->count());  // A separator is also an item
+  m_functions_parameters.push_back(QStringLiteral(""));            // Keep the indexes aligned
+  // Logic/Math:
+  AddFunction("if");
+  AddFunction("not");
+  AddFunction("min");
+  AddFunction("max");
+  AddFunction("clamp");
+  AddFunction("minus");
+  AddFunction("pow");
+  AddFunction("sqrt");
+  AddFunction("sin");
+  AddFunction("cos");
+  AddFunction("tan");
+  AddFunction("asin");
+  AddFunction("acos");
+  AddFunction("atan");
+  AddFunction("atan2");
+  m_functions_combo->insertSeparator(m_functions_combo->count());
+  m_functions_parameters.push_back(QStringLiteral(""));
+  // State/time based:
+  AddFunction("onPress");
+  AddFunction("onRelease");
+  AddFunction("onChange");
+  AddFunction("cache");
+  AddFunction("hold");
+  AddFunction("toggle");
+  AddFunction("tap");
+  AddFunction("relative");
+  AddFunction("smooth");
+  AddFunction("pulse");
+  AddFunction("timer");
+  m_functions_combo->insertSeparator(m_functions_combo->count());
+  m_functions_parameters.push_back(QStringLiteral(""));
+  // Stick helpers:
+  AddFunction("deadzone");
+  AddFunction("antiDeadzone");
+  AddFunction("bezierCurve");
+  AddFunction("antiAcceleration");
 
   // Devices
   m_main_layout->addWidget(m_devices_combo);
 
   // Range
   auto* range_hbox = new QHBoxLayout();
-  range_hbox->addWidget(new QLabel(tr("Range")));
+  range_hbox->addWidget(new QLabel(tr("Range (%)")));
   range_hbox->addWidget(m_range_slider);
   range_hbox->addWidget(m_range_spinbox);
   m_range_slider->setMinimum(-500);
   m_range_slider->setMaximum(500);
   m_range_spinbox->setMinimum(-500);
   m_range_spinbox->setMaximum(500);
-  m_main_layout->addLayout(range_hbox);
+  m_range_slider->setToolTip(
+      tr("Multiplies (not clamp) the final result. Usually defaults to 100"));
+  m_range_spinbox->setToolTip(m_range_slider->toolTip());
+  // If this is an input setting, the range won't be saved in our config,
+  // nor it would make sense to change it, so just hide it
+  if (m_type != Type::InputSetting)
+    m_main_layout->addLayout(range_hbox);
 
   // Options (Buttons, Outputs) and action buttons
 
   m_option_list->setTabKeyNavigation(false);
 
-  if (m_type == Type::Input)
+  if (m_type == Type::Input || m_type == Type::InputSetting)
   {
     m_option_list->setColumnCount(2);
     m_option_list->setColumnWidth(1, 64);
@@ -350,7 +383,7 @@ void IOWindow::CreateMainLayout()
 
   button_vbox->addWidget(m_select_button);
 
-  if (m_type == Type::Input)
+  if (m_type == Type::Input || m_type == Type::InputSetting)
   {
     m_test_button->hide();
     button_vbox->addWidget(m_detect_button);
@@ -363,7 +396,7 @@ void IOWindow::CreateMainLayout()
 
   button_vbox->addWidget(m_operators_combo);
 
-  if (m_type == Type::Input)
+  if (m_type == Type::Input || m_type == Type::InputSetting)
     button_vbox->addWidget(m_functions_combo);
   else
     m_functions_combo->hide();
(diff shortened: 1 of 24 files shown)

### CODE REVIEW

File: Source/Core/InputCommon/ControllerEmu/ControlGroup/Cursor.cpp [hunk 1]
```diff
@@ -71,80 +79,82 @@ Cursor::ReshapeData Cursor::GetReshapableState(bool adjusted)
...
     {
-      m_state.x = std::clamp(m_state.x + input.x * max_step, -1.0, 1.0);
-      m_state.y = std::clamp(m_state.y + input.y * max_step, -1.0, 1.0);
+      // If we are using a mouse axis to drive the cursor (there are reasons to), we want the
+      // step to be independent from the game speed, otherwise it would move at a different speed
+      // depending on the game speed.
+      // In other words, we want to be indipendent from time, and just use it as an absolute cursor.
+      // Of course if the emulation can't keep up with the wii mote
+      // emulation rate, the absolute time won't be accurate.
+      bool use_absolute_time = m_relative_absolute_time_setting.GetValue() && !is_ui;
```
jordan-woyak, line 124: I also thought about this when noticing poor behavior with relative mouse mapped to relative pointing.

Instead of adding this option I think we can just always use the new "time elapsed" argument.

Users are already confused enough when mapping pointing. I really don't want to add another relative input related checkbox.

Source/Core/InputCommon/ControllerEmu/ControlGroup/Cursor.h
```diff
@@ -47,22 +54,21 @@ class Cursor : public ReshapableInput
   static constexpr int AUTO_HIDE_MS = 2500;
   static constexpr double AUTO_HIDE_DEADZONE = 0.001;
 
-  // Not adjusted by width/height/center:
-  StateData m_state;
-
-  // Adjusted:
-  StateData m_prev_result;
+  // Not adjusted by width/height/center.
```
jordan-woyak, line 57: Instead of `Cursor` holding dual state (which is needed) it would be cleaner if there was a struct to pass along to `GetState`. Mappping indicators and WiimoteEmu would each have their own instance of course.
This could be done later though. Your fix works for now.

Source/Core/DolphinQt/Config/Mapping/IOWindow.cpp
```diff
@@ -270,55 +279,79 @@ void IOWindow::CreateMainLayout()
...
+  AddFunction("cos");
+  AddFunction("tan");
+  AddFunction("asin");
+  AddFunction("acos");
+  AddFunction("atan");
+  AddFunction("atan2");
+  m_functions_combo->insertSeparator(m_functions_combo->count());
+  m_functions_parameters.push_back(QStringLiteral(""));
+  // State/time based:
+  AddFunction("onPress");
```
jordan-woyak, line 311: I think I'd prefer snake_case over camelCase, but I'll settle for camels if it's what everyone else wants.

Source/Core/DolphinQt/Config/Mapping/IOWindow.cpp
```diff
@@ -270,55 +279,79 @@ void IOWindow::CreateMainLayout()
...
   m_range_slider->setMaximum(500);
   m_range_spinbox->setMinimum(-500);
   m_range_spinbox->setMaximum(500);
-  m_main_layout->addLayout(range_hbox);
+  m_range_slider->setToolTip(
+      tr("Multiplies (not clamp) the final result. Usually defaults to 100"));
+  m_range_spinbox->setToolTip(m_range_slider->toolTip());
+  // If this is an input setting, the range won't be saved in our config,
+  // nor it would make sense to change it, so just hide it
+  if (m_type != Type::InputSetting)
```
jordan-woyak, line 347: Unfortunately the (poorly named) "Range" setting IS needed for outputs. Steering wheel forces (which are sent as values from -1 to +1) often need reversing using a negative range.

jordan-woyak (commented):
Cool changes overall.
I skimmed through and left some comments.
I'll take a deeper look at the new expression functions a bit later.

Source/Core/InputCommon/ControllerEmu/ControlGroup/Cursor.cpp
```diff
@@ -71,80 +79,82 @@ Cursor::ReshapeData Cursor::GetReshapableState(bool adjusted)
   const ControlState y = controls[0]->GetState() - controls[1]->GetState();
   const ControlState x = controls[3]->GetState() - controls[2]->GetState();
 
-  // Return raw values. (used in UI)
+  // Return raw values
   if (!adjusted)
     return {x, y};
 
-  return Reshape(x, y, 0.0);
+  return Reshape(x, y, 0.0, std::numeric_limits<ControlState>::infinity());
```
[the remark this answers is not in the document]
Filoppi: Because with a relative axis, like a mouse, the offset from the previous frame can, and should, be up to any number. It made no sense that the maximum speed between two frames was 1.

Source/Core/InputCommon/ControllerEmu/ControlGroup/Cursor.cpp
Filoppi: Well the tooltip explains it, and I do like the choice as it depends on what you want to do. You might not always want it on depending on the relative input source.

Source/Core/InputCommon/ControllerEmu/ControlGroup/Cursor.h
Filoppi: Yeah I know it's not the best way of implementing it but it's fine. If more is needed it can be done later.

Source/Core/DolphinQt/Config/Mapping/IOWindow.cpp
Filoppi: Didn't i just follow the previous naming convention?

Source/Core/InputCommon/ControllerEmu/ControlGroup/Cursor.cpp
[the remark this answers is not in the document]
jordan-woyak: You're right. This value is processed further by `GetState` and should not be clamped here.
I think you should add a comment explaining this though.

Source/Core/DolphinQt/Config/Mapping/IOWindow.cpp
Filoppi: Range is still used for outputs, this is for input settings, which is a sub category of inputs with a fixed value. They can't use range as it's never read nor serealized.

Source/Core/DolphinQt/Config/Mapping/IOWindow.cpp
jordan-woyak: I think you introduced the first multi word function names.

Source/Core/InputCommon/ControllerEmu/ControlGroup/Cursor.cpp
```diff
@@ -62,6 +62,8 @@ Cursor::Cursor(std::string name_, std::string ui_name_)
 
   AddSetting(&m_relative_setting, {_trans("Relative Input")}, false);
   AddSetting(&m_autohide_setting, {_trans("Auto-Hide")}, false);
+  numeric_settings.back().get()->GetInputReference().default_range = AUTO_HIDE_MS / 1000.0;
+  numeric_settings.back().get()->GetInputReference().range = AUTO_HIDE_MS / 1000.0;
```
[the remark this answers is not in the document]
Filoppi: This is setting the default value, which I believe was called range. But I'm not sure now I'm from my phone, will look tomorrow.

Source/Core/InputCommon/ControllerEmu/ControlGroup/ControlGroup.cpp
```diff
@@ -44,7 +44,7 @@ void ControlGroup::AddDeadzoneSetting(SettingValue<double>* value, double maximu
               // i18n: The percent symbol.
               _trans("%"),
               // i18n: Refers to the dead-zone setting of gamepad inputs.
-              _trans("Input strength to ignore.")},
+              _trans("Circular input strength to ignore and remap.")},
```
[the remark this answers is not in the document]
Filoppi: Ok. What's a better word? Just none?

Source/Core/InputCommon/ControllerEmu/ControlGroup/Cursor.cpp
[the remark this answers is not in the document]
Filoppi: Should I? It made no sense to clamp it in the first place 😅.

Source/Core/DolphinQt/Config/Mapping/IOWindow.cpp
Filoppi: I think for an end user camelCase would be nicer

Source/Core/InputCommon/ControllerEmu/ControlGroup/Cursor.cpp
jordan-woyak: I really don't think the old timing is needed. The new timing might require tweaked "range" settings or whatever to restore old behavior but after that it should only make a difference when the game isn't running at 100% speed.

The advanced user want's their mouse movement of one pixel to do the same thing every time. (new timing)
The average user just want a dirty way to move the "cursor" around the screen with an analog stick and they don't care.

Source/Core/InputCommon/ControllerEmu/ControlGroup/ControlGroup.cpp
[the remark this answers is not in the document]
jordan-woyak: You could detect the shape (circle, square, star) to update the tooltip! jk
I'm not sure what a better word is. This is why I gave it such a vague description I think. :P

Source/Core/InputCommon/ControllerEmu/ControlGroup/Cursor.cpp
Filoppi: It does matter if you are using an analog stick to drive the relative wiimote cursor. In that case you don't want to use that setting to actually have the cursor to move independently from the game speed. That's why I made it off by default.

mbc07:
Maybe you could try using the custom tooltips from the graphics config window instead of the default OS tooltips? I didn't dig too deep at the code but one of the design goals of the custom tooltips was making them reusable on other places AFAIR...

Filoppi:
> Maybe you could try using the custom tooltips from the graphics config window instead of the default OS tooltips? I didn't dig too deep at the code but one of the design goals of the custom tooltips was making them reusable on other places AFAIR...

I think I wrote this code before the new tooltips were a thing, but anyway, I do understand and agree, but unless I'm enforced, I'd like to leave that task for someone else, this is already an improvement and they are not the only "old" style tooltips around. I just want to be done, I have a lot of other things to merge.

Source/Core/InputCommon/ControllerEmu/ControlGroup/Cursor.cpp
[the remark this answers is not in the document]
Filoppi: Right, no, this is useless, I think I was just testing whether the range would be visualized correctly in the UI and then forgot the code wasn't doing anything. Though I had hoped to just set the numeric setting range, and then read that back instead of reading AUTO_HIDE_MS in the cursors update, so that if we ever introduced range on fixed input numeric settings, it would already be changeable. Anyway doesn't matter.

Source/Core/DolphinQt/Config/Mapping/IOWindow.cpp
jordan-woyak: Ah. I misread what this was for. :P

Source/Core/InputCommon/ControllerEmu/ControlGroup/Cursor.cpp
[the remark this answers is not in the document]
Filoppi: Looking at this again, seems like it's now also not clamped when NOT using relative input. I think I had thought of that already but I need to test again.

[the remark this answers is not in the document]
jordan-woyak: That should be fine as long as the final `GetState` function clamps I think.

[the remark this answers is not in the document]
Filoppi: It does behave differently without the clamp, it can go over the borders in wiimote Widget and in the game I guess, Not sure what the consequences would be. I don't even understand why it's reshaped when using the relative movement setting.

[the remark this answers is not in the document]
jordan-woyak: `Reshape` also processes the calibration so it should still be used with relative input.
And on second though it should be clamped in the end so relative input doesn't let the pointer just float off the screen forever.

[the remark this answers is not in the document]
Filoppi: The relative one is already clamped in GetState(). The non relative one isn't clamped anymore, not as before, but it might still be "clamped" by other means in the Reshapre function. Will confirm tomorrow as I feel like there is a reason I have changed it for both relative and non relative settings.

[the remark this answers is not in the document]
Filoppi: All right, I think it should still be clamped when not in relative mode, otherwise you could go over the max specified angle, which usually corresponds to the borders of the game view and the borders of your window, if you want to go over these values, it's better to just increase the max angles instead of accepting values over abs(1).
And it should always be clamped when in relative mode to avoid the cursor drifting over the game view.
Will fix later.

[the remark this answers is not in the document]
Filoppi: Fixed

Miksel12:
The new input expressions work quite nice when using a mouse. I do think an explaination would be nice though I don't know where that could fit. Maybe a link to a wiki page? (https://wiki.dolphin-emu.org/index.php?title=Input_Syntax)
I think I agree with jordan that snake_case would look nicer.

The Relative Input Absolute Time seems like a nice feature but I think for an end user, having 2 options can be detrimental. The point section is already a lot to take in for new users and I can imaging that users only realistically need one option. I expect that the vast majority runs their games at 100% or sometimes almost full speed. When users fast forward, they are probably using it to skip something and are not giving inputs. So I think in most cases the difference is insignificant.

Filoppi:
@Miksel12 Thanks for testing.
In my current branch I've added a UI exposed description (tooltip) to every function so adding it to that page wouldn't be needed anymore I guess.

I prefer camelCase but honestly I don't care that much so if everybody wants snake_case I can change it.

I actually play games at increased speeds to go past boring gameplay sections faster, and that's why I added that option, the cursor wouldn't work as nice without it. The tooltip is there to explain it and it's greyed out when not using "Relative Input" so I don't think it adds that much complication, not enough to outweigh the benefits.

Miksel12:
Tooltips sounds like a good way to explain all the options. 

About the relative input, I can't think of any situations where non absolute relative input would be more beneficial than absolute input.

Filoppi:
> About the relative input, I can't think of any situations where non absolute relative input would be more beneficial than absolute input.

As I already mentioned above, when using a controller stick to drive the wii mote cursor (with "Relative Input" on of course), you need to turn the "Absolute Time" setting off to actually make it move independently from the game speed.
While when using a mouse or any relative axis (a controller stick is not a relative axis) you need it on.

Source/Core/InputCommon/ControllerEmu/ControlGroup/Cursor.cpp
```diff
@@ -69,80 +77,92 @@ Cursor::ReshapeData Cursor::GetReshapableState(bool adjusted)
...
+  int i = is_ui ? 1 : 0;
+
   // Kill this after state is moved into wiimote rather than this class.
   const auto now = Clock::now();
   const auto ms_since_update =
-      std::chrono::duration_cast<std::chrono::milliseconds>(now - m_last_update).count();
-  m_last_update = now;
-
-  const double max_step = STEP_PER_SEC / 1000.0 * ms_since_update;
+      std::chrono::duration_cast<std::chrono::microseconds>(now - m_last_update[i]).count() /
```
iwubcode, line 103:
```
// Include <ratio>
using milli_with_remainder = std::chrono::duration<double, std::milli>;

const auto ms_since_update = std::chrono::duration_cast<milli_with_remainder>(now - m_last_update[i]).count();
```

I can't find the link to the SO post but I believe that is more accurate

Source/Core/InputCommon/ControllerEmu/ControlGroup/Cursor.cpp
Filoppi: Yes I confirm what I said above, when using an analog stick to drive the relative wiimote cursor, you want the absolute time setting to be off, I tested it.
Also relative mouse movement gets much more precise with this setting as the chrono time delta was extremely variable between frames.

Source/Core/InputCommon/ControllerEmu/ControlGroup/Cursor.cpp
Filoppi: Done that in: https://github.com/dolphin-emu/dolphin/pull/9489, will either backport it here or merge it in a new PR, depending how we proceed.

Source/Core/DolphinQt/Config/Mapping/IOWindow.cpp
```diff
@@ -270,55 +279,79 @@ void IOWindow::CreateMainLayout()
...
-  m_functions_combo->addItem(QStringLiteral("asin"));
-  m_functions_combo->addItem(QStringLiteral("acos"));
-  m_functions_combo->addItem(QStringLiteral("atan"));
-  m_functions_combo->addItem(QStringLiteral("atan2"));
-  m_functions_combo->addItem(QStringLiteral("sqrt"));
-  m_functions_combo->addItem(QStringLiteral("pow"));
-  m_functions_combo->addItem(QStringLiteral("min"));
-  m_functions_combo->addItem(QStringLiteral("max"));
-  m_functions_combo->addItem(QStringLiteral("clamp"));
+  // This might an empty unselectable whitespace at the end of the ComboBox selection
```
Rumi-Larry, line 289: `//This might an empty...` I think you are missing a word here.

Filoppi: No that's the way I wanted to write it, but I will reword it. Basically if you add too many white spaces, Qt will end up adding a white/emtpy line at the bottom.

Source/Core/InputCommon/ControllerEmu/ControlGroup/Cursor.cpp
Filoppi: m_relative_absolute_time_setting won't be needed anymore given I have made an input function to do the same thing (per mapping). It will be in a future PR but at least this one will be smaller.

PR #9423, opened 2021-01-03
UnitTests: Add MOVI2R test

MerryMage:
Simple codegen tests.

### DIFF
--- a/Source/Core/Common/Random.h
+++ b/Source/Core/Common/Random.h
@@ -5,12 +5,37 @@
 #pragma once
 
 #include <cstddef>
+#include <memory>
 #include <type_traits>
 
 #include "Common/CommonTypes.h"
 
 namespace Common::Random
 {
+/// Cryptographically secure pseudo-random number generator, with explicit seed.
+class PRNG final
+{
+public:
+  explicit PRNG(u64 seed) : PRNG(&seed, sizeof(u64)) {}
+  PRNG(void* seed, std::size_t size);
+  ~PRNG();
+
+  void Generate(void* buffer, std::size_t size);
+
+  template <typename T>
+  T GenerateValue()
+  {
+    static_assert(std::is_arithmetic<T>(), "T must be an arithmetic type in GenerateValue.");
+    T value;
+    Generate(&value, sizeof(value));
+    return value;
+  }
+
+private:
+  struct Impl;
+  std::unique_ptr<Impl> m_impl;
+};
+
 /// Fill `buffer` with random bytes using a cryptographically secure pseudo-random number generator.
 void Generate(void* buffer, std::size_t size);
 
--- a/Source/UnitTests/Core/PowerPC/JitArm64/MovI2R.cpp
+++ b/Source/UnitTests/Core/PowerPC/JitArm64/MovI2R.cpp
@@ -1,0 +1,95 @@
+// Copyright 2021 Dolphin Emulator Project
+// Licensed under GPLv2+
+// Refer to the license.txt file included.
+
+#include <cstddef>
+#include <type_traits>
+
+#include "Common/Arm64Emitter.h"
+#include "Common/Assert.h"
+#include "Common/BitUtils.h"
+#include "Common/Random.h"
+
+#include <gtest/gtest.h>
+
+namespace
+{
+using namespace Arm64Gen;
+
+class TestMovI2R : public ARM64CodeBlock
+{
+public:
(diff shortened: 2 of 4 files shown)


Source/UnitTests/Core/PowerPC/JitArm64/MovI2R.cpp
```diff
@@ -0,0 +1,88 @@
+// Copyright 2021 Dolphin Emulator Project
+// Licensed under GPLv2+
+// Refer to the license.txt file included.
+
+#include "Common/Arm64Emitter.h"
+#include "Common/Random.h"
+
+#include <fmt/format.h>
```
JosJuice, line 8: Is this include used for anything?

Source/UnitTests/Core/PowerPC/JitArm64/MovI2R.cpp
```diff
@@ -0,0 +1,88 @@
...
+class TestMovI2R : public ARM64CodeBlock
+{
+public:
+  TestMovI2R() { AllocCodeSpace(4096); }
+
+  void Check32(u32 value)
+  {
+    ResetCodePtr();
+
+    auto fn = (u32(*)())GetCodePtr();
```
JosJuice, line 24: Please use C++ style casts here and below.

Source/UnitTests/Core/PowerPC/JitArm64/MovI2R.cpp
```diff
@@ -0,0 +1,88 @@
...
+    test.Check64(i);
+  }
+}
+
+TEST(JitArm64, MovI2R_Rand)
+{
+  TestMovI2R test;
+  for (u64 i = 0; i < 0x100000; i++)
+  {
+    test.Check64(Common::Random::GenerateValue<u64>());
```
JosJuice, line 64: For reproducibility (e.g. if you're trying to fix a bug and you want to be sure that the reason the tests started passing actually is because your fix works), I think it would be nice to use a PRNG that takes a constant seed. I don't think we have any in `Common`, but the `<random>` header has some.

Source/UnitTests/Core/PowerPC/JitArm64/MovI2R.cpp
```diff
@@ -0,0 +1,128 @@
...
+#include "Common/Assert.h"
+#include "Common/BitUtils.h"
+#include "Common/Random.h"
+
+#include <gtest/gtest.h>
+#include <mbedtls/hmac_drbg.h>
+
+namespace
+{
+class PRNG
```
leoetlino, line 18: Could we put this in Common/Random.h? I guess it'd have to be separate from the existing class as that one uses mbedtls_entropy_context to seed. The pimpl idiom should probably be used to avoid exposing mbedtls details directly.

MerryMage: Done.

leoetlino (approved):

Source/Core/Common/Random.h
```diff
@@ -5,12 +5,37 @@
...
 
 #include "Common/CommonTypes.h"
 
 namespace Common::Random
 {
+/// Cryptographically secure pseudo-random number generator, with explicit seed.
+class PRNG final
+{
+public:
+  explicit PRNG(u64 seed = 0) : PRNG(&seed, sizeof(u64)) {}
```
JosJuice, line 19: The seed probably shouldn't have a default value, lest someone accidentally set the seed to 0 when they wanted a random seed.

JosJuice:
How fast is `MovI2R_32BitValues` for you? I tried running it for a few minutes on my phone, but it didn't terminate. `MovI2R_Rand` took about 5 seconds, so... since `MovI2R_32BitValues` has 4096 times more iterations, I suppose it would take a few hours to finish?

Source/Core/Common/Random.h
leoetlino: Oh yeah, we should remove the default value to encourage the use of Random::Generate instead.

MerryMage:
Ah, yes, it takes a while. While I enjoy exhaustive testing, I guess it would be better to text a few 32 bit values instead.

JosJuice:
Feel free to keep the current test as a compile-time option if you want to, but yeah, we would need something faster for the cases where someone wants to run all the tests and isn't working on MOVI2R specifically.

JosJuice (approved):

PR #9424, opened 2021-01-04
Software: Remove normalization special case

Pokechu22:
The special case doesn't appear to make a significant difference in any games, and the current implementation has a (minor, fixable) issue that breaks Super Mario Sunshine (both with a failed assertion ([bug 11742](https://bugs.dolphin-emu.org/issues/11742)) and a rendering issue ([bug 7476](https://bugs.dolphin-emu.org/issues/7476))).  Hardware testing wasn't able to reproduce the special case, either, so it may just not exist.

PR #9315 contains a fixed implementation of the special case on all video backends, and can serve as a basis for it being reintroduced if it is found to exist under more specific circumstances.  For now, I don't see a reason to keep it present.


leoetlino:
Since hardware testing has been done and the FifoCI diff looks sane, I plan to merge this tomorrow unless somebody objects.

nyanpasu64:
Still not merged? :thinking:

PR #9426, opened 2021-01-07
Externals: Update Vulkan headers and glslang

iMonZ:
Update external dependencies for better BigSur support and less crashes


iMonZ:
The only thing that is missing is MoltenVK but it has a strange size.
I hope someone can do that.

leoetlino:
This fails to build on *any* platform so we cannot merge this. Please make sure this builds at least locally and then re-open a PR...

PR #9427, opened 2021-01-07
Update ja.po

HIJKLM:
Notation fluctuation correction.


JosJuice:
Sorry, but we can only accept translation updates through [Transifex](https://www.transifex.com/delroth/dolphin-emu/dashboard/). Any edits that made manually to the translation files would get overwritten the next time we sync with Transifex.

PR #9428, opened 2021-01-08
Android: Use old folder picker on Android TV

JosJuice:
See the comment I added to the code. This is a rather serious issue for Android TV users from what I've heard.

For reference, we switched folder pickers in PR #9318.

### DIFF
--- a/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/ui/main/MainActivity.java
+++ b/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/ui/main/MainActivity.java
@@ -163,8 +163,15 @@ public final class MainActivity extends AppCompatActivity implements MainView
   @Override
   public void launchFileListActivity()
   {
-    Intent intent = new Intent(Intent.ACTION_OPEN_DOCUMENT_TREE);
-    startActivityForResult(intent, MainPresenter.REQUEST_DIRECTORY);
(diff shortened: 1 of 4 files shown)


Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/ui/main/MainActivity.java
```diff
@@ -163,8 +163,15 @@ public void launchSettingsActivity(MenuTag menuTag)
   @Override
   public void launchFileListActivity()
   {
-    Intent intent = new Intent(Intent.ACTION_OPEN_DOCUMENT_TREE);
-    startActivityForResult(intent, MainPresenter.REQUEST_DIRECTORY);
+    if (DirectoryInitialization.preferOldFolderPicker(this))
```
leoetlino, line 166: Any opinion on moving this into openDirectoryPicker?

JosJuice: That would be inappropriate, because the `openDirectoryPicker` call from `SettingsAdapter` must not use the new file picker. I suppose I could create a second variant of `openDirectoryPicker` for this, but it somewhat obscures the logic of how the intent the activity returns as a result should be handled (whether `getSelectedPath` should be used or not).

leoetlino (approved):

Ebola16 (approved):
Makes sense to me, untested.

PR #9434, opened 2021-01-10
UWP Version of Dolphin will not build (5.0)

GK6475:
VS2017 Will not build the UWP version of Dolphin, I’ve tried over and over and have everything necessary installed to build it but it just won’t. I also got one error eventually that said the license had expired. Please let me know if anything that Im doing is wrong here.


JosJuice:
This is not an issue tracker, and we also don't provide any support for this UWP fork anyway. You'll have to ask them instead.

PR #9438, opened 2021-01-12
DolphinQT: Gives option to add desktop shortcut

shuffle2:
When a game is selected, the option to add a shortcut of the game to the desktop is given. Uses native Windows API since Qt lacks support for adding shortcuts.

This is a modification of #9365 , using WIL and handling all error/memory leak cases correctly (hopefully 🙂)

### DIFF
--- a/Source/Core/DolphinQt/GameList/GameList.cpp
+++ b/Source/Core/DolphinQt/GameList/GameList.cpp
@@ -53,6 +53,20 @@
 
 #include "UICommon/GameFile.h"
 
+#ifdef _WIN32
+#include <QCoreApplication>
+#include <shlobj.h>
+#include <wil/com.h>
+
+// This file uses some identifiers which are defined as macros in Windows headers.
+// Undefine them for the intellisense parser to solve some red squiggles.
+#ifdef __INTELLISENSE__
+#ifdef DeleteFile
+#undef DeleteFile
+#endif
+#endif
+#endif
+
 GameList::GameList(QWidget* parent) : QStackedWidget(parent), m_model(this)
 {
   m_list_proxy = new ListProxyModel(this);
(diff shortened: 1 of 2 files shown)


Source/Core/DolphinQt/GameList/GameList.cpp
```diff
@@ -631,6 +654,41 @@ void GameList::OpenGCSaveFolder()
     ModalMessageBox::information(this, tr("Information"), tr("No save data found."));
 }
 
+#ifdef _WIN32
+bool GameList::AddShortcutToDesktop()
+{
+  auto init = wil::CoInitializeEx_failfast(COINIT_APARTMENTTHREADED);
```
CookiePLMonster, line 660: Not sure if `failfast` is desirable here - COM may already be initialized in a different threading model, and I don't think this code **must** use an apartment threaded model?

shuffle2: That is (I think?) the only wrapper provided by WIL. Multithreaded would not work here as it's on a GUI thread. Specifying multithreaded resulted in an assert, so I think it's working as intended? However, I didn't investigate further to find out what is calling CoInitialize on this thread beforehand.

All CoInitialize* calls on a specific thread _should_ specify the same threading model, even though it's possible to specify multiple threading models (and ignore the returned error code). I think it's better to be explicit about the expectation (which is probably why WIL only exposes a _failfast version of the API).

CookiePLMonster: It's possible - but if you don't want to make a strong assumption like this, you could. My WASAPI PR realizes this by calling `CoInitializeEx` "manually" and then activating/deactivating the RAII `CoUnitialize` call basing on the result.

shuffle2: I noticed [cubeb has something similar](https://github.com/dolphin-emu/dolphin/blob/0029ca84b07f44e3ce8c80a44fa0285a966b34d3/Externals/cubeb/src/cubeb_wasapi.cpp#L138). I'd rather use what's provided by WIL, and I don't think making a strong assumption is a problem here. Ideally, all usages of CoInitialize would make strong assumptions (therefor if they trigger fastfail, it's a good indication some code is actually being used incorrectly).
If you want to make a wrapper like cubeb's, maybe consider moving it to some more common location in dolphin?

leoetlino:
ah, this looks much nicer than the non-WIL version!

PR #9440, opened 2021-01-13
DolphinQt: Qt 6.0 forward-compatibility changes

lioncash:
Makes changes that should ease us into transitioning over to Qt 6.0 in the future. Qt 6 brings a few breaking changes with it, and thankfully, replacements for most of those deprecated or removed features were introduced within Qt 5, so we can replace those obsoleted features.

This way when we move to Qt 6, way less things will break due to APIs being removed or changed.

### DIFF
--- a/Source/Core/DolphinQt/Config/GameConfigEdit.cpp
+++ b/Source/Core/DolphinQt/Config/GameConfigEdit.cpp
@@ -11,6 +11,7 @@
 #include <QMenu>
 #include <QMenuBar>
 #include <QPushButton>
+#include <QRegularExpression>
 #include <QScrollBar>
 #include <QStringListModel>
 #include <QTextCursor>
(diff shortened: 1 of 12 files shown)


BhaaLseN (approved):

iwubcode (approved):

RinMaru:
Will this result in the GUI looking diff?

nbohr1more:
@RinMaru GitHub is a developer platform. If you wish to discuss Dolphin changes, go to the dolphin emulator forums: 

https://forums.dolphin-emu.org/Forum-development-discussion

or IRC

#dolphin-emu @ irc.freenode.net

QT framework changes should not affect any visual elements. There are rare cases where a legacy QT feature is used then removed with no replacement where the developers would have to implement a "from scratch" equivalent but this replacement would be on the backend behind the visual elements so it would only (possibly) affect animations \ menu transitions \ etc... not the actual visual appearance.

PR #9442, opened 2021-01-14
Contributing.md: Code comment for do-while sample

Stevoisiak:


Stevoisiak:
> [FifoCI](https://fifo.ci/about/) detected that this change impacts graphical rendering.

Considering this PR doesn't change any code, I suspect this might be a false positive.

jordan-woyak (approved):
This change is consistent with the other examples.

We probably could just replace this whole file with:
TLDR: Just use `clang-format` with our included `.clang-format` file.

PR #9446, opened 2021-01-16
Arm64Emitter: Convert ShiftType to enum class

Dentomologist:
Straightforward conversion of ShiftType from enum to enum class.

### DIFF
--- a/Source/Core/Common/Arm64Emitter.h
+++ b/Source/Core/Common/Arm64Emitter.h
@@ -277,12 +277,16 @@ constexpr ARM64Reg EncodeRegToQuad(ARM64Reg reg)
   return static_cast<ARM64Reg>(reg | 0xC0);
 }
 
-enum ShiftType
+enum class ShiftType
 {
-  ST_LSL = 0,
-  ST_LSR = 1,
-  ST_ASR = 2,
-  ST_ROR = 3,
+  // Logical Shift Left
+  LSL = 0,
+  // Logical Shift Right
+  LSR = 1,
+  // Arithmetic Shift Right
+  ASR = 2,
+  // Rotate Right
+  ROR = 3,
 };
 
 enum class IndexType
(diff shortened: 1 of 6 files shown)


BhaaLseN:
You can (probably) ignore the Windows fail, thats on our CI...but the Android one needs to be looked at. `m_shifttype` is now a typed enum class and doesn't implicitly convert into `int`, but you can just cast that inside `GetData`.

Dentomologist:
I'm not sure why Visual Studio let that compile in the first place, but I've added the cast.

JosJuice:
@dolphin-emu-bot rebuild

BhaaLseN (approved):

Source/Core/Common/Arm64Emitter.h
```diff
@@ -277,12 +277,12 @@ constexpr ARM64Reg EncodeRegToQuad(ARM64Reg reg)
...
 }
 
-enum ShiftType
+enum class ShiftType
 {
-  ST_LSL = 0,
-  ST_LSR = 1,
-  ST_ASR = 2,
-  ST_ROR = 3,
+  LogicalShiftLeft = 0,
```
lioncash, line 282: I think it'd still be informative as `LSL`, `LSR`, `ASR`, and `ROR` as the names for the shift types. These are what they're referred to as by ARM in official documentation, and the long-form makes code (imo) slightly more verbose to read and write.

Dentomologist: Given that you've made roughly half the commits touching Arm64Emitter in the last few years, I'm especially inclined to defer to your judgement. I've reverted the naming changes.

Said changes were inspired by my not knowing what the terms referred to until I did some googling. In order to help anyone else in a similar situation I've added some comments in the declaration with the full name, which at least in Visual Studio appear in the hover tooltip for the respective enum values.

JosJuice:
@dolphin-emu-bot rebuild

JosJuice (approved):

lioncash (approved):

PR #9449, opened 2021-01-16
Qt: Fix Shake Mapping Indicator not showing deadzone

Filoppi:
QRectF was missing from drawRect which meant the deazone was casted to an int, flooring its value (it goes from 0 to 1).


iwubcode (approved):
Confirmed the deadzone indicator is fixed.  LGTM

PR #9450, opened 2021-01-17
Qt: Fix stock input profiles being deletable

Filoppi:
Also avoid files without a name before the extension (name: ".ini") from being added to the list because then they wouldn't be saveable and it would appear with an empty name anyway.

On Windows GetSysDirectory() was returning a path with `//` while being checked against one that had been converted to `\`, so the check would fail and you'd be able to delete system profiles.

### DIFF
--- a/Source/Core/DolphinQt/Config/Mapping/MappingWindow.cpp
+++ b/Source/Core/DolphinQt/Config/Mapping/MappingWindow.cpp
@@ -16,6 +16,7 @@
 
 #include "Core/Core.h"
 
+#include "Common/CommonPaths.h"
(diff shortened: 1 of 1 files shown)


Techjar:
I am disappointed that none of our Windows developers noticed this.

jordan-woyak (approved):

Filoppi:
> I am disappointed that none of our Windows developers noticed this. Also I hate that Windows uses backslash. Thanks for catching this.

I wonder if GetSysDirectory() should directly return a path with "\" so it doesn't need to get patched every time? But I didn't want to make that larger change of course.

PR #9452, opened 2021-01-18
Android: Add "Import Wii Save"

JosJuice:
Exporting will have to wait for another PR, because I need to implement more SAF stuff first.

### DIFF
--- a/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/ui/main/MainPresenter.java
+++ b/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/ui/main/MainPresenter.java
@@ -31,6 +31,7 @@ public final class MainPresenter
   public static final int REQUEST_GAME_FILE = 2;
   public static final int REQUEST_SD_FILE = 3;
   public static final int REQUEST_WAD_FILE = 4;
+  public static final int REQUEST_WII_SAVE_FILE = 5;
 
   private final MainView mView;
   private final Context mContext;
@@ -98,6 +99,11 @@ public final class MainPresenter
         new AfterDirectoryInitializationRunner().run(context, true,
                 () -> mView.launchOpenFileActivity(REQUEST_WAD_FILE));
         return true;
+
+      case R.id.menu_import_wii_save:
+        new AfterDirectoryInitializationRunner().run(context, true,
+                () -> mView.launchOpenFileActivity(REQUEST_WII_SAVE_FILE));
+        return true;
     }
 
     return false;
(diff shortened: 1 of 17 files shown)


Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/ui/main/MainPresenter.java
```diff
@@ -177,10 +210,14 @@ private void runOnThreadAndShowResult(int progressMessage, Supplier<String> f)
       {
         progressDialog.dismiss();
 
-        AlertDialog.Builder builder = new AlertDialog.Builder(mContext, R.style.DolphinDialogBase);
-        builder.setMessage(result);
-        builder.setPositiveButton(R.string.ok, (dialog, i) -> dialog.dismiss());
-        builder.show();
+        if (result != null)
```
leoetlino, line 245: maybe return early instead to reduce the indentation level

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/utils/WiiUtils.java
```diff
@@ -2,5 +2,13 @@
 
 public final class WiiUtils
 {
+  public static final int RESULT_SUCCESS = 0;
```
leoetlino, line 5: This seems a bit fragile considering nothing in the C++ source for the enum class suggests the order of the existing values must not be changed and new values must be added at the end. I can easily see somebody (me, for example :) ) forgetting to update the Java code.

What do you think of changing the C++ JNI code to fetch the value of these constants and map the enum class value -> Java enum value?

Source/Android/jni/WiiUtils.cpp
```diff
@@ -21,4 +22,14 @@ JNIEXPORT jboolean JNICALL Java_org_dolphinemu_dolphinemu_utils_WiiUtils_install
   const std::string path = GetJString(env, jFile);
   return static_cast<jboolean>(WiiUtils::InstallWAD(path));
 }
+
+JNIEXPORT jint JNICALL Java_org_dolphinemu_dolphinemu_utils_WiiUtils_importWiiSave(JNIEnv* env,
+                                                                                   jclass,
+                                                                                   jstring jFile)
+{
+  const std::string path = GetJString(env, jFile);
+  const auto can_overwrite = []() { return true; };  // TODO
```
leoetlino: () is not necessary

Source/Android/jni/WiiUtils.cpp
```diff
@@ -0,0 +1,24 @@
...
+
+#include <string>
+
+#include <jni.h>
+
+#include "jni/AndroidCommon/AndroidCommon.h"
+
+#include "Core/WiiUtils.h"
+
+#ifdef __cplusplus
```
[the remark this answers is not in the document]
JosJuice: No, not really. I just did it to match the other files, but I might as well remove it.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/ui/main/MainPresenter.java
JosJuice: It might be a little confusing when reading the code quickly since this is a lambda function.

leoetlino: Hmm that's true, since Java lambdas are less visible than C++ lambdas... Feel free to ignore this then

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/utils/WiiUtils.java
JosJuice: I agree with you, but then I'd need to put some constants in WiiUtils.cpp that are identical to the constants in WiiUtils.java, since I can't access Java constants without jumping through hoops :/

But I suppose it does solve the problem to some degree.

JosJuice: Done.

leoetlino (approved):

PR #9455, opened 2021-01-19
TraversalProto: Minor cleanup

lioncash:
Mainly consists of converting enums over to `enum class` to prevent implicit conversions and namespace pollution, along with converting typedefs into using aliases.

### DIFF
--- a/Source/Core/Common/TraversalClient.h
+++ b/Source/Core/Common/TraversalClient.h
@@ -6,6 +6,7 @@
 #include <list>
 #include <memory>
 #include <string>
+#include <string_view>
 
 #include <enet/enet.h>
 
@@ -51,7 +52,7 @@ public:
   bool IsConnected() const { return m_State == State::Connected; }
 
   void Reset();
-  void ConnectToClient(const std::string& host);
+  void ConnectToClient(std::string_view host);
   void ReconnectToServer();
   void Update();
   void HandleResends();
--- a/Source/Core/Common/TraversalClient.h
+++ b/Source/Core/Common/TraversalClient.h
@@ -19,7 +19,7 @@ public:
   virtual ~TraversalClientClient() = default;
   virtual void OnTraversalStateChanged() = 0;
   virtual void OnConnectReady(ENetAddress addr) = 0;
-  virtual void OnConnectFailed(u8 reason) = 0;
+  virtual void OnConnectFailed(TraversalConnectFailedReason reason) = 0;
 };
 
 class TraversalClient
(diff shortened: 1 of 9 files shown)


iwubcode (approved):

Source/Core/Common/TraversalClient.h
```diff
@@ -46,8 +47,12 @@ class TraversalClient
   State GetState() const;
   FailureReason GetFailureReason() const;
 
+  bool HasFailed() const { return GetState() == State::Failure; }
+  bool IsConnecting() const { return GetState() == State::Connecting; }
+  bool IsConnected() const { return GetState() == State::Connected; }
```
BhaaLseN, line 52: Is there any particular reason for using `GetState()` instead of `m_state` here? Not necessarily asking you to change it, just trying to understand and learn about things I'd do differently otherwise.

Source/Core/Common/TraversalProto.h
```diff
@@ -1,44 +1,43 @@
...
   // [s->c] ...who asks the game host to send a UDP packet to the
   // client... (an ack implies success)
-  TraversalPacketPleaseSendPacket = 5,
+  PleaseSendPacket = 5,
   // [s->c] ...which the central server relays back to the client.
-  TraversalPacketConnectReady = 6,
+  ConnectReady = 6,
   // [s->c] Alternately, the server might not have heard of this host.
-  TraversalPacketConnectFailed = 7
+  ConnectFailed = 7
```
BhaaLseN, line 31: Do we want to add the trailing comma while we're at it? Or i the packet type final enough so we know for sure we wont need another?
```suggestion
  ConnectFailed = 7,
```

Same for pretty much all the other enums

Source/Core/Common/TraversalServer.cpp
```diff
@@ -185,8 +185,9 @@ static const char* SenderName(sockaddr_in6* addr)
 static void TrySend(const void* buffer, size_t size, sockaddr_in6* addr)
 {
 #if DEBUG
-  printf("-> %d %llu %s\n", ((TraversalPacket*)buffer)->type,
-         (long long)((TraversalPacket*)buffer)->requestId, SenderName(addr));
+  const auto* packet = static_cast<const TraversalPacket*>(buffer);
+  printf("-> %d %llu %s\n", static_cast<int>(packet->type),
```
BhaaLseN, line 189: I wonder if "replace `printf` with a `xxxx_LOG` statement" falls under "minor cleanup" - but that certainly shouldn't block this or anything. Especially when the Log Statement could get rid of the `#if DEBUG` (unless it means something significant for performance)

BhaaLseN (approved):

PR #9456, opened 2021-01-19
Breakpoint from string flag parsing using token

trevor403:
When searching the breakpoint string for `"b"` for the `break_on_hit` flag you can get a false hit because the hex address may include the letter `b`. For example, `80,0cb,46c nl` was incorrectly enabling break on hit.

If you read the next token from the string stream, you will have the flags alone. You can then search these for each letter and you won't get a false hit.

This is quite a neat bug :smile:

### DIFF
--- a/Source/Core/Core/PowerPC/BreakPoints.cpp
+++ b/Source/Core/Core/PowerPC/BreakPoints.cpp
@@ -65,12 +65,13 @@ void BreakPoints::AddFromStrings(const TBreakPointsStr& bp_strings)
   for (const std::string& bp_string : bp_strings)
   {
     TBreakPoint bp;
-    std::stringstream ss;
-    ss << std::hex << bp_string;
-    ss >> bp.address;
-    bp.is_enabled = bp_string.find('n') != bp_string.npos;
-    bp.log_on_hit = bp_string.find('l') != bp_string.npos;
-    bp.break_on_hit = bp_string.find('b') != bp_string.npos;
+    std::string flags;
+    std::istringstream iss(bp_string);
+    iss >> std::hex >> bp.address;
+    iss >> flags;
+    bp.is_enabled = flags.find('n') != flags.npos;
+    bp.log_on_hit = flags.find('l') != flags.npos;
+    bp.break_on_hit = flags.find('b') != flags.npos;
     bp.is_temporary = false;
     Add(bp);
   }


JosJuice:
@dolphin-emu-bot rebuild

Source/Core/Core/PowerPC/BreakPoints.cpp
```diff
@@ -65,12 +65,14 @@ void BreakPoints::AddFromStrings(const TBreakPointsStr& bp_strings)
   for (const std::string& bp_string : bp_strings)
   {
     TBreakPoint bp;
+    std::string flags;
     std::stringstream ss;
```
sepalani, line 69: Can't this part be replaced with:
```c++
std::istringstream iss(bp_string);
iss >> std::hex >> bp.address;
iss >> flags;
```

sepalani (commented):
This change make sense to me. By "80,0cb,46c nl" you meant `800cb46c nl`, right? I'm not sure we support comma separated hex values.

Untested, will test later today.

Source/Core/Core/PowerPC/BreakPoints.cpp
trevor403: It certainly can simplified to that!

trevor403:
> This change make sense to me. By "80,0cb,46c nl" you meant `800cb46c nl`, right? I'm not sure we support comma separated hex values.
> 
> Untested, will test later today.

For some reason when saving breakpoints, my GameSettings contains comma separated hex values. They Load just fine too. Maybe this is a quirk of `std::hex`? I am not sure.

sepalani:
@dolphin-emu-bot rebuild

sepalani (approved):
Tested and LGTM.

I suspect the comma is coming from your locale, that would explain why you have no issues when saving and loading.

This locale/comma issue should be addressed in its own PR as it needs to fix loading and saving BPs, on top of other saved config I guess. I'll probably address it by next week if I manage to find some time.

lioncash (approved):

lioncash:
LGTM once changes are squashed into a single commit

trevor403:
@lioncash should be ready to merge once the checks run

trevor403:
It seems as though checks are not running on my squashed commit

sepalani:
@dolphin-emu-bot rebuild

trevor403:
The Pending pr-android status check seems to have run successfully on the Buildbot. Maybe it did not properly report the test status back to GitHub

smurf3tte:
This bug was also fixed in #9334...

leoetlino:
#9334 will take longer to review so I think we should merge this first

PR #9457, opened 2021-01-20
Fix CMake MSVC build

JosJuice:
Also removing an undef for a define that PR #9435 removed.

### DIFF
--- a/Source/Core/Common/BitUtils.h
+++ b/Source/Core/Common/BitUtils.h
@@ -378,7 +378,7 @@ constexpr int CountLeadingZeros(uint64_t value)
 {
 #if defined(__GNUC__)
   return value ? __builtin_clzll(value) : 64;
-#elif defined(_MSC_VER)
+#elif defined(_MSC_VER) && __cplusplus >= 201803L
   if (std::is_constant_evaluated())
   {
     return CountLeadingZerosConst(value);
@@ -397,7 +397,7 @@ constexpr int CountLeadingZeros(uint32_t value)
 {
 #if defined(__GNUC__)
   return value ? __builtin_clz(value) : 32;
-#elif defined(_MSC_VER)
+#elif defined(_MSC_VER) && __cplusplus >= 201803L
   if (std::is_constant_evaluated())
   {
     return CountLeadingZerosConst(value);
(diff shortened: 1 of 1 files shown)


Source/Core/Common/BitUtils.h
```diff
@@ -378,7 +378,7 @@ constexpr int CountLeadingZeros(uint64_t value)
 {
 #if defined(__GNUC__)
   return value ? __builtin_clzll(value) : 64;
-#elif defined(_MSC_VER)
+#elif defined(_MSC_VER) && __cplusplus >= 201803L
```
iwubcode, line 381: Was `is_constant_evaluated()` not available before that version?

shuffle2:
Doesn't this mean the cmake build isn't hitting this check? https://github.com/dolphin-emu/dolphin/blob/master/Source/PCH/pch.h#L10

Would prefer not to pepper random version checks all over the place.

shuffle2:
I see. the problem is here: https://github.com/dolphin-emu/dolphin/blob/c8cb330df04431b8414efaffaee01cf984572d33/Source/CMakeLists.txt#L19

Please synchronize cmake settings to match msbuild

Source/Core/Common/BitUtils.h
iwubcode: shuffle mentioned it was a C++20 feature, I didn't realize that..

JosJuice:
So... I should add an `if (MSVC)` check in CMakeLists.txt and set `CMAKE_CXX_STANDARD` to 20 for MSVC? Or does CMake have a way to set it to "latest" like we're using for MSBuild?

I tried doing the opposite in PR #8856, but not everyone liked the idea.

shuffle2:
It seems that would work ( https://gitlab.kitware.com/cmake/cmake/-/merge_requests/1892/diffs#12f7e66fe561d1b417b2eca25624b335c52623b6 )

JMC47:
Is this stalled?  Or can it be finished/merged?

JosJuice:
Personally I like the PR the way it is, but I guess shuffle2's way is also okay. I'll open a new PR with shuffle2's approach and let someone else decide which one to merge.

JosJuice:
Opened it as PR #9556.

leoetlino:
Closing in favour of #9556

PR #9460, opened 2021-01-20
WiimoteReal: Fix crash on real Wii Remote disconnect on Windows.

jordan-woyak:
Code for attaching a real Wii remote's HID interface to our emulated Bluetooth device is a bit messy.
A dangling pointer in the disconnect routine was causing a crash on Windows.

Fixes:
https://bugs.dolphin-emu.org/issues/12329
https://bugs.dolphin-emu.org/issues/12343
https://bugs.dolphin-emu.org/issues/12351

Possibly fixes this issue on macOS?
https://bugs.dolphin-emu.org/issues/9520

### DIFF
--- a/Source/Core/Core/HW/WiimoteReal/WiimoteReal.cpp
+++ b/Source/Core/Core/HW/WiimoteReal/WiimoteReal.cpp
@@ -916,7 +916,9 @@ static void TryToConnectBalanceBoard(std::unique_ptr<Wiimote> wm)
 static void HandleWiimoteDisconnect(int index)
 {
   Core::RunAsCPUThread([index] {
-    g_wiimotes[index] = nullptr;
+    // The Wii Remote object must exist through the call to UpdateSource
+    // to prevent WiimoteDevice from having a dangling HIDWiimote pointer.
+    const auto temp_real_wiimote = std::move(g_wiimotes[index]);
     WiimoteCommon::UpdateSource(index);
   });
 }


iwubcode (approved):

Source/Core/Core/HW/WiimoteReal/WiimoteReal.cpp
```diff
@@ -916,7 +916,9 @@ static void TryToConnectBalanceBoard(std::unique_ptr<Wiimote> wm)
 static void HandleWiimoteDisconnect(int index)
 {
   Core::RunAsCPUThread([index] {
-    g_wiimotes[index] = nullptr;
+    // The Wii Remote object must exist through the call to UpdateSource
+    // to prevent WiimoteDevice from having a dangling HIDWiimote pointer.
+    const auto temp_real_wiimote = std::move(g_wiimotes[index]);
```
iwubcode, line 921: I guess alternatively we could have just swapped these calls?

jordan-woyak: No. Unfortunately `UpdateSource` gives `WiimoteDevice` a new pointer by grabbing it from `g_wiimotes`.

shuffle2: Isn't `UpdateSource` going to reach here: https://github.com/dolphin-emu/dolphin/blob/0029ca84b07f44e3ce8c80a44fa0285a966b34d3/Source/Core/Core/HW/Wiimote.cpp#L75 and `.get()` the moved-from index?

shuffle2: ok, the comment is a bit confusing to me for some reason. anyway lgtm

StLouisCPhT:
Still hangs and crashes in MacOS 11.0

PR #9466, opened 2021-01-23
DolphinQt: Move common controller interface logic to its own widget

iwubcode:
Pulled from #8823 to make it easier to merge

Should have no functional or visual changes

### DIFF
--- a/Source/Core/DolphinQt/Config/CommonControllersWidget.cpp
+++ b/Source/Core/DolphinQt/Config/CommonControllersWidget.cpp
@@ -1,0 +1,68 @@
+// Copyright 2021 Dolphin Emulator Project
+// Licensed under GPLv2+
+// Refer to the license.txt file included.
+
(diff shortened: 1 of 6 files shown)


Source/Core/DolphinQt/Config/CommonControllersWidget.cpp
```diff
@@ -0,0 +1,68 @@
...
+  m_common_layout = new QVBoxLayout();
+  m_common_bg_input = new QCheckBox(tr("Background Input"));
+  m_common_configure_controller_interface = new QPushButton(tr("Alternate Input Sources"));
+
+  m_common_layout->addWidget(m_common_bg_input);
+  m_common_layout->addWidget(m_common_configure_controller_interface);
+
+  m_common_box->setLayout(m_common_layout);
+
+  auto* layout = new QVBoxLayout;
```
iwubcode, line 37: This layout block is the only difference from the current code

PR #9467, opened 2021-01-23
DolphinQt: Move gamecube controller ui logic to its own widget

iwubcode:
Pulled from #8823 to make it easier to merge

Should have no functional or visual changes

### DIFF
--- a/Source/Core/DolphinQt/Config/GamecubeControllersWidget.cpp
+++ b/Source/Core/DolphinQt/Config/GamecubeControllersWidget.cpp
@@ -1,0 +1,198 @@
+// Copyright 2021 Dolphin Emulator Project
+// Licensed under GPLv2+
+// Refer to the license.txt file included.
+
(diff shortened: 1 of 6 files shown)


Source/Core/DolphinQt/Config/GamecubeControllersWidget.cpp
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    int controller_row = m_gc_layout->rowCount();
+    m_gc_layout->addWidget(gc_label, controller_row, 0);
+    m_gc_layout->addWidget(gc_box, controller_row, 1);
+    m_gc_layout->addWidget(gc_button, controller_row, 2);
+  }
+  m_gc_box->setLayout(m_gc_layout);
+
+  auto* layout = new QVBoxLayout;
```
iwubcode, line 81: This layout block is the only difference from the current code

iwubcode:
@leoetlino - this one should be good to go!

PR #7412, opened 2018-09-13
Stitch together overlapping XFB regions in order of XFB copy creation.

AdmiralCurtiss:
This is a fun one!

So, the observable issue here is that if you try to play Rogue Leader in Dolphin and happen to pause during gameplay, the output image starts flickering between a properly rendered frame and something along the lines of this:

_wrong_output

What you're seeing in the lower part here is a pretty old frame that played during the last widescreen cutscene. How did that end up here? We should be instead seeing this instead!

_right_output

Well, I'm slightly simplfying here (all of this is double-buffered to two separate memory addresses and only one of them visibly shows the problem), but here's the basic idea.

During the cutscene, Rogue Leader makes XFB copies at 0x2bbae0. When the cutscene eventually ends (or you skip it) and the game switches to a more 4:3 output, it switches the location of its XFB copies to 0x25cae0. At this point, it only outputs a single full XFB copy per frame, so we just take it directly from the cache when outputting.

When you then pause the game during gameplay, the XFB copy method is switched again in order to do the half XFB copy stuff for antialiasing or whatever, this time to both 0x25cae0 and 0x2a2ae0. These copies have a size of 0x46a00 bytes each, which together make a full frame. That occupies a memory region from 0x25cae0 to 0x2e94e0.

Note how 0x2e94e0 is after 0x2bbae0, the old location of the cutscene XFB. When time comes to scan out the image, Dolphin attempts to stitch together all XFBs in the 0x25cae0-0x2e94e0 region, and finds this old cutscene XFB right in the middle of it. Finding nothing wrong with it, it proceeds to stitch that image into the output, and that's how we end up with the wrong image.

If this is a bit confusing, I've made a little visualization.

_cutscene

At some point in the past, the game made an XFB copy of a cutscene.

_first_xfb
_second_xfb_correct

When the pause menu is rendered, the two XFB copies only partially overwrite the old cutscene XFB. The game expects the output to be something like this, and only scans out the newly rendered parts to screen.

_second_xfb_wrong

But Dolphin both doesn't realize the cutscene copy has been partially overwritten and also fails to keep the order in mind. Instead, it stitches them together in the order they come back from the multimap iterator, which is memory address order.

[...]

PR #7408, opened 2018-09-11
Do not consider XFB copies for invalidating textures when loading a new texture.

AdmiralCurtiss:
This fixes severe image flickering in some cutscenes of Twin Snakes. The game appears to sometimes load a previously made XFB copy as a texture before it is actually rendered to the screen, which we took as an invitation to invalidate the XFB copy.

Someone more familiar with the texture cache code should probably take a look at this, but it seems like a pretty safe change.

PR #9470, opened 2021-01-24
Disable XFBToTexture in Skies of Arcadia

Stevoisiak:
Fixes an issue where screen will sometimes flash purple and green when loading save files, during certain cutscenes, and when entering a different area.

### DIFF
--- a/Data/Sys/GameSettings/GEA.ini
+++ b/Data/Sys/GameSettings/GEA.ini
@@ -18,4 +18,5 @@ SafeTextureCacheColorSamples = 512
 
 [Video_Hacks]
 EFBToTextureEnable = False
+XFBToTextureEnable = False
 


JMC47:
Does changing this setting result in a performance issue that outweighs the benefit of fixing the visual bugs?

As always, whenever there's a potential for flashing screens, I like the err on the side of caution and fix it, but I also like to keep an eye on performance changes due to many mobile devices struggling with Dolphin.

Stevoisiak:
At a quick glance, I don't notice a significant performance increase while running around Pirates' Isle. Stays between 9 - 12% CPU usage whether enabled or disabled. (Windows 10, Intel Core i5-7600K, Nvidia GTX 1070)

For reference, below is a screenshot of the flashing screen.

flashing green and pink screen

JosJuice:
The performance impact of XFB copies to RAM should be measured by how high of a framerate you can achieve with the framerate limiter disabled, not the CPU usage, since the CPU is idle when waiting for the GPU to supply an XFB copy.

Hmm... We already have `EFBToTextureEnable = False` for this game. @JMC47, didn't you tell me once that `XFBToTextureEnable = False` doesn't have too much of a performance impact if you're already using `EFBToTextureEnable = False` anyway?

Stevoisiak:
With framelimiter off, performance in the intro cutscene seems similar. (100 to 200 FPS)

iwubcode:
@JosJuice - I'll take another look at the code (it's had some cleanups since I worked on it) but AFAIK XFB and EFB are completely different code pathes, setting one shouldn't affect the other

JosJuice:
Yes, they're different code paths, but I think the logic was something like that if you have to stall for the GPU many times per frame anyway then stalling an extra time per frame doesn't hurt much.

iwubcode:
@JosJuice - ah possibly, that's more low level than I care to get :P

dirextric:
Just wanted to note here that the issue prompting this PR and [#12392](https://bugs.dolphin-emu.org/issues/12392) on the Dolphin bug tracker look like they're the same issue. I've been facing purple screens as well in The Adventures of Tintin: The Game, with exactly the same symptoms and solution. According to the bug report, this is a regression since 5.0-10000, which means it probably shouldn't be worked around in this way. @Stevoisiak, would you be able to test with default settings on a version of Dolphin before 5.0-10000 to confirm this is indeed the same issue?

Stevoisiak:
@dirextric I still get the purple flashing in Dolphin 5.0-9995 if XFB to Texture is enabled.

dirextric:
It seems I may have sprang to a conclusion too quickly with my previous assessment. I finally got time to test The Adventures of Tintin, and like @Stevoisiak, I still encountered purple flashing on Dolphin 5.0-9974. However, I had no such issues on Dolphin 5.0. I was able to trace my issue to 5.0-5874 as the first broken build, and I suspect the issue here also stems from this version. This version implemented Hybrid XFB, which (to my understanding) means that my issue (and probably this one, too) is an unfortunate consequence of the new implementation but not really a regression. All of this is to say that this issue and [#12392](https://bugs.dolphin-emu.org/issues/12392) are definitely different and the workaround in this PR is probably the best way to handle the situation. I'll open up a separate PR for The Adventures of Tintin, but before I do, I have a question for someone who knows this part of the code: in The Adventures of Tintin, the purple flashing is solved either by disabling Store XFB Copies to Texture Only or by enabling Immediately Present XFB (or doing both, for that matter). Which one of these is the "best" solution, so to speak?

JosJuice:
Disabling Store XFB Copies to Texture Only is the accurate solution and enabling Immediately Present XFB is the solution with better performance. For what it's worth, the default behavior in 5.0 is similar to enabling Immediately Present XFB, which is probably why it worked for you in 5.0.

AdmiralCurtiss:
'Purple flickering when XFB to Tex' reminds me of #7408 and #7412. I wonder if we're invalidating something we shouldn't.

JMC47:
That's most likely what's happened.  It's just figuring out what?  Once we merge this, it becomes much harder to keep track of what games need to be tested.

JMC47:
This is fixed by https://github.com/dolphin-emu/dolphin/pull/9694 as far as I can tell.  Please move testing over to that pull request.

PR #9471, opened 2021-01-24
Android: Fix motion control settings not saving to file

sspacelynx:
Motion controls are not saved to file when closing the dialog, so when changing this setting, then closing and re-opening the app, it gets back to its last saved position (which could be any since it is not saved anywhere).

I don't know if this is the correct way to do this, but it works as expected as far my testing goes.

### DIFF
--- a/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/activities/EmulationActivity.java
+++ b/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/activities/EmulationActivity.java
@@ -912,7 +912,8 @@ public final class EmulationActivity extends AppCompatActivity
               updateWiimoteNewImuIr(indexSelected);
               NativeLibrary.ReloadWiimoteConfig();
             });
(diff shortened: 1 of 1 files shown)


JosJuice:
We already have a call to `saveSettings` in `onStop`. Is that call not functioning?

sspacelynx:
It doesn't seem to work at least on my end.

JosJuice:
Have you checked with a debugger whether it is being called and whether the call successfully saves settings to disk?

sspacelynx:
From initial debugging, `onStop` is called correctly. The weird thing is even with my line added (which is redundant at this point) settings are not being saved so there might be something else. At this point I'm not sure what to do with this PR.

Ebola16:
Super behind on some tasks, sorry. I can confirm onStop has been causing me issue for saving settings in the Wii Overlay PR. I'll work on setting up a simpler example within the next day.

Ebola16:
@JosJuice I think we had this issue once before where settings were being saved in the C++ side and neither of us looked into where specifically since the code seemed to work fine. I can't look into this right now but I hope to soon.

PR #9472, opened 2021-01-24
JitArm64: Update registers last used before start of instruction

JosJuice:
Let's reset `m_last_used` for each register that will be used in an instruction before we start allocating any of them, so that one of the earlier allocations doesn't spill a register that we want in a later allocation.

See the individual commits for details.

### DIFF
--- a/Source/Core/Core/PowerPC/JitArm64/Jit.cpp
+++ b/Source/Core/Core/PowerPC/JitArm64/Jit.cpp
@@ -694,6 +694,15 @@ void JitArm64::DoJit(u32 em_address, JitBlock* b, u32 nextPC)
     if (!SConfig::GetInstance().bEnableDebugging)
       js.downcountAmount += PatchEngine::GetSpeedhackCycles(js.compilerPC);
 
+    // Skip calling UpdateLastUsed for lmw/stmw - it usually hurts more than it helps
+    if (op.inst.OPCD != 46 && op.inst.OPCD != 47)
+      gpr.UpdateLastUsed(op.regsIn | op.regsOut);
+
+    BitSet32 fpr_used = op.fregsIn;
+    if (op.fregOut >= 0)
+      fpr_used[op.fregOut] = true;
+    fpr.UpdateLastUsed(fpr_used);
+
     // Gather pipe writes using a non-immediate address are discovered by profiling.
     bool gatherPipeIntCheck = js.fifoWriteAddresses.find(op.address) != js.fifoWriteAddresses.end();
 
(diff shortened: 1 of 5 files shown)


Source/Core/Core/PowerPC/JitArm64/Jit.cpp
```diff
@@ -694,6 +694,15 @@ void JitArm64::DoJit(u32 em_address, JitBlock* b, u32 nextPC)
     if (!SConfig::GetInstance().bEnableDebugging)
       js.downcountAmount += PatchEngine::GetSpeedhackCycles(js.compilerPC);
 
+    // Skip calling UpdateLastUsed for lmw/stmw - it usually hurts more than it helps
+    if (op.inst.OPCD == 46 || op.inst.OPCD == 47)
```
leoetlino, line 698: Shouldn't this conditional be inverted? (wish we had an enum for the opcodes, had to check whether 46 and 47 were really lmw/stmw)

JosJuice: Ah, yes :(

The tricky part of making something like an enum is that many instructions are identified by not just the `OPCD` field but also other fields, and it would feel weird to add something that doesn't work for all instructions.

leoetlino: Yeah, that's true... Maybe calling it PrimaryOpcode (to match the interpreter's terminology) would make it clear enough  :man_shrugging: either way, that's out of the scope of this PR.

Source/Core/Core/PowerPC/JitArm64/JitArm64_FloatingPoint.cpp
```diff
@@ -220,15 +230,20 @@ void JitArm64::fselx(UGeckoInstruction inst)
     m_float_emit.FCMPE(EncodeRegToDouble(VA));
   }
 
-  const bool single = fpr.IsSingle(b, true) && fpr.IsSingle(c, true);
-  const RegType type = single ? RegType::LowerPairSingle : RegType::LowerPair;
-  const auto reg_encoder = single ? EncodeRegToSingle : EncodeRegToDouble;
+  const bool bc_single = fpr.IsSingle(b, true) && fpr.IsSingle(c, true);
```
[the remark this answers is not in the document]
JosJuice: I should draw a line somewhere for which ones I should use a lambda for (since using it for a simple `fpr.IsSingle(a)` is a bit silly), and I think this is simpler enough compared to `fpr.IsSingle(a, !packed) && (!use_b || fpr.IsSingle(b, !packed)) && (!use_c || fpr.IsSingle(c, !packed))` that it makes sense to not use a lambda here. I can rename it, though.

leoetlino (approved):

JosJuice:
Apparently the asserts I added are getting triggered in certain games (e.g. Wind Waker) :/

Will try to investigate.

PR #9473, opened 2021-01-25
Texture lookup fix

Miksel12:
Fixes Texture lookup by applying point sampling.

Old:
This adds an epsilon in the texture sampling stage to prevent rounding errors.

An epsilon of 0.5 gave the same results (when it comes to artifacts)as point sampling without actually point sampling so there is no behavioral change. 0.5 was chosen as that is the lowest value that can always be represented when converting a 24bit signed integer to a float.
The only fifolog that this doesn't fix is Just Dance 2 but the same artifacts can be found when point sampling and this depends on the hardware, so this is probably a rounding error(https://github.com/dolphin-emu/dolphin/pull/9473#issuecomment-768564342, https://github.com/dolphin-emu/dolphin/pull/9473#issuecomment-767937443).
But Just Dance 2 is improved compared to master, at least on my hardware (Nvidia).

OLD:
This solution was provided by u/fortsnek47 on reddit (https://www.reddit.com/r/EmuDev/comments/l0y7sg/gc_fixing_those_vp6_videos/).
I don't have any games with texture lookup issues so I can't test these changes.

Games that can be tested: https://wiki.dolphin-emu.org/index.php?title=Template:Problems/VP6_Videos

### DIFF
--- a/Source/Core/VideoCommon/GXPipelineTypes.h
+++ b/Source/Core/VideoCommon/GXPipelineTypes.h
@@ -20,7 +20,7 @@ namespace VideoCommon
 // As pipelines encompass both shader UIDs and render states, changes to either of these should
 // also increment the pipeline UID version. Incrementing the UID version will cause all UID
 // caches to be invalidated.
-constexpr u32 GX_PIPELINE_UID_VERSION = 2;  // Last changed in PR 9122
+constexpr u32 GX_PIPELINE_UID_VERSION = 3;  // Last changed in PR 9473
 
 struct GXPipelineUid
 {
--- a/Source/Core/VideoCommon/PixelShaderGen.cpp
+++ b/Source/Core/VideoCommon/PixelShaderGen.cpp
@@ -16,6 +16,7 @@
 #include "VideoCommon/LightingShaderGen.h"
 #include "VideoCommon/NativeVertexFormat.h"
 #include "VideoCommon/RenderState.h"
+#include "VideoCommon/SamplerCommon.h"
 #include "VideoCommon/VertexLoaderManager.h"
 #include "VideoCommon/VideoCommon.h"
 #include "VideoCommon/VideoConfig.h"
@@ -808,7 +809,15 @@ ShaderCode GeneratePixelShaderCode(APIType api_type, const ShaderHostConfig& hos
       }
 
       out.Write("\tint3 iindtex{} = ", i);
-      SampleTexture(out, "float2(tempcoord)", "abg", texmap, stereo, api_type);
+
+      FourTexUnits& texUnit = bpmem.tex[(texmap >> 2) & 1];
+      TexMode0& tm0 = texUnit.texMode0[texmap & 3];
+
+      SampleTexture(out,
+                    SamplerCommon::IsBpTexMode0PointFiltering(tm0) ?
+                        "((float2(tempcoord >> 7) + 0.5f) * 128.0f)" :
+                        "(float2(tempcoord))",
+                    "abg", texmap, stereo, api_type);
     }
   }
 
@@ -1201,7 +1210,15 @@ static void WriteStage(ShaderCode& out, const pixel_shader_uid_data* uid_data, i
         out.Write("\ttevcoord.xy = int2(0, 0);\n");
     }
     out.Write("\ttextemp = ");
-    SampleTexture(out, "float2(tevcoord.xy)", texswap, stage.tevorders_texmap, stereo, api_type);
+
+    FourTexUnits& texUnit = bpmem.tex[(stage.tevorders_texmap >> 2) & 1];
+    TexMode0& tm0 = texUnit.texMode0[stage.tevorders_texmap & 3];
+
+    SampleTexture(out,
+                  SamplerCommon::IsBpTexMode0PointFiltering(tm0) ?
+                      "((float2(tevcoord.xy >> 7) + 0.5f) * 128.0f)" :
+                      "float2(tevcoord.xy)",
+                  texswap, stage.tevorders_texmap, stereo, api_type);
   }
   else
   {


Miksel12:
Ok, that doesn't seem right. Guess I should have copied the code 1:1.

JMC47:
@dolphin-emu-bot rebuild

Miksel12:
I have been testing and with the solution from u/fortsnek47 you get very low precision textures, which makes sense as you are throwing away precision. But some experimentation show better results.

JMC47:
seems like fifoci doesn't like it so far, though.

Miksel12:
Really weird, I don't have any issues. u/fortsnek47 also says he has no issues: https://www.reddit.com/r/EmuDev/comments/l0y7sg/gc_fixing_those_vp6_videos/gkqjk5f?utm_source=share&utm_medium=web2x&context=3

To circumvent the loss of precision I tried this: https://github.com/Miksel12/dolphin/commit/b33207d2696a5846bf3226af13c8e021393c9dcf
Which seemed to work but randomly started showing different behaviour. It all seems a bit iffy.

Miksel12:
Okay, this seems to work now. I tested 2 random vp6 fifo logs which are now perfectly rendered (not sure if that is luck) and there is no loss of precision. 1 << 6 comes from 0.5 * 128.

Miksel12:
I do have some very weird inconsistencies. It seems to come from running master before running a fifo log on this build.
Running fifo log on this build, great result -> Test on master, expected result from master -> Test on this build again, ton of artefacts.

JMC47:
Okay, I know what's wrong.  You're having shaders carry between builds I bet.

mbc07:
Didn't we have a shader cache version somewhere in the code? Bumping the version should force Dolphin to regenerate the shader cache...

Miksel12:
Yeah I was looking for that but could find it.

endrift:
Zero percent surprised it's this broken, since the reddit version seems to be just substituting the epsilon for a very large value instead. Not sure the author realized that this "recentering" was applied _after_ normalization instead of before.

Miksel12:
> Zero percent surprised it's this broken, since the reddit version seems to be just substituting the epsilon for a very large value instead. Not sure the author realized that this "recentering" was applied _after_ normalization instead of before.

Well, the current implementation is my way to prevent precision loss but is wrong. `((float2(tevcoord.xy >> 7) + 0.5f) * 128.0f)` is correct for point-sampled textures, meaning that the filtering should be different for different sample type.  You can view the discussion on the reddit post. That is way too complex for me so I'll be closing. Though a FifoCI test run with the 'correct' point-sample might be interesting to see what it does to the vp6 fifologs.

Miksel12:
FifoCI should be able to run now, I was using C-style casts by accident.

mbc07:
@Miksel12 I think [that's the line](https://github.com/dolphin-emu/dolphin/blob/master/Source/Core/VideoCommon/GXPipelineTypes.h#L23) that controls shader cache version. Bumping it should invalidate existing caches...

Miksel12:
Ah yes, all inconsistencies are gone now. `float2(tevcoord.xy + (1 << 6))` clearly breaks vp6 video's, so it seemed to work sometime because it was using the cache which used `((float2(tevcoord.xy >> 7) + 0.5f) * 128.0f)` in those cases.

@mbc07 could you test your Just Dance games? Those clearly showed a epsilon doesn't work so I'm interested to see if this works. You should ignore all the low res texture for now and maybe run portable as this will invalidate the shader cache (though I assume only for games your boot in this build).

mbc07:
Can someone trigger the build bot? I can't compile this PR manually at the moment...

JMC47:
The current version of the PR does seem to fix things in the cases I know about.

Miksel12:
I also tested some texture seams fifo's I found on the issue tracker and it seems to fix those.

mbc07:
The affected Just Dance games seems to behave exactly the same as in PR #9407: completely fixed on Intel, improved/worsened on NVIDIA, depending of the game. If it helps, I recorded FIFO Logs of the same tests I've done in the other PR:

* [JD-fifolog.zip](https://github.com/dolphin-emu/dolphin/files/5876130/JD-fifolog.zip)

Miksel12:
Hmm, that is a shame.

JMC47:
Even if there is a performance penalty, I'd love to see an integer texture lookup option to see if it fixes everything.

Miksel12:
> Even if there is a performance penalty, I'd love to see an integer texture lookup option to see if it fixes everything.

Yeah, that would be awesome to see.

fortsnek9348:
> Even if there is a performance penalty, I'd love to see an integer texture lookup option to see if it fixes everything.

Shouldn't be much different from rounding the coordinates. For point-sampled textures. But for HLSL `Load`, docs say there's no wrapping.

> The affected Just Dance games seems to behave exactly the same as in PR #9407

Can any clues be gleamed from RenderDoc or something? Might be caused by other sources of inaccuracy like triangle interpolation or the pixel offset.

mbc07:
Is it possible to hook RenderDoc in Dolphin? If affirmative, the FIFO Logs I attached should be enough to verify that. I recorded 4 frames on each log, you can play them with the [FIFO Player](https://dolphin-emu.org/docs/guides/fifo-player-documentation-testers-and-developers/)...

JMC47:
You can hook renderdoc into Dolphin, yeah.  That's how we solved the Monster Hunter Tri bloom offset.

fortsnek9348:
Looking at JD3-dancer. 6 textures, clamped, point-sampled, 5 TEV stages, 5 texgens, 2 IND stages. Similar to the VP6 videos I've seen.

JMC47:
One of the things that'd be nice is to get this consistent regardless of driver, which I'd hope an integer solution would do.

fortsnek9348:
Oh. That's nice. RenderDoc allows you to edit shaders.

It seems applying either `float2(texcoord.xy | int2(1, 1))` or `float2(texcoord.xy >> 7) * 128.0f + 64.0f` to both the indirect texcoords and tev texcoords is enough to fix glitch in `JD3-dancer`. But `JD2-dancer` still has some artifacts.

fortsnek9348:
It seems the inaccuracy comes from triangle interpolation, or the vertex offsets. In `JD2-dancer`, the sample I'm looking at has a `tex3.y` value of `0.167'499'185`, which is `66.999'674` when scaled by `texdim[3].w/128`, which was why a `int2(0, 1)` TEV bias worked. Rounding the fixed-point texcoords (or using `Load`/`texelFetch`) probably won't fix that. The correct value of `tex3.y` should probably be at least `0.167'5`.

Miksel12:
Does that explain why it only happens on Nvidia hardware?

mbc07:
So, I did some more testing and there's some points I found interesting:

* **DLC songs:** if the song you're playing is a DLC, the dancers won't have any corruption, even when running on master! I confirmed this behavior both in Just Dance 2 and 3 (there's no DLC for the first Just Dance), only the songs that ships in the game disc will have the corrupted lines. 
* **Just Dance 4 loading screen:** on Just Dance 4, there's a loading screen that appears only one time (between the Ubisoft logo and the "Press A" screen), that loading screen also has very minor corruption (similar to JD3). Everything else from that game is corruption-free, though.

I recorded a new set of FIFO Logs from those edge cases: [JD-fifolog2.zip](https://github.com/dolphin-emu/dolphin/files/5877043/JD-fifolog2.zip)

-----

Analyzing the game files on disc, Just Dance 1, 2 and 3 uses Bink video format (.bik) for the dancers. The DLC files also uses Bink, I checked that by unpacking one of the DLC files of each game. Just Dance 4 switched to Mobiclip video format (.mo) and has no corruption issues, with the exception of the first loading screen, which still uses Bink video and has minor corruption. 

Finally, with Just Dance 2014 Ubisoft switched (again) to VP8 codec in a standard WebM container and kept that way until the very last Just Dance game they released on the Wii. JD games using VP8/WebM also has zero corruption issues...

fortsnek9348:
> Does that explain why it only happens on Nvidia hardware?

If it's the side effect of rolling the dice of rounding errors/differences in triangle interpolation. Or the vertex offset/pixel center is not accurate enough. I don't think it's just (one) FP rounding error, though, as the difference between `0.167'499'185` and `0.167'5` is more than the lowest bit (`3E2B8515` vs `3E2B851F`).

Could do with more eyes on this. Checking the same pixel in the software rasteriser would be nice, but it seems to be quite broken for this FIFO log.

JMC47:
@Pokechu22 could see if we could get the software renderer could be fixed.

mbc07:
> Checking the same pixel in the software rasteriser would be nice, but it seems to be quite broken for this FIFO log.

By broken you mean "doesn't render at all" or that the dancers are way more glitched? If it's the latter, I just gave a try by playing the game directly with the software renderer and indeed the dancers looked as glitched as in the FIFO log  ¯\\\_(ツ)_/¯

fortsnek9348:
More glitched. A different kind of glitch. Multiple vertical strips. I don't have the game so I'm only looking at FIFO logs.

mbc07:
Yep, this happens when playing the game directly with the software renderer too (just checked). It's not exclusive to the FIFO Logs...

Miksel12:
> If it's the side effect of rolling the dice of rounding errors/differences in triangle interpolation. Or the vertex offset/pixel center is not accurate enough. I don't think it's just (one) FP rounding error, though, as the difference between `0.167'499'185` and `0.167'5` is more than the lowest bit (`3E2B84E8` vs `3E2B851F`).

Couldn't integer math solve this issue?

fortsnek9348:
Well, I'm not sure how Flipper interpolates vertex attributes, and I'm pretty sure you can't "integer math" attribute interpolation on GPUs. But I'm sure somebody will get some idea involving compute shaders or manually computing barycentric coordinates in the pixel shader. But... if I try to do the interpolation in, say, wxMaxima, I don't get the desired value. Instead, I get `1.674967609189695733...e-1`.

Or the inaccuracy is caused by vertex rounding, if that happens. Does Flipper round vertices? Do modern GPUs round vertices? The quad in `JD2-dancer` is not aligned to screen edges. It's vertex positions seem arbitrary as if they were placed by an artist.

But if I switch to 6x internal resolution, remove the 1/12 vertex offset, and round to pixel corners, the artifacts are not removed where I thought they would be.

Or maybe Flipper rounds its barycentric coordinates. Not sure how likely that is. Would Flipper have a full 24 bits of FP precision there?

fortsnek9348:
Another attempt. Looking at the pixel shader, how should those `fixpoint_uv0` variables be rounded? Currently, it's just the default `int2` cast.

But, with the `float2(texcoord.xy | int2(1, 1))`  fudge for all texture lookups and using `floor` to round `fixpoint_uv*`, I haven't found any artifacts in the examples I have.

	out.Write("\tint2 fixpoint_uv%d = int2(floor(", i);
	out.Write("(tex%d.z == 0.0 ? tex%d.xy : tex%d.xy / tex%d.z)", i, i, i, i);
	out.Write(" * " I_TEXDIMS "[%d].zw));\n", i);

JMC47:
I'd be interested to see what fifoci says about that :)

JMC47:
@dolphin-emu-bot rebuild

Miksel12:
Ah, didn't see your comment @fortsnek9348, I pushed a different approach based on the software renderer. Now it only uses point sampling when it should.

JMC47:
I'd like to try to get fifoci to test as many solutions as possible still.  Finding a good solution to this would be a huge boon.

Miksel12:
This change has quite a big impact because before texture that should be point sampled were not but now they are. Meaning that you'll be seeing more jaggies but that is actually more faithfull.

JMC47:
Could that explain why certain textures had artifacting, too?

Miksel12:
You mean before? Yeah, that was the opposite, textures that shouldn't be point sampled were point sampled.

fortsnek9348:
> FourTexUnits& texUnit = bpmem.tex[(stage.tevorders_texmap >> 2) & 1];

Are you sure about that? Nowhere else in `PixelShaderGen.cpp` is bpmem used, except `GetPixelShaderUid`. I would have thought the necessary info would have to be in `pixel_shader_uid_data`.

Miksel12:
This is purely based on the software renderer: https://github.com/dolphin-emu/dolphin/blob/master/Source/Core/VideoBackends/Software/TextureSampler.cpp

JMC47:
Leoetlino pointed this one out to me - looks like a regression?  https://fifo.ci/compare/5877241-5873479/

fortsnek9348:
The shader system works on a caching system with invalidation coming from FIFO opcodes. If invalidation doesn't happen, whatever does the drawing won't know that point-sampling requires a different shader.

fortsnek9348:
@JMC47 are those textures supposed to be linear-filtered?

JMC47:
It looks like that particular fifolog is bugged out.  That does not occur when I tested the game.  However, Sexy Poker still has the lines on the faces of characters.  However... I think I tested that game on console and if you pay very close attention it has some artifacting there, though it's different than Dolphin.

mbc07:
The most recent commit still behaves exactly the same as #9407 with Just Dance...

Miksel12:
> The most recent commit still behaves exactly the same as #9407 with Just Dance...

That is correct, @fortsnek9348 showed that that doesn't happen in the sampling code.

fortsnek9348:
Recent commit does not have the `fixpoint_uv*` change that I'm wondering about.

Miksel12:
@fortsnek9348 did you test if the vertex change is enough, or did you only test in combination with the sampling epsilon?

fortsnek9348:
I did adjustments to the vertex offset, but didn't find any that would work. I've tested if 6x resolution can generate the correct pixels, to try and bypass any GPU-dependent vertex rounding with the assumption that Flipper would round to a 6x6 subpixel grid, but that didn't work either, assuming I did it correctly.

Miksel12:
Ok, but with both the epsilon and the vertex adjustment you didn't see any artifacts right?

fortsnek9348:
The fix above is:

> With the `float2(texcoord.xy | int2(1, 1))` fudge for all texture lookups and using `floor` to round `fixpoint_uv*`, I haven't found any artifacts in the examples I have.

For point-sampled textures, rounding to texel centers should also work as the fudge.

No vertex adjustment here.

Miksel12:
I tried those changes but I still get artifacts in Just Dance.

fortsnek9348:
Same FIFO logs?

Miksel12:
JD3-dancer from mbc07.

fortsnek9348:
`JD2-dancer` is the difficult one. Didn't need to change `fixpoint_uv*` for `JD3-dancer`.

Can't really do much about artifacts I can't see. Cleared your shader cache?

Miksel12:
Yes, but we shouldn't forget that this issue doesn't happen on Intel so it hardware seems to have impact. I'm using a GTX 1060.

fortsnek9348:
Pretty similar. GTX1050 Ti.

Miksel12:
Damn, that is weird. You are using windows 7 right?

fortsnek9348:
Win7, D3D11, GL, Vulkan, driver 460.89. Must've got lucky with the rounding errors. If you get artifacts, then maybe you would want to hop into RenderDoc. You can fiddle with the pixel shader without recompiling Dolphin, if you enable Dolphin's shader debugging.

Miksel12:
Where do I enable Dolphin's shader debugging?

fortsnek9348:
Ensure `D3DCOMPILE_DEBUG` is in the flags: `https://github.com/dolphin-emu/dolphin/blob/master/Source/Core/VideoBackends/D3DCommon/Shader.cpp#L99. Or maybe change that `bEnableValidationLayer` setting.

fortsnek9348:
Just realised that you might not be applying any correction to indirect stage texcoords.

Miksel12:
Since it doesn't ring a bell, I think that might be the case...

fortsnek9348:
Never mind, I'm just not seeing any `fixpoint_uv*` in the changes here.

fortsnek9348:
Indirect stage coords is the `tempcoord` variable in the shader.

Miksel12:
What kind of correction did you use on tempcoord?

fortsnek9348:
The same. Just another place where scaling by `texdim` is another source of inaccuracy.

Miksel12:
Ok, by flooring `fixpoint_uv`, and adding an epsilon to both `tempcoord` and `texcoord` all issues are gone in my fifolog. Except for sexy poker but JMC mentioned that those seams might be on console as well.

fortsnek9348:
Sounds good, but I wouldn't know if `floor` is correct for rounding texcoord inputs. It just happens to work here.

The "epsilon" bitwise-or should always be correct for point-filtering. It just adjusts the texcoord to avoid texel boundaries before converting to FP and scaling. So, this adjustment should not be a cause of those seams, but there may be an inaccuracy somewhere else. And if those seams use linear-filtered textures, adjustments for point-filtering need not apply.

Miksel12:
> Sounds good, but I wouldn't know if `floor` is correct for rounding texcoord inputs. It just happens to work here.
> 
> The "epsilon" bitwise-or should always be correct for point-filtering. It just adjusts the texcoord to avoid texel boundaries before converting to FP and scaling. So, this adjustment should not be a cause of those seams, but there may be an inaccuracy somewhere else. And if those seams use linear-filtered textures, adjustments for point-filtering need not apply.

I implemented the same point sample code for `tempcoord` so that all texture that need point sampling are actually point sampled and now all fifologs work correctly **without** epsilon. Except for JD3, so it seems you are right.

JMC47:
I can provide fifologs of sexy poker if anyone can get it working on the hw fifoplayer.  Or if you were lucky enough to buy it before the servers went down, you can also run it on console.

Miksel12:
I have a fifolog of sexy poker.

JMC47:
if I remember correctly, some of the texture seams on it should be visible in software renderer, which should match console.

fortsnek9348:
> I implemented the same point sample code for tempcoord so that all texture that need point sampling are actually point sampled and now all fifologs work correctly without epsilon. Except for JD3, so it seems you are right.

Not sure what you're referring to there. Rounding to texel centers instead of using the bitwise-or "epsilon"? Should be the same result. Or some other epsilon?

And for JD3, I only needed to fix the `texdim` scaling rounding error. Didn't need to `floor` the `fixpoint` values. Seems to be a bit of inconsistency there.

Miksel12:
> Not sure what you're referring to there. Rounding to texel centers instead of using the bitwise-or "epsilon"? Should be the same result. Or some other epsilon?

      FourTexUnits& texUnit = bpmem.tex[(texmap >> 2) & 1];
      TexMode0& tm0 = texUnit.texMode0[texmap & 3];

      SampleTexture(out,
                    SamplerCommon::IsBpTexMode0PointFiltering(tm0) ?
                        "((float2(tempcoord >> 7) + 0.5f) * 128.0f)" :
                        "float2(tempcoord)",
                    "abg", texmap, stereo, api_type);

fortsnek9348:
For testing this fix with point-sampled textures, it would be best to leave out the `IsBpTexMode0PointFiltering` check until that info is cached properly, afaik. The idea of the bitwise-or method is so that you can keep textures smooth whilst doing this, without having a special case.

Miksel12:
I know, this is purely to test and see what the 'correct' output is.

Miksel12:
About the epsilon, isn't a small floating point offset more correct? A bitwise-or will change the sample points differently while an offset would offset all the sample points.

fortsnek9348:
You'd have to find a correct offset. They are 24-bit values, and there might not be enough precision to add something like `0.5` to the fixed-point coordinate. But if it's 23-bit + twos-complement, then it should be possible. Well, it'll probably work for this case.

But for maximal performance, you wouldn't apply the correction to linear-filtered textures anyway.

Miksel12:
I opted to not include the point sampling code, while it does seem to work correctly I can't really take responsibility for it as I don't know how it exactly works. I just found this function in the software renderer and plugged in. 
It also point samples texture which were previously not point sampled, which is presumably more correct but quite a departure from what Dolphin has been doing. 
A good way to implement point sampling could be by linking it to 'Force Texture Filtering' so that people can disable it, though I think someone with more knowledge should look into that.

Instead I went for an epsilon of 0.5. This gave the same results for point sampling without point sampling so there is no behaviour change. 0.5 was chosen as that is the lowest value that can always be represented when converting a 24bit signed integer to a float. 
The only fifolog that this doesn't fix is Just Dance 2 but the same artifacts can be found when point sampling and this depends on the hardware, so this is probably a rounding error(https://github.com/dolphin-emu/dolphin/pull/9473#issuecomment-768564342, https://github.com/dolphin-emu/dolphin/pull/9473#issuecomment-767937443).
But Just Dance 2 is improved compared to master, at least on my hardware (Nvidia).

Miksel12:
`tempcoord` is actually an int, there is a todo which questions if this should be a s24: https://github.com/dolphin-emu/dolphin/blob/906fbf6c8e33b1f0247376b1f5d8f0ead5368c18/Source/Core/VideoCommon/PixelShaderGen.cpp#L788
Not sure what I should do with this. Converting it to a s24 didn't change anything in my testing..

Miksel12:
@mbc07 Could you do another round with the Just Dance games? I think they should all be better than master now (on Nvidia), the fifo's I tested were completely fixed or beter than master at least. There shouldn't be any regressions on Intel but probably worth a short test round.

mbc07:
Apart from a few horizontal lines remaining on Just Dance 2 (still an improvement, compared to master), everything else is fixed, both on Intel and NVIDIA...

-----
(I noticed another odd behavior that happens only with DLC songs on JD2 and JD3: they simply don't show any corruption, no matter what, that's even on master. Even when overriding the Game INI settings and enabling "Force Texture Filtering", which is known to cause severe corruption in those games, the DLC songs still renders with zero corruptions)

JMC47:
The DLC songs are likely compressed differently due to having to fit on the Wii NAND.  Perhaps smaller than on disc songs.

mbc07:
@JMC47 based on the results of the previous commits of this PR, as well as of #9407, perhaps we could add `JD2-dancer` and `JD3-dancer` logs to FifoCI? These two seems sufficient to cover all cases known to have issues in the affected JD games...

JMC47:
I've added a just dance 2 fifolog

Miksel12:
Do we need more test cases? Otherwise I'll fix the lint issue and then this will be ready to go, unless more people want to review the code.

mbc07:
Apart from games from EA using VP6 videos and the earlier Just Dance titles. I'm personally not aware of other games that might be affected by this PR...

PatrickFerry:
**Note:**
A previous revision of this PR had improved at least two other fifologs from the list in a not insignificant way which are not improved now.

These logs are both vegas-party-depth and melee-depth.

melee-depth had been improved by removing the coloured ringing around the leaf.
vegas-party-depth had shown the trees in full including the large and smaller branches

fortsnek9348:
> melee-depth had been improved by removing the coloured ringing around the leaf.
>
> vegas-party-depth had shown the trees in full including the large and smaller branches

Which is correct and are those textures linear-filtered? Some of the commits rounded texture coordinates for linear-filtered textures too.

PatrickFerry:
> > melee-depth had been improved by removing the coloured ringing around the leaf.
> > vegas-party-depth had shown the trees in full including the large and smaller branches
> 
> Which is correct and are those textures linear-filtered? Some of the commits rounded texture coordinates for linear-filtered textures too.

I haven't looked into them. That would make sense though.

Miksel12:
Those two are indeed incorrect this way. In the previous FifoCI run, a build with point sampling was tested. Which means that those textures should be point sampled while they are linearly sampled in this iteration and every other Dolphin build. 
The reason why I chose for an epsilon fix, is that users are used to everything being linearly sampled and me not being sure if the logic I used was correct, though I'm probably a bit overly cautious as it is pretty clear that the code I used worked correctly.

While testing SMG2, I saw many textures that I didn't remember being so jaggy, which is the correct look of those textures but still different.
This way, the old behaviour is maintained while fixing the issue.

PatrickFerry:
Alright, thanks for the explanation.
Hopefully a future PR with that feature that can be enabled with a toggle might be a good improvement for those that want dolphin to match the console's behaviour, or disabled for those who prefer the current look.

Miksel12:
I thought about that but I think it might be a bit niche. Though an option in the advanced section wouldn't hurt.

Miksel12:
Fixed linter.

Miksel12:
Ok, after discussion on IRC I'm gonna go for the correct point sampling.

mbc07:
Fair enough but I think the current linear sampling behavior should still be available as an optional enhancement (like "Force 24-bit color" or "Disable Copy Filter", for example). The image quality was significantly degraded on one of the initial iterations of this PR where everything was being point sampled...

JMC47:
Yep, we discussed that.  Adding in the Linearly sample everything + the epsilon will be **much** easier to accept as an enhancement when we have correct emulation to fallback on.

fortsnek9348:
> Linearly sample everything

Am I missing something? Including point-sampled textures? With just an epsilon?

Miksel12:
No, I'm gonna implement your solution for point sampling and make a fallback/enhancement with linear sampling + epsilon.

iwubcode:
I can't imagine proper point sampling would have a huge impact on image quality but if we need to add an option to fallback to the current solution (no modification of the texture coordinate) we can.

I hadn't heard of "point sampling" until now, I didn't realize there was a term at least.  Remember it's not the same as "point filtering" which an option was added to force this in this PR: #8645 .

Miksel12:
Whoops, something went wrong...

Miksel12:
> Edit: Text. Text is very different with point sampling. I don't have any issue with that look but I think many users won't like the point sampled look.

Ok lol, I thought textures with point sampling look very blocky but I just realised that I am so accustomed to the HD texture that I didn't realise I was just looking at the normal textures :P 

And considering that the FifoCI showed only a few changes I think making this the default and not having a fall back is perfectly ok.

Miksel12:
Point sampling will now be correctly applied. No epsilon in sight :D
Be sure to wipe caches if you tested a previous build. Haven't done ubershaders.

JMC47:
@dolphin-emu-bot rebuild

JMC47:
https://fifo.ci/compare/5887239-5873434/ looks weird to me?   Is this correct?

Miksel12:
Looks correct, clearly point sampled vs bilinearly sampled.

JMC47:
So there may be an issue with the detection?  Extrems posted a hardware test that has a modified pixel center and tests bilinear sampling.  Dolphin Master fails as expected, and old versions of this Pull Request also fail.  The new version of the Pull Request succeeds... however it's supposed to be bilinearly filtering.  This would suggest our detection is wrong.  Note that this uses a modified pixel center to fix a bug on hardware so that this renders correctly.  I don't know if that mucks up with detection.

The Hardware Test

https://files.extremscorner.org/gamecube/apps/bilineartest.zip

Current version of this PR

image

Master

image

iwubcode:
So I was pretty disappointed with the results.  I tested a number of issues I've seen and none of them were fixed.  A few were admittedly higher resolution issues (ex: [10401](https://bugs.dolphin-emu.org/issues/10401)) but there were some other minor ones too.

One downside to the current solution is upscaling to 6x for Spiderman Shattered Dimensions now looks really bad.  You used to get this smooth outline, now it is all pixely.

I don't know if we need to be modifying this offset value based on resolution (not sure if that makes sense) but that's at least one game that looks worse with this change.

Miksel12:
@JMC47 Is the correct rendering, like master but with all sides covered by red? Did you test on console?

@iwubcode I tested that fifolog in a previous attempt which fixed it at native (though not at higher resolutions), can you share some of your other test cases? And yes, some games/scenes will probably be more pixelated.

Miksel12:
I think the main problem now comes from detection. I pretty sure all the info we need is contained here: https://github.com/dolphin-emu/dolphin/blob/b22073ef595a5c186eccb54c83ccbe61cf33019b/Source/Core/VideoCommon/BPStructs.cpp#L650-L652

Which means we need to determine if a texture should be point sampled based on these values:
https://github.com/dolphin-emu/dolphin/blob/b22073ef595a5c186eccb54c83ccbe61cf33019b/Source/Core/VideoCommon/BPMemory.h#L446-L466

At the moment, I am using this function:

https://github.com/dolphin-emu/dolphin/blob/b22073ef595a5c186eccb54c83ccbe61cf33019b/Source/Core/VideoCommon/SamplerCommon.h#L17-L20

I think it should be doable to hardware test this. Or just by viewing what for values Extrems test case have to see what for parameters we can set.

JMC47:
The correct rendering is like the Pull Request.  However, the hardware test is bilinearly sampled according to extrems.

Miksel12:
Euh, so if you would run this on a Wii it would look like this PR or not?

JMC47:
It would render like this PR, but supposedly linearly sampled?

Miksel12:
I changed some logic and it now gets linearly sampled, it doesn't look as bad as master which I can't really explain as the non point sample path has remained te same. The funny thing is that the picture changes over time. It starts looking like it does in this PR and slowly changes to look more like master. Seems like constant rounding errors.

Miksel12:
> I tested a number of issues I've seen and none of them were unfortunately fixed :(. A few were admittedly higher resolution issues but some were at native too (ex: [10401](https://bugs.dolphin-emu.org/issues/10401)).

Something weird is going on, I tested the fifolog and got the same result. I later plugged it into RenderDoc and found that it was working correctly. Vulkan also seemed to work, after rebuilding all the backend were correct except for Vulkan. I have no idea what this could be but something seems to persist.

JMC47:
Any time there's something not changed, I'm guessing it's due to shaders getting set up.

Miksel12:
I did some testing to see if the detection parameters are correct by comparing to console and it seems to be correct, documentation also seems to agree. Sadly, I'm now stuck on inconsistent behaviour (again) with backends acting different from each other.

JMC47:
Did you push the changes that seemingly make the backends inconsistent so I can mess with it?  Maybe we can sort out why.

Miksel12:
The changes that cause inconsistency have been part of this PR for 2 days, so the tests you did were with those changes. 
For example, you showed that Extrems test was wrong on this build. I get the same result as you did but when I run portable I get the right result. I thought incrementing `GX_PIPELINE_UID_VERSION` would fix any shader cache issues so maybe another `VERSION` I should increment?

I used this fifolog for testing: https://bugs.dolphin-emu.org/issues/10401 
I ran the fifolog through RenderDoc and found that the missile icon should be point sampled, the texture has min_filter = 0 and mag_filter = 0. The detection code is` tm0.min_filter < 4 && !tm0.mag_filter; ` so it should catch it. But when I run it in portable mode it doesn't point sample it but it does when run not in portable mode...

Taking portable as baseline as it has no relation to previous builds/settings, it should handle both cases but does in one case but not the other?

PatrickFerry:
The caches will only get invalidated and rebuilt once.  Then they recreated with version 3. You made further changes in this PR but the caches won't get invalidated automatically because they are still version 3. So only the yourself and the people testing this are going to see that issue.

Miksel12:
I actually tested in multiple portable builds so this wasn't the issue for me but JMC did find a weird mismatch between shader and uidcache.

mbc07:
I know that at least the NVIDIA drivers for Windows has their own shader caches. Maybe it's related to that?

(they're located at `%LocalAppData%\NVIDIA\GLCache` and `%LocalAppData%\NVIDIA Corporation\NV_Cache`)

fortsnek9348:
Shader caching problems? It seems you can disable caching by setting `ShaderCache = False` in GFX.ini `[Settings]`.

Still got shader problems? Well, that direct reference of `bpmem` inside shader codegen might have something to do with it. I've got my build running with shader invalidation at `BPMEM_TX_SETMODE0(_4)`, and the point filtering info is put inside some spare bits in `PixelShaderUid`. Also, as the texcoord correction is only applied to point-filtered textures, I use the bitwise-or method, assuming it's faster

JMC47:
The problem in the PR currently is that the saved UIDs are broken.  When we wipe the UIDs, it's fine.  

If someone can fix that, that'd probably help.  We're in a bit over our head.

Miksel12:
> Shader caching problems? It seems you can disable caching by setting `ShaderCache = False` in GFX.ini `[Settings]`.
> 
> Still got shader problems? Well, that direct reference of `bpmem` inside shader codegen might have something to do with it. I've got my build running with shader invalidation at `BPMEM_TX_SETMODE0(_4)`, and the point filtering info is put inside some spare bits in `PixelShaderUid`.

I tried that but quickly realised that I just don't have enough knowledge to do something like that.

Miksel12:
Implementing this correctly is clearly over my head. I think someone with more knowledge and experience should be able to get the gist of it and make a correct implementation by looking at this PR. 
If no one implements a correct implementation in the future, I think I'll PR an epsilon solution which was also used in a previous revision in this PR which had no known regression.

PR #9474, opened 2021-01-26
Analytics refactoring

Dentomologist:
Relatively minor refactoring of various parts of Core/Analytics.cpp. Some functions are extracted, variables are made constant, etc.

Contains some static_asserts making sure the mapping between GameQuirks and names is maintained but otherwise there shouldn't be any functionality changes.


Source/Core/Core/Analytics.cpp
```diff
@@ -147,12 +243,46 @@ constexpr std::array<const char*, 14> GAME_QUIRKS_NAMES{"icache-matters",
                                                         "mismatched-gpu-colors-between-xf-and-bp",
                                                         "uses-uncommon-wd-mode",
                                                         "uses-wd-unimplemented-ioctl"};
+constexpr bool ValidateQuirkMatchesName(GameQuirk quirk, std::string_view name)
```
lioncash, line 246:
```suggestion
static constexpr bool ValidateQuirkMatchesName(GameQuirk quirk, std::string_view name)
```

otherwise this has external linkage

Source/Core/Core/Analytics.cpp
```diff
@@ -205,14 +335,20 @@ void DolphinAnalytics::ReportPerformanceInfo(PerformanceSample&& sample)
   }
 }
 
+u64 DolphinAnalytics::GetSamplingJitterMicroseconds() const
```
lioncash, line 338: Given the refactoring, maybe make use of `std::chrono::microseconds`, which should eliminate the need for conversion code, since the API handles that.

Source/Core/Core/Analytics.cpp
Dentomologist: Oops, good catch!

While I'm at it is there any reason I shouldn't add static to GAME_QUIRK_NAMES?

lioncash: const or constexpr on variables at file scope makes them internally linked by default, so it's not necessary

Dentomologist:
I've pushed an update to this PR.  The biggest changes are extracting the logic for collecting performance samples into a new PerformanceSampleAggregator class, and making the timing code use std::chrono.  

I also updated the original PR description with a more complete list of changes (both original and new).

Other changes:
Remove additional unused headers
Reformat another comment
Add static qualifier to ValidateQuirkMatchesName
Reformat GameQuirk enum class members to CamelCase
Extract PerformanceSample to new header to avoid circular dependencies with PerformanceSampleAggregator

Source/Core/Core/Analytics.cpp
Dentomologist: Done.

Dentomologist:
I was missing a header include for Common/CommonTypes.h in PerformanceSampleAggregator.h.

Source/Core/Core/PerformanceSampleAggregator.h
```diff
@@ -21,21 +23,33 @@ struct CompletedReport
 class PerformanceSampleAggregator
 {
 public:
-  static std::chrono::microseconds GetInitialSamplingStartTimestamp();
-  static std::chrono::microseconds GetRepeatSamplingStartTimestamp();
-  static std::chrono::microseconds GetCurrentMicroseconds();
-  static CompletedReport GetReportFromSamples(const std::vector<PerformanceSample>& samples);
+  // Called on game start / title switch.
```
leoetlino, line 26: Could we reduce the amount of static state? It feels like some of those (what used to be member variables in DolphinAnalytics) shouldn't be static.

Source/Core/Core/DolphinAnalytics.cpp
```diff
@@ -146,8 +146,42 @@ constexpr std::array<const char*, 14> GAME_QUIRKS_NAMES{"icache-matters",
...
+  if (index >= GAME_QUIRKS_NAMES.size())
+  {
+    return false;
+  }
+  return name == GAME_QUIRKS_NAMES[index];
+}
+
 static_assert(GAME_QUIRKS_NAMES.size() == static_cast<u32>(GameQuirk::COUNT),
               "Game quirks names and enum definition are out of sync.");
+static_assert(ValidateQuirkMatchesName(GameQuirk::ICACHE_MATTERS, "icache-matters"));
```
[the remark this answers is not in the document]
Dentomologist: Ok, I'll drop that commit.

Source/Core/Core/PerformanceSampleAggregator.h
```diff
@@ -5,13 +5,26 @@
 #pragma once
 
 #include <chrono>
+#include <vector>
+
+#include "Common/CommonTypes.h"
+
+#include "Core/PerformanceSample.h"
+
+struct CompletedReport
```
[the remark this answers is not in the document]
Dentomologist: I'll move it into PerformanceSampleAggregator.

Dentomologist:
I've made the suggested changes.

JMC47:
@dolphin-emu-bot rebuild

Source/Core/Core/PerformanceSampleAggregator.h
leoetlino: I think I might not have been clear enough in my previous comment. To clarify:

* s_samples, s_next_starting_timestamp, InitializePerformanceSampling, AddSampleIfSamplingInProgress, PopReportIfComplete seem like they can be made non-static
* DolphinAnalytics should probably have PerformanceSampleAggregator as a member variable.

Dentomologist: Done.  The other variables were only static because PerformanceSampleAggregator was too, but putting it into DolphinAnalytics fixes that problem.

leoetlino (approved):
This looks good to me, but I'd like to wait for a second approval just to lower the chances of missing an issue and messing up analytics :)

Dentomologist:
Added name/case changes for members of GameQuirk enum added by #9576.

dreamsyntax:
Is this still wanted or is it redundant at 4+ years later?

Dentomologist:
I've updated the PR which is now considerably smaller. Some of the removed changes have been implemented in other PRs in the meantime (changing `GameQuirk` to PascalCase and removing redundant data accesses), and I changed my mind about a couple of them (splitting `MakeBaseBuilder` into separate functions and moving the sample aggregation to a separate class).

The main remaining change is using `std::chrono` for timekeeping.

Source/Core/Core/DolphinAnalytics.cpp
```diff
@@ -45,7 +46,25 @@
...
+static constexpr auto PERFORMANCE_SAMPLING_BASE_INTERVAL_WAIT = 30min;
+static constexpr std::chrono::microseconds PERFORMANCE_SAMPLING_WAIT_MAX_JITTER = 3min;
+
+std::chrono::microseconds GetWaitJitter()
+{
+  std::random_device device;
+  std::default_random_engine engine(device());
+  std::uniform_int_distribution<> distribution(0, PERFORMANCE_SAMPLING_WAIT_MAX_JITTER.count());
+
+  return std::chrono::microseconds(distribution(engine));
```
jordan-woyak:
```suggestion
  std::uniform_int_distribution<> distribution(0, PERFORMANCE_SAMPLING_WAIT_MAX_JITTER.count());

  std::random_device device;
  return std::chrono::microseconds(distribution(device));
```
I suppose you don't need the `default_random_engine` if you're just seeding it with the device every call.
You can just use the device directly.

PR #9478, opened 2021-01-27
JitArm64: Fix false positive "turned singles into doubles" asserts

JosJuice:
Fixes a regression from PR #9472.


Source/Core/Core/PowerPC/JitArm64/JitArm64_FloatingPoint.cpp
```diff
@@ -219,7 +221,7 @@ void JitArm64::fselx(UGeckoInstruction inst)
   const u32 d = inst.FD;
 
   const bool a_single = fpr.IsSingle(a, true);
-  if (fpr.IsSingle(a, true))
+  if (a_single)
```
JosJuice, line 224: This is not a functional change, but all the other changes are.

leoetlino (approved):

PR #9092, opened 2020-09-17
msbuild: Merge "Core" libs into single DolphinLib

shuffle2:
For a long time, dolphin (on msbuild at least) has been building all of the "Dolphin Core" code with identical compilation settings.
This PR takes the next logical step and actually sends all source files to the compiler in a single batch. This itself is a build speedup.

Another build speedup comes from cutting out some unnecessary `lib` calls, since all the resulting "core" object files get packaged into a single lib file. This should especially help buildbot/machines with few threads, since otherwise the lib step acts like a bubble in the msbuild pipeline (this is a legitimate failing of msbuild!).

Another side effect is that the msbuild project dependency graph becomes almost flat, which also improves build parallelism (really, can be attributed to the same msbuild fail as the lib thing).

I'm currently investigating if it's possible for ninja to perform batched compiler invocations, and while I see people requesting/working on the feature over the years, I don't think it's actually implemented yet(?). Anyway, hopefully we can eventually do it for ninja/msvc. The batching would still be a speedup for ninja builds, but the speedup from lib reduction would probably not be as noticeable.

A prerequisite for bundling all the source files together is that they all have unique basenames. All of the commits except the final one are just accomplishing that. In most cases I think it also results in better style anyway.

I think with this PR and #9081 , the only realistic + meaningful remaining speedup for windows buildbot (without giving it better compute resources) is passing `/m` to msbuild (because currently `link` of multiple projects is serialized), but that change would have to be done in sadm.

PR #9479, opened 2021-01-28
msbuild: re-enable HAS_OPENGL and HAS_VULKAN

shuffle2:
fixes regression from #9092


mbc07:
Is the OpenGL condition correct? If I'm reading this correctly, HAS_OPENGL will only be enabled when the platform is not ARM64, however, nowadays Dolphin can be built for Windows 10 ARM and AFAICT it has OpenGL support...

iwubcode:
I don't test ARM so I can't say but if we update this, this cmake / comment can be updated:  https://github.com/dolphin-emu/dolphin/blob/master/CMakeLists.txt#L554

shuffle2:
windows on arm does not natively support (full) opengl. There is a (just released) translation layer https://devblogs.microsoft.com/directx/announcing-the-opencl-and-opengl-compatibility-pack-for-windows-10-on-arm/ but I don't forsee dolphin wanting to use that, really...

PR #9480, opened 2021-01-28
MathUtil: Add SaturatingCast to cast floats more safely

leoetlino:
A modified version of #8871 with the clamp function added to MathUtil along with tests.

The second commit makes VertexLoaderTest use it in order to avoid undefined behavior:

> A prvalue of a floating-point type can be converted to a prvalue of
> an integer type. The conversion truncates; that is, the fractional
> part is discarded. The behavior is undefined if the truncated value
> cannot be represented in the destination type. ([conv.fpint]/1)

### DIFF
--- a/Source/Core/Common/MathUtil.h
+++ b/Source/Core/Common/MathUtil.h
@@ -6,6 +6,7 @@
 
 #include <algorithm>
 #include <cmath>
+#include <limits>
 #include <type_traits>
 #include <vector>
 
@@ -28,6 +29,48 @@ template <typename T, typename F>
 constexpr auto Lerp(const T& x, const T& y, const F& a) -> decltype(x + (y - x) * a)
 {
   return x + (y - x) * a;
+}
+
+// Casts the specified value to a Dest. The value will be clamped to fit in the destination type.
+// Warning: The result of SaturatingCast(NaN) is undefined.
+template <typename Dest, typename T>
+constexpr Dest SaturatingCast(T value)
+{
+  static_assert(std::is_integral<Dest>());
+
+  constexpr Dest lo = std::numeric_limits<Dest>::lowest();
+  constexpr Dest hi = std::numeric_limits<Dest>::max();
+
+  // T being a signed integer and Dest unsigned is a problematic case because the value will
+  // be converted into an unsigned integer, and u32(...) < 0 is always false.
+  if constexpr (std::is_integral<T>() && std::is_signed<T>() && std::is_unsigned<Dest>())
+  {
+    static_assert(lo == 0);
+    if (value < 0)
+      return lo;
+    // Now that we got rid of negative values, we can safely cast value to an unsigned T
+    // since unsigned T can represent any positive value signed T could represent.
+    // The compiler will then promote the LHS or the RHS if necessary.
+    if (std::make_unsigned_t<T>(value) > hi)
+      return hi;
+  }
+  else if constexpr (std::is_integral<T>() && std::is_unsigned<T>() && std::is_signed<Dest>())
+  {
+    // value and hi will never be negative, and hi is representable as an unsigned Dest.
+    if (value > std::make_unsigned_t<Dest>(hi))
+      return hi;
+  }
+  else
+  {
+    // Do not use std::clamp or a similar function here to avoid overflow.
+    // For example, if Dest = s64 and T = int, we want integer promotion to convert value to a s64
+    // instead of changing lo or hi into an int.
+    if (value < lo)
+      return lo;
+    if (value > hi)
+      return hi;
+  }
+  return static_cast<Dest>(value);
 }
 
 template <typename T>
--- a/Source/UnitTests/Common/MathUtilTest.cpp
+++ b/Source/UnitTests/Common/MathUtilTest.cpp
@@ -26,3 +26,45 @@ TEST(MathUtil, NextPowerOf2)
   EXPECT_EQ(8U, MathUtil::NextPowerOf2(6));
   EXPECT_EQ(0x40000000U, MathUtil::NextPowerOf2(0x23456789));
 }
+
+TEST(MathUtil, SaturatingCast)
+{
+  // Cast from an integer type to a smaller type
+  EXPECT_EQ(255u, (MathUtil::SaturatingCast<u8, int>(1000)));
+  EXPECT_EQ(255u, (MathUtil::SaturatingCast<u8, u16>(1000u)));
+  EXPECT_EQ(255u, (MathUtil::SaturatingCast<u8, std::size_t>(1000)));
+
+  // Cast from a signed integer type
+  EXPECT_EQ(0u, (MathUtil::SaturatingCast<u8, int>(-1)));
+  EXPECT_EQ(0u, (MathUtil::SaturatingCast<u8, int>(-1000)));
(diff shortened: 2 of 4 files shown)


JosJuice:
Could you remove this in favor of the new `SaturatingCast`? https://github.com/dolphin-emu/dolphin/blob/3e66db2fd754fe04eb244860f18ae8388f9e1333/Source/Core/DiscIO/WIACompression.cpp#L169-L172

leoetlino:
@JosJuice Done. The tests also pass now; apparently using std::clamp was not a good idea because it forces `value`, `lo` and `hi` to be the same type T and converting `lo`/`hi` to T can already result in overflow.

Dentomologist:
SaturatingCast<unsigned int>(-1) returns 4294967295 instead of 0, which I don't think is intended.

sepalani:
@Dentomologist 
You're right and this should produce a compiler warning `signed/unsigned mismatch`. Maybe a static_assert can be added to prevent edgy cases like that.

Source/Core/Common/MathUtil.h
```diff
@@ -30,6 +31,32 @@ constexpr auto Lerp(const T& x, const T& y, const F& a) -> decltype(x + (y - x)
...
 
+template <typename Dest, typename T>
+constexpr Dest SaturatingCast(T value)
+{
+  static_assert(std::is_integral<Dest>());
+
+  constexpr Dest lo = std::numeric_limits<Dest>::lowest();
+  constexpr Dest hi = std::numeric_limits<Dest>::max();
+
+  // T being an unsigned integer and Dest unsigned is a problematic case because the value will
```
Dentomologist, line 42: -> T being a signed integer

leoetlino: Whoops, good catch. The `value > hi` comparison is also still incorrect, I'll fix this when I can...

Source/UnitTests/Common/MathUtilTest.cpp
```diff
@@ -26,3 +26,39 @@ TEST(MathUtil, NextPowerOf2)
...
+  EXPECT_EQ(127, MathUtil::SaturatingCast<s8>(5678.0));
+  EXPECT_EQ(-128, MathUtil::SaturatingCast<s8>(-5678.0));
+  EXPECT_EQ(65535, MathUtil::SaturatingCast<u16>(999999.0));
+
+  // Edge cases
+  EXPECT_EQ(std::numeric_limits<s32>::max(),
+            MathUtil::SaturatingCast<s32>(std::numeric_limits<float>::infinity()));
+  EXPECT_EQ(std::numeric_limits<s32>::min(),
+            MathUtil::SaturatingCast<s32>(-std::numeric_limits<float>::infinity()));
+  // 16777217 = 2^16 + 1 is the first integer that cannot be represented correctly with a f32.
```
MerryMage, line 61:
```suggestion
  // 16777217 = 2^17 + 1 is the first integer that cannot be represented correctly with a f32.
```

leoetlino: I must have been tired when I wrote this, this should be 2^24 + 1 :man_facepalming:  Thanks for catching this

Source/Core/Common/MathUtil.h
```diff
@@ -30,6 +31,45 @@ constexpr auto Lerp(const T& x, const T& y, const F& a) -> decltype(x + (y - x)
   return x + (y - x) * a;
 }
 
+template <typename Dest, typename T>
+constexpr Dest SaturatingCast(T value)
+{
+  static_assert(std::is_integral<Dest>());
+
+  constexpr Dest lo = std::numeric_limits<Dest>::lowest();
```
sepalani, line 39: Shouldn't these need to be renamed as `LO` and `HI` according to Dolphin's coding style?

leoetlino: I think it'd be weird here because there's no constant number on the right-hand side (e.g. `constexpr float PI = 3.1415;`) and it's a constant that varies (oxymoron, I know) depending on the Dest type.

If we followed the rule literally, all constexpr variables would need to be written in all caps -- in constexpr functions, most/all variables would need to be in all caps, which is super ugly and verbose when there's no risk of confusion anyway.

If I recall correctly @lioncash said that the rule should probably be updated to reflect modern C++ so I'm inclined to leaving the variable names as is.

Source/Core/Common/MathUtil.h
```diff
@@ -30,6 +31,45 @@ constexpr auto Lerp(const T& x, const T& y, const F& a) -> decltype(x + (y - x)
...
+  constexpr Dest hi = std::numeric_limits<Dest>::max();
+
+  // T being a signed integer and Dest unsigned is a problematic case because the value will
+  // be converted into an unsigned integer, and u32(...) < 0 is always false.
+  if constexpr (std::is_integral<T>() && std::is_signed<T>() && std::is_unsigned<Dest>())
+  {
+    if (value < 0)
+      return lo;
```
Pokechu22, line 47: I feel like there should be a `static_assert(lo == 0)` or something like that here.

leoetlino: Or perhaps replace `return lo;` with `return 0;`? What do you think would be clearer?

Pokechu22: I personally think the `static_assert` would be clearer, since it makes the `u32(...) < 0` assumption more explicit.

leoetlino: Works for me

Source/UnitTests/Common/MathUtilTest.cpp
```diff
@@ -26,3 +26,39 @@ TEST(MathUtil, NextPowerOf2)
...
+  EXPECT_EQ(0u, MathUtil::SaturatingCast<u8>(-1234.0));
+  EXPECT_EQ(127, MathUtil::SaturatingCast<s8>(5678.0));
+  EXPECT_EQ(-128, MathUtil::SaturatingCast<s8>(-5678.0));
+  EXPECT_EQ(65535u, MathUtil::SaturatingCast<u16>(999999.0));
+
+  // Edge cases
+  EXPECT_EQ(std::numeric_limits<s32>::max(),
+            MathUtil::SaturatingCast<s32>(std::numeric_limits<float>::infinity()));
+  EXPECT_EQ(std::numeric_limits<s32>::min(),
+            MathUtil::SaturatingCast<s32>(-std::numeric_limits<float>::infinity()));
```
Pokechu22, line 60: What about NaN?  (Not sure if there's a reasonable way to handle NaN).

It might also be worth adding a test case for -0, though -0 is already covered in VertexLoaderTest.cpp.  Say, `EXPECT_EQ(0u, MathUtil::SaturatingCast<u8>(-0.0));` and `EXPECT_EQ(0, MathUtil::SaturatingCast<s8>(-0.0));`

leoetlino: Yeah, I'm unsure how NaN should be treated. Tempted to just leave it undefined, because it's unlikely to come up in practice and it doesn't make sense to return either lo or hi.

I'll add tests for -0.0, good idea.

Pokechu22: A comment saying "the result of SaturatingCast with a value of NaN is undefined" would be fine by me.

Pokechu22 (approved):

PR #9485, opened 2021-01-30
JitArm64: Use STP for pc/npc

JosJuice:


leoetlino (approved):

manuelh2410:
An issue with NBA LIVe ROM on WII U . 
specifically at the screen where you enter your initials. 
the bug does not allow you to enter your initials ,and prevents you from starting a match

PR #9487, opened 2021-02-03
DolphinQt: Fix regression in input expressions

iwubcode:
A regression caused by #9440 prevents users from putting in an actual gamepad control because all backticks get stripped.  I think an exact replacement would be just wrapping the expression with "^[a-zA-Z]$".  This is what the [exactMatch docs](https://doc.qt.io/archives/qt-4.8/qregexp.html#exactMatch) seems to indicate.

However, it wasn't immediately obvious what was going on, so I opted for a slightly different expression (and negated the if) along with a comment.  Please let me know if you'd prefer the other.


lioncash (approved):

lioncash:
Thanks for catching this 👍

PR #9488, opened 2021-02-03
InputCommon: fix potential dynamic input texture crash and an optimization

iwubcode:
There was the possibility that a dynamic input texture change would cause a reload of textures while the prefetcher was sorting through them.  This would cause a crash.

Also for packs with multiple configurations, we were calling force reload too many times

### DIFF
--- a/Source/Core/VideoCommon/HiresTextures.cpp
+++ b/Source/Core/VideoCommon/HiresTextures.cpp
@@ -56,14 +56,7 @@ void HiresTexture::Init()
 
 void HiresTexture::Shutdown()
 {
-  if (s_prefetcher.joinable())
-  {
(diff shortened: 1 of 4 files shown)


Source/Core/VideoCommon/HiresTextures.cpp
```diff
@@ -147,6 +147,11 @@ void HiresTexture::Update()
 
 void HiresTexture::Clear()
 {
+  if (s_prefetcher.joinable())
```
leoetlino, line 150: Could you modify Shutdown to call Clear?

iwubcode: Ha, didn't even notice that, will do!

iwubcode:
@leoetlino - done!

PR #9489, opened 2021-02-03
Input caching, relative input, new input funcs, window focus refactor, input widget polish, devices population safety fixes

Filoppi:
Just published it in case anyone wanted to have a look or try it before I split it up in more PRs.
Code is final and can be reviewed if anyone dares.
**This is currently being split into multiple PRs, 2 or 3 at a time, then once they are merged, I will make more. Pending PRs:**
**Parts of this that have already been merged (marked in italic):**
https://github.com/dolphin-emu/dolphin/pull/9581
https://github.com/dolphin-emu/dolphin/pull/9688
https://github.com/dolphin-emu/dolphin/pull/9689
https://github.com/dolphin-emu/dolphin/pull/9697
https://github.com/dolphin-emu/dolphin/pull/9700
https://github.com/dolphin-emu/dolphin/pull/9702
https://github.com/dolphin-emu/dolphin/pull/9703
https://github.com/dolphin-emu/dolphin/pull/9771
https://github.com/dolphin-emu/dolphin/pull/9795
https://github.com/dolphin-emu/dolphin/pull/9987

**Binaries available here:
https://github.com/Filoppi/dolphin/releases/tag/1**

**Main changes:**
- Input are now cached once per input frame and then read from cached values (doesn't add a frame of latency). Outputs are now set (cached) once per frame and applied at the end of the frame.
- "Perfect" Relative input support (not hacked in as before) (mapping a mouse to the right analog stick of FPS is greatly improved).
- New input/output focus code.
- Loads of new input/output function expressions (a lot of them for speed run aid and relative input).
- Large polish and improvements to Qt Input Mapping Widgets.
- _ControllerInterface devices population safety and robustness._

Input and Output "caching":
- We can now implement a ton of new input/output functions which are reliably updated ONCE per input frame, and can be time based (game time, not world time).
- Input mappings can't read outputs anymore. Output mappings can't read input anymore. This avoids useless problems with threading and was just wrong/weird. We can achieve the same result with variable expressions.
- All outputs are now automatically reset on device disconnection/reset and game pause/stop.
- This should also improve performance as the new thread locking/dependency model is more independent, and probably more CPU cache friendly.
- Outputs (devices outputs) now have a collection of states, so they can be mapped to multiple mappings/controllers, and they will take the sum of the states.
- Input devices cached inputs values are now protected from being read and written at the same time by multiple threads.

Relative Input:
- Implement correct relative input mouse on most (if not all) mouse backends (platforms/OSes). This greatly increases its accuracy and allows FPS games to be played with it.
- Relative Inputs are reset on game pause/stop.
- WiiMote Cursor moved now scales with emulation speed correctly. Also improved its reset.

Qt Input Mappings improvements (greatly improved user friendliness of all input widgets, no PhD needed anymore):
- Everything has got tooltips now, including every single Function Expression, which have got a detailed description
- Function Expressions are now automatically added with their minimum required commas (depending on the number of params they have)
- You can now specify no default device
- Improved devices list (some of them were not updated correctly)
- Improved clarity of the settings to detect input from all devices in mappings
- All control names in expressions are now wrapped around ``
- Add inputs and outputs to the expression by double clicking on them
- Added button to test the selected output without having to add it to the expression first
- _Expose Control Expression variables to mappings UI and fix memory leak with them_
- Fixed/fully restored controls range slider and modifiers
- NumericSetting(s) value in the UI is now normalized between their min and max
- Increased frequency of UI updates of mapping widgets from 30 to the current monitor refresh rate
- Added ability to grey out a controller setting based on another
- Fix a lot of timing/order dependency bugs between UI widgets and actual controllers values. Same for ControlExpressions and NumericSettings (their simiplification was acting weird sometimes).
- Mapping widget now shows their required focus settings
- Mapping buttons now also show the preview state of outputs and highlight expressions that have failed to parse, and shorten the expression text for a better/quicker preview of what it contains.
- Nicely split input functions into categories (e.g. Math, Logic, Timed, Relative Input, ...)
- Split the state of the WiiMote Cursor between emulation and game (they were polluting each other)
- Polished code and improved robustness of config changed events etc etc...

New major input/output function expressions (mostly to aid mapping a mouse to the analog stick for FPS games, and for Relative Input in general):
- onPress, onRelease, onChange (on toggle)
- cache: can cache a value based on a condition
- relativeToSpeed: turns a relative axis into a rate of change (speed). Can be used to map a mouse axis to an analog stick for example
- interval: returns 1 for "duration_frames" every "delay_frames"
- average: returns the average of input values over a specified number of frames
- sum: returns the sum of input values over a specified number of frames
- record: records the value of the input every frame, when the record condition is true. Plays it back when the playback condition is true.
- sequences: plays back a specified sequences of values (frame by frame)
- lag: lags the input by the specified number of frames
- antiDeadzone: undoes the deadzone applied by games (for mapping sticks to a mouse)
- bezierCurve: undoes the bezier curve applied by some games to input to move the game camera (for mapping sticks to a mouse)
- antiAcceleration: undoes the acceleration applied by some games to input to move the game camera (for mapping sticks to a mouse)
- gameSpeed: returns the current game/emulation speed (1 is full speed)
- timeToInputFrames: converts a time to a number of input frames, depending on the current game and controller
- videoToInputFrames: converts video frames to input frames, depending on the current game and controller
- hasFocus: returns if the render widget has got focus
- ignoresFocus, ignoreOnFocusChange, requireFocus: focus modifiers
- scaleSet: scales the value of a single output. Will probably be made redundant once Billiard is done with the eval function.

ControllerInterface and input backends:
- _Fix Add/Remove/Refresh device safety, devices could be added and removed at the same time, causing missing or duplicated devices (rare but possible)_
- _use PlatformPopulateDevices() on all backends to fix mentioned above and more_
- _Make backends add devices async to not stall the main thread_
- _Fix devices not being released by the UI in time, preventing them from being recreated (fixes issue 12460)_
- _Thread safety fix for a lot of backends_
- _Dinput KeyboardAndMouse reliability improvements (e.g. device lost events)_
- _Workaround for issue 12478 and 11702 (Windows keyboard breaking on render window change): do not recreate devices when the window changes, just refresh their window ptr_

DS4 UDP:
- _Improve robustness when a large amount of devices are connected_
- _Connection disconnection detection improvements_
- _Fix UDP thread often stalling the main thread_
- _Make sure inputs start from their resting position and not always 0_
- _Add battery level_
- Add touch pad relative input
- Add touch pad calibration (the standard has got no min/max value, other emulators also offer calibration for them) (defaults to DS4 calibration)
- Add controller type to give inputs a different default name (defaults to DS4). Improved default inputs names
- Multiple widgets fixes and improvements
- _Fix concurrent/unsafe access to variables_

Input focus and mouse locking:
- _Added support for mouse capture/locking within the game window (following the game aspect ratio)_ (only implemented on Windows, if anyone wants to implement it on Linux and Mac OS X, they are welcome)
- Refactored code that checks for input focus to be more reactive (fixes game window accepting inputs after focus loss)
- Added a new focus setting that ignores input unless the mouse has been captured
- Added ability to ignore inputs from events that causes a window focus change (e.g. click on the dolphin emulation window with a LMB won't trigger the game input bound to the LMB)
- Every input mapping can now divide whether they ignore focus or not. For example, battery level inputs should never ignore focus, this previously was hacked around the code
- _Spawn the "Are you sure you want to stop the emulation" message from the game window if we press the stop emulation from there (to avoid losing focus)_
- Outputs now also support and follow focus requirement settings

Other input improvements:
- _Fixed FPS counter and game window speed breaking on pause/unpause_
- _Added CONTROLLERINTERFACE log (instead of using SERIALINTERFACE)_
- Small clarifying comments around the code
- Small fixes, const correctness

**Testing on Linux and Mac OS X:**
-Start and stop game a lot, see if it hangs and if input still works (HAS BEEN TESTED ON LINUX)
-When the game is running (or not) open the input mapping of a controller and click on "Refresh Devices". Try to spam it a lot. (HAS BEEN TESTED ON LINUX)
-Try the new relative mouse axes

Notes:
This is the result of a year of work.
I've been testing everything deeply and for a long time, so I don't expect any major problems, but you never know.
It would be cool if the changes could be tested on Linux and MacOS. Android is not affected by any of these changes.
Undoes the relative input and input channel changes from PR 8747, not because they were bad, but because my implementation was already done as it more feature rich and deeply rooted all around.
Splitting this up in many PRs is gonna be hard and I believe we are gonna need to work together and possibly make some compromises (like pushing some large PRs, because a lot of these features are entangled).

Screenshots:
Mapping1
Mapping2
MouseLocking

### DIFF
--- a/Source/Core/InputCommon/ControllerInterface/DualShockUDPClient/DualShockUDPClient.cpp
+++ b/Source/Core/InputCommon/ControllerInterface/DualShockUDPClient/DualShockUDPClient.cpp
@@ -42,7 +42,16 @@ const Config::Info<bool> SERVERS_ENABLED{{Config::System::DualShockUDPClient, "S
 // Clock type used for querying timeframes
 using SteadyClock = std::chrono::steady_clock;
 
-class Device final : public Core::Device
+std::atomic<int> g_calibration_device_index = -1;
+bool g_calibration_device_found = false;
+s16 g_calibration_touch_x_min = std::numeric_limits<s16>::max();
+s16 g_calibration_touch_y_min = std::numeric_limits<s16>::max();
+s16 g_calibration_touch_x_max = std::numeric_limits<s16>::min();
+s16 g_calibration_touch_y_max = std::numeric_limits<s16>::min();
+
+constexpr ControlState TOUCH_SPEED = 0.0125;  // Just a value that seemed good
+
+class UDPDevice final : public Core::Device
 {
 private:
   template <class T>
@@ -62,12 +71,13 @@ private:
     const T m_mask;
   };
 
-  template <class T>
+  template <class T = ControlState>
   class AnalogInput : public Input
   {
   public:
-    AnalogInput(const char* name, const T& input, ControlState range, ControlState offset = 0)
-        : m_name(name), m_input(input), m_range(range), m_offset(offset)
+    AnalogInput(const char* name, const T& input, ControlState range, ControlState offset = 0,
+                std::string_view old_name = "")
+        : m_name(name), m_input(input), m_range(range), m_offset(offset), m_old_name(old_name)
     {
     }
     std::string GetName() const final override { return m_name; }
@@ -75,15 +85,33 @@ private:
     {
       return (ControlState(m_input) + m_offset) / m_range;
     }
+    bool IsMatchingName(std::string_view name) const override
+    {
+      if (Input::IsMatchingName(name))
+        return true;
+
+      return m_old_name == name;
+    }
 
   private:
     const char* m_name;
     const T& m_input;
     const ControlState m_range;
     const ControlState m_offset;
+    const std::string_view m_old_name;
   };
 
-  class TouchInput final : public AnalogInput<int>
+  class RelativeTouchInput final : public RelativeInput<s16>
+  {
+  public:
+    RelativeTouchInput(const char* name, ControlState scale) : RelativeInput(scale), m_name(name) {}
+    std::string GetName() const final override { return m_name; }
+
+  private:
+    const char* m_name;
+  };
+
+  class TouchInput final : public AnalogInput<>
   {
   public:
     using AnalogInput::AnalogInput;
@@ -127,10 +155,67 @@ private:
     const BatteryState& m_battery;
   };
 
+  // Names are based on DS4 for simplicity
+  struct DeviceInputNames
+  {
+    DeviceInputNames(const char* up, const char* down, const char* left, const char* right,
+                     const char* square, const char* cross, const char* circle,
+                     const char* triangle, const char* l1, const char* r1, const char* l2,
+                     const char* r2, const char* l3, const char* r3, const char* share,
+                     const char* options, const char* ps)
+        : m_up(up), m_down(down), m_left(left), m_right(right), m_square(square), m_cross(cross),
+          m_circle(circle), m_triangle(triangle), m_l1(l1), m_r1(r1), m_l2(l2), m_r2(r2), m_l3(l3),
+          m_r3(r3), m_share(share), m_options(options), m_ps(ps)
+    {
+    }
+    const char* m_up;
+    const char* m_down;
+    const char* m_left;
+    const char* m_right;
+    const char* m_square;
+    const char* m_cross;
+    const char* m_circle;
+    const char* m_triangle;
+    const char* m_l1;
+    const char* m_r1;
+    const char* m_l2;
+    const char* m_r2;
+    const char* m_l3;
+    const char* m_r3;
+    const char* m_share;
+    const char* m_options;
+    const char* m_ps;
+  };
+
+  static constexpr std::string_view DS4_INPUT_NAME = "DS4";
+  static constexpr std::string_view DUALSENSE_INPUT_NAME = "DualSense";
+  static constexpr std::string_view SWITCH_INPUT_NAME = "Switch";
+  static constexpr std::string_view GENERIC_INPUT_NAME = "Generic";
+
+  // Sure it would be nice if they were filled in from a config but this is fine
+  const std::map<std::string_view, const DeviceInputNames> g_input_names_by_device = {
+      {DS4_INPUT_NAME,
+       DeviceInputNames("Up", "Down", "Left", "Right", "Square", "Cross", "Circle", "Triangle",
+                        "L1", "R1", "L2", "R2", "L3", "R3", "Share", "Options", "PS")},
+      {DUALSENSE_INPUT_NAME,
+       DeviceInputNames("Up", "Down", "Left", "Right", "Square", "Cross", "Circle", "Triangle",
+                        "L1", "R1", "L2", "R2", "L3", "R3", "Create", "Options", "PS")},
+      {SWITCH_INPUT_NAME,
+       DeviceInputNames("Up", "Down", "Left", "Right", "Y", "B", "A", "X", "L", "R", "ZL", "ZR",
+                        "Left Stick", "Right Stick", "-", "+", "Home")},
+      // Mandatory. Assume orientation of D-Pad and "action" buttons but not of central buttons
+      {GENERIC_INPUT_NAME,
+       DeviceInputNames("D-Pad N", "D-Pad S", "D-Pad W", "D-Pad E", "Button W", "Button S",
+                        "Button E", "Button N", "Left Bumper", "Right Bumper", "Left Trigger",
+                        "Right Trigger", "Left Stick", "Right Stick", "Special Button 1",
+                        "Special Button 2", "Special Button 3")}};
+
 public:
   void UpdateInput() override;
 
-  Device(std::string name, int index, std::string server_address, u16 server_port);
+  UDPDevice(std::string name, int index, std::string server_address, u16 server_port,
+            std::string device_type = "", std::string calibration = "",
+            bool for_calibration = false);
 
   std::string GetName() const final override;
   std::string GetSource() const final override;
@@ -145,13 +230,15 @@ private:
   const int m_index;
   sf::UdpSocket m_socket;
   SteadyClock::time_point m_next_reregister = SteadyClock::time_point::min();
-  Proto::MessageType::PadDataResponse m_pad_data{};
-  Proto::Touch m_prev_touch{};
-  bool m_prev_touch_valid = false;
-  int m_touch_x = 0;
-  int m_touch_y = 0;
+  SteadyClock::time_point m_last_update;
+  Proto::MessageType::PadDataResponse m_pad_data;
+  ControlState m_touch_x;
+  ControlState m_touch_y;
+  Proto::Touch m_prev_touches[u8(InputChannel::Max)];
+  RelativeTouchInput* m_relative_touch_inputs[4];
   std::string m_server_address;
   u16 m_server_port;
+  bool m_for_calibration;
 
   s16 m_touch_x_min;
   s16 m_touch_y_min;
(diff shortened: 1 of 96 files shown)


Source/Core/InputCommon/ControllerInterface/Device.h
```diff
@@ -73,37 +117,129 @@ class Device
...
+  // Keeps the last 2 absolute states and returns their difference, divided by the time delta.
+  // It has one state per input channel, as otherwise one SetState() would break
+  // GetState() on the other channels. 
+  // For example the SerialInterface updates inputs at irregular intervals, so to get consistent
+  // results, we need to consider the elapsed time (which also helps keeping in keeping
+  // the values similar between different channels).
+  // You don't have to use this implementation.
+  //
+  template <typename T = ControlState>
+  class RelativeInput : public Input
```
iwubcode, line 159: I'm not an inputs person so take this whole comment with that in mind but to me this feels overly complex (having seen @jordan-woyak 's [solution](https://github.com/dolphin-emu/dolphin/pull/8747/commits/dbc5d16e3d9b26a265757356bbb23dac8deec42f) ).

In particular, the timing feels like it should be based off the device poll rate, not some general time delta.  After all, we want the relative value between actions.  Plus, we are using seconds precision when it seems we'd want more granularity.

There's a lot of code here so I might just be missing the advantages this provides but wanted to mention for any reviewers that there is another implementation of this feature (my assumption is @jordan-woyak will review so he knows but just in case!).

Filoppi: The device poll rate is equal to the input channel delta time as devices. Devices are always updated before reading/caching inputs.

Highly doubt having a double as seconds instead of milliseconds makes any difference (especially given most game engines use float seconds as time delta), but I don't mind changing that.

iwubcode: I don't think we should be assuming any sort of usage here.  The devices should be providing the most accurate value they can.  So in this case we should use nanoseconds or whatever precision we have available.

I'm still not sure this is relative input, this feels more like relative rate of change which gives a different value.

Filoppi: The delta time I multiply for is fixed for every input channel other that the serial interface, which means it will not have influence on the value (it will not turn it into a speed) other than scaling it.

The serial interface updates at irregular intervals, so if you moved the mouse at a constant speed, the relative input would return a different value for every update, and that is something we don't want (because whatever you actually map it to on GC, it wouldn't need the actual movement, it would need the movement speed). You can see all the thinking behind it in the comment above the time multiplication, the design there is still pending as there isn't a perfect solution unless we add a few settings.

iwubcode: Thanks for the explanation @Filoppi !

> Not exactly sure what you mean with nanoseconds but I don't think it applies.

Sorry, yeah, it does not apply.  I had assumed by the name `GetCurrentInputDeltaSeconds()` it was some sort of time diff when in fact it's more of a poll rate for that channel (not actual time delta).

> (because whatever you actually map it to on GC controller, it wouldn't need the actual movement offset, it would need the movement speed)

I still think your value isn't a true relative value so it might be surprising to users.  How do you know users aren't wanting the actual movement offset of the GC controller for their expression?  Feels like we should add some other input function if you want a rate of change (that the user would multiply against).

As a different example, as you might recall I actually want the relative offset of the mouse for FreeLook.  That matches today's master implementation which is what users expect.  Not that I'm that worried about my implementation, I can divide out the scalar or do other workarounds if really necessary.

Anyway, I don't mean to ramble on this. I'm just one person with a view on one tiny part of the code.  I'll let others chime in if they have thoughts.  Great work overall!

Filoppi: As I said, yes this would as of now return a rate of change, but given that the time delta is fixed, the rate of change would be the same as the pure change, except scaled, so it does not make a difference anywhere except on the SI.

But yes, theoretically, we should have two mouses axis, one that is the real relative change (for freelook and the wiimote cursor), and one that is the rate of change (to be mapped to analog sticks), but I'm afraid that would be even more confusing. Or sure, we could let users multiply their own relative axis by the input channel time delta, but how would they know they need to do that?

Filoppi: I decided to go for the input function to convert a relative axis into a rate of change only when needed. It's a better solution for a lot of reasons and we can explain users they have to do it in one of the help UI tooltips/buttons I added (or on the wiki).
The relative axis code is now cleaned of that massive WIP comment that I needed to figure out what to do with it. Still polishing but using relative axes as input for games has never worked as good as this.

See the "relativeToSpeed(input)" function expression if you want to understand.

Pokechu22:
@dolphin-emu-bot rebuild

Pokechu22:
@dolphin-emu-bot rebuild

ds22x:
Hey Feloppi.
Pardon me for asking, but can you remember what commit(s) specifically fixed https://bugs.dolphin-emu.org/issues/12478 ?

Filoppi:
> Hey Feloppi.
> Pardon me for asking, but can you remember what commit(s) specifically fixed https://bugs.dolphin-emu.org/issues/12478 ?

I don't think I pushed the final fix I made to this PR yet, because I wasn't done with it, that's also why I had uploaded a custom build for you. That said, it's not just a commit, but it's a series of changes that stacked on each other. I'm taking some time away from dolphin these days but once I'm back on of the first things I plan to do is making a PR just for that issue.

ds22x:
More out of curiosity, but I was looking into itself myself to see what exactly it was that fixed it, and if the result was replicable when applied onto upstream source.
Thanks for answering so quickly btw.

Filoppi:
> More out of curiosity, but I was looking into itself myself to see what exactly it was that fixed it, and if the result was replicable when applied onto upstream source.
> Thanks for answering so quickly btw.

Note that this isn't a fix but it's a workaround. I've been investigating the bug for months and off, I down to the function that is broken, but I can't debug it as it's in DInput8.dll. The workaround should be enough though. Unless you are an expert in DInput, I'd advise against waisting time on it 😅.
You should be able to copy that method in a new branch and make it work, but I've also got in a ton of changes that helped make everything safer around the devices code.

ds22x:
Looking into it a bit deeper, it seems to be quite layered indeed.
In any case, I'll eagerly await the day you return and submit the PR with the keyboard workaround.

JMC47:
@dolphin-emu-bot rebuild

Source/Core/DolphinQt/Config/ToolTipControls/ToolTipWidget.h
```diff
@@ -49,6 +51,7 @@ class ToolTipWidget : public Derived
 
   virtual QPoint GetToolTipPosition() const = 0;
 
+protected:
```
iwubcode, line 54: These variables shouldn't be protected.  It was rather intentional that they were private because the private functions (or public ones) are the only thing that should manage these variables.

Filoppi: Yeah, that's a mistake, I had assumed they were public.

Source/Core/InputCommon/ControllerInterface/DualShockUDPClient/DualShockUDPClient.cpp
```diff
@@ -356,61 +358,42 @@ static void HotplugThreadFunc()
...
     // Receive controller port info within a time from our request.
     // Run this even if we sent no new requests, to disconnect devices,
     // sleep (wait) the thread and catch old responses.
     for (auto& server : s_servers)
     {
-      auto current_timeout_ms = adjusted_timeout_ms;
-      if (auto port_info = msg.CheckAndCastTo<Proto::MessageType::PortInfo>())
+      // ReceiveWithTimeout treats a timeout of zero as infinite timeout, which we don't want.
+      // Also define a max wait time for every server, again, to avoid long waits.
+      const auto adujusted_timout_ms =
```
iwubcode: Still wrapping my  head around why this will be helpful over a 1ms wait that is shared by all sockets but this logic isn't dependent on the server, it can be moved out of the loop

Filoppi: You are reviewing an old version, this isn't there anymore.

Filoppi:
I've squashed the whole commit history as it was becoming impossible to rebase on master every time a small part of this branch got merged through other PRs.

JosJuice:
This is both super big and hasn't been active in a long time, so I'm closing it. We can reopen it later if you see a reason to do so.

PR #9490, opened 2021-02-04
Config: Fix expression window scroll wheel spam

Dentomologist:
Fixes the expression window being spammed with the first entry in the
Operators or Functions select menus when scrolling the mouse wheel while
hovering over them.

Fixes https://bugs.dolphin-emu.org/issues/12405


Dentomologist:
I just added a small change to pass a parent pointer to m_functions_combo to match prior behavior.

Speaking of which, this is out of scope for this PR but IOWindow seems very inconsistent about which children get passed parent pointers and which don't.  Is there a reason I'm missing for the current pattern or is that something that should be fixed at some point?

leoetlino (approved):
Apparently this is really the only way to disable this behavior... setWheelPolicy isn't as useful as you'd expect...

Yes, I don't think there's any reason for passing `this` inconsistently.

PR #9491, opened 2021-02-04
Fix mapping window title bar becoming unresponsive

DacodaDragon:
Fixes https://bugs.dolphin-emu.org/issues/12064

The timer called the lambda expression by interval, causing weird behavior for release functions.
Setting the timer to be one-shot fixes this issue.


DacodaDragon:
The amount of time before the input is released seems annoyingly long though, we might consider reducing that time from 500ms to 100ms or so, depending on feedback.

Miksel12:
Thanks! Very happy to see this annoying bug fixed, I don't know why not more people reported this issue.

PR #8221, opened 2019-06-29
Core: fix X11 includes and preprocessor nits

Shizmob:
This adds a missing Xlib include and fixes compilation to work when RandR support is not available.

With these changes and the one in #8220, Dolphin compiles successfully on Alpine! 🎉

PR #9492, opened 2021-02-04
Cleanup X11 and XRANDR Macros

nolange:
This fixes build with X11 enabled and XRANDR disabled.


Rumi-Larry:
https://github.com/dolphin-emu/dolphin/pull/8221 I'm guessing this makes that pr obsolete?

nolange:
> #8221 I'm guessing this makes that pr obsolete?

Yeah, I guess too.

Source/Core/DolphinQt/MainWindow.cpp
```diff
@@ -1463,7 +1463,7 @@ void MainWindow::UpdateScreenSaverInhibition()
 
   m_is_screensaver_inhibited = inhibit;
 
-#if defined(HAVE_XRANDR) && HAVE_XRANDR
+#if defined(HAVE_X11) && HAVE_X11
```
leoetlino, line 1466: Could you simplify this to `#ifdef HAVE_X11`? We never define it to 0.

We should probably do the same thing for HAVE_XRANDR after removing this line from the CMakeLists.txt: https://github.com/dolphin-emu/dolphin/blob/ddacbf83f65b9642f3c9cee9e075b7ec3d473321/CMakeLists.txt#L444

Source/Core/UICommon/UICommon.cpp
```diff
@@ -39,7 +39,7 @@
 #include "UICommon/UICommon.h"
 #include "UICommon/USBUtils.h"
 
-#if defined(HAVE_XRANDR) && HAVE_XRANDR
+#if defined(HAVE_X11) && HAVE_X11
```
leoetlino, line 42: `#ifdef HAVE_X11`

Source/Core/UICommon/UICommon.cpp
```diff
@@ -394,7 +394,7 @@ bool TriggerSTMPowerEvent()
   return true;
 }
 
-#if defined(HAVE_XRANDR) && HAVE_XRANDR
+#if defined(HAVE_X11) && HAVE_X11
```
leoetlino, line 397: `#ifdef HAVE_X11`

Source/Core/UICommon/UICommon.h
```diff
@@ -13,7 +13,7 @@ namespace UICommon
 void Init();
 void Shutdown();
 
-#if defined(HAVE_XRANDR) && HAVE_XRANDR
+#if defined(HAVE_X11) && HAVE_X11
```
leoetlino, line 16: `#ifdef HAVE_X11`

Source/Core/DolphinQt/MainWindow.cpp
nolange: Ok, did add another commit cleaning up the HAVE_XRANDR guards

leoetlino (approved):

PR #9493, opened 2021-02-04
Stricter test for compiler flags

nolange:
Some flags do nothing but generate dozens of warnings later,
add '-Werror' during testing to catch those early.

Main culprit is '-fvisibility-inlines-hidden' which is invalid for C files

### DIFF
--- a/CMake/CheckAndAddFlag.cmake
+++ b/CMake/CheckAndAddFlag.cmake
@@ -29,20 +29,27 @@ function(check_and_add_flag var flag)
 
   set(is_c "$<COMPILE_LANGUAGE:C>")
   set(is_cxx "$<COMPILE_LANGUAGE:CXX>")
+  set(test_flags_c)
+  set(test_flags_cxx)
 
   # The Visual Studio generators don't support COMPILE_LANGUAGE
   # So we fail all the C flags and only actually test CXX ones
   if(CMAKE_GENERATOR MATCHES "Visual Studio")
     set(is_c "0")
     set(is_cxx "1")
+  else()
+    # Otherwise assume the compile follows GCC syntax
+    # and fail when the option is known but invalid.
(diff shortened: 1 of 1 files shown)


leoetlino:
Do we really need to test if -Werror can be used? All GCC and Clang versions we care about should have it...

orbea:
> Do we really need to test if -Werror can be used? All GCC and Clang versions we care about should have it...

Probably not, even if its not supported the compiler will probably print a warning instead of an error so that the test passes...

nolange:
> Do we really need to test if -Werror can be used? All GCC and Clang versions we care about should have it...

Are we sure that a non-MSVC compiler has GCC compatible syntax?
I dropped the check for now, probably best to just see if this raises issues.

leoetlino:
@nolange This seems to build fine, but clang on the macOS builder still complains about unused arguments. Do you know if this warning can be fixed as well?

clang: warning: argument unused during compilation: '-fno-objc-exceptions' [-Wunused-command-line-argument]

nolange:
> @nolange This seems to build fine, but clang on the macOS builder still complains about unused arguments. Do you know if this warning can be fixed as well?
> 
> clang: warning: argument unused during compilation: '-fno-objc-exceptions' [-Wunused-command-line-argument]

Dunno, where's this option coming from ? It's either the wrong language or maybe it's added to the linker step?
`-Qunused-arguments` can silence them, but I am not certain that this is a good solution.

JosJuice:
It's coming from cb14b65.

nolange:
> It's coming from [cb14b65](https://github.com/dolphin-emu/dolphin/commit/cb14b65aad860a79f949fa1e6986cffeba5e3509).

I mean I dont know anything of why and where objc++ flags are necessary and how this should work.
This seems to be an entirely different "fix" to this PR anyway.

leoetlino (approved):

PR #9494, opened 2021-02-06
Arm64Gen: Convert ARM64Reg to enum class

Dentomologist:
Most changes are just adding ARM64Reg:: in front of the constants.  

Add operator functions to reduce type casting spam

Make DecodeReg return int instead of ARM64Reg

Add EncodeRegTo32 with previous behavior of DecodeReg

Despite appearances, this does not change the value of 
ARM64Reg::INVALID_REG.  The previous value 0xFFFFFFFF was being cast to
an int and wrapping around to -1, this just makes that value explicit.

### DIFF
--- a/Source/Core/Common/Arm64Emitter.cpp
+++ b/Source/Core/Common/Arm64Emitter.cpp
@@ -507,8 +507,8 @@ void ARM64XEmitter::EncodeCompareBranchInst(u32 op, ARM64Reg Rt, const void* ptr
   ASSERT_MSG(DYNA_REC, distance >= -0x40000 && distance <= 0x3FFFF,
              "%s: Received too large distance: %" PRIx64, __func__, distance);
 
-  Rt = DecodeReg(Rt);
-  Write32((b64Bit << 31) | (0x34 << 24) | (op << 24) | (((u32)distance << 5) & 0xFFFFE0) | Rt);
+  Write32((b64Bit << 31) | (0x34 << 24) | (op << 24) | (((u32)distance << 5) & 0xFFFFE0) |
+          DecodeReg(Rt));
 }
 
 void ARM64XEmitter::EncodeTestBranchInst(u32 op, ARM64Reg Rt, u8 bits, const void* ptr)
@@ -524,9 +524,8 @@ void ARM64XEmitter::EncodeTestBranchInst(u32 op, ARM64Reg Rt, u8 bits, const voi
   ASSERT_MSG(DYNA_REC, distance >= -0x3FFF && distance < 0x3FFF,
              "%s: Received too large distance: %" PRIx64, __func__, distance);
 
-  Rt = DecodeReg(Rt);
   Write32((b64Bit << 31) | (0x36 << 24) | (op << 24) | (bits << 19) |
-          (((u32)distance << 5) & 0x7FFE0) | Rt);
+          (((u32)distance << 5) & 0x7FFE0) | DecodeReg(Rt));
 }
 
 void ARM64XEmitter::EncodeUnconditionalBranchInst(u32 op, const void* ptr)
@@ -546,8 +545,7 @@ void ARM64XEmitter::EncodeUnconditionalBranchInst(u32 op, const void* ptr)
 
 void ARM64XEmitter::EncodeUnconditionalBranchInst(u32 opc, u32 op2, u32 op3, u32 op4, ARM64Reg Rn)
 {
-  Rn = DecodeReg(Rn);
-  Write32((0x6B << 25) | (opc << 21) | (op2 << 16) | (op3 << 10) | (Rn << 5) | op4);
+  Write32((0x6B << 25) | (opc << 21) | (op2 << 16) | (op3 << 10) | (DecodeReg(Rn) << 5) | op4);
 }
 
 void ARM64XEmitter::EncodeExceptionInst(u32 instenc, u32 imm)
@@ -561,7 +559,8 @@ void ARM64XEmitter::EncodeExceptionInst(u32 instenc, u32 imm)
 
 void ARM64XEmitter::EncodeSystemInst(u32 op0, u32 op1, u32 CRn, u32 CRm, u32 op2, ARM64Reg Rt)
 {
-  Write32((0x354 << 22) | (op0 << 19) | (op1 << 16) | (CRn << 12) | (CRm << 8) | (op2 << 5) | Rt);
+  Write32((0x354 << 22) | (op0 << 19) | (op1 << 16) | (CRn << 12) | (CRm << 8) | (op2 << 5) |
+          DecodeReg(Rt));
 }
 
 void ARM64XEmitter::EncodeArithmeticInst(u32 instenc, bool flags, ARM64Reg Rd, ARM64Reg Rn,
@@ -569,11 +568,9 @@ void ARM64XEmitter::EncodeArithmeticInst(u32 instenc, bool flags, ARM64Reg Rd, A
 {
   bool b64Bit = Is64Bit(Rd);
 
-  Rd = DecodeReg(Rd);
-  Rn = DecodeReg(Rn);
-  Rm = DecodeReg(Rm);
   Write32((b64Bit << 31) | (flags << 29) | (ArithEnc[instenc] << 21) |
-          (Option.IsExtended() ? (1 << 21) : 0) | (Rm << 16) | Option.GetData() | (Rn << 5) | Rd);
+          (Option.IsExtended() ? (1 << 21) : 0) | (DecodeReg(Rm) << 16) | Option.GetData() |
+          (DecodeReg(Rn) << 5) | DecodeReg(Rd));
 }
 
 void ARM64XEmitter::EncodeArithmeticCarryInst(u32 op, bool flags, ARM64Reg Rd, ARM64Reg Rn,
@@ -581,10 +578,8 @@ void ARM64XEmitter::EncodeArithmeticCarryInst(u32 op, bool flags, ARM64Reg Rd, A
 {
   bool b64Bit = Is64Bit(Rd);
 
-  Rd = DecodeReg(Rd);
-  Rm = DecodeReg(Rm);
-  Rn = DecodeReg(Rn);
-  Write32((b64Bit << 31) | (op << 30) | (flags << 29) | (0xD0 << 21) | (Rm << 16) | (Rn << 5) | Rd);
+  Write32((b64Bit << 31) | (op << 30) | (flags << 29) | (0xD0 << 21) | (DecodeReg(Rm) << 16) |
+          (DecodeReg(Rn) << 5) | DecodeReg(Rd));
 }
 
 void ARM64XEmitter::EncodeCondCompareImmInst(u32 op, ARM64Reg Rn, u32 imm, u32 nzcv, CCFlags cond)
@@ -594,9 +589,8 @@ void ARM64XEmitter::EncodeCondCompareImmInst(u32 op, ARM64Reg Rn, u32 imm, u32 n
   ASSERT_MSG(DYNA_REC, !(imm & ~0x1F), "%s: too large immediate: %d", __func__, imm);
   ASSERT_MSG(DYNA_REC, !(nzcv & ~0xF), "%s: Flags out of range: %d", __func__, nzcv);
 
-  Rn = DecodeReg(Rn);
   Write32((b64Bit << 31) | (op << 30) | (1 << 29) | (0xD2 << 21) | (imm << 16) | (cond << 12) |
-          (1 << 11) | (Rn << 5) | nzcv);
+          (1 << 11) | (DecodeReg(Rn) << 5) | nzcv);
 }
 
 void ARM64XEmitter::EncodeCondCompareRegInst(u32 op, ARM64Reg Rn, ARM64Reg Rm, u32 nzcv,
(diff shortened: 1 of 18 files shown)


Source/Core/Core/PowerPC/JitArm64/JitArm64_LoadStoreFloating.cpp
```diff
@@ -164,9 +164,9 @@ void JitArm64::lfXX(UGeckoInstruction inst)
 
   BitSet32 regs_in_use = gpr.GetCallerSavedUsed();
   BitSet32 fprs_in_use = fpr.GetCallerSavedUsed();
-  regs_in_use[W0] = 0;
+  regs_in_use[static_cast<size_t>(ARM64Reg::W0)] = 0;
   fprs_in_use[0] = 0;  // Q0
-  fprs_in_use[VD - Q0] = 0;
+  fprs_in_use[static_cast<size_t>(VD - ARM64Reg::Q0)] = 0;
```
JosJuice, line 169: Probably preferable:

```suggestion
  fprs_in_use[DecodeReg(VD)] = 0;
```

(Same for all similar lines, like `regs_in_use[DecodeReg(ARM64Reg::W0)] = 0;`)

Source/Core/Core/PowerPC/JitArm64/JitArm64_Integer.cpp
```diff
@@ -894,8 +894,8 @@ void JitArm64::subfex(UGeckoInstruction inst)
     ARM64Reg WA = gpr.GetReg();
     if (js.carryFlagSet)
     {
-      MOVI2R(WA, ~i + j, gpr.R(d));
-      ADC(gpr.R(d), WA, WZR);
+      MOVI2R(WA, ~i + j, static_cast<bool>(gpr.R(d)));
```
JosJuice, line 897: Bug. The third parameter should just be removed

Source/Core/Core/PowerPC/JitArm64/JitArm64_Integer.cpp
```diff
@@ -1403,15 +1403,15 @@ void JitArm64::srawx(UGeckoInstruction inst)
 
     MOVI2R(WA, 32);
     SUB(WC, WA, WC);
-    LSL(WC, RS, WC);
+    LSL(WC, RS, static_cast<int>(WC));
```
JosJuice, line 1406: Bug. Should use LSLV (as you will probably find out if you rebase on latest master)

Source/Core/Core/PowerPC/JitArm64/JitArm64_Integer.cpp
Dentomologist: Done.

Source/Core/Core/PowerPC/JitArm64/JitArm64_Integer.cpp
Dentomologist: Done.

Source/Core/Core/PowerPC/JitArm64/JitArm64_LoadStoreFloating.cpp
Dentomologist: I can't follow this suggestion as given, because DecodeReg returns an ARM64Reg which (since it's now an enum class) doesn't implicitly convert to a size_t.  In order to use it as an index BitSet::operator[] I'd either have to cast it to a size_t anyway or write a function to perform that cast, or something else along those lines.

Given that DecodeReg masks its argument to [0,31] and most of the values in question are constants already in that range, the only purpose I can see to using DecodeReg in those cases would be for expressing intent.  Would you say it's worth adding in addition to a cast, or should I just stick with the cast?

As for replacing VD - Q0 with DecodeReg(VD), is VD guaranteed to be in the range [192, 223]?  Its value is set by fpr.RW(), whose behavior is rather opaque to me.

JosJuice: Ah, somehow I thought it was returning an `int` :( Feel free to ignore my suggestion then (either by only using `DecodeReg` on lines that were doing this kind of subtraction in master, or by not using `DecodeReg` at all).

> As for replacing VD - Q0 with DecodeReg(VD), is VD guaranteed to be in the range [192, 223]? Its value is set by fpr.RW(), whose behavior is rather opaque to me.

Yes, in practice it always is. But in case it were not in that range (as you say, it isn't obvious that it is) then using `DecodeReg` would be necessary to get the correct behavior.

Source/Core/Common/Arm64Emitter.h
```diff
@@ -224,9 +224,65 @@ enum ARM64Reg
   WZR = WSP,
   ZR = SP,
 
-  INVALID_REG = 0xFFFFFFFF
+  INVALID_REG = -1,
```
Dentomologist, line 227: This doesn't actually change the value of INVALID_REG; 0xFFFFFFFF was overflowing the int underlying type and wrapping around to -1.

Source/Core/Common/Arm64Emitter.h
```diff
@@ -224,9 +224,65 @@ enum ARM64Reg
...
+  return static_cast<ARM64Reg>(static_cast<int>(lhs) - static_cast<int>(rhs));
+}
+constexpr int operator<<(const ARM64Reg& reg, const int distance)
+{
+  return static_cast<int>(reg) << distance;
+}
+constexpr int operator>>(const ARM64Reg& reg, const int distance)
+{
+  return static_cast<int>(reg) >> distance;
+}
```
Dentomologist, line 285: I don't like the inconsistency of returning an int for the bitwise and shift operators and ARM64Reg for + and -, but the point of these operators is to reduce line noise by getting rid of a bunch of casts, and those return types reflect the expected type in most of those usages.

Source/Core/Core/PowerPC/JitArm64/JitArm64_RegCache.cpp
```diff
@@ -744,7 +744,7 @@ BitSet32 Arm64FPRCache::GetCallerSavedUsed() const
   for (const auto& it : m_host_registers)
   {
     if (it.IsLocked())
-      registers[it.GetReg() - Q0] = true;
+      registers[GetUsedRegisterIndex(it.GetReg())] = true;
```
Dentomologist, line 747: I believe the values in m_host_registers will all be Q registers which have values [192, 223].  Values in this range are mapped to [0, 31] both before and after this change.

Source/Core/Common/Arm64Emitter.h
JosJuice: I think this is reasonable for the most part, but why would someone want to add an (or subtract) an ARM64Reg with an ARM64Reg? ARM64Reg and int makes sense to me, but not two ARM64Reg.

Source/Core/Core/PowerPC/JitArm64/JitArm64_RegCache.cpp
JosJuice: Yes, this change is correct. The function name `GetUsedRegisterIndex` sounds a bit unwieldy, though... Now that I think about it, is there actually any reason why `DecodeReg` shouldn't return an `int`?

Dentomologist: There are a number of places where DecodeReg is called and the return value is passed into another function that expects an ARM64Reg, so I wouldn't want to get rid of it.

On the other hand, there are a number of places where the return value is used in bitwise operations which effectively treat it as an int.

I'm not sure if this is best done in this PR or moved to another one, but we may want to create another function and have one return an Arm64Reg and the other an int.  I don't know what we would name them; DecodeReg already sounds a bit opaque to me even though I know what it does, so that might be an opportunity to give it a clearer name, but perhaps that name already is clear to people like you who are familiar with ARM assembly.

Source/Core/Common/Arm64Emitter.h
Dentomologist: Adding two ARM64Regs is done in MOVI2FDUP and EncodeRegToSingle.  I don't think anything actually calls MOVI2FDUP so we could remove it, and if we made DecodeReg return an int (see other comment for more on the subject) then EncodeRegToSingle wouldn't need to add two ARM64Regs either.

Subtracting two registers was only done in the context of statements like fprs_in_use[VD - Q0], but since I'm replacing those anyway I suppose I could get rid of the subtraction operators.

Source/Core/Core/PowerPC/JitArm64/JitArm64_RegCache.cpp
JosJuice: > There are a number of places where DecodeReg is called and the return value is passed into another function that expects an ARM64Reg, so I wouldn't want to get rid of it.

After a quick look (I may have missed some instances), the functions I could find that were being called with the return value of `DecodeReg` were `UnlockRegister` (from JitArm64_RegCache) and `MOV`/`SXTW` (from JitArm64_Integer.cpp). In both of these cases, the `DecodeReg` call doesn't make sense and should just be removed as far as I can tell. (Things end up working anyway because the register in question is always a 32-bit GPR.)

Dentomologist: I started converting DecodeReg to return an int but found the change ended up being much more extensive than anticipated.

JitArm64::JumpIfCRFieldBit: calls DecodeReg and passes the result to either CBZ or CBNZ, which each insert the reg into a FixupBranch which is returned up to the bcx and bclrx functions, which pass the FixupBranch to SetJumpTarget, where the reg value is decoded a couple times and also passed to Is64Bit.

JitArm64::mcrxr: 
	calls LDRB, which calls one of two overloads of EncodeLoadStoreIndexedInst, which each call Is64Bit, IsVector, and DecodeReg. 
	calls ADD, which calls EncodeArithmeticInst, which calls Is64Bit and DecodeReg
	Initializes ArithOption, which initializes a member variable with it.

There are similar chains for crXXX, mfcr, mtcrf, cmpl, and cmpli in JitArm64, among others I haven't taken the time to investigate.

It would be possible to make all those changes, but it starts to approach the point of getting rid of ARM64Reg entirely.

JosJuice: At least for `JumpIfCRFieldBit` and `mcrxr`, the appropriate thing to do would be to create a function `EncodeRegTo32` which works identically to how `DecodeReg` currently works. (Decoding a register and converting a register to its 32-bit form are two rather distinct operations semantically, but because the enum happens to start with `W0`, `reg & 0x1F` can be used to implement both.)

Also, a correction: In my previous comment, I didn't take into account that CR registers normally are mapped to 64-bit GPRs. Instead of removing the usages of `DecodeReg` that I mentioned in that comment, please replace them with `EncodeRegTo32`.

JosJuice: Or in other words: Let's have both `DecodeReg` and `EncodeRegTo32`, with the former returning `int` and the latter returning `ARM64Reg`.

Source/Core/Common/Arm64Emitter.cpp
```diff
@@ -1223,44 +1177,44 @@ void ARM64XEmitter::_MSR(PStateField field, ARM64Reg Rt)
   int o0 = 0, op1 = 0, CRn = 0, CRm = 0, op2 = 0;
   ASSERT_MSG(DYNA_REC, Is64Bit(Rt), "MSR: Rt must be 64-bit");
   GetSystemReg(field, o0, op1, CRn, CRm, op2);
-  EncodeSystemInst(o0, op1, CRn, CRm, op2, DecodeReg(Rt));
+  EncodeSystemInst(o0, op1, CRn, CRm, op2, Rt);
```
Dentomologist, line 1180: Rt is decoded inside EncodeSystemInst, so there's no need to do it here.

Source/Core/Common/Arm64Emitter.h
```diff
@@ -256,17 +268,21 @@ constexpr bool IsGPR(ARM64Reg reg)
...
+  return static_cast<ARM64Reg>(DecodeReg(reg));
 }
 constexpr ARM64Reg EncodeRegTo64(ARM64Reg reg)
 {
   return static_cast<ARM64Reg>(reg | 0x20);
 }
 constexpr ARM64Reg EncodeRegToSingle(ARM64Reg reg)
 {
-  return static_cast<ARM64Reg>(DecodeReg(reg) + S0);
+  return static_cast<ARM64Reg>(ARM64Reg::S0 | DecodeReg(reg));
```
Dentomologist, line 285: S0 = 64 and DecodeReg return value is in range [0, 31] so these are equivalent.

Dentomologist:
As per previous discussion:

Made DecodeReg return int, and added EncodeRegTo32 which returns ARM64Reg
Replaced DecodeReg with EncodeRegTo32 where appropriate
Removed GetUsedRegisterIndex function and replaced calls with DecodeReg
Removed extraneous operator functions

Dentomologist:
One last thing before I remove draft status: this builds properly and isn't supposed to change any functionality, but since I don't have an ARM device to test it myself it would be helpful for someone who does to run a few games as a sanity check.

JosJuice:
I did some quick tests of a few games. Things seem to be working fine.

JosJuice (approved):
Code changes LGTM, but you should probably remove the lines of the commit message that come after "Most changes are just adding ARM64Reg:: in front of the constants."

Dentomologist:
> Code changes LGTM, but you should probably remove the lines of the commit message that come after "Most changes are just adding ARM64Reg:: in front of the constants."

Done, and draft status removed.

JosJuice:
Gonna take the opportunity to merge this before someone merges one of my open JitArm64 PRs and causes a merge conflict. degasus was the one who suggested making this into an enum class so I'm going to assume he doesn't mind this being merged

PR #9496, opened 2021-02-08
Use QGridLayout for mapping settings

DacodaDragon:
Takes a wack at https://bugs.dolphin-emu.org/issues/11707#change-741102

Mapping button resizing grid

Picking at the low priority issues since I still lack experience doing more significant things. 

I'm considering making a separate class that inherits from QGridLayout to support the "AddRow" functionality in QFormLayout.

### DIFF
--- a/Source/Core/DolphinQt/Config/Mapping/MappingNumeric.cpp
+++ b/Source/Core/DolphinQt/Config/Mapping/MappingNumeric.cpp
@@ -83,8 +83,6 @@ MappingBool::MappingBool(MappingWidget* parent, ControllerEmu::NumericSetting<bo
 
   connect(parent, &MappingWidget::ConfigChanged, this, &MappingBool::ConfigChanged);
   connect(parent, &MappingWidget::Update, this, &MappingBool::Update);
-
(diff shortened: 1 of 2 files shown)


Miksel12:
Looks great, I got stuck on the "Connect MotionPlus" widget. I assume that is the reason why you need to add "AddRow" functionality?

DacodaDragon:
> Looks great, I got stuck on the "Connect MotionPlus" widget. I assume that is the reason why you need to add "AddRow" functionality?

image

I'm currently tinkering around with this. It would be nice if it would somewhat follow the behavior of the other ones or stay aligned to the left entirely.

Dentomologist:
I think a better approach would be to keep using the form layout but remove the maximum width on the buttons.

In addition to that being more consistent with most of the config buttons, your .gif shows that even max-width buttons can't fit input expressions longer than a certain length. While there will always be expressions too long to fit in a given width, we can at least let that be dictated by the window width/user's monitor rather than an arbitrary limit.

jordan-woyak:
The mentioned issue has instead been solved by PR #11271.

PR #9497, opened 2021-02-09
Graphics refactoring + add names and descriptions in FIFO analyzer

Pokechu22:
The big thing here is names and detailed descriptions for most XF, BP, and CP registers, and the names being shown in the overview, along with refactoring of the graphics code to use `BitField` and `enum class`.  ~~Additionally, the way the frame and object ranges work has been tweaked to be inclusive (for frames) and not require selecting 2 objects.~~

For the most part, I've done the descriptions by creating custom [`fmt::formatter`s](https://fmt.dev/latest/api.html#formatting-user-defined-types), and then using `fmt::format` or `fmt::to_string` on the objects.  I construct the objects to print using e.g. `LitChannel{.hex = value}` (since most of them do not have constructors).  For enums, I've also introduced a new `EnumFormatter` that makes it easy to create `fmt::formatter`s for the enums (including clean handling of invalid values).  I also *tried* to expand most of the abbreviations, but I may have done some of them wrong.  There may be typos or other mistakes in the results.

<details><summary>Screenshots</summary>

XFMEM_SETCHAN_ALPHA example
BPMEM_GENMODE example
CP_VAT_REG_A example
</details>

~~Two things that I'd like to do that I haven't included here: 1, Show state changes before the draw instead of after (see the unintuitive warning [in the wiki](https://github.com/dolphin-emu/dolphin/wiki/FIFO-Player-Overview#how-to-analyze-a-fifo-log)), and 2, make the FIFO player a separate window that can be shown over the render window (currently, at least on Windows, the main dolphin window will always show up behind the FIFO player, blocking the render window).~~ Further changes to the FIFO analyzer itself are found in #9540, which has a weak dependency on this PR.


Source/Core/Core/FifoPlayer/FifoPlayer.cpp
```diff
@@ -189,6 +186,18 @@ bool FifoPlayer::IsRunningWithFakeVideoInterfaceUpdates() const
   return m_File->ShouldGenerateFakeVIUpdates();
 }
 
+u32 FifoPlayer::GetMaxObjectCount() const
+{
+  u32 result = 0;
+  for (u32 i = 0; i < m_FrameInfo.size(); i++)
```
lioncash, line 192: Would this not be able to be a ranged for loop? If not, the iteration variable should likely be typed as a `size_t` instead.

Source/Core/Core/FifoPlayer/FifoPlayer.cpp
```diff
@@ -189,6 +186,18 @@ bool FifoPlayer::IsRunningWithFakeVideoInterfaceUpdates() const
...
+  {
+    u32 count = static_cast<u32>(m_FrameInfo[i].objectStarts.size());
```
lioncash, line 194:
```suggestion
    const u32 count = static_cast<u32>(m_FrameInfo[i].objectStarts.size());
```

Source/Core/DolphinQt/FIFO/FIFOAnalyzer.cpp
```diff
@@ -484,11 +495,34 @@ void FIFOAnalyzer::UpdateDescription()
   }
   else if (*cmddata == OpcodeDecoder::GX_LOAD_CP_REG)
   {
+    u8 cmd = *(cmddata + 1);
```
lioncash, line 498: Locals should be marked as const where applicable. Ditto where applicable elsewhere.

Source/Core/VideoCommon/BPStructs.cpp
```diff
@@ -753,61 +753,808 @@ void LoadBPRegPreprocess(u32 value0)
...
-void GetBPRegInfo(const u8* data, std::string* name, std::string* desc)
+template <>
+struct fmt::formatter<GenMode>
 {
-  const char* no_yes[2] = {"No", "Yes"};
+  constexpr auto parse(format_parse_context& ctx) { return ctx.begin(); }
+  template <typename FormatContext>
+  auto format(const GenMode& mode, FormatContext& ctx)
+  {
+    const char* cullmodes[] = {"None", "Back-facing primitives only",
```
lioncash, line 763:
```suggestion
    static constexpr std::array cullmodes = {"None", "Back-facing primitives only",
```

Ditto elsewhere where applicable. Otherwise MSVC generally constructs the arrays on the stack every time the function is executed, as opposed to just storing the data in the R/O segment.

Source/Core/VideoCommon/BPStructs.cpp
```diff
@@ -753,61 +753,808 @@ void LoadBPRegPreprocess(u32 value0)
...
+struct fmt::formatter<UPE_Copy>
+{
+  constexpr auto parse(format_parse_context& ctx) { return ctx.begin(); }
+  template <typename FormatContext>
+  auto format(const UPE_Copy& copy, FormatContext& ctx)
+  {
+    const char* no_yes[] = {"No", "Yes"};
+    const char* frame_to_field[] = {"Progressive", "Invalid", "Interlaced (even lines)",
+                                    "Interlaced (odd lines)"};
+    std::string clamp;
```
lioncash, line 1020:
```suggestion
    std::string_view clamp;
```

given this is only ever assigned string literals. Ditto for other relevant variables.

Source/Core/VideoCommon/BPStructs.cpp
```diff
@@ -753,61 +753,808 @@ void LoadBPRegPreprocess(u32 value0)
...
+template <>
+struct fmt::formatter<GenMode>
 {
-  const char* no_yes[2] = {"No", "Yes"};
+  constexpr auto parse(format_parse_context& ctx) { return ctx.begin(); }
+  template <typename FormatContext>
+  auto format(const GenMode& mode, FormatContext& ctx)
+  {
+    const char* cullmodes[] = {"None", "Back-facing primitives only",
+                               "Front-facing primitives only", "All primitives"};
```
lioncash, line 764:
```suggestion
                               "Front-facing primitives only", "All primitives",
                           };
```

A comma after the last element of the array will make the ending brace get formatted in a more immediately visable manner. Ditto for other relevant occurrences.

Source/Core/VideoCommon/XFStructs.h
```diff
@@ -3,5 +3,11 @@
 // Refer to the license.txt file included.
 
 #pragma once
+#include <string>
```
lioncash, line 6: There should be a newline after the pragma

Source/Core/VideoCommon/XFStructs.cpp
```diff
@@ -300,3 +300,511 @@ void PreprocessIndexedXF(u32 val, int refarray)
...
+  case XFMEM_SETMATRIXINDA:
+    return std::make_pair(RegName(XFMEM_SETMATRIXINDA),
+                          fmt::format("Matrix index A: {}", TMatrixIndexA{.Hex = value}));
+  case XFMEM_SETMATRIXINDB:
+    return std::make_pair(RegName(XFMEM_SETMATRIXINDB),
+                          fmt::format("Matrix index B: {}", TMatrixIndexB{.Hex = value}));
+
+  case XFMEM_SETVIEWPORT:
+    return std::make_pair(RegName(XFMEM_SETVIEWPORT + 0),
+                          fmt::format("Viewport width: {}", reinterpret_cast<float&>(value)));
```
lioncash, line 560: Consider using BitCast within Common/BitUtils for conversions like this. This invokes undefined behavior as is. Ditto for other occurrences below.

Source/Core/VideoCommon/XFStructs.cpp
```diff
@@ -300,3 +300,511 @@ void PreprocessIndexedXF(u32 val, int refarray)
...
+  if (baseAddress >= 0x1000)
+  {
+    name += fmt::format("Write {} XF regs at {:04x}", endAddress - baseAddress, baseAddress);
+
+    for (u32 address = baseAddress; address < endAddress; address++)
+    {
+      u32 value = Common::swap32(data);
+
+      auto [regname, regdesc] = GetXFRegInfo(address, value);
+      desc += regname + "\n" + regdesc + "\n";
```
lioncash, line 803:
```suggestion
      desc += fmt::format("{}\n{}\n", regname, regdesc);
```

So this doesn't churn string allocations.

Source/Core/VideoCommon/XFStructs.cpp
```diff
@@ -300,3 +300,511 @@ void PreprocessIndexedXF(u32 val, int refarray)
...
+    {
+      if (i > 0)
+        desc += '\n';
+      desc += GetXFMemDescription(xfMemBase + i, Common::swap32(data));
+      data += 4;
+    }
+
+    if (endAddress > 1000)
+    {
+      name += " ";
```
lioncash, line 789:
```suggestion
      name += ' ';
```

Source/Core/VideoCommon/XFStructs.cpp
```diff
@@ -300,3 +300,511 @@ void PreprocessIndexedXF(u32 val, int refarray)
...
+  {
+    return fmt::format("Unknown memory {:04x} = {:08x}", address, value);
+  }
+}
+
+std::pair<std::string, std::string> GetXFTransferInfo(const u8* data)
+{
+  const u32 cmd = Common::swap32(data);
+  data += 4;
+  u32 baseAddress = cmd & 0xFFFF;
```
lioncash, line 739: `snake_case` for new code. Ditto elsewhere this is applicable.

Source/Core/DolphinQt/FIFO/FIFOAnalyzer.cpp
```diff
@@ -241,14 +246,16 @@ void FIFOAnalyzer::UpdateDetails()
         const u8* stream_start = objectdata;
         const u8* stream_end = stream_start + streamSize * 4;
 
-        new_label = QStringLiteral("XF  %1  ").arg(cmd2, 8, 16, QLatin1Char('0'));
+        new_label = QStringLiteral("XF  %1  ").arg(cmd2, 8, 16, QLatin1Char('0'));;
```
lioncash, line 249: Redundant semicolon?

Source/Core/VideoCommon/BPStructs.cpp
Pokechu22: The ending comma causes it to be formatted like this:

```C++
    const char* cullmodes[] = {
        "None",
        "Back-facing primitives only",
        "Front-facing primitives only",
        "All primitives",
    };
```

I somewhat prefer having it on fewer lines, especially for arrays where each element is only a few characters and it fits on a single line (e.g. `{"None", "S", "T", "ST", "U", "SU", "TU", "STU"}`).  Should I change all of those, or should I leave the single-line ones as-is?

lioncash: I'm fine with keeping one liner arrays, where everything is more or less the same predictable length, I think it's fine to place it on a single line. Ones that have very different string lengths, like `cullmodes` I think should have a comma, since it makes more sense to make those vertically identifiable, so they aren't accidentally scanned over when reading left to right.

Source/Core/DolphinQt/FIFO/FIFOAnalyzer.cpp
Pokechu22: Most of these have been done, though I think some more in FIFOAnalyzer could be cleaned up even if I didn't directly change them.

Source/Core/VideoCommon/XFStructs.cpp
```diff
@@ -300,3 +301,431 @@ void PreprocessIndexedXF(u32 val, int refarray)
...
+      break;
+    case XF_SRCTEX6_INROW:
+      sourcerow = "Tex 6";
+      break;
+    case XF_SRCTEX7_INROW:
+      sourcerow = "Tex 7";
+      break;
+    }
+    return format_to(ctx.out(),
+                     "Projection: {}\nInput form: {}\nTex gen type: {}\n"
```
iwubcode, line 439: Should this be size instead of 'projection'?  I think of different things when I hear 'projection' but maybe I'm misunderstanding

Source/Core/VideoCommon/XFStructs.cpp
```diff
@@ -300,3 +301,431 @@ void PreprocessIndexedXF(u32 val, int refarray)
...
+  case XFMEM_VTXSPECS:  //__GXXfVtxSpecs, wrote 0004
+    return std::make_pair(RegName(XFMEM_VTXSPECS),
+                          fmt::format("Vertex specs: {}", INVTXSPEC{.hex = value}));
+
+  case XFMEM_SETNUMCHAN:
+    return std::make_pair(RegName(XFMEM_SETNUMCHAN),
+                          fmt::format("Number of color channels: {}", value & 3));
+    break;
+
+  case XFMEM_SETCHAN0_AMBCOLOR:  // Channel Ambient Color
```
iwubcode, line 521: Too bad we can't somehow pass the color from  here to Qt.  It'd be nice to somehow see the color in addition to the hex code

Source/Core/VideoCommon/XFStructs.cpp
```diff
@@ -300,3 +301,431 @@ void PreprocessIndexedXF(u32 val, int refarray)
...
+}
+
+std::pair<std::string, std::string> GetXFTransferInfo(const u8* data)
+{
+  const u32 cmd = Common::swap32(data);
+  data += 4;
+  u32 base_address = cmd & 0xFFFF;
+  const u32 transfer_size = ((cmd >> 16) & 15) + 1;
+
+  if (base_address > 0x1058)
```
iwubcode, line 743: I'm not really good with this low level stuff.  But where do these hex values come from?  Maybe a comment unless it's just normal known information

Source/Core/VideoCommon/BPStructs.cpp
```diff
@@ -753,61 +753,396 @@ void LoadBPRegPreprocess(u32 value0)
...
+      texture_format = "C8";
+      break;
+    case TextureFormat::C14X2:
+      texture_format = "C14X2";
+      break;
+    case TextureFormat::CMPR:
+      texture_format = "CMPR";
+      break;
+    case TextureFormat::XFB:
+      // Is this actually a real value, or just an implementation convenience for Dolphin?
```
iwubcode, line 1229: This is not a real format, implementation convenience

Source/Core/VideoCommon/XFStructs.cpp
Pokechu22: YAGCD calls it projection, while libogc's [`GX_SetTexCoordGen`](https://libogc.devkitpro.org/gx_8h.html#a7d3139b693ace5587c3224e7df2d8245) calls it `tgen_typ` and has [`GX_TG_MTX3x4` and `GX_TG_MTX2x4`](https://libogc.devkitpro.org/group__texgentyp.html#ga37b44995ede4932a201bc8edc2578c9e); the 3 by 4 version divides by `Q` which I think is the actual projection step.  But I don't think calling it projection is particularly obvious (not is typ(e)); size might be better.

Source/Core/VideoCommon/XFStructs.cpp
Pokechu22: Short answer: they come from copying code in `XFRegWritten` and `LoadXFReg`.  Longer answer: it's `XFMEM_SETPOSTMTXINFO + 7` + 1, or one after the last register.  Probably there isn't actually a reason to treat it specially (since there's already a warning when writing to unknown registers), but there probably should be constants for that (`XFMEM_REGISTERS_END` maybe) and the start of the registers too.

Source/Core/VideoCommon/BPStructs.cpp
Pokechu22: Removed (along with `EFBCopyFormat::XFB`), but I think it's somewhat odd that `XFB` has the value 0xF allocated to it, when that value is within the normal bounds.  `-1` or `0xffffffff` or something would make it more obvious that it's not something that games would be able to trigger.

Source/Core/VideoCommon/XFStructs.cpp
iwubcode: @lioncash - slightly off-topic but I've been curious about this for a while.  I wasn't able to find any documentation about fmt vs string concatenation in terms of memory allocations.

It makes sense to me for the line you mentioned.  Or even `fmt::format("{}\n", str);`.

But I'm curious, what about `fmt::format("{}{}", str1, str2);`?  Any idea on how fmt handles memory allocation?

Source/Core/VideoCommon/RenderBase.cpp
```diff
@@ -1697,7 +1697,7 @@ bool Renderer::UseVertexDepthRange() const
     return false;
 
   // We need a full depth range if a ztexture is used.
-  if (bpmem.ztex2.type != ZTEXTURE_DISABLE && !bpmem.zcontrol.early_ztest)
+  if (bpmem.ztex2.op != ZTexOp::Disabled && !bpmem.zcontrol.early_ztest)
```
Pokechu22, line 1700: This seems to have been a typo originally; `type` was the format (previously `TEV_ZTEX_TYPE_U8`/`_U16`/`_U24`).  I'm not sure if that affected anything though.

Source/Core/VideoCommon/BPMemory.h
```diff
@@ -354,169 +469,384 @@ struct TevStageCombiner
...
+};
 
 union TexMode0
 {
-  enum TextureFilter : u32
+  BitField<0, 2, WrapMode> wrap_s;
+  BitField<2, 2, WrapMode> wrap_t;
+  BitField<4, 1, FilterMode> mag_filter;
+  BitField<5, 2, MipMode> min_mip;
+  BitField<7, 1, FilterMode> min_filter;
```
Pokechu22, line 698: This was split from what was originally `mag_filter : 1` and `min_filter : 3`.  I think the split makes sense (the code that used it treated the `min_filter` bit differently from the other 2, though `min_mip` is probably a bad name.

Source/Core/VideoCommon/BPMemory.h
```diff
@@ -354,169 +469,384 @@ struct TevStageCombiner
...
-    u32 rid : 8;
-  };
+  BitField<0, 3, u32> bi0;  // Indirect tex stage 0 ntexmap
+  BitField<3, 3, u32> bc0;  // Indirect tex stage 0 ntexmap
+  BitField<6, 3, u32> bi1;
+  BitField<9, 3, u32> bc1;
+  BitField<12, 3, u32> bi2;
+  BitField<15, 3, u32> bc3;  // Typo?
+  BitField<18, 3, u32> bi4;
+  BitField<21, 3, u32> bc4;
```
Pokechu22, line 609: These names are somewhat odd since it skips `bc2` and `bi3`; I don't think it should have gone up to 4 at all.  I've left this unchanged in case there's some intended meaning though.

Source/Core/VideoCommon/RenderState.cpp
```diff
@@ -169,38 +182,39 @@ void BlendingState::Generate(const BPMemory& bp)
...
   {
     bool subtract;
-    BlendMode::BlendFactor srcfactor;
-    BlendMode::BlendFactor dstfactor;
+    SrcBlendFactor srcfactor;
+    DstBlendFactor dstfactor;
   };
+  // TODO: This previously had a warning about SRC and DST being aliased and not to mix them,
+  // but INVSRCCLR and INVDSTCLR were also aliased and were mixed.
+  // Thus, NOR, EQUIV, INVERT, COPY_INVERTED, and OR_INVERTED duplicate(d) other values.
```
Pokechu22, line 193: I split `BlendFactor` into `SrcBlendFactor` and `DstBlendFactor` since `DstClr`/`InvDstClr` and `SrcClr`/`InvSrcClr` have differences in behavior and are worth treating differently, and I found that some entries in this table use `InvDstClr`/`InvSrcClr` even though they're not actually available.  I'm not sure exactly what to do here, but the behavior here may be (and have been) wrong.

Source/Core/DolphinNoGUI/PlatformX11.cpp
```diff
@@ -46,7 +46,7 @@ class PlatformX11 : public Platform
 
   Display* m_display = nullptr;
   Window m_window = {};
-  Cursor m_blank_cursor = None;
+  Cursor m_blank_cursor = 0L;  // None
```
Pokechu22, line 49: This is some jank due to me needing to undefine `None` which `X.h` defines to be 0L, since that define interferes with some enums.  I'm not sure if there's a better workaround.

Pokechu22 (commented):
This is much closer to a final version (though I still need to rebase and tidy up the commits).

I've converted a lot of graphics code to use `BitField` and `enum class`es, and have created `fmt::formatter`s for most of those enums (and a `EnumFormatter` to simplify creating such fomatters) and the corresponding structures.  Those formatters are used in the fifo analyzer but also can be used in logging, which is fairly convenient (and thus I've put them into the header file instead of the cpp files used by the fifo analyzer).

One issue caused by the use of `enum class` is that there are many `std::array`s that map some enum into different host API equivalents or code snippets for shaders.  These can't be indexed using an `enum class`, so I need to cast them; I've been doing e.g. `u32(state.logicmode.Value())` (the `Value()` part is needed because of `BitField`; it cannot cast directly to `u32`).  This is a bit messy, but I'm not sure if there's a better solution.

PatrickFerry:
Note:
This currently is crashing to desktop in windows.

As an example: https://bugs.dolphin-emu.org/attachments/8076/nes%20nekkotsu.dff

Pokechu22:
More specifically it plays correctly the first time, but after stopping playback, trying to play again or analyze the opened FIFO causes a crash.  (I'm _guessing_ that that's also what's breaking the radeon fifoci builder, since it's segfaulting, but I haven't confirmed).

Source/Core/VideoCommon/VertexLoader.cpp
```diff
@@ -318,17 +318,17 @@ void VertexLoader::CompileVertexTranslator()
...
       ASSERT_MSG(VIDEO, DIRECT <= tc[i] && tc[i] <= INDEX16,
                  "Invalid texture coordinates!\n(tc[i] = %d)", (u32)tc[i]);
-      ASSERT_MSG(VIDEO, FORMAT_UBYTE <= format && format <= FORMAT_FLOAT,
-                 "Invalid texture coordinates format!\n(format = %d)", format);
-      ASSERT_MSG(VIDEO, 0 <= elements && elements <= 1,
-                 "Invalid number of texture coordinates elements!\n(elements = %d)", elements);
+      ASSERT_MSG(VIDEO, ComponentFormat::UByte <= format && format <= ComponentFormat::Float,
+                 "Invalid texture coordinates format!\n(format = %d)", (u32)format);
+      ASSERT_MSG(VIDEO, elements == TexComponentCount::S || elements == TexComponentCount::ST,
+                 "Invalid number of texture coordinates elements!\n(elements = %d)", (u32)elements);
```
Pokechu22, line 331: Note the `(u32)format` here; ideally I'd just let the enum formatter handle this, but there doesn't seem to be an fmt equivalent of ASSERT_MSG.

Pokechu22:
I've fixed a typo that was causing rendering differences (044b07832e844a260fda713c0a73007c82ebe002), including [slightly different rendering in Need For Speed: Underground](https://fifo.ci/compare/5895123-5892167/) (JMC confirmed that the blue windows are what show up on hardware, even though they do look somewhat worse).  I've also rebased to the latest master, which includes #9501, so FifoCI should run properly now.

I've also fixed the crashes with the fifo player.  One (which triggered on playing back a fifolog a second time) was introduced in 17186b0c9162b27a2c2ee49eeb36333a6d9f84c9, and fixed in a1273206f12aafc229dda69c8beae7593635da28 by fixing a few places where I didn't update to `m_File->GetFrameCount() - 1` for inclusive maximums.  A second occurred when trying to analyze the fifolog after playback stopped and the file was closed; this was fixed in 9a32a135a2fc385f8a7065e0b44a26f72816028a (note that the call to `m_analyzer->Update()` sets the text to "No recording loaded.", but the file is closed a bit before that happens, so the checks for `FifoPlayer::GetInstance().IsPlaying()` are also needed).  That bug affected 5.0-13603.  I also (hopefully) fixed a 3rd bug (in a23b63892a43fa884142c57be5abd3d28bb92362), where on stopping playback the player would deadlock in `WriteFifo`, [waiting for `IsHighWatermarkSet`](https://github.com/dolphin-emu/dolphin/blob/a2fa89b15ef67eea63ef6a3c7e0cc0e3629ea71e/Source/Core/Core/FifoPlayer/FifoPlayer.cpp#L376-L380); this was much less consistent but I did manage to reproduce it in 5.0-13603.

Pokechu22:
I've created a new `BitFieldArray` type that can be used to access multiple sequential bitfields (for instance, the `Tex0Coord`-`Tex7Coord` fields in `TVtxDesc`), and I've also exposed `BitField`'s `StorageType` (it still defaults to the same thing it did before, but now it can be manually specified, allowing `BitField`s of `bool`).

I've also changed the fifo player window to be a separate window (that can be maximized and doesn't cause the main window to cover the render window), and done a bit more `BitField`-related adjustment of some of the video registers (in particular, `TVtxDesc` now uses `BitField`, and is packed in a way that matches how the values are stored on console (17 and 16 bits) instead of being packed tightly (32 bits and *1* bit); this was changed in a way that should hopefully not break savestates or shader UIDs though I haven't tested it).

What remains before I mark this ready for review is fixing the most recent compile errors and [fifoci regressions](https://fifo.ci/version/79cbc7aa9100307661c705b22723affc294a1f71/), and then rebasing everything so that it's in a more sane series of commits where each builds and the changes are specific and complete (so that the massive diff isn't as much of an issue).

Pokechu22:
I've made it so that register updates are shown before the object's primitive information, instead of after the primitive information in the previous object.  Besides making things easier to understand, this also shows the register updates for object 0, which were previously not visible(!).  This required a bit of refactoring in FIFOAnalyzer and its search functionality.

I've also added descriptions for GX_LOAD_INDX_A/B/C/D (stating what parts of xfmem were updated and from where in memory, though not the new values since I'm not sure how to get the appropriate parts of memory), and also made it so that multiple NOPs in a row are shown only as a single entry.

Source/Core/DolphinQt/FIFO/FIFOAnalyzer.cpp
```diff
@@ -158,155 +164,251 @@ void FIFOAnalyzer::UpdateTree()
...
 
-      case OpcodeDecoder::GX_CMD_CALL_DL:
-        // The recorder should have expanded display lists into the fifo stream and skipped the
-        // call to start them
-        // That is done to make it easier to track where memory is updated
-        ASSERT(false);
-        objectdata += 8;
-        new_label = QStringLiteral("CALL DL");
-        break;
+        const u16 vertex_count = Common::swap16(fifo_frame.fifoData[object_offset]);
```
Pokechu22, line 360: This is a typo, for two reasons: one, it should use `object[object_offset]`, but two, it's passing a u8 to `swap16` when it should be passing a `u8*`.  I've made this mistake several times and it is stupidly easy to miss unfortunately.  This one caused the number of vertices to usually be zero, which really confused me.

Source/Core/Common/EnumFormatter.h
```diff
@@ -0,0 +1,91 @@
...
+  constexpr auto parse(fmt::format_parse_context& ctx)
+  {
+    auto it = ctx.begin(), end = ctx.end();
+    // 'u' for user display, 's' for shader generation
+    if (it != end && (*it == 'u' || *it == 's'))
+      formatting_for_shader = (*it++ == 's');
+
+    // TODO: Fix this check
+    // if (it != end && *it != '}')
+    //   throw fmt::format_error("invalid format");
```
Pokechu22, line 65: This check is actually unneeded as fmt will [trigger an error anyways](https://github.com/fmtlib/fmt/blob/7e72673d87cc44340f4b808d0b536e4e191342b5/include/fmt/format.h#L3193-L3194).  I'll remove it.

Pokechu22:
I've rebased, reordered, and squashed together commits to make something more reviewable (and that builds on every single commit, at least in MSVC).  For CP, XF, and BP, there is one commit that converts to enum class + bitfields and introduces formatters for the new enums (as some of them are used in shaders), and a second commit that creates formatters for each struct/union and uses that to create a description in the FIFO analyzer.  Overall, it should be possible to review this by looking over each commit in sequence.

Source/Core/VideoCommon/VertexLoader.cpp
```diff
@@ -312,38 +306,42 @@ void VertexLoader::CompileVertexTranslator()
...
-                 "Invalid texture coordinates!\n(tc[i] = %d)", (u32)tc[i]);
-      ASSERT_MSG(VIDEO, FORMAT_UBYTE <= format && format <= FORMAT_FLOAT,
-                 "Invalid texture coordinates format!\n(format = %d)", format);
-      ASSERT_MSG(VIDEO, 0 <= elements && elements <= 1,
-                 "Invalid number of texture coordinates elements!\n(elements = %d)", elements);
+      ASSERT_MSG(VIDEO,
+                 VertexComponentFormat::NotPresent <= m_VtxDesc.high.TexCoord[i] &&
+                     m_VtxDesc.high.TexCoord[i] <= VertexComponentFormat::Index16,
+                 "Invalid texture coordinates!\n(tc[i] = %d)",
+                 (u32)m_VtxDesc.high.TexCoord[i].Value());
```
Pokechu22, line 324: Typo here; should use `VertexComponentFormat::Direct` instead (though in practice it won't matter due to the if statement).  Not sure if the if these assertions really matter either.

Rumi-Larry:
I think a good first step to getting this merged, would be to split all the changes that are cleanup or minor improvements, and another pr for refactoring and the new functions in other prs

Source/Core/VideoCommon/XFMemory.h
```diff
@@ -17,75 +21,164 @@ constexpr size_t NUM_XF_COLOR_CHANNELS = 2;
...
+struct fmt::formatter<TexGenType> : EnumFormatter<TexGenType::ColorSTRGBC1>
+{
+  static constexpr array_type names = {
+      "Regular",
+      "Emboss map (used when bump mapping)",
+      "Color STRGBC0",
+      "Color STRGBC1",
+  };
+  formatter() : EnumFormatter(names) {}
 };
```
Pokechu22, line 77: I've just realized that `STRGBC0` is not supposed to be parsed as "STR" "GBC" "0" (store gbc 0) but instead "ST" "RGB" "C0" (Possibly still store, but I think it's actually the normal ST, using color 0's RGB).  It's also a bit odd since there currently is an assumption that it's only used in AB11 (in the software renderer), so only the R and G components are used.  Maybe it should be named just `Color0` or `Color0RG`.

Pokechu22:
I don't think splitting it is feasible.  I could split off the `FIFOAnalyzer` specific changes, but the ones that add descriptions strongly depend on the conversion to `enum class` and `BitField`; it's not actually possible to write `formatter`s like I did when using raw bitfields, and without the enums it's much more annoying to write them.

Source/Core/VideoCommon/TextureDecoder.h
```diff
@@ -32,8 +33,17 @@ enum class TextureFormat
   // Special texture format used to represent YUVY xfb copies.
   // They aren't really textures, but they share so much hardware and usecases that it makes sense
   // to emulate them as part of texture cache.
+  // This is not a real value that can be used on console; it exists only for eas of implementation.
```
Pokechu22, line 36: Typo, should be `ease`.  Same applies to line 95.

Source/Core/DolphinQt/FIFO/FIFOAnalyzer.cpp
```diff
@@ -478,14 +478,13 @@ void FIFOAnalyzer::UpdateDescription()
   QString text;
   if (*cmddata == OpcodeDecoder::GX_LOAD_BP_REG)
   {
-    std::string name;
-    std::string desc;
-    GetBPRegInfo(cmddata + 1, &name, &desc);
+    const u8 cmd = *(cmddata + 1);
+    const u32 value = Common::swap32(cmddata + 1) & 0xFFFFFF;
+    auto [name, desc] = GetBPRegInfo(cmd, value);
```
Pokechu22: Should be const.  Same for previous commit (175e9a02685c5a67f4309d45aedc4895e0f4797c, which I think should be merged with this one, and I'm not sure actually compiles when it's before this).  Also should probably use `Common::swap24` here.

Source/Core/VideoCommon/BPMemory.h
```diff
@@ -949,6 +1327,19 @@ union FieldMask
...
+struct fmt::formatter<FieldMask>
+{
+  constexpr auto parse(format_parse_context& ctx) { return ctx.begin(); }
+  template <typename FormatContext>
+  auto format(const FieldMask& mask, FormatContext& ctx)
+  {
+    return format_to(ctx.out(),
+                     "Enable odd field: {}\n"
+                     "Enable even field: {}",
+                     mask.odd ? "Yes" : "No", mask.even ? "Yes" : "No");
```
Pokechu22, line 1340: These descriptions might be backwards based on the comment above:

https://github.com/dolphin-emu/dolphin/blob/950cfcec9bd8c3253b11bca64426b3886c5456dc/Source/Core/VideoCommon/BPMemory.h#L1323-L1329

Source/Core/VideoCommon/BPStructs.cpp
```diff
@@ -745,61 +745,60 @@ void LoadBPRegPreprocess(u32 value0)
...
+    u32 matrix_num = (cmd - BPMEM_IND_MTXA) / 3;
+    u32 matrix_col = (cmd - BPMEM_IND_MTXA) % 3;
+    // These all use the same structure, though the meaning is *slightly* different;
+    // for conveninece implement it only once
+    s32 row0 = cmddata & 0x0007ff;           // ma or mc or me
+    s32 row1 = (cmddata & 0x3ff800) >> 11;   // mb or md or mf
+    u32 scale = (cmddata & 0xc00000) >> 22;  // 2 bits of a 6-bit field for each column
+
+    float row0f = static_cast<float>(row0) / (1 << 10);
+    float row1f = static_cast<float>(row0) / (1 << 10);
```
Pokechu22, line 787: These should be `const`

Source/Core/VideoCommon/BPStructs.cpp
```diff
@@ -893,359 +892,302 @@ void GetBPRegInfo(const u8* data, std::string* name, std::string* desc)
...
 
   case BPMEM_COPYFILTER0:  // 0x53
-    SetRegName(BPMEM_COPYFILTER0);
-    // TODO: Description
-    break;
+  {
+    u32 w0 = (cmddata & 0x00003f);
+    u32 w1 = (cmddata & 0x000fc0) >> 6;
+    u32 w2 = (cmddata & 0x03f000) >> 12;
+    u32 w3 = (cmddata & 0xfc0000) >> 18;
```
Pokechu22, line 987: `const`

Source/Core/VideoCommon/BPStructs.cpp
```diff
@@ -869,9 +870,8 @@ void GetBPRegInfo(const u8* data, std::string* name, std::string* desc)
   case BPMEM_TREF + 5:
   case BPMEM_TREF + 6:
   case BPMEM_TREF + 7:
-    SetRegName(BPMEM_TREF);
-    // TODO: Description
-    break;
+    return std::make_pair(fmt::format("BPMEM_IREF number {}", cmd - BPMEM_TREF),
```
Pokechu22, line 873: Should say `BPMEM_TREF`.  The `RegName` macro usually covers things like this (though I mainly used a macro like that because one was used before; I'm not fully convinced it really helps), but I didn't use it in cases where `fmt` was used.

Source/Core/VideoCommon/BPStructs.cpp
```diff
@@ -869,9 +870,8 @@ void GetBPRegInfo(const u8* data, std::string* name, std::string* desc)
...
+                          fmt::format("Configuration: {}", TevStageIndirect{.fullhex = cmddata}));
```
Pokechu22, line 874: ... and it's using the wrong type; it should be `TwoTevStageOrders` here.

Source/Core/DolphinQt/FIFO/FIFOPlayerWindow.cpp
```diff
@@ -297,14 +297,14 @@ void FIFOPlayerWindow::OnFIFOLoaded()
...
+  auto object_count = FifoPlayer::GetInstance().GetMaxObjectCount();
   auto frame_count = file->GetFrameCount();
 
   m_frame_range_to->setMaximum(frame_count);
-  m_object_range_to->setMaximum(object_count);
+  m_object_range_to->setMaximum(object_count - 1);
 
   m_frame_range_to->setValue(frame_count);
-  m_object_range_to->setValue(object_count);
+  m_object_range_to->setValue(object_count - 1);
```
Pokechu22: These (/ `FIFOPlayerWindow::OnFIFOLoaded()` in general) are also called when the FIFO is unpaused, which turns out to be rather annoying since it changes the upper bound.  (Also a problem with 5.0-13603)

Source/Core/VideoCommon/XFStructs.cpp
```diff
@@ -300,3 +303,246 @@ void PreprocessIndexedXF(u32 val, int refarray)
...
+  case XFMEM_SETPOSTMTXINFO + 1:
+  case XFMEM_SETPOSTMTXINFO + 2:
+  case XFMEM_SETPOSTMTXINFO + 3:
+  case XFMEM_SETPOSTMTXINFO + 4:
+  case XFMEM_SETPOSTMTXINFO + 5:
+  case XFMEM_SETPOSTMTXINFO + 6:
+  case XFMEM_SETPOSTMTXINFO + 7:
+    return std::make_pair(
+        fmt::format("XFMEM_SETPOSTMTXINFO Matrix {}", address - XFMEM_SETPOSTMTXINFO),
+        fmt::format("Matrix {}: {}", address - XFMEM_SETPOSTMTXINFO, TexMtxInfo{.hex = value}));
```
Pokechu22, line 458: Should be `PostMtxInfo`

Source/Core/VideoCommon/XFStructs.cpp
```diff
@@ -300,3 +303,246 @@ void PreprocessIndexedXF(u32 val, int refarray)
...
+    return DescriptionlessReg(XFMEM_CLOCK);
+  case XFMEM_SETGPMETRIC:
+    return DescriptionlessReg(XFMEM_SETGPMETRIC);
+
+  case XFMEM_CLIPDISABLE:
+    return std::make_pair(
+        RegName(XFMEM_CLIPDISABLE),
+        fmt::format("{} clipping detection, {} trivial rejection, {} cpoly clipping acceleration",
+                    (value & 1) ? "Enable" : "Disable", (value & 2) ? "enable" : "disable",
+                    (value & 3) ? "enable" : "disable"));
```
Pokechu22, line 339: These are backwards (if the bit is set, it's disabled).  This should also be a union (even though dolphin doesn't currently do anything with these values).

Source/Core/VideoCommon/BPStructs.cpp
```diff
@@ -893,359 +892,302 @@ void GetBPRegInfo(const u8* data, std::string* name, std::string* desc)
...
-    break;
 
   case BPMEM_EFB_TL:  // 0x49
   {
-    SetRegName(BPMEM_EFB_TL);
     X10Y10 left_top;
     left_top.hex = cmddata;
-    *desc = fmt::format("Left: {}\nTop: {}", u32(left_top.x), u32(left_top.y));
+    return std::make_pair(RegName(BPMEM_EFB_TL), fmt::format("EFB Left: {}\nEFB Top: {}",
+                                                             u32(left_top.x), u32(left_top.y)));
```
Pokechu22, line 938: The `u32(left_top.x)`-type casts are no longer needed as `BitField` has a formatter for it (they were needed when using raw bitfields though).  Also, it can be made `const` using `const X10Y10 left_top{.hex = cmddata}`.

Source/Core/VideoCommon/BPMemory.h
Pokechu22: I've changed it to `mipmap_filter` as that name is [already used in RenderState](https://github.com/dolphin-emu/dolphin/blob/9d0983c9c988f81c73c9a4ed61835ca9ac20db06/Source/Core/VideoCommon/RenderState.cpp#L221).

Some reverse-engineering I did found that Nintendo also treats it as a single 3-bit field, but the functions that set it (`gx::GXInitTexObjLOD` (8035fd94) and `JSystem::loadTexNo` (802eea04) in Super Mario Sunshine) convert the parameter using a 3-bit lookup table (`gx::GX2HWFiltConv` (8040cee0) and `JSystem::GX2HWFiltConv` (8040cc70)) with contents [0, 4, 1, 5, 2, 6, 0, 0].  This seems like unnecessary jank that doesn't need to be kept around.

Source/Core/VideoCommon/BPMemory.h
Pokechu22: They weren't backwards (I misread the comment), but I've created an enum for these fields (and for `FieldMode`) to avoid confusion in the future.

Pokechu22:
I've split the fifo analyzer changes into another pull request (#9540).  This one now only contains the refactoring and the addition of names and descriptions for BP, CP, and XF registers.

Source/Core/VideoCommon/CPMemory.h
```diff
@@ -18,176 +20,239 @@ enum
...
 
-  MASK_INDEXED = 2,
+enum class ComponentFormat
+{
+  UByte = 0,   // 2 Cmp - Invalid for normals
+  Byte = 1,    // 3 Cmp
+  UShort = 2,  // Invalid for normals
+  Short = 3,
+  Float = 4,
+};
```
Pokechu22, line 47: Anyone know what the `2 Cmp` and `3 Cmp` comments might be referring to?  They've existed since [the initial megacommit](https://github.com/dolphin-emu/dolphin/blame/775dc8a9c0c83f7a2f776e97f5b90385c759535a/Source/Plugins/Plugin_VideoDX9/Src/CPStructs.h#L68-L69) (along with the `NA` comment on `FORMAT_16B_565` which I've removed).  The best guess I have is that "cmp" refers to "component" and the labels should instead apply to `CoordComponentCount`'s `XY` and `XYZ`, but that seems unlikely.  [Yagcd](http://hitmen.c02.at/files/yagcd/yagcd/chap5.html#sec5.11.2) doesn't mention anything like this either (ctrl+f Vertex Attribute Data Formats)

Source/Core/VideoCommon/CPMemory.cpp
```diff
@@ -33,3 +33,62 @@ void CopyPreprocessCPStateFromMain()
...
 }
+
+std::pair<std::string, std::string> GetCPRegInfo(u8 cmd, u32 value)
+{
+  // Most things here seem to contradict YAGCD; for example LoadCPReg seems to assume that there is
+  // only one vertex descriptor and anything starting with 0x50 or 0x60 writes it, while YAGCD says
+  // there are 8 of them.  Not sure what the deal with that is.
+  switch (cmd & 0xf0)
+  {
+  case 0x30:
```
iwubcode, line 44: Can we turn these magic numbers into a variable?  I'm surprised they don't exist somewhere already?

iwubcode: I guess the string is enough if these don't already exist in Dolphin, just surprised

Source/Core/VideoCommon/CPMemory.h
```diff
@@ -166,6 +170,65 @@ struct TVtxDesc
     high.Hex = value >> 17;
   }
 };
+template <>
+struct fmt::formatter<TVtxDesc::Low>
+{
+  constexpr auto parse(format_parse_context& ctx) { return ctx.begin(); }
+  template <typename FormatContext>
+  auto format(const TVtxDesc::Low& desc, FormatContext& ctx)
```
iwubcode, line 178: These formatters are so clean, love it!

Source/Core/VideoCommon/XFMemory.h
```diff
@@ -152,63 +245,58 @@ union LitChannel
   }
 };
 
+union ClipDisable
```
iwubcode, line 264: Nice!  Interested about this one.  I wonder if any games use it..

Source/Core/VideoCommon/XFStructs.cpp
```diff
@@ -93,7 +93,7 @@ static void XFRegWritten(int transferSize, u32 baseAddress, DataReader src)
       break;
 
     case XFMEM_DUALTEX:
-      if (xfmem.dualTexTrans.enabled != (newValue & 1))
+      if (xfmem.dualTexTrans.enabled != ((newValue & 1) != 0))
```
iwubcode, line 99: What happened here?

Source/Core/VideoCommon/BPMemory.h
```diff
@@ -104,140 +111,209 @@ enum
...
+  One = 12,
+  Half = 13,
+  Konst = 14,
+  Zero = 15
+};
+template <>
+struct fmt::formatter<TevColorArg> : EnumFormatter<TevColorArg::Zero>
+{
+  static constexpr array_type names = {
+      "prev.rgb", "prev.aaa", "c0.rgb",  "c0.aaa",  "c1.rgb", "c1.aaa", "c2.rgb",    "c2.aaa",
```
iwubcode, line 190: Is there a reason you went with `aaa`?  Could just be my time with shader languages but I was expecting `prev.a`

Source/Core/VideoCommon/BPMemory.h
```diff
@@ -544,109 +653,153 @@ struct FourTexUnits
...
+  CopyInverted = 12,
+  OrInverted = 13,
+  Nand = 14,
+  Set = 15
+};
+template <>
+struct fmt::formatter<LogicOp> : EnumFormatter<LogicOp::Set>
+{
+  static constexpr array_type names = {
+      "0",        "s & d",    "s & ~d", "s",      "~s & d", "d",      "s ^ d",    "s | d",
```
iwubcode, line 1059: This is nice but I'm wondering if the text in the enum would be more easily digested at a glance

Pokechu22: These names were [already present](https://github.com/dolphin-emu/dolphin/blob/9d0983c9c988f81c73c9a4ed61835ca9ac20db06/Source/Core/VideoCommon/BPStructs.cpp#L922-L924) before.  I agree that they're a bit hard to read at a glance; I could replace it with something like this:

```C++
  static constexpr array_type names = {
      "Clear (0)",
      "And (src & dst)",
      "And Reverse (src & ~dst)",
      "Copy (src)",
      "And Inverted (~src & dst)",
      "NoOp (dst)",
      "Xor (src ^ dst)",
      "Or (src | dst)",
      "Nor (~(src | dst))",
      "Equiv (~(src ^ dst))",
      "Invert (~dst)",
      "Or Reverse (src | ~dst)",
      "Copy Inverted (~src)",
      "Or Inverted (~src | dst)",
      "Nand (~(src & dst))",
      "Set (1)",
  };
```

Source/Core/VideoCommon/BPMemory.h
Pokechu22: These names [were already present before](https://github.com/dolphin-emu/dolphin/blob/9d0983c9c988f81c73c9a4ed61835ca9ac20db06/Source/Core/VideoCommon/BPStructs.cpp#L1289-L1292), so I didn't think about them too much.

I'm not super familiar with shader languages; is something like `color.rgb = prev.a` legal (meaning the same thing as `color.rgb = prev.aaa`)?  I know you can write `color.rgb = float3(prev.a)` (or something like that) at least.

Source/Core/VideoCommon/XFStructs.cpp
Pokechu22: In visual studio, it produces a signed-unsigned mismatch warning otherwise (which is then treated as an error).  That happens even if `(newValue & 1u)` or `u32(newValue & 1)` is used though `s32(newValue & 1)` works, as does the more natural `bool(newValue & 1)`.  The same also happens if `enabled.Value()` is used &mdash; I'm guessing it's because it's promoting the `bool` to an `int` and then getting mad when it's not unsigned.

Source/Core/VideoCommon/BPMemory.h
iwubcode: If you don't mind, I think that would be a good change.

Source/Core/VideoCommon/BPMemory.h
iwubcode: Ah, looking at the previous code, I think the triple-a is probably intentional to show that the value is a 3 float component all filled with the alpha.  Sorry, wasn't thinking about this very clearly last night.  This can be marked resolved.

Source/Core/VideoCommon/CPMemory.h
iwubcode: My take is that these were comments made in observation during reverse engineering.  I don't think they have any direct relevance to the datatype format.  But @hrydgard is still around.  Maybe he'll remember :D

Source/Core/VideoCommon/BPStructs.cpp
iwubcode: Honestly, I believe I just took value from phire 's [original code](https://github.com/phire/dolphin/blob/cb84c1e015b0e51bc8cc76e8195c0000dd39550d/Source/Core/VideoCommon/TextureDecoder.h#L65).  I can't say offhand whether it'd be bad to change this to something else, I'd have to go look at the usage, it's been too long!

Source/Core/VideoCommon/XFStructs.cpp
iwubcode: I guess we could use rich text tags like I did for the tooltips.  Kinda muddies the description text though..

Source/Core/DolphinQt/FIFO/FIFOAnalyzer.cpp
```diff
@@ -224,17 +226,22 @@ void FIFOAnalyzer::UpdateDetails()
         u32 cmd2 = *objectdata++;
         u32 value = Common::swap32(objectdata);
         objectdata += 4;
+        const auto [name, desc] = GetCPRegInfo(cmd2, value);
+        ASSERT(!name.empty());
 
-        new_label = QStringLiteral("CP  %1  %2")
+        new_label = QStringLiteral("CP  %1  %2   %3")
```
PatrickFerry, line 232:
```suggestion
        new_label = QStringLiteral("CP  %1  %2  %3")
```

Source/Core/DolphinQt/FIFO/FIFOAnalyzer.cpp
```diff
@@ -278,11 +287,17 @@ void FIFOAnalyzer::UpdateDetails()
...
-                        .arg(cmd2 >> 24, 2, 16, QLatin1Char('0'))
-                        .arg(cmd2 & 0xFFFFFF, 6, 16, QLatin1Char('0'));
+        const u8 cmd2 = *objectdata++;
+        const u32 cmddata = Common::swap24(objectdata);
+        objectdata += 3;
+
+        const auto [name, desc] = GetBPRegInfo(cmd2, cmddata);
+        ASSERT(!name.empty());
+
+        new_label = QStringLiteral("BP  %1 %2   %3")
```
PatrickFerry, line 297:
```suggestion
        new_label = QStringLiteral("BP  %1  %2  %3")
```

Source/Core/VideoCommon/CPMemory.h
hrydgard: I honestly don't remember. You're probably best off just deleting those comments, much wasn't understood properly back then...

Source/Core/VideoCommon/CPMemory.cpp
Pokechu22: I've created constants for these and used them in all the contexts that I could find.  I've also moved my comments about inconsistencies with YAGCD to those constants.

Miksel12:
The latest build fixes: https://bugs.dolphin-emu.org/issues/10931 
But when starting Toy Story 3, the message: "Unknown register 20 set to 00000000" shows up.

Pokechu22:
I don't think it should *fix* that, but I did add an additional panic alert in 83c9c99a30cc494165d3f69ff3acc9d96fc86062 that triggers when an unused CP command is used.  The asserts from issue 10931 still exist (and should still trigger in the same way); it's just that the other one is also triggering first; you should run into the other asserts if you ignore the first one.  ([Yagcd says](http://hitmen.c02.at/files/yagcd/yagcd/chap5.html#sec5.11.2) that 0x20 is `?` but notes its existence; it's possible that it does something that just hasn't been figured out).

However, both [BP](https://github.com/dolphin-emu/dolphin/blob/9d0983c9c988f81c73c9a4ed61835ca9ac20db06/Source/Core/VideoCommon/BPStructs.cpp#L715-L716) and [XF](https://github.com/dolphin-emu/dolphin/blob/9d0983c9c988f81c73c9a4ed61835ca9ac20db06/Source/Core/VideoCommon/XFStructs.cpp#L191-L194) use `WARN_LOG` for unknown registers instead of asserting; I could convert both the newly added case and those other assertions to use that.  That'd "fix" toy story 3 (but not really because if there's a reason it's writing those values, then that case is unhandled by dolphin).

Source/Core/VideoCommon/CPMemory.h
```diff
@@ -50,176 +52,239 @@ enum
...
 
-  // Easily index into the Position..Tex7Coord fields.
-  u32 GetVertexArrayStatus(int idx) { return (Hex >> (9 + idx * 2)) & 0x3; }
+  Low low;
+  High high;
+
+  // This structure was originally packed into bits 0..32, using 33 total bits.
+  // The actual format has 17 bits in the low one and 16 bits in the high one,
+  // but the old format is still supported for compatibility.
+  u64 GetLegacyHex() const { return (low.Hex & 0x1FFFF) | (high.Hex << 17); }
```
Pokechu22, line 195: This is implemented wrong, as it does 32-bit math and then promotes to u64 afterwards.  `high.Hex` should be `u64(high.Hex)`.

Source/Core/VideoCommon/VertexLoader.cpp
```diff
@@ -312,38 +306,42 @@ void VertexLoader::CompileVertexTranslator()
...
     m_native_vtx_decl.texcoords[i].type = VAR_FLOAT;
     m_native_vtx_decl.texcoords[i].integer = false;
 
-    const int format = m_VtxAttr.texCoord[i].Format;
-    const int elements = m_VtxAttr.texCoord[i].Elements;
+    const auto format = m_VtxAttr.texCoord[i].Format;
+    const auto elements = m_VtxAttr.texCoord[i].Elements;
 
-    if (tc[i] != NOT_PRESENT)
+    if (m_VtxDesc.high.TexCoord[i] != VertexComponentFormat::NotPresent)
```
Pokechu22, line 318: This code is just a bit more complicated looking because `m_VtxDesc.high.TexCoord[i]` is repeated a bit; I'll extract it into a variable named `tc` like `format` and `elements` above.

Source/Core/VideoCommon/VertexLoaderManager.cpp
```diff
@@ -72,16 +72,32 @@ void UpdateVertexArrayPointers()
...
+      cached_arraybases[i + 2] = Memory::GetPointer(g_main_cp_state.array_bases[i + 2]);
+  }
+
+  for (size_t i = 0; i < g_main_cp_state.vtx_desc.high.TexCoord.Size(); i++)
   {
-    // Only update the array base if the vertex description states we are going to use it.
-    if (g_main_cp_state.vtx_desc.GetVertexArrayStatus(i) & MASK_INDEXED)
-      cached_arraybases[i] = Memory::GetPointer(g_main_cp_state.array_bases[i]);
+    if (is_indexed(g_main_cp_state.vtx_desc.high.TexCoord[i]))
+      cached_arraybases[i + 4] = Memory::GetPointer(g_main_cp_state.array_bases[i + 4]);
```
Pokechu22: The different array numbers (0, 1, 2, 4, also 12 (not seen here)) should probably be extracted into constants.

Pokechu22: Looks like they actually are already constants, just not ones I used:

https://github.com/dolphin-emu/dolphin/blob/79ee6aae0a6a7add698bb7fbc0ef25aa166ccdec/Source/Core/VideoCommon/CPMemory.h#L10-L18

Source/Core/VideoBackends/Software/Tev.cpp
```diff
@@ -455,56 +425,56 @@ void Tev::Indirect(unsigned int stageNum, s32 s, s32 t)
...
     break;
-  case ITBA_T:
+  case IndTexBumpAlpha::T:
     AlphaBump = indmap[TextureSampler::BLU_SMP];
     break;
-  case ITBA_U:
+  case IndTexBumpAlpha::U:
     AlphaBump = indmap[TextureSampler::GRN_SMP];
     break;
   }
```
Pokechu22, line 440: This switch should have a `default` case with a PanicAlert (as basically all of the other ones in this file have one, even if it's guaranteed from the size of the BitField that they won't be hit).

Source/Core/VideoBackends/Vulkan/VKPipeline.cpp
```diff
@@ -85,28 +85,28 @@ static VkPipelineDepthStencilStateCreateInfo GetVulkanDepthStencilState(const De
...
     break;
-  case ZMode::GEQUAL:
+  case CompareMode::GEqual:
     compare_op = inverted_depth ? VK_COMPARE_OP_LESS_OR_EQUAL : VK_COMPARE_OP_GREATER_OR_EQUAL;
     break;
-  case ZMode::ALWAYS:
+  case CompareMode::Always:
     compare_op = VK_COMPARE_OP_ALWAYS;
     break;
   default:
```
Pokechu22, line 112: `PanicAlert` here in the default case

Source/Core/Common/EnumFormatter.h
```diff
@@ -0,0 +1,86 @@
...
+template <auto last_member, typename T = decltype(last_member),
+          size_t size = static_cast<size_t>(last_member) + 1,
+          std::enable_if_t<std::is_enum_v<T>, bool> = true>
+struct EnumFormatter
+{
+  // This is needed because std::array deduces incorrectly if nullptr is included in the list
+  using array_type = std::array<const char*, size>;
+  const array_type m_names;
+  bool formatting_for_shader = false;
+  EnumFormatter(const array_type names) : m_names(names) {}
```
iwubcode, line 54: explicit?

Also, if this wasn't const, you could use `std::move` on the names

Source/Core/Common/BitField.h
```diff
@@ -202,3 +203,299 @@ struct fmt::formatter<BitField<position, bits, T, S>>
...
+  constexpr BitFieldArrayIterator(BitFieldArray<position, bits, size, T, S>* array, size_t index)
+      : m_array(array), m_index(index)
+  {
+  }
+
+public:
+  // Required by std::input_or_output_iterator
+  constexpr BitFieldArrayIterator() = default;
+  // Required by LegacyIterator
+  constexpr BitFieldArrayIterator(const BitFieldArrayIterator& other) = default;
```
iwubcode, line 396: Do you expect move construction? You should explicitly define the move constructor/operator either way

Source/Core/Common/EnumFormatter.h
Pokechu22: `explicit` is probably a good idea, and the constructor should probably also be `protected` (and the fields also `private`) to more clearly convey how it should be used.  I think the constructor could also be `constexpr` but it doesn't look liek that really makes a difference.

`std::array` seems to support move construction even when const (at least when it's an array of `const char *`), but looking at compiler explorer, it doesn't seem to make a difference here.  I do think undue copies are a mild concern (I'm not 100% sure whether fmt constructs it each time and thus a new array is created each time, or if the array is constructed once, or what exactly); theoretically, I could instead have it take a const reference to a static constexpr field, but that would mean that a field is always needed.

Source/UnitTests/Common/BitFieldTest.cpp
```diff
@@ -252,3 +251,295 @@ TEST(BitField, Fmt)
...
+
+  for (s32 value : object.arr)
+  {
+    EXPECT_EQ(value, 0);
+  }
+
+  object.arr[0] = -2;
+  EXPECT_EQ(object.arr[0], -2);
+  EXPECT_EQ(object.a, -2);
+  EXPECT_EQ(object.hex, 0b00'00'10'00000);
```
iwubcode, line 354: Surprised here, there's no signed bit?  Need to look again..

Pokechu22: It's two's complement, so for a 2-bit field, `00` => 0, `01` => 1, `10` => -2, and `11` => -1.  The `10` there corresponds to the -2, while the other two fields are still 0.

Source/UnitTests/Common/BitFieldTest.cpp
```diff
@@ -2,10 +2,20 @@
 // Licensed under GPLv2+
 // Refer to the license.txt file included.
 
+#include <array>
```
Pokechu22: This include is unused (and was removed in the following commit)

Source/Core/VideoCommon/VertexLoaderManager.cpp
```diff
@@ -72,16 +72,35 @@ void UpdateVertexArrayPointers()
   if (!g_main_cp_state.bases_dirty)
     return;
 
+  // Only update the array base if the vertex description states we are going to use it.
+  auto is_indexed = [](VertexComponentFormat f) {
```
iwubcode, line 76: I've seen this multiple places.  Would it make sense to turn this into a free-form function that lives next to the enum?

Pokechu22: I think that could be nice; the one reason why I've been hesitant to do so is because BP/CP/XFMemory are all in global scope, and adding `IsIndexed` to the global scope seems a bit jank.  (The proper solution would be to refactor them into a namespace, which I guess isn't any harder if a function is added though.)

Source/Core/Core/FifoPlayer/FifoRecordAnalyzer.cpp
```diff
@@ -44,14 +45,28 @@ void FifoRecordAnalyzer::WriteVertexArray(int arrayIndex, const u8* vertexData,
...
+    arrayType = s_CpMem.vtxDesc.low.Normal;
+  else if (arrayIndex == 2 || arrayIndex == 3)
+    arrayType = s_CpMem.vtxDesc.low.Color[arrayIndex - 2];
+  else if (arrayIndex < 12)
+    arrayType = s_CpMem.vtxDesc.high.TexCoord[arrayIndex - 4];
+  else
+  {
+    PanicAlertFmt("Invalid arrayType {}", arrayType);
+    return;
+  }
```
Pokechu22, line 61: The PanicAlert should use arrayIndex, and also the `ARRAY_POSITION`/etc constants should be used here.

Source/Core/VideoCommon/UberShaderVertex.cpp
```diff
@@ -403,27 +403,27 @@ static void GenVertexShaderTexGens(APIType api_type, u32 num_texgen, ShaderCode&
...
-  out.Write("  case {}u: // XF_SRCBINORMAL_B_INROW\n", XF_SRCBINORMAL_B_INROW);
+  out.Write("  case {:s}:\n", SourceRow::BinormalB);
   out.Write(
       "    coord.xyz = ((components & {}u /* VB_HAS_NRM2 */) != 0u) ? rawnorm2.xyz : coord.xyz;",
       VB_HAS_NRM2);
   out.Write("    break;\n\n");
   for (u32 i = 0; i < 8; i++)
   {
-    out.Write("  case {}u: // XF_SRCTEX{}_INROW\n", XF_SRCTEX0_INROW + i, i);
+    out.Write("  case {}u: // SourceRow::Tex{}\n", static_cast<u32>(SourceRow::Tex0) + i, i);
```
iwubcode, line 426: Seems like you could do:

`case {:s}:\n", static_cast<SourceRow>(static_cast<int>(SourceRow::Tex0) + i))`

Just to keep consistent formatting?

Source/Core/VideoCommon/UberShaderVertex.cpp
```diff
@@ -462,14 +462,14 @@ static void GenVertexShaderTexGens(APIType api_type, u32 num_texgen, ShaderCode&
...
+  out.Write("  case {:s}:\n", TexGenType::Color0);
   out.Write("    output_tex.xyz = float3(o.colors_0.x, o.colors_0.y, 1.0);\n"
             "    break;\n\n");
-  out.Write("  case {}u: // XF_TEXGEN_COLOR_STRGBC1\n", XF_TEXGEN_COLOR_STRGBC1);
+  out.Write("  case {:s}:\n", TexGenType::Color1);
   out.Write("    output_tex.xyz = float3(o.colors_1.x, o.colors_1.y, 1.0);\n"
             "    break;\n\n");
-  out.Write("  default:  // Also XF_TEXGEN_REGULAR\n"
-            "    {{\n");
+  out.Write("  default:  // Also {}\n", TexGenType::Regular);
```
iwubcode, line 471: ` out.Write("  default:  // Also {:s}\n", TexGenType::Regular);`

?

Source/Core/VideoCommon/VertexLoaderX64.cpp
```diff
@@ -115,7 +116,7 @@ int VertexLoaderX64::ReadVertex(OpArg data, u64 attribute, int format, int count
 
   X64Reg coords = XMM0;
 
-  int elem_size = 1 << (format / 2);
+  int elem_size = 1 << (u32(format) / 2);
```
Pokechu22, line 119: This could also be isolated into a `size_t GetElementSize(ComponentFormat format)` since it appears in a few places (with the same issue that it would be in the global namespace for the time being).

Source/Core/VideoCommon/UberShaderVertex.cpp
Pokechu22: Semi-intentional, as it's showing up in a comment so `default:  // Also Regular (0)` looks better than `default:  // Also 0x0u /* Regular */`.  But perhaps just adding a `case 0x0u /* Regular */:` before the default would more clearly express the intent.

Source/Core/VideoBackends/Software/Clipper.cpp
```diff
@@ -489,13 +489,15 @@ bool CullTest(const OutputVertexData* v0, const OutputVertexData* v1, const Outp
 
   backface = normalZDir <= 0.0f;
 
-  if ((bpmem.genMode.cullmode & 1) && !backface)  // cull frontfacing
+  if ((bpmem.genMode.cullmode == CullMode::Back || bpmem.genMode.cullmode == CullMode::All) &&
```
iwubcode, line 492: These enum choices are odd with the comment.  Just checking to make sure that's intentional?

Source/Core/VideoCommon/UberShaderPixel.cpp
```diff
@@ -367,21 +367,21 @@ ShaderCode GenPixelShader(APIType ApiType, const ShaderHostConfig& host_config,
       "// which are common to both color and alpha channels\n"
       "bool tevCompare(uint op, int3 color_A, int3 color_B) {{\n"
       "  switch (op) {{\n"
-      "  case 0u: // TEVCMP_R8_GT\n"
+      "  case 0u: // TevCompareMode::R8, TevComparison::GT\n"
```
iwubcode, line 370: Surprised these aren't using the `case {:s}:\n` trick

Pokechu22: It's because there are two enums, instead of one pseudo-enum/set of random values that don't really make sense.  The values were previously these (note how they don't match what's in the switch statement):

```
TEVCMP_R8_GT = 8,
TEVCMP_R8_EQ = 9,
TEVCMP_GR16_GT = 10,
TEVCMP_GR16_EQ = 11,
TEVCMP_BGR24_GT = 12,
TEVCMP_BGR24_EQ = 13,
TEVCMP_RGB8_GT = 14,
TEVCMP_RGB8_EQ = 15,
TEVCMP_A8_GT = TEVCMP_RGB8_GT,
TEVCMP_A8_EQ = TEVCMP_RGB8_EQ
```

and the enum was used by `((cc.shift << 1) | cc.op | 8)`.  I decided to just split it up into two enums; for the software renderer I was able to split it into two parts (acting on `cc.compare` and then looking at whether it was using eq or gt), but for the hardware backends I left it in the combined form since it would require further restructuring for no real gain.

It does look like there are other switch statements in this file that hardcode enum values, though; those could be changed.

Source/Core/VideoBackends/Software/Clipper.cpp
Pokechu22: The mask is correct (1 is back, 2 is front, 3 is all), but the behavior seems odd to me.  YAGCD says that the bit is for what should be rejected (REJECT_EN in GEN_MODE near [§5.11.1](http://hitmen.c02.at/files/yagcd/yagcd/chap5.html#sec5.11.1)).  It's possible that the software renderer's definition of backface is reversed from the hardware renderer's, which was canceled out here.  (The changes needed for #9350 were less than I expected at least).

Source/Core/VideoCommon/UberShaderPixel.cpp
Pokechu22: Due to the binary search tree structure used for the other enums (in addition to the switch statements), I've decided not to deal with the hardcoded enum values in this PR; it can be handled in a later one.

Source/Core/VideoBackends/Software/Clipper.cpp
Pokechu22: I've added a TODO comment for this; it might be an inconsistency with the software renderer but probably shouldn't be addressed here.

Source/Core/DolphinLib.vcxproj
```diff
@@ -27,6 +27,12 @@
       <Project>{41279555-f94f-4ebc-99de-af863c10c5c4}</Project>
     </ProjectReference>
   </ItemGroup>
+  <ItemGroup>
+    <ClCompile Include="VideoCommon\VertexLoaderARM64.cpp" />
```
iwubcode, line 31: I think this should go in the props file?

Pokechu22: That shouldn't have been commited at all; I added it because VS didn't want to format the code unless it was included in the solution, and right-clicking it to add it to the solution puts it there...

Source/Core/VideoCommon/UberShaderVertex.cpp
```diff
@@ -462,13 +462,14 @@ static void GenVertexShaderTexGens(APIType api_type, u32 num_texgen, ShaderCode&
...
-  out.Write("  case {}u: // XF_TEXGEN_COLOR_STRGBC0\n", XF_TEXGEN_COLOR_STRGBC0);
+  out.Write("  case {:s}:\n", TexGenType::Color0);
   out.Write("    output_tex.xyz = float3(o.colors_0.x, o.colors_0.y, 1.0);\n"
             "    break;\n\n");
-  out.Write("  case {}u: // XF_TEXGEN_COLOR_STRGBC1\n", XF_TEXGEN_COLOR_STRGBC1);
+  out.Write("  case {:s}:\n", TexGenType::Color1);
   out.Write("    output_tex.xyz = float3(o.colors_1.x, o.colors_1.y, 1.0);\n"
             "    break;\n\n");
-  out.Write("  default:  // Also XF_TEXGEN_REGULAR\n"
+  out.Write("  case {:s}:\n", TexGenType::Regular);
```
iwubcode, line 471: Sorry, should have realized this earlier after your comment, maybe add:

```
[[fallthrough]];
```

iwubcode (approved):

PR #9498, opened 2021-02-10
Core/DSP: Fix improper uses of offsetof

leoetlino:
According to the C standard, an offsetof expression must evaluate to an
address constant, otherwise it's undefined behavior.

Fixes https://bugs.dolphin-emu.org/issues/12409

See also https://gcc.gnu.org/bugzilla/show_bug.cgi?id=95942

There are still improper uses of offsetof (mostly in JitArm64) but
fixing that will take more effort since there's a PPCSTATE_OFF wrapper
macro that is sometimes used with non-array members and sometimes used
with arrays and variable indices... Let's keep that for another PR.

### DIFF
--- a/Source/Core/Core/DSP/Jit/x64/DSPEmitter.cpp
+++ b/Source/Core/Core/DSP/Jit/x64/DSPEmitter.cpp
@@ -494,12 +494,13 @@ Gen::OpArg DSPEmitter::M_SDSP_external_interrupt_waiting()
 
 Gen::OpArg DSPEmitter::M_SDSP_r_st(size_t index)
 {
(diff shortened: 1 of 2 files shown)


Mystro256:
Thanks for the quick fix. I just did a build test on my Fedora 34 machine and it looks great! I haven't done tests running anything, but it compiles fine.

Mystro256:
FYI, I followed up with a build log of the issue with PPC if that helps. I realize you already said to keep this for another PR.
Let me know if you need any build feedback for PPC in case you don't have a machine available.

iwubcode (approved):
Just eyeballing, this seemed reasonable to me

PR #9500, opened 2021-02-11
PCAP: Improve TCP session logging

sepalani:
This PR improves TCP sessions logging for games using multiple sockets at the same time (for instance the Internet channel). That way, each socket tracks its sequence numbers and avoid sequence number mismatch when multiple of them are used.

Ready to reviewed & merged.

### DIFF
--- a/Source/Core/Core/NetworkCaptureLogger.h
+++ b/Source/Core/Core/NetworkCaptureLogger.h
@@ -6,6 +6,7 @@
 
 #include <array>
 #include <cstddef>
+#include <map>
 #include <memory>
 
 #ifdef _WIN32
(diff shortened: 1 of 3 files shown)


Source/Core/Core/NetworkCaptureLogger.h
```diff
@@ -43,6 +44,8 @@ class NetworkCaptureLogger
   NetworkCaptureLogger& operator=(NetworkCaptureLogger&&) = delete;
   virtual ~NetworkCaptureLogger();
 
+  virtual void LogSocket(s32 socket) = 0;
```
leoetlino, line 47: I don't think this function name reflects what PCAPSSLCaptureLogger::LogSocket actually does. I'm bad at naming, so I'm not sure what to suggest, but maybe "OnSocketAdded" which would IMO make it clearer that you *must* call this function when a socket is added for things to work properly? or register socket, I don't know

sepalani: Done. I chose `OnNewSocket` regarding the naming. The `AddSocket` function is just a helper used in the `NewSocket` and when `accept` returns a new socket.

PR #9501, opened 2021-02-11
hidapi: Use LIBUSB_LIBRARIES to link to libusb

leoetlino:
Because we have an old-style find script that does not define a proper
CMake target for libusb, we need to use ${LIBUSB_LIBRARIES}
rather than "usb".

Ideally, we would fix the find script to define a target. However,
this issue breaks the fifoci-ogl-lin-mesa builder and I'd prefer it
to be fixed sooner rather than later.

### DIFF
--- a/Externals/hidapi/CMakeLists.txt
+++ b/Externals/hidapi/CMakeLists.txt
@@ -14,7 +14,7 @@ else()
     target_link_libraries(hidapi PRIVATE udev)
   else()
     target_sources(hidapi PRIVATE libusb/hid.c)
-    target_link_libraries(hidapi PRIVATE usb)
+    target_link_libraries(hidapi PRIVATE ${LIBUSB_LIBRARIES})
   endif()
 endif()
 


iwubcode:
> Because we have an old-style find script that does not define a proper
CMake target for libusb, we need to use ${LIBUSB_LIBRARIES}
rather than "usb".

I'm fairly familiar with CMake but I don't know much about Dolphin's build system.  I didn't realize we were exposing these to the find functionality.  I assumed the `add_library` like defined [here](https://github.com/dolphin-emu/dolphin/blob/603668037665582af498e474697ecc1df3cf04d1/Externals/libusb/CMakeLists.txt#L1) was good enough.  Where do expose these external's to CMake's search path?

leoetlino:
The issue isn't in our Externals CMakeLists.txt, but in FindLibUSB.cmake, which doesn't define a real CMake target.

MerryMage (approved):
works locally, thanks

PR #9504, opened 2021-02-12
IOS: Use less ambiguous names for classes

leoetlino:
Some of the device names can be ambiguous and require fully or partly
qualifying the name (e.g. IOS::HLE::FS::) in a somewhat verbose way.

Additionally, insufficiently qualified names are prone to breaking.
Consider the example of IOS::HLE::FS:: (namespace) and
IOS::HLE::Device::FS (class). If we use FS::Foo in a file that doesn't
know about the class, everything will work fine. However, as soon as
Device::FS is declared via a header include or even just forward
declared, that code will cease to compile because FS:: now resolves
to Device::FS if FS::Foo was used in the Device namespace.

It also leads to having to write IOS::ES:: to access ES types and
utilities even for code that is already under the IOS namespace.

The fix for this is simple: rename the device classes and give them
a "device" suffix in their names if the existing ones may be ambiguous.
This makes it clear whether we're referring to the device class or to
something else.

This is not any longer to type, considering it lets us get rid of the
Device namespace, which is now wholly unnecessary.

There are no functional changes in this commit.

A future commit will fix unnecessarily qualified names.

### DIFF
--- a/Source/Core/Core/IOS/WFS/WFSI.h
+++ b/Source/Core/Core/IOS/WFS/WFSI.h
@@ -32,12 +32,10 @@ private:
   std::vector<u8> m_whole_file;
 };
 
-namespace Device
-{
-class WFSI : public Device
+class WFSIDevice : public Device
 {
 public:
-  WFSI(Kernel& ios, const std::string& device_name);
+  WFSIDevice(Kernel& ios, const std::string& device_name);
 
   IPCCommandResult IOCtl(const IOCtlRequest& request) override;
 
@@ -125,5 +123,4 @@ private:
     IOCTL_WFSI_CHECK_HAS_SPACE = 0x95,
   };
 };
-}  // namespace Device
 }  // namespace IOS::HLE
--- a/Source/Core/Core/IOS/WFS/WFSSRV.h
+++ b/Source/Core/Core/IOS/WFS/WFSSRV.h
@@ -29,12 +29,10 @@ enum
   WFS_FILE_IS_OPENED = -10032,  // Cannot perform operation on an opened file.
 };
 
-namespace Device
-{
-class WFSSRV : public Device
+class WFSSRVDevice : public Device
 {
 public:
-  WFSSRV(Kernel& ios, const std::string& device_name);
+  WFSSRVDevice(Kernel& ios, const std::string& device_name);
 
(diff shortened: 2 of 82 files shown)


iwubcode (approved):
Mostly looked at the diff but the few files I checked in full didn't seem to be missing any updated references.  LGTM

iwubcode:
On an unrelated note, I noticed there were a few device classes that could potentially be marked `final`

leoetlino:
Yeah, most of them are in older code. That can be done in a follow-up PR -- this one is already large enough :P

Source/Core/Core/IOS/WFS/WFSI.h
```diff
@@ -32,8 +32,6 @@ class ARCUnpacker
   std::vector<u8> m_whole_file;
 };
 
-namespace Device
-{
 class WFSI : public Device
```
BhaaLseN, line 35: Do we want to name this one `WFSIDevice` for consistency?

Source/Core/Core/IOS/WFS/WFSSRV.h
```diff
@@ -29,8 +29,6 @@ enum
   WFS_FILE_IS_OPENED = -10032,  // Cannot perform operation on an opened file.
 };
 
-namespace Device
-{
 class WFSSRV : public Device
```
BhaaLseN, line 32: Do we want to name this one `WFSSRVDevice` for consistency?

BhaaLseN (approved):
Pretty verbose diff, but thats expected on such a change. And it reads nicer I'd say, especially when theres a `Device` class inside a `Device` namespace and other `Device` subclasses using the `-Device` suffix already...
It feels a bit weird that you left the WFS stuff, but I don't insist on the changes if you had good reason to leave them (plus `OH0` threw me off before I noticed we already had a `OH0Device` as well).

Source/Core/Core/IOS/WFS/WFSI.h
leoetlino: Oh whoops, I forgot to rename this. I left the USB classes alone to avoid having everything called a "device" in the USB code (considering we have IOS devices and lots of USB devices there...) For WFS, I don't think there's any particular reason not to use the suffix.

PR #9508, opened 2021-02-13
IOS: Clean up the way IPC replies are constructed

leoetlino:
Instead of constructing IPCCommandResult with static member functions
in the Device class, we can just add the relevant constructors to the
reply struct itself. Makes more sense than putting it in Device when the
struct is also used in the kernel code and doesn't use any Device
specific members...

This commit also changes the IPC command handlers to return an optional
IPCCommandResult rather than an IPCCommandResult. This removes the need
for a separate boolean that indicates whether the "result" is actually
a reply, and also avoids the need to set dummy result values and ticks.

It also makes it really obvious which commands can result in no reply
being generated.

Finally, this commit renames IPCCommandResult to IPCReply since the
struct is now only used for actual replies. This new name is less
verbose in my opinion.

The diff is quite large since this touches every command handler, but
the only functional change is that I fixed EnqueueIPCReply to
take a s64 for cycles_in_future to match IPCReply.

### DIFF
--- a/Source/Core/Core/IOS/DI/DI.cpp
+++ b/Source/Core/Core/IOS/DI/DI.cpp
@@ -57,13 +57,13 @@ void DIDevice::DoState(PointerWrap& p)
   p.Do(m_last_length);
 }
 
-IPCCommandResult DIDevice::Open(const OpenRequest& request)
+std::optional<IPCReply> DIDevice::Open(const OpenRequest& request)
 {
   InitializeIfFirstTime();
   return Device::Open(request);
 }
 
-IPCCommandResult DIDevice::IOCtl(const IOCtlRequest& request)
+std::optional<IPCReply> DIDevice::IOCtl(const IOCtlRequest& request)
 {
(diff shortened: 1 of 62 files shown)


Source/Core/Core/IOS/DI/DI.cpp
```diff
@@ -57,13 +57,13 @@ void DIDevice::DoState(PointerWrap& p)
   p.Do(m_last_length);
 }
 
-IPCCommandResult DIDevice::Open(const OpenRequest& request)
+std::optional<IPCReply> DIDevice::Open(const OpenRequest& request)
```
leoetlino, line 60: Quick guide to reviewing:

* command handlers return `std::optional<IPCReply>` or `IPCReply`
* IPCCommandResult -> IPCReply
* GetNoReply -> std::nullopt
* GetDefaultReply -> IPCReply

BhaaLseN (approved):

PR #9509, opened 2021-02-13
NetworkWidget: Add hostname to SSL table

sepalani:
This PR adds the hostname column to the SSL table in the Network widget. This name is registered when initialising a new SSL context and will be used to verify the certificate CN if the verify option is enabled.


leoetlino (approved):
Pretty straightforward.

PR #9511, opened 2021-02-14
IOS: Emulate FS timings for ES content IPC commands

leoetlino:
Filesystem accesses aren't magically faster when they are done by ES,
so this commit changes our content wrapper IPC commands to take FS
access times and read operations into account.

This should make content read timings a lot more accurate and closer
to console. Note that the accuracy of the timings are limited to the
accuracy of the emulated FS timings, and currently performance
differences between IOS9-IOS28 and newer IOS versions are not emulated.

Part 1 of fixing https://bugs.dolphin-emu.org/issues/11346
(part 2 will involve emulating those differences)

### DIFF
--- a/Source/Core/Core/IOS/ES/ES.h
+++ b/Source/Core/Core/IOS/ES/ES.h
@@ -105,10 +105,10 @@ public:
   std::vector<std::array<u8, 20>> GetSharedContents() const;
 
   // Title contents
-  s32 OpenContent(const ES::TMDReader& tmd, u16 content_index, u32 uid);
-  ReturnCode CloseContent(u32 cfd, u32 uid);
-  s32 ReadContent(u32 cfd, u8* buffer, u32 size, u32 uid);
-  s32 SeekContent(u32 cfd, u32 offset, SeekMode mode, u32 uid);
+  s32 OpenContent(const ES::TMDReader& tmd, u16 content_index, u32 uid, Ticks ticks = {});
+  s32 CloseContent(u32 cfd, u32 uid, Ticks ticks = {});
+  s32 ReadContent(u32 cfd, u8* buffer, u32 size, u32 uid, Ticks ticks = {});
+  s32 SeekContent(u32 cfd, u32 offset, SeekMode mode, u32 uid, Ticks ticks = {});
 
   // Title management
   enum class TicketImportType
@@ -371,7 +371,7 @@ private:
   struct OpenedContent
   {
     bool m_opened = false;
-    FS::Fd m_fd;
+    u64 m_fd;
     u64 m_title_id = 0;
     ES::Content m_content;
     u32 m_uid = 0;
--- a/Source/Core/Core/IOS/ES/ES.h
+++ b/Source/Core/Core/IOS/ES/ES.h
@@ -87,8 +87,8 @@ public:
     No = false,
   };
 
-  ES::TMDReader FindImportTMD(u64 title_id) const;
-  ES::TMDReader FindInstalledTMD(u64 title_id) const;
+  ES::TMDReader FindImportTMD(u64 title_id, Ticks ticks = {}) const;
+  ES::TMDReader FindInstalledTMD(u64 title_id, Ticks ticks = {}) const;
   ES::TicketReader FindSignedTicket(u64 title_id) const;
 
   // Get installed titles (in /title) without checking for TMDs at all.
@@ -364,9 +364,9 @@ private:
   void FinishStaleImport(u64 title_id);
   void FinishAllStaleImports();
 
-  std::string GetContentPath(u64 title_id, const ES::Content& content,
-                             const ES::SharedContentMap& map) const;
-  std::string GetContentPath(u64 title_id, const ES::Content& content) const;
+  std::string GetContentPath(u64 title_id, const ES::Content& content, Ticks ticks = {}) const;
+
+  static constexpr u64 IPC_OVERHEAD_TICKS = 2700_tbticks;
 
   struct OpenedContent
   {
--- a/Source/Core/Core/IOS/ES/ES.h
+++ b/Source/Core/Core/IOS/ES/ES.h
@@ -366,8 +366,6 @@ private:
 
   std::string GetContentPath(u64 title_id, const ES::Content& content, Ticks ticks = {}) const;
 
-  static constexpr u64 IPC_OVERHEAD_TICKS = 2700_tbticks;
-
   struct OpenedContent
   {
     bool m_opened = false;
--- a/Source/Core/Core/IOS/ES/Formats.cpp
+++ b/Source/Core/Core/IOS/ES/Formats.cpp
@@ -27,6 +27,7 @@
 #include "Core/CommonTitles.h"
 #include "Core/IOS/Device.h"
 #include "Core/IOS/FS/FileSystem.h"
+#include "Core/IOS/FS/FileSystemProxy.h"
 #include "Core/IOS/IOS.h"
 #include "Core/IOS/IOSC.h"
 #include "Core/IOS/Uids.h"
@@ -521,17 +522,21 @@ struct SharedContentMap::Entry
 };
 
 constexpr char CONTENT_MAP_PATH[] = "/shared1/content.map";
-SharedContentMap::SharedContentMap(std::shared_ptr<HLE::FS::FileSystem> fs) : m_fs{fs}
+SharedContentMap::SharedContentMap(std::shared_ptr<HLE::FSDevice> fs)
+    : m_fs_device{fs}, m_fs{fs->GetFS()}
 {
   static_assert(sizeof(Entry) == 28, "SharedContentMap::Entry has the wrong size");
 
   Entry entry;
-  const auto file = fs->OpenFile(PID_KERNEL, PID_KERNEL, CONTENT_MAP_PATH, HLE::FS::Mode::Read);
-  while (file && file->Read(&entry, 1))
+  s64 fd = fs->Open(PID_KERNEL, PID_KERNEL, CONTENT_MAP_PATH, HLE::FS::Mode::Read, {}, &m_ticks);
+  if (fd < 0)
+    return;
+  while (fs->Read(fd, &entry, 1, &m_ticks) == sizeof(entry))
   {
     m_entries.push_back(entry);
     m_last_id++;
   }
+  fs->Close(fd, &m_ticks);
 }
 
 SharedContentMap::~SharedContentMap() = default;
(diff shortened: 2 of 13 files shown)


iwubcode:
@leoetlino - anything in particular to test or anything to watch out for when testing?

JMC47:
The Wii System Menu is crashing when I try to access my Channel List in the Data Management screen.  This does not happen in master.

leoetlino:
@iwubcode just making sure the system menu and NAND (including WiiWare) titles still work, nothing too specific

JMC said most games don't care about the timing changes, it's just VC games that take longer to load now.

BhaaLseN (approved):
Code seems good, nothing that sticks out - but untested.

JMC47:
We solved the crash.  Netplay corrupted my NAND at some point.

leoetlino:
yep, turned out there was a missing TMD for Dokapon Kingdom. It's why we enforce the requirement that games *must* be launched at least once before its save can be imported, but I guess this doesn't happen with netplay.

Source/Core/Core/IOS/ES/TitleContents.cpp
```diff
@@ -11,43 +11,61 @@
...
 
-    auto file = m_ios.GetFS()->OpenFile(PID_KERNEL, PID_KERNEL, GetContentPath(title_id, content),
-                                        FS::Mode::Read);
-    if (!file)
-      return FS::ConvertResult(file.Error());
+    if (content.IsShared())
+      ticks += SmallFileReadTicks;
+
+    const auto [fs_reply, fd] = m_ios.GetFSDevice()->Open(
+        PID_KERNEL, PID_KERNEL, GetContentPath(title_id, content), FS::Mode::Read);
```
Pokechu22, line 46: Idea: wouldn't it be better to make the various calls out to the FS device return the number of ticks they take, instead of needing to compute the number of ticks in ES?

leoetlino:
After thinking about this a bit more, I'm not entirely satisfied with the way ticks are kept track of in the ES code -- I think that ideally ES shouldn't have to increment the number of ticks manually after each call to FS. I'll try to find a way to make that cleaner...

leoetlino:
Still WIP, but I changed the way ticks were accumulated. For the new version of the third and fourth commits, I recommend reviewing with "Hide whitespace changes" enabled to make the diff less noisy.

Source/Core/Core/IOS/ES/TitleContents.cpp
```diff
@@ -11,11 +11,19 @@
 #include "Common/MsgHandler.h"
 #include "Core/HW/Memmap.h"
 #include "Core/IOS/ES/Formats.h"
+#include "Core/IOS/FS/FileSystemProxy.h"
 #include "Core/IOS/Uids.h"
 
 namespace IOS::HLE
 {
-s32 ESDevice::OpenContent(const ES::TMDReader& tmd, u16 content_index, u32 uid)
+// IPC overhead caused by internal ES <-> FS communications.
```
leoetlino, line 19: to be renamed, as this is now used to model ES's own overhead

Source/Core/Core/IOS/ES/TitleContents.cpp
```diff
@@ -11,11 +11,19 @@
...
+constexpr u64 IPCOverheadTicks = 2700_tbticks;
+
+// Small file read (TMD or content map) overhead. This accounts for the open command,
+// the read (which is assumed to be approximately CLUSTER_READ_TICKS to simplify) and the close.
+constexpr u64 SmallFileReadTicks = 2 * IPCOverheadTicks + 120000_tbticks;
```
leoetlino, line 24: to be removed once GetContentPath can update the timing information directly

Source/Core/Core/IOS/IOS.h
```diff
@@ -45,6 +46,38 @@ struct IPCReply
   u64 reply_delay_ticks;
 };
 
+// Used to make it more convenient for functions to return timing information
+// without having to explicitly keep track of ticks in callers.
+class Ticks
+{
+public:
+  Ticks(u64* ticks = nullptr) : m_ticks(ticks) {}
```
iwubcode, line 54: `explicit` ?

Source/Core/Core/IOS/FS/FileSystemProxy.cpp
```diff
@@ -105,59 +112,78 @@ static IPCReply GetReplyForSuperblockOperation(ResultCode result)
...
+    return fd;
   }
 
-  const u64 ticks = EstimateFileLookupTicks(request.path, FileLookupMode::Normal);
+  ticks.AddTimeBaseTicks(EstimateFileLookupTicks(path, FileLookupMode::Normal));
 
-  auto backend_fd = m_ios.GetFS()->OpenFile(request.uid, request.gid, request.path,
-                                            static_cast<Mode>(request.flags & 3));
-  LogResult(backend_fd, "OpenFile({})", request.path);
+  auto backend_fd = m_ios.GetFS()->OpenFile(uid, gid, path, mode);
```
iwubcode, line 142: `ticks` doesn't seem to be used.  Did you mean to pass it into `OpenFile` here?

iwubcode: Actually, I see this quite a bit.  What is the point of updating the ticks when they are passed in as a copy?

leoetlino: The number of ticks is estimated and `ticks` is updated two lines above -- the underlying backend doesn't know anything about timings.

`ticks` isn't an integer, it's a small wrapper around a `u64*`, which is why it is passed by value.

iwubcode: Thanks @leoetlino for your explanation.  I believe in my second comment I was confused ^^. In my first comment,  I was aware of the `u64*` but thought the ticks were used in an underlying call somewhere.

Re-reading, I missed an important part.  The ticks are ultimately passed in to the `IPCReply` structure via the `MakeIPCReply` function which I think helps determine the delay call for the command?  Either the delay time is built up from 0 or it is built up from `IPCOverheadTicks` depending on the call.

Sorry for all the dumb comments!

Source/Core/Core/IOS/IOS.h
leoetlino: In this case, we actually want implicit conversions to Ticks as this class is for all intents and purposes just a wrapper around a u64* and having to write Ticks{...} every time would be very verbose.

Source/Core/Core/IOS/FS/FileSystemProxy.cpp
leoetlino: Yes, that's correct. MakeIPCReply is used to help construct an `IPCReply` with timing information and without having to duplicate the `IPCReply{..., ticks}` for every return.

Source/Core/Core/IOS/ES/TitleContents.cpp
```diff
@@ -110,6 +110,7 @@ IPCReply ESDevice::ReadContent(u32 uid, const IOCtlVRequest& request)
   const u32 size = request.io_vectors[0].size;
   const u32 addr = request.io_vectors[0].address;
 
+  INFO_LOG_FMT(IOS_ES, "ReadContent(uid={:x}, cfd={}, size={}, addr={:08x})", uid, cfd, size, addr);
```
sepalani: Shouldn't `uid={:#x}` be used to be consistent with the previous commit? If the hex prefix isn't or shouldn't be used I think it might be a good idea to set its length `08x` like `addr` instead.

Source/Core/Core/IOS/ES/Formats.cpp
```diff
@@ -600,32 +605,34 @@ bool SharedContentMap::WriteEntries() const
          HLE::FS::ResultCode::Success;
 }
 
-static std::pair<u32, u64> ReadUidSysEntry(const HLE::FS::FileHandle& file)
+static std::pair<u32, u64> ReadUidSysEntry(HLE::FSDevice& fs, u64 fd, u64* ticks)
 {
   u64 title_id = 0;
-  if (!file.Read(&title_id, 1))
+  if (fs.Read(fd, &title_id, 1, ticks) != sizeof(title_id))
```
sepalani, line 611: Can't a constant be used to indicate an invalid read or an invalid `title_id` rather than relying on `sizeof`?

Source/Core/Core/IOS/ES/Formats.cpp
```diff
@@ -600,32 +605,34 @@ bool SharedContentMap::WriteEntries() const
...
     return {};
 
   u32 uid = 0;
-  if (!file.Read(&uid, 1))
+  if (fs.Read(fd, &uid, 1, ticks) != sizeof(uid))
```
sepalani, line 615: Ditto for `sizeof` and `uid`.

Source/Core/Core/IOS/ES/Formats.cpp
```diff
@@ -600,32 +605,34 @@ bool SharedContentMap::WriteEntries() const
          HLE::FS::ResultCode::Success;
 }
 
-static std::pair<u32, u64> ReadUidSysEntry(const HLE::FS::FileHandle& file)
+static std::pair<u32, u64> ReadUidSysEntry(HLE::FSDevice& fs, u64 fd, u64* ticks)
```
sepalani, line 608: Most functions use `std::shared_ptr`, is there a reason to switch to a  `non-const reference` rather a pointer. It's less obvious that's it's being modified and also violates Dolphin's coding style.

Source/Core/Core/IOS/ES/NandUtils.cpp
```diff
@@ -17,35 +16,38 @@
...
 #include "Common/StringUtil.h"
 #include "Core/IOS/ES/ES.h"
 #include "Core/IOS/ES/Formats.h"
+#include "Core/IOS/FS/FileSystemProxy.h"
 #include "Core/IOS/Uids.h"
 
 namespace IOS::HLE
 {
-static ES::TMDReader FindTMD(FS::FileSystem* fs, u64 title_id, const std::string& tmd_path)
+static ES::TMDReader FindTMD(FSDevice& fs, const std::string& tmd_path, Ticks ticks)
```
sepalani, line 28: Ditto regarding the `non-const reference`.

Source/Core/Core/IOS/ES/ES.h
```diff
@@ -364,9 +364,9 @@ class ESDevice final : public Device
   void FinishStaleImport(u64 title_id);
   void FinishAllStaleImports();
 
-  std::string GetContentPath(u64 title_id, const ES::Content& content,
-                             const ES::SharedContentMap& map) const;
-  std::string GetContentPath(u64 title_id, const ES::Content& content) const;
+  std::string GetContentPath(u64 title_id, const ES::Content& content, Ticks ticks = {}) const;
+
+  static constexpr u64 IPCOverheadTicks = 2700_tbticks;
```
sepalani, line 369: The constant name should be capitalised not to be mistaken by a function.

sepalani (changes requested):

Source/Core/Core/IOS/ES/TitleContents.cpp
leoetlino: Good catch. I chose `#x` for consistency and because UIDs typically look like `0x1xxx`.

Source/Core/Core/IOS/ES/Formats.cpp
leoetlino: sizeof(title_id) *is* that constant. I could hardcode 4 but I think that's worse than using sizeof.

Source/Core/Core/IOS/ES/Formats.cpp
leoetlino: There's no point to passing a shared_ptr by copy considering this function doesn't share ownership. It just needs a `fs` instance that actually exists and pointer semantics aren't useful here since `fs` cannot be null (and *must not* be null) and it's already clear that the instance is being modified in some way.

The guideline about using pointers rather than references is meant for situations where the value that is modified is e.g. an integer or a string, when it's otherwise unclear that the value is modified. In this case, since a device is an opaque handle anyway, what additional information would `ReadUidSysEntry(fs.get(), fd, &m_ticks);` give you over `ReadUidSysEntry(*fs, fd, &m_ticks);`? Absolutely nothing, but with a pointer you open the door to ReadUidSysEntry being misused and called with nullptr when a reference tells you at the type system level that the function requires a valid fs.

sepalani: These explanations sound fine to me.

Source/Core/Core/IOS/ES/Formats.cpp
sepalani: I see, I just find odd that the `Read` function can have an error state and it's not being used to handle the error. I understand the reasoning behind this use of `sizeof` in the case of partially read. Maybe that's just me but I find it strange to read.

leoetlino: If there is an error, Read() returns a negative value so the comparison will be false in that case. Or am I misunderstanding your comment?

sepalani: I was just saying that having `if (Read(..., &var, ...) != READ_SIZE)` rather than `if (Read(..., &var, ...) < 0)`/`if (Read(..., &var, ...) == ERROR` seems a bit odd to read to me.

Feel free to disregard that comment.

leoetlino: Oh, I see what you mean. But if we do `if (Read(..., &var, ...) < 0)` we still also need to check for partial reads, so we'd have to do something like

```cpp
const int ret = Read(...);
if (ret < 0 || ret != READ_SIZE)
  ...
```
which seems redundant to me, since READ_SIZE is positive.

PR #9512, opened 2021-02-14
CodeViewWidget: Add WithDetailedUpdate to update CodeWidget

sepalani:
This PR adds another enum value to also update the function calls and callers which are CodeWidget views.

Ready to be reviewed.


dreamsyntax:
I've just noticed this does not update the Callstack. Opening a new PR with that change, would we be fine to use the same enum for that?

PR #4734, opened 2017-01-23
DolphinWX: Show the required IOS version in info panel

leoetlino:
This is useful to know which IOS version is required by a title without having to look at the TMD manually.

The IOS version row will only appear if there is a TMD.

if there is a TMD

PR #9513, opened 2021-02-14
Qt: Re-add IOS version to the game info tab

leoetlino:
This was accidentally removed during the Qt migration: https://github.com/dolphin-emu/dolphin/pull/4734

~~Also adds the TMD region to the game info tab, as it's useful for figuring out what region Dolphin is ultimately using for Wii games~~ The country field should suffice, as it shows the value of TypicalCountryForRegion(region) and TypicalCountryForRegion is bijective

### DIFF
--- a/Source/Core/DolphinQt/Config/InfoWidget.cpp
+++ b/Source/Core/DolphinQt/Config/InfoWidget.cpp
@@ -17,6 +17,7 @@
 
 #include "DiscIO/Blob.h"
 #include "DiscIO/Enums.h"
+#include "DiscIO/Volume.h"
 
 #include "DolphinQt/Config/InfoWidget.h"
 #include "DolphinQt/QtUtils/ImageConverter.h"
@@ -25,6 +26,8 @@
 
 InfoWidget::InfoWidget(const UICommon::GameFile& game) : m_game(game)
 {
+  m_volume = DiscIO::CreateVolume(m_game.GetFilePath());
+
   QVBoxLayout* layout = new QVBoxLayout();
 
   layout->addWidget(CreateFileDetails());
@@ -35,6 +38,8 @@ InfoWidget::InfoWidget(const UICommon::GameFile& game) : m_game(game)
 
   setLayout(layout);
 }
+
+InfoWidget::~InfoWidget() = default;
 
 QGroupBox* InfoWidget::CreateFileDetails()
 {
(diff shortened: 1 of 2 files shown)


Source/Core/DolphinQt/Config/InfoWidget.cpp
```diff
@@ -25,6 +26,8 @@
 
 InfoWidget::InfoWidget(const UICommon::GameFile& game) : m_game(game)
 {
+  m_volume = DiscIO::CreateVolume(m_game.GetFilePath());
```
leoetlino, line 29: I wasn't sure if it was worth invalidating game list caches and adding the TMD to GameFile, so I opted for a simpler method: open the volume ourselves in InfoWidget.

JosJuice:
Isn't the TMD's region field inaccurate for discs? At least I believe we don't use it when determining the region of a disc.

leoetlino:
Oh, right, it's only used for NAND titles. I'll add a "is disc title" check.

leoetlino:
Actually, I'll just use Volume::GetRegion :)

Source/Core/DolphinQt/Config/InfoWidget.cpp
```diff
@@ -121,6 +126,19 @@ QGroupBox* InfoWidget::CreateGameDetails()
   if (!m_game.GetApploaderDate().empty())
     layout->addRow(tr("Apploader Date:"), CreateValueDisplay(m_game.GetApploaderDate()));
 
+  if (m_volume)
+  {
+    layout->addRow(tr("Region:"), CreateValueDisplay(fmt::format("{}", m_volume->GetRegion())));
```
iwubcode, line 131: NBD but is there a way to make the region into text?  We could do:  `<region text> (<region code>)` which would be more understandable than just the number

iwubcode: I guess "country" is good enough...?

iwubcode (approved):
Tested a handful of Wii / Wiiware games.  Code LGTM

Source/Core/DolphinQt/Config/InfoWidget.cpp
leoetlino: Yeah I guess the country field is good enough since it's the result of TypicalCountryForRegion(region)... I'll drop the same commit

PR #9514, opened 2021-02-14
JitArm64: Fix improper uses of offsetof

JosJuice:
Using a non-constant array index inside offsetof is not standards compliant, and is rejected by GCC 11.

https://bugs.dolphin-emu.org/issues/12409#note-5

### DIFF
--- a/Source/Core/Core/PowerPC/JitArm64/JitArm64_RegCache.h
+++ b/Source/Core/Core/PowerPC/JitArm64/JitArm64_RegCache.h
@@ -6,6 +6,7 @@
 
 #include <cstddef>
 #include <memory>
+#include <type_traits>
 #include <vector>
 
 #include "Common/Arm64Emitter.h"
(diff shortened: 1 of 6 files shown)


Source/Core/Core/PowerPC/JitArm64/JitArm64_RegCache.h
```diff
@@ -25,10 +25,19 @@ constexpr Arm64Gen::ARM64Reg DISPATCHER_PC = Arm64Gen::W26;
 
 #define PPCSTATE_OFF(elem) (offsetof(PowerPC::PowerPCState, elem))
 
+#define PPCSTATE_OFF_ARRAY(elem, i)                                                                \
+  (offsetof(PowerPC::PowerPCState, elem[0]) + sizeof(PowerPC::PowerPCState::elem[0]) * (i))
+
+#define PPCSTATE_OFF_GPR(i) PPCSTATE_OFF_ARRAY(gpr, i)
+#define PPCSTATE_OFF_PS0(i) (PPCSTATE_OFF_ARRAY(ps, i) + offsetof(PowerPC::PairedSingle, ps0))
```
leoetlino, line 32: Could we add a static_assert to check that PowerPCState::ps[0] is a PairedSingle?

leoetlino (approved):

PR #9515, opened 2021-02-14
IOS/FS: Implement timings for older IOS versions as well

leoetlino:
According to hwtests, older versions of IOS are slower at performing
various filesystem operations:

https://docs.google.com/spreadsheets/d/1OKo9IUuKCrniz4m0kYIaMP_qFtOCmAzHZ_zAmobvBcc/edit

(courtesy of JMC)

021-read-16384-bytes-cached

(More plots available at https://dolphin.leolam.fr/fs_timing/)

A quick glance at IOS9 reveals that older versions of IOS have a
simplistic implementation of memcpy that does not optimize large copies
by copying 16 bytes or 32 bytes per chunk, which makes cached reads
and writes noticeably slower -- the difference was significant enough
that the OoT speedrunning community noticed that IOS9 (the IOS that
is used for the OoT VC title) was slower.

### DIFF
--- a/Source/Core/Core/IOS/FS/FileSystemProxy.cpp
+++ b/Source/Core/Core/IOS/FS/FileSystemProxy.cpp
@@ -28,12 +28,60 @@ static IPCReply GetFSReply(s32 return_value, u64 extra_tb_ticks = 0)
   return IPCReply{return_value, (2700 + extra_tb_ticks) * SystemTimers::TIMER_RATIO};
 }
 
-/// Amount of TB ticks required for a superblock write to complete.
-constexpr u64 SUPERBLOCK_WRITE_TICKS = 3370000;
-/// Amount of TB ticks required to write a cluster (for a file).
-constexpr u64 CLUSTER_WRITE_TICKS = 300000;
-/// Amount of TB ticks required to read a cluster (for a file).
-constexpr u64 CLUSTER_READ_TICKS = 115000;
+/// Duration of a superblock write (in timebase ticks).
+constexpr u64 GetSuperblockWriteTbTicks(int ios_version)
+{
+  if (ios_version == 28 || ios_version == 80)
+    return 3350000;
+
+  if (ios_version < 28)
+    return 4100000;
+
+  return 3170000;
+}
+
+/// Duration of a file cluster write (in timebase ticks).
+constexpr u64 GetClusterWriteTbTicks(int ios_version)
+{
+  // Based on the time it takes to write two clusters (0x8000 bytes), divided by two.
+  if (ios_version < 28)
+    return 370000;
+
+  return 300000;
+}
+
+/// Duration of a file cluster read (in timebase ticks).
+constexpr u64 GetClusterReadTbTicks(int ios_version)
+{
+  if (ios_version == 28 || ios_version == 80)
+    return 125000;
+
+  if (ios_version < 28)
+    return 165000;
+
+  return 115000;
+}
+
+constexpr u64 GetFSModuleMemcpyTbTicks(int ios_version, u64 copy_size)
+{
+  // FS handles cached read/writes by doing a simple memcpy on an internal buffer.
+  // The following equations were obtained by doing cached reads with various read sizes
+  // on real hardware and performing a linear regression. R^2 was close to 1 (0.9999),
+  // which is not entirely surprising.
+
+  // Older versions of IOS have a simplistic implementation of memcpy that does not optimize
+  // large copies by copying 16 bytes or 32 bytes per chunk.
+  if (ios_version < 28)
+    return 1.0 * copy_size + 3.0;
+
+  return 0.636 * copy_size + 150.0;
+}
+
+constexpr u64 GetFreeClusterCheckTbTicks()
+{
+  return 1000;
+}
+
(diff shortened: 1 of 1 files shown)


Source/Core/Core/IOS/FS/FileSystemProxy.cpp
```diff
@@ -211,7 +254,9 @@ u64 FSDevice::EstimateTicksForReadWrite(const Handle& handle, const ReadWriteReq
       const u32 start = offset - m_cache_chain_index * CLUSTER_DATA_SIZE;
       copy_length = std::min<u32>(CLUSTER_DATA_SIZE - start, count);
       // FS essentially just does a memcpy from/to an internal buffer.
-      ticks += 0.63f * copy_length + (request.command == IPC_CMD_WRITE ? 1000 : 0);
+      ticks += GetFSModuleMemcpyTbTicks(m_ios.GetVersion(), copy_length);
+      if (is_write)
+        ticks += 1000;
```
BhaaLseN, line 259: Would it be the time to give this number a name? Guessing its some sort of cache flushing since its only on write.

leoetlino: Done, cached writes seem to take slightly longer because of the free cluster check

Source/Core/Core/IOS/FS/FileSystemProxy.cpp
```diff
@@ -29,11 +29,59 @@ static IPCReply GetFSReply(s32 return_value, u64 extra_tb_ticks = 0)
...
+  if (ios_version == 28 || ios_version == 80)
+    return 3350000;
+
+  if (ios_version < 28)
+    return 4100000;
+
+  return 3170000;
+}
+
 /// Amount of TB ticks required to write a cluster (for a file).
```
iwubcode, line 43: Just curious, what is "TB"?

JMC47: I'm guessing timebase?

leoetlino: Yep, timebase. I reworded the comments to make that clearer.

iwubcode (approved):

Source/Core/Core/IOS/FS/FileSystemProxy.cpp
```diff
@@ -28,12 +28,60 @@ static IPCReply GetFSReply(s32 return_value, u64 extra_tb_ticks = 0)
...
+  if (ios_version < 28)
+    return 370000;
+
+  return 300000;
+}
+
+/// Duration of a file cluster read (in timebase ticks).
+constexpr u64 GetClusterReadTbTicks(int ios_version)
+{
+  if (ios_version == 28 || ios_version == 80)
```
BhaaLseN, line 56: It seems weird that IOS 80 is an outlier here. Do we know why?

BhaaLseN (approved):

Source/Core/Core/IOS/FS/FileSystemProxy.cpp
leoetlino: It's not clear to me why IOS80 is slower, but it appears to use a different branch of the FS module:

```
IOS9: $IOSVersion:  FS: 12/23/08 17:11:51 64M $
IOS13: $IOSVersion:  FS: 12/23/08 17:51:08 64M $
IOS21 (same)
IOS22 (same)
IOS28: $IOSVersion:  FS: 12/24/08 12:57:39 64M $

IOS36: $IOSVersion:  FFS: 12/23/08 17:26:21 64M $

IOS55: $IOSVersion:  FFSP: 12/24/08 13:48:17 64M $
IOS56: $IOSVersion:  FFSP: 11/24/08 15:36:10 64M $
IOS59 (same)
IOS62 (same)
IOS59 (Wii U): $IOSVersion:  FFSP: 04/03/12 12:47:01 64M $

IOS80: $IOSVersion:  FSP: 11/24/08 15:36:10 64M $
```

Maybe FSP is closer to FS than to FFSP? I haven't looked at 80 so I don't know how they differ.

flagrama: IOS9 was used by the System Menu 1.0, and IOS80 is used by System Menu 4.3. Other IOSes used by the System Menu are:
* IOS11 for 2.0 and 2.1
* IOS20 for 2.2
* IOS30 for 3.0 - 3.3
* IOS50 for 3.4
* IOS60 for 4.0 and 4.1
* IOS70 for 4.2

I'm not sure if you can check those for the relevant info the same way you did for IOS9 or IOS80, but maybe it's possible that they needed to keep certain timings the same for the System Menu or something?

I'm pretty sure these aren't used by any third-party software because they are replaced with non-functioning versions on each subsequent update so would cause issues if third-party software tried to use them. I'm not sure if older versions of Nintendo's own Wii channels used them as well though.

leoetlino: Oh yeah, it sounds plausible that other System Menu IOS versions would have the same timing characteristics. Because those IOSes are stubbed on an up-to-date console, they weren't tested as part of the timing tests, though.

I don't think the timing differences were introduced on purpose for the system menu, it's probably just because of the system menu IOSes all coming from one branch.

PR #9516, opened 2021-02-15
InputCommon: move some dynamic input texture logic and add specification option

iwubcode:
This will come across as a big change but really I'm just moving some code around and adding a "specification" option in json, so that I can add a new set of features in a future PR.

### DIFF
--- a/Source/Core/InputCommon/DynamicInputTextureManager.h
+++ b/Source/Core/InputCommon/DynamicInputTextureManager.h
@@ -11,7 +11,10 @@
 
 namespace InputCommon
 {
-class DynamicInputTextureConfiguration;
+namespace DynamicInputTextures
+{
+class Configuration;
+}
 class DynamicInputTextureManager
 {
 public:
@@ -21,7 +24,7 @@ public:
   void GenerateTextures(const IniFile::Section* sec, const std::string& controller_name);
 
 private:
-  std::vector<DynamicInputTextureConfiguration> m_configuration;
+  std::vector<DynamicInputTextures::Configuration> m_configuration;
   std::string m_config_type;
 };
 }  // namespace InputCommon
--- a/Source/Core/InputCommon/DynamicInputTextures/DITConfiguration.cpp
+++ b/Source/Core/InputCommon/DynamicInputTextures/DITConfiguration.cpp
@@ -2,7 +2,7 @@
 // Licensed under GPLv2+
 // Refer to the license.txt file included.
 
-#include "InputCommon/DynamicInputTextureConfiguration.h"
+#include "InputCommon/DynamicInputTextures/DITConfiguration.h"
 
 #include <optional>
 #include <sstream>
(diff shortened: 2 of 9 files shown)


iwubcode:
@leoetlino - thanks for your comment on IRC.  I would have never thought of the name clash!  I updated the names, it's a little redundant I guess but I wanted to keep the namespace.

Source/Core/InputCommon/DynamicInputTextureManager.h
```diff
@@ -11,7 +11,10 @@
 
 namespace InputCommon
 {
-class DynamicInputTextureConfiguration;
+namespace DynamicInputTextures
+{
+class DITConfiguration;
```
leoetlino, line 16: Maybe just Configuration? The namespace identifier is already very long and DIT is redundant. The prefix is useful to avoid file name conflicts but it's less useful in the source code ^^

iwubcode:
Thanks @leoetlino .  I agree with the request, just didn't know if the filename and class name needed to match.  Anyway, fixed up all the classes.

Source/Core/InputCommon/DynamicInputTextures/DITConfiguration.cpp
```diff
@@ -54,197 +53,25 @@ Configuration::Configuration(const std::string& json_file)
...
 
-  const std::string& game_id = SConfig::GetInstance().GetGameID();
-  std::string generated_folder_name = fmt::format("{}_Generated", game_id);
-  if (generated_folder_name_json.is<std::string>())
+  const picojson::value& specification_json = root.get("specification");
+  u8 specification = 1;
+  if (specification_json.is<double>())
   {
-    generated_folder_name = generated_folder_name_json.get<std::string>();
+    specification = static_cast<u8>(specification_json.get<double>());
```
leoetlino, line 62: This is undefined behavior if the JSON value happens to be non-representable as a `u8`. See https://github.com/dolphin-emu/dolphin/pull/9480 (MathUtil::SaturatingCast) for an example of the clamping you need to do...

iwubcode: Good catch.  I was shocked when `picojson` didn't have integer support.

Guess I'll check if it's between the range of 0 and `std::numeric_limts<u8>::max()` and if not mark it invalid.

Source/Core/InputCommon/DynamicInputTextures/DITSpecification.cpp
```diff
@@ -0,0 +1,202 @@
...
+// Licensed under GPLv2+
+// Refer to the license.txt file included.
+
+#include "InputCommon/DynamicInputTextures/DITSpecification.h"
+
+#include <fmt/format.h>
+
+#include "Common/FileUtil.h"
+#include "Common/IOFile.h"
+
```
leoetlino, line 11: nitpick: there's a weird empty line here that should be removed

leoetlino (approved):

PR #9518, opened 2021-02-15
JitArm64: Workaround for GCC ICE

JosJuice:
https://github.com/dolphin-emu/dolphin/pull/9412#issuecomment-779329306

Untested, since I don't have a GCC install.


dovart:
@JosJuice it worked with GCC 8.3!

MerryMage (approved):

PR #9519, opened 2021-02-16
IOS: Add missing IPC return codes

sepalani:
This PR adds missing IPC return codes. While I managed to find some undocumented ones, I quickly came to the realisation that they seem to perfectly match [WiiU's IOSU ones](https://wiiubrew.org/wiki/IOSU#Error_codes). I managed to verified some of them and seem to make sense. I'll leave this PR as a draft until there are enough evidences showing that these codes are correct.

Ready to be reviewed.


Source/Core/Core/IOS/Network/NCD/Manage.cpp
```diff
@@ -72,11 +72,11 @@ std::optional<IPCReply> NetNCDManageDevice::IOCtlV(const IOCtlVRequest& request)
     if (m_ipc_fd == request_handle)
     {
       m_ipc_fd = 0;
-      common_result = 0;
+      common_result = IPC_SUCCESS;
     }
     else
     {
-      common_result = -3;
+      common_result = IPC_EINTR;
```
leoetlino, line 79: I'm not sure IPC_EINTR makes sense here, since there is no interrupted system call. I wonder if this is actually a typo'd -4 (invalid) or if -3 meant something different in IOS...

sepalani: Regarding -3 and WFSI, it's used a lot when checking the max path length. It seems returned when in an invalid case or unimplemented (yet?).

I wonder if there is/was a retail/debug version of IOS that can have interrupted syscalls with some kind of signals which could explain this return value. Maybe they were following some kind of RFC/specification that requires this return value in these cases.

PR #9520, opened 2021-02-16
Fix point sampling

Miksel12:
Second attempt at fixing point sampling. 
Previous attempt (#9473 ) didn't use the UID correctly.
This doesn't implement the fix in the UberShader yet.

### DIFF
--- a/Source/Core/VideoCommon/GXPipelineTypes.h
+++ b/Source/Core/VideoCommon/GXPipelineTypes.h
@@ -20,7 +20,7 @@ namespace VideoCommon
 // As pipelines encompass both shader UIDs and render states, changes to either of these should
 // also increment the pipeline UID version. Incrementing the UID version will cause all UID
 // caches to be invalidated.
-constexpr u32 GX_PIPELINE_UID_VERSION = 2;  // Last changed in PR 9122
+constexpr u32 GX_PIPELINE_UID_VERSION = 3;  // Last changed in PR 9520
 
 struct GXPipelineUid
 {
--- a/Source/Core/VideoCommon/PixelShaderGen.cpp
+++ b/Source/Core/VideoCommon/PixelShaderGen.cpp
@@ -16,6 +16,7 @@
 #include "VideoCommon/LightingShaderGen.h"
 #include "VideoCommon/NativeVertexFormat.h"
 #include "VideoCommon/RenderState.h"
+#include "VideoCommon/SamplerCommon.h"
 #include "VideoCommon/VertexLoaderManager.h"
 #include "VideoCommon/VideoCommon.h"
 #include "VideoCommon/VideoConfig.h"
@@ -273,7 +274,10 @@ PixelShaderUid GetPixelShaderUid()
       uid_data->stagehash[n].tevksel_swap2c = bpmem.tevksel[i * 2].swap2;
       uid_data->stagehash[n].tevksel_swap1d = bpmem.tevksel[i * 2 + 1].swap1;
       uid_data->stagehash[n].tevksel_swap2d = bpmem.tevksel[i * 2 + 1].swap2;
-      uid_data->stagehash[n].tevorders_texmap = bpmem.tevorders[n / 2].getTexMap(n & 1);
+      const int texmap = bpmem.tevorders[n / 2].getTexMap(n & 1);
+      uid_data->stagehash[n].tevorders_texmap = texmap;
+      uid_data->stagehash[n].tevorders_pointsample = SamplerCommon::IsBpTexMode0PointFiltering(
+          bpmem.tex[(texmap >> 2) & 1].texMode0[texmap & 3]);
     }
 
     if (cc.a == TEVCOLORARG_KONST || cc.b == TEVCOLORARG_KONST || cc.c == TEVCOLORARG_KONST ||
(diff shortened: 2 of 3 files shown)


Miksel12:
I did some testing and it seems that the results are the same as previous attempt, so VP6 video's fixed and Just Dance 2 still has some artifacts but less than master. Software renderer scrambles Just Dance 2 even more so that shows that Just Dance 2 has a different issue, though that was already concluded in the previous PR. 
 I guess this also shows that the 0.5 offset doesn't have much impact result wise. Meaning that copying the software renderer is probably preferable.

JMC47:
@dolphin-emu-bot rebuild

shuffle2:
Fwiw If this is rebased on master it will be faster on the build but

Miksel12:
Rebased on master

mbc07:
Hmm, wasn't this supposed to fix EA VP6 too? FifoCI shows no rendering differences...

Miksel12:
ogl-lin-radeon is the test enviroment which shows the issue, but that bot didn't run.

JMC47:
This one is actually an interesting change you can see

https://fifo.ci/compare/5908638-5906640/

Do we know if this is accurate or not?

Miksel12:
It does seem to diverge more from the software render which is weird but I don't know what is correct.

iwubcode:
JMC's comment on IRC:

>I wonder if it's possible that the pixel offset is different between bilinear and point filtered textures

(based on the bisect comment on https://bugs.dolphin-emu.org/issues/9464 )

JMC47:
it should have been *sampling* not filtering.  I misspoke, though the point still applies.

Miksel12:
I'm pretty sure the software renderer uses the same offset for both, I'll be looking into it more this weekend and will try to implement this fix in the ubershader.

Miksel12:
Correction, in the software renderer there is a -0.5 offset for linear samples and no offset for nearest samples. In master there is no offset for linear and nearest sampling and this PR has a +0.5 offset for nearest sampling, which seems to fix all point sampling issues. Implementing the same behaviour as software makes things worse so it seems there is 0.5 coordinate offset difference between the software and hardware backends which is corrected in this PR.

I tried to implement the fix in the Ubershader but that seems to be significantly harder (for me) than I thought. I could put more time into it but so long we don't know if this is correct way to handle it, I think that would be a bit of a waste of time.

Pokechu22:
With regards to whether software's -0.5 or the others' +0.5 is correct, I did a lot of screenshots on hardware for #9651 and all of them are *slightly* off.  `F00_0NN.png` from [addcoordstest](https://gist.github.com/Pokechu22/8396169343eab4f759cbaa279fb66d78) works well for this (it's just a square without any of the quirks being tested).  I don't know what sampling mode is in use (I didn't explicitly set any), though.  This PR doesn't affect the results for that at all.

<details><summary>Result on hardware</summary>

HW
</details>
<details><summary>Result on software renderer</summary>

SW
</details>
<details><summary>Comparison for hardware and software renderer</summary>

SW_compare
</details>
<details><summary>Result on OGL</summary>

OGL
</details>
<details><summary>Comparison for hardware and OGL</summary>

OGL_compare
</details>

As for ubershaders, the equivalent to `SampleTexture` is [here](https://github.com/dolphin-emu/dolphin/blob/0f563ffd59857424fd6281e4630d5564c9a7164e/Source/Core/VideoCommon/UberShaderPixel.cpp#L226-L254), and these are its uses:

https://github.com/dolphin-emu/dolphin/blob/0f563ffd59857424fd6281e4630d5564c9a7164e/Source/Core/VideoCommon/UberShaderPixel.cpp#L305-L308
https://github.com/dolphin-emu/dolphin/blob/0f563ffd59857424fd6281e4630d5564c9a7164e/Source/Core/VideoCommon/UberShaderPixel.cpp#L901-L902

There isn't a way to use `IsBpTexMode0PointFiltering` directly, and I don't think the needed information is currently provided.  You'd need to add it in [`PixelShaderConstants.pack2`](https://github.com/dolphin-emu/dolphin/blob/0f563ffd59857424fd6281e4630d5564c9a7164e/Source/Core/VideoCommon/ConstantManager.h#L47) (probably as `z`), and then set it in [PixelShaderManager](https://github.com/dolphin-emu/dolphin/blob/0f563ffd59857424fd6281e4630d5564c9a7164e/Source/Core/VideoCommon/PixelShaderManager.cpp#L175); you could use it as a boolean for whether `IsBpTexMode0PointFiltering` returns true, and then declare a thing to access it like [this](https://github.com/dolphin-emu/dolphin/blob/14959a1087b05500c5704b59e7d694d941d4a5cf/Source/Core/VideoCommon/PixelShaderGen.cpp#L438-L442).  This part is more awkward and I can see why you couldn't figure it out (I'm still not entirely sure what the best way of doing it is).

Miksel12:
Thanks for the information! Seems more doable now, I'll probably take another look at it this week.

Miksel12:
Looks like I have hit a dead end again. I did some research on graphics api's and I'm pretty sure this is not the right way of handling it. I did look into integerize texture lookups like endrift mentioned in https://bugs.dolphin-emu.org/issues/7193 and I got pretty far with texelFetch (https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/texelFetch.xhtml) but that needs a lod which I have no idea how to provide, hardcoding it showed that it correctly point samples though. There are also probably inaccuracies in other parts of the pixelshader which was discussed in #9473. 
I would love to see this issue fixed but this isn't the right solution. If no one comes with a fix before a potential 6.0 release, then an epsilon hack would be acceptable I think. Testing showed that the games that weren't fixed by an epsilon (#9407) are also broken on software (https://fifo.ci/dff/jd2-fmv/).

PR #8310, opened 2019-08-13
Core/PowerPC/MMU: Add address translation override flags to the Host functions.

AdmiralCurtiss:
Idea from phire to fix issue of Cheat Search pausing the CPU while it's in an exception, and thus has MMU turned off, which then causes all memory reads at known addresses to fail: Just allow overriding the MMU state to be always on.

Figured I might as well make this a separate PR to review.

PR #6913, opened 2018-05-20
MMU: always enable address translation for Host* memory functions

Felk:
Beware: I don't understand this code _too_ well, so someone who does should probably take a good look and verify that my thought here even makes sense.

The Host* memory functions in the MMU read and write memory by logical address, which get translated to physical memory. However, if the MSR.DR bit is set, it won't get translated. This causes whatever calls the Host* memory functions (memory view, cheats, etc.) to sometimes poke at physical memory.

While honoring the DR bit during normal emulation makes sense, I don't think this was intended for the Host* memory functions. This change makes those functions _always_ translate addresses. I poked around in the memory view and everything seems to work fine.

PR #9522, opened 2021-02-18
Core/MMU: Add HostTryRead/HostTryWrite functions to allow specifying the desired address space, and return whether it succeeded.

AdmiralCurtiss:
Alright, let's give this another shot. I'll summarize here, but if you want the previous attempts, see #6913, #8310, #8682, and sorta #9361.

Basically, our Cheat Search has a big problem where your search results may break (usually completely disappear, but sometimes other weird stuff if you do comparisons with 0) because the code calling the `HostRead_*()` functions cannot differentiate between them returning 0 because the actual value at that memory address is zero, or returning 0 because the address translation failed (eg. because the CPU is currently in an exception state).

To help with this, I've added new `Try` variants of the HostRead and HostWrite functions that allow specifying the expected address space of the given address and return whether the read or write succeeded or not.

### DIFF
--- a/Source/Core/Core/PowerPC/MMU.cpp
+++ b/Source/Core/Core/PowerPC/MMU.cpp
@@ -3,6 +3,7 @@
 
 #include "Core/PowerPC/MMU.h"
 
+#include <cassert>
 #include <cstddef>
 #include <cstring>
 #include <string>
@@ -427,6 +428,36 @@ u32 HostRead_Instruction(const u32 address)
   return ReadFromHardware<XCheckTLBFlag::OpcodeNoException, u32>(address);
 }
 
+TryReadResult<u32> HostTryReadInstruction(const u32 address, RequestedAddressSpace space)
+{
+  if (!HostIsInstructionRAMAddress(address, space))
+    return TryReadResult<u32>();
+
+  switch (space)
+  {
+  case RequestedAddressSpace::Effective:
+  {
+    const u32 value = ReadFromHardware<XCheckTLBFlag::OpcodeNoException, u32>(address);
+    return TryReadResult<u32>(!!MSR.DR, value);
+  }
+  case RequestedAddressSpace::Physical:
+  {
+    const u32 value = ReadFromHardware<XCheckTLBFlag::OpcodeNoException, u32, true>(address);
+    return TryReadResult<u32>(false, value);
+  }
+  case RequestedAddressSpace::Virtual:
+  {
+    if (!MSR.DR)
+      return TryReadResult<u32>();
+    const u32 value = ReadFromHardware<XCheckTLBFlag::OpcodeNoException, u32>(address);
+    return TryReadResult<u32>(true, value);
+  }
+  }
+
+  assert(0);
+  return TryReadResult<u32>();
+}
+
 static void Memcheck(u32 address, u32 var, bool write, size_t size)
 {
   if (PowerPC::memchecks.HasAny())
@@ -497,6 +528,73 @@ float Read_F32(const u32 address)
   const u32 integral = Read_U32(address);
 
   return Common::BitCast<float>(integral);
+}
+
+template <typename T>
+static TryReadResult<T> HostTryReadUX(const u32 address, RequestedAddressSpace space)
+{
+  if (!HostIsRAMAddress(address, space))
+    return TryReadResult<T>();
+
+  switch (space)
+  {
+  case RequestedAddressSpace::Effective:
+  {
+    T value = ReadFromHardware<XCheckTLBFlag::NoException, T>(address);
+    return TryReadResult<T>(!!MSR.DR, std::move(value));
+  }
+  case RequestedAddressSpace::Physical:
+  {
+    T value = ReadFromHardware<XCheckTLBFlag::NoException, T, true>(address);
+    return TryReadResult<T>(false, std::move(value));
+  }
+  case RequestedAddressSpace::Virtual:
+  {
+    if (!MSR.DR)
+      return TryReadResult<T>();
+    T value = ReadFromHardware<XCheckTLBFlag::NoException, T>(address);
+    return TryReadResult<T>(true, std::move(value));
+  }
+  }
+
+  assert(0);
+  return TryReadResult<T>();
+}
+
+TryReadResult<u8> HostTryReadU8(u32 address, RequestedAddressSpace space)
+{
+  return HostTryReadUX<u8>(address, space);
+}
+
+TryReadResult<u16> HostTryReadU16(u32 address, RequestedAddressSpace space)
+{
+  return HostTryReadUX<u16>(address, space);
+}
+
+TryReadResult<u32> HostTryReadU32(u32 address, RequestedAddressSpace space)
+{
+  return HostTryReadUX<u32>(address, space);
+}
+
+TryReadResult<u64> HostTryReadU64(u32 address, RequestedAddressSpace space)
+{
+  return HostTryReadUX<u64>(address, space);
+}
+
+TryReadResult<float> HostTryReadF32(u32 address, RequestedAddressSpace space)
+{
+  const auto result = HostTryReadUX<u32>(address, space);
+  if (!result)
+    return TryReadResult<float>();
+  return TryReadResult<float>(result.translated, Common::BitCast<float>(result.value));
+}
+
+TryReadResult<double> HostTryReadF64(u32 address, RequestedAddressSpace space)
+{
+  const auto result = HostTryReadUX<u64>(address, space);
+  if (!result)
+    return TryReadResult<double>();
+  return TryReadResult<double>(result.translated, Common::BitCast<double>(result.value));
 }
 
 u32 Read_U8_ZX(const u32 address)
@@ -639,6 +737,27 @@ std::string HostGetString(u32 address, size_t size)
   return s;
 }
 
+TryReadResult<std::string> HostTryReadString(u32 address, size_t size, RequestedAddressSpace space)
+{
+  auto c = HostTryReadU8(address, space);
+  if (!c)
+    return TryReadResult<std::string>();
+  if (c.value == 0)
+    return TryReadResult<std::string>(c.translated, "");
+
+  std::string s;
+  s += static_cast<char>(c.value);
+  while (size == 0 || s.length() < size)
+  {
+    ++address;
+    const auto res = HostTryReadU8(address, space);
+    if (!res || res.value == 0)
+      break;
+    s += static_cast<char>(res.value);
+  }
+  return TryReadResult<std::string>(c.translated, std::move(s));
+}
+
 bool IsOptimizableRAMAddress(const u32 address)
 {
   if (PowerPC::memchecks.HasAny())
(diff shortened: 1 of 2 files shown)


Source/Core/Core/ActionReplay.cpp
```diff
@@ -640,14 +642,15 @@ static bool ZeroCode_MemoryCopy(const u32 val_last, const ARAddr& addr, const u3
...
+      const u32 ptr_dest = PowerPC::HostRead_U32(addr_dest).value_or(0);
       LogInfo("Resolved Dest Address to: {:08x}", ptr_dest);
-      const u32 ptr_src = PowerPC::HostRead_U32(addr_src);
+      const u32 ptr_src = PowerPC::HostRead_U32(addr_src).value_or(0);
       LogInfo("Resolved Src Address to: {:08x}", ptr_src);
       for (int i = 0; i < num_bytes; ++i)
       {
-        PowerPC::HostWrite_U8(PowerPC::HostRead_U8(ptr_src + i), ptr_dest + i);
-        LogInfo("Wrote {:08x} to address {:08x}", PowerPC::HostRead_U8(ptr_src + i), ptr_dest + i);
+        PowerPC::HostWrite_U8(PowerPC::HostRead_U8(ptr_src + i).value_or(0), ptr_dest + i);
```
BhaaLseN, line 651: Thoughts on storing the result of `PowerPC::HostRead_U8(ptr_src + i).value_or(0)` in a local for use with both the `HostWrite_U8` and the `LogInfo` that follows?
Or is the overhead of calling it twice neglible with the changes in `HostRead_UX`?

Source/Core/DolphinQt/Debugger/WatchWidget.cpp
```diff
@@ -170,11 +170,11 @@ void WatchWidget::Update()
     {
       if (PowerPC::HostIsRAMAddress(entry.address))
       {
-        hex->setText(QStringLiteral("%1").arg(PowerPC::HostRead_U32(entry.address), 8, 16,
-                                              QLatin1Char('0')));
-        decimal->setText(QString::number(PowerPC::HostRead_U32(entry.address)));
+        hex->setText(QStringLiteral("%1").arg(PowerPC::HostRead_U32(entry.address).value_or(0), 8,
```
BhaaLseN, line 173: Same down here, where `HostRead_U32` is called twice with the same parameters etc.

BhaaLseN (commented):
Just checking, this looks like _every_ place so far was just appended `.value_or(N)` and "good to go" - is this correct, or should there be places that check for `std::nullopt` returns instead (aside from `ReadString` for obvious reasons)?

AdmiralCurtiss:
Correct, that's all I did so far. There are definitely places here that should now check the return value and branch on it, but I didn't want to waste time thinking about it before I had an 'okay' that this is a good approach to the general issue.

AdmiralCurtiss:
Speaking of the ReadString(), I was wondering how to deal with that one. For consistency it would make sense to also return a nullopt (instead of an empty string) when the address is invalid, but at the same time there might be possible (unlikely, to be fair) that there is a string that goes right up to a memory address boundary that you might want returned. Maybe return nullopt if the initial address is invalid, but return the built up string if we encounter an invalid address while advancing?

BhaaLseN:
No idea, I stared at it for a while, and I'm not entirely sure if it makes much ofa difference.
For consistency, it should probably do that though - that is, return `std::nullopt` if the initial address is already invalid. Crossing over is probably fine when it simply ends there.

Source/Core/Core/Debugger/OSThread.cpp
```diff
@@ -192,9 +192,10 @@ std::string OSThreadView::GetSpecific() const
 
   for (u32 addr : m_thread.specific)
   {
-    if (!PowerPC::HostIsRAMAddress(addr))
+    const auto name = PowerPC::HostGetString(addr);
+    if (!name)
       break;
-    specific += fmt::format("{:08x} \"{}\"\n", addr, PowerPC::HostGetString(addr));
+    specific += fmt::format("{:08x} \"{}\"\n", addr, *name);
```
iwubcode, line 198: Not related to this PR directly but this function could use a `fmt::memory_buffer` to avoid some of the overhead of string concatenation.

Source/Core/Core/HLE/HLE_OS.cpp
```diff
@@ -225,7 +225,7 @@ std::string GetStringVA(u32 str_reg, ParameterType parameter_type)
       {
       case 's':
         result += StringFromFormat(ArgumentBuffer.c_str(),
-                                   PowerPC::HostGetString(ap->GetArgT<u32>()).c_str());
+                                   PowerPC::HostGetString(ap->GetArgT<u32>()).value_or("").c_str());
```
iwubcode, line 228: While you're here, isn't this `c_str()` unnecessary in `StringFromFormat`?

AdmiralCurtiss: Seems to be required:
`warning C4840: non-portable use of class 'std::basic_string<char,std::char_traits<char>,std::allocator<char>>' as an argument to a variadic function`

iwubcode: Ah sorry.  Thought this was one of the functions converted to fmt.  Guess not

iwubcode: Oh.  Looking back through previous reviews, that must be a spot that hasn't been converted yet.  `StringFromFormat` was getting switched to `fmt::format` slowly.  I was remembering wrong and thought  that `StringFromFormat` itself was getting converted to use `fmt`.

Sorry about that.

sepalani:
@AdmiralCurtiss
My issue with this approach is that you're repeating what you're trying to avoid:
>  functions cannot differentiate between them returning 0 because the actual value at that memory address is zero, or returning 0 because the address translation failed (eg. because the CPU is currently in an exception state).

In some of the debugging stuff, like HLE or other widgets, the address (or range) is checked (using `HostIsRAMAddress`) prior to read values from RAM. With this change, each Read/Write call will check redundantly the RAM address which isn't ideal. Using `value_or` also hides PanicAlert messages in places where it might have been intended.

What do you think about adding to the interface your functions renamed as `Host_TryReadUX` (or similar) which behaves exactly like your current implementation (using RAM address check, address space and optional) and using it where you think it's relevant? For instance, we already have a similar approach with the `TryReadInstruction` function.

AdmiralCurtiss:
Mhm, yeah I understand what you mean. That probably is a better idea indeed, let's see...

AdmiralCurtiss:
So, like this?

sepalani:
Yes like that. Is the `assert(0)` needed assuming it's ok for the "Try" functions to fail?

AdmiralCurtiss:
It's for ensuring no one passes a nonsense enum value. I could go either way though.

Source/Core/Core/PowerPC/MMU.h
```diff
@@ -15,14 +15,55 @@ namespace PowerPC
...
+
+template <typename T>
+struct TryReadResult
+{
+  bool success;
+  bool translated;
+  T value;
+
+  TryReadResult() : success(false) {}
+  TryReadResult(bool translated_, T value_) : success(true), translated(translated_), value(value_)
```
iwubcode, line 44: I believe you want `value(std::forward(value_))`

Also your function definition should be `T&&`.  Otherwise your string moves are actually a copy

Source/Core/Core/PowerPC/MMU.h
```diff
@@ -15,14 +15,55 @@ namespace PowerPC
...
+  {
+  }
+  operator bool() const { return success; }
+};
+
+TryReadResult<u8> HostTryReadU8(u32 address,
```
iwubcode, line 50: AFAICT all these functions that use `TryReadResult` are only used in the cpp, could they be moved to file local static (as well as `TryReadResult` itself) or are you planning on using it elsewhere later?

AdmiralCurtiss: `TryReadResult` is going to be used from elsewhere in the future, yeah.

Source/Core/Core/PowerPC/MMU.h
AdmiralCurtiss: This doesn't work, it prevents use of constant values as parameters, as far as I can tell. I just added an `std::move()` for the value, should be good enough in practice...

iwubcode: > This doesn't work, it prevents use of constant values as parameters, as far as I can tell.

It should, it's perfect forwarding so const parameters would just get passed in as if you had a normal constructor...

> I just added an std::move() for the value, should be good enough in practice...

NBD but I believe that will still incur a copy in some cases

AdmiralCurtiss:
~~~
  TryReadResult(bool translated_, T&& value_)
      : success(true), translated(translated_), value(std::forward(value_))
  {
  }
~~~

~~~
    const u32 value = ReadFromHardware<XCheckTLBFlag::OpcodeNoException, u32, true>(address);
    return TryReadResult<u32>(false, value);
~~~

~~~
2>Source\Core\Core\PowerPC\MMU.cpp(444,43): error C2440: '<function-style-cast>': cannot convert from 'initializer list' to 'PowerPC::TryReadResult<u32>'
2>Source\Core\Core\PowerPC\MMU.cpp(444,43): error C2440:     return TryReadResult<u32>(false, value);
2>Source\Core\Core\PowerPC\MMU.cpp(444,43): error C2440:                                           ^
2>Source\Core\Core\PowerPC\MMU.cpp(444,43): message : No constructor could take the source type, or constructor overload resolution was ambiguous
2>Source\Core\Core\PowerPC\MMU.cpp(444,43): message :     return TryReadResult<u32>(false, value);
2>Source\Core\Core\PowerPC\MMU.cpp(444,43): message :                                           ^
~~~

iwubcode: Huh, I guess I always used a templated method in this scenario (where `const` would be forwarded), not a templated class, so I didn't notice.

Here's a variation of your current solution and should handle both the move and const variants with no copies:

```
  TryReadResult(bool translated_, T&& value_)
      : success(true), translated(translated_), value(std::move(value_))
  {
  }
  TryReadResult(bool translated_, const T& value_)
      : success(true), translated(translated_), value(value_)
  {
  }
```

Source/Core/Core/PowerPC/MMU.cpp
```diff
@@ -497,6 +528,73 @@ float Read_F32(const u32 address)
...
+static TryReadResult<T> HostTryReadUX(const u32 address, RequestedAddressSpace space)
+{
+  if (!HostIsRAMAddress(address, space))
+    return TryReadResult<T>();
+
+  switch (space)
+  {
+  case RequestedAddressSpace::Effective:
+  {
+    const T value = ReadFromHardware<XCheckTLBFlag::NoException, T>(address);
```
iwubcode, line 541: In order to handle the string cases more efficiently, since `const` can't be moved, I'd suggest:

```
T value = ReadFromHardware<XCheckTLBFlag::NoException, T>(address);
return TryReadResult<T>(!!MSR.DR, std::move(value));
```

(and the other variations below)

Source/Core/Core/PowerPC/MMU.h
AdmiralCurtiss: Did you mean std::move() in that first constructor? Or should that be std::forward()?

Source/Core/Core/PowerPC/MMU.h
```diff
@@ -15,30 +15,115 @@ namespace PowerPC
...
+  TryReadResult() : success(false) {}
+  TryReadResult(bool translated_, T&& value_)
+      : success(true), translated(translated_), value(std::move(value_))
+  {
+  }
+  TryReadResult(bool translated_, const T& value_)
+      : success(true), translated(translated_), value(value_)
+  {
+  }
+  operator bool() const { return success; }
```
leoetlino, line 57:
```suggestion
  explicit operator bool() const { return success; }
```

Source/Core/Core/PowerPC/MMU.h
```diff
@@ -15,30 +15,115 @@ namespace PowerPC
...
-std::string HostGetString(u32 address, size_t size = 0);
+struct TryWriteResult
+{
+  bool success;     // whether the write succeeded; if false, the other fields should not be touched
+  bool translated;  // whether the address had to be translated (given address was treated as
+                    // virtual) or not (given address was treated as physical)
+
+  TryWriteResult() : success(false) {}
+  TryWriteResult(bool translated_) : success(true), translated(translated_) {}
+  operator bool() const { return success; }
```
leoetlino, line 100:
```suggestion
  explicit operator bool() const { return success; }
```

Source/Core/Core/PowerPC/MMU.h
```diff
@@ -15,30 +15,115 @@ namespace PowerPC
...
 u64 HostRead_U64(u32 address);
 float HostRead_F32(u32 address);
 double HostRead_F64(u32 address);
 u32 HostRead_Instruction(u32 address);
+std::string HostGetString(u32 address, size_t size = 0);
+
+template <typename T>
+struct TryReadResult
+{
+  bool success;     // whether the read succeeded; if false, the other fields should not be touched
```
leoetlino, line 43: Could you put this comment above `success` (and same for the other members)? Otherwise, clang-format will try to align all these comments which means that any change that modifies the column width of the members will result in a very noisy diff.

leoetlino (changes requested):

Source/Core/Core/PowerPC/MMU.h
AdmiralCurtiss: Done.

sepalani:
This still looks fine to me.

I guess you can add a 3rd commit regarding your use-case (Cheat Search) for these functions.

leoetlino (approved):

PR #9524, opened 2021-02-19
Debugger: Fix issue were loaded memory breakpoints were not being triggered (break)

InusualZ:
This was caused, because we were saving the `break_on_hit` flag with the letter `p`. Then while loading the breakpoints, we read the flag with the letter `b`, resulting in the `break_on_hit` flag being always false.


sepalani:
@dolphin-emu-bot rebuild

sepalani (approved):
LGTM and matches code breakpoints behaviour.

AdmiralCurtiss:
Is... there a reason this is stored in a string instead of, like, a struct or a bitfield?

sepalani:
Probably because they are saved to and loaded from an INI file.

AdmiralCurtiss:
I feel like this should be changed to not a string with magic letters at some point, but this is free to merge in the meantime IMO.

AdmiralCurtiss (approved):

PR #9525, opened 2021-02-20
Fix IPC_OVERHEAD_TICKS Typo

flagrama:
Fixes Majora's Mask VC taking longer to load than on console (reported on https://bugs.dolphin-emu.org/issues/11346#note-11). Majora's Mask VC and Ocarina of Time VC match my console perfectly on load with this fix (ignoring the ES_Launch timings that are not yet implemented in Dolphin)

### DIFF
--- a/Source/Core/Core/IOS/IOS.h
+++ b/Source/Core/Core/IOS/IOS.h
@@ -46,7 +46,7 @@ struct IPCReply
   u64 reply_delay_ticks;
 };
 
-constexpr u64 IPC_OVERHEAD_TICKS = 2700_tbticks;
+constexpr u64 IPC_OVERHEAD_TICKS = 2700;
 
 // Used to make it more convenient for functions to return timing information
 // without having to explicitly keep track of ticks in callers.


JMC47:
@dolphin-emu-bot rebuild

BhaaLseN:
You might be on to something (ticks vs. time-base ticks), but it might not be the only place (or the correct place) to change this.

For example, this one used to be `2700 * SystemTimers::TIMER_RATIO` before the change; and if you change the `2700_tbticks` to `2700` this wont be correct anymore:
https://github.com/dolphin-emu/dolphin/blob/93f9d67d2fd8e0de89ea9c6703a53c25f213c330/Source/Core/Core/IOS/DI/DI.cpp#L112

Ping @leoetlino since this was changed in #9511.

leoetlino:
Yeah, I opened a new PR (#9526) to fix it and make it harder to mix up units

BhaaLseN:
Whoops, I guess  #9526 is the better PR for this - missed it also addresses this due to me visiting this one first.

JMC47:
Superceded by https://github.com/dolphin-emu/dolphin/pull/9526

PR #9526, opened 2021-02-20
Core: Fix time base unit mixup

leoetlino:
And add a strongly typed integer type so that making this kind of mistake is more difficult


JMC47:
Fixes an issue in Majora's Mask where it was loading far too slow.

BhaaLseN (approved):

PR #9527, opened 2021-02-20
WatchWidget: Add new/delete/clear toolbar items

sepalani:
This PR adds to the Watch widget toolbar items to create, delete and clear watches.

Ready to reviewed & merged.

### DIFF
--- a/Source/Core/DolphinQt/Debugger/WatchWidget.cpp
+++ b/Source/Core/DolphinQt/Debugger/WatchWidget.cpp
@@ -5,6 +5,7 @@
 #include "DolphinQt/Debugger/WatchWidget.h"
 
 #include <QHeaderView>
+#include <QInputDialog>
 #include <QMenu>
 #include <QTableWidget>
 #include <QToolBar>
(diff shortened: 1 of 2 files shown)


Source/Core/DolphinQt/Debugger/WatchWidget.cpp
```diff
@@ -111,6 +118,10 @@ void WatchWidget::ConnectWidgets()
 
 void WatchWidget::UpdateIcons()
 {
+  // TODO: Create a "debugger_add_watch" icon
```
iwubcode, line 121: Who would create this?  How do we keep track of this other than a code "TODO"?  Maybe an issue in the issue tracker?  I have some icons needed too..

JosJuice: @MayImilae would be the person to talk to.

iwubcode: @JosJuice - is there a way we can track this or must we block this PR until MayImilae has time to create said icons?

JosJuice: I don't think we really have a standard way to track icon requests... But I suppose the issue tracker could make sense.

JMC47:
May said she doesn't have time to work on an icon right now, so I think it is fine to merge it without one for now.

PR #9528, opened 2021-02-20
Debugger: Allow to edit breakpoints

InusualZ:
**Depends on #9524**

## This PR do 4 things:
1. Removed the current action when selecting a breakpoint
2. Allow to toggle breakpoints between on/off
3. Add a `QTableWidget::itemDoubleClicked` listener that would directly open the edit breakpoint dialog
4. Add a Context Menu for the selected breakpoint, which add the following actions
   - Enable/Disable - Toogle between on/off
   - Go to - Old behavior of selecting a breakpoint while the game is paused
   - Edit - Launch the edit dialog

To not duplicate code, I added support to the `NewBreakpointDialog` to be able to edit existing breakpoint

### Context Menu
image
### Edit Instruction Breakpoint
image
### Edit Memory Breakpoint
image

This is my first _major_ change to the codebase. Please, point out any mistake I have made or code that I should refactor in a better way.

### DIFF
--- a/Source/Core/Core/PowerPC/BreakPoints.h
+++ b/Source/Core/Core/PowerPC/BreakPoints.h
@@ -29,7 +29,7 @@ struct TMemCheck
   u32 start_address = 0;
   u32 end_address = 0;
 
-  bool is_enabled = false;
+  bool is_enabled = true;
 
   bool is_ranged = false;
 
--- a/Source/Core/Core/PowerPC/BreakPoints.h
+++ b/Source/Core/Core/PowerPC/BreakPoints.h
@@ -29,6 +29,8 @@ struct TMemCheck
   u32 start_address = 0;
   u32 end_address = 0;
 
+  bool is_enabled = false;
+
   bool is_ranged = false;
 
   bool is_break_on_read = false;
@@ -52,6 +54,7 @@ public:
   using TBreakPointsStr = std::vector<std::string>;
 
   const TBreakPoints& GetBreakPoints() const { return m_breakpoints; }
+  TBreakPoints& GetBreakPoints() { return m_breakpoints; }
   TBreakPointsStr GetStrings() const;
   void AddFromStrings(const TBreakPointsStr& bp_strings);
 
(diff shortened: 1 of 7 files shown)


sepalani:
@dolphin-emu-bot rebuild

Source/Core/Core/PowerPC/BreakPoints.h
```diff
@@ -52,6 +54,7 @@ class BreakPoints
   using TBreakPointsStr = std::vector<std::string>;
 
   const TBreakPoints& GetBreakPoints() const { return m_breakpoints; }
+  TBreakPoints& GetBreakPoints() { return m_breakpoints; }
```
sepalani, line 57: I'm pretty sure that's unwanted as the `const reference` is here to prevent changes outside of the exposed interface. It tends to avoid undefined behaviours like holding to a pointer that can lead to crash later if the structure has changed. You should change the interface to expose the function you need (especially if they are only useful for GUI) like `SetBreakPointEnabled` or along those lines.

Source/Core/Core/PowerPC/BreakPoints.h
```diff
@@ -83,6 +86,7 @@ class MemChecks
   using TMemChecksStr = std::vector<std::string>;
 
   const TMemChecks& GetMemChecks() const { return m_mem_checks; }
+  TMemChecks& GetMemChecks() { return m_mem_checks; }
```
sepalani, line 89: Ditto.

Source/Core/DolphinQt/Debugger/NewBreakpointDialog.h
```diff
@@ -56,4 +63,8 @@ class NewBreakpointDialog : public QDialog
 
   QDialogButtonBox* m_buttons;
   BreakpointWidget* m_parent;
+
+  bool m_edit;
+  TBreakPoint* m_instruction_breakpoint{nullptr};
```
sepalani, line 68: Holding pointers here sounds like a recipe for crashes. Can't you pass an optional, a variant, a reference to an overload or getting the right BP based on its address when needed rather than relying on theses pointers?

sepalani (changes requested):
To make things simpler and not interdependent, I strongly suggest you split this PR into smaller PRs. Especially your 2nd point which should have a PR of its own (i.e. `Allow to toggle breakpoints between on/off`).

sepalani:
For reference, an older attempt to enable/disable breakpoints: https://github.com/dolphin-emu/dolphin/pull/6728

InusualZ:
This PR would be splitted.

PR #9529, opened 2021-02-20
Android: Broadcast update before updating additional metadata

JosJuice:
In master, the game scanning process looks like this:

1. Scan for games
2. Scan for additional metadata (icon.png and meta.xml)
3. Save the cache if needed
4. Update the game list with the results

This change makes the game scanning process look like this:

1. Scan for games
2. Update the game list with the results
3. Scan for additional metadata (icon.png and meta.xml)
4. Update the game list with the results
5. Save the cache if needed

Updating the game list as soon as possible means the user has to wait less before their games show up. The new behavior matches what DolphinWX did before it was removed. (DolphinQt has an even fancier approach where games get added one by one.)


leoetlino (approved):
LGTM but a rebase is required

PR #7081, opened 2018-06-05
Patches window added

sepalani:
This PR adds a Patches window which allows to manage (enable/disable, delete, clear) patches such as replaced instructions.

Ready to be reviewed & merged.

image

PR #9531, opened 2021-02-22
DolphinQt: Patches window added

sepalani:
This PR adds a widget to handle memory patches like replaced instructions. (based on https://github.com/dolphin-emu/dolphin/pull/7081)

At the moment, it looks like this:
image

There are few things I need to address before making this PR ready to merge:
 - [ ] Fix a crash when `ClearPatches()` is called on shutdown
 - [ ] Show both the original value and the patched value in the table
 - [ ] Add buttons to load from and save all patches to a file
 - [ ] Add button to inject a file into a specific memory address

Ready to reviewed.


Source/Core/Common/Debug/MemoryPatches.cpp
```diff
@@ -90,6 +98,87 @@ void MemoryPatches::RemovePatch(std::size_t index)
...
+      DisablePatch(index);
+  }
+}
+
+std::vector<std::string> MemoryPatches::SaveToStrings() const
+{
+  std::vector<std::string> patches;
+  for (const auto& patch : m_patches)
+  {
+    std::ostringstream oss;
```
iwubcode, line 170: I wonder if it makes sense to make a `fmt::formatter` for `MemoryPatch` then  you just do:  `fmt::to_string(patch)`...

Source/Core/Core/Debugger/PPCDebugInterface.cpp
```diff
@@ -24,19 +24,27 @@
...
+  if (patch.patch_value.empty())
     return;
 
   const u32 address = patch.address;
-  const std::size_t size = patch.value.size();
+  const std::size_t size = patch.patch_value.size();
   if (!PowerPC::HostIsRAMAddress(address))
     return;
 
+  const bool is_enabled = patch.is_enabled == Common::Debug::MemoryPatch::State::Enabled;
```
iwubcode, line 35: I've seen this a couple times, maybe it makes sense to make a function?

Source/Core/Common/Debug/MemoryPatches.cpp
Pokechu22: Another option is to implement `std::ostream& operator<<` and use `fmt`'s [ostream support](https://fmt.dev/latest/api.html#ostream-api), if that'd make it easier to be consistent with the parsing code.

iwubcode:
One other idea, maybe adding a 'note' field or 'description' would be useful.

AdmiralCurtiss:
Can you rebase this on master?

sepalani:
Sure. However, I'm not satisfied with the current PR state anymore. There are some improvements I'd like to make first, so I'll switch this PR as a draft meanwhile.

PR #9532, opened 2021-02-23
Hackfix for debug cubes

Pokechu22:
This change hides the debug cubes in Super Mario Sunshine, while showing the other things that were previously broken by that change ([reflections in Super Monkey Ball](https://bugs.dolphin-emu.org/issues/12366), [trucks in Monster Jam](https://bugs.dolphin-emu.org/issues/12420), and [course elements in modded Mario Kart Wii courses](https://bugs.dolphin-emu.org/issues/12379)).  It does this by using a game-specific setting for the default color channel value when the color is not included in the vertex.

#9122 previously used the heuristic of setting alpha to 0 if the color format was ARGB (Color0Elements == 1), and setting alpha to 255 if the color format was RGB (Color0Elements == 0).  The R, G, and B components were always set to 255.  However, this heuristic turned out not to be correct, as it broke other games.

[My various experiments](https://gist.github.com/Pokechu22/f617a6b9486f0b53cc9343b926999a33#file-update5-md) have found that the cubes are actually colored transparent black (ARGB = 0x00000000) instead of transparent white (ARGB = 0x00FFFFFF), by setting `xfmem.alpha[0].matsource` to 0 (which means that the material color register's alpha value, which is 0xFF, replaces the previous color) and observing that the cubes showed up but were black.  Thus, this change makes the (invisible) cubes more hardware accurate, though this can only be seen by using the same patch I used in testing.

Note that since the fifo player does not use game-specific settings, the cubes will still show up in it (and thus on fifo.ci).  A proper fix would not use game-specific settings and thus would have them be invisible in the fifo player, but the exact conditions that make the cubes invisible are still unknown.

~~This PR is submitted for later use if I can't figure out what actually causes the cubes to be invisible, but shouldn't be merged unless it seems like that's not going to be solved at all.  (In addition, the shader UID version needs to be bumped, which I haven't done yet; when the shader UID version _is_ bumped I'd also want to remove some of the backwards-compatibility functionality in #9497 which has not yet been merged.)~~


phire:
If we are going to do a hack-fix, I think a "enabled by default" cheat-code which patches the game code to not emit ill-formed GX commands might be a better approach. 

At least that would get baked into any future fifologs that get recorded.

JMC47:
What would we do about romhacks and the such that don't use conventional GameIDs or boot from dols/homebrew channel?

phire:
Those are equally ugly with the per-game config hack approach.

This is why we really don't like per game fixes.

JMC47:
Right now we could go back to the old behavior *except* for Super Mario Sunshine and everything would work.  To our knowledge.  So a code for Super Mario Sunshine might be enough, but this behavior is theoretically correct for Super Mario Sunshine...?

Pokechu22:
I don't think it's feasible to do a game patch, as from what I remember the misconfiguration happens in level data and not in code, and I don't think there's one place in the code where it can be detected (for my testing, I forced use of the material color for _everything_, but that has side effects for other things and we'd also need to change the current material color to transparent black if we wanted to use it to hide the cubes.)  It might be possible, but it'd require a fair bit of research.

I don't think romhacks are a concern; if a hack has a new gameid, it probably has different levels, so the debug cubes in one specific level shouldn't be a problem.

phire:
We can totally patch level data too, but that might be getting a bit on the hard side.

Rumi-Larry:
Out of curiosity, does the PAL version of the game behave the same?

Pokechu22:
I haven't tested it (as I only own a US copy), and [bug 8059](https://bugs.dolphin-emu.org/issues/8059) doesn't say.  But probably, the US, PAL, Japanese, and Korean versions are all affected, as the Nintendo Switch version was affected and that would probably be built from the latest codebase.

Source/Core/Core/Config/GraphicsSettings.cpp
```diff
@@ -148,6 +148,7 @@ const Info<bool> GFX_HACK_COPY_EFB_SCALED{{System::GFX, "Hacks", "EFBScaledCopy"
 const Info<bool> GFX_HACK_EFB_EMULATE_FORMAT_CHANGES{
     {System::GFX, "Hacks", "EFBEmulateFormatChanges"}, false};
 const Info<bool> GFX_HACK_VERTEX_ROUDING{{System::GFX, "Hacks", "VertexRounding"}, false};
+const Info<u8> GFX_HACK_MISSING_COLOR_VALUE{{System::GFX, "Hacks", "MissingColorValue"}, 0xFF};
```
phire, line 151: For future flexibility, it might be wiser to allow the game ini to specify a 4 channel RGBA color. 

Because whatever is happening in underlying hardware, I really doubt it replicates the same value across all four channels.

phire (approved):

Source/Core/Core/Config/GraphicsSettings.cpp
Pokechu22: Done (and tested by confirming that `MissingColorValue = 0xFF00FFFF` results in magenta debug cubes).

Source/Core/VideoCommon/GXPipelineTypes.h
```diff
@@ -20,8 +20,7 @@ namespace VideoCommon
 // As pipelines encompass both shader UIDs and render states, changes to either of these should
 // also increment the pipeline UID version. Incrementing the UID version will cause all UID
 // caches to be invalidated.
-// TODO: Remove PixelShaderUid hasindstage on the next UID version bump
-constexpr u32 GX_PIPELINE_UID_VERSION = 2;  // Last changed in PR 9122
+constexpr u32 GX_PIPELINE_UID_VERSION = 2;  // Last changed in PR 9532
```
iwubcode, line 23: Did you mean to bump this?

Pokechu22: Yes, thanks for pointing that out.

iwubcode:
> Note that since the fifo player does not use game-specific settings, the cubes will still show up in it (and thus on fifo.ci). A proper fix would not use game-specific settings and thus would have them be invisible in the fifo player

If the fifologs stored the game id inside the file structure, we could _technically_ read that and use it as the game id for specific graphic settings?  Not sure what else that would break though

phire (approved):

Drflash55:
Oh, cool to see that a fix for this was finally able to be done and implemented into Rev 14402.
Finally, we can see the Monster Trucks again!

image

PR #9533, opened 2021-02-24
MMU: Fix IsRAMAddress not working

sepalani:
This PR fixes the `IsRAMAddress` function returning true when the RAM is NULL. It can happen very rarely when the hardware is shutting down and there is something interacting with the RAM at the same time like some debugger widgets.

Ready to be reviewed.

### DIFF
--- a/Source/Core/Core/PowerPC/MMU.cpp
+++ b/Source/Core/Core/PowerPC/MMU.cpp
@@ -207,7 +207,7 @@ static T ReadFromHardware(u32 em_address)
 
   // TODO: Make sure these are safe for unaligned addresses.
 
-  if ((em_address & 0xF8000000) == 0x00000000)
+  if (Memory::m_pRAM && (em_address & 0xF8000000) == 0x00000000)
   {
     // Handle RAM; the masking intentionally discards bits (essentially creating
     // mirrors of memory).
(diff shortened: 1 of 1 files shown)


AdmiralCurtiss:
This smells race-y. Do you have an example of a problem caused by this?

AdmiralCurtiss:
Aaah yeah okay that makes sense. And yeah, I agree, this patch looks sensible, I was just wondering if there wasn't a deeper issue hidden here.

Source/Core/Core/PowerPC/MMU.cpp
```diff
@@ -665,14 +665,15 @@ static bool IsRAMAddress(u32 address, bool translate)
...
+  if (Memory::m_pRAM && segment == 0x0 && (address & 0x0FFFFFFF) < Memory::GetRamSizeReal())
     return true;
   else if (Memory::m_pEXRAM && segment == 0x1 &&
            (address & 0x0FFFFFFF) < Memory::GetExRamSizeReal())
     return true;
   else if (Memory::m_pFakeVMEM && ((address & 0xFE000000) == 0x7E000000))
     return true;
-  else if (segment == 0xE && (address < (0xE0000000 + Memory::GetL1CacheSize())))
+  else if (Memory::m_pL1Cache && segment == 0xE &&
+           (address < (0xE0000000 + Memory::GetL1CacheSize())))
```
lioncash, line 676: this should be braced given the condition travels more than one line

AdmiralCurtiss:
You may want to add the same checks in `ReadFromHardware()` and `WriteToHardware()` too.

sepalani:
@AdmiralCurtiss 
Wouldn't that add an additional overhead and defeat the purpose of the `IsRAMAddress` function?

AdmiralCurtiss:
But in that case we're being quire inconsistent, ReadFromHardware() and WriteToHardware() do check `m_pEXRAM` and `m_pFakeVMEM`.

sepalani:
You've a point, I'll add the check to them as well, then.

PR #9534, opened 2021-02-24
Core/MemoryWatcher: Use appropriate memory read function in ChasePointer().

AdmiralCurtiss:
The MemoryWatcher probably shouldn't generate emulated exceptions or trigger memory breakpoints.

### DIFF
--- a/Source/Core/Core/MemoryWatcher.cpp
+++ b/Source/Core/Core/MemoryWatcher.cpp
@@ -8,9 +8,9 @@
 #include <unistd.h>
 
 #include "Common/FileUtil.h"
-#include "Core/HW/Memmap.h"
 #include "Core/HW/SystemTimers.h"
 #include "Core/MemoryWatcher.h"
+#include "Core/PowerPC/MMU.h"
 
 MemoryWatcher::MemoryWatcher()
 {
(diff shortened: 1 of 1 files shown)


AdmiralCurtiss:
Ohh I see what I'm missing here, `Memory::Read_U32()` != `PowerPC::Read_U32()`. But wait, in that case, does this even work at all? `PowerPC::HostIsRAMAddress()` wants a virtual address, but `Memory::Read_U32()` wants a physical address...

AdmiralCurtiss:
Okay I think I get it, this class comes from before we had proper MMU support at all, and `Memory::Read_U32()` masks off the top bits, so physical and virtual address resulted in the same thing in most cases. So, uh...

AdmiralCurtiss:
This is probably better?

sepalani (approved):
This change LGTM. However, I've never used the MemoryWatcher.

iwubcode:
Dumb question...how is the memory watcher different than the breakpoint logic and the watch window we have?

JosJuice:
It exposes similar features, but for scripting purposes rather than in a GUI.

leoetlino (approved):

PR #9535, opened 2021-02-25
DebugInterface: Check visited addresses in GetThreads

sepalani:
This PR is an attempt to fix https://bugs.dolphin-emu.org/issues/12410.

### DIFF
--- a/Source/Core/Core/Debugger/PPCDebugInterface.cpp
+++ b/Source/Core/Core/Debugger/PPCDebugInterface.cpp
@@ -4,9 +4,11 @@
 
 #include "Core/Debugger/PPCDebugInterface.h"
 
+#include <algorithm>
 #include <array>
 #include <cstddef>
 #include <string>
+#include <vector>
 
 #include <fmt/format.h>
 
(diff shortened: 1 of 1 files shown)


Source/Core/Core/Debugger/PPCDebugInterface.cpp
```diff
@@ -181,9 +183,13 @@ Common::Debug::Threads PPCDebugInterface::GetThreads() const
   if (!active_thread->IsValid())
     return threads;
 
-  const auto insert_threads = [&threads](u32 addr, auto get_next_addr) {
+  std::vector<u32> visited_addr{{active_thread->GetAddress()}};
```
iwubcode, line 186: Would a set or another container be a better choice?

AdmiralCurtiss: `vector<>` tends to perform really well for small sizes on modern systems thanks to cache locality, and I can't imagine the amount of addresses here reaching even three digits. So I can't imagine this mattering much.

sepalani:
Apparently, it does fix the issue: https://bugs.dolphin-emu.org/issues/12410#note-2

Source/Core/Core/Debugger/PPCDebugInterface.cpp
leoetlino, line 186: Nitpick: 
```suggestion
  std::vector<u32> visited_addrs{{active_thread->GetAddress()}};
```

sepalani: Done.

leoetlino (approved):

PR #9536, opened 2021-02-25
Don't call OnConfigChanged() unless config actually changed

Filoppi:
Also DualShock UDP Client was the only place in the code that assumed OnConfigChanged() was called at least once on startup or it won't load up the setting, so I took care of that.

OnConfigChanged() repopulates all the devices and it can be slow (and just plain unnecessary).
This might also help with any race conditions with devices population code, which as of now are unsafe and can break on some machines.


leoetlino (approved):

iwubcode:
Testing now, it seems like it might break Dynamic Input Textures

iwubcode (approved):

PR #9537, opened 2021-02-27
Core/Memmap: Give more detailed error messages if memory mapping fails.

AdmiralCurtiss:
Mostly I have a build to suggest over here to see if this could be a failure in InitFastmemArena(), which is silent on master: https://forums.dolphin-emu.org/Thread-wii-games-extremely-slow-after-revision-5-0-11476

### DIFF
--- a/Source/Core/Core/HW/Memmap.cpp
+++ b/Source/Core/Core/HW/Memmap.cpp
@@ -18,6 +18,7 @@
 #include "Common/CommonTypes.h"
 #include "Common/Logging/Log.h"
 #include "Common/MemArena.h"
+#include "Common/MsgHandler.h"
 #include "Common/Swap.h"
 #include "Core/Config/MainSettings.h"
 #include "Core/ConfigManager.h"
@@ -299,7 +300,9 @@ void Init()
 
     if (!*region.out_pointer)
     {
-      PanicAlertFmt("MemoryMap_Setup: Failed finding a memory base.");
+      PanicAlertFmt(
+          "Memory::Init(): Failed to create view for physical region at 0x{:08X} (size 0x{:08X}).",
+          region.physical_address, region.size);
       exit(0);
     }
   }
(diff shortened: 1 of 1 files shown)


Source/Core/Core/HW/Memmap.cpp
```diff
@@ -379,7 +387,10 @@ void UpdateLogicalMemory(const PowerPC::BatTable& dbat_table)
           void* mapped_pointer = g_arena.CreateView(position, mapped_size, base);
           if (!mapped_pointer)
           {
-            PanicAlertFmt("MemoryMap_Setup: Failed finding a memory base.");
+            PanicAlertFmt(
+                "Memory::UpdateLogicalMemory(): Failed to create view for physical region at "
+                "0x{:08X} (size 0x{:08X}).",
+                intersection_start - mapping_address, mapped_size);
```
AdmiralCurtiss, line 393: The physical address here might be wrong, I admit I don't fully understand what this code is doing...

Source/Core/Core/HW/Memmap.cpp
```diff
@@ -299,7 +299,9 @@ void Init()
 
     if (!*region.out_pointer)
     {
-      PanicAlertFmt("MemoryMap_Setup: Failed finding a memory base.");
+      PanicAlertFmt(
```
leoetlino, line 302: MsgHandler.h needs to be included.

leoetlino (changes requested):

AdmiralCurtiss:
@dolphin-emu-bot rebuild

Source/Core/Core/HW/Memmap.cpp
```diff
@@ -333,6 +339,9 @@ bool InitFastmemArena()
 
     if (base != view)
     {
+      PanicAlertFmt("Memory::InitFastmemArena(): Failed to map memory region at 0x{:08X} (size "
```
leoetlino, line 342: Nitpicking, but can the `(size ` be moved to the next line to make this more pleasant? (I assume clang-format is what reformatted it like this.)

AdmiralCurtiss: Yup that was clang-format. Should be nicer now.

leoetlino (approved):

PR #9540, opened 2021-02-28
Fifo analyzer quality of life improvements

Pokechu22:
Part 2 of #9497.  ~~Currently, the first commit stubs the changes it makes that are depended upon by later commits; once that is eventually emerged it should cleanly rebase with that commit dropped.~~

### DIFF
--- a/Source/Core/Core/FifoPlayer/FifoPlayer.cpp
+++ b/Source/Core/Core/FifoPlayer/FifoPlayer.cpp
@@ -189,6 +189,18 @@ bool FifoPlayer::IsRunningWithFakeVideoInterfaceUpdates() const
   return m_File->ShouldGenerateFakeVIUpdates();
 }
 
+u32 FifoPlayer::GetMaxObjectCount() const
+{
+  u32 result = 0;
+  for (auto& frame : m_FrameInfo)
+  {
+    const u32 count = static_cast<u32>(frame.objectStarts.size());
+    if (count > result)
+      result = count;
+  }
+  return result;
+}
+
 u32 FifoPlayer::GetFrameObjectCount() const
 {
   if (m_CurrentFrame < m_FrameInfo.size())
--- a/Source/Core/Core/FifoPlayer/FifoPlayer.cpp
+++ b/Source/Core/Core/FifoPlayer/FifoPlayer.cpp
@@ -384,6 +384,8 @@ void FifoPlayer::WriteFifo(const u8* data, u32 start, u32 end)
   {
     while (IsHighWatermarkSet())
     {
+      if (CPU::GetState() != CPU::State::Running)
+        break;
       CoreTiming::Idle();
       CoreTiming::Advance();
     }
--- a/Source/Core/Core/FifoPlayer/FifoPlayer.cpp
+++ b/Source/Core/Core/FifoPlayer/FifoPlayer.cpp
@@ -51,7 +51,7 @@ bool FifoPlayer::Open(const std::string& filename)
   {
     FifoPlaybackAnalyzer::AnalyzeFrames(m_File.get(), m_FrameInfo);
 
-    m_FrameRangeEnd = m_File->GetFrameCount();
+    m_FrameRangeEnd = m_File->GetFrameCount() - 1;
   }
 
   if (m_FileLoadedCb)
@@ -131,13 +131,10 @@ private:
 
 CPU::State FifoPlayer::AdvanceFrame()
 {
-  if (m_CurrentFrame >= m_FrameRangeEnd)
+  if (m_CurrentFrame > m_FrameRangeEnd)
   {
     if (!m_Loop)
       return CPU::State::PowerDown;
-    // If there are zero frames in the range then sleep instead of busy spinning
-    if (m_FrameRangeStart >= m_FrameRangeEnd)
-      return CPU::State::Stepping;
 
     // When looping, reload the contents of all the BP/CP/CF registers.
     // This ensures that each time the first frame is played back, the state of the
@@ -215,9 +212,9 @@ void FifoPlayer::SetFrameRangeStart(u32 start)
 {
   if (m_File)
   {
-    u32 frameCount = m_File->GetFrameCount();
-    if (start > frameCount)
-      start = frameCount;
+    const u32 lastFrame = m_File->GetFrameCount() - 1;
+    if (start > lastFrame)
+      start = lastFrame;
 
     m_FrameRangeStart = start;
     if (m_FrameRangeEnd < start)
@@ -232,9 +229,9 @@ void FifoPlayer::SetFrameRangeEnd(u32 end)
 {
   if (m_File)
   {
-    u32 frameCount = m_File->GetFrameCount();
-    if (end > frameCount)
-      end = frameCount;
+    const u32 lastFrame = m_File->GetFrameCount() - 1;
+    if (end > lastFrame)
+      end = lastFrame;
 
     m_FrameRangeEnd = end;
     if (m_FrameRangeStart > end)
--- a/Source/Core/Core/FifoPlayer/FifoPlayer.cpp
+++ b/Source/Core/Core/FifoPlayer/FifoPlayer.cpp
@@ -198,14 +198,19 @@ u32 FifoPlayer::GetMaxObjectCount() const
   return result;
 }
 
-u32 FifoPlayer::GetFrameObjectCount() const
+u32 FifoPlayer::GetFrameObjectCount(u32 frame) const
 {
-  if (m_CurrentFrame < m_FrameInfo.size())
+  if (frame < m_FrameInfo.size())
   {
-    return (u32)(m_FrameInfo[m_CurrentFrame].objectStarts.size());
+    return static_cast<u32>(m_FrameInfo[frame].objectStarts.size());
   }
 
   return 0;
+}
+
+u32 FifoPlayer::GetCurrentFrameObjectCount() const
+{
+  return GetFrameObjectCount(m_CurrentFrame);
 }
 
 void FifoPlayer::SetFrameRangeStart(u32 start)
--- a/Source/Core/DolphinQt/FIFO/FIFOAnalyzer.cpp
+++ b/Source/Core/DolphinQt/FIFO/FIFOAnalyzer.cpp
@@ -165,6 +165,9 @@ void FIFOAnalyzer::UpdateDetails()
   m_detail_list->clear();
   m_object_data_offsets.clear();
 
+  if (!FifoPlayer::GetInstance().IsPlaying())
+    return;
+
   auto items = m_tree_widget->selectedItems();
 
   if (items.isEmpty() || items[0]->data(0, OBJECT_ROLE).isNull())
@@ -316,6 +319,9 @@ void FIFOAnalyzer::BeginSearch()
 {
   QString search_str = m_search_edit->text();
 
+  if (!FifoPlayer::GetInstance().IsPlaying())
+    return;
+
   auto items = m_tree_widget->selectedItems();
 
   if (items.isEmpty() || items[0]->data(0, FRAME_ROLE).isNull())
@@ -461,6 +467,9 @@ void FIFOAnalyzer::ShowSearchResult(size_t index)
 void FIFOAnalyzer::UpdateDescription()
 {
   m_entry_detail_browser->clear();
+
+  if (!FifoPlayer::GetInstance().IsPlaying())
+    return;
 
   auto items = m_tree_widget->selectedItems();
 
--- a/Source/Core/DolphinQt/FIFO/FIFOAnalyzer.cpp
+++ b/Source/Core/DolphinQt/FIFO/FIFOAnalyzer.cpp
@@ -271,21 +271,37 @@ void FIFOAnalyzer::UpdateDetails()
     break;
 
     case OpcodeDecoder::GX_LOAD_INDX_A:
-      new_label = QStringLiteral("LOAD INDX A");
+    {
+      const auto [desc, written] =
+          GetXFIndexedLoadInfo(ARRAY_XF_A, Common::swap32(&object[object_offset]));
       object_offset += 4;
-      break;
+      new_label = QStringLiteral("LOAD INDX A   %1").arg(QString::fromStdString(desc));
+    }
+    break;
     case OpcodeDecoder::GX_LOAD_INDX_B:
-      new_label = QStringLiteral("LOAD INDX B");
+    {
+      const auto [desc, written] =
+          GetXFIndexedLoadInfo(ARRAY_XF_B, Common::swap32(&object[object_offset]));
       object_offset += 4;
-      break;
+      new_label = QStringLiteral("LOAD INDX B   %1").arg(QString::fromStdString(desc));
+    }
+    break;
     case OpcodeDecoder::GX_LOAD_INDX_C:
-      new_label = QStringLiteral("LOAD INDX C");
+    {
+      const auto [desc, written] =
+          GetXFIndexedLoadInfo(ARRAY_XF_C, Common::swap32(&object[object_offset]));
       object_offset += 4;
-      break;
+      new_label = QStringLiteral("LOAD INDX C   %1").arg(QString::fromStdString(desc));
+    }
+    break;
     case OpcodeDecoder::GX_LOAD_INDX_D:
-      new_label = QStringLiteral("LOAD INDX D");
+    {
+      const auto [desc, written] =
+          GetXFIndexedLoadInfo(ARRAY_XF_D, Common::swap32(&object[object_offset]));
       object_offset += 4;
-      break;
+      new_label = QStringLiteral("LOAD INDX D   %1").arg(QString::fromStdString(desc));
+    }
+    break;
 
     case OpcodeDecoder::GX_CMD_CALL_DL:
       // The recorder should have expanded display lists into the fifo stream and skipped the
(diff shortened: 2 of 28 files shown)


Source/Core/DolphinQt/FIFO/FIFOPlayerWindow.cpp
```diff
@@ -297,14 +297,14 @@ void FIFOPlayerWindow::OnFIFOLoaded()
...
+  auto object_count = FifoPlayer::GetInstance().GetMaxObjectCount();
   auto frame_count = file->GetFrameCount();
 
   m_frame_range_to->setMaximum(frame_count);
-  m_object_range_to->setMaximum(object_count);
+  m_object_range_to->setMaximum(object_count - 1);
 
   m_frame_range_to->setValue(frame_count);
-  m_object_range_to->setValue(object_count);
+  m_object_range_to->setValue(object_count - 1);
```
Pokechu22: These (/ `FIFOPlayerWindow::OnFIFOLoaded()` in general) are also called when the FIFO is unpaused, which turns out to be rather annoying since it changes the upper bound. (Also a problem with 5.0-13603)

(moved from [here](https://github.com/dolphin-emu/dolphin/pull/9497#discussion_r581567296))

Source/Core/Core/FifoPlayer/FifoPlayer.cpp
```diff
@@ -189,6 +189,18 @@ bool FifoPlayer::IsRunningWithFakeVideoInterfaceUpdates() const
   return m_File->ShouldGenerateFakeVIUpdates();
 }
 
+u32 FifoPlayer::GetMaxObjectCount() const
+{
+  u32 result = 0;
```
iwubcode, line 191: Nothing wrong with what you have but here's an alternative:

```
auto comparison = [](const AnalyzedFrameInfo& a, const AnalyzedFrameInfo& b)
{
  return a.objectStarts.size() < b.objectStarts.size();
 };

return std::max_element(m_FrameInfo.begin(), m_FrameInfo.end(), comparison);
```

Source/Core/VideoCommon/XFStructs.cpp
```diff
@@ -300,3 +304,98 @@ void PreprocessIndexedXF(u32 val, int refarray)
...
+
+std::string GetXFMemDescription(u32 address, u32 value)
+{
+  return fmt::format("{} = {:08x}", GetXFMemName(address), value);
+}
+
+#define XFMEM_REGISTERS_START 0x1000
+#define XFMEM_REGISTERS_END 0x1058
+
+std::pair<std::string, std::string> GetXFTransferInfo(const u8* data)
```
iwubcode, line 330: Just a suggestion but since you're appending to a string with `fmt::format`, it would likely be more performant to use `fmt::memory_buffer` with `fmt::format_to`.  Probably negligible but just thought I'd mention :)

Pokechu22: This code actually comes from PR 9497, but I've changed it to use memory_buffer there (and updated the stub here to match).

Source/Core/DolphinQt/FIFO/FIFOAnalyzer.cpp
```diff
@@ -550,6 +597,51 @@ void FIFOAnalyzer::UpdateDescription()
...
+    u32 vertex_size;
+    if (vertex_count != 0)
+    {
+      if ((primitive_size % vertex_count) == 0)
+      {
+        vertex_size = primitive_size / vertex_count;
+      }
+      else
+      {
+        text += tr("NOTE: Stream size doesn't match actual data length");
```
iwubcode, line 693: Just want to say, I saw this quite a bit in one of my fifo logs.  Not sure if that's valid or not?

Pokechu22: That message existed beforehand (and should occur with the same frequency as before).  I've taken a brief look and it seems like sometimes there can be multiple draw commands in a row that get bundled into one object (but then the size doesn't match), and there also sometimes is NOP padding at the end (which will also cause the size not to match).  It's something that could be handled much more elegantly if the vertex format were known at that point in the code, but it's not known.  (I still need to look at what code is responsible for breaking the FIFO into objects and how it can be tweaked to give better results/how it handles these things.)

Source/Core/Core/FifoPlayer/FifoPlayer.cpp
Pokechu22: I don't think that would work.  It'd need to dereference the iterator and then look at that value's `objectStarts.size()`, and I'm not 100% sure that `m_FrameInfo` can't be empty at this point so it'd also need to check that it's not `m_FrameInfo.end()`.  That'd just make it a bunch more complicated; I don't think it's worth it.

Source/Core/DolphinQt/FIFO/FIFOAnalyzer.cpp
Pokechu22: I've since changed it so that the object size is known; this issue no longer happens.  Since the vertex size is known, the number of bytes each draw command uses also is known, so multiple draw commands in a row are also handled better.  (One other fun edge case is that you can have a draw command that draws 0 vertices... which the code previously wasn't handling in any sane way, because it couldn't even do this size check in that case.  Now it's just treated the exact same as the other cases.)

Pokechu22:
Here's what the new descriptions for indexed XF loads and for primitive operations look like:

<details><summary>Super Mario Sunshine</summary>

image
image
image
</details>
<details><summary>Super Monkey Ball</summary>

image
image
</details>

You still need to look at other commands to figure out exactly what the different parts of the primitive commands mean, but the spaces should still make it a lot easier.

iwubcode:
Person who merges this, this fixes:  https://bugs.dolphin-emu.org/issues/12452 !

iwubcode:
@Pokechu22 - I'm assuming this is ready to go?

Pokechu22:
Yes, it is ready.

Source/Core/VideoCommon/VertexLoader.cpp
```diff
@@ -205,98 +205,16 @@ void VertexLoader::CompileVertexTranslator()
     m_native_vtx_decl.colors[i].components = 4;
     m_native_vtx_decl.colors[i].type = VAR_UNSIGNED_BYTE;
     m_native_vtx_decl.colors[i].integer = false;
-    switch (m_VtxDesc.low.Color[i])
```
iwubcode, line 208: It'd be a pain but seeing as this class already has some unit-tests, I wonder if it'd be beneficial to write some tests before removing this with the improved replacement.

Source/Core/VideoCommon/VertexLoaderBase.h
```diff
@@ -77,11 +77,10 @@ class VertexLoaderBase
...
 
 protected:
-  VertexLoaderBase(const TVtxDesc& vtx_desc, const VAT& vtx_attr);
-  void SetVAT(const VAT& vat);
+  VertexLoaderBase(const TVtxDesc& vtx_desc, const VAT& vtx_attr)
+      : m_VtxDesc{vtx_desc}, m_VtxAttr{vtx_attr} {};
 
   // GC vertex format
-  TVtxAttr m_VtxAttr;  // VAT decoded into easy format
+  VAT m_VtxAttr;
```
iwubcode: Now easier format :)

Source/Core/VideoCommon/VertexLoaderBase.cpp
```diff
@@ -91,6 +97,62 @@ class VertexLoaderTester : public VertexLoaderBase
   std::vector<u8> buffer_b;
 };
 
+u32 VertexLoaderBase::GetVertexSize(const TVtxDesc& vtx_desc, const VAT& vtx_attr)
+{
+  u32 size = 0;
+  if (vtx_desc.low.PosMatIdx)
+    size++;
+  for (auto texmtxidx : vtx_desc.low.TexMatIdx)
```
iwubcode, line 105: Can you not do `size += vtx_desc.low.TexMatIdx.size();`?

Pokechu22: That would always add 8, but we want to count the number that are set.

Source/Core/DolphinQt/FIFO/FIFOPlayerWindow.cpp
```diff
@@ -6,9 +6,13 @@
 
 #include <QCheckBox>
 #include <QDialogButtonBox>
+#include <QEvent>
```
iwubcode, line 9: Your commit comment.

>This way, it can be focused with the render window behind it

Did you mean?

>This way, it can be focused with the render window in front of it

?

You then go on to say:

>instead of having the main window show up and cover the render window

Which seems opposite to the first comment.

Source/Core/DolphinQt/FIFO/FIFOAnalyzer.cpp
```diff
@@ -330,42 +348,54 @@ void FIFOAnalyzer::UpdateDetails()
...
+                        .arg(object_prim_size);
 
-  if (vertex_count != 0 && (object_prim_size % vertex_count) != 0)
-  {
-    new_label += QLatin1Char{'\n'};
-    new_label += tr("NOTE: Stream size doesn't match actual data length");
+        // It's not really useful to have a massive unreadable hex string for the object primitives.
+        // Put it in the description instead.
+
+        // new_label += QStringLiteral("   ");
```
iwubcode, line 379: Still a todo or should this be removed?

Pokechu22: More of a leftover/sample implementation in case it's something that is wanted for the future.  Perhaps `#if 0` would be better than commenting it out?

Source/Core/DolphinQt/FIFO/FIFOPlayerWindow.cpp
Pokechu22: Nope, I meant exactly what I said.  Before, you could only have this:

image

or this:

image

i.e. the main window is always directly behind the FIFO player; you can't have the FIFO player and then the render window without the main window in the way (it's much more annoying when the main window is in fullscreen; the workaround I used before was to drag the main window into the corner so it's mostly invisible).  Perhaps it's more of a problem for me due to having a fairly small display.

iwubcode: Ah thanks.  I see what you mean now and your description makes sense just needed pictures to fully grasp it :)

Source/Core/DolphinQt/FIFO/FIFOAnalyzer.cpp
Pokechu22: I've replaced it with a `#ifdef` (and also adjusted the implementation, since it wasn't actually correct for multiple objects).

Source/Core/VideoCommon/VertexLoader.cpp
Pokechu22: I don't feel like writing tests exhaustive enough to cover every combination of color settings; although they'd be useful to have, it seems like a fair bit of effort which I don't want to include in this PR.  The refactoring I did here is basically just converting the switch statement into two arrays, without changing the functions themselves, so I don't think there's too much I could have broken.

It's also worth noting that the regular `VertexLoader` is not actually used by anything currently (only `VertexLoaderARM64` and `VertexLoaderX64` are used in practice since Dolphin only supports X64 and ARM64); it can be used by editing `VertexLoaderBase::CreateVertexLoader` (and in particular by defining `COMPARE_VERTEXLOADERS`; I've run the existing unit tests with that enabled and they all pass and some non-exhaustive gameplay testing also showed no issues).

Source/Core/Core/FifoPlayer/FifoPlayer.cpp
```diff
@@ -203,9 +212,9 @@ void FifoPlayer::SetFrameRangeStart(u32 start)
 {
   if (m_File)
   {
-    u32 frameCount = m_File->GetFrameCount();
-    if (start > frameCount)
-      start = frameCount;
+    u32 lastFrame = m_File->GetFrameCount() - 1;
```
Dentomologist, line 215: This can be const.

Source/Core/Core/FifoPlayer/FifoPlayer.cpp
```diff
@@ -220,9 +229,9 @@ void FifoPlayer::SetFrameRangeEnd(u32 end)
 {
   if (m_File)
   {
-    u32 frameCount = m_File->GetFrameCount();
-    if (end > frameCount)
-      end = frameCount;
+    u32 lastFrame = m_File->GetFrameCount() - 1;
```
Dentomologist, line 232: const

Source/Core/VideoCommon/VertexLoaderBase.h
Pokechu22: On a related note, I'm not sure if the comment below really makes sense anymore:

https://github.com/dolphin-emu/dolphin/blob/c350da023ded841c859e01ef6a38f32f7e4ea829/Source/Core/VideoCommon/VertexLoaderBase.h#L90-L91

I don't think there's any way to avoid use of `m_VtxDesc`; the comment only makes sense if the goal was to also include that info in TVtxAttr, in which case the comment doesn't apply anymore.

Pokechu22: Comment removed (and also, `m_VtxAttr` and `m_VtxDesc` are now changed to `const` in "Eliminate TVtxAttr" instead of "Move vertex size and component calculation to VertexLoaderBase").

Source/Core/VideoCommon/XFStructs.cpp
```diff
@@ -256,12 +256,19 @@ void LoadXFReg(u32 transferSize, u32 baseAddress, DataReader src)
   }
 }
 
+constexpr std::tuple<u32, u32, u32> ExtractIndexedXF(u32 val)
+{
+  u32 index = val >> 16;
+  u32 address = val & 0xFFF;  // check mask
+  u32 size = ((val >> 12) & 0xF) + 1;
```
Pokechu22, line 263: can be `const`

Source/Core/DolphinQt/FIFO/FIFOAnalyzer.cpp
```diff
@@ -160,168 +165,256 @@ void FIFOAnalyzer::UpdateTree()
...
 
+static std::string GetPrimitiveName(u8 cmd)
+{
+  if ((cmd & 0xC0) != 0x80)
+  {
+    PanicAlertFmt("Not a primitive command: {:#04x}", cmd);
+    return "";
+  }
+  u8 vat = cmd & OpcodeDecoder::GX_VAT_MASK;  // Vertex loader index (0 - 7)
+  u8 primitive = (cmd & OpcodeDecoder::GX_PRIMITIVE_MASK) >> OpcodeDecoder::GX_PRIMITIVE_SHIFT;
```
Pokechu22, line 176: Can be `const`

Source/Core/DolphinQt/FIFO/FIFOAnalyzer.cpp
```diff
@@ -160,168 +165,256 @@ void FIFOAnalyzer::UpdateTree()
...
+        // It's not really useful to have a massive unreadable hex string for the object primitives.
+        // Put it in the description instead.
+
+// #define INCLUDE_HEX_IN_PRIMITIVES
+#ifdef INCLUDE_HEX_IN_PRIMITIVES
+        new_label += QStringLiteral("   ");
+        for (u32 i = 0; i < object_prim_size; i++)
+        {
+          u32 byte = object[object_offset++];
+          new_label += QStringLiteral("%1").arg(byte, 2, 16, QLatin1Char('0'));
```
Pokechu22, line 385: Can be `const`, but it also doesn't need to be a `u32` here (`u8` would be better, or just not having a variable for it).

Source/Core/VideoCommon/VertexLoaderX64.cpp
```diff
@@ -53,7 +53,7 @@ VertexLoaderX64::VertexLoaderX64(const TVtxDesc& vtx_desc, const VAT& vtx_att)
   GenerateVertexLoader();
   WriteProtect();
 
-  const std::string name = ToString();
+  std::string name = fmt::format("VertexLoaderX64\nVtx desc: \n{}\nVAT:\n{}", vtx_desc, vtx_att);
```
iwubcode, line 53: Why was the `const` removed here?

Pokechu22: I don't think there's any reason why; it probably was by accident but I'm not entirely sure how I managed to do that.

Pokechu22:
I've fixed an issue where the same object count was used for all frames (based on the object count of the current frame at the time the FIFO player loaded, which could be effectively random in some situations); this could lead to some frames missing objects or more objects being listed than actually exist (the latter leading to crashes when those objects are selected).  This bug existed since the Qt FIFO player was added; the fix was 888762a4fa1983861acb925105a7c9a84e6e3132.

I've also made it so that the number before each command is the offset within the current frame, instead of the offset within the object; this matches the previous behavior.  (This change was made in 678d567fe4 by using `object_start + object_offset` instead of just `object_offset`).

JMC47:
Since this fixes a crash in the fifoplayer, should we have this merged before the next beta?  It also is a pretty good improvement.  I just want to make sure we aren't possibly merging in any performance regressions.

I did test this and ran through some fifologs and did analyzation stuff.  I don't know how the analyzer works though so I just kind of hit buttons and randomly clicked through stuff.

Functionality LGTM.

iwubcode (approved):

Source/Core/DolphinQt/FIFO/FIFOAnalyzer.cpp
```diff
@@ -356,36 +449,35 @@ void FIFOAnalyzer::BeginSearch()
...
+    for (const u8* ptr = cmd_start_ptr; ptr < cmd_end_ptr - length + 1; ptr++)
+    {
+      if (std::equal(search_val.begin(), search_val.end(), ptr))
       {
-        if (ptr < start_ptr + m_object_data_offsets[cmd_nr])
-        {
-          result.cmd = cmd_nr - 1;
-          break;
-        }
+        m_search_results.push_back(SearchResult(frame_nr, object_nr, cmd_nr));
```
Pokechu22, line 478: This can be `emplace_back`.

Source/Core/VideoCommon/VertexLoader_Color.h
```diff
@@ -6,23 +6,9 @@
...
-void Color_ReadIndex16_24b_888(VertexLoader* loader);
-void Color_ReadIndex16_32b_888x(VertexLoader* loader);
-void Color_ReadIndex16_16b_4444(VertexLoader* loader);
-void Color_ReadIndex16_24b_6666(VertexLoader* loader);
-void Color_ReadIndex16_32b_8888(VertexLoader* loader);
+class VertexLoader_Color
+{
+public:
+  static u32 GetSize(VertexComponentFormat type, ColorFormat format);
+  static TPipelineFunction GetFunction(VertexComponentFormat type, ColorFormat format);
```
Pokechu22, line 13: This requires forward declaration of `VertexComponentFormat` and `ColorFormat`, and including `VertexLoader.h` for `TPipelineFunction`.  The forward declaration of `VertexLoader` is also no longer needed.

Source/Core/DolphinQt/FIFO/FIFOAnalyzer.cpp
```diff
@@ -139,189 +144,277 @@ void FIFOAnalyzer::UpdateTree()
...
-
-  m_detail_list->addItem(new_label);
-  m_object_data_offsets.push_back(0);
-
-  // Between objectdata_end and next_objdata_start, there are register setting commands
-  if (object_nr + 1 < static_cast<int>(frame_info.objectStarts.size()))
+  // Note that frame_info.objectStarts[object_nr] is the start of the primitive data,
+  // but we want to start with the register updates which happen before that.
+  const u32 object_start = (object_nr == 0 ? 0 : frame_info.objectEnds[object_nr - 1]);
+  const u32 object_size = frame_info.objectEnds[object_nr] - object_start;
```
phire, line 216: Ok... but now you are missing the register updates after the final object in the frame.

Also, the terminology is getting mixed up here. In the fifo playback code, Objects are explictly primitive draws and just the primitive draws. They are only defined so that the object skipping code can skip the objects and not the register writes around it.

Register writes being connected to objects is something that only happens in this UI code

phire:
> I did test this and ran through some fifologs and did analyzation stuff. I don't know how the analyzer works though so I just kind of hit buttons and randomly clicked through stuff.

Actually a pretty good way to test it. If it's broken you, dolphin will crash or panicalert.

phire (approved):

lioncash (approved):

PR #9541, opened 2021-02-28
Symbols: Fix RSO Modules detection algorithm

InusualZ:
The current algorithm for detecting a RSO modules have a flaw. In which RSO modules are only detected if the byte before the name of the module is a null byte `\0`. This is wrong because before the name can be any arbitrary data.

I modify the approach. Instead of looking for the first null byte before the name. I count the bytes before a non-ascii character is found. After that, I look for a `uint32` offset that point to the name's bytes. Also, to be even more sure about it, after we have found the offset, the next 4 bytes are the module's name length and I also check against that. 

Also, even after this fix, we are not going to be capable of detecting every RSO module, since modules can come without name ([File Format Specification](http://www.metroid2002.com/retromodding/wiki/RSO_(File_Format)))

## Before
image

## After
image

image

I had to add a loading bar, because now the process is considerably slower. Now the user know that some work is being done.

### DIFF
--- a/Source/Core/DolphinQt/MenuBar.cpp
+++ b/Source/Core/DolphinQt/MenuBar.cpp
@@ -5,6 +5,7 @@
 #include "DolphinQt/MenuBar.h"
 
 #include <cinttypes>
+#include <future>
 
 #include <QAction>
 #include <QActionGroup>
@@ -52,6 +53,7 @@
 #include "DolphinQt/AboutDialog.h"
 #include "DolphinQt/Host.h"
 #include "DolphinQt/QtUtils/ModalMessageBox.h"
+#include "DolphinQt/QtUtils/ParallelProgressDialog.h"
 #include "DolphinQt/Settings.h"
 #include "DolphinQt/Updater.h"
 
@@ -1297,48 +1299,26 @@ void MenuBar::GenerateSymbolsFromRSO()
 
 void MenuBar::GenerateSymbolsFromRSOAuto()
 {
-  constexpr std::array<std::string_view, 2> search_for = {".elf", ".plf"};
-  const AddressSpace::Accessors* accessors =
-      AddressSpace::GetAccessors(AddressSpace::Type::Effective);
-  std::vector<std::pair<u32, std::string>> matches;
+  ParallelProgressDialog progress(tr("Modules found: %1").arg(0), tr("Cancel"), 0, 0, this);
+  progress.GetRaw()->setWindowTitle(tr("Detecting RSO Modules"));
+  progress.GetRaw()->setMinimumDuration(1000 * 10);
+  progress.GetRaw()->setWindowModality(Qt::WindowModal);
 
-  // Find filepath to elf/plf commonly used by RSO modules
-  for (const auto& str : search_for)
-  {
-    u32 next = 0;
-    while (true)
-    {
-      auto found_addr =
-          accessors->Search(next, reinterpret_cast<const u8*>(str.data()), str.size() + 1, true);
+  auto future = std::async(std::launch::async, [&progress, this]() -> RSOVector {
+    progress.SetValue(0);
+    auto matches = DetectRSOModules(progress);
+    progress.Reset();
 
-      if (!found_addr.has_value())
-        break;
-      next = *found_addr + 1;
+    return matches;
+  });
+  progress.GetRaw()->exec();
 
-      // Get the beginning of the string
-      found_addr = accessors->Search(*found_addr, reinterpret_cast<const u8*>(""), 1, false);
-      if (!found_addr.has_value())
-        continue;
-
-      // Get the string reference
-      const u32 ref_addr = *found_addr + 1;
-      const std::array<u8, 4> ref = {static_cast<u8>(ref_addr >> 24),
-                                     static_cast<u8>(ref_addr >> 16),
-                                     static_cast<u8>(ref_addr >> 8), static_cast<u8>(ref_addr)};
-      found_addr = accessors->Search(ref_addr, ref.data(), ref.size(), false);
-      if (!found_addr.has_value() || *found_addr < 16)
-        continue;
-
-      // Go to the beginning of the RSO header
-      matches.emplace_back(*found_addr - 16, PowerPC::HostGetString(ref_addr, 128));
-    }
-  }
+  auto matches = future.get();
 
   QStringList items;
   for (const auto& match : matches)
   {
     const QString item = QLatin1String("%1 %2");
-
     items << item.arg(QString::number(match.first, 16), QString::fromStdString(match.second));
   }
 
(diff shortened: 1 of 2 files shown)


sepalani:
@dolphin-emu-bot rebuild

sepalani:
@dolphin-emu-bot rebuild

sepalani:
@dolphin-emu-bot rebuild

Source/Core/DolphinQt/MenuBar.cpp
```diff
@@ -1313,24 +1313,64 @@ void MenuBar::GenerateSymbolsFromRSOAuto()
...
+      };
 
-      // Get the string reference
-      const u32 ref_addr = *found_addr + 1;
-      const std::array<u8, 4> ref = {static_cast<u8>(ref_addr >> 24),
-                                     static_cast<u8>(ref_addr >> 16),
-                                     static_cast<u8>(ref_addr >> 8), static_cast<u8>(ref_addr)};
-      found_addr = accessors->Search(ref_addr, ref.data(), ref.size(), false);
-      if (!found_addr.has_value() || *found_addr < 16)
+      const auto nameLength = get_string_len_until_non_ascii();
```
sepalani, line 1401: Should be renamed as `max_name_length` or similar.

Source/Core/DolphinQt/MenuBar.cpp
```diff
@@ -1313,24 +1313,64 @@ void MenuBar::GenerateSymbolsFromRSOAuto()
...
+      auto found = false;
+      u32 module_name_length = 0;
+
+      // Look for the Module Name Offset Field
```
sepalani, line 1405:
```C++
// Look for the Module Name Offset Field based on each possible length
```

Source/Core/DolphinQt/MenuBar.cpp
```diff
@@ -1313,24 +1313,64 @@ void MenuBar::GenerateSymbolsFromRSOAuto()
...
       if (!found_addr.has_value())
         break;
+
       next = *found_addr + 1;
 
-      // Get the beginning of the string
-      found_addr = accessors->Search(*found_addr, reinterpret_cast<const u8*>(""), 1, false);
-      if (!found_addr.has_value())
-        continue;
+      auto get_string_len_until_non_ascii = [found_addr]() {
```
sepalani, line 1379: This should probably be renamed since it grabs the maximum length of a module name (and excludes non printable characters not all non ascii ones). A description should be added to emphasis why it's important.

```C++
// Non-null data can precede the module name.
// Get the maximum name length that a module could have.
auto get_max_module_name_len = 
```

sepalani (changes requested):
Otherwise, this new heuristic is a better than the previous one.

Source/Core/DolphinQt/MenuBar.cpp
```diff
@@ -1297,48 +1299,26 @@ void MenuBar::GenerateSymbolsFromRSO()
...
-    u32 next = 0;
-    while (true)
-    {
-      auto found_addr =
-          accessors->Search(next, reinterpret_cast<const u8*>(str.data()), str.size() + 1, true);
-
-      if (!found_addr.has_value())
-        break;
-      next = *found_addr + 1;
+  ParallelProgressDialog progress(tr("Modules Found %1").arg(0), tr("Cancel"), 0, 0, this);
```
leoetlino, line 1302:
```suggestion
  ParallelProgressDialog progress(tr("Modules found: %1").arg(0), tr("Cancel"), 0, 0, this);
```

Source/Core/DolphinQt/MenuBar.cpp
```diff
@@ -1368,6 +1348,106 @@ void MenuBar::GenerateSymbolsFromRSOAuto()
...
+      if (!found)
+        continue;
+
+      const auto module_name_offset = accessors->ReadU32(*found_addr);
+
+      // Go to the beginning of the RSO header
+      matches.emplace_back(*found_addr - 16,
+                           PowerPC::HostGetString(module_name_offset, module_name_length));
+
+      progress.SetLabelText(tr("Modules Found %1").arg(matches.size()));
```
leoetlino, line 1444:
```suggestion
      progress.SetLabelText(tr("Modules found: %1").arg(matches.size()));
```

Source/Core/DolphinQt/MenuBar.cpp
```diff
@@ -1313,24 +1313,66 @@ void MenuBar::GenerateSymbolsFromRSOAuto()
...
-      // Get the beginning of the string
-      found_addr = accessors->Search(*found_addr, reinterpret_cast<const u8*>(""), 1, false);
-      if (!found_addr.has_value())
-        continue;
+      // Non-null data can precede the module name.
+      // Get the maximum name length that a module could have.
+      auto get_max_module_name_len = [found_addr] {
+        u32 len = 0;
+
+        while (true)
```
leoetlino, line 1384: The infinite loop here makes me a bit uncomfortable -- do we have an upper bound on the maximum length?

Source/Core/DolphinQt/MenuBar.cpp
```diff
@@ -1313,24 +1313,66 @@ void MenuBar::GenerateSymbolsFromRSOAuto()
...
+
+        // Get the field (Module Name Offset) that point to the string
+        const auto module_name_offset_addr =
+            accessors->Search(lookup_addr, ref.data(), ref.size(), false);
+        if (!module_name_offset_addr.has_value())
+          continue;
+
+        // The next 4 bytes should be the module name length
+        module_name_length = accessors->ReadU32(*module_name_offset_addr + 4);
+        if (module_name_length == ((max_name_length - i) + str.length()))
```
leoetlino, line 1429: Most of the parentheses are unnecessary and don't help with readability: `if (module_name_length == max_name_length - i + str.length())`

Source/Core/DolphinQt/MenuBar.cpp
InusualZ: The only limit that I can think of is the max value of a `uint` as that's the type of `module_name_length`. But, we could give a upper limit of `MAX_PATH`. Thoughts?

sepalani: Assuming that most module names are Windows paths, you could create a constant similar to `MAX_PATH` on Windows with a limit of 260 characters, I guess.

leoetlino: Using MAX_PATH=260 sounds fine to me.

InusualZ: I opted to name the constant `MODULE_NAME_MAX_LENGTH`, to avoid collision with the Windows's `MAX_PATH` constant

sepalani:
@dolphin-emu-bot rebuild

sepalani (approved):
LGTM. Tested on Monster Hunter 3, Kirby TV Channel and the System Menu 3.2E.

PR #9544, opened 2021-03-01
MemArena: Use memory placeholders for fastmem on Windows to ensure nothing allocates within the fastmem space.

AdmiralCurtiss:
This allocates and handles our fastmem address region as a placeholder on Windows. This avoids the possibility that something else allocates memory within the fastmem address region after it has been set up.

This should fix the bug (on Windows at least) where large texture packs break emulation, such as [12404](https://bugs.dolphin-emu.org/issues/12404) or [11994](https://bugs.dolphin-emu.org/issues/11994), but I cannot 100% confirm it.

- [x] Use old logic as fallback if VirtualAlloc2/MapViewOfFile3 are unavailable (as they would be on eg. Win7).
- [x] Make sure we load the VirtualAlloc2/MapViewOfFile3 from the correct dll. (see [Windows API sets](https://docs.microsoft.com/en-us/windows/win32/apiindex/windows-apisets?redirectedfrom=MSDN); maybe [this one...?](https://docs.microsoft.com/en-us/uwp/win32-and-com/win32-apis#apis-from-api-ms-win-core-memory-l1-1-6dll))
- [x] Memory cleanup on shutdown.
- [x] Error handling.
- [x] 32bit support.

References:
[Test code](https://gist.github.com/AdmiralCurtiss/0ba59f15710d919735f9f73ab20fb518)
[Docs on VirtualAlloc2](https://docs.microsoft.com/en-us/windows/win32/api/memoryapi/nf-memoryapi-virtualalloc2)
[Docs on VirtualFree](https://docs.microsoft.com/en-us/windows/win32/api/memoryapi/nf-memoryapi-virtualfree)
[Docs on MapViewOfFile3](https://docs.microsoft.com/en-us/windows/win32/api/memoryapi/nf-memoryapi-mapviewoffile3)
[Docs on UnmapViewOfFileEx](https://docs.microsoft.com/en-us/windows/win32/api/memoryapi/nf-memoryapi-unmapviewoffileex)

### DIFF
--- a/Source/Core/Common/MemArena.h
+++ b/Source/Core/Common/MemArena.h
@@ -4,15 +4,16 @@
 #pragma once
 
 #include <cstddef>
-
-#ifdef _WIN32
-#include <windows.h>
-#endif
+#include <vector>
 
 #include "Common/CommonTypes.h"
 
 namespace Common
 {
+#ifdef _WIN32
+struct WindowsMemoryRegion;
+#endif
+
 // This class lets you create a block of anonymous RAM, and then arbitrarily map views into it.
 // Multiple views can mirror the same section of the block, which makes it very convenient for
 // emulating memory mirrors.
@@ -99,7 +100,15 @@ public:
 
 private:
 #ifdef _WIN32
-  HANDLE hMemoryMapping;
+  WindowsMemoryRegion* EnsureSplitRegionForMapping(void* address, size_t size);
+  bool JoinRegionsAfterUnmap(void* address, size_t size);
+
+  std::vector<WindowsMemoryRegion> m_regions;
+  void* m_reserved_region = nullptr;
+  void* m_memory_handle = nullptr;
+  void* m_api_ms_win_core_memory_l1_1_6_handle = nullptr;
+  void* m_address_VirtualAlloc2 = nullptr;
+  void* m_address_MapViewOfFile3 = nullptr;
 #else
   int fd;
 #endif
--- a/Source/Core/Common/MemArenaWin.cpp
+++ b/Source/Core/Common/MemArenaWin.cpp
@@ -3,70 +3,410 @@
 
 #include "Common/MemArena.h"
 
+#include <algorithm>
 #include <cstddef>
 #include <cstdlib>
-#include <set>
 #include <string>
 
 #include <windows.h>
 
+#include "Common/Assert.h"
 #include "Common/CommonFuncs.h"
 #include "Common/CommonTypes.h"
+#include "Common/DynamicLibrary.h"
 #include "Common/Logging/Log.h"
 #include "Common/MsgHandler.h"
 #include "Common/StringUtil.h"
 
+using PVirtualAlloc2 = PVOID(WINAPI*)(HANDLE Process, PVOID BaseAddress, SIZE_T Size,
+                                      ULONG AllocationType, ULONG PageProtection,
+                                      MEM_EXTENDED_PARAMETER* ExtendedParameters,
+                                      ULONG ParameterCount);
+
+using PMapViewOfFile3 = PVOID(WINAPI*)(HANDLE FileMapping, HANDLE Process, PVOID BaseAddress,
+                                       ULONG64 Offset, SIZE_T ViewSize, ULONG AllocationType,
+                                       ULONG PageProtection,
+                                       MEM_EXTENDED_PARAMETER* ExtendedParameters,
+                                       ULONG ParameterCount);
+
+using PIsApiSetImplemented = BOOL(APIENTRY*)(PCSTR Contract);
+
 namespace Common
 {
-MemArena::MemArena() = default;
-MemArena::~MemArena() = default;
+struct WindowsMemoryRegion
+{
+  u8* m_start;
+  size_t m_size;
+  bool m_is_mapped;
+
+  WindowsMemoryRegion(u8* start, size_t size, bool is_mapped)
+      : m_start(start), m_size(size), m_is_mapped(is_mapped)
+  {
+  }
+};
+
+MemArena::MemArena()
+{
+  // Check if VirtualAlloc2 and MapViewOfFile3 are available, which provide functionality to reserve
+  // a memory region no other allocation may occupy while still allowing us to allocate and map
+  // stuff within it. If they're not available we'll instead fall back to the 'legacy' logic and
+  // just hope that nothing allocates in our address range.
+  DynamicLibrary kernelBase{"KernelBase.dll"};
+  if (!kernelBase.IsOpen())
+    return;
+
+  void* const ptr_IsApiSetImplemented = kernelBase.GetSymbolAddress("IsApiSetImplemented");
+  if (!ptr_IsApiSetImplemented)
+    return;
+  if (!static_cast<PIsApiSetImplemented>(ptr_IsApiSetImplemented)("api-ms-win-core-memory-l1-1-6"))
+    return;
+
+  const HMODULE handle = LoadLibraryW(L"api-ms-win-core-memory-l1-1-6.dll");
+  if (!handle)
+    return;
+
+  void* const address_VirtualAlloc2 = GetProcAddress(handle, "VirtualAlloc2FromApp");
+  void* const address_MapViewOfFile3 = GetProcAddress(handle, "MapViewOfFile3FromApp");
+  if (address_VirtualAlloc2 && address_MapViewOfFile3)
+  {
+    m_api_ms_win_core_memory_l1_1_6_handle = handle;
+    m_address_VirtualAlloc2 = address_VirtualAlloc2;
+    m_address_MapViewOfFile3 = address_MapViewOfFile3;
+  }
+  else
+  {
+    // at least one function is not available, use legacy logic
+    FreeLibrary(handle);
+  }
+}
+
+MemArena::~MemArena()
+{
+  ReleaseMemoryRegion();
+  ReleaseSHMSegment();
+  if (m_api_ms_win_core_memory_l1_1_6_handle)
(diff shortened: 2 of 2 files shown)


AdmiralCurtiss:
Alright, I think I'm at a point where I want some external input on this. Lemme have it.

And for reference, yes, this does appear to fix at least one variant of the large texture pack issue. Though it's hard to say exactly due to the sporadic nature of it.

Source/Core/Common/MemArena.cpp
```diff
@@ -29,6 +29,19 @@
 #endif
 #endif
 
+#ifdef _WIN32
+typedef PVOID(WINAPI* PVirtualAlloc2)(HANDLE Process, PVOID BaseAddress, SIZE_T Size,
```
Warepire, line 33: Perhaps put these in the header file so you don't need to cast the pointer every time you make call to one of these?

Warepire, line 33: Prefer `using` over `typedef`

Source/Core/Common/MemArena.cpp
```diff
@@ -67,12 +80,47 @@ static int AshmemCreateFileMapping(const char* name, size_t size)
...
+  if (!IsApiSetImplemented("api-ms-win-core-memory-l1-1-6"))
+    return;
+
+  HMODULE handle = LoadLibraryW(L"api-ms-win-core-memory-l1-1-6.dll");
+  if (!handle)
+    return;
+
+  m_address_VirtualAlloc2 = GetProcAddress(handle, "VirtualAlloc2FromApp");
+  m_address_MapViewOfFile3 = GetProcAddress(handle, "MapViewOfFile3FromApp");
+  if (!m_address_VirtualAlloc2 || !m_address_MapViewOfFile3)
```
Warepire, line 99: Will there be a problem if only one of them is null? Or is it even possible for one of them to be null?

Warepire (commented):
Perhaps something for later, it doesn't seem like many class methods with #ifdef in them share code, would it make sense to split the class into OS-specific implementations rather than having everything in the same implementation?

Source/Core/Common/MemArena.cpp
AdmiralCurtiss: I didn't do that to avoid including the horrible `Windows.h` in the header, but ymmv, especially given that it was already there before.

Source/Core/Common/MemArena.cpp
AdmiralCurtiss: It seems unlikely for that to ever be the case, but either way, having one without the other isn't useful, so I just fall back to the old code if either is not here.

Warepire: I got confused by your if-guards, perhaps it's worth adding a note or a bool based on the handle to clarify how it is checked.

Source/Core/Common/MemArena.cpp
Warepire: Good point, either is OK for me.

Source/Core/Common/MemArena.cpp
Warepire: Now it's very clear.

AdmiralCurtiss:
Okay, I've updated this:
- The MemArena has been split up into three separate files for each OS, which is indeed way cleaner.
- The Windows MemArena now handles the splitting/joining stuff internally, so you can just map and unmap wherever without any setup (as long as those regions are free, of course).

As far as I'm concerned, *this is done now*, but it still needs testing on an old Windows system that does not have VirtualAlloc2 and MapViewOfFile3. A Linux and Android implementation of a similar thing would also be good, but I'd leave that to a followup PR if that's okay.

Source/Core/Common/MemArenaAndroid.cpp
```diff
@@ -59,6 +59,12 @@ static int AshmemCreateFileMapping(const char* name, size_t size)
   return fd;
 }
 
+MemArena::MemArena()
```
leoetlino, line 62: Can this be `= default`?

Source/Core/Common/MemArenaUnix.cpp
```diff
@@ -22,6 +22,12 @@
 
 namespace Common
 {
+MemArena::MemArena()
```
leoetlino, line 25: = default?

Source/Core/Common/MemArenaWin.cpp
```diff
@@ -16,60 +17,395 @@
...
+  }
+};
+
 MemArena::MemArena()
 {
+  // Check if VirtualAlloc2 and MapViewOfFile3 are available, which provide functionality to reserve
+  // a memory region no other allocation may occupy while still allowing us to allocate and map
+  // stuff within it. If they're not available we'll instead fall back to the 'legacy' logic and
+  // just hope that nothing allocates in our address range.
+  HMODULE kernelBaseHandle = LoadLibraryW(L"KernelBase.dll");
```
leoetlino, line 53: Can we use DynamicLibrary (from Common) here instead?

AdmiralCurtiss: Oh, I didn't realize we had that! Yes, probably, let's see...

PatrickFerry:
I've asked a few people with the issue with texture packs to use this build and every one of them has had success in resolving the issue. They were all modern windows users though.

What's the most recent windows version that doesn't support VirtualAlloc2 and MapViewOfFile3?

Source/Core/Common/MemArenaUnix.cpp
```diff
@@ -0,0 +1,93 @@
...
+#include <string>
+
+#include <fcntl.h>
+#include <sys/mman.h>
+#include <unistd.h>
+
+#include "Common/CommonFuncs.h"
+#include "Common/CommonTypes.h"
+#include "Common/Logging/Log.h"
+#include "Common/MemArena.h"
```
Pokechu22, line 18: Since this file is implementing the class in `MemArena.h`, I think it would be best to have it at the top (that's what I did with #10270).  `MemArenaAndroid.cpp` already has it at the top, but `MemArenaWin.cpp` would need to be changed too.

Source/Core/Common/MemArenaAndroid.cpp
```diff
@@ -141,24 +100,10 @@ u8* MemArena::FindMemoryBase()
...
-    PanicAlertFmt("Failed to map enough memory space: {}", GetLastErrorString());
-    return nullptr;
-  }
-  VirtualFree(base, 0, MEM_RELEASE);
-  return base;
-#else
-#ifdef ANDROID
   // Android 4.3 changed how mmap works.
   // if we map it private and then munmap it, we can't use the base returned.
   // This may be due to changes in them support a full SELinux implementation.
```
Pokechu22, line 91: This isn't your comment, but I believe this is should be "due to changes in them <ins>to</ins> support a full SELinux implementation".

Source/Core/Common/MemArena.h
```diff
@@ -15,18 +15,57 @@ namespace Common
...
 // This class lets you create a block of anonymous RAM, and then arbitrarily map views into it.
 // Multiple views can mirror the same section of the block, which makes it very convenient for
-// emulating
-// memory mirrors.
+// emulating memory mirrors.
 class MemArena
 {
 public:
+  MemArena();
+  ~MemArena();
```
Pokechu22, line 24: Perhaps this should be `class MemArena final`?

Source/Core/Common/MemArenaWin.cpp
```diff
@@ -1,71 +1,410 @@
...
+    {
+      size_t mapped_region_count = 0;
+      if (r.m_is_mapped)
+        ++mapped_region_count;
+      if (mapped_region_count > 0)
+      {
+        PanicAlertFmt("Error while releasing fastmem region: {} regions are still mapped!",
+                      mapped_region_count);
+      }
+    }
```
Pokechu22, line 170: This logic is wrong; `mapped_region_count`'s declaration and the panic alert should be outside of the loop.  With the current logic, `mapped_region_count` will only ever be 0 or 1, and if multiple regions were not unmapped, you'll get multiple panic alerts.

Source/Core/Common/MemArenaWin.cpp
```diff
@@ -1,71 +1,410 @@
...
+
+    // update tracked mappings and return the first of the two
+    it->m_size = size;
+    u8* const new_mapping_start = address + size;
+    const size_t new_mapping_size = mapping_size - size;
+    regions.insert(it + 1, WindowsMemoryRegion(new_mapping_start, new_mapping_size, false));
+    return &regions[mapping_index];
+  }
+
+  assert(mapping_address < address);
```
Pokechu22, line 244: This should use Dolphin's `ASSERT` macro in `Common/Assert.h` instead (and the include of `cassert` should be removed).

Pokechu22:
Apart from my comments above, the first two commits look good.  I don't know enough about the Windows API to say that the 3rd commit is good, but nothing seems wrong with it either.

AdmiralCurtiss:
Appreciate it.

JMC47:
This has been sitting for a very long time and no one else has reviewed it.  Some of the most popular modern texture packs, like the Wind Waker Hytapia one.  I continue to see people on the unofficial discord running into this crash, so I think it's about time that we merged it.  I'll leave this for a few more hours to give people a chance, but if no one speaks up that they plan on reviewing it, I'll just merge.

Performance shouldn't be lower, and games seem to crash less.

Warepire (approved):

AdmiralCurtiss:
Much appreciated.

AhayriSG:
> ck if this works with Window

"I'll check if this works with Windows 7"

Minimum supported client | Windows 8 (https://docs.microsoft.com/en-us/windows/win32/api/memoryapi/nf-memoryapi-unmapviewoffileex)

image

PR #9545, opened 2021-03-01
IOS: Improve timing accuracy for title launches (both ARM and PPC)

leoetlino:
(alternative title: make ES_Launch slower and potentially break it)

This is a series of commits to make ES launch timings closer to console:

* In BootstrapPPC, we no longer neglect the amount of time it takes to read the DOL/ELF into memory.
* IOS and ES's slow boot time (>400ms) is now taken into account.
* IOS reloads are no longer instant. This fixes [reloadtest (from this HackMii post)](https://hackmii.com/2009/08/timing-is-everything-the-case-of-the-unsoftmoddable-wii/) as it makes the reasonable assumption that ES launches for IOS titles do not complete instantly.
* We now simulate what IOS does with /sys/launch.sys when launching a PPC title because FS operations that modify the NAND are particularly slow.
 
(those commits can be reviewed one at a time)

This PR also simplifies IPC reinitialisation after an IOS relaunch and makes it more accurate. Contrary to what the current code suggests, there is no complicated ack queue or anything. The extra IPC ack is actually triggered by a syscall that is invoked in ES's main function; the syscall literally just sets Y2, IX1 and IX2 in HW_IPC_ARMCTRL.

Needs careful testing because ES_Launch is cursed

### DIFF
--- a/Source/Core/Core/IOS/IOS.cpp
+++ b/Source/Core/Core/IOS/IOS.cpp
@@ -64,7 +64,6 @@ namespace IOS::HLE
 static std::unique_ptr<EmulationKernel> s_ios;
 
 constexpr u64 ENQUEUE_REQUEST_FLAG = 0x100000000ULL;
-constexpr u64 ENQUEUE_ACKNOWLEDGEMENT_FLAG = 0x200000000ULL;
 static CoreTiming::EventType* s_event_enqueue;
 static CoreTiming::EventType* s_event_sdio_notify;
 static CoreTiming::EventType* s_event_finish_ppc_bootstrap;
@@ -289,9 +288,6 @@ EmulationKernel::EmulationKernel(u64 title_id) : Kernel(title_id)
     return;
   }
 
-  // IOS re-inits IPC and sends a dummy ack during its boot process.
-  EnqueueIPCAcknowledgement(0);
-
   AddCoreDevices();
   AddStaticDevices();
 }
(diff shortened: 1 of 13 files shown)


Source/Core/Core/State.cpp
```diff
@@ -74,7 +74,7 @@ static Common::Event g_compressAndDumpStateSyncEvent;
 static std::thread g_save_thread;
 
 // Don't forget to increase this after doing changes on the savestate system
-constexpr u32 STATE_VERSION = 129;  // Last changed in PR 9511
+constexpr u32 STATE_VERSION = 130;  // Last changed in PR 953x
```
sepalani, line 77: Just a reminder as the PR number needs to be updated to 9545.

Source/Core/Core/IOS/IOS.cpp
```diff
@@ -137,6 +138,7 @@ static bool SetupMemory(u64 ios_title_id, MemorySetupType setup_type)
     return true;
   }
 
+  Memory::Memset(0, 0, 0x3fff);
```
BhaaLseN, line 141: Should we introduce a few named constants for those (or at least leave a comment)?

Source/Core/Core/State.cpp
BhaaLseN, line 77: Close enough, this is 9545 ;)

Source/Core/Core/IOS/IOS.cpp
leoetlino: Sure, I've added named constants for the region address and size + a comment.

Source/Core/Core/State.cpp
leoetlino: Whoops, good catch... Fixed in the last commit now

BhaaLseN:
Changes seem good, but this time I'll wait with approving until testing is done.

iwubcode (approved):
Wasn't 100% sure what to test so tested a handful of things:  wiiware titles, wii games, system menu and various channels, and a VC game.

BhaaLseN:
You still have `fixup!` commits in there btw.

leoetlino:
Squashed and rebased

BhaaLseN (approved):
Code LGTM, untested

JMC47:
I forgot to post I tested this with Ocarina of Time, Super Smash Bros. Brawl, and Metroid Prime Trilogy among other titles.

PR #9546, opened 2021-03-02
INI Changes

Miksel12:
Fixes:
https://bugs.dolphin-emu.org/issues/12437 (Also happens on Wii version of Barnyard)
https://bugs.dolphin-emu.org/issues/12419

Dualcore disabled for SNK Arcade Classics Volume 1 and The King of Fighters Collection: The Orochi Saga to fix  the clitching sprites.
EFB to RAM Battle of Giants: Dinosaurs Strike in order to display the player dinosaur texture.

### DIFF
--- a/Data/Sys/GameSettings/PM4.ini
+++ b/Data/Sys/GameSettings/PM4.ini
@@ -1,0 +1,5 @@
+# PM4E01 - Mario Kart: Double Dash‼ Bonus Disc
+
+[Core]
+MMU = True
+FPRF = True
--- a/Data/Sys/GameSettings/RFP.ini
+++ b/Data/Sys/GameSettings/RFP.ini
@@ -1,7 +1,13 @@
 # RFPE01, RFPJ01, RFPR01, RFPP01, RFPK01, RFPW01 - Wii Fit Plus and Wii Fit Plus Channel
 
+[Core]
+# Dualcore causes hangs in the Rhythm Parade mini game.
+CPUThread = False
+
 [Video_Hacks]
 # The Wii Fit Plus Channel shows Mii faces and a graph; however, EFBToTextureEnable must be disabled for this to work.
 # Since both the channel and the main game update these images, it must be disabled for both.  (They both have the same GameID, though.)
 # However, this has a performance impact, so if the Wii Fit Channel is not going to be used, this does not need to be disabled.
 #EFBToTextureEnable = False
+# Having EFBToTextureEnable disabled causes hangs in mini games if DeferEFBCopies is enabled.
+DeferEFBCopies = False
--- a/Data/Sys/GameSettings/GYA.ini
+++ b/Data/Sys/GameSettings/GYA.ini
@@ -14,3 +14,4 @@
 
 [Video_Hacks]
 EFBToTextureEnable = False
+DeferEFBCopies = False
--- a/Data/Sys/GameSettings/RBY.ini
+++ b/Data/Sys/GameSettings/RBY.ini
@@ -14,3 +14,4 @@
 
 [Video_Hacks]
 EFBToTextureEnable = False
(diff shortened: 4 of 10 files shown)


PatrickFerry:
I'd rather Hell's Kitchen get a different fix. Hell's Kitchen, I think, makes a case for partial XFB support. It's not just a one-off issue during loading screens in some games it's an issue throughout this game.

Miksel12:
I'm pretty sure something like partial XFB is non trivial to implement.

Data/Sys/GameSettings/RFP.ini
```diff
@@ -1,7 +1,13 @@
...
 
+[Core]
+# Dualcore causes hangs in the Rhythm Parade mini game.
+CPUThread = False
+
 [Video_Hacks]
 # The Wii Fit Plus Channel shows Mii faces and a graph; however, EFBToTextureEnable must be disabled for this to work.
 # Since both the channel and the main game update these images, it must be disabled for both.  (They both have the same GameID, though.)
 # However, this has a performance impact, so if the Wii Fit Channel is not going to be used, this does not need to be disabled.
 #EFBToTextureEnable = False
```
Rumi-Larry, line 11: Seems like there's a space missing after '#'

PatrickFerry:
Right, but I think this project has a tendency to now reject changes that affect the issue that are not the correct fix.

I'm OK with this if the developers are instead saying they will not write Partial XFB Support at any point down the line.

Data/Sys/GameSettings/RFP.ini
JosJuice: I believe that's intentional, as it's a commented-out line of code rather than a normal comment.

Miksel12:
> Right, but I think this project has a tendency to now reject changes that affect the issue that are not the correct fix.
> 
> I'm OK with this if the developers are instead saying they will not write Partial XFB Support at any point down the line.

XFB to RAM is correct as it emulates what the Wii/GC is doing. Partial XFB would be a nice addition but isn't more correct. If Dolphin was really going for correctness/accuracy than we XFB Real would still be an option.

PatrickFerry:
Yes, "correct" was the wrong word.

What I am trying to say is that if Partial XFB Support was to happen there would likely need to be a case to bother to do develop it. Initially it was believed to be only an issue in a couple games and even then they were just brief loading screens. There is now a lot more games that we know about that display purple/magenta screens and this one which uses it throughout the game. If this was resolved with XFB to Texture(and RAM) there would be less of a case for Partial XFB Support being developed.

JosJuice:
I really don't think that whether we set this option in the INI will affect how much motivation there is for developing partial XFB support. Just mark it with a comment so that we know that we can remove the line later if/once we set partial XFB support.

JMC47:
Partial XFB copies would be important to making this run at higher resolutions, so it's not a complete fix.  If anyone would be motivated by the games that require support, I think this wouldn't affect that.

PatrickFerry:
Alright, I withdraw my objection to that change then so.

JMC47:
These all look good to me after looking through the changes.

Data/Sys/GameSettings/PM4.ini
```diff
@@ -0,0 +1,5 @@
+# PM4E01 - Mario Kart: Double Dash‼ Bonus Disc
+
+[Core]
+MMU = True
+FPRF = True
```
SuperSamus, line 5: What's the motivation for these?

Miksel12: MMU is needed for Rebel Strike and FPRF is needed for F Zero. https://wiki.dolphin-emu.org/index.php?title=Mario_Kart:_Double_Dash%E2%80%BC_Bonus_Disc&diff=next&oldid=167651

PR #9548, opened 2021-03-02
Core/Memmap: Memory mapping logic fixes.

AdmiralCurtiss:
This fixes two potential (if unlikely to be relevant in practice) bugs:

- Remember which regions are mapped instead of re-checking via the flags each time. This fixes a potential issue where a memory region would not be unmapped when booting in Wii mode and then switching to GC mode by eg. using the Disc channel and a GC disc. (first commit)
- We didn't check which regions are active in `UpdateLogicalMemory()`, leading to an odd BAT config potentially attempting to map memory regions that haven't been allocated. (second commit)

### DIFF
--- a/Source/Core/Core/HW/Memmap.cpp
+++ b/Source/Core/Core/HW/Memmap.cpp
@@ -368,6 +368,9 @@ void UpdateLogicalMemory(const PowerPC::BatTable& dbat_table)
       u32 translated_address = dbat_table[i] & PowerPC::BAT_RESULT_MASK;
       for (const auto& physical_region : s_physical_regions)
       {
+        if (!physical_region.active)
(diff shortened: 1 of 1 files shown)


AdmiralCurtiss:
Oh, I also removed the branch for 32 bit builds, since I don't think we can build on those anymore anyway. I can readd if that's not the case.

JosJuice:
Dolphin still is supposed to build and run correctly on 32-bit systems. (32-bit performance isn't worth caring about, though.)

AdmiralCurtiss:
Fair enough then!

JMC47:
Is this good to go then?

leoetlino (approved):

PR #9549, opened 2021-03-02
JitArm64: Fix unsigned/signed argument/parameter mismatch

Dentomologist:
The result of calls to PPCSTATE_OFF_PS0/1 were being cast to u32 and passed to functions expecting s32 parameters. This changes the casts to s32 instead.

One location was missing a cast and generated a warning with VS which is now fixed.


JosJuice:
I don't think casting to `s32` inside the macro makes sense. There are situations where you might want the macro to return some other type, even if that isn't the case in Dolphin's code right now. Please replace and add casts as needed in code that uses the macro instead.

Dentomologist:
> I don't think casting to `s32` inside the macro makes sense. There are situations where you might want the macro to return some other type, even if that isn't the case in Dolphin's code right now. Please replace and add casts as needed in code that uses the macro instead.

Done.

JosJuice:
Please change the commit description – it still mentions casting inside the macro. LGTM otherwise.

JosJuice (approved):

PR #9550, opened 2021-03-03
SI/DeviceGBA: Ensure data socket isn't backed up

endrift:
When reading a reply from a message sent to the data socket there is the possibility that the other side gets sent multiple messages before replying to any of them, which can lead to multiple replies sent in a row. Though this only happens when things time out, it's quite possible for these timeouts to happen or build up over time, especially when initiating the connection.

This change makes sure to flush any pending bytes that have not been read yet out of the socket after a successful POLL reply is received, since that is the most common time when backups occur, and as well as using the exact number of bytes in an expected reply, to ensure the received data and the message it's replying to do not get out of sync.

---

An alternate approach not explored in this PR would be to modify `Receive` to read `bytes + 1` and do the flush if it receives `bytes + 1` instead of `bytes`, but I'm not sure which is better, if either.

### DIFF
--- a/Source/Core/Core/HW/SI/SI_DeviceGBA.cpp
+++ b/Source/Core/Core/HW/SI/SI_DeviceGBA.cpp
@@ -253,7 +253,7 @@ void GBASockServer::Send(const u8* si_buffer)
     Disconnect();
 }
 
-int GBASockServer::Receive(u8* si_buffer)
+int GBASockServer::Receive(u8* si_buffer, u8 bytes)
 {
   if (!m_client)
     return 0;
@@ -267,8 +267,7 @@ int GBASockServer::Receive(u8* si_buffer)
 
   size_t num_received = 0;
   std::array<u8, RECV_MAX_SIZE> recv_data;
-  sf::Socket::Status recv_stat =
-      m_client->receive(recv_data.data(), recv_data.size(), num_received);
+  sf::Socket::Status recv_stat = m_client->receive(recv_data.data(), bytes, num_received);
   if (recv_stat == sf::Socket::Disconnected)
   {
     Disconnect();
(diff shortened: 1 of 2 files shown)


JMC47:
The games that work still work.  The games that didn't work still don't work.  Seems fine to me.

endrift:
I'm worried it could make POLL commands to take longer, leading to lagging and timeout issues, but I'm not 100% sure that's realistic.

JMC47:
@dolphin-emu-bot rebuild

JMC47:
The website had a bit of an issue earlier today, maybe that's the cause of the build failure.

JMC47:
Tested this PR with various games.  It doesn't seem to affect compatibility.  FF:CC still broken.  Pokemon still finicky.  Everyone else seems to work fine.  Full testing list

Animal Crossing
Crash Bandicoot: Wrath of Cortex
Fire Emblem: Path of Radiance
The Legend of Zelda: Four Swords Adventures
The Legend of Zelda: The Wind Waker
Pokemon Colosseum
Pokemon XD
Mario Golf: Toadstool Tour
Mega Man X: Command Mission

I couldn't get Pokemon Box or Pokemon Bonus discs working in either master or this pull request.

Source/Core/Core/HW/SI/SI_DeviceGBA.cpp
```diff
@@ -329,7 +343,23 @@ int CSIDevice_GBA::RunBuffer(u8* buffer, int request_length)
 
   case NextAction::ReceiveResponse:
   {
-    int num_data_received = m_sock_server.Receive(buffer);
+    u8 bytes = 1;
+    switch (m_last_cmd)
+    {
+    case CMD_RESET:
+    case CMD_STATUS:
+      bytes = 3;
```
leoetlino, line 351: Is it worth defining named constants for these?

endrift: I'm not really sure. It''d just end up being e.g. CMD_RESET_BYTES_OUT, CMD_RESET_BYTES_IN, CMD_STATUS_BYTES_OUT, etc etc. Some could be reused with the transfer interval code but not all.

leoetlino (approved):

PR #9551, opened 2021-03-03
SI: Generate NOREP on timeout instead of generating Dolphin SDK reply

endrift:
I've noticed that NOREP timeouts aren't actually generated in Dolphin when there's no reply, instead favoring generating the reply that I _believe_ is actually created by the Dolphin SDK instead. I don't have proof that this behavior is correct, but I'm also not sure where to begin investigating.

### DIFF
--- a/Source/Core/Core/HW/SI/SI_DeviceGBA.cpp
+++ b/Source/Core/Core/HW/SI/SI_DeviceGBA.cpp
@@ -332,10 +332,9 @@ int CSIDevice_GBA::RunBuffer(u8* buffer, int request_length)
     }
     else
     {
-      u32 reply = Common::swap32(SI_ERROR_NO_RESPONSE);
-      std::memcpy(buffer, &reply, sizeof(reply));
-      return sizeof(reply);
+      return -1;
     }
+
     m_last_cmd = buffer[0];
     m_timestamp_sent = CoreTiming::GetTicks();
     m_next_action = NextAction::WaitTransferTime;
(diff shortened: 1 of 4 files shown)


Source/Core/Core/HW/SI/SI_DeviceGBA.cpp
```diff
@@ -306,11 +306,8 @@ int CSIDevice_GBA::RunBuffer(u8* buffer, int request_length)
       m_sock_server.Send(buffer);
     }
     else
-    {
-      u32 reply = Common::swap32(SI_ERROR_NO_RESPONSE);
-      std::memcpy(buffer, &reply, sizeof(reply));
-      return sizeof(reply);
-    }
+      return -1;
```
BhaaLseN, line 309: This hould still be braced since the `if` has curlies.

phire:
Well, we have two choices, hardware test, or just merge it and see if anything is reported as broken.

JMC47:
I don't know what it says about me that I feel like the second one would be more effective.  Every time I rely on hardware tests it turns out that the hardware test was missing some weird edge case.  When we merge without thinking, users report broken stuff, *then* we make a hardware test because we have to and it has more cases which is good.

Bonta0:
This PR with this commit https://github.com/Bonta0/dolphin/commit/820fb0c9aa719b902501fb9019afc8b99df341ad fix FFCC and 4 Swords+ not noticing disconnected GBAs. This has been confirmed to be HW accurate: COMERR should only be updated upon transfer completion
There are still cases where devices ``RunBuffer`` will return ``0`` by default. Those should probably be changed to return an error as well since returning 0 without setting up ``TransferInterval`` would only freeze Dolphin

endrift:
I'm going to cherry-pick that patch into this PR, but the `RunBuffer` issue should probably be handled separately.

JMC47:
So is this good to merge then?  I assume the windows buildbots are the way they are because of the lack of a rebase.

Hardware testing has been done.

PR #9553, opened 2021-03-03
Tools: Dolphin symbol map ghidra scripts added

sepalani:
This PR ports IDA scripts to ghidra.

### DIFF
--- a/Tools/ghidra/LoadDolphinMap.py
+++ b/Tools/ghidra/LoadDolphinMap.py
@@ -1,0 +1,55 @@
+# Copyright 2021 Dolphin Emulator Project
+# Licensed under GPLv2+
+# Refer to the license.txt file included.
+
+#@category Dolphin
+
+from collections import namedtuple
+
+
+DolphinSymbol = namedtuple("DolphinSymbol", [
+    "section", "addr", "size", "vaddr", "align", "name"
+])
+
+
+def load_dolphin_map(filepath):
+    with open(filepath, "r") as f:
+        section = ""
+        symbol_map = []
+        for line in f.readlines():
+            t = line.strip().split(" ", 4)
+            if len(t) == 3 and t[1] == "section" and t[2] == "layout":
+                section = t[0]
+                continue
+            if not section or len(t) != 5:
+                continue
+            symbol_map.append(DolphinSymbol(section, *t))
+        return symbol_map
+
+
+def ghidra_main():
+    from ghidra.program.model.symbol import SourceType
+    from ghidra.program.model.symbol import SymbolUtilities
+
+    f = askFile("Load a Dolphin emulator symbol map", "Load")
+    symbol_map = load_dolphin_map(f.getPath())
+
+    for symbol in symbol_map:
+        addr = toAddr(int(symbol.vaddr, 16))
+        size = int(symbol.size, 16)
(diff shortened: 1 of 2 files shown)


Tools/ghidra/LoadDolphinMap.py
```diff
@@ -0,0 +1,55 @@
...
+        name = SymbolUtilities.replaceInvalidChars(symbol.name, True)
+        if symbol.section in [".init", ".text"]:
+            createFunction(addr, symbol.name);
+            success = getFunctionAt(addr) is not None
+            if not success:
+                print("Can't apply properties for symbol:"
+                      " {0.vaddr} - {0.name}\n".format(symbol))
+                createLabel(addr, name, True)
+            else:
+                getFunctionAt(addr).setName(name, SourceType.USER_DEFINED)
```
Pokechu22, line 49: This should probably be `IMPORTED` (though I'm not really sure how much it matters).

Pokechu22 (commented):
Code looks good to me; however, Ghidra-GameCube-Loader already has its own [SymbolLoader](https://github.com/Cuyler36/Ghidra-GameCube-Loader/blob/master/src/main/java/gamecubeloader/common/SymbolLoader.java) and [SymbolMapExporter](https://github.com/Cuyler36/Ghidra-GameCube-Loader/blob/master/src/main/java/gamecubeloader/plugins/SymbolMapExporterPlugin.java).  I've only used the symbol loader, and IIRC it can only be run on first startup (which is a bit annoying), but I'm not quite sure what the advantage of having two implementations is (though I don't think there's too much of a downside to having our own implementation tied to Dolphin's symbol format either).

iwubcode:
As an aside, it'd be nice if someone knowledgeable could make a small README to describe what these scripts are for?  What is "IDA", what is "ghidra"?  When would I use them with symbols dumped from Dolphin?  etc

Pokechu22:
One other thought: the IDA version currently marks names staring with `zz_` as auto and weak:

https://github.com/dolphin-emu/dolphin/blob/7712f0831fdc1d573f0a0726ba86dd3ecdb9541e/Tools/IDA/LoadDolphinMap.py#L51-L56

I don't know if ghidra has an equivalent (the closest is `SourceType.ANALYSIS` but I don't know if that does anything really); it could also just not be imported as ghidra assigns its own names for those.

sepalani:
@iwubcode 
What would need to be done rather is writing, somewhere, on the wiki or something that serves the same purpose as an instruction leaflet: *how to use dolphin, its debuggers, its widgets and what are they for*.

Basically, if you know what a symbol map is for, the script name is quite self-explanatory. IDA/Ghidra are disassembler/decompiler/reverse engineering tools, quick search using the keyword "IDA/ghidra script" provides more details as well.

@Pokechu22 
The reason these scripts can be useful is because the gamecube loader doesn't support RAW analysis (with (A)RAM dump, for instance) nor ELF. I chose `SourceType.USER_DEFINED` because it has a higher priority:
https://ghidra.re/ghidra_docs/api/ghidra/program/model/symbol/SourceType.html
Unless, I misunderstood how the priority work I think `SourceType.USER_DEFINED` should be fine.

leoetlino:
Yeah, it's useful to be able to load a symbol map with a standalone script. I don't see any downside to merging this.

PR #9555, opened 2021-03-03
Android: Some JNI cleanup

JosJuice:
Details in the individual commits.


Source/Android/jni/AndroidCommon/IDCache.cpp
```diff
@@ -271,10 +271,6 @@ jmethodID GetBooleanSupplierGet()
 
 }  // namespace IDCache
 
-#ifdef __cplusplus
-extern "C" {
-#endif
-
 jint JNI_OnLoad(JavaVM* vm, void* reserved)
```
leoetlino, line 274: This should probably be marked JNIEXPORT?

Source/Android/jni/AndroidCommon/IDCache.cpp
```diff
@@ -271,10 +271,6 @@ jmethodID GetBooleanSupplierGet()
 
 }  // namespace IDCache
 
-#ifdef __cplusplus
-extern "C" {
```
leoetlino, line 275: Are you sure everything still works fine even with C++ symbol mangling? I'd expect the `extern "C"` to still be necessary...

JosJuice: Ah, oops, I was supposed to always include that and not never :(

leoetlino (approved):

PR #9556, opened 2021-03-03
CMake: Build with -std:c++latest for MSVC

JosJuice:
Making CMake match MSBuild in this aspect will hopefully reduce how often we accidentally break the MSVC CMake build.

Suggested by @shuffle2 as an alternative to PR #9457.

This PR actually replaces our current MSVC CMake build problem with another! A warning in qpixelformat.h (`operator '+': deprecated between enumerations of different types` on line 79) gets promoted to an error. Does anyone know why we are not getting this warning with MSBuild?


leoetlino:
> A warning in qpixelformat.h (`operator '+': deprecated between enumerations of different types` on line 79) gets promoted to an error. Does anyone know why we are not getting this warning with MSBuild?

Yes, that warning is disabled for Qt in the VS project files.

https://github.com/dolphin-emu/dolphin/commit/33c0abd1ab1444a70af649d80174ab8436b08300

leoetlino:
Anyway, I think I prefer this approach because it makes CMake builds consistent with msbuild and several people (myself included) preferred to keep a builder on the latest C++ standard to catch compatibility issues as they come rather than all at once every 3 years.

JosJuice:
Thanks, I've disabled that warning now. Unfortunately the MSVC CMake build is still broken ("Cannot open include file: 'wil/com.h'" in GameList.cpp line 59). This is due to PR #9438 and probably also exists in PR #9457 if you rebase that PR on latest master, so I think it would make sense to handle that in a separate PR.

leoetlino (approved):

PR #9557, opened 2021-03-03
PowerPC: fix WSACleanup() calls not mirroring WSAStartup()

Filoppi:
Now checking sockets against -1 for errors because Windows, Linux and Max OS all return -1 on failure, otherwise our gdb_active() check might return the wrong value.

Fixes https://bugs.dolphin-emu.org/issues/12429 and also netplay breaking after stopping the emulation for the first time.


leoetlino:
As stated on IRC, I'd prefer to see the WSA context wrapped into a RAII class (instead of duplicating the WSACleanup call and adding a manual check).

Filoppi:
I'm happy with this, don't think think it's worth making a RAII class to literally save 2 lines of code, and this will likely never be extended. In any case, I don't need need to take credit for fixing it so if you want to do that, you can make another PR.

JosJuice:
Closing in favor of PR #9559.

PR #9558, opened 2021-03-04
InputCommon: block on DSU servers collectively instead of individually

iwubcode:
While looking at the DSU server stuff for the GDB stub, I noticed that the DSU logic has each server waiting individually for a timeout.  That has the potential to cause issues, so instead we should wait once for all servers.

Alternatively if the thread won't be spin-waiting, it may be more beneficial to have these sockets not block at all.  However, I decided to keep the original blocking timeout for now.

### DIFF
--- a/Source/Core/InputCommon/ControllerInterface/DualShockUDPClient/DualShockUDPClient.cpp
+++ b/Source/Core/InputCommon/ControllerInterface/DualShockUDPClient/DualShockUDPClient.cpp
@@ -202,18 +202,6 @@ static bool IsSameController(const Proto::MessageType::PortInfo& a,
          std::tie(b.pad_id, b.pad_state, b.model, b.connection_type, b.pad_mac_address);
 }
 
-static sf::Socket::Status ReceiveWithTimeout(sf::UdpSocket& socket, void* data, std::size_t size,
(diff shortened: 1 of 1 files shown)


Source/Core/InputCommon/ControllerInterface/DualShockUDPClient/DualShockUDPClient.cpp
```diff
@@ -242,32 +230,48 @@ static void HotplugThreadFunc()
...
+    sf::SocketSelector selector;
+    for (auto& server : s_servers)
+    {
+      selector.add(server.m_socket);
+    }
+
+    using namespace std::chrono;
+    using namespace std::chrono_literals;
+    const auto timeout = s_next_listports - SteadyClock::now();
+    const auto timeout_ms = std::max(duration_cast<milliseconds>(timeout), 1ms);
```
Filoppi, line 242: Keep that old "// ReceiveWithTimeout treats a timeout of zero as infinite timeout, which we don't want" comment here

iwubcode:
Marked as draft for now, until I can test properly (after GDBStub gets fixed)

Filoppi:
Tested for multiple pads and servers. Works and the code is much better than before.

PR #9559, opened 2021-03-04
Common / Core: add raii object that cleans up WSA on destruction in gdb-stub

iwubcode:
(untested)

### DIFF
--- a/Source/Core/Core/PowerPC/GDBStub.cpp
+++ b/Source/Core/Core/PowerPC/GDBStub.cpp
@@ -4,6 +4,7 @@
 
 // Originally written by Sven Peter <sven@fail0verflow.com> for anergistic.
 
+#include <optional>
 #include <signal.h>
 #include <stdint.h>
 #include <stdio.h>
@@ -23,6 +24,7 @@ typedef SSIZE_T ssize_t;
 #endif
 
 #include "Common/Logging/Log.h"
+#include "Common/SocketContext.h"
 #include "Core/HW/CPU.h"
 #include "Core/HW/Memmap.h"
 #include "Core/Host.h"
@@ -30,6 +32,11 @@ typedef SSIZE_T ssize_t;
 #include "Core/PowerPC/Gekko.h"
 #include "Core/PowerPC/PPCCache.h"
 #include "Core/PowerPC/PowerPC.h"
+
+namespace
+{
+std::optional<Common::SocketContext> s_socket_context;
+}  // namespace
 
 #define GDB_BFR_MAX 10000
 #define GDB_MAX_BP 10
@@ -791,10 +798,6 @@ void gdb_handle_exception()
   }
 }
 
-#ifdef _WIN32
-WSADATA InitData;
-#endif
-
 // exported functions
 
 static void gdb_init_generic(int domain, const sockaddr* server_addr, socklen_t server_addrlen,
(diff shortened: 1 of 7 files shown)


iwubcode:
Alternative to #9557 @leoetlino

Source/Core/Core/PowerPC/GDBStub.cpp
```diff
@@ -31,6 +32,29 @@ typedef SSIZE_T ssize_t;
 #include "Core/PowerPC/PPCCache.h"
 #include "Core/PowerPC/PowerPC.h"
 
+namespace
+{
+#ifdef _WIN32
+class WSA_RAII
```
leoetlino, line 38: I'd put this in a separate header in Common/ (SocketContext?) and put the ifdefs around the function calls and the member, like this:

```cpp
namespace Common
{
class SocketContext
{
public:
#ifdef _WIN32
  SocketContext() { static_cast<void>(...); }
  ~SocketContext() { ... }
#else
  SocketContext() = default;
#endif
  // ...

private:
#ifdef _WIN32
  WSADATA m_data;
#endif
};
}  // namespace Common
```

The advantage of doing it like this is that it would abstract the WSA thing away from everything that just wants to use sockets. You would be able to remove all the ifdefs from GDBStub.cpp but also from IOS/Network/IP/Top.cpp.

Source/Core/Core/PowerPC/GDBStub.cpp
```diff
@@ -31,6 +32,29 @@ typedef SSIZE_T ssize_t;
...
+{
+public:
+  WSA_RAII() { (void)WSAStartup(MAKEWORD(2, 2), &m_data); }
```
leoetlino, line 41:
```suggestion
  WSA_RAII() { static_cast<void>(WSAStartup(MAKEWORD(2, 2), &m_data)); }
```

Source/Core/Core/PowerPC/GDBStub.cpp
```diff
@@ -31,6 +32,29 @@ typedef SSIZE_T ssize_t;
...
+  WSA_RAII(WSA_RAII&&) = delete;
+
+  WSA_RAII& operator=(const WSA_RAII&) = delete;
+  WSA_RAII& operator=(WSA_RAII&&) = delete;
+
+private:
+  WSADATA m_data;
+};
+
+std::unique_ptr<WSA_RAII> s_wsa;
```
leoetlino, line 54: A unique_ptr is kind of wasteful for this -- a std::optional would suffice if all you want is being able to construct/destruct at will.

Source/Core/Core/PowerPC/GDBStub.cpp
```diff
@@ -834,9 +854,8 @@ static void gdb_init_generic(int domain, const sockaddr* server_addr, socklen_t
                              sockaddr* client_addr, socklen_t* client_addrlen)
 {
 #ifdef _WIN32
-  WSAStartup(MAKEWORD(2, 2), &InitData);
+  s_wsa = std::make_unique<WSA_RAII>();
```
leoetlino, line 857: With the SocketContext thing mentioned above, you would be able to remove the ifdef here.

Source/Core/Core/PowerPC/GDBStub.cpp
```diff
@@ -883,10 +902,6 @@ void gdb_deinit()
     shutdown(sock, SHUT_RDWR);
     sock = -1;
   }
-
-#ifdef _WIN32
```
leoetlino, line 887: `s_wsa.reset();` (works with an std::optional too)

Source/Core/Core/PowerPC/GDBStub.cpp
iwubcode: Heh, I almost did that.  I just wasn't sure what would be better

Source/Core/Core/PowerPC/GDBStub.cpp
iwubcode: Nice.  Yeah, I thought of this too but wasn't sure what was more preferred (always having the object or only on windows).  I can definitely make this change!

leoetlino (approved):
LGTM. It might be worth using the wrapper in NetIPTopDevice too?

iwubcode:
@leoetlino - done.  Needs testing though

JMC47:
Resolves the DSU service issue as far as I can tell.  Booted up a bunch of games, closed, rebooted.  Device list correctly refreshes.

PR #9560, opened 2021-03-04
CMake: Include WIL headers

JosJuice:
MSBuild does this, so CMake should too. Fixes the MSVC build when combined with PR #9457 or PR #9556.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -805,6 +805,7 @@ else()
 endif()
 
 if (WIN32)
+  include_directories(Externals/WIL/include)
   include_directories(Externals/OpenAL/include)
 endif()
 
--- a/Source/Core/DolphinQt/GameList/GameList.cpp
+++ b/Source/Core/DolphinQt/GameList/GameList.cpp
@@ -2,6 +2,18 @@
 // Licensed under GPLv2+
 // Refer to the license.txt file included.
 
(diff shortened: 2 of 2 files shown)


CMakeLists.txt
```diff
@@ -805,6 +805,7 @@ else()
 endif()
 
 if (WIN32)
+  include_directories(Externals/WIL/include)
```
leoetlino, line 808: Apparently WIL has a CMakeLists.txt file. Does `add_subdirectory(Externals/WIL)` + `target_link_libraries` against WIL not work?

JosJuice: No, I got a whole load of errors with `add_subdirectory(Externals/WIL)`, starting on line 2 (`project(WIL)`). I'm not sure if the CMakeLists.txt file they provide is intended for use by projects including it as a library or only by them. They don't set `include_directories` (except in WIL/tests/CMakeLists.txt), for one.

leoetlino: Oh, turns out we are using an older version of WIL that doesn't have the target_include_directories call here :/

https://github.com/microsoft/wil/blob/86353933d5fba0caf61503c8b09e10e182dccdfb/CMakeLists.txt#L36-L40

leoetlino (approved):

PR #9563, opened 2021-03-04
BreakpointWidget: Emit BreakpointsChanged to update views

sepalani:
This PR fixes (M)BP icon not being updated when changed from the BreakpointWidget.

Ready to be review & merged.

### DIFF
--- a/Source/Core/DolphinQt/Debugger/BreakpointWidget.cpp
+++ b/Source/Core/DolphinQt/Debugger/BreakpointWidget.cpp
@@ -250,6 +250,7 @@ void BreakpointWidget::OnDelete()
   PowerPC::memchecks.Remove(address);
   Settings::Instance().blockSignals(false);
 
+  emit BreakpointsChanged();
   Update();
 }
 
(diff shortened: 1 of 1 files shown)


Source/Core/DolphinQt/Debugger/BreakpointWidget.cpp
```diff
@@ -250,6 +250,7 @@ void BreakpointWidget::OnDelete()
   PowerPC::memchecks.Remove(address);
   Settings::Instance().blockSignals(false);
 
+  emit BreakpointsChanged();
   Update();
```
leoetlino, line 254: Thoughts on connecting BreakpointsChanged to Update?

sepalani: If by that you mean moving `emit BreakpointsChanged()` in the `Update` function, it will update each view on `showEvent` even when breakpoints haven't changed since it calls `Update`.

leoetlino: No, I meant that emitting BreakpointsChanged would call Update. But after thinking about it more that'd lead to Update being called more than once if we added another change signal so forget about this suggestion :)

leoetlino (approved):

PR #9567, opened 2021-03-05
Fix various build warnings

Dentomologist:
Fix a number of build warnings on various platforms.

- Unused variables in Arm64Gen, FreeLookCamera and GraphicsWidget
- Variable shadowing in SDIOSlot0Device and NetworkCaptureLogger
- Missing maybe_unused attribute for unused but documenting constants in WiimoteDevice
- Missing override specifiers in Touchscreen subclasses and FreeLookCamera
- Spurious gcc warning about discarding return values with warn_unused_result attribute
- Signed/unsigned comparison in InputConfig
- Move Arm64Gen namespace variable into only consuming function (and make constexpr).

### DIFF
--- a/Source/Core/Core/IOS/USB/Bluetooth/WiimoteDevice.cpp
+++ b/Source/Core/Core/IOS/USB/Bluetooth/WiimoteDevice.cpp
@@ -725,11 +725,11 @@ void WiimoteDevice::SendConfigurationRequest(u16 cid, u16 mtu, u16 flush_time_ou
   SendCommandToACL(L2CAP_CONFIG_REQ, L2CAP_CONFIG_REQ, offset, buffer);
 }
 
-constexpr u8 SDP_UINT8 = 0x08;
-constexpr u8 SDP_UINT16 = 0x09;
+[[maybe_unused]] constexpr u8 SDP_UINT8 = 0x08;
+[[maybe_unused]] constexpr u8 SDP_UINT16 = 0x09;
 constexpr u8 SDP_UINT32 = 0x0A;
 constexpr u8 SDP_SEQ8 = 0x35;
-constexpr u8 SDP_SEQ16 = 0x36;
+[[maybe_unused]] constexpr u8 SDP_SEQ16 = 0x36;
 
 void WiimoteDevice::SDPSendServiceSearchResponse(u16 cid, u16 transaction_id,
                                                  u8* service_search_pattern,
--- a/Source/Core/Common/Arm64Emitter.cpp
+++ b/Source/Core/Common/Arm64Emitter.cpp
@@ -27,7 +27,6 @@ namespace Arm64Gen
 namespace
 {
 const int kWRegSizeInBits = 32;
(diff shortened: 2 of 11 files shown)


Dentomologist:
Ubuntu is still reporting an unused variable error for the structured bindings.  Looking into it again, it looks like the error might be a gcc bug in versions prior to 8.0.  If that's the case the maybe_unused attributes I added in those locations wouldn't be doing anything and I can revert them.

Dentomologist:
Other remaining warnings: 

- FreeLookCamera.cpp:315:75: warning: 'old_type' may be used uninitialized in this function [-Wmaybe-uninitialized]

old_type is initialized by m_current_type which is an optional. I don't know yet whether it's possible for that optional to be nullopt, but apparently the compiler thinks so.

- clang: warning: argument unused during compilation: '-fno-objc-exceptions' [-Wunused-command-line-argument]

There's a todo comment where that flag is set noting that it should only be set when compiling with objective c, but that conditional isn't implemented yet.

- /Users/osxguest/buildslave/pr-osx-x64/build/Source/Core/MacUpdater/Main.storyboard:global: warning: This file is set to build for a version older than the deployment target. Functionality may be limited. [9]
- Source/Core/DolphinQt/dolphin-emu_autogen/mocs_compilation.cpp:75:10: warning: non-portable path to file '"PUA4RWX4XS/moc_FreelookWidget.cpp"'; specified path differs in case from file name on disk [-Wnonportable-include-path]

Also the same error for "PUA4RWX4XS/moc_FreelookWindow.cpp" and "Y64OOCBRBB/moc_FreelookGeneral.cpp"

- ../Source/Core/VideoCommon/LightingShaderGen.h:40:23: warning: no previous extern declaration for non-static variable 's_lighting_struct' [-Wmissing-variable-declarations]

That variable is declared constexpr inline which some research seems to indicate should handle the problem, but apparently not.

- Not strictly a warning, but CMake on android complains about not having "Findfmt.cmake".

Source/Core/Core/IOS/USB/Bluetooth/WiimoteDevice.cpp
```diff
@@ -154,7 +154,7 @@ void WiimoteDevice::SetBasebandState(BasebandState new_state)
 
 const WiimoteDevice::SChannel* WiimoteDevice::FindChannelWithPSM(u16 psm) const
 {
-  for (auto& [cid, channel] : m_channels)
+  for ([[maybe_unused]] auto& [cid, channel] : m_channels)
```
leoetlino, line 157: This is actually considered a GCC bug: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=81767

I think this commit should be dropped as it is fixed in newer GCC versions and it hurts readability.

leoetlino (changes requested):

Dentomologist:
I've dropped the commit adding [[maybe_unused]] to structured bindings and rebased on master.

leoetlino (approved):

PR #9570, opened 2021-03-05
DiscIO: Fix reading certain WIA chunks with many exceptions

JosJuice:
The loop in `WIARVZFileReader::Chunk::Read` could terminate prematurely if the size argument was smaller than the size of an exception list which had only been partially loaded.

### DIFF
--- a/Source/Core/DiscIO/WIABlob.cpp
+++ b/Source/Core/DiscIO/WIABlob.cpp
@@ -653,7 +653,7 @@ bool WIARVZFileReader<RVZ>::Chunk::Read(u64 offset, u64 size, u8* out_ptr)
     return false;
   }
 
-  while (offset + size > m_out.bytes_written - m_out_bytes_used_for_exceptions)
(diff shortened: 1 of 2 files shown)


Source/Core/DiscIO/WIABlob.cpp
```diff
@@ -653,7 +653,11 @@ bool WIARVZFileReader<RVZ>::Chunk::Read(u64 offset, u64 size, u8* out_ptr)
     return false;
   }
 
-  while (offset + size > m_out.bytes_written - m_out_bytes_used_for_exceptions)
+  const auto out_bytes_written_excluding_exceptions = [this] {
```
leoetlino, line 656: Maybe `get_out_bytes_written_excluding_exceptions` to make it clearer it's a lambda and not an integer?

JosJuice: I've turned it into a member function `GetOutBytesWrittenExcludingExceptions` instead.

PR #9571, opened 2021-03-06
NetPlay: Sync more settings

JosJuice:
Fixes certain netplay desyncs, https://bugs.dolphin-emu.org/issues/10328, and a bug where the Wii language stored in a DTM file would be ignored. Prerequisite for PR #9388.

### DIFF
--- a/Source/Core/Core/BootManager.cpp
+++ b/Source/Core/Core/BootManager.cpp
@@ -433,13 +433,14 @@ bool BootCore(std::unique_ptr<BootParameters> boot, const WindowSystemInfo& wsi)
   if (StartUp.bWii && DiscIO::IsNTSC(StartUp.m_region) && Config::Get(Config::SYSCONF_PAL60))
     Config::SetCurrent(Config::SYSCONF_PAL60, false);
 
+  Core::UpdateWantDeterminism(/*initial*/ true);
+
   if (StartUp.bWii)
   {
-    const bool want_determinism = Movie::IsMovieActive() || NetPlay::IsNetPlayRunning();
-    Core::InitializeWiiRoot(want_determinism);
+    Core::InitializeWiiRoot(Core::WantsDeterminism());
 
     // Ensure any new settings are written to the SYSCONF
-    if (!want_determinism)
+    if (!Core::WantsDeterminism())
     {
       Core::BackupWiiSettings();
       ConfigLoaders::SaveToSYSCONF(Config::LayerType::Meta);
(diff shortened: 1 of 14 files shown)


Source/Core/Core/NetPlayClient.cpp
```diff
@@ -654,11 +654,18 @@ unsigned int NetPlayClient::OnData(sf::Packet& packet)
...
-      packet >> m_net_settings.m_ProgressiveScan;
-      packet >> m_net_settings.m_PAL60;
       packet >> m_net_settings.m_DSPEnableJIT;
       packet >> m_net_settings.m_DSPHLE;
       packet >> m_net_settings.m_WriteToMemcard;
+      packet >> m_net_settings.m_Mem1Size;
+      packet >> m_net_settings.m_Mem2Size;
+
+      {
+        int tmp;
```
leoetlino, line 664: Maybe `std::underlying_type_t<DiscIO::Region>` just in case it's changed?

Source/Core/Core/BootManager.cpp
```diff
@@ -433,11 +433,23 @@ bool BootCore(std::unique_ptr<BootParameters> boot, const WindowSystemInfo& wsi)
...
+    // Ensure any new settings are written to the SYSCONF
+    if (!want_determinism)
+    {
+      Core::BackupWiiSettings();
+      ConfigLoaders::SaveToSYSCONF(Config::LayerType::Meta);
+    }
+    else
+    {
+      ConfigLoaders::SaveToSYSCONF(Config::LayerType::Meta, [](const Config::Location& location) {
+        return Config::GetActiveLayerForConfig(location) >= Config::LayerType::Movie;
```
leoetlino, line 450: Could we add a comment to LayerType to mention that layer types must be listed in reverse search order? Otherwise it's not obvious that the enum order actually matters.

Source/Core/Core/BootManager.cpp
```diff
@@ -433,11 +433,23 @@ bool BootCore(std::unique_ptr<BootParameters> boot, const WindowSystemInfo& wsi)
   if (StartUp.bWii && DiscIO::IsNTSC(StartUp.m_region) && Config::Get(Config::SYSCONF_PAL60))
     Config::SetCurrent(Config::SYSCONF_PAL60, false);
 
-  // Ensure any new settings are written to the SYSCONF
   if (StartUp.bWii)
   {
-    Core::BackupWiiSettings();
-    ConfigLoaders::SaveToSYSCONF(Config::LayerType::Meta);
+    const bool want_determinism = Movie::IsMovieActive() || NetPlay::IsNetPlayRunning();
```
leoetlino, line 438: Is it possible to use Core::UpdateWantDeterminism(true) and Core::WantsDeterminism here?

edit: If we use UpdateWantDeterminism, the call would have to be outside the bWii conditional.

JosJuice: I've added a commit that does this.

leoetlino (approved):

PR #9572, opened 2021-03-07
VolumeVerifier: Align partition reads to groups

JosJuice:
Dependent on PR #8625.

This improves the speed of verifying Wii WIA/RVZ files. For me, the verification speed for LZMA2-compressed files has gone from 11-12 MiB/s to 13-14 MiB/s.

One thing VolumeVerifier does to achieve parallelism is to compute hashes for one chunk of data while reading the next chunk of data. In master, when reading data from a Wii partition, each such chunk is 32 KiB. This is normally fine, but with WIA and RVZ it leads to rather lopsided read times (without the compute times being lopsided): The first 32 KiB of each 2 MiB takes a long time to read, and the remaining part of the 2 MiB can be read nearly instantly. (The WIA/RVZ code has to read the entire 2 MiB in order to compute hashes which appear at the beginning of the 2 MiB, and then caches the result afterwards.) This leads to us at times not doing much reading and at other times not doing much computation. To improve this, this change makes us use 2 MiB chunks instead of 32 KiB chunks when reading from Wii partitions.

(block = 32 KiB, group = 2 MiB)

### DIFF
--- a/Source/Core/DiscIO/Volume.h
+++ b/Source/Core/DiscIO/Volume.h
@@ -126,7 +126,7 @@ public:
   virtual bool IsNKit() const = 0;
   virtual bool SupportsIntegrityCheck() const { return false; }
   virtual bool CheckH3TableIntegrity(const Partition& partition) const { return false; }
-  virtual bool CheckBlockIntegrity(u64 block_index, const std::vector<u8>& encrypted_data,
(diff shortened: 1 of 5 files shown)


lioncash (approved):

lioncash:
Feel free to merge once conflicts are fixed.

JosJuice:
Without merging PR #8625 first?

PR #9574, opened 2021-03-07
EnumFormatter: fix signed/unsigned comparison warnings

Pokechu22:
A mistake from #9497 &mdash; I added a unit test to catch this kind of issue, but didn't actually check whether it gave warnings on platforms other than MSVC (which it does).  See https://godbolt.org/z/PhMKo3.

### DIFF
--- a/Source/Core/Common/EnumFormatter.h
+++ b/Source/Core/Common/EnumFormatter.h
@@ -60,22 +60,23 @@ public:
   template <typename FormatContext>
   auto format(const T& e, FormatContext& ctx)
   {
-    const auto value = static_cast<std::underlying_type_t<T>>(e);
+    const auto value_s = static_cast<std::underlying_type_t<T>>(e);      // Possibly signed
+    const auto value_u = static_cast<std::make_unsigned_t<T>>(value_s);  // Always unsigned
+    const bool has_name = value_s >= 0 && value_u < size && m_names[value_u] != nullptr;
 
     if (!formatting_for_shader)
     {
-      if (value >= 0 && value < size && m_names[value] != nullptr)
-        return fmt::format_to(ctx.out(), "{} ({})", m_names[value], value);
+      if (has_name)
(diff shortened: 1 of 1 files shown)


Source/Core/Common/EnumFormatter.h
```diff
@@ -60,22 +60,22 @@ class EnumFormatter
...
   {
-    const auto value = static_cast<std::underlying_type_t<T>>(e);
+    const auto value_s = static_cast<std::underlying_type_t<T>>(e);
+    const auto value_u = static_cast<std::make_unsigned_t<T>>(value_s);
 
     if (!formatting_for_shader)
     {
-      if (value >= 0 && value < size && m_names[value] != nullptr)
-        return fmt::format_to(ctx.out(), "{} ({})", m_names[value], value);
+      if (value_s >= 0 && value_u < size && m_names[value_u] != nullptr)
```
Dentomologist, line 68: I'd suggest factoring this if expression into a bool above and using that bool here and on line 75.

Source/Core/Common/EnumFormatter.h
```diff
@@ -60,22 +60,23 @@ class EnumFormatter
   template <typename FormatContext>
   auto format(const T& e, FormatContext& ctx)
   {
-    const auto value = static_cast<std::underlying_type_t<T>>(e);
+    const auto value_s = static_cast<std::underlying_type_t<T>>(e);      // Possibly signed
+    const auto value_u = static_cast<std::make_unsigned_t<T>>(value_s);  // Always unsigned
+    const bool has_name = value_s >= 0 && value_u < size && m_names[value_u] != nullptr;
```
leoetlino, line 65: The value_s >= 0 check isn't actually useful for checking the bounds (because of the value_u < size), but I guess it doesn't hurt to check that the index is valid.

leoetlino (approved):

PR #9575, opened 2021-03-07
Fix FreeLook warnings

Filoppi:
m_enabled was never used.


Dentomologist:
FYI, these changes are also in #9567.  

We did them the same way (not that there's really another way to do them), so there shouldn't be a conflict whichever way these get merged.

iwubcode (approved):

leoetlino:
This has been fixed in #9567 (apologies for taking so long to merge it and causing duplicate work)

PR #3242, opened 2015-11-08
XFMem: Don't warn on writes of zero to unknown registers.

phire:
This gets rid of these extra warnings in FifoCI:

    ../Source/Core/VideoCommon/XFStructs.cpp:187 W[Video]: Unknown XF Reg: 1007=0
    ../Source/Core/VideoCommon/XFStructs.cpp:187 W[Video]: Unknown XF Reg: 1013=0
    ../Source/Core/VideoCommon/XFStructs.cpp:187 W[Video]: Unknown XF Reg: 1014=0
    ../Source/Core/VideoCommon/XFStructs.cpp:187 W[Video]: Unknown XF Reg: 1015=0
    ../Source/Core/VideoCommon/XFStructs.cpp:187 W[Video]: Unknown XF Reg: 1016=0
    ../Source/Core/VideoCommon/XFStructs.cpp:187 W[Video]: Unknown XF Reg: 1017=0
    ../Source/Core/VideoCommon/XFStructs.cpp:187 W[Video]: Unknown XF Reg: 1027=0
    ../Source/Core/VideoCommon/XFStructs.cpp:187 W[Video]: Unknown XF Reg: 1028=0
    ../Source/Core/VideoCommon/XFStructs.cpp:187 W[Video]: Unknown XF Reg: 1029=0
    ../Source/Core/VideoCommon/XFStructs.cpp:187 W[Video]: Unknown XF Reg: 102a=0
    ../Source/Core/VideoCommon/XFStructs.cpp:187 W[Video]: Unknown XF Reg: 102b=0
    ../Source/Core/VideoCommon/XFStructs.cpp:187 W[Video]: Unknown XF Reg: 102c=0
    ../Source/Core/VideoCommon/XFStructs.cpp:187 W[Video]: Unknown XF Reg: 102d=0
    ../Source/Core/VideoCommon/XFStructs.cpp:187 W[Video]: Unknown XF Reg: 102e=0
    ../Source/Core/VideoCommon/XFStructs.cpp:187 W[Video]: Unknown XF Reg: 102f=0
    ../Source/Core/VideoCommon/XFStructs.cpp:187 W[Video]: Unknown XF Reg: 1030=0
    ../Source/Core/VideoCommon/XFStructs.cpp:187 W[Video]: Unknown XF Reg: 1031=0
    ../Source/Core/VideoCommon/XFStructs.cpp:187 W[Video]: Unknown XF Reg: 1032=0
[...]

PR #9576, opened 2021-03-08
Add game quirks for unknown BP/CP/XF commands

Pokechu22:
See e.g. [bug 10931](https://bugs.dolphin-emu.org/issues/10931).  #9497 removed assertions for unknown CP commands, and there were only warnings for BP and XF before.  Since there are a few commands marked as maybe unknown, this is something worth tracking.

I have separate quirks for completely unknown CP commands (`USES_UNKNOWN_CP_COMMAND`) and CP commands where the lower nybble is not in the expected range (`USES_MAYBE_INVALID_CP_COMMAND`); this is because YAGCD disagrees with Dolphin's current implementation and it would be good to find out which is correct.  Based on bug 10931, Toy Story 3 will trigger both of those cases, but other games might only hit one.


Source/Core/Core/DolphinAnalytics.h
```diff
@@ -61,6 +61,15 @@ enum class GameQuirk
 
   USES_WD_UNIMPLEMENTED_IOCTL,
 
+  // Some games use invalid/unknown graphics commands (see e.g. bug 10931).
+  // These are different from unknown opcodes: it is known that BP/CP/XF is being used to,
```
leoetlino, line 65: is being used to what?

Pokechu22: er, I guess that should be "being written to", or perhaps just "being used".

Pokechu22:
I've created a separate quirk for 0x00/0x10/0x20, which are unknown commands that appear to be related to performance queries (they're used by e.g. Mario Golf ([bug 12461](https://bugs.dolphin-emu.org/issues/12461)).  Note that the SDK (and libogc, but perhaps unsurprisingly not anything by datel) issues 0x20 with argument 0 on startup, which I've ignored here as otherwise it would completely drown everything out.  Libogc also has other examples with both 0x20 and 0x00, but nothing with 0x10 (`GX_SetVCacheMetric` seems suspicious).  I haven't attempted to name them.  For performance reasons, uses of these commands are `DEBUG_LOG`ged so that there is no impact in release builds; I've left the other ones as warnings.

I've also added logging for inexact matches to `MATINDEX_A`/`MATINDEX_B`/`VCD_LO`/`VCD_HI` (0x30/0x40/0x50/0x60) rather than just reporting a quirk in that case, as looking at `__GXSetMatrixIndex` and `__GXSetVAT` in Super Mario Sunshine shows that those are used without any offset.

Source/Core/VideoCommon/XFStructs.cpp
```diff
@@ -192,6 +194,7 @@ static void XFRegWritten(int transferSize, u32 baseAddress, DataReader src)
     case 0x1017:
 
     default:
+      DolphinAnalytics::Instance().ReportGameQuirk(GameQuirk::USES_UNKNOWN_XF_COMMAND);
```
iwubcode, line 197: We want to report 0 here?  Any idea why it ignored for the warn?

Source/Core/VideoCommon/VertexLoaderManager.cpp
```diff
@@ -321,45 +323,96 @@ void LoadCPReg(u32 sub_cmd, u32 value, bool is_preprocess)
...
     if (update_global_state)
       VertexShaderManager::SetTexMatrixChangedA(value);
     break;
 
   case MATINDEX_B:
+    if (sub_cmd != MATINDEX_B)
+    {
+      DolphinAnalytics::Instance().ReportGameQuirk(GameQuirk::USES_MAYBE_INVALID_CP_COMMAND);
+      WARN_LOG_FMT(VIDEO,
+                   "CP MATINDEX_B: Command does not match exactly ({:02x} instead of {:02x})",
```
iwubcode, line 356: I'm gonna be really nitpicky here.  I feel like this is worded oddly.  The expectation and the value given seem backward.  Here's my take:

```
CP MATINDEX_B:  an exact value of {MATINDEX_0} was expected but instead a value of {sub_cmd} was seen
```

iwubcode:
I had no idea that Dolphin did tracking on this level, very cool.

Source/Core/VideoCommon/XFStructs.cpp
Pokechu22: Looks like 636bedb207784db2d58b9986464f6a863677b59e / #3242 hid those logs, specifically because of warnings on FifoCI, but doesn't specify what game did it.  Assuming it was a single game, IMO it's still useful to track it.

Source/Core/VideoCommon/VertexLoaderManager.cpp
Pokechu22: Changed.  It definitely was oddly worded, but I couldn't think of a better phrasing.

Source/Core/VideoCommon/XFStructs.cpp
Pokechu22: I've added c702180fe7e1a15421533fd1e477c5dfe1c79703 temporarily to get a list of these from FifoCI.

Pokechu22: Looks like the source of the problem is that the FIFO player copies all of the registers on startup, which includes the unused ones; this produces log messages for everything (but shouldn't affect normal playback).  1099475d461f859ebde6e01e871973d38b0d37bd is a hacky workaround where the unused ones just aren't copied, but I don't think it's a great implementation.

0x1007 in particular is odd, because YAGCD says it's `Perf1` (0x1006 is `Perf0` which Dolphin calls `XFMEM_SETGPMETRIC`); games don't seem to use it though.  It's marked as known in the actual XFMemory struct, but there is no constant for it.  And weirder still is this code (which seems very unlikely to be correct):

https://github.com/dolphin-emu/dolphin/blob/8d2b0fff8a066281365072f827321e47bfed8c72/Source/Core/VideoCommon/XFStructs.cpp#L41-L48

Pokechu22: Having looked at other parts of the code, it turns out `nextAddress` advances address forward, but also updates the count, so it's more of an optimisation that happens to be rather unnecessary in this case, but makes more sense for e.g. `XFMEM_SETPROJECTION` where usually all 7 registers are updated at once, but it would be bad to update `g_vertex_manager` et al 7 times for that.

My earlier hackfix was backwards (I was skipping everything *but* the unknown registers...); 7ea201cf1cedd305af7c3b6008b7e2e2eff9187a is a correct implementation.  Note that that produces [rendering differences](https://fifo.ci/version/ae0f6042606541a063df68227e63d7416b6f1601/) but it seems that those are only from timing changes (if instead of skipping unknown registers, the last valid register is repeated (2bd038e9d5852e0df6d62b2556365ce11f3c30d9), [there are no differences](https://fifo.ci/version/2bd038e9d5852e0df6d62b2556365ce11f3c30d9/)); since this also only happened on the first frame in the fifo player, I'm not going to worry about it.

With skipping unknown registers, it should be fine to log a warning even when it's set to 0; none of the testcases on fifo.ci have any unknown registers in the fifo itself.

JMC47:
Looks reasonable to me.  Not a coder ✔️

Source/Core/VideoCommon/VertexLoaderManager.cpp
```diff
@@ -321,45 +323,102 @@ void LoadCPReg(u32 sub_cmd, u32 value, bool is_preprocess)
...
+    break;
+
   case MATINDEX_A:
+    if (sub_cmd != MATINDEX_A)
+    {
+      DolphinAnalytics::Instance().ReportGameQuirk(GameQuirk::USES_MAYBE_INVALID_CP_COMMAND);
+      WARN_LOG_FMT(VIDEO,
+                   "CP MATINDEX_A: an exact value of {:02x} was expected "
+                   "but instead a value of {:02x} was seen",
+                   sub_cmd, MATINDEX_A);
```
leoetlino, line 345: Shouldn't this be MATINDEX_A and sub_cmd (in this order) instead?
```suggestion
                   MATINDEX_A, sub_cmd);
```

Source/Core/VideoCommon/VertexLoaderManager.cpp
```diff
@@ -321,45 +323,102 @@ void LoadCPReg(u32 sub_cmd, u32 value, bool is_preprocess)
...
     break;
 
   case MATINDEX_B:
+    if (sub_cmd != MATINDEX_B)
+    {
+      DolphinAnalytics::Instance().ReportGameQuirk(GameQuirk::USES_MAYBE_INVALID_CP_COMMAND);
+      WARN_LOG_FMT(VIDEO,
+                   "CP MATINDEX_B: an exact value of {:02x} was expected "
+                   "but instead a value of {:02x} was seen",
+                   sub_cmd, MATINDEX_B);
```
leoetlino, line 359: ditto

Source/Core/VideoCommon/VertexLoaderManager.cpp
```diff
@@ -321,45 +323,102 @@ void LoadCPReg(u32 sub_cmd, u32 value, bool is_preprocess)
...
     break;
 
   case VCD_LO:
+    if (sub_cmd != VCD_LO)  // Stricter than YAGCD
+    {
+      DolphinAnalytics::Instance().ReportGameQuirk(GameQuirk::USES_MAYBE_INVALID_CP_COMMAND);
+      WARN_LOG_FMT(VIDEO,
+                   "CP VCD_LO: an exact value of {:02x} was expected "
+                   "but instead a value of {:02x} was seen",
+                   sub_cmd, VCD_LO);
```
leoetlino, line 373: ditto

Source/Core/VideoCommon/VertexLoaderManager.cpp
```diff
@@ -321,45 +323,102 @@ void LoadCPReg(u32 sub_cmd, u32 value, bool is_preprocess)
...
     break;
 
   case VCD_HI:
+    if (sub_cmd != VCD_HI)  // Stricter than YAGCD
+    {
+      DolphinAnalytics::Instance().ReportGameQuirk(GameQuirk::USES_MAYBE_INVALID_CP_COMMAND);
+      WARN_LOG_FMT(VIDEO,
+                   "CP VCD_HI: an exact value of {:02x} was expected "
+                   "but instead a value of {:02x} was seen",
+                   sub_cmd, VCD_HI);
```
leoetlino, line 388: ditto

Source/Core/VideoCommon/VertexLoaderManager.cpp
Pokechu22: Yep.  Probably a leftover from the original message I used.

Source/Core/Core/DolphinAnalytics.cpp
```diff
@@ -133,20 +133,27 @@ void DolphinAnalytics::ReportGameStart()
...
+    "uses-DVDLowRequestDiscStatus",
+    "uses-DVDLowRequestRetryNumber",
+    "uses-DVDLowSerMeasControl",
+    "uses-different-partition-command",
+    "uses-di-interrupt-command",
+    "mismatched-gpu-texgens-between-xf-and-bp",
+    "mismatched-gpu-colors-between-xf-and-bp",
+    "uses-uncommon-wd-mode",
+    "uses-wd-unimplemented-ioctl",
+    "uses-unknown-bp-command",
```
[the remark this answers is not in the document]
Pokechu22: Probably not, at least for these strings, since the old ones can't be changed.  The programmer-facing enum could be adjusted, and there could be friendly names and categorizations on a site that actually displays the analytics though. (I don't think there's any site that actually displays them in a friendly way currently.)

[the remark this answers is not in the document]
leoetlino: Yeah I was thinking of doing some server-side postprocessing but I guess it's saner to just leave the old strings untouched...

PR #9577, opened 2021-03-09
DVDInterface: Make the WBFS/CISO hack only affect WBFS/CISO

JosJuice:
PR #8561 added a small hack to DVDInterface to keep WBFS and CISO files working with Nintendo's `Error #001` anti-piracy check. Unfortunately I don't think it's possible to support WBFS and CISO without any kind of hack or heuristic, but what we can do is replace the PR #8561 hack (which applies regardless of file format) with a hack that only is active when using WBFS or CISO.

This change is similar to PR #8557, but the disc size is calculated in a different way.

cc @Pokechu22


Source/Core/DiscIO/DiscUtils.h
```diff
@@ -0,0 +1,43 @@
...
+{
+class FileInfo;
+struct Partition;
+class Volume;
+
+constexpr u64 MINI_DVD_SIZE = 1459978240;  // GameCube
+constexpr u64 SL_DVD_SIZE = 4699979776;    // Wii retail
+constexpr u64 SL_DVD_R_SIZE = 4707319808;  // Wii RVT-R
+constexpr u64 DL_DVD_SIZE = 8511160320;    // Wii retail
+constexpr u64 DL_DVD_R_SIZE = 8543666176;  // Wii RVT-R
```
Pokechu22, line 23: Do gamecube devkit NR discs have a different size from retail discs?

JosJuice: They have the same size as far as I know.

Source/Core/DiscIO/DiscUtils.cpp
```diff
@@ -0,0 +1,203 @@
...
+
+  // If a partition doesn't seem to contain any valid data, skip it.
+  // This can happen when certain programs that create WBFS files scrub the entirety of
+  // the Masterpiece partitions in Super Smash Bros. Brawl without removing them from
+  // the partition table. https://bugs.dolphin-emu.org/issues/8733
+  constexpr u32 WII_MAGIC = 0x5D1C9EA3;
+  const auto it =
+      std::remove_if(partitions.begin(), partitions.end(), [&](const Partition& partition) {
+        return volume.ReadSwapped<u32>(0x18, partition) != WII_MAGIC;
+      });
```
Pokechu22, line 137: This constant duplicates `WII_DISC_MAGIC`.

Source/Core/Core/HW/DVD/DVDInterface.cpp
```diff
@@ -425,6 +427,33 @@ void Shutdown()
...
 
+static u64 GetDiscEndOffset(const DiscIO::VolumeDisc& disc)
+{
+  u64 size = disc.GetSize();
+
+  if (disc.IsSizeAccurate())
+  {
+    if (size == DiscIO::MINI_DVD_SIZE)
+      return DiscIO::MINI_DVD_SIZE;
+  }
```
Pokechu22, line 438: Why not just `return size` here?  Why explicitly check if it's the mini DVD size?

JosJuice: I suppose the biggest practical consequence of not rounding is that we would no longer show a panic alert if the user tries to run a 4 GiB dump of a 4.37 GiB game (due to not having merged the files outputted by CleanRip). But the user will get a message from the game itself about the disc read failing, so it's not too big of a deal.

Pokechu22: Ah, the deal with it is that `IsSizeAccurate` can be wrong if it's a bad/undersized dump; `IsSizeAccurate` indicates whether the _format_ uses/should use accurate size information, not whether the dump itself uses accurate size information.  Right?

Pokechu22 (approved):
Looks good to me.

There _is_ one remaining use of the Wii magic value, which could maybe be replaced except it's in a comment for future use:

https://github.com/dolphin-emu/dolphin/blob/72a6fff36c42989c71765012e26285943085b8c5/Source/Core/Core/IOS/DI/DI.cpp#L150-L163

It's probably not worth replacing the usage there, since it'd require either including DiscIO/DiscUtils.h or leaving the include missing; if/when that functionality is implemented the constant could be used.

Source/Core/Core/HW/DVD/DVDInterface.cpp
JosJuice: Yes, exactly.

PR #9578, opened 2021-03-10
Rename emulation speed limit hotkey to clarify that it must be held

Calinou:
It might not be obvious at a glance that this hotkey must be held to have an effect, rather than just pressed once like a toggle key. (I'm personally more used to Turbo keys working as a toggle, but I think the current behavior is fine too.)

### DIFF
--- a/Source/Core/Core/HotkeyManager.cpp
+++ b/Source/Core/Core/HotkeyManager.cpp
@@ -44,7 +44,10 @@ constexpr std::array<const char*, 125> s_hotkey_labels{{
 
     _trans("Decrease Emulation Speed"),
     _trans("Increase Emulation Speed"),
-    _trans("Disable Emulation Speed Limit"),
+    // i18n: Here, "(Hold)" refers to the fact that the key must be held
+    // continuously to have an effect, rather than pressing it once
+    // and pressing it again later.
+    _trans("Disable Emulation Speed Limit (Hold)"),
 
     _trans("Frame Advance"),
     _trans("Frame Advance Decrease Speed"),


Source/Core/Core/HotkeyManager.cpp
```diff
@@ -44,7 +44,10 @@ constexpr std::array<const char*, 125> s_hotkey_labels{{
 
     _trans("Decrease Emulation Speed"),
     _trans("Increase Emulation Speed"),
-    _trans("Disable Emulation Speed Limit"),
+    // i18n: Here, "(Hold)" refers to the fact that the key must be held
+    // continuously to have an effect, rather than pressing it once
+    // and pressing it again later.
+    _trans("Disable Emulation Speed Limit (Hold)"),
```
leoetlino, line 50: For the record, this will reset the hotkey binding for existing configs since the setting key has been changed. I personally don't think this is a major issue, but I'd like to hear what other people think.

MayImilae:
Is this really necessary? Any confusion about how it works will be resolved within a second of the user pressing it and seeing that it isn't a toggle. ┐(´-｀)┌

JMC47:
Do note that you can actually just turn it into a toggle anyway?

image

Calinou:
> Do note that you can actually just turn it into a toggle anyway?
> 
> image

I didn't see this was actually possible – this makes this PR irrelevant then :slightly_smiling_face:

1337haXXor:
It took me quite a while to find this, and only then in a passing comment on a reddit post, but I could not figure out *how* to get to those menus.

You apparently have to right click on one any of the inputs that you've set, and it brings up this entire other menu. I definitely could not find that documented anywhere significant, and it seems like it's a pretty important. I came here in anticipation of the RetroAchievements release for Wii, and was wondering how to toggle the speed turbo.

Maybe it would be valuable to add as a quick line on the Syntax page (https://wiki.dolphin-emu.org/index.php?title=Input_Syntax)?

PR #9580, opened 2021-03-11
[Updater] Check whether we've already had an update triggered

Antidote:
Currently Dolphin just assumes that it only has one update running, if a user were to "accidentally" check for updates, check the "Update when dolphin closes" checkbox multiple times then closes dolphin, you get a ton of updaters running at once.

The solution is to only ever allow dolphin to trigger the update once then wait for dolphin to close.

### DIFF
--- a/Source/Core/UICommon/AutoUpdate.cpp
+++ b/Source/Core/UICommon/AutoUpdate.cpp
@@ -31,6 +31,7 @@
 
 namespace
 {
+bool s_update_triggered = false;
 #ifdef _WIN32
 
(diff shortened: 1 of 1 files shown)


Source/Core/UICommon/AutoUpdate.cpp
```diff
@@ -31,6 +31,7 @@
 
 namespace
 {
+bool s_update_triggered = false;
```
leoetlino, line 34: (I was trying to see if we could avoid adding static state, but I guess spawning a process *is* modifying system global state anyway...)

leoetlino (approved):

PR #9581, opened 2021-03-11
Expose Control Expression variables to mappings UI

Filoppi:
-add a way to reset their value (from the mappings UI)
-fix "memory leak" where they would never be cleaned, one would be created every time you wrote a character after a "$"
-fix ability to create variables with an empty string by just writing "$" (+ added a parsing error for the case)
-add $ operator to the UI operators list, to expose this functionality even more

The UI will now show all the custom variables used.
Their value is kept as long as there is at least one mapping (control expression) referencing them. You can ever change devices or input profile and as long as the variable was and is still referenced, the value will be kept.
Still, the ability to reset values without restarting Dolphin or being forced to change their names can be useful.
UI is always kept to date with the actual values underneath.

The control ref variables were a bit obscure and confusing and hidden away from users, so this should also help with making them aware of the feature.

Tested everything. There should be no negative consequences to this.

Pic is from my work branch, but you can understand what changed:
unknown

### DIFF
--- a/Source/Core/DolphinQt/Config/Mapping/IOWindow.cpp
+++ b/Source/Core/DolphinQt/Config/Mapping/IOWindow.cpp
@@ -275,6 +275,7 @@ void IOWindow::CreateMainLayout()
     m_operators_combo->addItem(tr("^ Xor"));
   }
   m_operators_combo->addItem(tr("| Or"));
+  m_operators_combo->addItem(tr("$ User Variable"));
   if (m_type == Type::Input)
   {
     m_operators_combo->addItem(tr(", Comma"));
@@ -304,6 +305,15 @@ void IOWindow::CreateMainLayout()
   m_functions_combo->addItem(QStringLiteral("min"));
   m_functions_combo->addItem(QStringLiteral("max"));
   m_functions_combo->addItem(QStringLiteral("clamp"));
+
+  m_variables_combo = new QComboBoxWithMouseWheelDisabled(this);
+  m_variables_combo->addItem(tr("User Variables"));
+  m_variables_combo->setToolTip(
+      tr("User defined variables usable in the control expression.\nYou can use them to save or "
+         "retrieve values between\ninputs and outputs of the same parent controller."));
+  m_variables_combo->insertSeparator(m_variables_combo->count());
+  m_variables_combo->addItem(tr("Reset Values"));
+  m_variables_combo->insertSeparator(m_variables_combo->count());
 
   // Devices
   m_main_layout->addWidget(m_devices_combo);
@@ -366,6 +376,8 @@ void IOWindow::CreateMainLayout()
     button_vbox->addWidget(m_test_button);
   }
 
+  button_vbox->addWidget(m_variables_combo);
+
   button_vbox->addWidget(m_operators_combo);
 
   if (m_type == Type::Input)
@@ -425,8 +437,26 @@ void IOWindow::ConnectWidgets()
   connect(m_expression_text, &QPlainTextEdit::textChanged,
           [this] { UpdateExpression(m_expression_text->toPlainText().toStdString()); });
 
+  connect(m_variables_combo, qOverload<int>(&QComboBox::activated), [this](int index) {
+    if (index == 0)
+      return;
+
+    // Reset button. 1 and 3 are separators.
+    if (index == 2)
+    {
+      const auto lock = ControllerEmu::EmulatedController::GetStateLock();
+      m_controller->ResetExpressionVariables();
+    }
+    else
+    {
+      m_expression_text->insertPlainText(QLatin1Char('$') + m_variables_combo->currentText());
+    }
+
+    m_variables_combo->setCurrentIndex(0);
+  });
+
   connect(m_operators_combo, qOverload<int>(&QComboBox::activated), [this](int index) {
-    if (0 == index)
+    if (index == 0)
       return;
 
     m_expression_text->insertPlainText(m_operators_combo->currentText().left(1));
@@ -435,7 +465,7 @@ void IOWindow::ConnectWidgets()
   });
 
   connect(m_functions_combo, qOverload<int>(&QComboBox::activated), [this](int index) {
-    if (0 == index)
+    if (index == 0)
       return;
 
     m_expression_text->insertPlainText(m_functions_combo->currentText() + QStringLiteral("()"));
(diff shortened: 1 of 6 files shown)


Source/Core/DolphinQt/Config/Mapping/IOWindow.cpp
```diff
@@ -425,6 +437,23 @@ void IOWindow::ConnectWidgets()
   connect(m_expression_text, &QPlainTextEdit::textChanged,
           [this] { UpdateExpression(m_expression_text->toPlainText().toStdString()); });
 
+  connect(m_variables_combo, qOverload<int>(&QComboBox::activated), [this](int index) {
+    if (0 == index)
```
lioncash, line 441:
```suggestion
    if (index == 0)
```

To be consistent with the line below. We also generally don't invert conditionals like this in the codebase.

Source/Core/DolphinQt/Config/Mapping/IOWindow.cpp
```diff
@@ -425,6 +437,23 @@ void IOWindow::ConnectWidgets()
...
+      return;
+
+    if (index == 2)  // Reset button. 1 and 3 are separators.
+    {
+      const auto lock = ControllerEmu::EmulatedController::GetStateLock();
+      m_controller->ResetExpressionVariables();
+    }
+    else
+    {
+      m_expression_text->insertPlainText(QStringLiteral("$") + m_variables_combo->currentText());
```
lioncash, line 451:
```suggestion
      m_expression_text->insertPlainText(QLatin1Char('$') + m_variables_combo->currentText());
```

Source/Core/DolphinQt/Config/Mapping/IOWindow.cpp
```diff
@@ -563,6 +592,16 @@ void IOWindow::UpdateExpression(std::string new_expression, UpdateMode mode)
   const auto error = m_reference->SetExpression(std::move(new_expression));
   const auto status = m_reference->GetParseStatus();
   m_controller->UpdateSingleControlReference(g_controller_interface, m_reference);
+  
+  // This is the only place where we need to update the shown user variables:
+  while (m_variables_combo->count() > 4)  // Keep the previous items
```
lioncash, line 597: We generally prefer comments either above the line or within the statement, but not on the side, since they can make future refactoring a tiny bit more annoying if the line ever changes.

Source/Core/InputCommon/ControlReference/ExpressionParser.h
```diff
@@ -140,7 +140,14 @@ class ControlQualifier
 class ControlEnvironment
 {
 public:
-  using VariableContainer = std::map<std::string, ControlState>;
+  struct VariableState
+  {
+    VariableState() : state(0), reference_counter(0) {}
```
lioncash, line 145: The member variables can be initialized directly in the struct instead of in the initializer list, which cuts down on code a little bit.

Source/Core/DolphinQt/Config/Mapping/IOWindow.cpp
Filoppi: Actually I don't like the inverted condition check but that's how it's done in 2 places below that are doing the same exact thing. I was about to change them but then I left them and decided to keep consistency with them.

Source/Core/DolphinQt/Config/Mapping/IOWindow.cpp
Filoppi: The only comments I put on the right are generally "extra" comments that we could do without. That comment isn't worth taking a line of its own IMO.

Source/Core/DolphinQt/Config/Mapping/IOWindow.cpp
Filoppi: Ok i've changed the ones below as well, looks better now.

Source/Core/DolphinQt/Config/Mapping/IOWindow.cpp
Filoppi: removed all comments on the right of code.

Filoppi: If this is an actual rule, it should be here:
https://github.com/dolphin-emu/dolphin/blob/master/Contributing.md
but I think it's good to allow it in some cases (like this)

JMC47:
@dolphin-emu-bot rebuild

Source/Core/DolphinQt/Config/Mapping/IOWindow.cpp
```diff
@@ -305,6 +306,16 @@ void IOWindow::CreateMainLayout()
   m_functions_combo->addItem(QStringLiteral("max"));
   m_functions_combo->addItem(QStringLiteral("clamp"));
 
+  m_variables_combo = new QComboBoxWithMouseWheelDisabled(this);
+  m_variables_combo->addItem(tr("User Variables"));
+  m_variables_combo->setToolTip(
+      tr("User defined variables usable in the control expression.\nYou can use them to save or "
+         "retrive values,\nor even to share states between inputs and outputs.\nThey can only be "
```
JMC47, line 313: Should be "retrieve" instead of retrive.

Source/Core/DolphinQt/Config/Mapping/IOWindow.cpp
```diff
@@ -305,6 +306,16 @@ void IOWindow::CreateMainLayout()
   m_functions_combo->addItem(QStringLiteral("max"));
   m_functions_combo->addItem(QStringLiteral("clamp"));
 
+  m_variables_combo = new QComboBoxWithMouseWheelDisabled(this);
+  m_variables_combo->addItem(tr("User Variables"));
+  m_variables_combo->setToolTip(
+      tr("User defined variables usable in the control expression.\nYou can use them to save or "
```
JMC47, line 312: should probably be something more like.

"You can use them to save or retrieve values, and can even share states between inputs and outputs.

The double or, while technically correct in this particular case, reads weird.

Source/Core/DolphinQt/Config/Mapping/IOWindow.cpp
```diff
@@ -564,6 +595,16 @@ void IOWindow::UpdateExpression(std::string new_expression, UpdateMode mode)
   const auto status = m_reference->GetParseStatus();
   m_controller->UpdateSingleControlReference(g_controller_interface, m_reference);
 
+  // This is the only place where we need to update the user variables. keep the first 4 items.
```
JMC47, line 598: Should keep be capitalized?

Source/Core/InputCommon/ControllerEmu/ControllerEmu.h
```diff
@@ -191,6 +191,11 @@ class EmulatedController
   // which happens while handling a hotplug event because a control reference's State()
   // could be called before we have finished updating the reference.
   [[nodiscard]] static std::unique_lock<std::recursive_mutex> GetStateLock();
+  const ciface::ExpressionParser::ControlEnvironment::VariableContainer&
+  GetExpressionVariables() const;
+
+  // Resets the values while keeping the list
```
JMC47, line 197: Other comments you added have had a period and this one doesn't.

JMC47 (commented):
I can't really say too much about the code, but I went through it and it looked clean.  There were a few minor typos.  The only one that's probably major is the user facing one.

Filoppi:
> I can't really say too much about the code, but I went through it and it looked clean. There were a few minor typos. The only one that's probably major is the user facing one.

Fixed all

jordan-woyak (approved):
Code LGTM. Untested.

JMC47 (approved):

PR #9582, opened 2021-03-13
SI/DeviceGBA: Improve link stability

endrift:
Building on #9550 I found some more issues that could cause the link to get out of sync.

The original flush technique didn't work properly; it didn't work before the first RESET was sent, which generally meant the first RESET would be junk too. This replaces it with flushing _before_ sending any message. This approach isn't perfect but clears out the junk much more reliably.

This also adds a minor wait for the "non-booted" state (polling for status instead of actual data transfer) to improve the stability further.

### DIFF
--- a/Source/Core/Core/HW/SI/SI_DeviceGBA.cpp
+++ b/Source/Core/Core/HW/SI/SI_DeviceGBA.cpp
@@ -220,7 +220,11 @@ void GBASockServer::ClockSync()
 bool GBASockServer::Connect()
 {
   if (!IsConnected())
+  {
     m_client = GetNextSock();
+    if (m_client)
+      m_client->setBlocking(false);
+  }
   return IsConnected();
 }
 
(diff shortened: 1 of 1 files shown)


JMC47:
Fixes connecting to Final Fantasy: Crystal Chronicles.

Source/Core/Core/HW/SI/SI_DeviceGBA.cpp
```diff
@@ -258,12 +261,12 @@ int GBASockServer::Receive(u8* si_buffer, u8 bytes)
   if (!m_client)
     return 0;
 
+  sf::SocketSelector selector;
+  selector.add(*m_client);
   if (m_booted)
```
BhaaLseN, line 266: I wonder if its worth documenting why `m_booted` affects how long we can/should wait for a socket, and if those numbers have any significance (vs. being guesses that just work in practise)

endrift: Apparently the number didn't even work on Windows, so I replaced it with redefining what `m_booted` represents.

JMC47:
This does improve linking with Final Fantasy: Crystal Chronicles.  I do not know of any regressions.

BhaaLseN (approved):
Code looks good, but untested. Plus, I don't know enough of the protocol to say whether there's other things looming that might pop out later.

endrift:
I know I've been testing it and I think @JMC47 has been too, so "untested" might not be the most accurate description here.

BhaaLseN:
Untested by _me_ ;)

PR #9583, opened 2021-03-14
Fix loading save states with bluetooth passthrough

michaelszubartowicz:
When a save state is loaded, the IOS device serving bluetooth is cast as BluetoothEmuDevice. If, however, a real Wiimote with BT passthrough is used, this caused the game to crash. Now the proper device class is used.

### DIFF
--- a/Source/Core/Core/HW/Wiimote.cpp
+++ b/Source/Core/Core/HW/Wiimote.cpp
@@ -53,6 +53,9 @@ void UpdateSource(unsigned int index)
   if (!ios)
     return;
 
+  if (s_wiimote_sources[index] != WiimoteSource::Emulated)
+    return;
+
   const auto bluetooth = std::static_pointer_cast<IOS::HLE::BluetoothEmuDevice>(
       ios->GetDeviceByName("/dev/usb/oh1/57e/305"));
   if (!bluetooth)


Source/Core/Core/HW/Wiimote.cpp
```diff
@@ -53,12 +54,22 @@ void UpdateSource(unsigned int index)
   if (!ios)
     return;
 
-  const auto bluetooth = std::static_pointer_cast<IOS::HLE::BluetoothEmuDevice>(
-      ios->GetDeviceByName("/dev/usb/oh1/57e/305"));
-  if (!bluetooth)
-    return;
+  if (s_wiimote_sources[index] == WiimoteSource::Real)
+  {
+    const auto bluetooth = std::static_pointer_cast<IOS::HLE::BluetoothRealDevice>(
```
iwubcode, line 59: This branch seems unnecessary in its current state...it's accessing the bluetooth and doing nothing with it.

Suggest fixing it (not sure what the correct solution would be) or just removing it if there's nothing to do:

```
if (s_wiimote_sources[index] == WiimoteSource::Emulated)
{
    const auto bluetooth = std::static_pointer_cast<IOS::HLE::BluetoothEmuDevice>(
        ios->GetDeviceByName("/dev/usb/oh1/57e/305"));
    if (!bluetooth)
      return;
bluetooth->AccessWiimoteByIndex(index)->SetSource(GetHIDWiimoteSource(index));
  }
```

michaelszubartowicz:
Thanks for the quick reply. I fixed it.

Source/Core/Core/HW/Wiimote.cpp
```diff
@@ -53,12 +54,15 @@ void UpdateSource(unsigned int index)
   if (!ios)
     return;
 
-  const auto bluetooth = std::static_pointer_cast<IOS::HLE::BluetoothEmuDevice>(
-      ios->GetDeviceByName("/dev/usb/oh1/57e/305"));
-  if (!bluetooth)
-    return;
+  if (s_wiimote_sources[index] == WiimoteSource::Emulated)
```
leoetlino, line 57:
```suggestion
  if (s_wiimote_sources[index] != WiimoteSource::Emulated)
    return;
```

leoetlino (changes requested):

leoetlino (approved):
(I took the liberty of rebasing your PR and squashing the commits)

PR #9584, opened 2021-03-14
FreeLook: support exact rotation (allowing for more accurate headtracking)

iwubcode:
This adds support for Free Look to specify exact camera rotation in radians or via an input expression:

image

There are a couple primary use cases for this:

* In theory this has the ability to support something called ["FlickStick"](https://www.youtube.com/watch?v=C5L_Px3dFtE) .  When I started thinking of this feature, "FlickStick" was what I had mind.  It wasn't until I started my implementation that I found out a community had already came up with something similar!  Flickstick is fairly complex in nature but this feature can allow us to get a bit closer to it by allowing the controller's stick angle to act as the angle for Free Look
* In the DSU spec, technically rotations are supposed to be from a gyro (relative rotation in angular velocity) but this allows supporting other forms of input.  For example, my [OpenTrackDSU](https://github.com/iwubcode/OpenTrackDSUServer) was hacked to support relative values but [OpenTrack](https://github.com/opentrack/opentrack) actually supports exact pitch/yaw/roll.  So by providing this, the data will be more exact.

### DIFF
--- a/Source/Core/Core/FreeLookManager.cpp
+++ b/Source/Core/Core/FreeLookManager.cpp
@@ -220,7 +220,7 @@ void FreeLookController::Update()
   const auto gyro_motion_quat =
       Common::Quaternion::RotateXYZ(gyro_motion_rad_velocity_converted * dt);
 
-  g_freelook_camera.Rotate(gyro_motion_quat);
+  g_freelook_camera.RotateIncremental(gyro_motion_quat);
   if (m_move_buttons->controls[MoveButtons::Up]->GetState<bool>())
     g_freelook_camera.MoveVertical(-g_freelook_camera.GetSpeed() * dt);
 
--- a/Source/Core/Core/FreeLookManager.cpp
+++ b/Source/Core/Core/FreeLookManager.cpp
@@ -74,6 +74,16 @@ enum GyroButtons
   YawRight,
 };
 }
+
+namespace RotationButtons
+{
+enum RotationButtons
+{
+  Pitch,
+  Roll,
+  Yaw,
+};
+}
 }  // namespace
 
 FreeLookController::FreeLookController(const unsigned int index) : m_index(index)
@@ -106,6 +116,19 @@ FreeLookController::FreeLookController(const unsigned int index) : m_index(index
 
   groups.emplace_back(m_rotation_gyro = new ControllerEmu::IMUGyroscope(
                           _trans("Incremental Rotation"), _trans("Incremental Rotation")));
+
+  groups.emplace_back(m_exact_rotation_group = new ControllerEmu::ControlGroup(
+                          _trans("Exact Rotation"), _trans("Exact Rotation")));
+
+  m_exact_rotation_group->AddSetting(
+      &m_pitch, {_trans("Pitch"), _trans("rad"), _trans("The exact pitch of the camera")}, 0.0,
+      -MathUtil::TAU, MathUtil::TAU);
+  m_exact_rotation_group->AddSetting(
+      &m_yaw, {_trans("Yaw"), _trans("rad"), _trans("The exact yaw of the camera")}, 0.0,
+      -MathUtil::TAU, MathUtil::TAU);
+  m_exact_rotation_group->AddSetting(
+      &m_roll, {_trans("Roll"), _trans("rad"), _trans("The exact yaw of the camera")}, 0.0,
+      -MathUtil::TAU, MathUtil::TAU);
 }
 
 std::string FreeLookController::GetName() const
@@ -187,6 +210,8 @@ ControllerEmu::ControlGroup* FreeLookController::GetGroup(FreeLookGroup group) c
     return m_other_buttons;
   case FreeLookGroup::Rotation:
     return m_rotation_gyro;
+  case FreeLookGroup::RotationExact:
+    return m_exact_rotation_group;
   default:
     return nullptr;
   }
@@ -208,6 +233,17 @@ void FreeLookController::Update()
              .count();
   }
   m_last_free_look_rotate_time = std::chrono::steady_clock::now();
+
+  if (m_pitch.GetValue() != m_last_pitch || m_yaw.GetValue() != m_last_yaw ||
+      m_roll.GetValue() != m_last_roll)
+  {
+    m_last_pitch = m_pitch.GetValue();
+    m_last_yaw = m_yaw.GetValue();
+    m_last_roll = m_roll.GetValue();
+    g_freelook_camera.RotateExact(Common::Vec3{static_cast<float>(m_pitch.GetValue()),
+                                               static_cast<float>(m_yaw.GetValue()),
+                                               static_cast<float>(m_roll.GetValue())});
+  }
 
   const auto gyro_motion_rad_velocity =
       m_rotation_gyro->GetState() ? *m_rotation_gyro->GetState() : Common::Vec3{};
--- a/Source/Core/Core/FreeLookManager.cpp
+++ b/Source/Core/Core/FreeLookManager.cpp
@@ -114,7 +114,7 @@ FreeLookController::FreeLookController(const unsigned int index) : m_index(index
   m_fov_buttons->AddInput(ControllerEmu::Translate, _trans("Increase Y"));
   m_fov_buttons->AddInput(ControllerEmu::Translate, _trans("Decrease Y"));
 
-  groups.emplace_back(m_rotation_gyro = new ControllerEmu::IMUGyroscope(
+  groups.emplace_back(m_incremental_rotation_gyro = new ControllerEmu::IMUGyroscope(
                           _trans("Incremental Rotation"), _trans("Incremental Rotation")));
 
   groups.emplace_back(m_exact_rotation_group = new ControllerEmu::ControlGroup(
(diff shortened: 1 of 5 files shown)


iwubcode:
The main reason I wanted to support "FlickStick" (besides the extra mobility for a gamepad with gyro) was also because I thought it might be useful for games with a fixed camera (ex: Bomberman) where you might want to support a first-person view.  By mapping the FreeLook input to be the same as the controller, you can get a more useable first person solution!

@MichaelJW - let me know if this works for you!

---

Marked as draft for now, the relative and exact rotations don't play 100% nicely with each other (setting the exact value cancels the previous relative values out even if unrelated)

Source/Core/Core/FreeLookManager.cpp
```diff
@@ -125,6 +168,35 @@ void FreeLookController::LoadDefaults(const ControllerInterface& ciface)
                                       hotkey_string({"Shift", "`Axis Z+`"}));
   m_fov_buttons->SetControlExpression(FieldOfViewButtons::DecreaseY,
                                       hotkey_string({"Shift", "`Axis Z-`"}));
+
+#if defined HAVE_X11 && HAVE_X11
```
Rumi-Larry, line 172: Was this intended? Seems like you are checki6for HAVE_X11 twice.

iwubcode: Update the description, this is unrelated to this review.  Also, this is how it works in `Wiimote` code.

Rumi-Larry: Weird, but ok 🤷

JosJuice: The two checks are different – one is for `defined HAVE_X11` and one is for `HAVE_X11`. I would assume that checking just `HAVE_X11` would be enough, but if this is how we're doing it in other places in the code then I'd say it's okay to do the same here.

Source/Core/Core/FreeLookManager.h
```diff
@@ -4,23 +4,30 @@
...
 
 enum class FreeLookGroup
 {
   Move,
   Speed,
   FieldOfView,
-  Other
+  Other,
+  Rotation,
+  RotationExact,
```
Rumi-Larry, line 30: It might be better to rename the name of "Rotation" to "RotationIncremental" for consistency.

MichaelJW:
> The main reason I wanted to support "FlickStick" (besides the extra mobility for a gamepad with gyro) was also because I thought it might be useful for games with a fixed camera (ex: Bomberman) where you might want to support a first-person view. By mapping the FreeLook rotational input to be the same controller stick as the controller stick that handles movement, you can get a more useable first person solution!
> 
> @MichaelJW - let me know if this works for you!

Whoa, this is amazing! It works very well indeed, with **Camera 1:** `First Person`. Thank you for tagging me so I could check it out!

I haven't quite cracked how to record VR well yet, but here are a couple of videos if you're interested:

- https://youtu.be/9XPwOBpKzlg - holding the two virtual controllers still so you get an idea of how I'm moving my head
- https://youtu.be/ZMSciYtGe0g - similar but with the bound values visible

When I was testing your incremental rotation solution yesterday, I couldn't consciously notice the drift (except that when I returned my head to neutral I was no longer looking "forwards" in-game), but I did feel some VR nausea. With your new solution here I don't feel any nausea so far :)

I was surprised to notice in the videos that Yaw doesn't appear to do anything, because it feels natural with the headset on. (This made sense once I thought about it, though; the virtual screen is rotating so that does the job.) I see in the tooltip for First Person you say that "the camera can rotate and travel, but roll is impossible" - should this actually be "yaw is impossible"?

Also, I didn't notice a difference between Orbital and First Person, but I did find that Six Axis immediately went haywire, spinning all over the place very quickly.

Anyway, amazing work. I see you've also submitted #9587 which I'm very excited about, will check that out next.

iwubcode:
@MichaelJW - great to hear!  Thank  you for the videos.

>When I was testing your incremental rotation solution yesterday, I couldn't consciously notice the drift (except that when I returned my head to neutral I was no longer looking "forwards" in-game), but I did feel some VR nausea. With your new solution here I don't feel any nausea so far :)

Glad to hear you weren't seeing any drift with the other solution but so sorry for causing the nausea.  Keep me updated if you feel any problems with this solution.

> I see in the tooltip for First Person you say that "the camera can rotate and travel, but roll is impossible" - should this actually be "yaw is impossible"?

No it is roll.  Yaw would be equivalent to moving the head left to right (pitch is moving up and down, like a nodding rotation) and the extreme for roll would be if you went completely upside down, like break dancing on your head).

> Also, I didn't notice a difference between Orbital and First Person, but I did find that Six Axis immediately went haywire, spinning all over the place very quickly.

Yes, six-axis allows for roll but when moving either pitch or yaw, you will inadvertently get some rotation on the third axis (roll).  This is due to how we apply the rotation.  It may be a bug on our end but all the online resources seem to indicate this is normal behavior.

Orbital's difference would be more apparent if you move forward or backward.   Doing so, you should 'orbit' around a central point.  Not sure how useful it would be for VR.  It's meant for spectating games in netplay.

> Anyway, amazing work. I see you've also submitted #9587 which I'm very excited about, will check that out next.

Thanks again!  I'm actually reworking that one a bit.  So please wait a bit longer :).  I'll mention you when there's something finished.

Source/Core/DolphinQt/HotkeyScheduler.cpp
```diff
@@ -139,10 +139,22 @@ void HotkeyScheduler::Run()
...
       g_controller_interface.UpdateInput();
+      FreeLook::UpdateInput();
+    }
+
+    Common::SleepCurrentThread(1000 / 60);
+
+    g_controller_interface.SetCurrentInputChannel(ciface::InputChannel::Host);
+    g_controller_interface.UpdateInput();
 
     if (!HotkeyManagerEmu::IsEnabled())
```
Filoppi, line 159: HotkeyManagerEmu::IsEnabled() should not disable freelook IMO.

iwubcode: It doesn't.  It only disables the ability to toggle FreeLook on/off via hotkey

Filoppi: right, you move the freelook update above, nevermind.

Source/Core/DolphinQt/HotkeyScheduler.cpp
```diff
@@ -139,10 +139,22 @@ void HotkeyScheduler::Run()
...
-    if (Core::GetState() == Core::State::Uninitialized || Core::GetState() == Core::State::Paused)
+    if (now - m_last_freelook_ticked > interval)
+    {
+      m_last_freelook_ticked = now;
+      g_controller_interface.SetCurrentInputChannel(ciface::InputChannel::FreeLook);
       g_controller_interface.UpdateInput();
+      FreeLook::UpdateInput();
+    }
+
+    Common::SleepCurrentThread(1000 / 60);
```
Filoppi, line 154: If this sleeps the thread for ~16ms, how is freelook going to update at 200Hz? `if (now - m_last_freelook_ticked > interval)` is always gonna be true.

Source/Core/DolphinQt/HotkeyScheduler.cpp
```diff
@@ -139,10 +139,22 @@ void HotkeyScheduler::Run()
 
   while (!m_stop_requested.IsSet())
   {
-    Common::SleepCurrentThread(1000 / 60);
+    const u64 freelook_update_freq = 200;
+    const u64 interval = SystemTimers::GetTicksPerSecond() / freelook_update_freq;
+    const u64 now = CoreTiming::GetTicks();
 
-    if (Core::GetState() == Core::State::Uninitialized || Core::GetState() == Core::State::Paused)
+    if (now - m_last_freelook_ticked > interval)
```
Filoppi, line 146: Guessing it should be >= interval?

Source/Core/Core/FreeLookManager.h
iwubcode: Thank you!   Agreed.  I missed this in my initial pass but plan on fixing this in a subsequent pass.

Source/Core/DolphinQt/HotkeyScheduler.cpp
```diff
@@ -139,10 +139,22 @@ void HotkeyScheduler::Run()
 
   while (!m_stop_requested.IsSet())
   {
-    Common::SleepCurrentThread(1000 / 60);
+    const u64 freelook_update_freq = 200;
```
Filoppi, line 142: I'd move to a constexpr above, together with the hotkeys 60Hz update rate (I've also done the same change, it's nicer and highlights that there shouldn't be any negative consequence in changing them).

Source/Core/DolphinQt/HotkeyScheduler.cpp
```diff
@@ -139,10 +139,22 @@ void HotkeyScheduler::Run()
...
+    const u64 interval = SystemTimers::GetTicksPerSecond() / freelook_update_freq;
+    const u64 now = CoreTiming::GetTicks();
```
Filoppi, line 144: CoreTiming::GetTicks() read might not be atomic on all architectures? Why don't you hook this inside the emulation thread give it depends on emulation timings? It's weird it's in the hotkeys thread if then it needs to depend on the current game speed.

Source/Core/DolphinQt/HotkeyScheduler.cpp
Filoppi: Why is the refresh rate 200Hz? Maybe it should be hinted. Is it because you want the input polling to be as up to date as possible when a new game frame starts? Or can freelook actually output more frames that the game naturally would (I don't know how it works)?

MichaelJW:
> @MichaelJW - great to hear! Thank you for the videos. Are the white things the controllers?

Yes - each the size of a Wii remote, to give you a sense of scale.
 
> Glad to hear you weren't seeing any drift with the other solution but so sorry for causing the nausea. Keep me updated if you feel any problems with this solution.

Oh, no need to apologise! I only mentioned it to explain my appreciation for this exact rotate change.

> No it is roll. Yaw would be equivalent to moving the head left to right (pitch is moving up and down, like a nodding rotation) and the extreme for roll would be if you went completely upside down, like break dancing on your head).

Oh, duh. I just watched the video back and you're right, of course - I got the two mixed up.

> Yes, six-axis allows for roll but when moving either pitch or yaw, you will inadvertently get some rotation on the third axis (roll). This is due to how we apply the rotation. It may be a bug on our end but all the online resources seem to indicate this is normal behavior.

Just to clarify: as soon as I switch to six-axis in the menu, the game camera moves as if it's rotating all around at very very high speed - that's what I meant by going haywire. I've run out of time to get a video today but I'll see if I can recreate this tomorrow.

Actually, thinking about it, maybe I was still binding DSU/acceleration to Up/Down/Left/Right but was now passing position via those values instead. That would explain it! I'll test again with Move properly bound the next chance I get.

> Orbital's difference would be more apparent if you move forward or backward. Doing so, you should 'orbit' around a central point. Not sure how useful it would be for VR. It's meant for very casual spectating games in netplay.

Oh nice, that's a cool feature :)

iwubcode:
Finally had a chance to try this out with opentrack DSU.  Worked perfectly with the test tracker!  Very cool.  That opens up a whole bunch of inputs to control via DSU (ex: FreePie).

jordan-woyak:
Just commenting on your screenshot.
I don't know if radians should be exposed in the UI.

Current angular velocity inputs are in deg/s.
Radians make math nice but I don't think they are nice for mortals to look at.

iwubcode:
There were two reasons I chose radians both very related.

1) The other input was in radians.
2) Both controls are in radians because one type of input provides radians (DSU is scaled to radians)

Also keep in mind that the point of this wasn't to support users specifying exact values but to support the two use cases I mentioned.  However, I know a lot of users will likely want to enter exact angles to get the perfect screenshot.

Because of that, I am thinking about a UI that allows you to swap between "views".  Internally it would store radians but externally you could view as either.  I'll see about coming up with a proof-of-concept.

iwubcode:
This is superseded by #9671 .  Unfortunately, this doesn't support FlickStick or matching movement with rotations (ex: First person in bomberman).  However, I think I'll wait and revisit that at a later date...

PR #9585, opened 2021-03-14
JitArm64: Skip calculating carry flag when not needed

JosJuice:


JosJuice:
Actually... I was assuming that this would be faster on some CPUs, but now that I look, I can't find any specific CPUs where it would be faster. Should we do this anyway?

degasus:
Even if this has no effect on the execution units performance, it might have one on the out-of-order decoder. So yes, I think it is good not to write any state which we don't need.

degasus (approved):

Source/Core/Core/PowerPC/JitArm64/JitArm64_Integer.cpp
```diff
@@ -16,6 +16,15 @@
 
 using namespace Arm64Gen;
 
+#define CARRY_IF_NEEDED(inst_without_carry, inst_with_carry, ...)                                  \
```
lioncash, line 19: Is this not able to be made into a templated member function of JitArm64?

JosJuice: It would be rather annoying, because most of the functions we want to pass as the first two arguments are overloaded, and I'm not of any way that a templated function would be able to automatically determine the correct overload without the caller having to explicitly specify the intended overload.

PR #9586, opened 2021-03-14
JitArm64: Optimize general case of srawx

JosJuice:
Same approach as Jit64. A lot simpler, don't you think? :)


degasus (approved):
A lot better. Thank you.

PR #9587, opened 2021-03-14
FreeLook: allow for specifying the exact field of view multiplier for the camera

iwubcode:
This updates Dolphin to support specifying direct Free Look field of view multipliers via configuration or the UI using a 'Configure Options' button:

image

the settings are then exposed as spin-boxes:

image

This allows for game specific fov to be specified.  Also makes it easier to pinpoint the exact value you are interested in.

More options are planned in the future...


iwubcode:
Probably should call the options "multiplier" (to match the description).

Rumi-Larry:
Assuming the screenshots still represent the UI, I would slightly increase the width of the windows so the text isn't quite cramped.

iwubcode:
While changing the field of view to angles would require breaking configs, I feel like doing it now would make this PR larger than it needs to be.  I'm also still uncertain whether switching to angles makes sense from a game interaction perspective (even if it's nicer for the user).  Open to comments from the rest of the team.

In the meantime, I'll open this up as having the ability to directly set these (or set these in per-game) is useful.

iwubcode:
Field of view would likely be better as a graphics mod.  Closing.

PR #9588, opened 2021-03-14
JitArm64: Improve srawx special case carry calculation

JosJuice:
At a first glance it may look like a part of the code I added to srawx in efeda3b has a bug when a == s. The code actually happens to work correctly, but in the interest of making the code easier to reason about, I'd like to change the way it's implemented. This change should improve the pipelining a little in the a == s case too.


leoetlino (approved):
Personally it seemed clear that arithmetic right shifts preserved the sign bit, but why not.

PR #4164, opened 2016-08-30
VertexShaderGen: Move the sonic epsilon hack to the vertex shader.

Armada651:
In the vertex shader we have control over the depth clipping planes, so we don't have to offset every floating point value and lose accuracy. This fixes z fighting in the Metroid Prime visor. It also make slow depth a lot more accurate.

It has also been confirmed that the only reason the SW backend rendered the Sonic Adventure/Unleashed UIs correctly was because of its own version of the epsilon hack: https://github.com/dolphin-emu/dolphin/blob/master/Source/Core/VideoBackends/Software/TransformUnit.cpp#L60

We can't use the exact same hack as the SW backend because our depth comparison is done in floats instead of integers and is therefore sensitive to rounding errors introduced by adding an epsilon to all depth values.

---
This change is [Reviewable](https://reviewable.io/reviews/dolphin-emu/dolphin/4164)

PR #9591, opened 2021-03-16
VertexShaderGen: Sonic epsilon hack for OpenGL ES

blaahaj:
https://github.com/dolphin-emu/dolphin/pull/4164 moved the "Sonic epsilon hack" to vertex shaders. However, it was only done for targets with depth clamping. If this is not available, for example the target is OpenGL ES, the Sonic problem appears (https://bugs.dolphin-emu.org/issues/11897).

A version of the "Sonic epsilon hack" is added for targets without depth clamping.

### DIFF
--- a/Source/Core/VideoCommon/VertexShaderGen.cpp
+++ b/Source/Core/VideoCommon/VertexShaderGen.cpp
@@ -498,6 +498,12 @@ ShaderCode GenerateVertexShaderCode(APIType api_type, const ShaderHostConfig& ho
                 "o.clipDist1 = clipDist1;\n");
     }
   }
+  else
+  {
+    // Same depth adjustment for Sonic. Without depth clamping, it unfortunately
+    // affects non-clipping uses of depth too.
+    out.Write("o.pos.z = o.pos.z * (1.0 - 1e-7);\n");
+  }
 
   // Write the true depth value. If the game uses depth textures, then the pixel shader will
   // override it with the correct values if not then early z culling will improve speed.


Source/Core/VideoCommon/VertexShaderGen.cpp
```diff
@@ -498,6 +498,11 @@ ShaderCode GenerateVertexShaderCode(APIType api_type, const ShaderHostConfig& ho
                 "o.clipDist1 = clipDist1;\n");
     }
   }
+  else
+  {
+    // Same depth adjustment for Sonic. Unfortunately not just for clipping.
+    out.Write("o.pos.z = o.pos.z * (1.0 - 1e-7);\n");
```
Pokechu22, line 504: The "unfortunately not just for clipping" message doesn't entirely make sense to me.  Would something like "This hack needs to be applied even when depth control is not supported" make more sense?

blaahaj: "unfortunately" because it affects depth for purposes other than clipping too.

Pokechu22: Alright, maybe something like "Unfortunately, this path will affect depth for purposes beyond clipping; there is no way to affect clipping only without support for backend_depth_clamp" then?

(Also, you mention that depth clamping coincides with clip control [here](https://bugs.dolphin-emu.org/issues/11897#note-2); does this hack work correctly if `backend_clip_control` is supported but `backend_depth_clamp` isn't or if `backend_depth_clamp` is supported but `backend_clip_control` isn't?)

blaahaj: > you mention that depth clamping coincides with clip control

That may have been a mistake. I will check this again.

blaahaj: I made a mistake. It is independent of clip control. Perhaps I was confusing it with clip distance.

blaahaj: I updated the comment. I hope it is better.

Pokechu22 (approved):

JMC47:
This has been tested on multiple Sonic Games.  Appears to work.  It's a small change so if there is some kind of weird regression, it's easy to change back.  Looks good to me.

SN902:
This hack fixes black screens in Sonic Unleashed and Bpack Knight.

mirh:
https://stackoverflow.com/questions/7408855/clipping-planes-in-opengl-es-2-0
User-defined clipping can also be found natively in GLES devices supporting EXT_clip_cull_distance (my Adreno 630 phone does for example)

JMC47:
Oh that's neat!  Maybe we can fix this properly on GLES then.

blaahaj:
There already can be found a mention of this extension in the Dolphin source: https://github.com/dolphin-emu/dolphin/blob/5513d5f4f732fb1e436765ab87e7d60ba02b1ad6/Source/Core/VideoBackends/OGL/OGLRender.cpp#L506

But the comment implies there is a missing piece.

blaahaj:
> But does there exist an "extension for depth clamping"?

GL_EXT_depth_clamp. [It does not have a large number of supporting devices.](https://opengles.gpuinfo.org/listreports.php?extension=GL_EXT_depth_clamp)

mirh:
Because it was basically just shipped for mesa purposes https://github.com/KhronosGroup/OpenGL-Registry/pull/240
Not that it couldn't probably be implemented in most phones, but idk how much GLES is still being maintained there. 

Btw https://github.com/hrydgard/ppsspp/issues/11399 and https://github.com/hrydgard/ppsspp/pull/14042 sound interesting.

PR #9593, opened 2021-03-16
JitArm64: Constant carry flag optimizations

JosJuice:
If we know at compile time that the PPC carry flag definitely has a certain value, we can bake that value into the emitted code and skip having to read from PPCState.

### DIFF
--- a/Source/Core/Core/PowerPC/JitArm64/JitArm64_Integer.cpp
+++ b/Source/Core/Core/PowerPC/JitArm64/JitArm64_Integer.cpp
@@ -32,7 +32,7 @@ void JitArm64::ComputeRC0(u64 imm)
 
 void JitArm64::ComputeCarry(ARM64Reg reg)
 {
-  js.carryFlagSet = false;
+  js.carryFlag = CarryFlag::InPPCState;
 
   if (!js.op->wantsCA)
     return;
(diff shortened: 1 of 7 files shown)


Source/Core/Core/PowerPC/JitArm64/JitArm64_Integer.cpp
```diff
@@ -1103,19 +1207,65 @@ void JitArm64::addex(UGeckoInstruction inst)
...
+      // upload the carry state
       ARM64Reg WA = gpr.GetReg();
       LDRB(IndexType::Unsigned, WA, PPC_REG, PPCSTATE_OFF(xer_ca));
       CMP(WA, 1);
       gpr.Unlock(WA);
+      [[fallthrough]];
+    }
+    case CarryFlag::InHostCarry:
+    {
+      if (js.op->wantsCA)
```
degasus, line 1229: Isn't this redundant with #9585?

JosJuice: Oops, yes. I had a bunch of manual checks like these in this PR and thought I had removed them all after creating PR #9585, but seems like I missed this one.

degasus (approved):

PR #9594, opened 2021-03-18
InputCommon: avoid updating devices that aren't used

Filoppi:
This is an optimization.
All pointers to devices in Dolphin are std::shared_ptr.
ControlReferences keep a ptr to a device when they use an input (or output) from that device.
Some Mapping Widgets also keep ptrs to devices when they are open.

This means that we can safely skip updating devices that are currently not in use.
This would still update devices referenced by wiimotes when playing a GC game, but it's still better than nothing.

Also, updating devices tries to lock the devices mutex, and if it fails, it will skip that update. This should decrease the amount of time devices are locked for, and so the occurrence of devices updates being skipped.

A device update generally just queries the device and caches the new returned input values.
Unless some types of devices automatically disconnect when not queried for a long time, which I highly doubt, this should be absolutely fine. Sure in the first instance devices are updated again, a mouse device might return a huge relative movement, but that's also fine.

This is going to be fine even with the current implementations of relative input in a bunch of WIP PRs.
I was actually planning to get rid of shared_ptr for devices, given they are troubling, but even if we proceeded on that road, we could still implement a custom ref counter, if this turned out to be a worth optimization.

This would also help https://github.com/dolphin-emu/dolphin/pull/8747 as it updates devices at 200Hz, and it would be a shame to update devices which aren't used, especially if you have a lot connected to your machine.

Tested with XInput and Dinput and Keyboard/Mouse device on Windows.

### DIFF
--- a/Source/Core/InputCommon/ControllerInterface/ControllerInterface.cpp
+++ b/Source/Core/InputCommon/ControllerInterface/ControllerInterface.cpp
@@ -270,7 +270,13 @@ void ControllerInterface::UpdateInput()
   {
     std::lock_guard lk(m_devices_mutex, std::adopt_lock);
     for (const auto& d : m_devices)
-      d->UpdateInput();
+    {
+      // Don't update devices that aren't referenced anywhere else
+      if (d.use_count() > 1)
+      {
+        d->UpdateInput();
+      }
+    }
   }
 }
 


Filoppi:
To be tested with DSU, as it would accumulate a huge amount of packages on the socket.
If that is a problem, we could simply make a `virtual bool AlwaysUpdate()` method and set it to true on such devices.

jordan-woyak:
> To be tested with DSU, as it would accumulate a huge amount of packages on the socket.
> If that is a problem, we could simply make a `virtual bool AlwaysUpdate()` method and set it to true on such devices.

The "Wiimote" input backend will have the same problem.

Filoppi:
> > To be tested with DSU, as it would accumulate a huge amount of packages on the socket.
> > If that is a problem, we could simply make a `virtual bool AlwaysUpdate()` method and set it to true on such devices.
> 
> The "Wiimote" input backend will have the same problem.

Nevermind, the DSU client (dolphin) keeps telling the DSU server (e.g. DS4Window) to keep sending update packages within it's UpdateInput() call. If we don't update input, then packages won't be sent and they won't accumulate.

Would the Wiimote pile up BT packages or what? Would it disconnect if not updated?

jordan-woyak:
"a mouse device might return a huge relative movement"
It might be "fine", but is that really desirable? Maybe relative states should be cleared when it's determined a device will begin to receive regular "Update" calls. But that's something to worry about in your relative input PR I suppose.

Filoppi:
> "a mouse device might return a huge relative movement"
> It might be "fine", but is that really desirable? Maybe relative states should be cleared when it's determined a device will begin to receive regular "Update" calls. But that's something to worry about in your relative input PR I suppose.

Yeah I've already got the reset of relative inputs in my PR in all the necessary cases. But I've also converted the mouse in Windows to return global values and not relative ones from the last call. Anyway, something to not worry about now (for this PR).

Filoppi:
Under very specific (and unlikely) circumstances, this could mess with input detection.
Devices should be somewhat updated or reset when starting to be referenced again, and that road is just not worth pursuing for such a small gain. Too much code and testing involved.

PR #9595, opened 2021-03-19
VertexLoaderTest: Fix memset assignment warning

Dentomologist:
Testing a fix for Debian unit test warning -Wclass-memaccess.

### DIFF
--- a/Source/UnitTests/VideoCommon/VertexLoaderTest.cpp
+++ b/Source/UnitTests/VideoCommon/VertexLoaderTest.cpp
@@ -23,9 +23,7 @@ TEST(VertexLoaderUID, UniqueEnough)
   std::unordered_set<VertexLoaderUID> uids;
 
   TVtxDesc vtx_desc;
-  memset(&vtx_desc, 0, sizeof(vtx_desc));
   VAT vat;
-  memset(&vat, 0, sizeof(vat));
   uids.insert(VertexLoaderUID(vtx_desc, vat));
 
   vtx_desc.SetLegacyHex(0xFEDCBA9876543210ull);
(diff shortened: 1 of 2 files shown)


Pokechu22:
As an alternative to specifying default values in CPMemory.h, this seems to be valid (but is much more awkward looking than when used for single structs):

```C++
  TVtxDesc vtx_desc{.low{.Hex = 0}, .high{.Hex = 0}};
  VAT vat{.g0{.Hex = 0}, .g1{.Hex = 0}, .g2{.Hex = 0}};
```

Source/UnitTests/VideoCommon/VertexLoaderTest.cpp
```diff
@@ -50,8 +48,12 @@ class VertexLoaderTest : public testing::Test
     memset(input_memory, 0, sizeof(input_memory));
     memset(output_memory, 0xFF, sizeof(input_memory));
 
-    memset(&m_vtx_desc, 0, sizeof(m_vtx_desc));
-    memset(&m_vtx_attr, 0, sizeof(m_vtx_attr));
+    m_vtx_desc.low.Hex = 0;
+    m_vtx_desc.high.Hex = 0;
+
+    m_vtx_attr.g0.Hex = 0;
```
lioncash, line 54: This can be shortened to:

```suggestion
    m_vtx_attr = {};
```

Source/UnitTests/VideoCommon/VertexLoaderTest.cpp
```diff
@@ -50,8 +48,12 @@ class VertexLoaderTest : public testing::Test
     memset(input_memory, 0, sizeof(input_memory));
     memset(output_memory, 0xFF, sizeof(input_memory));
 
-    memset(&m_vtx_desc, 0, sizeof(m_vtx_desc));
-    memset(&m_vtx_attr, 0, sizeof(m_vtx_attr));
+    m_vtx_desc.low.Hex = 0;
```
lioncash, line 51: This can be shortened to:

```suggestion
    m_vtx_desc = {};
```

Dentomologist:
Changed m_vtx_attr and m_vtx_desc to use value initialization.

Source/UnitTests/VideoCommon/VertexLoaderTest.cpp
Dentomologist: This doesn't work because the copy assignment operator for BitField has been deleted.  

I tried creating a copy assignment operator for TVtxDesc and VAT to copy the respective member .Hex fields, but kept getting errors (I believe from the BitFieldArrays in TVtxDesc .low and .high) about the operation not being sane.

PR #9597, opened 2021-03-19
Open the properties window when alt+enter is used on the game list

Pokechu22:
This is a keyboard shortcut on Windows at least, and I've accidentally started games several times trying to use it in Dolphin.


MayImilae:
I'm hesitant to support the addition of a hotkey that isn't in the hotkey settings and can't be changed, especially when it overlaps one of our most used hotkeys! But the notion of "just add it to the hotkey settings" doesn't really fit here, since to support it in a user-understandable way would mean adding a game list tab to the hotkey settings. That sounds very excessive.

Well, as long as this doesn't impact the functionality of alt-enter for the game window (regular or render to main window), I think it's alright.

leoetlino (approved):

PR #9598, opened 2021-03-19
JitArm64: Fix broken format strings in Arm64RegCache

JosJuice:


BhaaLseN (approved):
Seems like a missed opportunity to switch em over to `ASSERT_MSG_FMT`, but I guess we can just as well put those in another PR instead.

JosJuice:
There is no `ASSERT_MSG_FMT`. A bug in GCC is blocking us from adding it.

PR #9600, opened 2021-03-19
Full GBA Controllers Integration with mGBA

Bonta0:
This replaces the current GBA devices implementation that relied on VBA-M connections and fully emulates them with libmgba instead. This one has multiple advantages: it works, it's deterministic, more convenient and makes it trivial to support SaveStates, NetPlay and Movies for those devices
image
image
New GBA profiles:
image
NetPlay: Device type can be configured where you would normally assign controllers
image
The host has the option to make players see only GBAs that correspond to a local pad since most games will expose secrets to the players on the GBA screen
image
Additional settings:
image
mGBA was added as a git submodule unlike other libraries in this project. If copying the entire tree is preferred for consistency I can do it but I believe this is more appropriate. A CMake option `USE_MGBA` was added to exclude the library if desired.
Rendering is only implemented in Qt.

### Known Issues:
- ~~DSP HLE: Booting with multiple GBAs will fail. This is a DSP issue, the same behavior would happen with VBA-M if you could somehow connect multiple instances simultaneously. In the meantime either use DSPLLE or connect GBAs manually one by one~~
- ~~Final Fantasy Crystal Chronicles: Game will randomly drop inputs when the minimap is active on the GBA. While this still uses the same inaccurate timings as the previous implementation, using more accurate ones does not fix it, I suspect that there are more inaccuracies in Dolphin's SI implementation that also need to be fixed~~
- ~~Sonic Adventure 2: Does not connect with Sonic Advance~~


Source/Core/InputCommon/InputConfig.h
```diff
@@ -25,7 +25,14 @@ class InputConfig
 
   ~InputConfig();
 
-  bool LoadConfig(bool isGC);
+  enum class InputType
```
Rumi-Larry, line 28: I would name this InputClass, type should be reserved for other subdivisions, like the different gc controllers.

Source/Core/Core/Movie.h
```diff
@@ -75,8 +75,8 @@ struct DTMHeader
   std::array<char, 6> gameID;  // The Game ID
   bool bWii;                   // Wii game
 
-  u8 controllers;  // Controllers plugged in (from least to most significant,
-                   // the bits are GC controllers 1-4 and Wiimotes 1-4)
+  u16 controllers;  // Controllers plugged in (from least to most significant,
+                    // the bits are GC controllers 1-4, Wiimotes 1-4 and GBAs 1-4)
```
JosJuice, line 79: This change breaks DTM comparability, as it causes other fields to be offset by one byte. Any new data you want to add needs to be placed in a byte that currently is reserved (and preferably `reserved` instead of `reserved3` since `reserved3` used to have a different meaning).

JMC47:
So Sonic Adventure not connecting is because the most common guide appears to be incorrect.  I need to get around to doing what the booklet says to do lol.

JMC47:
@endrift this might interest you as well

Source/Core/Core/HW/GBACore.cpp
```diff
@@ -0,0 +1,517 @@
...
+}
+
+Core::Core(int device_number, u64 gc_ticks)
+    : m_device_number(device_number), m_last_gc_ticks(gc_ticks)
+{
+  mLogSetDefaultLogger(&s_dummy_logger);
+  m_core = mCoreCreate(mPlatform::mPLATFORM_GBA);
+  m_core->init(m_core);
+
+  mCoreInitConfig(m_core, nullptr);
```
endrift, line 61: You should pass an actual string here, like "dolphin", to identify the frontend.

Source/Core/Core/HW/GBACore.cpp
```diff
@@ -0,0 +1,517 @@
...
+        break;
+      }
+
+      vf->close(vf);
+      vf = nullptr;
+    } while (unzGoToNextFile(zip) == UNZ_OK);
+    unzClose(zip);
+  }
+  return vf;
+}
```
endrift, line 210: mGBA has minizip integration built in already. Why did you add a second version?

Source/Core/Core/HW/GBACore.cpp
```diff
@@ -0,0 +1,517 @@
...
+
+void Core::RunUntil(u64 gc_ticks)
+{
+  const u64 gc_frequency = SystemTimers::GetTicksPerSecond();
+  mTimingSchedule(
+      m_core->timing, &m_event,
+      static_cast<s32>((gc_ticks - m_last_gc_ticks) * GBA_ARM7TDMI_FREQUENCY / gc_frequency));
+  m_waiting_for_event = true;
+
+  s32 start_time = mTimingCurrentTime(m_core->timing);
```
endrift, line 436: You may want to use `mTimingGlobalTime` which is a u64.

Source/Core/Core/HW/GBACore.h
```diff
@@ -0,0 +1,106 @@
...
+#include <mgba/core/core.h>
+#include <mgba/gba/interface.h>
+
+#include "Common/CommonTypes.h"
+
+class PointerWrap;
+
+namespace HW::GBA
+{
+static constexpr auto BITS_PER_SECOND = 115200;
```
endrift, line 27: This has been shown to be wrong--see #9552.

RinMaru:
could this be used to get Gameboy Player to work in the future?

Externals/mGBA/make_version.c.js
```diff
@@ -0,0 +1,134 @@
...
+var gitexe = GetGitExe();
+var commit      	= GetFirstStdOutLine(gitexe + cmd_commit);
+var commit_short    = GetFirstStdOutLine(gitexe + cmd_commit_short);
+var branch  		= GetFirstStdOutLine(gitexe + cmd_branch);
+var rev	        	= GetFirstStdOutLine(gitexe + cmd_rev);
+var tag 	    	= GetFirstStdOutLine(gitexe + cmd_tag);
+var binary_name     = "mgba";
+var project_name    = "mGBA";
+var version_abi     = "0.9";
+var version_string  = "0.9.0";
```
endrift, line 87: Yikes.

endrift (changes requested):
This seems _exceedingly fragile_ and inclined to break at any given moment when I update, well, pretty much anything upstream, especially since some of the APIs used are going to be changing in the future. While having it as an option seems neat, I'm _extremely unfond_ of wholesale replacing the old SI_DeviceGBA since it means that your version of Dolphin and your version of mGBA are tightly integrated. This has a huge number of issues, including causing a large burden on bug reporting for games that have been fixed upstream but not in your downstream, making packaging difficult for distros, and a lot more. Furthermore, as evidenced I've done a bunch of work lately on _fixing_ the old version to work better (with still more room to go) so this kind of feels like a slap in the face.

endrift:
> could this be used to get Gameboy Player to work in the future?

Short answer: not really, but it's complicated
Long answer: the biggest thing holding back GBP was a good way to attach libmgba, which this solves. I already have a lot of work on a branch that integrates GBP but its libmgba integration was rather poorly done, so having that done here would help, especially since I could pretty much file a PR immediately after rebasing (though sound will be garbage if I do it that way).

JosJuice:
I agree that we should keep the old way of connecting to a GBA emulator as an option.

endrift:
Especially since connecting via libmgba is enabled or disabled at compile time, removing the old approach seems like a bad idea.

Source/Core/Core/Config/MainSettings.cpp
```diff
@@ -147,6 +147,18 @@ const Info<std::string> MAIN_RESOURCEPACK_PATH{{System::Main, "General", "Resour
...
+// Main.GBA
+
+const Info<std::string> MAIN_GBA_BIOS_PATH{{System::Main, "GBA", "BIOS"}, ""};
+const std::array<Info<std::string>, 4> MAIN_GBA_ROM_PATHS{
+    Info<std::string>{{System::Main, "GBA", "Rom1"}, ""},
+    Info<std::string>{{System::Main, "GBA", "Rom2"}, ""},
+    Info<std::string>{{System::Main, "GBA", "Rom3"}, ""},
+    Info<std::string>{{System::Main, "GBA", "Rom4"}, ""}};
+const Info<std::string> MAIN_GBA_SAVES_PATH{{System::Main, "GBA", "SavesPath"}, ""};
+const Info<bool> MAIN_GBA_SAVES_IN_ROM_PATH{{System::Main, "GBA", "SavesInRomPath"}, false};
```
Rumi-Larry, line 159: What is the distinction between the first saves and the saves in ROM? right now it seems like the first contains or includes the second

Source/Core/Core/HW/GBACore.cpp
```diff
@@ -0,0 +1,517 @@
...
+    return;
+
+  std::string rom_path =
+      Config::Get(Config::MAIN_GBA_ROM_PATHS[static_cast<size_t>(m_device_number)]);
+  if (!rom_path.empty())
+  {
+    if (!LoadROM(rom_path.c_str()))
+      return;
+
+    std::string save_path = fmt::format("{}_{}.sav", rom_path.substr(0, rom_path.find_last_of('.')),
```
endrift, line 76: Although mGBA doesn't have per-player saves yet, other places that have similar infixes use `-`, so you might want to do that here for when mGBA adds them.

Source/Core/AudioCommon/AudioCommon.cpp
```diff
@@ -66,13 +65,10 @@ void InitSoundStream()
...
   }
+}
 
-  // Ideally these two calls would be done in AudioInterface::Init so that we don't
-  // need to have a dependency on AudioInterface here, but this has to be done
-  // after creating g_sound_stream (above) and before starting audio dumping (below)
-  g_sound_stream->GetMixer()->SetDMAInputSampleRate(AudioInterface::GetAIDSampleRate());
-  g_sound_stream->GetMixer()->SetStreamInputSampleRate(AudioInterface::GetAISSampleRate());
-
+void PostInitSoundStream()
```
BhaaLseN, line 70: I wonder if this could/should be in its own PR, since it addresses a comment thats been around for quite a while.

Source/Core/Core/HW/SI/SI.cpp
```diff
@@ -368,8 +369,35 @@ void DoState(PointerWrap& p)
...
+}
+
+static void RegisterEvents()
+{
+  s_change_device_event = CoreTiming::RegisterEvent("ChangeSIDevice", ChangeDeviceCallback);
+  s_tranfer_pending_event = CoreTiming::RegisterEvent("SITransferPending", RunSIBuffer);
+
+  for (int i = 0; i < MAX_SI_CHANNELS; ++i)
+    s_device_events[i] =
+        CoreTiming::RegisterEvent(fmt::format("SIEventChannel{}", i), DeviceEventCallback);
```
BhaaLseN, line 384: Since this spans multiple lines, please add some curlies:
```suggestion
  for (int i = 0; i < MAX_SI_CHANNELS; ++i)
  {
    s_device_events[i] =
        CoreTiming::RegisterEvent(fmt::format("SIEventChannel{}", i), DeviceEventCallback);
  }
```

Source/Core/Core/HW/SI/SI.cpp
```diff
@@ -381,7 +409,9 @@ void Init()
     {
       s_desired_device_types[i] = SIDEVICE_NONE;
 
-      if (Movie::IsUsingPad(i))
+      if (Movie::IsUsingGBA(i))
+        s_desired_device_types[i] = SIDEVICE_GC_GBA;
```
BhaaLseN, line 413: We prefer balanced curlies, and since the `else` has some the `if` should have them as well.
```suggestion
      if (Movie::IsUsingGBA(i))
      {
        s_desired_device_types[i] = SIDEVICE_GC_GBA;
      }
```

Source/Core/Core/Movie.cpp
```diff
@@ -463,10 +474,10 @@ void ChangeWiiPads(bool instantly)
 
   for (int i = 0; i < MAX_WIIMOTES; ++i)
     if (WiimoteCommon::GetSource(i) != WiimoteSource::None)
-      controllers |= (1 << i);
+      controllers |= (1 << (i + 4));
```
BhaaLseN, line 477: I know this isn't originally your code, but would you mind adding curlies here around the `for`? That would make it more obvious which statements belong to the loop.

Source/Core/Core/Movie.h
BhaaLseN: Maybe `reserved3` should've been documented then :)
But thats for another PR.

Source/Core/DolphinQt/GBAWidget.cpp
```diff
@@ -0,0 +1,334 @@
...
+  resize(width, height);
+  setVisible(visible);
+
+  QSettings& settings = Settings::GetQSettings();
+  auto key = QString::fromStdString(fmt::format("gbawidget/geometry{}", m_geometry_slot + 1));
+  if (settings.contains(key))
+    restoreGeometry(settings.value(key).toByteArray());
+  else
+    move(x() - (frameGeometry().width() / (m_geometry_slot & 1 ? -2 : 2)),
+         y() - (frameGeometry().height() / (m_geometry_slot & 2 ? -2 : 2)));
```
BhaaLseN, line 79: Curlies please, since the `else` spans over multiple lines.
```suggestion
  if (settings.contains(key))
  {
    restoreGeometry(settings.value(key).toByteArray());
  {
  else
  {
    move(x() - (frameGeometry().width() / (m_geometry_slot & 1 ? -2 : 2)),
         y() - (frameGeometry().height() / (m_geometry_slot & 2 ? -2 : 2)));
  }
```

Source/Core/DolphinQt/Settings/GameCubePane.cpp
```diff
@@ -390,5 +498,17 @@ void GameCubePane::SaveSettings()
...
+  // GBA Settings
+  Config::SetBaseOrCurrent(Config::MAIN_GBA_THREADS, m_gba_threads->isChecked());
+  Config::SetBaseOrCurrent(Config::MAIN_GBA_BIOS_PATH, m_gba_bios_edit->text().toStdString());
+  Config::SetBaseOrCurrent(Config::MAIN_GBA_SAVES_IN_ROM_PATH, m_gba_save_rom_path->isChecked());
+  Config::SetBaseOrCurrent(Config::MAIN_GBA_SAVES_PATH, m_gba_saves_edit->text().toStdString());
+  File::SetUserPath(F_GBABIOS_IDX, Config::Get(Config::MAIN_GBA_BIOS_PATH));
+  File::SetUserPath(D_GBASAVES_IDX, Config::Get(Config::MAIN_GBA_SAVES_PATH));
+  for (size_t i = 0; i < m_gba_rom_edits.size(); ++i)
+    Config::SetBaseOrCurrent(Config::MAIN_GBA_ROM_PATHS[i],
+                             m_gba_rom_edits[i]->text().toStdString());
```
BhaaLseN, line 511: Curlies.
```suggestion
  for (size_t i = 0; i < m_gba_rom_edits.size(); ++i)
  {
    Config::SetBaseOrCurrent(Config::MAIN_GBA_ROM_PATHS[i],
                             m_gba_rom_edits[i]->text().toStdString());
  }
```

Source/Core/Core/HW/GBACore.cpp
```diff
@@ -0,0 +1,517 @@
...
+void Core::RunFor(u64 gc_ticks)
+{
+  RunUntil(m_last_gc_ticks + gc_ticks);
+}
+
+void Core::DoState(PointerWrap& p)
+{
+  Flush();
+
+  p.Do(m_video_buffer);
```
BhaaLseN, line 455: Any chance to save states should also increment the version over in `State.cpp`:
https://github.com/dolphin-emu/dolphin/blob/bcd572a820427af472147f6238e907ec2ec2cab4/Source/Core/Core/State.cpp#L77

BhaaLseN (commented):
I'm pleasently surprised how little code this needed. Also, welcome to the project Bonta _(first you conquered ALTTPR, now you conquer emulation? Nice.)_

I do wonder if we should keep around the other GBA implementation for..._something_. Mainly because it had the loose coupling of a GBA implementation (vs. locking us into a specific version of mGBA.)
If we do, I feel like we should make this the user-facing default tho, since it has the It-Just-Works™ factor many end-users seek.

JMC47:
I do think we keep around the old path if possible, but at the same time there is a level of usability, especially on Windows, that is lacking with the old protocol.  Linux definitely performs better than Windows with the old setup with separate emulators (verified in Crystal Chronicles) but with this everyone is brought to an even playing field and it even works on netplay between different OSes in my experience!

There are definitely some protocol bugs, as endrift has already mentioned.  We have a lot of unknowns with the protocol as well.  Also there's the fact that by removing the old support, we'd be essentially removing support for connecting to VBA-M from Dolphin immediately.  They'd still be able to connect to old builds and releases, so it's not a *huge* problem.  And you could talk me into that from a user standpoint considering how big of a usability upgrade this actually is; this makes it super easy to setup and adds a ton of features that users really want.

I don't see this as an end-point to GBA <-> GCN, however.  I see it as a transitional point with the real tests ahead.  This makes things more consistent, more deterministic, and more useable which should result in us being able to track down the remaining bugs more easily without having to worry about latency being unrealistically high like with the old connectivity on Windows.

Bonta0:
I've restored the previous GBA implementation, a force-push was necessary to keep a coherent commit history, sorry about that
This PR now has a dependency on #9603 to address a comment

Source/Core/Core/HW/GBACore.cpp
Bonta0: Building with LIBMGBA_ONLY will disable every dependency, except for lzma for some reason which is convenient in our case since Dolphin doesn't have 7z support
If anything I'd rather go the opposite direction and disable lzma in mGBA and have Dolphin handle it instead

Source/Core/Core/HW/GBACore.cpp
Bonta0: Turns out that is only available when USE_DEBUGGERS is defined, this is not an issue since we frequently sync devices and already use an s32 for the event anyway

Source/Core/Core/HW/GBACore.h
Bonta0: That was mentioned in this PR, the goal isn't to improve compatibility but address the usability of this feature. Once more accurate timings get merged I will rebase

Externals/mGBA/make_version.c.js
Bonta0: This script reproduces the behavior of your version.cmake to make it consistent across platforms, this isn't even exposed to the user. May I know what the issue is here ?

Source/Core/Core/Config/MainSettings.cpp
Bonta0: That is only true when all your roms are in the same directory. Most emulators will offer this choice

Source/Core/Core/HW/GBACore.h
endrift: Yes I only noticed that after I posted this, apologies for the noise.

Source/Core/Core/HW/GBACore.cpp
endrift: If Dolphin is already linking against minizip it would be preferable to use mGBA's existing minizip code and just point it to the right headers.

Externals/mGBA/make_version.c.js
endrift: The issue is that this will almost certainly forget to get updated when the external is updated. If you don't pull it directly from the version.cmake file then it has an extremely large chance of getting out of sync. One source of truth is preferable.

JosJuice:
Don't worry about the force push, we do that all the time in pull requests to keep the history tidy.

Rumi-Larry:
I would split the "Movies" and "Netplay" to a different pr(s), to reduce the size of the diff. This would also narrow down what needs to be tested, reducing the possibility of unnoticed bugs from being merged.

JMC47:
Considering it's adding a new feature and things are fairly well segmented between the different modules, I'm not certain that's necessary?  The way it is now makes it so everything can be tested in one place and it is pretty explicit about what it adds.

JMC47:
On the latest version of this Pull Request downloaded from buildbot, I can't seem to run any games regardless of controller settings.  Hard crash.  I will run from Visual Studio to see if I get the same crash and get a stack trace.

JMC47:
Here's the stack trace along with Visual Studio saying what happened.

image

image

JMC47:
@dolphin-emu-bot rebuild

JMC47:
After retesting this and testing both the internal GBA and GBA over TCP, I'm pleased with the performance of this build.  I'm able to do things like use savestates, record movies, and even play netplay.  And if the old way to do things is always there if I want to connect over TCP, for say playing across the network on a separate device.

I know there's a lot of code review to happen and things to discuss on that front, but I'm extremely happy with how well it works.  Usage wise, huge thumbs up.

Source/Core/Core/HW/GBACore.cpp
```diff
@@ -0,0 +1,518 @@
...
+  m_stream.core = this;
+  m_stream.videoDimensionsChanged = nullptr;
+  m_stream.postVideoFrame = nullptr;
+  m_stream.postAudioFrame = nullptr;
+  m_stream.postAudioBuffer = [](struct mAVStream* stream, struct blip_t* left,
+                                struct blip_t* right) {
+    auto device = static_cast<AVStream*>(stream)->core;
+    std::vector<s16> buffer(SAMPLES * 2);
+    blip_read_samples(left, &buffer[0], SAMPLES, 1);
+    blip_read_samples(right, &buffer[1], SAMPLES, 1);
```
BhaaLseN, line 311: This looks weird to me, isn't this reading into `buffer[0]` with `left`, overwrite `buffer[1..512]` with `right` and leave `buffer[513..]` unused?

Source/Core/Core/HW/SI/SI_DeviceGBAEmu.h
```diff
@@ -0,0 +1,50 @@
+// Copyright 2008 Dolphin Emulator Project
```
BhaaLseN, line 1:
```suggestion
// Copyright 2021 Dolphin Emulator Project
```

Source/Core/DolphinQt/GBAWidget.cpp
```diff
@@ -0,0 +1,316 @@
...
+#include "DolphinQt/Resources.h"
+#include "DolphinQt/Settings.h"
+
+GBAWidget::GBAWidget(int device_number, std::string_view current_rom, std::string_view title,
+                     u32 width, u32 height, QWidget* parent, Qt::WindowFlags flags)
+    : QWidget(parent, flags), m_device_number(device_number), m_geometry_slot(device_number),
+      m_current_rom(current_rom), m_game_title(title), m_width(width), m_height(height),
+      m_volume(0), m_muted(false)
+{
+  bool visible = true;
```
BhaaLseN, line 34: You're not changing the value of `visible`, this looks a bit like you intended to do something with it, then either removed that code or forgot to actually implement it.

Maybe hide remote players or so?

Source/Core/Core/HW/GBACore.cpp
endrift: No, the last parameter tells it to only write every other location (it's the `stereo` parameter that tells it the buffer is interlaced)

BhaaLseN: Oh, got it. Thanks 👍

Bonta0:
I have included the commit from #9608 for now so this can be tested DSP HLE as well

JMC47:
@dolphin-emu-bot rebuild

JMC47:
Reported issues with Pokemon Box were not related to GBA <-> GCN and were fixed here -> https://github.com/dolphin-emu/dolphin/pull/9625

This should mean that once this is merged along with the timing fixes, all GBA <-> GCN games will be fully supported except one.  The true final boss of the GameCube and the last game not to reach gameplay: The GameBoy Player.

endrift (commented):
I apparently wrote this some time ago and forgot to hit submit review.

Source/Core/Core/HW/SI/SI_DeviceGBAEmu.cpp
```diff
@@ -0,0 +1,167 @@
...
+namespace SerialInterface
+{
+namespace
+{
+int s_num_connected = 0;
+}  // namespace
+
+// --- GameBoy Advance "Link Cable" ---
+
+static int GetTransferTime(u8 cmd)
```
endrift, line 27: This function should be shared with the same function in SI_DeviceGBA.cpp, which should be exposed instead of static.

Source/Core/DolphinQt/GBAWidget.cpp
```diff
@@ -0,0 +1,316 @@
...
+      m_geometry_slot = client->InGamePadToLocalPad(m_device_number);
+    }
+    else
+    {
+      int ingame_slot = client->LocalPadToInGamePad(m_device_number);
+      if (ingame_slot < 4)
+        m_geometry_slot = ingame_slot;
+      if (client->GetNetSettings().m_HideRemoteGBAs)
+        visible = false;
+      if (!IsMuted())
```
v1993, line 59: Should remote GBAs be unconditionally muted? I'd either make this depend on `HideRemoteGBAs` (so all visible GBAs play sound and all invisible do not) or even separate into its own option.

JMC47:
Now that mGBA's 0.9.1 release and the major regressions are out of the way, I would like to start moving toward getting this ready to merge along with the other GBA fixes.

Bonta0:
Rebased & updated mGBA to 0.9.1, this should fix every known issue so far

Source/Core/Core/HW/GBACore.cpp
```diff
@@ -0,0 +1,518 @@
...
+
+Core::Core(int device_number, u64 gc_ticks)
+    : m_device_number(device_number), m_last_gc_ticks(gc_ticks)
+{
+  mLogSetDefaultLogger(&s_dummy_logger);
+  m_core = mCoreCreate(mPlatform::mPLATFORM_GBA);
+  m_core->init(m_core);
+
+  mCoreInitConfig(m_core, "dolphin");
+  mCoreConfigSetValue(&m_core->config, "idleOptimization", "detect");
```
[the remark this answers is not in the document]
Bonta0: This can be more efficient than the default ``"remove"`` in some games. I left it there as a way to potentially avoid stalling Dolphin or lower CPU usage

JMC47:
The only issue I know of that still happens in GBA <-> GCN is a freeze *after* every match of Navi Trackers.  Reproduced on an earlier version of this Pull Request, but not the latest.  I'll be able to test it again in a few hours.  The good news with the hang is that it happens after a save so it's whatever.

PatrickFerry:
> The only issue I know of that still happens in GBA <-> GCN is a freeze _after_ every match of Navi Trackers. Reproduced on an earlier version of this Pull Request, but not the latest. I'll be able to test it again in a few hours. The good news with the hang is that it happens after a save so it's whatever.

There was a comment in the forum https://forums.dolphin-emu.org/Thread-mgba-now-supports-dolphin-gba-link-and-it-works-with-sonic-adventure-2

Which demo disk has GBA connections? It would be worth testing too.

t27duck:
>
> There was a comment in the forum https://forums.dolphin-emu.org/Thread-mgba-now-supports-dolphin-gba-link-and-it-works-with-sonic-adventure-2
> 
> Which demo disks have GBA connections? It would be worth testing too.

Most likely referring to this disc which had 2 GBA demos on it https://wiki.dolphin-emu.org/index.php?title=Nintendo_GameCube_Preview_Disc

fancythedeveloper:
I'm assuming the only real downside is losing compatibility with VBA-M, even if that? Even then, mGBA is generally a better emulator anyways so I don't think I would really be losing anything if that's the case.

JMC47:
VBA-M support is not dropped with this change.  This adds a new option that lets you bypass using an external GBA emulator.

Source/Core/Core/Config/MainSettings.cpp
v1993: To clarify, first is a path to GBA saves, second is a bool (checkbox) to override the first and store saves in same directory as ROM file.

Source/Core/Core/HW/GBAPadEmu.cpp
```diff
@@ -0,0 +1,106 @@
...
+  m_buttons->SetControlExpression(3, "W");  // R
+#ifdef _WIN32
+  m_buttons->SetControlExpression(4, "BACK");    // Select
+  m_buttons->SetControlExpression(5, "RETURN");  // Start
+#else
+  // OS X/Linux
+  // Start
+  m_buttons->SetControlExpression(4, "Backspace");
+  m_buttons->SetControlExpression(5, "Return");
+#endif
```
v1993, line 99:
```suggestion
#ifdef _WIN32
  m_buttons->SetControlExpression(4, "BACK");    // Select
  m_buttons->SetControlExpression(5, "RETURN");  // Start
#else
  m_buttons->SetControlExpression(4, "Backspace");  // Select
  m_buttons->SetControlExpression(5, "Return");     // Start
#endif
```

It seems that second section have wrong button comments. Also, `// OS X/Linux` is both wrong (FreeBSD, anyone?) and obvious from code.

Source/Core/DolphinQt/Config/GamecubeControllersWidget.cpp
```diff
@@ -25,23 +26,32 @@
...
+    {SerialInterface::SIDEVICE_NONE, "None"},
+    {SerialInterface::SIDEVICE_GC_CONTROLLER, "Standard Controller"},
+    {SerialInterface::SIDEVICE_WIIU_ADAPTER, "GameCube Adapter for Wii U"},
+    {SerialInterface::SIDEVICE_GC_STEERING, "Steering Wheel"},
+    {SerialInterface::SIDEVICE_DANCEMAT, "Dance Mat"},
+    {SerialInterface::SIDEVICE_GC_TARUKONGA, "DK Bongos"},
+#ifdef HAS_LIBMGBA
+    {SerialInterface::SIDEVICE_GC_GBA_EMULATED, "GBA (Emulated)"},
+#endif
+    {SerialInterface::SIDEVICE_GC_GBA, "GBA (TCP)"},
```
v1993, line 39: I think that this naming can be a little confusing (this makes it sound like TCP options refers to physical GBA as opposed to emulated one which is obviously very wrong). Maybe "GBA (built-in, recommended)"/"GBA (separate emulator)"?

v1993 (changes requested):

fancythedeveloper:
> VBA-M support is not dropped with this change. This adds a new option that lets you bypass using an external GBA emulator.

Ah I see, alright.

fancythedeveloper:
Also, wouldn't this being deterministic allow for possible Netplay support, if that's not already a thing?

AdmiralCurtiss:
yes, read the first comment in this PR, netplay is specifically mentioned as working.

JMC47:
@dolphin-emu-bot rebuild

fancythedeveloper:
Anyways, I'm quite excited for what this PR means, as like what JMC said, could be the first step in finally emulating the GBP.

JMC47:
One thing to add is that Navi Trackers still freezes when you save the game in this pull request.

All connected GBAs reset, the game stops (as if it's waiting for them to reconnect?) and the GBAs play the BIOS but then nothing happens.

Source/Core/DolphinQt/Config/GamecubeControllersWidget.cpp
JMC47: I like how it is now, but I can see how it may be confusing.  What if it was like this.

GBA (Integrated) and GBA (External)

I don't know if that fixes what you want though, but it would make it clear what each one did.  I do like how TCP leaves it more consistent with the type of connectivity though.

v1993: I think your wording is fine: people will look things up online instead of making wrong assumptions should they get confused if we'll go with it.

Of course current wording is *technically correct*, but it isn't very informative to your random causal user.

v1993:
Sorry for being Capt. Obvious, but was it tested what happens on console when you try to save in Navi Trackers? Do GBAs also reset there?

AdmiralCurtiss:
Yes, the reset is intentional, but it's supposed to reconnect shortly afterwards.

See https://www.youtube.com/watch?v=4Gnmt8L-k80 for an example.

AdmiralCurtiss:
For reference, I just checked how it looks in Dolphin right now, and I think 'it doesn't reconnect' is actually an incorrect diagnostic. The GC game is supposed to go back to its stage select screen after resetting the GBAs, but never does for some reason. My guess is that it's expecting a 'yes i have resetted myself' reply from the GBA side that never comes, but that's pure speculation.

Bonta0:
Rebased and updated mGBA to fix another savestate issue
The last 3 commits are only there for testing purposes, they belong in separate PRs. Those address the FFCC and 4S+ issues

JMC47:
@dolphin-emu-bot rebuild

Bonta0:
Sure, go ahead since HW tests confirmed it is accurate

Source/Core/Core/HW/SI/SI_DeviceGCController.cpp
```diff
@@ -318,15 +312,12 @@ void CSIDevice_GCController::SendCommand(u32 command, u8 poll)
       INFO_LOG_FMT(SERIALINTERFACE, "PAD {} set to mode {}", m_device_number, m_mode);
     }
   }
-  break;
-
-  default:
+  else if (controller_command.command != 0x00)
   {
+    // Costis sent 0x00 in some demos :)
```
Rumi-Larry, line 317: Seems to be a typo here

JMC47:
@dolphin-emu-bot rebuild

JMC47:
All issues should be fixed with GBA <-> GCN.  I tested the games broken in the previous version of this PR, and since fixes have been merged.

Previous games should not be affected.

Source/Core/Core/HW/SI/SI_DeviceGCController.cpp
BhaaLseN: I'm fairly certain their name is Costis, or didn you mean something else? I don't see any other potential typo.

Rumi-Larry: It seemed that they meant "Cost is" I don't know who Costis is but anyhow the meaning isn't clear.

BhaaLseN: They made a few homebrew things like SDLoader/PSOLoader. Assuming that comment is about them.

JMC47:
@dolphin-emu-bot rebuild

JMC47:
@dolphin-emu-bot rebuild

Source/Core/Core/Movie.cpp
```diff
@@ -432,15 +437,21 @@ void ChangePads()
   {
     if (SerialInterface::SIDevice_IsGCController(SConfig::GetInstance().m_SIDevice[i]))
       controllers |= (1 << i);
+    if (SConfig::GetInstance().m_SIDevice[i] == SerialInterface::SIDEVICE_GC_GBA_EMULATED)
+      controllers |= (1 << (i + 8));
   }
 
-  if ((s_controllers & 0x0F) == controllers)
+  if ((s_controllers & 0x0F0F) == controllers)
```
leoetlino, line 444: Doing all the masking and bitwise stuff manually seems a bit error prone. Could we add a wrapper to manipulate the controller mask more safely and avoid having to expose those details to non-Movie code?

(I realise the problem already existed prior to your changes, but this PR exacerbates it.)

Source/Core/DolphinQt/Settings/GameCubePane.cpp
```diff
@@ -124,8 +128,51 @@ void GameCubePane::CreateWidgets()
...
+  gba_layout->addWidget(new QLabel(tr("Bios:")), gba_row, 0);
+  gba_layout->addWidget(m_gba_bios_edit, gba_row, 1);
+  gba_layout->addWidget(m_gba_browse_bios, gba_row, 2);
+  gba_row++;
+
+  for (size_t i = 0; i < m_gba_rom_edits.size(); ++i)
+  {
+    m_gba_rom_edits[i] = new QLineEdit();
+    m_gba_browse_roms[i] = new QPushButton(QStringLiteral("..."));
+    gba_layout->addWidget(new QLabel(tr("Port %1 Rom:").arg(i + 1)), gba_row, 0);
```
JosJuice, line 153: ROM should preferably be capitalized (here and below).

Source/Core/Core/HW/GBAPadEmu.cpp
```diff
@@ -0,0 +1,115 @@
...
+  groups.emplace_back(m_buttons = new ControllerEmu::Buttons(_trans("Buttons")));
+  for (const char* named_button : named_buttons)
+  {
+    ControllerEmu::Translatability translate = ControllerEmu::DoNotTranslate;
+    std::string ui_name = named_button;
+
+    if (named_button == std::string("Start"))
+    {
+      translate = ControllerEmu::Translate;
+      ui_name = _trans("START");
```
JosJuice, line 31: It would probably be simpler to just make the normal name uppercase as well. The reason why that isn't the case for GC controllers is because we didn't want to break compatibility with old INIs where it isn't in uppercase. Though I guess there is an argument to be made that it's nice to have GBA INIs look the same as GC INIs, so don't treat this as a required change.

Source/Core/Core/NetPlayClient.cpp
```diff
@@ -1057,6 +1094,29 @@ unsigned int NetPlayClient::OnData(sf::Packet& packet)
...
+        return 0;
+
+      u8 slot;
+      packet >> slot;
+
+      const std::string path =
+          fmt::format("{}{}{}.sav", File::GetUserPath(D_GBAUSER_IDX), GBA_SAVE_NETPLAY, slot + 1);
+      if (File::Exists(path) && !File::Delete(path))
+      {
+        PanicAlertFmtT("Failed to delete NetPlay GBA{0} Save file. Verify your write permissions.",
```
JosJuice, line 1109: Lowercase `save`

JosJuice (commented):
JMC asked me to check if this contains anything that would break things on Android. I couldn't find anything (of course the new feature doesn't actually work on Android as of now, but that's a different problem), but I have some minor comments on translatable strings.

JMC47:
Over the past couple of weeks, I've probably used this pull request as much as I have master.  Earlier today, I did verify that movie recordings seemed to work, but I'm not a TASer, so I don't know how to use rerecording to really stress test it.  

Netplay, on the other hand, I tested in a multitude of situations.  I tested Wind Waker on netplay, using a second player to control the Tingle Tuner, Pac-Man Vs., Final Fantasy Crystal Chronicles, and even more complicated titles like Pokemon Colosseum where you use multiple different GBA roms and savefiles.  Everything was rock solid.  I even organized 4 player sessions of Four Swords Adventures and recorded it for a video.  There were no crashes on netplay, and we were able to run multiple netplay sessions without having to exit the netplay room.

In terms of other functionality, there's really no words for how polished the experience was.  Over the past few weeks, I've gone through every GBA connectivity title that I could, from demos, to Wario World, to *e-Reader titles*.  Every single title that I tried works.  I'd go through the list, but honestly it'd be more games listed than the list of GBA Connectivity titles on the Wiki.

I never experienced any crashes or issues regarding setup, there is a warning regarding the lack of BIOS.  I've tested importing/exporting saves and savestates to mGBA proper (one of which is in the video) and I've confirmed that TCP GBAs can be run alongside Emulated GBAs.

Performance wise, it runs like a dream.  There's not really much of a performance impact having an Emulated GBA running alongside Dolphin, especially compared to TCP GBAs.

It's still sinking in for me; I got to play Four Swords Adventures on netplay with friends across the continent.  And it wasn't some hacky, *hope this will work* situation.

JMC47 (approved):

endrift:
I'll try to find time to go over the PR today to see if I find anything that could use fixing.

BhaaLseN (approved):
Did another quick pass on the source, nothing that sticks out (and wasn't there before), so its probably good to go (if theres nothing mGBA specific to be addressed). As usual, I only stared at code and didn't test it myself, though.

Source/Core/DolphinQt/Config/GamecubeControllersWidget.cpp
endrift: I agree that this wording is confusing. Both are emulated after all, just one internal to Dolphin and one external. Perhaps "Integrated" and "Remote".

endrift (approved):
Looks good to me, though it has brought some potential upstream issues to my attention.

Source/Core/DolphinQt/Config/GamecubeControllersWidget.cpp
JMC47: I'm fine with Integrated and Remote if we want to go that route.  Just let me know so I can re-render things.

JosJuice: Integrated plus TCP would be my pick. If we do go with Remote, I wouldn't mind, but please put an i18n comment above the string so that translators don't accidentally think that it means the same kind of "remote" as in "Wii Remote".

JMC47: Just someone @ me if a change is made so I can rerender footage 👍

Source/Core/Core/HW/GBACore.cpp
Bonta0: Dolphin's minizip doesn't have the zipping functions required for mGBA's ``vfs-zip``. We could update it but that would only introduce dead code

CMakeLists.txt
```diff
@@ -826,6 +827,14 @@ if(USE_DISCORD_PRESENCE)
   include_directories(Externals/discord-rpc/include)
 endif()
 
+if(NOT ENABLE_QT)
+  set(USE_MGBA 0)
+endif()
+if(USE_MGBA)
+  message(STATUS "Using static libmgba from Externals")
```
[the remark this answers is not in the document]
Bonta0: That could be a issue for NetPlay or Movies, we may have to enforce a specific version for those cases

Source/Core/Core/HW/GBACore.cpp
```diff
@@ -0,0 +1,704 @@
...
+namespace
+{
+mLogger s_stub_logger = {
+    [](mLogger*, int category, mLogLevel level, const char* format, va_list args) {}, nullptr};
+}  // namespace
+
+constexpr auto SAMPLES = 512;
+constexpr auto SAMPLE_RATE = 48000;
+
+// libmGBA does not return the correct frequency for some GB models
```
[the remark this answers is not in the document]
Bonta0: https://github.com/mgba-emu/mgba/blob/617bb0a09310c5dbeefd263c2efd7975cb481243/src/gb/core.c#L712-L713 That is a TODO and requires more work than simply returning the expected value

Source/Core/DolphinQt/Config/GamecubeControllersWidget.cpp
Bonta0: Integrated and TCP would be my pick as well, "Remote" is a bit vague and would need to be changed if a shared memory implementation ever comes up

JMC47: I'll assume we're changing over to Integrated and re-render then.  Seems we have a consensus!

CMakeLists.txt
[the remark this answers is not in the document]
endrift: That's a good point. Distros can get angry about it though.

AdmiralCurtiss:
@dolphin-emu-bot rebuild

leoetlino (changes requested):
> Reviewed 15 of 15 files at r1, 4 of 4 files at r2, 2 of 2 files at r3, 7 of 7 files at r4, 2 of 2 files at r5, 3 of 3 files at r6, 2 of 2 files at r7, 15 of 15 files at r8, 10 of 10 files at r9, 3 of 3 files at r10, 6 of 6 files at r11, 4 of 4 files at r12, 9 of 9 files at r13, 7 of 7 files at r14, 8 of 8 files at r15, 9 of 9 files at r16, 16 of 16 files at r17, 5 of 5 files at r18.
> *[Reviewable](https://reviewable.io/reviews/dolphin-emu/dolphin/9600)* status: all files reviewed, 7 unresolved discussions (waiting on @Bonta0)

___
*[Source/Core/DolphinQt/GBAHost.cpp, line 28 at r15](https://reviewable.io/reviews/dolphin-emu/dolphin/9600#-MeSM_EaE2mROHzfidaq:-MeSM_EaE2mROHzfidar:bj58x8z) ([raw file](https://github.com/dolphin-emu/dolphin/blob/ea7ca8265b7e61ba2600356caae9299a63522af2/Source/Core/DolphinQt/GBAHost.cpp#L28)):*
> ```cpp
>   u32 width, height;
>   core_ptr->GetVideoDimensions(&width, &height);
>   QMetaObject::invokeMethod(
> ```

This wasn't immediately obvious to me, but for anyone wondering the reason invokeMethod is used (rather than a direct call or emit) is for thread safety reasons.
___
*[Source/Core/DolphinQt/GBAHost.cpp, line 42 at r15](https://reviewable.io/reviews/dolphin-emu/dolphin/9600#-MeSLoW8EiNfuJ2czZ5q:-MeSLoW9E2kUtvsYFYGN:b-9jnqba) ([raw file](https://github.com/dolphin-emu/dolphin/blob/ea7ca8265b7e61ba2600356caae9299a63522af2/Source/Core/DolphinQt/GBAHost.cpp#L42)):*
> ```cpp
> {
>   auto core_ptr = m_core.lock();
>   if (!core_ptr->IsStarted())
> ```

Should we check or assert that m_core is still valid (or equivalently, that core_ptr is not null)?
___
*[Source/Core/DolphinQt/GBAWidget.cpp, line 233 at r15](https://reviewable.io/reviews/dolphin-emu/dolphin/9600#-MeSOF64999WxwTdFkM0:-MeSOF64999WxwTdFkM1:b-xxoywb) ([raw file](https://github.com/dolphin-emu/dolphin/blob/ea7ca8265b7e61ba2600356caae9299a63522af2/Source/Core/DolphinQt/GBAWidget.cpp#L233)):*
> ```cpp
>   else
>   {
>     move(x() - (frameGeometry().width() / (m_local_pad & 1 ? -2 : 2)),
> ```

`% 2 != 0`? not sure I understand the check here
___
*[Source/Core/DolphinQt/GBAWidget.cpp, line 234 at r15](https://reviewable.io/reviews/dolphin-emu/dolphin/9600#-MeSOPRX-FpVNvMbwsR0:-MeSOPRX-FpVNvMbwsR1:bji28bi) ([raw file](https://github.com/dolphin-emu/dolphin/blob/ea7ca8265b7e61ba2600356caae9299a63522af2/Source/Core/DolphinQt/GBAWidget.cpp#L234)):*
> ```cpp
>   {
>     move(x() - (frameGeometry().width() / (m_local_pad & 1 ? -2 : 2)),
>          y() - (frameGeometry().height() / (m_local_pad & 2 ? -2 : 2)));
> ```

and here
___
*[Source/Core/DolphinQt/GBAWidget.cpp, line 262 at r15](https://reviewable.io/reviews/dolphin-emu/dolphin/9600#-MeSOWW84THvBS0EntEI:-MeSOWW84THvBS0EntEJ:b-yvdk6o) ([raw file](https://github.com/dolphin-emu/dolphin/blob/ea7ca8265b7e61ba2600356caae9299a63522af2/Source/Core/DolphinQt/GBAWidget.cpp#L262)):*
> ```cpp
> void GBAWidget::contextMenuEvent(QContextMenuEvent* event)
> {
>   QMenu* menu = new QMenu(this);
> ```

these can be `auto` too
___
*[Source/Core/DolphinQt/GBAWidget.cpp, line 409 at r15](https://reviewable.io/reviews/dolphin-emu/dolphin/9600#-MeSOeUN1z04BywDh3Pl:-MeSOeUN1z04BywDh3Pm:b-9ca2v5) ([raw file](https://github.com/dolphin-emu/dolphin/blob/ea7ca8265b7e61ba2600356caae9299a63522af2/Source/Core/DolphinQt/GBAWidget.cpp#L409)):*
> ```cpp
> }
> 
> GBAWidgetController::GBAWidgetController()
> ```

`GBAWidgetController::GBAWidgetController() = default;`
___
*[Source/Core/DolphinQt/Config/GamecubeControllersWidget.cpp, line 29 at r14](https://reviewable.io/reviews/dolphin-emu/dolphin/9600#-MeSKtA34OAi4uLwoVpR:-MeSKtA34OAi4uLwoVpS:biq51j0) ([raw file](https://github.com/dolphin-emu/dolphin/blob/c63e31f86f77c5e5fcd906a9373104b44a905fc0/Source/Core/DolphinQt/Config/GamecubeControllersWidget.cpp#L29)):*
> ```cpp
> 
> static const std::vector<std::pair<SerialInterface::SIDevices, const char*>> s_gc_types = {
>     {SerialInterface::SIDEVICE_NONE, "None"},
> ```

These strings should be marked as translatable manually with `_trans`, like this: `_trans("None")`

(String literals that are passed to tr() are automatically marked as translatable, but here you need to do this manually by using the macro.)
___
*[Source/Core/DolphinQt/Config/GamecubeControllersWidget.cpp, line 117 at r14](https://reviewable.io/reviews/dolphin-emu/dolphin/9600#-MeSLMGJ278O-8sLPosS:-MeSLMGJ278O-8sLPosT:burgq1u) ([raw file](https://github.com/dolphin-emu/dolphin/blob/c63e31f86f77c5e5fcd906a9373104b44a905fc0/Source/Core/DolphinQt/Config/GamecubeControllersWidget.cpp#L117)):*
> ```cpp
>     {
>       const SerialInterface::SIDevices si_device = FromGCMenuIndex(box->currentIndex());
>       m_gc_buttons[i]->setEnabled(si_device != SerialInterface::SIDEVICE_NONE &&
> ```

nit: might be worth making a function for this check? It's repeated 3 times ^^

leoetlino:
(Used Reviewable because it was getting difficult to track changes across pushes with GitHub's web interface. Even 4 years later the built-in review tools are still mostly unusable for large PRs.)

Bonta0 (commented):
> *[Reviewable](https://reviewable.io/reviews/dolphin-emu/dolphin/9600)* status: all files reviewed, 6 unresolved discussions (waiting on @leoetlino)

___
*[Source/Core/DolphinQt/GBAHost.cpp, line 28 at r15](https://reviewable.io/reviews/dolphin-emu/dolphin/9600#-MeSM_EaE2mROHzfidaq:-MeVGbYJ9BV8S_-zE9Hm:bv1mrkz) ([raw file](https://github.com/dolphin-emu/dolphin/blob/ea7ca8265b7e61ba2600356caae9299a63522af2/Source/Core/DolphinQt/GBAHost.cpp#L28)):*
<details><summary><i>Previously, leoetlino (Léo Lam) wrote…</i></summary><blockquote>

This wasn't immediately obvious to me, but for anyone wondering the reason invokeMethod is used (rather than a direct call or emit) is for thread safety reasons.
</blockquote></details>

I wasn't aware of QueueOnObject at the time, I've replaced those invokeMethod to use that instead to be more consistent with the rest of the code base
___
*[Source/Core/DolphinQt/GBAHost.cpp, line 42 at r15](https://reviewable.io/reviews/dolphin-emu/dolphin/9600#-MeSLoW8EiNfuJ2czZ5q:-MeVGU9eD-l8NLZ8g8EJ:b734l) ([raw file](https://github.com/dolphin-emu/dolphin/blob/ea7ca8265b7e61ba2600356caae9299a63522af2/Source/Core/DolphinQt/GBAHost.cpp#L42)):*
<details><summary><i>Previously, leoetlino (Léo Lam) wrote…</i></summary><blockquote>

Should we check or assert that m_core is still valid (or equivalently, that core_ptr is not null)?
</blockquote></details>

This is called directly by the core itself so it can't happen. I've added the check anyway
___
*[Source/Core/DolphinQt/GBAWidget.cpp, line 233 at r15](https://reviewable.io/reviews/dolphin-emu/dolphin/9600#-MeSOF64999WxwTdFkM0:-MeVFBGK6fZL1EP9Wskb:b-ggsopa) ([raw file](https://github.com/dolphin-emu/dolphin/blob/ea7ca8265b7e61ba2600356caae9299a63522af2/Source/Core/DolphinQt/GBAWidget.cpp#L233)):*
<details><summary><i>Previously, leoetlino (Léo Lam) wrote…</i></summary><blockquote>

`% 2 != 0`? not sure I understand the check here
</blockquote></details>

The point was to arrange the windows in a nice square pattern on the first run but Qt does a good enough job on its own so I'll just let Qt handle it and remove this entirely
___
*[Source/Core/DolphinQt/GBAWidget.cpp, line 234 at r15](https://reviewable.io/reviews/dolphin-emu/dolphin/9600#-MeSOPRX-FpVNvMbwsR0:-MeVGQR_B0iJQmP23wk4:b-896fix) ([raw file](https://github.com/dolphin-emu/dolphin/blob/ea7ca8265b7e61ba2600356caae9299a63522af2/Source/Core/DolphinQt/GBAWidget.cpp#L234)):*
<details><summary><i>Previously, leoetlino (Léo Lam) wrote…</i></summary><blockquote>

and here
</blockquote></details>

Done.
___
*[Source/Core/DolphinQt/GBAWidget.cpp, line 262 at r15](https://reviewable.io/reviews/dolphin-emu/dolphin/9600#-MeSOWW84THvBS0EntEI:-MeVHQxj-P09NTL-T4UQ:b-896fix) ([raw file](https://github.com/dolphin-emu/dolphin/blob/ea7ca8265b7e61ba2600356caae9299a63522af2/Source/Core/DolphinQt/GBAWidget.cpp#L262)):*
<details><summary><i>Previously, leoetlino (Léo Lam) wrote…</i></summary><blockquote>

these can be `auto` too
</blockquote></details>

Done.
___
*[Source/Core/DolphinQt/GBAWidget.cpp, line 409 at r15](https://reviewable.io/reviews/dolphin-emu/dolphin/9600#-MeSOeUN1z04BywDh3Pl:-MeVHRQKD1S5A-CFnH98:b-896fix) ([raw file](https://github.com/dolphin-emu/dolphin/blob/ea7ca8265b7e61ba2600356caae9299a63522af2/Source/Core/DolphinQt/GBAWidget.cpp#L409)):*
<details><summary><i>Previously, leoetlino (Léo Lam) wrote…</i></summary><blockquote>

`GBAWidgetController::GBAWidgetController() = default;`
</blockquote></details>

Done.
___
*[Source/Core/DolphinQt/Config/GamecubeControllersWidget.cpp, line 29 at r14](https://reviewable.io/reviews/dolphin-emu/dolphin/9600#-MeSKtA34OAi4uLwoVpR:-MeVHRy183YtvlyU7TA1:b-896fix) ([raw file](https://github.com/dolphin-emu/dolphin/blob/c63e31f86f77c5e5fcd906a9373104b44a905fc0/Source/Core/DolphinQt/Config/GamecubeControllersWidget.cpp#L29)):*
<details><summary><i>Previously, leoetlino (Léo Lam) wrote…</i></summary><blockquote>

These strings should be marked as translatable manually with `_trans`, like this: `_trans("None")`

(String literals that are passed to tr() are automatically marked as translatable, but here you need to do this manually by using the macro.)
</blockquote></details>

Done.
___
*[Source/Core/DolphinQt/Config/GamecubeControllersWidget.cpp, line 117 at r14](https://reviewable.io/reviews/dolphin-emu/dolphin/9600#-MeSLMGJ278O-8sLPosS:-MeVHVeM3EuU7ybXjnZk:b-896fix) ([raw file](https://github.com/dolphin-emu/dolphin/blob/c63e31f86f77c5e5fcd906a9373104b44a905fc0/Source/Core/DolphinQt/Config/GamecubeControllersWidget.cpp#L117)):*
<details><summary><i>Previously, leoetlino (Léo Lam) wrote…</i></summary><blockquote>

nit: might be worth making a function for this check? It's repeated 3 times ^^
</blockquote></details>

Done.

leoetlino (approved):
> Reviewed 1 of 85 files at r19, 85 of 85 files at r20.
> *[Reviewable](https://reviewable.io/reviews/dolphin-emu/dolphin/9600)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Bonta0)

Externals/mGBA/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
+set(LIBMGBA_ONLY ON)
+set(USE_LZMA ON)
+add_subdirectory(mgba EXCLUDE_FROM_ALL)
+
+if(NOT MSVC)
+  target_compile_options(mgba PRIVATE -Wno-unused-parameter -Wno-unused-result -Wno-unused-variable)
+endif()
+
+if(ANDROID)
+  target_compile_definitions(mgba PRIVATE -Dfutimes=futimens)
```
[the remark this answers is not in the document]
endrift: This should be fixed upstream now.

JMC47:
A feature I never thought would be... is.

PR #9602, opened 2021-03-20
Avoid using panic alerts in DiscIO

JosJuice:
Panic alerts in DiscIO can potentially be very annoying since large amounts of them can pop up when loading the game list if you have some particularly weird files in your game list.

This was a much bigger problem back in 5.0 with its "Tried to decrypt data from a non-Wii volume" panic alert, but I figured I would take it all the way and remove the remaining panic alerts that can show up when loading the game list.

I have exempted uses of ASSERT/ASSERT_MSG since they indicate a bug in Dolphin rather than a malformed file.

### DIFF
--- a/Source/Core/DiscIO/CompressedBlob.cpp
+++ b/Source/Core/DiscIO/CompressedBlob.cpp
@@ -83,7 +83,7 @@ u64 CompressedBlobReader::GetBlockCompressedSize(u64 block_num) const
   else if (block_num == m_header.num_blocks - 1)
     return m_header.compressed_data_size - start;
   else
-    PanicAlertFmt("{} - illegal block number {}", __func__, block_num);
+    ERROR_LOG_FMT(DISCIO, "{} - illegal block number {}", __func__, block_num);
   return 0;
 }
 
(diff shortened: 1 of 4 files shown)


Source/Core/DiscIO/CompressedBlob.cpp
```diff
@@ -133,7 +134,7 @@ bool CompressedBlobReader::GetBlock(u64 block_num, u8* out_ptr)
     z.avail_in = comp_block_size;
     if (z.avail_in > m_header.block_size)
     {
-      PanicAlertFmt("We have a problem");
+      ERROR_LOG_FMT(DISCIO, "We have a problem");
```
BhaaLseN, line 137: Can we make this log entry more useful while we're at it? As it is now, it doesn't help anyone (other than letting us know we hit this line)

BhaaLseN (approved):
Nobody reads those anyways, do they? :D
Replacements are straight-forward, untested but LGTM.

Source/Core/DiscIO/CompressedBlob.cpp
JosJuice: I've replaced it with a message that I *think* is correct.

lioncash (approved):

PR #9603, opened 2021-03-21
AudioCommon: Split Initialization

Bonta0:
This splits SoundStream initialization in two parts, allowing HW::Init to set sample rates and addresses a comment that's been around for a while

### DIFF
--- a/Source/Core/AudioCommon/AudioCommon.cpp
+++ b/Source/Core/AudioCommon/AudioCommon.cpp
@@ -15,7 +15,6 @@
 #include "Common/FileUtil.h"
 #include "Common/Logging/Log.h"
 #include "Core/ConfigManager.h"
-#include "Core/HW/AudioInterface.h"
 
 // This shouldn't be a global, at least not here.
 std::unique_ptr<SoundStream> g_sound_stream;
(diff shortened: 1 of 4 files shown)


JosJuice:
Perhaps it would be good to leave a comment about why `PostInitSoundStream` is a separate function that has to be called later, but LGTM otherwise.

BhaaLseN:
LGTM, aside from Jos' remark.

Source/Core/AudioCommon/AudioCommon.cpp
```diff
@@ -66,13 +65,11 @@ void InitSoundStream()
...
 
-  // Ideally these two calls would be done in AudioInterface::Init so that we don't
-  // need to have a dependency on AudioInterface here, but this has to be done
-  // after creating g_sound_stream (above) and before starting audio dumping (below)
-  g_sound_stream->GetMixer()->SetDMAInputSampleRate(AudioInterface::GetAIDSampleRate());
-  g_sound_stream->GetMixer()->SetStreamInputSampleRate(AudioInterface::GetAISSampleRate());
-
+void PostInitSoundStream()
+{
+  // This needs to be called after HW::Init where input sample rates are set
```
BhaaLseN, line 72: This should probably mention `AudioInterface::Init`, should it? `HW::Init` isn't wrong (because it calls `AudioInterface::Init`), but it should only be a secondary mention due to the connection in `EmuThread`.
Thoughts?

Bonta0: I didnt want to be too specific about `AudioInterface::Init` since #9600 also takes advantage of this which would make the comment no longer correct, so I guess I'll just update the comment in that PR as well

PR #9606, opened 2021-03-24
Qt: avoid queuing ConfigChanged() more than once

Filoppi:
This often happens when changing certain settings and starting/stopping a game and it's just a waste.

I'm not even sure how safe it is to not call "OnConfigChanged()" inline when editing settings from the Qt UI with a cursor, given that a lot of widgets states are refreshed within that callback. But I guess a user couldn't interact with two widgets before Qt had actually run its next update and called that queued callback.

### DIFF
--- a/Source/Core/DolphinQt/Settings.cpp
+++ b/Source/Core/DolphinQt/Settings.cpp
@@ -4,6 +4,8 @@
 
 #include "DolphinQt/Settings.h"
 
+#include <atomic>
+
 #include <QApplication>
 #include <QDir>
 #include <QFile>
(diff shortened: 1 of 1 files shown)


Source/Core/DolphinQt/Settings.cpp
```diff
@@ -40,8 +42,17 @@ Settings::Settings()
     QueueOnObject(this, [this, new_state] { emit EmulationStateChanged(new_state); });
   });
 
-  Config::AddConfigChangedCallback(
-      [this] { QueueOnObject(this, [this] { emit ConfigChanged(); }); });
+  Config::AddConfigChangedCallback([this] {
+    static std::atomic<bool> do_once{true};
+    if (do_once)
+    {
+      do_once = false;
```
JosJuice, line 49: This doesn't seem very race-free to me. I believe you should at least make the following change, though I'm not sure if it's sufficient or if something also needs to be changed at the line where you're assigning `false`:

```suggestion
    if (do_once.exchange(false))
    {
```

Filoppi: It should be, as long as do_once is set to true before emitting the OnConfigChanged() I don't see any problem. Even that exchange wasn't necessary for safety, but it's an optimization if multiple threads changed settings at the same time.

JosJuice: I believe you're correct in that the only harmful effect of this race is that we will emit ConfigChanged more times than needed.

Filoppi: When "Config" invokes the ConfigChanged() callbacks, the settings have already changed, so even if the host/Qt thread callstack was within that queued lambda for a previous config change, the setting change will either be broadcast immediately within the current Qt OnConfigChanged() or queued for later.
The only race condition possible is that the optimization would fail to apply and OnConfigChanged() is called more than needed.

PR #9607, opened 2021-03-24
Allow binding exact position and orientation of Wiimote IR camera relative to sensor bar

MichaelJW:
Hi folks, I'd like to suggest adding a new panel in the **Motion Input** tab that allows the user to bind the x-, y-, and z-coordinates of the Wiimote's IR camera, as well as its pitch, roll and yaw:

image

My personal main motivation for seeing this feature is for use with VR, since (unlike with, say, a Joy-Con) we can track the exact position and orientation of motion controllers, as well as the accelerometer and gyro values that many modern game controllers support.

At the moment, I believe (and please correct me if I'm wrong) that in order to simulate a sensor bar in Dolphin, I have two options:

- Raycast from the VR controller to get a point on the VR TV screen, and bind this to the **Motion Simulation** - **Point** inputs.
    - However, this doesn't support roll movements, unless I also add some **Tilt** bindings, at which point I would have a strange combination of **Motion Simulation** and **Motion Input** bindings.
    - Also, I'm guessing that this could cause some problems with games that use both accelerometer and IR camera inputs to figure out what the controller is doing - not sure.
- Use the **Motion Input** - **Point** panel's simulated pointer position
    - Although this is really good, it's not quite perfect, due to (what I appreciate is unavoidable) integration errors that blow up quickly.

I considered having inputs for the two sensor bar IR LEDs' positions and brightness, which the VR app would simulate and pass through, but since Dolphin already has a function to convert x/y/z/pitch/roll/yaw into these values, I thought it would be simpler to just pass those to that existing function.

Using these six values is also convenient for other existing software. For instance, [opentrack](https://github.com/opentrack/opentrack) can retrieve the six values from a number of different trackers. [I've made a small modification here](https://github.com/MichaelJW/OpenTrackDSUServer) to @iwubcode's excellent OpenTrackDSUServer that allows passing these raw values rather than calculated accelerometer and gyro values, so in theory someone could, for example, tape a [Delanclip](https://delanengineering.com/) to a Joy-Con, run both BetterJoy and OpenTrackDSUServer, and retrieve accel, gyro, button, _and_ IR position and orientation data, for very accurate Wiimote emulation.

I don't have a Delanclip, but I have tested this with Oculus Touch controllers and also with a simple Android app that uses the OS's built-in [AR image tracking](https://developers.google.com/ar/develop/c/augmented-images) to track the position and orientation of a comically large paper tracker taped to an Xbox controller:

image

- See video here: https://www.youtube.com/watch?v=Cv8hhsi_Mnc
    - 0.00: 2006-style setup, with sensor bar beneath 40-inch TV, both 2m away from user
    - 0.52: Since it's VR, we can move the TV closer and make it bigger without it blocking the simulated IR line of sight
    - 1.18: Image tracker + Xbox One controller. This pad has no IMU; the motion and pointing is calculated from the AR paper tracker's position and orientation.

Since the AR example is pretty flaky in practice, I added an **Input Confidence** input, the idea being that the DSU server would pass through a number expressing how much confidence it had in the current packet of values, with the number being lower in low light, or if the tracker is further away, or moving quickly, or similar. 

At the moment, the implementation is simple: assuming both **Point** and **Sensor Bar** panels are enabled, then if **Input Confidence** is `95%` or above the **Sensor Bar** values are used, otherwise the calculated gyro values are used.

Some technical notes:

- This is my first time trying to contribute to a big C++ project like this; I'm unfamiliar with both the language and the codebase so while I have tried to match the coding style I'm sure I've made some mistakes - sorry!
- I've named the panel and the relevant bits of code "sensor bar" even though this is really about the IR camera, because there's already references to "cursor" and "IMUCamera" and so on and I thought this might help express that the purpose is to simulate the IR camera's ability to see the sensor bar. I appreciate it's confusing though.
- I may have introduced some redundancies by adding `orientation` and `position` variables when there are already `rotation` and `m_point_state.position` etc; I was struggling to figure out exactly how it all tied together so that I could fit this new stuff in without breaking what was already there.
    - Untangling this would be really helpful for improving the "fall back to gyro pointing" feature, but isn't strictly necessary.

Hope this is OK, looking forward to your feedback, thanks.

### DIFF
--- a/Source/Core/Core/HW/WiimoteEmu/Dynamics.h
+++ b/Source/Core/Core/HW/WiimoteEmu/Dynamics.h
@@ -15,6 +15,7 @@
 #include "InputCommon/ControllerEmu/ControlGroup/IMUAccelerometer.h"
 #include "InputCommon/ControllerEmu/ControlGroup/IMUCursor.h"
 #include "InputCommon/ControllerEmu/ControlGroup/IMUGyroscope.h"
+#include "InputCommon/ControllerEmu/ControlGroup/SensorBar.h"
 #include "InputCommon/ControllerEmu/ControlGroup/Tilt.h"
 
 namespace WiimoteEmu
@@ -45,6 +46,10 @@ struct IMUCursorState
 
   // Rotation of world around device.
   Common::Quaternion rotation;
+  // Orientation of device in world, relative to sensor bar
+  Common::Vec3 orientation;
+  // Position of device in world, relative to sensor bar
+  Common::Vec3 position;
 
   float recentered_pitch = {};
 };
@@ -86,7 +91,9 @@ void EmulateSwing(MotionState* state, ControllerEmu::Force* swing_group, float t
 void EmulatePoint(MotionState* state, ControllerEmu::Cursor* ir_group, float time_elapsed);
 void EmulateIMUCursor(IMUCursorState* state, ControllerEmu::IMUCursor* imu_ir_group,
                       ControllerEmu::IMUAccelerometer* imu_accelerometer_group,
-                      ControllerEmu::IMUGyroscope* imu_gyroscope_group, float time_elapsed);
+                      ControllerEmu::IMUGyroscope* imu_gyroscope_group,
+                      ControllerEmu::SensorBar* sensor_bar_group,
+                      float time_elapsed);
 
 // Convert m/s/s acceleration data to the format used by Wiimote/Nunchuk (10-bit unsigned integers).
 WiimoteCommon::AccelData ConvertAccelData(const Common::Vec3& accel, u16 zero_g, u16 one_g);
--- a/Source/Core/Core/HW/WiimoteEmu/WiimoteEmu.cpp
+++ b/Source/Core/Core/HW/WiimoteEmu/WiimoteEmu.cpp
@@ -47,6 +47,7 @@
 #include "InputCommon/ControllerEmu/ControlGroup/IMUCursor.h"
 #include "InputCommon/ControllerEmu/ControlGroup/IMUGyroscope.h"
 #include "InputCommon/ControllerEmu/ControlGroup/ModifySettingsButton.h"
+#include "InputCommon/ControllerEmu/ControlGroup/SensorBar.h"
 #include "InputCommon/ControllerEmu/ControlGroup/Tilt.h"
 
 namespace WiimoteEmu
@@ -238,6 +239,9 @@ Wiimote::Wiimote(const unsigned int index) : m_index(index)
                         _trans("Camera field of view (affects sensitivity of pointing).")},
                        fov_default.y, 0.01, 180);
 
+  groups.emplace_back(m_sensor_bar =
+                          new ControllerEmu::SensorBar("SensorBar", _trans("Sensor Bar")));
+
   // Extension
   groups.emplace_back(m_attachments = new ControllerEmu::Attachments(_trans("Extension")));
   m_attachments->AddAttachment(std::make_unique<WiimoteEmu::None>());
@@ -337,6 +341,8 @@ ControllerEmu::ControlGroup* Wiimote::GetWiimoteGroup(WiimoteGroup group) const
     return m_imu_gyroscope;
   case WiimoteGroup::IMUPoint:
     return m_imu_ir;
+  case WiimoteGroup::SensorBar:
+    return m_sensor_bar;
   default:
     assert(false);
     return nullptr;
@@ -724,7 +730,7 @@ void Wiimote::StepDynamics()
   EmulateTilt(&m_tilt_state, m_tilt, 1.f / ::Wiimote::UPDATE_FREQ);
   EmulatePoint(&m_point_state, m_ir, 1.f / ::Wiimote::UPDATE_FREQ);
   EmulateShake(&m_shake_state, m_shake, 1.f / ::Wiimote::UPDATE_FREQ);
-  EmulateIMUCursor(&m_imu_cursor_state, m_imu_ir, m_imu_accelerometer, m_imu_gyroscope,
+  EmulateIMUCursor(&m_imu_cursor_state, m_imu_ir, m_imu_accelerometer, m_imu_gyroscope, m_sensor_bar,
                    1.f / ::Wiimote::UPDATE_FREQ);
 }
 
@@ -745,7 +751,7 @@ Common::Vec3 Wiimote::GetAngularVelocity(Common::Vec3 extra_angular_velocity) co
                              m_point_state.angular_velocity + extra_angular_velocity);
 }
 
-Common::Matrix44 Wiimote::GetTransformation(const Common::Matrix33& extra_rotation) const
+Common::Matrix44 Wiimote::GetTransformation(const Common::Matrix33& extra_rotation, const Common::Vec3& position) const
 {
   // Includes positional and rotational effects of:
   // Point, Swing, Tilt, Shake
@@ -755,7 +761,7 @@ Common::Matrix44 Wiimote::GetTransformation(const Common::Matrix33& extra_rotati
          Common::Matrix44::FromMatrix33(extra_rotation * GetRotationalMatrix(-m_tilt_state.angle) *
                                         GetRotationalMatrix(-m_point_state.angle) *
                                         GetRotationalMatrix(-m_swing_state.angle)) *
-         Common::Matrix44::Translate(-m_swing_state.position - m_point_state.position);
+         Common::Matrix44::Translate(position - m_swing_state.position - m_point_state.position);         
 }
 
 Common::Quaternion Wiimote::GetOrientation() const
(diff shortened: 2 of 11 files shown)


iwubcode:
@MichaelJW - thanks for this!!  At your earliest convenience please fix the conflict.  Then we can build this for people to try out.


Also @jordan-woyak , this is what I was talking about.  Please have a look and see what you think.

Source/Core/DolphinLib.vcxproj
```diff
@@ -27,6 +27,12 @@
       <Project>{41279555-f94f-4ebc-99de-af863c10c5c4}</Project>
     </ProjectReference>
   </ItemGroup>
+  <ItemGroup>
+    <ClCompile Include="InputCommon\ControllerEmu\ControlGroup\SensorBar.cpp" />
```
iwubcode, line 31: These includes should actually go in https://github.com/dolphin-emu/dolphin/blob/master/Source/Core/DolphinLib.props

Source/Core/InputCommon/ControllerEmu/ControlGroup/SensorBar.h
```diff
@@ -0,0 +1,38 @@
+#pragma once
```
iwubcode, line 1: This pragma can be removed since it is already below

Source/Core/Core/HW/WiimoteEmu/WiimoteEmu.cpp
```diff
@@ -782,9 +788,8 @@ Common::Vec3 Wiimote::GetTotalAngularVelocity() const
 
 Common::Matrix44 Wiimote::GetTotalTransformation() const
 {
-  return GetTransformation(Common::Matrix33::FromQuaternion(
-      m_imu_cursor_state.rotation *
-      Common::Quaternion::RotateX(m_imu_cursor_state.recentered_pitch)));
+  return GetTransformation(GetRotationalMatrix(m_imu_cursor_state.orientation),
```
iwubcode, line 791: If the cursor rotation isn't used here where is it used?  Seems like you need to calculate orientation (for when sensor bar isn't enabled) up above.

Source/Core/Core/HW/WiimoteEmu/Dynamics.h
```diff
@@ -45,6 +46,10 @@ struct IMUCursorState
 
   // Rotation of world around device.
   Common::Quaternion rotation;
```
iwubcode, line 48: See my other comments as well.  If you're going to use orientation for the cursor, you probably want to remove this.  Or at least, if I'm understanding correctly, I wouldn't have both.

Source/Core/Core/HW/WiimoteEmu/Dynamics.h
```diff
@@ -45,6 +46,10 @@ struct IMUCursorState
...
+  // Orientation of device in world, relative to sensor bar
+  Common::Vec3 orientation;
```
iwubcode, line 50: Feels like this should be a `Quaternion`

iwubcode:
Left some comments.  In general I think this looks fine but @jordan-woyak is the keeper of the Wiimote code :)

One other thing, I think the "Input Confidence" is too specific to one case and will likely be very confusing to users for general use.  I don't have a great solution on an alternative.  Only thing I can think of is adding a toggle hotkey instead?  It would require more configuration by the user but it's a very specific situation so that seems ok.  jordan may have a better idea.

Source/Core/InputCommon/ControllerEmu/ControlGroup/SensorBar.h
```diff
@@ -0,0 +1,38 @@
...
+public:
+  using StateData = Common::Vec3;
+
+  SensorBar(std::string name, std::string ui_name);
+
+  Common::Vec3 GetIRCameraDisplacement();
+  Common::Vec3 GetIRCameraOrientation();
+  double GetInputConfidence();
+
+  SettingValue<double> m_ir_camera_displacement_x_setting;
```
lioncash, line 30: Shouldn't these member variables be private?

Source/Core/InputCommon/ControllerEmu/ControlGroup/SensorBar.h
```diff
@@ -0,0 +1,38 @@
...
+namespace ControllerEmu
+{
+class SensorBar : public ControlGroup
+{
+public:
+  using StateData = Common::Vec3;
+
+  SensorBar(std::string name, std::string ui_name);
+
+  Common::Vec3 GetIRCameraDisplacement();
```
lioncash, line 26:
```suggestion
  Common::Vec3 GetIRCameraDisplacement() const;
```

Ditto for the other two getters.

MichaelJW:
Thank you both! I will take a look at these the next chance I get, possibly tonight.

RinMaru:
Could this be used to remove the need of  censor bar for wiimote plus? Or is this just intended for regular controllers?

JosJuice:
Dolphin already has support for emulating the pointer using MotionPlus to get rid of the need for a sensor bar.

MichaelJW:
Yeah, this does almost the opposite - it lets you _add_ input derived from an external sensor, rather than emulating the pointer.

jordan-woyak:
I don't think the names of these mappings you've introduced should mention "IR" or "Camera" or "Sensor Bar" or anything like that.
They should just be "Position" and "Orientation" (or maybe "Rotation").
OpenXR input integration is also going to need these mappings as it provides position and orientation but no accel/gyro data.
This brings up another point. I think changes in these values should generate accel/gyro data if the raw accel/gyro mappings are not active. OpenXR will need this. I suppose that can be added later.

All other bidirectional mappings in Dolphin provide one mapping for each direction to avoid the confusion of "which way is positive and which way is negative".
This pairs with input backends providing an input for each direction.
I realize doing this would create a giant wall of mappings but I don't like the inconsistency.
Grouping them into "Position" and "Orientation" would at least create another column.

ControllerInterface's Wiimote backend (aka Hybrid Wiimote 2.0) currently has no feasible way to have IR passthrough. The IR data comes from the camera mounted on the remote and is not equivalent to pitch/yaw/roll. We should be able to pass raw IR directly to the emulated remote without applying orientation corrections and and the reverse corrections within the emulated remote code. We probably should have mappings to pass through the IR data but this is not really the same as position and orientation which you've introduced. This can of course be done in another PR, but IR is not position/orientation. :P

jordan-woyak:
I'm concerned about the order of applying the rotations. There isn't really a standard order of pitch/roll/yaw. You're indirectly using `GetRotationalMatrix` which chose an order just to go with emulated movements making some sense. Euler angles are not an ideal way to describe an orientation but I don't know if there is a better solution here. I suppose there could be an option to choose a different axis order if it becomes a problem.

MichaelJW:
Thank you @jordan-woyak, this is really helpful feedback.

I've reflected on all the above and here's what I'm thinking now:

* I agree with the problems of using Euler angles as well; I'm converting from Quaternion to Euler on the DSU server side already, so I would prefer to pass Quaternion values through, but that feels like it would be unintuitive for most users, and also difficult to implement because a DSU packet only offers six float fields.
* Naming this simply "Position" and "Orientation" rather than referring to IR or sensor bar or whatever makes a lot of sense, much more intuitive. 
    * However, I do think that the IR camera's position needs to be considered separately _somewhere_. For instance, if you "yaw" a Wiimote, the camera could move a few cm while the accelerometer barely moves at all. `CameraLogic::Update()` needs to know the position of the IR camera specifically - but using that position to derive accelerometer values would be incorrect.
    * (This could be solved by stating that position is based on the IMU position, say, and then applying an within the code to get the IR camera position for the camera logic, but it feels like things are getting even messier at this point.)
* I really like the idea of using this absolute position/rotation data to generate accelerometer and gyroscope values if no IMU is present, but realistically I think only VR controllers are likely to provide enough fidelity to make this work - the paper tracker setup I was testing certainly wouldn't be good enough for real usage.

It feels like trying to use position and orientation "for simplicity" has ended up making things much more complicated than necessary, so IMO the best thing to do next here is scrap the whole idea of passing through absolute position and orientation (and of derving accelerometer, gyroscope, and IR camera data from that). As you note, passing through that data will need to be done as part of any OpenXR implementation later anyway, and like I say it's unlikely that any non-XR controller would be capable of supporting this.

Instead, I suggest a new panel with six binds, three values for each of two camera points: `x` (either 0-1023 or 0-1), `y` (either 0-767 or 0-1), and `size` (either 0-15 or 0-1). If this panel is enabled, then `CameraLogic::Update` would use these values directly instead of calculating them based on the camera transform matrix.

This puts the responsibility for dealing with the IR camera position offset and the correct rotation order on the DSU server, rather than on Dolphin. Both my VR interface and @iwubcode's OpenTrackDSUServer could be modified to generate this "points" data without too much trouble - the logic from `CameraLogic::Update()` can be re-used here. The main drawback I can see is that inputs like "shake" that modify the camera transform matrix would no longer function - but "shake" at least would be fairly simple to modify to work with this setup.

What do you think? I will try to implement a demo of this over the weekend if I can find the time.

PR #9608, opened 2021-03-26
DSPHLE: Don't restore the last UCode when the crc doesn't match

Bonta0:
This performs a sanity check in case that assumption proves to be wrong, which can happen when connecting multiple GBAs simultaneously

### DIFF
--- a/Source/Core/Core/HW/DSPHLE/DSPHLE.cpp
+++ b/Source/Core/Core/HW/DSPHLE/DSPHLE.cpp
@@ -79,15 +79,16 @@ void DSPHLE::SwapUCode(u32 crc)
 {
   m_mail_handler.Clear();
 
-  if (m_last_ucode == nullptr)
+  if (m_last_ucode && UCodeInterface::GetCRC(m_last_ucode.get()) == crc)
   {
-    m_last_ucode = std::move(m_ucode);
-    m_ucode = UCodeFactory(crc, this, m_wii);
-    m_ucode->Initialize();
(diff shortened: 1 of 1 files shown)


JMC47:
@dolphin-emu-bot rebuild

Source/Core/Core/HW/DSPHLE/DSPHLE.cpp
```diff
@@ -79,9 +79,10 @@ void DSPHLE::SwapUCode(u32 crc)
 {
   m_mail_handler.Clear();
 
-  if (m_last_ucode == nullptr)
+  if (!m_last_ucode || UCodeInterface::GetCRC(m_last_ucode.get()) != crc)
```
leoetlino, line 82: For clarity, maybe invert the conditional to avoid the extra negations
```suggestion
  if (m_last_ucode && UCodeInterface::GetCRC(m_last_ucode.get()) == crc)
  {
    m_ucode = std::move(m_last_ucode);
  }
  ...
```

leoetlino (approved):

Source/Core/Core/HW/DSPHLE/DSPHLE.cpp
```diff
@@ -79,15 +79,16 @@ void DSPHLE::SwapUCode(u32 crc)
 {
```
shuffle2, line 79: as the comment above states, this was always kinda hacky. Now the behavior doesn't really match the comment, either. Maybe a list should be kept (instead of implicit 2-entry list)?

PR #9609, opened 2021-03-26
Fix PanicAlert when recording FIFOs

Pokechu22:
The typo (from f749fcfa9f42cda679ff7f4288c8418d2b56b08d) was `ARRAY_POSITION` being used instead of `ARRAY_TEXCOORD0`.

### DIFF
--- a/Source/Core/VideoCommon/CPMemory.h
+++ b/Source/Core/VideoCommon/CPMemory.h
@@ -51,9 +51,14 @@ enum
 {
   ARRAY_POSITION = 0,
   ARRAY_NORMAL = 1,
-  ARRAY_COLOR = 2,
-  ARRAY_COLOR2 = 3,
+  ARRAY_COLOR0 = 2,
+  NUM_COLOR_ARRAYS = 2,
   ARRAY_TEXCOORD0 = 4,
+  NUM_TEXCOORD_ARRAYS = 8,
+
+  // Number of arrays related to vertex components (position, normal, color, tex coord)
+  // Excludes the 4 arrays used for indexed XF loads
+  NUM_VERTEX_COMPONENT_ARRAYS = 12,
 };
 
 // Vertex components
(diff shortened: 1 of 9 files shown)


Source/Core/VideoCommon/CPMemory.h
```diff
@@ -42,6 +42,7 @@ enum
   CP_COMMAND_MASK = 0xf0,
   CP_NUM_VAT_REG = 0x08,
   CP_VAT_MASK = 0x07,
+  CP_NUM_REGULAR_ARRAYS = 12,  // Excludes the arrays used for indexed XF loads
```
leoetlino, line 45: IMO it would make more sense to have this be a separate (constexpr) constant rather than part of the enum, especially since all the other values are written in base 16.

Pokechu22: It could be a separate constant, or maybe it could be in the `enum` below.  Also, looking at it, perhaps `NUM_VERTEX_ARRAYS` would be better (as each of those 12 arrays are vertex components). #9540 also adds constants for the indexed XF ones as well.

leoetlino: I'd opt for a constant, but a separate enum with only NUM_* constants sounds fine to me too.

Pokechu22: How do you feel about the `NUM_COLOR_ARRAYS` and `NUM_TEXCOORD_ARRAYS` constants I added next to `ARRAY_COLOR0` and `ARRAY_TEXCOORD0` in the first commit then?

leoetlino: Ambivalent about it -- on the one hand I don't really like indices and sizes/numbers being in the same enum, on the other hand it makes sense to have the number of arrays right next to the starting index :man_shrugging:

Pokechu22: I tried a separate enum, but it turns out MSVC doesn't like it when you add elements of different enums.

Converting all of those into scoped enums or proper constants (and having them not be in the global namespace) is something that should be done eventually, but will be a bit more refactoring than I want to handle here.

leoetlino: Fair enough. Let's fix the regression and do further cleanup in a followup PR :)

leoetlino (approved):

PR #9610, opened 2021-03-27
Fix shortcut creation

CookiePLMonster:
This PR fixes the following:
1. Selected shortcut names did not use the Title Database, making them consistent with Dolphin's game list and sometimes picking rather unfortunate names, depending on game's metadata.
2. Shortcuts of games containing invalid NTFS characters in their name (most commonly `:`) would create incorrectly. These characters are now being removed from the file name.


Source/Core/DolphinQt/GameList/GameList.cpp
```diff
@@ -685,7 +685,18 @@ bool GameList::AddShortcutToDesktop()
   if (FAILED(SHGetKnownFolderPath(FOLDERID_Desktop, KF_FLAG_NO_ALIAS, nullptr, &desktop)))
     return false;
 
-  const auto& game_name = game->GetLongName();
+  std::string game_name = game->GetName(Core::TitleDatabase());
+  // Sanitize the string by removing all characters that cannot be used in NTFS file names
+  game_name.erase(std::remove_if(game_name.begin(), game_name.end(),
+                                 [](char ch) {
+                                   const char illegal_characters[] = {'<',  '>', ':', '\"', '/',
```
lioncash, line 692: Perhaps make this `static constexpr char`, so this array doesn't get constructed on the stack for every execution of the lambda (which is currently what MSVC does)

leoetlino (approved):

PR #9616, opened 2021-03-30
Core/HotkeyManager: Use hotkey string for all keys

Techjar:
This fixes the issue with the defaults where unrelated keys will trigger some hotkeys, such as Alt+F4 loading a save state.

Should fix https://bugs.dolphin-emu.org/issues/12466


Techjar:
This will not work until the issue where hotkeys are not aware of modifiers with no bindings is fixed.

Techjar:
I have been informed that this does essentially nothing, so I guess I just didn't fully understand how the hotkey expression works. A new solution has been proposed: making a Qt keyboard/mouse backend that won't have the issue of reading OS hotkeys.

PR #9617, opened 2021-03-31
Allow binding "camera points" directly

MichaelJW:
Based on feedback to #9607, this PR takes a different approach to solving the same problem of wanting to accurately simulate the way a Wiimote's IR camera sees the sensor bar for situations where the hardware allows for that accuracy.

This adds a new panel, **Sensor Bar**, to the **Motion Input** settings:

image

If this group is enabled, then `CameraLogic` draws size and coords of the two "camera points" directly from these bindings, rather than calculating where the points would appear based on position and orientation of the IR camera relative to the sensor bar.

This commit splits the existing `CameraLogic::Update` function into three: `CameraLogic::UpdateFromCameraTransform()` (containing existing behaviour), `CameraLogic::UpdateFromSensorBarPoints()` (containing new behaviour), and `CameraLogic::Update()` (containing what the two have in common).

Technical points and questions:

- The `camera_points` array size is defined in different ways in different places: in one place it's hard-coded as `2`, in another it's based on the size of `leds`, and elsewhere it's based on `NUM_POINTS`. I appreciate this is messy, but does it need to be cleaned up? If so, should it make a hard assumption that the array will always have two elements, or should the array be padded with "blanks" to make the length always `NUM_POINTS`?
- I would like to add a widget similar to that in **Motion Simulation > Point** but haven't got my head around how to do that yet. Is that necessary for this PR, or just a nice-to-have?
- Because the camera transform is not used at all if this panel is enabled, anything that simulates controller motion (like "Shake") will not affect the camera input either.


BhaaLseN:
You have a lot of unrelated history in there, could you please run a rebase and get rid of those?

----
Please rebase your branch instead of merging (`git fetch upstream; git rebase upstream/master sensor-bar-2` - replace `upstream` with the name of the dolphin remote). This also gives you the option of [combining commits](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History#_squashing) if you use `git rebase -i` instead.

jordan-woyak (commented):
I don't think I care for exposing the position of individual points.

No hardware generates IR data like this other than actual Wii Remotes.
The only reasonable use of this feature is in conjunction with ControllerInterface's Wii Remote input backend for passthrough of IR data.

Any other motion hardware is going to use Accel/Gyro or Position/Orientation bindings.
If you were planning on having specialized input servers generate simulated IR data I think that is a weird/bad idea.
Dolphin is already generating this data in a manner that is coherent with other motion data.
Duplicating this logic and exposing the internal features of the Wii Remote is unnecessary in this situation.
Supplying Position/Orientation makes more sense and it's a less confusing user interface.
It also allows the "FOV" settings to still have an effect.

My plan for IR passthrough was to expose bindings just for the center of all IR points (Up Down Left Right), bindings for IR tilt (Left and Right), and a setting for the distance from the sensor bar in meters (something that could also be useful outside of IR passthrough).
I have the required changes for Controller Interface's Wii Remote to expose these inputs implemented locally.
This would allow for near-perfect IR passthrough without too ugly of a user interface and minimal intermediary calculations on both ends of the interface.

Source/Core/InputCommon/ControllerEmu/ControlGroup/ControlGroup.h
```diff
@@ -43,7 +43,8 @@ enum class GroupType
   Shake,
   IMUAccelerometer,
   IMUGyroscope,
-  IMUCursor
+  IMUCursor,
+  SensorBar
```
Rumi-Larry, line 47: An extra comma is recommended for future diffs

MichaelJW:
> My plan for IR passthrough was to expose bindings just for the center of all IR points (Up Down Left Right), bindings for IR tilt (Left and Right), and a setting for the distance from the sensor bar in meters (something that could also be useful outside of IR passthrough).
> I have the required changes for Controller Interface's Wii Remote to expose these inputs implemented locally.
> This would allow for near-perfect IR passthrough without too ugly of a user interface and minimal intermediary calculations on both ends of the interface.

Nice! Yeah, agreed, that's way better. I look forward to seeing that implemented 🙂

PR #9618, opened 2021-04-01
JitArm64: Allow DoJit at address 0 (fix launching Wii titles)

JosJuice:
`JitArm64::DoJit` contains a check where it prints a warning and tries to pause emulation if instructed to compile code at address 0. I'm assuming this was done in order to provide a nicer error behavior in cases where PC was accidentally set to null. Unfortunately, it has started causing us problems recently, as 688bd61 writes and runs some code at address 0 to simulate the PPC being held in reset. What makes this worse is that calling `Core::SetState` from the CPU thread is actually not allowed and will cause a deadlock instead of  the intended behavior. I don't believe there is anything on a real console that would stop you from executing code at address 0 (as long as the MMU has been set up to allow it), and `Jit64::DoJit` doesn't contain any check like this, so let's remove the check.

### DIFF
--- a/Source/Core/Core/PowerPC/JitArm64/Jit.cpp
+++ b/Source/Core/Core/PowerPC/JitArm64/Jit.cpp
@@ -623,12 +623,6 @@ void JitArm64::Jit(u32)
 
 void JitArm64::DoJit(u32 em_address, JitBlock* b, u32 nextPC)
 {
-  if (em_address == 0)
-  {
-    Core::SetState(Core::State::Paused);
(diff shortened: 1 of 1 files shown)


leoetlino (approved):
Whoops, didn't notice this was in JitArm64, sorry :/

degasus:
0x0 is a valid memory address, both with IR=0 and (maybe) IR=1.

StLouisCPhT:
> Whoops, didn't notice this was in JitArm64, sorry :/

Honestly, that is completely understandable. JitArm64 is just as cursed as ES_Launch :)

PR #9619, opened 2021-04-01
IOS/FS: Add a scoped FD class to make it harder to leak FDs

leoetlino:
This changes FileSystemProxy::Open to return a file descriptor wrapper
that will ensure the FD is closed when it goes out of scope.

By using such a wrapper we make it more difficult to forget to close
file descriptors.

This fixes a leak in ReadBootContent. I should have added such a class
from the beginning... In practice, I don't think this would have caused
any obvious issue because ReadBootContent is only called after an IOS
relaunch -- which clears all FDs -- and most titles do not get close
to the FD limit.

### DIFF
--- a/Source/Core/Core/IOS/FS/FileSystemProxy.h
+++ b/Source/Core/Core/IOS/FS/FileSystemProxy.h
@@ -24,12 +24,40 @@ constexpr FS::Fd INVALID_FD = 0xffffffff;
 class FSDevice : public Device
 {
 public:
+  class ScopedFd
+  {
+  public:
+    ScopedFd(FSDevice* fs, s64 fd, Ticks tick_tracker = {})
+        : m_fs{fs}, m_fd{fd}, m_tick_tracker{tick_tracker}
+    {
+    }
+
+    ~ScopedFd()
+    {
+      if (m_fd >= 0)
+        m_fs->Close(m_fd, m_tick_tracker);
+    }
+
+    ScopedFd(const ScopedFd&) = delete;
+    ScopedFd(ScopedFd&&) = delete;
+    ScopedFd& operator=(const ScopedFd&) = delete;
+    ScopedFd& operator=(ScopedFd&&) = delete;
+
+    s64 Get() const { return m_fd; }
+    s64 Release() { return std::exchange(m_fd, -1); }
+
+  private:
+    FSDevice* m_fs{};
+    s64 m_fd = -1;
+    Ticks m_tick_tracker{};
+  };
+
   FSDevice(Kernel& ios, const std::string& device_name);
 
   // These are the equivalent of the IPC command handlers so IPC overhead is included
   // in timing calculations.
-  s64 Open(FS::Uid uid, FS::Gid gid, const std::string& path, FS::Mode mode,
-           std::optional<u32> ipc_fd = {}, Ticks ticks = {});
+  ScopedFd Open(FS::Uid uid, FS::Gid gid, const std::string& path, FS::Mode mode,
+                std::optional<u32> ipc_fd = {}, Ticks ticks = {});
   s32 Close(u64 fd, Ticks ticks = {});
   s32 Read(u64 fd, u8* data, u32 size, std::optional<u32> ipc_buffer_addr = {}, Ticks ticks = {});
   s32 Write(u64 fd, const u8* data, u32 size, std::optional<u32> ipc_buffer_addr = {},
--- a/Source/Core/Core/IOS/ES/ES.cpp
+++ b/Source/Core/Core/IOS/ES/ES.cpp
@@ -403,9 +403,9 @@ bool ESDevice::LaunchPPCTitle(u64 title_id)
   // To keep track of the PPC title launch, a temporary launch file (LAUNCH_FILE_PATH) is used
   // to store the title ID of the title to launch and its TMD.
   // The launch file not existing means an IOS reload is required.
-  const auto launch_file_fd = m_ios.GetFSDevice()->Open(PID_KERNEL, PID_KERNEL, LAUNCH_FILE_PATH,
-                                                        FS::Mode::Read, {}, &ticks);
-  if (launch_file_fd < 0)
+  if (const auto launch_file_fd = m_ios.GetFSDevice()->Open(
+          PID_KERNEL, PID_KERNEL, LAUNCH_FILE_PATH, FS::Mode::Read, {}, &ticks);
+      launch_file_fd.Get() < 0)
   {
     if (WriteLaunchFile(tmd, &ticks) != IPC_SUCCESS)
     {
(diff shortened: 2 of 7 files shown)


Source/Core/Core/IOS/FS/FileSystemProxy.h
```diff
@@ -24,12 +24,50 @@ constexpr FS::Fd INVALID_FD = 0xffffffff;
...
+    {
+    }
+
+    ~ScopedFd()
+    {
+      if (m_fd >= 0)
+        m_fs->Close(m_fd, m_tick_tracker);
+    }
+
+    ScopedFd(ScopedFd&& other) noexcept { *this = std::move(other); }
```
iwubcode, line 41: Can you implement or delete the copy constructor?

Source/Core/Core/IOS/FS/FileSystemProxy.h
```diff
@@ -24,12 +24,50 @@ constexpr FS::Fd INVALID_FD = 0xffffffff;
...
+      return *this;
+    }
+
+    s64 Get() const { return m_fd; }
+
+    s64 Release()
+    {
+      const s64 fd = m_fd;
+      m_fd = -1;
+      return fd;
```
Dentomologist, line 56:
```
return std::exchange(m_fd, -1);
```

Source/Core/Core/IOS/FS/FileSystemProxy.h
```diff
@@ -24,12 +24,50 @@ constexpr FS::Fd INVALID_FD = 0xffffffff;
...
+    {
+      m_fs = other.m_fs;
+      m_fd = other.Release();
+      m_tick_tracker = other.m_tick_tracker;
+      return *this;
+    }
+
+    s64 Get() const { return m_fd; }
+
+    s64 Release()
```
iwubcode, line 52: Can this be moved to be `private` ?

iwubcode: nm, used elsewhere

Source/Core/Core/IOS/FS/FileSystemProxy.h
leoetlino: Both the copy constructor and copy assignment operators are automatically deleted because there is a user-defined move constructor or move assignment operator.

iwubcode: True.  I generally feel like it's good practice to specify them all for completeness, see: https://en.cppreference.com/w/cpp/language/rule_of_three (rule of five)

Source/Core/Core/IOS/FS/FileSystemProxy.h
```diff
@@ -24,12 +24,50 @@ constexpr FS::Fd INVALID_FD = 0xffffffff;
...
+    }
+
+    ScopedFd(ScopedFd&& other) noexcept { *this = std::move(other); }
+    ScopedFd& operator=(ScopedFd&& other) noexcept
+    {
+      m_fs = other.m_fs;
+      m_fd = other.Release();
+      m_tick_tracker = other.m_tick_tracker;
+      return *this;
+    }
```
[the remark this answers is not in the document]
leoetlino: The purpose of implementing the move constructor is to make moves do the right thing (clear the FD in the move source to prevent a double close) and make this a proper move only type, not to optimise.

iwubcode:
General thoughts on the class.  I think the `Release()` function is a bit odd.  Like you have to remember to call it to properly close the object.  Maybe it doesn't make sense from an emulation standpoint but I feel like code wise it would make more sense to move the `ScopedFd` object to any dependent objects and then just have the destructor naturally call close.

Ex:

```
 ~ScopedFd()
    {
        m_fs->Close(m_fd, m_tick_tracker);
    }
```

```
return MakeIPCReply([&](Ticks t) {
    return Open(request.uid, request.gid, request.path, static_cast<Mode>(request.flags & 3),
                request.fd, t);
                request.fd, t)
  });
```

and

```
entry.m_opened = true;
    entry.m_fd = std::move(fd);  // m_fd is now a ScopedFd instead of an int
    entry.m_content = content;
    entry.m_title_id = title_id;
    entry.m_uid = uid;
    INFO_LOG_FMT(IOS_ES,
                 "OpenContent: title ID {:016x}, UID {:#x}, content {:08x} (index {}) -> CFD {}",
                 title_id, uid, content.id, content_index, i);
    return static_cast<s32>(i);
```

leoetlino:
> Like you have to remember to call it to properly close the object

No, Release does the opposite: it's like `unique_ptr<T>::release`, it relinquishes control of the file descriptor so that the caller is now responsible for closing the file (with unique_ptr you'd need to delete the pointer yourself after calling release).

It is necessary when handles are directly exposed to the emulated console, because what is responsible for closing the handle is emulated code, *not* Dolphin.

Source/Core/Core/IOS/FS/FileSystemProxy.h
```diff
@@ -24,12 +24,50 @@ constexpr FS::Fd INVALID_FD = 0xffffffff;
...
+  public:
+    ScopedFd(FSDevice* fs, s64 fd, Ticks tick_tracker = {})
+        : m_fs{fs}, m_fd{fd}, m_tick_tracker{tick_tracker}
+    {
+    }
+
+    ~ScopedFd()
+    {
+      if (m_fd >= 0)
+        m_fs->Close(m_fd, m_tick_tracker);
```
iwubcode, line 38: Should we add an error here like we do for `WriteSystemFile` ?

leoetlino: IMO no, it's not really useful. If we are relying on the scoped behavior, we pretty much don't care whether Close() fails, as long as the resource is cleaned up :)

leoetlino:
All comments addressed, I think.

Source/Core/Core/IOS/FS/FileSystemProxy.h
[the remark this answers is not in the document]
Dentomologist: What I meant to say is that I don't believe the functions are being used now; I tried deleting them and it compiled fine. Is that on purpose and you implemented them for future support?

For the move assignment operator, suppose somebody does the following:
```
auto foobar = fs->Open(/* file foo.txt */);
foobar = fs->Open(/* file bar.txt */);
```
As implemented the descriptor for foo.txt will be left open, which might be an unintended leak of the sort ScopedFd is intended to prevent, but changing it to close foo.txt might be unexpected behavior in other cases.

I think the semantics of the assignment operator in that regard are fundamentally unclear and so it should be deleted.  It's not used at the moment, but if you want to keep the ability to assign (as opposed to copy constructing) ScopedFd values I would create CloseAndOverwrite and ReleaseAndOverwrite (or similar) and use those instead.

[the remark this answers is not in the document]
leoetlino: Ah, good point. I'll just delete it then since it's not used at the moment.

PR #9621, opened 2021-04-01
Replace uses of cassert with Common/Assert.h

Pokechu22:
This replaces the standard library `assert` with Dolphin's `ASSERT` macro.

I haven't changed any uses of cassert or assert.h in Externals, nor in pch (because the precompiled headers are used for externals, I'm pretty sure).

### DIFF
--- a/Source/Core/Core/HW/WiimoteEmu/Extension/Guitar.cpp
+++ b/Source/Core/Core/HW/WiimoteEmu/Extension/Guitar.cpp
@@ -5,10 +5,10 @@
 #include "Core/HW/WiimoteEmu/Extension/Guitar.h"
 
 #include <array>
-#include <cassert>
 #include <cstring>
 #include <map>
 
+#include "Common/Assert.h"
 #include "Common/BitUtils.h"
 #include "Common/Common.h"
 #include "Common/CommonTypes.h"
@@ -164,7 +164,7 @@ ControllerEmu::ControlGroup* Guitar::GetGroup(GuitarGroup group)
   case GuitarGroup::SliderBar:
     return m_slider_bar;
   default:
-    assert(false);
+    ASSERT(false);
     return nullptr;
   }
 }
--- a/Source/Core/Core/HW/WiimoteEmu/WiimoteEmu.cpp
+++ b/Source/Core/Core/HW/WiimoteEmu/WiimoteEmu.cpp
@@ -5,12 +5,12 @@
 #include "Core/HW/WiimoteEmu/WiimoteEmu.h"
 
 #include <algorithm>
-#include <cassert>
 #include <memory>
 #include <string_view>
 
 #include <fmt/format.h>
 
+#include "Common/Assert.h"
 #include "Common/CommonTypes.h"
 #include "Common/Config/Config.h"
 #include "Common/FileUtil.h"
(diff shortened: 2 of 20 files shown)


Source/Core/DiscIO/VolumeVerifier.cpp
```diff
@@ -696,7 +695,7 @@ bool VolumeVerifier::ShouldHaveChannelPartition() const
       "RFNE01", "RFNJ01", "RFNK01", "RFNP01", "RFNW01", "RFPE01", "RFPJ01", "RFPK01", "RFPP01",
       "RFPW01", "RGWE41", "RGWJ41", "RGWP41", "RGWX41", "RMCE01", "RMCJ01", "RMCK01", "RMCP01",
   };
-  assert(std::is_sorted(channel_discs.cbegin(), channel_discs.cend()));
+  DEBUG_ASSERT(std::is_sorted(channel_discs.cbegin(), channel_discs.cend()));
```
Pokechu22, line 698: This (along with `ShouldBeDualLayer`) is the only place where I used `DEBUG_ASSERT`, as this isn't an assertion that needs to be run in most cases.  In C++20, `is_sorted` becomes `constexpr`, so it could be replaced with `static_assert`.

Dentomologist:
Removing the <cassert> in pch.h shouldn't be an issue; the externals don't depend on Dolphin, so anything that needs assert already includes it directly.  I cleaned and rebuilt the VS solution to double check and it worked fine.

There's one assert left in AudioCommon/PulseAudioStream.cpp and a handful in VideoCommon/TextureDecoder_x64.cpp, though the latter are all in inactive include guards.

Pokechu22:
I'm pretty sure it would still build if cassert was removed from pch.h, but to my understanding the precompiled headers are used when building all projects (not just DolphinLib) in which case we'd still want cassert to be precompiled for faster build times.

I've fixed the PulseAudioStream and TextureDecoder_x64 ones (and tested to make sure that it works with `CHECK` defined).

Dentomologist:
Ah, I misunderstood what you were referring to in the PR.

Source/Core/Core/HW/WiimoteEmu/Extension/Guitar.cpp
```diff
@@ -164,7 +163,7 @@ ControllerEmu::ControlGroup* Guitar::GetGroup(GuitarGroup group)
   case GuitarGroup::SliderBar:
     return m_slider_bar;
   default:
-    assert(false);
+    ASSERT(false);
```
leoetlino, line 166: Missing include

Source/Core/Core/HW/WiimoteEmu/WiimoteEmu.cpp
```diff
@@ -338,7 +337,7 @@ ControllerEmu::ControlGroup* Wiimote::GetWiimoteGroup(WiimoteGroup group) const
   case WiimoteGroup::IMUPoint:
     return m_imu_ir;
   default:
-    assert(false);
+    ASSERT(false);
```
leoetlino, line 340: Missing include

leoetlino (approved):

Source/Core/DiscIO/VolumeVerifier.cpp
mitaclaw: Done by https://github.com/dolphin-emu/dolphin/pull/13084.

PR #9622, opened 2021-04-02
Check that zstd headers can be found before using a shared zstd library

skylersaleh:
Some users were struggling to build the Mac build of dolphin because Homebrew had installed a header-less version of the zstd library with Qt and were getting compilation errors that were difficult to track down. 

This pull requests modifies the CMakeLists.txt so that it checks for header files for zstd before using the shared zstd library.


JosJuice:
Where is `HAVE_ZSTD_H` getting defined?

skylersaleh:
@JosJuice Sorry, about the embarrassing mistake. It is now defined in the commit.

Rumi-Larry:
I assume having the headers without the library isn't possible. But if it is, then such a special case need to be handled, just in case.

JosJuice:
There's no reason to have special handling for that case. It'll go to "Shared zstd not found, falling back to the static library", which is the correct thing to do.

PR #9623, opened 2021-04-03
Core: Fix PauseAndLock segfaulting under certain circumstances

Techjar:
This can happen if it's called before the core has fully initialized, as is happening in https://bugs.dolphin-emu.org/issues/12473.

There might be other side effects from this, so it should probably be tested more thoroughly. Maybe need a mutex?


CookiePLMonster:
Partially overlaps with #8335, your solution is more localized though.

Techjar:
Ah, so we both stumbled across the same bug. In both PRs though, I see the potential for a very tight race condition where RunAsCPUThread can execute non-thread-safe code while the core is still booting, which might be a problem.

leoetlino:
#8335 is a more complete solution and addresses other problems but I guess it doesn't hurt to fix this more common case first.

PR #8268, opened 2019-07-28
[WIP] Balance Board emulation

Pokechu22:
This PR adds support for balance board emulation.

This isn't completely finished, and the UI for it isn't too great.  In particular, there isn't any way to specify the current weight on the main UI, only to adjust the weight in terms of percentages, though it does exist on the TAS UI.  The UI also doesn't support all possible inputs, only cases where actual balance is used; this means that e.g. walking doesn't work too well (or the step game).  The balance board is treated as a regular extension and can be set on any remote, and also needs to be manually set for the actual balance board.  Also, games that don't support the balance board will continuously disconnect it, and dolphin will then repeatedly attempt to reconnect it.  But it is perfectly usable.

One thing I haven't tested is whether this works with movies; the [documentation](http://tasvideos.org/EmulatorResources/DTM.html) ([rev 31](http://tasvideos.org/wiki.exe?page=EmulatorResources/DTM&rev=31) specifically) mentions that a single byte is used to indicate the connected controllers, and adding a 5th pseudo-wiimote probably causes issues there.  I currently have a commit that adds an excessive number of bounds-checks for movies, which ideally should be replaced with proper support.

PR #3033, opened 2015-09-12
Use official names for GameCube controller sticks

JosJuice:
Main Stick is changed to Control Stick and C-Stick is changed to C Stick.

A new ui_name variable is added to ControlGroup so that the UI strings in DolphinWX can be updated without breaking backwards compatibility with config INIs and other things that use names as IDs.

PR #9624, opened 2021-04-05
Add new "input override" system for TAS input and Android touch controls

JosJuice:
Dolphin currently has two parts of the codebase that directly map UI elements to emulated controls: TAS input in DolphinQt, and the touch overlay on Android. The two accomplish this using entirely different methods, both of which have their own drawbacks. TAS input manually pokes around inside Wii Remote input reports instead of letting this be handled by the normal input report creation code, and does so in a way that isn't always reliable. The Android touch overlay requires a fixed mapping to be present in GCPadNew.ini and WiimoteNew.ini, meaning that these two files can't be used for storing the user's actual controller mappings, which has led to the creation of a second controller mapping system implemented in Java which lacks many of the features present on PC.

This pull request tries to improve the situation by replacing both of these two systems with a new "input override" framework. The basis of it is that each `EmulatedController` lets you set a callback function which can inspect and replace the input coming from InputCommon. TAS input and Android both register such callback functions.

Getting rid of the Android touch overlay's dependency on ButtonManager like this PR does is a prerequisite for getting rid of the controller indirection on Android. I'm aiming to work on fully getting rid of the controller indirection around July if this gets merged.

One important note: This pull request breaks a certain hack that some Android users have been using in which they intentionally desync the controller used by the emulation core and the controller used by the touch overlay by first configuring one Wii Remote extension in a game INI and then configuring a different Wii Remote extension in the touch overlay settings. Normally this would result in the touch overlay not working, but what these Android users then do is to edit the fixed mappings in WiimoteNew.ini (typically by using a profile instead of by editing the file directly) to let them play games that use a Wii Remote or Wii Remote + Nunchuk using an on-screen Classic Controller, which has more buttons. (This is especially popular for Skyward Sword, because of its complex motion controls.) With this PR, the contents of WiimoteNew.ini no longer affect the touch controller, so this "trick" no longer works. Since editing the fixed mappings in WiimoteNew.ini isn't something you're intended to do, and since this PR is necessary for future improvements of the controller code on Android, I am not going to try to support this hack. The proper way to go would be to add a way for users to add custom touch overlay buttons with custom mappings, but that would have to be a future project.

@jordan-woyak I would appreciate if you could review this. Not sure if this approach is the best one possible, but I would really like to be able to get rid of the controller indirection on Android.

### DIFF
--- a/Source/Core/Core/HW/GCPadEmu.cpp
+++ b/Source/Core/Core/HW/GCPadEmu.cpp
@@ -138,14 +138,15 @@ GCPadStatus GCPad::GetInput() const
   const auto lock = GetStateLock();
   GCPadStatus pad = {};
 
-  if (!(m_always_connected_setting.GetValue() || IsDefaultDeviceConnected()))
+  if (!(m_always_connected_setting.GetValue() || IsDefaultDeviceConnected() ||
+        m_input_override_function))
   {
     pad.isConnected = false;
     return pad;
   }
 
   // buttons
-  m_buttons->GetState(&pad.button, button_bitmasks);
+  m_buttons->GetState(&pad.button, button_bitmasks, m_input_override_function);
 
   // set analog A/B analog to full or w/e, prolly not needed
   if (pad.button & PAD_BUTTON_A)
@@ -154,20 +155,20 @@ GCPadStatus GCPad::GetInput() const
     pad.analogB = 0xFF;
 
   // dpad
-  m_dpad->GetState(&pad.button, dpad_bitmasks);
+  m_dpad->GetState(&pad.button, dpad_bitmasks, m_input_override_function);
 
   // sticks
-  const auto main_stick_state = m_main_stick->GetState();
+  const auto main_stick_state = m_main_stick->GetState(m_input_override_function);
   pad.stickX = MapFloat<u8>(main_stick_state.x, GCPadStatus::MAIN_STICK_CENTER_X, 1);
   pad.stickY = MapFloat<u8>(main_stick_state.y, GCPadStatus::MAIN_STICK_CENTER_Y, 1);
 
-  const auto c_stick_state = m_c_stick->GetState();
+  const auto c_stick_state = m_c_stick->GetState(m_input_override_function);
   pad.substickX = MapFloat<u8>(c_stick_state.x, GCPadStatus::C_STICK_CENTER_X, 1);
   pad.substickY = MapFloat<u8>(c_stick_state.y, GCPadStatus::C_STICK_CENTER_Y, 1);
 
   // triggers
   std::array<ControlState, 2> triggers;
-  m_triggers->GetState(&pad.button, trigger_bitmasks, triggers.data());
+  m_triggers->GetState(&pad.button, trigger_bitmasks, triggers.data(), m_input_override_function);
   pad.triggerLeft = MapFloat<u8>(triggers[0], 0);
   pad.triggerRight = MapFloat<u8>(triggers[1], 0);
 
--- a/Source/Core/Core/HW/GCPadEmu.cpp
+++ b/Source/Core/Core/HW/GCPadEmu.cpp
@@ -36,62 +36,47 @@ static const u16 trigger_bitmasks[] = {
 static const u16 dpad_bitmasks[] = {PAD_BUTTON_UP, PAD_BUTTON_DOWN, PAD_BUTTON_LEFT,
                                     PAD_BUTTON_RIGHT};
 
-static const char* const named_buttons[] = {"A", "B", "X", "Y", "Z", "Start"};
-
-static const char* const named_triggers[] = {
-    // i18n: The left trigger button (labeled L on real controllers)
-    _trans("L"),
-    // i18n: The right trigger button (labeled R on real controllers)
-    _trans("R"),
-    // i18n: The left trigger button (labeled L on real controllers) used as an analog input
-    _trans("L-Analog"),
-    // i18n: The right trigger button (labeled R on real controllers) used as an analog input
-    _trans("R-Analog")};
-
 GCPad::GCPad(const unsigned int index) : m_index(index)
 {
   // buttons
-  groups.emplace_back(m_buttons = new ControllerEmu::Buttons(_trans("Buttons")));
-  for (const char* named_button : named_buttons)
+  groups.emplace_back(m_buttons = new ControllerEmu::Buttons(BUTTONS_GROUP));
+  for (const char* named_button : {A_BUTTON, B_BUTTON, X_BUTTON, Y_BUTTON, Z_BUTTON})
   {
-    const bool is_start = named_button == std::string("Start");
-    const ControllerEmu::Translatability translate =
-        is_start ? ControllerEmu::Translate : ControllerEmu::DoNotTranslate;
-    // i18n: The START/PAUSE button on GameCube controllers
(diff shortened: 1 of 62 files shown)


Source/Core/Core/HW/GCPadEmu.h
```diff
@@ -50,6 +52,31 @@ class GCPad : public ControllerEmu::EmulatedController
...
   static constexpr ControlState C_STICK_GATE_RADIUS = 0.7221375;
 
+  static constexpr const char* BUTTONS_GROUP = _trans("Buttons");
+  static constexpr const char* MAIN_STICK_GROUP = "Main Stick";
+  static constexpr const char* C_STICK_GROUP = "C-Stick";
+  static constexpr const char* DPAD_GROUP = _trans("D-Pad");
+  static constexpr const char* TRIGGERS_GROUP = _trans("Triggers");
+  static constexpr const char* RUMBLE_GROUP = _trans("Rumble");
+  static constexpr const char* MIC_GROUP = _trans("Microphone");
+  static constexpr const char* OPTIONS_GROUP = _trans("Microphone");
```
Rumi-Larry, line 62: There seems to be a mistake on the user facing string

Rumi-Larry:
Why the terms "Main Stick" and "C-stick" not translatable?

JosJuice:
The translatable terms are "Control Stick" and "C Stick". See PR #3033.

Rumi-Larry:
> 
> 
> The translatable terms are "Control Stick" and "C Stick". See PR #3033.

Interesting, than you!

Source/Core/InputCommon/ControllerEmu/ControlGroup/ControlGroup.h
```diff
@@ -27,6 +30,9 @@ class NumericSetting;
 template <typename T>
 class SettingValue;
 
+using InputOverrideFunction = std::function<std::optional<ControlState>(
+    const std::string& group_name, const std::string& control_name, ControlState state)>;
```
AdmiralCurtiss, line 34: Both of these strings should probably be string_views.

Source/Core/InputCommon/ControllerEmu/ControlGroup/Cursor.cpp
```diff
@@ -82,15 +82,30 @@ ControlState Cursor::GetGateRadiusAtAngle(double ang) const
...
-  }
+Cursor::StateData Cursor::GetState(const bool adjusted,
+                                   const ControllerEmu::InputOverrideFunction& override_func)
+{
+  if (!override_func)
+    return GetState(adjusted);
 
-  const auto input = GetReshapableState(true);
+  const ReshapeData input = GetReshapableState(adjusted);
+  StateData state = adjusted ? UpdateState(input) : StateData{input.x, input.y};
```
AdmiralCurtiss, line 97: Can't this be simplified to

```
StateData state = GetState(adjusted);
if (!override_func)
  return state;
```

? Same for other inputs.

AdmiralCurtiss:
The rest of the non-Android code looks fine from what I can tell, though I'm not super familiar with the input code.

Source/Core/Core/HW/GCPadEmu.cpp
```diff
@@ -138,14 +138,15 @@ GCPadStatus GCPad::GetInput() const
   const auto lock = GetStateLock();
   GCPadStatus pad = {};
 
-  if (!(m_always_connected_setting.GetValue() || IsDefaultDeviceConnected()))
+  if (!(m_always_connected_setting.GetValue() || IsDefaultDeviceConnected() ||
+        m_input_override_function))
```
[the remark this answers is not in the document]
JosJuice: I designed it this way because it's convenient for Android. Currently on Android, both GC controller 1 and Wii Remote 1 count as being connected basically all the time. This leads to a problem in games like Sonic Colors that refuse to use the Wii Remote if there is a GC controller connected – if the touch overlay is set to use the Wii Remote (which is the default), it seems to the user like the controls simply don't work. (The game does show a prompt saying "GameCube Controller", but you know how it is with Android users and figuring out these kinds of things on their own.) This PR makes it so that an input override function is only registered for the GameCube controller when the touch overlay is set to use the GameCube controller. If a follow-up PR then makes it so the controllers don't get treated as constantly connected on Android, this problem should be solved... I hope.

You do have a point about TAS input. But if it's fine, I'd like to leave it this way for now and then look into changing it when/if we implement that for TAS input. (Perhaps unregistering the override function would work for TAS input too? I haven't fully looked into it. Or we could send the "connected" signal to the override function as if it was an input...)

Source/Core/InputCommon/ControllerEmu/ControlGroup/Cursor.cpp
JosJuice: I've applied this simplification to AnalogStick and Cursor, but for the other ones it wouldn't apply cleanly due to GetState doing extra processing after the point where the override function would be called.

JosJuice:
Fixes https://bugs.dolphin-emu.org/issues/12894.

AdmiralCurtiss:
Can we get a review from @jordan-woyak on this? I'd like to see this merged so we can inch closer to fixing that ridiculous Android input system.

I suspect this can also be leveraged to improve (non-GC) controller behavior on netplay...

saminamo (approved):
Looks good to me

jordan-woyak (approved):
Code looks good. Untested, but LGTM.

JMC47:
whelp, I see approval, I see green button.

Source/Core/DolphinQt/TAS/TASInputWindow.cpp
```diff
@@ -151,7 +216,6 @@ QSpinBox* TASInputWindow::CreateSliderValuePair(QBoxLayout* layout, int default_
   slider->setRange(0, max);
   slider->setValue(default_);
   slider->setFocusPolicy(Qt::ClickFocus);
-  slider->setInvertedAppearance(invert);
```
Pokechu22, line 154: Why was this (and the `invert` parameter in general) removed? The IR widget used them. (However, it seems like things still behave correctly - the sliders for the IR widget still match the position and aren't reversed, at least on my machine, so I'm not sure what it does; [the docs for `setInvertedAppearance`](https://doc.qt.io/qt-6/qabstractslider.html#invertedAppearance-prop) don't really clarify either.)

Source/Core/DolphinQt/TAS/IRWidget.h
```diff
@@ -34,5 +34,7 @@ public slots:
 };
 
 // Should be part of class but fails to compile on mac os
+static const u16 ir_min_x = 0;
+static const u16 ir_min_y = 0;
 static const u16 ir_max_x = 1023;
 static const u16 ir_max_y = 767;
```
Pokechu22, line 40: Note that these can be in the class (at least if they're changed to `constexpr`). This is a change I make in the balance board PR (#8268).

Source/Core/DolphinQt/TAS/TASInputWindow.cpp
JosJuice: Previously, the IR sliders were more or less directly mapped to where on the IR camera the sensor bar lights are visible. When you move a Wii Remote right, the lights appear to move to the left from the perspective of the camera – which means that when you move the slider right, we should invert the input so that the light points will actually move to the left. With the new input override API, the sliders are mapped to the Point action, and WiimoteEmu takes care of translating from "location you're pointing at" to "light locations".

PR #9626, opened 2021-04-06
Freelook: add position offset binding

iwubcode:
This adds the ability to offset the FreeLook camera directly using x,y,z coordinates.  These values can even be bound to an input expression:

all-aboard

~~In theory this allows for headtracking using the positional element in OpenTrack or VR devices (untested at the moment)~~ (working, see post below!)

--

UI has changed, move is now put in a "Position" tab:

image

here is what the tab looks like:

image

### DIFF
--- a/Source/Core/Core/FreeLookManager.h
+++ b/Source/Core/Core/FreeLookManager.h
@@ -9,6 +9,7 @@
 
 #include "Common/CommonTypes.h"
 #include "InputCommon/ControllerEmu/ControllerEmu.h"
+#include "InputCommon/ControllerEmu/Setting/NumericSetting.h"
 
 class InputConfig;
 
@@ -26,6 +27,7 @@ enum class FreeLookGroup
   FieldOfView,
   Other,
   Rotation,
+  PositionOffset,
 };
 
 namespace FreeLook
@@ -54,6 +56,19 @@ public:
 
 private:
   ControllerEmu::Buttons* m_move_buttons;
+
+  // The following position offsets are doubles
+  // but used as floats in the camera system
+  // This is done because the settings and qt
+  // controls are in double format
+  ControllerEmu::SettingValue<double> m_pos_x;
+  double m_pos_last_x;
+  ControllerEmu::SettingValue<double> m_pos_y;
+  double m_pos_last_y;
+  ControllerEmu::SettingValue<double> m_pos_z;
+  double m_pos_last_z;
+  ControllerEmu::ControlGroup* m_position_offset_group;
+
   ControllerEmu::Buttons* m_speed_buttons;
   ControllerEmu::Buttons* m_fov_buttons;
   ControllerEmu::Buttons* m_other_buttons;
--- a/Source/Core/Core/FreeLookManager.cpp
+++ b/Source/Core/Core/FreeLookManager.cpp
@@ -87,6 +87,19 @@ FreeLookController::FreeLookController(const unsigned int index) : m_index(index
   m_move_buttons->AddInput(ControllerEmu::Translate, _trans("Forward"));
   m_move_buttons->AddInput(ControllerEmu::Translate, _trans("Backward"));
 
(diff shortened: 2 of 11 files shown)


iwubcode:
For "First Person" this allows you to modify the positional offset.

For "Orbital" this allows you to modify the target point!

"Six Axis" is still to be tested, marked as draft for now.

@MichaelJW - thought you might be interested

iwubcode:
I updated Six Axis to split it into a separate rotation and position component (mimic'ing the other controllers).  It should still behave the same as master though.  This allows the position to be set without effecting the rotation.

I fixed orbital, it wasn't updating its focal point before but now it is.

This is now open for review

Source/Core/Core/FreeLookManager.h
```diff
@@ -54,6 +56,17 @@ class FreeLookController final : public ControllerEmu::EmulatedController
...
   ControllerEmu::Buttons* m_move_buttons;
+
+  ControllerEmu::SettingValue<double> m_pos_x;
+  double m_pos_last_x;
+
+  ControllerEmu::SettingValue<double> m_pos_y;
+  double m_pos_last_y;
+
+  ControllerEmu::SettingValue<double> m_pos_z;
+  double m_pos_last_z;
```
Dentomologist, line 67: I believe these are cast to float whenever they're used. If there's a reason they need to be doubles anyway I would add a comment explaining why.

iwubcode: Numeric settings and also qt spin boxes all operate in terms of doubles but our Free Look code is in terms of floats.  I'll add a comment

iwubcode:
Finally had a chance to test this with open track:

all-aboard-2

It's completely ready for users to use!

Rumi-Larry:
I do wonder, why there isn't a button to reset the field of view?

MayImilae:
> I do wonder, why there isn't a button to reset the field of view?

You use Reset View to reset Field of View.

Speaking of which, I assume the reset doesn't affect the offset, I need to check that when I review this.

Rumi-Larry:
> > I do wonder, why there isn't a button to reset the field of view?
> 
> You use Reset View to reset Field of View.
> 
> Speaking of which, I assume the reset doesn't affect the offset, I need to check that when I review this.
If it is possible then the UI isn't clear about that. 
I get that, but shouldn't one be able to reset the position of the camera, swparate from the field of view?

iwubcode:
> I get that, but shouldn't one be able to reset the position of the camera, separate from the field of view? If it is possible then the UI isn't clear about that.

It may make sense to split "Reset View" into 3 separate entities:

* Reset Fov
* Reset position
* Reset rotation

For backwards compatibility we could assign those all to the same key.

I'll let @MayImilae comment on that.

However, if applicable, that would be done in a separate review.

> Speaking of which, I assume the reset doesn't affect the offset, I need to check that when I review this.

@MayImilae - thanks for bringing that up.  That's not how it works at the moment but was something I was thinking about. I wasn't sure how to make this work but I just thought of an idea.  I'll implement that tonight so you can review.  Thanks!

MayImilae:
I think that standard FOV, Position, and Rotation should not have separate reset commands. Even if there was still a Reset All mapping by default (I would insist on it) I think it's a LOT of needless hotkeys to add. There are no circumstances I can think of where a user would want to reset position without also reseting rotation. FOV is a bit more complex as someone may want to adjust the FOV of free look and leave it to widen the FOV of a game, but thinking it through, that user would not be changing the position or rotation at all. So they also wouldn't benefit from separate reset hotkeys.

As for a reset offset hotkey, I don't think it is necessary. This offset is definitely a set and leave kind of thing, and since it is all controlled through the UI anyway, a user will know to return to the UI to reset them.

iwubcode:
This is superceded by #9671 so closing this out for now.  However, this does have a couple changes I'd like to get in as a separate PRs (SixAxis rework + moving movement to a "Position" tab).

PR #9627, opened 2021-04-06
VideoInterface.cpp: simplify expression

QuestionableEthics:
Anything that isn't 0 will be true, so the expression can be simplified.


JosJuice:
I don't think this makes the code more readable. WPL isn't a register that enables or disables something, it's a register that we're checking for 0 to avoid a division by 0.

delroth:
+1 to @JosJuice's comment.

PR #9628, opened 2021-04-06
WiiUtils: Remove reference qualifier

Dentomologist:
Remove & to avoid reference to a temporary subobject.  Fixes error C2131 on upcoming version of msbuild.


JosJuice:
Linking the issue for reference: https://bugs.dolphin-emu.org/issues/12468

BhaaLseN (approved):

PR #9629, opened 2021-04-06
Exclude initial memory/register setup from timing for first FIFO frame

Pokechu22:
I do this by waiting a frame so that the timing is reset.

Currently a draft to see what this does to FifoCI.  I expect this to make e.g. the software renderer produce less weird first frames, but I don't know what will happen for sure (it might introduce a garbage frame to the output).  (The hardware renderers are also affected by this initial timing &mdash; see https://github.com/dolphin-emu/dolphin/pull/9576#discussion_r602822069 &mdash; but the software renderer is much more affected).


Pokechu22:
Closing this because it didn't fix the software renderer issues I hoped it would.  If I figure out what's causing those, I'll reopen it.

PR #9630, opened 2021-04-08
Gameini: Disable ICache for Indiana Jones and the Staff of Kings

PatrickFerry:
The option to disable ICache is recent but it allows Indiana Jones and the Staff of Kings to proceed where it would otherwise crash.

This game got past the first loading screen and in-game in 3.0-195(1387da790062f67ea79db470999e37d0ed43c57f), however at certain parts the game will crash. Recent logs will state that an ICache read returned stale data.

### DIFF
--- a/Data/Sys/GameSettings/RJ8.ini
+++ b/Data/Sys/GameSettings/RJ8.ini
@@ -1,0 +1,21 @@
+# RJ8P64, RJ8E64 - Indiana Jones and the Staff of Kings
+
+[Core]
+# Values set here will override the main Dolphin settings.
+# The JIT cache causes problems with emulated icache invalidation in this game resulting in areas failing to load
+DisableICache = True
+
+[OnLoad]
+# Add memory patches to be loaded once on boot here.
+
+[OnFrame]
+# Add memory patches to be applied every frame here.
+
+[ActionReplay]
+# Add action replay cheats here.
+
+[Video_Settings]
+SafeTextureCacheColorSamples = 2048
+
+[Video_Hacks]
+EFBEmulateFormatChanges = True


MayImilae:
> [Video_Settings]
> SafeTextureCacheColorSamples = 2048

> [Video_Hacks]
> EFBEmulateFormatChanges = True

What is your reasoning for these two ancillary changes? You only describe the one.

PatrickFerry:
Some scenes in game are either fuzzy or have a different tint(blue) when "Ignore Format Changes" is set.

The intro screen(LucasArts) is choppy even with Texture Cache at "Medium" but a higher value makes it look correct.

iwubcode:
@PatrickFerry - please add the game name to the title for others reference

Pokechu22:
Have you tested the game with the (uncached) interpreter core on master?  ICache is implemented in a way that should be accurate when using the uncached interpreter so it would be good to make sure that a JIT-ICache interaction is actually at fault.

Miksel12:
The ini file is RJ8E64 only while it says RJ8P64 and RJ8E64 in the file itself. So you should rename the file to RJ8 or drop RJ8P64.

JosJuice:
Yes, the file should be renamed RJ8.ini.

PatrickFerry:
> Have you tested the game with the (uncached) interpreter core on master? ICache is implemented in a way that should be accurate when using the uncached interpreter so it would be good to make sure that a JIT-ICache interaction is actually at fault.

It doesn't crash if uncached interpreter is used.

I also renamed the file to work for any region.

Pokechu22:
@dolphin-emu-bot rebuild

Pokechu22 (approved):
I don't own this game, so I haven't tested myself, but this seems like a reasonable change to me.  

Since it's a Wii game, `HLE_BS2 = True` isn't required (as only the GameCube Main Menu has issues with ICache being disabled), so it not being included is correct.

JMC47 (approved):
The fact pure interpreter works signals that this is the best we can do to fix the issue for now.  All changes to the INI make sense.

PR #9634, opened 2021-04-11
Build: Fix -fno-strict-aliasing on Xcode builds

skylersaleh:
Xcode is configured by default to turn on the strict aliasing optimization through a separate project setting. This would cause the compiler to be called with both -fno-strict-aliasing(from the CMakeLists.txt) and -fstrict-aliasing(appended to the compiler arguments by Xcode) flags. This change turns off this default project setting, causing only the -fno-strict-aliasing flag to be provided to the compiler.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -291,6 +291,9 @@ if(CMAKE_SYSTEM_NAME MATCHES "Darwin")
     list(APPEND CMAKE_PREFIX_PATH "/usr")
   endif()
 
+  # Prevents Xcode from overriding the -fno-strict-aliasing flag
+  set(CMAKE_XCODE_ATTRIBUTE_GCC_STRICT_ALIASING NO)
+
   # Specify target CPUs.
   check_and_add_flag(HAVE_MSSSE3 -mssse3)
   check_and_add_flag(HAVE_ARCH_CORE2 -march=core2)


JMC47:
@dolphin-emu-bot rebuild

StLouisCPhT:
It doesn't solve the issue with pugixml & FIFO gfx crashes. It also breaks the speech bubbles in Paper Mario Thousand Year Door causing them to be blank.

skylersaleh:
> It doesn't solve the issue with pugixml & FIFO gfx crashes. It also breaks the speech bubbles in Paper Mario Thousand Year Door causing them to be blank.

It would be very unlikely for this change to cause functional bugs. All this change does is turn off the strict aliasing optimization in Xcode to match the settings used for the make file based builds for Mac OS and Linux. Is it possible the Paper Mario bug was caused by something else?

StLouisCPhT:
> > It doesn't solve the issue with pugixml & FIFO gfx crashes. It also breaks the speech bubbles in Paper Mario Thousand Year Door causing them to be blank.
> 
> It would be very unlikely for this change to cause functional bugs. All this change does is turn off the strict aliasing optimization in Xcode to match the settings used for the make file based builds for Mac OS and Linux. Is it possible the Paper Mario bug was caused by something else?

it could be another coincidental bug like pugixml. I do know that it only occurs when building with this specific code change added.

StLouisCPhT:
Seems okay to me.

PR #9635, opened 2021-04-11
Vulkan: Work around AMD exclusive fullscreen bug (21.3+)

stenzek:
AMD's binary driver as of 21.3 seems to return exclusive fullscreen lost even when it was never requested, so long as the caller requested it to be application controlled. Handle this ignoring the lost result and just continuing as normal if we never acquired it.

References:
 - https://community.amd.com/t5/graphics/vulkan-spec-violation-vk-error-full-screen-exclusive-mode-lost/td-p/462283
 - https://github.com/mpv-player/mpv/issues/8008#issuecomment-813747217

### DIFF
--- a/Source/Core/VideoBackends/Vulkan/VKRenderer.cpp
+++ b/Source/Core/VideoBackends/Vulkan/VKRenderer.cpp
@@ -288,9 +288,24 @@ void Renderer::BindBackbuffer(const ClearColor& clear_color)
     m_swap_chain->SetNextFullscreenState(m_swap_chain->GetCurrentFullscreenState());
   }
 
-  VkResult res = g_command_buffer_mgr->CheckLastPresentFail() ?
-                     g_command_buffer_mgr->GetLastPresentResult() :
-                     m_swap_chain->AcquireNextImage();
+  const bool present_fail = g_command_buffer_mgr->CheckLastPresentFail();
+  VkResult res = present_fail ? g_command_buffer_mgr->GetLastPresentResult() :
+                                m_swap_chain->AcquireNextImage();
+
+  if (res == VK_ERROR_FULL_SCREEN_EXCLUSIVE_MODE_LOST_EXT &&
+      !m_swap_chain->GetCurrentFullscreenState())
+  {
+    // AMD's binary driver as of 21.3 seems to return exclusive fullscreen lost even when it was
+    // never requested, so long as the caller requested it to be application controlled. Handle
+    // this ignoring the lost result and just continuing as normal if we never acquired it.
+    res = VK_SUCCESS;
+    if (present_fail)
+    {
+      // We still need to acquire an image.
+      res = m_swap_chain->AcquireNextImage();
+    }
+  }
+
   if (res != VK_SUCCESS)
   {
     // Execute cmdbuffer before resizing, as the last frame could still be presenting.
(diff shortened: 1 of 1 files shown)


Source/Core/VideoBackends/Vulkan/VKRenderer.cpp
```diff
@@ -300,12 +315,12 @@ void Renderer::BindBackbuffer(const ClearColor& clear_color)
     if (res == VK_ERROR_FULL_SCREEN_EXCLUSIVE_MODE_LOST_EXT)
     {
       // The present keeps returning exclusive mode lost unless we re-create the swap chain.
-      INFO_LOG_FMT(VIDEO, "Lost exclusive fullscreen.");
+      INFO_LOG(VIDEO, "Lost exclusive fullscreen.");
```
BhaaLseN, line 318: Please don't change those back. Mid- to long-term we want to get rid of `xxx_LOG` and rename `xxx_LOG_FMT` so everything uses `fmt`.

Source/Core/VideoBackends/Vulkan/VKRenderer.cpp
```diff
@@ -300,12 +315,12 @@ void Renderer::BindBackbuffer(const ClearColor& clear_color)
...
       m_swap_chain->RecreateSwapChain();
     }
     else if (res == VK_SUBOPTIMAL_KHR || res == VK_ERROR_OUT_OF_DATE_KHR)
     {
-      INFO_LOG_FMT(VIDEO, "Resizing swap chain due to suboptimal/out-of-date");
+      INFO_LOG(VIDEO, "Resizing swap chain due to suboptimal/out-of-date");
```
BhaaLseN, line 323: Same here.

Source/Core/VideoBackends/Vulkan/VKRenderer.cpp
```diff
@@ -315,9 +330,9 @@ void Renderer::BindBackbuffer(const ClearColor& clear_color)
     }
 
     res = m_swap_chain->AcquireNextImage();
+    if (res != VK_SUCCESS)
```
BhaaLseN, line 333: Did you move this inside the `if` intentionally?

Source/Core/VideoBackends/Vulkan/VKRenderer.cpp
stenzek: Ah, sure. I haven't been keeping up to date, and assumed "_FMT" was for a "formatted variant".

Source/Core/VideoBackends/Vulkan/VKRenderer.cpp
stenzek: Yep. The outer block already checks for a failed result, so it would be impossible for it to fail twice outside of it.

Source/Core/VideoBackends/Vulkan/VKRenderer.cpp
Pokechu22: The `INFO_LOG_FMT` version means the `fmt` library is used, but it's not like there's a performance benefit to not using it; the `INFO_LOG` version uses C-style format strings (which are, IIRC, slower than fmt).

RinMaru:
how did this issue pop up just in 5.0-14002 but nothing before it?

JMC47:
People were reporting it happening in older builds than that.  It's a driver bug and nothing caused by anything dolphin changed.

PR #9636, opened 2021-04-11
DriverDetails: Fix broken vector bitwise AND on Mali drivers

sspacelynx:
Mali GPUs have been plagued by this issue since the dawn of time of Dolphin on Android (though I'm not sure if other OSes are also affected by this). This makes Super Mario Galaxy 1 and 2 completely unplayable on those GPUs, and I'm not aware of any othe games that are affected by this.

Here's the before and after:
comparison smg1

### DIFF
--- a/Source/Core/VideoCommon/DriverDetails.h
+++ b/Source/Core/VideoCommon/DriverDetails.h
@@ -288,6 +288,16 @@ enum Bug
   // Mali Vulkan driver, causing high CPU usage in the __pi___inval_cache_range kernel
   // function. This flag causes readback buffers to select the coherent type.
   BUG_SLOW_CACHED_READBACK_MEMORY,
+
+  // BUG: Apparently ARM Mali GLSL compiler managed to break bitwise AND operations between
+  // two integers vectors, when one of them is non-constant (though the exact cases of when
+  // this occurs are still unclear). The resulting vector from the operation will be the
+  // constant vector.
+  // Easy enough to fix, just do the bitwise AND operation on the vector components first and
+  // then construct the final vector.
+  // Started version: -1
+  // Ended version: -1
+  BUG_BROKEN_VECTOR_BITWISE_AND,
 };
 
 // Initializes our internal vendor, device family, and driver version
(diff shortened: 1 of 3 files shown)


Source/Core/VideoCommon/DriverDetails.h
```diff
@@ -288,6 +288,16 @@ enum Bug
   // Mali Vulkan driver, causing high CPU usage in the __pi___inval_cache_range kernel
   // function. This flag causes readback buffers to select the coherent type.
   BUG_SLOW_CACHED_READBACK_MEMORY,
+
+  // BUG: Apparently ARM Mali GLSL compiler managed to break bitwise AND operations between
+  // two integers vectors, when one of them is non-constant (though the exact cases of when
+  // this occurs are still unclear to me). The resulting vector from the operation will be
```
lioncash, line 294: This description should probably avoid the self-reference here, particularly since it's not clear out of context who "me" is.

sspacelynx: sure

sspacelynx:
Fixed lint errors

BhaaLseN:
I'm afraid you didn't :)
Curlies go on the next line, specifically the one on the `if`.

sspacelynx:
Damm it

JosJuice:
This line is also using a bitwise AND: https://github.com/dolphin-emu/dolphin/blob/aba9cae5ab5749f1a86cdf56f3e6c7b6906d8d8c/Source/Core/VideoCommon/PixelShaderGen.cpp#L926

Could you check if rendering breaks on Mali when disabling Force 24-Bit Color? If not, then LGTM as it is.

sspacelynx:
Sure thing, will take a look.

sspacelynx:
I've yet to investigate the exact circumstances in which this happen, as I'm planning to report this in depth. For now the issue seems to be related to certain ivec4 only, as ANDing ivec3 and lower components count vectors results in the correct behavior.
Disabling Force 24-Bit Color has no effect on rendering.

JosJuice (approved):

PR #9638, opened 2021-04-12
WiiUtils: Add helper functions to get emulated/real Bluetooth device

leoetlino:
This adds a function to get the emulated or real Bluetooth device for
an active emulation instance. This lets us deduplicate all the
`ios->GetDeviceByName("/dev/usb/oh1/57e/305")` calls that are currently
scattered in the codebase and ensures Bluetooth passthrough is being
handled correctly.

This also fixes the broken check in WiimoteCommon::UpdateSource.
There was a confusion between "emulated Bluetooth" (as opposed to
"real Bluetooth" aka Bluetooth passthrough) and "emulated Wiimote".

### DIFF
--- a/Source/Core/Core/HW/Wiimote.cpp
+++ b/Source/Core/Core/HW/Wiimote.cpp
@@ -16,6 +16,7 @@
 #include "Core/IOS/USB/Bluetooth/WiimoteDevice.h"
 #include "Core/Movie.h"
 #include "Core/NetPlayClient.h"
+#include "Core/WiiUtils.h"
 
 #include "InputCommon/ControllerEmu/ControlGroup/ControlGroup.h"
 #include "InputCommon/InputConfig.h"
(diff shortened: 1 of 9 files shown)


Source/Core/Core/Movie.cpp
```diff
@@ -469,16 +470,13 @@ void ChangeWiiPads(bool instantly)
...
-                            ios->GetDeviceByName("/dev/usb/oh1/57e/305")) :
-                        nullptr;
+  const auto bt = WiiUtils::GetBluetoothEmuDevice();
   for (int i = 0; i < MAX_WIIMOTES; ++i)
   {
     const bool is_using_wiimote = IsUsingWiimote(i);
 
     WiimoteCommon::SetSource(i, is_using_wiimote ? WiimoteSource::Emulated : WiimoteSource::None);
-    if (!SConfig::GetInstance().m_bt_passthrough_enabled && bt)
+    if (bt)
```
BhaaLseN, line 479: Do we want to change this into `bt != nullptr` for consistency while we're at it?

Source/Core/Core/HW/Wiimote.cpp
```diff
@@ -49,15 +50,7 @@ void SetSource(unsigned int index, WiimoteSource source)
...
-  if (!ios)
-    return;
-
-  if (s_wiimote_sources[index] != WiimoteSource::Emulated)
-    return;
-
-  const auto bluetooth = std::static_pointer_cast<IOS::HLE::BluetoothEmuDevice>(
-      ios->GetDeviceByName("/dev/usb/oh1/57e/305"));
+  const auto bluetooth = WiiUtils::GetBluetoothEmuDevice();
   if (!bluetooth)
```
BhaaLseN, line 54: Same here, `if (bluetooth == nullptr) return`

Source/Core/Core/Movie.cpp
leoetlino: Sure, why not. Also changed MenuBar for consistency

BhaaLseN (approved):

JMC47 (approved):
Verified to fix the issue.

iwubcode (approved):

PR #9639, opened 2021-04-12
Update pngpread.c to address vulnerability

gstrick2:
This is a patch to the vulnerability addressed and fixed here: https://github.com/glennrp/libpng/commit/347538efbdc21b8df684ebd92d37400b3ce85d55


Pokechu22:
Thanks for the PR.

Dolphin uses libpng 1.6.37, which [came out after that change](https://github.com/glennrp/libpng/commits/v1.6.37/pngpread.c), and it seems like that change isn't present in the current version of pngpread.c.  In particular look at glennrp/libpng@2dca15686fadb1b8951cb29b02bad4cae73448da, which moved the length check into `png_check_chunk_length` in pngrutil.c.  So I don't think this change is actually needed.

PR #9641, opened 2021-04-14
Remove spurious memory allocations in TextureCacheBase::SerializeTexture and DeserializeTexture

lynlevenick:
I'm doing a project with Dolphin which wants to save state incredibly frequently (~every frame). Reallocating `texture_data` in `TextureCacheBase::SerializeTexture` takes up a non-negligible amount of the serialization time in my environment. For my own project I might be able to implement a check to see if textures have changed in the last frame and avoid re-serializing them, but this is general enough to save everybody some fractions of a second when they save state - maybe TASers would appreciate it?

This PR fixes the issue by adding a method to `PointerWrap` which sets aside some space for the caller to read/write/verify, then using it to set aside one large chunk of space all at once in the texture cache. This should not change the save-state format.

Additionally, the new method is used when deserializing textures as well to avoid a single memory allocation/copy/deallocation, but that's of much less concern and could be removed pretty easily.

I'm not a huge fan of the method I went with here - all of the paths that hit `DoExternal` need to handle the `mode` of the `PointerWrap`, but at present there's no enforcement of that. I've personally verified that only read and write operations take place to the pointers, but that's no guard against future changes so it's pretty reasonable to say `DoExternal` is going too far/breaking abstraction. If so, let me know: I may be able to change this to refactor `SerializeTexture` to do one allocation instead of worst-case layers * levels copies/reallocations.

If save state performance is important, I've got a few other lower-impact changes which I could also try to contribute if desired.

### DIFF
--- a/Source/Core/Common/ChunkFile.h
+++ b/Source/Core/Common/ChunkFile.h
@@ -200,6 +200,16 @@ public:
     DoArray(arr, static_cast<u32>(N));
   }
 
+  // The caller is required to inspect the mode of this PointerWrap
+  // and deal with the pointer returned from this function themself.
+  [[nodiscard]] u8* DoExternal(u32& count)
+  {
+    Do(count);
+    u8* current = *ptr;
+    *ptr += count;
+    return current;
+  }
+
   void Do(Common::Flag& flag)
   {
     bool s = flag.IsSet();
--- a/Source/Core/VideoCommon/TextureCacheBase.cpp
+++ b/Source/Core/VideoCommon/TextureCacheBase.cpp
@@ -438,28 +438,48 @@ void TextureCacheBase::SerializeTexture(AbstractTexture* tex, const TextureConfi
   const bool skip_readback = p.GetMode() == PointerWrap::MODE_MEASURE;
   p.DoPOD(config);
 
-  std::vector<u8> texture_data;
   if (skip_readback || CheckReadbackTexture(config.width, config.height, config.format))
   {
-    // Save out each layer of the texture to the staging texture, and then
-    // append it onto the end of the vector. This gives us all the sub-images
-    // in one single buffer which can be written out to the save state.
+    // First, measure the amount of memory needed.
+    u32 total_size = 0;
     for (u32 layer = 0; layer < config.layers; layer++)
     {
       for (u32 level = 0; level < config.levels; level++)
       {
         u32 level_width = std::max(config.width >> level, 1u);
         u32 level_height = std::max(config.height >> level, 1u);
-        auto rect = tex->GetConfig().GetMipRect(level);
-        if (!skip_readback)
+
+        u32 stride = AbstractTexture::CalculateStrideForFormat(config.format, level_width);
+        u32 size = stride * level_height;
+
+        total_size += size;
+      }
+    }
+
+    // Set aside total_size bytes of space for the textures.
+    // When measuring, this will be set aside and not written to,
+    // but when writing we'll use this pointer directly to avoid
+    // needing to allocate/free an extra buffer.
+    u8* texture_data = p.DoExternal(total_size);
+
+    if (!skip_readback)
+    {
+      // Save out each layer of the texture to the pointer.
+      for (u32 layer = 0; layer < config.layers; layer++)
+      {
+        for (u32 level = 0; level < config.levels; level++)
+        {
+          u32 level_width = std::max(config.width >> level, 1u);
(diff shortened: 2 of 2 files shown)


JMC47:
Savestate performance is always appreciated.

JosJuice:
Yes. It's the kind of change that won't be noticed by most users but would be greatly appreciated by a small subset of users. If you have more optimizations you can make, I'm interested in seeing those as well.

All three commits should be squashed into one, not just the last two.

I agree that adding this kind of `DoExternal` function is a bit "risky" in terms of code design, but I don't really have a better suggestion for how to do it, and I think the performance benefit outweighs the downsides.

Source/Core/Common/ChunkFile.h
```diff
@@ -200,6 +200,14 @@ class PointerWrap
     DoArray(arr, static_cast<u32>(N));
   }
 
+  void* DoExternal(u32& count)
```
JosJuice, line 203:
```suggestion
  [[nodiscard]] void* DoExternal(u32& count)
```

Source/Core/VideoCommon/TextureCacheBase.cpp
```diff
@@ -438,48 +438,64 @@ void TextureCacheBase::SerializeTexture(AbstractTexture* tex, const TextureConfi
...
+      }
+    }
+
+    u8* texture_data = static_cast<u8*>(p.DoExternal(total_size));
+
+    if (!skip_readback)
+    {
+      // Save out each layer of the texture to the staging texture, and then
+      // append it onto the end of the region. This gives us all the sub-images
+      // in one single buffer which can be written out to the save state.
```
JosJuice, line 465: This comment is a bit out of date now. It would be nice to replace the mention of "one single buffer" (this buffer doesn't actually exist anymore) with something saying that the reason we write directly into the `PointerWrap` is because we want to avoid allocating memory for performance reasons.

Source/Core/Common/ChunkFile.h
lioncash: This should also return `u8*`, rather than hiding the type

lynlevenick:
Addressed comments and squashed the commits together.

I'll open up some other changes as separate PRs once I've pulled them out of my branch.

Pokechu22:
@dolphin-emu-bot rebuild

Pokechu22:
@dolphin-emu-bot rebuild

Source/Core/Common/ChunkFile.h
```diff
@@ -200,6 +200,16 @@ class PointerWrap
     DoArray(arr, static_cast<u32>(N));
   }
 
+  // The caller is required to inspect the mode of this PointerWrap
+  // and deal with the pointer returned from this function themself.
```
Rumi-Larry, line 204:
```suggestion
  // The caller is required to inspect the mode of this PointerWrap
  // and deal with the pointer returned from this function itself.
```

lynlevenick: I suppose I'm humanizing the caller here as a "person who is writing the function calling this code". Is there any style guide for comments in general? I'm not really opposed to this but wouldn't be able to get around to making this change and squashing it until possibly tomorrow.

Rumi-Larry: That would only make sense if "The Caller" was a proper noun, in which case it must be capitalized like so and between parenthesis to make it explicit, but since it's not, then it should be "itself"

Pokechu22: Both ways seem awkward to me.  But this:

> The caller is required to inspect the mode of this PointerWrap and deal with the returned pointer by itself.

sounds a lot better than this:

> The caller is required to inspect the mode of this PointerWrap and deal with the returned pointer by themself.

Rumi-Larry: > 
> 
> ...what? I don't understand why assuming the caller is a human implies that it needs to be a proper noun. If you replace this with something like 'The doctor is required...' the word 'doctor' doesn't need to be capitalized either.
> 
> If you ask me the comment is fine as-is anyway, but I'm really not following your train of thought here.

Because "themself" is intended for people, and people have proper nouns.

JosJuice: People can be referred to using nouns that are not proper nouns. And capitalizing a noun doesn't necessarily make it a proper noun either.

JosJuice (approved):

PR #9642, opened 2021-04-15
Fix crash when stopping a game that does not use the BBA when XLink Kai BBA is selected in configuration

CrunchBite:
This PR fixes a bug reported here https://bugs.dolphin-emu.org/issues/12273
Crash is caused by a thread being terminated that never started since the BBA wasn't activated.

### DIFF
--- a/Source/Core/Core/HW/EXI/BBA/XLINK_KAI_BBA.cpp
+++ b/Source/Core/Core/HW/EXI/BBA/XLINK_KAI_BBA.cpp
@@ -58,6 +58,10 @@ bool CEXIETHERNET::XLinkNetworkInterface::Activate()
 
 void CEXIETHERNET::XLinkNetworkInterface::Deactivate()
 {
+  // Is the BBA Active? If not skip shutdown
+  if (!IsActivated())
+    return;
+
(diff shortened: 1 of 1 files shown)


Pokechu22:
@dolphin-emu-bot rebuild

Pokechu22:
@dolphin-emu-bot rebuild

Source/Core/Core/HW/EXI/BBA/XLINK_KAI_BBA.cpp
```diff
@@ -58,6 +58,12 @@ bool CEXIETHERNET::XLinkNetworkInterface::Activate()
 
 void CEXIETHERNET::XLinkNetworkInterface::Deactivate()
 {
+  // Is the BBA Active? If not skip shutdown
+  if (!this->IsActivated())
+  {
+    return;
+  }
```
JosJuice, line 65:
```suggestion
  if (!IsActivated())
    return;
```

JosJuice (commented):
Please squash the two commits into one.

JMC47 (approved):
Appears to fix the issue.  I can no longer crash with the XLink Kai backend by closing the game shortly after boot.

PR #7214, opened 2018-07-05
Fifo: Run/sync with the GPU on command processor register access

stenzek:
In Dolphin, we don't emulate the CPU and GPU in lock-step, for performance reasons. The CPU drives events, with the GPU emulation being one of these events which is fired periodically. This is the behavior in single core mode. Dual-core mode is completely non-deterministic here, and runs whenever Fifo::RunGpu() is called. Thus, I am mainly concerned about single-core mode.

Due to the GPU only being run every 1000 cycles or so, to a game, if it polled the command processor registers (for example, the read pointer, or distance), the GPU would appear to be stalled. Testing would suggest this is what causes FIFO resets/unknown opcodes in games such as F-Zero GX, and Rogue Squadron 3.

To work around this, each time these registers are accessed, we run the GPU for its time quantum. This way, to a game, the GPU is making progress (as it would on the console). While this isn't necessarily accurate to the hardware in terms of cycles executed, we don't *really* emulate GPU timings anyway, so executing a few extra GPU cycles doesn't really have any impact.

In dual core, it syncs with the GPU thread, and ensures that the GPU thread isn't too far behind. Again, this is not deterministic, but dual core isn't to begin with, and has numerous stability issues as a result.

Long-term, I'm planning on refactoring the FIFO in a way which resembles the current state of deterministic dual core - read from memory on the CPU thread, but process the commands on the GPU thread. This will lead to determinism for command processor behavior (e.g. FIFO breakpoints, hi/low watermark interrupts), except EFB copies to RAM, and BP token interrupts.

However, this is a much more significant task, with far larger chances of regressions, so I pulled this change out to fix RS3, in the meantime, at least. And perhaps new motivation to make full MMU faster ;).

PR #9643, opened 2021-04-17
GameINI: Fix transitions and crashes in Rogue Squadron 3

JMC47:
Needs to come after https://github.com/dolphin-emu/dolphin/pull/7214 or this doesn't really do anything.

Transitions are fixed by disabling "Store XFB Copies to Texture Only"

Forcing Single Core, in cooperation with https://github.com/dolphin-emu/dolphin/pull/7214, fixes random crashes on the intro dance, main menu, ship selection, pausing, at the end of stages, and in multiplayer.

The performance implications are minor because the game runs like crap on the beefiest machines anyway.  I saw maybe a 1 FPS difference disabling Store XFB Copies to Texture Only.  It's Rogue Squadron 3, if you're playing it in Dolphin, it's mostly out of masochism or curiosity more than to try to get a playable full speed experience.

SyncGPU is not enough to prevent the crashes.  While they are rarer, I still crashed on ship selection 1/3 times, which I do not consider playable.  I want people to be able to boot this up and sort of see it play before realizing the monstrosity that Factor 5 created.

### DIFF
--- a/Data/Sys/GameSettings/GLR.ini
+++ b/Data/Sys/GameSettings/GLR.ini
@@ -3,6 +3,7 @@
 [Core]
 # Values set here will override the main Dolphin settings.
 MMU = 1
+CPUThread = False
 
 [OnLoad]
 # Add memory patches to be loaded once on boot here.
@@ -20,4 +21,5 @@ MMU = 1
 [Video_Hacks]
 EFBToTextureEnable = False
 ImmediateXFBEnable = False
+XFBToTextureEnable = False
 DeferEFBCopies = False


StLouisCPhT (approved):
LGTM

JMC47:
Note that this can't be merged until after 7214, as it will make the game not work at all until that point.

StLouisCPhT:
You could also add EFBAccessEnable = False to the ini I suppose. It fixes the transparency issue seen in some of the cutscenes (most notably the Imperial transports flying through the Yavin system). Haven't noticed any issues during gameplay with it enabled.

JMC47:
Let me check.  If that's true, then I'll add it.  Last I checked the transparency issues didn't get fixed in RS3, only RS2.

JosJuice:
Should DLS.ini be updated as well?

JMC47:
@JosJuice it uses a different version of the engine afaik, it needs separate testing.

JMC47:
I don't think we should enable it.  It trades one glitchiness for another.

StLouisCPhT:
> I don't think we should enable it. It trades one glitchiness for another.

What's the other glitchiness?

JMC47:
The engine flares will show up through solid objects.

JMC47:
I'm not sure the demo needs single core, it may work ok in Dualcore with syncGPU.  I need to play it longer.

JMC47:
EFB Access can be added to this once https://github.com/dolphin-emu/dolphin/pull/9673 is merged.

JMC47:
So, I probably should close this for now and go through all the demos of RS3 and whatnot to try and figure out what exactly is needed in each one.  I noticed in some of the demos that apparently XFB works differently and maybe we can get away with less emulation.  Dualcore may also be ok in one of the demos I tried?  It's tough to tell but more testing is required.

JMC47:
I've been too lazy to figure out what each version needs and honestly I don't know if I'll have time to go through each of the demo discs to figure it out.

PR #9644, opened 2021-04-17
Jits: Fix interpreter fallback handling of discarded registers

JosJuice:
When the interpreter writes to a discarded register, its type must be changed so that it is no longer considered discarded.

Fixes a PR #9351 regression.

### DIFF
--- a/Source/Core/Core/PowerPC/Jit64/Jit.cpp
+++ b/Source/Core/Core/PowerPC/Jit64/Jit.cpp
@@ -419,15 +419,23 @@ void Jit64::FallBackToInterpreter(UGeckoInstruction inst)
 {
   gpr.Flush();
   fpr.Flush();
+
   if (js.op->opinfo->flags & FL_ENDBLOCK)
   {
     MOV(32, PPCSTATE(pc), Imm32(js.compilerPC));
     MOV(32, PPCSTATE(npc), Imm32(js.compilerPC + 4));
   }
+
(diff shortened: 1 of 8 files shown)


JMC47 (approved):
Fixes the issue

Source/Core/Core/PowerPC/Jit64/Jit.cpp
```diff
@@ -419,15 +419,24 @@ void Jit64::FallBackToInterpreter(UGeckoInstruction inst)
...
+
   Interpreter::Instruction instr = PPCTables::GetInterpreterOp(inst);
   ABI_PushRegistersAndAdjustStack({}, 0);
   ABI_CallFunctionC(instr, inst.hex);
   ABI_PopRegistersAndAdjustStack({}, 0);
+
+  // If the instruction wrote to any registers which were marked as discarded,
+  // we must mark them as no longer discarded
+  gpr.Reset(js.op->regsOut);
+  if (js.op->fregOut)
```
leoetlino, line 437: Shouldn't this be `js.op->fregOut >= 0`? -1 seems to be used to signal "no out float register". Also wonder if it's worth adding a "GetFRegOutBitSet".

JosJuice: Good catch. I've pushed a fix.

leoetlino (approved):

PR #9645, opened 2021-04-17
FifoPlayer: Copy data with memcpy instead of one byte at a time

leoetlino:
Copying with memcpy/std::copy is a lot faster than writing one byte at a time. And the behavior should be fully equivalent.

All FastWrite8 does is write a byte and increment the pointer:

https://github.com/dolphin-emu/dolphin/blob/5513d5f4f732fb1e436765ab87e7d60ba02b1ad6/Source/Core/Core/HW/GPFifo.cpp#L144-L148

Before: https://f.leolam.fr/out-rs3.svg

After: https://f.leolam.fr/out-rs3_v2.svg


JMC47:
This makes fifologs recorded from Rogue Squadron 3 run "full speed" and also makes a lot of other fifologs run at a much higher potential framerate.

iwubcode (approved):

Pokechu22 (approved):

PR #9646, opened 2021-04-17
SW: Fix alignedWidth in TextureEncoder

PatrickFerry:
This was causing issues in Software Renderer.

Attempting to solve https://bugs.dolphin-emu.org/issues/11487

### DIFF
--- a/Source/Core/VideoBackends/Software/TextureEncoder.cpp
+++ b/Source/Core/VideoBackends/Software/TextureEncoder.cpp
@@ -221,7 +221,7 @@ static void SetSpans(int sBlkSize, int tBlkSize, s32* tSpan, s32* sBlkSpan, s32*
 {
   // width is 1 less than the number of pixels of width
   u32 width = bpmem.copyTexSrcWH.x >> bpmem.triggerEFBCopy.half_scale;
-  u32 alignedWidth = Common::AlignUp(width, sBlkSize);
+  u32 alignedWidth = Common::AlignUp(width + 1, sBlkSize);
 
   u32 readStride = 3 << bpmem.triggerEFBCopy.half_scale;
 


shuffle2:
for which values of `width`, `sBlkSize` do you see this change producing different result from `Common::AlignUp`? That there is any difference suggests a kinda catastrophic bug in AlignUp

shuffle2:
It also seems like sBlkSize and tBlkSize are u16s that only store a power of 2 value, so the type of the args to `SetSpans` should be u16, and `ENCODE_LOOP_BLOCKS` should be changed to make `t`/`s` u32s?

shuffle2:
Oh, I missed that your replacement is `(value + align) & ~(align-1)` instead of the typical `(value + align - 1) & ~(align-1)`. I guess this is because `width is 1 less than the number of pixels of width`

PatrickFerry:
I would be happy for someone else to take this over since I'm not sure what the alignedWidth value should be in this case.

width = 388
sBlkSize = 4

(Broken)
`Common::AlignUp(width, sBlkSize) = 388`

(Seems to work)
`Common::AlignUp(width, sBlkSize-1) = 390`

(Seems to work and was how this written before https://github.com/dolphin-emu/dolphin/pull/9651/commits/7cf27b913ac8c3e3e03f965183b3f8843178f1b7)
`(width + sBlkSize) & (~(sBlkSize - 1)) = 392`

shuffle2:
Yea I think this PR is fine as-is, just took me a while to realize that https://github.com/dolphin-emu/dolphin/commit/d7dc854b505b5c3c604ed07b7c128726737c391c caused that comment (`width is 1 less than the number of pixels of width`) to go out of sync with what the code was actually doing.

PatrickFerry:
Oh, I could rewrite it like this

leoetlino (approved):
Whoops, my bad :sweat_smile:

PR #9648, opened 2021-04-17
Cache bounding box values between register reads in DX11/GL

stenzek:
This was present in the Vulkan and DX12 renderers, but not DX11/GL.

Quick test shows 60fps->90fps at 1x, and 55fps->80fps at 3x on an AMD GPU.

Bounding box is broken on OpenGL/AMD due to what seems to be a driver bug - the atomic only works on one of the fields, so I can't test it.

### DIFF
--- a/Source/Core/VideoBackends/D3D/D3DBoundingBox.cpp
+++ b/Source/Core/VideoBackends/D3D/D3DBoundingBox.cpp
@@ -2,18 +2,25 @@
 // Licensed under GPLv2+
 // Refer to the license.txt file included.
 
-#include "VideoBackends/D3D/D3DBoundingBox.h"
+#include <algorithm>
+#include <array>
+
 #include "Common/CommonTypes.h"
 #include "Common/MsgHandler.h"
+#include "VideoBackends/D3D/D3DBoundingBox.h"
 #include "VideoBackends/D3D/D3DState.h"
 #include "VideoBackends/D3DCommon/D3DCommon.h"
 #include "VideoCommon/VideoConfig.h"
 
 namespace DX11
 {
+static constexpr u32 NUM_BBOX_VALUES = 4;
 static ComPtr<ID3D11Buffer> s_bbox_buffer;
 static ComPtr<ID3D11Buffer> s_bbox_staging_buffer;
 static ComPtr<ID3D11UnorderedAccessView> s_bbox_uav;
+static std::array<s32, NUM_BBOX_VALUES> s_bbox_values;
+static std::array<bool, NUM_BBOX_VALUES> s_bbox_dirty;
+static bool s_bbox_valid = false;
 
 ID3D11UnorderedAccessView* BBox::GetUAV()
 {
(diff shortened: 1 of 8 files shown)


Source/Core/VideoBackends/D3D/D3DBoundingBox.cpp
```diff
@@ -22,42 +33,45 @@ ID3D11UnorderedAccessView* BBox::GetUAV()
...
+  CHECK(SUCCEEDED(hr), "Create BoundingBox Staging Buffer.");
+  D3DCommon::SetDebugObjectName(s_bbox_staging_buffer.Get(), "BoundingBox Staging Buffer");
+
+  // UAV is required to allow concurrent access.
+  D3D11_UNORDERED_ACCESS_VIEW_DESC UAVdesc = {};
+  UAVdesc.Format = DXGI_FORMAT_R32_SINT;
+  UAVdesc.ViewDimension = D3D11_UAV_DIMENSION_BUFFER;
+  UAVdesc.Buffer.FirstElement = 0;
+  UAVdesc.Buffer.Flags = 0;
+  UAVdesc.Buffer.NumElements = 4;
```
iwubcode, line 67: Does it make sense to use `NUM_BBOX_VALUES` here?

Source/Core/VideoBackends/OGL/OGLBoundingBox.cpp
```diff
@@ -36,43 +50,101 @@ void BoundingBox::Shutdown()
   glDeleteBuffers(1, &s_bbox_buffer_id);
 }
 
-void BoundingBox::Set(int index, int value)
+void BoundingBox::Flush()
```
iwubcode, line 53: I don't know enough about BoundingBox stuff to know if this is worth it but longer term it makes me wonder if we could split some of this out to a `AbstractBoundingBox` class that does some of this common logic and calls into the backend specific functionality.  This is near identical to the `D3D11::Flush()` .  I haven't looked at the other backends to know if they are similar but you mentioned there is caching there too..

Source/Core/VideoBackends/D3D/D3DBoundingBox.cpp
```diff
@@ -67,23 +81,73 @@ void BBox::Shutdown()
...
+
+    D3D::context->Unmap(s_bbox_staging_buffer.Get(), 0);
   }
-  D3D::context->Unmap(s_bbox_staging_buffer.Get(), 0);
-  return data;
+
+  s_bbox_valid = true;
+}
+
+void BBox::Set(int index, int value)
```
iwubcode, line 137: The `Set()` and `Get()` here don't do:

```
if (!g_ActiveConfig.backend_info.bSupportsBBox)
    return;
```

like OpenGL does.  Intentional?

Techjar:
This makes bounding box effects much faster in Paper Mario TTYD on D3D11 with an NVIDIA GPU. OpenGL got a marginal performance gain as well. However, despite what the PR description seems to imply, D3D12 is very slow.

Source/Core/VideoBackends/D3D/D3DBoundingBox.cpp
```diff
@@ -2,18 +2,29 @@
...
 #include "Common/CommonTypes.h"
 #include "Common/MsgHandler.h"
+#include "VideoBackends/D3D/D3DBoundingBox.h"
 #include "VideoBackends/D3D/D3DState.h"
 #include "VideoBackends/D3DCommon/D3DCommon.h"
 #include "VideoCommon/VideoConfig.h"
 
 namespace DX11
 {
+enum : u32
```
leoetlino, line 17: `static constexpr u32 NUM_BBOX_VALUES = 4;` (thanks to constexpr, there's no need for the enum trick to get true compile-time constants :D)

Source/Core/VideoBackends/D3D/D3DBoundingBox.cpp
phire: The check isn't needed for either backend. 

VideoCommon only requests Bounding Box reads when bounding box is enabled and supported by the backend.

phire (approved):
One thing I'm noticing (and it seems to extend to the VK and DX12 cache implementations too) is that bounding box behaviour is ill-defined during save-state load. 

At the very least, perhaps the caches should be flushed after a save-state load to prevent dirty cached bbox values from being loaded. But even that could produce invalid results if there is a save state between the finish of a draw and the bbox values being read.

I'm not sure if this is worth worrying about. This shouldn't effect regular users, just people doing debugging with save-states, which is already a hairy proposition with lots of edge cases

Techjar:
@stenzek Can you please rebase this?

stenzek:
^ I think this fifoci diff is a false herring, bounding box shouldn't affect anything here.

PR #9649, opened 2021-04-17
Make it possible to enable auto-updates by default with CMake builds

leoetlino:
This adds a CMake option (DOLPHIN_DEFAULT_UPDATE_TRACK) to allow configuring SCM_UPDATE_TRACK_STR. This is needed to enable auto-updates in Windows CMake builds by default.


CMakeLists.txt
```diff
@@ -17,6 +17,8 @@ project(dolphin-emu)
 # unique name here.
 set(DISTRIBUTOR "None" CACHE STRING "Name of the distributor.")
 
+set(DOLPHIN_DEFAULT_UPDATE_TRACK "" CACHE STRING "Name of the default update track. If empty, disables auto-update.")
```
iwubcode, line 20: My assumption is this approach let's you set the value via the command line if you don't want the default?

leoetlino: Yes, and it gives you a nice description for the option (just like DISTRIBUTOR).

iwubcode (approved):

spycrab (approved):
Looks sensible to me but I'm a bit out of the loop so take it with a grain of salt. (Although the change is so minimal I doubt it will break anything)

PR #9650, opened 2021-04-17
Put x86_64 Windows binaries in Binary/x64 for consistency with ARM64

leoetlino:
...and with our existing Visual Studio project files and buildbot config.


iwubcode (approved):

CMakeLists.txt
```diff
@@ -115,7 +115,9 @@ set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/Binaries)
 if (WIN32)
   set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_SOURCE_DIR}/Binary)
 
-  if (CMAKE_SYSTEM_PROCESSOR STREQUAL "aarch64")
+  if (CMAKE_SYSTEM_PROCESSOR MATCHES "x86_64|amd64|AMD64")
```
iwubcode, line 118: Not sure if it makes sense to use a temp variable with a `TO_LOWER` and compare once?

leoetlino: I don't think it's worth adding a temp variable, especially given how variable scope works in CMake. Efficiency is not a concern here.

PR #8296, opened 2019-08-05
PixelShaderGen: fix OOB indirect texcoord indices

Tilka:
Previously we set the indirect texture coordinate to zero, now we set the indirect texture coordinate *index* to zero. This fixes the ripple effect of the Mario painting in Luigi's Mansion ([issue 11462](https://bugs.dolphin-emu.org/issues/11462)).

Needs some hardware tests.

PR #9651, opened 2021-04-18
Fix out of bounds tex coord behavior; always apply fb_addprev and tex coord wrapping

Pokechu22:
Fixes the Mario portrait blur effect in Luigi's Mansion ([issue 11462](https://bugs.dolphin-emu.org/issues/11462)).  Closes #8296 (the older implementation).

When a tex coord is greater than or equal to the number of tex gens, the behavior is undefined on console, but it generally appears to be equivalent to using tex coord 0 (instead of clamping it, or setting the value to (0, 0), or just using whatever garbage is already around).  This applies to both direct and indirect textures.  There does seem to be some unpredictable garbage data that also can be used, but I don't know the details of it and that doesn't seem to be an issue with the Mario portrait.

There is an exception to this: when 0 tex gens are enabled, I always got a black image on console; ~~I didn't implement this in Dolphin because I couldn't think of a clean way of doing it (however, ubershaders seem to already do it)~~ this has been implemented as well.  I don't think this is something that games would do by accident since it's easy to notice.

For the Mario portrait specifically, the blur effect is achieved with indirect textures; they have a 128 by 128 texture (generated from an EFB (so at higher IRs, it will be a higher resolution, but the scale factor is still 128 by 128)) that [shows Mario](https://gist.githubusercontent.com/Pokechu22/897780fcb026f106f537e0cf0803fef0/raw/6857cff633c1d944c395f7286595c321b944a5d5/tar364_stage0_map0_mip0.png), and another 64 by 64 texture that contains [random greyscale noise](https://gist.githubusercontent.com/Pokechu22/897780fcb026f106f537e0cf0803fef0/raw/6857cff633c1d944c395f7286595c321b944a5d5/noise.png).  They also have separate texture coordinates for both of them (the Mario texture uses texture coordinate 0, with values ranging horizontally from .1 to .9 and vertically from 0 to 1, while the noise texture uses texture coordinate 1, with both axes ranging from 0 to 1).  Since the texture coordinates get multiplied by the texture scale (more or less; it's actually a separate parameter the game controls), texture coordinate 0 more strictly ranges from 12.8 to 115.2 on one axis and 0 to 128 on the other, and texture coordinate 1 ranges from 0 to 64 on both axes.

The only problem is that Nintendo forgot to set the number of tex gens to 2, so only texture coordinate 0 is valid, triggering this undefined behavior.  For the portrait, it (apparently) always renders using texture coordinate 0, and since the only purpose of the noise texture is to distort the image, that it was using the wrong texture coordinate presumably wasn't noticed.  The only difference this causes is that the noise is more fine-grained than it would have been, since texture coordinate 0's scale is twice that of texture coordinate 1, causing the texture to be repeated twice in each direction, but that (at least in my opinion) looks better than what it would look like with 2 tex gens.  (To be clear, with 2 tex gens, it would render correctly without this PR, and this PR does not change how it renders with 2 tex gens.)

Screenshots:

<details><summary>Broken portrait effect</summary>

MarioPortraitBroken
</details>
<details><summary>Fixed portrait effect</summary>

MarioPortraitFixed
</details>
<details><summary>"Intended" portrait effect (by hex-editing the FIFO to use 2 tex gens)</summary>

MarioPortraitIntended
</details>

Here is [my hardware test](https://gist.github.com/Pokechu22/dac1a542f6b99ef0b4140564b5d118fc), and here are some [fragmentary, partially incorrect notes](https://gist.github.com/Pokechu22/897780fcb026f106f537e0cf0803fef0).

### DIFF
--- a/Source/Core/VideoBackends/Software/Tev.cpp
+++ b/Source/Core/VideoBackends/Software/Tev.cpp
@@ -485,20 +485,16 @@ void Tev::Indirect(unsigned int stageNum, s32 s, s32 t)
 
   // matrix multiply - results might overflow, but we don't care since we only use the lower 24 bits
   // of the result.
-  const int indmtxid = indirect.mid & 3;
-  if (indmtxid)
+  if (indirect.matrix_index != IndMtxIndex::Off)
   {
-    const IND_MTX& indmtx = bpmem.indmtx[indmtxid - 1];
-    const int scale =
-        ((u32)indmtx.col0.s0 << 0) | ((u32)indmtx.col1.s1 << 2) | ((u32)indmtx.col2.s2 << 4);
+    const IND_MTX& indmtx = bpmem.indmtx[static_cast<u32>(indirect.matrix_index.Value()) - 1];
 
-    int shift;
+    const int shift = 17 - indmtx.GetScale();
 
-    switch (indirect.mid & 12)
+    switch (indirect.matrix_id)
     {
-    case 0:
+    case IndMtxId::Indirect:
       // matrix values are S0.10, output format is S17.7, so divide by 8
-      shift = (17 - scale);
       indtevtrans[0] = (indmtx.col0.ma * indcoord[0] + indmtx.col1.mc * indcoord[1] +
                         indmtx.col2.me * indcoord[2]) >>
                        3;
@@ -506,24 +502,28 @@ void Tev::Indirect(unsigned int stageNum, s32 s, s32 t)
                         indmtx.col2.mf * indcoord[2]) >>
                        3;
       break;
-    case 4:  // s matrix
+    case IndMtxId::S:
       // s is S17.7, matrix elements are divided by 256, output is S17.7, so divide by 256. - TODO:
       // Maybe, since s is actually stored as S24, we should divide by 256*64?
-      shift = (17 - scale);
       indtevtrans[0] = s * indcoord[0] / 256;
       indtevtrans[1] = t * indcoord[0] / 256;
       break;
-    case 8:  // t matrix
-      shift = (17 - scale);
+    case IndMtxId::T:
       indtevtrans[0] = s * indcoord[1] / 256;
       indtevtrans[1] = t * indcoord[1] / 256;
       break;
     default:
+      PanicAlertFmt("Invalid indirect matrix ID {}", indirect.matrix_id);
       return;
     }
 
     indtevtrans[0] = shift >= 0 ? indtevtrans[0] >> shift : indtevtrans[0] << -shift;
     indtevtrans[1] = shift >= 0 ? indtevtrans[1] >> shift : indtevtrans[1] << -shift;
+  }
+  else
+  {
+    // If matrix_index is Off (0), matrix_id should be Indirect (0)
+    ASSERT(indirect.matrix_id == IndMtxId::Indirect);
   }
 
   if (indirect.fb_addprev)
--- a/Source/Core/VideoBackends/Software/Tev.cpp
+++ b/Source/Core/VideoBackends/Software/Tev.cpp
@@ -6,6 +6,7 @@
 
 #include <algorithm>
 #include <cmath>
+#include <cstring>
 
 #include "Common/ChunkFile.h"
 #include "Common/CommonTypes.h"
@@ -559,8 +560,15 @@ void Tev::Draw()
     const int stageNum2 = stageNum >> 1;
     const int stageOdd = stageNum & 1;
 
-    const u32 texcoordSel = bpmem.tevindref.getTexCoord(stageNum);
+    u32 texcoordSel = bpmem.tevindref.getTexCoord(stageNum);
     const u32 texmap = bpmem.tevindref.getTexMap(stageNum);
+
+    // Quirk: when the tex coord is not less than the number of tex gens (i.e. the tex coord does
+    // not exist), then tex coord 0 is used (though sometimes glitchy effects happen on console).
+    // This affects the Mario portrait in Luigi's Mansion, where the developers forgot to set
+    // the number of tex gens to 2 (bug 11462).
+    if (texcoordSel >= bpmem.genMode.numtexgens)
+      texcoordSel = 0;
 
     const TEXSCALE& texscale = bpmem.texscale[stageNum2];
     const s32 scaleS = stageOdd ? texscale.ss1 : texscale.ss0;
@@ -592,8 +600,13 @@ void Tev::Draw()
     const TevStageCombiner::ColorCombiner& cc = bpmem.combiners[stageNum].colorC;
     const TevStageCombiner::AlphaCombiner& ac = bpmem.combiners[stageNum].alphaC;
 
-    const int texcoordSel = order.getTexCoord(stageOdd);
-    const int texmap = order.getTexMap(stageOdd);
+    u32 texcoordSel = order.getTexCoord(stageOdd);
+    const u32 texmap = order.getTexMap(stageOdd);
+
+    // Quirk: when the tex coord is not less than the number of tex gens (i.e. the tex coord does
+    // not exist), then tex coord 0 is used (though sometimes glitchy effects happen on console).
+    if (texcoordSel >= bpmem.genMode.numtexgens)
+      texcoordSel = 0;
 
     Indirect(stageNum, Uv[texcoordSel].s, Uv[texcoordSel].t);
 
@@ -603,8 +616,17 @@ void Tev::Draw()
       // RGBA
       u8 texel[4];
 
-      TextureSampler::Sample(TexCoord.s, TexCoord.t, TextureLod[stageNum], TextureLinear[stageNum],
-                             texmap, texel);
+      if (bpmem.genMode.numtexgens > 0)
+      {
+        TextureSampler::Sample(TexCoord.s, TexCoord.t, TextureLod[stageNum],
+                               TextureLinear[stageNum], texmap, texel);
+      }
+      else
+      {
+        // It seems like the result is always black when no tex coords are enabled, but further
+        // hardware testing is needed.
+        std::memset(texel, 0, 4);
+      }
 
 #if ALLOW_TEV_DUMPS
       if (g_ActiveConfig.bDumpTevTextureFetches)
--- a/Source/Core/VideoCommon/PixelShaderGen.cpp
+++ b/Source/Core/VideoCommon/PixelShaderGen.cpp
@@ -991,7 +991,7 @@ static void WriteStage(ShaderCode& out, const pixel_shader_uid_data* uid_data, i
       // TODO: Should we reset alphabump to 0 here?
     }
 
-    if (tevind.mid != 0)
+    if (tevind.matrix_index != IndMtxIndex::Off)
     {
       // format
       static constexpr std::array<const char*, 4> tev_ind_fmt_mask{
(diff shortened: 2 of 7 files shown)


Pokechu22:
The difference in Fortune Street is/should be correct; the one part of the water references a nonexistant tex coord on stage 1.  I'm not entirely sure why there aren't any differences for it on OGL.  Note that it only has one tex coord in the vertex spec (coord 0), with coords 1 and 2 coming from the normals.

<details>
For object 644 (commands are in 643 at the moment):

000d1049:
<pre>
BP register BPMEM_TREF number 0
Stage 0 texmap: 0
Stage 0 tex coord: 0
Stage 0 enable texmap: Yes
Stage 0 color channel: Color chan 0 (0)
Stage 1 texmap: 3
Stage 1 tex coord: 3
Stage 1 enable texmap: Yes
Stage 1 color channel: Color chan 1 (1)
</pre>

000d11f9:
<pre>
BP register BPMEM_GENMODE
Num tex gens: 3
Num color channels: 2
Unused bit: 0
Flat shading (unconfirmed): No
Multisampling: No
Num TEV stages: 5
Cull mode: Back-facing primitives only (1)
Num indirect stages: 0
ZFreeze: No
</pre>
</details>

I also implemented the 0 tex coords making things black case.

Lastly, I [hardware tested](https://gist.github.com/Pokechu22/8396169343eab4f759cbaa279fb66d78) it, and `fb_addprev` applies even if the indirect stage's matrix is off or if the number of indirect stages is 0; currently that works correctly for the software renderer but not specialized shaders or ubershaders (specifically for the indirect stages = 0 case).  (Note that that hardware test also exhibits the other undefined behavior where it uses an indirect stage when the number of indirect stages is 0; you can see a slightly different pattern in the `0YY` and `0YN` cases.  That was only due to lazyness on my part but it's still interesting.)

Pokechu22:
I did one more hardware test and confirmed that in addition to `fb_addprev`, the indirect texture wrap functions also work when the indirect stage is not active ([here's the test](https://gist.github.com/Pokechu22/55cb365518defd58d6a294c22e566787)).  I don't think there's a situation when they'd be *useful* (regular textures already wrap properly (as long as they're a power of 2, but indirect texture wrap only does powers of 2 as well)); I think it's only useful when you have an indirect texture to implement tilemaps/spritesheets/something like that where the indirect texture selects the tile/sprite.  I also found that a value of 7 behaves the same as 6 (`ITW_0`), in that it results in the texture coord always being 0; shader generation currently *crashes* in that case.  (Side note, the comment about the value being `1` and not `1 << 7` was wrong; the value should be 1 because `1 - 1 = 0` which would lead to `x & 0` => 0, just like the first value is `0` which leads to `0 - 1 = -1` so `x & -1` = `x`.  Except all of those are pointless, because the array isn't used in those cases.  The comment was added in 3a6389992e6 for reference.)

Source/Core/VideoCommon/UberShaderPixel.cpp
```diff
@@ -150,15 +150,23 @@ ShaderCode GenPixelShader(APIType ApiType, const ShaderHostConfig& host_config,
   // Uniform index -> texture coordinates
   if (numTexgen > 0)
   {
+    for (u32 i = 0; i < numTexgen; i++)
+    {
+      out.Write("\tint2 fixpoint_uv{} = int2(", i);
+      out.Write("(tex{}.z == 0.0 ? tex{}.xy : tex{}.xy / tex{}.z)", i, i, i, i);
+      out.Write(" * " I_TEXDIMS "[{}].zw);\n", i);
+      // TODO: S24 overflows here?
```
iwubcode, line 162: What's the story here?

Source/Core/VideoCommon/PixelShaderGen.cpp
```diff
@@ -240,9 +237,12 @@ PixelShaderUid GetPixelShaderUid()
   {
     uid_data->stagehash[n].tevorders_texcoord = bpmem.tevorders[n / 2].getTexCoord(n & 1);
 
+    // hasindstage previously was used as a criterion to set tevind to 0, but there are variables in
+    // tevind that are used even if the indirect stage is disabled, so now it is only left in to
+    // avoid breaking existing UIDs (in most cases, games will have 0 in tevind anyways)
+    // TODO: Remove hasindstage on the next UID version bump
```
iwubcode, line 243: Wondering if this TODO should be next to the UID with a reference to this location?

Source/Core/VideoCommon/UberShaderPixel.cpp
Pokechu22: The same comment exists in specialized shaders:

https://github.com/dolphin-emu/dolphin/blob/4d37dad20d230b003b74e38ff4c5af68e6ed232e/Source/Core/VideoCommon/PixelShaderGen.cpp#L786-L792

Ubershaders already had (basically) the same code, just in a different place and without the comment.

The software renderer doesn't seem to worry about it at all:

https://github.com/dolphin-emu/dolphin/blob/4d37dad20d230b003b74e38ff4c5af68e6ed232e/Source/Core/VideoBackends/Software/Rasterizer.cpp#L228-L238

https://github.com/dolphin-emu/dolphin/blob/4d37dad20d230b003b74e38ff4c5af68e6ed232e/Source/Core/VideoBackends/Software/Rasterizer.cpp#L117-L123

It's something that might need further investigation, given the inconsistent results I got with my first attempt at hardware testing where I ended up overflowing really hard.  The [zww-water](https://fifo.ci/dff/zww-water/) test case is also incorrect in the software renderer so it might be wrong.  But those are something to come back to in the future.

Source/Core/VideoCommon/PixelShaderGen.cpp
Pokechu22: Good idea, done.

Source/Core/VideoBackends/Software/Tev.cpp
```diff
@@ -603,8 +615,17 @@ void Tev::Draw()
...
+      if (bpmem.genMode.numtexgens > 0)
+      {
+        TextureSampler::Sample(TexCoord.s, TexCoord.t, TextureLod[stageNum],
+                               TextureLinear[stageNum], texmap, texel);
+      }
+      else
+      {
+        // It seems like the result is always black when no tex coords are enabled, but further
+        // hardware testing is needed.
+        memset(texel, 0, 4);
```
leoetlino, line 627: `std::memset(texel, 0, sizeof(texel))` (and `<cstring>` needs to be included)

Source/Core/VideoCommon/PixelShaderGen.cpp
```diff
@@ -953,11 +963,8 @@ static void WriteStage(ShaderCode& out, const pixel_shader_uid_data* uid_data, i
   if (!has_tex_coord)
     texcoord = 0;
 
-  if (stage.hasindstage)
   {
-    TevStageIndirect tevind;
-    tevind.hex = stage.tevind;
-
+    const TevStageIndirect tevind{.hex = stage.tevind};
```
leoetlino, line 967: Designated initializers are a C++20 feature. GCC and Clang support it in C++17 but only as a nonstandard extension.

Pokechu22: It's also supported by MSVC for C++17.  I already used them in several places in #9497 (which has been merged, and would be a fair bit more awkward to adjust there because I use it to avoid variables), so I'd rather leave it like this.

leoetlino: Hmm, fair. I guess we'll be able to use C++20 soon anyway, assuming macOS and Android don't hold us back...

Source/Core/VideoCommon/PixelShaderGen.cpp
```diff
@@ -1038,11 +1038,14 @@ static void WriteStage(ShaderCode& out, const pixel_shader_uid_data* uid_data, i
                   tev_ind_bias_add[u32(tevind.fmt.Value())]);
       }
 
+      // Multiplied by 2 because each matrix has two rows.
+      // Note also that the 4th column of the matrix contains the scale factor.
+      const u32 mtxidx = 2 * (static_cast<u32>(tevind.matrix_index.Value()) - 1);
```
iwubcode, line 1047: Is the index one based?  Why did you subtract 1?

iwubcode: (I do see that this was similar to the previous solution, just curious if you know why)

Source/Core/VideoCommon/PixelShaderGen.h
```diff
@@ -133,6 +133,7 @@ struct pixel_shader_uid_data
     u32 pad1 : 6;
 
     // TODO: Clean up the swapXY mess
+    // TODO: remove hasindstage, as it no longer does anything useful
```
iwubcode, line 136: Is it being used anymore?  Is there a reason we don't just remove it now?

iwubcode:
The last two commits change the pixelshader/ubershader to always run the indirect stage logic.  Was that already present in the software renderer or possibly not applicable here?

Source/Core/VideoCommon/PixelShaderGen.cpp
Pokechu22: Yes, it's one-based; see the enum:

https://github.com/dolphin-emu/dolphin/blob/c37b5fe144c9bdf21622822212127cbd26651cae/Source/Core/VideoCommon/BPMemory.h#L303-L314

(See also [the defines in libogc](https://github.com/devkitPro/libogc/blob/bc4b778d558915aa40676e33514c4c9ba2af66b8/gc/ogc/gx.h#L819-L832), which uses a combined `indtexmtx` field; note how 4 and 8 are skipped.  I haven't actually tested what happens with 4 and 8 (or anything with the "s-type" or "t-type" matrices), but I have to assume it still uses an all-zero matrix so it's equivalent to 0.)

iwubcode: Ah right, my apologies, I had worked out all the details the other day but completely spaced tonight.  Still a bit out of it.

Pokechu22:
Yes, the software renderer always ran it (in `Tev::Indirect`, which is always called even if no indirect stage is enabled).  There actually was a comment about the software renderer doing that in PixelShaderManager (with the `stage < bpmem.genMode.numindstages` condition above):

https://github.com/dolphin-emu/dolphin/blob/ac679eb24d4882c48be4b47971427f834f8b09c2/Source/Core/VideoCommon/PixelShaderManager.cpp#L161-L162

iwubcode (approved):

JMC47 (approved):
I went through and played a few games with this pull request.  I made sure nothing was really broken.  As was said on IRC, we have pretty good coverage on indirect textures so there's not really a high chance anything will break.  If it does, we could fix it.

JMC47:
Finally verified how Mario looks in the final cutscene.  It looks identical to console on a zoom in.

GLME01_2021-05-09_14-43-24

https://gist.githubusercontent.com/Pokechu22/897780fcb026f106f537e0cf0803fef0/raw/6857cff633c1d944c395f7286595c321b944a5d5/Screenshot3h13m29s00f.png

Pokechu22:
4:3 images:

<details><summary>Broken portrait effect</summary>

MarioPortraitBroken
</details>
<details><summary>Fixed portrait effect</summary>

MarioPortraitFixed
</details>
<details><summary>"Intended" portrait effect (by hex-editing the FIFO to use 2 tex gens)</summary>

MarioPortraitIndented
</details>

PR #9652, opened 2021-04-18
FreeLook: add separate camera for modifying orthographic projections from games

iwubcode:
This adds logic to Dolphin that allows you to modify orthographic projections from games ("2D").

Not all games will be able to take advantage of this (as some do weird things with 2d) but in theory it can remove bloom or disable the HUD by moving them off the "screen".

Possibly, it can be used to support modifying the aspect ratio of the 2d elements so that VR or ultra-wide monitors can feel more comfortable.

Best case, it can be used for some cool screenshots on certain games (though some games will need added post processing effects to fix missing bloom..coming soon™).

Before:

image

After:

image

--

Initial 2d edit screen:

image

2d / ortho freelook camera controller screen (no defaults, these are what I used to test):

image

### DIFF
--- a/Source/Core/Core/Config/FreeLookSettings.cpp
+++ b/Source/Core/Core/Config/FreeLookSettings.cpp
@@ -14,6 +14,8 @@ namespace Config
 // Configuration Information
 const Info<bool> FREE_LOOK_ENABLED{{System::FreeLook, "General", "Enabled"}, false};
 
+const Info<bool> FREE_LOOK_2D_ENABLED{{System::FreeLook, "General", "2DEnabled"}, false};
+
 // FreeLook.Controller1
 const Info<FreeLook::ControlType> FL1_CONTROL_TYPE{{System::FreeLook, "Camera1", "ControlType"},
                                                    FreeLook::ControlType::SixAxis};
--- a/Source/Core/Core/Config/FreeLookSettings.h
+++ b/Source/Core/Core/Config/FreeLookSettings.h
@@ -16,6 +16,7 @@ namespace Config
 // Configuration Information
 
 extern const Info<bool> FREE_LOOK_ENABLED;
+extern const Info<bool> FREE_LOOK_2D_ENABLED;
 
 // FreeLook.Controller1
 extern const Info<FreeLook::ControlType> FL1_CONTROL_TYPE;
--- a/Source/Core/Core/FreeLookConfig.cpp
+++ b/Source/Core/Core/FreeLookConfig.cpp
@@ -32,6 +32,7 @@ Config::Config()
 {
   camera_config.control_type = ControlType::SixAxis;
   enabled = false;
+  enabled_2d = false;
 }
 
 void Config::Refresh()
@@ -44,5 +45,6 @@ void Config::Refresh()
 
   camera_config.control_type = ::Config::Get(::Config::FL1_CONTROL_TYPE);
   enabled = ::Config::Get(::Config::FREE_LOOK_ENABLED);
+  enabled_2d = ::Config::Get(::Config::FREE_LOOK_2D_ENABLED);
 }
 }  // namespace FreeLook
--- a/Source/Core/Core/FreeLookConfig.h
+++ b/Source/Core/Core/FreeLookConfig.h
@@ -31,6 +31,7 @@ struct Config final
 
   CameraConfig camera_config;
   bool enabled;
+  bool enabled_2d;
 };
 
 Config& GetConfig();
--- a/Source/Core/Core/FreeLookManager.cpp
+++ b/Source/Core/Core/FreeLookManager.cpp
@@ -16,6 +16,7 @@
 #include "InputCommon/InputConfig.h"
 
 #include "VideoCommon/FreeLookCamera.h"
+#include "VideoCommon/FreeLookCamera2D.h"
 #include "VideoCommon/OnScreenDisplay.h"
 
 namespace
@@ -72,6 +73,26 @@ enum GyroButtons
   RollRight,
   YawLeft,
   YawRight,
+};
+}
+
+namespace StretchButtons
+{
+enum StretchButtons
+{
+  IncreaseX,
+  DecreaseX,
+  IncreaseY,
+  DecreaseY,
+};
+}
+
+namespace ZoomButtons
+{
+enum ZoomButtons
+{
+  Increase,
+  Decrease,
 };
 }
 }  // namespace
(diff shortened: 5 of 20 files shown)


iwubcode:
@MichaelJW - this is going to be one of the last of the Free Look PRs.  I have one more waiting on a couple of the others to be merged (the "surprise" I mentioned).

Please let me know what you think of this.  It may not be as useful for you as you hoped but I think we can build upon it in the future.

iwubcode:
While we technically have the anti-blur code for Tales of Symphonia, if you didn't have that you could use this!

Before:

image

After:

image

MichaelJW:
🤯 

This works _incredibly well_ in combination with FOV adjustment. OK, granted, I've only tried it with one game so far, but it worked immediately - I could set the FOV to 0.5 in both directions to see all around, and then stretch and move the orthographic camera controller so that the HUD remained visible in the centre of view.

Amazing work!

I am of course going to keep playing with this and I'll let you know how I get on. Please let me know if there's anything specific you'd like me to test.

iwubcode:
Great to hear @MichaelJW !

I spent some of this morning trying to find a solution for the games that this doesn't work well on.  I found one but it's _really_ klunky.  Basically have to say "skip x 2d elements".  This is fine for screenshots but for normal gameplay it really varies on the game and even for the scene.

Something like Mario Kart Wii can get by with barely changing the value.  However for the Last Story, just change the camera slightly and you're suddenly drawing more 2d elements.  Games like Rune Factory have less 2d elements indoors than outdoors.

This isn't a very good solution.  Luckily, I've been looking into this for months now (off and on) as I wanted something similar for a different feature.  I have another idea I'll try and see if I can make work.

If I can get it working, it'll still be game specific but won't require scene-by-scene changes.  If I can't then the skip solution will be better than nothing.

DrBoiffard:
Hi,

Sorry for the intromission . Could this feature let me configure a "expanded 4:3" view for playing in 3 monitors?
https://forums.dolphin-emu.org/Thread-feature-request-thread?page=186

If I've understund correctly I could move/stretch the 2d layer as desired, am I correct? I don't know if it'd be possible to "preconfigure" this "expanded" 4:3 i was trying to describe there

Thank you very much

iwubcode:
> Could this feature let me configure a "expanded 4:3" view for playing in 3 monitors?

I wasn't really sure what you meant by expanded black bars.  I don't believe this feature will impact letterboxing, though I primarily play Wii games that have 16:9 support already.

> If I've understood correctly, I could move/stretch the 2d layer as desired, am I correct?

Correct.  It currently applies to all 2d elements drawn.  In the future, I hope to be able to tweak individual 2d layers.

> I woudn't be using the freelook itself, so I don't know if it'd be possible to "preconfigure" this "expanded" 4:3 i was trying to describe there, some generalisation indepently of how wide the screen is, or the freelook being intensively used.

With this PR, there's now 2d freelook and 3d freelook.  When you say you wouldn't be using freelook, I assume you mean the 3d freelook.  You'd have to use the 2d freelook to modify the 2d layers.

I can likely expose some settings to allow you to preconfigure this per-game.  Though I'll likely do that as a follow-up PR once the feature is usable by more games.  There's no plans to support specific layouts at the moment, as the thought is there will be a variety of use cases:

* Removing hud for nicer screenshots or more immersive gameplay
* Removing bloom or other effects so get a cleaner image and can enhance it with post-processing
* Tweaking hud for ultrawide / multi-monitor / VR experiences

DrBoiffard:
Thank you very much for your responses! 

Right now I only use the 3 monitor config with racing games (specially F-ZERO GX, Mario Kart and R:Racing for the moment).  

I never think of this behaviour it as a "2d freelook", but as an option that'd appear when using "aspect ratio= fill to screen", cause that's when needed. Whith this aspect ratio option, the 3d works perfect in widescreen, but for the 2d layer, a natural "aspect ratio" and "fill to screen" are contraries, so it does just the later: The 2d (GUI and menus layer) stretches to fill the screen, so the horizontal stretching it's enormous.

The idea it'd be changing the "horzontal stretching" and draw the 2d layer unstretched in the corresponding "central 4:3 area" of the screen.  As long as the natural 4:3 (or 16:9 when present) proportions are constant for a vertical value, and the "centered position" is easily calculable for any given aspect ratio, I think it can be easy to generalizate independtly of the "wide" of the screen. So this "presetting" I was thinking would be just precalculate the values which you'd usually change with the buttons used to config this option.

Maybe I'm not explaining very well, sorry. At the end the desired behaviour it's just like supermodel's: central 4:3 area is what it'd show in a 4:3 display, and the rest of the screen is more 3D instead of black bars. It does the same for 16:9, 21:9, 48:9....
https://www.youtube.com/watch?v=56mNv-Rv-So
 
Thanks,

DrBoiffard:
Maybe I'm too selfish cause I'm only thinking in racing games, but a preconfigured option in which the central area of the screen coincide 100% with what it'd be shown in a 4:3, black-bar way can be useful in other games genres. maybe RPGs with pixelart in textboxes, that I guess are ultrastretched

leoetlino:
@DrBoiffard please refrain from editing your comment so excessively often. All those edits are producing a large number of notifications in our IRC channel.

DrBoiffard:
Ok I'm so sorry

DrBoiffard:
I've succesfully compiled the branch. It's pretty cool, it works exactly as I imagined. 
imagen

imagen

DrBoiffard:
Anyways as the buttons make smaller leaps in both placing and stretching,  I think I haven't centered it perfect, it's a little left-leanint, and maybe I don't have it unstretched to it's natural H/V proportion.

DrBoiffard:
In Mario Kart there's some rare "3d reflections" mixed with the 2D layer. If I reset the option so it goes back to the default stretching, it seems to dissappear, so it appears to be taken out of the corrected 3D aspect ratio without reason. 

Here I moved it down to see more clearly.
imagen

iwubcode:
> In Mario Kart there's some rare "3d reflections" mixed with the 2D layer. If I reset the option so it goes back to the default stretching, it seems to dissappear, so it appears to be taken out of the corrected 3D aspect ratio without reason.

This is just the first attempt at a solution to get something merged into master.  Games do weird things with 2d drawing (ex: drawing bloom, color correction, etc).  In your Mario Kart screenshot you can see the shadows and other elements are drawn as 2d and modified when  you stretch / move the content around.

Dolphin has no way of generically knowing what the HUD is (or any other 2d element), it just knows it is drawing _some_ 2d element.

In a future PR, I hope to have a solution that allows users to call out specific 2d elements.  However, this will be done on a game-by-game basis.

> Thank you very much for your responses!

You are welcome.  To respond to your other comments, I currently have no plans for a generic auto-correction mode for the 2d elements based on aspect ratio.  The current solution is meant to be used for a variety of use cases, one of which is multi-monitor / ultra-wide aspect ratio correction.

DrBoiffard:
Thank you very much! 

It's ok like this, I don't mind what to touch to get here.  The only problem I see, as said before,  is the "steps" of each button assigned to "stretch" and "move" the layer are considerably big. So it's impossible to center it  perfect. And I suspect the same is happening for the stretching, I stop when I see the less deformed possible but I'm not sure if its the native 4:3 h/V proportion or just the closest I can get with the buttons.  

If  there's not going to be an auto-correction based on aspect-ratio: could it be then an extra option in the option menu to make smaller the stretching/moving of each button pressing? So it's possible to adjust it more precisely.

Thanks

DrBoiffard:
I think it'd be enough with an on/off button to change from the present "increment jumps" into "pixel-by-pixel" scale, affecting the other 8 buttons. So you can place it more or less in the place you want, like it is now, and then you'd press this button and pass to a fine adjustment with the same 8 buttons.

iwubcode:
> I think it'd be enough with an on/off button to change from the present "increment jumps" into "pixel-by-pixel" scale

I'll have to double check and see but my initial thought is we can't move based on pixels because we're actually applying a matrix transformation to the vertices of the 2d elements.

We already have a way to reduce the movement rate, that's "speed" but we could add that for the stretching capability as well.

iwubcode:
No functional changes.  I updated the code to mimic the flow defined in #9670 .  This prepares this camera to be controlled by other means (ex: UDP).

I also renamed all code references from "Ortho" to "2D".  This is just on the off chance we want to implement an actual orthographic camera for the perspective projection calls (I've thought about it but not sure it would be very useful).  All user facing references are also "2D" (settings and qt).

iwubcode:
Superseded by #10518

PR #9655, opened 2021-04-18
GameINI: Add speed hack for Korean TP

PPLToast:
Closes https://bugs.dolphin-emu.org/issues/7825


ImSpecial:
Why is this outside the GameSettings folder?

leoetlino:
Oh, doh. Was so focused on the contents of the patch that I missed the file path being incorrect. I'll fix it.

PR #9657, opened 2021-04-19
DataReport: Amend conditional test for data reports in IsValidMode

lioncash:
This particular range check is kind of odd, and would only interpret interleave mode 2 as a valid mode, while rejecting interleave mode 1 and the extension byte mode.

As far as I know, and, based off the information on Wiibrew, we should be considering all three values within this range as valid.


leoetlino:
That check does look odd... @jordan-woyak?

BhaaLseN:
The change makes sense (now), but I couldn't find the equivalent piece of code before the big refactoring.

lioncash:
@jordan-woyak Sorry for the repeat ping, but would you be able to take a look at this?

jordan-woyak (approved):

jordan-woyak:
Sorry, missed the first ping. Looks good to me. Thanks for catching this.

PR #9661, opened 2021-04-21
InputCommon: Add a new function ntimes(input, n)

MichaelJW:
Example: `ntimes(A, 5)` will output 0.0 until A is pressed, then alternately output 1.0 and 0.0 five times (when polled), and then output 0.0 until A is released and pressed again.

Example: `ntimes(0.7, 3)` will output `0.7 0.0 0.7 0.0 0.7 0.0 0.0 0.0 0.0 ...`.

This is very similar to `toggle()`, except that (a) the first argument does not have to be greater than 0.5 to trigger it, just non-zero, and (b) it won't trigger again until an input of zero has been received again.

This might not have any practical use in games that isn't already covered by other functions (especially `toggle()`). My interest is in using it to make precise changes to Dolphin settings. For instance, setting *Decrease FOV X* and *Decrease FOV Y* to `ntimes(A, 20)` lets me press a single button to change both FOV settings to exactly 0.500. I am not aware of another way to do this via a controller binding.

### DIFF
--- a/Source/Core/InputCommon/ControlReference/FunctionExpression.cpp
+++ b/Source/Core/InputCommon/ControlReference/FunctionExpression.cpp
@@ -636,6 +636,57 @@ private:
   mutable Clock::time_point m_release_time = Clock::now();
 };
 
+// usage: ntimes(input, n = 1)
+class NTimesExpression : public FunctionExpression
+{
+private:
+  ArgumentValidation
+  ValidateArguments(const std::vector<std::unique_ptr<Expression>>& args) override
+  {
+    if (args.size() == 1 || args.size() == 2)
+      return ArgumentsAreValid{};
+    else
+      return ExpectedArguments{"input, times = 1"};
+  }
+
+  ControlState GetValue() const override
+  {
+    if (m_running)
+    {
+      m_loops++;
+      if (m_loops >= m_times * 2 - 1)
+        m_running = false;
+
+      if (m_loops % 2 == 1)
+        return m_input;
+      else
+        return 0.0;
+    }
+    else if (m_zeroed && GetArg(0).GetValue() != 0.0)
+    {
+      m_input = GetArg(0).GetValue();
+      m_times = GetArg(1).GetValue();
+      m_running = true;
+      m_loops = 1;
+      m_zeroed = false;
+      return m_input;
+    }
+    else
+    {
+      if (GetArg(0).GetValue() == 0.0)
+        m_zeroed = true;
+      return 0.0;
+    }
+  }
+
+private:
+  mutable bool m_zeroed = true;
+  mutable bool m_running = false;
+  mutable int m_loops = 0;
+  mutable ControlState m_input = 0.0;
+  mutable int m_times = 0;
+};
+
 std::unique_ptr<FunctionExpression> MakeFunctionExpression(std::string_view name)
 {
   if (name == "not")
@@ -684,6 +735,8 @@ std::unique_ptr<FunctionExpression> MakeFunctionExpression(std::string_view name
     return std::make_unique<RelativeExpression>();
   if (name == "pulse")
     return std::make_unique<PulseExpression>();
+  if (name == "ntimes")
+    return std::make_unique<NTimesExpression>();
 
   return nullptr;
 }
--- a/Source/Core/InputCommon/ControlReference/FunctionExpression.cpp
+++ b/Source/Core/InputCommon/ControlReference/FunctionExpression.cpp
@@ -652,7 +652,7 @@ private:
   ControlState GetValue() const override
   {
     const ControlState input = GetArg(0).GetValue();
-    const int times = (GetArgCount() >= 2) ? int(GetArg(1).GetValue()) : 1;
+    const u32 times = GetArgCount() >= 2 ? u32(std::max(GetArg(1).GetValue() + 0.5, 0.0)) : 1;
 
     if (m_running)
     {
@@ -685,9 +685,9 @@ private:
 private:
   mutable bool m_zeroed = true;
   mutable bool m_running = false;
-  mutable int m_loops = 0;
+  mutable u32 m_loops = 0;
   mutable ControlState m_input = 0.0;
-  mutable int m_times = 0;
+  mutable u32 m_times = 0;
 };
 
 std::unique_ptr<FunctionExpression> MakeFunctionExpression(std::string_view name)
--- a/Source/Core/InputCommon/ControlReference/FunctionExpression.cpp
+++ b/Source/Core/InputCommon/ControlReference/FunctionExpression.cpp
@@ -646,13 +646,13 @@ private:
     if (args.size() == 1 || args.size() == 2)
       return ArgumentsAreValid{};
     else
-      return ExpectedArguments{"input, [times = 1]"};
+      return ExpectedArguments{"input, times = 1"};
   }
 
   ControlState GetValue() const override
   {
     const ControlState input = GetArg(0).GetValue();
-    const ControlState times = GetArg(1).GetValue();
+    const int times = (GetArgCount() >= 2) ? int(GetArg(1).GetValue()) : 1;
 
     if (m_running)
     {
--- a/Source/Core/InputCommon/ControlReference/FunctionExpression.cpp
+++ b/Source/Core/InputCommon/ControlReference/FunctionExpression.cpp
@@ -636,7 +636,7 @@ private:
   mutable Clock::time_point m_release_time = Clock::now();
 };
 
-// usage: ntimes(input, n = 1)
+// usage: ntimes(input, times = 1)
 class NTimesExpression : public FunctionExpression
 {
 private:
@@ -646,11 +646,14 @@ private:
     if (args.size() == 1 || args.size() == 2)
       return ArgumentsAreValid{};
     else
-      return ExpectedArguments{"input, times = 1"};
+      return ExpectedArguments{"input, [times = 1]"};
   }
 
   ControlState GetValue() const override
   {
+    const ControlState input = GetArg(0).GetValue();
+    const ControlState times = GetArg(1).GetValue();
+
     if (m_running)
     {
       m_loops++;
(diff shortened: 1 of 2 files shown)


Filoppi:
This can't accurately work at the moment, GetValue() on function expressions can be called any number of times within a frame, you can't use a counter there.
This is why I've been re-writing part of the input system in this PR https://github.com/dolphin-emu/dolphin/pull/9489/ (that's what input caching means, it guarantees GetValue() to only be called once per frame).

Source/Core/InputCommon/ControlReference/FunctionExpression.cpp
```diff
@@ -636,6 +636,57 @@ class PulseExpression : public FunctionExpression
   mutable Clock::time_point m_release_time = Clock::now();
 };
 
+// usage: ntimes(input, n = 1)
```
Filoppi, line 639: @jordan-woyak camel case 😂?

Nitpick but it should be
`// usage: ntimes(input, times = 1)`
to be aligned with the expected arguments.

Filoppi: Nitpick but it should be
// usage: ntimes(input, times = 1)
to be aligned with the expected arguments.

Source/Core/InputCommon/ControlReference/FunctionExpression.cpp
```diff
@@ -636,6 +636,57 @@ class PulseExpression : public FunctionExpression
...
+
+      if (m_loops % 2 == 1)
+        return m_input;
+      else
+        return 0.0;
+    }
+    else if (m_zeroed && GetArg(0).GetValue() != 0.0)
+    {
+      m_input = GetArg(0).GetValue();
+      m_times = GetArg(1).GetValue();
```
Filoppi, line 668: For consistency, all function expressions should read and cache all their possible arguments at the top.
This is in case the arg is also another expression which modifies its state in its GetValue(), so it needs to be called for every call of this function GetValue(), not only when some conditions are true.

Filoppi (commented):
Review

Filoppi (commented):
Review

Source/Core/InputCommon/ControlReference/FunctionExpression.cpp
MichaelJW: Thanks, I think I see what you mean. I've inserted new `const ControlState`s `input` and `times` at the top of `GetValue()` to store the current inputs while retaining the original inputs in `m_input` and `m_times` - does that cover it?

Source/Core/InputCommon/ControlReference/FunctionExpression.cpp
MichaelJW: D'oh, thank you. I couldn't decide whether the second argument should be called `times` or `n` and apparently went for both without noticing 🤦

Source/Core/InputCommon/ControlReference/FunctionExpression.cpp
```diff
@@ -636,6 +636,60 @@ class PulseExpression : public FunctionExpression
   mutable Clock::time_point m_release_time = Clock::now();
 };
 
+// usage: ntimes(input, times = 1)
```
Filoppi, line 639: should now be [times = 1] 😂

Source/Core/InputCommon/ControlReference/FunctionExpression.cpp
```diff
@@ -636,6 +636,60 @@ class PulseExpression : public FunctionExpression
...
+    if (args.size() == 1 || args.size() == 2)
+      return ArgumentsAreValid{};
+    else
+      return ExpectedArguments{"input, [times = 1]"};
+  }
+
+  ControlState GetValue() const override
+  {
+    const ControlState input = GetArg(0).GetValue();
+    const ControlState times = GetArg(1).GetValue();
```
Filoppi, line 655: You forgot to fallback to 1 (default value) if the arg isn't defined.

Filoppi (commented):
Looks good otherwise. Haven't tested though. Probably worth to squash the commits.

Source/Core/InputCommon/ControlReference/FunctionExpression.cpp
```diff
@@ -636,6 +636,60 @@ class PulseExpression : public FunctionExpression
...
+class NTimesExpression : public FunctionExpression
+{
+private:
+  ArgumentValidation
+  ValidateArguments(const std::vector<std::unique_ptr<Expression>>& args) override
+  {
+    if (args.size() == 1 || args.size() == 2)
+      return ArgumentsAreValid{};
+    else
+      return ExpectedArguments{"input, [times = 1]"};
```
Filoppi, line 649: you don't need [], that is used for arguments which can be missing but are ignored (like if they have no fallback/default value). `times = 1` is good

Filoppi (commented):
review

MichaelJW:
> This can't accurately work at the moment, GetValue() on function expressions can be called any number of times within a frame, you can't use a counter there.

Ah right, I see. (Does that mean that `toggle()` doesn't work correctly at the moment either?)

> This is why I've been re-writing part of the input system in this PR #9489 (that's what input caching means, it guarantees GetValue() to only be called once per frame).

Oh, nice, I saw your PR before but didn't grasp the significance of it for this.

Thanks for the feedback!

Filoppi:
> > This can't accurately work at the moment, GetValue() on function expressions can be called any number of times within a frame, you can't use a counter there.
> 
> Ah right, I see. (Does that mean that `toggle()` doesn't work correctly at the moment either?)
> 
> > This is why I've been re-writing part of the input system in this PR #9489 (that's what input caching means, it guarantees GetValue() to only be called once per frame).
> 
> Oh, nice, I saw your PR before but didn't grasp the significance of it for this.
> 
> Thanks for the feedback!

toggle() and your function will probably work most of the times, but they are not reliable, and depending on which binding/button they are linked to on the emulated controller, they might be read more than once per input frame.

Still I don't understand the point of this function.
If you want to change the FOV every x frames by y, why don't you just change it every frame by y/x?

MichaelJW:
> Still I don't understand the point of this function.
> If you want to change the FOV every x frames by y, why don't you just change it every frame by y/x?

That would be ideal, but I couldn't see a way to do that reliably. 

I'm working on a VR interface that uses DSU, and I would like to be able to say, essentially, "when X is pressed, decrease horizontal and vertical FOV to 0.5" (and also modify the orthographic cameras similarly using the changes in @iwubcode's PR #9652). 

However, if I map "decrease horizontal FOV" to `X`, then when I pass a DSU packet containing an "X is pressed" signal it keeps triggering "decrease horizontal FOV" until I pass an "X is not pressed" DSU packet, and I haven't been able to get it to trigger a precise number of "decreases" reliably by changing the time between the two DSU packets.

Adding a new function that does the work of ensuring the command is triggered exactly n times seemed like the simplest way to achieve this, and could apply to FOV, to @iwubcode's new orthographic camera configs, and to other config settings like stereoscopic depth and convergence, increase/decrease internal resolution, etc, without having to make any changes to those settings.

(I appreciate this is all very niche!)

If there is another way of doing this that I've missed please do let me know though 😃

Source/Core/InputCommon/ControlReference/FunctionExpression.cpp
```diff
@@ -646,13 +646,13 @@ class NTimesExpression : public FunctionExpression
...
     else
-      return ExpectedArguments{"input, [times = 1]"};
+      return ExpectedArguments{"input, times = 1"};
   }
 
   ControlState GetValue() const override
   {
     const ControlState input = GetArg(0).GetValue();
-    const ControlState times = GetArg(1).GetValue();
+    const int times = (GetArgCount() >= 2) ? int(GetArg(1).GetValue()) : 1;
```
Filoppi, line 655: I think in general, when casting to int, all the other functions do something like this:
`const u32 times = GetArgCount() >= 2 ? u32(std::max(GetArg(1).GetValue() + 0.5, 0.0)) : 1;`

Filoppi (commented):
review

Filoppi:
> > Still I don't understand the point of this function.
> > If you want to change the FOV every x frames by y, why don't you just change it every frame by y/x?
> 
> That would be ideal, but I couldn't see a way to do that reliably.
> 
> I'm working on a VR interface that uses DSU, and I would like to be able to say, essentially, "when X is pressed, decrease horizontal and vertical FOV to 0.5" (and also modify the orthographic cameras similarly using the changes in @iwubcode's PR #9652).
> 
> However, if I map "decrease horizontal FOV" to `X`, then when I pass a DSU packet containing an "X is pressed" signal it keeps triggering "decrease horizontal FOV" until I pass an "X is not pressed" DSU packet, and I haven't been able to get it to trigger a precise number of "decreases" reliably by changing the time between the two DSU packets.
> 
> Adding a new function that does the work of ensuring the command is triggered exactly n times seemed like the simplest way to achieve this, and could apply to FOV, to @iwubcode's new orthographic camera configs, and to other config settings like stereoscopic depth and convergence, increase/decrease internal resolution, etc, without having to make any changes to those settings.
> 
> (I appreciate this is all very niche!)
> 
> If there is another way of doing this that I've missed please do let me know though 😃

In my PR there will be other functions to achieve this exact same result without this new functions, but there isn't a way for now.
Though the idea seems hacky in general, another of my PR has got an onPress(input) function, which is similar to toggle(), but the PR is on hold as as previously mentioned, GetState() isn't reliably called once per input frame.

I think it would be better to just live with the fact that DSU doesn't update the input every input frame (or devices refresh), given that it works with its own custom intervals. This function won't fully fix the problem anyway, as it's still frame based on random intervals which are also not reliable.

iwubcode (changes requested):
I know why you want this and I _do_ want to support that endeavor.

However, this part of the comment:

>This might not have any practical use in games that isn't already covered by other functions (especially toggle()). My interest is in using it to make precise changes to Dolphin settings. For instance, setting Decrease FOV X and Decrease FOV Y to ntimes(A, 20) lets me press a single button to change both FOV settings to exactly 0.500. I am not aware of another way to do this via a controller binding.

makes me feel like we are trying to add an input-expression for a single use case.  I just can't get behind that.  Obviously, if there are other use cases or the other dolphin devs feel this would be a good general enhancement, then it's fine.  But bloating out our input-expressions to support this scenario seems wrong to me.

I feel like we need to handle this problem in a separate way.  Hacking the DSU protocol to support exact values is also not ideal since it wasn't designed for that (in case Dolphin changes in the future).  While users may still want to be able to control exact position or rotation via the UI (so my PRs are possibly still relevant), it may be best to add a UDPController for your purposes.

MichaelJW:
OK, understood. I get it - I am totally trying to hack things to do things they weren't supposed to do, and for a single use case, no argument there. Cheers for the feedback :)

PR #9662, opened 2021-04-22
Dolphin iOS

jkcoxson:
This is a full port of Dolphin for iOS. You don’t need jailbreak anymore, which makes this a viable port for Dolphin now. It’s based off the same JIT that Android uses, to in terms of breaking things the chances are very low. The GUI is completely independent, so breaking that is not a concern. Does not modify upstream code, so maintenance is slim.


OatmealDome:
I'm really sorry, but I don't think this is a good idea. Dolphin upstream and myself agreed that it isn't best that my iOS fork is merged in #8492. (There's also a bunch of hacks in my fork which probably shouldn't be merged into upstream. A cleaner branch would need to be created.)

Perhaps it would be best if a discussion takes place on IRC or something before another attempt at iOS support is added?

JosJuice:
> The GUI is completely independent, so breaking that is not a concern. Does not modify upstream code, so maintenance is slim.

This doesn't mean that there are no maintenance concerns. For instance, I'm currently working on replacing the old janky controller system we're using on Android (see PR #9624) as it's severely limiting our ability to make improvements. I believe the iOS port uses the same system. If I were to completely replace the system while only updating the Android GUI and not the iOS GUI, the iOS port would completely break. So if we ever are to merge iOS support (I don't know if this is something that people would be fine with or not), we need to have someone who can maintain it.

jkcoxson:
As far as I can see, iOS has completely reimplemented the UI in Swift, not taking anything from upstream.

delroth:
Two things here, since I've regularly commented on iOS related stuff in the past:

- I don't think it's worth maintaining an iOS port unless we think it will see usage. If it can't be allowed on the App Store, the usage will be fringe and not worth the support. Forks are a fine solution to this.
- If we were to actually support iOS, merging back the fork is IMO the right solution, but this will have to be done in smaller individual chunks that can be reviewed and merged over time. A PR like this is unmergeable, as it is someone would probably need > 2 weeks of focused review time to give useful feedback, and you likely need more than one person looking at this (because of specialized expertise domains).

JosJuice:
> As far as I can see, iOS has completely reimplemented the UI in Swift, not taking anything from upstream.

Yes, the UI is reimplemented, but it still has the emulator core as a dependency. So any large change in the emulator core's interface might break the iOS build unless you make sure to update the iOS-specific code.

PR #9663, opened 2021-04-22
Fix IPL crash when launching MIOS-patched games

leoetlino:
MIOS (the IPL code running on the PPC, not the ARM part) puts patch data in low MEM1 (0x1800-0x3000) for its own use. Overwriting data in this range can cause the IPL to crash when launching games that get patched by MIOS. See https://bugs.dolphin-emu.org/issues/11952 for more info.

Not applying the Gecko HLE patches means that Gecko codes will not work under MIOS, but this is better than the alternative of having specific games crash.

---

This also fixes an oversight from #9545, which moved the "new game has been loaded" logic to a separate OnNewTitleLoad function that has to be called explicitly *after* a title has loaded.

Those two changes fix Wind Waker and other MIOS-patched games when they are launched from the System Menu.

### DIFF
--- a/Source/Core/Core/HLE/HLE.cpp
+++ b/Source/Core/Core/HLE/HLE.cpp
@@ -76,6 +76,14 @@ void Patch(u32 addr, std::string_view func_name)
 
 void PatchFixedFunctions()
 {
+  // MIOS puts patch data in low MEM1 (0x1800-0x3000) for its own use.
+  // Overwriting data in this range can cause the IPL to crash when launching games
+  // that get patched by MIOS. See https://bugs.dolphin-emu.org/issues/11952 for more info.
+  // Not applying the Gecko HLE patches means that Gecko codes will not work under MIOS,
+  // but this is better than the alternative of having specific games crash.
+  if (SConfig::GetInstance().m_is_mios)
+    return;
+
   // HLE jump to loader (homebrew).  Disabled when Gecko is active as it interferes with the code
   // handler
   if (!SConfig::GetInstance().bEnableCheats)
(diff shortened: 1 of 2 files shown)


Pokechu22:
On console, do Gecko codes work with MIOS-based games?  (I'm not entirely sure how Gecko codes even work on console to be honest.)

JMC47:
Gecko hooking does not work with MIOS on my Wii.

JMC47 (approved):
This is better than crashing regardless of what happens on console.  Our emulation of MIOS is limited as it is.

Pokechu22 (approved):

mimimi085181:
Just throwing in some thoughts to consider:
The list of games that are patched through MIOS is very limited:
https://wiibrew.org/wiki/MIOS#After-read_patches
and it's a fixed list, that won't change in the future. On the other hand, if gecko codes were enabled for all the other games, users would be confused why codes don't work on those specific games. Especially when asking for help and not mentioning how they boot the games.

I don't really see a use case for gecko codes and running a game from the wii menu's disc channel. If there is, then it should be considered. Maybe even as a hard option even for the patched games, because the MIOS patches can be replaced with gecko codes or other forms of patching.

As far as i know, NeoGamma is the only software that can run gecko codes on retail gamecube discs on a wii. But it doesn't even work for wii games in Dolphin, at least last time i checked. And it requires cMIOS support, which i guess is not implemented in Dolphin. cMIOS basically does 2 things, if i remember correctly, allow reading DVD-Rs(which is irrelevant for Dolphin and there should be a cMIOS variant without it), and running homebrew software, that has been loaded into memory before starting MIOS(or BC). NeoGamma comes with its own gamecube executable that handles booting the gamecube games, and that one writes the gecko code handler into memory. So, even for NeoGamma this PR should be irrelevant, because the code from the PR would be executed before booting the homebrew software from memory, if i'm not mistaken.

JMC47:
Gecko OS can also hook into GameCube games, but only if you've done a MIOS replacement I believe?  the neogamma thing makes more sense though.  I do remember that.

PR #9664, opened 2021-04-23
Fix F-Zero AX Mishandling

Zopolis4:
Removes the f-zero ax gameini, and modifies the f-zero gx one to no longer override it. Previously, F-Zero AX would be labelled as GX, as well as the banner and ID.
image


JosJuice:
The GFZ.ini file is still applied to F-Zero AX, and the one setting you removed from GFZJ8P.ini is present in GFZ.ini, so the behavior of the emulator is the exact same after this change as before this change. (Lines that start with `#` are entirely ignored by Dolphin.) I don't see the point of the change.

Zopolis4:
Oh, how would I stop the GFZ.ini file from applying to F-Zero AX then?

JosJuice:
You can't. If there is a setting that needs to be set for GFZ.ini but must not be set for GFZJ8P, then GFZJ8P.ini must explicitly set the setting to the opposite value of what GFZ.ini sets it to. But there are no such settings for this game as far as I know.

To be clear, there is nothing in GFZ.ini that affects which game ID, name or banner is shown for a particular ISO. Dolphin showing the wrong name could be explained by how TitleDatabase works, but there is no mechanism in Dolphin which causes the shown game ID to be overridden. Are you sure that what you have actually is a copy of F-Zero AX?

Zopolis4:
Checking further, it seems to be GX. Sorry.

PR #9665, opened 2021-04-24
[WIP] [NEED HELP] Add ability to convert discs with cli

beyondmeat:
This is a WIP, a PoC. It works, but might contains bugs. Only tested under linux with Iso to RVZ conversion. The underlying conversion code is still the DiscIO library. I basically took the Converter class and ripped out the UI code to make it usable from a terminal.

I don't plan on contributing further to this PR and thus I would welcome anyone to take what I started and finish it. This is enough that it works and should do all the conversions, but I only tested the RVZ path even though I did the others too.

ToDos

- Expose cli options in dolphinQT?
- detect compressed formats and decompress first before recompressing.
- Detect NKit ISO and warn that you can't just convert those to RVZ (unless more work is done to restore NKit back to 1:1 ISO). it's technically not the same thing as NKit -> ISO -> RVZ. Not sure if RVZ (NKit) is just as good as RVZ or still suffers from NKit's longer load times?

Binaries for this change:

November 2021 New build with compression options:
[Windows No GUI Build](https://github.com/dolphin-emu/dolphin/files/7544785/DolphinNoGUI-win-x64.zip)
[Ubuntu No GUI Build](https://github.com/dolphin-emu/dolphin/files/7544786/dolphin-emu-nogui-ubuntu-20.04.zip)


JosJuice:
If you won't be contributing further to this PR, then I guess I will be the one to take it over. It will probably take a little while before I get around to doing so, so if someone else would like to jump in before then and do it instead of me, just let me know. I'll leave the pull request open for the time being (even though it isn't mergeable in its current state).

beyondmeat:
I have attached a linux-x64 binary of dolphin-no-gui. Built on Ubuntu 20.04
[dolphin-emu-nogui-linux-x64.zip](https://github.com/dolphin-emu/dolphin/files/6370184/dolphin-emu-nogui-linux-x64.zip)

mjrawesome:
@beyondmeat 

Many thanks for creating and sharing your work on this. Your code seems to work quite well for me, but there is one problem... 

**There is no option to change the compression algorithm (to LZMA2 for best possible compression)**.

I don't suppose that you could add this feature? If you did, then this would be a workable solution for a lot of people with headless machines. Dolphin is a great emulator and I'm not complaining, but as far as this issue is concerned there has been no progress even though a few people requested it ([example](https://forums.dolphin-emu.org/Thread-possible-to-compress-files-to-the-rvz-format-from-command-line?highlight=rvz)).

mbc07:
> IIRC that's because it's inferior to Z-Standard, which is much faster at decompression than LZMA2, which is important for the Dolphin use-case. There's no real benefit to using LZMA2 over Z-Standard, the difference in size is negligible, while the slower decompression speed and higher CPU requirements have a much higher impact. This is one of the reasons why RVZ was made, fast decompression = fast read speeds. The dolphin team recommends using Z-Std w/ RVZ. It's just the better compression algorithm to use in this case, so I didn't bother to make it configurable.

That's not entirely accurate. ZStandard is the default because it's faster to compress/decompress and provides a good enough compression ratio. While it's true that LZMA2 is slower to compress, it can save more space than ZStandard. If you compare a single ISO, the size difference won't be big between ZStandard vs LZMA2, but when you consider converting an entire library to RVZ with LZMA2, the small differences with each ISO adds up and becomes noticeable. 

Also, as long as you keep the default block size, you will **not** notice any difference during gameplay when using RVZ with LZMA2, despite the reportedly bigger CPU requirements for decompression. It perhaps might be noticeable on Android devices or other low power devices, but on desktops, definitely not. That's from personal experience running Dolphin on an now ancient Pentium G3258 (4th gen, Haswell architecture) with just two cores and no hyperthreading. I've converted my entire library to RVZ with LZMA2 and 128 KB block size and there's no noticeable difference to RVZ with ZStandard, even when using Speed Up Disc Transfer...

beyondmeat:
Ideally this should have been a cli app. I was going to fork nkit and enhance it to do rvz, give it some much needed updates, but was discouraged from doing so. End of 2021 and nkit is still dead...

Sadly I no longer have the free time to devote to doing so. I may be able to update this PR 
with customizable compression settings by the end of the year.

JosJuice:
There is an update of NKit in progress which includes RVZ support, it's just that it's taking quite some time and there's very little public communication.

mjrawesome:
> I may be able to update this PR with customizable compression settings by the end of the year.

> There is an update of NKit in progress which includes RVZ support, it's just that it's taking quite some time and there's very little public communication.

I'd be happy to help test this on W7(64x) and, most importantly, on Ubuntu 18.04 (where I want to re-convert my entire GC/Wii library). Any cli support for LZMA2 with lvl 9 compression/decompression/recompression would be **greatly** appreciated.

If there is a separate repo for NKit where this is being worked on, I'd appreciate a link. I only found two ([1](https://github.com/Nanook/NKit), [2](https://github.com/koree/nkit-nintendont-batch)), both of which seem pretty much dead.

Source/Core/DiscIO/DiscConverter.cpp
```diff
@@ -0,0 +1,133 @@
+#include "DiscConverter.h"
```
Rumi-Larry, line 1: You are missing the legal header.

beyondmeat:
I spent way more than I liked and took more time than i thought. I was originally just going to go a build with lzma2 set. Instead I implemented the compression options.

If you used my prior build, please take note that I changed the cli option switches. The options have changed, if you based a script off my build, you will need to update it.

@mjrawesome to
Please keep in mind, I did this update out of kindness and generosity. I wasn't planning on touching this but decided to do it after the request and thinking it wasn't much work. I had to reinstall lots of stuff just to provide a build, as I don't normally dev c++. Spent an entire day and then some  to bring you these new builds. I made sure to rebase my changes to the latest upstream code.

I did not test this update, but I'm confident enough to think it will work. I'm providing the binaries for Windows and Ubuntu 20.04. I have no idea if it will run on 18.04 and tbh I rather not install that old OS. So if it don't work, upgrade your LTS, same goes for Win7. If you do run into an issue, reply here about it.

I have updated the binaries above. RVZ, Lzma2, level 9 with 128kb block size are the defaults. Only need to specify the --input, everything else should be optional to change.

If you supply an option that doesn't work with the --format or --compression, it will just be ignored.

The remove junk option is not implemented! Shouldn't use it anyway with rvz but I digress. Even though I added it as an option it does nothing.

beyondmeat:
November 2021 New build with compression options:
[Windows No GUI Build](https://github.com/dolphin-emu/dolphin/files/7544785/DolphinNoGUI-win-x64.zip)
[Ubuntu No GUI Build](https://github.com/dolphin-emu/dolphin/files/7544786/dolphin-emu-nogui-ubuntu-20.04.zip)

mjrawesome:
@beyondmeat I tested your new build on W7(64x) and so far it works great. While I've only converted a few GC/Wii games so far, the compression/decompression output for them does match Dolphin/Redump perfectly. Unfortunately, my linux server died a few days ago, so I couldn't do any tests on it yet. If I run into any issues, I'll report them here.

It seems like I'll finally be able to convert my GC/Wii library to the new format. **Thank you for spending extra time on the project to make this possible**. There are a few people that I know for sure will also want to use it, so I'll point them here.

JosJuice:
Closing in favor of PR #10252.

PR #9666, opened 2021-04-24
Jit: Optimize block link queries by using hash tables

leoetlino:
Repeated erase() + iteration on a std::multimap is extremely slow. Slow enough that it causes a 7 second long stutter during some transitions in F-Zero X (a N64 VC game that triggers many, many icache invalidations).

And slow enough that JitBaseBlockCache::DestroyBlock shows up on a flame graph as taking >50% of total CPU time on the CPU-GPU thread.

**Master**:

flame

This PR optimises those block link queries by replacing the std::multimap (which is typically implemented with red-black trees) with hash tables, which seem better suited to our repeated insert/erase/iteration workload.

**PR**: ~0.7s stutters (10x less), which is pretty close to 5.0 stable. (5.0-2021 introduced the performance regression and it is especially noticeable when branch following is disabled, which is the case for all N64 VC games since 5.0-8377.)

flame pr

---

I briefly considered using a faster implementation (e.g. Swiss tables) since this is a pretty hot path, but that "only" led to a 4-5% improvement (stutters were just 30ms shorter)...

### DIFF
--- a/Source/Core/Core/PowerPC/JitCommon/JitCache.cpp
+++ b/Source/Core/Core/PowerPC/JitCommon/JitCache.cpp
@@ -127,7 +127,7 @@ void JitBaseBlockCache::FinalizeBlock(JitBlock& block, bool block_link,
   {
     for (const auto& e : block.linkData)
     {
-      links_to.emplace(e.exitAddress, &block);
+      links_to[e.exitAddress].insert(&block);
     }
 
     LinkBlock(block);
@@ -299,13 +299,14 @@ void JitBaseBlockCache::LinkBlockExits(JitBlock& block)
 void JitBaseBlockCache::LinkBlock(JitBlock& block)
 {
   LinkBlockExits(block);
-  auto ppp = links_to.equal_range(block.effectiveAddress);
+  const auto it = links_to.find(block.effectiveAddress);
+  if (it == links_to.end())
+    return;
 
-  for (auto iter = ppp.first; iter != ppp.second; ++iter)
+  for (JitBlock* b2 : it->second)
   {
-    JitBlock& b2 = *iter->second;
-    if (block.msrBits == b2.msrBits)
-      LinkBlockExits(b2);
+    if (block.msrBits == b2->msrBits)
+      LinkBlockExits(*b2);
   }
 }
 
@@ -318,14 +319,15 @@ void JitBaseBlockCache::UnlinkBlock(const JitBlock& block)
   }
 
   // Unlink all exits of other blocks which points to this block
-  auto ppp = links_to.equal_range(block.effectiveAddress);
-  for (auto iter = ppp.first; iter != ppp.second; ++iter)
+  const auto it = links_to.find(block.effectiveAddress);
+  if (it == links_to.end())
+    return;
+  for (JitBlock* sourceBlock : it->second)
   {
-    JitBlock& sourceBlock = *iter->second;
-    if (sourceBlock.msrBits != block.msrBits)
+    if (sourceBlock->msrBits != block.msrBits)
       continue;
 
-    for (auto& e : sourceBlock.linkData)
+    for (auto& e : sourceBlock->linkData)
     {
       if (e.exitAddress == block.effectiveAddress)
       {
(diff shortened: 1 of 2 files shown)


JMC47 (approved):
Improves N64 VC.  Instant Approve.

Rumi-Larry:
> I briefly considered using a faster implementation (e.g. Swiss tables) since this is a pretty hot path, but that "only" led to a 4-5% improvement (stutters were just 30ms shorter)...

I mean if it's not difficult to implement and maintain, I would go for it anyway 🤷‍♂️

iwubcode (approved):
Seems reasonable.  Untested

leoetlino:
> I mean if it's not difficult to implement and maintain, I would go for it anyway

That would require adding an external library, which is not trivial at the moment since the Windows builder doesn't use CMake :) I'm also not sure it's really worth it.

Rumi-Larry:
> 
> 
> > I mean if it's not difficult to implement and maintain, I would go for it anyway
> 
> That would require adding an external library, which is not trivial at the moment since the Windows builder doesn't use CMake :) I'm also not sure it's really worth it.

That makes sense, hopefully CMake can be implemented to the builders soon. Imho, it is still worth it, because it would get Dolphin closer to being viable for speedrunning F-Zero X

theofficialgman:
@JosJuice 
I noticed performance regressions with master today compared to my previous builds from the start of april.

I have bisected the issues to one of the PRs merged on April 25th (not necessarily this one) with changes to the ARM64 Jit (using a nintendo switch, effectively a TX1 devkit) for testing.
Performance is somewhere around 10% lower in the games I have tested

If necessary, I can further bisect the issue to determine the specific PR where the performance degradation occured

JosJuice:
If you have found a regression and bisected it down to a specific version, please report it at https://bugs.dolphin-emu.org/projects/emulator.

Though I will say in advance, if the slowdown was caused by 5.0-14066, then it is to be expected and an issue report for it would likely be closed as "Working as intended". That change is an accuracy improvement which does cost some performance. I had hoped it would be less than 10%, though. Hopefully some upcoming PRs like fres/frsqrte can bring some of that performance back.

theofficialgman:
tomorrow I will further bisect to determine the version and will report back with the findings. I've seen your in progress/in waiting PRs and I just want to say thank you for your and others continued effort on arm64 and the jit in general

theofficialgman:
> If you have found a regression and bisected it down to a specific version, please report it at https://bugs.dolphin-emu.org/projects/emulator.
> 
> Though I will say in advance, if the slowdown was caused by 5.0-14066, then it is to be expected and an issue report for it would likely be closed as "Working as intended". That change is an accuracy improvement which does cost some performance. I had hoped it would be less than 10%, though. Hopefully some upcoming PRs like fres/frsqrte can bring some of that performance back.

you were correct.  5.0-14066 is the version where the main performance difference comes from. I tested both 14066 and the previous version 14053 to determine that. The performance drop about 13% in the wario coliseum map which I was testing in double dash. Unfortunate but I guess its good to know that the other changes which were made aren't affecting performance negatively

JMC47:
If you report an issue for that build lowering performance, the issue report will be closed.  It fixes potential hangs/crashes and other bugs in games.

As JosJuice said, sometimes being correct is a bit slower.  The Nintendo Switch has a horribly out of date processor paired with a fairly decent GPU for Dolphin.  It's one of the few devices that would actually show significant performance problems JitArm64 CPU changes.

These changes are necessary to make JitArm64 work with more games and eventually be as compatible as Jit64.

theofficialgman:
Yes I'm aware it will be closed, JosJuice already informed me of that, I was just going to report it so that it goes on documented. 

There are a few thousand people using android/linux on switch with the majority of the running dolphin, I imagine that at least on the linux side the nintendo switch actually makes up the majority of arm64 users compared to any other platform. The cortex A57 does not shine its strongest, I am aware.

I'm curious though, do you have a specific game/scenario which the jit was changed to fix? I don't think I've run into hangs or crashes in anything I've played so far.

JosJuice:
Sonic Unleashed is the game which was most obviously broken, but it also fixed problems in specific parts of Sonic Colors, Paper Mario: The Thousand-Year Door, The Legend of Zelda: Skyward Sword, and possibly more games that we don't know about.

By the way, the fres/frsqrte PR was merged just a few hours ago, so the performance should be better now by some amount.

theofficialgman:
Hmm... if its better than I can't really measure it. Master current seems to be roughly equal if not slightly worse than 5.0-14066
I guess that could be due to some of the other jitarm64 commits from last week

PR #9668, opened 2021-04-24
CMake: minor Windows fixes

leoetlino:
~~Two~~ ~~five~~ Several small commits to fix CMake on Windows builds.


shuffle2:
>LTO is supposed to be enabled by default for VS Release builds
according to the VS prop files but a build log from JMC reveals
that /GL and /LTCG are not actually passed to cl.exe/link.exe
for some reason...

It _should_ be enabled by default (the codegen really is quite better). But it's disabled (hidden behind DolphinRelease flag) because...:

> LTO also leads to *extremely* and unacceptably slow build times
when using link.exe, so let's disable it by default to actually
match the project files.

(this is why)

The idea was that we would compile with ltcg (and maybe do PGO, but that is more effort than just passing a flag) for "actual releases". But dolphin doesn't really release much, so...

shuffle2:
btw, if you want to compare args being passed to msvc tools, the easiest way is probably just something like `msbuild -v:n` (`msbuild -v:n -m -p:Platform=<ARM64|x64>,Configuration=<Release|Debug> <the vxcproj or sln>`). You can also specify `-t:Rebuild`

CMakeLists.txt
```diff
@@ -259,6 +260,10 @@ if(CMAKE_C_COMPILER_ID MATCHES "MSVC")
   )
 
   string(APPEND CMAKE_EXE_LINKER_FLAGS " /NXCOMPAT")
+  # Generate debug data
+  string(APPEND CMAKE_EXE_LINKER_FLAGS " /DEBUG:FASTLINK")
```
shuffle2, line 264: Even tho this appears in the options passed to link.exe, it is actually coming from the defaulted value. It's not explicitly set because of things like this: https://devblogs.microsoft.com/cppblog/playground-games-and-turn-10-studios-see-18-2x-and-4-95x-link-time-improvements-respectively-on-visual-studio-2019/#comment-1554 where /DEBUG:full is preferable and will probably start defaulting back to that...

i.e. just pass `/DEBUG`

leoetlino: Done

PR #9669, opened 2021-04-25
Add support for building against Qt 6

spycrab:


Source/Core/DolphinQt/CMakeLists.txt
```diff
@@ -1,15 +1,24 @@
-if (NOT Qt5_DIR AND MSVC)
+if (Qt6_DIR)
```
spycrab, line 1: For compatibility reasons since it's relatively common to use ``Qt5_DIR`` / ``Qt6_DIR`` to specify what Qt version to used (and we ourselves used the former in the past)

Source/Core/DolphinQt/CMakeLists.txt
```diff
@@ -1,15 +1,24 @@
...
+	set(Qt_DIR "${Qt6_DIR}")
+endif()
+
+if (Qt5_DIR)
```
spycrab, line 5: See above.

leoetlino (approved):
Changes LGTM and match what Qt docs recommend: https://doc-snapshots.qt.io/qt6-dev/cmake-qt5-and-qt6-compatibility.html

Source/Core/DolphinQt/CMakeLists.txt
```diff
@@ -1,15 +1,24 @@
...
+	set(Qt_DIR "${Qt5_DIR}")
+endif()
+
+if (NOT Qt_DIR AND MSVC)
   if(_M_ARM_64)
-    set(Qt5_DIR "${CMAKE_SOURCE_DIR}/Externals/Qt/Qt5.15.0/msvc2019_arm64/lib/cmake/Qt5")
+    set(Qt_DIR "${CMAKE_SOURCE_DIR}/Externals/Qt/Qt5.15.0/msvc2019_arm64/lib/cmake/Qt5")
   else()
-    set(Qt5_DIR "${CMAKE_SOURCE_DIR}/Externals/Qt/Qt5.15.0/msvc2019_64/lib/cmake/Qt5")
+    set(Qt_DIR "${CMAKE_SOURCE_DIR}/Externals/Qt/Qt5.15.0/msvc2019_64/lib/cmake/Qt5")
```
BhaaLseN, line 13: I'm guessing this stays on Qt5 for now, until the submodule is updated?

Source/Core/DolphinQt/CMakeLists.txt
```diff
@@ -1,15 +1,24 @@
...
   endif()
 endif()
 
-find_package(Qt5 5.9 REQUIRED COMPONENTS Gui Widgets)
+find_package(Qt NAMES Qt6 Qt5 COMPONENTS Core Gui Widgets REQUIRED)
+find_package(Qt${Qt_VERSION_MAJOR} 5.9 COMPONENTS Core Gui Widgets REQUIRED)
```
BhaaLseN, line 18: Are those nested identifiers "the way to go" or can you just alias `Qt5`/`Qt6` to `Qt` (just like we do with `Qt_DIR`) and use things like `Qt::Core`, `Qt::Widgets` etc. instead?

spycrab: That would be cool, sadly
> We then do load the package for the determined Qt version again by creating the name Qt${QT_VERSION_MAJOR} on the fly. This is needed because CMAKE_AUTOMOC expects the package name to be either Qt5 or Qt6, and will print an error otherwise.

https://doc-snapshots.qt.io/qt6-dev/cmake-qt5-and-qt6-compatibility.html#supporting-older-qt-5-versions

I assume this extends to using ``Qt`` instead of ``Qt${Qt_VERSION_NUMBER}`` since both the Qt docs and Qt creator both use the ``${Qt_VERSION_NUMBER}`` approaches. It's probably best to not stray from that path or that might risk some breakages, even if it works right now.

Versionless Qt modules [will be possible](https://doc-snapshots.qt.io/qt6-dev/cmake-qt5-and-qt6-compatibility.html#versionless-targets) once we restrict support to Qt 5.15+ though.

Source/Core/DolphinQt/CMakeLists.txt
spycrab: Correct. From reading some release notes, I don't see any need to upgrade very soon unless something breaks in a new macOS release to be honest.

spycrab:
Still needs some adaption for macOS to work properly.

StLouisCPhT:
> Still needs some adaption for macOS to work properly.

It looks like it shouldn't break Skyler's UB work either if adapted correctly.
I'm able to build melonDS for ARM macOS with qt6.

shuffle2:
Is there some reason to keep Qt5 support after updating? It would be nicer to raise the minimum version uniformly on all platforms at the same time.

Techjar:
> Is there some reason to keep Qt5 support after updating? It would be nicer to raise the minimum version uniformly on all platforms at the same time.

Not really an option at this time, as quite a few Linux distros have yet to even introduce a Qt 6 package, most notably Ubuntu.

shuffle2:
It seems Ubuntu still does not package any Qt 6.x version even for Ubuntu 21.10 and 22.04. Is this actually some licensing issue, and Ubuntu will never progress past qt5? If so, we should probably setup a workaround and get on with it...

shuffle2:
Looks like Qt6 may make it into ubuntu jammy? https://launchpad.net/ubuntu/+source/qt6-base 👀 which is scheduled for april 21st: https://discourse.ubuntu.com/t/jammy-jellyfish-release-schedule/23906

shuffle2:
fwiw I was able to build dolphin on windows against Qt 6.2.4 with this commit and a small change (which should probably be incorporated into this PR):
```diff
diff --git a/Source/Core/DolphinQt/GBAWidget.cpp b/Source/Core/DolphinQt/GBAWidget.cpp
index db05c04367..90aa309315 100644
--- a/Source/Core/DolphinQt/GBAWidget.cpp
+++ b/Source/Core/DolphinQt/GBAWidget.cpp
@@ -510,7 +510,7 @@ void GBAWidget::mouseMoveEvent(QMouseEvent* event)
 {
   if (!m_moving)
     return;
-  move(event->globalPos() - m_move_pos - (geometry().topLeft() - pos()));
+  move(event->pos() - m_move_pos - (geometry().topLeft() - pos()));
 }

 void GBAWidget::paintEvent(QPaintEvent* event)
```
I'm not sure if that's the correct fix.

Source/Core/DolphinQt/DolphinEmu.entitlements
```diff
@@ -13,5 +13,7 @@
 	<!-- This is needed to use adhoc signed linked libraries -->
  	<key>com.apple.security.cs.disable-library-validation</key> 
   	<true/>
+    <key>com.apple.security.get-task-allow</key>
```
OatmealDome, line 16: Is there a reason why this is here? We should be avoiding this entitlement if possible. (Plus, the notary service may reject the app if this is present.)

shuffle2: @OatmealDome should we drop this change when merging Qt6?
i'm not familiar with xcode and such, but from googling it looks like it was probably just added by mistake?

OatmealDome: @shuffle2 It should be dropped, yeah. If we need it, something is probably very wrong. (macOS requires processes to have this entitlement to allow a debugger to be attached to the process, so I guess this was added for testing.)

AdmiralCurtiss:
I think we can close this?

shuffle2:
yes

PR #9670, opened 2021-04-25
Core / VideoCommon: refactor FreeLookCamera

iwubcode:
This just refactors the `FreeLookCamera` so things that it previously managed (dirty state, field of view, etc) come from the controller.  Also the controller has been split into the base controller and a new controller called `CameraControllerInput` which all controllers derive from.  This will be used in a future PR.

### DIFF
--- a/Source/Core/Core/FreeLookManager.cpp
+++ b/Source/Core/Core/FreeLookManager.cpp
@@ -197,6 +197,15 @@ void FreeLookController::Update()
   if (!g_freelook_camera.IsActive())
     return;
 
+  auto* camera_controller = g_freelook_camera.GetController();
+  if (camera_controller->SupportsInput())
+  {
+    UpdateInput();
+  }
+}
+
+void FreeLookController::UpdateInput()
+{
   const auto lock = GetStateLock();
 
   float dt = 1.0;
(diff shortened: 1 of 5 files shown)


Source/Core/Core/FreeLookManager.cpp
```diff
@@ -220,48 +229,50 @@ void FreeLookController::Update()
   const auto gyro_motion_quat =
       Common::Quaternion::RotateXYZ(gyro_motion_rad_velocity_converted * dt);
 
-  g_freelook_camera.Rotate(gyro_motion_quat);
+  auto* camera_controller = static_cast<CameraControllerInput*>(g_freelook_camera.GetController());
```
Pokechu22, line 232: I feel it might be better to have `UpdateInput` take a `CameraControllerInput*` as a parameter; that way, the cast would be in the same place as the `camera_controller->SupportsInput()` check.

Source/Core/VideoCommon/FreeLookCamera.cpp
```diff
@@ -194,123 +206,101 @@ class OrbitalController final : public CameraController
...
+  p.Do(m_fov_x);
+  p.Do(m_fov_y);
 }
 
-Common::Vec2 FreeLookCamera::GetFieldOfView() const
+void CameraControllerInput::IncreaseFovX(float fov)
 {
-  return Common::Vec2{m_fov_x, m_fov_y};
+  m_fov_x += fov;
+  m_fov_x = std::clamp(m_fov_x, m_min_fov_multiplier, m_fov_x);
```
Pokechu22, line 224: I don't understand why `std::clamp` is being used here.  `clamp` is basically:

```C++
if (m_fov_x < m_min_fov_multiplier)
    return m_min_fov_multiplier;
else if (m_fov_x > m_fov_x)
    return m_fov_x;
else
    return m_fov_x;
```

Wouldn't `std::max(m_fov_x, m_min_fov_multiplier)` be better?  The same applies for `m_fov_y` and `m_speed`.  (I do see that this code existed before this change, but it's still odd).

Source/Core/VideoCommon/FreeLookCamera.h
```diff
@@ -36,50 +57,41 @@ class CameraController
...
+
+  void ModifySpeed(float multiplier);
+  void ResetSpeed();
+  float GetSpeed() const;
+
+private:
+  float m_fov_x = 1.0f;
+  float m_fov_y = 1.0f;
+
+  float m_min_fov_multiplier = 0.025f;
```
Pokechu22, line 72: It seems odd to me that this isn't a constant; I'd expect `static constexpr float MIN_FOV_MULTIPLIER = 0.025f` instead.  (It also seems odd that this one is specified as a multiplier while the others aren't.)

Source/Core/VideoCommon/FreeLookCamera.cpp
```diff
@@ -200,34 +200,35 @@ class OrbitalController final : public CameraControllerInput
...
 private:
-  float m_distance = 0;
+  static constexpr float MIN_DISTANCE = 0.0f;
+  float m_distance = MIN_DISTANCE;
   Common::Vec3 m_rotation = Common::Vec3{};
   Common::Quaternion m_rotate_quat = Common::Quaternion::Identity();
 };
 }  // namespace
 
 Common::Vec2 CameraControllerInput::GetFieldOfView() const
```
Pokechu22, line 210: Perhaps this function should also be named `GetFieldOfViewMultiplier` for consistency.

Source/Core/VideoCommon/FreeLookCamera.h
```diff
@@ -36,50 +57,43 @@ class CameraController
...
-  void ModifySpeed(float multiplier);
-  void ResetSpeed();
-  float GetSpeed() const;
-
   bool IsActive() const;
 
+  CameraController* GetController() const;
+
 private:
   bool m_dirty = false;
```
Pokechu22, line 94: This field is now unused (a warning is generated on [osx](https://dolphin.ci/#/builders/3/builds/3721) and [bsd](https://dolphin.ci/#/builders/13/builds/3731) builders); it should be removed.

Source/Core/VideoCommon/FreeLookCamera.cpp
```diff
@@ -319,10 +310,7 @@ void FreeLookCamera::DoState(PointerWrap& p)
   else
   {
     const auto old_type = m_current_type;
```
Pokechu22, line 312: There's also a warning (on the `const std::string old_type_name = old_type ? to_string(*old_type) : ""`) on the [ubuntu builder](https://dolphin.ci/#/builders/19/builds/3724) that `old_type` may be uninitialized.  This [might be a GCC bug](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=80635), but it's odd.  (It also isn't new to this PR).  Possibly switching to an `EnumFormatter` here would work, though I'm not sure.  It's probably fine to just ignore this.

Source/Core/VideoCommon/FreeLookCamera.h
iwubcode: Whoops.  Thanks for catching this!

Pokechu22 (approved):

PR #9671, opened 2021-04-25
FreeLook: allow controlling exact position, rotation, field-of-view multiplier by UDP

iwubcode:
This PR allows Free Look to be controlled by UDP packets using a newly defined UDP spec that is supported by Dolphin.  Unlike the input controllers which deal with incremental values, this controller takes absolute data over a socket.

With this, you can write any script or tool to control Free Look externally!!  I will be including a formal document describing the spec for interested users.

Here's an example of OpenTrack using [OpenTrackFreeLook](https://github.com/iwubcode/OpenTrackFreeLook):

freelook-opentrack

Controlling via a scripting language is also pretty easy, here's an example of moving the Free Look camera to position 30.0 along the x axis and rotating its pitch to 20°:

```
import socket
import struct

ip = "127.0.0.1"
port = 36760

print(f"UDP target IP: {ip}")
print(f"UDP target port: {port}")
euler_pos_struct = struct.pack("<HBIBdddddd", 1, 1, 48, 1, 0.0, 20.0, 0.0, 30.0, 0.0, 0.0)

sock = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
sock.sendto(bytes(euler_pos_struct), (ip, port))
```

This should allow Free Look to be hooked up to the many inputs that [FreePie](https://andersmalmgren.github.io/FreePIE/) supports.

Want to do something crazy?  You wouldn't be the first.  Maybe you revive the script to [control Free Look via Blender](https://github.com/John10v10/-Useless-DolphinToolForBlender)!

### DIFF
--- a/Source/Core/DolphinQt/Config/FreeLookWidget.cpp
+++ b/Source/Core/DolphinQt/Config/FreeLookWidget.cpp
@@ -12,7 +12,9 @@
 #include "Core/Config/FreeLookSettings.h"
 #include "Core/ConfigManager.h"
 #include "Core/Core.h"
+#include "Core/FreeLookConfig.h"
 
+#include "DolphinQt/Config/FreeLookOptionsWindow.h"
 #include "DolphinQt/Config/Graphics/GraphicsChoice.h"
 #include "DolphinQt/Config/Mapping/MappingWindow.h"
 #include "DolphinQt/Config/ToolTipControls/ToolTipCheckBox.h"
@@ -35,9 +37,10 @@ void FreeLookWidget::CreateLayout()
       tr("Allows manipulation of the in-game camera.<br><br><dolphin_emphasis>If unsure, "
          "leave this unchecked.</dolphin_emphasis>"));
   m_freelook_controller_configure_button = new QPushButton(tr("Configure Controller"));
+  m_freelook_options_configure_button = new QPushButton(tr("Configure Options"));
 
-  m_freelook_control_type = new GraphicsChoice({tr("Six Axis"), tr("First Person"), tr("Orbital")},
-                                               Config::FL1_CONTROL_TYPE);
+  m_freelook_control_type = new GraphicsChoice(
+      {tr("Six Axis"), tr("First Person"), tr("Orbital"), tr("UDP")}, Config::FL1_CONTROL_TYPE);
   m_freelook_control_type->SetTitle(tr("Free Look Control Type"));
   m_freelook_control_type->SetDescription(tr(
       "Changes the in-game camera type during Free Look.<br><br>"
@@ -47,7 +50,9 @@ void FreeLookWidget::CreateLayout()
       "First Person: Controls the free camera similarly to a first person video game. The camera "
       "can rotate and travel, but roll is impossible. Easy to use, but limiting.<br><br>"
       "Orbital: Rotates the free camera around the original camera. Has no lateral movement, only "
-      "rotation and you may zoom up to the camera's origin point."));
+      "rotation and you may zoom up to the camera's origin point."
+      "<br><br>"
+      "UDP: Receives camera movement and rotation from UDP packets. "));
 
   auto* description =
       new QLabel(tr("Free Look allows for manipulation of the in-game camera. "
@@ -64,6 +69,7 @@ void FreeLookWidget::CreateLayout()
   hlayout->addWidget(new QLabel(tr("Camera 1")));
   hlayout->addWidget(m_freelook_control_type);
   hlayout->addWidget(m_freelook_controller_configure_button);
+  hlayout->addWidget(m_freelook_options_configure_button);
 
   layout->addWidget(m_enable_freelook);
   layout->addLayout(hlayout);
@@ -76,6 +82,8 @@ void FreeLookWidget::ConnectWidgets()
 {
   connect(m_freelook_controller_configure_button, &QPushButton::clicked, this,
           &FreeLookWidget::OnFreeLookControllerConfigured);
+  connect(m_freelook_options_configure_button, &QPushButton::clicked, this,
+          &FreeLookWidget::OnFreeLookOptionsConfigured);
   connect(m_enable_freelook, &QCheckBox::clicked, this, &FreeLookWidget::SaveSettings);
   connect(&Settings::Instance(), &Settings::ConfigChanged, this, [this] {
     const QSignalBlocker blocker(this);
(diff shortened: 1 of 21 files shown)


iwubcode:
Marked as draft for now.  It's working but I do want to rework the spec (and define it in a document) a little bit.  And I need to add the UDP server dialog.

Posting for @MichaelJW to see!

Source/Core/VideoCommon/FreeLookUDPServer.cpp
```diff
@@ -0,0 +1,113 @@
...
+void FreeLookUDPServer::ProcessReceivedData(const u8* data, std::size_t packet_size)
+{
+  const std::size_t header_size = sizeof(PacketHeader);
+  if (packet_size < header_size)
+    return;
+
+  PacketHeader header;
+  memcpy(&header, data, header_size);
+
+  if (header.version != FREE_LOOK_PROTO_VERSION)
```
Filoppi, line 51: If I recall, standard UDP packages already come with a version number, which I guess is 0 or 1, so why don't you officially propose this new version of the standard and implement all of this as a switch based on package version in the InputCommon UDP server?

Filoppi (commented):
Quick look. Seems cool and well coded.

jordan-woyak:
@iwubcode Instead of adding another "UDP Protocol" (one is already unfortunate enough) maybe we could define some extensions to the DSU Protocol?

iwubcode:
@jordan-woyak - yeah.  That was sorta my original thought process.  And it was for that reason that I hacked around DSU by mistreating its fields as exact position/rotation and had those previous PRs.  However, in the end I wasn't a fan of going that way as it  has a number of short comings.

1) If you actually played around with my previous PRs that provided exact position/rotation, the UI was very klunky.  This type of data doesn't really fit well into our input system.  It doesn't make sense to use an exact value since the input binding is one directional.  The way I worked around this was to have it be an offset.  That works but could be confusing to the user and the input system has a flaw in that the input binding is dependent on what will be bound to it (it'd be nice if we had a conversion system).  Finally, the user has control over the final result, not the external software (I guess this could be seen as a pro or a con).

2) Free Look is more than just position and rotation.  At the moment, there's only field of view but the new orthographic stretch is another feature that if merged would be specific to Dolphin.  These changes shouldn't go in the DSU spec because the DSU is meant to be generic.  If we start adding emulator specific stuff, then other emulators could follow.  I don't think that's a good place to be.

3) At the moment this Free Look spec allows sending a quaternion + position, a matrix, or euler angles + position.  This allows us to handle almost any use case.  While we could add that to the DSU spec, these are all very specific to Dolphin and not sure it makes sense for the spec to be bloated just for our use cases.

4) We don't own the DSU spec.  We can add extensions but if someone else adds a different extension that doesn't build on ours and we want that extension, then we have a mess on our hands.


As an aside, I don't really understand your comment about it being "unfortunate".  Outside of shared memory or a centralized system (like a database), there aren't many ways to share information across programs (at least that I'm aware of).  I think UDP is a reasonable choice for that.

I'll just add, I'm not saying I'm opposed to DSU spec extensions.  I do think there are some things that a DSU extension would make sense for, ex: sending down button names but I'm not sure it's as great of a fit here.

If you still think a DSU spec enhancement makes the most sense, please let me know how we would handle the above situations.  I don't necessarily think there's anything wrong with having a second udp specification.

Filoppi:
> 2. Free Look is more than just position and rotation.  At the moment, there's only field of view but the new orthographic stretch is another feature that if merged would be specific to Dolphin.  These changes shouldn't go in the DSU spec because the DSU is meant to be generic.  If we start adding emulator specific stuff, then other emulators could follow.  I don't think that's a good place to be.
> 3. At the moment this Free Look spec allows sending a quaternion + position, a matrix, or euler angles + position.  This allows us to handle almost any use case.  While we could add that to the DSU spec, these are all very specific to Dolphin and not sure it makes sense for the spec to be bloated just for our use cases.

Why not return the list of returned inputs by the UDP device when first queried for availability?
You could add all the angular and positional data, together with their types and input names to a list, essentially achieving a "dynamic" extensions to UDP. Sure, it's still more clunky to set up by the end user, but you could make a pre-filled default profile/ini. Does that make sense?

iwubcode:
@Filoppi - I'm sorry, I'm not sure I understand.

What I think you're suggesting is having the DSU send a message on startup that tells Dolphin what kind of data to accept?  When sending messages, the DSU sends a generic blob of some arbitrary amount of bytes and Dolphin then reads that and converts it an appropriate input dynamically using some hardcoded spec.  Or possibly pulls it in via the ini in a similar fashion somehow?

If that is what you are suggesting, that is technically feasible, but I'm not sure how I feel about that.  If you can get arbitrary data from any source, then both client and server have to know what the other produces.  It's basically a spec without the formality of it.

Filoppi:
@iwubcode
When you query for the available UDP devices to the server, it returns some data about them. We could add a new field to that data, describing additional custom data returned by every UDP device. It could just be a name of the input and the type/size of the data, or it could be a pre-defined set of data (like angular data, positional data, ...). Then based on what info the device gave when first queried, we decode the actual message with input they send based on that. Users would then have to manually map them in the device profile (which I know is not as straight forward as your implementation).
Then we could also implement some other output data, like rumble and such.

iwubcode:
Thanks for clarifying @Filoppi !  That was kinda what I thought.

The issue with that is either DSU or emulator could support custom fields.  If DSU provided something, emulators would have to follow or the data would be ignored.  If emulators wanted something, DSUs would have to follow.  Each DSU would have to look at all emulators to see what was supported.  Each emulator would have to look at all DSUs to see what to support.  That sounds like chaos.

Doing a DSU extension update to formally support new messages or data doesn't necessarily completely buffer against that (there could still be multiple versions of the spec in flight) but it's a little more formal (it is versioned) and both clients/servers know exactly what to expect.  It can be hosted in one place ([like it is now](https://v1993.github.io/cemuhook-protocol/)), so anyone can easily reference it.

Source/Core/VideoCommon/FreeLookUDPServer.cpp
Filoppi: What if someone customized the UDP protocol in the future and used the same version number as you did here? Which then could also have the same packet size.

iwubcode: On a general level, what you're describing is something that is discussed when managing any sort of wire transfer.  How much validation can we do?  It's usually a trade off of speed vs ensuring correctness.  At some point though, you have to take whatever you're given and assume it is correct.

If the protocol version is an expected value and the packet size / message type are appropriate, we assume the data is appropriate.  You're right though, we could be given a character array of the same size or 2 floats instead of every double, etc.

I don't know how important it is to do more validation than message type/size and version check.

Source/Core/DolphinQt/Config/FreeLookWidget.cpp
```diff
@@ -89,10 +92,17 @@ void FreeLookWidget::OnFreeLookControllerConfigured()
   if (m_freelook_controller_configure_button != QObject::sender())
     return;
   const int index = 0;
-  MappingWindow* window = new MappingWindow(this, MappingWindow::Type::MAPPING_FREELOOK, index);
-  window->setAttribute(Qt::WA_DeleteOnClose, true);
-  window->setWindowModality(Qt::WindowModality::WindowModal);
-  window->show();
+  QDialog* window;
+  if (Config::Get(Config::FL1_CONTROL_TYPE) == FreeLook::ControlType::UDP)
+  {
```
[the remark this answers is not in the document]
iwubcode: This is still in draft state.  Eventually this will trigger a dialog that has the UDP connection information.  Right now I have it hard coded.

Source/Core/VideoCommon/FreeLookUDPServer.cpp
Filoppi: It's fine as it is. My second comment didn't make much sense.

iwubcode:
Code wise this is now ready for review

iwubcode:
No major changes, just updated the copyright date of some files and rebased on latest master.

iwubcode:
There didn't seem to be much appetite for this and I've since moved on.  Maybe one day we can revisit the VR and Flight-stick dream.

PR #9673, opened 2021-04-26
Implement EFB Peeks for compressed z16 formats

phire:
Dolphin doesn't correctly emulate 3xMSAA mode. It probably never will as it's kind of pointless when users can just force everything to have 4xMSAA or 16xMSAA or even stupidly high Internal Resolutions.

Besides, Not many games uses the 3xMSAA mode. It's a real pain to use when it cuts the vertical EFB resolution in half to 640x264, requiring multiple passes to get a complete 4:3 frame. 

But Rogue Squadron uses it. Actually, Factor 5 cheap out and only uses it during cutscenes. They cheap out even further by rendering many of those cutscenes in a "cinematic widescreen" aspect ratio with a resolution of 512x264 that just so happens to fit inside a single 3xMSAA EFB buffer. 

Dolphin's trickery is usually hidden from games. Except RS3 also directly accesses the EFB framebuffer to check the occlusion of engine exhaust plumes. This was previously working in main gameplay, but in cutscenes dolphin would return the wrong depth and the occlusion query would always pass, resulting in the lens flare effect shining though solid objects. 

Dolphin had some previous code to automatically convert from Z24 to Z16 formats during an efb peek, but it was hardcoded to the Linear format, while RS3 uses the ZFAR compressed Z16 format.

This PR implements the compressed Z16 formats, which fixes this visual glitch in RS3 cutscenes. 

# Screenshots:

### Before
hanger-before
### After
hanger-after

### Before

GLRE64_2021-04-26_22-30-42
### After
GLRE64_2021-04-26_22-21-44

### DIFF
--- a/Source/Core/VideoCommon/VideoCommon.h
+++ b/Source/Core/VideoCommon/VideoCommon.h
@@ -4,7 +4,12 @@
 
 #pragma once
 
+#include <algorithm>
+
+#include "Common/BitUtils.h"
 #include "Common/CommonTypes.h"
+
+#include "VideoCommon/BPMemory.h"
 
 // These are accurate (disregarding AA modes).
 constexpr u32 EFB_WIDTH = 640;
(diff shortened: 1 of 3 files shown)


Rumi-Larry:
Perhaps the second example should be included into Fifoci

JMC47:
I cannot be, for the results get baked into the fifolog.  We already checked for that.

JMC47:
This seems to only affect Rogue Squadron 3.  I pushed this through a big part of my library of games that do EFB Access, and it just doesn't seem to catch anything at all.  No fixes, no regressions, no nothing.  Just Rogue Squadron 3.

PatrickFerry:
Fixes https://bugs.dolphin-emu.org/issues/8859

stenzek (approved):
I'm unfamiliar with how the hardware works here, but the in-source explanation makes sense to me.

phire:
To be clear, I didn't do hardware testing here. 

The data format was documented from REed Z16 decompression/compression functions decompiled from game code (that appear to be originally from the Gamecube SDK)

I don't know if it's 100% correct, but It's close enough to make RE3 behave correctly.

Source/Core/VideoCommon/VideoCommon.h
```diff
@@ -59,3 +64,71 @@ inline u32 Z24ToZ16ToZ24(u32 src)
 {
   return (src & 0xFFFF00) | (src >> 16);
 }
+
+inline u32 CompressZ16(u32 Z24depth, DepthFormat format)
+{
+  // Flipper offers a number of choices for 16bit Z formats that adjust
+  // where the bulk of the percision lies.
```
leoetlino, line 71:
```suggestion
  // where the bulk of the precision lies.
```

Source/Core/VideoCommon/VideoCommon.h
```diff
@@ -59,3 +64,71 @@ inline u32 Z24ToZ16ToZ24(u32 src)
...
+  {
+    // This is just a linear depth buffer with 16 bits of percision
+    return Z24depth >> 8;
+  }
+
+  // ZNEAR/ZMID/ZFAR are custom floating point formats with 2/3/4 bits of exponent
+  // The exponent is simply the number of leadings ones that have been removed
+  // The first zero bit is skipped and not stored. The mantissa contains the next 14/13/12 bits
+  // If exponent is at the MAX (3, 7, or 12) then the next bit might still be a one, and can't
+  // be skipped, so the mantissa simply containes the next 14/13/12 bits
```
leoetlino, line 83:
```suggestion
  // be skipped, so the mantissa simply contains the next 14/13/12 bits
```

Source/Core/VideoCommon/VideoCommon.h
```diff
@@ -59,3 +64,71 @@ inline u32 Z24ToZ16ToZ24(u32 src)
...
+
+  unsigned leadingOnes = Common::CountLeadingZeros((~Z24depth) << 8);
```
leoetlino, line 85: u32 leading_ones

Source/Core/VideoCommon/VideoCommon.h
```diff
@@ -59,3 +64,71 @@ inline u32 Z24ToZ16ToZ24(u32 src)
...
+  bool nextBitIsOne = false;  // AKA: Did we clamp leadingsOnes?
```
leoetlino, line 86: next_bit_is_one, etc.

PatrickFerry:
There is a TODO for checking for compressed z16 format changes in BPFunctions.cpp -> OnPixelFormatChange()

I am assuming "special compression format" means ZFAR, if that's the case could you add Star Wars:RS3 to the comment?

phire:
> There is a TODO for checking for compressed z16 format changes in BPFunctions.cpp -> OnPixelFormatChange()
> 
> I am assuming "special compression format" means ZFAR, if that's the case could you add Star Wars:RS3 to the comment?

ZFAR, ZMID and ZNEAR are the 3 options for "special compression format", on top of ZLINEAR. 

I've updated the TODO to be more accurate and have more up-to-date infomation.

leoetlino (approved):
LGTM, the first three commits should probably be squashed together

phire:
Squashed, ready for merge

PR #9677, opened 2021-04-27
DolphinQt: fix input profile drop down not resizing

iwubcode:
This fixes the input profile drop down not resizing.  This regression was due to #8867 where I made that change because I was trying to fix the device drop down from disappearing with a small window (in order to do that I changed the size policy from `Ignored` to `Expanding`).  However, this  made the device drop down take up all the space!  So now I've tried changing the size policy from `Expanding` to `Minimum`.

Please test!


Miksel12:
Fixes the issue and it doesn't dissapear!

PR #9678, opened 2021-04-28
Make all Qt ToolTips look similar to BalloonTips

Filoppi:
Dolphin still has a lot of tooltips which haven't been converted to BalloonTips, which is fine, and some are way harder to convert as there is no code for them yet (e.g. ComboBox per item ToolTips).

With this simple addition to the qApp style sheet (mostly using existing code), **ToolTips**, and only ToolTips, **will visualize themselves with the exact same colors and padding size of BalloonTips**. Plus, now parts of their code is shared, so it can safely be changed in one place to reflect to both.

Tested. Works with high and not high contrast, or any theme brightness. Also works on top of any other style sheet (the ToolTip style is only overridden if not already specified).
There should be no possible regression from this, and BalloonTips will still look exactly as before (and they will not follow the user custom style sheet, just like before, which I think is a bit weird but it's fine).

This **also fixes a bug** (?) with Qt 5.15.0 on Windows 10 where calling SetCurrentUserStyle(...) before the MainWindow has been created, with a style sheet string different from null/emtpy (even an empty space is enough), causes some widgets to appear with a different style (e.g. the log widget background changes from black to white).

Notes:
-Making parts of the text (emphasis) take another color is not possible on default ToolTips of course
-I've tried to round the borders of default ToolTips but the background is drawn in a squared fashion behind the borders.

High contrast standard **Qt ToolTip**:
image

High contrast Dolphin **BalloonTip**:
image

### DIFF
--- a/Source/Core/DolphinQt/Settings.cpp
+++ b/Source/Core/DolphinQt/Settings.cpp
@@ -10,6 +10,7 @@
 #include <QFileInfo>
 #include <QFontDatabase>
 #include <QSize>
+#include <QWidget>
 
 #include "AudioCommon/AudioCommon.h"
 
@@ -45,8 +46,6 @@ Settings::Settings()
 
   g_controller_interface.RegisterDevicesChangedCallback(
       [this] { QueueOnObject(this, [this] { emit DevicesChanged(); }); });
-
-  SetCurrentUserStyle(GetCurrentUserStyle());
 }
 
 Settings::~Settings() = default;
@@ -80,10 +79,13 @@ QString Settings::GetCurrentUserStyle() const
   return QFileInfo(GetQSettings().value(QStringLiteral("userstyle/path")).toString()).fileName();
 }
 
+// Calling this before the main window has been created breaks the style of some widgets on
+// Windows 10/Qt 5.15.0. But only if we set a stylesheet that isn't an empty string.
 void Settings::SetCurrentUserStyle(const QString& stylesheet_name)
 {
   QString stylesheet_contents;
 
+  // If we haven't found one, we continue with an empty (default) style
   if (!stylesheet_name.isEmpty() && AreUserStylesEnabled())
   {
     // Load custom user stylesheet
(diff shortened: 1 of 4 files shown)


Source/Core/DolphinQt/Settings.cpp
```diff
@@ -94,6 +96,26 @@ void Settings::SetCurrentUserStyle(const QString& stylesheet_name)
...
+    GetToolTipStyle(window_color, text_color, unused_text_emphasis_color, border_color, palette,
+                    palette);
+
+    const auto tooltip_stylesheet =
+        QStringLiteral("QToolTip { background-color: #%1; color: #%2; padding: 8px; "
+            "border: 1px; border-style: solid; border-color: #%3; }")
+            .arg(window_color.rgba(), 0, 16)
+            .arg(text_color.rgba(), 0, 16)
+            .arg(border_color.rgba(), 0, 16);
+    stylesheet_contents.append(QStringLiteral("%1").arg(tooltip_stylesheet));
```
Filoppi, line 116: Appending a space here is not needed, if someone was wondering.

Source/Core/DolphinQt/Settings.h
```diff
@@ -54,6 +54,10 @@ class Settings final : public QObject
   void SetUserStylesEnabled(bool enabled);
   bool AreUserStylesEnabled() const;
 
+  void GetToolTipStyle(QColor& window_color, QColor& text_color, QColor& emphasis_text_color,
```
Filoppi, line 57: This could be have static for now but in the future it might depend on other settings.

Filoppi (commented):
Notes

Source/Core/DolphinQt/Settings.cpp
```diff
@@ -109,6 +131,32 @@ void Settings::SetUserStylesEnabled(bool enabled)
...
+  const auto theme_window_color = palette.color(QPalette::Base);
+  const auto theme_window_hsv = theme_window_color.toHsv();
+  const auto brightness = theme_window_hsv.value();
+  const bool brightness_over_threshold = brightness > 128;
+  const QColor emphasis_text_color_1 = Qt::yellow;
+  const QColor emphasis_text_color_2 = QColor(QStringLiteral("#0090ff")); // ~light blue
+  if (Config::Get(Config::MAIN_USE_HIGH_CONTRAST_TOOLTIPS))
+  {
+    window_color = brightness_over_threshold ? QColor(72, 72, 72) : Qt::white;
+    text_color = brightness_over_threshold ? Qt ::white : Qt::black;
```
BhaaLseN, line 147: Any reason this has a space?
```suggestion
    text_color = brightness_over_threshold ? Qt::white : Qt::black;
```

Filoppi: Thanks. Fixed

MayImilae:
I'd like for all tooltips to move to the modern balloon tooltips eventually, but this seems like a good temporary solution to make the two tooltip types coexist a bit better.

JMC47:
Seems like people are generally in favor of this.  Has been reviewed and May says it's an improvement.  Tested myself in the default GUI style and it worked fine.

PR #9679, opened 2021-04-28
DolphinQt: Disable verify button when emulation is running

JosJuice:
Verifying a Wii game creates an instance of IOS, and Dolphin can't handle more than one instance of IOS at the same time. Properly supporting it is probably more effort than it's worth.

Fixes https://bugs.dolphin-emu.org/issues/12494.


iwubcode (approved):
LGTM.  Untested

Miksel12:
Maybe a popup when hovering over the greyed out button telling why it is disabled would be nice, though probably not that important.

leoetlino (approved):

PR #9681, opened 2021-05-01
VideoCommon: move all texture calculations to a "TextureInfo" class

iwubcode:
More stuff can be moved over if desired later, I tried to stick to the basic information needed to calculate the texture id.  This cleans up the code ever so slightly, the current TextureCacheBase code is hard to follow because it changes some of the parameters it passes in but not all.  This centralizes the code and makes its members const, hopefully making it easier to read.

Additionally, and more importantly, texture data is currently loaded and bound in [VertexManagerBase](https://github.com/dolphin-emu/dolphin/blob/master/Source/Core/VideoCommon/VertexManagerBase.cpp#L481).  By providing this class, the texture data can be leveraged in other areas of the code base, allowing us to provide new enhancements.

### DIFF
--- a/Source/Core/VideoCommon/TextureCacheBase.cpp
+++ b/Source/Core/VideoCommon/TextureCacheBase.cpp
@@ -261,7 +261,7 @@ void TextureCacheBase::SetBackupConfig(const VideoConfig& config)
 }
 
 TextureCacheBase::TCacheEntry*
-TextureCacheBase::ApplyPaletteToEntry(TCacheEntry* entry, u8* palette, TLUTFormat tlutfmt)
+TextureCacheBase::ApplyPaletteToEntry(TCacheEntry* entry, const u8* palette, TLUTFormat tlutfmt)
 {
   DEBUG_ASSERT(g_ActiveConfig.backend_info.bSupportsPaletteConversion);
 
@@ -739,7 +739,7 @@ void TextureCacheBase::TCacheEntry::DoState(PointerWrap& p)
 }
 
 TextureCacheBase::TCacheEntry*
-TextureCacheBase::DoPartialTextureUpdates(TCacheEntry* entry_to_update, u8* palette,
+TextureCacheBase::DoPartialTextureUpdates(TCacheEntry* entry_to_update, const u8* palette,
                                           TLUTFormat tlutfmt)
 {
   // If the flag may_have_overlapping_textures is cleared, there are no overlapping EFB copies,
@@ -946,11 +946,6 @@ void TextureCacheBase::DumpTexture(TCacheEntry* entry, std::string basename, uns
     return;
 
   entry->texture->Save(filename, level);
-}
-
-static u32 CalculateLevelSize(u32 level_0_size, u32 level)
-{
-  return std::max(level_0_size >> level, 1u);
 }
 
 static void SetSamplerState(u32 index, float custom_tex_scale, bool custom_tex,
@@ -1182,23 +1177,9 @@ TextureCacheBase::TCacheEntry* TextureCacheBase::Load(const u32 stage)
     return bound_textures[stage];
   }
 
-  const FourTexUnits& tex = bpmem.tex[stage >> 2];
-  const u32 id = stage & 3;
-  const u32 address = (tex.texImage3[id].image_base /* & 0x1FFFFF*/) << 5;
-  u32 width = tex.texImage0[id].width + 1;
-  u32 height = tex.texImage0[id].height + 1;
-  const TextureFormat texformat = tex.texImage0[id].format;
-  const u32 tlutaddr = tex.texTlut[id].tmem_offset << 9;
-  const TLUTFormat tlutfmt = tex.texTlut[id].tlut_format;
-  const bool use_mipmaps = SamplerCommon::AreBpTexMode0MipmapsEnabled(tex.texMode0[id]);
-  u32 tex_levels = use_mipmaps ? ((tex.texMode1[id].max_lod + 0xf) / 0x10 + 1) : 1;
-  const bool from_tmem = tex.texImage1[id].cache_manually_managed != 0;
-  const u32 tmem_address_even = from_tmem ? tex.texImage1[id].tmem_even * TMEM_LINE_SIZE : 0;
-  const u32 tmem_address_odd = from_tmem ? tex.texImage2[id].tmem_odd * TMEM_LINE_SIZE : 0;
+  TextureInfo texture_info = TextureInfo::FromStage(stage);
 
-  auto entry = GetTexture(address, width, height, texformat,
-                          g_ActiveConfig.iSafeTextureCache_ColorSamples, tlutaddr, tlutfmt,
-                          use_mipmaps, tex_levels, from_tmem, tmem_address_even, tmem_address_odd);
+  auto entry = GetTexture(g_ActiveConfig.iSafeTextureCache_ColorSamples, texture_info);
 
   if (!entry)
     return nullptr;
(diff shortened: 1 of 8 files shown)


Source/Core/VideoCommon/TextureInfo.cpp
```diff
@@ -0,0 +1,325 @@
...
+}
+
+u32 TextureInfo::GetLevelCount() const
+{
+  return static_cast<u32>(m_mip_levels.size()) + 1;
+}
+
+const TextureInfo::MipLevel* TextureInfo::GetMipMapLevel(u32 level) const
+{
+  if (m_mip_levels.size() > level)
```
lioncash, line 262: you may want to make the check in terms of the would-be in-bounds level like:

```cpp
if (level < m_mip_levels.size())
```

Which makes it a little more explicit that it's checking against an upper bound.

Source/Core/VideoCommon/TextureInfo.cpp
```diff
@@ -0,0 +1,325 @@
...
+    return &m_mip_levels.at(level);
```
lioncash, line 263:
```suggestion
    return &m_mip_levels[level];
```

Given the bounds is already checked before this access.

Source/Core/VideoCommon/TextureInfo.cpp
```diff
@@ -0,0 +1,325 @@
...
+
+  const u8*& data = from_tmem ? ((level % 2) ? ptr_odd : ptr_even) : src_data;
+  m_data = data;
+  data += m_texture_size;
+}
+
+u32 TextureInfo::GetFullLevelSize() const
+{
+  u32 all_mips_size = 0;
+  for (std::size_t i = 0; i < m_mip_levels.size(); i++)
```
lioncash, line 289: Can't this be condensed into a ranged for loop like

```cpp
for (const auto& mip_map : m_mip_levels)
```

This also avoids the copy-by-value that currently occurs within the loop.

Source/Core/VideoCommon/TextureInfo.h
```diff
@@ -0,0 +1,111 @@
+// Copyright 2021 Dolphin Emulator Project
+// Licensed under GPLv2+
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <optional>
+#include <string>
```
lioncash, line 8: `<vector>` needs to be included.

Source/Core/VideoCommon/TextureInfo.h
```diff
@@ -0,0 +1,111 @@
...
+
+#include "Common/CommonTypes.h"
+#include "VideoCommon/TextureDecoder.h"
```
lioncash, line 11: Would this be able to be turned into a set of forward declarations.

Source/Core/VideoCommon/TextureInfo.cpp
iwubcode: Yeah, the reason this is done is this is a `const` function and `[]` is not const.

iwubcode: (I do the check because Dolphin doesn't support exceptions, so I think this would be a hard crash??)

Pokechu22: There are both const and non-const versions of `[]`: https://en.cppreference.com/w/cpp/container/array/operator_at (one returning a `T&` and the other returning `const T&`).

iwubcode: But not for `std::vector` until C++20:  https://en.cppreference.com/w/cpp/container/vector/operator_at

iwubcode: Ah.  I was thinking of map.  Great, I'll switch back to `[]`.  Thanks all

Source/Core/VideoCommon/TextureCacheBase.cpp
```diff
@@ -1214,86 +1195,58 @@ TextureCacheBase::TCacheEntry* TextureCacheBase::Load(const u32 stage)
...
-                             const int textureCacheSafetyColorSampleSize, u32 tlutaddr,
-                             TLUTFormat tlutfmt, bool use_mipmaps, u32 tex_levels, bool from_tmem,
-                             u32 tmem_address_even, u32 tmem_address_odd)
+TextureCacheBase::GetTexture(const int textureCacheSafetyColorSampleSize, TextureInfo& texture_info)
 {
-  // TexelSizeInNibbles(format) * width * height / 16;
-  const unsigned int bsw = TexDecoder_GetBlockWidthInTexels(texformat);
-  const unsigned int bsh = TexDecoder_GetBlockHeightInTexels(texformat);
+  unsigned int expandedWidth = texture_info.GetExpandedWidth();
+  unsigned int expandedHeight = texture_info.GetExpandedHeight();
```
Pokechu22, line 1201: This should be `expanded_width`, and possibly also `u32` instead of `unsigned int`.  I also find the logic around this a bit confusing (both that and `width` are updated by the hires texture logic, but in a way I don't find super obvious).

Source/Core/VideoCommon/TextureInfo.cpp
```diff
@@ -0,0 +1,325 @@
...
+
+  m_base_name = fmt::format("{}{}x{}{}_{:016x}", format_prefix, m_raw_width, m_raw_height,
+                            m_mip_levels.empty() ? "" : "_m", tex_hash);
+  m_tlut_name = tlut_size ? fmt::format("_{:016x}", tlut_hash) : "";
+  m_format_name = fmt::format("_{}", static_cast<int>(m_texture_format));
+
+  m_name_cached = true;
+}
+
+std::optional<std::string> TextureInfo::GetFullName() const
```
Pokechu22, line 153: I find it odd that `GetFullName` is optional, but the others aren't.

Source/Core/VideoCommon/TextureInfo.cpp
```diff
@@ -0,0 +1,325 @@
...
+
+    for (u32 i = 0; i < mip_count; i++)
+    {
+      MipLevel mip_level(i + 1, *this, m_from_tmem, src_data, ptr_even, ptr_odd);
+      m_mip_levels.push_back(std::move(mip_level));
+    }
+  }
+}
+
+void TextureInfo::CalculateTextureName()
```
Pokechu22, line 87: It seems odd that this is a public function.  I'd design it by having the various name getters compute the data if needed, and use the `mutable` specifier on `m_name_cached` and the names.  At least as far as I can tell, `TextureInfo`'s actual data is immutable, so once the name has been computed once it will never need to be computed again (which is consistent with `m_name_cached` never being set to false again), so I think the actual design is fine.

Pokechu22 (commented):
Some initial thoughts; I haven't read through this in detail yet.

Source/Core/VideoCommon/TextureCacheBase.cpp
iwubcode: Agreed.  TextureCacheBase is very confusing.  This PR isn't mean to completely solve that, just take a step in cleaning things up a bit (though the main purpose is for an enhancement).

Source/Core/VideoCommon/TextureInfo.cpp
iwubcode: I personally hate the `mutable` keyword.  I just think it brings about confusion.  Hey this function is `const` but you have to look at the member variables to see if something actually changed!

That being said, I like your idea.  The only downside would be the getters would incur slight overhead by having multiple `if ( !m_name_cached ) CalculateTextureName()` calls.  I thought of making this function return a `tuple` or a `struct` instead and removing the getters.  Not sure if you'd prefer that or keeping the current design.

Pokechu22: I was thinking that instead of having `m_tlut_name`, `tlut_size` being nonzero would control whether or not the tlut hash (and the underscore before it) were included in the output.  But I don't think there's actually a good way of doing that.

Source/Core/VideoCommon/TextureInfo.cpp
```diff
@@ -0,0 +1,325 @@
...
+  const u32 tmem_address_even = m_from_tmem ? tex.texImage1[id].tmem_even * TMEM_LINE_SIZE : 0;
+  const u32 tmem_address_odd = m_from_tmem ? tex.texImage2[id].tmem_odd * TMEM_LINE_SIZE : 0;
+  if (m_from_tmem)
+  {
+    m_data = &texMem[tmem_address_even];
+    m_tmem_odd = &texMem[tmem_address_odd];
+  }
+  else
+  {
+    m_data = Memory::GetPointer(m_raw_address);
```
phire, line 52: `m_data` is a bad name. Implies that it's a owned copy of the data, while this is just a pointer. The underlying data might change.

`m_ptr` might be better? 

Or a `u32 m_address` and every usage can just do a `Memory::GetPointer(m_address)`?

If *somebody* ever gets around to implementing locking (*glares at self*), then storing pointers into emulated memory is a bad idea that can potentially break things.

Source/Core/VideoCommon/TextureInfo.cpp
```diff
@@ -0,0 +1,325 @@
...
+  m_expanded_width = Common::AlignUp(m_raw_width, parent.GetBlockWidth());
+  m_expanded_height = Common::AlignUp(m_raw_height, parent.GetBlockHeight());
+
+  // load mips - TODO: Loading mipmaps from tmem is untested!
+
+  m_texture_size = TexDecoder_GetTextureSizeInBytes(m_expanded_width, m_expanded_height,
+                                                    parent.GetTextureFormat());
+
+  const u8*& data = from_tmem ? ((level % 2) ? ptr_odd : ptr_even) : src_data;
+  m_data = data;
```
phire, line 283: Same as above, should be `m_ptr` or `m_address`

Source/Core/VideoCommon/TextureInfo.cpp
```diff
@@ -0,0 +1,325 @@
...
+
+#include <xxhash.h>
+
+#include "Common/Align.h"
+#include "Core/HW/Memmap.h"
+#include "VideoCommon/BPMemory.h"
+#include "VideoCommon/SamplerCommon.h"
+#include "VideoCommon/TextureDecoder.h"
+
+TextureInfo::TextureInfo(u32 stage)
```
phire, line 15: I'm not sure I like how this constructor pulls a whole bunch of info out of the emulated state. Seems wrong that something called `TextureInfo` has a constructor that is anything other than pure. 

Instead, I recommend having a static method called something like `TextureInfo::FromCurrentState(u32 stage)` that returns a constructed TextureInfo.

Source/Core/VideoCommon/TextureInfo.cpp
phire: > Well, I was thinking I'd still cache it. There's a possibility the name can be used multiple times in one cycle

As I pointed on in my own review, you are burning a full 76 bytes for every `TextureInfo` object, just so you can save a few cpu cycles in a few rare code paths. I highly doubt the increased memory overhead is worth it.

Source/Core/VideoCommon/TextureCacheBase.cpp
```diff
@@ -1637,46 +1602,34 @@ TextureCacheBase::GetTexture(u32 address, u32 width, u32 height, const TextureFo
...
-        const u32 decoded_mip_size = expanded_mip_width * sizeof(u32) * expanded_mip_height;
-        TexDecoder_Decode(dst_buffer, mip_src_data, expanded_mip_width, expanded_mip_height,
-                          texformat, tlut, tlutfmt);
-        entry->texture->Load(level, mip_width, mip_height, expanded_mip_width, dst_buffer,
-                             decoded_mip_size);
+        const u32 decoded_mip_size =
+            mip_level->GetExpandedWidth() * sizeof(u32) * mip_level->GetExpandedHeight();
+        TexDecoder_Decode(dst_buffer, mip_level->GetData(), mip_level->GetExpandedWidth(),
+                          mip_level->GetExpandedHeight(), texture_info.GetTextureFormat(),
+                          texture_info.GetTlutAddress(), texture_info.GetTlutFormat());
```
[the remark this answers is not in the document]
iwubcode: Thanks phire.  I think all 3 of these are great suggestions.  However, yeah, I'd prefer to save this cleanup for a separate PR (maybe even separate them out, depending on how involved they are).

Source/Core/VideoCommon/TextureInfo.cpp
iwubcode: Yeah, I think I'll stick with `m_ptr` for now.  Only because certain things become a bit more painful once we say we can't store pointers.  If locking manifests, we will definitely want to revisit.

Source/Core/VideoCommon/TextureInfo.h
```diff
@@ -0,0 +1,115 @@
...
+  u32 GetTextureSize() const;
+
+  u32 GetBlockWidth() const;
+  u32 GetBlockHeight() const;
+
+  u32 GetExpandedWidth() const;
+  u32 GetExpandedHeight() const;
+
+  u32 GetRawWidth() const;
+  u32 GetRawHeight() const;
```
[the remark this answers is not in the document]
lioncash: Note that public const variables would prevent move assignment from being able to occur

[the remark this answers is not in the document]
iwubcode: @lioncash - yeah thought of that.  Haven't decided whether to do this or not but if I do would you want me to be explicit and delete the move operators in that case?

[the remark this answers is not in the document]
iwubcode: If there's no objections, I decided to keep it as is for now.  Once I have a clearer idea of how this class will be used (if moves aren't useful for instance), we can revisit.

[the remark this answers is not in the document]
phire: Fair enough

phire (approved):

lioncash (approved):

PR #9682, opened 2021-05-02
IOS: Don't let Kernel initialize WiiRoot if already initialized

JosJuice:
The SaveToSYSCONF call in BootManager.cpp was unintentionally overriding the temporary NAND set by the preceding InitializeWiiRoot call. Fixes https://bugs.dolphin-emu.org/issues/12500.

### DIFF
--- a/Source/Core/Core/WiiRoot.cpp
+++ b/Source/Core/Core/WiiRoot.cpp
@@ -33,6 +33,7 @@ namespace Core
 namespace FS = IOS::HLE::FS;
 
 static std::string s_temp_wii_root;
+static bool s_wii_root_initialized = false;
 
 static bool CopyBackupFile(const std::string& path_from, const std::string& path_to)
 {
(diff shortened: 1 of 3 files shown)


Source/Core/Core/WiiRoot.cpp
```diff
@@ -198,6 +199,8 @@ void InitializeWiiRoot(bool use_temporary)
   {
     File::SetUserPath(D_SESSION_WIIROOT_IDX, File::GetUserPath(D_WIIROOT_IDX));
   }
+
+  s_wii_root_initialized = true;
```
leoetlino, line 203: Should we add an assertion that !s_wii_root_initialized in InitializeWiiRoot?

leoetlino (approved):

Source/Core/Core/WiiRoot.cpp
JosJuice: Done.

leoetlino (approved):

BhaaLseN (approved):

JMC47:
This breaks netplay recorded dtms.

JosJuice:
Breaks them in what way? A desync (presumably due to the starting NAND differing)?

jordan-woyak:
> Breaks them in what way? A desync (presumably due to the starting NAND differing)?

Yeah, at least that's the claim of this issue report: https://bugs.dolphin-emu.org/issues/13732

PR #9683, opened 2021-05-02
UnitTests: Put PowerPC test values in a separate file

JosJuice:
We use more or less the same values for all PowerPC float unit tests. Let's put them in one place instead of duplicating them.


Source/UnitTests/Core/PowerPC/Jit64Common/Frsqrte.cpp
```diff
@@ -60,38 +61,9 @@ TEST(Jit64, Frsqrte)
...
-      0xFFF0'0000'0000'0001,  // first negative SNaN
-      0xFFF7'FFFF'FFFF'FFFF,  // last negative SNaN
-      0xFFF8'0000'0000'0000,  // first negative QNaN
-      0xFFFF'FFFF'FFFF'FFFF,  // last negative QNaN
-  };
-
   UReg_FPSCR fpscr;
 
-  for (u64 ivalue : special_values)
+  for (u64 ivalue : double_test_values)
```
Dentomologist, line 66: While you're here, ivalue could be const.

Source/UnitTests/Core/PowerPC/TestValues.h
```diff
@@ -0,0 +1,96 @@
...
+    0x3680'1234'5678'9ABC, 0x36A0'1234'5678'9ABC, 0x36B0'1234'5678'9ABC, 0xB680'1234'5678'9ABC,
+    0xB6A0'1234'5678'9ABC, 0xB6B0'1234'5678'9ABC,
+
+    // Some typical numbers
+    0x3FF8'0000'0000'0000,  // 1.5
+    0x408F'4000'0000'0000,  // 1000
+    0xC008'0000'0000'0000,  // -3
+};
+
+const std::array<u32, 29> single_test_values{
```
lioncash, line 63: This can be constexpr. LGTM otherwise

JosJuice: Fixed.

leoetlino (approved):

PR #9685, opened 2021-05-02
Remove Input Mapping Delay

JMC47:
While this opens up the possibility to create more complex expressions, the half second delay was the source of much confusion and frustration for users who didn't know what the emulator was doing.  The emulator essentially hard locked for 500ms and sometimes would get all kinds of weird inputs because they'd start spamming buttons trying to unfreeze it.

This makes the delay 0, returning the behavior to the way it was before.  This means you cannot do | statements without using the advanced input expression anymore.

### DIFF
--- a/Source/Core/DolphinQt/Config/Mapping/MappingCommon.cpp
+++ b/Source/Core/DolphinQt/Config/Mapping/MappingCommon.cpp
@@ -21,7 +21,7 @@
 namespace MappingCommon
 {
 constexpr auto INPUT_DETECT_INITIAL_TIME = std::chrono::seconds(3);
-constexpr auto INPUT_DETECT_CONFIRMATION_TIME = std::chrono::milliseconds(500);
+constexpr auto INPUT_DETECT_CONFIRMATION_TIME = std::chrono::milliseconds(0);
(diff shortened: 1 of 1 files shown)


JosJuice:
While I agree with getting rid of the delay, we should probably do a proper removal of it from the code instead of setting the timeout value to 0.

JMC47:
I didn't want to change things too drastically.  If someone can do it better, that'd be fine.  I think everyone agrees that the current setup is not good.  Or if you'd rather me do it, I can try but I can't guarantee it'd be correct.

Filoppi:
> While I agree with getting rid of the delay, we should probably do a proper removal of it from the code instead of setting the timeout value to 0.

Look at the source, setting the value to 0 doesn't create any special case, it should always work perfectly.
This could be back based on a UI setting, like "Detect Input Over Time" or something like that, so I don't think it should be removed. It's literally a sum in the code.

Filoppi (approved):

JosJuice:
Okay, yeah, after looking at what `DetectInput` does more closely I'd say it's okay to just set it to 0.

jordan-woyak (approved):

PR #9687, opened 2021-05-02
Fix config descriptions grammar

Dentomologist:


Source/Core/DolphinQt/Config/Graphics/GeneralWidget.cpp
```diff
@@ -200,15 +200,15 @@ void GeneralWidget::AddDescriptions()
 #if defined(_WIN32)
```
leoetlino, line 200: Why do we even have two different strings for Windows and non-Windows? Can't we just unify both with "any of the other backends"?

The non-Windows description is also missing the "Different games and different GPUs will behave differently" bit.

Dentomologist: Done.

I originally did add the "Different games" bit to non-Windows platforms, but apparently that got lost while I was rewriting the history before the PR.

lioncash (approved):

PR #9688, opened 2021-05-04
Input cleanup

Filoppi:
Skimming of: https://github.com/dolphin-emu/dolphin/pull/9489

Nothing here should really make any behavioural difference over the code. It's just:
-const correctness
-improved commenting
-added a new log category for the controller interface
-follow coding conventions and grammar errors
-expose mapping bool tooltips
-extremely tiny fixes

Most commits don't overlap the same files, but I suggest looking at commits individually when reviewing.


Filoppi:
@dolphin-emu-bot rebuild

Source/Core/VideoCommon/RenderBase.cpp
```diff
@@ -770,6 +770,7 @@ void Renderer::UpdateDrawRectangle()
   const float win_width = static_cast<float>(m_backbuffer_width);
   const float win_height = static_cast<float>(m_backbuffer_height);
 
+  // TODO: fix, this breaks at very low widget sizes
```
Rumi-Larry, line 773: This should be a "FIXME:"

Source/Core/Core/HW/GCPadEmu.h
```diff
@@ -35,7 +35,7 @@ class GCPad : public ControllerEmu::EmulatedController
 {
 public:
   explicit GCPad(unsigned int index);
-  GCPadStatus GetInput() const;
+  GCPadStatus GetInput();
```
iwubcode, line 38: Any particular reason why this isn't `const` anymore?

Filoppi: Mistake

Source/Core/Core/HW/GCPadEmu.cpp
```diff
@@ -93,10 +93,12 @@ GCPad::GCPad(const unsigned int index) : m_index(index)
 
   // options
   groups.emplace_back(m_options = new ControllerEmu::ControlGroup(_trans("Options")));
-  m_options->AddSetting(&m_always_connected_setting,
-                        // i18n: Treat a controller as always being connected regardless of what
-                        // devices the user actually has plugged in
```
BhaaLseN, line 98: This `i18n` comment still makes sense IMO, any particular reason why you removed it? Translators only see the string "Always Connected" and this comment makes sure they also get some context on what it _means_.

Source/Core/InputCommon/ControlReference/ExpressionParser.cpp
```diff
@@ -371,6 +369,7 @@ class BinaryExpression : public Expression
     }
     case TOK_ASSIGN:
     {
+      // You can do crazy things with this
```
BhaaLseN, line 372: This comment seems a bit off, should it serve any particular purpose? It doesn't really explain _why_ you can do crazy things (or even more specifically, _what_ crazy things), so I don't see much point in adding it like that.

Source/Core/InputCommon/ControlReference/ExpressionParser.cpp
```diff
@@ -945,6 +947,7 @@ static std::unique_ptr<Expression> ParseBarewordExpression(const std::string& st
   qualifier.control_name = str;
   qualifier.has_device = false;
 
+  // This will only work with the default device
```
BhaaLseN, line 950: Care to elaborate? What happens with a non-default device? Does it explode, does it do _nothing_?

Source/Core/InputCommon/ControllerEmu/ControlGroup/ControlGroup.cpp
```diff
@@ -44,7 +44,7 @@ void ControlGroup::AddDeadzoneSetting(SettingValue<double>* value, double maximu
               // i18n: The percent symbol.
               _trans("%"),
               // i18n: Refers to the dead-zone setting of gamepad inputs.
-              _trans("Input strength to ignore.")},
+              _trans("Input strength to ignore and remap.")},
```
BhaaLseN, line 47: Not sure if "dead zone" is _also_ used to _remap_ anything. Care to elaborate?

Source/Core/InputCommon/ControllerEmu/ControlGroup/Cursor.h
```diff
@@ -25,9 +25,10 @@ class Cursor : public ReshapableInput
 
   Cursor(std::string name, std::string ui_name);
 
-  ReshapeData GetReshapableState(bool adjusted) final override;
+  ReshapeData GetReshapableState(bool adjusted) const final override;
   ControlState GetGateRadiusAtAngle(double ang) const override;
 
+  // Modifies the state
   StateData GetState(bool adjusted);
```
BhaaLseN, line 32: This seems like a weird choice, having _some_ `GetState` members as `const` and others as _not_ `const`. By the name, they probably _should_ be, but I guess this would be out of scope here since it seems a bit more involved for the Cursor input.

Source/Core/InputCommon/ControllerEmu/ControlGroup/IMUGyroscope.cpp
```diff
@@ -40,7 +40,7 @@ IMUGyroscope::IMUGyroscope(std::string name_, std::string ui_name_)
               // i18n: "°/s" is the symbol for degrees (angular measurement) divided by seconds.
               _trans("°/s"),
               // i18n: Refers to the dead-zone setting of gyroscope input.
-              _trans("Angular velocity to ignore.")},
+              _trans("Angular velocity to ignore and remap.")},
```
BhaaLseN, line 43: Same here for "remap"

Source/Core/Core/HW/GCPadEmu.cpp
```diff
@@ -181,64 +183,60 @@ void GCPad::LoadDefaults(const ControllerInterface& ciface)
...
 #else
   // OS X/Linux
   // Start
-  m_buttons->SetControlExpression(5, "Return");
+  m_buttons->SetControlExpression(5, "`Return`");
 #endif
 
-  // stick modifiers to 50 %
-  m_main_stick->controls[4]->control_ref->range = 0.5f;
-  m_c_stick->controls[4]->control_ref->range = 0.5f;
```
BhaaLseN, line 199: Did you intentionally remove those lines?

Source/Core/InputCommon/ControllerEmu/ControlGroup/ModifySettingsButton.h
```diff
@@ -18,10 +18,11 @@ class ModifySettingsButton : public Buttons
...
+  // Modifies the state
   void GetState();
 
-  const std::vector<bool>& isSettingToggled() const;
-  const std::vector<bool>& getSettingsModifier() const;
+  const std::vector<bool>& IsSettingToggled() const;
+  const std::vector<bool>& GetSettingsModifier() const;
 
 private:
   std::vector<bool> threshold_exceeded;  // internal calculation (if "state" was above threshold)
```
BhaaLseN, line 28: You renamed members in the `ExpressionParser` to have the `m_` prefix, but didn't touch those. Is this an oversight, or something you intend to do in another PR?

Source/Core/InputCommon/ControllerEmu/ControlGroup/ModifySettingsButton.h
```diff
@@ -18,10 +18,11 @@ class ModifySettingsButton : public Buttons
 
   void AddInput(std::string button_name, bool toggle = false);
 
+  // Modifies the state
   void GetState();
```
BhaaLseN, line 22: Same here, being weird as non `const` (and probably out of scope)

Source/Core/InputCommon/InputConfig.h
```diff
@@ -34,7 +34,8 @@ class InputConfig
     m_controllers.emplace_back(std::make_unique<T>(std::forward<Args>(args)...));
   }
 
-  ControllerEmu::EmulatedController* GetController(int index);
+  ControllerEmu::EmulatedController* GetController(int index) const;
+  int GetControllersNum() const;
```
BhaaLseN, line 38: This seems unused, did you mean to put this in another PR instead?

BhaaLseN (commented):
Some changes make sense, others appear a bit random; so I'm not completely sure what you meant with some of the changes and whether they even should be here.
Or I'm just misunderstanding what you're trying to do...

Source/Core/InputCommon/InputConfig.h
Filoppi: No I specified I will use in another PR in the commit (input caching), it was next to another change so I just thought I'd push this as well. Not too fuzzed about it, but it could have other uses as well.

Source/Core/Core/HW/GCPadEmu.cpp
Filoppi: No this is a mistake that happened when splitting up changes. thanks.

Source/Core/InputCommon/ControlReference/ExpressionParser.cpp
Filoppi: Definitely agree, I'm not the type to write these comments and I knew someone would notice it, I will clarify it.

Source/Core/InputCommon/ControllerEmu/ControlGroup/ModifySettingsButton.h
Filoppi: It's not like I specifically looked for anything that didn't have m_, I just fixed the naming of things I ended up directly working with. I don't see the problem here.

Source/Core/InputCommon/ControllerEmu/ControlGroup/ControlGroup.cpp
Filoppi: The deadzone remaps the values after the deadzone.
If deadzone is 0.1 and you pass in 0.5, the outcome won't be 0.5, it will be 0.55, so there is a remap on going (applied differently depending on the kind of deadzone), just saying "ignore" is not enough in my opinion, it's actually misleading.

Source/Core/InputCommon/ControlReference/ExpressionParser.cpp
Filoppi: I will make this clearer

Source/Core/Core/HW/GCPadEmu.cpp
Filoppi: So what do I need to do? Copy and paste the tooltip I added:
`If this is false and you have a default device,\nit will link the emulated controller connection state\nto the real device connection state.`
as `i18n`?

Source/Core/InputCommon/ControllerEmu/ControlGroup/ModifySettingsButton.h
Filoppi: I didn't write this method nor removed the const, I just think it's good to clarify that a function called GetState() does modify the state internally.

Source/Core/InputCommon/ControllerEmu/ControlGroup/Cursor.h
Filoppi: See my answer to the other similar comment you made. Still don't understand when you say: out of scope.
Worth nothing that I tried to quickly make these const, but it was too complicated, it can be done in the future.

Source/Core/InputCommon/ControllerEmu/ControlGroup/ControlGroup.cpp
JosJuice: While the change is technically correct, I think it's going to be hard for users to understand what you mean. I think you at the very least should expand the i18n comment to explain what's meant by remap, so that the string doesn't get mistranslated.

Source/Core/Core/HW/GCPadEmu.cpp
JosJuice: I think the tooltip is a little hard to understand. But if it wasn't for that, then either putting a copy of the tooltip as an i18n comment or keeping the existing i18n comment would be fine.

Source/Core/InputCommon/ControllerEmu/ControlGroup/ControlGroup.cpp
Filoppi: I can do that. If anyone has got a better word than remap, I'm happy to change it (maybe scale?). "Ignore" alone is technically wrong.

BhaaLseN: I didn't realize this also happened here, thanks for clarifying!

Source/Core/InputCommon/ControllerEmu/ControlGroup/Cursor.h
BhaaLseN: With "out of scope" i meant "rewriting `GetState` to be `const`-able. Your other PR does something with a `bool update` parameter that makes it clear; and it is probably better suited for something like that (if you _do_ want to make this more explicit than just a comment)

Source/Core/InputCommon/InputConfig.h
BhaaLseN: No preference either way, it just raises an eyebrow here when you add a method but never use it. Makes it appear as incomplete change (which either misses the call, or was meant for something else).

Filoppi: I can take it out if it's a problem, but I'm fairly sure the rest of the input caching will be merged, PR is coming soon.

Source/Core/InputCommon/ControllerEmu/ControlGroup/Cursor.h
Filoppi: For this specific GetState() on the Cursor, I have also made changes to it, which were already in this PR:
https://github.com/dolphin-emu/dolphin/blob/d41f881b5b1e1579fb4a3bc40d769696bee71e00/Source/Core/InputCommon/ControllerEmu/ControlGroup/Cursor.h#L37
but will come in a more polished version in one of my upcoming PRs.

So it will kind of also have a `bool update`

Source/Core/InputCommon/ControllerEmu/ControlGroup/ModifySettingsButton.h
Filoppi: Renamed to UpdateState() because the name GetState() made no sense here

Source/Core/Core/HW/GCPadEmu.cpp
Filoppi: Re-written the tooltip, put back the `i18n`

Filoppi:
@BhaaLseN Thanks for the review. I think I corrected or improved everything you have mentioned.

Source/Core/VideoCommon/RenderBase.cpp
```diff
@@ -770,6 +770,7 @@ void Renderer::UpdateDrawRectangle()
   const float win_width = static_cast<float>(m_backbuffer_width);
   const float win_height = static_cast<float>(m_backbuffer_height);
 
+  // FIXME: fix, this breaks at very low widget sizes
```
Rumi-Larry, line 773: You can erase the word "fix" after the colon

Source/Core/InputCommon/ControllerEmu/StickGate.cpp
```diff
@@ -285,6 +286,9 @@ ReshapableInput::ReshapeData ReshapableInput::Reshape(ControlState x, ControlSta
   x -= m_center.x;
   y -= m_center.y;
 
+  // We run this even if both x and y will be zero.
+  // The angle value will be random but dist will stay 0
+
   // TODO: make the AtAngle functions work with negative angles:
   ControlState angle = std::atan2(y, x) + MathUtil::TAU;
```
Pokechu22, line 294: I don't think it's entirely accurate to say the angle value will be _random_; [it's actually specified](https://en.cppreference.com/w/cpp/numeric/math/atan2#Error_handling) to be either +0, -0, +Pi, or -Pi depending on the sign of x and y.  I'm not quite sure what phrasing would be best.

Source/Core/Core/HW/GCPadEmu.cpp
```diff
@@ -96,7 +96,11 @@ GCPad::GCPad(const unsigned int index) : m_index(index)
   m_options->AddSetting(&m_always_connected_setting,
                         // i18n: Treat a controller as always being connected regardless of what
                         // devices the user actually has plugged in
-                        _trans("Always Connected"), false);
+                        {_trans("Always Connected"), _trans(""),
+                         _trans("Always connected if true.\nIt will link the emulated controller "
+                                "connetion state to the\n"
+                                "real default device connetion state if false (if there is one).")},
```
AdmiralCurtiss, line 102: Typo, 'connetion' -> 'connection'. Two instances.

AdmiralCurtiss: Also rereading that this is not a good comment to expose to a non-programmer user like that... Maybe replace the 'true' and 'false' with 'checked' and 'unchecked'?

Source/Core/InputCommon/ControllerEmu/StickGate.cpp
Pokechu22: Maybe "`std::atan2(0, 0)` returns a valid non-NaN value, but the exact value (which depends on the signs of x and y) does not matter here as dist is zero"?

AdmiralCurtiss:
Two commits still have typos in their commit messages. ("NumericSettings support a mas" and "fix some related grammar erros")

Filoppi:
> Two commits still have typos in their commit messages. ("NumericSettings support a mas" and "fix some related grammar erros")

Fixed both. thx

Source/Core/Core/HW/GCPadEmu.cpp
```diff
@@ -93,10 +93,15 @@ GCPad::GCPad(const unsigned int index) : m_index(index)
...
-                        // devices the user actually has plugged in
-                        _trans("Always Connected"), false);
+  m_options->AddSetting(
+      &m_always_connected_setting,
+      // i18n: Treat a controller as always being connected regardless of what
+      // devices the user actually has plugged in
+      {_trans("Always Connected"), _trans(""),
+       _trans("Always connected if checked.\nIf unchecked, it will link the emulated controller "
+              "connection state\nto the "
+              "real default device connection state (if there is one).")},
```
Pokechu22, line 103: Minor, but I think this would look better like this (it's the same number of lines, but it more closely reflects what's actually shown):

```
       _trans("Always connected if checked.\n"
              "If unchecked, it will link the emulated controller connection state\n"
              "to the real default device connection state (if there is one).")},
```

Filoppi: True, I will change it, but if we are gonna worry about these nitpick, we are never gonna get my other main PRs merged 😅

Rumi-Larry: Not necessarily, I think that with this change, all natural languages segments of the diff are accounted for.

phire (approved):

JMC47 (approved):
My controllers all work.  I love the new log type actually.  It was a good decision.

PR #9689, opened 2021-05-04
Input cleanup 2

Filoppi:
Skimming of: https://github.com/dolphin-emu/dolphin/pull/9489

A few fixes/improvements/cleanup for input related stuff, nicely described in commits.
Separated from https://github.com/dolphin-emu/dolphin/pull/9688 because this one actually has behavioural changes and needs to be reviewed.

### DIFF
--- a/Source/Core/InputCommon/ControllerEmu/ControlGroup/IMUAccelerometer.cpp
+++ b/Source/Core/InputCommon/ControllerEmu/ControlGroup/IMUAccelerometer.cpp
@@ -4,13 +4,12 @@
 
 #include "InputCommon/ControllerEmu/ControlGroup/IMUAccelerometer.h"
 
+#include <algorithm>
 #include <memory>
 
 #include "Common/Common.h"
 #include "Core/HW/WiimoteEmu/WiimoteEmu.h"
-#include "InputCommon/ControlReference/ControlReference.h"
 #include "InputCommon/ControllerEmu/Control/Control.h"
-#include "InputCommon/ControllerEmu/Control/Input.h"
 
 namespace ControllerEmu
 {
@@ -25,9 +24,15 @@ IMUAccelerometer::IMUAccelerometer(std::string name_, std::string ui_name_)
   AddInput(Translate, _trans("Backward"));
 }
 
+bool IMUAccelerometer::AreInputsBound() const
+{
+  return std::all_of(controls.begin(), controls.end(),
+                     [](const auto& control) { return control->control_ref->BoundCount() > 0; });
+}
+
 std::optional<IMUAccelerometer::StateData> IMUAccelerometer::GetState() const
 {
-  if (controls[0]->control_ref->BoundCount() == 0)
+  if (!AreInputsBound())
     return std::nullopt;
 
   StateData state;
--- a/Source/Core/InputCommon/ControllerEmu/ControlGroup/IMUGyroscope.h
+++ b/Source/Core/InputCommon/ControllerEmu/ControlGroup/IMUGyroscope.h
@@ -23,7 +23,8 @@ public:
   IMUGyroscope(std::string name, std::string ui_name);
 
   StateData GetRawState() const;
-  std::optional<StateData> GetState() const;
+  // Also updates the state by default
+  std::optional<StateData> GetState(bool update = true);
 
   // Value is in rad/s.
   ControlState GetDeadzone() const;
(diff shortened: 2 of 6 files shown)


Source/Core/InputCommon/ControllerEmu/ControlGroup/IMUAccelerometer.cpp
```diff
@@ -25,9 +25,16 @@ IMUAccelerometer::IMUAccelerometer(std::string name_, std::string ui_name_)
...
+{
+  return controls[0]->control_ref->BoundCount() && controls[1]->control_ref->BoundCount() &&
+         controls[2]->control_ref->BoundCount() && controls[3]->control_ref->BoundCount() &&
+         controls[4]->control_ref->BoundCount() && controls[5]->control_ref->BoundCount();
+}
+
 std::optional<IMUAccelerometer::StateData> IMUAccelerometer::GetState() const
 {
-  if (controls[0]->control_ref->BoundCount() == 0)
+  if (!AreInputsBound())
```
iwubcode, line 37: FWIW I did the following in `IMUGyroscope`:

```
if (std::all_of(controls.begin(), controls.end(),
                  [](const auto& control) { return control->control_ref->BoundCount() == 0; }))
```

(and include `algorithm`)

iwubcode: Ah, I see you changed `IMUGyropscope`.  I'll need to see why but I'd still suggest matching behavior if possible.

Source/Core/InputCommon/ControllerEmu/ControlGroup/IMUGyroscope.h
```diff
@@ -23,7 +23,8 @@ class IMUGyroscope : public ControlGroup
   IMUGyroscope(std::string name, std::string ui_name);
 
   StateData GetRawState() const;
-  std::optional<StateData> GetState() const;
+  // Also updates the state
+  std::optional<StateData> GetState(bool update = false);
```
iwubcode, line 27: Where is this called with `update = true`?  Assuming this was added for a future PR...?

Filoppi: Good find, I should have changed the place that calls it to true like it's in the future PR. Forgot that.

Source/Core/InputCommon/ControllerEmu/ControlGroup/IMUAccelerometer.cpp
Filoppi: True, I wrote this before someone implemented the std::all_of in IMUGyroscope and then didn't update it.

Source/Core/InputCommon/ControllerEmu/ControlGroup/IMUGyroscope.h
```diff
@@ -33,14 +34,16 @@ class IMUGyroscope : public ControlGroup
...
-  void UpdateCalibration(const StateData&) const;
+  bool AreInputsBound() const;
+  bool CanCalibrate() const;
+  void RestartCalibration();
+  void UpdateCalibration(const StateData&);
 
   SettingValue<double> m_deadzone_setting;
   SettingValue<double> m_calibration_period_setting;
 
-  mutable StateData m_calibration = {};
```
Filoppi, line 42: Just think it's nicer to have `GetState(bool update)` instead than `GetState() const` with mutable stuff, which does not make it clear externally that GetState() is infact an update.

AdmiralCurtiss (commented):
Looks fine otherwise, as far as I can tell anyway.

AdmiralCurtiss:
Oh, actually, two typos in commit messages that should probably be fixed!

`Similar to the guitar, only control[0] was check, and that feels random.` -> `checked`
`make GetState update optinional` -> `optional`

Filoppi:
> Oh, actually, two typos in commit messages that should probably be fixed!
> 
> `Similar to the guitar, only control[0] was check, and that feels random.` -> `checked`
> `make GetState update optinional` -> `optional`

done everything

AdmiralCurtiss:
b02281259fd1d5b8c5eaa86366df11197395931e still has the typo in the commit message. It's quite minor, but since we're already here...

AdmiralCurtiss (approved):

leoetlino (approved):
Changes look fine. Could you rebase this on master to fix the Windows build failures?

PR #9690, opened 2021-05-05
Jit64: divwux - Prefer three-operand IMUL

Sintendo:
Another tiny optimization.

By taking advantage of three-operand IMUL, we can eliminate a MOV instruction. This is a small code size win. However, due to IMUL sign extending the immediate value to 64 bits, we can only apply this when the magic number's most significant bit is zero.

To ensure this can actually happen, we also minimize the magic number by checking for trailing zeroes.

<details><summary>Unsigned division by 18</summary>

Before:
```
41 BE E4 38 8E E3    mov         r14d,0E38E38E4h
4D 0F AF F5          imul        r14,r13
49 C1 EE 24          shr         r14,24h
```

After:
```
4D 69 F5 39 8E E3 38 imul        r14,r13,38E38E39h
49 C1 EE 22          shr         r14,22h
```
</details>

<details><summary>Unsigned division by 52</summary>

Before:
```
B8 9E D8 89 9D       mov         eax,9D89D89Eh
48 0F AF F8          imul        rdi,rax
48 C1 EF 25          shr         rdi,25h
```

After:
```
48 69 FF 4F EC C4 4E imul        rdi,rdi,4EC4EC4Fh
48 C1 EF 24          shr         rdi,24h
```
</details>

<details><summary>Unsigned division by 125000</summary>

Before:
```
BF 06 BD 37 86       mov         edi,8637BD06h
49 0F AF FF          imul        rdi,r15
48 C1 EF 30          shr         rdi,30h
```

After:
```
49 69 FF 83 DE 1B 43 imul        rdi,r15,431BDE83h
48 C1 EF 2F          shr         rdi,2Fh
```
</details>

### DIFF
--- a/Source/Core/Core/PowerPC/Jit64/Jit_Integer.cpp
+++ b/Source/Core/Core/PowerPC/Jit64/Jit_Integer.cpp
@@ -1271,14 +1271,29 @@ void Jit64::divwux(UGeckoInstruction inst)
           RCX64Reg Rd = gpr.Bind(d, RCMode::Write);
           RegCache::Realize(Ra, Rd);
 
-          if (d == a)
+          magic++;
+
+          // Use smallest magic number and shift amount possible
+          while ((magic & 1) == 0 && shift > 0)
           {
-            MOV(32, R(RSCRATCH), Imm32(magic + 1));
+            magic >>= 1;
+            shift--;
+          }
+
+          // Three-operand IMUL sign extends the immediate to 64 bits, so we may only
+          // use it when the magic number has its most significant bit set to 0
+          if ((magic & 0x80000000) == 0)
+          {
+            IMUL(64, Rd, Ra, Imm32(magic));
+          }
+          else if (d == a)
+          {
+            MOV(32, R(RSCRATCH), Imm32(magic));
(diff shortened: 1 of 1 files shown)


Source/Core/Core/PowerPC/Jit64/Jit_Integer.cpp
```diff
@@ -1271,14 +1271,29 @@ void Jit64::divwux(UGeckoInstruction inst)
           RCX64Reg Rd = gpr.Bind(d, RCMode::Write);
           RegCache::Realize(Ra, Rd);
 
-          if (d == a)
+          magic++;
+
+          // Use smallest magic number and shift amount possible
+          while ((magic & 1) == 0)
```
MerryMage, line 1277: `&& shift > 0`

Sintendo: Don't think there is any risk of this being an infinite loop since divisions by powers of 2 are already handled elsewhere, but let's err on the side of caution.

leoetlino (approved):
looks fine to me, but would like a second opinion

lioncash (approved):

PR #9692, opened 2021-05-06
Android: Don't set android:debuggable="true"

JosJuice:
In PR #9654, I explicitly set `android:debuggable="true"` for Dolphin. (By default, it's set for debug builds but not release builds.) The reason I made the change is because debuggable must be set to true in order to allow `adb backup` to be used with apps which target Android 12. We have no reason to want to stop users from debugging Dolphin and certainly no reason to stop users from using `adb backup` (especially not after forced storage is going to force us to store the User folder in app-specific external storage!), but, it turns out that setting debuggable to true is forbidden by Google Play "for security reasons". Go figure. The beta build which was tagged earlier today was rejected because of this.

### DIFF
--- a/Source/Android/app/src/main/AndroidManifest.xml
+++ b/Source/Android/app/src/main/AndroidManifest.xml
@@ -38,9 +38,7 @@
         android:allowBackup="false"
         android:supportsRtl="true"
         android:isGame="true"
-        android:banner="@drawable/banner_tv"
-        android:debuggable="true"
-        tools:ignore="HardcodedDebugMode">
+        android:banner="@drawable/banner_tv">
         <meta-data
             android:name="android.max_aspect"
             android:value="2.1"/>


Rumi-Larry:
It really seems that developing for Android is more trouble than it's worth. What is the share of users for Android anyway? Most Android devices aren't powerful enough for a consistent decent performance in the most popular games. 
In any case, I assume that making the user set the app to be debuggable isn't straight foward.
It may be wise to gauge the community opinion about removing Android support ahead of time, because it seems like something that will eventually you'll be forced to do.

JosJuice:
> What is the share of users for Android anyway?

Relatively large. I don't have the exact numbers, though.

> Most Android devices aren't powerful enough for a consistent decent performance in the most popular games.

There is a good number of them that are powerful enough nowadays, and it will only get better in the future.

> In any case, I assume that making the user set the app to be debuggable isn't straight foward.

I don't believe it's possible to do so after the app that has been built. But not being able to use `adb backup` is really a minor thing, especially compared to scoped storage as a whole.

> It may be wise to gauge the community opinion about removing Android support ahead of time, because it seems like something that will eventually you'll be forced to do.

I have no plans to remove Android support and I don't think others on the team are expressing any desire to remove it either.

leoetlino (approved):
Unfortunate, but I guess we don't really have a choice

JosJuice:
This also fixed a crash that a user was getting with the development builds. So I suppose that even if were to set up making different builds for Google Play and for dolphin-emu.org, it's best to not enable this.

mbc07:
AFAIK the property that dictates is `adb backup` is allowed or not is `android:allowBackup`, which is currently set to false. Any particular reason to not have it set to true?

JosJuice:
I believe someone set it to false because Dolphin can do weird stuff if you back up the internal app-specific storage but not the User folder, like Wii controls not working until you use Change Controller. I'm planning to change that one to true when I enable using the external app-specific storage as the User folder.

PR #7925, opened 2019-03-24
TextureCache: Simplify XFB reconstruction

stenzek:
Currently, we don't consider the stride from the video interface when loading the XFB texture from memory. This results in textures which are twice the width being created, with both fields side-by-side. The data is then ignored as only the first half of the texture is actually displayed, but it's still incorrect. 

This PR changes things to create two cache entries, one for each field, and display them as the real hardware would.

ARM also had no CPU XFB decoding function, and with the GPUs not supporting texel buffers large enough, GPU decoding wasn't possible. So on Android, you likely just saw black screens for in-memory XFBs.

Lastly, we're still reliant on the force progressive hack, unless XFB2RAM is enabled. If XFB2RAM is disabled, you'll just get a purple screen, as for interlaced scanout, the stride won't match (copy stride will be half the scanout stride, as it's only displaying every second line). Otherwise, you can disable the hack, and see the half-vertical-resolution interlaced video in all its glory. See the comment in TextureCacheBase.cpp in StitchXFBCopy() for more info.

PR #9693, opened 2021-05-07
[Do not merge] Test fix for texture invalidation

Miksel12:
This seems to fix https://bugs.dolphin-emu.org/issues/12392?next_issue_id=11956&prev_issue_id=12435 and could fix other regressions from https://github.com/dolphin-emu/dolphin/pull/7925 but I can only test on fifologs.

I'm not sure if this is remotely correct but it doesn't seem to cause any issues.

### DIFF
--- a/Source/Core/VideoCommon/BPStructs.cpp
+++ b/Source/Core/VideoCommon/BPStructs.cpp
@@ -240,7 +240,7 @@ static void BPWritten(const BPCmd& bp)
     // known for configuring these out-of-range copies.
     u32 copy_width = srcRect.GetWidth();
     u32 copy_height = srcRect.GetHeight();
-    if (srcRect.right > EFB_WIDTH || srcRect.bottom > EFB_HEIGHT)
+    if (0 && (srcRect.right > EFB_WIDTH || srcRect.bottom > EFB_HEIGHT))
     {
       WARN_LOG_FMT(VIDEO, "Oversized EFB copy: {}x{} (offset {},{} stride {})", copy_width,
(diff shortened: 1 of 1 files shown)


JMC47:
@dolphin-emu-bot rebuild

iwubcode:
This basically just removes the logic done in https://github.com/dolphin-emu/dolphin/commit/378b605669641ec9675e137d04f4037bc621cd46 .  That commit says that it was hardware tested, so I can't imagine this is correct behavior.

JMC47:
Where is our bug in the texture cache, though?

Miksel12:
> This basically just removes the logic done in [378b605](https://github.com/dolphin-emu/dolphin/commit/378b605669641ec9675e137d04f4037bc621cd46) . That commit says that it was hardware tested, so I can't imagine this is correct behavior.

I see, though the fact that TMNT has a green bar that disappears when removing the logic makes me think that the logic is not completely correct.

iwubcode:
Well we are creating a new xfb/efb with a modified width/height.  Which means it will hit the logic that changed in 7925 where we recompute the hash and compare.  It's possible the hash we are computing is wrong or something in that logic is wrong..

iwubcode:
Possibly the `destStride` is wrong?

Or the other very likely possibility is the `GetXFBFromCache` is using `GetHash64` always.  However, the XFB hash generation logic has a check to see if its stride matches the bytesperrow (if it doesn't it does something else).  We're not consistent...

iwubcode:
JMC says https://github.com/dolphin-emu/dolphin/pull/9694 resolved the issue.

PR #9694, opened 2021-05-07
VideoCommon: update TextureCache logic for finding oversized XFBs

iwubcode:
Previously, when looking for XFBs in the texture cache, if their dimensions didn't match VI we were recomputing the hash.  However, when we did so we weren't using the same hash logic as we did when actually hashing the entry.

Additionally, when dealing with oversized XFBs (which are resized on creation) the hashing algorithm is actually different than the one used to hash the VI registers that we compare against.  This causes these copies to be invalidated, meaning we will fall back to stitching from guest memory (if it exists).

This change fixes both of those issues.  This seems to resolve [12392](https://bugs.dolphin-emu.org/issues/12392)

### DIFF
--- a/Source/Core/VideoCommon/TextureCacheBase.cpp
+++ b/Source/Core/VideoCommon/TextureCacheBase.cpp
@@ -1722,12 +1722,8 @@ TextureCacheBase::GetXFBTexture(u32 address, u32 width, u32 height, u32 stride,
     return nullptr;
   }
 
-  // Compute total texture size. XFB textures aren't tiled, so this is simple.
-  const u32 total_size = height * stride;
-  const u64 hash = Common::GetHash64(src_data, total_size, 0);
-
   // Do we currently have a version of this XFB copy in VRAM?
-  TCacheEntry* entry = GetXFBFromCache(address, width, height, stride, hash);
+  TCacheEntry* entry = GetXFBFromCache(address, width, height, stride);
   if (entry)
   {
     if (entry->is_xfb_container)
@@ -1743,11 +1739,16 @@ TextureCacheBase::GetXFBTexture(u32 address, u32 width, u32 height, u32 stride,
   // Create a new VRAM texture, and fill it with the data from guest RAM.
   entry = AllocateCacheEntry(TextureConfig(width, height, 1, 1, 1, AbstractTextureFormat::RGBA8,
                                            AbstractTextureFlag_RenderTarget));
+
+  // Compute total texture size. XFB textures aren't tiled, so this is simple.
+  const u32 total_size = height * stride;
   entry->SetGeneralParameters(address, total_size,
                               TextureAndTLUTFormat(TextureFormat::XFB, TLUTFormat::IA8), true);
   entry->SetDimensions(width, height, 1);
-  entry->SetHashes(hash, hash);
   entry->SetXfbCopy(stride);
+
+  const u64 hash = entry->CalculateHash();
+  entry->SetHashes(hash, hash);
   entry->is_xfb_container = true;
   entry->is_custom_tex = false;
   entry->may_have_overlapping_textures = false;
@@ -1784,7 +1785,7 @@ TextureCacheBase::GetXFBTexture(u32 address, u32 width, u32 height, u32 stride,
 }
 
 TextureCacheBase::TCacheEntry* TextureCacheBase::GetXFBFromCache(u32 address, u32 width, u32 height,
-                                                                 u32 stride, u64 hash)
+                                                                 u32 stride)
 {
   auto iter_range = textures_by_address.equal_range(address);
   TexAddrCache::iterator iter = iter_range.first;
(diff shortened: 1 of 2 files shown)


iwubcode:
@stenzek - curious if you have any thoughts on this one!

Source/Core/VideoCommon/TextureCacheBase.cpp
```diff
@@ -1795,15 +1829,18 @@ TextureCacheBase::TCacheEntry* TextureCacheBase::GetXFBFromCache(u32 address, u3
...
+    if (entry->is_xfb_copy && entry->memory_stride == stride && entry->native_width != width &&
+        entry->native_height != height && !entry->may_have_overlapping_textures)
     {
       // But if the dimensions do differ, we must compute the hash on the sub-rectangle.
       u64 check_hash = hash;
       if (entry->native_width != width || entry->native_height != height)
       {
-        check_hash = Common::GetHash64(Memory::GetPointer(entry->addr),
-                                       entry->memory_stride * entry->native_height, 0);
+        check_hash = CalculateTextureHash(
```
phire, line 1841: You should be able to simplify this to just `check_hash = entry->CalculateHash();`

If not, we have another bug.

phire (changes requested):

Source/Core/VideoCommon/TextureCacheBase.cpp
iwubcode: We are using `entry->memory_stride * entry->native_height` here though.  If `native_height` does not equal `NumBlocksY` then this will be different than:  https://github.com/dolphin-emu/dolphin/blob/master/Source/Core/VideoCommon/TextureCacheBase.cpp#L2905 .

phire:
Also, looking though texture cache code. I find it a little suspicious that we have separate stitching functions for xfb and efb textures (`StitchXFBCopy` and `DoPartialTextureUpdates`). It's a bad sign that neither function is generic enough. 

The original intention of merging the xfb cache into texture cache was so that only one function to stitch copies would be needed. That it would be generic and able to handle every single usecase. 

There is no fundamental difference between efb and xfb copies. They just have a different texture format.

phire:
Looks like I left a long comment on this topic back when the separate path was originally PRed: https://github.com/dolphin-emu/dolphin/pull/7412#issuecomment-427620197

Fixing the fundamental issue (that dolphin's texture cache was implemented at the level of whole textures and any kind of handling for non-matching transformations has been hacked on) probably requires major refactors.

Source/Core/VideoCommon/TextureCacheBase.cpp
iwubcode: Ah perfect.  That does simplify this a lot.  I'll make that change.  Thanks phire!

iwubcode:
> Also, looking though texture cache code. I find it a little suspicious that we have separate stitching functions for xfb and efb textures (StitchXFBCopy and DoPartialTextureUpdates). It's a bad sign that neither function is generic enough.

Yeah, when originally doing HybridXFB, there was a reason I couldn't consolidate the two completely.  Well and `DoPartialTextureUpdates` is a horrible name :P.

I believe the difference was in the NES zelda in Zelda Collector's Edition.  Something about there being some bounds difference?  Maybe there have been enough changes to warrant another look.

> Looks like I left a long comment on this topic back when the separate path was originally PRed

I remember that.  I agree, that would get rid of all the weird EFB specific checks we do right now.  All the mess in `GetTexture`.  Though, I think we'd have more work to do to handle hires textures.

I'd be interested in looking into that.  Maybe once all my enhancement features are finished.

phire:
> and `DoPartialTextureUpdates` is a horrible name :P.

Because when it was originally created, it did one thing only. Handle partial texture updates to the main texture atlas in NSMB Wii.

It's been expanded and has grown into a monster since then.

iwubcode:
@JMC47 - please retest when you have time!

Source/Core/VideoCommon/TextureCacheBase.cpp
```diff
@@ -1795,8 +1795,10 @@ TextureCacheBase::TCacheEntry* TextureCacheBase::GetXFBFromCache(u32 address, u3
 
     // The only thing which has to match exactly is the stride. We can use a partial rectangle if
     // the VI width/height differs from that of the XFB copy.
-    if (entry->is_xfb_copy && entry->memory_stride == stride && entry->native_width >= width &&
-        entry->native_height >= height && !entry->may_have_overlapping_textures)
+    // One way this can happen is if there were oversized XFB copies that got reduced on
+    // creation.  Example: Teenage Mutant Ninja Turtles (Wii)
+    if (entry->is_xfb_copy && entry->memory_stride == stride && entry->native_width != width &&
+        entry->native_height != height && !entry->may_have_overlapping_textures)
```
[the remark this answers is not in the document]
iwubcode: Good point.  Updated

JMC47:
This no longer fixes the games in the latest revision.

JMC47:
It is now fixing the games again.

phire (changes requested):

phire (approved):

JMC47:
Let's get this merged early in the month in case there are some unknown edge cases beyond what I'm aware of that crop up.

PR #9699, opened 2021-05-09
Fix backslash parsing issue in SplitPath

retroNUC:
From a bug mentioned here that I also ran into - https://forums.dolphin-emu.org/Thread-automatically-changing-discs-second-take?pid=508806#pid508806

For the "Change Discs Automatically" functionality, loading a .m3u file referencing multiple discs would allow both relative and absolute paths when loaded via GUI, but only absolute paths when using the -e/--exec command line option.

File paths provided via command line options could be using backslashes instead of forward slashes, which is the Windows path standard. This wasn't being considered in SplitPath, and caused issues down the line for functions that needed to re-append the folder path for relative file locations (such as ReadM3UFile in boot.cpp). Adding detection of the backslash character on _WIN32 fixes this.

First time contributing (and first time ever doing a pull request at all, actually) so please let me know if I've made any mistakes. Thank you!

### DIFF
--- a/Source/Core/Common/StringUtil.cpp
+++ b/Source/Core/Common/StringUtil.cpp
@@ -303,8 +303,9 @@ bool SplitPath(std::string_view full_path, std::string* path, std::string* filen
 
   size_t dir_end = full_path.find_last_of("/"
 // Windows needs the : included for something like just "C:" to be considered a directory
+// It also needs to parse escaped backslashes properly
 #ifdef _WIN32
-                                          ":"
+                                          ":\\"
 #endif
   );
   if (std::string::npos == dir_end)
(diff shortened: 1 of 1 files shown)


JosJuice:
Thanks for finding the cause! I tried to look into it once but didn't find anything, because I was assuming it was a problem with the relative path handling code, which had been written by me fairly recently before the reports came in...

I believe the reason why `SplitPath` doesn't handle backslashes properly is because we've decided to not use backslashes in paths internally in Dolphin. (See [FileSearch.cpp:150](https://github.com/dolphin-emu/dolphin/blob/eb5cd9be78c76b9ccbab9e5fbd1721ef6876cd68/Source/Core/Common/FileSearch.cpp#L150).) So I would suggest modifying the command-line argument parsing code to replace backslashes with forward slahes instead.

Alternatively, if we go with what you've currently implemented, you should remove the word "escaped" from the comment. A backslash separator in a path isn't escaped (but the representation of it in C++ code is).

retroNUC:
> I believe the reason why SplitPath doesn't handle backslashes properly is because we've decided to not use backslashes in paths internally in Dolphin.

Sure, makes sense to just make sure all platforms handle it the same way. I've corrected the code comment, but I'll take a look at the command line options code. Switching it early would make more sense if there's other places in code not handling backslashes.

PR #9700, opened 2021-05-09
Simplify the preview of control mappings by removing the device path from controls

Filoppi:
The picture should explain everything. Simply, if the mapping did not use a default device when it was made, its preview was completely "useless".
Splitting by "**`**" is 100% safe as special characters have a distinct and unique meaning in controls expressions, so there couldn't be any false positives from this code.
The only downside is not seeing what device a control corresponds to, but you can finally see the control at least, which is still better than only seeing (part of) the device name. We now add the prefix "**:**" if a control has got a device specifier in front of them.
We also remove white spaces and \n \t \r from the mapping preview.

image

**Extras:**
**bugfix**: fix EmulatedController::GetStateLock() not being acquired when reading the expression reference
**bugfix**: MappingButton::UpdateIndicator() calling State(0) on outputs, breaking ongoing rumbles if a game was running.
**Improvement**: make expressions previews appear in Italic if they failed to parse correctly (highlighted in **yellow** in the **screenshot**) (expressions are still evaluated and become bold if their state is "true")

### DIFF
--- a/Source/Core/DolphinQt/Config/Mapping/MappingButton.cpp
+++ b/Source/Core/DolphinQt/Config/Mapping/MappingButton.cpp
@@ -21,11 +21,48 @@
 
 constexpr int SLIDER_TICK_COUNT = 100;
 
-// Escape ampersands and remove ticks
-static QString ToDisplayString(QString&& string)
+// Escape ampersands and simplify the text for a short preview
+static QString RefToDisplayString(ControlReference* ref)
 {
-  return string.replace(QLatin1Char{'&'}, QStringLiteral("&&"))
-      .replace(QLatin1Char{'`'}, QString{});
+  const bool expression_valid =
+      ref->GetParseStatus() != ciface::ExpressionParser::ParseStatus::SyntaxError;
+  QString expression;
+  {
+    const auto lock = ControllerEmu::EmulatedController::GetStateLock();
+    expression = QString::fromStdString(ref->GetExpression());
+  }
+
+  // Split by "`" so that we can give a better preview of control,
+  // without including their device in front of them, which is usually
+  // too long to actually see the control.
+  QStringList controls = expression.split(QLatin1Char{'`'});
+  // Do try to simplify controls if the parsing had failed, as it might create false positives.
+  if (expression_valid)
+  {
+    for (int i = 0; i < controls.size(); i++)
+    {
+      // We have two ` for control so make sure to only consider the odd ones.
+      if (i % 2)
+      {
+        // Use the code from the ControlQualifier instead of duplicating it.
+        ciface::ExpressionParser::ControlQualifier qualifier;
+        qualifier.FromString(controls[i].toStdString());
+        // If the control has got a device specifier/path, add ":" in front of it, to make it clear.
+        controls[i] = qualifier.has_device ? QStringLiteral(":") : QString();
+        controls[i].append(QString::fromStdString(qualifier.control_name));
+      }
+    }
+  }
+  // Do not re-add "`" to the final string, we don't need to see it.
+  expression = controls.join(QStringLiteral(""));
+
+  expression.remove(QLatin1Char{' '});
+  expression.remove(QLatin1Char{'\t'});
+  expression.remove(QLatin1Char{'\n'});
+  expression.remove(QLatin1Char{'\r'});
+  expression.replace(QLatin1Char{'&'}, QStringLiteral("&&"));
+
+  return expression;
 }
 
 bool MappingButton::IsInput() const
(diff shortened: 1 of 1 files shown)


MayImilae:
I honestly disagree with this PR. Knowing that it is not the default device is very important to the point that I'm ok with it being clunky. Plus if you are overriding the default device anyway you already have been in the advanced menu and know to look there. It's a feature for advanced users, so, it's ok.

The original idea is hella clunky, oh yea, but I think we need to find a decent alternative rather than strip it out entirely.

Filoppi:
> I honestly disagree with this PR. Knowing that it is not the default device is very important to the point that I'm ok with it being clunky. Plus if you are overriding the default device anyway you already have been in the advanced menu and know to look there. It's a feature for advanced users, so, it's ok.
> 
> The original idea is hella clunky, oh yea, but I think we need to find a decent alternative rather than strip it out entirely.

Unfortunately your argument is invalid, as whether the device path is in front of the control or not does not mean the device isn't the default one now, it just means that it wasn't the default one when you mapped it.

Also, even if the above was true, what's the value in knowing to what device the first control of the expression belongs to?
All you want to do from the mapping window is hack a quick overview of everything.
Plus you might have two controls mapped into one expression/mapping, like **`DInput/0/KeyboardMouse:A`|`DSU/0/DualShot4:X`**, so from the preview you'd only see the first control anyway.

A few people complained about this on IRC and it's something I had also thought about in the past, so I decided to do it.

jordan-woyak:
I agree that trying to display the full "path" in the little button is silly but it should be made clear that it has an explicit device specification.
Maybe you should keep a leading backtick or something?

Filoppi:
> I agree that trying to display the full "path" in the little button is silly but it should be made clear that it has an explicit device specification.
> Maybe you should keep a leading backtick or something?

That does sound like a good compromise. Will do.

MayImilae:
Works for me 👍

Filoppi:
Actually the best design would be to add a prefix number to the list of device [0], [1], ... and add that number to the preview of the mapping if that control has got a device path in front of it. Not sure I will do that, as the **:A** is already a decent solution.

Filoppi:
@MayImilae @jordan-woyak Actually the best design would be to add a prefix number to the list of device [0], [1], ... and add that number to the preview of the mapping if that control has got a device path in front of it. Not sure I will do that, as the :A is already a decent solution.

jordan-woyak:
`:A` Sounds good to me too.

MayImilae:
I'm kind of more in favor of the `:A` option. Having the inputs numbered means it would be odd to not have the selected input numbered while having unselected numbered. Going with `:` as a simple notification of a device override is better, imo.

iwubcode:
Just personally, the ":" seems like it's really hard to see.

I was thinking what if we had some logic to ensure that the key was shown but then showed up to some maximum number of symbols.  We could "..." the rest.  Ex:

`DInput/0/.../B`

This could be used in other places in our GUI as well to denote longer strings than we have space for.

MayImilae:
> Just personally, the ":" seems like it's really hard to see.
> 
> I was thinking what if we had some logic to ensure that the key was shown but then showed up to some maximum number of symbols. We could "..." the rest. Ex:
> 
> `DInput/0/.../B`
> 
> This could be used in other places in our GUI as well to denote longer strings than we have space for.

Unfortunately I don't think that is any better than the current situation. So you can just enter anything into a controller slot and it'll just appear. It won't work, but you can see it. And when I put `Quartz/0/.../Up` I get this.

Screen Shot 2021-05-09 at 18 26 30

It's the same for Dinput.

Screen Shot 2021-05-09 at 18 26 57

It's just a very small space. It wouldn't achieve the desired goal of showing the input and not just the device.

It may vary depending on the device and the text size, but fortunately we can all just try it. Just type in `Xinput/.../A` or whatever and ignore the errors, and it will appear in the slot. It's good enough to see how it fits!

iwubcode:
> Unfortunately I don't think that is any better than the current situation. So you can just enter anything into a controller slot and it'll just appear. It won't work, but you can see it. And when I put Quartz/0/.../Up I get this.

Hmm.  Just to clarify my point.  I was thinking we would dynamically determine how much text was viewable and then based on that remove from the device name as much as necessary.  We'd always fit the key in.

So if we had `Gamepad B` and had 15 total characters available, we'd only have 6 characters left.  Subtract the 3 for "..." and that means the device name would show 3 characters:

```
XIn...Gamepad B
```
(you could tweak the algorithm if you always wanted a slash at the end)

But this would at least show that we're not binding a simple key to the current device.  It would also be more obvious than ":".

MayImilae:
Unfortunately the length isn't character based. Filling with `a` allows 10 characters, the 11th causes truncating. Using `A` allows 8 characters and the 9th triggers truncation. It's not a lot of characters to work with. :/

Filoppi:
@MayImilae done

JMC47:
Will this affect currently made controller profiles or configs?  If it makes people have to reconfigure things it'd be annoying to them.

Filoppi:
> Will this affect currently made controller profiles or configs? If it makes people have to reconfigure things it'd be annoying to them.

No, this just affects the way Qt prints some text on screen, serialization isn't affected.

JMC47 (approved):
LGTM

JMC47:
I've asked a couple of times in IRC if anyone else has any qualms with this and haven't gotten any responses.  This is an incredibly small change where I can mostly understand what the code is doing at a glance.

PR #9702, opened 2021-05-10
Controller Interface refactor

Filoppi:
Controller Interface and devices backends:
-Made devices population 100% safe as it had tons of race conditions. For example, under rare circumstances we could end up with duplicate devices or missing ones
-Made all suitable devices backend use "ControllerInterface::PlatformPopulateDevices()" to avoid for thread safety and avoid sending a ton of devices changed events
-Made most devices populations async, to increase performance and avoid hanging the Qt thread on manual devices refresh
-Added a sort priority method to devices, so we can prioritize some backends over others
-Avoid re-creating all devices when dolphin is being shut down
-Avoid the ControllerInterface sending Devices Changed events if devices haven't actually changed
-Fix not thread safe access to multiple variables in devices backends
-Fix DInput keyboard device failing to return the state when re-created (we avoid recreating the device now even when the render window changes)
-Improve a few things with DInput, which should make devices population and re-population safer. Also added comments regarding the remaining issues
-Remove the window handle from the OSX backend, as it wasn't used
-Fixed multiple race conditions in the DSU UDP backend
-Reworked the DSU UDP backend to never hang the Host thread for more than a really small time (e.g. when disabling DSU its thread now closes almost immediately)
-Fix DSU using a different UID for every controller of the same server (that's not how its supposed to work)
-Make sure DSU inputs start from their resting pose instead of 0
-Add devices disconnection detection to DSU, previously they'd stay there forever until manually refreshed
-Improve logging and add clarifying comments to all devices backends

Qt:
-Fix IOWindow keeping a shared ptr to devices even after them being removed by the ControllerInterface. This prevented some devices from being recreated correctly, as they were exclusive (e.g. DInput Joysticks)
-Other fixes for performance and thread safety
-Refactored the way IOWindow lists devices to make it clearer and hold onto disconnected devices.

This has been in the work for months and it is deeply tested.
Could use https://github.com/dolphin-emu/dolphin/pull/9688 to be merged first as it brings down the number of changes due to a log category rename.


Source/Core/InputCommon/ControllerInterface/DInput/DInputKeyboardMouse.cpp
```diff
@@ -230,6 +230,12 @@ std::string KeyboardMouse::GetSource() const
   return DINPUT_SOURCE_NAME;
 }
 
+// Give this device a higher priority to make sure it shows first
```
leoetlino, line 259: Other platforms (XInput2::KeyboardMouse, QuartzKeyboardAndMouse, Touch, possibly more) should receive the same treatment.

Source/Core/InputCommon/ControllerInterface/ControllerInterface.cpp
```diff
@@ -134,6 +134,28 @@ void ControllerInterface::RefreshDevices(bool because_of_window_change)
   // This wasn't thread safe in multiple device sources.
   std::lock_guard lk_population(m_devices_pupulation_mutex);
 
+#ifdef CIFACE_USE_WIN32
```
leoetlino, line 137: `#if defined(CIFACE_USE_WIN32) && !defined(CIFACE_USE_XLIB) && !defined(CIFACE_USE_OSX)`

Source/Core/InputCommon/ControllerInterface/Win32/Win32.cpp
```diff
@@ -19,20 +19,27 @@
 
 constexpr UINT WM_DOLPHIN_STOP = WM_USER;
 
-static Common::Event s_done_populating;
-static std::atomic<HWND> s_hwnd;
-static HWND s_message_window;
+static Common::Event s_received_device_change_event;
+static HWND s_hwnd;            // Dolphin's render window
```
leoetlino, line 23: Each of these comments should be *above* their respective lines to avoid clang-format reformatting these lines when new comments are added. e.g.

```cpp
// Dolphin's render window
static HWND s_hwnd;
```

Source/Core/InputCommon/ControllerInterface/Win32/Win32.cpp
```diff
@@ -19,20 +19,27 @@
...
+static HWND s_message_window;  // Windows messaging window (hidden)
 static std::thread s_thread;
+static std::atomic<bool> s_first_pupulate_devices_asked;
```
leoetlino, line 26: This should use Common::Flag.

leoetlino: Also there's a typo. (p**O**pulate)

Source/Core/InputCommon/ControllerInterface/Win32/Win32.cpp
```diff
@@ -19,20 +19,27 @@
...
 static LRESULT CALLBACK WindowProc(HWND hwnd, UINT message, WPARAM wparam, LPARAM lparam)
 {
   if (message == WM_INPUT_DEVICE_CHANGE)
   {
-    g_controller_interface.PlatformPopulateDevices([] {
-      ciface::DInput::PopulateDevices(s_hwnd);
-      ciface::XInput::PopulateDevices();
-    });
-    s_done_populating.Set();
+    // Windows automatically sends this message before we ask for it and "ready" to listen for it.
```
leoetlino, line 32: "and **before we are** ready to listen for it"

Source/Core/InputCommon/ControllerInterface/Win32/Win32.cpp
```diff
@@ -19,20 +19,27 @@
...
+    if (s_first_pupulate_devices_asked)
+    {
+      s_received_device_change_event.Set();
+      // TODO: we could easily use the message passed alongside this event, which tells
+      // whether a device was added or revoved, to avoid removing old, still connected, devices
```
leoetlino, line 37: revoved -> removed

leoetlino, line 37: "still connected, devices" -> "still connected devices"

Source/Core/InputCommon/ControllerInterface/Win32/Win32.cpp
```diff
@@ -125,10 +132,14 @@ void ciface::Win32::PopulateDevices(void* hwnd)
...
     s_hwnd = static_cast<HWND>(hwnd);
-    s_done_populating.Reset();
+    s_first_pupulate_devices_asked = true;
+    s_received_device_change_event.Reset();
+    // Do this forced devices refresh in the messaging thread so it won't cause any race conditions
     PostMessage(s_message_window, WM_INPUT_DEVICE_CHANGE, 0, 0);
-    if (!s_done_populating.WaitFor(std::chrono::seconds(10)))
-      ERROR_LOG_FMT(CONTROLLERINTERFACE, "win32 timed out when trying to populate devices");
+    std::thread([] {
+      if (!s_received_device_change_event.WaitFor(std::chrono::microseconds(10)))
```
leoetlino, line 140: 10µs seems very, very short for Windows.

Source/Core/InputCommon/ControllerInterface/DInput/DInputJoystick.cpp
```diff
@@ -60,37 +66,40 @@ void InitJoystick(IDirectInput8* const idi8, HWND hwnd)
...
+        // only add if it has some inputs/outputs.
+        // Don't even add it to our static list in case we first created it without a window handle,
+        // failing to get exclusive mode, and then later managed to obtain it, which mean it
+        // could now have some outputs if it didn't before.
         if (js->Inputs().size() || js->Outputs().size())
-          g_controller_interface.AddDevice(std::move(js));
+        {
+          if (g_controller_interface.AddDevice(std::move(js)))
+          {
+            std::lock_guard<std::mutex> lk(s_guids_mutex);
```
leoetlino, line 89: `std::lock_guard<std::mutex>` -> `std::lock_guard` (no need to write the template argument explicitly thanks to C++17's CTAD)

Source/Core/InputCommon/ControllerInterface/DInput/DInputKeyboardMouse.cpp
```diff
@@ -123,8 +123,12 @@ KeyboardMouse::KeyboardMouse(const LPDIRECTINPUTDEVICE8 kb_device,
 {
   s_keyboard_mouse_exists = true;
 
-  m_kb_device->Acquire();
-  m_mo_device->Acquire();
+  if (FAILED(m_kb_device->Acquire()))
```
leoetlino, line 126: This if statement should be braced because it spans more than one line.

Source/Core/InputCommon/ControllerInterface/DInput/DInputKeyboardMouse.cpp
```diff
@@ -123,8 +123,12 @@ KeyboardMouse::KeyboardMouse(const LPDIRECTINPUTDEVICE8 kb_device,
...
+    WARN_LOG_FMT(CONTROLLERINTERFACE,
+                 "Keyboard device failed to be acquired (we will retry later on)");
+  if (FAILED(m_mo_device->Acquire()))
```
leoetlino, line 129: Ditto

Source/Core/InputCommon/ControllerInterface/DInput/DInputKeyboardMouse.cpp
```diff
@@ -207,16 +213,15 @@ void KeyboardMouse::UpdateInput()
...
-  if (DIERR_INPUTLOST == kb_hr || DIERR_NOTACQUIRED == kb_hr)
-    m_kb_device->Acquire();
-
   if (DIERR_INPUTLOST == mo_hr || DIERR_NOTACQUIRED == mo_hr)
-    m_mo_device->Acquire();
-
-  if (SUCCEEDED(mo_hr))
+  {
+    INFO_LOG_FMT(CONTROLLERINTERFACE, "Mouse device failed to get state");
+    if (FAILED(m_mo_device->Acquire()))
```
leoetlino, line 220: ditto (braces for the if statement)

Source/Core/InputCommon/ControllerInterface/DInput/DInputKeyboardMouse.cpp
```diff
@@ -227,8 +232,17 @@ void KeyboardMouse::UpdateInput()
...
     // copy over the buttons
     std::copy_n(tmp_mouse.rgbButtons, std::size(tmp_mouse.rgbButtons), m_state_in.mouse.rgbButtons);
+  }
 
-    UpdateCursorInput();
+  HRESULT kb_hr = m_kb_device->GetDeviceState(sizeof(m_state_in.keyboard), &m_state_in.keyboard);
+  if (kb_hr == DIERR_INPUTLOST || kb_hr == DIERR_NOTACQUIRED)
+  {
+    INFO_LOG_FMT(CONTROLLERINTERFACE, "Keyboard device failed to get state");
+    if (SUCCEEDED(m_kb_device->Acquire()))
```
leoetlino, line 241: ditto (braces)

Source/Core/InputCommon/ControllerInterface/DualShockUDPClient/DualShockUDPClient.cpp
```diff
@@ -347,6 +352,8 @@ static void ConfigChanged()
 
   if (servers_enabled != s_servers_enabled || servers_setting != new_servers_setting)
   {
+    StopHotplugThread();  // Stop the thread before writing to s_servers
```
leoetlino, line 409: The comment should be above its associated line.

leoetlino (commented):
The DSU changes should probably be in their own PRs as they're not really related to refactors --  they're separate bug fixes and additions.

Source/Core/InputCommon/ControllerInterface/ControllerInterface.h
```diff
@@ -62,11 +62,18 @@ class ControllerInterface : public ciface::Core::DeviceContainer
 
   ControllerInterface() : m_is_init(false) {}
   void Initialize(const WindowSystemInfo& wsi);
-  void ChangeWindow(void* hwnd);
-  void RefreshDevices();
+  // Not thread safe
```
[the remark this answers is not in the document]
Filoppi: Yeah I meant that differently from the other methods, it can't be called concurrently by multiple threads. I made the comment clearer.

Source/Core/InputCommon/ControllerInterface/DInput/DInputKeyboardMouse.cpp
Filoppi: Mac and Linux Keyboard and Mouse are always first because they are not populated asynchronously, only windows (DInput/XInput) devices are. They will appear in the same order they always appeared, and the order is decided by the ControllerInterface backend "PopulateDevices()" calls.

Source/Core/InputCommon/ControllerInterface/Win32/Win32.cpp
Filoppi: Subjective, I think it's correct the way it is. Not a big deal I hope anyway...

Source/Core/InputCommon/ControllerInterface/Win32/Win32.cpp
Filoppi: Good spot. This was a mistake, it used to be 10 seconds, I didn't mean to change it by that much.

Source/Core/InputCommon/ControllerInterface/ControllerInterface.cpp
```diff
@@ -207,13 +234,48 @@ void ControllerInterface::Shutdown()
 #ifdef CIFACE_USE_DUALSHOCKUDPCLIENT
   ciface::DualShockUDPClient::DeInit();
 #endif
+
+  // Make sure no devices had been added within Shutdown() in the time
+  // between checking they checked atomic m_is_init bool and we changed it.
+  // We couldn't have locked m_devices_mutex nor m_devices_pupulation_mutex for the whole Shutdown()
+  // as they could cause deadlocks. Note that this is still not 100% safe as some backends are
+  // shutdown in other places, possibly adding devices after we have shut down, but the chances of
```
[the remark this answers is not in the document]
Filoppi: Grammatically correct but there are 3 or 4 other similar comments that already used shutdown as verb in the file, I'd keep consistency.

Source/Core/InputCommon/ControllerInterface/ControllerInterface.h
```diff
@@ -62,11 +62,18 @@ class ControllerInterface : public ciface::Core::DeviceContainer
...
+  void ChangeWindow(void* hwnd, bool is_exit = false);
```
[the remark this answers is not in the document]
Filoppi: There is only one call site and likely won't be anymore, plus it's false by default so it's not even specified in other places.
Sure it would be better, but I don't think its worth bothering honestly.

Filoppi:
> The DSU changes should probably be in their own PRs as they're not really related to refactors -- they're separate bug fixes and additions.

I've moved the DSU changes in https://github.com/dolphin-emu/dolphin/pull/9771 in separate commits. I guess you already reviewed them? There are no changes.

I've fixed everything else.

Source/Core/InputCommon/ControllerInterface/ControllerInterface.cpp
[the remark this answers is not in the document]
leoetlino: I brought it up because the very next sentence fragment (correctly) uses "shut down" rather than "shutdown", which makes the whole thing even more inconsistent. Might as well fix the grammar so that the new comment is *at least* consistent with itself.

Source/Core/InputCommon/ControllerInterface/ControllerInterface.cpp
```diff
@@ -319,6 +319,19 @@ bool ControllerInterface::AddDevice(std::shared_ptr<ciface::Core::Device> device
 
     NOTICE_LOG_FMT(CONTROLLERINTERFACE, "Added device: {}", device->GetQualifiedName());
     m_devices.emplace_back(std::move(device));
+
+    // We can't (and don't want) to control the order in which devices are added, but we
+    // need their order to be consistent, and we need the same one to always be the first, where
+    // present (the keyboard and mouse device usually). This is because when defaulting a
+    // controller profile, it will automatically select the first device in the list as its default.
+    std::sort(m_devices.begin(), m_devices.end(),
```
leoetlino, line 347: `std::sort` is not guaranteed to preserve the order of elements that are equivalent. `std::stable_sort` should be used here.

Filoppi:
@leoetlino done

leoetlino (approved):

ds22x:
Sorry to interject, but hotkeys don't work when having "render to main window" selected in the graphics settings.
Keyboard controller inputs work fine though, just the hotkeys fail to do anything.

JMC47:
That's why we merge these things after beta, so that people can report issues.  Also, I'm still surprised people like Render to Main.  I'm glad people are using it though, it helps maintain it.

Dentomologist:
> Sorry to interject, but hotkeys don't work when having "render to main window" selected in the graphics settings.
> Keyboard controller inputs work fine though, just the hotkeys fail to do anything.

If that's happening when fullscreen is enabled, [PR 9794](https://github.com/dolphin-emu/dolphin/pull/9794) has a fix for it.

ds22x:
https://github.com/dolphin-emu/dolphin/pull/9794 did indeed fix it.

JMC47:
Nice!

PR #9703, opened 2021-05-11
Fix serialization of control expressions with line breaks

Filoppi:
The control expression editor allows line breaks, but the serialization was losing anything after the first line break (/r /n).

Instead of opting to encode them and decode them on serialization,
which I tried but was not safe, as it would lose `\n` written in the expression by users,
I opted to replace them with a space, to at least maintain some of the spacing the user originally put in.
Completely removing them could also have caused the expression to be parsed differently.
This cannot cause any breaking differences on the expression, and its in any way better than before, as before part of the string could just get lost.

image

### DIFF
--- a/Source/Core/Common/StringUtil.cpp
+++ b/Source/Core/Common/StringUtil.cpp
@@ -219,7 +219,7 @@ std::string ArrayToString(const u8* data, u32 size, int line_len, bool spaces)
   return oss.str();
 }
 
-// Turns "  hello " into "hello". Also handles tabs.
+// Turns "\n\r\t hello " into "hello" (trims at the start and end but not inside).
 std::string_view StripSpaces(std::string_view str)
 {
   const size_t s = str.find_first_not_of(" \t\r\n");
@@ -239,6 +239,13 @@ std::string_view StripQuotes(std::string_view s)
     return s.substr(1, s.size() - 2);
   else
     return s;
+}
+
+// Turns "\n\rhello" into "  hello".
+void ReplaceBreaksWithSpaces(std::string& str)
+{
+  std::replace(str.begin(), str.end(), '\r', ' ');
+  std::replace(str.begin(), str.end(), '\n', ' ');
 }
 
 bool TryParse(const std::string& str, bool* const output)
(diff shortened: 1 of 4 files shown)


Source/Core/InputCommon/ControllerEmu/Setting/NumericSetting.h
```diff
@@ -115,7 +115,12 @@ class NumericSetting final : public NumericSettingBase
     if (IsSimpleValue())
       section.Set(group_name + m_details.ini_name, GetValue(), m_default_value);
     else
-      section.Set(group_name + m_details.ini_name, m_value.m_input.GetExpression(), "");
+    {
```
BhaaLseN, line 118: The `if` also needs curlies if you add them to the `else`.

Source/Core/Common/StringUtil.cpp
```diff
@@ -219,7 +219,7 @@ std::string ArrayToString(const u8* data, u32 size, int line_len, bool spaces)
   return oss.str();
 }
 
-// Turns "  hello " into "hello". Also handles tabs.
+// Turns "\n\r\t hello " into "hello".
```
BhaaLseN, line 222: Since you're changing the comment, maybe it should include "at the start/end but not inside". This appears to be what "trim" is in other languages, not "normalize space" (which is my immediate thought going by the name).

Filoppi:
> I know very little about this code but I'm also confused. I don't see why we can't just serialize the exact string the user provides and deserialize the exact string the user provides.
> 
> Then it just becomes a parsing problem (no different whether they load from save or type something in).

You can't serialize actual line breaks in a dolphin config. Every config is just one line. Or so it seems.

jordan-woyak:
Implementing typical backslash escaping will break current configs with device names containing a backslash unless the escaping is implemented at the expression parser level.

The real problem is our ini-file code being bad. It shouldn't be the job of input code to escape newlines so the config can be saved.
I think this is a good enough fix that can be easily removed if ini-files are ever fixed/removed.

jordan-woyak (approved):

lioncash (approved):

PR #9704, opened 2021-05-11
WIP: Don't inline Display Lists into Fifologs

phire:
Built on top of #9540 -- Only last two commits are relevant.

This is a WIP for feedback and comment. 

-----------

There are several reasons why I want to make this change:

1. To make it possible to investigate how games utilise display lists
2. I'm considering making dolphin cache display lists. 
   - Investigation would be useful before starting on this
   - Once implemented, it would be nice to have fifoci stress display list caching. 
3. This makes fifologs smaller (at least when uncompressed) 
4. This can theoretically speed up the capturing of fifologs, which could be useful for future "paused free-look" or "graphics debugger" functionality. 

**This PR is current a draft, some questions to consider before it can be ready:**

 - [ ] How do we want displaylists to be showed in the Analyzer UI?

In the current implementation, all displaylists are displayed after all objects for a frame. Some kind of interleaved or tree view might be better.

image

 - [ ] Should this be an optional capture mode, or the new default?

Unfortunately, old versions of dolphin can't quite read these new fifologs, but only because they have ASSERT(false) in their fifolog read path when they encounter a CALL DL command

 - [ ] How do we want first/last object to interact with displaylists?

Do we want to treat a call to a displaylist with multiple inner objects as a single object? Or inline the objects counts.

Currently, this PR doesn't support skipping objects inside Displaylists at all. They always draw (making it possible to see just how many objects a drawn via displaylists). But it's totally possible to implement skipping of individual objects within displaylists.

### DIFF
--- a/Source/Core/Core/FifoPlayer/FifoAnalyzer.cpp
+++ b/Source/Core/Core/FifoPlayer/FifoAnalyzer.cpp
@@ -258,37 +258,37 @@ u32 AnalyzeCommand(const u8* data, DecodeMode mode)
 
 void LoadCPReg(u32 subCmd, u32 value, CPMemory& cpMem)
 {
-  switch (subCmd & 0xF0)
+  switch (subCmd & CP_COMMAND_MASK)
   {
-  case 0x50:
+  case VCD_LO:
     cpMem.vtxDesc.low.Hex = value;
     break;
 
-  case 0x60:
+  case VCD_HI:
     cpMem.vtxDesc.high.Hex = value;
     break;
 
-  case 0x70:
-    ASSERT((subCmd & 0x0F) < 8);
-    cpMem.vtxAttr[subCmd & 7].g0.Hex = value;
+  case CP_VAT_REG_A:
+    ASSERT(subCmd - CP_VAT_REG_A < CP_NUM_VAT_REG);
+    cpMem.vtxAttr[subCmd & CP_VAT_MASK].g0.Hex = value;
     break;
 
-  case 0x80:
-    ASSERT((subCmd & 0x0F) < 8);
-    cpMem.vtxAttr[subCmd & 7].g1.Hex = value;
+  case CP_VAT_REG_B:
+    ASSERT(subCmd - CP_VAT_REG_B < CP_NUM_VAT_REG);
+    cpMem.vtxAttr[subCmd & CP_VAT_MASK].g1.Hex = value;
     break;
 
-  case 0x90:
-    ASSERT((subCmd & 0x0F) < 8);
-    cpMem.vtxAttr[subCmd & 7].g2.Hex = value;
+  case CP_VAT_REG_C:
+    ASSERT(subCmd - CP_VAT_REG_C < CP_NUM_VAT_REG);
+    cpMem.vtxAttr[subCmd & CP_VAT_MASK].g2.Hex = value;
     break;
 
-  case 0xA0:
-    cpMem.arrayBases[subCmd & 0xF] = value;
+  case ARRAY_BASE:
+    cpMem.arrayBases[subCmd & CP_ARRAY_MASK] = value;
     break;
 
-  case 0xB0:
-    cpMem.arrayStrides[subCmd & 0xF] = value & 0xFF;
+  case ARRAY_STRIDE:
+    cpMem.arrayStrides[subCmd & CP_ARRAY_MASK] = value & 0xFF;
     break;
   }
 }
--- a/Source/Core/Core/FifoPlayer/FifoAnalyzer.cpp
+++ b/Source/Core/Core/FifoPlayer/FifoAnalyzer.cpp
@@ -147,14 +147,14 @@ std::array<int, 21> CalculateVertexElementSizes(int vatIndex, const CPMemory& cp
 }
 }  // Anonymous namespace
 
-bool s_DrawingObject;
 FifoAnalyzer::CPMemory s_CpMem;
 
-u32 AnalyzeCommand(const u8* data, DecodeMode mode, bool in_display_list)
+u32 AnalyzeCommand(const u8* data, DecodeMode mode, bool in_display_list, CmdInfo *info_out)
 {
   const u8* dataStart = data;
 
   int cmd = ReadFifo8(data);
+  CmdInfo::Type type = CmdInfo::NORMAL;
 
   switch (cmd)
   {
@@ -165,8 +165,6 @@ u32 AnalyzeCommand(const u8* data, DecodeMode mode, bool in_display_list)
 
   case OpcodeDecoder::GX_LOAD_CP_REG:
   {
-    s_DrawingObject = false;
-
     u32 cmd2 = ReadFifo8(data);
     u32 value = ReadFifo32(data);
     LoadCPReg(cmd2, value, s_CpMem);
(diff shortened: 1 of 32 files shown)


Pokechu22:
<details><summary>Patch to get it to build on windows (may change more than actually needed)</summary>

```diff
diff --git a/Source/Core/Core/FifoPlayer/FifoPlaybackAnalyzer.cpp b/Source/Core/Core/FifoPlayer/FifoPlaybackAnalyzer.cpp
index 3e4f3f9e4a..2f69dfc947 100644
--- a/Source/Core/Core/FifoPlayer/FifoPlaybackAnalyzer.cpp
+++ b/Source/Core/Core/FifoPlayer/FifoPlaybackAnalyzer.cpp
@@ -143,7 +143,7 @@ void FifoPlaybackAnalyzer::AnalyzeFrames(FifoDataFile* file,
 
           if (it == std::end(analyzed.displayLists))
           {
-            u32 id = analyzed.displayLists.size();
+            u32 id = u32(analyzed.displayLists.size());
             analyzed.displayLists.emplace_back(DisplayList{id, info.display_list_address, 1, {}, {}});
             it = --std::end(analyzed.displayLists);
 
@@ -172,7 +172,7 @@ void FifoPlaybackAnalyzer::AnalyzeFrames(FifoDataFile* file,
 
       // Complete the final object
       if (!objects.empty() && objects.back().end == 0)
-        objects.back().end = length;
+        objects.back().end = u32(length);
     };
 
     AnalyzeCommands(frame.fifoData.data(), frame.fifoData.size(), AnalyzePrimaryList, analyzed.objects);
diff --git a/Source/Core/Core/FifoPlayer/FifoRecorder.cpp b/Source/Core/Core/FifoPlayer/FifoRecorder.cpp
index a5242a2725..93abab2dee 100644
--- a/Source/Core/Core/FifoPlayer/FifoRecorder.cpp
+++ b/Source/Core/Core/FifoPlayer/FifoRecorder.cpp
@@ -192,14 +192,14 @@ void FifoRecorder::EndFrame(u32 fifoStart, u32 fifoEnd)
     m_IsRecording = false;
 
 
-    u32 fifo_bytes = 0;
-    u32 update_bytes = 0;
-    u32 xf_bytes = 0;
-    u32 tex_bytes = 0;
-    u32 vert_bytes = 0;
-    u32 tmem_bytes = 0;
-    u32 dl_bytes = 0;
-    u32 update_overhead = 0;
+    size_t fifo_bytes = 0;
+    size_t update_bytes = 0;
+    size_t xf_bytes = 0;
+    size_t tex_bytes = 0;
+    size_t vert_bytes = 0;
+    size_t tmem_bytes = 0;
+    size_t dl_bytes = 0;
+    size_t update_overhead = 0;
 
     for (u32 i=0; i < m_File->GetFrameCount(); i++)
     {
diff --git a/Source/Core/DolphinQt/FIFO/FIFOAnalyzer.cpp b/Source/Core/DolphinQt/FIFO/FIFOAnalyzer.cpp
index 10fbf9e1ab..eac7b678bb 100644
--- a/Source/Core/DolphinQt/FIFO/FIFOAnalyzer.cpp
+++ b/Source/Core/DolphinQt/FIFO/FIFOAnalyzer.cpp
@@ -164,7 +164,7 @@ void FIFOAnalyzer::UpdateTree()
     }
 
     auto frame_info = FifoPlayer::GetInstance().GetAnalyzedFrameInfo(frame);
-    const u32 list_count = frame_info.displayLists.size();
+    const u32 list_count = u32(frame_info.displayLists.size());
     for (u32 list = 0; list < list_count; list++)
     {
       auto* list_item = new QTreeWidgetItem({tr("Display List %1").arg(list)});
```
</details>

How does this change interact with memory updates (specifically memory updates for things within a display list)?  Is that actually a concern?  (I could see a display list being used with e.g. vertex data in memory being changed between its uses; does that still work properly?)

phire:
> How does this change interact with memory updates 

Memory updates should all be handled.

When recording, display lists are thrown into the same `UseMemory` path as textures and vertex buffers which does a memcmp to check the memory backing the display list hasn't been updated, then adds a memory update if it has.

On Playback Analyze, I've added code to reconstruct the display list memory and special handling to make sure two lists at the same address but with different memory contents are logged independently.   
In the future it would be nice to show in the UI how the display lists are related and what bytes have changed.

> probably less confusing a script to convert the data formats

Ah yes, thats probably the way to go.

phire:
Some notes on size reduction:

### Wind Waker

*60 frames at outset island, immediately after loading from safe file*

 * Texture updates: 580KB
 * Vertex updates: 1.2MB
 * TMEM updates: 10KB
 * DisplayList updates: 730KB
 * Primary FIFO: 18MB

Before this PR, FIFO is instead 43MB.   
For the usecase of streaming a FIFO over network this is a bandwidth drop from 172mbit to 72mbit (before any compression)

### Mario Kart Wii

*60 frames Luigi circuit, starting line*

 * Texture updates: 2.3MB
 * Vertex updates: 2.5MB
 * XFmem updates: 530KB
 * DisplayList updates: 250KB
 * Primary FIFO: 3.7MB

Before this PR, FIFO is instead 17MB.   
For the usecase of streaming a FIFO over network this is a bandwidth drop from 136mbit to 30mbit (before any compression)

### THPS4

*60 frames Collage, starting position*

 * Texture updates: 2.8MB
 * Vertex updates: 39MB
 * TMEM updates: 2KB
 * DisplayList updates: 683KB
 * Primary FIFO: 9.7MB

Before this PR, FIFO is instead 57MB.   
This game will require something to be done about the vertex updates. Perhaps skinning is happening on the CPU, creating fresh vertices each frame.

### Super Smash Bros. Melee

*60 frames Hyrule Temple*

 * Texture updates: 1.6MB
 * Vertex updates: 1.5MB
 * Tmem updates: 24KB
 * DisplayList updates: 250KB
 * Primary FIFO: 11MB

Before this PR, FIFO is instead 26MB.   
For the usecase of streaming a FIFO over network this is a bandwidth drop from 208mbit to 88mbit (before any compression)

------------------------

All numbers are standing still. Moving around will likely cause more vertex and texture updates as object culling reveals different objects.

Source/Core/Core/FifoPlayer/FifoAnalyzer.cpp
```diff
@@ -199,12 +200,18 @@ u32 AnalyzeCommand(const u8* data, DecodeMode mode)
   }
 
   case OpcodeDecoder::GX_CMD_CALL_DL:
-    // The recorder should have expanded display lists into the fifo stream and skipped the call to
-    // start them
-    // That is done to make it easier to track where memory is updated
-    ASSERT(false);
-    data += 8;
+  {
+    const u32 address = ReadFifo32(data);
```
iwubcode: Do we want to bump the fifo version for this?  And then if the fifo version is older than the support for the DL, revert back to the previous behavior?

phire: Bump fifo version, yes. 

I don't really feel like reverting to the old behaviour of crashing when it encounters a display list call. Old files shouldn't have them anyway.

phire: If it wasn't for that single ASSERT(false) being in the read path, we wouldn't even need to bump the version number.

iwubcode: Yeah, I was more thinking about the case of reading an old fifo log.  I guess we're safe with old fifos because they embedded the display list values in them?

PR #7360, opened 2018-08-25
NetPlay host input authority mode

Techjar:
Currently, each player buffers their own inputs and sends them to the host. The host then relays those inputs to everyone else. Every player waits on inputs from all players to be buffered before continuing. What this means is all clients run in lockstep, and the total latency of inputs cannot be lower than the sum of the 2 highest client ping times in the game (in 3+ player sessions with people across the world, the latency can be very high).

Host input authority mode changes it so players no longer buffer their own inputs, and only send them to the host. The host stores only the most recent input received from a player. The host then sends inputs for all pads at the SI poll interval, similar to the existing code. If a player sends inputs to slowly, their last received input is simply sent again. If they send too quickly, inputs are dropped. This means that the host has full control over what inputs are actually read by the game, hence the name of the mode. Also, because the rate at which inputs are received by SI is decoupled from the rate at which players are sending inputs, clients are no longer dependent on each other. They only care what the host is doing. This means that they can set their buffer individually based on their latency to the host, rather than the highest latency between any 2 players, allowing someone with lower ping to the host to have less latency than someone else.

This is a catch to this: as a necessity of how the host's input sending works, the host has 0 latency. There isn't a good way to fix this, as input delay is now solely dependent on the real latency to the host's server. Having differing latency between players would be considered unfair for competitive play, but for casual play we don't really care. For this reason though, combined with the potential for a few inputs to be dropped on a bad connection, the old mode will remain and this new mode is entirely optional.

PR #9706, opened 2021-05-12
Netplay: Auto Pad Buffer Option

SirMangler:
Automatically set the pad buffer for users who do not know how to configure it or would rather automate the calculation used to estimate the optimal buffer size.

~~Calculation (No Host Input Authority):~~
~~`((total_latency * (number_of_players - 1)) / 15) + 1`~~

~~Calculation for Host Input Authority:~~
~~`(client_latency / 8) + 1`~~

Thanks to JMC for the calculations!

image

### DIFF
--- a/Source/Core/Core/NetPlayServer.cpp
+++ b/Source/Core/Core/NetPlayServer.cpp
@@ -1269,7 +1269,14 @@ bool NetPlayServer::StartGame()
 
   // no change, just update with clients
   if (!m_host_input_authority)
+  {
+    if (Config::Get(Config::NETPLAY_AUTO_BUFFER))
+    {
+      Config::SetBase(Config::NETPLAY_BUFFER_SIZE, CalculatePadBuffer());
+    }
+
     AdjustPadBufferSize(m_target_buffer_size);
+  }
 
   m_current_golfer = 1;
   m_pending_golfer = 0;
@@ -2157,5 +2164,42 @@ void NetPlayServer::ChunkedDataAbort()
   m_abort_chunked_data = true;
   m_chunked_data_event.Set();
   m_chunked_data_complete_event.Set();
+}
+
+// The calculations in this function are based off the discussions in this pr: 
+// https://github.com/dolphin-emu/dolphin/pull/9706#issuecomment-840262908
+u32 NetPlayServer::CalculatePadBuffer() const
+{
+  u32 total_latency = 0;
+  u32 player1_ping = 0;
+  u32 player2_ping = 0;
+
+  // Gather the two highest latencies of non-spectators.
+  for (const auto& player_entry : m_players)
+  {
+    // Check if the player has a mapped controller, otherwise they're a spectator.
+    const bool pad_mapped =
+      std::find(m_pad_map.begin(), m_pad_map.end(), player_entry.first) != m_pad_map.end();
+
+    const bool wiimote_mapped =
+      std::find(m_wiimote_map.begin(), m_wiimote_map.end(), player_entry.first) != m_wiimote_map.end();
+
+    if (pad_mapped || wiimote_mapped)
+    {
+      if (player_entry.second.ping > player1_ping)
+      {
+        player2_ping = player1_ping;
+        player1_ping = player_entry.second.ping;
+      }
+    }
+  }
+
+  // If the session only contains 2 players (1 host, 1 client),
+  // player2_ping will implicitly be 0.
+  total_latency = player1_ping + player2_ping;
+
+  constexpr u32 polling_rate = (1.0 / (120.0 / 2.0)) * 1000;
+
+  return (total_latency / polling_rate) + 1;
 }
 }  // namespace NetPlay
--- a/Source/Core/DolphinQt/NetPlay/NetPlayDialog.cpp
+++ b/Source/Core/DolphinQt/NetPlay/NetPlayDialog.cpp
@@ -7,6 +7,7 @@
 #include <QAction>
 #include <QActionGroup>
 #include <QApplication>
+#include <QCheckBox>
 #include <QClipboard>
 #include <QComboBox>
 #include <QGridLayout>
@@ -98,6 +99,8 @@ void NetPlayDialog::CreateMainLayout()
   m_main_layout = new QGridLayout;
   m_game_button = new QPushButton;
   m_start_button = new QPushButton(tr("Start"));
+  m_auto_buffer_box = new QCheckBox;
+  m_auto_buffer_label = new QLabel(tr("Auto Buffer:"));
   m_buffer_size_box = new QSpinBox;
   m_buffer_label = new QLabel(tr("Buffer:"));
   m_quit_button = new QPushButton(tr("Quit"));
@@ -192,10 +195,13 @@ void NetPlayDialog::CreateMainLayout()
   auto* options_widget = new QGridLayout;
 
   options_widget->addWidget(m_start_button, 0, 0, Qt::AlignVCenter);
-  options_widget->addWidget(m_buffer_label, 0, 1, Qt::AlignVCenter);
-  options_widget->addWidget(m_buffer_size_box, 0, 2, Qt::AlignVCenter);
-  options_widget->addWidget(m_quit_button, 0, 3, Qt::AlignVCenter | Qt::AlignRight);
-  options_widget->setColumnStretch(3, 1000);
+  options_widget->addWidget(m_auto_buffer_label, 0, 1, Qt::AlignVCenter);
+  options_widget->addWidget(m_auto_buffer_box, 0, 2, Qt::AlignVCenter);
+  options_widget->addWidget(m_buffer_label, 0, 3, Qt::AlignVCenter);
+  options_widget->addWidget(m_buffer_size_box, 0, 4, Qt::AlignVCenter);
+  options_widget->addWidget(m_quit_button, 0, 5, Qt::AlignVCenter | Qt::AlignRight);
+  options_widget->setColumnStretch(2, 50);
+  options_widget->setColumnStretch(5, 1000);
 
   m_main_layout->addLayout(options_widget, 2, 0, 1, -1, Qt::AlignRight);
   m_main_layout->setRowStretch(1, 1000);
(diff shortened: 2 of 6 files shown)


JosJuice:
Since the length in milliseconds of "one buffer" is different for SI devices (GameCube controllers) and Wii Remotes, I assume the calculation is only correct for one out of the two?

skylersaleh:
@dolphin-emu-bot rebuild

mimimi085181:
The formulas need the input polling rate in them as a factor. And, as far as i remember, the 2 formulas would need to be one for 2 players and one for 3/4 players. In 2 player mode, both talk directly to each other, so the buffer just needs to factor in the latency between both players. In 3 and 4 player mode, all players send their inputs to the host, which then sends them to the other players. So, for example, they have the latencies 0(host), 10, 30 and 70, the buffer would need to factor in the latency of 100, the worst case: sending the inputs from player 3 to player 4 or from 4 to 3. The latency of player 2 is irrelevant here, because their latency doesn't affect the latency of the communication from player 3 to 4. Keep in mind, i haven't looked at the code in ages, and this is just out of memory.

A suggestion for an alternative:
Instead of setting a fixed buffer size, set a target buffer length in ms. A nice auto-option on top of that wouldn't hurt of course. Every time the inputs are polled, each client calculates the technical buffer size at that time, which would be really useful if the polling rate can change midgame or is not known until the game is started. And it would be way easier to understand for users to set: 2 players: Set it to the latency between the 2 computers. 3 or 4 players: Set it to the sum of the 2 worst latencies between host and players. Maybe with the option to set the buffer size the old way, for people being used to it, or if games do stupid things and constantly change the polling rate in transitions or something.

Either way, the way i see it(like i said, i could be wrong), the formula would be:
buffersize = latency*pollingrate/1000
Then you could discuss whether to always round up, or round down but add 1, so that the buffer always holds inputs for the target time or a little longer. Without my suggestion, you'd insert the sum of the 2 worst latencies as the latency in the formula.

Oh, and btw, does Dolphin support gamecube and wii inputs at the same time for netplay? If yes, you'd either need 2 different buffer sizes for it work nicely(different pr...), or the suggested target buffer length in ms would just fix it as a side effect.

Techjar:
This should not be touching Host Input Authority at all. The buffer there is a client-side buffer akin to how video streaming sites buffer data, in order to smooth out the playback at the cost of additional latency. It should always be 1 unless the user is on a high jitter connection.

Techjar:
> Every time the inputs are polled, each client calculates the technical buffer size at that time, which would be really useful if the polling rate can change midgame or is not known until the game is started.

I'm not sure this is a good idea. It would likely have some undesirable side effects, like runaway buffer conditions or dropped inputs, because of the way that the buffer is padded or drained to adjust its size.

mimimi085181:
> I'm not sure this is a good idea. It would likely have some undesirable side effects, like runaway buffer conditions or dropped inputs, because of the way that the buffer is padded or drained to adjust its size.

Hmm, i guess you are right. Filling the buffer with empty inputs would be somewhat fine, but the other way around is more complicated than i thought.

Techjar:
> Oh, and btw, does Dolphin support gamecube and wii inputs at the same time for netplay?

Yes it does actually, though to my knowledge that's rarely done beyond Dokapon Kingdom. While I'm fond of the idea of having the buffer settings as milliseconds instead of raw size, the fact that polling rates can be variable makes this impossible. I also kind of prefer exposing the raw value, just because it gives the player more direct control. I don't really like the idea of having two buffer settings either, but I'm not sure of a better solution.

Source/Core/Core/NetPlayServer.cpp
```diff
@@ -1269,7 +1269,23 @@ bool NetPlayServer::StartGame()
...
+    if (Config::Get(Config::NETPLAY_AUTO_BUFFER))
+    {
+      unsigned int latency = 0;
+      for (const auto& player_entry : m_players)
+      {
+        latency += player_entry.second.ping;
+      }
+
+      const unsigned int pad_buffer =
+          ((latency * (static_cast<int>(m_players.size()) - 1)) / 15) + 1;
```
lioncash, line 1282: Maybe this calculation should be placed into a reusable function, since it's used once here in core code, and also used in the frontend code.

Note that this code also results in a signed/unsigned implicit conversion, since player size is converted to an int and then that side of the expression. Maybe make the signedness conversion explicit (or avoidable if possible)

Source/Core/Core/NetPlayServer.cpp
```diff
@@ -1269,7 +1269,23 @@ bool NetPlayServer::StartGame()
 
   // no change, just update with clients
   if (!m_host_input_authority)
+  {
+    if (Config::Get(Config::NETPLAY_AUTO_BUFFER))
+    {
+      unsigned int latency = 0;
```
lioncash, line 1275: Ideally, prefer using `u32` instead since it's much shorter

SirMangler:
@Techjar @mimimi085181 I appreciate the feedback on the calculations, I will talk to @JMC47 about what's been said and we will look at factoring in the extra considerations.

> This should not be touching Host Input Authority at all. The buffer there is a client-side buffer akin to how video streaming sites buffer data, in order to smooth out the playback at the cost of additional latency. It should always be <=1 unless the user is on a high jitter connection.

Like the standard pad buffer, my intention was only to assign that value automatically using the same method an experienced user would. For clarification, are you saying that with the exception of unique scenarios, the value should be 0? If this is the case I will remove it and leave it to manual configuration. If not, I would appreciate input on a better calculation.

Source/Core/Core/NetPlayServer.cpp
SirMangler: This is a good point, and per the next comment, I forgot to use u32. I will look at a cleaner implementation.

Techjar:
> For clarification, are you saying that with the exception of unique scenarios, the value should be 0?

Yeah, the default value is pretty much always what you want. There's no need to auto-configure it to a higher value, as the "padding" comes from the actual link latency, since there's no lockstep and the client is strictly using inputs from the host. The client can lag behind in time from the host, and will unlimit the emulation speed to try to catch up until the buffer gets down to the specified value. You can find a more detailed description of how it works in the original PR #7360.

SirMangler:
Pushed for further testing and review. We have found that with 2 players, the pad buffer doesn't need to be changed with an increased polling rate (tested with wiimotes mapped and gamecube mapped independently). For now I am factoring in a 120Hz rate, and have taken into account the rest of the suggested changes. The brief testing I have done with JMC has shown this works quite well, but ideally we should test this with a full lobby.

Source/Core/Core/NetPlayServer.cpp
```diff
@@ -2158,4 +2165,41 @@ void NetPlayServer::ChunkedDataAbort()
   m_chunked_data_event.Set();
   m_chunked_data_complete_event.Set();
 }
+
+// The calculations in this function are based off the discussions in this pr: 
+// https://github.com/dolphin-emu/dolphin/pull/9706#issuecomment-840262908
+u32 NetPlayServer::CalculatePadBuffer() const
+{
+  u32 total_latency = 0;
```
Techjar, line 2173: This declaration can be moved down to its useful assignment.

Source/Core/DolphinQt/NetPlay/NetPlayDialog.cpp
```diff
@@ -98,6 +99,8 @@ void NetPlayDialog::CreateMainLayout()
   m_main_layout = new QGridLayout;
   m_game_button = new QPushButton;
   m_start_button = new QPushButton(tr("Start"));
+  m_auto_buffer_box = new QCheckBox;
+  m_auto_buffer_label = new QLabel(tr("Auto Buffer:"));
```
Techjar, line 103: Let's get rid of this separate label and use the QCheckBox's native label instead, like the rest of the UI code. :)

```c++
m_auto_buffer_box = new QCheckBox(tr("Auto Buffer"));
```

Also not sure whether I prefer the checkbox to the left or right side of the spinbox.

Source/Core/DolphinQt/NetPlay/NetPlayDialog.cpp
```diff
@@ -292,6 +298,24 @@ void NetPlayDialog::ConnectWidgets()
...
+    if (value)
+    {
+      auto server = Settings::Instance().GetNetPlayServer();
+      if (server && !m_host_input_authority)
+      {
+        const u32 pad_buffer = server->CalculatePadBuffer();
+
+        Config::SetBase(Config::NETPLAY_BUFFER_SIZE, pad_buffer);
+        server->AdjustPadBufferSize(pad_buffer);
+        m_buffer_size_box->setValue(pad_buffer);
```
Techjar, line 314: I'm pretty sure you can remove the `SetBase` and `AdjustPadBufferSize` calls, as `setValue` should fire the `valueChanged` signal that already does the right thing.

Source/Core/DolphinQt/NetPlay/NetPlayDialog.cpp
```diff
@@ -292,6 +298,24 @@ void NetPlayDialog::ConnectWidgets()
           [this] { m_chat_send_button->setEnabled(!m_chat_type_edit->text().isEmpty()); });
 
   // Other
+  connect(m_auto_buffer_box, &QCheckBox::clicked, this, [this](bool value) {
```
Techjar, line 301: `&QCheckBox::toggled`

Also please make the bool `checked` instead of `value` for consistency.

Techjar (changes requested):

Source/Core/DolphinQt/NetPlay/NetPlayDialog.cpp
```diff
@@ -945,17 +969,20 @@ void NetPlayDialog::OnHostInputAuthorityChanged(bool enabled)
 
     if (is_hosting)
     {
-      m_buffer_size_box->setEnabled(enable_buffer);
       m_buffer_label->setEnabled(enable_buffer);
```
Techjar, line 972: Removing this isn't the correct way, the logic should instead be amended to also check for auto buffer being enabled, otherwise the buffer spinbox will get enabled when it's not supposed to be.

Source/Core/DolphinQt/NetPlay/NetPlayDialog.cpp
```diff
@@ -1089,6 +1117,8 @@ void NetPlayDialog::LoadSettings()
   const bool golf_mode_overlay = Config::Get(Config::NETPLAY_GOLF_MODE_OVERLAY);
 
   m_buffer_size_box->setValue(buffer_size);
+  m_buffer_size_box->setEnabled(!auto_buffer);
```
Techjar, line 1120: This doesn't make sense to have here because the state depends on session-specific variables. The other call I reviewed should take care of it.

Techjar (changes requested):

jordan-woyak:
The manual buffer size is not ideal but this PR has aged and has some unaddressed comments. I'm closing this for now, but we can re-open on new information.

PR #9707, opened 2021-05-13
Remove Atomic.h

JosJuice:
The STL has everything we need nowadays.

I have tried to not alter any behavior or semantics with this change wherever possible. In particular, `WriteLow` and `WriteHigh` in  `CommandProcessor` retain the ability to accidentally undo another thread's write to the upper half or lower half respectively. If that should be fixed, it should be done in a separate commit for clarity. One thing did change: The places where we were using `+=` on a volatile variable (not an atomic operation) are now using `fetch_add` (actually an atomic operation).

Tested with single core and dual core on x86-64 and AArch64.


lioncash:
You absolutely love to see it =). I think with this, we finally have most of the volatile occurrences removed from the codebase (aside from one in DSP code off the top of my head)

JosJuice:
I've actually only removed 7 out of the 16 volatiles in `SCPFifoStruct`. The remaining ones shouldn't be hard to get rid of, but I'll handle them separately so the PR doesn't get too big.

Source/Core/VideoCommon/CommandProcessor.cpp
```diff
@@ -91,21 +90,15 @@ void DoState(PointerWrap& p)
   p.Do(s_interrupt_waiting);
 }
 
-static inline void WriteLow(volatile u32& _reg, u16 lowbits)
+static inline void WriteLow(std::atomic<u32>& reg, u16 lowbits)
 {
-  Common::AtomicStore(_reg, (_reg & 0xFFFF0000) | lowbits);
+  reg.store((reg.load(std::memory_order_relaxed) & 0xFFFF0000) | lowbits,
```
BhaaLseN, line 95: Just wondering, is this _still_ atomic in the end (with a seperate `load` then `store`)?
_Then again, thats what the old code did, so same issue there...?_

JosJuice: This is indeed not atomic. I mentioned this in the PR description.

BhaaLseN: Fair enough, I read that and dismissed it from my head while looking at `LowPart`/`HighPart` (only remembering the `Low`/`High` part).

leoetlino (approved):

lioncash (approved):

PR #9708, opened 2021-05-13
DSP: Change external_interrupt_waiting from volatile to atomic

JosJuice:
Making this volatile accomplishes... Well, nothing in practice. Making it atomic, on the other hand, lets us enforce a memory ordering.

### DIFF
--- a/Source/Core/Core/DSP/DSPCore.cpp
+++ b/Source/Core/Core/DSP/DSPCore.cpp
@@ -195,7 +195,7 @@ void SDSP::SetException(ExceptionType exception)
 
 void SDSP::SetExternalInterrupt(bool val)
 {
-  external_interrupt_waiting = val;
+  external_interrupt_waiting.store(val, std::memory_order_release);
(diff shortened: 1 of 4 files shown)


lioncash:
Will this work for the DSP JIT? Its offset is used in `M_SDSP_external_interrupt_waiting` to perform a CMP [here](https://github.com/dolphin-emu/dolphin/blob/5513d5f4f732fb1e436765ab87e7d60ba02b1ad6/Source/Core/Core/DSP/Jit/x64/DSPEmitter.cpp#L443), and isn't guaranteed to be 8 bits in size any longer

JosJuice:
Wait, it's used there? Why didn't Visual Studio's "Find Usages" tell me about that usage :/

I'll put a static assert on it to ensure that `std::atomic<bool>` is 8-bit and lock-free.

leoetlino (approved):

lioncash (approved):

PR #9715, opened 2021-05-15
DolphinQt: ignore warnings from qt headers on buildbot

iwubcode:
Uses the [experimental external feature](https://devblogs.microsoft.com/cppblog/broken-warnings-theory/) to ignore warnings in header.  This is experimental right now but the [latest preview release notes](https://docs.microsoft.com/en-us/visualstudio/releases/2019/release-notes-preview#16.10.0.pre.3.0) seem to indicate it's going out of experimental status.

This should fix the build errors we're seeing.

I wasn't sure if there was a better place for this.  Also, ideally this should be applied to _all_ our external library headers but that's a much larger change.

### DIFF
--- a/Source/VSProps/QtCompile.props
+++ b/Source/VSProps/QtCompile.props
@@ -28,10 +28,11 @@
       <AdditionalIncludeDirectories>$(QtIncludeDir)QtWidgets;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
       <AdditionalIncludeDirectories Condition="'$(Platform)'=='ARM64'">$(QtIncludeDir)QtANGLE;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
       <!--
-      Negate the previously enabled warning (set in Base.props). Not compatible with QtCore\qtmap.h
-      This isn't too bad since we live with the fact that Qt-using projects will have different compiler args already.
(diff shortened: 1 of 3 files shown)


iwubcode:
Updated to also fix the Visual Studio solution files.

And since I upgraded VS locally, I noticed a warning treated as an error with a qt file "assignment within conditional expression".  I fixed that as well.

Source/VSProps/QtCompile.props
```diff
@@ -32,6 +32,12 @@
       This isn't too bad since we live with the fact that Qt-using projects will have different compiler args already.
       -->
       <AdditionalOptions>%(AdditionalOptions) /wd4946</AdditionalOptions>
```
Pokechu22, line 34: Is this line still needed if all warnings are suppressed?  (Same in CMakeLists.txt)

iwubcode: Is that for the qt headers or for the project itself?  I assumed it was for the project itself.

iwubcode: Looks like it is for the headers.  Going to try it and confirm

shuffle2:
fwiw here is qt tracking bug https://bugreports.qt.io/browse/QTBUG-93748

lioncash (approved):

PR #9717, opened 2021-05-16
Fix Android controller mappings accidentally being in IsSettingSaveable

JosJuice:
`Settings.SECTION_INI_ANDROID` and `Settings.SECTION_BINDINGS` both have the value `"Android"`, but we only want the former to be marked as being handled by the new config system.

This change fixes a problem where controller settings were not being properly saved to Dolphin.ini.


JMC47:
This fixes the longstanding controller configuration issues on Android.

JosJuice:
Nah, this only fixes one of the several issues :P

The one it fixes is arguably the biggest issue, though.

lioncash (approved):

PR #9718, opened 2021-05-16
Fifo analyzer improvements, part 3

Pokechu22:
Features:
* GP opcodes are decoded in a single place, `OpcodeDecoding.cpp`, instead of also being decoded for the FIFO player and recorder and in the FIFO analyzer.
* The new decoder uses a callback which has different methods for each type of decoded command.  The callback also has a function that gets a reference to the current CP state (and updates that for CP commands), so that it can properly decode vertices.  It also can handle display lists decently well (by re-using the same callback).
* Commands after the final object (the copy to the XFB) is now shown.  Other copies also create new objects.  Unfortunately, this affects both the object range and the analyzer view; since the objects for copies don't have any vertices, excluding or including them in the range has no effect.  But it makes analyzing a bit easier.
* Object info for each frame is now stored in a vector of structs containing the start position, offset to primitive data, size, and CP memory state.  This, combined with the extra final object, simplifies the playback code a fair bit.

-----

This is currently a draft to make sure this builds on all platforms.  I still need to do a _lot_ of cleanup, and may use `EnumMap` in a few more places and/or convert some other things (e.g. the BP and XF enums) to `enum class`es.  Feedback is still appreciated (in particular, I have mixed feelings about what I needed to do to get nested `EnumMap`s in the various VertexLoaders, and opinions on the whole OpcodeDecoding callback structure would also be useful).


Source/Core/Common/EnumMap.h
```diff
@@ -0,0 +1,68 @@
...
+#include <type_traits>
+
+#include "Common/TypeUtils.h"
+
+template <std::size_t position, std::size_t bits, typename T, typename StorageType>
+struct BitField;
+
+// A type that allows lookup of values associated with an enum as the key.
+// Designed for enums whose numeric values start at 0 and increment continuously with few gaps.
+template <typename V, auto last_member>
```
lioncash, line 17: This should be placed in the Common namespace.

Maybe also a unit test or two as well in the tests.

Source/Core/Core/FifoPlayer/FifoPlayer.cpp
```diff
@@ -31,6 +31,159 @@
...
 
+namespace
+{
+class FifoPlaybackAnalyzer : private OpcodeDecoder::Callback
+{
+public:
+  static void AnalyzeFrames(FifoDataFile* file, std::vector<AnalyzedFrameInfo>& frame_info);
+
+private:
+  FifoPlaybackAnalyzer(const u32* cpmem) : m_cpmem(cpmem) {}
```
lioncash, line 42:
```suggestion
  explicit FifoPlaybackAnalyzer(const u32* cpmem) : m_cpmem(cpmem) {}
```

Source/Core/Core/FifoPlayer/FifoRecorder.h
```diff
@@ -48,6 +50,37 @@ class FifoRecorder
   static FifoRecorder& GetInstance();
 
 private:
+  class FifoRecordAnalyzer : public OpcodeDecoder::Callback
+  {
+  public:
+    FifoRecordAnalyzer(FifoRecorder* owner) : m_owner(owner) {}
```
lioncash, line 56:
```suggestion
    explicit FifoRecordAnalyzer(FifoRecorder* owner) : m_owner(owner) {}
```

Source/Core/Core/FifoPlayer/FifoRecorder.h
```diff
@@ -48,6 +50,37 @@ class FifoRecorder
...
+    FifoRecordAnalyzer(FifoRecorder* owner, const u32* cpmem) : m_owner(owner), m_cpmem(cpmem) {}
```
lioncash, line 57:
```suggestion
    explicit FifoRecordAnalyzer(FifoRecorder* owner, const u32* cpmem) : m_owner(owner), m_cpmem(cpmem) {}
```

Source/Core/Core/FifoPlayer/FifoRecorder.h
```diff
@@ -48,6 +50,37 @@ class FifoRecorder
...
+      // Should have been inlined by the recorder
+      ASSERT(false);
+    }
+    void OnNop(u32 count) override {}
+    void OnUnknown(u8 opcode, const u8* data) override {}
+
+    CPState& GetCPState() override { return m_cpmem; }
+
+  private:
+    const FifoRecorder* m_owner;
```
lioncash, line 76: This should be below with the other member variable

Source/Core/DolphinQt/FIFO/FIFOAnalyzer.cpp
```diff
@@ -164,24 +173,124 @@ void FIFOAnalyzer::UpdateTree()
...
 }
 
-static std::string GetPrimitiveName(u8 cmd)
+namespace
+{
+class DetailCallback : public OpcodeDecoder::Callback
 {
-  if ((cmd & 0xC0) != 0x80)
+public:
+  DetailCallback(CPState cpmem) : m_cpmem(cpmem) {}
```
lioncash, line 181:
```suggestion
  explicit DetailCallback(CPState cpmem) : m_cpmem(cpmem) {}
```

Source/Core/DolphinQt/FIFO/FIFOAnalyzer.cpp
```diff
@@ -164,24 +173,124 @@ void FIFOAnalyzer::UpdateTree()
...
+  {
+    const auto [desc, written] = GetXFIndexedLoadInfo(array, index, address, size);
+    text = QStringLiteral("LOAD INDX %1   %2")
+               .arg("ABCD"[array - ARRAY_XF_A])
+               .arg(QString::fromStdString(desc));
+  }
+  void OnPrimitiveCommand(OpcodeDecoder::Primitive primitive, u8 vat, u32 vertex_size,
+                          u16 num_vertices, const u8* vertex_data) override
+  {
+    const auto name = fmt::to_string(primitive);
```
lioncash, line 233:
```suggestion
    const auto name = QString::number(primitive);
```

Avoids the need for conversion from std::string to QString.

Source/Core/DolphinQt/FIFO/FIFOAnalyzer.cpp
```diff
@@ -164,24 +173,124 @@ void FIFOAnalyzer::UpdateTree()
...
+    {
+      text += QStringLiteral("%1").arg(vertex_data[i], 2, 16, QLatin1Char('0'));
+    }
+#endif
+  }
+
+  void OnDisplayList(u32 address, u32 size) override
+  {
+    // Should have been inlined by the recorder
+    ASSERT(false);
```
lioncash, line 263: Note that this can cause false-positives with code relating to static analysis, since this will always be executed and consider the following code as dead.

Source/Core/DolphinQt/FIFO/FIFOAnalyzer.cpp
```diff
@@ -538,156 +483,159 @@ void FIFOAnalyzer::ShowSearchResult(size_t index)
...
-  const u32 entry_start = m_object_data_offsets[entry_nr];
-
-  const u8* cmddata = &fifo_frame.fifoData[object_start + entry_start];
-
-  // TODO: Not sure whether we should bother translating the descriptions
+// TODO: Not sure whether we should bother translating the descriptions
+class DescriptionCallback : public OpcodeDecoder::Callback
+{
+public:
+  DescriptionCallback(const CPState& cpmem) : m_cpmem(cpmem) {}
```
lioncash, line 492:
```suggestion
  explicit DescriptionCallback(const CPState& cpmem) : m_cpmem(cpmem) {}
```

Source/Core/DolphinQt/FIFO/FIFOAnalyzer.cpp
```diff
@@ -538,156 +483,159 @@ void FIFOAnalyzer::ShowSearchResult(size_t index)
...
         text += QLatin1Char{' '};
       }
     }
   }
-  else
+
+  void OnDisplayList(u32 address, u32 size) override
+  {
+    // Should have been inlined by the recorder
+    ASSERT(false);
```
lioncash, line 595: Ditto

Source/Core/VideoCommon/CPMemory.h
```diff
@@ -621,32 +622,29 @@ class VertexLoaderBase;
...
-  VAT vtx_attr[CP_NUM_VAT_REG];
+  std::array<VAT, CP_NUM_VAT_REG> vtx_attr;
 
-  // Attributes that actually belong to VertexLoaderManager:
-  BitSet32 attr_dirty;
-  bool bases_dirty;
-  VertexLoaderBase* vertex_loaders[CP_NUM_VAT_REG];
-  int last_id;
+  CPState() = default;
+  CPState(const u32* memory);
```
lioncash, line 634:
```suggestion
  explicit CPState(const u32* memory);
```

Source/Core/VideoCommon/CPMemory.h
```diff
@@ -621,32 +622,29 @@ class VertexLoaderBase;
...
   // Most games only use the first VtxAttr and simply reconfigure it all the time as needed.
-  VAT vtx_attr[CP_NUM_VAT_REG];
+  std::array<VAT, CP_NUM_VAT_REG> vtx_attr;
 
-  // Attributes that actually belong to VertexLoaderManager:
-  BitSet32 attr_dirty;
-  bool bases_dirty;
-  VertexLoaderBase* vertex_loaders[CP_NUM_VAT_REG];
-  int last_id;
+  CPState() = default;
```
lioncash, line 633: Functions go above member variables.

Source/Core/VideoCommon/VertexLoaderManager.h
```diff
@@ -51,4 +58,10 @@ extern u32 position_matrix_index[4];
 
 // VB_HAS_X. Bitmask telling what vertex components are present.
 extern u32 g_current_components;
+
+extern BitSet8 g_main_vat_dirty, g_preprocess_vat_dirty;
```
lioncash, line 62: Ideally place variables on their own line to clearly separate them from one another for reading.

Source/Core/VideoCommon/VertexLoaderManager.cpp
```diff
@@ -49,14 +45,19 @@ static std::mutex s_vertex_loader_map_lock;
 static VertexLoaderMap s_vertex_loader_map;
 // TODO - change into array of pointers. Keep a map of all seen so far.
 
-u8* cached_arraybases[NUM_VERTEX_COMPONENT_ARRAYS];
+std::array<u8*, NUM_VERTEX_COMPONENT_ARRAYS> cached_arraybases;
+
+BitSet8 g_main_vat_dirty, g_preprocess_vat_dirty;
+bool g_bases_dirty;  // Main only
+u8 g_current_vat;    // Main only
+std::array<VertexLoaderBase*, CP_NUM_VAT_REG> g_main_vertex_loaders, g_preprocess_vertex_loaders;
```
lioncash, line 53: Ditto

Source/Core/VideoCommon/VertexLoaderManager.cpp
```diff
@@ -49,14 +45,19 @@ static std::mutex s_vertex_loader_map_lock;
 static VertexLoaderMap s_vertex_loader_map;
 // TODO - change into array of pointers. Keep a map of all seen so far.
 
-u8* cached_arraybases[NUM_VERTEX_COMPONENT_ARRAYS];
+std::array<u8*, NUM_VERTEX_COMPONENT_ARRAYS> cached_arraybases;
+
+BitSet8 g_main_vat_dirty, g_preprocess_vat_dirty;
```
lioncash, line 50: Ditto

Source/Core/Core/FifoPlayer/FifoRecorder.h
```diff
@@ -48,6 +50,37 @@ class FifoRecorder
...
+    void OnXF(u16 address, u8 count, const u8* data) override {}
+    void OnCP(u8 command, u32 value) override {}
+    void OnBP(u8 command, u32 value) override {}
+    void OnIndexedLoad(u8 array, u32 index, u16 address, u8 size) override;
+    void OnPrimitiveCommand(OpcodeDecoder::Primitive primitive, u8 vat, u32 vertex_size,
+                            u16 num_vertices, const u8* vertex_data) override;
+    void OnDisplayList(u32 address, u32 size) override
+    {
+      // Should have been inlined by the recorder
+      ASSERT(false);
```
lioncash, line 68: Maybe implement this in the cpp file, so the assert header doesn't need to be included in another header.

lioncash (commented):
Just a very preliminary review. Feel free to disregard any that would otherwise be addressed during cleanup

Source/Core/DolphinQt/FIFO/FIFOAnalyzer.cpp
Pokechu22: Does that apply do Dolphin's `ASSERT`?  I'm pretty sure that Dolphin's one will return normally if the user chooses to ignore it.  (I can definitely see it causing problems with C's `abort` which is called by `assert` when it's enabled, since `abort` does not return)

Pokechu22:
One other thing I did was switch from

```c++
template <auto last_member, typename T = decltype(last_member),
          size_t size = static_cast<size_t>(last_member) + 1,
          std::enable_if_t<std::is_enum_v<T>, bool> = true>
class EnumFormatter {}
```

to

```c++
template <auto last_member>
class EnumFormatter
{
  using T = decltype(last_member);
  static_assert(std::is_enum_v<T>);
}
```

but it turns out that that causes issues with GCC prior to version 8; it treats different enums with the same numeric value as the same and then can't find the right things in the template ([minimal case](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAM1QDsCBlZAQwBtMQBGAFlJvoCqAZ0wAFAB4gA5AAYppAFZdSrZrVDIApACYAQjt2kR7ZATx1KmWugDCqVgFcAtrRDaA7KSvoAMnlqYAHLOAEaYxCAAzKQADqhChOa0do4ubp5xCWZ0fgHBTmER0caYpkkMBMzEBCnOrh5GmCbZtBVVBLlBoeFRRpXVtWkNQv0d/l0FPZEAlEaoDsTIHFI6kf7IjlgA1JqRNgQAnjGYAPoExMyEQrvYmjIAgnf3BJhOMaovO3vMDkRbqiMTk5XoUbk8RsQHKYtrpmCIdu59A8tlsHAl1FsACpfAAiWywG0OxwgAIIQJB4WmuyR9xRI2YZmQJzhImqEBG6BAIDwQhOVmcJwAbrsbJiblTIjTNO4cdSni83h9MF99kcrMxgViwQ8NiytjjwnhBZh0HKHk9%2BU4trqhEItgAxVCoLYgLb%2BAgImko2m7PEyUhPFG6XFbTgB80ys3PV7vBnKkXa%2B4QqEeg3EI0mkWO1A3F0wuGYLNOrn6SLYT3S2WSp4W2jOa0Au2w4h590V5FbGwh/2B/UhsM1yPVh4K2OfBNl8HnFP6w3G017Zu512wkQi5tcqvl6X6IdSh5tpyXWgQabt2lbemM5m28IEdkETnc3lCDWnYV7NMZhc2bM3LmYqQ%2BZrns2Ylv%2BIBimWEpepelTXiyd4Pk%2BPInK%2BwJCiKX7zuuVQQYBwGFouVSbvh4pRnS8F4EyiFshyXKoeh764SBv7FiApbYABQF/tBFFwQy1E3qy970c%2BaFvphxGsRuIBbtx%2BbEORw4XleQm0aJj4MS%2BkkfjY2GZqBTr4Txxl8SplGCTRt50Vp4lMVJ%2BlzoZNhLmWCludgMGDrKDxSLMrDSAArPIriyPIqDSDYBgGJe8yLPG2iRJw8gENIcjTLMADWIDcAAHAAdAAbNwMjuDIQWREVkTlclQVBSo0jcKF6URdI8hCCA/ppeFsxwLAMCICgqBvHg7BkBQEBoKN40gMA7gNVQY0vMQnUQCErWkCE/hVAc0gpaQ03AvQADytCsHt4WkFgR7qOwm34MQpRmManVXZg4ilL8ywHe6TSbaweAhBcxAHHYWCbeceBOPt/l8HQjAsPdPDw4IIgSJtShhqo6goDFegqEDnWQLMqAxC0b0ALQcr6Wh6AYnDuB1TTPUkljWIMrhht4nT5IUyiZIkdCcwL8RC7QvPdBEYYlGUdBtAM9h1MosstArYx5FLKujCLMujJLkzS7MQjxUsXABcFLVXZFUjiHlRWUyVWzAMgyBbO4BVBVsEC4IQJA7ElYadiNMRjeEAfJWe0X03oqWtX1SDTaH43kJQSdhxEwB5WGS2sCta0bVd220LtsOHSNx0EGdF0Pa8ajAPdV2Pazr2bR9X0vGXf2BVdgPA7t4PLHIpBQzDUgpbM/CI2wHAo1PwhiJIV1KNoKj13jMeGH3xOnqQZMU9I1OPrT%2BO6IzzPNGzEDeLrXjWAb/NhoLLS38/SQPz0Mss3LrQ60raRf0vvLfW4w%2Bafz6O0W%2B9JqgfyNnMBYZtOAWykCFUgYVh42ztg7J2Ls3aFU4N7X2RAWyrCDnYGa4dVjaCjqfOOvVsogHqgVTgMgiqVQAJyRBkJw5KkQ8qNSkM1UgMMarMPYUVbQRVOBFTytobQ3BtDsPYWgzaNsOpdRHvHUg/UhrzAIDEX4qcpohwzsoE0ftpao2nsjXg890ZL2HgAdwuDEWGyDUHoLaisQOWxHGEAQFsLBjtuDO1du7T2dCMraKQHogxBAjHp1mqvcxxDlBTyYDPLgtiEYLwxldZxzBXHjzhj3DxqjpBUN8f4wJ9tgmhLwcwyJ/lZgIEwMwLAERd49yETDCRBU8oyBqnw5K7gFrlWkSo627UjAaJ6hlBhQVOAFUiEFdhijuDVQkeIvKDUe6RCthg6ZczmkCO0AcrxTTMqkGNKtNm3AgA%3D%3D%3D)).  I've fixed this by changing it to `template <auto last_member, typename = decltype(last_member)>` ([godbolt](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAM1QDsCBlZAQwBtMQBGAFlJvoCqAZ0wAFAB4gA5AAYppAFZdSrZrVDIApACYAQjt2kR7ZATx1KmWugDCqVgFcAtrRABmAGykr6ADJ5aTAA5ZwAjTGIQLwAHVCFCc1o7Rxd3GLiEun9AkKdwyK9jTFNEhgJmYgJk51dPI0wTMzoyioJs4LCIqKNyyurUuqFetoCOvK6PAEojVAdiZA4pHTcA5EcsAGpNNxsCAE9ozAB9AmJmQiFt7E0ZAEEb24JMJ2jVJ62d5gciDdUho6cz3ypA2%2B0OtGYgI%2BABENlg1mDMBA/gQAUCIpMrg8hsQHKYNrpmCItgB2fR3DYbBzxdQbAAqMLhxVYiORRNRgPGxExbnJt0pQ2YZmQRyJIkqECG6BAIDwQiOVmcRwAbtsbHSrjy%2BZoSdDtnyni83pgPrsDlZISaNW5rnc1mKNtCInhlZh0PqHg9FU4NvahEINgAxVCoDYgDYBAikvmU/nbWEyUgPSm6RmcJN3HV63kPQ2vIUmtVYu44vFRp3EF1utXB1BXMMEomYGshmX6G3RrMezN3b2%2Bv4BwnEBuRzsUjY2RmJ5OOtMZ%2B667uPZ7595Fm3Y05lx3O13unZD%2BvhwkiNVDmXZ7Cdxc5nu3UdOc60CCTMf8jaC4Wi/0RAiSgjSrK8pCJaKpqhWVb7jYtZXDKdIgiezY7LWbawSA1rYFqM6fngIpir%2B/6AXKRwgYCYE7BBe5nhUaHwY2p4HhUF60ZqS4CuUX74RKUoysRpHHKqjEMdBrYgO22BwSCME2lh444XhP7cQBvHAaBgk2Ih1GRCAl6SY2xCsbe77yd%2B4p/jxQEkWp4G7tWyEhrRUkOTJbEfhxuGmQRFl8dZFG2VBh42npgWYUuXZ3FI0ysNIACs8iuLI8ioNINgGAYH6zPMhbaG4nDyAQ0hyJM0wANYgNwAAcAB0HjcDIJIyDFnhuA1uUxTFKjSNw8WFUl0jyEIICJgViXTHAsAwIgKCoC8eDsGQFAQGgs3zSAwAkh1VBzU8xCDRAoS9aQoQBBUezSHlpDLYC9AAPK0KwZ2JaQWCPuo7CHfgxDFGYrqDU9mDiMU3yLBdkYNIdrB4KEZzEHsdhYIdpx4E452RXwdCMCw708OjggiBIh1KOmqjqCgaV6CoUODZA0yoNETS0H9AC0UrxloegGJwJIDQ032JJY1j9K46Y%2BO0uT5MosTxAzQuSxkDNi50kTpkUJTNMMssq7zau0C0lSK1yyiCn09g1EbwwGxLnDTEImULFwUWxT1T3JVI4gVR4TO1RswDIMgGwklVMUbBAuCECQWw5emE4zdEc0RJHuWvqlHN6PlvVjUgy1x/N5CUNn8eRMAFXpltLIRHtB1PcdEKw6jl0zddBB3Q9H3PGowDvU9n1879h0A0DTz12D0VPZD0OnfDixyKQSMo1IeXTPwmNsBwOPL8IYiSE9SjaCoHdk6nhjj9TL6kHTDPM6zbjQuz%2Bh6FzPONPzEA%2BJr3jWJbXTplLmRJKbqQf7y0SF/ZW9Rn7q1aO/VWDM9YjByErc2UCAHCx6K0UBDsZhzHttbTqUg4qkASjPV27tPbe19v7aqnAQ5hyIMOZY0c7ArQTssbQydya6HTqNUqIB2pVU4DIDwTUACcbgZCiNym4CqeDuqkBRi1fhwiPDaA8JwDwFVtDaG4NoYRwjCGHVdgNIas8M6kHGlNWYBBojfDzktWOhdlBunDmA5eTBV5cF4BvfG28Z4AHczjRFRo7fBztiHSAYRsXxhAEAbFIV7bgPs/YByDlwoqZikCWOsQQWxBdVp7ycXQ5QrisZr08RjTeBMnr%2BOYIEheaNR4EKIX1JYOVtCROibEj28TEmUP4akyK0wECYGYFgSIZ9R6yJRsoqqFUZAtSkblEkG0GpqP0S7fqRhjEjSKjwmKnAqpuBisInR3APCeB0eojqo83ChOaf04qeDtC3PPhs7ZAzSCul2vzbgQA%3D%3D%3D)), but this is a bit of an awkward fix.  The problem is solved on GCC 8 (though I can't find a specific bug report for it; updating to GCC 8 would also remove all of the misleading "'vertex_shader_uid_data::<unnamed struct>::texgentype' is too small to hold all values of 'enum class TexGenType'" warnings ([bug 61414](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=61414)).

Source/Core/Core/FifoPlayer/FifoRecorder.cpp
```diff
@@ -224,3 +226,117 @@ FifoRecorder& FifoRecorder::GetInstance()
...
+        if (index > max_index)
+          max_index = index;
+      }
+    }
+  }
+
+  const u32 array_start = m_cpmem.array_bases[array_index];
+  const u32 array_size = m_cpmem.array_strides[array_index] * (max_index + 1);
+
+  FifoRecorder::GetInstance().UseMemory(array_start, array_size, MemoryUpdate::VERTEX_STREAM);
```
phire, line 341: Just thinking out aloud here.

This is going to result in creating individual memory update for every single triangle strip or other primitive group out of the same vertex buffer. 

It might be smarter to simply cache the min_index and max_index from function and make the memory as used the next time CPmem is updated to point at new vertex buffers.

phire:
One thing I'm slightly worried about is a performance impact from rewriting OpcodeDecoder to be more generic, especially in single-threaded mode. It's a quite hot bit of code.

Pokechu22:
I've done a bit of inline hackery to tell the compiler to inline everything in the main case for OpcodeDecoder (it generates `OpcodeDecoder::RunFifo<0>` for the main fifo and then `OpcodeDecoder::Run0<OpcodeDecoder::RunCallback<0>,void>` for the display list call, checked using Visual Studio's disassembler (alt+8 while debugging)).  The new uses in the fifo player/recorder/analyzer don't have the same aggressive inlining, but I don't think they need to be as optimized.  I haven't done any actual performance testing (as my machine isn't really well suited to that), though.

Source/Core/DolphinQt/FIFO/FIFOAnalyzer.cpp
Pokechu22: `fmt::to_string` is being used to go through the EnumFormatter for e.g. `GX_DRAW_QUADS`, not to create a number.  So that wouldn't do the right thing.

Source/Core/Core/FifoPlayer/FifoRecorder.h
Pokechu22: I've moved `FifoRecordAnalyzer` to the cpp file; however, I also moved the main OpcodeDecoding logic into the header for inlining, which includes other asserts (that are probably more useful).

Pokechu22: Oh, I should add: I replaced *this* assert with a warning log, as it's in the recording code, which currently assumes that display lists will be inlined so one should never reach here.  But I did remove the asserts on analysis and playback.

Source/Core/Core/FifoPlayer/FifoRecordAnalyzer.cpp
```diff
@@ -1,104 +0,0 @@
...
-    FifoAnalyzer::LoadCPReg(CP_VAT_REG_A + i, cpMem[CP_VAT_REG_A + i], s_CpMem);
-
-  const u32* const bases_start = cpMem + ARRAY_BASE;
-  const u32* const bases_end = bases_start + s_CpMem.arrayBases.size();
-  std::copy(bases_start, bases_end, s_CpMem.arrayBases.begin());
-
-  const u32* const strides_start = cpMem + ARRAY_STRIDE;
-  const u32* const strides_end = strides_start + s_CpMem.arrayStrides.size();
-  std::copy(strides_start, strides_end, s_CpMem.arrayStrides.begin());
-}
```
Pokechu22, line 32: Note that this code failed to load `CP_VAT_REG_B` and `CP_VAT_REG_C`, which caused errors when recording for some games where the initial CP state matters (for instance, Need for Speed: Most Wanted was affected).  This has been fixed with the new code, which also made this kind of loading happen in only one place.

Source/Core/Common/EnumMap.h
Pokechu22: Moved to `Common`.  I don't think there are any tests I can reasonably write; the `EnumFormatter` test covers it pretty well and most of the other logic is something that would produce a compile-time error when violated (which I don't think can be handled in a test).

Source/Core/VideoCommon/BPMemory.h
```diff
@@ -491,6 +491,81 @@ struct fmt::formatter<TevStageCombiner::ColorCombiner>
   template <typename FormatContext>
   auto format(const TevStageCombiner::ColorCombiner& cc, FormatContext& ctx)
   {
+    auto out = ctx.out();
+    if (cc.bias == TevBias::Compare)
+    {
+      // TODO
```
Rumi-Larry, line 497: This TODO was left blank

Source/Core/VideoCommon/BPMemory.h
```diff
@@ -512,7 +587,70 @@ struct fmt::formatter<TevStageCombiner::AlphaCombiner>
   template <typename FormatContext>
   auto format(const TevStageCombiner::AlphaCombiner& ac, FormatContext& ctx)
   {
-    return format_to(ctx.out(),
+    auto out = ctx.out();
+    if (ac.bias == TevBias::Compare)
+    {
+      // TODO
```
Rumi-Larry, line 593: Ditto

Source/Core/VideoCommon/CPMemory.h
```diff
@@ -53,24 +55,44 @@ enum
...
 // Vertex array numbers
-enum
+enum class CPArray : u8
+{
+  Position = 0,
+  Normal = 1,
+  Color0 = 2,
+  Color1 = 3,
+  TexCoord0 = 4,
+  TexCoord1 = 5,
```
Rumi-Larry, line 65: Adding extra empty lines after 61 and 64 would make it consistent with the separation after line 71

Source/Core/VideoCommon/OpcodeDecoding.cpp
```diff
@@ -81,202 +41,229 @@ void Init()
...
+  {
+    if (static_cast<Opcode>(opcode) == Opcode::GX_UNKNOWN_RESET)
+    {
+      // Datel software uses this command
+      m_cycles += 6;
+      DEBUG_LOG_FMT(VIDEO, "GX Reset?");
+    }
+    else if (static_cast<Opcode>(opcode) == Opcode::GX_CMD_UNKNOWN_METRICS)
+    {
+      // zelda 4 swords calls it and checks the metrics registers after that
```
Rumi-Larry, line 203:
```suggestion
      // 'Zelda Four Swords' calls it and checks the metrics registers after that
```

Source/Core/VideoCommon/OpcodeDecoding.cpp
```diff
@@ -81,202 +41,229 @@ void Init()
...
+    else
+      return g_main_cp_state;
+  }
+
+  u32 m_cycles;
+  bool m_in_display_list;
+};
+
+// This code is very hot (as everything in the FIFO passes through it and it needs to work
+// realtime). Thus, some hackery is used above to inline a lot of code into it.
```
Rumi-Larry, line 251:
```suggestion
// This code is very hot; as everything in the FIFO passes through it and it needs to work
// in realtime. Thus, some hackery is used above to inline a lot of code into it.
```
Shouldn't this comment be next to the relevant code _above_ anyway?

Source/Core/VideoCommon/XFStructs.cpp
```diff
@@ -44,7 +37,7 @@ static void XFRegWritten(int transferSize, u32 baseAddress, DataReader src)
     case XFMEM_STATE1:  // internal state 1
     case XFMEM_CLOCK:
     case XFMEM_SETGPMETRIC:
-      nextAddress = 0x1007;
+      // Not implemented
```
Rumi-Larry, line 40: What isn't implemented?

Pokechu22: All of the above XF commands (`XFMEM_ERROR` through `XFMEM_SETGPMETRIC`).  They weren't implemented before, but I added the comment here to make that clear (and also I got rid of the system where it would skip over multiple commands since it was unnecessarily complicated for no real benefit.)

Source/Core/VideoCommon/OpcodeDecoding.cpp
Pokechu22: I've removed the comment.  It made more sense before when there were inline versions of `Run` and `RunCommand` right above in `OpcodeDecoding.cpp`, but I've since moved them to the header so that _all_ uses of them can be inlined.  The current, more relevant comment is [here](https://github.com/Pokechu22/dolphin/blob/8e468820a5ef6fa5d013872e030bb9cca989862f/Source/Core/VideoCommon/OpcodeDecoding.h#L66-L69) (though perhaps I should add a bit more info about inlining there).

Source/Core/VideoCommon/XFStructs.cpp
Rumi-Larry: Then, the comment makes more sense above the commands.

Source/Core/VideoCommon/BPMemory.h
Pokechu22: I don't think there's really a good way of formatting it when in compare mode; compare mode doesn't have that much of an equation, but what it does have is awkward to format.  I'll just get rid of the TODO comment.

Source/Core/VideoBackends/Software/SWVertexLoader.cpp
```diff
@@ -32,32 +32,25 @@ SWVertexLoader::SWVertexLoader() = default;
...
 SWVertexLoader::~SWVertexLoader() = default;
 
+DataReader SWVertexLoader::PrepareForAdditionalData(OpcodeDecoder::Primitive primitive, u32 count,
+                                                    u32 stride, bool cullall)
+{
+  // Prevent multiple groups of vertices from being batched, as that causes issues with SetupUnit
+  Flush();
+  return VertexManagerBase::PrepareForAdditionalData(primitive, count, stride, cullall);
+}
+
```
Pokechu22, line 42: Unfortunately this causes problems, as (at least as far as I can tell) fifologs have the memory update after _all_ primitive data, as they get recorded when textures are loaded, and textures get loaded in the original `Flush()` call which is at the end of the batch.  Thus, the first object that uses a texture won't have the texture loaded, resulting in most of the fifoci differences seen here.  I don't think there's a way of avoiding this.

Pokechu22: I've removed the relevant commit.  (It's archived [here](https://github.com/pokechu22/dolphin/tree/archive/better-fifo-analyzer-part-3-sw-setupunit-raw-primitive-data)).  There are now no fifoci differences.

phire (changes requested):

Source/Core/VideoCommon/BPMemory.h
```diff
@@ -491,6 +491,81 @@ struct fmt::formatter<TevStageCombiner::ColorCombiner>
...
+    {
+      // Generate an equation view, simplifying out addition of zero and multiplication by 1
+      // dest = (d (OP) ((1 - c)*a + c*b) + bias) * scale
+
+      // Note that this is simplified with regards to how lerping works; see PixelShaderGen's
+      // WriteTevRegular for more details.
+
+      static constexpr Common::EnumMap<const char*, TevColorArg::Zero> alt_names = {
+          "prev.rgb", "prev.aaa", "c0.rgb",  "c0.aaa",  "c1.rgb", "c1.aaa", "c2.rgb",    "c2.aaa",
+          "tex.rgb",  "tex.aaa",  "ras.rgb", "ras.aaa", "1",      ".5",     "konst.rgb", "0",
```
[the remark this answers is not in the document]
Pokechu22: Unfortunately, we only have access to the alpha combiner or color combiner itself, and not the rest of the BP state (due to both how the fifo analyzer works and how `fmt::formatter` works).  So that's not possible.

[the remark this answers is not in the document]
phire: Might be worth doing some refactoring later to make this possible.

Source/Core/VideoCommon/BPMemory.h
```diff
@@ -512,7 +587,69 @@ struct fmt::formatter<TevStageCombiner::AlphaCombiner>
...
+
+      if (ac.dest == TevOutput::Prev)
+        out = format_to(out, "dest = ");
+      else
+        out = format_to(out, "{:n} = ", ac.dest);
+
+      if (has_scale)
+        out = format_to(out, "(");
+      if (has_d)
+        out = format_to(out, "{:n}", ac.d);
```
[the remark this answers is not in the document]
Pokechu22: It doesn't need it; the names used by `TevAlphaArg` are `"prev", "c0", "c1", "c2", "tex", "ras", "konst", "ZERO"`.  Of these, only "ZERO" is bad, but any case where zero would show up is instead special-cased to exclude the term.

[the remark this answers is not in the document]
phire: I was thinking it could be worth it to have `prev.a`, `c0.a`, `c1.a`, `c2.a` `ras.a` and `konst.a` for consistency with color combiner.

Source/Core/VideoCommon/BPMemory.h
```diff
@@ -491,6 +491,81 @@ struct fmt::formatter<TevStageCombiner::ColorCombiner>
...
+      {
+        if (cc.c == TevColorArg::Zero)
+          out = format_to(out, "{}", alt_names[cc.a]);
+        else if (cc.c == TevColorArg::Half)
+          out = format_to(out, ".5*{}", alt_names[cc.a]);
+        else
+          out = format_to(out, "(1 - {})*{}", alt_names[cc.c], alt_names[cc.a]);
+      }
+      if (has_a && has_b)
+        out = format_to(out, " {} ", op);
```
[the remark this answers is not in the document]
Pokechu22: Yep, this should be hardcoded to `+`.  To put it another way, what is currently implemented is this:

```
dest = (d (OP) ((1 - c)*a (OP) c*b) + bias) * scale
```

but it should be this:

```
dest = (d (OP) ((1 - c)*a + c*b) + bias) * scale
```

Good catch.

Source/Core/VideoCommon/BPMemory.h
```diff
@@ -491,6 +491,81 @@ struct fmt::formatter<TevStageCombiner::ColorCombiner>
...
+        if (has_a || has_b || has_d)
+          out = format_to(out, cc.bias == TevBias::AddHalf ? " + .5" : " - .5");
+        else
+          out = format_to(out, cc.bias == TevBias::AddHalf ? ".5" : "-.5");
+      }
+      else
+      {
+        // If nothing has been written so far, add a zero
+        if (!(has_a || has_b || has_d))
+          out = format_to(out, "0");
```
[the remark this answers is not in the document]
Pokechu22: I don't think that situation is common enough to worry about and would result in a fair bit of additional complexity; similarly, it would be possible to eliminate the parentheses if there's only one term, but that would make the code a lot more messy.

Source/Core/VideoCommon/BPMemory.h
```diff
@@ -491,6 +491,81 @@ struct fmt::formatter<TevStageCombiner::ColorCombiner>
...
+      static constexpr Common::EnumMap<const char*, TevColorArg::Zero> alt_names = {
+          "prev.rgb", "prev.aaa", "c0.rgb",  "c0.aaa",  "c1.rgb", "c1.aaa", "c2.rgb",    "c2.aaa",
+          "tex.rgb",  "tex.aaa",  "ras.rgb", "ras.aaa", "1",      ".5",     "konst.rgb", "0",
+      };
+
+      const bool has_d = cc.d != TevColorArg::Zero;
+      // If c is one, (1 - c) is zero, so (1-c)*a is zero
+      const bool has_a = cc.a != TevColorArg::Zero && cc.c != TevColorArg::One;
+      // If either b or c is zero, b*c is zero
+      const bool has_b = cc.b != TevColorArg::Zero && cc.c != TevColorArg::Zero;
```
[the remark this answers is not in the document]
Pokechu22: I think `has_b` is a bit clearer, in that the variable is indicating whether the `b` term shows up in the result, without saying anything about `c` (since it could appear in the `a` term instead).

`has_b_and_c` would be more accurate than `or` (since both b and c need to be nonzero), but there isn't a corresponding good name for `has_a`.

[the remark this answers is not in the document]
phire: Hmm, when you put it like that the current naming does make more sense. 

You could go more abstract with `have_left` and `have_right` for the two sides of the equation.

Source/Core/VideoCommon/BPMemory.h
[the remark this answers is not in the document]
Pokechu22: Actually, no, this is correct; it generates

```
dest = (d (OP) (1 - c)*a (OP) c*b + bias) * scale
```

which is equivalent; it's stripped away a layer of parentheses and distributed the subtraction (if OP is `-`).  I believe I did this to simplify the code, but I should probably add a comment.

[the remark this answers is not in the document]
phire: I think that format is a bit hard for the reader to parse, especially since these equations are never represented like that anywhere else.

What do you think about my suggestion to represent this as just

     dest = (d (OP) lerp(a, b, c) + bias) * scale

when a, b and c are all non-zero/non-one. Would also avoid the need for that extra set of parentheses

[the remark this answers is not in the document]
Pokechu22: Yeah, I think that lerp would help.

Source/Core/VideoCommon/BPMemory.h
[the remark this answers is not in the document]
Pokechu22: I've ended up going with `has_bc` and `has_ac`, because on further thought that does still fit well enough in both cases.

Source/Core/VideoCommon/BPMemory.h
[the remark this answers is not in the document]
Pokechu22: I've manually added `.a` to the format strings instead (and added a comment).  Partially, this avoids redundant data, but also, using `alt_names` with the `lerp` change results in this rather bothersome automatic formatting:

```C++
        if (has_a && has_b)
        {
          out =
              format_to(out, "lerp({}, {}, {})", alt_names[ac.a], alt_names[ac.b], alt_names[ac.c]);
        }
```

whereas this is what is obtained otherwise:

```C++
        if (has_a && has_b)
        {
          out = format_to(out, "lerp({:n}.a, {:n}.a, {:n}.a)", ac.a, ac.b, ac.c);
        }
```

Source/Core/Core/FifoPlayer/FifoPlayer.cpp
```diff
@@ -31,6 +31,158 @@
...
+
+    u32 object_start = 0;
+    CPState cpmem;
+    u32 primitive_start = 0;
+
+    while (offset < frame.fifoData.size())
+    {
+      // Add memory updates that have occurred before this point in the frame
+      // TODO: Is this really needed?  Couldn't we just copy all of the frame memory updates?  We're
+      // not associating memory updates with commands...
```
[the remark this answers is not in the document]
Pokechu22: I did a quick test to see what happens if this is changed to just copy all memory updates at the same time, and fifoci shows no differences.  Note that this is used during analysis (when converting `FifoFrameInfo` into `AnalyzedFrameInfo`), not playback.  I don't think the memory update timings issue comes from here.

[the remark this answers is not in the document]
Pokechu22: Actually, on further inspection, I've realized that there's no reason for `AnalyzedFrameInfo` to have its own copy of `memoryUpdates`, as the frame's memory updates are identical.  I've removed it.

Pokechu22:
I've refactored the object code further so that it's possible to use different numbering for objects in terms of primitive data, and for the list of entries in the fifo analyzer (or rather, EFB copies and regular objects are counted separately).  This means that the object counts are no longer changed by this PR, and that objects that don't actually have primitive data no longer are showing up in the fifo player's object range.  This code still needs to be squashed into existing commits, though.

Source/Core/Core/FifoPlayer/FifoPlayer.cpp
```diff
@@ -414,38 +392,33 @@ void FifoPlayer::WriteFrame(const FifoFrameInfo& frame, const AnalyzedFrameInfo&
...
-      WriteFramePart(object.m_start, last_object.m_start + last_object.m_size, memory_update,
-                     frame);
+      // We always include commands and EFB copies, as commands from earlier objects still apply to
+      // later ones (games generally do not reconfigure everything for each object)
+      show_part = true;
     }
+
+    if (show_part)
+      WriteFramePart(part, &memory_update, frame);
   }
```
Pokechu22: The new code, although a lot simpler, results in [fifoci differences](https://fifo.ci/version/6fd946c70f69f8d69ab52f80f133c4d54ab15b57/) (most noticeable with [ntfu-reflections](https://fifo.ci/compare/7043712-7041897/)).  These are due to different timings caused by using several calls to `WriteFramePart` instead of one big call to `WriteFramePart` for the main object range (particularly the common case where all objects are enabled).

I've switched back to one big `WriteFramePart` call for now, but it might be better to just accept the timing differences.

phire (approved):
You have a minor lint issue, but otherwise looks good to me.

Source/Core/Core/FifoPlayer/FifoPlayer.cpp
Pokechu22: I switched back to the newer code again.

Source/Core/Core/FifoPlayer/FifoRecorder.cpp
```diff
@@ -6,13 +6,168 @@
...
+  OPCODE_CALLBACK(void OnCommand(const u8* data, u32 size)) {}
+
+  OPCODE_CALLBACK(CPState& GetCPState()) { return m_cpmem; }
+
+private:
+  void ProcessVertexComponent(CPArray array_index, VertexComponentFormat array_type,
+                              u32 component_offset, u32 vertex_size, u16 num_vertices,
+                              const u8* vertex_data);
+
+  const FifoRecorder* m_owner;
```
Pokechu22, line 53: This field is currently unused (the code that should use it instead uses `FifoRecorder::GetInstance()`).

JMC47:
This has been tested locally and has seen a lot of reviews from trusted developers.

shehzman:
Ever since this branch, Mario Galaxy water shaders don't seem to load properly in Sea Slide Galaxy and Loopdeswoop/Loopdeloop Galaxy. Attached is the dump file and a screenshot of what the error looks like.
[bad_ps_5_0_D3D_0.txt](https://github.com/dolphin-emu/dolphin/files/7860073/bad_ps_5_0_D3D_0.txt)
Screenshot (5)
.

Pokechu22:
Thanks; I've created #10366 which should fix that.

PR #9719, opened 2021-05-17
DolphinQt: Add quotes around QtIncludeDir on Windows

Dentomologist:
As of 99a724361abb5c9728761dfc13b9f58a0b4fff8b Visual Studio gets hopelessly confused if QtIncludeDirectory has any spaces in the path. Adding quotes around the include directive solves the issue.

### DIFF
--- a/Source/VSProps/QtCompile.props
+++ b/Source/VSProps/QtCompile.props
@@ -7,7 +7,8 @@
     <QTDIR Condition="Exists('$(QTDIR)') And !HasTrailingSlash('$(QTDIR)')">$(QTDIR)\</QTDIR>
     <QtDirValid>false</QtDirValid>
     <QtDirValid Condition="Exists('$(QTDIR)')">true</QtDirValid>
-    <QtIncludeDir>$(QTDIR)include\</QtIncludeDir>
+    <QtIncludeDirWithoutTrailingSeparator>$(QTDIR)include</QtIncludeDirWithoutTrailingSeparator>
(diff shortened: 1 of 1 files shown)


Source/VSProps/QtCompile.props
```diff
@@ -32,7 +33,7 @@
       -->
       <AdditionalOptions>%(AdditionalOptions) /experimental:external</AdditionalOptions>
       <AdditionalOptions>%(AdditionalOptions) /external:W0</AdditionalOptions>
-      <AdditionalOptions>%(AdditionalOptions) /external:I$(QtIncludeDir)</AdditionalOptions>
+      <AdditionalOptions>%(AdditionalOptions) /external:I "$(QtIncludeDirWithoutTrailingSeparator)"</AdditionalOptions>
```
Dentomologist, line 36: If I put quotes around QtIncludeDir the trailing separator escapes the second quote and causes VS to barf in new and exciting ways.  I could put \" afterward so the trailing separator escapes the \ instead, which ends up with the right result, but I think this way is clearer (if more verbose).

shuffle2:
lgtm

JosJuice:
Do we need this for CMake as well?

Either way, this change makes sense, so I'll merge this.

Dentomologist:
> Do we need this for CMake as well?

As far as I know CMake doesn't have an issue with this particular path, but some testing shows it has a different space-in-path issue I'm looking into.

PR #9720, opened 2021-05-17
NetPlay: Move NetSettings initialization to backend

Techjar:
It doesn't really belong in frontend code, and this change will make it easier to eventually support NetPlay on other UI frameworks. This should have no effect on current functionality, as it's more or less a refactor.

### DIFF
--- a/Source/Core/Core/NetPlayServer.cpp
+++ b/Source/Core/Core/NetPlayServer.cpp
@@ -33,8 +33,10 @@
 #include "Common/UPnP.h"
 #include "Common/Version.h"
 #include "Core/ActionReplay.h"
+#include "Core/Config/GraphicsSettings.h"
 #include "Core/Config/MainSettings.h"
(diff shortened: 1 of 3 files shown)


JMC47:
Didn't completely break netplay as far as I can tell.

Source/Core/Core/NetPlayServer.cpp
```diff
@@ -1210,9 +1212,112 @@ bool NetPlayServer::AbortMD5()
 }
 
 // called from ---GUI--- thread
-void NetPlayServer::SetNetSettings(const NetSettings& settings)
+bool NetPlayServer::SetupNetSettings()
 {
+  NetPlay::NetSettings settings;
```
lioncash, line 1217: This can be moved closer to its usage after the if statement

lioncash (commented):
LGTM otherwise

lioncash (approved):

PR #9721, opened 2021-05-17
Fix some warnings found with gcc 11 and ffmpeg master

linkmauve:
These are all mostly obvious.

The main remaining ones are:
- `DSP::SDSP` not being pod yet `offsetof()` being used on it, same for `memcpy()` on `CPMemory`.
- Some signed comparison issues where I’m not sure in which direction to cast.
- Some invalid size copies in `IOS/ES/Formats.cpp` which made me notice the addition of `ticket_version` in `Ticket` which isn’t there in `TicketView`, making me wonder just how this ever worked.

### DIFF
--- a/Source/Core/Common/CRC32.cpp
+++ b/Source/Core/Common/CRC32.cpp
@@ -1,0 +1,19 @@
+// Copyright 2021 Dolphin Emulator Project
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+#include <zlib.h>
+
+#include "Common/CRC32.h"
+
+namespace Common
+{
+u32 ComputeCRC32(std::string_view data)
+{
+  const Bytef* buf = reinterpret_cast<const Bytef*>(data.data());
+  uInt len = static_cast<uInt>(data.size());
+  // Use zlibs crc32 implementation to compute the hash
+  u32 hash = crc32(0L, Z_NULL, 0);
+  hash = crc32(hash, buf, len);
+  return hash;
+}
+}  // namespace Common
--- a/Source/Core/Common/CRC32.h
+++ b/Source/Core/Common/CRC32.h
@@ -1,0 +1,11 @@
+// Copyright 2021 Dolphin Emulator Project
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+#include <string_view>
+
+#include "Common/CommonTypes.h"
+
+namespace Common
+{
+u32 ComputeCRC32(std::string_view data);
+}  // namespace Common
--- a/Source/Core/Core/IOS/USB/Bluetooth/BTReal.cpp
+++ b/Source/Core/Core/IOS/USB/Bluetooth/BTReal.cpp
@@ -546,8 +546,11 @@ void BluetoothRealDevice::SaveLinkKeys()
     oss << Common::MacAddressToString(address);
     oss << '=';
     oss << std::hex;
-    for (const u16& data : entry.second)
-      oss << std::setfill('0') << std::setw(2) << data;
+    for (u8 data : entry.second)
+    {
+      // We cast to u16 here in order to have it displayed as two nibbles.
+      oss << std::setfill('0') << std::setw(2) << static_cast<u16>(data);
+    }
     oss << std::dec << ',';
   }
   std::string config_string = oss.str();
--- a/Source/Core/DolphinQt/QtUtils/SignalDaemon.cpp
+++ b/Source/Core/DolphinQt/QtUtils/SignalDaemon.cpp
@@ -34,8 +34,9 @@ void SignalDaemon::OnNotifierActivated()
   m_term->setEnabled(false);
 
   char tmp;
-  if (read(s_sigterm_fd[1], &tmp, sizeof(char)))
+  if (read(s_sigterm_fd[1], &tmp, sizeof(char)) != sizeof(char))
   {
+    // Not much we can do here.
   }
 
   m_term->setEnabled(true);
@@ -45,10 +46,14 @@ void SignalDaemon::OnNotifierActivated()
 
 void SignalDaemon::HandleInterrupt(int)
 {
-  write(STDERR_FILENO, message, sizeof(message));
+  if (write(STDERR_FILENO, message, sizeof(message)) != sizeof(message))
+  {
+    // Not much we can do here.
+  }
 
   char a = 1;
-  if (write(s_sigterm_fd[0], &a, sizeof(a)))
+  if (write(s_sigterm_fd[0], &a, sizeof(a)) != sizeof(a))
   {
+    // Not much we can do here.
   }
 }
--- a/Source/Core/DolphinQt/Resources.cpp
+++ b/Source/Core/DolphinQt/Resources.cpp
@@ -21,9 +21,10 @@ QList<QPixmap> Resources::m_platforms;
 QList<QPixmap> Resources::m_countries;
 QList<QPixmap> Resources::m_misc;
 
-QIcon Resources::GetIcon(const QString& name, const QString& dir)
+QIcon Resources::GetIcon(std::string_view name, const QString& dir)
 {
-  QString base_path = dir + QLatin1Char{'/'} + name;
+  QString name_owned = QString::fromLatin1(name.data(), static_cast<int>(name.size()));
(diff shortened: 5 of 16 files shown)


lioncash:
offsetof on non-POD types is mostly fine, given the standard allows it to be conditionally supported by implementations (and all implementations that we care about support it).

Maybe -Wreorder should be added into the CMakeLists as an error for the codebase, given there's never a scenario where it's desirable for them to be out of order (and at worst, indicate a bug).

Source/Core/Common/CRC32.h
```diff
@@ -0,0 +1,9 @@
+// Copyright 2021 Dolphin Emulator Project
+// Licensed under GPLv2+
+// Refer to the license.txt file included.
+
+#include <string>
+
+#include "Common/CommonTypes.h"
+
+u32 ComputeCRC32(const std::string& data);
```
JosJuice, line 9: `std::string_view` would be preferable here.

Source/Core/Common/CRC32.cpp
```diff
@@ -0,0 +1,15 @@
+// Copyright 2021 Dolphin Emulator Project
+// Licensed under GPLv2+
+// Refer to the license.txt file included.
+
+#include <zlib.h>
+
+#include "Common/CRC32.h"
+
+u32 ComputeCRC32(const std::string& data)
```
lioncash, line 9:
```suggestion
u32 ComputeCRC32(std::string_view data)
```

Would allow this to be used in a non-allocating manner as well.

Source/Core/Common/CRC32.h
lioncash, line 9: This should be in the Common namespace.

Source/Core/Core/IOS/USB/Bluetooth/BTReal.cpp
```diff
@@ -547,7 +547,7 @@ void BluetoothRealDevice::SaveLinkKeys()
     oss << Common::MacAddressToString(address);
     oss << '=';
     oss << std::hex;
-    for (const u16& data : entry.second)
+    for (u8 data : entry.second)
```
lioncash, line 550: Note that this will print out as a character, rather than as a value, like it would with u16. Maybe cast data to u16 in body if that matters.

Source/Core/Common/CRC32.h
linkmauve: Thanks, fixed!

Source/Core/Common/CRC32.cpp
linkmauve: Thanks, fixed!

Source/Core/Core/IOS/USB/Bluetooth/BTReal.cpp
linkmauve: I went with the cast, with a comment to explain why.

Source/Core/VideoBackends/Software/Tev.cpp
```diff
@@ -815,6 +815,9 @@ void Tev::Draw()
       fog = 1.0f - fog;
       fog = pow(2.0f, -8.0f * fog * fog);
       break;
+    case FogType::Linear:
+    case FogType::Off:
+      ASSERT(0);
```
linkmauve, line 820: Most probable reason for the rendering failures would be this assert, which I added in order to see which games might use them since they weren’t actually handled.

But fifoci seems to be telling me that they are actually used. :)

Source/Core/VideoCommon/FrameDump.cpp
```diff
@@ -336,10 +336,9 @@ void FrameDump::ProcessPackets()
 {
   while (true)
   {
-    AVPacket pkt;
-    av_init_packet(&pkt);
+    AVPacket* pkt = av_packet_alloc();
```
Warepire, line 339: Might return NULL on error

Source/Core/VideoCommon/FrameDump.cpp
```diff
@@ -336,10 +336,9 @@ void FrameDump::ProcessPackets()
...
 
-    const int receive_error = avcodec_receive_packet(m_context->codec, &pkt);
+    const int receive_error = avcodec_receive_packet(m_context->codec, pkt);
 
     if (receive_error == AVERROR(EAGAIN) || receive_error == AVERROR_EOF)
     {
```
Warepire, line 344: This statement causes `pkt` to leave scope without being freed, same with any other statement that breaks this loop. I suggest wrapping `pkt` in a smart pointer.

Source/Core/VideoCommon/FrameDump.cpp
linkmauve: Indeed, fixed in the latest version.

Source/Core/VideoCommon/FrameDump.cpp
linkmauve: Indeed, that’s now fixed with a `std::unique_ptr` with a deleter.

Source/Core/VideoCommon/FrameDump.cpp
```diff
@@ -336,10 +336,17 @@ void FrameDump::ProcessPackets()
 {
   while (true)
   {
-    AVPacket pkt;
-    av_init_packet(&pkt);
+    std::unique_ptr<AVPacket, std::function<void(AVPacket*)>> pkt =
```
lioncash, line 339: Maybe use auto here like:

```
auto pkt = ...
```

to reduce the duplication of the type a little.

As an aside, perhaps this should be moved out of the loop so the packet is only allocated once and then reused throughout the iterations to avoid heap churn a little bit.

Tilka:
Please rebase and drop the -Wreorder commit as that's been fixed (sorry, I didn't check open PRs).

Source/Core/VideoCommon/FrameDump.cpp
Dentomologist: As a less verbose alternative to unique_ptr's deleter syntax:
```
  AVPacket* pkt = av_packet_alloc();
  Common::ScopeGuard pkt_guard{[&pkt]() { av_packet_free(&pkt); }};
```
... which also frees you from having to call get() everywhere. (ScopeGuard is in Common/ScopeGuard.h)

Source/Core/DolphinQt/QtUtils/SignalDaemon.cpp
```diff
@@ -35,8 +35,9 @@ void SignalDaemon::OnNotifierActivated()
   m_term->setEnabled(false);
 
   char tmp;
-  if (read(s_sigterm_fd[1], &tmp, sizeof(char)))
+  if (read(s_sigterm_fd[1], &tmp, sizeof(char)) != sizeof(char))
```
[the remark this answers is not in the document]
leoetlino: No, that doesn't work. GCC will warn even with an explicit cast to void. (And I don't think a C-style cast gives us such a readability improvement that warrants its use here.)

Source/Core/DolphinQt/Resources.cpp
```diff
@@ -76,13 +76,13 @@ QPixmap Resources::GetScaledPixmap(const std::string& name)
...
 {
-  for (const std::string& platform :
+  for (const char* platform :
        {"Platform_Gamecube", "Platform_Wii", "Platform_Wad", "Platform_File"})
   {
     m_platforms.append(GetScaledPixmap(platform));
   }
 
-  for (const std::string& country :
+  for (const char* country :
```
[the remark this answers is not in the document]
Warepire: In that case maybe prefer a `std::string_view` over a pure `char*`.

linkmauve:
I have now fixed all of the comments you made, sorry for the huge delay in doing so…

linkmauve:
There, fixed all your comments.

leoetlino (approved):
> Reviewed 16 of 16 files at r8, 5 of 5 files at r9, 2 of 2 files at r10, 3 of 3 files at r11, 2 of 2 files at r12, 2 of 2 files at r13, 2 of 2 files at r14, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/dolphin-emu/dolphin/9721)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @linkmauve)

PR #9722, opened 2021-05-18
Gamelist: Add 4x resolution nobanner image

Dentomologist:
When a game hasn't generated a GameList banner yet Dolphin displays an X placeholder image. This PR adds a 4x resolution version of the placeholder. It also improves centering of 2x and android banners, and adjusts 1x banner to increase consistency with other resolutions while maintaining sharpness.


Dentomologist:
Android nobanner now changed as well.

MayImilae:
"Add antialiasing to 2x" it wasn't aliased, it was just sharp. :/ And I can provide a sharp 4x as well if you were worried about them matching.

Dentomologist:
So it turns out the 2x image already had antialiasing, but the color change used was subtle enough that I missed it. The 1x image had a larger color gap which I copied for the other resolutions, which ended up looking blurrier than what the 2x had before.

I've redone the other resolutions to match the 2x banner colors.

Note: The file size reduction isn't present in this version; I want to address any other issues first before redoing that.

MayImilae:
Here you go, nobanner 1x, 2x, and 4x, all perfectly consistent and sharp as they all originate from the same vector file. Also optimized with optifine so no shrinking is necessary.

[nobanner.zip](https://github.com/dolphin-emu/dolphin/files/6506916/nobanner.zip)

Dentomologist:
Thank you for the images, I've updated the build with them. 

The 4x and android versions had their rightmost column slightly transparent, so I made them opaque and reran them through optipng.

MayImilae (approved):

lioncash (approved):

PR #9723, opened 2021-05-19
[DO NOT MERGE!!] Don't clear cache on state load.

AdmiralCurtiss:
Very dangerous nonsense to test whether this could be related to TAS desyncing issues. Don't try to use this on any game that dynamically loads code into main RAM.


AdmiralCurtiss:
FYI this did not turn out to be related -- or if it was, it wasn't the only culprit.

PR #9724, opened 2021-05-19
GameINI: Fix Final Fantasy Crystal Chronicles GBA race condition

Bonta0:
There is a race condition in FFCC when entering any level: GBAs will be stuck on a flickering "Waiting for connection..." screen for a random amount of time
The race happens on the GBA side during the handshake process. If the game also tries to send data during that process it will be pushed into a queue and transfers won't resume once the connection is established, causing it to timeout after 10 frames before trying again
Although this happens on HW, Dolphin is much more likely to lose the race repeatedly due to timing inaccuracies and that can last up to a minute depending on your luck and configuration which is why I believe this should be enabled by default
These patches apply the necessary changes to the rom at ``/dvd/gba/ffcc_cli.bin`` to clear the transfer queue once connected

### DIFF
--- a/Data/Sys/GameSettings/GCCE01.ini
+++ b/Data/Sys/GameSettings/GCCE01.ini
@@ -6,6 +6,29 @@
 $Fix buffer overrun bug (crash at Goblin Wall)
 0x80017B16:word:0x00003430
 
+# Fix a race condition that makes GBAs take longer to connect when entering a level.
+$Fix GBA connections
+0x800b1828:dword:0x4bf534d9
+0x80004d00:dword:0x7ce802a6
+0x80004d04:dword:0x428007f1
+0x80004d08:dword:0x3c608000
+0x80004d0c:dword:0x38834d28
+0x80004d10:dword:0x807f10a0
+0x80004d14:dword:0x38632504
+0x80004d18:dword:0x38a00018
+0x80004d1c:dword:0x428007d9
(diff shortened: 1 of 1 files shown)


JMC47:
@dolphin-emu-bot rebuild

JMC47:
For those unaware, this race condition is unlikely to be fixable on Dolphin or mGBA's end.

Miksel12:
Applying a patch every frame for a (niche?) mode, which isn't strictly necessary, seems a bit much imo. I have no idea how popular the GBA functionality is (going to be) but I have the feeling that most people will never use the GBA mode.

JMC47:
It's actually the opposite case.  There's no reason to really play this game *without* the GBA mode.  The single player mode is very gimped and limited in comparison.

Miksel12:
Ah I see, I didn't know that. I guess that makes this a no brainer.

JMC47:
Final Fantasy Crystal Chronicles and The Legend of Zelda: Four Swords Adventures are the only two games where I'd say the *main* game mode is the one involving GBAs and the only reason they included a mode without it was because getting all the stuff together was expensive.

JMC47:
I did a performance test at the request of @MayImilae and we discussed the merit of the patch being enabled by default versus not.  We both agreed that the patch should be enabled by default as it brings us closer to the console experience and Dolphin is unlikely to ever emulate whatever timing issues are happening on console to allow the race condition to work.

PR #9729, opened 2021-05-20
Fix minor grammar error

Peter0x44:
This is quite self explanatory, I noticed this minor mistake while using dolphin.


Peter0x44:
Sorry, I didn't notice that.

PR #9730, opened 2021-05-20
AudioPane: Remove main layout vertical stretching

Dentomologist:
Fixes AudioPane's layout to prevent it from vertically stretching to fill the window, which looks bad and was inconsistent with the other panes.

Before:
Before

After:
After


MayImilae:
LGTM.

leoetlino (approved):

PR #9710, opened 2021-05-14
Remove all remaining volatile qualifiers

JosJuice:

PR #9731, opened 2021-05-21
Fix perf query regression

JosJuice:
When trying to do a small optimization in PR #9710, I failed to take into account that `WeakFlush` and `FlushOne` update `m_query_count`.

Only D3D11 and OGL had this problem, not D3D12 and Vulkan.

### DIFF
--- a/Source/Core/VideoBackends/D3D/D3DPerfQuery.cpp
+++ b/Source/Core/VideoBackends/D3D/D3DPerfQuery.cpp
@@ -27,16 +27,20 @@ PerfQuery::~PerfQuery() = default;
 
 void PerfQuery::EnableQuery(PerfQueryGroup type)
 {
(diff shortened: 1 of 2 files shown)


JosJuice:
Fixes bad performance in Need for Speed: Most Wanted and a freeze in Super Mario Sunshine.

Techjar:
Can confirm this fixes both the performance loss on OpenGL and the crash on D3D11, in Super Mario Sunshine.

Techjar:
For anyone curious, after analysis of the code, I have determined that the underlying reason why this broke is that we would begin a query with the wrong id, then later when we tried to get the result with the correct id, bad things would happen (read: driver gets angry).

JMC47 (approved):

lioncash (approved):

PR #9732, opened 2021-05-21
Avoid including scmrev.h except in Version.cpp

JosJuice:
Any file which includes scmrev.h must be rebuilt when scmrev.h is regenerated. By not including scmrev.h from any file other than Version.cpp, incremental builds become a little faster.

### DIFF
--- a/Source/Core/UICommon/AutoUpdate.cpp
+++ b/Source/Core/UICommon/AutoUpdate.cpp
@@ -12,7 +12,7 @@
 #include "Common/HttpRequest.h"
 #include "Common/Logging/Log.h"
 #include "Common/StringUtil.h"
-#include "Common/scmrev.h"
+#include "Common/Version.h"
 #include "Core/ConfigManager.h"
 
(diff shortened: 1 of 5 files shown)


Dentomologist:
If we defined scm_update_track_str in Version we could get rid of the scmrev.h include in ConfigManager.cpp as well.

JosJuice:
Done. I didn't actually know about that one.

lioncash (approved):

Source/Core/UICommon/AutoUpdate.cpp
```diff
@@ -157,7 +157,7 @@ void AutoUpdateChecker::CheckForUpdate()
 #endif
 
   std::string version_hash = SConfig::GetInstance().m_auto_update_hash_override.empty() ?
-                                 SCM_REV_STR :
+                                 Common::scm_rev_str :
```
Dentomologist, line 160: This should be scm_rev_git_str; scm_rev_str is defined as something else in Version.

PR #9733, opened 2021-05-21
JitArm64: Remove fast path for converting single to double

JosJuice:
The fast path is of course faster in the cases where we can take it, but the added check for whether to use the fast path gives us additional overhead in the slow case, and the extra code has a negative impact on the icache.

I'm submitting this as a pull request so that we can do performance testing on it. Do not merge this unless it's shown to be faster!

### DIFF
--- a/Source/Core/Core/PowerPC/JitArm64/Jit.h
+++ b/Source/Core/Core/PowerPC/JitArm64/Jit.h
@@ -170,11 +170,9 @@ public:
   void ConvertDoubleToSinglePair(size_t guest_reg, Arm64Gen::ARM64Reg dest_reg,
                                  Arm64Gen::ARM64Reg src_reg);
   void ConvertSingleToDoubleLower(size_t guest_reg, Arm64Gen::ARM64Reg dest_reg,
-                                  Arm64Gen::ARM64Reg src_reg,
-                                  Arm64Gen::ARM64Reg scratch_reg = Arm64Gen::ARM64Reg::INVALID_REG);
+                                  Arm64Gen::ARM64Reg src_reg);
   void ConvertSingleToDoublePair(size_t guest_reg, Arm64Gen::ARM64Reg dest_reg,
-                                 Arm64Gen::ARM64Reg src_reg,
-                                 Arm64Gen::ARM64Reg scratch_reg = Arm64Gen::ARM64Reg::INVALID_REG);
+                                 Arm64Gen::ARM64Reg src_reg);
 
   void FloatCompare(UGeckoInstruction inst, bool upper = false);
 
--- a/Source/Core/Core/PowerPC/JitArm64/JitArm64_FloatingPoint.cpp
+++ b/Source/Core/Core/PowerPC/JitArm64/JitArm64_FloatingPoint.cpp
@@ -678,11 +678,8 @@ void JitArm64::ConvertDoubleToSinglePair(size_t guest_reg, ARM64Reg dest_reg, AR
   ABI_PopRegisters(gpr_saved);
 }
 
-void JitArm64::ConvertSingleToDoubleLower(size_t guest_reg, ARM64Reg dest_reg, ARM64Reg src_reg,
-                                          ARM64Reg scratch_reg)
+void JitArm64::ConvertSingleToDoubleLower(size_t guest_reg, ARM64Reg dest_reg, ARM64Reg src_reg)
 {
-  ASSERT(scratch_reg != src_reg);
-
   if (js.fpr_is_store_safe[guest_reg])
   {
     m_float_emit.FCVT(64, 32, EncodeRegToDouble(dest_reg), EncodeRegToDouble(src_reg));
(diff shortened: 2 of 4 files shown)


JosJuice:
@theofficialgman Please test how the performance of this pull request compares to 5.0-14246 on your Switch. You can get an APK of it by clicking the Details link next to the pr-android builder in the list of build checks.

theofficialgman:
@JosJuice yes I build on linux already though I've never actually had to test a PR. I could just fetch the your fork branch associated with this but what would be the best way of using dolphin master and then applying this PR over it using git?

Dentomologist:
Using 'upstream' as the remote name:
git fetch upstream pull/9733/head:jitarm64-remove-cstd-fast-path
git checkout jitarm64-remove-cstd-fast-path

JosJuice:
> I could just fetch the your fork branch associated with this but what would be the best way of using dolphin master and then applying this PR over it using git?

You would have to first check out the fork branch (as described by Dentomologist) and then either `git merge master` or `git rebase master`. But the fork branch is already based on a fairly recent version of master (5.0-14246), so unless you're going to play Super Mario Sunshine or Need for Speed: Most Wanted (which use performance queries, which was fixed in 5.0-14248), you can just use the fork branch.

theofficialgman:
> @theofficialgman Please test how the performance of this pull request compares to 5.0-14246 on your Switch. You can get an APK of it by clicking the Details link next to the pr-android builder in the list of build checks.
> 
> EDIT: Ah no, wait, I forgot you aren't using Android... In that case, I assume you know how to fetch and build the code of Dolphin already?
 Performance seems pretty much identical

JosJuice:
Okay, thanks for testing.

That probably means we don't have any reason to merge this PR. I would expect the fast path to be a bit better on modern phones than on the Switch in terms of how many times faster than the slow path it is, due to the Cortex-A57's slower floating point instructions (the slow path is mostly integer instructions). So if the performance is the same on the Switch with this PR, the performance is probably a little worse on modern phones with this PR. But I'll leave this open a little while in case anyone else has anything to say.

Gamemulatorer:
Can I test this to my Redmi Note 9 android phone?

JosJuice:
Sure, feel free to test it if you want to. But do keep in mind that in many cases phones are GPU bottlenecked rather than CPU bottlenecked, so ideally you should set the video backend to Null if you want to test how this affects the performance.

Gamemulatorer:
> Sure, feel free to test it if you want to. But do keep in mind that in many cases phones are GPU bottlenecked rather than CPU bottlenecked, so ideally you should set the video backend to Null if you want to test how this affects the performance.

Where I can download the apk?
sorry for asking 😅

JosJuice:
I described it earlier:

> You can get an APK of it by clicking the Details link next to the pr-android builder in the list of build checks.

Gamemulatorer:
Performance is improve a little bit on Vulkan
On OpenGL performance still the same
Null and software doesn't work on my phone.

JosJuice:
Null doesn't work on your phone? That's strange...

Just to be clear, the intended behavior of Null is that nothing is shown on the screen, but the game still runs and accepts inputs and outputs audio.

> Performance is improve a little bit on Vulkan
On OpenGL performance still the same

And you are comparing to 5.0-14246?

Gamemulatorer:
> Null doesn't work on your phone? That's strange...
> 
> Just to be clear, the intended behavior of Null is that nothing is shown on the screen, but the game still runs and accepts inputs and outputs audio.
> 
> > Performance is improve a little bit on Vulkan
> > On OpenGL performance still the same
> 
> And you are comparing to 5.0-14246?

Ah yes I can hear audio on null but slow and blackscreen
I only tested on this game.
Buildbot beta apk
Screenshot_2021-05-22-15-44-51-479_org dolphinemu dolphinemu
Debug beta apk
Screenshot_2021-05-22-15-45-02-139_org dolphinemu dolphinemu debug

JosJuice:
Okay, so it's compared to 5.0-14245. That also works.

Surprising how it would make Vulkan faster but not OpenGL... You'd think that Vulkan would have more of a GPU thread load than OpenGL but not more of a CPU thread load.

Gamemulatorer:
> Okay, so it's compared to 5.0-14245. That also works.
> 
> Surprising how it would make Vulkan faster but not OpenGL... You'd think that Vulkan would have more of a GPU thread load than OpenGL but not more of a CPU thread load.

But on Vulkan that game Naruto is flickering I also tried the playstore version same issue.

theofficialgman:
Hopefully you can manage to makeup that performance somewhere else. At least how I can measure it, it seems like the x86_64 jit is about 1.5X faster than the aarch64 one. I know comparing cpu architectures is hard but I've at minimum normalized geekbench scores between these two cpus (i5-8600k @ 0.97Ghz and A57 @2.091Ghz) https://browser.geekbench.com/v5/cpu/compare/7825399?baseline=8096105 and the intel machine runs games in dolphin 50% faster.

theofficialgman:
yet even as I say this, luabench.dol performs faster on aarch64 than my x86_64 machine (testing latest master dolphin)... which makes no sense given the in game scenario stated before (its a total time of 18min 36sec on the A57, and 20min 27sec on the underclocked 8600K)

@JosJuice do you have any idea why this (including the previous comment) might be the case?

JosJuice:
I guess luabench isn't representative of a normal game? I don't really know beyond that.

dvessel:
Rebased from latest master and tested. All default settings and ran through f-zero intro and a few seconds of gameplay. Looks like there’s no difference.

| builds | samples | min  | max | median | stddev |
| ---: | --- | --- | --- | --- | --- |
| master first run | 14361.000 | 1.213 | 939.965 | 6.975 | 9.841 |
| master second run | 14361.000 | 1.175 | 875.353 | 6.940 | 9.402 |
| master third run | crashed | ??? | ??? | ??? | ??? |
| master fourth run | 14360.000 | 1.164 | 611.686 | 7.218 | 7.349 |
| 9733 first run | 14361.000 | 1.205 | 939.239 | 7.019 | 9.936 |
| 9733 second run | 14361.000 | 1.180 | 933.051 | 6.927 | 9.906 |
| 9733 third run | 14361.000 | 1.182 | 931.542 | 6.915 | 9.904 |
| 9733 fourth run | 14361.000 | 1.171 | 935.850 | 6.928 | 9.887 |

PR #9734, opened 2021-05-21
DolphinQt: AutoUpdate.cpp should use scm_rev_git_str

JosJuice:
PR #9732 regression.


Dentomologist (approved):

JosJuice:
Oops, this should have said UICommon instead of DolphinQt. Oh well. That we got the fix out quickly is what was important, since it affects the updater

PR #9735, opened 2021-05-21
Socket: Bind socket before sending data

sepalani:
This PR is an attempt to address https://bugs.dolphin-emu.org/issues/12507.

A side effect of binding to the default network interface is it prevents the use of the loopback interface.
I added an extra check to choose the loopback interface if used.

It is not hardware accurate but convenient.
I can add it as a config option and disable it by default if it matters.

Ready to be tested.


sepalani:
I'm switching this to draft for now. I suspect that this approach might not work as it might not fix the interface changing part between two different sockets. A better approach would be to involve the SocketManager so it can keep track of the interface currently being used.

PR #9736, opened 2021-05-22
VideoCommon: Account for pixel quads in bounding box calculation

Techjar:
The GC/Wii GPU rasterizes in 2x2 pixel groups, so bounding box values will be rounded to the extents of these groups, rather than the exact pixel. To account for this, we'll round the top/left down to even and the bottom/right up to odd. I have verified that the values resulting from this change exactly match a real Wii.

I would have implemented this in the pixel shaders, but the NVIDIA driver seems to be outsmarting me and optimizes out all my attempts to perform these operations in there, as it works on Vulkan but not OpenGL.

I used this modified homebrew to perform the hardware test: https://qimg.techjargaming.com/f/vRk5NA28/triangle.elf
Source code (ignore the mess, I've used it for a number of tests): https://gist.github.com/Techjar/4f38aa7807d0f5334011ef14ec721219
The values it outputs are as follows:

### DIFF
--- a/Source/Core/VideoCommon/RenderBase.h
+++ b/Source/Core/VideoCommon/RenderBase.h
@@ -211,9 +211,9 @@ public:
   virtual u32 AccessEFB(EFBAccessType type, u32 x, u32 y, u32 poke_data);
   virtual void PokeEFB(EFBAccessType type, const EfbPokeData* points, size_t num_points);
 
-  virtual u16 BBoxRead(int index) = 0;
-  virtual void BBoxWrite(int index, u16 value) = 0;
-  virtual void BBoxFlush() {}
+  u16 BBoxRead(int index);
+  void BBoxWrite(int index, u16 value);
+  void BBoxFlush();
 
   virtual void Flush() {}
   virtual void WaitForGPUIdle() {}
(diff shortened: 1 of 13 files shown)


Source/Core/VideoCommon/RenderBase.h
```diff
@@ -211,7 +211,7 @@ class Renderer
   virtual u32 AccessEFB(EFBAccessType type, u32 x, u32 y, u32 poke_data);
   virtual void PokeEFB(EFBAccessType type, const EfbPokeData* points, size_t num_points);
 
-  virtual u16 BBoxRead(int index) = 0;
+  u16 BBoxRead(int index);
   virtual void BBoxWrite(int index, u16 value) = 0;
```
lioncash, line 215: BBoxWrite and BBoxFlush should also follow this change as well to make the API consistent.

It's not consistent for BBoxWrite to be renamed to BBoxWriteImpl's in all the interfaces but BBoxRead and BBoxFlush are as is.

Techjar: Well, it's redundant for those, but I guess if you want for the sake of consistency.

lioncash: Thanks!

Techjar: Done. I also split the changes up for organizational purposes.

BhaaLseN:
> I would have implemented this in the pixel shaders, but the NVIDIA driver seems to be outsmarting me and optimizes out all my attempts to perform these operations in there, as it works on Vulkan but not OpenGL.

I wonder if this is something that should be left as comment next to the implementation of `BBoxRead` (just to make it more visible)

JMC47:
This fixes a crash in Ultimate Spider-Man during missions with lots of smoke, notably the Rhino mission in https://bugs.dolphin-emu.org/issues/12516

phire (approved):

lioncash (approved):

PR #9738, opened 2021-05-22
Interpreter: Fix fctiwx rounding

JosJuice:
The interpreter implementation of fctiwx was treating rounding mode 0 as "round to nearest, ties towards zero", which is not an actual IEEE-754 rounding mode. The IBM document mentioned in a comment at the top of the function, on the other hand, treats rounding mode 0 as "round to nearest, ties to even", which makes more sense.

This fixes one of JMC's console-recorded F-Zero GX replays on JitArm64. (JitArm64 uses an interpreter fallback for fctiwx.)


JMC47:
Confirmed to sync up the replay on JitArm64 on my phone.  Doesn't seem to be a performance loss either.

lioncash (approved):

PR #9739, opened 2021-05-22
Use Correct mask for Fake VMem

phire:
Shouldn't have any behaviour change for regular usage as both masks are 32MB by default.

But fixes a theoretical buffer overrun when memory size override is larger than 32MB.


lioncash (approved):

shuffle2:
Does this still work? https://github.com/phire/dolphin/commit/cc858c63b8d04c5ce407c7efd0ae8abc654e73a0

PR #9740, opened 2021-05-22
Remove remaining hardcoded masks

phire:
Dolphin now supports overriding EXRAM size, so hardcoded masks will be incorrect

As for L1Cache, I don't think it ever makes sense to override it's size, but some of the masks were incorrect anyway.

### DIFF
--- a/Source/Core/Core/PowerPC/MMU.cpp
+++ b/Source/Core/Core/PowerPC/MMU.cpp
@@ -221,7 +221,7 @@ static T ReadFromHardware(u32 em_address)
       (em_address & 0x0FFFFFFF) < Memory::GetExRamSizeReal())
   {
     T value;
-    std::memcpy(&value, &Memory::m_pEXRAM[em_address & 0x0FFFFFFF], sizeof(T));
+    std::memcpy(&value, &Memory::m_pEXRAM[em_address & Memory::GetExRamMask()], sizeof(T));
     return bswap(value);
   }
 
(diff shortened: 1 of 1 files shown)


Source/Core/Core/PowerPC/MMU.cpp
```diff
@@ -320,7 +320,8 @@ static void WriteToHardware(u32 em_address, const T data)
       (em_address < (0xE0000000 + Memory::GetL1CacheSize())))
   {
     const T swapped_data = bswap(data);
-    std::memcpy(&Memory::m_pL1Cache[em_address & 0x0FFFFFFF], &swapped_data, sizeof(T));
+    std::memcpy(&Memory::m_pL1Cache[em_address & Memory::GetL1CacheMask()], &swapped_data,
```
BhaaLseN, line 323: This changed from `0x0FFFFFFF` to `Memory::GetL1CacheMask()`, but every other occurrence used to be `0x3FFFF`. Is this correct, or should this be `Memory::GetExRamMask()` instead?

phire: It certainly shouldn't be `GetExRamMask()`

Taking a better look at this code (and the code ExRam above), it doesn't actually implement mirroring, as the if statement is only matches for the first mirror.

All the `& 0x0fffffff` is doing is lopping off the `0x10000000` or `0xe000000`

Tilka:
The file still mentions 0x0FFFFFFF a few times. Are all remaining occurrences meant to be static?

phire:
I've decided against this cleanup

PR #9741, opened 2021-05-22
DSP: Approximate INT_AID timing based on hwtest

sepalani:
This PR fixes this issue where Syobon Action isn't booting properly: https://dolp.in/i12498

After the interrupt is triggered the game can't resume its normal execution. The interrupt callback is called too frequently preventing it to load its splash screen and continue its normal execution.

I wrote a [hardware test](https://github.com/dolphin-emu/dolphin/files/6526913/wii-dma-start.zip) to approximate **roughly** the timing of this interrupt:
hw_dma

This test tries to imitate the behaviour of an [old version of libasnd](https://github.com/devkitPro/libogc/blob/36f6bef9e1e8cb021a03d041ce0b47b568154d27/libasnd/asndlib.c#L268) that Syobon Action probably used. I say **roughly** as the timing is biased and taken at the beginning of libogc's DMA callback not right after the interrupt is triggered. Regardless, it gives a rough idea of the timing window of this interrupt.

The games I own are still working but the ones sensitive to this timing change should be tested.

### DIFF
--- a/Source/Core/Core/HW/DSP.cpp
+++ b/Source/Core/Core/HW/DSP.cpp
@@ -397,8 +397,11 @@ void RegisterMMIO(MMIO::Mapping* mmio, u32 base)
 
           // TODO: need hardware tests for the timing of this interrupt.
           // Sky Crawlers crashes at boot if this is scheduled less than 87 cycles in the future.
-          // Other Namco games crash too, see issue 9509. For now we will just push it to 200 cycles
-          CoreTiming::ScheduleEvent(200, s_et_GenerateDSPInterrupt, INT_AID);
+          // Other Namco games crash too, see issue 9509.
+          // Real hardware takes roughly ~68 microseconds to trigger libogc DMA callback based on
+          // https://dolp.in/pr9741 hardware test. The amount of cycles for this interrupt is
+          // approximated to reach this timing value.
+          CoreTiming::ScheduleEvent(17000, s_et_GenerateDSPInterrupt, INT_AID);
         }
       }));
 


Source/Core/Core/HW/DSP.cpp
```diff
@@ -397,8 +397,9 @@ void RegisterMMIO(MMIO::Mapping* mmio, u32 base)
 
           // TODO: need hardware tests for the timing of this interrupt.
```
lioncash, line 398: This TODO can probably be removed if the hardware test is added to the [hwtests](https://github.com/dolphin-emu/hwtests) repository, so its not lost track of

Source/Core/Core/HW/DSP.cpp
```diff
@@ -397,8 +397,9 @@ void RegisterMMIO(MMIO::Mapping* mmio, u32 base)
...
           // Sky Crawlers crashes at boot if this is scheduled less than 87 cycles in the future.
-          // Other Namco games crash too, see issue 9509. For now we will just push it to 200 cycles
-          CoreTiming::ScheduleEvent(200, s_et_GenerateDSPInterrupt, INT_AID);
+          // Other Namco games crash too, see issue 9509. For now we will just push it to 200 cycles.
```
lioncash, line 400: This comment is out of sync now, since it's no longer 200 cycles.

Source/Core/Core/HW/DSP.cpp
sepalani: IMO, it still needs a proper hardware test that measure the cycle it takes for this interrupt. Since my test is based on the libogc's DMA callback its result is biased. Moreover, Dolphin seems to execute instructions slightly faster than a real Wii.

The timing I chose for this interrupt is a "close-enough" value that reach the timing expected by the hardware test without taking into account Dolphin's speed. In sum, this test doesn't measure the cycles it takes but roughly approximates it.

JMC47:
Would this potentially affect other interrupt crashes?  Daring Game for Girls, Carnival whatever Wii?

sepalani:
@JMC47 It could affect them, especially if the interrupt was fired too quickly. Meanwhile, I'll write a proper hardware test to get a more accurate cycle value.

JMC47:
This does indeed fix The Daring Game for Girls.  However, it does not fix Active Life: Magical Carnival, which means those issues are indeed different.

sepalani:
Closed in favor of https://github.com/dolphin-emu/dolphin/pull/9774 and https://github.com/dolphin-emu/dolphin/pull/9776.

PR #9742, opened 2021-05-23
Software: Fix bad backbuffer size

Pokechu22:
`MAX_XFB_WIDTH`/`HEIGHT` are the largest XFB sizes seen in practice, but do not make sense to use for the backbuffer size, which should be the size of the window.  The old code created screenshots with a size of 720x540 on NTSC games when "Dump Frames at Internal Resolution" is unchecked; now, the window size is used.  See [bug 12519](https://bugs.dolphin-emu.org/issues/12519) (though note that this does not entirely fix that).

With "Auto-Adjust Window Size" checked, screenshots of Super Mario Sunshine's title screen are now 640 by 480 with or without "Dump Frames at Internal Resolution" checked.  With "Dump Frames at Internal Resolution" unchecked, resizing the window results in the screenshot being saved at a 4:3 aspect ratio based on the window's height (the software renderer itself also stretches the image, unlike other renderers which add bars on the side to maintain 4:3).

This change was based on the following code from `OGLRenderer`:

https://github.com/dolphin-emu/dolphin/blob/95aadff0e7f8e5f6bbcc0d7d650e45040b417e6b/Source/Core/VideoBackends/OGL/OGLRender.cpp#L325-L328
https://github.com/dolphin-emu/dolphin/blob/95aadff0e7f8e5f6bbcc0d7d650e45040b417e6b/Source/Core/VideoBackends/OGL/OGLRender.cpp#L1105-L1114

### DIFF
--- a/Source/Core/VideoBackends/Software/SWRenderer.cpp
+++ b/Source/Core/VideoBackends/Software/SWRenderer.cpp
@@ -27,7 +27,8 @@
 namespace SW
 {
 SWRenderer::SWRenderer(std::unique_ptr<SWOGLWindow> window)
-    : ::Renderer(static_cast<int>(MAX_XFB_WIDTH), static_cast<int>(MAX_XFB_HEIGHT), 1.0f,
+    : ::Renderer(static_cast<int>(std::max(window->GetContext()->GetBackBufferWidth(), 1u)),
+                 static_cast<int>(std::max(window->GetContext()->GetBackBufferHeight(), 1u)), 1.0f,
                  AbstractTextureFormat::RGBA8),
       m_window(std::move(window))
 {
@@ -54,6 +55,18 @@ SWRenderer::CreateFramebuffer(AbstractTexture* color_attachment, AbstractTexture
 {
   return SWFramebuffer::Create(static_cast<SWTexture*>(color_attachment),
                                static_cast<SWTexture*>(depth_attachment));
+}
+
+void SWRenderer::BindBackbuffer(const ClearColor& clear_color)
+{
+  // Look for framebuffer resizes
+  if (!m_surface_resized.TestAndClear())
+    return;
+
+  GLContext* context = m_window->GetContext();
+  context->Update();
+  m_backbuffer_width = context->GetBackBufferWidth();
+  m_backbuffer_height = context->GetBackBufferHeight();
 }
 
 class SWShader final : public AbstractShader
--- a/Source/Core/VideoBackends/Software/SWRenderer.h
+++ b/Source/Core/VideoBackends/Software/SWRenderer.h
@@ -27,6 +27,8 @@ public:
   std::unique_ptr<AbstractFramebuffer>
   CreateFramebuffer(AbstractTexture* color_attachment, AbstractTexture* depth_attachment) override;
 
+  void BindBackbuffer(const ClearColor& clear_color = {}) override;
+
   std::unique_ptr<AbstractShader> CreateShaderFromSource(ShaderStage stage,
                                                          std::string_view source) override;
   std::unique_ptr<AbstractShader> CreateShaderFromBinary(ShaderStage stage, const void* data,


phire:
> screenshots of Super Mario Sunshine's title screen are now 640 by 480 

They shouldn't be. 
For NTSC Super Mario Sunshine, the OpenGL backend does screenshots at 640x476.

It derives this by taking the original xfb/efb resolution of 640x448 and multiplying that by a calculated aspect ratio of 1.3429594, while preserving width: `640 / 1.3429594 = 476.56` and rounding that down to 476.

phire:
The 1.3429594 aspect ratio (aka 4.03:3) is an Image Aspect Ratio, not a Screen Aspect Ratio that can be assumed to be 4:3 (aka 1.3333333) 

It comes from the fact that Super Mario Sunshine doesn't use the full NTSC resolution of (from memory) 710.5 by 486 and adds black bars of 38 pixels vertically and 70.5 horizontally. It's the 710.5 by 486 area that is defined as the Screen Aspect Ratio of 4:3 and with these black bars "removed" the result image is no-longer 4:3. 

While SMS ends up being really close to 4:3, a lot of other games end up further away. The most extreme example are the cut-scenes in Rogue Squadron 2 and 3 that use an Image Aspect Ratio of ~2:1

Pokechu22:
> They shouldn't be.
> For NTSC Super Mario Sunshine, the OpenGL backend does screenshots at 640x476.

That's odd.  I was previously testing by using a fifolog, not by loading the game.  Here's some more data.  &dagger; indicates the default size based on auto-adjust window size, but it changes when the window is resized (which I tested in all cases, even with dump at internal resolution on).

Master, running SMS.

<table>
<tr><th rowspan="2">Renderer</th><th colspan="2">Dump Frames at Internal Resolution</th></tr>
<tr><th>Off</th><th>On</th></tr>
<tr><th>OpenGL</th><td>640 by 476&dagger;</td><td>640 by 476</td></tr>
<tr><th>Software</th><td>720 by 536</td><td>640 by 476</td></tr>
</table>

Master, using a SMS fifolog:

<table>
<tr><th rowspan="2">Renderer</th><th colspan="2">Dump Frames at Internal Resolution</th></tr>
<tr><th>Off</th><th>On</th></tr>
<tr><th>OpenGL</th><td>640 by 480&dagger;</td><td>640 by 480</td></tr>
<tr><th>Software</th><td>720 by 540</td><td>640 by 480</td></tr>
</table>

This PR, running SMS.

<table>
<tr><th rowspan="2">Renderer</th><th colspan="2">Dump Frames at Internal Resolution</th></tr>
<tr><th>Off</th><th>On</th></tr>
<tr><th>OpenGL</th><td>640 by 476&dagger;</td><td>640 by 476</td></tr>
<tr><th>Software</th><td><ins>640 by 476&dagger;</ins></td><td>640 by 476</td></tr>
</table>

This PR, using a SMS fifolog.

<table>
<tr><th rowspan="2">Renderer</th><th colspan="2">Dump Frames at Internal Resolution</th></tr>
<tr><th>Off</th><th>On</th></tr>
<tr><th>OpenGL</th><td>640 by 480&dagger;</td><td>640 by 480</td></tr>
<tr><th>Software</th><td><ins>640 by 480&dagger;</ins></td><td>640 by 480</td></tr>
</table>

My conclusion is that there's some weird bug with the way the fifo player chooses its size (which affects the size of the window, not just screenshots), but that's separate from the issue this PR fixes: the screenshot size when "Dump Frames at Internal Resolution" is off being different between OGL and Software.

phire:
> I was previously testing by using a fifolog, not by loading the game. 

> My conclusion is that there's some weird bug with the way the fifo player chooses its size 

The fifo player is limited because the dff file format doesn't save the VI registers, which are essential for calculating the correct aspect ratio.

So fifo player just guesses.

phire (approved):

Rumi-Larry:
> 
> 
> > I was previously testing by using a fifolog, not by loading the game.
> 
> > My conclusion is that there's some weird bug with the way the fifo player chooses its size
> 
> The fifo player is limited because the dff file format doesn't save the VI registers, which are essential for calculating the correct aspect ratio.
> 
> So fifo player just guesses.

Well, I guess that making the FIFO player more accurate, with the IV registers is more trouble than it's worth.

PR #9743, opened 2021-05-23
UnitTests: Add Version header tests

Dentomologist:
Adds tests to ensure the values in Common/Version.h are generated correctly.


JosJuice:
I'm not sure about this. This feels like relatively fragile code for something we don't have much reason to test.

Dentomologist:
This PR has two primary motivations.

One is that I think that wherever practical we should have tests for the updater and anything that interacts with it. There's still a lot more to do on that front, but this PR is one small piece of that.

The other is that the names of scm_rev_str and scm_desc_str are really vague and I lost track of how many times during this PR alone I had to double check what exactly each of them referred to. I'd like to make a PR renaming them, and having these tests present would help me and reviewers know I didn't break anything.

I meant for the tests to be fragile to changes. If a change to a variable's format is accidental then naturally we want to catch that so it doesn't cause other bugs. Meanwhile, to the best of my knowledge intentional changes here are rare, so the tests should rarely need to be changed and if they do the test changes will help document the difference in formats.

Dentomologist:
In hindsight, I agree this is unnecessary.

PR #9745, opened 2021-05-23
Add support for Apple's GameController.framework as an input device

dhannema:
Add support for input devices to be detected from Apple's GameController.framework.

So far, I have only tested this with a DualShock 4 and an Xbox One S controller.

This will allow Mac users to directly use accelerometer and gyroscope motion input (if available) without need for external DSU applications:
image

And will also allow Mac users to use the touchpad of the DS4 controller:
image


Source/Core/InputCommon/ControllerInterface/GameController/GameControllerDevice.mm
```diff
@@ -0,0 +1,109 @@
...
+{
+using MathUtil::GRAVITY_ACCELERATION; // 9.80665
+
+Controller::Controller(const GCController* controller) : m_controller(controller)
+{
+  std::vector<std::pair<std::string, GCControllerButtonInput*>> pairs;
+    
+  for (GCControllerButtonInput* item in controller.physicalInputProfile.allButtons)
+  {
+    pairs.push_back(std::make_pair(std::string([item.aliases.allObjects[0] UTF8String]), item));
```
lioncash, line 20:
```suggestion
    pairs.emplace_back(std::string([item.aliases.allObjects[0] UTF8String], item);
```

Source/Core/InputCommon/ControllerInterface/GameController/GameControllerDevice.mm
```diff
@@ -0,0 +1,109 @@
...
+  }
+    
+  sort(pairs.begin(), pairs.end(), [=](std::pair<std::string, GCControllerButtonInput*>& a, std::pair<std::string, GCControllerButtonInput*>& b)
```
lioncash, line 23:
```suggestion
  std::sort(pairs.begin(), pairs.end(), [](const auto& lhs, const auto& rhs)
```

Source/Core/InputCommon/ControllerInterface/GameController/GameControllerDevice.mm
```diff
@@ -0,0 +1,109 @@
...
+    return a.first.compare(b.first)<0;
+  }
+  );
+    
+  for (size_t i = 0; i < pairs.size(); ++i) {
+    INFO_LOG_FMT(SERIALINTERFACE, "Found button '{}'", pairs[i].first);
+    AddInput(new Button(pairs[i].second));
+  }
+    
+  if (controller.motion)
```
lioncash, line 34:
```suggestion
  if (!controller.motion)
    return;
```

Source/Core/InputCommon/ControllerInterface/GameController/GameControllerDevice.h
```diff
@@ -0,0 +1,63 @@
...
+
+#include "InputCommon/ControllerInterface/CoreDevice.h"
+#include "InputCommon/ControllerInterface/ControllerInterface.h"
+
+namespace ciface::GameController
+{
+
+class Controller : public Core::Device
+{
+private:
```
lioncash, line 15: Private interface should be below the public interface.

Source/Core/InputCommon/ControllerInterface/GameController/GameControllerDevice.h
```diff
@@ -0,0 +1,63 @@
...
+    ControlState GetState() const override;
+
+  private:
+    const GCControllerButtonInput* m_item;
+  };
+
+  class Motion : public Input
+  {
+  public:
+    enum button
```
lioncash, line 30:
```suggestion
    enum class ButtonAxis
```

Source/Core/InputCommon/ControllerInterface/GameController/GameControllerDevice.h
```diff
@@ -0,0 +1,63 @@
...
+    {
+      accel_x = 0,
```
lioncash, line 32:
```suggestion
      AccelX = 0,
```

ditto for similar formatting on the rest of the entries.

Source/Core/InputCommon/ControllerInterface/GameController/GameControllerDevice.h
```diff
@@ -0,0 +1,63 @@
...
+namespace ciface::GameController
+{
+
+class Controller : public Core::Device
+{
+private:
+  class Button : public Input
+  {
+  public:
+    Button(GCControllerButtonInput* item) : m_item(item) {}
```
lioncash, line 19:
```suggestion
    explicit Button(GCControllerButtonInput* item) : m_item(item) {}
```

Source/Core/InputCommon/ControllerInterface/GameController/GameControllerDevice.h
```diff
@@ -0,0 +1,63 @@
...
+    {
+      accel_x = 0,
+      accel_y,
+      accel_z,
+      gyro_x,
+      gyro_y,
+      gyro_z
+    };
+      
+    Motion(const GCController* item, std::string name, button button, double multiplier) : m_item(item), m_name(name), m_button(button), m_multiplier(multiplier) {}
```
lioncash, line 40:
```suggestion
    explicit Motion(const GCController* item, std::string name, button button, double multiplier) : m_item(item), m_name(name), m_button(button), m_multiplier(multiplier) {}
```

Source/Core/InputCommon/ControllerInterface/GameController/GameControllerDevice.h
```diff
@@ -0,0 +1,63 @@
...
+
+  private:
+    const GCController* m_item;
+    const std::string m_name;
+    const button m_button;
+    const double m_multiplier;
+  };
+
+public:
+  Controller(const GCController* controller);
```
lioncash, line 52:
```suggestion
  explicit Controller(const GCController* controller);
```

Source/Core/InputCommon/ControllerInterface/GameController/GameControllerDevice.mm
```diff
@@ -0,0 +1,109 @@
...
+}
+
+ControlState Controller::Button::GetState() const
+{
+  return m_item.value;
+}
+
+ControlState Controller::Motion::GetState() const
+{
+  double r = 0.0;
```
lioncash, line 77:
```suggestion
  ControlState r = 0.0;
```

Source/Core/InputCommon/ControllerInterface/GameController/GameController.mm
```diff
@@ -0,0 +1,80 @@
...
+{
+  CFNotificationCenterAddObserver(CFNotificationCenterGetLocalCenter(), s_observer, &onControllerConnect, CFSTR("GCControllerDidConnectNotification"), NULL, CFNotificationSuspensionBehaviorDeliverImmediately);
+
+  CFNotificationCenterAddObserver(CFNotificationCenterGetLocalCenter(), s_observer, &onControllerDisconnect, CFSTR("GCControllerDidDisconnectNotification"), NULL, CFNotificationSuspensionBehaviorDeliverImmediately);
+}
+
+void PopulateDevices(void* window)
+{
+  // I was unable to use [GCController controllers] in Objective-C++ and therefore needed to access this class method using objc_msgSend instead
+  typedef NSArray<GCController*>* (*send_type)(id, SEL);
```
lioncash, line 64:
```suggestion
  using SendType = NSArray<GCController*>* (*)(id, SEL);
```

Source/Core/InputCommon/ControllerInterface/GameController/GameController.mm
```diff
@@ -0,0 +1,80 @@
...
+#include "InputCommon/ControllerInterface/GameController/GameController.h"
+#include "InputCommon/ControllerInterface/GameController/GameControllerDevice.h"
+
+#include "Common/Logging/Log.h"
+
+namespace ciface::GameController
+{
+static char *s_observer="";
+
+void addController(GCController* controller)
```
lioncash, line 18:
```suggestion
static void AddController(GCController* controller)
```

Source/Core/InputCommon/ControllerInterface/GameController/GameController.mm
```diff
@@ -0,0 +1,80 @@
...
+  std::string vendorName = std::string([controller.vendorName UTF8String]);
+  INFO_LOG_FMT(SERIALINTERFACE, "Controller '{}' connected with motion={}", vendorName, controller.motion.sensorsActive);
+  NOTICE_LOG_FMT(SERIALINTERFACE, "Controller '{}' connected with motion={} and id=", vendorName, controller.motion.sensorsActive);
+
+  std::shared_ptr<ciface::GameController::Controller> j=std::make_shared<Controller>(controller);
+    
+  g_controller_interface.AddDevice(j);
+}
+
+static void onControllerConnect(CFNotificationCenterRef center, void* observer, CFStringRef name, const void* object, CFDictionaryRef userInfo)
```
lioncash, line 31:
```suggestion
static void OnControllerConnect(CFNotificationCenterRef center, void* observer, CFStringRef name, const void* object, CFDictionaryRef userInfo)
```

Source/Core/InputCommon/ControllerInterface/GameController/GameController.mm
```diff
@@ -0,0 +1,80 @@
...
+{
+  GCController* controller = (GCController*)object;
+  addController(controller);
+}
+
+static void onControllerDisconnect(CFNotificationCenterRef center, void* observer, CFStringRef name, const void* object, CFDictionaryRef userInfo)
```
lioncash, line 37:
```suggestion
static void OnControllerDisconnect(CFNotificationCenterRef center, void* observer, CFStringRef name, const void* object, CFDictionaryRef userInfo)
```

Source/Core/InputCommon/ControllerInterface/GameController/GameController.mm
```diff
@@ -0,0 +1,80 @@
...
+  NOTICE_LOG_FMT(SERIALINTERFACE, "Controller '{}' connected with motion={} and id=", vendorName, controller.motion.sensorsActive);
+
+  std::shared_ptr<ciface::GameController::Controller> j=std::make_shared<Controller>(controller);
+    
+  g_controller_interface.AddDevice(j);
+}
+
+static void onControllerConnect(CFNotificationCenterRef center, void* observer, CFStringRef name, const void* object, CFDictionaryRef userInfo)
+{
+  GCController* controller = (GCController*)object;
```
lioncash, line 33:
```suggestion
  auto* controller = static_cast<GCController*>(object);
```

Source/Core/InputCommon/ControllerInterface/GameController/GameController.mm
```diff
@@ -0,0 +1,80 @@
...
+  addController(controller);
+}
+
+static void onControllerDisconnect(CFNotificationCenterRef center, void* observer, CFStringRef name, const void* object, CFDictionaryRef userInfo)
+{
+  GCController *inController = (GCController*)object;
```
lioncash, line 39:
```suggestion
  auto* in_controller = static_cast<GCController*>(object);
```

Source/Core/InputCommon/ControllerInterface/GameController/GameController.mm
```diff
@@ -0,0 +1,80 @@
...
+    
+  std::string vendorName = std::string([inController.vendorName UTF8String]);
```
lioncash, line 41:
```suggestion
  std::string vendor_name = std::string([inController.vendorName UTF8String]);
```

Source/Core/InputCommon/ControllerInterface/GameController/GameController.mm
```diff
@@ -0,0 +1,80 @@
...
+
+void addController(GCController* controller)
+{
+  controller.motion.sensorsActive = true;
+
+  std::string vendorName = std::string([controller.vendorName UTF8String]);
+  INFO_LOG_FMT(SERIALINTERFACE, "Controller '{}' connected with motion={}", vendorName, controller.motion.sensorsActive);
+  NOTICE_LOG_FMT(SERIALINTERFACE, "Controller '{}' connected with motion={} and id=", vendorName, controller.motion.sensorsActive);
+
+  std::shared_ptr<ciface::GameController::Controller> j=std::make_shared<Controller>(controller);
```
lioncash, line 26: This can just collapse into:

```suggestion
  g_controller_interface.AddDevice(std::make_shared<Controller>(controller));
```

Source/Core/InputCommon/ControllerInterface/GameController/GameController.mm
```diff
@@ -0,0 +1,80 @@
...
+
+namespace ciface::GameController
+{
+static char *s_observer="";
+
+void addController(GCController* controller)
+{
+  controller.motion.sensorsActive = true;
+
+  std::string vendorName = std::string([controller.vendorName UTF8String]);
```
lioncash, line 22:
```suggestion
  std::string vendor_name = std::string([controller.vendorName UTF8String]);
```

Source/Core/InputCommon/ControllerInterface/GameController/GameController.mm
```diff
@@ -0,0 +1,80 @@
...
+    if (controller && controller->IsSameDevice(inController))
+      return true;
+
+    return false;
+  });
+}
+
+void Init(void* window)
+{
+  CFNotificationCenterAddObserver(CFNotificationCenterGetLocalCenter(), s_observer, &onControllerConnect, CFSTR("GCControllerDidConnectNotification"), NULL, CFNotificationSuspensionBehaviorDeliverImmediately);
```
lioncash, line 56: Prefer using `nullptr` over NULL. Ditto for other applicable cases elsewhere.

Source/Core/InputCommon/ControllerInterface/GameController/GameControllerDevice.mm
```diff
@@ -0,0 +1,109 @@
...
+// Refer to the license.txt file included.
+
+#include "InputCommon/ControllerInterface/GameController/GameControllerDevice.h"
+
+#include "Common/Logging/Log.h"
+#include "Common/MathUtil.h"
+
+namespace ciface::GameController
+{
+using MathUtil::GRAVITY_ACCELERATION; // 9.80665
```
lioncash, line 12: Maybe move this into the constructor, given it's only used in there.

Source/Core/InputCommon/ControllerInterface/GameController/GameController.mm
```diff
@@ -0,0 +1,80 @@
...
+
+  CFNotificationCenterAddObserver(CFNotificationCenterGetLocalCenter(), s_observer, &onControllerDisconnect, CFSTR("GCControllerDidDisconnectNotification"), NULL, CFNotificationSuspensionBehaviorDeliverImmediately);
+}
+
+void PopulateDevices(void* window)
+{
+  // I was unable to use [GCController controllers] in Objective-C++ and therefore needed to access this class method using objc_msgSend instead
+  typedef NSArray<GCController*>* (*send_type)(id, SEL);
+  send_type func = (send_type)objc_msgSend;
+  NSArray<GCController*>* controllers=(NSArray<GCController*>*)func(objc_getClass("GCController"), sel_getUid("controllers"));
```
lioncash, line 66:
```suggestion
  auto* controllers = func(objc_getClass("GCController"), sel_getUid("controllers"));
```

Source/Core/InputCommon/ControllerInterface/GameController/GameController.h
```diff
@@ -0,0 +1,10 @@
+// Copyright 2010 Dolphin Emulator Project
+// Licensed under GPLv2+
+// Refer to the license.txt file included.
+
+namespace ciface::GameController
+{
+void Init(void* window);
+void PopulateDevices(void* window);
```
Filoppi, line 8: the window ptr is never used

Source/Core/InputCommon/ControllerInterface/GameController/GameController.mm
```diff
@@ -0,0 +1,80 @@
...
+
+  for (GCController* controller in controllers)
+  {
+    addController(controller);
+  }
+}
+
+void DeInit()
+{
```
Filoppi, line 75: Could you add some safety measure to make sure DeInit doesn't break if called when Init hasn't been called? Or is that the case already?

Source/Core/InputCommon/ControllerInterface/GameController/GameControllerDevice.mm
```diff
@@ -0,0 +1,109 @@
...
+      AddInput(new Motion(controller, "Gyro Yaw Left", Motion::gyro_z, 1));
+      AddInput(new Motion(controller, "Gyro Yaw Right", Motion::gyro_z, -1));
+    }
+ 
+  }
+}
+
+std::string Controller::GetSource() const
+{
+  return "GCF";
```
Filoppi, line 62: constexpr somewhere

Source/Core/InputCommon/ControllerInterface/GameController/GameControllerDevice.mm
Filoppi: Not like GRAVITY_ACCELERATION is likely to change, but why copy its value in the comment?

Source/Core/InputCommon/ControllerInterface/ControllerInterface.cpp
```diff
@@ -16,6 +16,7 @@
 #include "InputCommon/ControllerInterface/Xlib/XInput2.h"
 #endif
 #ifdef CIFACE_USE_OSX
+#include "InputCommon/ControllerInterface/GameController/GameController.h"
```
Filoppi, line 19: Is this PR WIP? Init/PopulateDevices/DeInit aren't plugged in in the ControllerInterface.
I don't understand how you even tested this.

Source/Core/InputCommon/ControllerInterface/GameController/GameController.mm
```diff
@@ -0,0 +1,80 @@
...
+#import <objc/message.h>
+
+#include "InputCommon/ControllerInterface/GameController/GameController.h"
+#include "InputCommon/ControllerInterface/GameController/GameControllerDevice.h"
+
+#include "Common/Logging/Log.h"
+
+namespace ciface::GameController
+{
+static char *s_observer="";
```
Filoppi, line 16: Why assigning a "" to a ptr?

Source/Core/InputCommon/ControllerInterface/GameController/GameController.mm
```diff
@@ -0,0 +1,80 @@
...
+
+void addController(GCController* controller)
+{
+  controller.motion.sensorsActive = true;
+
+  std::string vendorName = std::string([controller.vendorName UTF8String]);
+  INFO_LOG_FMT(SERIALINTERFACE, "Controller '{}' connected with motion={}", vendorName, controller.motion.sensorsActive);
+  NOTICE_LOG_FMT(SERIALINTERFACE, "Controller '{}' connected with motion={} and id=", vendorName, controller.motion.sensorsActive);
```
Filoppi, line 24: This is CONTROLLERINTERFACE now, I've recently added a new log type.

Source/Core/InputCommon/ControllerInterface/GameController/GameControllerDevice.mm
Filoppi: Personal preference I guess, but I'd use m_controller instead of controller inside this function.

Source/Core/InputCommon/ControllerInterface/GameController/GameControllerDevice.mm
Filoppi: Is item.aliases.allObjects[0] guaranteed to be valid?

Source/Core/InputCommon/ControllerInterface/GameController/GameControllerDevice.mm
```diff
@@ -0,0 +1,109 @@
...
+      INFO_LOG_FMT(SERIALINTERFACE, "Adding gyro inputs");
+            
+      AddInput(new Motion(controller, "Gyro Pitch Up", Motion::gyro_x, 1));
+      AddInput(new Motion(controller, "Gyro Pitch Down", Motion::gyro_x, -1));
+      AddInput(new Motion(controller, "Gyro Roll Left", Motion::gyro_y, 1));
+      AddInput(new Motion(controller, "Gyro Roll Right", Motion::gyro_y, -1));
+      AddInput(new Motion(controller, "Gyro Yaw Left", Motion::gyro_z, 1));
+      AddInput(new Motion(controller, "Gyro Yaw Right", Motion::gyro_z, -1));
+    }
+
```
Filoppi, line 56: Extra space

Source/Core/InputCommon/ControllerInterface/GameController/GameController.mm
Filoppi: I don't particularly care but I think this string be const

Source/Core/InputCommon/ControllerInterface/GameController/GameController.mm
Filoppi: Does this trigger the callback for devices that are connection on Init, or does it only call it for devices that will be connected after binding to it? I guess it's the second case, but please specify it in a comment.

Filoppi: Also there is a space between the two functions here, but there isn't in DeInit()

Source/Core/InputCommon/ControllerInterface/GameController/GameController.mm
```diff
@@ -0,0 +1,80 @@
+// Copyright 2013 Dolphin Emulator Project
+// Licensed under GPLv2+
+// Refer to the license.txt file included.
+
+#include <random>
```
Filoppi, line 5: This doesn't seem to be used?

Source/Core/InputCommon/ControllerInterface/GameController/GameController.mm
```diff
@@ -0,0 +1,80 @@
...
+
+#include "Common/Logging/Log.h"
+
+namespace ciface::GameController
+{
+static char *s_observer="";
+
+void addController(GCController* controller)
+{
+  controller.motion.sensorsActive = true;
```
Filoppi, line 20: What is this for? Why is it not "deactivated" on disconnection? Is it relevant? Persistent?

Source/Core/InputCommon/ControllerInterface/GameController/GameController.mm
Filoppi: The ID for the notice log doesn't seem to be specified?
Do we really need two different logs of different verbosity?

Source/Core/InputCommon/ControllerInterface/GameController/GameController.mm
```diff
@@ -0,0 +1,80 @@
...
+void PopulateDevices(void* window)
+{
+  // I was unable to use [GCController controllers] in Objective-C++ and therefore needed to access this class method using objc_msgSend instead
+  typedef NSArray<GCController*>* (*send_type)(id, SEL);
+  send_type func = (send_type)objc_msgSend;
+  NSArray<GCController*>* controllers=(NSArray<GCController*>*)func(objc_getClass("GCController"), sel_getUid("controllers"));
+
+  for (GCController* controller in controllers)
+  {
+    addController(controller);
```
Filoppi, line 70: You should remove all devices which have the same source name as ours before re-adding them.

Source/Core/InputCommon/ControllerInterface/GameController/GameControllerDevice.mm
dhannema: According to https://developer.apple.com/documentation/gamecontroller/gccontrollerelement, the aliases object is always present, but I will add a check if the array contains more than 0 entries

Source/Core/InputCommon/ControllerInterface/GameController/GameController.mm
dhannema: I have checked this and the CFNotificationCenterRemoveObserver methods in DeInit() can already be called multiple times without any problems, so it looks like an explicit safety is not needed.

Filoppi: Ok, maybe specify that in a comment, it wouldn't hurt.

Source/Core/InputCommon/ControllerInterface/ControllerInterface.cpp
dhannema: I'm sorry for the confusion. I own an M1 Mac and had to test/develop in skylersaleh's fork as I am unable to build Dolphin on the main branch.
I forgot to manually copy the other changes I made in ControllerInterface.cpp

Source/Core/InputCommon/ControllerInterface/GameController/GameController.mm
dhannema: This is to ensure s_observer is an actual pointer to something. Otherwise the pointer will always be 0.

Source/Core/InputCommon/ControllerInterface/GameController/GameController.mm
Filoppi: What does s_observer do? Is it assigned in Init? Maybe you could set it to null after DeInit?

Source/Core/InputCommon/ControllerInterface/GameController/GameController.mm
```diff
@@ -0,0 +1,80 @@
...
+static void onControllerDisconnect(CFNotificationCenterRef center, void* observer, CFStringRef name, const void* object, CFDictionaryRef userInfo)
+{
+  GCController *inController = (GCController*)object;
+    
+  std::string vendorName = std::string([inController.vendorName UTF8String]);
+  INFO_LOG_FMT(SERIALINTERFACE, "Controller '{}' disconnected", vendorName);
+    
+  g_controller_interface.RemoveDevice([&inController](const auto* obj)
+  {
+    const Controller* controller = dynamic_cast<const Controller*>(obj);
```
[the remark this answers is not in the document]
dhannema: I borrowed this code from the already existing OSX.mm, so I guess it is not a problem?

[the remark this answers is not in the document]
Filoppi: Yes I know it's there as well. Not a problem to me, but on MSVC if you try to add a dynamic_cast it refused to build as it's been disabled.

Source/Core/InputCommon/ControllerInterface/GameController/GameController.mm
Filoppi: Not sure I understand.
I've never allocated a ptr like that.
Does that create a new object and assign it to the pointer?

Filoppi: If I understand what this is, you should probably assign "dolphin" to that char array.

Filoppi: Why isn't this a constexpr or a const which you then pass as a reference to the callbacks bindings functions?

JosJuice: It ensures that there is an empty string in the constant data segment of the binary, and makes the pointer point to it. If this is to be done, the variable should be marked `const char*` instead of just `char*`, because writing to such a string is not allowed.

dhannema: I double checked the manual (https://developer.apple.com/documentation/corefoundation/1543316-cfnotificationcenteraddobserver?language=objc) and it looks like I can completely remove the s_observer variable.
I will remove it and pass nullptr to the callback functions!

Source/Core/InputCommon/ControllerInterface/GameController/GameController.mm
dhannema: @Filoppi during my tests, it seems like this is working properly. When I click 'Refresh' in the Device popup screen the controllers are added properly again.

Filoppi: It does work properly in the way the controller interface is currently setup, meaning that it cleans all devices before calling PopulateDevices() on each input backend, but it doesn't hurt to also remove your devices manually, and some input backends do. It would avoid duplicates if the controller interface design ever changed.

skylersaleh:
@dolphin-emu-bot rebuild

skylersaleh:
@dolphin-emu-bot rebuild

skylersaleh:
@dhannema Could you rebase your changes to the latest master to pick up the changes to support M1 builds, so that the osx-universal buildbot will be able to run?

Source/Core/InputCommon/ControllerInterface/GameController/GameController.mm
```diff
@@ -0,0 +1,92 @@
...
+#include "InputCommon/ControllerInterface/GameController/GameController.h"
+#include "InputCommon/ControllerInterface/GameController/GameControllerDevice.h"
+
+#include "Common/Logging/Log.h"
+
+namespace ciface::GameController
+{
+static void AddController(GCController* controller)
+{
+  if (!@available(macOS 11.0, *)) {
```
OatmealDome, line 16: You can't negate an ``@available`` check. An alternative might be
```objc
if (@available(macOS 11.0, *))
{
}
else
{
    INFO_LOG_FMT(...);
    return;
}
```

But admittedly, it is a bit ugly.

skylersaleh:
@dolphin-emu-bot rebuild

canesalato:
I tested this on my M1 MacBook and it seems to work well, both when running native (arm) and under Rosetta2

adityajs12321:
Tested this on my m1 mac and a DualShock4, can confirm it works well

kicksluci:
Can confirm this works even with a DS5 (DualSense) controller!

Rumi-Larry:
Protip: When making small changes, is always better to ammend the main commits than create a new one... Also hit the weak spot for massive damage!

jordan-woyak:
SDL should now provide accelerometer/gyroscope input.
Does this "GameController Famework" provide any additional functionality on macOS over our SDL input backend?
Unfortunately macOS-specific code can tend to quickly become unmaintained in dolphin so if SDL does everything we need I think it might be best to just rely on SDL for macOS gamepad input.

dhannema:
> Does this "GameController Famework" provide any additional functionality on macOS over our SDL input backend? 

You are correct. We are 3 years later and indeed SDL also provides this functionality now.
This PR can be closed !

PR #9748, opened 2021-05-24
Interpreter/Jit64: Emulate FMA accurately in more cases

JosJuice:
Thought my efforts to improve cross-architecture netplay sync would only involve making JitArm64 more accurate? Nope! It's time to make Jit64 more accurate now. (And the interpreter too, as a bonus.)

### DIFF
--- a/Source/Core/Core/PowerPC/Interpreter/Interpreter_FPUtils.h
+++ b/Source/Core/Core/PowerPC/Interpreter/Interpreter_FPUtils.h
@@ -238,7 +238,7 @@ inline FPResult NI_sub(UReg_FPSCR* fpscr, double a, double b)
 // inputs are checked for NaN is still a, b, c.
 inline FPResult NI_madd(UReg_FPSCR* fpscr, double a, double c, double b)
 {
-  FPResult result{a * c};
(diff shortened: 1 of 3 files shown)


Sintendo:
Commit message for the Jit64 commit appears to have been truncated.

JosJuice:
Fixed.

Source/Core/Core/PowerPC/Interpreter/Interpreter_FPUtils.h
```diff
@@ -268,7 +268,7 @@ inline FPResult NI_madd(UReg_FPSCR* fpscr, double a, double c, double b)
     return result;
   }
 
-  result.value += b;
+  result.value = std::fma(a, c, b);
```
degasus, line 271: Is there still a point in checking if `a*c` is a NaN? Especially as the `isnan(b)` handling is in there as well?

Tilka (approved):
lgtm, but please make sure this has been tested before merging

JosJuice:
I've tested that F-Zero GX runs correctly with all the Jit64 codepaths (hardware FMA, software FMA, no FMA).

PR #9749, opened 2021-05-24
Updater: Escape HTML characters in commit descriptions

Dentomologist:
Escapes &<>'" characters in the "Update available" modal's short commit descriptions.

Before:
Before

After:
After

Fixes https://bugs.dolphin-emu.org/issues/12521

### DIFF
--- a/Source/Core/Common/StringUtil.cpp
+++ b/Source/Core/Common/StringUtil.cpp
@@ -5,6 +5,7 @@
 #include "Common/StringUtil.h"
 
 #include <algorithm>
+#include <array>
 #include <codecvt>
 #include <cstdarg>
 #include <cstddef>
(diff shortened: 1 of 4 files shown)


Source/Core/Common/StringUtil.cpp
```diff
@@ -664,3 +665,19 @@ std::vector<std::string> CommandLineToUtf8Argv(const wchar_t* command_line)
   return argv;
 }
 #endif
+
+std::string GetEscapedHtml(std::string html)
+{
+  static constexpr std::array<std::array<const char*, 2>, 5> replacements{
+      // Escape ampersand first to avoid escaping the ampersands in other replacements
+      std::array<const char*, 2>{"&", "&amp;"},  std::array<const char*, 2>{"<", "&lt;"},
```
lioncash, line 673: You should be able to compress these down to just the type name like:

```suggestion
      std::array{"&", "&amp;"},  std::array{"<", "&lt;"},
```

Dentomologist: I tried to do this but freebsd and macOS wouldn't let me.

lioncash: Ah, both freebsd and macOS probably have old compilers installed on the buildbots then

Source/Core/Common/StringUtil.cpp
```diff
@@ -664,3 +665,19 @@ std::vector<std::string> CommandLineToUtf8Argv(const wchar_t* command_line)
   return argv;
 }
 #endif
+
+std::string GetEscapedHtml(std::string html)
+{
+  static constexpr std::array<std::array<const char*, 2>, 5> replacements{
```
leoetlino, line 671:
```cpp
static constexpr std::array<std::array<const char*, 2>, 5> replacements{{
    {{"&", "&amp;"}},
    {{"<", "&lt;"}},
    {{">", "&gt;"}},
    {{"\"", "&quot;"}},
    {{"'", "&apos;"}},
}};
```

Does that work? (It seems to compile fine in godbolt for GCC >=9 and Clang >= 9 at least)

Dentomologist: Yes it does.

PR #9751, opened 2021-05-25
JitArm64: Make fcmpX with flush-to-zero enabled less bad

JosJuice:
See the added code comment for details. Fixes Pokémon Battle Revolution not progressing past the title screen.

### DIFF
--- a/Source/Core/Common/ArmFPURoundMode.cpp
+++ b/Source/Core/Common/ArmFPURoundMode.cpp
@@ -2,8 +2,10 @@
 // Licensed under GPLv2+
 // Refer to the license.txt file included.
 
+#include "Common/CPUDetect.h"
 #include "Common/CommonTypes.h"
 #include "Common/FPURoundMode.h"
+#include "Common/Logging/Log.h"
 
 #ifdef _MSC_VER
 #include <intrin.h>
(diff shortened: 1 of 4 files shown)


Source/Core/Common/ArmFPURoundMode.cpp
```diff
@@ -45,8 +47,25 @@ void SetPrecisionMode(PrecisionMode mode)
...
+  // FZ controls flush-to-zero for outputs, and FIZ controls flush-to-zero for inputs.
   constexpr u32 FZ = 1 << 24;
+  constexpr u32 AH = 1 << 1;
+  constexpr u32 FIZ = 1 << 0;
+  constexpr u32 flush_to_zero_mask = FZ | AH | FIZ;
+
+  // Setting AH = 1, FZ = 1, FIZ = 0 emulates the GC/Wii CPU's "non-IEEE mode". Unfortunately,
+  // the AH and FIZ bits were not added until 2020, so we can't count on setting AH actually
+  // doing anything. But flushing both inputs and outputs seems to cause less problems than
+  // flushing nothing, so let's just set FZ and AH and roll with whatever behavior we get.
```
Rumi-Larry, line 60: Would this change in the near future? I'm confused about the relevance of the year.

JosJuice: In the future, CPUs will be released which support the AH bit. So for each year that passes, the percentage of users that have a CPU with support for it will increase.

JosJuice: Maybe I should clarify that "added in 2020" means added to AArch64, not to Dolphin.

JosJuice: Reworded.

Rumi-Larry: Ok, so it's a hardware limitation, not of drivers or AArch itself. Got it

PR #9752, opened 2021-05-25
Android: Disable building unit tests by default

JosJuice:
This reduces the build time for incremental builds from about 2 minutes to about 20 seconds. Most people never run the unit tests on Android anyway (I'm not aware of anyone other than me ever having done it).


leoetlino:
We should probably make sure they're still built on the Android buildbot, though?

JosJuice:
I was thinking that there wasn't as much reason to do so anymore now that we have the pr-osx-universal builder, but if it's possible to do it then I guess why not. I'll have to take a look at how that can be set later.

leoetlino:
Eh, you're right. I guess there's not much point to running the unit tests on Android.

PR #9753, opened 2021-05-26
VolumeVerifier: increase problem severity for incorrectly signed TMDs

mbc07:
To avoid problems, Dolphin currently enforces a 512 MB size limit to the emulated Wii NAND. Since this can be rather small (especially if you have a lot of Virtual Console or WiiWare titles), the Wii System Menu 4.0+ ability to move channels to the SD Card is extremely helpful, especially when you can create virtual SD Cards as big as 32 GB and have them working without issues with Dolphin.

While moving titles to the virtual SD Card, I stumbled upon a few that would behave oddly: Dolphin could successfully launch them if booted directly from the WAD, as well as if booted from inside the Wii System Menu while they were stored on the NAND. From the data management screen, System Menu would also successfully copy or move those channels to the SD Card without errors. However, the transfer was one-way only: once stored on the SD Card, Wii System Menu would refuse to launch them ("This channel failed to load") as well as refuse to move or copy them back to the NAND ("The data may not have been copied/moved").

After a couple of tests and checks between the WADs that worked and the ones that didn't, I found the culprit: an incorrectly signed TMD. When a channel is stored on the SD Card, Wii System Menu will temporarily import it to the NAND before launching from there, checking the TMD in the process and refusing contents with incorrectly signed TMDs. The same happens when copying/moving content from the SD Card to the NAND, contents with incorrectly signed TMDs are refused. It doesn't seem to care about incorrectly signed tickets, however, those could be launched just fine from the SD Card.

Dolphin also doesn't seem to care about an incorrectly signed TMD when booting or installing WADs and the Wii System Menu apparently doesn't check the TMD signature either for contents already installed on the NAND, but it becomes an issue in the other scenarios (launching from the SD Card and copying/moving from the SD Card to the NAND).

So, **TL;DR**, this PR increases the problem severity on VolumeVerifier for WADs with incorrectly signed TMDs to medium and updates its description. Ccing @JosJuice and @leoetlino since this is related to areas you've extensively worked in the past...

### DIFF
--- a/Source/Core/DiscIO/VolumeVerifier.cpp
+++ b/Source/Core/DiscIO/VolumeVerifier.cpp
@@ -971,7 +971,11 @@ void VolumeVerifier::CheckMisc()
         es->VerifyContainer(IOS::HLE::ESDevice::VerifyContainerType::TMD,
                             IOS::HLE::ESDevice::VerifyMode::DoNotUpdateCertStore, tmd, cert_chain))
     {
-      AddProblem(Severity::Low, Common::GetStringT("The TMD is not correctly signed."));
+      AddProblem(
+          Severity::Medium,
+          Common::GetStringT("The TMD is not correctly signed. If you move or copy this title to "
+                             "the SD Card, the Wii System Menu will not launch it anymore and will "
+                             "also refuse to copy or move it back to the NAND."));
     }
   }
 


JosJuice:
Have you confirmed that the issue happens when we have "The TMD is not correctly signed." rather than when we have "The specified common key index is {0} but should be {1}."? The two usually co-occur.

JosJuice:
> Yes. For instance, I had a WAD with only "The specified common key index is {0} but should be {1}." problem, that one installed fine and still launched once moved to the SD Card, as Dolphin seems to account for that when importing the WAD (as suggested by this comment).

How did you manage to get a correctly signed WAD that has this problem and that installs fine? I thought that was impossible...

> On a second test, I tried the same on another WAD with no problems ("This is a good dump."), after patching it got only "The TMD is not correctly signed." problem and failed to launch once moved to the SD Card.

Okay, it should be safe to assume that the relevant problem isn't "The specified common key index is {0} but should be {1}." then. Maybe leoetlino knows more about me why the signing of the TMD would be checked but not the ticket.

mbc07:
> How did you manage to get a correctly signed WAD that has this problem and that installs fine? I thought that was impossible...

Some WADs I have are several years old, from when I first started tinkering with Wii homebrew. I used to have a lot of bad dumps back then, but slowly improved as I learned more and switched to better dumping tools. The WAD with just the common key problem comes from that very first batch and I don't remember which tool produced it anymore, sorry.

> Okay, it should be safe to assume that the relevant problem isn't "The specified common key index is {0} but should be {1}." then. Maybe leoetlino knows more about me why the signing of the TMD would be checked but not the ticket.

For instance, I also tested a WAD that has both the common key as well as "the ticket is not correctly signed" problems, that one also launches fine once moved to the SD Card. From the bad WADs I tested, the only problem that triggers the issue seems to be a badly signed TMD...

leoetlino:
> From the bad WADs I tested, the only problem that triggers the issue seems to be a badly signed TMD...

This makes sense, because the system menu cannot access the ticket and cannot validate whether it is correctly signed. On the other hand, it *does* have access to the entire TMD so this change sounds good to me.

leoetlino (approved):

PR #9756, opened 2021-05-27
Replace MemoryCard251 setting with generic MemoryCardSize

Techjar:
This allows for a broader range of game-specific memory card sizes to be configured. Needed for Japanese version of Pokémon Box: Ruby & Sapphire, as that game requires a 59 block memory card.

Fixes https://bugs.dolphin-emu.org/issues/12185

The value follows a power of two scale, with 4 Mbit (59 blocks) as the base, so the sizes are as follows:
0 = 4 Mbit (59 blocks)
1 = 8 Mbit (123 blocks)
2 = 16 Mbit (251 blocks)
3 = 32 Mbit (507 blocks)
4 = 64 Mbit (1019 blocks)

Any value outside this range of 0-4 is simply ignored. 5 is technically valid, but we don't handle it specially as it would just be the default/maximum memory card size of 128 Mbit (2043 blocks).

### DIFF
--- a/Source/Core/Core/HW/EXI/EXI_DeviceMemoryCard.cpp
+++ b/Source/Core/Core/HW/EXI/EXI_DeviceMemoryCard.cpp
@@ -236,8 +236,11 @@ void CEXIMemoryCard::SetupRawMemcard(u16 size_mb)
       SConfig::GetDirectoryForRegion(SConfig::ToGameCubeRegion(SConfig::GetInstance().m_region));
   MemoryCard::CheckPath(filename, region_dir, is_slot_a);
 
-  if (size_mb == Memcard::MBIT_SIZE_MEMORY_CARD_251)
-    filename.insert(filename.find_last_of('.'), ".251");
+  if (size_mb < Memcard::MBIT_SIZE_MEMORY_CARD_2043)
+  {
+    filename.insert(filename.find_last_of('.'),
+                    fmt::format(".{}", Memcard::MbitToFreeBlocks(size_mb)));
+  }
 
   m_memory_card = std::make_unique<MemoryCard>(filename, m_card_index, size_mb);
 }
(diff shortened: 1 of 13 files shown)


Source/Core/Core/HW/EXI/EXI_DeviceMemoryCard.cpp
```diff
@@ -236,8 +236,11 @@ void CEXIMemoryCard::SetupRawMemcard(u16 size_mb)
       SConfig::GetDirectoryForRegion(SConfig::ToGameCubeRegion(SConfig::GetInstance().m_region));
   MemoryCard::CheckPath(filename, region_dir, is_slot_a);
 
-  if (size_mb == Memcard::MBIT_SIZE_MEMORY_CARD_251)
-    filename.insert(filename.find_last_of('.'), ".251");
+  if (size_mb < Memcard::MBIT_SIZE_MEMORY_CARD_2043)
+  {
+    filename.insert(filename.find_last_of('.'),
+                    fmt::format(".{}", size_mb * Memcard::MBIT_TO_BLOCKS - Memcard::MC_FST_BLOCKS));
```
lioncash, line 242: Perhaps GCMemcard.h should have a helper function like:

```cpp
// Or whatever name is more appropriate
constexpr u32 ToBlockSize(u32 size_mb) {
  return (size_mb * Memcard::MBIT_TO_BLOCKS) - Memcard::MC_FST_BLOCKS;
}
```

Given the expression is primarily the same in several different places, with the only difference being that the size_mb part of the expression is different.

Techjar: Heh, I was just thinking exactly the same thing after I submitted the PR.

Techjar: How's that?

Tilka (approved):

Rumi-Larry:
I do wonder if it would be possible to write some patches for the games to work with larger memory cards.
I would name the setting "MaximumMemoryCardSize" or "RequiredMemoryCardSize"

leoetlino (approved):
The path manip should probably be refactored, but the logic was already duplicated prior to this PR and this can be cleaned up in a follow-up PR. LGTM

PR #9761, opened 2021-05-28
JitArm64: Flush outputs to zero in software

JosJuice:
This is an alternative to PR #9751. This PR fixes all instances of inputs being flushed to zero, whereas that PR only fixes one specific instance of inputs being flushed to zero, but the performance impact of this PR is much greater.

Needs performance testing.


JMC47:
Maybe it's because I'm GPU limited but I don't really see much of a performance impact.  We'd probably need Skyler to test this.  Tested with F-Zero GX.

JosJuice:
This PR has a big flaw that I somehow didn't realize about until now: We flush to zero even if the game doesn't want it. Marking as a draft because of this problem.

PR #7039, opened 2018-05-31
Vulkan: macOS support via MoltenVK

stenzek:
It seems MoltenVK is now mature enough to handle the shaders Dolphin generates without completely falling over. It will be interesting to see how it performs compared to OpenGL, but more importantly, it enables mac users to use features blocked by the GL version, e.g. bounding box.

We'll also have to decide how to distribute MoltenVK. Currently, the loader searches for an environment variable named `LIBMOLTENVK_PATH`, and if present uses that. Otherwise, it looks in the application bundle directory, under `DolphinWx.app/Contents/Frameworks/libMoltenVK.dylib`.

I'd suggest building MoltenVK externally, and copying the library in as a buildbot step. IMO, this is better than including yet another huge externals dependency, for a single platform.

**Known limitations:**

- In single core mode, enabling multithreading in graphics options can lead to lower performance. In dual core mode, it seems similar. Probably best to disable it though, due to MoltenVK doing its own threading internally.
- Ubershaders cause the shader compiler to freak out. I haven't looked into whether this is a SPIRV-Cross (via MoltenVK) issue, or we need to update our glslang.

PR #9762, opened 2021-05-28
VideoCommon: Fix bounding box on AMD/OpenGL

Techjar:
According to @Pokechu22, using `int4` instead of 4 `int`s fixes bounding box being broken on AMD with OpenGL.


Techjar:
Metal can't deal with an `int4` here (see https://github.com/dolphin-emu/dolphin/pull/7039/commits/0c0d66809d9c029c0468982725ab63be48500081), and none of the other attempted workarounds fix the problem, so this will just have to remain broken.

PR #9763, opened 2021-05-29
NetPlay: Refactor some functions into a common header

Techjar:
This moves some functions to a common place so that both sides can use them, which may be needed for future features. Also reduces the amount of code in the client and server a little bit. I also removed an unused struct, and renamed/restructured some things.

### DIFF
--- a/Source/Core/Core/NetPlayCommon.cpp
+++ b/Source/Core/Core/NetPlayCommon.cpp
@@ -1,0 +1,227 @@
+// Copyright 2021 Dolphin Emulator Project
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+#include "Core/NetPlayCommon.h"
+
+#include <lzo/lzo1x.h>
+
+#include "Common/FileUtil.h"
+#include "Common/IOFile.h"
+#include "Common/MsgHandler.h"
+#include "Common/SFMLHelper.h"
+
+namespace NetPlay
+{
+constexpr u32 LZO_IN_LEN = 1024 * 64;
+constexpr u32 LZO_OUT_LEN = LZO_IN_LEN + (LZO_IN_LEN / 16) + 64 + 3;
+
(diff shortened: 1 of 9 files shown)


Source/Core/Core/NetPlayCommon.h
```diff
@@ -0,0 +1,25 @@
...
+#include <array>
+#include <optional>
+#include <string>
+#include <vector>
+
+#include "Common/CommonTypes.h"
+
+namespace NetPlay
+{
+constexpr u32 LZO_IN_LEN = 1024 * 64;
```
leoetlino, line 18: Does this actually need to be exposed publicly?

Techjar: It doesn't now I suppose.

Source/Core/Core/NetPlayCommon.cpp
```diff
@@ -0,0 +1,228 @@
...
+#include <lzo/lzo1x.h>
+
+#include "Common/FileUtil.h"
+#include "Common/IOFile.h"
+#include "Common/MsgHandler.h"
+#include "Common/SFMLHelper.h"
+
+namespace NetPlay
+{
+static constexpr u32 LZO_IN_LEN = 1024 * 64;
```
leoetlino, line 16: constexpr variables at namespace scope already have internal linkage by default, so `static` is not required here.

Source/Core/Core/NetPlayCommon.cpp
```diff
@@ -0,0 +1,228 @@
...
+static constexpr u32 LZO_OUT_LEN = LZO_IN_LEN + (LZO_IN_LEN / 16) + 64 + 3;
```
leoetlino, line 17: nor here

leoetlino (approved):

PR #9764, opened 2021-05-29
VideoCommon: Fix bounding box on AMD/OpenGL/Windows

Pokechu22:
At least on my (old) AMD device, bounding box does not work on OpenGL; using an `int4` works, while using 4 ints (either as an array or 4 separate fields) does not work.  The Metal shader compiler does not work with `int4`, though, so both variants need to be supported.

I don't know if this is an issue on newer AMD devices as well.

See also #9762.

### DIFF
--- a/Source/Core/VideoCommon/PixelShaderGen.cpp
+++ b/Source/Core/VideoCommon/PixelShaderGen.cpp
@@ -451,20 +451,37 @@ void WritePixelShaderCommonHeader(ShaderCode& out, APIType api_type,
 
   if (bounding_box)
   {
+    if (api_type == APIType::D3D)
+    {
+      out.Write("globallycoherent RWBuffer<int> bbox_data : register(u2);\n"
+                "#define atomicMin InterlockedMin\n"
+                "#define atomicMax InterlockedMax");
+    }
+    else
+    {
+      out.Write("SSBO_BINDING(0) buffer BBox {{\n");
+
+      if (DriverDetails::HasBug(DriverDetails::BUG_BROKEN_SSBO_FIELD_ATOMICS))
+      {
+        // AMD drivers on Windows seemingly ignore atomic writes to fields or array elements of an
+        // SSBO other than the first one, but using an int4 seems to work fine
+        out.Write("  int4 bbox_data;\n");
+      }
+      else
+      {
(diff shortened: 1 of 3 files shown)


Techjar:
I don't think this is worth the mess.

Source/Core/VideoCommon/PixelShaderGen.cpp
```diff
@@ -451,20 +451,35 @@ void WritePixelShaderCommonHeader(ShaderCode& out, APIType api_type,
...
+        out.Write("SSBO_BINDING(0) buffer BBox {{\n"
+                  "  int4 bbox_data;\n"
+                  "}};");
+      }
+      else
+      {
+        // The Metal shader compiler fails to compile the atomic instructions
+        // when operating on individual components of a vector
+        out.Write("SSBO_BINDING(0) buffer BBox {{\n"
+                  "  int bbox_data[4];\n"
```
Techjar, line 474: We don't actually know if this works on Metal either.

Source/Core/VideoCommon/PixelShaderGen.cpp
```diff
@@ -451,20 +451,35 @@ void WritePixelShaderCommonHeader(ShaderCode& out, APIType api_type,
 
   if (bounding_box)
   {
+    if (api_type == APIType::D3D)
+    {
+      out.Write("globallycoherent RWBuffer<int> bbox_data : register(u2);\n"
+                "#define atomicMin InterlockedMin\n"
+                "#define atomicMax InterlockedMax");
```
Techjar, line 458: Missing line break

JMC47:
Confirmed on discord that the bugs affect the latest drivers.

JMC47:
The situation is that a lot of our AMD users on Windows use OpenGL still.  Yes, AMD's Windows drivers are pretty slow for OpenGL, but it is still playable enough for a lot of users that they use it.  OpenGL also has our most accurate depth emulation which is sometimes necessary for games, but I'm not sure if that comes into play for Paper Mario.

I think we should try to avoid the crash.  This is a fairly innocuous hack that can be reverted if they fix the bug.  It doesn't affect performance on other drivers/devices.

phire (approved):

PR #9765, opened 2021-05-29
VideoCommon: Move bounding box pixel quads rounding to shader

Techjar:
This avoids rounding values that the game writes to the bounding box registers, especially the default values.

Fixes corruption/crash in Paper Mario TTYD
https://bugs.dolphin-emu.org/issues/12524

### DIFF
--- a/Source/Core/VideoCommon/RenderBase.cpp
+++ b/Source/Core/VideoCommon/RenderBase.cpp
@@ -185,21 +185,6 @@ void Renderer::ReinterpretPixelData(EFBReinterpretType convtype)
   g_framebuffer_manager->ReinterpretPixelData(convtype);
 }
 
-u16 Renderer::BBoxRead(int index)
-{
-  return BBoxReadImpl(index);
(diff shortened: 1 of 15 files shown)


JMC47:
@phire @stenzek can someone please review this?  It fixes a very critical crash in Paper Mario that causes save corruption.  This will block the progress report.

Source/Core/VideoCommon/RenderBase.cpp
```diff
@@ -187,18 +187,7 @@ void Renderer::ReinterpretPixelData(EFBReinterpretType convtype)
 
 u16 Renderer::BBoxRead(int index)
```
stenzek, line 188: Could these wrapper/impl functions be removed now?

Shader looks okay to me.

Techjar: I was thinking we'd just keep that around as a stepping stone for abstracting bbox more later.

PR #9766, opened 2021-05-29
OGL: Force memory barrier when reading back bounding box values

stenzek:
We also need to ensure the the CPU does not receive stale values which have been updated by the GPU. Apparently the buffer here is not coherent on NVIDIA drivers.

Not sure if this is a driver bug/spec violation or not, one would think that glGetBufferSubData() would invalidate any caches as needed, but this path is only used on NVIDIA anyway, so it's fine.

A point to note is that according to ARB_debug_report, it's moved from video to host memory, which would explain why it needs the cache invalidate.


Techjar (approved):
Fixes the flickering dining car window reflections in Paper Mario TTYD.

Techjar:
The incoherent access here is actually entirely within spec, and not a bug at all: https://www.khronos.org/opengl/wiki/Memory_Model#Incoherent_memory_access

PR #9767, opened 2021-05-29
JitArm64: Improve Arm64FPRCache::GetCallerSavedUsed

JosJuice:
If we're only using the lower 64 bits of a callee-saved register, GetCallerSavedUsed can return false for it.

### DIFF
--- a/Source/Core/Core/PowerPC/JitArm64/JitArm64_RegCache.cpp
+++ b/Source/Core/Core/PowerPC/JitArm64/JitArm64_RegCache.cpp
@@ -742,6 +742,17 @@ bool Arm64FPRCache::IsCalleeSaved(ARM64Reg reg) const
   return std::find(callee_regs.begin(), callee_regs.end(), reg) != callee_regs.end();
 }
 
+bool Arm64FPRCache::IsTopHalfUsed(ARM64Reg reg) const
+{
+  for (const OpArg& r : m_guest_registers)
+  {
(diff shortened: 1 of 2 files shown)


Source/Core/Core/PowerPC/JitArm64/JitArm64_RegCache.cpp
```diff
@@ -729,6 +729,17 @@ bool Arm64FPRCache::IsCalleeSaved(ARM64Reg reg) const
   return std::find(callee_regs.begin(), callee_regs.end(), reg) != callee_regs.end();
 }
 
+bool Arm64FPRCache::IsTopHalfUsed(ARM64Reg reg) const
+{
+  for (const OpArg& r : m_guest_registers)
+  {
+    if (r.GetReg() != ARM64Reg::INVALID_REG && DecodeReg(r.GetReg()) == DecodeReg(reg))
```
leoetlino, line 736: Could you explain how the DecodeReg check works? Not sure I understand how it lets you check whether only one half of the register is used.

JosJuice: DecodeReg strips away the upper bits of the ARM64Reg enum (which tell us whether the register is to be treated as wide, extra wide, single, double or quad), leaving us with a number from 0 to 31. The "half of the register" check has nothing to do with DecodeReg – it's handled by the line where we call GetType. (Actually, we might technically not need DecodeReg at all here, but I figured it would make sense to include it.)

PR #9768, opened 2021-05-29
JitArm64: Support keeping constants in FPR cache

JosJuice:
Sometimes, we need to put constants in float registers. If the same constant is used multiple times in a block, we only need to generate it once, not before every use like we do in master.

### DIFF
--- a/Source/Core/Core/PowerPC/JitArm64/JitArm64_RegCache.cpp
+++ b/Source/Core/Core/PowerPC/JitArm64/JitArm64_RegCache.cpp
@@ -5,6 +5,7 @@
 
 #include <algorithm>
 #include <cstddef>
+#include <cstring>
 #include <vector>
 
 #include "Common/Assert.h"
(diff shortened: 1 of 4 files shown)


Source/Core/Core/PowerPC/JitArm64/JitArm64_RegCache.h
```diff
@@ -323,6 +337,20 @@ class Arm64FPRCache : public Arm64RegCache
 
   void StoreRegisters(BitSet32 regs) { FlushRegisters(regs, false); }
 
+  Arm64Gen::ARM64Reg GetConstant(
+      u64 value, bool single, bool want_duplicated, bool generates_duplicated,
+      const std::function<void(Arm64Gen::ARM64Reg, Arm64Gen::ARM64FloatEmitter*)>& generate);
+
+  // Equivalent to calling GetConstant two times, except if both constants need to be generated,
+  // generate_both is used instead of generate1 and generate2.
+  std::pair<Arm64Gen::ARM64Reg, Arm64Gen::ARM64Reg> GetTwoConstants(
```
lioncash, line 346: Perhaps the function params should be assigned aliases to make this a little nicer to read. Currently this is quite noisy to visually scan through at once.

Source/Core/Core/PowerPC/JitArm64/JitArm64_RegCache.cpp
```diff
@@ -697,6 +718,15 @@ void Arm64FPRCache::GetAllocationOrder()
 
 void Arm64FPRCache::FlushByHost(ARM64Reg host_reg)
 {
+  for (HostReg& reg : m_host_registers)
```
lioncash, line 721: Perhaps a helper function should be introduced that operates like `find_if` for scanning host registers, given this loop pattern is used in multiple places

JosJuice: The specific pattern of looking for a host register for which `host_reg.GetReg()` is equal to a particular value only occurs in two places as far as I can tell. Would this helper function be just for those two cases, or more general? If it's the latter, I don't think a helper function would be much more convenient than calling `std::find_if` directly (you would just skip having to write `m_host_registers.begin(), m_host_registers.end()`).

merryhime:
I understand why there is duplication code but I feel this introduces more complexity than necessary.

JosJuice:
Closing in favor of PR #11178.

PR #9769, opened 2021-05-30
Fix -Wreorder warnings

Tilka:


leoetlino:
One of the commits in #9721 also contains the same changes but let's merge this first since that other PR still has unresolved review comments.

PR #9770, opened 2021-05-30
Jits: Fix accidentally setting GT in CR when clearing EQ

JosJuice:
Fixes https://bugs.dolphin-emu.org/issues/12526.


Tilka:
Do you mind writing a test for this? (+ some other conditions if convenient)

JosJuice:
I will do so.

JosJuice:
Here is the test: https://github.com/dolphin-emu/hwtests/pull/39

In master, the interpreter passes all subtests, but Jit64 has two problems: Issue 12526, and the SO bit being set in the same way as the LT bit instead of being fetched from XER. This PR fixes the former but not the latter.

PR #9771, opened 2021-05-30
DSU improvements

Filoppi:
- Reworked the DSU to never hang the Host thread for more than a really small time (e.g. when disabling DSU its thread now closes almost immediately)
- Fix DSU using a different UID for every controller of the same server (that's not how its supposed to work)
- Add battery level input
- Improve robustness when a large amount of devices are connected
- Make sure inputs start from their resting pose instead of 0
- Add devices disconnection detection, previously they'd stay there forever until manually refreshed

Deeply tested with one and multiple controllers.

More improvements will be coming after this, including touch pad calibration and touch pad relative input (mapped 1:1 to the window screen, for the wii mote cursor).

### DIFF
--- a/Source/Core/InputCommon/ControllerInterface/DualShockUDPClient/DualShockUDPClient.cpp
+++ b/Source/Core/InputCommon/ControllerInterface/DualShockUDPClient/DualShockUDPClient.cpp
@@ -114,7 +114,7 @@ private:
     {
       switch (m_battery)
       {
-      case BatteryState::Charging:
+      case BatteryState::Charging:  // We don't actually know the battery level in this case
       case BatteryState::Charged:
         return BATTERY_INPUT_MAX_VALUE;
       default:
@@ -138,6 +138,8 @@ public:
   std::optional<int> GetPreferredId() const final override;
 
 private:
+  void ResetPadData();
+
   const std::string m_name;
   const int m_index;
   sf::UdpSocket m_socket;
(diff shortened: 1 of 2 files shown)


JMC47:
Tested and confirmed to work well.

JMC47:
I've been using this in my local builds for a couple of days.  It's kind of hard to use DSU without it now that I'm used to it.

JMC47:
I looked through the code to the best of my ability.  It seems relatively simple and, as I've said, I've been using it for a while.  It's after the beta now so if a few regressions pop up, so be it.  There's only so many things I can test on it.

Techjar (approved):
Nothing immediately stands out to me as wrong in the code, and JMC has been using it without issues (improvements in fact), so LGTM.

midget35:
Thanks for your work on all this @Filoppi . I have been doing a lot of testing with multiplayer emulated wiimote input including DSU. This seems very promising.

A question regarding your plans to do 'touch pad calibration'... what exactly is this item? Don't know if related, but it would be terrific to be able to simulate wiimote forwards and backwards motion using the touchpad y axis.

Thanks again :)

Filoppi:
@midget35 thanks.
Touch pad calibration is for non DS4 devices using a touch screen/pad over DSU. There is no official rule for the min and max values for the touch pad, which on the DS4 is like 1920x1040, but if you use a Switch touchscreen over DSU, the max is different, so to map a wiimote cursor to the DSU device touchscreen, you need calibration. Yuzu has it.

Pretty sure you can already map a wiimote motion forward/backward axis to the touchpad now, it just won't be very accurate.

PR #9772, opened 2021-05-31
OGL: Disable bounding box emulation on AMD/Windows

Techjar:
AMD drivers on Windows only write to the first field of an SSBO binding in a shader when using atomics, which breaks bounding box emulation.


shuffle2:
Has this been reported to amd?

Techjar:
Closing in favor of #9764

PR #9773, opened 2021-05-31
VideoCommon: Add fallback handling for bounding box when disabled or unsupported

Techjar:
The SDK seems to write "default" bounding box values before every draw (1023 0 1023 0 are the only values encountered so far, which happen to be the extents allowed by the BP registers) to reset the registers for comparison in the pixel engine, and presumably to detect whether GX has updated the registers with real values. Handling these writes and returning them on read when bounding box emulation is disabled or unsupported, even without computing real values from rendering, seems to prevent games from corrupting memory or crashing.

This obviously does not fix any effects that rely on bounding box emulation, but having the game not clobber its own code/data or just outright crash is a definite improvement.

### DIFF
--- a/Source/Core/VideoCommon/RenderBase.h
+++ b/Source/Core/VideoCommon/RenderBase.h
@@ -394,6 +394,18 @@ private:
   u32 m_last_xfb_stride = 0;
   u32 m_last_xfb_height = 0;
 
+  // Nintendo's SDK seems to write "default" bounding box values before every draw (1023 0 1023 0
+  // are the only values encountered so far, which happen to be the extents allowed by the BP
+  // registers) to reset the registers for comparison in the pixel engine, and presumably to detect
+  // whether GX has updated the registers with real values.
+  //
+  // We can store these values when Bounding Box emulation is disabled and return them on read,
+  // which the game will interpret as "no pixels have been drawn"
+  //
+  // This produces much better results than just returning garbage, which can cause games like
+  // Ultimate Spider-Man to crash
+  std::array<u16, 4> m_bounding_box_fallback = {};
+
   // NOTE: The methods below are called on the framedumping thread.
   void FrameDumpThreadFunc();
   bool StartFrameDumpToFFMPEG(const FrameDump::FrameData&);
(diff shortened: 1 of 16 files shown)


JMC47:
This makes it so we can run Bounding Box titles without them crashing if Bounding Box is disabled, instead just breaking Visual Effects.  It's actually really useful for examining what they do.

JMC47 (approved):

phire (changes requested):
If these "default values" have been written by the sdk before this draw call, then we would have received them in SetBoundingBox.

Rather than hardcoded these "default values" if the desired result is to preserve what was written, then we should modify how BBdisabled works so it accepts writes and just never fetches new results from the host gpu.

Source/Core/VideoCommon/RenderBase.h
```diff
@@ -390,6 +394,10 @@ class Renderer
   u32 m_last_xfb_stride = 0;
   u32 m_last_xfb_height = 0;
 
+  // Provides a sink for bounding values reads and writes when the feature is disabled or
+  // unsupported, this way games shouldn't crash due to garbage values.
```
phire, line 398: Might be wise to expand upon this comment for the benefit of future developers.

Something like this?

```suggestion
  // Nintendo's SDK seems to write "default" nonsensical bounding box values before every
  // draw (1023 0 1023 0 are the only values encountered so far), which presumably it is
  // using to detect whether GX has updated the registers with real values.
  //
  // We can store these values when Bounding Box emulation is disabled and return them
  // on read, which the game will interpret as "no pixels have been drawn" 
  // 
  // This produces much better results than just returning garbage, which can cause games
  // like SpiderMan 2 to crash.  
```

phire (approved):

Source/Core/VideoCommon/RenderBase.h
```diff
@@ -390,6 +394,18 @@ class Renderer
   u32 m_last_xfb_stride = 0;
   u32 m_last_xfb_height = 0;
 
+  // Nintendo's SDK seems to write "default" bounding box values before every draw (1023 0 1023 0
```
JMC47, line 397: Missing ) at the end of the first line?

PR #9774, opened 2021-06-02
DSP: Fix cycles overhead

sepalani:
This PR fixes https://dolp.in/i12498. `ForceExceptionCheck` was increasing the current instruction's cycles too much making it way too slow.

I created this hardware test to measure the access timings of some instructions/registers used during Syobon Action boot :  
[wii-cycles-timing.zip](https://github.com/dolphin-emu/dolphin/files/6586410/wii-cycles-timing.zip)

<details><summary>HWTEST results on Wii</summary>
<p>

| BL/BLR   | Attempt 1       | Attempt 2       | Attempt 3       |
|----------|-----------------|-----------------|-----------------|
| BL/BLR   |       1/1       |       1/1       |       1/1       |

| SPRG     | Attempt 1       | Attempt 2       | Attempt 3       |
|----------|-----------------|-----------------|-----------------|
| SPRG0    |       3/3       |       3/3       |       3/3       |
| SPRG1    |       3/41      |       3/12      |       3/12      |
| SPRG2    |       3/3       |       3/3       |       3/3       |
| SPRG3    |       3/3       |       3/3       |       3/3       |

| GQR      | Attempt 1       | Attempt 2       | Attempt 3       |
|----------|-----------------|-----------------|-----------------|
| GQR0     |       5/3       |       5/3       |       5/3       |
| GQR1     |      44/3       |       5/3       |       5/3       |
| GQR2     |       5/43      |       5/3       |       5/3       |
| GQR3     |       5/3       |       5/3       |       5/3       |
| GQR4     |       5/3       |       5/3       |       5/3       |
| GQR5     |      43/3       |       5/3       |       5/3       |
| GQR6     |       5/43      |       5/3       |       5/3       |
| GQR7     |       5/3       |       5/3       |       5/3       |

| DSP REG  | Attempt 1       | Attempt 2       | Attempt 3       |
|----------|-----------------|-----------------|-----------------|
| 5        |      76/57      |      35/1       |      34/1       |
| 24       |      35/1       |      34/1       |      34/1       |
| 25       |      35/1       |      34/1       |      34/1       |
| 27       |      35/1       |      34/1       |      34/1       |

| AI REG   | Attempt 1       | Attempt 2       | Attempt 3       |
|----------|-----------------|-----------------|-----------------|
| 0        |     116/1       |     116/1       |     116/1       |
| 1        |     125/1       |     116/1       |     116/1       |
| 2        |     119/1       |     116/1       |     125/1       |
| 3        |     116/1       |     116/1       |     116/1       |

| PI REG   | Attempt 1       | Attempt 2       | Attempt 3       |
|----------|-----------------|-----------------|-----------------|
| 0        |      46/1       |      47/1       |      47/1       |
| 1        |      45/1       |      47/1       |      47/1       |

</p>
</details>

<details><summary>HWTEST results on Dolphin (custom build)</summary>
<p>

| BL/BLR   | Attempt 1       | Attempt 2       | Attempt 3       |
|----------|-----------------|-----------------|-----------------|
| BL/BLR   |       2/2       |       2/2       |       2/2       |

| SPRG     | Attempt 1       | Attempt 2       | Attempt 3       |
|----------|-----------------|-----------------|-----------------|
| SPRG0    |       3/4       |       3/4       |       3/4       |
| SPRG1    |       3/4       |       3/4       |       3/4       |
| SPRG2    |       3/4       |       3/4       |       3/4       |
| SPRG3    |       3/4       |       3/4       |       3/4       |

| GQR      | Attempt 1       | Attempt 2       | Attempt 3       |
|----------|-----------------|-----------------|-----------------|
| GQR0     |       3/4       |       3/4       |       3/4       |
| GQR1     |       3/4       |       3/4       |       3/4       |
| GQR2     |       3/4       |       3/4       |       3/4       |
| GQR3     |       3/4       |       3/4       |       3/4       |
| GQR4     |       3/4       |       3/4       |       3/4       |
| GQR5     |       3/4       |       3/4       |       3/4       |
| GQR6     |       3/4       |       3/4       |       3/4       |
| GQR7     |       3/4       |       3/4       |       3/4       |

| DSP REG  | Attempt 1       | Attempt 2       | Attempt 3       |
|----------|-----------------|-----------------|-----------------|
| 5        |      35/3       |      35/3       |      35/3       |
| 24       |      35/3       |      35/3       |      35/3       |
| 25       |      35/3       |      35/3       |      35/3       |
| 27       |      35/3       |      35/3       |      35/3       |

| AI REG   | Attempt 1       | Attempt 2       | Attempt 3       |
|----------|-----------------|-----------------|-----------------|
| 0        |     116/3       |     116/3       |     116/3       |
| 1        |     116/3       |     116/3       |     116/3       |
| 2        |     116/5593    |     116/5917    |     116/4117    |
| 3        |     116/517     |     116/11317   |     116/5593    |

| PI REG   | Attempt 1       | Attempt 2       | Attempt 3       |
|----------|-----------------|-----------------|-----------------|
| 0        |       3/3       |       3/3       |       3/3       |
| 1        |       3/3       |       3/3       |       3/3       |

</p>
</details>

I used a custom build based on [this branch](https://github.com/sepalani/dolphin/commit/c1b65471b74a52f3ca63e3ac74e2ec9266abee3c) which is really WIP but implements some kind of MMIO access with cycles adjustment based on real hardware timing. From this branch, I found that the cycles overhead is caused by the `ForceExceptionCheck` call in the `Core:Timing::ScheduleEvent` function. In this branch, the performance monitor was updated (without speed hack support) based on the downcount to see the overhead which was previously hidden by the hardcoded opinfo's cycles.

I'm open to suggestion regarding the way this fix should be implemented, especially if you've better alternative regarding the function name or its implementation.

### DIFF
--- a/Source/Core/Core/CoreTiming.cpp
+++ b/Source/Core/Core/CoreTiming.cpp
@@ -227,7 +227,8 @@ void ClearPendingEvents()
   s_event_queue.clear();
 }
 
-void ScheduleEvent(s64 cycles_into_future, EventType* event_type, u64 userdata, FromThread from)
+void ScheduleEvent(s64 cycles_into_future, EventType* event_type, u64 userdata, FromThread from,
+                   bool cycle_safe)
 {
   ASSERT_MSG(POWERPC, event_type, "Event type is nullptr, will crash now.");
 
@@ -249,7 +250,7 @@ void ScheduleEvent(s64 cycles_into_future, EventType* event_type, u64 userdata, 
     s64 timeout = GetTicks() + cycles_into_future;
 
     // If this event needs to be scheduled before the next advance(), force one early
-    if (!s_is_global_timer_sane)
+    if (!cycle_safe && !s_is_global_timer_sane)
       ForceExceptionCheck(cycles_into_future);
 
     s_event_queue.emplace_back(Event{timeout, s_event_fifo_id++, userdata, event_type});
@@ -268,6 +269,12 @@ void ScheduleEvent(s64 cycles_into_future, EventType* event_type, u64 userdata, 
     std::lock_guard lk(s_ts_write_lock);
     s_ts_queue.Push(Event{g.global_timer + cycles_into_future, 0, userdata, event_type});
   }
+}
+
+void ScheduleCycleSafeEvent(s64 cycles_into_future, EventType* event_type, u64 userdata,
+                            FromThread from)
+{
+  ScheduleEvent(cycles_into_future, event_type, userdata, from, true);
 }
 
 void RemoveEvent(EventType* event_type)
--- a/Source/Core/Core/CoreTiming.h
+++ b/Source/Core/Core/CoreTiming.h
@@ -70,7 +70,11 @@ enum class FromThread
 // is scheduled earlier than the current values (when scheduled from the CPU Thread only).
 // Scheduling from a callback will not update the downcount until the Advance() completes.
 void ScheduleEvent(s64 cycles_into_future, EventType* event_type, u64 userdata = 0,
-                   FromThread from = FromThread::CPU);
+                   FromThread from = FromThread::CPU, bool cycle_safe = false);
+
+// Same as ScheduleEvent but without the ForceExceptionCheck.
+void ScheduleCycleSafeEvent(s64 cycles_into_future, EventType* event_type, u64 userdata = 0,
+                            FromThread from = FromThread::CPU);
 
 // We only permit one event of each type in the queue at a time.
 void RemoveEvent(EventType* event_type);
--- a/Source/Core/Core/HW/AudioInterface.cpp
+++ b/Source/Core/Core/HW/AudioInterface.cpp
@@ -248,7 +248,7 @@ void RegisterMMIO(MMIO::Mapping* mmio, u32 base)
                    s_sample_counter = val;
                    s_last_cpu_time = CoreTiming::GetTicks();
                    CoreTiming::RemoveEvent(event_type_ai);
-                   CoreTiming::ScheduleEvent(GetAIPeriod(), event_type_ai);
+                   CoreTiming::ScheduleCycleSafeEvent(GetAIPeriod(), event_type_ai);
                  }));
 
   mmio->Register(base | AI_INTERRUPT_TIMING, MMIO::DirectRead<u32>(&s_interrupt_timing),
@@ -257,7 +257,7 @@ void RegisterMMIO(MMIO::Mapping* mmio, u32 base)
                                  PowerPC::ppcState.pc);
                    s_interrupt_timing = val;
                    CoreTiming::RemoveEvent(event_type_ai);
-                   CoreTiming::ScheduleEvent(GetAIPeriod(), event_type_ai);
+                   CoreTiming::ScheduleCycleSafeEvent(GetAIPeriod(), event_type_ai);
                  }));
 }
 
--- a/Source/Core/Core/HW/DSP.cpp
+++ b/Source/Core/Core/HW/DSP.cpp
@@ -398,7 +398,7 @@ void RegisterMMIO(MMIO::Mapping* mmio, u32 base)
           // TODO: need hardware tests for the timing of this interrupt.
           // Sky Crawlers crashes at boot if this is scheduled less than 87 cycles in the future.
           // Other Namco games crash too, see issue 9509. For now we will just push it to 200 cycles
-          CoreTiming::ScheduleEvent(200, s_et_GenerateDSPInterrupt, INT_AID);
+          CoreTiming::ScheduleCycleSafeEvent(200, s_et_GenerateDSPInterrupt, INT_AID);
         }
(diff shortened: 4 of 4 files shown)


JMC47:
I checked a couple other cycle timing sensitive games in this pull request, and it doesn't seem to affect anything else.  Doesn't affect slowdown in N64 VC, doesn't fix hangs in FMVs in demo discs, doesn't affect The Daring Game for Girls, Radiant Dawn, etc.

phire (changes requested):
Whoa... No. 

You haven't changed cycle timings at all, all you have done is scheduled an event and then told to maybe not service it until some unpredictable number of cycles later (fully dependant on what other events are happening).

If performance monitor is being updated incorrectly based on the forced exception check, (which doesn't effect actual cycles executed by dolphin, if you check timebase), then you should fix the performance monitor code, not add bugs over here.

phire:
I did warn you on IRC that the performance monitor cycles code in dolphin was buggy and incomplete. 

You should fix performance monitor before trying to fix any timing issues revealed by hwtests that read performance monitor.

sepalani:
Unless I'm totally off, the root cause of Syobon Action not booting is the fact it's stuck in some kind of interrupt/event happening too early preventing it to load its resources. I highly suspect [ForceExceptionCheck](https://github.com/dolphin-emu/dolphin/commit/2ebbfd6f856d7fad39676568529bc48f9369e103#diff-d2aeca6f94955b3772222bb77853f0dcc015b12148188cbe7e5771b42664fc7dR323) to trigger them too early. I'll try to have a MMIO timing branch ready and fix the PMC, especially for the JITs to see if it still happens. As far as my debugging went, it's spammed by these audio interrupts which prevents it to resume its normal execution.

> 
> 
> Whoa... No.
> 
> You haven't changed cycle timings at all, all you have done is scheduled an event and then told to maybe not service it until some unpredictable number of cycles later (fully dependant on what other events are happening).

My tests were run with the interpreter and the cycles were off there. If I understood the code properly, this PR should have fixed the cycles accuracy there, am I correct? Though, as you mentioned, it will delay the exception handling to a later point for the recompilers.

> 
> 
> I did warn you on IRC that the performance monitor cycles code in dolphin was buggy and incomplete.
> 
> You should fix performance monitor before trying to fix any timing issues revealed by hwtests that read performance monitor.

Yeah, that's why I used this [custom build](https://github.com/sepalani/dolphin/commit/c1b65471b74a52f3ca63e3ac74e2ec9266abee3c#diff-11cc45baa12b37c7e2e7630504d59d9b1619b12c84daa010d411fd44d1a65265R224) and the interpreter mode, which should be good for testing unless I missed something (like speedhack support which I didn't use). As you said, running this hardware test has to be done using the interpreter mode, otherwise it shows zeroes all over the place which is wrong.

@phire 
Assuming that my timing fix is sheer luck, shouldn't ForceExceptionCheck not be done if I disabled IRQ? It's another issue that I can easily fix and test.

phire:
> shouldn't ForceExceptionCheck not be done if I disabled IRQ? 

`ForceExceptionCheck` is perhaps a bad name for that function. It does both IRQs and dolphin's internal events. 

In the case of the call from `ScheduleEvent`, it's only really dealing with dolphin's internal events. 

-----

**Some background:**

`CoreTiming::Advance()` is the core of the Timing system. It checks the event queue, executes any events that are meant to be executed before or at the current cycle. It also checks for External Exceptions.

To improve performance, `CoreTiming::Advance()` is called when there is meant to be an event to execute, or an External exception to trigger an interrupt. The previous execution of `CoreTiming::Advance()` will calculate how many cycles until the next scheduled event that it knows about and sets downcounts to that value. 

A JIT slice will execute for at least `downcounts` cycles before return to Advance. Because the JIT can only end on block boundaries, it might overshoot quite a bit. 

But sometimes an MMIO access generates new events that are scheduled to execute before the current JIT slice is scheduled to end. Or an exception. In this case, `downcounts` was wrong. So `ForceExceptionCheck` actually re-calculates `downcounts` and updates it so the current slice will end early and exit. 

In the case of IRQs, it's always called with zero cycles, AKA exit as soon as possible, at the end of the current JIT block. 

But because we are dealing with events here, it can be scheduled much further into the future, or not at all if the current slice is already ending before that event.

--------------------------

> the root cause of Syobon Action not booting is the fact it's stuck in some kind of interrupt/event happening too early preventing it to load its resources. I highly suspect ForceExceptionCheck to trigger them too early

Only because `ForceExceptionCheck` has been told by someone else to trigger too early. Disabling `ForceExcpetionCheck` is kind of a brute force approach which delays ALL events all over dolphin, causing massive regressions to timings. 

What you need to do is find the exact call to `ScheduleEvent()` that has a `cycles_into_future` that is way too close. Change that and only that.

sepalani:
This PR isn't the right way to address the issue.

I'll try to find other ways (if any), based on what's already known: https://dolp.in/i12498#note-6.

PR #9775, opened 2021-06-02
GameINI: Fix GFXBackend crash

dbrookman:
Fixes a regression from [5.0-12066](https://dolphin-emu.org/download/dev/a660033e8c394634a22c02a031969685fcdb4fba/), where setting the GFXBackend variable to a backend other than the current global one would crash Dolphin upon launching that game.

Fixes [https://bugs.dolphin-emu.org/issues/12529](https://bugs.dolphin-emu.org/issues/12529)

### DIFF
--- a/Source/Core/Core/Core.cpp
+++ b/Source/Core/Core/Core.cpp
@@ -242,6 +242,9 @@ bool Init(std::unique_ptr<BootParameters> boot, const WindowSystemInfo& wsi)
 
   Host_UpdateMainFrame();  // Disable any menus or buttons at boot
 
+  // Manually reactivate the video backend in case a GameINI overrides the video backend setting.
+  VideoBackendBase::PopulateBackendInfo();
+
   // Issue any API calls which must occur on the main thread for the graphics backend.
   WindowSystemInfo prepared_wsi(wsi);
   g_video_backend->PrepareWindow(prepared_wsi);


Source/Core/Core/BootManager.cpp
```diff
@@ -279,6 +279,7 @@ bool BootCore(std::unique_ptr<BootParameters> boot, const WindowSystemInfo& wsi)
       config_cache.bSetVolume = true;
     dsp_section->Get("EnableJIT", &StartUp.m_DSPEnableJIT, StartUp.m_DSPEnableJIT);
     dsp_section->Get("Backend", &StartUp.sBackend, StartUp.sBackend);
+    VideoBackendBase::ActivateBackend(Config::Get(Config::MAIN_GFX_BACKEND));
```
leoetlino, line 282: The root issue is that the new graphics backend gets initialized too late after the WindowSystemInfo has been prepared for the *original* video backend (and not for the new backend).

https://github.com/dolphin-emu/dolphin/blob/4cd1567308ab921d0e37a83672ad91fa4932097f/Source/Core/Core/Core.cpp#L246-L253

Could you move this line to right before the above code snippet (before "// Issue any API calls...")? This will make it clearer that the manual ActivateBackend isn't redundant.

leoetlino (changes requested):

Source/Core/Core/BootManager.cpp
dbrookman: Just pushed with these changes!

Source/Core/Core/Core.cpp
```diff
@@ -242,6 +242,9 @@ bool Init(std::unique_ptr<BootParameters> boot, const WindowSystemInfo& wsi)
 
   Host_UpdateMainFrame();  // Disable any menus or buttons at boot
 
+  // Manually reactivate the video backend just in case a GameINI overides the video backend setting.
```
leoetlino, line 245:
```suggestion
  // Manually reactivate the video backend just in case a GameINI overrides the video backend setting.
```

leoetlino (commented):
LGTM after the typo is fixed and the commits are squashed together

Source/Core/Core/Core.cpp
dbrookman: Good eye on that typo, thanks!

leoetlino (approved):

PR #9776, opened 2021-06-03
CoreTiming: ForceExceptionCheck only when exception is enabled

sepalani:
An alternative to https://github.com/dolphin-emu/dolphin/pull/9774. It still fixes Syobon Action.

Ready to reviewed & tested.

**EDIT**: I had to enable exception manually for the CoreTimingTest. I'm open to suggestions and alternatives regarding that aspect such as:
 - Are there events where it doesn't apply?
 - A better way to fix the timing test?


phire (changes requested):
No.

See https://github.com/dolphin-emu/dolphin/pull/9774#issuecomment-853855506 for more details.

PR #9778, opened 2021-06-03
msvc: remove workarounds for old compiler issues

shuffle2:


Source/Core/Common/BitField.h
```diff
@@ -134,18 +134,13 @@ struct BitField
   // so that we can use this within unions
   constexpr BitField() = default;
 
-// Visual Studio (as of VS2017) considers BitField to not be trivially
-// copyable if we delete this copy assignment operator.
-// https://developercommunity.visualstudio.com/content/problem/101208/c-compiler-is-overly-strict-regarding-whether-a-cl.html
-#ifndef _MSC_VER
```
Dentomologist, line 140: According to the last post in that link this fix wasn't available in a release version until last week.  #9777 bumps the required visual studio version, but it looks like this PR depends on that too.

shuffle2: yes

Tilka: That's merged now. Please rebase.

PR #9779, opened 2021-06-04
Freelook: enhance 2d camera.  Add ability to remove bloom, stretch or smush HUD and more!

iwubcode:
This builds on top of #9652

It allows direct movement of individual 2d images using the texture-ids that texture pack owners know and love.

This allows the HUD to be moved offscreen for pretty screenshots or maybe just reorganize things to give more screenspace:

GIF 6-4-2021 12-53-04 AM

You can also stretch the HUD specifically so ultra wide monitors or triple monitor users have a more natural feel (HUD in center monitor).  Use the 3d fov hack for an even nicer experience!

image
(credit to @DrBoiffard for the image)

But wait...there's more!  Depending on the game, you might also be able to target the efb textures.  In order to make it easier for users to do that, I changed Dolphin's "dump EFB" to output the size and the texture format.

What does that allow you to do?  You can move bloom or other effects off-screen, without requiring any gecko codes.

RMCE01_2021-06-04_00-25-30
(lots of brightness)

RMCE01_2021-06-04_00-25-19
(goodbye!)

See [bloom removal album](https://imgur.com/a/5tRWtKl) for other games tested.

Unfortunately, some games will still be stubborn.  This will not work on games that use 3d HUDs.  It will also not work if your game is trying to draw multiple things in one call and you only want to move _one_ of them.

PR #9780, opened 2021-06-04
Common/FileUtil: Add some missing configs to D_CONFIG_IDX

Techjar:
This is needed otherwise the config won't be saved (except GCKEYBOARD which seems to be handled elsewhere, but it doesn't hurt to add it here anyways). Only affected Linux, somehow Windows saves just fine. Must be a different code path somewhere.


JMC47:
@iwubcode please review.

iwubcode:
Thanks @Techjar !  This looks ok to me.  I'm guessing windows is using the "USER" path:  https://github.com/dolphin-emu/dolphin/pull/9780/files#diff-25b1e5a1a176c6bfa1e888cde0940d6df39945077b6bd47b5890787f2456e19eL924  .  But that's just a guess. I can't recall how that's working.

Note you can't actually override freelook settings in user config  without an additional change.  I  made that in #9779 but it probably should be pulled out (see [this commit](https://github.com/dolphin-emu/dolphin/pull/9779/commits/fbcf914a1b72cb59d29cc1387a146bc1060381f4)) since that review is quite large.

JosJuice:
Could you add `F_DUALSHOCKUDPCLIENTCONFIG_IDX` too? LGTM otherwise.

JosJuice (approved):

PR #9781, opened 2021-06-05
VideoCommon: Round bounding box coordinates down and remove pixel center offset

Techjar:
Fragment coordinates always have a 0.5 offset from a whole integer, as that's where the pixel center is on modern GPUs. Therefore, we want to always round the fragment coordinates down for bounding box calculations. This also renders the pixel center offset useless, as 0.5 vs ~0.5833333 makes no difference when rounding down.

Fixes softlocks in certain cutscenes of Paper Mario: TTYD and Super Paper Mario, caused (presumably) by rounding up to a coordinate outside the render area.

### DIFF
--- a/Source/Core/VideoCommon/PixelShaderGen.cpp
+++ b/Source/Core/VideoCommon/PixelShaderGen.cpp
@@ -495,19 +495,9 @@ void UpdateBoundingBoxBuffer(int2 min_pos, int2 max_pos) {{
 }}
 
 void UpdateBoundingBox(float2 rawpos) {{
-  // The pixel center in the GameCube GPU is 7/12, not 0.5 (see VertexShaderGen.cpp)
-  // Adjust for this by unapplying the offset we added in the vertex shader.
-  const float PIXEL_CENTER_OFFSET = 7.0 / 12.0 - 0.5;
-  float2 offset = float2(PIXEL_CENTER_OFFSET, -PIXEL_CENTER_OFFSET);
-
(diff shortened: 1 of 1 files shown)


JMC47:
For OpenGL/Vulkan, this causes artifacts in Disney's Magical Mirror, Disney's Hide and Sneak, and whatever the third one I have is that's in Japanese that I don't feel like looking up. 

Super Paper Mario, Paper Mario TTYD hangs are fixed.  Ultimate Spider-Man is still fine.  Speed Challenge hangs before I can test Bounding Box.

JMC47:
I'm merging this because it literally can't make things worse at this point.

Source/Core/VideoCommon/PixelShaderGen.cpp
```diff
@@ -495,19 +495,9 @@ void UpdateBoundingBoxBuffer(int2 min_pos, int2 max_pos) {{
...
-
-#ifdef API_OPENGL
-  // OpenGL lower-left origin means that Y goes in the opposite direction.
-  offset.y = -offset.y;
-#endif
-
   // The rightmost shaded pixel is not included in the right bounding box register,
   // such that width = right - left + 1. This has been verified on hardware.
-  int2 pos = iround(rawpos * cefbscale + offset);
+  int2 pos = int2(rawpos * cefbscale);
```
blaahaj, line 500: Will this produce correct result for cefbscale > 1?

PR #9782, opened 2021-06-05
VideoCommon: Perform OpenGL bounding box inversion in pixel shader

Techjar:
Running the min/max operation on the upside down, quad-rounded pixel coordinates before inverting them to the standard upper-left origin produces wrong results. Therefore, we need to do the inversion before rounding to pixel quads.

I have included results from a hardware test to confirm that this is correct. The hardware can be downloaded here: https://qimg.techjargaming.com/f/Tzx3VrV2.7z

Hardware

Before Change

After Change


JMC47:
Code seems much cleaner and it fixes the issues.

stenzek:
LGTM

PR #9783, opened 2021-06-05
Vulkan: Disable subgroup reduction on NVIDIA

Techjar:
Subgroup reduction on NVIDIA drivers produces wrong results in some cases. This causes bounding box values to be off by several pixels.

OpenGL actually uses NV_thread_shuffle to implement a "subgroup reduction" so this bug check is not applicable there, and the extension is assumed and known to work without issues.

Here's a hardware test to demonstrate the wrong results and the fix: https://qimg.techjargaming.com/f/Tzx3VrV2.7z

Hardware

Vulkan (Broken)

Vulkan (Fixed)


Techjar:
User reports have confirmed that whatever is going on here also affects AMD, so it may not actually be a driver problem. I'm making this a draft while we investigate further.

PR #9784, opened 2021-06-05
Update buffer_storage warning message

PEmu2:
The current warning is confusing to users as performance may be fine despite the warning. Many users also do not understand the device vendor sentence and the attempt at shaming vendors into supporting the feature clearly hasn't worked at this point so let's just get rid of it.

### DIFF
--- a/Source/Core/VideoBackends/OGL/OGLRender.cpp
+++ b/Source/Core/VideoBackends/OGL/OGLRender.cpp
@@ -744,8 +744,7 @@ Renderer::Renderer(std::unique_ptr<GLContext> main_gl_context, float backbuffer_
     OSD::AddMessage(fmt::format("Your OpenGL driver does not support {}_buffer_storage.",
                                 m_main_gl_context->IsGLES() ? "EXT" : "ARB"),
                     60000);
-    OSD::AddMessage("This device's performance will be terrible.", 60000);
-    OSD::AddMessage("Please ask your device vendor for an updated OpenGL driver.", 60000);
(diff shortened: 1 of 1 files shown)


JosJuice:
Actually, the attempt at shaming vendors seems to have worked out for all vendors except Apple. Any decent Android phone you buy nowadays has this extension. But either way, I agree that there isn't that much point in having the line about vendors anymore.

JMC47:
Should we just remove the warning all together?  Some devices are strong enough that they can still get decent performance without buffer storage.  Sure, they lose out on a huge optimization, but they can make it up elsewhere.

Rumi-Larry:
> 
> 
> Should we just remove the warning all together? Some devices are strong enough that they can still get decent performance without buffer storage. Sure, they lose out on a huge optimization, but they can make it up elsewhere.

I don't believe that's the point. If Buffer Storage is in the mind of Dolphin users, then at least some of them may have some influence over Android's development.

mbc07:
I'm fine with the message rewording but still think the warning should stay, as it provides a far quicker way of checking if buffer_storage is supported (just launch any title with the OpenGL backend), compared to going to the device's app store to download another app to check the supported driver extensions...

JMC47:
Should we at least say 

EXT: Buffer Storage is missing.  Your device may have poor performance.

I think it's a bit weird to just say your performance will be poor without saying why.

JosJuice:
@dolphin-emu-bot rebuild

JosJuice (approved):

leoetlino (approved):

PR #9789, opened 2021-06-06
NetPlay: Sync the EFB access tile size as an integer instead of a boolean.

AdmiralCurtiss:
Fixes severe slowdown in affected titles (such as Rocket Power: Beach Bandits) when played in NetPlay.


JosJuice (approved):

leoetlino (approved):

Techjar:
Heh, that'd be my fault. Dunno how nobody noticed for 2 years though...

PR #9790, opened 2021-06-06
CheatsManager: Avoid recreating child widgets on every OnStateChanged(), and take running game info directly from SConfig.

AdmiralCurtiss:
While this does not fix the core issue, this does eliminate a big part of the lag you get after holding the Frame Advance button for a while -- enough for it to become unnoticeable on my system, but I suspect weaker systems might still run into it, though to a lesser degree.


leoetlino (approved):
CheatsManager was going through every game in the game list? o.O

PR #9793, opened 2021-06-07
MMIOHandlers: Move method definitions to MMIO.cpp

sepalani:
This PR moves MMIOHandlers method definitions to MMIO.cpp.

It prevents some compilation issues when editing the header file.


leoetlino:
JMC47 tested this and found no perf impact.

PR #9794, opened 2021-06-07
Fix input focus checks being wrong (issue 12540)

Filoppi:
When rendering to main and going full screen, we aren't using the main window handle as the code assumed, but the other, detached, render widget.

Fixes issue:
https://bugs.dolphin-emu.org/issues/12540
caused by https://github.com/dolphin-emu/dolphin/pull/9697


Dentomologist (approved):
Tested and LGTM.

PR #9795, opened 2021-06-07
Fix controllers after the first one not defaulting to the default device

Filoppi:
Also fixes this bug: At the moment, Qt does not support a controller not having any default device, so it will still show the first device from the devices list as default, but internally it won't actually be selected, so input detection won't work as expected, until the user manually changes the device from the drop down list.

Also added a check to avoid potential crashes if we had no controllers. That function is only called on the first start up so there is no need to set the default device on the controller unless the default device string is valid. There is also no need to update references as all bindings would be empty. 

This will work even better when other parts of my input improvements from https://github.com/dolphin-emu/dolphin/pull/9489 are merged.

### DIFF
--- a/Source/Core/InputCommon/InputConfig.cpp
+++ b/Source/Core/InputCommon/InputConfig.cpp
@@ -145,8 +145,23 @@ bool InputConfig::LoadConfig(bool isGC)
   }
   else
   {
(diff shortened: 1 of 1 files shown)


iwubcode (approved):
Code wise LGTM

PR #9796, opened 2021-06-07
Jit64: Don't use software FMA after all

JosJuice:
When writing 2c38d64, I was assuming that either the performance impact of software FMA wouldn't be too large, or CPUs that were too old to have FMA instructions were too slow to run Dolphin well anyway. This was wrong. To give an example, the netplay performance went from 60 FPS to 30 FPS in one case.

We have three options for how to get the performance back:

1. A simple revert - never FMA on x64, always FMA on AArch64
2. Never use FMA anywhere
3. Use FMA only on CPUs with hardware FMA

Option 1 is not appealing at all to me – I don't think maintaining the exact same results between all x64 CPUs is more important than maintaining the exact same results between most x64 CPUs and all AArch64 CPUs. AArch64 is only going to get more important from here out.

Option 2 is a possibility, but it really feels like being constrained by the past, and would require writing a bunch of annoying code for AArch64 just for the sake of making things less accurate. I was wondering if option 3 could be a valid option, so I tried implementing it in this commit. Considering that things seemed to sync okay between x64 and AArch64 before 2c38d64, and that replay physics are not affected, maybe the very slight differences in rounding from not using FMA aren't that much of a problem in practice. But this will need testing.

### DIFF
--- a/Source/Core/Core/PowerPC/Jit64/Jit_FloatingPoint.cpp
+++ b/Source/Core/Core/PowerPC/Jit64/Jit_FloatingPoint.cpp
@@ -241,10 +241,21 @@ void Jit64::fmaddXX(UGeckoInstruction inst)
   JITDISABLE(bJITFloatingPointOff);
   FALLBACK_IF(inst.Rc);
 
-  // While we don't know if any games are actually affected (replays seem to work with all the usual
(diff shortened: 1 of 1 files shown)


JosJuice:
This desyncs. I guess I will have to look into option 2 then. (Maybe with some way of detecting if all players have FMA support...?)

PR #9797, opened 2021-06-07
4.0 hotfixes

SeroQ0:
Good day, ever since the 5.0-14337 of the Dev dolphin version, I've got a steady dropping in frames (to around 30) with the Game, " Naruto SCON4" on the game cube. This occurs whenever I enter online play. Changing the buffer does nothing to fix the issue. My PC ram is not stressed at all and my processor is going easy. Once again, playing the game normally, it runs just fine, but going online (regardless of the ping) sees a massive drop in frames, and it's one steady too so it's not pulsing like those experienced in heat throttling.  I hope I posted this in the right place.


JosJuice:
This is not an issue tracker. Please use https://bugs.dolphin-emu.org/projects/emulator or the forums.

Or, as it happens... I actually created PR #9796 very recently, which is intended to (hopefully) address this. If you have further comments, I suppose leaving them on that PR is fine.

JosJuice:
Er, wait. Did you say it happens in **5.0-14337** and later, not 5.0-14348 like I got a report about elsewhere?

PR #9800, opened 2021-06-08
Vulkan: Fix subgroup reduction

Techjar:
It seems that we were writing values from helper invocations, which produces wrong results.

Here's a hardware test to demonstrate the wrong results and the fix: https://qimg.techjargaming.com/f/Tzx3VrV2.7z

Hardware

Vulkan (Broken)

Vulkan (Fixed)


stenzek:
Nice catch, LGTM

PR #9801, opened 2021-06-08
VideoCommon: Only include centered pixels in bounding box

Techjar:
At higher resolutions, our bounding box dimensions end up being slightly larger than original hardware in some cases. This is not necessarily wrong, it's just an artifact of rendering at a higher resolution, due to bringing out detail that wouldn't have appeared on original hardware. It causes a texel to fall partially on what would have been a single pixel at native resolution, resulting in the coordinates getting bumped up to the next valid value. In many cases, these slightly larger bounding boxes are perfectly fine, as games don't hard-code expected dimensions. It is problematic in Paper Mario TTYD though, for a somewhat complicated reason.

Paper Mario TTYD frequently uses EFB copies to pre-render a bunch of animation frames for a character sprite (especially in Chapter 2), so that it can then render 100 or more of them without bringing the GameCube to its knees. Based on my observation, the game seems to set aside a region of memory to store these EFB copies. This region is obviously fairly small, as the GameCube only has 24MB of RAM. There are 2 rooms in Chapter 2 where you fight a horde of as many as 100 Jabbies, which are also rendered using EFB copies, so in this room the game ends up making 130(!) EFB copies just for Puni and Jabbi sprites. This seems to nearly fill the region of memory it set aside for them. Unfortunately, our slightly larger bounding boxes at higher resolutions results in overflowing this memory, causing very strange behavior. Some EFB copies partially overlap game state, resulting in reading it as a garbage RGB5A3 texture that constantly changes. Others apparently somehow trigger a corner case in our persistent buffer mapping, causing them to partially overwrite earlier EFB copies.

What this change does is only include the screen coordinates that align with the equivalent native resolution pixel centers, which generally results in the bounding boxes being more in line with original hardware. It isn't perfect, but it's enough to fix Paper Mario TTYD's Jabbi rooms by avoiding the buffer overflow. Notably, it is more accurate at odd resolutions than at even resolutions. Native resolution is completely unaffected by this change, as should be the case. This change may also have a small positive impact on shader performance at higher resolutions, as there will be less atomic operations performed.

Here's some diffs comparing to master at the same resolution, as well as high resolution vs native resolution on this PR (native resolution is unchanged from master, as mentioned).
[Master vs This PR @ 4x Native](https://gist.github.com/Techjar/39f800e8cc4e34c75b2053e61ce61669/revisions#diff-702fdcbece63cbb1f9816d9511c65867a274bc527e38477ec2fa04359902b64a)
[Master vs This PR @ 5x Native](https://gist.github.com/Techjar/c29f0fd50c3164928f735ac95457693f/revisions#diff-07af3015a87bbd2b03f7cf6aeae965d2965579b0ce3238ba98e3a90874af68db)
[1x Native vs 4x Native](https://gist.github.com/Techjar/713b4830803de0afd35a50e3b177bd5d/revisions#diff-09f8e1d54ba0aff82e8240c003e8fd8f6e546df52e3258acc81997f2b1cc7d6b)
[1x Native vs 5x Native](https://gist.github.com/Techjar/f928d063a1fb666e339d5c775de41894/revisions#diff-7a67ea5e2624ef62466d9bedba31f33e8fa17d1167d0fc0cdb6fdd16fcb4d81d)

### DIFF
--- a/Source/Core/VideoCommon/PixelShaderGen.cpp
+++ b/Source/Core/VideoCommon/PixelShaderGen.cpp
@@ -495,9 +495,18 @@ void UpdateBoundingBoxBuffer(int2 min_pos, int2 max_pos) {{
 }}
 
 void UpdateBoundingBox(float2 rawpos) {{
+  // We only want to include coordinates for pixels aligned with the native resolution pixel centers.
+  // This makes bounding box sizes more accurate (though not perfect) at higher resolutions,
+  // avoiding EFB copy buffer overflow in affected games.
+  //
+  // For a more detailed explanation, see https://dolp.in/pr9801
+  int2 int_efb_scale = iround(1 / {efb_scale}.xy);
+  if (any(int2(rawpos) % int_efb_scale != int_efb_scale >> 1))  // divide by two
+    return;
+
   // The rightmost shaded pixel is not included in the right bounding box register,
   // such that width = right - left + 1. This has been verified on hardware.
-  int2 pos = int2(rawpos * cefbscale);
+  int2 pos = int2(rawpos * {efb_scale}.xy);
 
 #ifdef API_OPENGL
   // We need to invert the Y coordinate due to OpenGL's lower-left origin
@@ -506,8 +515,8 @@ void UpdateBoundingBox(float2 rawpos) {{
 
   // The GC/Wii GPU rasterizes in 2x2 pixel groups, so bounding box values will be rounded to the
   // extents of these groups, rather than the exact pixel.
-  int2 pos_tl = pos & ~1;
-  int2 pos_br = pos | 1;
+  int2 pos_tl = pos & ~1;  // round down to even
+  int2 pos_br = pos | 1;   // round up to odd
 
 #ifdef SUPPORTS_SUBGROUP_REDUCTION
   if (CAN_USE_SUBGROUP_REDUCTION) {{
@@ -526,7 +535,7 @@ void UpdateBoundingBox(float2 rawpos) {{
 }}
 
 )",
-              fmt::arg("efb_height", EFB_HEIGHT));
+              fmt::arg("efb_height", EFB_HEIGHT), fmt::arg("efb_scale", I_EFBSCALE));
   }
 }
 


JMC47:
Current status:  Disney Hide and Sneak + Disney Magical Mirror have no crashes nor visual artifacts in any backend + any resolution.

Still need to test the other games.

Source/Core/VideoCommon/PixelShaderGen.cpp
```diff
@@ -495,19 +495,33 @@ void UpdateBoundingBoxBuffer(int2 min_pos, int2 max_pos) {{
 }}
 
 void UpdateBoundingBox(float2 rawpos) {{
-  // The rightmost shaded pixel is not included in the right bounding box register,
-  // such that width = right - left + 1. This has been verified on hardware.
-  int2 pos = int2(rawpos * cefbscale);
+  // This is kind of a hack. This is used to bias the coordinates "inwards" to produce
+  // more accurate (or at least less broken) results at higher resolutions. We also subtract
+  // an epsilon to avoid floating point precision error.
```
JMC47, line 500: Should it be said that being slightly smaller is better than being slightly oversized with examples of issues?

JMC47 (approved):

Source/Core/VideoCommon/PixelShaderGen.cpp
Techjar: I don't wanna make the comment too massive, but I can try to briefly explain I suppose.

JMC47: I just think having an example for future devs who may not be familiar with this can know why the decisions were made.

a simple "see: dolp.in/i/xxxx for info"

phire (changes requested):
As discussed on IRC, I think this is the wrong approach.

Instead, I think we should adjust the pixel center correction code in VertexShaderGen to make sure that one of the pixels in the higher IR grid (2x2 for 2xIR, 3x3 for 3xIR etc) lines up with the pixel centre of the original 1xIR grid. Ideally this will be the center pixel, but for even IRs it will have to be off-center. 

Then you can simply adjust the pixel shader bounding box code it only updates the atomics on pixel shader invocations that on 1xIR grid. 

-----------

Rather than being a hack, this should result in almost identical results to running at 1xIR

Techjar:
@phire Thanks for the help. I think I've come up with an implementation of what you suggested, and it's producing significantly more accurate results than the previous attempt. I also looked at the pixel center correction, and it already seems to be as accurate as we can make it for higher-than-native resolutions, so there wasn't any changes to be made there.

phire:
Ok, the new version of this PR is much better IMO. 

Doesn't produce perfect results, that might require playing with the vertex shader code to get the pixel centres aligned correctly, but I think this is good enough for now.

phire (approved):
Approved dependant on testing from JMC47

PR #9803, opened 2021-06-09
VideoCommon: Abstract bounding box

Techjar:
This moves much of the duplicated bounding box code into VideoCommon, leaving only the specific buffer implementations in each backend.

I also accidentally fixed the bug where bounding box doesn't work on software renderer unless you select some other backend first.


Source/Core/VideoBackends/Software/SWBoundingBox.cpp
```diff
@@ -0,0 +1,72 @@
...
+  {
+    values[i] = BBoxManager::GetCoordinate(static_cast<BBoxManager::Coordinate>(index + i));
+  }
+
+  return values;
+}
+
+void SWBoundingBox::Write(int index, const std::vector<BBoxType> values)
+{
+  for (int i = 0; i < values.size(); i++)
```
lioncash, line 66:
```suggestion
  for (size_t i = 0; i < values.size(); i++)
```

Otherwise this is a sign mismatch warning in the comparison.

Source/Core/DolphinLib.props
```diff
@@ -531,7 +531,7 @@
     <ClInclude Include="VideoBackends\D3DCommon\D3DCommon.h" />
     <ClInclude Include="VideoBackends\D3DCommon\Shader.h" />
     <ClInclude Include="VideoBackends\D3DCommon\SwapChain.h" />
-    <ClInclude Include="VideoBackends\Null\NullPerfQuery.h" />
```
PatrickFerry, line 534: Is removing NullPerfQuery.h a mistake?

Techjar: Nope, it's a non-existent file. Must've been leftover accidentally from some past refactor.

Source/Core/VideoCommon/BoundingBox.h
```diff
@@ -4,40 +4,44 @@
...
 
-// Sets a particular coordinate for the bounding box.
-void SetCoordinate(Coordinate coordinate, u16 value);
+  virtual bool Initialize() = 0;
 
-// Updates all bounding box coordinates.
-void Update(u16 left, u16 right, u16 top, u16 bottom);
+protected:
+  virtual std::vector<BBoxType> Read(int index, int length) = 0;
+  virtual void Write(int index, const std::vector<BBoxType> values) = 0;
```
lioncash, line 37:
```suggestion
  virtual void Write(int index, const std::vector<BBoxType>& values) = 0;
```

Perhaps leave a general TODO or Note comment that this vector can likely be turned into a `std::span` whenever we update to C++20.

Source/Core/VideoCommon/BoundingBox.h
```diff
@@ -4,40 +4,44 @@
...
 
-// Save state
-void DoState(PointerWrap& p);
-}  // namespace BoundingBox
+private:
+  void Readback();
+
+  bool m_is_active;
```
lioncash, line 42: This should be initialized to false or true (whichever is more appropriate).

Otherwise, it's potentially possible to perform an uninitialized read by creating a bounding box instance and immediately calling `IsEnabled()`.

Source/Core/VideoCommon/BoundingBox.h
```diff
@@ -4,40 +4,44 @@
...
 
 class PointerWrap;
 
-// Bounding Box manager
-namespace BoundingBox
-{
-// Indicates a coordinate of the bounding box.
-enum class Coordinate
+using BBoxType = s32;
+static constexpr u32 NUM_BBOX_VALUES = 4;
```
lioncash, line 15:
```suggestion
constexpr u32 NUM_BBOX_VALUES = 4;
```

given constexpr internally links by default for values at file/namespace scope.

Source/Core/VideoCommon/BoundingBox.cpp
```diff
@@ -4,73 +4,109 @@
...
 }
 
-u16 GetCoordinate(Coordinate coordinate)
+void BoundingBox::Readback()
 {
-  return s_coordinates[static_cast<u32>(coordinate)];
+  auto read_values = Read(0, NUM_BBOX_VALUES);
+
+  // Preserve dirty values, that way we don't need to sync.
+  for (int i = 0; i < NUM_BBOX_VALUES; i++)
```
lioncash, line 60:
```suggestion
  for (u32 i = 0; i < NUM_BBOX_VALUES; i++)
```

Otherwise there's a sign-mismatch in the loop conditional.

Source/Core/VideoCommon/BoundingBox.cpp
```diff
@@ -4,73 +4,109 @@
...
+  // Games only ever seem to write all 4 values at once anyways.
+  for (int start = 0; start < NUM_BBOX_VALUES;)
+  {
+    if (!m_dirty[start])
+    {
+      start++;
+      continue;
+    }
+
+    int end = start + 1;
```
lioncash, line 41:
```suggestion
    u32 end = start + 1;
```
Ditto

Source/Core/VideoCommon/BoundingBox.cpp
```diff
@@ -4,73 +4,109 @@
...
 {
-  return s_is_active;
+  m_is_valid = false;
+
+  if (std::none_of(m_dirty.begin(), m_dirty.end(), [](bool dirty) { return dirty; }))
+    return;
+
+  // TODO: Does this make any difference over just writing all the values?
+  // Games only ever seem to write all 4 values at once anyways.
+  for (int start = 0; start < NUM_BBOX_VALUES;)
```
lioncash, line 33:
```suggestion
  for (u32 start = 0; start < NUM_BBOX_VALUES;)
```

Ditto

Source/Core/VideoBackends/D3D/D3DRender.cpp
```diff
@@ -264,19 +264,9 @@ void Renderer::UnbindTexture(const AbstractTexture* texture)
     D3D::stateman->ApplyTextures();
 }
 
-u16 Renderer::BBoxReadImpl(int index)
+std::unique_ptr<BoundingBox> CreateBoundingBox()
```
leoetlino, line 267: Shouldn't this be Renderer::CreateBoundingBox?

Techjar: Heh

Source/Core/VideoCommon/BoundingBox.h
Techjar: I thought `bool` was default-initialized to `false`.

JosJuice: No, if you default-initialize a primitive type, no value is set.

iwubcode:
This looks really good @Techjar , thanks for doing this!!

One silly comment, thoughts on renaming `VideoCommon/BoundingBox` to `VideoCommon/AbstractBoundingBox` to match the other classes?

Source/Core/VideoBackends/D3D/D3DBoundingBox.cpp
```diff
@@ -14,45 +14,39 @@
...
-  return s_bbox_uav.Get();
+  m_uav.Reset();
+  m_staging_buffer.Reset();
+  m_buffer.Reset();
 }
 
-void BBox::Init()
+bool D3DBoundingBox::Initialize()
 {
   if (!g_ActiveConfig.backend_info.bSupportsBBox)
```
iwubcode, line 26: This check happens in every backend (save for D3D12, Software, and Null).  Should it be moved to the abstract interface?  Vulkan supports a log message which might be nice in general.

Additionally, why is this returning `true`?  If it's not supported, seems like we would want to error out or possibly return a `NullBoundingBox` to avoid errors when it is being used.

iwubcode: Read/Write also check this same config option (but not in all backends).

Techjar: Returning `false` would cause us to abort backend initialization, which we don't want because games can still run without bounding box support.

iwubcode: Ah ok.  Well, I'd let me amend my suggestion then to modifying Read in `AbstractBoundingBox` to:

```
std::vector<BBoxType> Read(int index, int length)
{
  if (!g_ActiveConfig.backend_info.bSupportsBBox)
    return {};

  return ReadImpl(index, length);
}
```

with a

```
private:
virtual std::vector<BBoxType> ReadImpl(int index, int length) = 0;
```

and same with `Write()` so that all backends get this behavior in a single place.

And I'd add that check to D3D12's `Initialize()` for consistency with D3D11, OpenGL, and Vulkan.

Source/Core/VideoBackends/Null/NullBoundingBox.h
```diff
@@ -0,0 +1,23 @@
...
+
+namespace Null
+{
+class NullBoundingBox final : public BoundingBox
+{
+public:
+  bool Initialize() override { return true; }
+
+protected:
+  std::vector<BBoxType> Read(int index, int length) { return std::vector<BBoxType>(length); }
```
iwubcode, line 19: `override` for these?

Source/Core/VideoBackends/D3D/D3DBoundingBox.cpp
Techjar: I already addressed it in a slightly different way, but thanks. Didn't want to create forwarding functions, as the implementation here is pretty tight anyways.

D3D12 implicitly supports bounding box, so the check would be kinda pointless there. In any case I moved the initialization check to RenderBase.

Source/Core/VideoBackends/Null/NullBoundingBox.h
Techjar: Oh yea

Techjar:
> One silly comment, thoughts on renaming `VideoCommon/BoundingBox` to `VideoCommon/AbstractBoundingBox` to match the other classes?

I dunno. The naming of other things isn't entirely consistent either, something being AbstractX and others being XBase.

Rumi-Larry:
> > One silly comment, thoughts on renaming `VideoCommon/BoundingBox` to `VideoCommon/AbstractBoundingBox` to match the other classes?
> 
> I dunno. The naming of other things isn't entirely consistent either, something being AbstractX and others being XBase.

My personal preference would be XBase being consistently used. Any piece of code is already an abstraction :P

iwubcode:
> Any piece of code is already an abstraction

That's...not what this is referring to at all, see https://www.tutorialspoint.com/cplusplus/cpp_interfaces.htm

> I dunno. The naming of other things isn't entirely consistent either, something being AbstractX and others being XBase.

Well "Base" does not technically imply "Abstract" but I do understand what you're saying.  Honestly, not a big deal to leave it either, the backend specific abbreviations are prepended onto the derived classes, so the distinction is there.

JMC47:
Is there anything left for this?

iwubcode (approved):
LGTM.  Untested

Techjar:
> Shouldn't this default to {0x80, 0xA0, ...}?

Is this the default value of the registers on hardware?

leoetlino (approved):
> Reviewed 4 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/dolphin-emu/dolphin/9803)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Techjar)

PR #9804, opened 2021-06-09
NetPlay/Jit64: Avoid using software FMA

JosJuice:
When I added the software FMA path in 2c38d64 and made us use it when determinism is enabled, I was assuming that either the performance impact of software FMA wouldn't be too large or CPUs that were too old to have FMA instructions were too slow to run Dolphin well anyway. This was wrong. To give an example, the netplay performance went from 60 FPS to 30 FPS in one case.

This change makes netplay clients negotiate whether FMA should be used. If all clients use an x64 CPU that supports FMA, or AArch64, then FMA is enabled, and otherwise FMA is disabled. In other words, we sacrifice accuracy if needed to avoid massive slowdown, but not otherwise. When not using netplay, whether to enable FMA is simply based on whether the host CPU supports it.

The only remaining case where the software FMA path gets used under normal circumstances is when an input recording is created on a CPU with FMA support and then played back on a CPU without. This is not an especially common scenario (though it can happen), and TASers are generally less picky about performance and more picky about accuracy than other users anyway.

With this change, FMA desyncs are avoided between AArch64 and modern x64 CPUs (unlike before 2c38d64), but we do get FMA desyncs between AArch64 and old x64 CPUs (like before 2c38d64). This desync can be avoided by adding a non-FMA path to JitArm64 as an option, which I will wait with for another pull request so that we can get the performance regression fixed as quickly as possible.

https://bugs.dolphin-emu.org/issues/12542

### DIFF
--- a/Source/Core/Core/Config/SessionSettings.cpp
+++ b/Source/Core/Core/Config/SessionSettings.cpp
@@ -1,0 +1,14 @@
+// Copyright 2021 Dolphin Emulator Project
+// Licensed under GPLv2+
+// Refer to the license.txt file included.
+
+#include "Core/Config/SessionSettings.h"
+
+#include "Common/CPUDetect.h"
+#include "Common/CommonTypes.h"
+#include "Common/Config/Config.h"
+
+namespace Config
+{
+const Info<bool> SESSION_USE_FMA{{System::Session, "Core", "UseFMA"}, CPUInfo().bFMA};
+}  // namespace Config
--- a/Source/Core/Core/ConfigLoaders/GameConfigLoader.cpp
+++ b/Source/Core/Core/ConfigLoaders/GameConfigLoader.cpp
@@ -250,6 +250,9 @@ private:
       if (location.section.empty() && location.key.empty())
         continue;
 
+      if (location.system == Config::System::Session)
+        continue;
+
       layer->Set(location, value.second);
     }
   }
@@ -272,7 +275,7 @@ void INIGameConfigLayerLoader::Save(Config::Layer* layer)
     const Config::Location& location = config.first;
     const std::optional<std::string>& value = config.second;
 
-    if (!IsSettingSaveable(location))
+    if (!IsSettingSaveable(location) || location.system == Config::System::Session)
       continue;
 
     const auto ini_location = GetINILocationFromConfig(location);
--- a/Source/Core/Core/NetPlayClient.cpp
+++ b/Source/Core/Core/NetPlayClient.cpp
@@ -34,8 +34,10 @@
 #include "Common/StringUtil.h"
 #include "Common/Timer.h"
 #include "Common/Version.h"
+
 #include "Core/ActionReplay.h"
 #include "Core/Config/NetplaySettings.h"
+#include "Core/Config/SessionSettings.h"
 #include "Core/ConfigManager.h"
 #include "Core/GeckoCode.h"
 #include "Core/HW/EXI/EXI_DeviceIPL.h"
@@ -54,6 +56,7 @@
 #include "Core/Movie.h"
 #include "Core/PowerPC/PowerPC.h"
 #include "Core/SyncIdentifier.h"
+
 #include "InputCommon/ControllerEmu/ControlGroup/Attachments.h"
 #include "InputCommon/GCAdapter.h"
 #include "InputCommon/InputConfig.h"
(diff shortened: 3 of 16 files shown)


Source/Core/Core/Config/SessionSettings.cpp
```diff
@@ -0,0 +1,14 @@
...
+
+#include "Core/Config/SessionSettings.h"
+
+#include "Common/CPUDetect.h"
+#include "Common/CommonTypes.h"
+#include "Common/Config/Config.h"
+
+namespace Config
+{
+const Info<bool> SESSION_USE_FMA{{System::Session, "Core", "UseFMA"}, cpu_info.bFMA};
```
Techjar, line 13: `cpu_info` isn't properly initialized at this point, is it?

Source/Core/Core/NetPlayClient.cpp
```diff
@@ -609,10 +612,11 @@ unsigned int NetPlayClient::OnData(sf::Packet& packet)
     game_status_packet << static_cast<u32>(result);
     Send(game_status_packet);
 
-    sf::Packet ipl_status_packet;
-    ipl_status_packet << static_cast<MessageId>(NP_MSG_IPL_STATUS);
-    ipl_status_packet << ExpansionInterface::CEXIIPL::HasIPLDump();
-    Send(ipl_status_packet);
+    sf::Packet client_settings_packet;
```
Techjar, line 615: `client_caps_packet` would make more sense.

Source/Core/Core/Config/SessionSettings.cpp
JosJuice: Ah, now that I'm thinking about this a second time, I don't think this is actually safe :(

`cpu_info` does get initialized during dynamic initialization... But so does this, and the order isn't well-defined. That's annoying. I'm not sure what the best way to design this would be then...

Techjar: Would setting NetSettings/DTM field directly from `cpu_info` not work?

JosJuice: That does work, though I wonder if it's going to make things messy for us in the future when we try to migrate the movie config handling code to always using the new config system instead of requiring a handwritten line of code for each setting. It would also get rid of the ability to forcibly enable FMA through command-line options in case someone ever wants that – not that that would be a super important feature.

For now, I've made this line create a new temporary instance of `CpuInfo`. Simple and safe, and I don't believe there's any meaningful performance impact.

Techjar: Yeah, I suppose that's a good enough solution for now.

Source/Core/Core/ConfigLoaders/GameConfigLoader.cpp
```diff
@@ -272,7 +275,7 @@ void INIGameConfigLayerLoader::Save(Config::Layer* layer)
     const Config::Location& location = config.first;
     const std::optional<std::string>& value = config.second;
 
-    if (!IsSettingSaveable(location))
+    if (!IsSettingSaveable(location) || location.system == Config::System::Session)
```
Techjar, line 278: Shouldn't we just modify `IsSettingSaveable` to return false for `location.system == Config::System::Session` instead?

Techjar (approved):
Haven't tested this yet, but code looks good. I don't have pre-Haswell hardware in service to properly test.

Source/Core/Core/ConfigLoaders/GameConfigLoader.cpp
JosJuice: `IsSettingSaveable` marks settings that are still handled by the old config system, and based on code comments, it seems like we intend to drop `IsSettingSaveable` when the old config system is dead. The `Session` system is already handled by the new config system, and we want to keep not saving in the future when the old config system is gone.

JosJuice:
I asked people in the netplay community that I got the performance regression report from to test it, and they said things work with good performance and without desyncs regardless of whether the players' CPUs support FMA. I've also verified on my own that when you run netplay with only CPUs that support FMA, you do take the hardware FMA code path.

Source/Core/Core/ConfigLoaders/GameConfigLoader.cpp
Techjar: Ah, yeah okay. I guess `Session` is where anything that we don't want to be persistent should go. I'm gonna make a follow-up PR that moves a few of such settings there.

BhaaLseN: It still feels a bit weird that `Config::System::Session` shows up scattered around everywhere. Isn't this what the layered configuration system was designed to solve in first place? Have a layer that is session-only, doesn't save, and has that single setting?

JosJuice: The layered config system supports session-only *values*, but not session-only *settings*. I want to make it impossible for this setting to get loaded from Dolphin.ini so that nobody overrides the default by accident.

leoetlino (approved):

PR #9806, opened 2021-06-10
Display cutout support

Jokkaj:
Added <item name="android:windowLayoutInDisplayCutoutMode">shortEdges</item> to the main activity

### DIFF
--- a/Source/Android/app/src/main/res/values/styles.xml
+++ b/Source/Android/app/src/main/res/values/styles.xml
@@ -9,6 +9,7 @@
         <!-- Darker variant for the status bar and contextual app bars -->
         <item name="colorPrimaryDark">@color/dolphin_blue_dark</item>
         <item name="colorAccent">@color/dolphin_purple</item>
+        <item name="android:windowLayoutInDisplayCutoutMode">shortEdges</item>
 
         <!-- Enable window content transitions -->
         <item name="android:windowContentTransitions">true</item>


JosJuice:
Doesn't this have the potential of obscuring UI elements, like the app's version number?

Jokkaj:
> 
> 
> Doesn't this have the potential of obscuring UI elements, like the app's version number?

nope, the app version numbers (or other interface elements) are not impacted
https://imgur.com/a/Zor2ijm
it does improve immersion in-game

helps prevent black bar burn-in in the display for users who have an OLED and greatly improves immersion at the cost of a small cutout in the display, most modern smartphones now only have a small hole for the camera so the gameplay is not impacted.

JosJuice:
Not even if you have a phone that has the camera cutout in the corner and you use the main activity in landscape mode?

Also wait, does that mean this change affects the emulation activity as well? That's **definitely** going to cause problems unless you adapt the in-game menu to stay out of the way of the camera cutout.

Jokkaj:
> 
> 
> Not even if you have a phone that has the camera cutout in the corner and you use the main activity in landscape mode?
> 
> Also wait, does that mean this change affects the emulation activity as well? That's **definitely** going to cause problems with the in-game menu unless you adapt the in-game menu to stay out of the way of the camera cutout.

https://imgur.com/a/BjekgFf
as seen in the photos, there is a lot of wasted space with a single camera hole.
single-hole cameras also have little impact on the menu, although there could be problems with larger notches (even it would be enough to rotate the phone in reverse to not have the menu on the notch side)

JosJuice:
Yes, I see why you would like to be able to use the extra part of the screen for gameplay. I'm not opposed to that in itself, but I am opposed to merging this without adapting the in-game menu. Forcing people to rotate the phone isn't an especially good tradeoff.

JosJuice:
OSD messages would need to be moved out of the way of the cutout as well. `SetObscuredPixelsLeft` can be used for this, but you'd also need to add something similar for the FPS counter for the case where the cutout is on the right side of the screen.

DatEmulationGuy:
Do phones like Poco F2 Pro without a notch for the camera have this black bar in the notification area in Dolphin? Or it's using the whole screen?

I have a Poco x3 Pro with a small dot for the camera and get the black bar but in the Android settings it's possible to add any app to use this notch area and since my notch is very small it's not very distracting. It would be more convenient if there was an option in Dolphin to use the notch, I guess for people who don't about the Android settings, but if it's causing problems then RIP.

Too bad not many powerful phones have pop up or under display cameras to make the whole screen available for gaming and other stuff, maybe this will change in the future since I see them making the notches as small as possible or using new techniques like under display cameras.

Jokkaj:
> Do phones like Poco F2 Pro without a notch for the camera have this black bar in Dolphin? Or it's using the whole screen?
> 
> I have a Poco x3 Pro with a small dot for the camera and get the black bar but in the Android settings it's possible to add any app to use this notch area and since my notch is very small it's not very distracting. It would be more convenient if there was an option in Dolphin to use the notch, I guess for people who don't about the Android settings, but if it's causing problems then RIP.
> 
> Too bad not many powerful phones have pop up or under display cameras to make the whole screen available for gaming and other stuff, maybe this will change in the future since I see them making the notches as small as possible or using new techniques like under display cameras.

You can use your cutout area because the poco X3 has MIUI (wich is a custom skin for android) and MIUI adds the option to ignore the cutout, in stock android or other custom skins you just don't have that option.

blakbin:
Oh i get it. First i dont know whats is talking about ask to support cutout, since im MiUI user toremove that bar using screen ratio to strech in dolphin and make  the dolphin to always show notch and use fullscreen in my MiUI setting

Rumi-Larry:
> OSD messages would need to be moved out of the way of the cutout as well. `SetObscuredPixelsLeft` can be used for this, but you'd also need to add something similar for the FPS counter for the case where the cutout is on the right side of the screen.

Presumably that would be taking care of by the OS.

JosJuice:
It isn't. OSD messages and the FPS counter are custom UI elements drawn by Dolphin into a framebuffer. What the OS does do by default is prevent that framebuffer from overlapping the cutout by default, but getting rid of that behavior is exactly what this pull request does.

Rumi-Larry:
> It isn't. OSD messages and the FPS counter are custom UI elements drawn by Dolphin into a framebuffer. What the OS does do by default is prevent that framebuffer from overlapping the cutout by default, but getting rid of that behavior is exactly what this pull request does.

I see, well Android does allow to "crop" so that the image doesn't overlap the cutout, but that would necesitate to add the option in the UI. It would be somewhat annoying, but I think it's a good compromise.

JosJuice:
I don't see why a compromise would be needed, though. Implementing it properly isn't much more difficult than adding an option.

DatEmulationGuy:
Shouldn't it be an option? Some people might not want to always have this enabled by default

JosJuice:
Ah, because they want to make sure that none of the game image is hidden? Yeah, I suppose you're right. Though, it only matters if you're using Stretch to Window or if you're watching Rogue Squadron's super wide cutscenes.

PR #9807, opened 2021-06-11
VideoCommon: Define scalar any() and all() functions in GLSL

Techjar:
This fixes bounding box shaders failing to compile under Vulkan, due to differences between GLSL and HLSL in the return value of vector comparisons and what types these functions accept. I included all() for the sake of completeness.

### DIFF
--- a/Source/Core/VideoCommon/PixelShaderGen.cpp
+++ b/Source/Core/VideoCommon/PixelShaderGen.cpp
@@ -372,6 +372,16 @@ void WritePixelShaderCommonHeader(ShaderCode& out, APIType api_type,
             "int3 iround(float3 x) {{ return int3(round(x)); }}\n"
             "int4 iround(float4 x) {{ return int4(round(x)); }}\n\n");
 
+  // GLSL's any() and all() only accept vector types, while HLSL's also accept scalar types. We're
+  // adding these for convenience because while vector comparisons return a bool scalar in GLSL,
+  // allowing the results to be used directly in an if statement, they return a bool vector in HLSL,
+  // necessitating the use of any() or all() to reduce it to a scalar.
+  if (api_type == APIType::OpenGL || api_type == APIType::Vulkan)
+  {
+    out.Write("bool any(bool b) {{ return b; }}\n"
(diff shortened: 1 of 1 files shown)


Source/Core/VideoCommon/PixelShaderGen.cpp
```diff
@@ -372,6 +372,10 @@ void WritePixelShaderCommonHeader(ShaderCode& out, APIType api_type,
             "int3 iround(float3 x) {{ return int3(round(x)); }}\n"
             "int4 iround(float4 x) {{ return int4(round(x)); }}\n\n");
 
+  // GLSL's any() only accepts vector types, not scalar types
+  if (api_type == APIType::OpenGL || api_type == APIType::Vulkan)
+    out.Write("bool any(bool b) {{ return b; }}\n");
+
   if (api_type == APIType::OpenGL || api_type == APIType::Vulkan)
```
Rumi-Larry, line 379: These two "if" checks are identical. Is it intentional?

Techjar: The second one has to do with sampler bindings, so I didn't feel it was appropriate to shove this in there.

Pokechu22: Note that there is another identical if check below that handles uniform bindings, so it's consistent with the existing pattern.

Source/Core/VideoCommon/PixelShaderGen.cpp
```diff
@@ -372,6 +372,13 @@ void WritePixelShaderCommonHeader(ShaderCode& out, APIType api_type,
             "int3 iround(float3 x) {{ return int3(round(x)); }}\n"
             "int4 iround(float4 x) {{ return int4(round(x)); }}\n\n");
 
+  // GLSL's any() and all() only accept vector types, not scalar types
```
Pokechu22, line 375: Perhaps saying "while HLSL's also accept scalar types" would make it more obvious that this is for compatibility.

leoetlino (approved):

PR #9809, opened 2021-06-13
Move some NetPlay-specific settings to SessionSettings

Techjar:
This moves some settings that are only used during a NetPlay session to the Session config, so there's no possibility of them being permanently set by accident.

### DIFF
--- a/Source/Core/Core/Config/SessionSettings.h
+++ b/Source/Core/Core/Config/SessionSettings.h
@@ -10,4 +10,8 @@
 namespace Config
 {
 extern const Info<bool> SESSION_USE_FMA;
+extern const Info<bool> SESSION_LOAD_IPL_DUMP;
+extern const Info<bool> SESSION_GCI_FOLDER_CURRENT_GAME_ONLY;
+extern const Info<bool> SESSION_CODE_SYNC_OVERRIDE;
+extern const Info<bool> SESSION_WII_SD_CARD_WRITABLE;
 }  // namespace Config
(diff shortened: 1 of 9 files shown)


JosJuice:
`MAIN_WII_SD_CARD_WRITABLE` could be useful outside of netplay, so I don't think we should move that one over. Arguably the same applies to `MAIN_LOAD_IPL_DUMP`, but I'm less sure. Other than that, I agree with the changes.

Source/Core/Core/Config/SessionSettings.h
```diff
@@ -10,4 +10,8 @@
 namespace Config
 {
 extern const Info<bool> SESSION_USE_FMA;
+extern const Info<bool> SESSION_LOAD_IPL_DUMP;
+extern const Info<bool> SESSION_GCI_FOLDER_CURRENT_GAME_ONLY;
```
JosJuice, line 14: This one probably doesn't need "current game only" in the name anymore.

Techjar: I think you've misunderstood its purpose.

JosJuice: Yes, I have.

Techjar:
`MAIN_WII_SD_CARD_WRITABLE` is actually kind of a duplicate now, but I think we can address that when we get rid of the silly setting structs.

JosJuice:
Oh, with `MAIN_ALLOW_SD_WRITES`. Yeah, makes sense then.

JosJuice (approved):

leoetlino (approved):

PR #9811, opened 2021-06-13
Fix FPRF handling of denormal singles

JosJuice:
We were classifying denormal singles as if they were normals (because we converted singles to doubles and then classified them using the same code as for doubles), but this behavior has been confirmed incorrect by https://github.com/dolphin-emu/hwtests/pull/41.

When fixing this, I went with an approach of performing classification on 32-bit floats in all CPU emulation cores (i.e. classification is performed after forcing the result to single but before converting it back to double). It would be possible to perform classification after converting to double, but that would require us to check if the exponent is in a certain range instead of just checking if it is zero, so I think the approach I went with is better.

### DIFF
--- a/Source/Core/Core/PowerPC/Jit64Common/EmuCodeBlock.cpp
+++ b/Source/Core/Core/PowerPC/Jit64Common/EmuCodeBlock.cpp
@@ -727,34 +727,6 @@ void EmuCodeBlock::JitClearCA()
   MOV(8, PPCSTATE(xer_ca), Imm8(0));
 }
 
-void EmuCodeBlock::ForceSinglePrecision(X64Reg output, const OpArg& input, bool packed,
-                                        bool duplicate)
-{
-  // Most games don't need these. Zelda requires it though - some platforms get stuck without them.
-  if (m_jit.jo.accurateSinglePrecision)
-  {
-    if (packed)
-    {
-      CVTPD2PS(output, input);
-      CVTPS2PD(output, R(output));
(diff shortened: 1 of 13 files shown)


Source/Core/Core/PowerPC/Jit64Common/EmuCodeBlock.cpp
```diff
@@ -971,37 +954,55 @@ void EmuCodeBlock::SetFPRF(Gen::X64Reg xmm)
   else
   {
     MOVQ_xmm(R(RSCRATCH), xmm);
-    TEST(64, R(RSCRATCH), MConst(psDoubleExp));
+    if (single)
+      TEST(32, R(RSCRATCH), MConst(psFloatExp));
```
Sintendo, line 958: We can use `Imm32` for the constant rather than loading it from the pool in the `single` case.

JosJuice: Thanks! I'm not all that used to x64 yet :)

This is only possible for the non-SSE4.1 path, right?

Sintendo: No worries, you're doing just fine. 👍

Yeah, SSE instructions don't take full 32-bit immediates, so loading arbitrary constants from memory is the way the go.

Tilka (approved):

phire (approved):

PR #9813, opened 2021-06-13
AudioInterface: Initialize s_cpu_cycles_per_sample to a sane value.

AdmiralCurtiss:
I'm not sure what the previous default was, it doesn't seem to be anything sensible. I can trace it back to literally the first commit in the git but I haven't found an explanation for it.

Re-fixes the Naruto games after #9788.


leoetlino (approved):
Ooh, good catch... So the previous change revealed a latent bug, great :)

DatEmulationGuy:
Eternal Darkness game also fixed with this, was broken from 14420

PR #9815, opened 2021-06-14
JitArm64: Implement FMA-less path for FMA instructions

JosJuice:
For determinism compatibility with old x64 CPUs. Off by default.

I haven't tested this on netplay, but I've tested that the new code path doesn't break games in any obvious way. (Other than breaking existing F-Zero GX replays, which is unsurprising)

This is a follow-up to PR #9804.


merryhime (approved):
Code LGTM

PR #9817, opened 2021-06-15
Jit64: subfic - Optimize constants for d != a

Sintendo:
These optimizations were already present, but only when d == a. They also make sense when this condition does not hold.

---

<details><summary>imm == 0</summary>

Before:
```
41 BB 00 00 00 00    mov         r11d,0
45 2B DF             sub         r11d,r15d
```

After:
```
45 8B DF             mov         r11d,r15d
41 F7 DB             neg         r11d
```
</details>

<details><summary>imm == -1</summary>

Before:
```
41 BD FF FF FF FF    mov         r13d,0FFFFFFFFh
44 2B EE             sub         r13d,esi
0F 93 45 68          setae       byte ptr [rbp+68h]
```

After:
```
44 8B EE             mov         r13d,esi
41 F7 D5             not         r13d
C6 45 68 01          mov         byte ptr [rbp+68h],1
```
</details>

### DIFF
--- a/Source/Core/Core/PowerPC/Jit64/Jit_Integer.cpp
+++ b/Source/Core/Core/PowerPC/Jit64/Jit_Integer.cpp
@@ -898,28 +898,31 @@ void Jit64::subfic(UGeckoInstruction inst)
   RCX64Reg Rd = gpr.Bind(d, RCMode::Write);
   RegCache::Realize(Ra, Rd);
 
-  if (d == a)
+  if (imm == 0)
   {
-    if (imm == 0)
-    {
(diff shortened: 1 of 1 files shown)


degasus (approved):
Good catch, however I wonder if x64 has an 8bit sign extended immediate mov instruction. This should lead to the same code size and maybe with a lower latency.

Sintendo:
Unfortunately, there is no such instruction. In fact, of all the x64 'core' instructions, MOV and TEST are the only ones to lack a compact encoding for 8-bit signed immediates.

PR #9820, opened 2021-06-17
JitArm64: Simplify addex/subfex

JosJuice:
Some of the code used when the carry flag is known to be a constant value is really not much better than just setting the carry flag and then using the normal code, and with how rarely this code runs, it isn't well tested either. Might as well get rid of some of this code and simplify things.


degasus (approved):
Indeed, addex with carry=const_true was terrible.

PR #9822, opened 2021-06-17
JitArm64: Implement ps_cmpXX

JosJuice:


Source/Core/Core/PowerPC/JitArm64/JitArm64_FloatingPoint.cpp
```diff
@@ -405,12 +405,32 @@ void JitArm64::fcmpX(UGeckoInstruction inst)
...
 
+  ARM64Reg V0Q = ARM64Reg::INVALID_REG;
+  ARM64Reg V1Q = ARM64Reg::INVALID_REG;
+  if (upper_a)
+  {
+    V0Q = fpr.GetReg();
+    m_float_emit.DUP(singles ? 32 : 64, reg_encoder(V0Q), VA, 1);
+    VA = reg_encoder(V0Q);
+  }
+  if (upper_b)
```
degasus, line 416: We have a `a == b` optimization below. I doubt that it will be called often, but we could easily skip the DUP call here.

degasus (approved):
Feel free to merge with or without the `a == b` optimization.

Source/Core/Core/PowerPC/JitArm64/JitArm64_FloatingPoint.cpp
JosJuice: I don't see how we can skip the DUP call. We still need to perform the comparison, in case we have a NaN.

JosJuice: Oh wait, you mean we can skip *one* DUP call. Yeah, that makes sense.

JosJuice: Implemented.

PR #9824, opened 2021-06-18
RenderWidget: Set Window to Top for Cursor Lock

SirMangler:
Simple solution to the task bar being in the way when using the lock mouse cursor feature by setting the window to be topmost while focused. This is particularly annoying for users with the setting `Auto Hide Task Bar` enabled, because the mouse is able to enter the region for the taskbar to show up above the dolphin window. 

This calls `SetWindowPos` with the `NOMOVE`, `NOSIZE` and `NOACTIVATE` flags, the last flag is important to prevent a loop with popup boxes fighting for focus.

### DIFF
--- a/Source/Core/DolphinQt/RenderWidget.cpp
+++ b/Source/Core/DolphinQt/RenderWidget.cpp
@@ -254,6 +254,10 @@ void RenderWidget::SetCursorLocked(bool locked, bool follow_aspect_ratio)
   if (locked)
   {
 #ifdef _WIN32
+    // This will prevent the mouse from interacting with the task bar
+    // while in windowed / borderless fullscreen.
+    SetWindowPos((HWND) winId(), HWND_TOPMOST, 0, 0, 0, 0, SWP_NOMOVE | SWP_NOSIZE | SWP_NOACTIVATE);
+
     RECT rect;
     rect.left = render_rect.left();
     rect.right = render_rect.right();
@@ -280,6 +284,7 @@ void RenderWidget::SetCursorLocked(bool locked, bool follow_aspect_ratio)
   else
   {
 #ifdef _WIN32
+    SetWindowPos((HWND) winId(), HWND_NOTOPMOST, 0, 0, 0, 0, SWP_NOMOVE | SWP_NOSIZE | SWP_NOACTIVATE);
     ClipCursor(nullptr);
 #endif
 


Source/Core/DolphinQt/RenderWidget.cpp
```diff
@@ -254,6 +254,10 @@ void RenderWidget::SetCursorLocked(bool locked, bool follow_aspect_ratio)
   if (locked)
   {
 #ifdef _WIN32
+    // This will prevent the mouse from interacting with the task bar
+    // while in windowed / borderless fullscreen.
+    SetWindowPos((HWND) winId(), HWND_TOPMOST, 0, 0, 0, 0, SWP_NOMOVE | SWP_NOSIZE | SWP_NOACTIVATE);
```
leoetlino, line 259:
```suggestion
    SetWindowPos(static_cast<HWND>(winId()), HWND_TOPMOST, 0, 0, 0, 0, SWP_NOMOVE | SWP_NOSIZE | SWP_NOACTIVATE);
```
(or reinterpret_cast if the static_cast doesn't work -- either way C++ style casts are preferred)

Source/Core/DolphinQt/RenderWidget.cpp
```diff
@@ -280,6 +284,7 @@ void RenderWidget::SetCursorLocked(bool locked, bool follow_aspect_ratio)
   else
   {
 #ifdef _WIN32
+    SetWindowPos((HWND) winId(), HWND_NOTOPMOST, 0, 0, 0, 0, SWP_NOMOVE | SWP_NOSIZE | SWP_NOACTIVATE);
```
leoetlino, line 287: ditto

leoetlino (changes requested):

Source/Core/DolphinQt/RenderWidget.cpp
Filoppi, line 259: You should probably do this within the `if (ClipCursor(&rect))` condition, as if ClipCursor fails, you don't want to set the window to the topmost. Even if ClipCursor might never fail, it would still be better I'd say.

Filoppi, line 259: Have you checked if no other code changes the HWND_TOPMOST on the render widget/window handle? We don't want any conflicts there.

Filoppi, line 259: Have you tested this with "Render To Main Window"? I don't know what happens if you set HWND_TOPMOST to a window handle that is part of another window.

Filoppi: And could you explain the exact benefit of NOACTIVATE?

SirMangler: > Have you checked if no other code changes the HWND_TOPMOST on the render widget/window handle? We don't want any conflicts there.

My additions of `SetWindowPos` are the only two in the project file (aka DolphinQt folder).

> And could you explain the exact benefit of NOACTIVATE?

per my original post, 
> the last flag is important to prevent a loop with popup boxes fighting for focus.

The default behaviour of this function is to omit an `activate` event to the window, which will focus the window if not already. This creates a deadlock with any popup dialogs, such as the "Are you sure you want to exit the emulation" pop up, which also wants focus. They will continue to fight over who has focus until Dolphin is forcefully closed. There isn't any reason to trigger an activate event, this function is just used to set the window Z position, so removing it would be ideal anyway.

SirMangler: Interestingly, using Dolphin's built in `Keep Window on Top` itself doesn't seem to work with "Render To Main Window" in the first place. As such, this implementation doesn't work with that option either. The fix should be simple though.

Filoppi: I still don't really understand what you mean, as there was no problem like that with the exit confirmation window in my tests, but anyway, this sounds like a general improvement, actually unrelated from mouse locking (which is an optional feature)? So should the SWP_NOACTIVATE only be here?

SirMangler: > 
> 
> I still don't really understand what you mean, as there was no problem like that with the exit confirmation window in my tests, but anyway, this sounds like a general improvement, actually unrelated from mouse locking (which is an optional feature)? So should the SWP_NOACTIVATE only be here?

This problem definitely exists, the only way I could close Dolphin was by literally *logging off*, because the focus stealing was so intense I couldn't even end the process in task manager, and the problem would happen every time until I added the `SWP_NOACTIVATE` flag. Regardless of if you were able to successfully recreate it, the problem can happen and therefore this change should exist to prevent it, especially since there are no downsides. 

The point is, there is no reason to activate the window here, regardless of if the problem does or doesn't exist. This is only to change the window's level, we do not need to simulate the same events that occur when a user adjusts the window.

Rumi-Larry: > 
> 
> Interestingly, using Dolphin's built in `Keep Window on Top` itself doesn't seem to work with "Render To Main Window" in the first place. As such, this implementation doesn't work with that option either. The fix should be simple though.

That's a good idea for a follow up pr.

Filoppi: Not really, this fix should work with all dolphin settings out of the box.

Filoppi: True, the SWP_NOACTIVATE here is fine, but are you saying that the "focus stealing" but between the render and exit widgets already existed before this? Could you provide a video for that?

If you are interested, I think there is also a but that happens when going full screen, like about 1 time out of 3, full-screen doesn't get the keyboard focus and pressing exit does not open the confirmation widget, or maybe it does but it doesn't put it in the foreground, or something like that.

Rumi-Larry: I'm a little confused, but I'll take your word for it 😁

PR #9825, opened 2021-06-18
JitArm64: Implement mcrfs

JosJuice:
I'm also planning to implement the other system register instructions that deal with FPSCR. But you will have to merge this PR first :)


Tas-player:
Why can't dolphins load DLC play while recording input

PR #9826, opened 2021-06-18
VideoCommon/TextureInfo: Fix mipmap loading from tmem

Techjar:
Even address needs to be offset to the first mipmap entry.

Fixes https://bugs.dolphin-emu.org/issues/12556

### DIFF
--- a/Source/Core/VideoCommon/TextureInfo.cpp
+++ b/Source/Core/VideoCommon/TextureInfo.cpp
@@ -84,8 +84,10 @@ TextureInfo::TextureInfo(const u8* ptr, const u8* tlut_ptr, u32 address,
     const u32 limited_mip_count =
         std::min<u32>(IntLog2(std::max(width, height)) + 1, raw_mip_count + 1) - 1;
 
-    // load mips - TODO: Loading mipmaps from tmem is untested!
+    // load mips
     const u8* src_data = m_ptr + GetTextureSize();
+    if (tmem_even)
+      tmem_even += GetTextureSize();
 
     for (u32 i = 0; i < limited_mip_count; i++)
     {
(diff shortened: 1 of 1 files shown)


iwubcode (approved):
Whoops.  Thanks Techjar!

Source/Core/VideoCommon/TextureInfo.cpp
```diff
@@ -86,6 +86,7 @@ TextureInfo::TextureInfo(const u8* ptr, const u8* tlut_ptr, u32 address,
 
     // load mips - TODO: Loading mipmaps from tmem is untested!
     const u8* src_data = m_ptr + GetTextureSize();
+    tmem_even += m_texture_size;
```
iwubcode, line 89: Actually, I think we only want to do this if this isn't a nullptr?

iwubcode, line 89: We should only do this if `tmem_even` is not a `nullptr`

iwubcode (changes requested):
Actually,  I was a bit too quick I think?  I believe the ptr needs to be checked for `nullptr`

Source/Core/VideoCommon/TextureInfo.cpp
Techjar: Probably a good idea as a matter of principle, though it doesn't technically matter since we don't dereference in that case.

iwubcode (approved):

Techjar:
> Actually, I was a bit too quick I think? I believe the ptr needs to be checked for `nullptr`

`ptr` is just `src_data` if we're not reading from tmem.

Source/Core/VideoCommon/TextureInfo.cpp
```diff
@@ -86,6 +86,8 @@ TextureInfo::TextureInfo(const u8* ptr, const u8* tlut_ptr, u32 address,
 
     // load mips - TODO: Loading mipmaps from tmem is untested!
```
Rumi-Larry, line 87: Shouldn't this comment be modified?

Techjar: I dunno, I didn't put there.

Techjar: I guess this counts as testing it, since the fact this fixes the issue proves that the game is loading mipmaps from tmem.

iwubcode: Yeah, this was an old comment from the previous code, no idea why it was there.  I just left it in when I ported it over.

PR #9827, opened 2021-06-19
AudioInterface: Cleanup

sepalani:
This PRs refactors the AudioInterface a bit. It removes some underflows and uses the IsPlaying function more rather than relying directly on the control's state.

There shouldn't be any functional differences.

### DIFF
--- a/Source/Core/Core/HW/AudioInterface.cpp
+++ b/Source/Core/Core/HW/AudioInterface.cpp
@@ -338,10 +338,11 @@ static void Update(u64 userdata, s64 cycles_late)
 
 int GetAIPeriod()
 {
-  u64 period = s_cpu_cycles_per_sample * (s_interrupt_timing - s_sample_counter);
-  u64 s_period = s_cpu_cycles_per_sample * s_ais_sample_rate;
-  if (period == 0)
+  const u64 s_period = s_cpu_cycles_per_sample * s_ais_sample_rate;
+  if (s_interrupt_timing <= s_sample_counter)
     return static_cast<int>(s_period);
+
+  const u64 period = s_cpu_cycles_per_sample * (s_interrupt_timing - s_sample_counter);
   return static_cast<int>(std::min(period, s_period));
 }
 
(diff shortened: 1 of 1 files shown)


Source/Core/Core/HW/AudioInterface.cpp
```diff
@@ -287,7 +287,7 @@ static void IncreaseSampleCount(const u32 amount)
   const u32 old_sample_counter = s_sample_counter + 1;
   s_sample_counter += amount;
 
-  if ((s_interrupt_timing - old_sample_counter) <= (s_sample_counter - old_sample_counter))
```
leoetlino: This doesn't look equivalent to me.

Source/Core/Core/HW/AudioInterface.cpp
```diff
@@ -338,10 +338,11 @@ static void Update(u64 userdata, s64 cycles_late)
 
 int GetAIPeriod()
 {
-  u64 period = s_cpu_cycles_per_sample * (s_interrupt_timing - s_sample_counter);
-  u64 s_period = s_cpu_cycles_per_sample * s_ais_sample_rate;
-  if (period == 0)
+  const u64 s_period = s_cpu_cycles_per_sample * s_ais_sample_rate;
+  if (s_interrupt_timing <= s_sample_counter)
```
leoetlino, line 342: I'm not convinced we should fail silently and return a value that possibly doesn't make sense rather than add an assertion for `s_interrupt_timing >= s_sample_counter`.

sepalani: We don't fail here. If I understand the logic correctly, the point of this function is to return the shortest period before the next AI event.

If we've already reached the sample count value where it triggers an interrupt (i.e. when `AIINTVLD` is set) then the only event we handle is the next AI period. Otherwise, if we may reach that sample count because `s_sample_counter < s_interrupt_timing`, then we take the smaller period of those two.

PR #9828, opened 2021-06-19
Android: Apply opacity setting to pressed buttons too

JosJuice:


MayImilae:
I thought the pressed buttons opacity already reflected the opacity setting and designed for that, so we're good from the assets front!

PR #9834, opened 2021-06-21
MemArena: Keep fastmem memory region mapped on Linux to ensure nothing allocates within the fastmem space.

AdmiralCurtiss:
Same idea as #9544 but for Linux. As I understand correctly Linux allows this brutal clobbering of regions, but I honestly only have half an idea what I'm doing here. Please scrutinize.

This should fix the bug (on Linux) where large texture packs break emulation, such as [12404](https://bugs.dolphin-emu.org/issues/12404) or [11994](https://bugs.dolphin-emu.org/issues/11994), but I cannot 100% confirm it.

### DIFF
--- a/Source/Core/Common/MemArena.h
+++ b/Source/Core/Common/MemArena.h
@@ -113,7 +113,13 @@ private:
   void* m_address_VirtualAlloc2 = nullptr;
   void* m_address_MapViewOfFile3 = nullptr;
 #else
+#ifdef ANDROID
   int fd;
+#else
+  int m_shm_fd;
+  void* m_reserved_region;
+  std::size_t m_reserved_region_size;
+#endif
 #endif
 };
 
--- a/Source/Core/Common/MemArenaUnix.cpp
+++ b/Source/Core/Common/MemArenaUnix.cpp
@@ -28,30 +28,39 @@ MemArena::~MemArena() = default;
 void MemArena::GrabSHMSegment(size_t size)
 {
   const std::string file_name = "/dolphin-emu." + std::to_string(getpid());
-  fd = shm_open(file_name.c_str(), O_RDWR | O_CREAT | O_EXCL, 0600);
-  if (fd == -1)
+  m_shm_fd = shm_open(file_name.c_str(), O_RDWR | O_CREAT | O_EXCL, 0600);
+  if (m_shm_fd == -1)
   {
     ERROR_LOG_FMT(MEMMAP, "shm_open failed: {}", strerror(errno));
     return;
   }
   shm_unlink(file_name.c_str());
-  if (ftruncate(fd, size) < 0)
+  if (ftruncate(m_shm_fd, size) < 0)
     ERROR_LOG_FMT(MEMMAP, "Failed to allocate low memory space");
 }
 
 void MemArena::ReleaseSHMSegment()
 {
-  close(fd);
+  close(m_shm_fd);
 }
 
 void* MemArena::CreateView(s64 offset, size_t size)
 {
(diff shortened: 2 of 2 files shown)


Source/Core/Common/MemArena.h
```diff
@@ -101,7 +101,13 @@ class MemArena final
...
   HANDLE hMemoryMapping;
 #else
+#ifdef ANDROID
   int fd;
+#else
+  int m_shm_fd;
+  void* m_reserved_region;
+  std::size_t m_reserved_region_size;
+#endif
 #endif
```
v1993, line 111:
```suggestion
#if defined(_WIN32)
  HANDLE hMemoryMapping;
#elif defined(ANDROID)
  int fd;
#else
  int m_shm_fd;
  void* m_reserved_region;
  std::size_t m_reserved_region_size;
#endif
```

v1993: Additionally - `hMemoryMapping` and `fd` lack `m_` prefix, so they probably should be renamed for consistency (or avoid using prefix for new member variables too).

AdmiralCurtiss: Regarding the variable prefixes, I've tried to avoid touching multiple platforms in a single PR, so they'll be done one at a time. The Windows equivalent PR is #9544. There is no Android one yet because I'm not familiar with that platform.

Regarding the defines, our codebase seems to use both variants. Not sure whether it's better to check for the symbol or for a nonzero definition.

v1993: Does new Linux approach fail to work on Android?

AdmiralCurtiss: Genuinely, no idea. I don't have an Android device capable of running Dolphin to test it.

v1993: With android being Linux at its core I'd expect this approach to work, unless it is artificially restricted by permissions. Unfortunately, I don't have Dolphin-capable android device/VM/emulator either for quick testing.

JosJuice: shm is not available on Android.

Source/Core/Common/MemArenaUnix.cpp
```diff
@@ -28,30 +28,39 @@ MemArena::~MemArena() = default;
 void MemArena::GrabSHMSegment(size_t size)
 {
   const std::string file_name = "/dolphin-emu." + std::to_string(getpid());
-  fd = shm_open(file_name.c_str(), O_RDWR | O_CREAT | O_EXCL, 0600);
-  if (fd == -1)
+  m_shm_fd = shm_open(file_name.c_str(), O_RDWR | O_CREAT | O_EXCL, 0600);
```
Warepire, line 31: If dolphin crashes, this file will stay until the system is rebooted. This is because it has the pid appended to it, so no other instance can come later and clean it up.

AdmiralCurtiss: Not wrong, but it's unlinked literally the next line, so it would have to crash at a *very* specific point in time. Not sure if there's much we can do about this? This is also a pre-existing issue.

Warepire: I completely missed that, then it's ok. Feel free to disregard this.

AdmiralCurtiss:
If no one has any objections I'd like to merge this sometime soon. The Windows variant has been merged for a few weeks now and apart from one issue of no longer launching on Win7 (which is fixed) I haven't heard any issues with it. I've also re-tested this today (on Xubuntu 20.04) and have confirmed that it does indeed fix freezing issues with large texture packs.

PR #9839, opened 2021-06-24
WiimoteScannerDarwin: fix hang on quit and clean up

OatmealDome:
* Minor clean up on `WiimoteScannerDarwin`.
* Moved `IOBluetoothHostController` initialization to the constructor.
* Used `defaultController` to get an instance of `IOBluetoothHostController` instead of creating one manually (`[[IOBluetoothHostController alloc] init]` is not mentioned anywhere in documentation, seems to work by coincidence).
* Removes a hack that only applies to old macOS SDKs.
* Fixes a potential issue where Wiimote scanning would continue even if `[IOBluetoothDeviceInquiry start]` returns an error.
* Fixes Dolphin hanging on quit if Continuous Scanning is enabled and no Wiimote was ever connected. ([#12424](https://bugs.dolphin-emu.org/issues/12424))

Tested on MBP 16" (2019), macOS Big Sur 11.4, Xcode 12.5. Continuous Scanning still works and I was able to play a game of Wii Sports tennis just fine.

### DIFF
--- a/Source/Core/Core/HW/WiimoteReal/IOdarwin.mm
+++ b/Source/Core/Core/HW/WiimoteReal/IOdarwin.mm
@@ -21,27 +21,40 @@
 
 namespace WiimoteReal
 {
+WiimoteScannerDarwin::WiimoteScannerDarwin()
+{
+  m_host_controller = [IOBluetoothHostController defaultController];
+  if (![m_host_controller addressAsString])
+  {
+    WARN_LOG_FMT(WIIMOTE, "No Bluetooth host controller");
+    
+    [m_host_controller release];
+    m_host_controller = nil;
+    
+    return;
+  }
+  
+  [m_host_controller retain];
+}
+
 WiimoteScannerDarwin::~WiimoteScannerDarwin()
 {
+  [m_host_controller release];
+  m_host_controller = nil;
+  
   m_stop_scanning = true;
 }
 
 void WiimoteScannerDarwin::FindWiimotes(std::vector<Wiimote*>& found_wiimotes,
                                         Wiimote*& found_board)
 {
-  // TODO: find the device in the constructor and save it for later
-  IOBluetoothHostController* bth;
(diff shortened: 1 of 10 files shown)


Source/Core/Core/HW/WiimoteReal/IOdarwin.mm
```diff
@@ -85,16 +98,14 @@ @interface ConnectBT : NSObject
...
+  
   [bti release];
   [sbt release];
 }
 
 bool WiimoteScannerDarwin::IsReady() const
 {
-  // TODO: only return true when a BT device is present
-  return true;
+  return m_host_controller != nil;
```
OatmealDome, line 108: Is this even used? As far as I can tell, ``WiimoteScannerBackend::IsReady`` is called from ``WiimoteScanner::IsReady()`` but nothing ever calls that.

JMC47:
@dolphin-emu-bot rebuild

MayImilae:
Oh wow, this addresses a lot of very old bugs! I'll test this soon.

OatmealDome:
OK, I think I fixed lint issues.

(If there's more macOS bugs/issues feel free to ping btw)

MayImilae:
Tested on my MacBook. Hanging on quit with continuous scanning is fixed! Also Wii Remotes still work, though I'm pretty sure that would have been noticed if it was otherwise. I don't really play games on this so I don't really want to do stability testing though so I can't comment on stability improvements from these fixes, however I got through a full stage of Sonic Colors without difficulty.

Tested on:
MacBook Pro 13in mid-2018 with Touch Bar
macOS 11.2.2
Core i7-8559U @ 4.5ghz
Intel Iris Plus Graphics 655
16GB 2133MHz LPDDR3

JMC47:
We don't have many mac testers/maintainers, and this appears to approve the situation dramatically.  Billiard did say that non-boilerplate parts of the code looked okay on IRC.

PR #9840, opened 2021-06-24
JitArm64: Implement mffsx

JosJuice:
Part 2 of implementing the FPSCR system register instructions.


degasus (approved):
No clue about the `FPSCR.FEX = 0;` part as it differs from the interpreter, but the ARM code looks good to me.

JosJuice:
I'm not sure about that either, but I did it because Jit64 does it.

PR #9841, opened 2021-06-24
Partially revert "Android: Clean up some hardcoded strings"

JosJuice:
This partially reverts commit cbc4989095baff7b84694802bb195c0c79d56ebe of PR #9835 due to a crash: https://bugs.dolphin-emu.org/issues/12561

I can't debug what the cause of the crash is due to not having an Android TV device. Let's just revert this for now to fix the crash.

### DIFF
--- a/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/services/SyncChannelJobService.java
+++ b/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/services/SyncChannelJobService.java
@@ -85,7 +85,7 @@ public class SyncChannelJobService extends JobService
       }
       else
       {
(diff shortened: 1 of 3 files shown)


JosJuice:
@dolphin-emu-bot rebuild

mbc07:
It wasn't the most straightforward process but I remember being able to compile a JIT-less version of the Android port before. You should be able to use a JIT-less build with the Android TV images from the official Android emulator (Google only publishes x86 images for Android TV)...

JMC47:
Can we have this merged before an article in order to make sure a beta isn't broken on T.V. devices?

leoetlino:
This is a pretty severe regression, so let's merge this to fix the crash for now.

PR #9842, opened 2021-06-25
[PoC] SW: Implement uninitialized vertex component behavior

Pokechu22:
A proof of concept implementation for the uninitialized vertex behavior (what happens when you try to use color 0 or color 1 when it's not specified in the vertex descriptor and not overridden by a material color).  This implementation only exists for the software renderer.  [Here](https://gist.github.com/Pokechu22/1e79aadcf1d0d77c1816884ba725cb06) is the hardware test, but this is something where I could probably write an automated test for the [hwtests repo](https://github.com/dolphin-emu/hwtests/).

The basic explanation is that each component is tracked in an array with 16 elements, and if an uninitialized value is used, the value from 16 vertices ago is used (but the index only changes when the value is present).  This is tracked separately for each component.  I can't think of a reason why it'd be implemented that way on hardware, but it still seems to be.

This should produce the same results as #9532 in most cases, as in the majority of cases, the previous vertices were white.  The only exception I'm currently aware of are some custom smash characters, where the new result should be accurate to what happens on hardware.

Some notes:
* This **does not fix** the debug cubes in Super Mario Sunshine.  This kind of behavior seems like it's the perfect candidate for causing it, but it's inconsistent with the actual behavior; there's still some other thing that's not discovered.
* This doesn't implement the behavior when `GX_SetNumChans` (which affects the bp genmode) is less than the number of used colors either.  That has its own set of oddities, but I'm fairly sure it's not what affects Sunshine either.
* I have only tested this behavior with color 0 and color 1 on hardware, but assume it also affects position, texture coordinates, and normals.  I got bad results when applying it to the position and texture coordinate matrix indices (which seem to have default values specified in xfmem).

### DIFF
--- a/Source/Core/VideoBackends/Software/SWVertexLoader.cpp
+++ b/Source/Core/VideoBackends/Software/SWVertexLoader.cpp
@@ -4,6 +4,7 @@
 
 #include "VideoBackends/Software/SWVertexLoader.h"
 
+#include <algorithm>
 #include <cstddef>
 #include <limits>
 
@@ -28,10 +29,6 @@
 #include "VideoCommon/VertexLoaderManager.h"
 #include "VideoCommon/VideoConfig.h"
 #include "VideoCommon/XFMemory.h"
-
-SWVertexLoader::SWVertexLoader() = default;
-
-SWVertexLoader::~SWVertexLoader() = default;
 
 void SWVertexLoader::DrawCurrentBatch(u32 base_index, u32 num_indices, u32 base_vertex)
 {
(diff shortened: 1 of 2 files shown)


Source/Core/VideoBackends/Software/SWVertexLoader.cpp
```diff
@@ -134,76 +131,90 @@ static T ReadNormalized(I value)
...
+  const std::size_t component_size = (1ull << (format.type >> 1));
+  src.Skip(base_component * component_size);
+
+  for (std::size_t i = 0; i < std::min(format.components - base_component, components); i++)
+  {
+    const std::size_t i_dst = reverse ? components - i - 1 : i;
+    switch (format.type)
     {
-      int i_dst = reverse ? components - i - 1 : i;
-      dst[i_dst] = i == 3;
```
Pokechu22, line 175: Note: this loop (which applies to the components not affected by the previous loop) is somewhat confusing, but I think the intention is to supply default values for uninitialized components (e.g. if only XY are provided, it gives a default for Z, and if only RGB are provided, it gives a default for A).  As far as I can tell, it sets the 4th component (`i == 3`, which becomes `dst[0]` when reverse is enabled) to 1 and the other uninitialized ones to 0, i.e. it's trying to set the alpha component to 1 for colors.  However, colors use a `u8`, so if it were trying to set it to the max value, it should use 255 instead.

I've changed the relevant code to use `std::fill`/`std::fill_n` and just always sets 0, which doesn't seem to have made a difference; I haven't hardware tested what actually happens in that case though.

I don't think that case could even be hit, because `format.components` is always set to 4 for colors ([here](https://github.com/dolphin-emu/dolphin/blob/c77a5f7e3246398aeb1d520d7bbc0bfdf5787754/Source/Core/VideoCommon/VertexLoader.cpp#L146)/[here](https://github.com/dolphin-emu/dolphin/blob/c77a5f7e3246398aeb1d520d7bbc0bfdf5787754/Source/Core/VideoCommon/VertexLoaderX64.cpp#L467)).  But this code is very weird looking so it's worth mentioning it.

Source/Core/VideoBackends/Software/SWVertexLoader.h
```diff
@@ -14,11 +15,37 @@
...
+  std::array<T, CACHE_SIZE> data;
+  u8 index;
+
+  T Read() { return data[index]; }
+  void Write(const T& value)
+  {
+    data[index] = value;
+    index = (index + 1) % CACHE_SIZE;
+  }
+};
```
Pokechu22, line 34: I've been calling this a cache, but I don't know if that's the best description.  It's definitely unrelated to the vertex cache, which [libogc claims](https://github.com/devkitPro/libogc/blob/cb1934ccfde50373236afec894c21bb620dfe9ac/gc/ogc/gx.h#L1764-L1766) is not used for direct components.

Perhaps calling it a queue or pipeline would be better, along the lines of "each vertex component has a 16-element queue for values, written into when that component is enabled.  If the component is not enabled, the value in the index that would have been written to next (i.e. the value from 16 components ago) is used, and the index does not change".

I also looked at the various patents to try and understand things better; the closest I've found is "Data required to process a particular component is stored within a queue 410 having a depth that is proportional to memory latency." in [US6717577B1](https://patents.google.com/patent/US6717577B1/en) (see Fig. 4 / image 9), but there's no guarantee that exactly describes what is used in practice.

Pokechu22:
Closing this because I don't think the implementation is correct, and I don't think this is a good starting point for further work.  I may return to this at some point, but probably I'll start from scratch then.

PR #9846, opened 2021-06-26
Config: Add an ini setting to control real Wii Remote report duplication.

jordan-woyak:
It is enabled by default.

Untested.

This should unbreak input in Super Mario Galaxy.

It will also re-break gestures in "Jett Rocket" and "Sonic and the Secret Rings".
Game INIs should be updated.

### DIFF
--- a/Source/Core/Core/HW/WiimoteReal/WiimoteReal.cpp
+++ b/Source/Core/Core/HW/WiimoteReal/WiimoteReal.cpp
@@ -16,6 +16,7 @@
 #include "Common/IniFile.h"
 #include "Common/Swap.h"
 #include "Common/Thread.h"
+#include "Core/Config/MainSettings.h"
 #include "Core/ConfigManager.h"
 #include "Core/Core.h"
 #include "Core/HW/Wiimote.h"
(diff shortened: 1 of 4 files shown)


Source/Core/Core/HW/WiimoteReal/WiimoteReal.cpp
```diff
@@ -442,7 +443,8 @@ void Wiimote::Update()
   // We duplicate data reports to accomplish this.
   // Unfortunately this breaks detection of motion gestures in some games.
   // e.g. Sonic and the Secret Rings, Jett Rocket
```
JosJuice, line 445: It would probably be good to mention in the comment that there are things other than speaker data which break if the input rate isn't around 200 Hz.

jordan-woyak: Done.

iwubcode:
> It will also re-break gestures in "Jett Rocket" and "Sonic and the Secret Rings".
Game INIs should be updated.

Could this be done as part of this PR?

leoetlino:
let's add the INIs as a follow-up PR since Billiard is busy

niekvlessert:
Works great! Thanks!!

PR #9847, opened 2021-06-27
DVDInterface: Don't evict cache block i unless block i + 2 was read

JosJuice:
Intends to fix https://bugs.dolphin-emu.org/issues/12279. I have hardware tested the behavior, but I haven't tested the game.

### DIFF
--- a/Source/Core/Core/HW/DVD/DVDInterface.cpp
+++ b/Source/Core/Core/HW/DVD/DVDInterface.cpp
@@ -1365,6 +1365,9 @@ static void ScheduleReads(u64 offset, u32 length, const DiscIO::Partition& parti
   const u32 ticks_per_second = SystemTimers::GetTicksPerSecond();
   const bool wii_disc = DVDThread::GetDiscType() == DiscIO::Platform::WiiDisc;
 
+  // Whether we have performed a seek.
+  bool seek = false;
+
   // Where the DVD read head is (usually parked at the end of the buffer,
   // unless we've interrupted it mid-buffer-read).
   u64 head_position;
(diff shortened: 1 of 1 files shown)


Avasam:
As was mentioned on IRC, while not a "complete" test, my early results seem to show that timing is on par with GameCube, and maybe ever so slightly slower than Wii.
[Timing comparison](https://prnt.sc/174du8i)

This also fixes Pitfall: The Big Adventure (Wii port), which had the exact same issue. I can't really tell exact timing differences, because I don't have enough data. But as far as I, and the rest of the Pitfall Speed running community, is concerned, this looks good and pretty close to console.

Timing comparisons as compiled by our community: https://docs.google.com/spreadsheets/d/16kTAM4g4qUmREwehROz_tzjOS3DXJHLGa44FKWy7GNI/edit#gid=0

Source/Core/Core/HW/DVD/DVDInterface.cpp
```diff
@@ -1512,26 +1512,33 @@ static void ScheduleReads(u64 offset, u32 length, const DiscIO::Partition& parti
...
   } while (length > 0);
 
-  // Update the buffer based on this read. Based on experimental testing,
-  // we will only reuse the old buffer while reading forward. Note that the
-  // buffer start we calculate here is not the actual start of the buffer -
-  // it is just the start of the portion we need to read.
+  // Update the buffer based on this read. Based on hardware testing, the blocks which are kept are
+  // the most recently accessed block, the block immediately before it, and all blocks after it.
+  // If the block immediately before the most recently accessed block is not kept, we get too long
+  // loading screens in Pitfall: The Lost Expedition (https://bugs.dolphin-emu.org/issues/1227).
```
Pokechu22, line 1518: This link is wrong; it should be https://bugs.dolphin-emu.org/issues/12279 (it's missing the 9).

Also, "we get too long loading screens" sounds a bit awkward to me; maybe "loading takes longer than it should" would be better?

Pokechu22 (approved):
Code looks good to me.

It might be useful to implement `DICommand::Seek` at some point, but it doesn't seem to be used by many games in my experience.  As far as I know, it essentially acts as a read except it doesn't perform the DMA afterwards.  Unfortunately I'm not sure where I read that.  But it's not something that needs to be done for this PR.

JMC47:
I retested again to make sure everything looked fine.  I'm assuming this won't cause any regressions, but who knows.

PR #9848, opened 2021-06-27
Implement resetting DI via HW_RESETS

Pokechu22:
This PR implements resetting DI via [HW_RESETS](https://github.com/Pokechu22/dolphin/pull/new/di-hw-resets).  This is mainly used for Wii homebrew that bypasses IOS (for instance, CleanRip now runs correctly in Dolphin, in case you want to copy dumped files in a slow, inconvenient way), but I also changed the implementation of IOS's [DVDLowReset](https://wiibrew.org/wiki//dev/di#0x8A_DVDLowReset) to use the MMIOs directly.  This also meant I needed to implement the starlet [GPIOs](https://wiibrew.org/wiki/Hardware/Hollywood_GPIOs), which Wiibrew says use the same backing value as the broadway-accessible ones; I haven't done any hardware testing on that recently, but my implementation at least seems to work.

### DIFF
--- a/Source/Core/Core/HW/WII_IPC.cpp
+++ b/Source/Core/Core/HW/WII_IPC.cpp
@@ -43,6 +43,12 @@ enum
   GPIOB_DIR = 0xc4,
   GPIOB_IN = 0xc8,
 
+  GPIO_OUT = 0xe0,
+  GPIO_DIR = 0xe4,
+  GPIO_IN = 0xe8,
+
+  HW_RESETS = 0x194,
+
   UNK_180 = 0x180,
   UNK_1CC = 0x1cc,
   UNK_1D0 = 0x1d0,
@@ -103,6 +109,8 @@ static constexpr Common::Flags<GPIO> gpio_owner = {GPIO::SLOT_LED, GPIO::SLOT_IN
 static Common::Flags<GPIO> gpio_dir;
 Common::Flags<GPIO> g_gpio_out;
 
+static u32 resets;
+
 static CoreTiming::EventType* updateInterrupts;
 static void UpdateInterrupts(u64 = 0, s64 cyclesLate = 0);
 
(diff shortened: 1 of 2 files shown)


Source/Core/Core/HW/WII_IPC.cpp
```diff
@@ -200,13 +219,63 @@ void RegisterMMIO(MMIO::Mapping* mmio, u32 base)
...
+
+  // Note from WiiBrew: When switching owners, copying of the data is not necessary. For example, if
+  // pin 0 has certain configuration in the HW_GPIO registers, and that bit is then set in the
+  // HW_GPIO_OWNER register, those settings will immediately be visible in the HW_GPIOB registers.
+  // There is only one set of data registers, and the HW_GPIO_OWNER register just controls the
+  // access that the HW_GPIOB registers have to that data.
+  // Also: The HW_GPIO registers always have read access to all pins, but any writes (changes) must
+  // go through the HW_GPIOB registers if the corresponding bit is set in the HW_GPIO_OWNER
+  // register.
+  mmio->Register(base | GPIO_OUT, MMIO::DirectRead<u32>(&g_gpio_out.m_hex),
```
BhaaLseN, line 242: How does this even work, since `base | 0xc0` is the same for both `GPIO_OUT` and `GPIOB_OUT` - does the 2nd one simply overwrite the 1st MMIO handler, or do _both_ run?

...unless the enum values up top are incorrect, and it's just that the last registered handler wins (which are those three)?

Pokechu22: Yep, the enum values are wrong (they should be e0/e4/e8).  It looks like it's the most recently registered handler that wins, too.  Will fix...

BhaaLseN: Interresting that it does work though, perhaps the `GPIOB_*` ones are also slightly incorrect (enough to make a difference in some cases, but not enough to cause problems at large)?

Initially I wanted to comment on the enum values themselves, but then figured there might be some magic happening in the MMIO registration (so I put it down here)

Pokechu22: The GPIOs are only used for a few things -- the only things they handle are ejecting the disc and powering the sensor bar... and it turns out there was a big issue with this change already: the power sensor bar bit wasn't being set, so the cursor was always hidden.  Toggling power for the sensor bar is used in two places (blinking the sensor bar when adjusting sensitivity in the Wii menu, and pretending to be a TV remote for [the television friend channel](https://wiki.dolphin-emu.org/index.php?title=Television_Friend_Channel)).  (The eject function is either not used at all, or only used as an alternative path for the system menu where you boot, eject the disc, and then shut down; it's not something that dolphin can trigger normally.)  Thanks for catching this; it really would have broken things.

Source/Core/Core/IOS/DI/DI.cpp
```diff
@@ -267,8 +271,30 @@ std::optional<DIDevice::DIResult> DIDevice::StartIOCtl(const IOCtlRequest& reque
...
+    {
+      // This route will not generally be taken in Dolphin but is included for completeness
+      // Syscall 0x45 deassert_di_reset
+      HW_RESETS = HW_RESETS | (1 << 10);
+    }
+    else
+    {
+      // Syscall 0x44 assert_di_reset
+      HW_RESETS = HW_RESETS & ~(1 << 10);
+      // Normally IOS sleeps for 12 microseconds here, but we can't easilly emulate that
```
leoetlino, line 294:
```suggestion
      // Normally IOS sleeps for 12 microseconds here, but we can't easily emulate that
```

Source/Core/Core/HW/WII_IPC.cpp
```diff
@@ -200,13 +219,62 @@ void RegisterMMIO(MMIO::Mapping* mmio, u32 base)
...
+                 MMIO::ComplexWrite<u32>([](u32, u32 val) {
+                   g_gpio_out.m_hex =
+                       (g_gpio_out.m_hex & gpio_owner.m_hex) | (val & ~gpio_owner.m_hex);
+                   if (g_gpio_out[GPIO::DO_EJECT])
+                   {
+                     INFO_LOG_FMT(WII_IPC, "Ejecting disc due to GPIO write");
+                     DVDInterface::EjectDisc(DVDInterface::EjectCause::Software);
+                   }
+                   // SENSOR_BAR is checked by WiimoteEmu::CameraLogic
+                   // TODO: AVE, SLOT_LED
```
leoetlino, line 252: (How would handling SLOT_LED even work? :eyes:)

Pokechu22: Maybe for gaming PCs, making the excessive number of LEDs flash bright blue?  There probably isn't a reasonable way to actually implement it, but it's still worth noting that it isn't implemented yet.

leoetlino (approved):

PR #9850, opened 2021-06-28
PowerPC: move SO bit to a more convenient location

Tilka:
Spacing the LT and SO bits like in the native PowerPC format makes it easier to convert to that format (see mfcr).


JosJuice (approved):
LGTM. It might be possible to improve `JitArm64::mfcr` now, but I can do that in a separate PR.

PR #9853, opened 2021-06-28
SI: Don't cancel transfer with TSTART=0

Bonta0:
As far as I can tell there is no such thing as transfer cancelling, I don't know where that came from. This hasn't been an issue so far because Dolphin's transfer are instantaneous except for GBAs
This was confirmed on hardware with
```
printf("start csr %08X buf %08X%08X\n", *si_comcsr, si_buffer[0], si_buffer[1]);
si_buffer[0] = 0; *si_comcsr = 0x40010303;
printf("started csr %08X buf %08X%08X\n", *si_comcsr, si_buffer[0], si_buffer[1]);
*si_comcsr &= ~SICOMCSR_TSTART;
printf("cancelled csr %08X buf %08X%08X\n", *si_comcsr, si_buffer[0], si_buffer[1]);
while(*si_comcsr & SICOMCSR_TSTART);
printf("waited csr %08X buf %08X%08X\n", *si_comcsr, si_buffer[0], si_buffer[1]);
```
image

### DIFF
--- a/Source/Core/Core/HW/SI/SI.cpp
+++ b/Source/Core/Core/HW/SI/SI.cpp
@@ -511,12 +511,10 @@ void RegisterMMIO(MMIO::Mapping* mmio, u32 base)
                    // be careful: run si-buffer after updating the INT flags
                    if (tmp_com_csr.TSTART)
                    {
+                     if (s_com_csr.TSTART)
+                       CoreTiming::RemoveEvent(s_tranfer_pending_event);
                      s_com_csr.TSTART = 1;
                      RunSIBuffer(0, 0);
-                   }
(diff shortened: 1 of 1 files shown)


JMC47:
@dolphin-emu-bot rebuild

booto (approved):
Looks okay to me - nice that the code is simpler with this. If you're feeling motivated, a test somewhere in https://github.com/dolphin-emu/hwtests exercising the behavior could be great, but thanks for the contribution all the same!

JMC47:
I've tested this and verified the behavior on both GameCube and Wii.  I ran the aforementioned tests on Wii, with a variety of controller configurations in order to verify the behavior.  LGTM.

PR #9854, opened 2021-06-29
DiscIO: Store partition key directly in ReuseID

JosJuice:
~The main reason I'm making this change is to avoid the ICE mentioned in https://github.com/dolphin-emu/dolphin/commit/e5b9e1ba1f7a8c7a48671f947cf096e0f4cb0fe9#r51445090, but there is also the advantage that~ this lets us reuse blocks in the case where different partition entries use the same key (which probably isn't actually all that common).

### DIFF
--- a/Source/Core/DiscIO/WIABlob.cpp
+++ b/Source/Core/DiscIO/WIABlob.cpp
@@ -1305,7 +1305,7 @@ WIARVZFileReader<RVZ>::ProcessAndCompress(CompressThreadState* state, CompressPa
     std::vector<u8>& data = parameters.data;
 
     if (AllSame(data))
-      entry.reuse_id = ReuseID{nullptr, data.size(), false, data.front()};
+      entry.reuse_id = ReuseID{WiiKey{}, data.size(), false, data.front()};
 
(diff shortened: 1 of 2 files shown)


JosJuice:
Actually, it seems like this doesn't fix the ICE...

Erisa:
> Actually, it seems like this doesn't fix the ICE...

Same here, was about to comment. Same ICE happens with this code.

JosJuice:
Well, I'll leave this PR open in case someone thinks this PR is worth merging for its secondary benefit. But hmm, should we add shuffle2's workaround in a separate PR or should we just wait for the next version of MSVC...

Erisa:
I personally think it would be worth attempting some kind of fix, so anyone else trying to build for Windows ARM64 doesn't have any unpleasant surprises like I did, until the next version of MSVC (Which I have no idea when that would even be).
Even if the number of people building for that platform is incredibly low, it currently doesn't build for an issue that's relatively hard to diagnose and that's a problem.

JosJuice:
Submitted as PR #9855.

PR #9855, opened 2021-06-29
DiscIO: Add workaround for MSVC ARM64 ICE

JosJuice:


iwubcode (approved):

Rumi-Larry:
Just out of curiosity, what does 'ICE' means in this context?

sepalani:
It means *Internal Compiler Error*: https://docs.microsoft.com/en-us/cpp/build/reference/errorreport-report-internal-compiler-errors

PR #9860, opened 2021-07-04
Add sharp bilinear postprocessing shader.

AdmiralCurtiss:
Attempt at porting [this shader](https://github.com/libretro/slang-shaders/blob/master/interpolation/shaders/sharp-bilinear.slang) to Dolphin, for those who prefer a more pixel-y look. Honestly it looks pretty bad to me for 3D games, but ymmv.

I'm not sure this algorithm is completely correct; I attempted to do the equivalent with ImageMagick like this:

```
magick RSBP01_2021-07-04_01-56-21.png -scale 1920x1704! native_scaled_point.png
magick native_scaled_point.png -filter triangle -resize 1920x1096! native_scaled_final.png
```

and the resulting image is slightly different. So... not sure. I'll need to experiment a bit more so I actually understand how this code works, I think.

For reference:

[Native resolution screenshot](https://user-images.githubusercontent.com/4522237/124371142-36022680-dc7f-11eb-8114-5c30dd9c2d02.png)
[As displayed by Dolphin](https://user-images.githubusercontent.com/4522237/124371143-38fd1700-dc7f-11eb-9f0c-3015f0c90196.png)
[As displayed by Dolphin if we were to force point filtering](https://user-images.githubusercontent.com/4522237/124371144-3a2e4400-dc7f-11eb-8f60-bf8a67503cb3.png)
[As displayed by Dolphin with this shader](https://user-images.githubusercontent.com/4522237/124371146-3a2e4400-dc7f-11eb-8372-a958394b5dd9.png)
[As scaled by ImageMagick](https://user-images.githubusercontent.com/4522237/124371148-3ac6da80-dc7f-11eb-8f53-9c8843227151.png)

Oh, and the postproc shader config system seems to be completely broken, they fail to compile if they have an option. Someone should probably look at that...

### DIFF
--- a/Data/Sys/Shaders/sharp_bilinear.glsl
+++ b/Data/Sys/Shaders/sharp_bilinear.glsl
@@ -1,0 +1,47 @@
+// Based on https://github.com/libretro/slang-shaders/blob/master/interpolation/shaders/sharp-bilinear.slang
+// by Themaister, Public Domain license
+// Does a bilinear stretch, with a preapplied Nx nearest-neighbor scale,
+// giving a sharper image than plain bilinear.
+
+/*
+[configuration]
+[OptionRangeFloat]
+GUIName = Prescale Factor (set to 0 for automatic)
+OptionName = PRESCALE_FACTOR
+MinValue = 0.0
+MaxValue = 16.0
+StepAmount = 1.0
+DefaultValue = 0.0
+[/configuration]
+*/
+
+float CalculatePrescale(float config_scale) {
+  if (config_scale == 0.0) {
+    float2 source_size = GetResolution();
+    float2 window_size = GetWindowResolution();
+    return ceil(max(window_size.x / source_size.x, window_size.y / source_size.y));
+  } else {
+    return config_scale;
+  }
+}
+
+void main()
+{
+  float2 source_size = GetResolution();
+  float2 texel = GetCoordinates() * source_size;
+  float2 texel_floored = floor(texel);
+  float2 s = fract(texel);
+  float config_scale = GetOption(PRESCALE_FACTOR);
+  float scale = CalculatePrescale(config_scale);
+  float region_range = 0.5 - 0.5 / scale;
+
+  // Figure out where in the texel to sample to get correct pre-scaled bilinear.
+  // Uses the hardware bilinear interpolator to avoid having to sample 4 times manually.
+
+  float2 center_dist = s - 0.5;
+  float2 f = (center_dist - clamp(center_dist, -region_range, region_range)) * scale + 0.5;
+
+  float2 mod_texel = texel_floored + f;
+
+  SetOutput(SampleLocation(mod_texel / source_size));
+}


MayImilae:
Pixel peeping, true point filtered is definitely sharper than Sharp Bilinear in these results, though the difference isn't _that_ much. You can see it in the top most slit of the lamp, the water, and some in the rope of the below images. Is Sharp Bilinear more desirable than simply point filtering?


Point Filter
pointerfilter-zoom

Sharp Bilinear
shader-zoom

iwubcode:
> Oh, and the postproc shader config system seems to be completely broken, they fail to compile if they have an option. Someone should probably look at that...

Please wait just a little longer :)

AdmiralCurtiss:
> Pixel peeping, true point filtered is definitely sharper than Sharp Bilinear in these results, though the difference isn't _that_ much. The blurring occurs when there is a "half pixel" when point sampling is doing shenanigans to fit the pixel grind, and it's sharp in point filtering but blurred in sharp bilinear. you can see it in the lamp especially the top slit.
> 
> Is Sharp Bilinear's behavior more desirable than simply point filtering? Regardless either solution would probably be fine for sharp pixels fans, but just asking.

The reason you want something like Sharp-Bilinear instead of straight Point filtering is to avoid uneven pixel sizing, which can look quite bad in motion. This is probably less of a big deal at GC/Wii resolutions and 3D games, but here's an example of NES Super Mario Bros, scaled 2x vertically and slightly above 2x horizontally to achieve the aspect ratio a real NES would have appoximately had on a real CRT back in the day:

shimmer_point_test

Look at the floor. Notice how the outlines of the blocks create a distracting pattern in motion? This has become known as 'shimmering' artifacts in enthusiast circles and happens because the individual columns of source pixels are unevenly sized -- some are 3 pixels wide, some are 2 wide -- so a pattern moving across them stands out poorly. Now if we instead scale with a bilinear filter:

shimmer_bilin_test

The distracting effect is gone! But now the whole picture is blurry. So what if we instead want both? Well, the commonly accepted solution has become to first scale up to the nearest-but-not-smaller integer multiple resolution (individually per axis), so in this case 2x vertically and 3x horizontally, then scaling back down to the desired actual resolution with a bilinear filter. This ends up with:

shimmer_sharpbilin_test

There, nice as can be given the limitations of modern fixed pixel grid displays.

This shader, if I'm understanding it correctly, attempts to do this exact thing but in a single scaling step by fudging with the bilinear scaling coefficients so they end up at a position where they produce an identical result.

MayImilae:
That is an extremely good point and well explained, thanks.

Data/Sys/Shaders/sharp_bilinear.glsl
```diff
@@ -0,0 +1,49 @@
...
+OptionRangeFloat
+GUIName = Prescale Factor (set to 0 for automatic)
+OptionName = PRESCALE_FACTOR
+MinValue = 0.0
+MaxValue = 16.0
+StepAmount = 1.0
+DefaultValue = 0.0
+
+[/configuration]
+*/
```
TryTwo, line 18:
```suggestion
/*
[configuration]
[OptionRangeFloat]
GUIName = Prescale_Factor
OptionName = PRESCALE
MinValue = 0.0
MaxValue = 16.0
StepAmount = 1.0
DefaultValue = 0.0
[/configuration]
```
I just figured this out. You then call it with PRESCALE, but it cannot be a part of a condition.

MayImilae:
> The reason you want something like Sharp-Bilinear instead of straight Point filtering is to avoid uneven pixel sizing, which can look quite bad in motion. This is probably less of a big deal at GC/Wii resolutions and 3D games, but here's an example of NES Super Mario Bros, scaled 2x vertically and slightly above 2x horizontally to achieve the aspect ratio a real NES would have appoximately had on a real CRT back in the day:
> 
> shimmer_point_test [ shimmer_point_test ](https://user-images.githubusercontent.com/4522237/124672201-eefe7600-deb6-11eb-84ca-b25efd3f3201.gif) [ ](https://user-images.githubusercontent.com/4522237/124672201-eefe7600-deb6-11eb-84ca-b25efd3f3201.gif)
> 
> Look at the floor. Notice how the outlines of the blocks create a distracting pattern in motion? This has become known as 'shimmering' artifacts in enthusiast circles and happens because the individual columns of source pixels are unevenly sized -- some are 3 pixels wide, some are 2 wide -- so a pattern moving across them stands out poorly. Now if we instead scale with a bilinear filter:
> 
> shimmer_bilin_test [ shimmer_bilin_test ](https://user-images.githubusercontent.com/4522237/124672200-ee65df80-deb6-11eb-95a9-d0a94aa00203.gif) [ ](https://user-images.githubusercontent.com/4522237/124672200-ee65df80-deb6-11eb-95a9-d0a94aa00203.gif)
> 
> The distracting effect is gone! But now the whole picture is blurry. So what if we instead want both? Well, the commonly accepted solution has become to first scale up to the nearest-but-not-smaller integer multiple resolution (individually per axis), so in this case 2x vertically and 3x horizontally, then scaling back down to the desired actual resolution with a bilinear filter. This ends up with:
> 
> shimmer_sharpbilin_test [ shimmer_sharpbilin_test ](https://user-images.githubusercontent.com/4522237/124672198-edcd4900-deb6-11eb-9845-2ef5ebb3d821.gif) [ ](https://user-images.githubusercontent.com/4522237/124672198-edcd4900-deb6-11eb-9845-2ef5ebb3d821.gif)
> 
> There, nice as can be given the limitations of modern fixed pixel grid displays.

@AdmiralCurtiss These images are great! Can I have them? Ok seriously, where are they from and do they have any licensing attached? I want to know if I can use them in the report directly or if I need to recreate them in Dolphin. If I can reuse them, it would save me like 6 hours of recording and video editing. I'll do it if I have to but I'd rather not!

AdmiralCurtiss:
Yeah sure, you can use them. They're not from Dolphin but a NES emulator, but for explaining the concept it totally works.

(Specificially, I recorded a lossless 1x video and then did the scaling/filtering/gif export with Avisynth.)

PR #9861, opened 2021-07-04
Fix Zora eye coloration in Twilight Princess

Pokechu22:
This fixes the appearance of Zora eyes (and bodies) in Twilight Princess; see [bug 10346](https://bugs.dolphin-emu.org/issues/10346).  This originally regressed in #68, and I bisected the issue to cff952c397833873055e42a85b76ff8a621a0792 (though shaders from that commit will not run unless line 707 is changed by replacing `int2(round(dot(...), dot(...)))` with `int2(round(dot(...)), round(dot(...)))`).

Twilight princess uses a scale value of 47, which becomes 2^(47 - 17) = 2^30; this is too big to fit into a 24-bit value.  Using a float seems to fix it, but I'm not sure entirely what is being done on hardware (and I don't have any good ideas on how to investigate it).  I've changed it to a float on all backends.

### DIFF
--- a/Source/Core/VideoBackends/Software/Tev.cpp
+++ b/Source/Core/VideoBackends/Software/Tev.cpp
@@ -482,7 +482,10 @@ void Tev::Indirect(unsigned int stageNum, s32 s, s32 t)
     return;
   }
 
-  s32 indtevtrans[2] = {0, 0};
+  // Using u32 to emulate an S24 seems to give the wrong result for Zora eyes in Twilight Princess
+  // (see bug 10346).  It uses a scale of 47, which results in a 2^30 multiplier, which doesn't make
+  // sense for S24, but it's not clear what actually should be done.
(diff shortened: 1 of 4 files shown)


Pokechu22:
OK, this change definitely isn't correct; it [breaks NES emulation](https://fifo.ci/compare/6383905-6381236/), and also the [goop bubbles in SMS](https://fifo.ci/compare/6383905-6381236/) ([issue 540](https://bugs.dolphin-emu.org/issues/540), originally fixed by #68), and maybe also [Armos appearances in WW](https://fifo.ci/compare/6383988-6381248/) ([issue 6683](https://bugs.dolphin-emu.org/issues/6683), originally fixed by #68); these seem to be things that care about not using floats.

Pokechu22:
Closing this as this seems to be the wrong approach; will open a second PR shortly.

PR #9862, opened 2021-07-05
Use SPDX for Dolphin licensing info

delroth:
The open source community is converging on a standard way for projects to convey licensing information, using the SPDX standard (http://spdx.dev). Replace Dolphin's current ad-hoc way of providing license information and use SPDX tags instead through `Source/`. Also reflect in the top-level licensing info that we have a bunch of licenses in here, not everything is GPLv2+, though we atempt to make the aggregate GPLv3 compatible.

Along the way fix a few issues, like files that were not properly relicensed back in 2015 even though we got the author's permission to do so.

Note: this is just a first step, there is still plenty that could be improved: figuring out how to handle externals, adding licensing info for stuff outside of Source/, etc. Ideally we would follow the http://reuse.software guidelines imo.


delroth:
As a note, I found that we're currently distributing at least one file (TAP_Win32.h) which seems to be GPLv2-only. I think this might technically be a license violation, though it could easily be argued that the contents of this file are not copyrightable (it's mostly trivial API definitions and some constants). Not planning to fix this here.

JosJuice (approved):

lioncash (approved):

lioncash:
Looks good to me. Look out down below, takin' the plunge.

LICENSES/BSD-3-Clause.txt
```diff
@@ -0,0 +1,11 @@
+Copyright (c) <year> <owner>. All rights reserved.
```
PEmu2, line 1: Is the date and copyright owner supposed to be filled in here?

delroth: No.

PR #9864, opened 2021-07-05
Change "Wiimote" to "Wii Remote" in interface

PEmu2:


MayImilae:
Real should be capitalized to match the Real Wii Remote setting, but otherwise seems like a good whoops fix.

PEmu2:
I think it makes sense to leave real lower case as it's part of the description in this context. When a setting is directly referenced in the UI elsewhere it is in quotation marks.

JosJuice:
I'm unsure if we should capitalize "real" or not, but let's get the actual change merged and worry about "real" later.

PR #9865, opened 2021-07-05
PowerPC: Raise alignment exceptions in more situations

JosJuice:
~Intends to improve https://bugs.dolphin-emu.org/issues/12565.~

To avoid affecting the performance, the JITs will most of the time not raise alignment exceptions unless you enable the new INI-only setting `AlignmentExceptions`.


Source/Core/Core/PowerPC/Jit64Common/Jit64AsmCommon.cpp
```diff
@@ -340,6 +340,9 @@ alignas(16) static const float m_m128 = -128.0f;
 // Sizes of the various quantized store types
 constexpr std::array<u8, 8> sizes{{32, 0, 0, 0, 8, 16, 8, 16}};
 
+// TODO: Use the actual instruction being emulated (needed for alignment exception emulation)
+static const UGeckoInstruction ps_placeholder_instruction = 0;
```
JosJuice, line 344: Could someone who knows Jit64 better give me some advice here? Basically, I want to pass an additional argument to the quantized loadstore routines (so that the slowmem handler then can forward the argument when calling `Read_*` or `Write_*`), but all of `RSCRATCH`, `RSCRATCH2` and `RSCRATCH_EXTRA` are already used for arguments to the quantized loadstore routines. How should I pass the argument?

JosJuice:
While alignment exceptions turned out to have nothing to do with issue 12565, alignment exceptions are still something that would make sense for us to emulate (even if it doesn't affect any games). I've cleaned up the PR after the merge of the PR that fixed 12565, and it should be more or less done other than my earlier question about Jit64.

phire:
Do you have any rough measurements for how big of a performance hit checking for alignment exceptions is?

You are probably right about there being a hit, but I'd rather it was enabled by default (Or tied to the MMU option), if that's at all feasible.

JosJuice:
I tested the intro of Wind Waker on my laptop now, and unfortunately it's significant. At the most lightweight part of the intro I got around 80 FPS without checking for alignment exceptions, and around 50 FPS with. The difference in heavier parts of the intro was smaller – a difference of 5 to 10 FPS or so.

The part of the change that I think is the heaviest (I haven't benchmarked it) is that the JIT emits a few extra instructions for every float loadstore to manually check its alignment.

phire:
Yeah, that's large.

JosJuice:
@dolphin-emu-bot rebuild

dreamsyntax:
This might be expected behavior, but this never raises as an unhandled exception for me, and instead goes to a silent PPCHalt.

JosJuice:
Does that differ from the behavior on console?

dreamsyntax:
> Does that differ from the behavior on console?

Just checked, hardware throws
> Unhandled Exception 5

And verified Dolphin does too.

So it is working as expected. Panic Handlers do not get invoked however, and imo maybe they should? Though that would not be related to this PR.

dreamsyntax (approved):

JosJuice:
Please note that this PR is not done. My review comment from July 2024 still needs to be resolved in some way.

Source/Core/Core/PowerPC/Jit64Common/Jit64AsmCommon.cpp
JosJuice: Welp, it's been five years and I'm the Jit64 expert now. I'll just have to figure something out.

PR #9866, opened 2021-07-05
JitArm64: Don't hardcode exception constants

JosJuice:
Split out from PR #9865, since I don't think that PR is going to be merged any time soon.

Unlike most constants we emit in JitArm64, these constants are **not** inherent to the CPU we're emulating, and can have whatever  values we want, so let's handle them in a more robust way. I was rather confused when I found out that a previous iteration of PR #9865 broke JitArm64, because I hadn't thought that making some simple changes to the values of an enum would do it!

### DIFF
--- a/Source/Core/Core/PowerPC/JitArm64/Jit.cpp
+++ b/Source/Core/Core/PowerPC/JitArm64/Jit.cpp
@@ -748,7 +748,7 @@ void JitArm64::DoJit(u32 em_address, JitBlock* b, u32 nextPC)
 
       // Inline exception check
       LDR(IndexType::Unsigned, ARM64Reg::W30, PPC_REG, PPCSTATE_OFF(Exceptions));
-      TBZ(ARM64Reg::W30, 3, done_here);  // EXCEPTION_EXTERNAL_INT
+      TBZ(ARM64Reg::W30, IntLog2(EXCEPTION_EXTERNAL_INT), done_here);
       LDR(IndexType::Unsigned, ARM64Reg::W30, PPC_REG, PPCSTATE_OFF(msr));
       TBZ(ARM64Reg::W30, 11, done_here);
       MOVP2R(ARM64Reg::X30, &ProcessorInterface::m_InterruptCause);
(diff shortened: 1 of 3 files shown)


Source/Core/Core/PowerPC/JitArm64/Jit.cpp
```diff
@@ -748,7 +748,7 @@ void JitArm64::DoJit(u32 em_address, JitBlock* b, u32 nextPC)
 
       // Inline exception check
       LDR(IndexType::Unsigned, ARM64Reg::W30, PPC_REG, PPCSTATE_OFF(Exceptions));
-      TBZ(ARM64Reg::W30, 3, done_here);  // EXCEPTION_EXTERNAL_INT
+      TBZ(ARM64Reg::W30, IntLog2(EXCEPTION_EXTERNAL_INT), done_here);
```
leoetlino, line 751: This doesn't seem equivalent? EXCEPTION_EXTERNAL_INT is 4 and log2(4) = 2 (and IntLog2 does seem to return 2), but the original code used 3. Is that an pre-existing bug?

JosJuice: I would assume so. I haven't been able to observe any difference in behavior from changing it from 3 to 2, though.

Source/Core/Core/PowerPC/JitArm64/JitArm64_SystemRegisters.cpp
```diff
@@ -218,7 +218,7 @@ void JitArm64::twx(UGeckoInstruction inst)
   fpr.Flush(FlushMode::MaintainState);
 
   LDR(IndexType::Unsigned, WA, PPC_REG, PPCSTATE_OFF(Exceptions));
-  ORR(WA, WA, 24, 0);  // Same as WA | EXCEPTION_PROGRAM
+  ORRI2R(WA, WA, EXCEPTION_PROGRAM);
```
leoetlino, line 221: Worth noting that there was a bug in the previous code, which actually OR'd with EXCEPTION_PERFORMANCE_MONITOR (0x100), not EXCEPTION_PROGRAM (0x80).

leoetlino (approved):

Source/Core/Core/PowerPC/JitArm64/Jit.cpp
```diff
@@ -774,7 +774,7 @@ void JitArm64::DoJit(u32 em_address, JitBlock* b, u32 nextPC)
       ARM64Reg WA = gpr.GetReg();
       ARM64Reg XA = EncodeRegTo64(WA);
       LDR(IndexType::Unsigned, WA, PPC_REG, PPCSTATE_OFF(Exceptions));
-      FixupBranch NoExtException = TBZ(WA, 3);  // EXCEPTION_EXTERNAL_INT
+      FixupBranch NoExtException = TBZ(WA, IntLog2(EXCEPTION_EXTERNAL_INT));
```
JosJuice, line 777: This change specifically seems to have broken Pokémon XD (and presumably other games as well). Will investigate.

JosJuice: Fixed in PR #9872.

PR #9869, opened 2021-07-06
JitArm64: Encode logical immediates at compile-time where possible

JosJuice:
Manually encoding and decoding logical immediates is error-prone. Using ORRI2R and friends lets us avoid doing the work manually, but in exchange, there is a runtime performance penalty. It's probably rather small, but still, it would be nice if we could let the compiler do the work at compile-time. And that's exactly what this pull request does, so now I have no excuse for trying to manually write logical immediates anymore.

### DIFF
--- a/Source/Core/Common/Arm64Emitter.cpp
+++ b/Source/Core/Common/Arm64Emitter.cpp
@@ -28,11 +28,6 @@ namespace Arm64Gen
 {
 namespace
 {
-uint64_t LargestPowerOf2Divisor(uint64_t value)
-{
-  return value & -(int64_t)value;
-}
-
 // For ADD/SUB
 std::optional<std::pair<u32, bool>> IsImmArithmetic(uint64_t input)
 {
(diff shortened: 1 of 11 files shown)


Source/Core/Common/Arm64Emitter.h
```diff
@@ -920,6 +944,211 @@ class ARM64XEmitter
   bool TryORRI2R(ARM64Reg Rd, ARM64Reg Rn, u64 imm);
   bool TryEORI2R(ARM64Reg Rd, ARM64Reg Rn, u64 imm);
 
+  // For AND/TST/ORR/EOR etc
+  constexpr LogicalImm EncodeLogicalImm(u64 value, u32 width)
```
degasus, line 948: This sounds like a good constructor of `LogicalImm` to me. Feel free to chose between explicit or implicit one.

Source/Core/Common/Arm64Emitter.h
```diff
@@ -772,10 +789,15 @@ class ARM64XEmitter
...
   // Logical (immediate)
   void AND(ARM64Reg Rd, ARM64Reg Rn, u32 immr, u32 imms, bool invert = false);
+  void AND(ARM64Reg Rd, ARM64Reg Rn, LogicalImm imm);
   void ANDS(ARM64Reg Rd, ARM64Reg Rn, u32 immr, u32 imms, bool invert = false);
+  void ANDS(ARM64Reg Rd, ARM64Reg Rn, LogicalImm imm);
   void EOR(ARM64Reg Rd, ARM64Reg Rn, u32 immr, u32 imms, bool invert = false);
+  void EOR(ARM64Reg Rd, ARM64Reg Rn, LogicalImm imm);
   void ORR(ARM64Reg Rd, ARM64Reg Rn, u32 immr, u32 imms, bool invert = false);
+  void ORR(ARM64Reg Rd, ARM64Reg Rn, LogicalImm imm);
   void TST(ARM64Reg Rn, u32 immr, u32 imms, bool invert = false);
```
degasus, line 799: Is there a point in keeping the old versions? I assume we never want to generate this functions with immr/imms, and even if we want so, we can use the LogicalImm constructor.

Source/Core/Common/Arm64Emitter.cpp
```diff
@@ -784,6 +571,18 @@ void ARM64XEmitter::EncodeLogicalImmInst(u32 op, ARM64Reg Rd, ARM64Reg Rn, u32 i
...
+void ARM64XEmitter::EncodeLogicalImmInst(u32 op, ARM64Reg Rd, ARM64Reg Rn, LogicalImm imm)
+{
+  ASSERT_MSG(DYNAREC, imm.valid, "Invalid logical immediate");
+
+  // Sometimes Rd is fixed to SP, but can still be 32bit or 64bit.
+  // Use Rn to determine bitness here.
+  bool b64Bit = Is64Bit(Rn);
+
+  Write32((b64Bit << 31) | (op << 29) | (0x24 << 23) | (imm.n << 22) | (imm.r << 16) |
+          (imm.s << 10) | (DecodeReg(Rn) << 5) | DecodeReg(Rd));
```
degasus, line 583: Please just call the other implementation of `EncodeLogicalImmInst` instead. Having this constants twice makes it mor error prone.

Source/Core/Common/Arm64Emitter.h
```diff
@@ -903,14 +927,14 @@ class ARM64XEmitter
...
   void EORI2R(ARM64Reg Rd, ARM64Reg Rn, u64 imm, ARM64Reg scratch = ARM64Reg::INVALID_REG);
-  void CMPI2R(ARM64Reg Rn, u64 imm, ARM64Reg scratch = ARM64Reg::INVALID_REG);
 
   void ADDI2R_internal(ARM64Reg Rd, ARM64Reg Rn, u64 imm, bool negative, bool flags,
                        ARM64Reg scratch);
   void ADDI2R(ARM64Reg Rd, ARM64Reg Rn, u64 imm, ARM64Reg scratch = ARM64Reg::INVALID_REG);
   void ADDSI2R(ARM64Reg Rd, ARM64Reg Rn, u64 imm, ARM64Reg scratch = ARM64Reg::INVALID_REG);
   void SUBI2R(ARM64Reg Rd, ARM64Reg Rn, u64 imm, ARM64Reg scratch = ARM64Reg::INVALID_REG);
   void SUBSI2R(ARM64Reg Rd, ARM64Reg Rn, u64 imm, ARM64Reg scratch = ARM64Reg::INVALID_REG);
+  void CMPI2R(ARM64Reg Rn, u64 imm, ARM64Reg scratch = ARM64Reg::INVALID_REG);
```
degasus, line 937: What do you think about dropping the default value of `ARM64Reg::INVALID_REG`? If no register is specified, the caller should not use the I2R method.

degasus (approved):
I really approach this concept of having a constexpr ImmLogical object. Those immr/imms have confused me so often already...

Source/Core/Common/Arm64Emitter.h
JosJuice: I was planning to clean up the calls to the immr/imms version in a follow-up PR. I could remove it in that PR.

degasus: Fine. No need to make this one larger :)

Source/Core/Common/Arm64Emitter.h
JosJuice: There are actually some cases where the value is guaranteed to not need a scratch register even though the value isn't known at compile time, like here: https://github.com/dolphin-emu/dolphin/blob/cfcc994f6c6966172c108ee73347d7ad3f0ae717/Source/Core/Core/PowerPC/JitArm64/JitArm64_SystemRegisters.cpp#L717

JosJuice: Ah, but you mean it would be possible to just call `EncodeLogicalImm` at runtime... Yeah, I suppose that makes sense. I'll get rid of the default argument for the bitwise instructions. (But not for `CMPI2R`, because that one is not a bitwise instruction :))

Source/Core/Common/Arm64Emitter.cpp
```diff
@@ -784,6 +571,18 @@ void ARM64XEmitter::EncodeLogicalImmInst(u32 op, ARM64Reg Rd, ARM64Reg Rn, u32 i
           (DecodeReg(Rn) << 5) | DecodeReg(Rd));
 }
 
+void ARM64XEmitter::EncodeLogicalImmInst(u32 op, ARM64Reg Rd, ARM64Reg Rn, LogicalImm imm)
+{
+  ASSERT_MSG(DYNAREC, imm.valid, "Invalid logical immediate");
```
degasus, line 576: Could you also add an assert if the width of the imm matches the width of Rn?

JosJuice: I can add an assert that you don't use a 64-bit imm with a 32-bit register, but all other combinations of register size and imm width are valid.

PR #8544, opened 2020-01-02
MacOS: Dispatch GL calls to main thread to prevent crashes on Catalina

gtalusan:
Hello, everybody.  First time here.

I discovered Dolphin today and tried out the development build on Catalina.  It didn't work too well with a bunch of hanging, so I built it up and pulled in #8530 to fix some of the chronic hangs and non-responsive controllers.

After configuring my controllers, selecting a ROM immediately led to this crash in AttachContextToView:

```
Process:               Dolphin [78328]
Path:                  /Users/USER/*/Dolphin.app/Contents/MacOS/Dolphin
Identifier:            org.dolphin-emu.dolphin
Version:               5.0-11422 (5.0)
Code Type:             X86-64 (Native)
Parent Process:        zsh [12794]
Responsible:           Terminal [2118]
User ID:               501

Date/Time:             2020-01-01 21:26:09.146 -0500
OS Version:            Mac OS X 10.15.3 (19D49f)
Report Version:        12
Anonymous UUID:        82114560-50F3-87D4-9575-CCC256799B8D


Time Awake Since Boot: 1100000 seconds

System Integrity Protection: enabled

Crashed Thread:        21  Emuthread - Starting

Exception Type:        EXC_BAD_INSTRUCTION (SIGILL)
Exception Codes:       0x0000000000000001, 0x0000000000000000
Exception Note:        EXC_CORPSE_NOTIFY

Termination Signal:    Illegal instruction: 4
Termination Reason:    Namespace SIGNAL, Code 0x4
Terminating Process:   exc handler [78328]

Application Specific Information:
-[NSOpenGLContext setView:] must be called from the main thread.

Thread 0:: Dispatch queue: com.apple.main-thread
0   libsystem_kernel.dylib        	0x00007fff63a6a25a mach_msg_trap + 10
1   libsystem_kernel.dylib        	0x00007fff63a6a5d0 mach_msg + 60
2   com.apple.CoreFoundation      	0x00007fff2be13d0b __CFRunLoopServiceMachPort + 322
3   com.apple.CoreFoundation      	0x00007fff2be128e7 __CFRunLoopRun + 1695
4   com.apple.CoreFoundation      	0x00007fff2be11bd3 CFRunLoopRunSpecific + 499
5   com.apple.HIToolbox           	0x00007fff2a96765d RunCurrentEventLoopInMode + 292
6   com.apple.HIToolbox           	0x00007fff2a96739d ReceiveNextEventCommon + 600
[...]

PR #9870, opened 2021-07-06
macOS: Move UI API calls to the main thread

OatmealDome:
Just a minor cleanup that makes debugging in Xcode a better experience.

Some APIs must only be called from the main thread. The fact that these APIs were called from other threads and still worked without crashing / bugs is a coincidence. (The AGL GLInterface has apparently already been affected by this before, and was mostly fixed in #8544.)

Also, I define ``GL_SILENCE_DEPRECATION`` in all files that use the AppKit OpenGL APIs. This silences all warnings that say "OpenGL is deprecated, use Metal" when an OpenGL-related API is accessed. I think we all know that by now.

### DIFF
--- a/Source/Core/Common/GL/GLInterface/AGL.h
+++ b/Source/Core/Common/GL/GLInterface/AGL.h
@@ -3,6 +3,9 @@
 
 #pragma once
 
+// This define will silence all "OpenGL is deprecated, use Metal" warnings.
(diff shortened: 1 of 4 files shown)


Pokechu22:
@dolphin-emu-bot rebuild

leoetlino (approved):
LGTM, changes seem sane

PR #9874, opened 2021-07-07
JitArm64: Get rid of an unconditional branch to farcode

JosJuice:
If a branch is unconditional, its target should not be in farcode, since that defeats the purpose of farcode (putting seldom executed code in farcode to keep it out of the icache when possible).


Source/Core/Core/PowerPC/JitArm64/Jit.cpp
```diff
@@ -759,6 +759,7 @@ void JitArm64::DoJit(u32 em_address, JitBlock* b, u32 nextPC)
       fpr.Flush(FlushMode::MaintainState);
       WriteExceptionExit(js.compilerPC, true);
       SwitchToNearCode();
+      SetJumpTarget(no_ext_exception);
```
degasus, line 762: The above `B(CC_EQ, done_here);` should also jump here directly.

degasus (approved):

PR #9875, opened 2021-07-07
Fix a postprocessing bug and add scanline effect

blaahaj:
Yes, it is not perfectly accurate!

### DIFF
--- a/Data/Sys/Shaders/crt_scanlines.glsl
+++ b/Data/Sys/Shaders/crt_scanlines.glsl
@@ -1,0 +1,39 @@
+/*
+[configuration]
+
+[OptionRangeFloat]
+GUIName = Inactive Field Brightness
+OptionName = INACTIVE_BRIGHT
+MinValue = 0.0
+MaxValue = 1.0
+StepAmount = 0.1
+DefaultValue = 0.0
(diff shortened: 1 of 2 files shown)


iwubcode (changes requested):
Thank you for your contribution!

I'm going to push back against this.  Not because there's anything inherently wrong with this change but because _very soon_ there will be better alternatives.

I'm in the process of rewriting the post-processing framework into what I hope will be a much nicer framework.  As part of this, I'm purging a majority of the Dolphin shaders (at least that's the plan) and instead keeping the ones that are useful or interesting.  Each one that I keep may also require additional conversion to the new configuration structure.  Because of that, I'm hesitant to add new shaders if they will just be thrown away or require additional work for me to convert.

Finally, I think CRT shaders are a great addition but [I'd rather pick one of the many](http://filthypants.blogspot.com/2015/04/more-crt-shaders.html) that are "standard" solutions to this problem.

blaahaj:
Would it be ok to merge just the the fix for the UBO link error (the 1st commit), then?

blaahaj:
https://github.com/dolphin-emu/dolphin/pull/9881

PR #9876, opened 2021-07-07
Fix indirect textures when format is not ITF_8

Pokechu22:
This affects the skyward sword map.  As mentioned in [issue 9070](https://bugs.dolphin-emu.org/issues/9070#note-11), this regressed in #68 (tev_fixes_new), specifically in 3ea97f77301d57f7ffddefb45f7e34ca926d5098.  It turns out that this commit swapped the shifts used for "bump alpha" and the indirect offset; [per libogc](https://github.com/devkitPro/libogc/blob/bd24a9b3f59502f9b30d6bac0ae35fc485045f78/gc/ogc/gx.h#L790-L800), the indirect offsets are the high bits and the bump alpha comes from the low bits.  (This is further confounded by bump alpha [being only 5 bits](https://github.com/devkitPro/libogc/blob/bd24a9b3f59502f9b30d6bac0ae35fc485045f78/gc/ogc/gx.h#L3038-L3041); in the most common ITF_8 case where bump alpha and the indirect offset are both from the same value, it seems like it does use the top 5 bits, though I haven't hardware tested it).  The old code handled it mostly correctly, but because it was trying to implement bit manipulation using floats, it had various rounding issues (and was hard to understand).

### DIFF
--- a/Source/Core/VideoCommon/PixelShaderGen.cpp
+++ b/Source/Core/VideoCommon/PixelShaderGen.cpp
@@ -1019,7 +1019,7 @@ static void WriteStage(ShaderCode& out, const pixel_shader_uid_data* uid_data, i
           '3',  // ITF_3: 0bIIIAAAAA -> 0bAAAAA000, shift of 3
       };
 
-      out.Write("alphabump = (iindtex{}.{} << {}) & 248;\n", tevind.bt.Value(),
+      out.Write("\talphabump = (iindtex{}.{} << {}) & 248;\n", tevind.bt.Value(),
                 tev_ind_alpha_sel[u32(tevind.bs.Value())],
                 tev_ind_alpha_shift[u32(tevind.fmt.Value())]);
     }
--- a/Source/Core/VideoCommon/PixelShaderGen.cpp
+++ b/Source/Core/VideoCommon/PixelShaderGen.cpp
@@ -1006,17 +1006,22 @@ static void WriteStage(ShaderCode& out, const pixel_shader_uid_data* uid_data, i
           "z",
       };
 
-      // 0b11111000, 0b11100000, 0b11110000, 0b11111000
-      static constexpr std::array<const char*, 4> tev_ind_alpha_mask{
-          "248",
-          "224",
-          "240",
-          "248",
+      // According to libogc, the bump alpha value is 5 bits, and comes from the bottom bits of the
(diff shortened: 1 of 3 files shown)


Source/Core/VideoCommon/PixelShaderGen.cpp
```diff
@@ -1006,17 +1006,22 @@ static void WriteStage(ShaderCode& out, const pixel_shader_uid_data* uid_data, i
...
-          "224",
-          "240",
-          "248",
+      // According to libogc, the bump alpha value is 5 bits, and comes from the bottom bits of the
+      // component byte, except in the case of ITF_8, which presumably uses the top bits with a
+      // mask.
+      // https://github.com/devkitPro/libogc/blob/bd24a9b3f59502f9b30d6bac0ae35fc485045f78/gc/ogc/gx.h#L3038-L3041
+      // https://github.com/devkitPro/libogc/blob/bd24a9b3f59502f9b30d6bac0ae35fc485045f78/gc/ogc/gx.h#L790-L800
+
+      static constexpr std::array<const char*, 4> tev_ind_alpha_shift{
```
lioncash, line 1015: This can be changed into:

```suggestion
      static constexpr std::array<char, 4> tev_ind_alpha_shift{
```

Since they're all single character elements

Source/Core/VideoCommon/PixelShaderGen.cpp
```diff
@@ -1026,14 +1031,14 @@ static void WriteStage(ShaderCode& out, const pixel_shader_uid_data* uid_data, i
     if (has_ind_stage && tevind.matrix_index != IndMtxIndex::Off)
     {
       // format
-      static constexpr std::array<const char*, 4> tev_ind_fmt_mask{
-          "255",
-          "31",
-          "15",
-          "7",
+      static constexpr std::array<const char*, 4> tev_ind_fmt_shift{
```
lioncash, line 1034: Ditto

Pokechu22:
I've done a write-up of how the rendering on the map works, available here: https://gist.github.com/Pokechu22/dfebe2758d23faeba48bb7c9fb8557a2.  It should explain everything pretty clearly.

Tilka (approved):

philou-felin:
Thank you! I'm just an end user. Who fixed that? Pokechu22? This has to be highlighted in the next monthly report!

PR #9879, opened 2021-07-09
Core/MMU: Fix inverted condition in HostIsInstructionRAMAddress().

AdmiralCurtiss:
My bad. Untested, but this should fix https://bugs.dolphin-emu.org/issues/12571

Compare with previous code: https://github.com/dolphin-emu/dolphin/commit/b59fcae70abd7bd14d0a87aef9f07cd2481f905c#diff-4e481906f161c6ca9062f7b02499e2eca16f00bab97f94bb44ab8bb292c47385L699-R832

https://github.com/dolphin-emu/dolphin/blob/954f27c5d7289b2b806d0fb94dd105edaa1bb4be/Source/Core/Core/PowerPC/MMU.cpp#L698-L699
https://github.com/dolphin-emu/dolphin/blob/b59fcae70abd7bd14d0a87aef9f07cd2481f905c/Source/Core/Core/PowerPC/MMU.cpp#L830-L832


leoetlino:
oh, whoops

PR #9881, opened 2021-07-09
PostProcessing: Fix OpenGL UBO linking with configuration options

blaahaj:


stenzek (approved):
Nice catch. Bad design that the loading is part of `CreatePixelShader()`, it should really be its own function.

PR #9884, opened 2021-07-11
JitArm64: Improve psq_l/psq_st address checking

JosJuice:
psq_l and psq_st, which both do not use backpatching in the general case due to their reliance on asm routines, are rather lacking when it comes to address checking. psq_st assumes fixed BAT mappings, and psq_l has no address checking at all – if you give it a funny-looking address it will simply crash. This PR improves the situation by implementing Jit64's way of handling psq_l/psq_st address checking.


JosJuice:
@dolphin-emu-bot rebuild

PR #9889, opened 2021-07-12
DSPHLE: Prevent AX GC and AX Wii symbol mixups

leoetlino:
Putting AX functions from AXVoice.h in an anonymous namespace does
successfully prevent compilers from merging those functions and
allows us to avoid ODR violations.

However, tools such as gdb still mix up AX GC and AX Wii functions
and variables because those have the exact same symbol names.

This can be fixed by using inline namespaces which are transparent
at the source code level but forces AX GC and AX Wii symbols to be
different.


Tilka:
Can't the anonymous namespace be removed now?

leoetlino:
Technically it could, but we don't use those functions externally so it makes sense for them to be in an anonymous namespace and get internal linkage.

PR #9892, opened 2021-07-13
JitArm64: Add temp reg parameter to Arm64RegCache::Flush

JosJuice:
We currently have a bug when calling `Arm64GPRCache::Flush` with `FlushMode::MaintainState`, zero free host registers, and at least one guest register containing an immediate. We end up grabbing a temporary register from the register cache in order to be able to write the immediate to memory, but grabbing a temporary register when there are zero free registers causes the least recently used register to be flushed in a way which does not maintain the state of the register cache.

To get around this, require callers to pass in a temporary register in the GPR MaintainState case. In other cases, passing in a temporary register is not required but can help avoid spilling a register (if the caller already had a temporary register at hand anyway, which in particular will be the case in my upcoming memcheck pull request).

### DIFF
--- a/Source/Core/Core/PowerPC/JitArm64/Jit.cpp
+++ b/Source/Core/Core/PowerPC/JitArm64/Jit.cpp
@@ -145,8 +145,8 @@ void JitArm64::Shutdown()
 void JitArm64::FallBackToInterpreter(UGeckoInstruction inst)
 {
   FlushCarry();
-  gpr.Flush(FlushMode::All, js.op);
-  fpr.Flush(FlushMode::All, js.op);
+  gpr.Flush(FlushMode::All, ARM64Reg::INVALID_REG);
+  fpr.Flush(FlushMode::All, ARM64Reg::INVALID_REG);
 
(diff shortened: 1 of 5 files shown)


XiaomiRedmi9:
This will improve android performance?

JosJuice:
No, but it fixes a bug.

degasus:
Honestly `and at least one guest register containing an immediate` is a bug in the old implementation. The issue only happens when all host registers are allocated, so you have host register you can flush first and immediates later. Not the less, it might be good to have a tmp reg in the flush call, however I'm not so sure about having one in the flush_all one.

JosJuice:
> The issue only happens when all host registers are allocated, so you have host register you can flush first and immediates later.

But flushing registers with `FlushMode::MaintainState` does not actually free up the corresponding host register for use, so changing the order in which we flush does nothing to fix the bug.

> Not the less, it might be good to have a tmp reg in the flush call, however I'm not so sure about having one in the flush_all one.

Yes, in the `FlushMode::All` case there's no real benefit to it.

PR #9893, opened 2021-07-13
MenuBar: Check for updates on the current update track

OatmealDome:
Going to Help -> Check for Updates always checks for updates on the ``dev`` update track, regardless of what track the user has chosen. However, the error message for when no update is detected states that it checked on "this update track", presumably referring to the user's current one.

This seems like a bug? Maybe it's debug code that was unintentionally left in?

### DIFF
--- a/Source/Core/DolphinQt/MenuBar.cpp
+++ b/Source/Core/DolphinQt/MenuBar.cpp
@@ -554,11 +554,6 @@ void MenuBar::AddOptionsMenu()
 
 void MenuBar::InstallUpdateManually()
 {
-  auto& track = SConfig::GetInstance().m_auto_update_track;
-  auto previous_value = track;
-
-  track = "dev";
-
   auto* updater = new Updater(this->parentWidget());
 
   if (!updater->CheckForUpdate())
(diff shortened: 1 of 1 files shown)


Dentomologist:
My assumption was that this is intentional. If we ask a user to update to the latest version for a fix or testing purposes it's easier to just have them select "Check for Updates" rather than walking them through changing the update track to dev and then back afterward.

The message about "this update track" is unclear; we should change that to refer to the actually checked track (Hardcoding it to dev would work since that's the only track that can spawn that message, but would be fragile to future changes).

There are a number of things I'd like to rework/refactor about the updater; I'm in the middle of adding tests for it and then it'll be safer to make changes and refactor it.

OatmealDome:
Hmm, this behaviour is a bit confusing. I agree that the message should be changed -- I'll open a separate PR later for that. (It appears someone suggested it in the prior PR but it was never actually addressed?)

Perhaps this menu bar command can be changed to "Update to Development Build" (or something like that) and then a new "Check for Updates" can be added that checks for updates on the user's track?

PR #9894, opened 2021-07-13
GameINI: Set safe texture cache in WarioWare: Smooth Moves for Writer's Block and Brain Age microgames.

AdmiralCurtiss:
The line you're drawing in those games doesn't properly show up otherwise.

### DIFF
--- a/Data/Sys/GameSettings/ROD.ini
+++ b/Data/Sys/GameSettings/ROD.ini
@@ -16,3 +16,4 @@ EFBEmulateFormatChanges = True
 
 [Video_Settings]
 SuggestedAspectRatio = 2
+SafeTextureCacheColorSamples = 0


Pokechu22:
The wiki also [claims](https://wiki.dolphin-emu.org/index.php?title=WarioWare:_Smooth_Moves&oldid=177349#FMV_Graphics_Errors) that some of the FMVs in the game have graphical issues that are fixed by safe texture cache.  Can you confirm this?

JMC47:
Safe should fix anything that medium does.

AdmiralCurtiss:
> 
> 
> The wiki also [claims](https://wiki.dolphin-emu.org/index.php?title=WarioWare:_Smooth_Moves&oldid=177349#FMV_Graphics_Errors) that some of the FMVs in the game have graphical issues that are fixed by **medium** texture cache (though safe should work with them too). Can you confirm this?

I can't see any issues with the FMVs regardless of texture cache setting. If something is broken there then it's either non-obvious or I just happened to pick scenes that work fine.

PR #9895, opened 2021-07-13
GCAdapter: Simulate Dance Mat adapter

Dentomologist:
Experimental support for dance mats using GameCube Adapter Passthrough. I have neither an adaptor nor the dance mat so this is mostly untested.


JMC47:
Note that Mayflash is incorrect - this would work for any Wii U/Switch Adapter that Dolphin supports, including the official Nintendo one.

Dentomologist:
Changed the checkboxes to a combo box, and more importantly made the simulated adapters call the right version of MapPadStatus (which shuffles some pad bits around).

Dentomologist:
Apparently this already works now.

PR #9896, opened 2021-07-13
android phone with a Razer Kishi contoller

takeshineale128:
My android phone (Xiaomi Mi11 - M2011K2G) with a Razer Kishi contoller and when I map the contol stick and on Pokemon XD when I move the guy down it goes up


leoetlino:
This is not a bug tracker.

PR #9898, opened 2021-07-16
NetPlayIndex: Use scm_rev_str instead of scm_desc_str for version check

Techjar:
Some build systems (i.e. Nix) make it architecturally impossible to acquire the correct scm_desc_str that we normally use for builds,
making it basically impossible for users of those to use the NetPlay browser, unless they manually override the version string. Not relying on this string in the first place for any actual functionality beyond display to the user would be beneficial, and using the scm_rev_str instead matches the check on actually joining a session anyways, so it makes sense to do it this way.

This requires a backend change to the NetPlay lobby server, found at


Techjar:
Pinging @spycrab to review this since he manages the NetPlay lobby backend.

leoetlino:
Marking this as draft until the backend is updated.

Zopolis4:
Does this PR and the backend PR need to be merged simultaneously?

PR #9900, opened 2021-07-16
Updater: Display an appropriate error message on failure

OatmealDome:
`AutoUpdateChecker::CheckForUpdate()` is binary in its return result - either there is an update (true), or Dolphin is up to date or there's an error (false). This means that even if an error occurred while checking for an update, DolphinQt will always say that "you are running the latest version".

(I think this is one of the contributing factors to how macOS universal updates were broken and nobody ever noticed.)

This PR aims to solve this by adding an error callback where DolphinQt can display appropriate error messages. Here they are:

<details>
<summary>Screenshots</summary>

### Up to Date

image

### HTTP Request Failed

image

### JSON Response Invalid

image

### Updater Process Failed to Start

image

</details>

### DIFF
--- a/Source/Core/DolphinQt/Main.cpp
+++ b/Source/Core/DolphinQt/Main.cpp
@@ -284,7 +284,7 @@ int WINAPI wWinMain(HINSTANCE hInstance, HINSTANCE hPrevInstance, PWSTR pCmdLine
 
     if (!Settings::Instance().IsBatchModeEnabled())
     {
-      auto* updater = new Updater(&win);
+      auto* updater = new Updater(&win, Updater::ShouldSilentlyFail::Yes);
       updater->start();
     }
 
--- a/Source/Core/DolphinQt/Main.cpp
+++ b/Source/Core/DolphinQt/Main.cpp
@@ -284,7 +284,7 @@ int WINAPI wWinMain(HINSTANCE hInstance, HINSTANCE hPrevInstance, PWSTR pCmdLine
 
     if (!Settings::Instance().IsBatchModeEnabled())
     {
-      auto* updater = new Updater(&win, Updater::ShouldSilentlyFail::Yes);
+      auto* const updater = new Updater(&win, Updater::ShouldSilentlyFail::Yes);
       updater->start();
     }
 
--- a/Source/Core/DolphinQt/Updater.cpp
+++ b/Source/Core/DolphinQt/Updater.cpp
@@ -121,8 +121,8 @@ void Updater::OnUpdateAvailable(const NewVersionInformation& info) const
   if (choice && *choice == QDialog::Accepted)
   {
     bool updater_process_started =
-        TriggerUpdate(info, later ? AutoUpdateChecker::RestartMode::NO_RESTART_AFTER_UPDATE :
-                                    AutoUpdateChecker::RestartMode::RESTART_AFTER_UPDATE);
+        TriggerUpdate(info, later ? AutoUpdateChecker::RestartMode::NoRestartAfterUpdate :
+                                    AutoUpdateChecker::RestartMode::RestartAfterUpdate);
 
     if (!later && updater_process_started)
     {
--- a/Source/Core/DolphinQt/Updater.cpp
+++ b/Source/Core/DolphinQt/Updater.cpp
@@ -19,7 +19,7 @@
 
 // Refer to docs/autoupdate_overview.md for a detailed overview of the autoupdate process
 
-Updater::Updater(QWidget* parent, const ShouldSilentlyFail silently_fail)
+Updater::Updater(QWidget* const parent, const ShouldSilentlyFail silently_fail)
     : m_parent(parent), m_silently_fail(silently_fail)
 {
   connect(this, &QThread::finished, this, &QObject::deleteLater);
@@ -65,7 +65,7 @@ void Updater::OnErrorOccurred(const CheckError error) const
   }
 }
 
-void Updater::OnUpdateAvailable(const NewVersionInformation& info)
+void Updater::OnUpdateAvailable(const NewVersionInformation& info) const
 {
   bool later = false;
 
--- a/Source/Core/DolphinQt/Updater.cpp
+++ b/Source/Core/DolphinQt/Updater.cpp
@@ -13,33 +13,61 @@
 
 #include "Common/Version.h"
 
+#include "DolphinQt/QtUtils/ModalMessageBox.h"
 #include "DolphinQt/QtUtils/RunOnObject.h"
 #include "DolphinQt/Settings.h"
 
 // Refer to docs/autoupdate_overview.md for a detailed overview of the autoupdate process
 
-Updater::Updater(QWidget* parent) : m_parent(parent)
+Updater::Updater(QWidget* parent, const ShouldSilentlyFail silently_fail)
+    : m_parent(parent), m_silently_fail(silently_fail)
 {
   connect(this, &QThread::finished, this, &QObject::deleteLater);
 }
 
 void Updater::run()
 {
-  AutoUpdateChecker::CheckForUpdate();
+  CheckForUpdate();
 }
 
-bool Updater::CheckForUpdate()
+void Updater::OnErrorOccurred(const CheckError error) const
 {
-  m_update_available = false;
-  AutoUpdateChecker::CheckForUpdate();
+  if (m_silently_fail == ShouldSilentlyFail::Yes)
+  {
+    return;
+  }
 
-  return m_update_available;
+  switch (error)
+  {
+  case CheckError::AlreadyUpToDate:
+    ModalMessageBox::information(
+        m_parent, tr("Update"),
+        tr("You are running the latest version available on this update track."));
+    break;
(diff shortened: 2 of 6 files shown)


JosJuice:
@dolphin-emu-bot rebuild

Source/Core/DolphinQt/Updater.h
```diff
@@ -15,13 +15,13 @@ class Updater : public QThread, public AutoUpdateChecker
 {
   Q_OBJECT
 public:
-  explicit Updater(QWidget* parent);
+  explicit Updater(QWidget* parent, bool silently_fail);
```
leoetlino, line 18: Please use an enum class rather than a boolean to improve readability at call sites. Something like `enum class ShouldSilentlyFail { Yes, No }` would be a lot clearer than seeing a random false or a true.

Source/Core/DolphinQt/Updater.cpp
```diff
@@ -13,33 +13,58 @@
...
+    break;
+  case CheckError::RequestFailed:
+    ModalMessageBox::critical(m_parent, tr("Update failed"),
+                              tr("Could not download update information. "
+                                 "Please check your Internet connection and try again."));
+    break;
+  case CheckError::InvalidJson:
+    ModalMessageBox::critical(m_parent, tr("Update failed"),
+                              tr("The server returned an invalid response. "
+                                 "Please check your Internet connection and try again."));
```
BhaaLseN, line 55: I wonder if "Check your Internet connection" is a useful suggestion for "got an invalid response"...

I could only imagine a proxy blocking the request (where that suggestions _might_ remotely apply), or otherwise actual infrastructure issues (that the user probably can't influence anyways).

How about "Please try again in a few minutes" or something like that instead?

Source/Core/DolphinQt/Updater.h
BhaaLseN: Does this even have to be `explicit` when it has more than one parameter?

Source/Core/UICommon/AutoUpdate.h
```diff
@@ -18,6 +18,19 @@ class AutoUpdateChecker
...
+  enum class CheckError
+  {
+    // The HTTP request failed.
+    RequestFailed,
+
+    // The JSON received from the server is not valid.
+    InvalidJson,
+
+    // We're already up to date.
+    AlreadyUpToDate
```
BhaaLseN, line 31: We probably want the comma here even on the last member, right?

Source/Core/DolphinQt/Updater.cpp
OatmealDome: I went with "Please try again later". That sounds more natural to me, but I'm fine with "in a few minutes" too.

OatmealDome:
OK, here's what's changed...

- The code comments have been addressed.
- The suggestion in the HTTP request failure error message has been changed to "Please try again later."
- I added another error which appears if the updater process fails to launch in `AutoUpdateChecker::TriggerUpdate()`.
- Screenshots in the OP have been updated to reflect the latest changes.

Source/Core/UICommon/AutoUpdate.h
```diff
@@ -39,8 +61,9 @@ class AutoUpdateChecker
     NO_RESTART_AFTER_UPDATE = 0,
     RESTART_AFTER_UPDATE,
   };
-  void TriggerUpdate(const NewVersionInformation& info, RestartMode restart_mode);
+  bool TriggerUpdate(const NewVersionInformation& info, RestartMode restart_mode);
 
 protected:
   virtual void OnUpdateAvailable(const NewVersionInformation& info) = 0;
+  virtual void OnErrorOccurred(CheckError error) = 0;
```
Dentomologist, line 68: These functions can all be const (along with the overridden ones in Updater).

Dentomologist (changes requested):

Source/Core/UICommon/AutoUpdate.h
```diff
@@ -18,6 +18,28 @@ class AutoUpdateChecker
...
+    RequestFailed,
+
+    // The JSON received from the server is not valid.
+    InvalidJson,
+
+    // We're already up to date.
+    AlreadyUpToDate,
+
+    // Couldn't launch updater.
+    FailedToLaunchUpdater,
```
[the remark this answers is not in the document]
leoetlino: IMO InvalidJsonInServerResponse is more readable (considering PascalCase is used and caps are used to separate words). We don't have particularly consistent rules on this though...

Source/Core/DolphinQt/Main.cpp
```diff
@@ -284,7 +284,7 @@ int WINAPI wWinMain(HINSTANCE hInstance, HINSTANCE hPrevInstance, PWSTR pCmdLine
 
     if (!Settings::Instance().IsBatchModeEnabled())
     {
-      auto* updater = new Updater(&win);
+      auto* updater = new Updater(&win, Updater::ShouldSilentlyFail::Yes);
```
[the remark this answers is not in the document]
OatmealDome: If I make this `const auto*`, the compiler complains `QThread::start` isn't `const`. (Maybe there's something I'm missing?)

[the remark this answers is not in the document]
Dentomologist: Putting const on the left makes it a non-const pointer to a const variable, which isn't what we want.  Do "auto* const updater" instead to make it a const pointer to a non-const variable.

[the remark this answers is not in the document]
OatmealDome: Thanks, I was unaware this was a thing you could do. (Still new to C++...)

Source/Core/UICommon/AutoUpdate.h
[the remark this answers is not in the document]
OatmealDome: I changed the enums and removed the comments to match your suggestions. For JSON vs Json, I went with Json since I agreed that it looked more readable. (I couldn't find an example elsewhere where an acronym is used in an enum, but I'm probably blind.)

I also changed RestartMode lower down to PascalCase to make the styling consistent within the file. There seems to be a bunch of other enums where ``WHATEVER_THIS_IS_CALLED`` styling is used, but PascalCase seems to be prevalent?

Source/Core/UICommon/AutoUpdate.h
```diff
@@ -54,9 +54,9 @@ class AutoUpdateChecker
     NO_RESTART_AFTER_UPDATE = 0,
     RESTART_AFTER_UPDATE,
   };
-  bool TriggerUpdate(const NewVersionInformation& info, RestartMode restart_mode) const;
+  bool TriggerUpdate(const NewVersionInformation& info, const RestartMode restart_mode) const;
```
leoetlino, line 57: The const qualifier for restart_mode is kind of useless *in the declaration* and should be removed.

Source/Core/UICommon/AutoUpdate.h
```diff
@@ -54,9 +54,9 @@ class AutoUpdateChecker
...
 
 protected:
-  virtual void OnUpdateAvailable(const NewVersionInformation& info) = 0;
+  virtual void OnUpdateAvailable(const NewVersionInformation& info) const = 0;
   virtual void OnErrorOccurred(const CheckError error) const = 0;
```
leoetlino, line 61: ditto for error

leoetlino (changes requested):

Source/Core/UICommon/AutoUpdate.h
Dentomologist: I asked for this one. I presume your reasoning is that it's passed by value so the caller doesn't need to worry about the const? 

That makes sense to me.

Source/Core/DolphinQt/Updater.cpp
```diff
@@ -13,33 +13,61 @@
...
 #include "Common/Version.h"
 
+#include "DolphinQt/QtUtils/ModalMessageBox.h"
 #include "DolphinQt/QtUtils/RunOnObject.h"
 #include "DolphinQt/Settings.h"
 
 // Refer to docs/autoupdate_overview.md for a detailed overview of the autoupdate process
 
-Updater::Updater(QWidget* parent) : m_parent(parent)
+Updater::Updater(QWidget* parent, ShouldSilentlyFail silently_fail)
```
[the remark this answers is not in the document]
Dentomologist: Partial retraction: Remove the const from silently_fail.

Source/Core/UICommon/AutoUpdate.h
leoetlino: Yeah. const prevents `error` from being mutated, but what's going to mutate its value in the declaration? Nothing, so it's just noise in this case.

Dentomologist (commented):
Tested on Windows and macOS and things worked as designed.

OatmealDome:
Superseded by #11018.

Source/Core/UICommon/AutoUpdate.h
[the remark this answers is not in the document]
Minty-Meeo: SCREAMING_SNAKE_CASE, I think

PR #9901, opened 2021-07-17
Android Adding Guitar  Drum Overlay

blakbin:
Drum has two pedal but sampe button type, since i dont want to add extra key & shared preference using buttonID to maintain position, so the positioning is the same for both. default on pedal 2 is false, so enable from toggle button & move position manually.
drum landscape
drum portrait (Small)

Guitar overlay has two different size in Portrait & Landscape mode, so im modified inputoverlaybutton before draw.
guitar landscape
Guitar portrait (Small)

for both input im adding Action Move so when touch input slide enter bound button get pressed & outside bound button will released

Note : for MIUI users 
in some case guitar has three button must be press, in MIUI, if there are 3 pointers then other Actions will be cancelled, this is a shortcut set by defaut to take a screenshoot, that can make the three pressed buttons to not be released. 
To fix this, please disable the navigation gesture menu in Settings > Additional Settings > Button Shortcut, clean the shortcut for Take Screenshoot & set Automatically Disable Navigation Buttons for Dolphin Emulator


JosJuice:
@dolphin-emu-bot rebuild

JosJuice:
@dolphin-emu-bot rebuild

JosJuice:
@dolphin-emu-bot rebuild

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/overlay/InputOverlay.java
```diff
@@ -218,6 +220,39 @@ public boolean onTouch(View v, MotionEvent event)
...
+              {
+                button.setPressedState(true);
+                button.setTrackId(event.getPointerId(pointerIndex));
+                pressed = true;
+                NativeLibrary.onGamePadEvent(NativeLibrary.TouchScreenDevice, button.getId(),
+                        ButtonState.PRESSED);
+              }
+              else
+              {
+                if (button.getTrackId() == event.getPointerId(pointerIndex))
```
BhaaLseN, line 245: Any particular reason this isn't a combined `else if`?

BhaaLseN (commented):
Feel free to combine those commits into one, there is very little reason to keep them separate like that (especially when the commit message doesn't hold any additional detail specific to the commit)

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/overlay/InputOverlay.java
blakbin: no i think it just handle two condition for each touch pointer

Rumi-Larry:
The vertical orientation should be preferred, otherwise uñyour fingers cover the screen and you may miss notes.

blakbin:
> The vertical orientation should be preferred, otherwise one's fingers will cover the screen and you may miss notes.

when I made this I thought they should separate the layout, in Horizontal mode we can hold the phone like a guitar, 4 fingers on the frets and 1 finger for strumming. In Vertical mode I think it's confusing for me to place a layout like Horizontal mode with the game image position reversed. therefore the portrait position is intended to allow using two fingers, and when we press the button in the combined area will make both buttons pressed.

PR #9906, opened 2021-07-18
remove msvc warnings which no longer fire, and DSP::Profiler

shuffle2:
The profiler seems like it hasn't been used for a long time, and is trivial to recreate/restore if needed.


Source/Core/Core/HW/WiimoteEmu/EmuSubroutines.cpp
```diff
@@ -53,7 +53,7 @@ void Wiimote::InvokeHandler(H&& handler, const WiimoteCommon::OutputReportGeneri
     return;
   }
 
-  (this->*handler)(Common::BitCastPtr<T>(rpt.data));
+  (this->*handler)(Common::BitCastPtr<T>(&rpt.data[0]));
```
Tilka, line 56: why this?

shuffle2: personally I find it more clear, just a preference I guess.

PR #9907, opened 2021-07-18
make FPSCR.RN an enum

shuffle2:
This just addresses a comment I came across when grepping for `_MSC_VER`.
Compiles without warnings on msvc, unittests pass, and f-zero seems to run fine. The url in the comment is dead but I assume the issue was something relating to sign extension based on commit history and https://bugs.webkit.org/show_bug.cgi?id=134237

If this change isn't wanted anymore maybe we can just remove the comment instead?


Tilka:
As a follow-up, SetRoundMode() and SetSIMDMode() can be merged since we don't care about x87 rounding anymore.

PR #9908, opened 2021-07-18
merge SetRoundMode into SetSIMDMode

shuffle2:


Source/Core/Common/x64FPURoundMode.cpp
```diff
@@ -14,15 +14,12 @@ namespace FPURoundMode
...
 static const u32 default_sse_state = _mm_getcsr();
 
-void SetRoundMode(RoundMode mode)
+void SetSIMDMode(RoundMode rounding_mode, bool non_ieee_mode)
 {
   // Convert PowerPC to native rounding mode.
   static const int rounding_mode_lut[] = {FE_TONEAREST, FE_TOWARDZERO, FE_UPWARD, FE_DOWNWARD};
-  fesetround(rounding_mode_lut[mode]);
-}
+  fesetround(rounding_mode_lut[rounding_mode]);
```
Tilka, line 21: Sorry, what I meant is that you can get rid of fesetround() since it just sets x87 rounding and SSE rounding, but we only need the latter and already do it manually below.

PR #9912, opened 2021-07-18
Fix MEM2 in the memory viewer in physical mode

Pokechu22:
Previously, looking at MEM2 in the physical mode (at 10000000) would show all zeros.  However, the search feature would match the actual data.  This happened because ReadU8 and WriteU8 did not account for the accessor's base offset (though search did).  I'm somewhat surprised it didn't just crash in this case.

Note that effective mode (where MEM2 is at 90000000) worked properly before in most cases.  The one exception I know of is with the FIFO player, where nothing seems to be mapped at 90000000; this is probably because fifologs don't store BATs.  However, graphics-related logic all operates on physical addresses, so I'm not going to worry about that for now.

This change should have no impact on emulation, only on the memory viewer.

### DIFF
--- a/Source/Core/Core/HW/AddressSpace.cpp
+++ b/Source/Core/Core/HW/AddressSpace.cpp
@@ -246,22 +246,22 @@ struct CompositeAddressSpaceAccessors : Accessors
 
   u8 ReadU8(u32 address) const override
   {
-    auto it = FindAppropriateAccessor(address);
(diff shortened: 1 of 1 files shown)


Pokechu22:
Correction: the previous behavior won't always result in seeing all zeros, but you still won't see actual contents of MEM2 (just whatever happens to be 0x10000000 (268,435,456) bytes after the start of MEM2 in the host system).

PR #9914, opened 2021-07-18
Hackfix for clipping in Gormiti

Pokechu22:
This should fix [bug 12436](https://bugs.dolphin-emu.org/issues/12436).

I don't really know what I'm doing here, to be honest; I don't fully understand the clipping code.

What I do know is that the game deliberately draws a black rectangle over the entire screen (the first 3 values are the x/y/z coordinates, with the hex values first and then the decoded float values in parentheses, and the last value is the color, where alpha is 1):

```
Primitive GX_DRAW_TRIANGLE_STRIP (3) VAT 0

00000000 (0) 00000000 (0) 3f800000 (1)  00000001  
00000000 (0) 43f00000 (480) 3f800000 (1)  00000001  
44200000 (640) 00000000 (0) 3f800000 (1)  00000001  
44200000 (640) 43f00000 (480) 3f800000 (1)  00000001  
```

It also configures the alpha test and z-test to pass for these.  I can only assume that this is deliberate anti-emulation code, because there doesn't seem to be any other purpose to this rectangle.

Testing using the hardware fifoplayer showed that this rectangle did not render; I assume this is because it is passing the far plane (but I'm not 100% sure whether it's the far or near plane, as I don't have a perfect understanding of this part of rendering.  The z coordinate is transformed to -1 and then projected position's z-coordinate is 0, so I think it it's the far plane, though.)  My guess is that this is supposed to be a `>= 0` comparison for clipping instead of a `> 0` comparison, which I did with the software renderer, but I did some hacky stuff in the hardware renderers instead.

This PR also applies the sonic epsilon hack for `host_config.backend_depth_clamp = false` to UberShaders (it was implemented for specialized shaders in #9591).  I haven't tested anything for `backend_depth_clamp = false`.

### DIFF
--- a/Source/Core/VideoBackends/Software/Clipper.cpp
+++ b/Source/Core/VideoBackends/Software/Clipper.cpp
@@ -90,7 +90,7 @@ static inline int CalcClipMask(const OutputVertexData* v)
   if (pos.y + pos.w < 0)
     cmask |= CLIP_NEG_Y_BIT;
 
-  if (pos.w * pos.z > 0)
+  if (pos.w * pos.z >= 0)
     cmask |= CLIP_POS_Z_BIT;
 
   if (pos.z + pos.w < 0)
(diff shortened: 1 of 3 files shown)


Rumi-Larry:
Wouldn't it be better to patch the game to remove the rectangle?

Pokechu22:
That would probably be better than the current fix (given the side effects of breaking a bunch of games), but if a more accurate fix is found that'd be preferable to a game patch.

Rumi-Larry:
What would be the motivation of the devs to implement anti-emulation? Certainly it would be different than anti-piracy, right?

Pokechu22:
Closing this as this is not the correct fix (and breaks other games).  I will probably investigate this further later on and come up with a better fix (once I understand the problem better).

PR #9920, opened 2021-07-20
GameINI: Fix PoP:TT Shadows

JMC47:
At extreme angles, there is severe shadow zfighting with fast depth disabled.  Enabling it causes it work on all backends without issue.  Verified in OpenGL, D3D11, D3D12, and Vulkan.  I assume that this is the correct setting name, as I copied it out of the GFX config file.

### DIFF
--- a/Data/Sys/GameSettings/GKM.ini
+++ b/Data/Sys/GameSettings/GKM.ini
@@ -14,4 +14,4 @@
 
 [Video_Settings]
 SafeTextureCacheColorSamples = 512
-
+FastDepthCalc = True


Data/Sys/GameSettings/GKM.ini
```diff
@@ -15,3 +15,6 @@
 [Video_Settings]
 SafeTextureCacheColorSamples = 512
 
+[Video_Hacks]
```
JosJuice, line 18:
```suggestion
[Video_Settings]
```

Rumi-Larry: 'Video settings' is already written 3 lines above, so it would be repeated.

PR #9921, opened 2021-07-20
Software: Handle texture wrapping more accurately

Pokechu22:
See https://bugs.dolphin-emu.org/issues/12408.  When texture wrapping is enabled, the texture size should be a power of two per [libogc](https://github.com/devkitPro/libogc/blob/bd24a9b3f59502f9b30d6bac0ae35fc485045f78/gc/ogc/gx.h#L4242-L4243); yagcd mentions (near [bp register](http://hitmen.c02.at/files/yagcd/yagcd/chap5.html#sec5.11.1) 0x80) that repeat and mirror "[require] the texture size to be a power of two. (wrapping is implemented by a logical AND (SIZE-1))".

Although I initially targeted this at the the non-power-of-2 issue, it also impacts the software renderer when wrapping textures in negative coordinates (https://dolp.in/i11834/6 is an example).  This is probably due to the modulus operator `%` returning negative numbers for negative inputs (e.g. `-1 % 4` is `-1`), while using `&` gives positive values (e.g. `-1 & (4-1)` is `3`).

I haven't implemented the non-power-of-2 stuff in the hardware backends, as the graphics APIs themselves handle texture wrapping in the sampling process.  Dolphin could handle it manually in the pixel shader, but I suspect there would be a performance cost to it for something that basically no games do (though I haven't tried it and the performance impact may be minimal or it might even be faster due to allowing silly behavior for non-power-of-2 sizes).  One other possible complication to implementing it on the hardware backends is custom textures, as those might not be a power-of-2 size and using a bitwise and might also get rid of fractional parts of coordinates which would break higher resolution textures (though these too may not be issues).

I did also find that setting the wrap mode to 3 (which is invalid) seems to behave the same as setting it to 0 (clamp).  Dolphin previously treated it the same as 1 (repeat) instead; I've fixed that for both the hardware and software backends.

### DIFF
--- a/Source/Core/VideoBackends/Software/TextureSampler.cpp
+++ b/Source/Core/VideoBackends/Software/TextureSampler.cpp
@@ -18,29 +18,36 @@
 
 namespace TextureSampler
 {
-static inline void WrapCoord(int* coordp, WrapMode wrapMode, int imageSize)
+static inline void WrapCoord(int* coordp, WrapMode wrap_mode, int image_size)
 {
   int coord = *coordp;
-  switch (wrapMode)
+  switch (wrap_mode)
   {
   case WrapMode::Clamp:
-    coord = (coord > imageSize) ? imageSize : (coord < 0) ? 0 : coord;
+    coord = std::clamp(coord, 0, image_size - 1);
     break;
   case WrapMode::Repeat:
-    coord = coord % (imageSize + 1);
-    coord = (coord < 0) ? imageSize + coord : coord;
+    // Per YAGCD's info on TX_SETMODE1_I0 (et al.), mirror "requires the texture size to be a power
+    // of two. (wrapping is implemented by a logical AND (SIZE-1))".  So though this doesn't wrap
+    // nicely for non-power-of-2 sizes, that's how hardware does it.
+    coord = coord & (image_size - 1);
     break;
   case WrapMode::Mirror:
   {
-    const int sizePlus1 = imageSize + 1;
-    const int div = coord / sizePlus1;
-    coord = coord - (div * sizePlus1);
-    coord = (coord < 0) ? -coord : coord;
-    coord = (div & 1) ? imageSize - coord : coord;
+    // YAGCD doesn't mention this, but this seems to be the check used to implement mirroring.
+    // With power-of-2 sizes, this correctly checks if it's an even-numbered repeat or an
+    // odd-numbered one, and thus can decide whether to reflect.  It fails in unusual ways
+    // with non-power-of-2 sizes, but seems to match what happens on actual hardware.
+    if ((coord & image_size) != 0)
+      coord = ~coord;
+    coord = coord & (image_size - 1);
     break;
   }
   default:
-    PanicAlertFmt("Invalid wrap mode: {}", wrapMode);
+    // Hardware testing indicates that wrap_mode set to 3 behaves the same as clamp.
+    PanicAlertFmt("Invalid wrap mode: {}", wrap_mode);
+    coord = std::clamp(coord, 0, image_size - 1);
+    break;
   }
   *coordp = coord;
 }
@@ -131,8 +138,8 @@ void SampleMip(s32 s, s32 t, s32 mip, bool linear, u8 texmap, u8* sample)
     imageSrc = Memory::GetPointer(imageBase);
   }
 
-  int imageWidth = ti0.width;
-  int imageHeight = ti0.height;
+  int image_width_minus_1 = ti0.width;
+  int image_height_minus_1 = ti0.height;
 
   const int tlutAddress = texTlut.tmem_offset << 9;
   const u8* tlut = &texMem[tlutAddress];
@@ -141,15 +148,15 @@ void SampleMip(s32 s, s32 t, s32 mip, bool linear, u8 texmap, u8* sample)
   // move texture pointer to mip location
   if (mip)
   {
-    int mipWidth = imageWidth + 1;
-    int mipHeight = imageHeight + 1;
+    int mipWidth = image_width_minus_1 + 1;
+    int mipHeight = image_height_minus_1 + 1;
 
     const int fmtWidth = TexDecoder_GetBlockWidthInTexels(texfmt);
     const int fmtHeight = TexDecoder_GetBlockHeightInTexels(texfmt);
     const int fmtDepth = TexDecoder_GetTexelSizeInNibbles(texfmt);
 
-    imageWidth >>= mip;
-    imageHeight >>= mip;
+    image_width_minus_1 >>= mip;
+    image_height_minus_1 >>= mip;
     s >>= mip;
     t >>= mip;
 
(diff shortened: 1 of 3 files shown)


Pokechu22:
Old:

New:


Note that the new version also looks slightly more pixelated; this is because the texture is 512 by 191, which means it gets masked vertically by `(191 - 1) = 190 = 0b10111110`, i.e. it loses the bottom bit in addition to a bit in the middle.  This seems to be consistent with [the screenshot on hardware](https://bugs.dolphin-emu.org/attachments/8198/IMG_20210206_180615.jpg), though it's hard to be 100% sure.

Source/Core/VideoBackends/Software/TextureSampler.cpp
```diff
@@ -18,29 +18,32 @@
...
+    coord = std::clamp(coord, 0, image_size - 1);
     break;
   case WrapMode::Repeat:
-    coord = coord % (imageSize + 1);
-    coord = (coord < 0) ? imageSize + coord : coord;
+    // Per YAGCD's info on TX_SETMODE1_I0 (et al.), mirror "requires the texture size to be a power
+    // of two. (wrapping is implemented by a logical AND (SIZE-1))".  So though this doesn't wrap
+    // nicely for non-power-of-2 sizes, that's how hardware does it.
+    coord = coord & (image_size - 1);
+    coord = (coord < 0) ? image_size - 1 + coord : coord;
```
Pokechu22, line 34: The `(coord < 0)` case is no longer needed, as the bitwise and will always result in a positive value.

I suspect that this is also responsible for all of the rendering differences I mentioned above (and initially incorrectly attributed to size + 1), which should also bring the software renderer in line with the hardware renderers.

Pokechu22:
Hardware test results:

These were done by editing object 306 in the hardware fifo player, editing `BPMEM_TX_SETMODE0` at offset 00119e80.  The data there by default is `6180018195`, where the 5 at the end indicates that both U and V wrapping is set to `1` (repeat).  Since only the texture's width is a power of 2, we can set U to 0 for clamp without any differences.  Thus, I tested the v wrapping with `6180018190` (Clamp), `6180018194` (Repeat), `6180018198` (Mirror), and `618001819c` (Invalid).  I also edited the texture's height in `BPMEM_TX_SETIMAGE0` at offset 00119e8a.

<details><summary>Height = 191</summary>

Leave 00119e8a unchanged as `618852f9ff`.
|Mode       |Hardware                                                                                                                                   |Dolphin                                                                                                                               |
|-----------|-------------------------------------------------------------------------------------------------------------------------------------------|--------------------------------------------------------------------------------------------------------------------------------------|
|Clamp (0)  |Screenshot 2021-07-20 10-42-32_0 |OHBCHB_2021-07-20_12-42-28_0|
|Repeat (1) |Screenshot 2021-07-20 10-43-08_4 |OHBCHB_2021-07-20_12-42-12_4|
|Mirror (2) |Screenshot 2021-07-20 10-43-56_8 |OHBCHB_2021-07-20_12-42-45_8|
|Invalid (3)|Screenshot 2021-07-20 10-44-50_c |OHBCHB_2021-07-20_12-43-03_c|
</details>
<details><summary>Height = 192</summary>

Change 00119e8a to `618852fdff`.
|Mode       |Hardware                                                                                                                                   |Dolphin                                                                                                                               |
|-----------|-------------------------------------------------------------------------------------------------------------------------------------------|--------------------------------------------------------------------------------------------------------------------------------------|
|Clamp (0)  |Screenshot 2021-07-20 10-50-24_r0 |OHBCHB_2021-07-20_12-43-30_r0|
|Repeat (1) |Screenshot 2021-07-20 10-50-51_r4 |OHBCHB_2021-07-20_12-43-38_r4|
|Mirror (2) |Screenshot 2021-07-20 10-51-27_r8 |OHBCHB_2021-07-20_12-43-47_r8|
|Invalid (3)|Screenshot 2021-07-20 10-51-53_rc |OHBCHB_2021-07-20_12-43-58_rc|
</details>
<details><summary>Height = 96</summary>

Change 00119e8a to `61885181ff`.
|Mode       |Hardware                                                                                                                                   |Dolphin                                                                                                                               |
|-----------|-------------------------------------------------------------------------------------------------------------------------------------------|--------------------------------------------------------------------------------------------------------------------------------------|
|Clamp (0)  |Screenshot 2021-07-20 12-26-55_s0 |OHBCHB_2021-07-20_12-44-42_s0|
|Repeat (1) |Screenshot 2021-07-20 12-27-19_s4*|OHBCHB_2021-07-20_12-44-49_s4|
|Mirror (2) |Screenshot 2021-07-20 12-28-12_s8 |OHBCHB_2021-07-20_12-44-59_s8|
|Invalid (3)|Screenshot 2021-07-20 12-28-30_sc |OHBCHB_2021-07-20_12-45-09_sc|

*Repeat had some unusual behavior where one row changed each time it drew.  [Here's another image of it](https://user-images.githubusercontent.com/8334194/126389228-aa3fdb9a-5bba-4bf5-83dc-12dc0d2ab967.png).  The actual texture being drawn seems the same, apart from the places where no texture is drawn at all, so I think the wrapping logic is correct.
</details>

It's important to note that the original issue I was trying to fix was from a mod that unintentionally had texture wrapping enabled when it should have been disabled; the buggy-looking wrapped appearance is what they should have gotten in Dolphin (and got when they later tested it on actual hardware).

Pokechu22:
Notable differences (apart from just texture shifting):

* `jd2-fmv` ([version diff](https://fifo.ci/compare/6476571-6471861/) / [prev SW-OGL diff](https://fifo.ci/compare/6471861-6471342/) / [new SW-OGL diff](https://fifo.ci/compare/6476571-6476039/)): no longer a garbled mess; still has some artifacts (to the same degree as OGL, but in different places)
* `nsmbw-coins` ([version diff](https://fifo.ci/compare/6476620-6471885/) / [prev SW-OGL diff](https://fifo.ci/compare/6471885-6471360/) / [new SW-OGL diff](https://fifo.ci/compare/6476620-6475954/)): the background and brown platforms are slightly darker.  (There is no difference with the coins, which I think already render correctly?)
* `rs3-bumpmapping` ([version diff](https://fifo.ci/compare/6476589-6471858/) / [prev SW-OGL diff](https://fifo.ci/compare/6471858-6471330/) / [new SW-OGL diff](https://fifo.ci/compare/6476589-6475978/)): the snow is darker in some areas
* `tp-skin` ([version diff](https://fifo.ci/compare/6476600-6471892/) / [prev SW-OGL diff](https://fifo.ci/compare/6471892-6471321/) / [new SW-OGL diff](https://fifo.ci/compare/6476600-6476057/)): a small white line on the head is gone

Pokechu22:
The other eyelid also has some oddness going on with it:

Before this change:


After this change:


Both of these are OGL with specialized shaders.  This eyelid isn't visible normally, and there isn't any way to move the camera with the hardware fifoplayer, so it may just be messed up like this normally and you can't see it.

phire (approved):

phire:
It's important to note that Mirrored with height=96 produces a quite different result on hardware.

Pokechu22:
Here are some additional images from the hardware fifoplayer: [WrongTitle_191_192.7z.zip](https://github.com/dolphin-emu/dolphin/files/8039038/WrongTitle_191_192.7z.zip) [WrongTitle_96.7z.zip](https://github.com/dolphin-emu/dolphin/files/8039037/WrongTitle_96.7z.zip)

With height=96, when repeat is used there does seem to be some randomness (though it still matches the pattern behind it, as I found).  With mirror, it seems to be consistent (though the different pattern from what Dolphin shows still exists).

PR #9922, opened 2021-07-20
VideoCommon: Expand vector comparisons instead of overloading any()

Techjar:
For whatever stupid reason, Mali drivers do not allow overloading built-in functions.


JMC47:
it's android so ok

XiaomiRedmi9:
This improve mali gpu performance on android?

JosJuice:
No.

PR #9923, opened 2021-07-20
PixelShaderGen: Fix OpenGL ES bounding box compilation error

JosJuice:


JMC47:
lgtm, just waiting to compile.

PR #9925, opened 2021-07-20
GameINI:  Improve Rendering in Sin & Punishment: Star Successor

JMC47:
Sin & Punishment: Star Successor has rendering issues in the upper left part of the screen.  While we can't completely fix it, by disabling Store EFB Copies to Texture only, we can make it from a missing box to just a couple of lines being incorrectly visible that make up the copy region's borders.

This makes the game much more playable in the meantime.

### DIFF
--- a/Data/Sys/GameSettings/R2V.ini
+++ b/Data/Sys/GameSettings/R2V.ini
@@ -1,0 +1,18 @@
+# R2VE01, R2VP01, R2VJ01 - Sin & Punishment: Star Successor
+
+[Core]
+# Values set here will override the main Dolphin settings.
+
+[OnLoad]
+# Add memory patches to be loaded once on boot here.
+
+[OnFrame]
+# Add memory patches to be applied every frame here.
+
+[ActionReplay]
+# Add action replay cheats here.
+
+[Video_Hacks]
+# Store EFB Copies to Texture Only must be disabled to prevent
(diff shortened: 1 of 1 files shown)


JosJuice:
Preferably put a comment about this in the INI itself, in case the issue gets improved on in the future. (Actually, maybe putting comments in the INI is something that would be useful in other cases too...?)

JMC47:
The copy region issue will probably not get fixed.  The lines around it might.  That's why I didn't put a comment in the INI.

JMC47:
Updated with a comment about the copy region.

Data/Sys/GameSettings/R2V.ini
```diff
@@ -0,0 +1,20 @@
...
+
+[OnFrame]
+# Add memory patches to be applied every frame here.
+
+[ActionReplay]
+# Add action replay cheats here.
+
+[Video_Hacks]
+# Store EFB Copies to Texture Only must be disabled to prevent
+# the copy region from showing the incorrect graphics.
```
Rumi-Larry, line 17:
```suggestion
# Store EFB Copies to Texture Only must be disabled to prevent
# the copy region from showing incorrect graphics.
```

Rumi-Larry:
Also there are two extra empty lines at the very bottom.

PR #9929, opened 2021-07-21
DSPHLE: re-enable low-pass filter

Tilka:
Sonic Colors probably needs this for underwater audio: https://bugs.dolphin-emu.org/issues/10069

No idea whether it's still broken or not.

### DIFF
--- a/Source/Core/Core/HW/DSPHLE/UCodes/AXVoice.h
+++ b/Source/Core/Core/HW/DSPHLE/UCodes/AXVoice.h
@@ -419,9 +419,7 @@ void ProcessVoice(PB_TYPE& pb, const AXBuffers& buffers, u16 count, AXMixControl
   }
 
   // Optionally, execute a low pass filter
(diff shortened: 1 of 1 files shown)


JMC47:
This probably doesn't need to be a draft.  I took some time to test a couple dozen games this morning between the different microcode revisions and it seems fine.  Fixes a ton of minor audio errors, from Paper Mario to Sonic Colors.

BhaaLseN:
The comment mentions Super Monkey Ball, did you check that game? If theres nothing of note, I'd also opt to just re-enable it. Plus, if something still _is_ wrong with it, broken audio is more noticable than no audio.

JMC47:
Super Monkey Ball is not broken.  I'd prefer to merge this sooner than later so that we have time to catch any regressions before the next beta.

BhaaLseN (approved):
¯\\\_(ツ)_/¯

PR #9931, opened 2021-07-21
msbuild: reduce number of cl batches for mgba

shuffle2:
minor build perf improvement

### DIFF
--- a/Externals/mGBA/mgba.vcxproj
+++ b/Externals/mGBA/mgba.vcxproj
@@ -26,91 +26,63 @@
   </ItemDefinitionGroup>
   <ItemGroup>
     <ClCompile Include="mgba\src\core\bitmap-cache.c" />
-    <ClCompile Include="mgba\src\core\cache-set.c">
-      <ObjectFileName>$(IntDir)/src/core/cache-set.c.obj</ObjectFileName>
-    </ClCompile>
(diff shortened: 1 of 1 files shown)


Externals/mGBA/mgba.vcxproj
```diff
@@ -26,123 +26,95 @@
...
+    <ClCompile Include="mgba\src\sm83\decoder.c" />
     <ClCompile Include="mgba\src\sm83\isa-sm83.c" />
     <ClCompile Include="mgba\src\sm83\sm83.c" />
-    <ClCompile Include="mgba\src\gb\audio.c">
-      <ObjectFileName>$(IntDir)/src/gb/audio.c.obj</ObjectFileName>
-    </ClCompile>
+    <ClCompile Include="mgba\src\gb\audio.c" />
     <ClCompile Include="mgba\src\gb\cheats.c">
-      <ObjectFileName>$(IntDir)/src/gb/cheats.c.obj</ObjectFileName>
+      <ObjectFileName>$(IntDir)obj2\</ObjectFileName>
```
leoetlino, line 53: Could we call those `gb` and `gba`?

shuffle2: I decided not to because the (object) files aren't actually split up that way (because there are sometimes duplicate names outside of gb/gba dirs)

PR #9934, opened 2021-07-21
Qt: Un-maximize GBA widget before resizing it

Pokechu22:
This is more a "windows is silly" thing: if a window is maximized but then resized, it'll remain maximized and in the top-left corner of the screen, but the size will change.  However, if you attempt to move the window to a different place, it'll become un-maximized and the size it had before maximizing will be used (and it'll ignore the size you set when it was maximized).  Returning the window to a normal state fixes this.


Rumi-Larry:
I thought a maximized window couldn't be resized.

Pokechu22:
Users can't resize it, but the `resize` call in code still manages to resize it for some reason.  It's silly.

PR #9935, opened 2021-07-21
Android Add Drum Overlay Button

blakbin:
Adding Drums Overlay Button for Android
landscape
portrait

Tested in Guitar Hero World Tour


Rumi-Larry:
For the title I suggest "Android: Add overlays for the drums"

JosJuice:
@dolphin-emu-bot rebuild

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/features/settings/model/BooleanSetting.java
```diff
@@ -125,6 +125,27 @@
           "ButtonToggleNunchukZ", true),
   MAIN_BUTTON_TOGGLE_WII_10(Settings.FILE_DOLPHIN, Settings.SECTION_INI_ANDROID_OVERLAY_BUTTONS,
           "ButtonToggleNunchukStick", true),
+  //drum
```
JosJuice, line 128: We always put one space after the `//`.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/overlay/InputOverlay.java
```diff
@@ -218,6 +219,37 @@ public boolean onTouch(View v, MotionEvent event)
                     ButtonState.PRESSED);
           }
           break;
+        case MotionEvent.ACTION_MOVE:
+          // For Drum Overlay Button pressed immediately after touch pointer enter bound
```
JosJuice, line 223: Why? This doesn't match how playing on an actual drum controller works.

JosJuice:
"Android: Add overlays for the drums" isn't significantly better than "Android Add Drum Overlay Button". Though I do prefer the capitalization and punctuation of the former.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/overlay/InputOverlay.java
```diff
@@ -789,6 +882,13 @@ else if (mPreferences.getInt("wiiController", 3) == 4)
       else
         wiiClassicPortraitDefaultOverlay();
     }
+    else if (mPreferences.getInt("wiiController", 3) == 5)
+    {
+      if (isLandscape)
+        DrumsDefaultOverlay();
+      else
+        DrumsPortraitDefaultOverlay();
```
JosJuice, line 890: These two methods should start with lowercase letters to follow our Java coding style.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/overlay/InputOverlay.java
```diff
@@ -924,6 +1024,18 @@ private static InputOverlayDrawableButton initializeOverlayButton(Context contex
...
         scale = 0.25f;
         break;
+      case ButtonType.DRUMS_PAD_BASS:
+        scale = 0.425f;
+        break;
+      case ButtonType.DRUMS_PAD_GREEN:
+      case ButtonType.DRUMS_PAD_RED:
+        scale = 0.4f;
+        break;
+      case ButtonType.DRUMS_PAD_BLUE:
```
JosJuice, line 1034: Why is the blue pad smaller than the green and red pads? Those three are identical on a physical controller.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/overlay/InputOverlay.java
```diff
@@ -50,7 +50,8 @@
   public static final int OVERLAY_WIIMOTE_SIDEWAYS = 2;
   public static final int OVERLAY_WIIMOTE_NUNCHUK = 3;
   public static final int OVERLAY_WIIMOTE_CLASSIC = 4;
-  public static final int OVERLAY_NONE = 5;
+  public static final int OVERLAY_WIIMOTE_DRUM = 5;
+  public static final int OVERLAY_NONE = 6;
```
JosJuice, line 54: The value of `OVERLAY_NONE` should remain `5`, otherwise you'll break existing configs.

JosJuice (commented):
It feels a bit odd to me that there's a Dolphin logo in the bass drum image. We don't do anything like that for other controllers.

Source/Android/app/src/main/res/values/arrays.xml
```diff
@@ -433,6 +435,19 @@
...
+    <string-array name="drumsKit">
+        <item>-</item>
+        <item>+</item>
+        <item>Green</item>
+        <item>Red</item>
+        <item>Blue</item>
+        <item>Yellow</item>
+        <item>Orange</item>
+        <item>Bass 1</item>
+        <item>Bass 2</item>
```
JosJuice, line 447: I'm wondering whether having two bass drums in the overlay actually makes sense.

I'm assuming that the reason you made it this way is because it's possible to use two bass pedals with a real drum controller. However, the gameplay point of connecting two bass pedals is to gain the ability to play the bass drum more quickly by alternating between two limbs – something which is already possible for the 3 pads and 2 cymbals without any extra hardware (you just need to use both arms at once on the same pad). The latter is far more important for gameplay, since the game only expects you to do double bass if you're playing on Expert+, yet this pull request is more accommodating towards playing double bass than playing fast rolls on the pads/cymbals.

Additionally, I (at least as a Rock Band player – I must admit I'm not entirely familiar with how it works in Guitar Hero drum hardware) think it's natural for the user to expect that if they hit the first bass drum and then hit the second bass drum before having released the first one, the hit on the second bass drum registers. As I understand it, this is not the case in this implementation, since you have simply mapped both pads to the same input.

blakbin: the purpose of inserting two pedals is all about positioning. at least if the user needs it, then the default value is false, and it only works in Overlay and not for external gamepads.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/overlay/InputOverlay.java
blakbin: As far as I tried. it's easier that way, than following the original wii drum kit's appearance

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/overlay/InputOverlay.java
blakbin: it because we playin on the overlay, when there are 2 different buttons consecutively, we just slide the finger

blakbin:
> It feels a bit odd to me that there's a Dolphin logo in the bass drum image. We don't do anything like that for other controllers.

changed to plain

Source/Android/app/src/main/res/values/arrays.xml
JosJuice: Ah, so it's a matter of having a conveniently placed bass drum that you can hit no matter where your hand is? Hmm... I suppose that makes sense in a way.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/overlay/InputOverlay.java
JosJuice: That sounds much harder to time correctly than tapping, and the drum charts in the games are not designed in a way that would make this action useful most of the time (there is a lot more of hitting the same pad than moving your hand to an adjacent pad). I feel like the usefulness doesn't outweigh the risk of people accidentally hitting two pads when they meant to hit one.

JosJuice:
@dolphin-emu-bot rebuild

JosJuice:
The code seems OK to me other than the lint error. (Actually, there's a lot more boilerplate code than I would like, but the new code is just following the precedent set by the existing code, so this PR shouldn't be blocked because of it.)

I would like someone else's opinion on my UX comments above. Also, I would like @MayImilae's opinion on the new images.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/overlay/InputOverlay.java
Rumi-Larry: I think making all pads the smaller size and tweaking their position, would look fine.

Source/Android/app/src/main/res/values/arrays.xml
Rumi-Larry: > 
> 
> Ah, so it's a matter of having a conveniently placed bass drum that you can hit no matter where your hand is? Hmm... I suppose that makes sense in a way.

Wouldn't a wide button that can be reached by either finger make more sense then?

JosJuice: I think I agree.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/overlay/InputOverlay.java
blakbin: yes it rarely used, and mostly used when we hold phone with both hand and only using our two thumb. 
i dont mind if its removed

Source/Android/app/src/main/res/values/arrays.xml
blakbin: umm if we make the bass button wider like space button in real keyboard, the only way i can do is editing in the InputOverlayDrawableButton.java resizing before draw, since all image bound is treated as square, without resizing if image has empty space that will cause button got pressed when we tap its empty area, like the classic button L & R

JosJuice: I have submitted PR #9944 to fix this issue.

PR #9936, opened 2021-07-22
MSVC: Fix IntelliSense fmt error false positives

Dentomologist:
Use __VA_OPT__(,) to avoid MSVC IntelliSense "expected an expression" errors when calling log or alert functions with no replacement parameters. This error is triggered by the comma after the format string having an empty ##\_\_VA_ARGS__ after it, making it a trailing comma. The compiler ignores the comma when followed by ##\_\_VA_ARGS__ but IntelliSense does not.

This workaround unfortunately allows for calling log/alert functions in MSVC with a trailing comma when there are no replacement parameters, PanicAlertFmt("Like this",); , but those calls will fail on buildbot for other platforms so we don't need to worry about errant commas making their way into the codebase.

### DIFF
--- a/Source/Core/Common/Logging/Log.h
+++ b/Source/Core/Common/Logging/Log.h
@@ -143,6 +143,30 @@ void GenericLog(LOG_LEVELS level, LOG_TYPE type, const char* file, int line, con
 
 // fmtlib capable API
 
+#ifdef _MSC_VER
+// Use __VA_OPT__(,) to avoid MSVC IntelliSense "expected an expression" error when calling a log
+// function with no replacement parameters. This is caused by the comma after the format string
+// having an empty ##__VA_ARGS__ after it, making it a trailing comma. The compiler ignores the
+// comma when followed by ##__VA_ARGS__ but IntelliSense does not.
+//
+// This workaround unfortunately allows for calling log functions in MSVC with a trailing comma when
+// there are no replacement parameters, INFO_LOG_FMT(COMMON, "Like this",); , but those calls work
+// correctly so the tradeoff to get rid of false positive errors everywhere is worth it.
+//
+// TODO: Remove workaround if/when MSVC fixes the bug
+#define GENERIC_LOG_FMT(t, v, format, ...)                                                         \
+  do                                                                                               \
+  {                                                                                                \
+    if (v <= MAX_LOGLEVEL)                                                                         \
+    {                                                                                              \
+      /* Use a macro-like name to avoid shadowing warnings */                                      \
+      constexpr auto GENERIC_LOG_FMT_N = Common::CountFmtReplacementFields(format);                \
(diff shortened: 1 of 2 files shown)


Source/Core/Common/Logging/Log.h
```diff
@@ -143,6 +143,30 @@ void GenericLog(LOG_LEVELS level, LOG_TYPE type, const char* file, int line, con
...
+// Use __VA_OPT__(,) to avoid MSVC IntelliSense "expected an expression" error when calling a log
+// function with no replacement parameters. This is caused by the comma after the format string
+// having an empty ##__VA_ARGS__ after it, making it a trailing comma. The compiler ignores the
+// comma when followed by ##__VA_ARGS__ but IntelliSense does not.
+//
+// This workaround unfortunately allows for calling log functions in MSVC with a trailing comma when
+// there are no replacement parameters, INFO_LOG_FMT(COMMON, "Like this",); , but those calls work
+// correctly so the tradeoff to get rid of false positive errors everywhere is worth it.
+//
+// TODO: Remove workaround if/when MSVC fixes the bug
```
shuffle2, line 156: It is not msvc bug, rather intellisense. Has the issue been reported to them?
Generally, I would say live with the intellisense annoyance until it's fixed instead of making a workaround which doesn't seem required..

Pokechu22: It was reported here a while back, but hasn't been re-reported since then: https://developercommunity.visualstudio.com/t/intellisense-reports-false-positive-error-in-varia/560370

IMO, it's worth keeping this fix as technically `__VA_OPT__(,)` is a standard way of handling this, while the behavior with `##__VA_ARGS__` removing a preceding comma is a non-standard extension (though it's done in all compilers we use).  The only reason why multiple versions are needed is that `__VA_OPT__` is a C++20 feature, but Dolphin already uses C++20 for MSVC (or rather Dolphin uses the latest standard instead of having a set one), while other compilers aren't using C++20 yet.

Dentomologist: It was reported [here](https://developercommunity.visualstudio.com/t/intellisense-reports-false-positive-error-in-varia/560370) two years ago and they eventually closed it saying it wasn't important enough to fix.

It's not the end of the world, but IMO it's annoying and been a bug for a long time. I know for me it makes the red squiggly lines a lot less useful via boy-who-cried-wolf.

shuffle2: I'm confused. It sounds like the "workaround" is actually the existing code. So the comment about removing code should instead be on the non-msvc code?

Dentomologist: This commit is the workaround. My intention was that it would eventually be removed if MS fixed the IntelliSense bug, but based on Pokechu's point maybe I'll just remove that part of the comment.

Dentomologist:
Closing in favor of #11008.

PR #9937, opened 2021-07-22
NetPlay: Split save and SD card write settings

Techjar:
This actually eliminates any setting pertaining to SD cards from the NetPlay dialog, as it would effectively just be a duplicate of the setting in the Wii pane, potentially causing confusion.

This also enables save data writing by default, as this is probably what most players want, and should avoid them losing hours of progress because they forgot to tick a checkbox.


JMC47 (approved):
Seems to work for me.

lioncash (approved):

PR #9938, opened 2021-07-22
GBA: Add "Scan e-Reader Card(s)" context menu item

Pokechu22:
See [issue 12588](https://bugs.dolphin-emu.org/issues/12588).  This doesn't introduce any more complicated functionality; it only exposes the functionality that was already available via drag and drop in a more discoverable way.

I haven't tested this at all.

### DIFF
--- a/Source/Core/DolphinQt/GBAWidget.cpp
+++ b/Source/Core/DolphinQt/GBAWidget.cpp
@@ -50,6 +50,16 @@ static void RestartCore(const std::weak_ptr<HW::GBA::Core>& core, std::string_vi
       false);
 }
 
+static void QueueEReaderCard(const std::weak_ptr<HW::GBA::Core>& core, std::string_view card_path)
+{
+  Core::RunOnCPUThread(
+      [core, card_path = std::string(card_path)] {
+        if (auto core_ptr = core.lock())
+          core_ptr->EReaderQueueCard(card_path);
+      },
+      false);
+}
+
 GBAWidget::GBAWidget(std::weak_ptr<HW::GBA::Core> core, int device_number,
                      std::string_view game_title, int width, int height, QWidget* parent,
                      Qt::WindowFlags flags)
@@ -164,6 +174,16 @@ void GBAWidget::UnloadROM()
     return;
 
   RestartCore(m_core);
+}
+
+void GBAWidget::PromptForEReaderCards()
+{
+  const QStringList card_paths = QFileDialog::getOpenFileNames(
+      this, tr("Select e-Reader Cards"), QString(), tr("e-Reader Cards (*.raw);;All Files (*)"),
+      nullptr, QFileDialog::Options());
+
+  for (const QString& card_path : card_paths)
+    QueueEReaderCard(m_core, QDir::toNativeSeparators(card_path).toStdString());
 }
 
 void GBAWidget::ResetCore()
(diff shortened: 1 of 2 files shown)


JMC47:
I guess I'll have to test this lol

JMC47 (approved):
Was able to load e-Reader cards via the context menu and unlock the content.

Source/Core/DolphinQt/GBAWidget.cpp
```diff
@@ -166,6 +166,23 @@ void GBAWidget::UnloadROM()
   RestartCore(m_core);
 }
 
+void GBAWidget::QueueEReaderCard()
+{
+  QString card_path = QDir::toNativeSeparators(QFileDialog::getOpenFileName(
```
Bonta0, line 171: This should call ``QFileDialog::getOpenFileNames`` instead to add multiple files at once for convenience

Source/Core/DolphinQt/GBAWidget.cpp
```diff
@@ -166,6 +166,23 @@ void GBAWidget::UnloadROM()
...
+
+  if (card_path.isEmpty())
+    return;
+
+  Core::RunOnCPUThread(
+      [core = m_core, card_path = card_path.toStdString()] {
+        if (auto core_ptr = core.lock())
+          core_ptr->EReaderQueueCard(card_path);
+      },
+      false);
```
Bonta0, line 183: Since this shows up more than once now it might be better to move into its own function

Source/Core/DolphinQt/GBAWidget.cpp
```diff
@@ -281,6 +298,10 @@ void GBAWidget::contextMenuEvent(QContextMenuEvent* event)
   unload_action->setEnabled(CanControlCore() && !m_game_title.empty());
   connect(unload_action, &QAction::triggered, this, &GBAWidget::UnloadROM);
 
+  auto* card_action = new QAction(tr("&Load e-Reader Card"), menu);
+  card_action->setEnabled(CanControlCore() && !m_game_title.empty());
```
Bonta0, line 302:
```suggestion
  card_action->setEnabled(CanControlCore());
```

Pokechu22: Don't you need to load the e-Reader ROM first?  Or does that happen automatically?

Source/Core/DolphinQt/GBAWidget.cpp
```diff
@@ -281,6 +298,10 @@ void GBAWidget::contextMenuEvent(QContextMenuEvent* event)
   unload_action->setEnabled(CanControlCore() && !m_game_title.empty());
   connect(unload_action, &QAction::triggered, this, &GBAWidget::UnloadROM);
 
+  auto* card_action = new QAction(tr("&Load e-Reader Card"), menu);
```
Pokechu22, line 301: Thoughts on changing this to `&Scan e-Reader Card`?

JMC47: Hard agree.  Because most e-Reader cards are in pairs, you have to scan two at once.

Source/Core/DolphinQt/GBAWidget.cpp
JMC47: e-Reader itself has to be loaded in order to scan cards.  It's like a normal game in how it is loaded in mGBA.

Source/Core/DolphinQt/GBAWidget.cpp
JMC47: Yeah, as e-Reader cards are in pairs, loading two at once would probably be nice.

Source/Core/DolphinQt/GBAWidget.cpp
Pokechu22: Alright, then I think it makes sense to still require a non-empty game title if you can't load a card when the e-Reader itself is loaded.  (It'd be better to check if it's the e-Reader itself that's loaded, but I'm not sure how to do that.)

Source/Core/DolphinQt/GBAWidget.cpp
Pokechu22: I've implemented loading multiple cards at once.  Would it make sense to change it to `&Scan e-Reader Card(s)`, or is just `&Scan e-Reader Card` fine?

JMC47:
Should the Pull Request's name be changed then?

lioncash (approved):

PR #9939, opened 2021-07-22
Readme: Copy the submodule note to non-Windows sections

JosJuice:
Pulling submodules is necessary now that mGBA is integrated.

Perhaps having the same submodule note repeated three times in the different sections of the readme file is a bit odd, but I don't know of a fitting section to put it in that isn't OS-specific...

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -63,6 +63,11 @@ bundled with Dolphin and used if they're not installed on your system. CMake
 will inform you if a bundled library is used or if you need to install any
 missing packages yourself.
 
+Make sure to pull submodules before building:
+```sh
+git submodule update --init
+```
+
 ### macOS Build Steps:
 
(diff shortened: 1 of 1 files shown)


JosJuice:
Hm, pulling all submodules will result in the Windows Qt submodule getting pulled, even though it's pointless on systems other that Windows... Should we recommend something other than pulling all submodules?

BhaaLseN:
If it wasn't for Windows, I'd suggest writing a shell script that checks the system and excludes Qt if not needed (or maybe even have CMake do it).
Or, the straight-forward way, just add the submodule path to the command invocation and hope that we don't forget to update the Readme next time we add one.

leoetlino:
eh, I don't think it's worth the trouble when our non-Windows buildbots also just pull all submodules

PR #9940, opened 2021-07-22
GameINI:  Fix de Blob 2 visual issues.

JMC47:
de Blob 2 has the same engine as de blob and shares a lot of the same issues.  I copied the settings used for de blob.  It seems defer EFB copies was the main culprit of the black box issue returning, but force texture filtering could also cause problems.


iwubcode (approved):
Confirmed fixes the black box

lioncash (approved):

PR #6219, opened 2017-11-25
PixelShaderGen: only use dual-source if bpmem requires dual-src blending

ligfx:
This fixes [Issue 9373: Depth issues on Mario Kart Double Dash](https://bugs.dolphin-emu.org/issues/9373).

There seems to be some interaction between having `ocol1` in the fragment shader and calling `glBindFragDataLocationIndexed`: removing either one fixes the issue. Adding stronger dual source requirements to the fragment shader seemed like the cleaner option.

Before:

screen shot 2017-11-25 at 2 18 01 pm

After:

screen shot 2017-11-25 at 2 17 32 pm

PR #4402, opened 2016-10-29
OGL: Fall back to the old dual-source blending behaviour.

Armada651:
This allows us to still use dual-source blending in cases where it is required for accurate emulation and not just for small precision enhancements where it could crash some driver.

I also added OS X Intel to the bug, as those drivers seem to be affected by a similar bug. The OS X Nvidia drivers are not affected. I've tested it on my own MacBook Pro and confirmed this PR works correctly.

---

This change is [Reviewable](https://reviewable.io/reviews/dolphin-emu/dolphin/4402)

PR #9945, opened 2021-07-23
DriverDetails: Add broken dual source blending bug to MoltenVK on Intel GPUs

OatmealDome:
The Metal driver on Intel GPUs appears to have a bug where if ``ocol1`` is present in the fragment shader and dual source blending is enabled, entire primitives will not be written to the depth buffer if any fragment is discarded.

This causes many games to exhibit odd layering issues that look similar to Z-fighting. The problem is especially pronounced in the Paper Mario games, where it affects basically every sprite-like object.

Apparently, a similar-ish bug also involving ``ocol1`` is also present in the OpenGL driver: see #4402, #6219.

Fixes https://bugs.dolphin-emu.org/issues/12137 and https://bugs.dolphin-emu.org/issues/12317.

(I'm not very experienced in graphics programming, so the comments / explanation might be wrong. Feel free to correct.)

<details>
  <summary>Depth buffer comparison</summary>

Dual source blending enabled:

image

Dual source blending disabled:

image

</details>


JosJuice:
@dolphin-emu-bot rebuild

JMC47:
All of this sounds pretty logical based on the bugs we've seen.

OatmealDome:
Rebased and modified to be more in line with how OpenGL handles ``BUG_BROKEN_DUAL_SOURCE_BLENDING`` (i.e. only enable dual source blending when necessary to emulate destination alpha).

phire (approved):

PR #9947, opened 2021-07-23
Re-scale of buttons

Gamer64ytb:


MayImilae:
You've not stated why you want this change. The scaling for the on-screen buttons was set as they are intentionally, so, what is your reason for wanting to change some of them?

Gamer64ytb:
> You've not stated why you want this change. The scaling for the on-screen buttons was set as they are intentionally, so, what is your reason for wanting to change some of them?

It is a more comfortable scaling for users, it looks like this:

Screenshot_2021-07-24-15-52-04-114_org dolphinemu gamerytb

Bankaimaster999:
Is it really? I don't notice that much of a difference in my opinion ....

AdmiralCurtiss:
Why not instead allow users to scale them themselves rather than hardcoding a different arbitrary scale? We already allow custom positioning I think, so just extend that.

Bankaimaster999:
@AdmiralCurtiss There is general (ALL) button scaling already implemented in dolphin so this really doesn't matter.

jordan-woyak:
This PR has aged and is perhaps unnecessary so I'm going to close it for now. Feel free to re-open upon new information.

PR #9949, opened 2021-07-24
DiscIO: Fix broken workaround for MSVC ARM64 ICE

JosJuice:
The workaround added in PR #9855 caused a regression where Dolphin incorrectly replaced runs of one byte with runs of another byte when writing WIA and RVZ files. `ReuseID::operator<` was always returning false unless the `ReuseID`s being compared had different partition keys, which caused `std::map<ReuseID, GroupEntry>` to treat all `ReuseID`s with the same partition key as equal.

### DIFF
--- a/Source/Core/DiscIO/WIABlob.h
+++ b/Source/Core/DiscIO/WIABlob.h
@@ -240,22 +240,26 @@ private:
 
   struct ReuseID
   {
-    // This is a workaround for an ICE in Visual Studio 16.10.0 when making an ARM64 Release build.
-    // Once the ICE has been fixed upstream, we can move partition_key inside the tie.
-#define COMPARE_TIED(op)                                                                           \
-  (partition_key op other.partition_key) &&                                                        \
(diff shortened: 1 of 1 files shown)


shuffle2:
how about this? :D (no ICE here, but ofc would be great to check the logic)

~~~c++
#define COMPARE_TIED_INNER(op, f, next) (f != other.f) ? (f op other.f) : next

#define COMPARE_TIED(op) \
  COMPARE_TIED_INNER(op, partition_key, \
    COMPARE_TIED_INNER(op, data_size, \
      COMPARE_TIED_INNER(op, encrypted, \
        COMPARE_TIED_INNER(op, value, true))))
~~~

JosJuice:
I can't find anything wrong when looking through the logic, but it seems to break the reuse optimization in the opposite way (it's applied less often than it should be, rather than more often than it should be).

JosJuice:
While it's not exactly pretty, it's the only solution so far that both works correctly (after fixing the bug in it) and doesn't trigger the ICE. Implemented.

PR #9950, opened 2021-07-24
Netplay: Adjust text to help with controls

JMC47:
Ever since we added GBA support on netplay, there's been a lot of confusion about how controls are assigned on netplay.  Pretty much *everyone* has a controller configured to slot 1, so this didn't come up as much.  I added clarification to how Dolphin loads controller configurations to the netplay information in the connect menu.

image


JosJuice:
Is there any particular reason why it says "profile" instead of "input configuration" for Wii Remotes? I feel like we should use the same term for it in both sentences.

JMC47:
I was trying to fit it, but let me see

Source/Core/DolphinQt/NetPlay/NetPlaySetupDialog.cpp
```diff
@@ -110,17 +110,20 @@ void NetPlaySetupDialog::CreateMainLayout()
...
-             "If DSP LLE is used, DSP ROMs must be identical between players.\n"
-             "If a game is hanging on boot, it may not support Dual Core Netplay."
-             " Disable Dual Core.\n"
-             "If connecting directly, the host must have the chosen UDP port open/forwarded!\n"
-             "\n"
-             "Wii Remote support in netplay is experimental and may not work correctly.\n"
-             "Use at your own risk.\n")),
+      new QLabel(tr(
+          "ALERT:\n\n"
+          "Netplay will use the input configuration from controller port 1 first,\n"
```
Techjar, line 115: Make this "NetPlay" for consistency :)

Source/Core/DolphinQt/NetPlay/NetPlaySetupDialog.cpp
```diff
@@ -110,17 +110,20 @@ void NetPlaySetupDialog::CreateMainLayout()
...
+          "regardless of the port you are assigned during the netplay session.\n"
+          "Wii Remotes, on the other hand, use the configuration from the assigned netplay slot.\n"
```
Techjar, line 117: Same here.

Source/Core/DolphinQt/NetPlay/NetPlaySetupDialog.cpp
```diff
@@ -110,17 +110,20 @@ void NetPlaySetupDialog::CreateMainLayout()
...
+          "All players must use the same Dolphin version.\n"
+          "If enabled, SD cards must be identical between players.\n"
+          "If DSP LLE is used, DSP ROMs must be identical between players.\n"
+          "If a game is hanging on boot, it may not support Dual Core Netplay."
+          " Disable Dual Core.\n"
+          "If connecting directly, the host must have the chosen UDP port open/forwarded!\n"
+          "\n"
+          "Wii Remote support in netplay is experimental and may not work correctly.\n"
+          "Use at your own risk.\n")),
```
Techjar, line 126: Might as well fix all the case inconsistencies while we're messing with this. :)

I also cleaned up the phrasing a little bit, and changed some spacing.

```suggestion
          "ALERT:\n\n"
          "GameCube Controllers will use the input configuration starting from controller port 1,"
          " regardless of the port you are assigned during the NetPlay session.\n"
          "Wii Remotes, on the other hand, use the configuration corresponding to your"
          " assigned port in the NetPlay session.\n"
          "\n"
          "All players must use the same Dolphin version.\n"
          "If enabled, SD cards must be identical between players.\n"
          "If DSP LLE is used, DSP ROMs must be identical between players.\n"
          "If a game is hanging on boot, it may not support Dual Core NetPlay."
          " Disable Dual Core.\n"
          "If connecting directly, the host must have the chosen UDP port open/forwarded!\n"
          "\n"
          "Wii Remote support in NetPlay is experimental and may not work correctly.\n"
          "Use at your own risk.\n")),
```

Techjar (changes requested):

iwubcode:
I'm probably stating the obvious and I know that it's a lot more work but just maybe something to note for the future.  It feels like most of these comments should be a validation step done by netplay, with a very explicit error to the user if they've done something wrong:

> SD cards not identical between player x and  player y!

Possibly even with a "fix now" or "fix for this session" or something (if applicable).

Otherwise, this is a lot for users to remember...

Dentomologist:
Is there a reason we can't make Wiimotes work the same way as other controllers, or have we just not done that yet?

JMC47:
It was problems with settings like extensions.  It's definitely fixable but requires some work.

Techjar:
> I'm probably stating the obvious and I know that it's a lot more work but just maybe something to note for the future. It feels like most of these comments should be a validation step done by netplay, with a very explicit error to the user if they've done something wrong:
> 
> > SD cards not identical between player x and  player y!
> 
> Possibly even with a "fix now" or "fix for this session" or something (if applicable). (in this case, if you wanted to go the extra mile, SD card could be sent over internet and old one backed up but then restored on session exit)
> 
> Otherwise, this is a lot for users to remember...

The reason this hasn't been done is because of how large the SD card images tend to be. It would be quite a long wait every time you joined a session, so I think it may be best to just let users sort it out.

As for validation, there is an option in the MD5 checksum menu. Again, the SD card images can be rather large, so validating automatically on every start may be rather inconvenient.

JMC47:
Could we have a setting to disable dualcore on netplay?  I can probably have hundreds of games to add to that.

iwubcode:
> The reason this hasn't been done is because of how large the SD card images tend to be. It would be quite a long wait every time you joined a session, so I think it may be best to just let users sort it out.
> As for validation, there is an option in the MD5 checksum menu. Again, the SD card images can be rather large, so validating automatically on every start may be rather inconvenient.

Are the sizes always the same?  If not a size check would be at least a simple sanity check.  If the sizes are the same, well, then not a lot to do I agree.  I'm a bit surprised that a MD5 computation would be that slow?  Are we talking seconds or minutes?

> Dual Core compatibility can't be pre-validated. Some games will just hang and others will run fine, so users will have to test it themselves.

I mean, if we don't want to force it like @JMC47 said, we could at least have a setting that says "netplayrequiresdualcoreoff" and any time we see that, we error if they have it on.

> We could probably validate that the DSP ROMs match, though. Not sure how much of a problem that is in practice.

Yeah, it's more about narrowing down the issue when there's a problem and trying to provide a nicer user experience.

--

Dolphin version seems like it'd be rather simple to error if it doesn't match.

PR #9952, opened 2021-07-24
WiimoteReal: Remove unused IsReady() functions

Dentomologist:
The backends have IsReady() functions which would be called from WiimoteScanner::IsReady(), but that function itself is never called. I went digging through history, and as far as I can tell it never was...?

The Linux backend did call WiimoteScannerLinux::IsReady() from its destructor, but in that context I think the code is clearer if it's inlined so I did.

### DIFF
--- a/Source/Core/Core/HW/WiimoteReal/IOAndroid.h
+++ b/Source/Core/Core/HW/WiimoteReal/IOAndroid.h
@@ -44,7 +44,6 @@ class WiimoteScannerAndroid final : public WiimoteScannerBackend
 public:
   WiimoteScannerAndroid() = default;
   ~WiimoteScannerAndroid() override = default;
-  bool IsReady() const override { return true; }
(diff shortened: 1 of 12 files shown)


leoetlino:
This was used in the past to inform the user no Bluetooth adapter was detected if IsReady() returned false. I guess this didn't survive during the Qt migration?

https://github.com/dolphin-emu/dolphin/blob/6ff06ed41d49971859036999b8892c7a6c862dca/Source/Core/DolphinWX/ControllerConfigDiag.cpp#L321-L324

leoetlino:
Closing this because we do actually want to re-add the warning. I'll make a PR for that tomorrow

Dentomologist:
This PR was closed in favor of re-adding the warning, which happened in #11601. Now that #14595 removed the warning this PR makes sense again.

PR #9954, opened 2021-07-24
DSPHLE: fix problems with older ucodes

Tilka:
This fixes https://bugs.dolphin-emu.org/issues/12515.


Rumi-Larry:
Above on the lines 246-248 where it says:
`  // Capcom vs. SNK 2, Naruto 2, Lost Kingdoms, Star Fox, Mario Party 4, Mortal Kombat,
  // Smugglers Run Warzone, Smash Brothers, Sonic Mega Collection, ZooCube
  // nddemo, Star Fox`
There seems to be some issues with the indentation. If some uCodes are supposed to be there, then some kind of placeholder should be placed.

Rumi-Larry:
Also, on line 234 there's this `//CARD` comment that's missing from the other cases like GBA or ROMS. I think it can be removed.

Source/Core/Core/HW/DSPHLE/UCodes/AXWii.cpp
```diff
@@ -214,7 +214,7 @@ void AXWiiUCode::HandleCommandList()
 
       case CMD_OUTPUT:
       case CMD_OUTPUT_DPL2:
-        volume = m_cmdlist[curr_idx++];
+        volume = m_crc == 0xd9c4bf34 ? 0x8000 : m_cmdlist[curr_idx++];
```
leoetlino, line 217: Where does this value come from? Is it hardcoded in that particular µcode?

leoetlino:
@Rumi-Larry while the formatting should probably be fixed (in a separate PR), those changes are not relevant to this PR at all.

Source/Core/Core/HW/DSPHLE/UCodes/AXWii.cpp
Tilka: 0x8000 is full volume. I'm just copying the behavior of CMD_OUTPUT_OLD. I haven't independently verified that this is correct.

leoetlino: CMD_OUTPUT in µcode 0xd9c4bf34:

image

Seems to match what we do in OutputSamples (the whole clamping loop is a no-op considering the volume never changes from 0x8000).

leoetlino (approved):

PR #9955, opened 2021-07-25
DSPHLE: fix and re-enable polyphase resampling

Tilka:
This fixes the title screen music of Snowpack Park (https://bugs.dolphin-emu.org/issues/12499).

### DIFF
--- a/Source/Core/Core/HW/DSPHLE/UCodes/AXVoice.h
+++ b/Source/Core/Core/HW/DSPHLE/UCodes/AXVoice.h
@@ -231,11 +231,8 @@ u32 ResampleAudio(std::function<s16(u32)> input_callback, s16* output, u32 count
 {
   int read_samples_count = 0;
 
-  // TODO(delroth): find out why the polyphase resampling algorithm causes
-  // audio glitches in Wii games with non integral ratios.
-
   // If DSP DROM coefficients are available, support polyphase resampling.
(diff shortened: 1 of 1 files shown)


JMC47 (changes requested):
This doesn't work with savestates, currently.

Source/Core/Core/HW/DSPHLE/UCodes/AXVoice.h
```diff
@@ -264,7 +261,7 @@ u32 ResampleAudio(std::function<s16(u32)> input_callback, s16* output, u32 count
 
       s64 samp = (t0 * c[0] + t1 * c[1] + t2 * c[2] + t3 * c[3]) >> 15;
 
-      output[i] = (s16)samp;
+      output[i] = (s16)std::clamp(samp, s64(-0x8000), s64(0x7FFF));
```
sepalani, line 264: Can't the [SaturingCast](https://github.com/dolphin-emu/dolphin/blob/02309bd0acf030f5f641d427e55b73d59455c4b1/Source/Core/Common/MathUtil.h#L36) helper function be used here?

JMC47:
This fixes the HLE audio issues in Snowpack Park and the two Pokemon games.

JMC47 (approved):
The Savestate issue as a pre-existing condition.

PR #9956, opened 2021-07-25
VideoCommon: Manually handle texture wrapping and sampling

Pokechu22:
This PR reimplements texture sampling logic in the pixel shaders, rather than using the video backend's built-in texture sampling functionality.  This fixes a number of bugs that occur in various drivers (such as the EA VP6 issue), and also allows implementing a few additional features that were not handled before (diagonal LOD and the jank wrap logic used by actual hardware and implemented for the software renderer in #9921).  Currently, anisotropic filtering is not implemented, but that can be handled in a later PR.


JMC47:
This does fix the EA VP6 issues on my NVIDIA.

iwubcode:
@Pokechu22  - just to confirm, is this ready to go?

Pokechu22:
It's ready to be reviewed, yes.  There still are some particularly bad performance decreases with mesa that I haven't investigated (fifoci takes ~45 minutes instead of a more normal 10-20 minutes for ogl-lin-mesa (compare this to ogl-lin-radeon, which went from ~4 minutes to ~7 minutes)), which may require the fifoci configuration to be changed.  But apart from that, all features are implemented and the code has been reorganized in a way that shouldn't be too bad to review.

Source/Core/DolphinQt/Config/Graphics/HacksWidget.cpp
```diff
@@ -106,11 +106,14 @@ void HacksWidget::CreateWidgets()
   m_vertex_rounding = new GraphicsBool(tr("Vertex Rounding"), Config::GFX_HACK_VERTEX_ROUDING);
   m_save_texture_cache_state =
       new GraphicsBool(tr("Save Texture Cache to State"), Config::GFX_SAVE_TEXTURE_CACHE_TO_STATE);
+  m_native_texture_sampling =
```
iwubcode, line 109: Since we already use "Native" to mean the emulated system (ex: "Native Resolution"), I wonder if this should be called something else

iwubcode:
I did some preliminary testing to see if the issues I've seen on a variety of games were resolved.  Unfortunately not.

That being said, I didn't notice anything regress either.  I'm curious what the incompatibility is with 'Anisotropic Filtering'?  I had it on and off with both the GPU sampling and the manual sampling and didn't notice any issues.

Source/Core/DolphinQt/Config/Graphics/HacksWidget.cpp
```diff
@@ -276,6 +279,15 @@ void HacksWidget::AddDescriptions()
       "higher internal resolutions. This setting has no effect when native internal "
       "resolution is used.<br><br><dolphin_emphasis>If unsure, leave this "
       "unchecked.</dolphin_emphasis>");
+  static const char TR_NATIVE_TEXTURE_SAMPLING_DESCRIPTION[] = QT_TR_NOOP(
+      "Use the video backend's built-in texture sampling functionality instead of a manual "
+      "implementation.<br><br>This can introduce graphical issues in some games on some GPUs (such "
+      "as breaking pre-rendered VP6 videos in some EA titles), but will generally perform "
+      "better.<br><br>Disabling this option uses manual texture sampling, which more accurately "
+      "handles texture wrapping and can bypass GPU problems.  Some of the accuracy improvements do "
```
iwubcode, line 287: > Some of the accuracy improvements do not apply when using custom textures or when using a higher internal resolution with Scaled EFB Copy enabled

I'd be curious what accuracy improvements aren't available and whether this new implementation causes issues with higher IR / custom textures..

Pokechu22:
I looked through all of the fifoci changes above.  Only a few had actually interesting differences that could be checked on hardware (though it's still hard to see due to analog signal stuff and the hardware fifoplayer having some vertex explosion issues).

Test | New | Hardware | Old
-|-|-|-
Melty Motlen Galaxy |  | MMGCrop | 
Baby Park (diagonal LOD) |  | _BabyPark | 
Wario's Gold Mine (diagonal LOD) |  | _mkwii_bluebox | 
Need for Speed |  | _nfsu | 
RS2 |  | _RS2_frame1 | 
Galaxy 2 clouds |  | _marioeyes | 
Tony Hawk |  | _THPS | 

Most things seem to be improvements.  However, RS2 has a red color on the tube (which is very hard to see), [old](https://user-images.githubusercontent.com/8334194/130728730-c4b2b163-7d70-461f-bde8-8caacb272314.png) / [new](https://user-images.githubusercontent.com/8334194/130728726-a078a89a-d4c4-4eab-b570-e296aca08a31.png).  Need for speed underground has some new blue dots in the headlights which are odd.  One triangle in Tony Hawk changed in brightness (but I can't tell what's right).  Also the shadows in Tony Hawk (both 3 and 4) flicker now which seems like a bigger problem - this happens even when looping frame 0 only.

------

> I'm curious what the incompatibility is with 'Anisotropic Filtering'? I had it on and off with both the GPU sampling and the manual sampling and didn't notice any issues.

Anisotropic filtering is a form of texture filtering (related to mipmapping).  It applies when performing texture sampling.  I haven't implemented it for manual texture sampling, so it basically acts as if the value is set to 1x.

Source/Core/DolphinQt/Config/Graphics/HacksWidget.cpp
Pokechu22: The only accuracy improvement that isn't available is emulating the broken way the console handles wrapping when the texture size isn't a power of 2 (see [bug 12408](https://bugs.dolphin-emu.org/issues/12408)), which is disabled when using custom textures/higher IR it's possible to have textures with a non-power-of-2 size that were originally a power of 2 size.  It's pretty minor and shouldn't affect any official games.  The fixes for bad GPU behavior should still apply.

Pokechu22:
I've fixed the rendering regressions for Need for Speed, THPS4, and the lego star wars crane; they were caused by computing LOD based on `round(log2(max_delta) * 16)` when it should have been `floor(log2(max_delta) * 16)`.  I also fixed the small with RS2, which was caused by trying to compute `log2(0)` when the texture coordinates on all vertices of a triangle were the same (combined with integer underflow); I [hardware tested it](https://gist.github.com/Pokechu22/fabbd02f7ce17cb3779c3b8d5dfed213) and found that it uses the min lod value in that case (see [res_hw.zip](https://github.com/dolphin-emu/dolphin/files/7095392/res_hw.zip)).  Changes from this fix compared to an earlier version of the PR can be found [here](https://fifo.ci/version/4c7845b1dcbf97130ef4ce0059c1f585c038e0e9/).

I also did a bit of testing to confirm mipmapping behavior on hardware:

<table>
<tr><th>Test</th><th width="30%">Diagonal LOD, 45 deg</th><th width="30%">Edge LOD, 45 deg</th><th width="30%">Edge LOD, 0 deg</th></tr>
<tr><th>Hardware</th><td>NPN_A1_D_N_1 5_45_1 0</td><td>NPN_A1_E_N_1 5_45_1 0</td><td>NPN_A1_E_N_1  5_0_1 0</td></tr>
<tr><th>Software, new</th><td>NPN_A1_D_N_1 5_45_1 0</td><td>NPN_A1_E_N_1 5_45_1 0</td><td>NPN_A1_E_N_1  5_0_1 0</td></tr>
<tr><th>Vulkan, new</th><td>NPN_A1_D_N_1 5_45_1 0</td><td>NPN_A1_E_N_1 5_45_1 0</td><td>NPN_A1_E_N_1  5_0_1 0</td></tr>
<tr><th>Software, old</th><td>NPN_A1_D_N_1 5_45_1 0</td><td>NPN_A1_E_N_1 5_45_1 0</td><td>NPN_A1_E_N_1  5_0_1 0</td></tr>
<tr><th>Vulkan, old</th><td>NPN_A2_D_N_1 5_45_1 0</td><td>NPN_A1_E_N_1 5_45_1 0</td><td>NPN_A1_E_N_1  5_0_1 0</td></tr>
</table>

You can see that things are a lot closer to correct, particularly with diagonal LOD.  There is some difference in brightness, but that seems to be a vertex lighting-related issue (probably also related to clipping; the test scene here uses very large triangles).

The shadows in THPS3 still seem to have regressed; they randomly flicker on OGL (even when looping frame 0).  I don't know what's causing that.

Pokechu22:
I did a bit of debugging with RenderDoc, and it seems that the shadow in THPS3 isn't rendering because _sometimes_ for the shadow texture, `pack2.w` is 0 (indicating min lod and max lod is 0), but sometimes it's 65280 (indicating min lod of 0 and max lod of 255).  When the max lod is 255, the top-left of the shadow texture is used (even though there is only a single actual mipmap level).  So it ends up being mostly transparent, and the alpha test fails.

I don't know why it's doing that, though.  It must be a race condition of some sort (and 255 indicates it's probably from `RenderState::GetPointSamplerState` or `RenderState::GetLinearSamplerState`; the object before the shadow has a min and max lod of 0 too so it's not that).

iwubcode:
I took another pass at the code.  Going to do some more testing this weekend.  In particular, I want to throw a number of custom texture packs at it to see if there are any issues (since fifoci doesn't test that).  I'll try and finish up my review this weekend.  Nice work pokechu

Source/Core/DolphinQt/Config/Graphics/HacksWidget.cpp
Pokechu22: Maybe "Fast texture sampling" (with the alternative still being "Manual texture sampling") would work; it's not _necessarily_ true (since sometimes manual texture sampling is faster) but it'd make things more obvious to end-users who assume that fast means less accurate.

Source/Core/VideoCommon/RenderState.h
```diff
@@ -93,42 +93,64 @@ union BlendingState
...
+    tm0.hex = rhs.tm0.hex;
+    tm1.hex = rhs.tm1.hex;
+    return *this;
+  }
+
+  bool operator==(const SamplerState& rhs) const { return Hex() == rhs.Hex(); }
+  bool operator!=(const SamplerState& rhs) const { return Hex() != rhs.Hex(); }
+  bool operator<(const SamplerState& rhs) const { return Hex() < rhs.Hex(); }
 
-  enum class AddressMode : StorageType
```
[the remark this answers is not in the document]
iwubcode: After thinking some more about this, from a generic high level, I do wish we hadn't bled the emulation state into our somewhat generic graphics api.  Obviously there is some overlap and this is relatively minor but I'm not sure this was necessary.  My assumption is the point of `Generate` was to make that conversion from emulation to graphics backend.  However, not suggesting you revert or modify this, just something I think we should avoid in the future.

To your point, adding a separate field might work but would make for a very klunky solution (if both are set, which do I choose?).  For this particular use case, wouldn't we have enough bits to add a new enum flag?  Again, no big deal, I will likely drop the feature (it's not critical) but thought it'd be worth discussing.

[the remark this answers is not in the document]
Pokechu22: There is an unused value in the enum (0, 1, and 2 correspond to valid values, while 3 is invalid), but hardware testing indicated that its behavior matches clamp (0).

Leaving the separate `AddressMode` enum would work, but then there would be an enum value that's not implemented in the actual shaders (since it doesn't exist on console), which feels weird to me.

Source/Core/DolphinQt/Config/Graphics/HacksWidget.cpp
iwubcode: Thanks, I just saw this.  I can't speak for anyone else but I like this a lot better!

heikadog (commented):
If it helps for adding the new texture sampling method to the GameINIs, Rabbids Go Home (RGWX41, RGWE41, RGWJ41, RGWP41) also uses VP6, as shown by the response to my bug report a while back: https://bugs.dolphin-emu.org/issues/12559

RGW.ini should be the file to edit judging from the other ones.

Pokechu22:
I looked at the fifolog on that issue, and it doesn't seem to be using the VP6 codec specifically.  The rendering issues there also only happen at 2x IR or higher, and it's horizontal lines rather than vertical lines.  Red Steel 2 is probably the same based on [this issue](https://bugs.dolphin-emu.org/issues/8457).  Probably Tornado Outbreak (R6TEA4) and Dancing with the Stars (RD8E52) are codecs other than VP6 as well, but I don't have any way to test it.

This PR does seem to fix Rabbids Go Home, though, so I'll add it to the gameini with a separate note along the lines of "Fixes prerendered videos".

It's a bit of an organizational issue to blame all prerendered video issues on VP6, since different videos do get decoded in different ways.  In particular, it seems like this PR doesn't entirely fix Just Dance's videos, though it impacts them slightly.

Pokechu22:
I did a bit more testing, and it seems like this doesn't fully fix issues with Bink Video ([#12666](https://bugs.dolphin-emu.org/issues/12666)) on Mesa - though it does improve it in the majority of cases.  There are still problems with the Excite Truck video and with 640x473 test case on issue 12666.  The Just Dance videos (which also use Bink) also aren't perfect, but are consistently broken on all platforms (including the software renderer) now instead of having additional lines on some platforms.

I'm reasonably sure that every game using Bink is actually affected; there's a chance that some use Bink exclusively with videos that don't cause issues (e.g. some videos in Shrek the Third are fine, but not all of them), but it's easier to just add it for all games.  However, it'll take a while to go through all games on the list linked in that issue, and there are still more games that aren't on that list that will need to be added as people find them.

iwubcode:
@Pokechu22  - sorry I've been doing a zillion things lately so I haven't gotten back to this.  Are you wanting to handle the Blink issues in a separate PR or do you plan to update this one?

Pokechu22:
I'm going to try to handle GameINI updates in this PR (it's already been started, but there are a lot more games), but the remaining graphical issues will probably be a separate PR.

iwubcode:
Alright, just let me know when it's ready!

iwubcode:
It's possible my shaders are just borked but the first game I tried with custom textures, this seemed to make many things invisible:

rune-factory-invis

I was able to resolve this in one of a few ways:

* Switching off custom textures
* Turning on "Fast Texture Sampling"
* Crank up the resolution (at least fixed the HUD, the character was never resolved)

In this game, I only had used custom textures for a few elements.

To confirm, I checked another game:

last-story-opening

As you can see, it also has issues.

Pokechu22:
Thanks, I've confirmed that on my machine.  Not sure what causes it yet.

Side note, I started looking into updating GameINIs for all Bink games, and after a few hours I've only made it through the 'A's - a ton of developers used it, and I have to go through all games for that developer and look at the back of the box on gametdb to try and determine if Bink is being used.  So probably I'll just get rid of the commit updating GameINIs for Bink in this PR and instead handle it in separate PR.

Pokechu22:
Figured it out: it's this code:

https://github.com/dolphin-emu/dolphin/blob/4d1bd54917f097510e814d0ffcd625de6867fcc0/Source/Core/VideoCommon/TextureCacheBase.cpp#L989-L991

The shader then assumes that there are more levels of detail than actually exist, and invalid ones seem to result in an invisible texture.  The easiest workaround is to check the number of levels in the shader itself, but that requires using `textureQueryLevels` which [isn't available on all opengl versions](https://www.khronos.org/opengl/wiki/Sampler_(GLSL)#Texture_mipmap_retrieval).

Pokechu22:
Fixed.

I did notice one other issue: when using custom textures or a higher IR with D3D11 (but not D3D12 or other backends), the following is repeatedly logged:

```
26:56:473 VideoBackends\D3DCommon\Shader.cpp:129 W[Video]: ps_5_0 compilation succeeded with warnings:
C:\Users\Pokechu22\source\repos\dolphin\Source\Core\DolphinQt\Shader@0x000002222D5A6B80(102,11-51): warning X3203: signed/unsigned mismatch, unsigned assumed
C:\Users\Pokechu22\source\repos\dolphin\Source\Core\DolphinQt\Shader@0x000002222D5A6B80(135,10-40): warning X3556: integer divides may be much slower, try using uints if possible.
C:\Users\Pokechu22\source\repos\dolphin\Source\Core\DolphinQt\Shader@0x000002222D5A6B80(136,10-40): warning X3556: integer divides may be much slower, try using uints if possible.
```

pointing to these lines:
https://github.com/dolphin-emu/dolphin/blob/4802753298e28288290622f9bc76de45727be34f/Source/Core/VideoCommon/PixelShaderGen.cpp#L584
https://github.com/dolphin-emu/dolphin/blob/4802753298e28288290622f9bc76de45727be34f/Source/Core/VideoCommon/PixelShaderGen.cpp#L696-L698

I'm not sure where it's getting a signed/unsigned mismatch (as `coord` and `size` are both signed), and since `uv` is signed the division needs to be signed (making the sizes unsigned and using `uv.x * (size_s / native_size_s)` is less flexible for custom texture sizes).

I've also removed the Bink video GameINIs for now.  Of course it's possible to manually disable fast texture sampling (and in fact it's disabled by default currently) so the GameINIs are only needed to ensure correct behavior if the user has changed the setting.

iwubcode:
More testing.  Custom textures are fixed but can sometimes be incorrect when "Fast Texture Sampling" is turned off (see opening of Last Story)

Source/Core/DolphinQt/Config/Graphics/HacksWidget.cpp
```diff
@@ -276,6 +279,23 @@ void HacksWidget::AddDescriptions()
...
+      "Use the video backend's built-in texture sampling functionality instead of a manual "
+      "implementation.<br><br>"
+      "This can introduce graphical issues in some games on some GPUs (such as breaking "
+      "pre-rendered VP6 videos in some EA titles), but will generally (though not always) perform "
+      "better.<br><br>"
+      "Disabling this option uses Manual Texture Sampling, which more accurately handles level of "
+      "detail calculation, and can bypass GPU driver problems.<br><br>"
+      "Manual Texture Sampling also emulates the GameCube/Wii's unusual behavior when wrapping "
+      "textures with non-power-of-2 sizes, but this is disabled when using custom textures or when "
+      "using a higher internal resolution with Scaled EFB Copy enabled.  The accurate texture "
```
iwubcode, line 292: > The accurate texture wrapping logic is not known to fix any officially released game, but will perform slightly better than the version compatible with custom textures

This is confusing when you previously said:

> This can introduce graphical issues in some games on some GPUs (such as breaking pre-rendered VP6 videos in some EA titles), but will generally (though not always) perform better

Are you just saying that the custom texture code will be less performant than the non-custom texture code for the manual texture sampling?

Pokechu22:
> Custom textures are fixed but can sometimes be incorrect when "Fast Texture Sampling" is turned off (see opening of Last Story)

I don't have Last Story; can you give some screenshots and maybe a fifolog that I can test with?

Source/Core/DolphinQt/Config/Graphics/HacksWidget.cpp
Pokechu22: Yes.  The code when custom textures are not in use with manual texture sampling is both faster _and_ more accurate than the code when custom textures are in use with manual texture sampling, since actual hardware uses bitwise and (`&`) to implement texture wrapping, which is faster than modulo (`%`) but behaves weirdly when texture sizes aren't powers of 2.  With custom textures, the assumption that textures will be powers of 2 can be violated, so modulo is used instead, which is slightly slower.

On the other hand, actual game mods might violate the power of 2 size, but it's better for them to get the strange wrapping behavior because that's the behavior they would get on real hardware.  See [issue 12408](https://bugs.dolphin-emu.org/issues/12408) for an example.

Miksel12: Modulo arithmetic is also used when Scaled EFB Copy is enabled but shouldn't that only be necessary when the internal resolution is not a power of 2?

Source/Core/VideoCommon/VideoConfig.h
```diff
@@ -238,6 +239,16 @@ struct VideoConfig final
     return backend_info.bSupportsGPUTextureDecoding && bEnableGPUTextureDecoding;
   }
   bool UseVertexRounding() const { return bVertexRounding && iEFBScale != 1; }
+  bool ManualTextureSamplingWithHiResTextures() const
+  {
+    // Hi-res textures (including hi-res EFB copies, but not native-resultion EFB copies at higher
```
Miksel12, line 244:
```suggestion
    // Hi-res textures (including hi-res EFB copies, but not native-resolution EFB copies at higher
```

Source/Core/DolphinQt/Config/Graphics/HacksWidget.cpp
Pokechu22: Theoretically we could get away with using bitwise arithmetic with powers of 2 with scaled EFB copy, yes.  But there's also some rescaling that needs to happen so that texture coordinates are in the right range for the old and new textures, and that can't be skipped.  I think it'd be unnecessarily complicated to special-case the power-of-2 IRs (and special-casing them would also result in more cases where shaders need to be recompiled; currently that only happens when changing between 1x IR and other IRs or when enabling custom textures.)

Miksel12: That makes sense.

iwubcode:
Here's the screenshots that show the issue:

**Fast**

fast

**Manual**

manual

--

Will upload a fifo in a bit

iwubcode:
Fifo

[last_story_title2.zip](https://github.com/dolphin-emu/dolphin/files/7190686/last_story_title2.zip)

Pokechu22:
Fixed.  The main challenge is that both GLSL and HLSL make modulo when the dividend is negative and the divisor is positive undefined (GLSL: page 123 of [this](https://www.khronos.org/registry/OpenGL/specs/gl/GLSLangSpec.4.60.pdf), "Results are undefined if one or both operands are negative."; HLSL: [here](https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-operators#additive-and-multiplicative-operators), "The % operator is defined only in cases where either both sides are positive or both sides are negative.").  I knew that and had attempted to work around it by using `(x & 0x7fffffff) % size`, but that breaks if `x` is negative.  Suppose `size` was 5; then `0 % 5` is 0, and we want `-1 % 5` to be 4.  But `-1 & 0x7fffffff` is `0x7fffffff`, and `0x7fffffff % 5` is 2.  I couldn't find an easy workaround, but Wikipedia's [page on modulo mentions different definitions](https://en.wikipedia.org/wiki/Modulo_operation#Variants_of_the_definition), and the "floored division" definition is easy enough to implement and gives results that work (`FlooredModulo(-1, 5)` is 4).

Things seemed to work before because the only context in which I had tested non-power-of-2 sizes was with EFB copies, and games generally don't wrap them (and when they do, as with Battle for Bikini Bottom, they still don't have negative coordinates usually).

This change also resolves the signed-unsigned mismatch warning on D3D11 mentioned in https://github.com/dolphin-emu/dolphin/pull/9956#issuecomment-921151165 (though the two integer division warnings are still present).

Note that Dolphin actually loads some warnings for [that pack](https://forums.dolphin-emu.org/Thread-the-last-story-hd-texture-pack) about the textures not being powers of two (and also note that `tex1_256x256_2a6221813b7a0adf_14.png` is in both `Environments/Gurak Castle` and `Environments/Reptids Cave`):

```
13:13:231 VideoCommon\HiresTextures.cpp:115 E[Video]: One or more textures at path 'C:/Users/Pokechu22/Documents/Dolphin Emulator/Load/Textures/00000000' were already inserted
13:13:233 Core\Boot\Boot.cpp:539 N[BOOT]: Booting DFF: C:/Users/Pokechu22/Documents/last_story_title2.dff
13:13:337 VideoCommon\HiresTextures.cpp:345 E[Video]: Invalid custom texture size 1632x1224 for texture C:/Users/Pokechu22/Documents/Dolphin Emulator/Load/Textures/00000000/Environments/Reptids Cave/tex1_256x256_m_d1d7d990afdb608e_14.png. The aspect differs from the native size 256x256.
13:13:337 VideoCommon\HiresTextures.cpp:354 E[Video]: Invalid custom texture size 1632x1224 for texture C:/Users/Pokechu22/Documents/Dolphin Emulator/Load/Textures/00000000/Environments/Reptids Cave/tex1_256x256_m_d1d7d990afdb608e_14.png. Please use an integer upscaling factor based on the native size 256x256.
13:13:490 VideoCommon\HiresTextures.cpp:345 E[Video]: Invalid custom texture size 3264x2448 for texture C:/Users/Pokechu22/Documents/Dolphin Emulator/Load/Textures/00000000/Environments/Gurak Castle/tex1_256x256_2a6221813b7a0adf_14.png. The aspect differs from the native size 256x256.
13:13:491 VideoCommon\HiresTextures.cpp:354 E[Video]: Invalid custom texture size 3264x2448 for texture C:/Users/Pokechu22/Documents/Dolphin Emulator/Load/Textures/00000000/Environments/Gurak Castle/tex1_256x256_2a6221813b7a0adf_14.png. Please use an integer upscaling factor based on the native size 256x256.
13:13:822 VideoCommon\HiresTextures.cpp:345 E[Video]: Invalid custom texture size 3264x2448 for texture C:/Users/Pokechu22/Documents/Dolphin Emulator/Load/Textures/00000000/Environments/Reptids Cave/tex1_256x256_m_b1a186f501229fe3_14.png. The aspect differs from the native size 256x256.
13:13:822 VideoCommon\HiresTextures.cpp:354 E[Video]: Invalid custom texture size 3264x2448 for texture C:/Users/Pokechu22/Documents/Dolphin Emulator/Load/Textures/00000000/Environments/Reptids Cave/tex1_256x256_m_b1a186f501229fe3_14.png. Please use an integer upscaling factor based on the native size 256x256.
```

Dolphin still loads the pack despite those messages (though on the other hand it refuses to load mipmaps if they are not the right sizes).  Perhaps they should be converted to warnings?

iwubcode:
It's pretty close but there's likely a slight bug?  I'll have to look at the code and see if I can spot anything...

You can see the weird grey line in the stone texture now:

grey line


Also, the pack I was testing (I believe) with was this one:  https://forums.dolphin-emu.org/Thread-the-last-story-hd-hi-texture-pack-19-07-11

Source/Core/VideoCommon/PixelShaderGen.cpp
```diff
@@ -567,7 +567,42 @@ int4 readTextureLinear(in Texture2DArray tex, uint2 uv1, uint2 uv2, int layer, i
...
+  return dividend - divisor * int(floor(float(dividend) / divisor));
+}}
+
+uint WrapCoord(int coord, uint wrap, int size) {{
+  switch (wrap) {{
+    case {:s}:
+    default: // confirmed that clamp is used for invalid (3) via hardware test
+      return uint(clamp(coord, 0, size - 1));
+    case {:s}:
+      return uint(FlooredModulo(coord, size));  // coord % size
```
iwubcode, line 592: Should this be:

```
return uint(FlooredModulo(coord, size - 1));
```

?

Pokechu22: No, it is correct; `FlooredModulo` works in the half-open interval `[0, size)` (i.e. the closed interval `[0, size - 1]`) already.

Pokechu22:
Thanks, that pack explains some of the inconsistencies between what I was seeing compared to your sizes and screenshots (and a 768x768 texture is fairly reasonable for upscaling a 256x256 texture since it's 3x; there's no warnings logged about the texture sizes with that pack, though there is a duplicate file warning).

I'm unable to reproduce the line, though.  At least with the fifolog, everything looks fine for me.  Do the lines stay at the same thickness at higher IRs, or do they become 1px wide regardless of IR?  And what graphics backend and drivers are you using?

I also see a green line at the bottom of your screenshots which seems odd.  I'm not getting that with the fifoplayer.

iwubcode:
> I'm unable to reproduce the line, though. At least with the fifolog, everything looks fine for me. Do the lines stay at the same thickness at higher IRs, or do they become 1px wide regardless of IR?  And what graphics backend and drivers are you using?

I'm able to reproduce with the fifolog.  I use the textures from the pack `tex1_256x256_m_931f5deb219a5739_14` and its mips.  I was using D3D11 but I also tried Vulkan and saw the same issue.  The line does not seem to increase with IR but stays the same.

> I also see a green line at the bottom of your screenshots which seems odd. I'm not getting that with the fifoplayer.

Yes, does not show on fifoplayer which I guess could mean it is something XFB related?  I've never looked into it but there are weird issues like that in some games.

Source/Core/VideoCommon/PixelShaderGen.cpp
```diff
@@ -567,7 +567,42 @@ int4 readTextureLinear(in Texture2DArray tex, uint2 uv1, uint2 uv2, int layer, i
...
+    {
+      // This is slower, and doesn't result in the same odd behavior that happens on console when
+      // wrapping with non-power-of-2 sizes, but it's fine for custom textures to have non-console
+      // behavior.
+      out.Write(R"(
+// Both GLSL and HLSL produce undefined values when the modulo operator (%) is used with a negative
+// dividend and a positive divisor.  We want a positive value such that FlooredModulo(-1, 3) is 2.
+// We don't really care about what happens when the divisor is negative, so Euclidean division would
+// also work, but floored division is easier to implement.  Note that when both the dividend and
+// divisor are positive, FlooredModulo(dividend, divisor) is euqal to dividend % divisor.
```
Tilka, line 580: typo: `euqal`

Data/Sys/GameSettings/GRZ.ini
```diff
@@ -0,0 +1,22 @@
+# GRZJ13 - Medal of Honor: Rising Sun
```
Rumi-Larry, line 1: Why is this ini file separate from the one above?

JosJuice: The game was released under this game ID in Japan for some reason.

Data/Sys/GameSettings/GW7.ini
```diff
@@ -12,7 +12,12 @@
 [ActionReplay]
```
Rumi-Larry, line 12: The "(tm)" can be removed on the title of the game, here.

JosJuice: That isn't especially important, especially considering it's a pre-existing title.

Pokechu22:
I can't reproduce the line with Nvidia or Intel; it also doesn't happen with Mesa (in WSL following [these instructions](https://github.com/gnsmrky/wsl-vulkan-mesa)).

I've pushed something that uses the native modulo function for floats, which may or may not fix it on AMD.  If it does work I'll tidy it up further.

iwubcode:
@Pokechu22 - that's surprising...

Regardless, it is now fixed on D3D11 but **not** Vulkan.  I also tested D3D12 and OpenGL, both were fine as well.

Pokechu22:
It happening with Vulkan but not OGL is weird.

Can you try out the textures and fifolog in [Last_story_test.zip](https://github.com/dolphin-emu/dolphin/files/7214405/Last_story_test.zip), and take a screenshot at 8x IR?  It'd help with figuring out where the line is getting its color from (hopefully).  (I recommend uploading the screenshot to GitHub, but then removing the `!` before the image so it renders as a link instead of an inline image.)

iwubcode:
Here you go:

[Screenshot](https://user-images.githubusercontent.com/15224722/134454181-e3da43c8-2c4f-4c3b-98b5-bfd703f0c734.png)

Seems like it isn't drawing the texture all the way?

Pokechu22:
Comparing that with the [result on my machine](https://user-images.githubusercontent.com/8334194/134565569-a6acec6e-d8af-4b88-8080-d871f265ade1.png), it seems like:

- The line is 1 pixel wide in the texture (it's about the same width as the random noise that I happened to get in it due to how I generated the texture)
- The line only appears on one axis (horizontally, when it transitions from purple to yellow or black to blue, not vertically when it transitions from yellow to black or blue to purple)
- Texture linear interpolation works properly on the line (it transitions from purple to grey, not purple to yellow with a sudden shift to grey)
- The line appears a second time in the background, so it's not related to a transition from positive to negative texture coordinates, and probably not related to mipmap levels
- The lines are the only significant difference; the textures aren't otherwise shifted (though there are some minor differences, such as one of the rocks in the distance having one extra pixel - that's probably not important).

As another test case, I've disabled texture filtering and added bounds-checking for textures (producing different colors if `u` is too high or too low).  Can you use the pack in [Last_story_test_2.zip](https://github.com/dolphin-emu/dolphin/files/7220358/Last_story_test_2.zip) with the fifolog from before, and get another 8x IR screenshot?

iwubcode:
Is this what you wanted?

[screenshot2](https://user-images.githubusercontent.com/15224722/134607430-60164d04-b920-4a6d-8dea-1f9ca7843489.png)

Pokechu22:
Yep, that's informative... the magenta lines indicate that `u >= size_s`, which is reasonable (though wrong).  The magenta shapes in the background are unexpected, though - do you still get them when "load custom textures" is disabled?

(For reference, [this](https://user-images.githubusercontent.com/8334194/134609078-40223cc1-f1fb-4ce6-8567-8883c6a93199.png) is what that test looks like on my machine.  The main difference is the absence of magenta stuff, but there are also some different patterns on the walls - not sure what the deal with those is.)

iwubcode:
Here's the old build with no custom textures:

[previous_build_no_custom](https://user-images.githubusercontent.com/15224722/134618557-6bb1582e-4bd1-4570-920b-4b0909fac58a.png)

Here's the new build:

[new_build](https://user-images.githubusercontent.com/15224722/134618628-577c507c-9023-42f8-ad52-c301d6fdac88.png)

and finally, if you're talking about the 'pink dots', I used FreeLook to zoom in on them:

[pink-dots](https://user-images.githubusercontent.com/15224722/134618877-3ea973eb-88e3-44b8-83f8-24316017811f.png)

They do not occur if I turn on "Fast Texture Sampling" or switch to "D3D".  (all my tests have been with vulkan fyi)

Pokechu22:
I was actually referring to stuff near the torch in the background, but the dots on the ground are also concerning (especially since they too show up without custom textures).

It's odd that that torch seems to differ between each capture, since this fifolog should have the same results on all frames:
<table>
<tr><th width="30%">screenshot2</th><th width="30%">previous_build_no_custom</th><th width="30%">new_build</th></tr>
<tr><td></td><td></td><td></td></tr>
</table>

I removed bloom by recording the hardware fifoplayer, so there might be something stupid from that.  But even then, I don't know why it'd behave that way.  Similar stuff seems to be happening on mesa fifoci, which is surprising (see [this diff between two versions of this PR](https://fifo.ci/version/8e397099a5c101dc9c9d5919987b1c93b4c08d32/ - smg-mmg and tsp3-pinkgrass are more significant examples).  Maybe that only happened due to me disabling stuff for test purposes, but I'm not sure.  (Radeon ubershader fifoci also broke [due to `std::length_error`](https://dolphin.ci/#/builders/24/builds/1163) which is its own kind of weirdness...).  I can reproduce the mesa issue on my machine with WSL+lavapipe, so I don't think I need you to do more experimentation with it for the moment.

iwubcode:
@Pokechu22 - it is a lot better.  The previous seams are gone.  Unfortunately there seems to be some "corruption" which is visible on the text/icons in the intro and the flame when starting, it only occurs on Vulkan (D3D11 is fine).  It doesn't occur if I turn off "Fast Texture Sampling".  Interestingly, resolution does effect it:

[Screenshot 3x](https://user-images.githubusercontent.com/15224722/136296482-e840b376-c008-48d8-833c-568c314401e7.png)
[Screenshot 5x](https://user-images.githubusercontent.com/15224722/136296698-d302b319-0b27-4e11-9f85-3f30b6253f86.png)
[Screenshot 6x](https://user-images.githubusercontent.com/15224722/136296758-972d082e-0cc3-4a44-9c5c-34b92d5bde7c.png)

Pokechu22:
Alright, can you record the fifolog I mentioned, and then while playing it back, increase the "from" object range (i.e. bring it closer to "to", so that fewer objects are rendered in the background)?  Mainly to determine if this is a depth-related issue or something else.

Also, does the issue happen when the cave is visible in the background, or only when there's a white screen? If it happens when the cave is visible, is the corrupt part of the text white, or transparent?  And does the corruption also affect space outside of where the text should be, or only where there was originally text?

iwubcode:
> Mainly to determine if this is a depth-related issue or something else.

Doesn't seem to be related to depth

[Screenshot](https://user-images.githubusercontent.com/15224722/136301336-4d4e0ace-1967-4264-826b-54b5f57b34e4.png)

> If it happens when the cave is visible, is the corrupt part of the text white, or transparent? And does the corruption also affect space outside of where the text should be, or only where there was originally text?

Not just the text that is impacted.  The fire in the background has visible black pixels.  I can't seem to capture that.

Here's the title screen though:

[Screenshot](https://user-images.githubusercontent.com/15224722/136301532-cbad8d0e-d07c-48b4-9bd1-3b98579f207d.png)

Here's another screenshot with fast sampling on:

[SLSEXJ_2021-10-06_19-25-47](https://user-images.githubusercontent.com/15224722/136301646-2370a21b-59ea-4935-ae0b-271bd3c1a393.png)

Pokechu22:
Since I don't have access to a working machine with an AMD GPU (my old laptop is pretty much dead from hardware failure, and no, I can't test this on my Wii :P), I don't have a good way of reproducing.

It seems like it's not just the text that's affected - there are also black splotches all over the walls.  Compare [this](https://user-images.githubusercontent.com/8334194/136716797-7c0ecb75-d224-41a7-a7b7-1493911b26f0.png) with [this](https://user-images.githubusercontent.com/8334194/136716805-cb5e6aac-97d5-4770-8915-de919f2ba73e.png) (both of these are your images).  The text is actually slightly transparent (which can be seen with freelook if you move in front of blackness and then one of the bright areas where light is shining), so it's possible that only the walls are affected (but it's also possible that both the walls and the text are affected).

Looking in earlier comments, the walls didn't have splotches in the first test (https://github.com/dolphin-emu/dolphin/pull/9956#issuecomment-925498592 / https://github.com/dolphin-emu/dolphin/pull/9956#issuecomment-926091522 / 95d4564e3b1e9695dd6b663177b280d44641f1da) but did in the second test (https://github.com/dolphin-emu/dolphin/pull/9956#issuecomment-926289051 / https://github.com/dolphin-emu/dolphin/pull/9956#issuecomment-926296754 / 862dd41600bc08b889e4a6a049b8ccfa57d1e0fa), though it's fairly subtle.  The only new code is what was added in 862dd41600bc08b889e4a6a049b8ccfa57d1e0fa (disabling texture filtering and adding some out-of-bounds stuff), but it also [was rebased ontop of a newer version of Dolphin](https://github.com/dolphin-emu/dolphin/compare/95d4564e3b1e9695dd6b663177b280d44641f1da...862dd41600bc08b889e4a6a049b8ccfa57d1e0fa) ([changes](https://github.com/dolphin-emu/dolphin/compare/95d4564e3b1e9695dd6b663177b280d44641f1da...0d8ad5f53a213bd8fb9a201eff42e73e80b540dd), though none seem relevant).  I'm not really sure why this would behave differently.

It also seems like there's _still_ a single-pixel seam on the wall.  I can reproduce that locally, and it seems to only happen with the custom textures.  I'm investigating that right now.

iwubcode:
> Unfortunately as that article mentions, the coarse/fine versions are only available with OpenGL 4.5 while Dolphin still technically supports OpenGL 3.3

Yikes, 3.3 is really old...

I feel like having different backends potentially behave different ways completely contradicts the desire of our 'VideCommon' rewrite.  Still, I can appreciate wanting to support old systems.  Therefore I'd request at least D3D be updated to be explicitly 'coarse' and then maybe some checks for Vulkan/OGL to see if they support 'coarse' or fallback to the un-explicit version with a warning.

> That's unexpected, as in the first build (95d4564 which did not have the issue) texture filtering was enabled but not forced, while the in second build (862dd41 which did have the issue) texture filtering was force-disabled.

Yes, I plan on trying to get back to your original build which doesn't have the black spots and slowly introduce your changes until I deduce the code causing it.  The difficult part is I started by only testing D3D, so it's possible I didn't notice this originally.  The LOD logic seems likely (given that the issue is affected by resolution) but commenting it out had no effect.

>  That might cause problems if the game says no mipmaps should be used on a texture in some cases and allows them in others, but that's not what's happening here.

Huh, I didn't notice that.  Also, I believe last night I confirmed that this occurs even without custom textures but I will double check.

> I recommend trying to use RenderDoc to debug further

Not sure how much will be gleaned from RenderDoc.  The Vulkan debugging is ok at best.  But I will try.  Thanks!

Pokechu22:
I've changed the default to Fast Texture Sampling and moved the option for Manual Texture Sampling to the advanced tab (under experimental).  This is a separate commit so that it can be reverted when we want to change the default.

I've also removed the gameini updates for now.  They can be added after further testing occurs.

Pokechu22:
The differences with the radeon builders are caused by 6bc42981c0d21676c072378a84a2002d0b229f97 &mdash; I think it's something to do with `a / x` and `a * (1 / x)` not always having the same results under floating-point math.  Given that that issue is fixed with manual texture sampling, I don't think it's that much of a concern.

JMC47:
Is this essentially done now, then?  With (almost) no effect when the setting is off.

iwubcode:
> That's a good point. Unfortunately as that article mentions, the coarse/fine versions are only available with OpenGL 4.5 while Dolphin still technically supports OpenGL 3.3, though I'm already using textureQueryLevels which is OpenGL 4.3 (but that's only on the custom textures path). That article also mentions that the coarse version isn't available on Metal (which presumably means that MoltenVK can't use implement it properly). But using coarse derivatives would more-accurately match the way the software renderer implements it (though I'm not sure if that's necessarily how hardware does it).

So I've been doing some other work and I too need OGL 4.3+ features.  I'm currently working around that with extensions and leveraging the backend info / opengl specific config.

What you could do is add either an option to the backend implementation.  Something like `bSupportsCoarseAndFineDerivatives`.  If it's false, just generate the old behavior (maybe with a warning?).  If it's true, use the new implementation.  The opengl version would dictate which functionality is used.  This would ensure correct behavior if the version supports it.

Pokechu22:
I've added handling of the absence of both `dFdxCoarse` and `textureQueryLevels`.  These are currently in two separate commits for ease of review, but I can squash them if wanted.

I'm currently checking for support of those functions like this:

https://github.com/dolphin-emu/dolphin/blob/d49d35fe43c9af5a4cef8975b3bdd7b94021359e/Source/Core/VideoBackends/OGL/OGLRender.cpp#L486-L489

I'm not sure if I need to check both the version and the extension; the extension documentation for [GL_ARB_derivative_control](https://www.khronos.org/registry/OpenGL/extensions/ARB/ARB_derivative_control.txt) and [GL_ARB_texture_query_levels](https://www.khronos.org/registry/OpenGL/extensions/ARB/ARB_texture_query_levels.txt) doesn't say anything about version support, while the [OpenGL 4.3 spec](https://www.khronos.org/registry/OpenGL/specs/gl/glspec43.core.pdf) says on page 680:

> G.3.3.130 Texture Query Levels
> The name string for texture query levels is `GL_ARB_texture_query_levels`.
> This extension is equivalent to new core functionality introduced in OpenGL 4.3,
> and is provided to enable this functionality in older drivers.

It's not clear to me whether that means that `GLExtensions;:Supports("GL_ARB_derivative_control")` will return true on 4.3 or not.  `GLExtensions::InitExtensionList` also seems to do some version-related stuff but it's not exactly clear what it's doing.

Source/Core/VideoCommon/RenderState.h
```diff
@@ -22,11 +22,18 @@ union RasterizationState
 {
   void Generate(const BPMemory& bp, PrimitiveType primitive_type);
 
-  RasterizationState& operator=(const RasterizationState& rhs);
+  RasterizationState() = default;
```
iwubcode, line 25: Not sure if we should be declaring the move semantics as well so that we can be explicit with expectations?

iwubcode: (same with `FramebufferState` )

Source/Core/VideoCommon/RenderState.h
```diff
@@ -22,11 +22,18 @@ union RasterizationState
...
+  RasterizationState(const RasterizationState&) = default;
+  RasterizationState& operator=(const RasterizationState& rhs)
```
iwubcode, line 27: I wonder if it'd be worthwhile to add `noexcept`

Pokechu22: Dolphin is built with exceptions disabled, isn't it?  Wouldn't `noexcept` be meaningless as such?

Source/Core/VideoCommon/RenderState.h
Pokechu22: I've added them, but they feel like a lot of boilerplate for something that doesn't actually behave differently when moved.

Source/Core/VideoCommon/RenderState.h
iwubcode: I'm being nitpicky here, it probably doesn't matter, I just thought I'd raise it as I was curious.  I've recently been getting into the habit of marking at least my constructors noexcept (assuming they are that way) just as a "good practice".

Stackoverflow seems to indicate that there's very minor gains for using `noexcept` and it only applies in very specific use cases.  Otherwise it's solely a worthwhile addition as a marker for the programmer (just like `const` ).

I'm fine leaving it as is, just thought it'd be a good debate :)

iwubcode (approved):
I can't speak to the emulation logic directly but the code LGTM.

Source/Core/DolphinQt/Config/Graphics/AdvancedWidget.cpp
```diff
@@ -266,6 +269,21 @@ void AdvancedWidget::AddDescriptions()
       "<br><br>May improve performance in some games which rely on CPU EFB Access at the cost "
       "of stability.<br><br><dolphin_emphasis>If unsure, leave this "
       "unchecked.</dolphin_emphasis>");
+  static const char TR_MANUAL_TEXTURE_SAMPLING_DESCRIPTION[] = QT_TR_NOOP(
+      "Use a manual implementation of texture sampling instead of the graphics backend's built-in "
+      "functionality.<br><br>"
```
JMC47, line 274: Users aren't that good at reading lots of things, so having such a long description might not be a good idea.  I suggest trying to break it down a bit.

"This setting can fix graphical issues in some games on certain GPUs, most commonly vertical lines on FMVs.  In addition to this, enabling manual texture sampling will allow for correct emulation of texture wrapping (at 1x IR or when scaled EFB is disabled) and better emulates Level of Detail calculation.

This comes at the cost of potentially worse performance, especially at higher internal resolutions, and Anisotropic Filtering is not compatible yet.

If unsure, leave this unchecked."

iwubcode: Don't have much to say about the new text but I'd make the last line:

`<dolphin_emphasis>If unsure, leave this unchecked.</dolphin_emphasis>`

Pokechu22: I've reworded this based on your suggestion.  Let me know how the new wording looks.

JMC47 (approved):

JMC47:
Tested and confirmed to work on Android.  I guess it's just about time.

CrusadingNinja:
With Manual Texture Sampling enabled, Dolphin seems to freeze when booting any game during the shader pre compilation process on my GTX 1070 when using D3D11 or D3D12. OpenGL and Vulkan are not affected.

JMC47:
This is probably because it needs to compile shaders.

CrusadingNinja:
> This is probably because it needs to compile shaders.

Same behavior even with shader pre compilation disabled.

JMC47:
I have a GTX 1070 as well and cannot reproduce the issue.

mbc07:
Although this PR didn't completely fix the vertical/horizontal lines issue on Just Dance 2 (they're less noticeable, though), it did eliminate all lines on Just Dance 1, 3 and derivates (Just Dance Wii 2, Just Dance Greatest Hits, etc.) when using a NVIDIA GPU. 

My question is, given that enabling "Manual Texture Sampling" has a minor impact on performance and that it's only needed for specific GPU drivers (mainly NVIDIA), should this setting be enabled by default on the Game INI of the affected games? 

(asking because I was about to push a small batch of Game INI updates and earlier Just Dance games are among this batch)

Pokechu22:
I plan on doing a larger batch of GameINI updates for all Bink Video and EA VP6 games, but it might be a while before I get around to it.  Those early Just Dance games do use Bink Video, but if you already plan on updating their GameINIs, it would make sense to enable it for them.

PR #9957, opened 2021-07-25
Jit: Perform BAT lookup in dcbf/dcbi/dcbst

JosJuice:
When PR #9314 fixed https://bugs.dolphin-emu.org/issues/12133, it did so by removing the broken address calculation entirely and always using the slow path. This caused a performance regression, https://bugs.dolphin-emu.org/issues/12477.

This pull request instead replaces the broken address calculation with a BAT lookup. If the BAT lookup succeeds, we can use the old fast path. Otherwise we use the slow path.

Intends to improve https://bugs.dolphin-emu.org/issues/12477.

@JMC47 Please test Happy Feet and some games with known performance problems.

Everyone else: I'm not entirely confident that I made no mistakes with the assembly, so don't let JMC merge this before it's been properly reviewed :)

### DIFF
--- a/Source/Core/Core/PowerPC/Jit64/Jit_LoadStore.cpp
+++ b/Source/Core/Core/PowerPC/Jit64/Jit_LoadStore.cpp
@@ -234,21 +234,37 @@ void Jit64::dcbx(UGeckoInstruction inst)
   JITDISABLE(bJITLoadStoreOff);
 
   X64Reg addr = RSCRATCH;
+  X64Reg value = RSCRATCH2;
   RCOpArg Ra = inst.RA ? gpr.Use(inst.RA, RCMode::Read) : RCOpArg::Imm32(0);
   RCOpArg Rb = gpr.Use(inst.RB, RCMode::Read);
-  RegCache::Realize(Ra, Rb);
+  RCX64Reg tmp = gpr.Scratch();
+  RegCache::Realize(Ra, Rb, tmp);
 
   MOV_sum(32, addr, Ra, Rb);
-  AND(32, R(addr), Imm8(~31));
 
+  // Check whether a JIT cache line needs to be invalidated.
+  LEA(32, value, MScaled(addr, SCALE_8, 0));  // addr << 3 (masks the first 3 bits)
+  SHR(32, R(value), Imm8(3 + 5 + 5));         // >> 5 for cache line size, >> 5 for width of bitset
+  MOV(64, R(tmp), ImmPtr(GetBlockCache()->GetBlockBitSet()));
+  MOV(32, R(value), MComplex(tmp, value, SCALE_4, 0));
+  SHR(32, R(addr), Imm8(5));
+  BT(32, R(value), R(addr));
+
+  FixupBranch c = J_CC(CC_C, true);
+  SwitchToFarCode();
+  SetJumpTarget(c);
   BitSet32 registersInUse = CallerSavedRegistersInUse();
   ABI_PushRegistersAndAdjustStack(registersInUse, 0);
   MOV(32, R(ABI_PARAM1), R(addr));
+  SHL(32, R(ABI_PARAM1), Imm8(5));
   MOV(32, R(ABI_PARAM2), Imm32(32));
   XOR(32, R(ABI_PARAM3), R(ABI_PARAM3));
   ABI_CallFunction(JitInterface::InvalidateICache);
   ABI_PopRegistersAndAdjustStack(registersInUse, 0);
   asm_routines.ResetStack(*this);
+  c = J(true);
+  SwitchToNearCode();
+  SetJumpTarget(c);
 }
 
 void Jit64::dcbt(UGeckoInstruction inst)
--- a/Source/Core/Core/PowerPC/Jit64/Jit_LoadStore.cpp
+++ b/Source/Core/Core/PowerPC/Jit64/Jit_LoadStore.cpp
@@ -240,31 +240,46 @@ void Jit64::dcbx(UGeckoInstruction inst)
   RCX64Reg tmp = gpr.Scratch();
   RegCache::Realize(Ra, Rb, tmp);
 
-  MOV_sum(32, addr, Ra, Rb);
+  // Translate effective address to physical address.
+  MOV_sum(32, value, Ra, Rb);
+  FixupBranch bat_lookup_failed;
+  if (MSR.IR)
+  {
+    MOV(32, R(addr), R(value));
+    bat_lookup_failed = BATAddressLookup(value, tmp);
+    AND(32, R(addr), Imm32(0x0001ffff));
+    AND(32, R(value), Imm32(0xfffe0000));
+    OR(32, R(value), R(addr));
+  }
+  MOV(32, R(addr), R(value));
 
   // Check whether a JIT cache line needs to be invalidated.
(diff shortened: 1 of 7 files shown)


JMC47:
Tested the afflicted games and they're fixed.  Happy Feet still works.

Source/Core/Core/PowerPC/JitCommon/JitCache.h
```diff
@@ -121,7 +109,19 @@ class ValidBlockBitSet final
...
+  ValidBlockBitSet()
+  {
+    m_valid_block.reset(new u32[VALID_BLOCK_ALLOC_ELEMENTS]);
+    ClearAll();
+  }
+
+  void Set(u32 bit) { m_valid_block[bit / 32] |= 1u << (bit % 32); }
+  void Clear(u32 bit) { m_valid_block[bit / 32] &= ~(1u << (bit % 32)); }
+  void ClearAll() { memset(m_valid_block.get(), 0, sizeof(u32) * VALID_BLOCK_ALLOC_ELEMENTS); }
+  bool Test(u32 bit) { return (m_valid_block[bit / 32] & (1u << (bit % 32))) != 0; }
```
iwubcode, line 124: I know this is just a move so maybe it should be reserved for a future PR but:

```
bool Test(u32 bit) const { return (m_valid_block[bit / 32] & (1u << (bit % 32))) != 0; }
```

?

Source/Core/Core/PowerPC/Jit64Common/EmuCodeBlock.cpp
```diff
@@ -91,6 +91,16 @@ void EmuCodeBlock::SwitchToNearCode()
   SetCodePtr(m_near_code, m_near_code_end, m_near_code_write_failed);
 }
 
+FixupBranch EmuCodeBlock::BATAddressLookup(X64Reg addr, X64Reg tmp)
+{
+  MOV(64, R(tmp), ImmPtr(&PowerPC::dbat_table[0]));
+  SHR(32, R(addr), Imm8(PowerPC::BAT_INDEX_SHIFT));
+  MOV(32, R(addr), MComplex(tmp, addr, SCALE_4, 0));
```
AdmiralCurtiss, line 98: This here compiles to

```
00000122AC983898  mov         rdi,offset PowerPC::dbat_table (07FF6F001F9F0h)  
00000122AC9838A2  shr         edx,11h  
00000122AC9838A5  mov         edx,dword ptr [rdi+rdx*4]  
```

Which seems to be not quite right; you need to clear the high bits of `rdx` somewhere so this is safe, I think? Unless that happens implicitly somewhere? Same thing happens again at the similar construct at `MOV(64, R(tmp), ImmPtr(GetBlockCache()->GetBlockBitSet()))`.

AdmiralCurtiss: Okay, apparently on x64 if you write into a 32-bit register the upper 32 bits of its corresponding 64 bit register get automatically cleared, so this is fine then.

Source/Core/Core/PowerPC/Jit64/Jit_LoadStore.cpp
```diff
@@ -234,21 +234,52 @@ void Jit64::dcbx(UGeckoInstruction inst)
...
+    OR(32, R(value), R(addr));
+  }
+  MOV(32, R(addr), R(value));
+
+  // Check whether a JIT cache line needs to be invalidated.
+  SHR(32, R(value), Imm8(5 + 5));  // >> 5 for cache line size, >> 5 for width of bitset
+  MOV(64, R(tmp), ImmPtr(GetBlockCache()->GetBlockBitSet()));
+  MOV(32, R(value), MComplex(tmp, value, SCALE_4, 0));
+  SHR(32, R(addr), Imm8(5));
+  BT(32, R(value), R(addr));
```
AdmiralCurtiss, line 261: The other thing I'm wondering here if this needs an extra `AND(32, R(addr), 0x1F)`? I suspect not as references online claim that `bt` does an implicit mod32 on the 2nd parameter, which seems logical, but curiously MSVC does emit this `and` in the bitset code, even in Release mode. Oh well, I think this is fine as-is.

AdmiralCurtiss:
The x64 implementation looks fine to me, can't say much about the ARM one.

Source/Core/Core/PowerPC/Jit64/Jit_LoadStore.cpp
phire: Yeah, x64 clears the upper bits of the register when you do a 32bit operation. 

This is fine.

leoetlino (approved):
The JitArm64 changes look fine to me

XiaomiRedmi9:
This will improve android performance?

Source/Core/Core/PowerPC/JitArm64/JitArm64_LoadStore.cpp
```diff
@@ -274,6 +274,19 @@ void JitArm64::SafeStoreFromReg(s32 dest, u32 value, s32 regOffset, u32 flags, s
   gpr.Unlock(ARM64Reg::W0, ARM64Reg::W1, ARM64Reg::W30);
 }
 
+FixupBranch JitArm64::BATAddressLookup(ARM64Reg addr_out, ARM64Reg addr_in, ARM64Reg tmp)
+{
+  tmp = EncodeRegTo64(tmp);
+
+  MOVP2R(tmp, PowerPC::dbat_table.data());
```
degasus, line 281: Do you think it is worth to place the BAT table somewhere within the PPC_State?

JosJuice: In that case we would need to have an addition instruction in here instead of the MOVP2R (since we wouldn't be putting the BAT table at the very beginning of PPCState). So we'd save maybe one instruction (if MOVP2R needs two instructions), which isn't all that much.

Source/Core/Core/PowerPC/JitArm64/JitArm64_LoadStore.cpp
```diff
@@ -539,34 +552,71 @@ void JitArm64::dcbx(UGeckoInstruction inst)
...
+  gprs_to_push[DecodeReg(WA)] = false;
 
   ABI_PushRegisters(gprs_to_push);
   m_float_emit.ABI_PushRegisters(fprs_to_push, ARM64Reg::X30);
 
-  MOVI2R(ARM64Reg::X1, 32);
-  MOVI2R(ARM64Reg::X2, 0);
-  MOVP2R(ARM64Reg::X3, &JitInterface::InvalidateICache);
-  BLR(ARM64Reg::X3);
+  MOVP2R(ARM64Reg::X8, &JitInterface::InvalidateICache);
```
degasus, line 603: The fallback now skips the `AND(addr, addr, LogicalImm(~31, 32));  // mask sizeof cacheline` call. Was this on purpose?

JosJuice: I checked the called C++ function, and it seems like it ignores the last 5 bits.

degasus: Perfect, Thanks a lot.

smurf3tte:
Very happy to see this go in! Thank you for taking care of the fallout of my earlier change.

PR #9959, opened 2021-07-25
Qt: GBA QoL improvements

Bonta0:
This adds features frequently requested by users to the GBA windows
- Borderless mode
- Always on top
- Interframe blending (see https://mgba.io/2020/01/21/mgba-0.8.0/#interframe-blending)
image
This also exposes Core features better to make a couple conditions in the context menu more accurate

### DIFF
--- a/Source/Core/Core/HW/GBACore.cpp
+++ b/Source/Core/Core/HW/GBACore.cpp
@@ -292,6 +292,25 @@ bool Core::IsStarted() const
   return m_started;
 }
 
+CoreInfo Core::GetCoreInfo() const
+{
+  CoreInfo info{};
+  info.device_number = m_device_number;
+  info.width = GBA_VIDEO_HORIZONTAL_PIXELS;
+  info.height = GBA_VIDEO_VERTICAL_PIXELS;
+
+  if (!IsStarted())
+    return info;
+
+  info.is_gba = m_core->platform(m_core) == mPlatform::mPLATFORM_GBA;
+  info.has_rom = !m_rom_path.empty();
+  info.has_ereader =
+      info.is_gba && static_cast<::GBA*>(m_core->board)->memory.hw.devices & HW_EREADER;
+  m_core->desiredVideoDimensions(m_core, &info.width, &info.height);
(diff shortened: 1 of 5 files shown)


JMC47:
@dolphin-emu-bot rebuild

JMC47 (approved):
Everything is tested to work.  Tested Animal Crossing and Nintendo Puzzle Collection for Interframe blending.  Tested fullscreen shenanigans for the always on top and borderless GBA.

WeedyWeedSmoker:
This commit breaks compatibility with macOS 10.13 (still works on the newer macOS 11 version, didn't test other macOSes…)

JMC47:
I can update the FAQ to change the minimum requirements.

JosJuice:
But I can't see anything in this PR that would cause macOS 10.13 to break... I don't think you should update the minimum requirements until the reason why it broke is understood.

WeedyWeedSmoker:
@JMC47 Let me test again, this specific commit now seems to work fine on 10.13, I'm checking which commit actually broke it

JMC47:
I just meant that as a nuclear option, we can raise the minimum requirements.  Obviously if it's something silly or unintended, we can try to fix it.

WeedyWeedSmoker:
@JMC47 @JosJuice 5.0-14748 crashes, while 5.0-14746 works fine… The culprit is between this commit (3e04cb6e6bd5e3606f07637711442130f4489bfb) and 7fe97b27de3cdcf6669f23f38971c889b5a1a3a2, I will try to compile those in-between commits to try and find the real culprit, while 46e331d000e0af8986189fb937f2b68f7dcb9bd1 seems to be a good candidate for a crash on macOS 10.13 somehow…

WeedyWeedSmoker:
@JMC47 @JosJuice I can't manage to compile Dolphin on macOS, finally got my Qt5 detected, but then got errors with `<optional>`, trying this : https://github.com/youknowone/xcode-libcxx but brew got certificate errors trying to download a required part of llvm…

Do you have a way to compile the aforementioned commits for mac ?

PR #9962, opened 2021-07-25
VideoBackendBase: Prefer Vulkan over OGL on macOS Mojave and newer

OatmealDome:
macOS's OpenGL support is terribly outdated - it's been stuck on OpenGL 4.1 since Mavericks (released in 2013!) and doesn't support key extensions like ``ARB_buffer_storage``. Plus, starting with macOS Mojave, OpenGL has been deprecated and may be removed in a future macOS.

Now that we have MoltenVK, it might be best to set the Vulkan backend as default for new users with Macs that run Mojave and newer. (High Sierra and older don't support some Metal features that Dolphin requires.)

### DIFF
--- a/Source/Core/VideoCommon/VideoBackendBase.cpp
+++ b/Source/Core/VideoCommon/VideoBackendBase.cpp
@@ -214,6 +214,10 @@ const std::vector<std::unique_ptr<VideoBackendBase>>& VideoBackendBase::GetAvail
     std::vector<std::unique_ptr<VideoBackendBase>> backends;
 
     // OGL > D3D11 > D3D12 > Vulkan > SW > Null
+    //
+    // On macOS Mojave and newer, we prefer Vulkan over OGL due to outdated drivers.
+    // However, on macOS High Sierra and older, we still prefer OGL due to its older Metal version
+    // missing several features required by the Vulkan backend.
 #ifdef HAS_OPENGL
     backends.push_back(std::make_unique<OGL::VideoBackend>());
 #endif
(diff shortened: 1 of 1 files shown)


JosJuice:
@dolphin-emu-bot rebuild

Source/Core/VideoCommon/VideoBackendBase.cpp
```diff
@@ -222,8 +238,15 @@ const std::vector<std::unique_ptr<VideoBackendBase>>& VideoBackendBase::GetAvail
     backends.push_back(std::make_unique<DX12::VideoBackend>());
 #endif
 #ifdef HAS_VULKAN
+#ifdef __APPLE__
+    if (!is_vulkan_preferred)
+    {
+      backends.push_back(std::make_unique<Vulkan::VideoBackend>());
+    }
```
Warepire, line 245:
```suggestion
#ifdef __APPLE_
    if (__builtin_available(macOS 10.14, *))
    {
      backends.emplace(backends.begin(), std::make_unique<Vulkan::VideoBackend>());
    }
    else
#endif
    {
      backends.push_bach(std::make_unique<Vulkan::VideoBackend>());
    }
```

And get rid of all the other changes except the comment block? It makes the change much more self-contained, but it does hide intent a bit. Not sure what is best way.

OatmealDome: I think this is definitely cleaner than how I did it and it doesn't seem to hide intent *that* much. Perhaps an extra comment will help with that.

XiaomiRedmi9:
Can you also do this on android?
I prefer vulkan over opengl in terms of performance and battery life of my phone.

JMC47:
For most android users, Vulkan performs worse so no.

XiaomiRedmi9:
> For most android users, Vulkan performs worse so no.

Ah ok :)
But maybe someday when vulkan is optimize on android...

nickp-ca:
For MacOS, what is criteria for pulling recent Khronos maintained source for MoltenVK, this is now showing 2 months stale with the libvulkan.dylib in Dolphin - a few key areas have been corrected specific to apple silicon use - https://github.com/KhronosGroup/MoltenVK is shadowing 1.2.182 of Vulkan whereas Dolphin is still.  Since this is a work in progress from contributors to that program and issues are being addressed, is there a better methodology to rebuild the dynamic library on a more regular cadence?

OatmealDome:
@nickp-ca

As far as I'm aware, there is no specific criteria. The last update happened because the older MoltenVK version was not compatible with Apple Silicon.

I hope to submit a PR soon which will remove the precompiled binary and integrate MoltenVK with Dolphin's CMake build system as an external project. This should make it easier to update in the future.

leoetlino (approved):
On macOS this seems like a no brainer considering how badly supported OpenGL is.

PR #9964, opened 2021-07-26
PowerPC: Implement broken masking for uncached unaligned writes

JosJuice:
This implements the behavior described in https://bugs.dolphin-emu.org/issues/12565.

The implementation requires writes to pass through slowmem in order to be affected. Since most users probably don't care about emulating this behavior and thus wouldn't want performance to get lower just for the sake of emulating it, I've added a new INI-only setting `AlignmentQuirks` which controls whether slowmem is forced for uncached memory.

Thank you to eigenform, delroth, phire, marcan, segher, and Extrems for all helping in one way or another with the efforts to reverse engineer this behavior, and to Rylie for reporting the issue.

### DIFF
--- a/Source/Core/Core/PowerPC/JitArm64/JitArm64_BackPatch.cpp
+++ b/Source/Core/Core/PowerPC/JitArm64/JitArm64_BackPatch.cpp
@@ -197,18 +197,16 @@ void JitArm64::EmitBackpatchRoutine(u32 flags, bool fastmem, bool do_farcode, AR
     }
     else if (flags & BackPatchInfo::FLAG_STORE)
     {
-      ARM64Reg temp = ARM64Reg::W0;
-      temp = ByteswapBeforeStore(this, temp, RS, flags, false);
-      if (temp != ARM64Reg::W0)
-        MOV(ARM64Reg::W0, temp);
+      const bool reverse = (flags & BackPatchInfo::FLAG_REVERSE) != 0;
 
       if (flags & BackPatchInfo::FLAG_SIZE_32)
-        MOVP2R(ARM64Reg::X8, &PowerPC::Write_U32);
+        MOVP2R(ARM64Reg::X8, reverse ? &PowerPC::Write_U32_Swap : &PowerPC::Write_U32);
       else if (flags & BackPatchInfo::FLAG_SIZE_16)
-        MOVP2R(ARM64Reg::X8, &PowerPC::Write_U16);
+        MOVP2R(ARM64Reg::X8, reverse ? &PowerPC::Write_U16_Swap : &PowerPC::Write_U16);
       else
         MOVP2R(ARM64Reg::X8, &PowerPC::Write_U8);
 
+      MOV(ARM64Reg::W0, RS);
       BLR(ARM64Reg::X8);
     }
     else if (flags & BackPatchInfo::FLAG_ZERO_256)
--- a/Source/Core/Core/PowerPC/MMU.cpp
+++ b/Source/Core/Core/PowerPC/MMU.cpp
@@ -11,6 +11,7 @@
 #include "Common/Assert.h"
 #include "Common/BitUtils.h"
 #include "Common/CommonTypes.h"
+#include "Common/Logging/Log.h"
 
 #include "Core/ConfigManager.h"
 #include "Core/HW/CPU.h"
@@ -277,6 +278,8 @@ static void WriteToHardware(u32 em_address, const u32 data, const u32 size)
     return;
   }
 
+  bool wi = false;
+
   if (!never_translate && MSR.DR)
   {
     auto translated_addr = TranslateAddress<flag>(em_address);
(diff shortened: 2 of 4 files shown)


JosJuice:
The code for raising the interrupt doesn't seem to be working properly, at least not when running my hardware test. But everything else should be working.

MrCheeze:
Trying this out in Ocarina of Time. Seems like it just crashes in the situation where duplicate writes happen on console now. The below screenshots are 1 breakpoint step apart:

https://cdn.discordapp.com/attachments/85238846538674176/869361246649057280/unknown.png
https://cdn.discordapp.com/attachments/85238846538674176/869361306787012608/unknown.png

phire:
How much performance would it cost to enable this feature by default?

If the cost is low (less than 1% in most games) it might be worth enabling it by default. Do games even do that many uncached writes by default?

Source/Core/Core/PowerPC/MMU.h
```diff
@@ -199,18 +199,21 @@ constexpr int BAT_INDEX_SHIFT = 17;
...
 extern BatTable ibat_table;
 extern BatTable dbat_table;
-inline bool TranslateBatAddess(const BatTable& bat_table, u32* address)
+inline bool TranslateBatAddess(const BatTable& bat_table, u32* address, bool* wi)
 {
   u32 bat_result = bat_table[*address >> BAT_INDEX_SHIFT];
   if ((bat_result & BAT_MAPPED_BIT) == 0)
     return false;
   *address = (bat_result & BAT_RESULT_MASK) | (*address & (BAT_PAGE_SIZE - 1));
+  *wi = bat_result & BAT_WI_BIT;
```
lioncash, line 213:
```suggestion
  *wi = (bat_result & BAT_WI_BIT) != 0;
```

Otherwise this is performing implicit truncation

Source/Core/Core/PowerPC/MMU.cpp
```diff
@@ -1511,10 +1578,12 @@ void IBATUpdated()
 template <const XCheckTLBFlag flag>
 static TranslateAddressResult TranslateAddress(u32 address)
 {
-  if (TranslateBatAddess(IsOpcodeFlag(flag) ? ibat_table : dbat_table, &address))
-    return TranslateAddressResult{TranslateAddressResult::BAT_TRANSLATED, address};
+  bool wi;
```
lioncash, line 1581:
```suggestion
  bool wi = false;
```

Even if it gets overwritten to ensure it always has an initialized value for its lifespan

Source/Core/Core/PowerPC/MMU.cpp
```diff
@@ -1424,19 +1472,38 @@ static void UpdateBATs(BatTable& bat_table, u32 base_spr)
...
-        else if (physical_address < Memory::GetRamSizeReal())
-          valid_bit |= BAT_PHYSICAL_BIT;
-        else if (Memory::m_pEXRAM && physical_address >> 28 == 0x1 &&
-                 (physical_address & 0x0FFFFFFF) < Memory::GetExRamSizeReal())
-          valid_bit |= BAT_PHYSICAL_BIT;
-        else if (physical_address >> 28 == 0xE &&
-                 physical_address < 0xE0000000 + Memory::GetL1CacheSize())
-          valid_bit |= BAT_PHYSICAL_BIT;
+
+        const bool wi = batl.WIMG & 0b1100;
```
lioncash, line 1480:
```suggestion
        const bool wi = (batl.WIMG & 0b1100) != 0;
```

Source/Core/Core/PowerPC/MMU.cpp
```diff
@@ -1252,6 +1297,7 @@ static TLBLookupResult LookupTLBPageAddress(const XCheckTLBFlag flag, const u32
       tlbe.recent = 1;
 
     *paddr = tlbe.paddr[1] | (vpa & 0xfff);
+    *wi = PTE2.WIMG & 0b1100;
```
lioncash, line 1300:
```suggestion
    *wi = (PTE2.WIMG & 0b1100) != 0;
```

Source/Core/Core/PowerPC/MMU.cpp
```diff
@@ -1230,16 +1273,18 @@ static TLBLookupResult LookupTLBPageAddress(const XCheckTLBFlag flag, const u32
       tlbe.recent = 0;
 
     *paddr = tlbe.paddr[0] | (vpa & 0xfff);
+    *wi = PTE2.WIMG & 0b1100;
```
lioncash, line 1276:
```suggestion
    *wi = (PTE2.WIMG & 0b1100) != 0;
```

Source/Core/Core/PowerPC/MMU.cpp
```diff
@@ -1037,7 +1077,8 @@ bool IsOptimizableGatherPipeWrite(u32 address)
   // Translate address, only check BAT mapping.
   // If we also optimize for TLB mappings, we'd have to clear the
   // JitCache on each TLB invalidation.
-  if (!TranslateBatAddess(dbat_table, &address))
+  bool wi;
```
lioncash, line 1080:
```suggestion
  bool wi = false;
```

Source/Core/Core/PowerPC/MMU.cpp
```diff
@@ -1015,7 +1054,8 @@ u32 IsOptimizableMMIOAccess(u32 address, u32 access_size)
   // Translate address
   // If we also optimize for TLB mappings, we'd have to clear the
   // JitCache on each TLB invalidation.
-  if (!TranslateBatAddess(dbat_table, &address))
+  bool wi;
```
lioncash, line 1057:
```suggestion
  bool wi = false;
```

Source/Core/Core/PowerPC/JitArm64/JitArm64_BackPatch.cpp
```diff
@@ -197,18 +197,16 @@ void JitArm64::EmitBackpatchRoutine(u32 flags, bool fastmem, bool do_farcode, AR
     }
     else if (flags & BackPatchInfo::FLAG_STORE)
     {
-      ARM64Reg temp = ARM64Reg::W0;
-      temp = ByteswapBeforeStore(this, temp, RS, flags, false);
-      if (temp != ARM64Reg::W0)
-        MOV(ARM64Reg::W0, temp);
+      const bool reverse = flags & BackPatchInfo::FLAG_REVERSE;
```
lioncash, line 200:
```suggestion
      const bool reverse = (flags & BackPatchInfo::FLAG_REVERSE) != 0;
```

Source/Core/Core/PowerPC/MMU.cpp
```diff
@@ -1368,6 +1414,8 @@ static TranslateAddressResult TranslatePageAddress(const u32 address, const XChe
         if (res != TLBLookupResult::UpdateC)
           UpdateTLBEntry(flag, PTE2, address);
 
+        *wi = PTE2.WIMG & 0b1100;
```
lioncash, line 1417:
```suggestion
        *wi = (PTE2.WIMG & 0b1100) != 0;
```

delroth:
> Thank you to eigenform, delroth, phire, marcan, and Extrems for all helping in one way or another with the efforts to reverse engineer this behavior.

Don't forget segher who spent like 4h explaining to me how the 60x bus works on IRC :-)

JosJuice:
> How much performance would it cost to enable this feature by default?
>
> If the cost is low (less than 1% in most games) it might be worth enabling it by default. Do games even do that many uncached writes by default?

@JMC47, would you like to test? (My laptop with its thermal throttling is probably not the best for testing small performance differences)

Something I want to note is that I plan to revive and fix PR #9865 at some point after this PR has been merged, and I think it would be natural to put the behavior implemented in that PR under the same `AlignmentQuirks` setting. In addition to needing slowmem for write-through/cache-inhibited memory, that PR also adds an extra branch to the fastmem implementation of memory accesses that cause exceptions when misaligned regardless of the type of memory (most notably all float loadstores). This will have an additional impact on the performance.

JosJuice:
After some discussion and some quick testing of how often games access uncached memory (100-1000 reads per frame and essentially no writes, it seems like), I've decided to remove the setting for now. Uncached reads and writes will now always go through slowmem.

JMC47:
Seems like everyone is in agreement.  Just be sure to watch this for potential performance regressions down the line.

JosJuice:
Noticed a typo in the third commit message just now. It should say "upper 16 bits" instead of "upper 32 bits", in case anyone finds this in the future and wonders about it.

PR #9965, opened 2021-07-27
IOS HLE: Load MEM2 sections from IOS binaries into memory

leoetlino:
While we currently do not LLE the Starlet, we should also load the MEM2
sections into memory because some titles (notably the Homebrew Channel)
require a valid IOS to be loaded.

Untested (beyond checking that Wii games do not crash).


leoetlino:
The kernel sections are still missing, and those are apparently what the (non-open) HBC checks.

PR #9967, opened 2021-07-28
Remove note about anisotropic filtering requiring edge LOD

Pokechu22:
This was added because YAGCD's info on MAXANISO (near TX_SETMODE0 in Section [5.11.1](http://hitmen.c02.at/files/yagcd/yagcd/chap5.html#sec5.11.1)) claims it's the case, but Extrems says it does work.  I haven't tested anything myself, and dolphin still does not actually implement anisotropic filtering based on this field.

### DIFF
--- a/Source/Core/VideoCommon/BPMemory.h
+++ b/Source/Core/VideoCommon/BPMemory.h
@@ -765,7 +765,7 @@ enum class MaxAnsio
 template <>
 struct fmt::formatter<MaxAnsio> : EnumFormatter<MaxAnsio::Four>
 {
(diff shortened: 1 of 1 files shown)


Source/Core/VideoCommon/BPMemory.h
```diff
@@ -765,7 +765,7 @@ enum class MaxAnsio
 template <>
 struct fmt::formatter<MaxAnsio> : EnumFormatter<MaxAnsio::Four>
```
leoetlino, line 766: something that you might want to fix in a follow-up PR: MaxAn**is**o seems to be misspelled

Pokechu22: Thanks; I'll add a fix for that to #9718.

PR #9968, opened 2021-07-28
DSPHLE: add support for dynamic range compression

Tilka:
The CPU-side AX library enables it by default and uses hardcoded parameters. CMD_COMPRESSOR_TABLE_ADDR (0x0A) was incorrect. It's always a nop on the GameCube and was probably confused with the Wii version.

Triggers in Pac-Man World Rally.


Source/Core/Core/HW/DSPHLE/UCodes/AX.cpp
```diff
@@ -508,6 +507,50 @@ void AXUCode::SetMainLR(u32 src_addr)
...
+    // one attack frame based on previous frame
+    table_offset = m_compressor_pos * frame_byte_size;
+    // next frame will start release
+    m_compressor_pos = release_frames;
+  }
+  else if (m_compressor_pos)
+  {
+    // I'm a bit confused that they do "dec $ACC0; sr @pos, $AC0.M".
+    // Seems like it should have been decm instead of dec? Very strange.
+    --m_compressor_pos;
```
BhaaLseN, line 537: That seems weird, it would decrement a part of the accumulator that isn't used in the store. And also means that we _shouldn't_ decrement here if it is indeed correct.

Maybe theres a bug in DSPSpy (or whatever you used for disassembly of the UCode)?

Tilka: Mystery solved: The reason it works anyway is that the lower bits are zero, so e.g. `0x00'1234'0000 - 1 = 0x00'1233'ffff`.

PR #9969, opened 2021-07-28
GameINI: Add Safe Texture Cache to Inazuma Eleven GO: Strikers 2013

JMC47:
Certain text characters do not render properly with any other texture cache setting.  We tested 2048 and 512, and that was not enough to completely remedy the issue.


JosJuice:
I thought this applied to non-Latin characters as well? LGTM other than that.

JMC47:
Latin characters don't work with 2048.

JMC47:
Editing since it doesn't seem 100% clear.

AidenSonic:
Thanks !

PR #9970, opened 2021-07-28
Jit64: Fix fmaddXX with accurate NaNs

JosJuice:
So it turns out you have to pass `XMM0` as the clobber register to `HandleNaNs`, because `HandleNaNs` uses `BLENDVPD` and `BLENDVPD` implicitly uses `XMM0`, and nobody noticed when I broke this in 2c38d64 because nobody plays the one game that needs accurate NaNs.

### DIFF
--- a/Source/Core/Core/PowerPC/Jit64/Jit_FloatingPoint.cpp
+++ b/Source/Core/Core/PowerPC/Jit64/Jit_FloatingPoint.cpp
@@ -257,7 +257,7 @@ void Jit64::fp_arith(UGeckoInstruction inst)
       avx_op(avxOp, sseOp, dest, Rop1, Rop2, packed, reversible);
     }
 
-    HandleNaNs(inst, Rd, dest);
+    HandleNaNs(inst, Rd, dest, XMM0);
     if (single)
       FinalizeSingleResult(Rd, Rd, packed, true);
     else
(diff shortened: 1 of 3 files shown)


Source/Core/Core/PowerPC/Jit64/Jit_FloatingPoint.cpp
```diff
@@ -307,7 +307,7 @@ void Jit64::fmaddXX(UGeckoInstruction inst)
   // is created on a CPU that has FMA instructions and then gets played back on a CPU that doesn't.
   // (Or if the user just really wants to override the setting and knows how to do so.)
   const bool use_fma = Config::Get(Config::SESSION_USE_FMA);
-  const bool software_fma = use_fma && !cpu_info.bFMA;
+  const bool software_fma = use_fma;
```
AdmiralCurtiss, line 310: Uuh I assume this was left in form debugging?

JosJuice: Oh oops, yes

iwubcode:
Which game needs this behavior?  Could that be put into the commit message and/or as a comment?

JosJuice:
It is Dragon Ball: Revenge of King Piccolo. This is mentioned in a comment in the implementation of `HandleNaNs`.

JMC47:
This fixes accurate NaNs when enabled.

lioncash (approved):

PR #9972, opened 2021-07-29
Recompile shaders when 'Enable API Validation Layers' is toggled

Pokechu22:
This causes shaders to be reloaded when 'Enable API Validation Layers' is toggled.  It also does some general fixing and cleanup of related code.


iwubcode (approved):
Code wise LGTM.  Untested.

PR #9973, opened 2021-07-29
Jit: Use accurate negation order for FMA instructions

JosJuice:
It was believed that this only mattered when the rounding mode was set to round to infinity, which games generally don't do, but it can also affect the sign of the output when the inputs are all zero.

### DIFF
--- a/Source/Core/Core/PowerPC/Jit64/Jit_FloatingPoint.cpp
+++ b/Source/Core/Core/PowerPC/Jit64/Jit_FloatingPoint.cpp
@@ -421,7 +421,7 @@ void Jit64::fmaddXX(UGeckoInstruction inst)
     }
     else
     {
-      if (single && round_input)
+      if (round_input)
         Force25BitPrecision(result_xmm, Rc, scratch_xmm);
       else
         MOVAPD(result_xmm, Rc);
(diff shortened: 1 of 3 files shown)


JMC47:
Fixes issue https://bugs.dolphin-emu.org/issues/12297

JMC47:
Performance doesn't seem to be any different from what I can tell.

Source/Core/Core/PowerPC/Jit64/Jit_FloatingPoint.cpp
```diff
@@ -345,13 +345,18 @@ void Jit64::fmaddXX(UGeckoInstruction inst)
     RegCache::Realize(Ra, Rb, Rc, Rd);
   }
 
-  X64Reg scratch_xmm = !use_fma && inst.SUBOP5 == 30 ? XMM1 : XMM0;
-  X64Reg result_xmm = scratch_xmm == XMM0 ? XMM1 : XMM0;
+  const bool subtract = inst.SUBOP5 == 28 || inst.SUBOP5 == 30;  // nsub, nmsub
```
booto, line 348: // **msub**, nmsub

Source/Core/Core/PowerPC/Jit64/Jit_FloatingPoint.cpp
```diff
@@ -399,116 +404,79 @@ void Jit64::fmaddXX(UGeckoInstruction inst)
...
+  if (negate)
+    XORPD(result_xmm, MConst(packed ? psSignBits2 : psSignBits));
+
   if (SConfig::GetInstance().bAccurateNaNs && result_xmm == XMM0)
   {
     // HandleNaNs needs to clobber XMM0
-    MOVAPD(XMM1, R(result_xmm));
-    result_xmm = XMM1;
+    MOVAPD(Rd, R(result_xmm));
+    result_xmm = Rd;
```
booto, line 479: Feel free to ignore, but I think you can lift the two HandleNaNs() below to here as well

JosJuice: Like this?

booto (approved):

lioncash (approved):

PR #9975, opened 2021-07-30
Enable anisotropic filtering when the game requests it

Pokechu22:
This PR implements the `TexMode0.max_aniso` field.  For now this is just a test to see what games actually use it on fifoci, since I'm not sure I've actually implemented it properly.


JMC47:
Is this actually accurate?  If so it shows some pretty crazy improvements at 1x IR.

Pokechu22:
Yes, I just now hardware-tested it.

Screenshot 2021-08-01 21-11-31
Screenshot 2021-08-01 21-21-14
Screenshot 2021-08-01 21-29-48
Screenshot 2021-08-01 21-37-16

For testing purposes, I also checked what happened when anisotropic filtering is disabled by editing the fifolog, which is equivalent to what you get in Dolphin currently.  Note that I only one object in most cases (e.g. I changed the railings in F-zero, but not the road itself):

Screenshot 2021-08-01 21-15-46
Screenshot 2021-08-01 21-22-37
Screenshot 2021-08-01 21-30-45
Screenshot 2021-08-01 21-38-25

<details><summary>Changes (for reference)</summary>

Pokemon 29b2a / obj 139 change to 61800000d5
F-Zero 58f7 / obj 22 change to 618101c0d4
Tony Hawk / obj 358 / 4f3da in obj 357 change to 61800000d5
TSP3 / obj 70 / 14d7f and obj 19 / 73d8 change to 61800000d5
TSP3 also has broken rendering unless I disable geometry for all objects, and then enable it for obj 71 and 20 (old numbering scheme where geometry data is at the start of the object).
</details>

One other thing to note is that there is _no_ difference on Mesa, probably because Mesa doesn't implement anisotropic filtering.

JMC47:
@dolphin-emu-bot rebuild

AdmiralCurtiss:
Would it be possible to revive this? This came up on Discord recently and I think it would be great to actually implement correctly.

Pokechu22:
The main problem with this PR (if I recall correctly) is that it doesn't interact nicely with dolphin's anisotropic options; if dolphin's one is in use then that overrides the game's settings, but there isn't a way to override it to always disable it. There's a side issue of using anisotropic filtering without linear filtering, which I believe I determined _does_ work/do something on real hardware (at least it pushes mipmaps further out), but apparently doesn't work super well on most modern GPUs.

I was also working on this at the same time as manual texture sampling (#9956), and was somewhat looking into how anisotropic filtering worked on real hardware for that. I never ended up implementing anisotropic filtering with manual texture sampling, though.

AdmiralCurtiss:
If it's just a config problem then we should just figure that out, because it does improve quality quite a bit under 'native' rendering settings.

jordan-woyak:
Okay, it sounds like the main blocker is the inability to explicitly disable anisotropic filtering?
"Default" used to mean "Disabled" but now "Default" means "Do what the game wants".

Should I implement that to get this PR merged?

1. Do we really need a "Disabled" state? I might argue, no.
I guess the value in the ini file should be `-1`? (`0` is currently for "Default")

2. Is there a reason the Anisotropic and MinMag settings are combined into one combobox?
I'll probably change it to a combobox for Anisotropic: default,2x,4x,8x,16x,disabled
And another combobox for the MinMag filter: default,nearest,linear

Or, again, do we really need to expose a "Disabled" state?

Pokechu22:
> Do we really need a "Disabled" state? I might argue, no.

It's mostly for users who have particularly weak devices; no game seems to truly depend on having anisotropic filtering, so forcing users to use it if the game would do it on real hardware seems unnecessary.

>  I guess the value in the ini file should be -1? (0 is currently for "Default")

That's probably fine, though it depends on what older versions of Dolphin would do with that value (and if it'd cause crashes on if we passed that to graphics backends)

> Is there a reason the Anisotropic and MinMag settings are combined into one combobox?
> I'll probably change it to a combobox for Anisotropic: default,2x,4x,8x,16x,disabled
> And another combobox for the MinMag filter: default,nearest,linear

See #11276; we couldn't come up to a consensus about how it should be done. Note that anisotropic filtering and forcing nearest produces weird, implementation-defined results (but it can technically be done).

jordan-woyak:
Closing since #13368 was merged. Thanks.

PR #9976, opened 2021-07-31
DSPHLE: include compressor state in save states

Tilka:
Forgot this.


JMC47:
Probably should be merged before we set things down for the Progress Report.

lioncash (approved):

PR #9978, opened 2021-07-31
Jit: Use ibat_table for dcbf/dcbi/dcbst address check

JosJuice:
Minor mistake in PR #9957. We should be using ibat_table instead of dbat_table, since we're dealing with invalidating icache.

### DIFF
--- a/Source/Core/Core/PowerPC/Jit64Common/EmuCodeBlock.cpp
+++ b/Source/Core/Core/PowerPC/Jit64Common/EmuCodeBlock.cpp
@@ -91,9 +91,9 @@ void EmuCodeBlock::SwitchToNearCode()
   SetCodePtr(m_near_code, m_near_code_end, m_near_code_write_failed);
 }
 
-FixupBranch EmuCodeBlock::BATAddressLookup(X64Reg addr, X64Reg tmp)
+FixupBranch EmuCodeBlock::BATAddressLookup(X64Reg addr, X64Reg tmp, const void* bat_table)
 {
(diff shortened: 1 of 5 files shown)


Source/Core/Core/PowerPC/Jit64Common/EmuCodeBlock.cpp
```diff
@@ -91,9 +91,9 @@ void EmuCodeBlock::SwitchToNearCode()
   SetCodePtr(m_near_code, m_near_code_end, m_near_code_write_failed);
 }
 
-FixupBranch EmuCodeBlock::BATAddressLookup(X64Reg addr, X64Reg tmp)
+FixupBranch EmuCodeBlock::BATAddressLookup(X64Reg addr, X64Reg tmp, const void* bat_table)
```
degasus, line 94: I'd prefer `const uint64_t* bat_table`, but I don't care much.

JosJuice: It would have to be `uint32_t` instead of `uint64_t` in order to not make calling this function awkward, since the BAT table type is defined as `using BatTable = std::array<u32, 1 << (32 - BAT_INDEX_SHIFT)>`.

Originally I was going to pass in `const BatTable&`, but then I realized I would need to include MMU.h in EmuCodeBlock.h (doing a simple `class BatTable;` forward declaration isn't possible with `using`), so I switched to this.

degasus: Indeed, my bad, u32.

PR #9441, opened 2021-01-13
Apple M1 Support for MacOS

skylersaleh:
This pull request adds support for compiling Dolphin for ARM on MacOS so that it can
run natively on the M1 processors without running through Rosseta2 emulation
providing a 30-50% performance speedup and less hitches from Rosseta2.

It consists of several key changes:

- Adding support for W^X allocation(MAP_JIT) for the ARM JIT
- Adding the machine context and config info to identify the M1 processor
- Additions to the build system and docs to support building universal binaries
- Adding code signing entitlements to access the MAP_JIT functionality
- Updating the MoltenVK libvulkan.dylib to a newer version with M1 support

Thanks everyone on IRC for the help in making this change!

PR #9981, opened 2021-07-31
MoltenVK: Use an external project instead of a pre-compiled dylib

OatmealDome:
Changes the build system to build MoltenVK instead of having a dylib in the repository. This also opens up the opportunity to modify MoltenVK to our needs if needed.

Also, the MoltenVK version has been updated to v1.1.5 (Vulkan SDK 1.2.189).

~~**EDIT**: This PR is being blocked by the presence of the Intel macOS buildbot (``pr-osx-x64``). The M1 Mac Mini can build MoltenVK perfectly fine for both Intel and ARM, so the Intel buildbot needs to be shut down before can be merged.~~

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -573,6 +573,10 @@ add_subdirectory(Externals/glslang)
 
 if(ENABLE_VULKAN)
   add_definitions(-DHAS_VULKAN)
+
+  if(APPLE)
+    add_subdirectory(Externals/MoltenVK)
+  endif()
 endif()
 
 if(NOT WIN32 OR (NOT (CMAKE_SYSTEM_PROCESSOR STREQUAL "aarch64")))
--- a/Externals/MoltenVK/CMakeLists.txt
+++ b/Externals/MoltenVK/CMakeLists.txt
@@ -1,0 +1,18 @@
+include(ExternalProject)
+
+ExternalProject_Add(MoltenVK
+  GIT_REPOSITORY https://github.com/KhronosGroup/MoltenVK.git
+  GIT_TAG v1.1.5
+
+  CONFIGURE_COMMAND <SOURCE_DIR>/fetchDependencies --macos
+
+  BUILD_COMMAND make -C <SOURCE_DIR> macos
+  BUILD_IN_SOURCE ON
+  BUILD_BYPRODUCTS <SOURCE_DIR>/Package/Release/MoltenVK/dylib/macOS/libMoltenVK.dylib
+
+  INSTALL_COMMAND ""
+
+  LOG_CONFIGURE ON
+  LOG_BUILD ON
+  LOG_OUTPUT_ON_FAILURE ON
+)
--- a/Externals/MoltenVK/version.txt
+++ b/Externals/MoltenVK/version.txt
@@ -1,1 +1,0 @@
-MoltenVK from https://vulkan.lunarg.com/sdk/home#mac, version 1.2.170.0
--- a/Source/Core/DolphinQt/CMakeLists.txt
+++ b/Source/Core/DolphinQt/CMakeLists.txt
@@ -528,8 +528,12 @@ if(APPLE)
   endforeach()
 
   # Copy MoltenVK into the bundle
-  target_sources(dolphin-emu PRIVATE "${CMAKE_SOURCE_DIR}/Externals/MoltenVK/libvulkan.dylib")
-  set_source_files_properties("${CMAKE_SOURCE_DIR}/Externals/MoltenVK/libvulkan.dylib" PROPERTIES MACOSX_PACKAGE_LOCATION Frameworks)
+  if(ENABLE_VULKAN)
+    add_dependencies(dolphin-emu MoltenVK)
+    ExternalProject_Get_Property(MoltenVK SOURCE_DIR)
+    target_sources(dolphin-emu PRIVATE "${SOURCE_DIR}/Package/Release/MoltenVK/dylib/macOS/libMoltenVK.dylib")
+    set_source_files_properties("${SOURCE_DIR}/Package/Release/MoltenVK/dylib/macOS/libMoltenVK.dylib" PROPERTIES MACOSX_PACKAGE_LOCATION Frameworks GENERATED ON)
+  endif()
 
   # Update library references to make the bundle portable
   include(DolphinPostprocessBundle)
--- a/Source/Core/VideoBackends/Vulkan/VulkanLoader.cpp
+++ b/Source/Core/VideoBackends/Vulkan/VulkanLoader.cpp
@@ -43,8 +43,8 @@ static bool OpenVulkanLibrary()
   if (libvulkan_env && s_vulkan_module.Open(libvulkan_env))
     return true;
 
-  // Use the libvulkan.dylib from the application bundle.
-  std::string filename = File::GetBundleDirectory() + "/Contents/Frameworks/libvulkan.dylib";
+  // Use the libMoltenVK.dylib from the application bundle.
+  std::string filename = File::GetBundleDirectory() + "/Contents/Frameworks/libMoltenVK.dylib";
   return s_vulkan_module.Open(filename.c_str());
 #else
   std::string filename = Common::DynamicLibrary::GetVersionedFilename("vulkan", 1);


JMC47:
@dolphin-emu-bot rebuild

Dentomologist:
@dolphin-emu-bot rebuild

kemenaran:
(This is largely outside of the scope of this PR - but for what is worth, gfx-portability as a drop-in replacement for MoltenVK performs way better for me (in terms of FPS and frame pacing). Is there anything preventing MoltenVK to be replaced by gfx-portability, or is it just some work to be done?)

OatmealDome:
@kemenaran As far as I'm aware, there are no plans to replace MoltenVK with gfx-portability.

Dentomologist (changes requested):
Builds and runs without issue.

Minor nitpick: The blank lines you added in CMakeLists.txt files have a couple extraneous spaces at the beginning; I'm surprised the linter didn't complain about them.

OatmealDome:
Didn't even notice, thanks. Xcode likes to add those.

I don't think the linter runs on CMakeLists?

Dentomologist (approved):

Dentomologist (commented):
(Approval conditional on upgrading or removing the macOS Intel builder)

nickp-ca:
Any reason the intel MacOS buildbot pr-osx-x64 is still needed considering pr-osx-universal will run on Intel?

OatmealDome:
@nickp-ca It isn't needed and is planned to be shut down (see the comment above yours).

leoetlino:
Could you rebase this PR please?

OatmealDome:
Rebased. Also, updated to the latest MoltenVK release.

The build time is a bit slow, even on incremental builds. I have some ideas for decreasing the build time, but I'll open a separate PR for that considering this one has already been reviewed and approved...

iljatanger:
Hello dear developer and thank you very much for your labour and such great emulator.

Started from this release, the performance in many games (SMG, SMG2, Metroid Prime, Super Mario bros.) etc got significantly lower (2-3 times!).

this is 100% tested on many releases (about 20 times). 

Initially I tried to update Dolphin on MacBook Air M1 from the beta version 15445 to the newest one 16101, because I had artifacts (black and transparent squares in the center of screen in Super Smash Bros. Brawl). And I was shocked what the performance became so poor. If at 15445  3X in Mario Galaxy and Metroid Prime was playable, now it became much worse, so only 1X and 2x (hardly) playable.  So after that I installed many releases and found, that such visible, significant performance drop had started from this release. The size of the release also was changed started from this version 15474 (38.8mb instead of 34.9 before, so I suggest it was some global recompiling of the code).

so now I stopped on version 15472 as the latest playable one. But unfortunately, I still have the same visual artifacts in Super Smash Bros...

Could you ever be so kind to check whether there is some mistake in this release starting from 15474?

bluesalt:
All the version started from this release have kept crash with vulkan mode in my macbook pro.  Like @iljatanger , the workable version stopped at 15472 in my macbook.

Process:               Dolphin 
Path:                  /Applications/Dolphin.app/Contents/MacOS/Dolphin
Identifier:            org.dolphin-emu.dolphin
Version:               5.0-15474 (5.0)
Code Type:             X86-64 (Native)
Parent Process:        ??? [1]
Responsible:           Dolphin [3953]


OS Version:            Mac OS X 10.15.7 (19H2)
Report Version:        12
Bridge OS Version:     4.6 (17P6610)



System Integrity Protection: enabled

Crashed Thread:        28  Emuthread - Starting

Exception Type:        EXC_BAD_ACCESS (SIGSEGV)
Exception Codes:       KERN_INVALID_ADDRESS at 0x0000000000000000
Exception Note:        EXC_CORPSE_NOTIFY

Termination Signal:    Segmentation fault: 11
Termination Reason:    Namespace SIGNAL, Code 0xb
Terminating Process:   exc handler

OatmealDome:
This is a known issue with MoltenVK and will be fixed as part of the next MoltenVK release. (https://bugs.dolphin-emu.org/issues/12840)

pizuz:
> Hello dear developer and thank you very much for your labour and such great emulator.
> 
> Started from this release, the performance in many games (SMG, SMG2, Metroid Prime, Super Mario bros.) etc got significantly lower (2-3 times!).

I can second that. At the moment, even the deprecated OpenGL seems to be faster than Vulkan in the games I tested. This performance regression affects all releases including the most recent one (at the time of writing 5.0-16831).

Late 2013 iMac with 3,1 GHz Quad-Core Intel Core i7 and NVIDIA GeForce GT 750M 1 GB running macOS 10.15.7.

OatmealDome:
@pizuz Here's an experiment to try, if you want.

Try running `LIBVULKAN_PATH=/path/to/libMoltenVK.dylib /path/to/DolphinBeforePerformanceRegression.app/Contents/MacOS/Dolphin` in the Terminal with a path to the [latest MoltenVK dylib from LunarG](https://vulkan.lunarg.com/sdk/home#mac).

pizuz:
Yup. MoltenVK 1.1.2 (which was the last update introduced in #9441) goes a lot faster. The performance regression must have been introduced somewhere between 1.1.3 and 1.1.5.

I'll try to bisect next.

OatmealDome:
@pizuz If I remove the code signatures from those files, the file hashes end up being exactly the same. Perhaps you made a mistake somewhere?

```
$ shasum dolphin.dylib 
6d6af070954de2a5049550c3f23ffcad43e9e506  dolphin.dylib
$ shasum lunarg.dylib
cc12d4c897ea01370c51c2dcd94bd62e6ddb2f03  lunarg.dylib
$ codesign --remove-signature dolphin.dylib 
$ codesign --remove-signature lunarg.dylib 
$ shasum dolphin.dylib 
3be651368e7d9c9940a907eb894728194a413106  dolphin.dylib
$ shasum lunarg.dylib 
3be651368e7d9c9940a907eb894728194a413106  lunarg.dylib
```

pizuz:
There is a difference in performance-light scenes such as splash screens which render twice as fast, now. In other instances, such as the attract screen in Super Mario Galaxy, there is no improvement, unfortunately. I'm still trying to find time to bisect those old PRs in MoltenVK to pinpoint the change that caused this. MVK 1.1.2 definitely runs a lot faster.

MayImilae:
While benchmarking Metal for the Progress Report, I noticed that MoltenVK's performance was way down from what I recorded previously. The decrease varies depending on the game and what it is doing, but with at least 10% down in all tests. But in GPU heavy workloads, I found ~40% worse performance!

Specifically, those worst numbers came from the Prime 3 Elysia Vista test (a particularly nasty GPU thread workout). 5.0-15472 was 61fps, while 5.0-15474 is 37-40fps. That matches latest master (Dolphin 5.0-17134), so it seems that this PR is the only drop.

Unfortunately this is kind of a problem in demonstrating Metal, because MoltenVK is our comparison point and if its performance is this much worse, it's going to make Metal look really, REALLY good. I wonder what caused this?

OatmealDome:
@MayImilae Could you try experimenting with the following command in the Terminal with latest dev and see if it improves things?

``MVK_ALLOW_METAL_FENCES=1 /path/to/Dolphin.app/Contents/MacOS/Dolphin``

MayImilae:
No change.

OatmealDome:
Interesting. I thought this would improve things since it did for pizuz above...

@tellowkrinkle Any ideas?

tellowkrinkle:
> While benchmarking Metal for the Progress Report, I noticed that MoltenVK's performance was way down from what I recorded previously. The decrease varies depending on the game and what it is doing, but with at least 10% down in all tests. But in GPU heavy workloads, I found ~40% worse performance!

Are those GPU heavy workloads ones that involve readbacks or no?  I know Metroid Prime has readbacks.

For games with readbacks (EFB to RAM), the issue is at least known and on the MVK bugtracker now.  You can try shrinking the pool size [here](https://github.com/dolphin-emu/dolphin/blob/b02653722d88577b059b51c8bb76add081158f04/Source/Core/VideoBackends/Vulkan/CommandBufferManager.cpp#L109) as a temporary workaround, though making it too small could also result in performance issues (for everyone, not just MVK), so ideally we leave it as is for production.

If there's issues with games that have EFB to RAM disabled, I'd be interested in looking into investigating those.  If you could grab some Metal System Traces in Instruments, that would be really helpful.

MayImilae:
Prime 3 does use readbacks for the visors, but the test in question doesn't use a visor. It shouldn't be affecting this test. Also, the performance drop was in all of my typical tests, including games that don't use readbacks at all.

I can do some more specific tests for readbacks versus not if you'd like.

pizuz:
@MayImilae, what happens if you replace the MoltenVK dylib with the old 1.1.2 version from #9441?

tellowkrinkle:
In that case, a Metal System Trace would be nice
Ideally one of `5.0-15472` and one of `5.0-15474`

tellowkrinkle:
> what happens if you replace the MoltenVK dylib with the old 1.1.2 version from #9441?

I don't think we try to verify our MVK supports our custom fbfetch modification before trying to use it... maybe we should though.

tellowkrinkle:
@MayImilae actually first, try messing with the pool size anyways.  It still might help, we still run one pool reset per frame even if you really start to feel the effects when there's multiple.

tellowkrinkle:
BTW are your GPU heavy traces heavy on actual GPU or heavy on the CPU's GPU command buffer encoding?  If they're actually GPU heavy, messing with pool size won't help, but with an M1 Max, I'd expect you to be running these at insane resolutions before you actually end up GPU heavy.

(Personally, I haven't seen MVK have much trouble when you're actually GPU limited, though I'm on an AMD GPU and MVK sets some settings that may negatively affect Apple GPUs more than others)

pizuz:
> > what happens if you replace the MoltenVK dylib with the old 1.1.2 version from #9441?
> 
> I don't think we try to verify our MVK supports our custom fbfetch modification before trying to use it... maybe we should though.

I‘m talking Vanilla MoltenVK.

tellowkrinkle:
> I‘m talking Vanilla MoltenVK.

Right.  I'm saying that we'll end up sending it shaders that it doesn't know what to do with, since it wasn't modified to support framebuffer fetch.

There's a chance vanilla MVK will handle those shaders fine.  I don't have an Apple GPU to try with, and on non-Apple GPUs we assume no framebuffer fetch support.  But I wouldn't exactly count on it.

MayImilae:
Confirmed the performance drop in Metroid Prime GC, which has no readbacks whatsoever. I also noticed some changes in behavior that were interesting, but I need to stop testing for the night. I'll report on that later or edit this post.

pizuz:
> > I‘m talking Vanilla MoltenVK.
> 
> Right. I'm saying that we'll end up sending it shaders that it doesn't know what to do with, since it wasn't modified to support framebuffer fetch.
> 
> There's a chance vanilla MVK will handle those shaders fine. I don't have an Apple GPU to try with, and on non-Apple GPUs we assume no framebuffer fetch support. But I wouldn't exactly count on it.

I was saying that regardless of being patched or not, any MoltenVK version after 1.1.2 is affected by this regression. I tried them all.

tellowkrinkle:
> I was saying that regardless of being patched or not, any MoltenVK version after 1.1.2 is affected by this regression. I tried them all.

Would you be willing to bisect MVK builds down to a specific commit?

pizuz:
> > I was saying that regardless of being patched or not, any MoltenVK version after 1.1.2 is affected by this regression. I tried them all.
> 
> Would you be willing to bisect MVK builds down to a specific commit?

I was planning to, but at the moment I don‘t have any time. Especially since the MVK team doesn‘t archive their CI builds for individual PRs. I‘ll have to build everything myself over and over again.

tellowkrinkle:
Okay.  I'm going to try to repro on my laptop now, and if successful, I'll bisect.

pizuz:
> Okay. I'm going to try to repro on my laptop now, and if successful, I'll bisect.

As I said earlier, the regression must have been introduced between 1.1.2 and 1.1.4 (1.1.3 freezes Dolphin).

tellowkrinkle:
Bisected to https://github.com/KhronosGroup/MoltenVK/commit/4371ef4d2b706d761acac49c1b7f9d413d0d15db, which matches up with the pool reset issues I reported.  If it is just this, reducing the size of our descriptor pool is a fine temporary solution for testing while we wait for MVK to properly fix this.

If you could check that to confirm that it fixes performance (or just confirm that reducing the descriptor pool size fixes performance), that would be ideal.

pizuz:
I‘ll try building it later today.

MayImilae:
@pizuz I'd love to try that build myself.

pizuz:
> @pizuz I'd love to try that build myself.

I'd love to share, but CMake still doesn't seem to let me build and I have no idea why that is. I'm getting no reasonable error message. It did work before.

[cmake error.txt](https://github.com/dolphin-emu/dolphin/files/9287092/cmake.error.txt)

OatmealDome:
@pizuz Building using static SDL from the submodule on macOS does not work at this time. Install SDL as a shared library onto your system (for example, with brew) and it should work.

pizuz:
Thanks for the hint. Now the Metal backend fails to build...

`Building CXX object Source...es/videometal.dir/MTLStateTracker.mm.o
FAILED: Source/Core/VideoBackends/Metal/CMakeFiles/videometal.dir/MTLStateTracker.mm.o 
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++ -DAUTOUPDATE=1 -DDATA_DIR=\"/usr/local/share/dolphin-emu/\" -DFMT_EXCEPTIONS=0 -DFMT_LOCALE -DFMT_SHARED -DHAS_LIBMGBA -DHAS_OPENGL -DHAS_VULKAN -DHAVE_CRC32 -DHAVE_SDL2=1 -DLZMA_API_STATIC -DSFML_STATIC -DSPIRV_CROSS_EXCEPTIONS_TO_ASSERTIONS -DUSE_ANALYTICS=1 -DUSE_MEMORYWATCHER=1 -DUSE_PIPES=1 -DUSE_UPNP -D_ARCH_64=1 -D_DEFAULT_SOURCE -D_FILE_OFFSET_BITS=64 -D_LARGEFILE_SOURCE -D_M_X86=1 -D_M_X86_64=1 -D__LIBUSB__ -D__STDC_CONSTANT_MACROS -D__STDC_LIMIT_MACROS -I/Users/pizuz/Downloads/Dolphin-dev/Dolphin/Externals/SFML/include -I/Users/pizuz/Downloads/Dolphin-dev/Dolphin/Source/Core -I/Users/pizuz/Downloads/Dolphin-dev/Dolphin/Externals/enet/include -I/Users/pizuz/Downloads/Dolphin-dev/Dolphin/External/minizip -I/Users/pizuz/Downloads/Dolphin-dev/Dolphin/Externals/ed25519 -I/Users/pizuz/Downloads/Dolphin-dev/Dolphin/Externals/soundtouch -I/Users/pizuz/Downloads/Dolphin-dev/Dolphin/Externals/mbedtls/include -I/Users/pizuz/Downloads/Dolphin-dev/Dolphin/Externals/discord-rpc/include -I/Users/pizuz/Downloads/Dolphin-dev/Dolphin/Externals/picojson -I/Users/pizuz/Downloads/Dolphin-dev/Dolphin/build/Source/Core -I/Users/pizuz/Downloads/Dolphin-dev/Dolphin/Externals/minizip/. -I/Users/pizuz/Downloads/Dolphin-dev/Dolphin/build/Externals/zlib-ng/zlib-ng -I/Users/pizuz/Downloads/Dolphin-dev/Dolphin/Externals/zlib-ng/zlib-ng -I/Users/pizuz/Downloads/Dolphin-dev/Dolphin/Externals/cubeb/include -I/Users/pizuz/Downloads/Dolphin-dev/Dolphin/build/exports -I/Users/pizuz/Downloads/Dolphin-dev/Dolphin/Externals/liblzma/api -I/Users/pizuz/Downloads/Dolphin-dev/Dolphin/Externals/zstd/lib -I/Users/pizuz/Downloads/Dolphin-dev/Dolphin/Externals/libusb/libusb/libusb -I/Users/pizuz/Downloads/Dolphin-dev/Dolphin/Externals/pugixml/. -I/Users/pizuz/Downloads/Dolphin-dev/Dolphin/Externals/mGBA/mgba/include -I/Users/pizuz/Downloads/Dolphin-dev/Dolphin/build/Externals/mGBA/mgba/include -I/Users/pizuz/Downloads/Dolphin-dev/Dolphin/Externals/hidapi/hidapi -I/Users/pizuz/Downloads/Dolphin-dev/Dolphin/Externals/spirv_cross/SPIRV-Cross/include -I/Users/pizuz/Downloads/Dolphin-dev/Dolphin/Externals/spirv_cross/SPIRV-Cross -isystem /usr/local/include -isystem /Users/pizuz/Downloads/Dolphin-dev/Dolphin/Externals/glslang/glslang/Public -isystem /Users/pizuz/Downloads/Dolphin-dev/Dolphin/Source/Core/VideoCommon/SYSTEM -isystem /Users/pizuz/Downloads/Dolphin-dev/Dolphin/Externals/rangeset/include -O3 -DNDEBUG -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX11.1.sdk -mmacosx-version-min=10.14.0 -msse2 -fdiagnostics-color -fno-strict-aliasing -fno-exceptions -fvisibility-inlines-hidden -fvisibility=hidden -fomit-frame-pointer -mssse3 -march=core2 -Wall -Wtype-limits -Wsign-compare -Wignored-qualifiers -Wuninitialized -Wshadow -Winit-self -Wmissing-declarations -Wmissing-variable-declarations -Werror=format -std=c++2a -MD -MT Source/Core/VideoBackends/Metal/CMakeFiles/videometal.dir/MTLStateTracker.mm.o -MF Source/Core/VideoBackends/Metal/CMakeFiles/videometal.dir/MTLStateTracker.mm.o.d -o Source/Core/VideoBackends/Metal/CMakeFiles/videometal.dir/MTLStateTracker.mm.o -c /Users/pizuz/Downloads/Dolphin-dev/Dolphin/Source/Core/VideoBackends/Metal/MTLStateTracker.mm
/Users/pizuz/Downloads/Dolphin-dev/Dolphin/Source/Core/VideoBackends/Metal/MTLStateTracker.mm:631:30: error: use of overloaded operator '!=' is ambiguous (with operand types 'Metal::DepthStencilSelector' and 'Metal::DepthStencilSelector')
    if (pipe->DepthStencil() != m_current.depth_stencil)
        ~~~~~~~~~~~~~~~~~~~~ ^  ~~~~~~~~~~~~~~~~~~~~~~~
/Users/pizuz/Downloads/Dolphin-dev/Dolphin/Source/Core/VideoBackends/Metal/MTLObjectCache.h:41:8: note: candidate function
  bool operator!=(const DepthStencilSelector& other) { return !(*this == other); }
       ^
/Users/pizuz/Downloads/Dolphin-dev/Dolphin/Source/Core/VideoBackends/Metal/MTLObjectCache.h:40:8: note: candidate function
  bool operator==(const DepthStencilSelector& other) { return value == other.value; }
       ^
/Users/pizuz/Downloads/Dolphin-dev/Dolphin/Source/Core/VideoBackends/Metal/MTLObjectCache.h:40:8: note: candidate function (with reversed parameter order)
1 error generated.
[1153/3570] Performing configure step for 'MoltenVK'
ninja: build stopped: subcommand failed.`

OatmealDome:
@MayImilae @pizuz A potential fix landed in MoltenVK upstream - could you test this build and see if it improves anything? https://dl.dolphin-emu.org/prs/65/fe/pr-10968-dolphin-latest-universal.dmg

BJS1206:
@OatmealDome Yes, this build improved performance for me drastically! M1 Macbook Air Mario Galaxy 2 @ 1080p ~40 fps -> 60 fps, no dips. I was wondering why my iPhone 11 was outpreforming my M1 Mac in Dolphin, that explains it.

MayImilae:
That is a massive performance improvement! However it is slightly less than anticipated. So in the Elysia test I described previously before the perf drop I would get 60fps in MoltenVK (metal in latest master gets 61fps). With pr10968 I am getting 58fps. I'll hella take that, but it's slightly less. Possibly from other changes in Dolphin or in MoltenVK, no way to know.

Either way, very nice. What change was it that fixed this?

OatmealDome:
Thanks for testing! MoltenVK will be releasing a new version on August 15 to coincide with a new Vulkan SDK release. Not sure if we should be waiting for that, or if I should just make a PR updating to the latest upstream ASAP for the report.

@MayImilae Tellow found that `vkResetDescriptorSets` was implemented in such a way where it will reset every descriptor set in a pool regardless if it was actually used. Dolphin's Vulkan backend allocates 100000 descriptor sets, but doesn't use all of them, so every call to `vkResetDescriptorSets` ends up taking a significant amount of CPU time. A PR was just merged (KhronosGroup/MoltenVK#1676) which improves the implementation to only reset what was used, which is very likely where the performance improvement comes from.

Hopefully I got that right. Tellow can chime in if I didn't \:P

pizuz:
It seems to improve the speed quite a bit (looks like it's on par with OpenGL again on the SMG splash screen), but the performance once achieved with MVK 1.1.2 was a lot better. I'm still seeing a substantial gap between Vulkan and Metal.

BJS1206:
@pizuz Do you know if it's possible to use MVK 1.1.2 with the latest version of dolphin (or can you direct me to the dolphin version that used 1.1.2)? I would like to test to see the difference in performance between the two versions.

pizuz:
> @pizuz Do you know if it's possible to use MVK 1.1.2 with the latest version of dolphin (or can you direct me to the dolphin version that used 1.1.2)? I would like to test to see the difference in performance between the two versions.

Just copy the dylib over (either from the last Dolphin version before this PR or from the official Vulkan SDK) and rename it to libvulkan.dylib. I don't know if the FBFetch patch is required for M1 macs, but since my nVidia card doesn't support that feature anyway, I never encountered any issues.

OatmealDome:
@pizuz Could you try the fixed build with this command that I brought up a while ago?

`MVK_ALLOW_METAL_FENCES=1 /path/to/Dolphin.app/Contents/MacOS/Dolphin`

I ask because a change was made to MoltenVK to better comply with the Vulkan specification which loses some performance, and they also had to workaround a bug in NVIDIA's GPU drivers which loses performance. Those two things combined might be the reason why you're still seeing a drop. Turning that flag on will restore the old behaviour.

tellowkrinkle:
@pizuz could you use Instruments to collect System Traces and Metal System Traces of MVK 1.1.2 vs latest?

<details>
<summary>How to collect traces in Instruments</summary>

1. Open Instruments, either by searching for it in spotlight or through Xcode → Open Developer Tool in Xcode's main menu
2. Select the trace type you want
image
3. Click "All Processes" on the top, select "Choose Target...", and select Dolphin.  You can also add `LIBVULKAN_PATH` to the environment variables to use a custom MoltenVK dylib.  Then select "Choose"
image
4. Click and hold the red button in the upper left corner, then select "Recording Options..."
image
5. In global options, you can optionally change the recording mode to "Last 2 seconds" to keep the trace short.  Traces get big fast.  I've had issues with the "Last X seconds" mode losing data in Metal System Traces though, so you might want to stick to Deferred for those.
image
6. Press record, do whatever you want to record a trace of in Dolphin, and then press the stop button in Instruments without quitting Dolphin.  Stopping can take a few seconds, and you don't want the last few seconds of trace to be dolphin quitting.
7. Cmd-S to save, and upload to some file transfer site because the trace will probably be too big to upload to GitHub or Discord.

</details>

pizuz:
I‘ll try over the weekend.

pizuz:
Alright, these are the [Metal traces](https://mega.nz/file/8cJXmLza#qinvrVCYKsub6Hvu7YMd0oF2LJCJGohDklm_W69hqVo) and the [system traces](https://mega.nz/file/EEQDjJ4D#9g6P9N5XmA0NWFIHsTtrCibsm9pq1QLsLs8UJFDl088).

I uploaded four different scenarios: MVK 1.1.2, MVK 1.1.11pre, OpenGL and Metal. I used the beginning of the Super Mario Galaxy attract screen, loaded from a save state and without a frame rate cap. Interestingly, the difference between MVK 1.1.11pre and 1.1.2 is less than I initially remembered.

tellowkrinkle:
It looks like MVK is stalling for about 6ms every other frame when we attempt to present a surface, stuck waiting for some other task from a previous frame to finish.  I've encountered this on my 750M laptop in PCSX2 as well, it can get really bad there.  I haven't encountered it on any non-Nvidia GPU.

<details>
<summary>Screenshots of MVK 1.1.11 doing bad things</summary>

image
image
</details>

I'll see if 1.1.2 runs PCSX2 faster on that laptop, and if it does, I'll bisect it there.

billhollings:
> Screenshots of MVK 1.1.11 doing bad things

1. I see only two surfaces in the trace. What is the value of `VkSwapchainCreateInfoKHR::minImageCount`? If it is in fact set to 2, does any improvement happen if you set it to 3?

2. I see a reference to `MVKSemaphoreEmulated`. I assume this is is because the default for `MVKConfiguration::semaphoreUseMTLEvent` (env var `MVK_ALLOW_METAL_EVENTS`) is disabled when using NVIDIA GPUs. Unfortunately, this ties Vulkan semaphores to CPU callbacks from command buffer completion, rather than leaving them in the GPU. What happens if you enable the above config flag or equivalent env var? And if that breaks things, the second choice would be to enable `MVKConfiguration::semaphoreUseMTLFence` (env var `MVK_ALLOW_METAL_FENCES`).

tellowkrinkle:
`MVK_ALLOW_METAL_FENCES=1` clears it up for me.  @pizuz?

billhollings:
> [MVK_ALLOW_METAL_EVENTS seems to be disabled regardless of the env variable on Nvidia GPUs](https://github.com/KhronosGroup/MoltenVK/blob/ea5a53d13c4ff7bad454ea44de9ca7e00113a3cd/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm#L4417), so it doesn't do anything. Is that intended?

Uggh. You're correct. 🤦🏻‍♂️

Are you in a position to override that line of code to force it on NVIDIA, and test whether it works properly? If it does work in some cases on NVIDIA, we could redesign the logic of the env vars to allow a kind of "force `MTLEvents` in all cases" option.

tellowkrinkle:
I removed the Nvidia check and ran it for ~10min in PCSX2 and ~5min in Dolphin and didn't get any crashes or other weird things.

billhollings:
> I removed the Nvidia check and ran it for ~10min in PCSX2 and ~5min in Dolphin and didn't get any crashes or other weird things.

Thanks for the testing and feedback. I've created a [MoltenVK issue](https://github.com/KhronosGroup/MoltenVK/issues/1695) to track changing MoltenVK to accommodate this situation.

tellowkrinkle:
@pizuz Can you test #11179 and see if it resolves your performance issues on Nvidia?

pizuz:
@tellowkrinkle, sorry, I didn't find the time to test it until now. You guys merged a ton of performance optimisations and I didn't have the time to bisect, yet, to see if updating to MVK 1.2.0 brought any benefit in itself. Anyway, latest master provides a huge performance boost to the point that MVK actually outperforms your Metal backend on my machine.

tellowkrinkle:
Nice.  (If you're wondering, [MoltenVK 1.2.0 recently got a new alternative to MTLEvents that works on Nvidia GPUs](https://github.com/KhronosGroup/MoltenVK/pull/1709), which is now enabled by default)

And with the now-working multithreaded submission, you can expect MVK performance to be similar to Metal in performance, depending on how fast your GPU's Metal driver is compared to MoltenVK (MVK does all the Metal work on the submission thread, so the Metal driver being slow doesn't slow down the emulator).  In my testing, AMD's Metal driver is fast, while Intel's isn't so much.  I guess Nvidia's is on the slower side of things.  (Metal will still be faster in games with a lot of synchronization, like Paper Mario)

pizuz:
In my limited testing capabilities, my nVidia Kepler card (iMac 14,3) is faster using MVK than using the Metal backend. Oh, and it‘s faster than my M2 MacBook Air.

PR #9983, opened 2021-08-01
Use MagicYUV as the lossless video codec

chungy:
A much faster video codec than FFV1, this enables high-resolution
dumps to be performed in much more reasonable time, and even realtime
recording on some lower resolutions.

Really motivated by wanting to do 1080p/2160p recordings faster. I can achieve exactly as such now in roughly 50% realtime, versus ~3% I got with FFV1.

### DIFF
--- a/Source/Core/Core/Config/GraphicsSettings.cpp
+++ b/Source/Core/Core/Config/GraphicsSettings.cpp
@@ -41,7 +41,7 @@ const Info<bool> GFX_CACHE_HIRES_TEXTURES{{System::GFX, "Settings", "CacheHiresT
 const Info<bool> GFX_DUMP_EFB_TARGET{{System::GFX, "Settings", "DumpEFBTarget"}, false};
 const Info<bool> GFX_DUMP_XFB_TARGET{{System::GFX, "Settings", "DumpXFBTarget"}, false};
 const Info<bool> GFX_DUMP_FRAMES_AS_IMAGES{{System::GFX, "Settings", "DumpFramesAsImages"}, false};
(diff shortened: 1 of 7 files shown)


JosJuice:
Please do not edit the .po/.pot files. Any manual changes you make to those files will be reverted the next time the translations are synced.

chungy:
Well alright

mbc07:
FifoCI doesn't seem happy with this change and AFAICT the ffmpeg build used by Dolphin is a stripped down version, are you sure that just changing the requested codec is sufficient?

chungy:
It seems sufficient for my purposes (faster lossless dumps). I suppose the CI server would need to build ffmpeg with magicyuv enabled so it'd work.

Techjar:
I'm gonna have to say I'm against this change, on the principle that this is a much lesser known codec, and may not be supported everywhere. You can always use whatever codec you want by turning off the FFV1 option and setting `DumpCodec` in `GFX.ini`.

chungy:
I'm not sure FFV1 would be any more widely supported, but hey, I didn't know it was possible to set a custom codec.

Techjar:
Well, in any case, I don't think this is the right approach. I believe we should rethink how the codec configuration works and make it more flexible.

PatrickFerry:
The lossless video codec default is now Ut Video (#13233)

jordan-woyak:
This PR has conflicts and has aged. We're now using Ut Video as a lossless codec so I'm going to close this for now, but we can re-open it if the situation changes. Thanks.

PR #9986, opened 2021-08-02
FilesystemWidget: Show more information about partitions

Pokechu22:
This adds the partition type, partition gameid, and partition title ID from the ticket into the filesystem widget.  This is mostly just information that's interesting but not particularly useful, but it does help with seeing how the different values differ.  Note that the partition gameid already shows up if you take screenshots while running something on a different partition (e.g. the Wii Fit Channel installer).

<details><summary>Screenshots</summary>

Super Mario Sunshine
Super Mario Galaxy
Wii Freeloader (US)
Wii Freeloader (EU)
Mario Kart Wii
Wii Fit
Wii Fit Plus
Super Smash Bros. Brawl
Super Smash Bros. Brawl - expanded
</details>

### DIFF
--- a/Source/Core/DolphinQt/Config/FilesystemWidget.cpp
+++ b/Source/Core/DolphinQt/Config/FilesystemWidget.cpp
@@ -16,6 +16,8 @@
 
 #include <future>
 
+#include "Common/StringUtil.h"
+
 #include "DiscIO/DiscExtractor.h"
 #include "DiscIO/DiscUtils.h"
 #include "DiscIO/Filesystem.h"
@@ -101,7 +103,7 @@ void FilesystemWidget::PopulateView()
 
   for (size_t i = 0; i < partitions.size(); i++)
   {
-    auto* item = new QStandardItem(tr("Partition %1").arg(i));
+    auto* item = new QStandardItem;
     item->setEditable(false);
 
     item->setIcon(Resources::GetScaledIcon("isoproperties_disc"));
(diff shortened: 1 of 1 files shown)


JMC47:
Looks good to me.

JosJuice:
I like the idea, but I think the lines are rather long now. What do you think about getting rid of the initial "Partition %1", or maybe the title ID as ASCII at the end?

Pokechu22:
The title ID as ASCII is interesting (it's where `.UPE` appears, for instance), but getting rid of the initial word "Partition" would work. (I think it might be useful to keep the number, if everything fits with it, but if there still isn't enough room I'll get rid of that too.)

JosJuice:
I worry a little that the number might be confusing. For instance, an update partition would normally be listed with one number saying "0" and one number saying "1", and I don't think it's clear to the user what these two numbers mean. (Though I guess it's true that even if we remove one of them, it's unclear what the remaining one means...)

Pokechu22:
Yeah, actually, it's probably not useful information.  Discs apparently have [4 lists of partitions](https://wiibrew.org/wiki/Wii_Disc#Partitions_information) (I don't really know why), so a single partition number isn't helpful for that purpose.  [DVDLowOpenPartition](https://wiibrew.org/wiki//dev/di#0x8B_DVDLowOpenPartition) uses the offset to the partition, so it's not useful for understanding what games are doing either.  I'll remove it.

Pokechu22:
With the "Partition %1" text removed, things fit in almost all cases, with the one exception of Super Smash Bros. Brawl's partitions, where they fit as long as none of the partitions are expanded, and the end gets truncated to make room for the size column if one is expanded (which I think is perfectly fine behavior).  I've updated the screenshots to show this.

Source/Core/DolphinQt/Config/FilesystemWidget.cpp
```diff
@@ -123,6 +125,58 @@ void FilesystemWidget::PopulateView()
...
                                          const DiscIO::Partition& partition)
 {
+  auto partition_type = m_volume->GetPartitionType(partition);
+  auto game_id = m_volume->GetGameID(partition);
+  auto title_id = m_volume->GetTitleID(partition);
+
+  QString text = root->text();
+
+  if (!text.isEmpty())
+    text += QString::fromLatin1(" - ");
```
JosJuice, line 135: You can probably replace your `fromLatin1`s with `QStringLiteral`.

BhaaLseN:
The order in those screenshots seems a bit weird. The "Data Partition 0" being the most interresting one (usually with the index 0), but _not_ being the first in the list feels off to me. Is this supposed to be in order of partition table?

Not sure if the screenshots are up2date though.

leoetlino (approved):

PR #9989, opened 2021-08-03
GameINI: SpongeBob SquarePants: Battle For Bikini Bottom Quality of Life settings

JMC47:
This enables Vertex Rounding Hack and has a game patch for two bugs resulting from game choices.  The game draws an EFB copy from texture coordinate 0.00195313 to 1.00195 repeating which only works due how to the numbers work out at low resolution.

The Vertex Rounding Hack is a null operation on 1x IR, so there should be no concern enabling it by default since higher resolutions are completely broken without it.

I spent all day testing the game and got nearly to the end with no crashes nor major graphical glitches at 4x internal resolution.  The PAL version of the patch I developed myself based on the NTSC version of the patch, which was based on the Action Replay code on the wiki - https://wiki.dolphin-emu.org/index.php?title=SpongeBob_SquarePants:_Battle_for_Bikini_Bottom

### DIFF
--- a/Data/Sys/GameSettings/GQPE78.ini
+++ b/Data/Sys/GameSettings/GQPE78.ini
@@ -1,0 +1,14 @@
+# GQPE78 - SpongeBob SquarePants: Battle for Bikini Bottom
+
+[OnFrame]
+$EFB Copy Fix
+0x803CD04C:byte:0x00000000
+
+[OnFrame_Enabled]
+# This game renders an EFB copy with texture repeating enabled
+# and it draws from texture coordinate 0.00195313 to 1.00195.
+# This only works on console and 1x IR due to low precision.
+# "EFB Copy Fix" adjusts the region to not cause bugs at higher
+# resolutions.  In order for this patch to fully work, the
+# Vertex Rounding Hack must be enabled.
+$EFB Copy Fix
--- a/Data/Sys/GameSettings/GQP.ini
+++ b/Data/Sys/GameSettings/GQP.ini
@@ -1,0 +1,18 @@
+# GQPP78, GQPE78 - SpongeBob SquarePants: Battle for Bikini Bottom
+
+[Core]
+# Values set here will override the main Dolphin settings.
+
+[OnLoad]
+# Add memory patches to be loaded once on boot here.
+
+[OnFrame]
+# Add memory patches to be applied every frame here.
+
+[ActionReplay]
+# Add action replay cheats here.
+
+[Video_Hacks]
+# Fixes shadows at higher resolution.
+# Option has no effect at 1x IR, so no reason not to enable.
+VertexRounding = True
--- a/Data/Sys/GameSettings/GQPP78.ini
+++ b/Data/Sys/GameSettings/GQPP78.ini
@@ -1,0 +1,14 @@
+# GQPP78 - SpongeBob SquarePants: Battle for Bikini Bottom
+
+[OnFrame]
+$EFB Copy Fix
+0x803CD414:byte:0x00000000
+
+[OnFrame_Enabled]
+# This game renders an EFB copy with texture repeating enabled
+# and it draws from texture coordinate 0.00195313 to 1.00195.
+# This only works on console and 1x IR due to low precision.
+# "EFB Copy Fix" adjusts the region to not cause bugs at higher
+# resolutions.  In order for this patch to fully work, the
+# Vertex Rounding Hack must be enabled.
+$EFB Copy Fix


JMC47:
For what it's worth, the addresses I'm overwriting has "0.001953125" or 3B 00 00 00 in Hex in them.  By overwriting them with 00 00 00 00, you fix the offset and make the game render correctly regardless of resolution.

Pokechu22:
What's specifically happening is that the game is rendering to a 256 by 256 area at the top-left of the screen to generate shadows, but at higher IRs, this is offset slightly.  The game then tries to render an EFB copy over the shadow thing it previously rendered, which also gets offset.  The game is actually trying to compensate for this offsetting by changing the texture coordinate by 1/512 (which is 0.001953125), which works except that it requires wrapping around on the texture.  Since the texture is configured to repeat, you get the stuff at the top of the screen instead, which is jarring; had they not configured it to repeat, there would still be some graphical artifacts, but less so.  **Importantly, all of these proposed result in identical rendering 1x IR**, so this change doesn't introduce accuracy problems.

Here are some high-resolution (8x IR) screenshots:

* Without any changes ([full image](https://user-images.githubusercontent.com/8334194/127968929-26133a91-99a3-4625-9571-3b734548ff93.png)):
  OHBCHB_2021-08-02_21-47-56_norm_crop 
* The shadow it's trying to cover ([full image](https://user-images.githubusercontent.com/8334194/127969571-cbb881c8-29a9-40c9-9b28-5042bfd7e121.png)):
  OHBCHB_2021-08-02_21-49-12_shadow_crop
* If wrapping were disabled ([full image](https://user-images.githubusercontent.com/8334194/127969669-27ea4559-a21d-4456-9722-6fe35bbd248d.png)):
  OHBCHB_2021-08-02_21-50-17_nowrap_crop
* With this patch ([full image](https://user-images.githubusercontent.com/8334194/127969899-421276c2-32b3-4f4c-b4f5-4959524f95b5.png)):
  OHBCHB_2021-08-02_21-54-47_patch_crop

And with vertex rounding enabled:
* Without any changes ([full image](https://user-images.githubusercontent.com/8334194/127970068-be112aac-7689-4dd0-bf9e-f305680e09aa.png))
  OHBCHB_2021-08-02_23-11-43_norm_v_crop
* The shadow it's trying to cover ([full image](https://user-images.githubusercontent.com/8334194/127970310-2083f7eb-5956-43ab-bc91-dd71a8f405c5.png))
  OHBCHB_2021-08-02_23-17-08_shadow_v_crop
* If wrapping were disabled ([full image](https://user-images.githubusercontent.com/8334194/127970441-7b8862a1-acf4-4759-ae4b-baad4eebfc9e.png)):
  OHBCHB_2021-08-02_23-12-49_nowrap_v_crop
* With this patch ([full image](https://user-images.githubusercontent.com/8334194/127970520-f6908d75-f817-4557-bd6c-8016abbc6eb1.png)):
  OHBCHB_2021-08-02_23-13-54_patch_v_crop

My conclusion here: the game is actually _trying_ to compensate for some sort of offsetting at higher IRs, and would succeed at it except for the effect not quite fitting properly and thus triggering texture wrapping.  However, when it creates an EFB for shadows, it picks up stuff outside of the region where it was drawing, resulting in bad graphics; thus, the Vertex Rounding hack needs to be used.  But if Vertex Rounding is enabled, the offset correction the game is already doing ends up being incorrect, since it's actually drawing at an offset of 0 now; thus, the game also needs to be patched to use an offset of 0.  This also fixes the texture wrapping issue.  In short, Vertex Rounding + the patch is the way to get correct rendering at higher IRs.

<details><summary>Testing note</summary>

To get consistent images, I used the hardware fifoplayer in Dolphin.  To disable wrapping, I wrote 6180200000 at 00017816 (in Object 78).  To apply the patch, I wrote 98000400000000000000003f7fffffffffffff000000000000000000000000438000003f7fffffffffffff000000003f80000043800000000000003f7fffffffffffff3f8000000000000043800000438000003f7fffffffffffff3f8000003f800000 at 000178f6 (in Object 79); this uses coordinates of 0 and 1 instead of 0.001953125 and 1.00195 and was copied from a fifolog recorded with the patch.  To see the shadows, I disabled geometry for Object 79.
</details>

Data/Sys/GameSettings/GQPE78.ini
```diff
@@ -0,0 +1,12 @@
+# GQPE78 - SpongeBob SquarePants: Battle for Bikini Bottom
+
+[OnFrame_Enabled]
+# This game renders an EFB copy with texture repeating enabled
+# and it draws from texture coordinate 0.00195313 to 1.00195.
+# This only works on console and 1x IR due to low precision.
+# CopyRegionFix adjusts the region to not cause bugs at higher
```
JosJuice, line 7: Shouldn't this say `EFBCopyFix`?

Also, EFBCopyFix is a slightly strange name in that it doesn't have any spaces. Normally we do use spaces in patch names.

Data/Sys/GameSettings/GQPE78.ini
```diff
@@ -0,0 +1,12 @@
...
+# resolutions.
+$EFBCopyFix
+[OnFrame]
+$EFBCopyFix
```
JosJuice, line 11: Normally we put the `OnFrame_Enabled` section below the `OnFrame` section.

PR #9990, opened 2021-08-03
VideoCommon: Add shader logic ops support for Apple Silicon GPUs

OatmealDome:
This PR adds a shader logic ops implementation for Apple Silicon GPUs. While Metal does not support logic ops itself, it *does* support framebuffer fetch, allowing us to implement it in the fragment shader. (Note that Intel, AMD, and NVIDIA GPUs do *not* support framebuffer fetch, so this PR will not solve the logic ops problem on Intel Macs.)

Fixes Kirby's Air Ride and more on M1 Macs.

**Depends on**: #9981, #9988

### DIFF
--- a/Source/Core/VideoCommon/PixelShaderGen.cpp
+++ b/Source/Core/VideoCommon/PixelShaderGen.cpp
@@ -318,7 +318,7 @@ PixelShaderUid GetPixelShaderUid()
   BlendingState state = {};
   state.Generate(bpmem);
 
-  if (state.usedualsrc && state.dstalpha && g_ActiveConfig.backend_info.bSupportsFramebufferFetch &&
+  if (state.usedualsrc && g_ActiveConfig.backend_info.bSupportsFramebufferFetch &&
       !g_ActiveConfig.backend_info.bSupportsDualSourceBlend)
   {
     uid_data->blend_enable = state.blendenable;
@@ -613,13 +613,8 @@ ShaderCode GeneratePixelShaderCode(APIType api_type, const ShaderHostConfig& hos
     }
   }
 
-  // Only use dual-source blending when required on drivers that don't support it very well.
-  const bool use_dual_source =
-      host_config.backend_dual_source_blend &&
-      (!DriverDetails::HasBug(DriverDetails::BUG_BROKEN_DUAL_SOURCE_BLENDING) ||
-       uid_data->useDstAlpha);
-  const bool use_shader_blend =
-      !use_dual_source && (uid_data->useDstAlpha && host_config.backend_shader_framebuffer_fetch);
+  const bool use_dual_source = host_config.backend_dual_source_blend;
+  const bool use_shader_blend = !use_dual_source && host_config.backend_shader_framebuffer_fetch;
 
   if (api_type == APIType::OpenGL || api_type == APIType::Vulkan)
   {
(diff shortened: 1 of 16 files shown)


Dentomologist:
@dolphin-emu-bot rebuild

pizuz:
Do you know whether nVidia supports this? I could test.

JMC47:
Confirmed on discord to fix things on Apple Silicon GPUs.

OatmealDome:
@pizuz

> Do you know whether nVidia supports this? I could test (late 2012 iMac with nVidia on macOS Catalina).

Sorry, I forgot NVIDIA existed. Only Apple Silicon GPUs support framebuffer fetch, so this PR will not work on NVIDIA GPUs.

Source/Core/VideoCommon/VideoConfig.h
```diff
@@ -182,6 +182,7 @@ struct VideoConfig final
   struct
   {
     APIType api_type;
+    APIType real_api_type; // the underlying API (for example, Metal if using MoltenVK)
```
iwubcode, line 185: I'm not that familiar with Metal but not a huge fan of this approach.  Right now it's rather simple but it seems like it could blow out the complexity later if we have multiple apis and 'real' apis.  It also means every time we check our api, we have to question "did someone mean to check the 'real' api or did they intentionally avoid it?".

I think I'd rather pick one of two solutions:

1.  Check for api Vulkan and also do a preprocessor check for APPLE.  This would match what we're already doing in a few places.  Only negative is this doesn't work if later we add a second API that loads off Vulkan but isn't locked to a specific hardware.
2. Introduce a new "Metal" API type.  It creates a vulkan backend but the api type is passed in and that's where you load the metal library.  Then everywhere in code, we check for the Metal api.  That means a lot of updates to be done but it becomes very clear what apis are doing what logic.

OatmealDome: Yeah, I think the best and least invasive solution would be 1. I added the "real" type on the off chance a scenario like you suggested would happen (second API type that loads off another) but I don't think the chances of that are very high right now.

Source/Core/VideoCommon/PixelShaderGen.cpp
```diff
@@ -620,10 +621,46 @@ ShaderCode GeneratePixelShaderCode(APIType api_type, const ShaderHostConfig& hos
 
   const bool use_dual_source = host_config.backend_dual_source_blend;
   const bool use_shader_blend = !use_dual_source && host_config.backend_shader_framebuffer_fetch;
+  const bool use_shader_logic_op =
+#ifdef __APPLE__
+      !host_config.backend_logic_op && host_config.backend_shader_framebuffer_fetch;
+#else
+      false;
```
OatmealDome, line 628: A separate PR to add shader logic ops support on OpenGL ES (Android) will be opened after this is merged.

quinton-ashley:
merge please!

OatmealDome:
@quinton-ashley This PR is currently being blocked by some necessary infrastructure work on the Dolphin Project's part and can't be merged until that is completed.

quinton-ashley:
ah okay, what's the eta on that being done?

OatmealDome:
Starting over on a new branch, will open another PR when it's ready.

PR #9991, opened 2021-08-03
Jit64: Fix BATAddressLookup bit test

JosJuice:
BT sets the carry flag, not the zero flag.

~~Hopefully fixes https://bugs.dolphin-emu.org/issues/12608 (I don't have the game for testing).~~


JMC47:
Does not fix https://bugs.dolphin-emu.org/issues/12608

PR #9993, opened 2021-08-03
VI: Implement post-scanout XFB output

Techjar:
This adds about a frame of latency, and since most games don't change VI registers during scanout, we can get away with outputting the XFB at the start of scanout. WWE Crush Hour is the (only currently known) exception, which has flickering problems when doing it this way.

This adds a path to perform the output at the end of scanout, and gates it behind an option which defaults to using the latency-reducing pre-scanout path.

I also included the appropriate INI change.

### DIFF
--- a/Source/Core/Core/HW/VideoInterface.cpp
+++ b/Source/Core/Core/HW/VideoInterface.cpp
@@ -13,6 +13,7 @@
 #include "Common/Config/Config.h"
 #include "Common/Logging/Log.h"
 
+#include "Core/Config/GraphicsSettings.h"
 #include "Core/Config/MainSettings.h"
 #include "Core/Config/SYSCONFSettings.h"
 #include "Core/ConfigManager.h"
@@ -757,7 +758,7 @@ static void LogField(FieldType field, u32 xfb_address)
                 GetTicksPerOddField());
 }
 
-static void BeginField(FieldType field, u64 ticks)
+static void OutputField(FieldType field, u64 ticks)
 {
   // Could we fit a second line of data in the stride?
   // (Datel's Wii FreeLoaders are the only titles known to set WPL to 0)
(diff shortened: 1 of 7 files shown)


iwubcode (approved):
Code LGTM.  Untested

JMC47:
Can't read the code, but tested.  Fixes WWE Crush Hour.

Source/Core/Core/HW/VideoInterface.cpp
```diff
@@ -811,16 +812,26 @@ static void BeginField(FieldType field, u64 ticks)
...
+  // going to change the VI registers while a frame is scanning out.
+  if (Config::Get(Config::GFX_HACK_EARLY_XFB_OUTPUT))
+    OutputField(field, ticks);
 }
 
-static void EndField()
+static void EndField(FieldType field, u64 ticks)
 {
+  // If the game does change VI registers while a frame is scanning out, we need to defer output
+  // until the end when all those changes have been collated.
```
phire, line 830: Not accurate. This code-path only uses the last update to VI registers, all other updates get discarded. 

In the case of WWE Crush Hour, there is probably a tiny bit of tearing that we are not emulating. The 

    // To correctly handle that case we would need to collate all changes
    // to VI during scanout

comment from before should be preserved for future reference, but I don't really think it's worth actually implementing that just to emulate a bit of glitchyness in this one game.

phire (changes requested):

Source/Core/Core/HW/VideoInterface.cpp
phire: Maybe also comment to state that this still isn't accurate.

phire:
To be clear, this isn't more accurate.

It's just inaccurate in a different way that produces an acceptable result with this one game.

Source/Core/Core/HW/VideoInterface.cpp
Techjar: Isn't that essentially what collating the changes is though?

phire: collating means collecting (and implies actually using them) rather than discarding.

Shouldn't be too hard to actually implement in a hardware implementation. Just a collect a list of  regions, with the vi properties and XFB buffers they refer to. Then issue a draw call for each region to draw that section of XFB into the final output buffer.

The rendering cost when there is only one update as expected will almost identical to the current implementation. You just pay a tiny bit of overhead for each additional region that needs to be drawn (and slightly more overhead again when a region starts outside of hblank)

But IMO, not worth implementing, unless some demoscene intro exploits it.

phire (approved):

PR #9994, opened 2021-08-04
GameINI: SpongeBob SquarePants Movie QoL Settings

JMC47:
SpongeBob SquarePants: The Movie suffers from all of the same issues as Battle for Bikini Bottom.  These patches are not based on any existing cheats or codes, this one isn't as loved as Battle for Bikini Bottom and didn't receive an action replay code to fix everything.  There *is* a cheat that disables shadowing, but that wasn't good enough, so I sought out to make a new patch that does the same thing as the one for Battle for Bikini Bottom.

Heavy guidance was given by Pokechu22, who helped walking me through ghidra to find the memory addresses.  The game is rendering efb copies with coordinates 0.00390625 and 1.00390625, which cause severe problems at higher resolutions.  This patch adjusts them to render at 0 and 1, and enables the Vertex Rounding Hack to repair things at higher resolutions.

I tested the first stage of each version of the game, but I'm tired and didn't test any further.  And for those reviewing, for some reason the PAL and NTSC addresses are identical.  I triple checked that.

### DIFF
--- a/Data/Sys/GameSettings/GGV.ini
+++ b/Data/Sys/GameSettings/GGV.ini
@@ -1,8 +1,20 @@
 # GGVD78, GGVE78, GGVP78, GGVX78 - The SpongeBob SquarePants Movie
+
 [Core]
+# Values set here will override the main Dolphin settings.
+
+[OnLoad]
+# Add memory patches to be loaded once on boot here.
+
 [OnFrame]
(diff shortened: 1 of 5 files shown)


leoetlino:
Related PR: #9989 (with a great explanation here: https://github.com/dolphin-emu/dolphin/pull/9989#issuecomment-891587356)

PR #9998, opened 2021-08-06
PixelShaderGen: Pass LOD bias to fragment shader and use when sampling texture on Metal and OpenGL ES

OatmealDome:
In Twilight Princess, the iris of Midna's eyes disappears as the camera moves further away from her model. This is done by using two mipmaps: one with her iris, and one with just the pale yellow of her sclera. The game sets the distance at which the transition between the two mipmaps occurs by setting the LOD bias value to a large negative value.

To emulate LOD bias in the Vulkan backend, ``mipLodBias`` in the ``VkSamplerCreateInfo`` for the texture's sampler is set. Unfortunately, Metal does not support setting LOD bias in a sampler, so `mipLodBias` is ignored by MoltenVK. On OpenGL ES, the sampler parameter for setting LOD bias (``GL_TEXTURE_LOD_BIAS``) is unsupported, unlike desktop OpenGL.

This causes Midna's iris to only appear when the camera is very close to her.

However, we can pass through a bias value in the ``texture()`` call in the fragment shader, which both OpenGL ES and Metal support. This is what this PR implements.

Fixes https://bugs.dolphin-emu.org/issues/11568 and https://bugs.dolphin-emu.org/issues/11993..

<details>
<summary>Screenshots (macOS)</summary>

Broken:

image

Fixed:

image

</details>

### DIFF
--- a/Source/Core/VideoBackends/D3D/D3DMain.cpp
+++ b/Source/Core/VideoBackends/D3D/D3DMain.cpp
@@ -106,6 +106,7 @@ void VideoBackend::FillBackendInfo()
   g_Config.backend_info.bSupportsSSAA = true;
   g_Config.backend_info.bSupportsShaderBinaries = true;
   g_Config.backend_info.bSupportsPipelineCacheData = false;
+  g_Config.backend_info.bSupportsLodBiasInSampler = true;
   g_Config.backend_info.bSupportsLogicOp = D3D::SupportsLogicOp(g_Config.iAdapter);
 
   g_Config.backend_info.Adapters = D3DCommon::GetAdapterNames();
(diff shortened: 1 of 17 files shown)


OatmealDome:
@pizuz Assuming you're the same pizuz that opened the bug report, could you test Twilight Princess with this PR?

https://dl.dolphin-emu.org/prs/72/8c/pr-9998-dolphin-latest-universal.dmg

pizuz:
Yup, that's me. Working as intended.

Current master:
Screenshot 2021-08-06 at 18 56 09

This build:
Screenshot 2021-08-06 at 19 01 50

There is one minor quibble, but this seems to be present on OpenGL, too, so this might as well happen on console: When summoning midna in human Link form as a shadow, for a split second before she turns to Link, the iris is kinda fading in. I cannot test on console, unfortunately. It looks like this:
Screenshot 2021-08-06 at 19 01 20

I'm going through Thorn's Let's Play, right now. Maybe I can find something.

Anyway, thanks for fixing this :)

OatmealDome:
Not really sure why ``ogl-lin-mesa`` is showing minuscule differences when a bias of ``0.0`` should be equivalent to not providing any at all (at least according to what I read).

I can't even see the difference without setting the diff to bright.

OatmealDome:
Closed in favour of #9956.

PR #9999, opened 2021-08-06
Jit: Re-add dcbx masking

JosJuice:
When making PR #9957, I checked whether the ~0x1f masking in dcbx actually was necessary. I came to the conclusion that it wasn't, so I removed it. However, I hadn't checked the second half of InvalidateICache closely enough – the masking is actually needed.

This commit re-adds the masking, but this time in C++ code instead of in jitted code in order to save icache. Though I suppose the difference doesn't matter all that much, since this is in farcode and all...

Hopefully fixes https://bugs.dolphin-emu.org/issues/12612 (not tested).


AdmiralCurtiss:
Yes, this fixes TP under MMU.

AdmiralCurtiss (approved):

lioncash (approved):

PR #7614, opened 2018-12-08
InputCommon: Add Windows.Gaming.Input to ControllerInterface.

jordan-woyak:
Why?
To support more than four Xbox 360/One controllers: https://bugs.dolphin-emu.org/issues/11400
To support Xbox One trigger rumble motors.

Xbox 360 and DS4 controllers (over USB) work as expected with analog triggers and rumble support.
Battery levels of Xbox controllers are also exposed.

PR #10000, opened 2021-08-06
Add SDL motion input and rumble support

jambonmcyeah:
Add motion input and rumble support for the SDL controller interface, inspired by yuzu-emu/yuzu#6244. This is useful for Windows and MacOS users since it removes the need to have a DSU server to get motion input.


JosJuice:
@dolphin-emu-bot rebuild

Source/Core/InputCommon/ControllerInterface/SDL/SDL.cpp
```diff
@@ -487,4 +521,14 @@ ControlState Joystick::Hat::GetState() const
 {
   return (SDL_JoystickGetHat(m_js, m_index) & (1 << m_direction)) > 0;
 }
+#ifdef USE_SDL_GAMECONTROLLER
+
+ControlState Joystick::MotionInput::GetState() const
+{
+  float data[3];
```
lioncash, line 528:
```suggestion
  std::array<float, 3> data{};
  SDL_GameControllerGetSensorData(m_gc, m_type, data.data(), data.size());
```

Source/Core/InputCommon/ControllerInterface/SDL/SDL.h
```diff
@@ -137,6 +147,40 @@ class Joystick : public Core::Device
   };
 #endif
 
+#if SDL_VERSION_ATLEAST(2, 0, 9)
+  class Motor : public Output {
+  public:
+    Motor(SDL_Joystick* js) : m_js(js) {};
```
lioncash, line 153:
```suggestion
    explicit Motor(SDL_Joystick* js) : m_js(js) {};
```

jambonmcyeah:
I'm not familiar with Visual Studio so my approach of adding SDL into Externals is probably wrong. Would appreciate if someone can help me with that. 😃

AdmiralCurtiss:
Yeah considering the size of the SDL, it's probably a better idea to instead add an actual git submodule of `https://github.com/libsdl-org/SDL.git`.

jambonmcyeah:
> Yeah considering the size of the SDL, it's probably a better idea to instead add an actual git submodule of `https://github.com/libsdl-org/SDL.git`.
Ok, I've done that and fixed the visual studio files also.

jambonmcyeah:
I think this is ready for review now

jordan-woyak:
@dolphin-emu-bot build.

OatmealDome:
Perhaps if SDL isn't found when ``Source/Core/InputCommon/CMakeLists.txt`` checks, the SDL from Externals should be used.

jambonmcyeah:
> Perhaps if SDL isn't found, the SDL from Externals should be used. (There are checks in `Source/Core/InputCommon/CMakeLists.txt`, but they might have to moved to the root CMakeLists?)

Just finished that as well. Let me know if there's more changes needed.

OatmealDome:
@dolphin-emu-bot rebuild

jambonmcyeah:
This should hopefully fix the FreeBSD build failure.

OatmealDome:
@dolphin-emu-bot rebuild

jambonmcyeah:
This should fix unit tests on Windows

jambonmcyeah:
This should be everything fixed, just need to run it through CI to confirm that.

OatmealDome:
@dolphin-emu-bot rebuild

Simonx22:
I have tested this on Windows 10 v20H1 with my DualShock 4 Controller:
- rumble works fine
- buttons can be mapped just fine
- motion control does not work

@OatmealDome also tried this on macOS 11 with a Switch Pro Controller, everything works except motion controls.

Is there anything special we need to do to get motion controls working?

jambonmcyeah:
Maybe it's some background programs that are interfering with it. If you guys have Steam, DS4Windows or other programs that interact with controllers try closing them.

Also, the system SDL on the macOS build bot might be too old, version 2.0.14 or newer is required for motion controls. 

I have personally tested on Windows 10 v21H1 and it works for me.

jambonmcyeah:
Also macOS might have permission issues, so the solution to that is going to `Security & Privacy” -> Privacy -> “Input Monitoring` and allowing Dolphin

OatmealDome:
Hmm, the universal buildbot uses SDL 2.0.14, so that should be fine. I tried my Pro Controller over Bluetooth and USB with Input Monitoring enabled for Dolphin, and motion isn't being registered in the controller configuration.

If I right click on the binding button to look at the list of inputs, none of the ``Accel`` or ``Gyro`` inputs appear in the list.

mbc07:
I tested with a DualShock 4 v2 on Windows 10 21H1 and it works (including motion) but:

* SDL doesn't detect the controller when connected over Bluetooth, only over USB
* Touchpad isn't accessible (only the touchpad click)
* It's not possible to access the rumble motors individually, SDL reports only one motor which activates both

Those probably are SDL limitations, but not working over Bluetooth kinda defeats the hability of reading the motion sensors. Regardless it's still a great addition especially for macOS, which has a single DSU Server available that I heard is quite finicky to use...

mbc07:
Ah, nevermind, SDL can see the DualShock 4 over BT on Win10 (I forgot to add Dolphin to the HidHide whitelist earlier), however it can't access the motion sensors over BT (they don't appear in the Advanced Mapping window), rumble also doesn't work over BT (it can be selected in the Advanced Mapping windows but the controller doesn't rumble at all)...

jambonmcyeah:
Adding touchpad support and individual motor rumble is doable. I'll add support for those after I figure out why it's not working for some people.
Bluetooth should work in theory, it works on both of my computers. I'll try to figure out why it's not working for you guys.

Simonx22:
Quick update on my issue: a reboot fixed the motion control issues. The pointer works fine now but the actual accelerometer itself shifts heavily to the right. I have tested this with Mario Kart Wii. Calibrating the motion controls doesn't seem to help much. It seems like the accelerometer is shifting further the longer I drive. Not sure if the shifting to the right is an SDL issue or something else with Dolphin.

mbc07:
@jambonmcyeah maybe it's the slightly communication differences between DualShock 4 v1 and v2 bitting again? The v2 (the one I have) is the newer model released with the PS4 Slim, that has the light bar visible through the touchpad, v1 is the model that launched with the PS4 "fat". PR #7614 also exhibited different behavior depending of the DualShock 4 model...

jambonmcyeah:
The one I have seems to be the V2 version which may explain things.

jambonmcyeah:
Also for the shifting problem I have it as well with the gyroscope. Have no idea what's causing the problem

jambonmcyeah:
It shouldn't be SDL's problem since from my testing it doesn't seem to happen with other emulator which uses SDL like yuzu and cemu. 
I looked through yuzu's SDL motion implementation and it doesn't seem they are doing anything special.

infirit:
Sdl2 gamepad also works on linux so I would like to suggest to also enable it on linux as well. I am playing mario odyssey right now with my pro controller, rumble and motion work without joycond and hidnintendo kernel modules.

v1993:
@infirit If enabled on Linux, this would render evdev nodes of controller unusable due to limits of HIDAPI. While I understand that not needing to install hid_nintendo is handy, breaking working approach is not nice. In addition, on Linux SDL does not build with HIDAPI support by default (it would need to be explicitly enabled during build like it's done in yuzu) exactly to avoid conflict with evdev, so that's something to consider as well.

Becuase dolphin already supports evdev motion, only hid_nintendo is needed. Joycond and joycond-cemuhook are not.

RinMaru:
While this gets rid of the need for cemuhook and better joy would it also be able to handle TN a wiimotes better since syncing is a problem?

phire:
> Becuase dolphin already supports evdev motion, only hid_nintendo is needed

Looks like native support for switch controllers will land in Linux 5.16, so you won't even need an extra kernel module. It will just work with dolphin's evdev.

Source/Core/DolphinQt/DolphinQt.vcxproj
```diff
@@ -42,6 +42,9 @@
     <Manifest>
       <AdditionalManifestFiles>DolphinQt.manifest;%(AdditionalManifestFiles)</AdditionalManifestFiles>
     </Manifest>
+    <PostBuildEvent>
```
shuffle2, line 45: use a Copy task instead (see the end of this file , and add it there)

Source/Core/DolphinLib.vcxproj
```diff
@@ -22,6 +22,11 @@
   <ImportGroup Condition="'$(Platform)'=='ARM64'">
     <Import Project="DolphinLib.ARM64.props" />
   </ImportGroup>
+  <ItemDefinitionGroup>
```
shuffle2, line 25: this belongs in base.props

.gitignore
```diff
@@ -24,6 +24,9 @@ Source/Core/Common/scmrev.h
 *.VC.db
 .vs*/
 /Source/enc_temp_folder/
+/Source/x64
```
shuffle2, line 27: what is this for? probably should be removed

Source/UnitTests/UnitTests.vcxproj
```diff
@@ -30,6 +30,9 @@
       <AdditionalLibraryDirectories Condition="'$(Platform)'=='x64'">$(ExternalsDir)ffmpeg\lib;%(AdditionalLibraryDirectories)</AdditionalLibraryDirectories>
       <SubSystem>Console</SubSystem>
     </Link>
+    <PostBuildEvent>
+      <Command>copy /Y "$(SolutionDir)$(Platform)\$(Configuration)\SDL2.dll" "$(OutDir)"</Command>
```
shuffle2, line 34: should be removed

Source/dolphin-emu.sln
```diff
@@ -75,290 +75,374 @@ Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "zstd", "..\Externals\zstd\z
 EndProject
 Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "mgba", "..\Externals\mGBA\mgba.vcxproj", "{864C4C8E-296D-3DBC-AD83-F1D5CB6E8EC6}"
 EndProject
+Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "SDL2", "..\Externals\SDL\VisualC\SDL\SDL.vcxproj", "{81CE8DAF-EBB2-4761-8E45-B71ABCCA8C68}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|ARM64 = Debug|ARM64
 		Debug|x64 = Debug|x64
+		Debug|x86 = Debug|x86
```
shuffle2, line 84: revert these x86 additions, remove x86 config from sdl

shuffle2:
tbh I'm not a fan of adding sdl, but i made some comments for improvements of the pr anyway
we take binary deps when they're *required*, but sdl just seems like bloat and cruft - and we haven't even built this version of sdl ourselves - not great.

shuffle2:
this PR should be closed now

SuperSamus:
> @infirit If enabled on Linux, this would render evdev nodes of controller unusable due to limits of HIDAPI. While I understand that not needing to install hid_nintendo is handy, breaking working approach is not nice. In addition, on Linux SDL does not build with HIDAPI support by default (it would need to be explicitly enabled during build like it's done in yuzu) exactly to avoid conflict with evdev, so that's something to consider as well.
>
> Becuase dolphin already supports evdev motion, only hid_nintendo is needed. Joycond and joycond-cemuhook are not.

I don't know if it's relevant, but on other emulators the SDL interface is enabled under Linux too.
For example, on DuckStation and PCSX2 (Qt), it's the _only_ input interface available on Linux.

SuperSamus:
> @infirit If enabled on Linux, this would render evdev nodes of controller unusable due to limits of HIDAPI. While I understand that not needing to install hid_nintendo is handy, breaking working approach is not nice. In addition, on Linux SDL does not build with HIDAPI support by default (it would need to be explicitly enabled during build like it's done in yuzu) exactly to avoid conflict with evdev, so that's something to consider as well.
>
> Becuase dolphin already supports evdev motion, only hid_nintendo is needed. Joycond and joycond-cemuhook are not.

I don't know if it's relevant, but on other emulators the SDL interface is enabled under Linux too.
For example, on [DuckStation](https://github.com/stenzek/duckstation/commit/68866f3174228194cd86c9caef1038e1721c96dc) and PCSX2 (Qt), it's the _only_ input interface available on Linux.
(For the former, only by default.)

PR #10004, opened 2021-08-06
FifoDataFile: Improve error handling when reading

Pokechu22:
The main thing I'm targeting here is opening a 0-byte file, as currently that produces the nonsensical error message "The DFF's minimum loader version (1) exceeds the version of this FIFO Player (5)" (due to the check on `header.fileId != FILE_ID`).  I often end up opening a 0-byte file if I try to open it while it's still downloading (as firefox creates a 0-byte file as well as a .tmp file).

The other checks were added for completeness; note that `Common::IOFile` uses RAII so the manual close calls aren't needed.  I removed them for `Load` but haven't touched anything in `Save`.


Pokechu22:
`ss-timestone` appears to be corrupt (it has a bad offset for memory updates on the 2nd frame, and the 2nd and 3rd frames in general seem messed up), and has since been disabled on fifo.ci.

PR #10005, opened 2021-08-07
GameINI:  Fix EFB Copy Offset in more games

JMC47:
SpongeBob SquarePants: Battle for Bikini Bottom and The Movie Game, along with the two Incredibles titles suffer from an EFB offset issue described in earlier commits.  This commit adds support for the Incredibles 1/2 revisions and adds additional inis for SpongeBob special releases.

To the best of my knowledge, no The Incredibles: Rise of the Underminer GIQP78 exists, but it is marked on our wiki so I added it to the base INI as a possible revision.  I did not provide a code for it, though.  These probably need a really good look over as this is probably the most changes I've ever packed into a single pull request.

### DIFF
--- a/Data/Sys/GameSettings/GICD78.ini
+++ b/Data/Sys/GameSettings/GICD78.ini
@@ -1,8 +1,8 @@
-# GICE78 - The Incredibles
+# GICD78 - The Incredibles
 
 [OnFrame]
 $EFB Copy Fix
-0x803D2AD4:dword:0x00000000
+0x803D2A94:dword:0x00000000
 
 [OnFrame_Enabled]
 # This game renders an EFB copy with texture repeating enabled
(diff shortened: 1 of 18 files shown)


JosJuice:
> To the best of my knowledge, no Incredible 2 GIQP78 exists, but it is marked on our wiki so I added it to the base INI as a possible revision.

In that case, I think it would be better to not list it in the INI. We already have several cases of an INI not listing game IDs that definitely exist, and listing something that doesn't exist can cause confusion in the future. (Right now you're choosing to include this game ID in the INI just based on what the wiki says, and maybe in the future someone else will choose to include the game ID somewhere else based on what the INI and the wiki says...)

JMC47:
I removed the GIQP78 mention.

Tilka:
It's still mentioned in the shared ini.

JMC47:
I must have removed one of the valid gameids because I'm blind.  will fix tomorrow.

JMC47:
Actually removed GIQP78 this time around.  Corrected Incredibles 2 to The Incredibles: Rise of the Underminer, because that's the actual name and Incredibles 2 is something else.

PR #10006, opened 2021-08-07
AudioCommon: fix bogus error + cleanup

Tilka:
This gets rid of the bogus "AudioCommon/AudioCommon.cpp:181 E[Audio]: Error stopping stream." errors.

### DIFF
--- a/Source/Core/AudioCommon/PulseAudioStream.h
+++ b/Source/Core/AudioCommon/PulseAudioStream.h
@@ -20,7 +20,7 @@ public:
   ~PulseAudio() override;
 
   bool Init() override;
-  bool SetRunning(bool running) override { return running; }
+  bool SetRunning(bool running) override { return true; }
   static bool isValid() { return true; }
   void StateCallback(pa_context* c);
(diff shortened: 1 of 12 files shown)


Source/Core/AudioCommon/PulseAudioStream.h
```diff
@@ -20,7 +20,7 @@ class PulseAudio final : public SoundStream
   ~PulseAudio() override;
 
   bool Init() override;
-  bool SetRunning(bool running) override { return running; }
+  bool SetRunning(bool running) override { return true; }
```
leoetlino, line 23: Should we add a comment to SoundStream::SetRunning to clarify what SetRunning is supposed to return (whether the operation succeeded, rather than whether the stream is now running)?

leoetlino (approved):

Source/Core/AudioCommon/PulseAudioStream.h
Tilka: meh^^

PR #10007, opened 2021-08-07
Jit64: dcbx loop detection for improved performance when invalidating large memory regions.

AdmiralCurtiss:
So unfortunately, I have to report that our recent dcbx fixes have once again made Mario Sunshine's boot sequence and A Boy and his Blob in general slow again.

I profiled this a bit and it turns out the big performance issue comes from a game calling dcbx, failing the BAT check on memory that is not mapped in the BAT, and then `JitCache_TranslateAddress()` having to go through the rather slow `TranslatePageAddress()`. This wouldn't be so bad once, but:

- Mario Sunshine calls this *on the entire address space*, one cache line at a time.
- A Boy and his Blob does this all the time in the background, on multiple ~266KB regions.

This, as it turns out, is slow. Profiler in Mario Sunshine's boot sequence shows >70% CPU usage in just the page translation code. Which makes sense when you think about it, we're effectively invoking InvalidateICache() about 134 million times.

This PR tries to detect such loops and transform them into larger invalidation calls, limited by the remaining downcount, instead of one per cache line.

Probably want an ARM implementation of this too.

### DIFF
--- a/Source/Core/Core/PowerPC/Jit64/Jit_LoadStore.cpp
+++ b/Source/Core/Core/PowerPC/Jit64/Jit_LoadStore.cpp
@@ -234,20 +234,94 @@ void Jit64::dcbx(UGeckoInstruction inst)
   INSTRUCTION_START
   JITDISABLE(bJITLoadStoreOff);
 
-  X64Reg addr = RSCRATCH;
-  X64Reg value = RSCRATCH2;
+  // Check if the next instructions match a known looping pattern:
+  // - dcbx rX
+  // - addi rX,rX,32
+  // - bdnz+ -8
+  const bool make_loop = inst.RA == 0 && inst.RB != 0 && CanMergeNextInstructions(2) &&
+                         (js.op[1].inst.hex & 0xfc00'ffff) == 0x38000020 &&
+                         js.op[1].inst.RA_6 == inst.RB && js.op[1].inst.RD_2 == inst.RB &&
+                         js.op[2].inst.hex == 0x4200fff8;
+
   RCOpArg Ra = inst.RA ? gpr.Use(inst.RA, RCMode::Read) : RCOpArg::Imm32(0);
-  RCOpArg Rb = gpr.Use(inst.RB, RCMode::Read);
+  RCX64Reg Rb = gpr.Bind(inst.RB, make_loop ? RCMode::ReadWrite : RCMode::Read);
   RCX64Reg tmp = gpr.Scratch();
   RCX64Reg effective_address = gpr.Scratch();
   RegCache::Realize(Ra, Rb, tmp, effective_address);
 
-  // Translate effective address to physical address.
+  RCX64Reg loop_counter;
+  if (make_loop)
+  {
+    // We'll execute somewhere between one single cacheline invalidation and however many are needed
+    // to reduce the downcount to zero, never exceeding the amount requested by the game.
+    // To stay consistent with the rest of the code we adjust the involved registers (CTR and Rb)
+    // by the amount of cache lines we invalidate minus one -- since we'll run the regular addi and
+    // bdnz afterwards! So if we invalidate a single cache line, we don't adjust the registers at
+    // all, if we invalidate 2 cachelines we adjust the registers by one step, and so on.
+
+    RCX64Reg& reg_cycle_count = tmp;
+    RCX64Reg& reg_downcount = effective_address;
+    loop_counter = gpr.Scratch();
+    RegCache::Realize(loop_counter);
+
+    // This must be true in order for us to pick up the DIV results and not trash any data.
+    static_assert(RSCRATCH == Gen::EAX && RSCRATCH2 == Gen::EDX);
+
+    // Alright, now figure out how many loops we want to do.
+    const u8 cycle_count_per_loop =
+        js.op[0].opinfo->numCycles + js.op[1].opinfo->numCycles + js.op[2].opinfo->numCycles;
+
+    // This is both setting the adjusted loop count to 0 for the downcount <= 0 case and clearing
+    // the upper bits for the DIV instruction in the downcount > 0 case.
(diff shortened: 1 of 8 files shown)


JMC47:
We probably need to do this before the end of the month for the progress report, since it'd be nice to have performance *and* compatibility.

Pokechu22:
For reference, [libogc's cache_asm.S](https://github.com/devkitPro/libogc/blob/master/libogc/cache_asm.S) is (I think) a disassembly of the relevant SDK functions.  You can see that almost all of the `Range` ones follow that pattern.  The only exceptions are [`__LCEnable`](https://github.com/devkitPro/libogc/blob/bd24a9b3f59502f9b30d6bac0ae35fc485045f78/libogc/cache_asm.S#L362-L365) and [`LCAllocTags`](https://github.com/devkitPro/libogc/blob/bd24a9b3f59502f9b30d6bac0ae35fc485045f78/libogc/cache_asm.S#L437-L440), which have two cache-related instructions per loop instead of one.

JMC47:
I can confirm there's no more stuttering in Arc Rise Fantasia, Pokemon Colosseum's startup, and Super Mario Sunshine.  Mario and Sonic at the London Olympic Games and Happy Feet are not crashing with this optimization.

Source/Core/Core/PowerPC/JitCommon/JitCache.cpp
```diff
@@ -183,13 +183,46 @@ const u8* JitBaseBlockCache::Dispatch()
...
+void JitBaseBlockCache::InvalidateICache(u32 initial_address, u32 initial_length, bool forced)
+{
+  u32 address = initial_address;
+  u32 length = initial_length;
+  while (length > 0)
+  {
+    const auto translated = PowerPC::JitCache_TranslateAddress(address);
+
+    const bool address_from_bat = translated.valid && translated.translated && translated.from_bat;
+    const int shift = address_from_bat ? PowerPC::BAT_INDEX_SHIFT : 12;
```
JosJuice, line 203: There is a constant `HW_PAGE_INDEX_SHIFT` in MMU.cpp. Perhaps it should be moved to MMU.h so we can use it here.

Source/Core/Core/PowerPC/Jit64/Jit_LoadStore.cpp
```diff
@@ -234,18 +234,89 @@ void Jit64::dcbx(UGeckoInstruction inst)
...
+    DIV(32, R(reg_cycle_count));                              // RSCRATCH = downcount / cycle_count
+    LEA(32, RSCRATCH2, MDisp(loop_counter, -1));              // RSCRATCH2 = CTR - 1
+    // ^ Note that this CTR-1 implicitly handles the CTR == 0 case correctly.
+    CMP(32, R(RSCRATCH), R(RSCRATCH2));
+    CMOVcc(32, RSCRATCH2, R(RSCRATCH), CC_B);  // RSCRATCH2 = min(RSCRATCH, RSCRATCH2)
+
+    // RSCRATCH2 now holds the amount of loops to execute minus 1, which is the amount we need to
+    // adjust downcount, CTR, and Rb by to exit the loop construct with the right values in those
+    // registers.
+    SUB(32, PPCSTATE_CTR, R(RSCRATCH2));  // CTR -= RSCRATCH2
```
JosJuice, line 292: Would it be faster to do this?

```suggestion
    SUB(32, loop_counter, R(RSCRATCH2));
    MOV(32, PPCSTATE_CTR, loop_counter);  // CTR -= RSCRATCH2
```

Source/Core/Core/PowerPC/Jit64/Jit_LoadStore.cpp
```diff
@@ -264,6 +335,14 @@ void Jit64::dcbx(UGeckoInstruction inst)
   BT(32, R(value), R(addr));
   FixupBranch invalidate_needed = J_CC(CC_C, true);
 
+  if (make_loop)
+  {
+    ADD(32, R(effective_address), Imm8(32));
+    MOV(32, R(value), R(effective_address));
+    SUB(32, R(loop_counter), Imm8(1));
```
JosJuice, line 342: Would it be possible/advantageous to use `DEC` here?

AdmiralCurtiss: https://github.com/dolphin-emu/dolphin/blob/ffdc8538a162b1ca56e3943c5a78c02358053b2a/Source/Core/Common/x64Emitter.h#L409-L413

Though who knows how accurate that even is anymore, that comment is from 2008. But either way it's not in our Jit so...

Source/Core/Core/PowerPC/Jit64/Jit_LoadStore.cpp
AdmiralCurtiss: I genuinely have no idea! Maybe? It technically avoids a read from memory, I suppose.

Source/Core/Core/PowerPC/Jit64/Jit_LoadStore.cpp
JosJuice: Let's leave it as it is then.

Source/Core/Core/PowerPC/Jit64/Jit_LoadStore.cpp
AdmiralCurtiss: Since I had an extra register available I also did this for the downcount memory write, along with moving some instructions into the `downcount > 0` branch. Results should still be identical, but please review again.

Source/Core/Core/PowerPC/JitCommon/JitCache.cpp
AdmiralCurtiss: Yup, did that.

JosJuice (approved):
LGTM. I will submit a JitArm64 version as a follow-up PR once this is merged.

PR #10008, opened 2021-08-08
Android: Remove the EmulationState class

JosJuice:
The purpose of this class is to keep track of state which the emulation core is also keeping track of. This is rather risky – if we update the state of one of the two without updating the other, the two become out of sync, leading to some rather confusing problems. I think we're better off only keeping this state in the emulation core.

In particular, this change fixes a panic alert deadlock when booting a file that isn't a valid GC/Wii file while holding the phone in an orientation which does not match the orientation set in the Dolphin settings.

### DIFF
--- a/Source/Android/jni/MainAndroid.cpp
+++ b/Source/Android/jni/MainAndroid.cpp
@@ -442,6 +442,13 @@ JNIEXPORT void JNICALL Java_org_dolphinemu_dolphinemu_NativeLibrary_SurfaceDestr
   }
 }
 
+JNIEXPORT jboolean JNICALL Java_org_dolphinemu_dolphinemu_NativeLibrary_HasSurface(JNIEnv*, jclass)
+{
+  std::lock_guard guard(s_surface_lock);
+
+  return s_surf ? JNI_TRUE : JNI_FALSE;
+}
+
 JNIEXPORT jfloat JNICALL Java_org_dolphinemu_dolphinemu_NativeLibrary_GetGameAspectRatio(JNIEnv*,
                                                                                          jclass)
 {
--- a/Source/Android/jni/MainAndroid.cpp
+++ b/Source/Android/jni/MainAndroid.cpp
@@ -23,6 +23,7 @@
 #include "Common/CommonTypes.h"
 #include "Common/Event.h"
 #include "Common/FileUtil.h"
+#include "Common/Flag.h"
 #include "Common/IniFile.h"
 #include "Common/Logging/LogManager.h"
 #include "Common/MsgHandler.h"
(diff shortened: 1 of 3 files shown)


Source/Android/jni/MainAndroid.cpp
```diff
@@ -81,6 +81,7 @@ Common::Event s_update_main_frame_event;
 std::mutex s_surface_lock;
 bool s_need_nonblocking_alert_msg;
 
+Common::Flag s_is_booting;
```
leoetlino, line 84: missing include

Source/Android/jni/MainAndroid.cpp
```diff
@@ -442,7 +442,25 @@ JNIEXPORT void JNICALL Java_org_dolphinemu_dolphinemu_NativeLibrary_SurfaceChang
 JNIEXPORT void JNICALL Java_org_dolphinemu_dolphinemu_NativeLibrary_SurfaceDestroyed(JNIEnv*,
                                                                                      jclass)
 {
-  std::lock_guard<std::mutex> guard(s_surface_lock);
+  {
+    // If emulation continues running without a valid surface, we will probably crash,
+    // so pause emulation until we get a valid surface again. EmulationFragment handles resuming.
+
+    std::unique_lock<std::mutex> host_identity_guard(s_host_identity_lock);
```
leoetlino, line 449: just `std::unique_lock` is fine (with C++17 CTAD)

Source/Android/jni/MainAndroid.cpp
```diff
@@ -442,7 +442,25 @@ JNIEXPORT void JNICALL Java_org_dolphinemu_dolphinemu_NativeLibrary_SurfaceChang
...
+      host_identity_guard.unlock();
+      std::this_thread::sleep_for(std::chrono::milliseconds(1));
+      host_identity_guard.lock();
+    }
+
+    if (Core::GetState() == Core::State::Running)
+      Core::SetState(Core::State::Paused);
+  }
+
+  std::lock_guard<std::mutex> surface_guard(s_surface_lock);
```
leoetlino, line 463: ditto

Source/Android/jni/MainAndroid.cpp
```diff
@@ -442,6 +472,13 @@ JNIEXPORT void JNICALL Java_org_dolphinemu_dolphinemu_NativeLibrary_SurfaceDestr
   }
 }
 
+JNIEXPORT jboolean JNICALL Java_org_dolphinemu_dolphinemu_NativeLibrary_HasSurface(JNIEnv*, jclass)
+{
+  std::lock_guard<std::mutex> guard(s_surface_lock);
```
leoetlino, line 477: `std::lock_guard`

PR #10009, opened 2021-08-08
GameINI: Force safe texture cache on Resident Evil 3

OatmealDome:
This is already recommended by the wiki, but I think it should be set by default.

Without safe texture cache accuracy, item and weapon icons can disappear when picking an item up or transferring an item to/from an item box. In addition, when equipping/unequipping a weapon, the previous weapon icon (or lack thereof) may stay in its place instead of changing. This can cause confusion when playing the game.

On fast accuracy, this issue occurs often. I also tried with middle accuracy, and the frequency of occurrence decreased, but it still sometimes happens. Safe accuracy appears to eliminate the problem entirely.

<details>
<summary>Example</summary>

Here, I have the shotgun equipped. However, the equipped weapon icon tells me I have the pistol equipped.

image
</details>

### DIFF
--- a/Data/Sys/GameSettings/GLE.ini
+++ b/Data/Sys/GameSettings/GLE.ini
@@ -1,0 +1,6 @@
+# GLEE08, GLEJ08, GLEP08 - Resident Evil 3: Nemesis
+
+[Video_Settings]
+# On middle and fast, item and weapon icons can disappear or not change when picking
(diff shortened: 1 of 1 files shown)


JMC47:
Confirmed.

JosJuice:
The setting should be in the three-character INI unless the issue has been confirmed to not happen in the Japanese version of the game.

JMC47:
Yeah, should be 3 letter version.

PR #10010, opened 2021-08-08
PowerPC: Minor cleanup around JitCache_TranslateAddress().

AdmiralCurtiss:
The only functional change is around `TranslateResult::from_bat`, which was previously true but is now false if `MSR.IR` is off... but since no one actually checks that variable at the moment even that doesn't actually change behavior. I'm keeping it around though because I found a use for it in #10007.

### DIFF
--- a/Source/Core/Core/PowerPC/MMU.h
+++ b/Source/Core/Core/PowerPC/MMU.h
@@ -189,9 +189,17 @@ bool IsOptimizableGatherPipeWrite(u32 address);
 
 struct TranslateResult
 {
-  bool valid;
-  bool from_bat;
-  u32 address;
+  bool valid = false;
+  bool translated = false;
+  bool from_bat = false;
+  u32 address = 0;
+
+  TranslateResult() = default;
+  explicit TranslateResult(u32 address_) : valid(true), address(address_) {}
+  TranslateResult(bool from_bat_, u32 address_)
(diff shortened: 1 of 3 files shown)


Source/Core/Core/PowerPC/MMU.h
```diff
@@ -190,8 +190,19 @@ bool IsOptimizableGatherPipeWrite(u32 address);
 struct TranslateResult
 {
   bool valid;
```
lioncash, line 192: May want to consider directly initializing these to false/0 directly. It'll simplify the implementations of two of the constructors below.

AdmiralCurtiss: Reasonable, yes.

Source/Core/Core/PowerPC/MMU.h
```diff
@@ -189,9 +189,17 @@ bool IsOptimizableGatherPipeWrite(u32 address);
...
 {
-  bool valid;
-  bool from_bat;
-  u32 address;
+  bool valid = false;
+  bool translated = false;
+  bool from_bat = false;
+  u32 address = 0;
+
+  TranslateResult() {}
```
lioncash, line 197:
```suggestion
  TranslateResult() = default;
```

JosJuice (approved):

PR #10011, opened 2021-08-09
Split out code for serializing/deserializing cheat lines

JosJuice:
Some refactoring to prepare for my upcoming Android cheats PR. I figured it would make sense to submit this as a separate PR, since C++-heavy and Java-heavy PRs tend to get reviewed by different people.

### DIFF
--- a/Source/Core/Core/GeckoCodeConfig.cpp
+++ b/Source/Core/Core/GeckoCodeConfig.cpp
@@ -4,6 +4,7 @@
 #include "Core/GeckoCodeConfig.h"
 
 #include <algorithm>
+#include <optional>
 #include <sstream>
 #include <string>
 #include <vector>
@@ -176,8 +177,10 @@ std::vector<GeckoCode> LoadCodes(const IniFile& globalIni, const IniFile& localI
       {
         GeckoCode::Code new_code;
         // TODO: support options
-        new_code.original_line = line;
-        ss >> std::hex >> new_code.address >> new_code.data;
+        if (std::optional<GeckoCode::Code> code = DeserializeLine(line))
+          new_code = *code;
+        else
+          new_code.original_line = line;
         gcode.codes.push_back(new_code);
       }
       break;
@@ -251,4 +254,23 @@ void SaveCodes(IniFile& inifile, const std::vector<GeckoCode>& gcodes)
   inifile.SetLines("Gecko_Enabled", enabled_lines);
   inifile.SetLines("Gecko_Disabled", disabled_lines);
 }
+
+std::optional<GeckoCode::Code> DeserializeLine(const std::string& line)
+{
+  std::vector<std::string> items = SplitString(line, ' ');
+
+  GeckoCode::Code code;
+  code.original_line = line;
+
+  if (items.size() < 2)
+    return std::nullopt;
+
+  if (!TryParse(items[0], &code.address, 16))
+    return std::nullopt;
+  if (!TryParse(items[1], &code.data, 16))
+    return std::nullopt;
+
+  return code;
+}
+
 }  // namespace Gecko
--- a/Source/Core/Core/PatchEngine.cpp
+++ b/Source/Core/Core/PatchEngine.cpp
@@ -14,6 +14,8 @@
 #include <string>
 #include <vector>
 
+#include <fmt/format.h>
+
 #include "Common/Assert.h"
 #include "Common/IniFile.h"
 #include "Common/StringUtil.h"
@@ -43,8 +45,8 @@ const char* PatchTypeAsString(PatchType type)
   return s_patch_type_strings.at(static_cast<int>(type));
 }
 
-void LoadPatchSection(const std::string& section, std::vector<Patch>& patches, IniFile& globalIni,
-                      IniFile& localIni)
+void LoadPatchSection(const std::string& section, std::vector<Patch>* patches,
+                      const IniFile& globalIni, const IniFile& localIni)
 {
   const IniFile* inis[2] = {&globalIni, &localIni};
 
@@ -64,7 +66,7 @@ void LoadPatchSection(const std::string& section, std::vector<Patch>& patches, I
         // Take care of the previous code
         if (!currentPatch.name.empty())
         {
-          patches.push_back(currentPatch);
+          patches->push_back(currentPatch);
         }
         currentPatch.entries.clear();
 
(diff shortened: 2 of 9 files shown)


Source/Core/Core/PatchEngine.cpp
```diff
@@ -43,8 +46,55 @@ const char* PatchTypeAsString(PatchType type)
...
+  if (static_cast<size_t>(type) == s_patch_type_strings.size())
+    return std::nullopt;
+  entry.type = static_cast<PatchType>(type);
+
+  return entry;
+}
+
+std::string SerializeLine(const PatchEntry& entry)
+{
+  if (!entry.conditional)
```
lioncash, line 84: This can be inverted to get rid of the `!`

Source/Core/Core/GeckoCodeConfig.cpp
```diff
@@ -251,4 +250,21 @@ void SaveCodes(IniFile& inifile, const std::vector<GeckoCode>& gcodes)
   inifile.SetLines("Gecko_Enabled", enabled_lines);
   inifile.SetLines("Gecko_Disabled", disabled_lines);
 }
+
+bool DeserializeLine(const std::string& line, GeckoCode::Code* out)
```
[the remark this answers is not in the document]
JosJuice: Not possible. The `DeserializeLine` call earlier in this file intentionally accepts invalid Gecko codes.

Source/Core/Core/GeckoCodeConfig.cpp
```diff
@@ -251,4 +250,21 @@ void SaveCodes(IniFile& inifile, const std::vector<GeckoCode>& gcodes)
...
+{
+  std::vector<std::string> pieces = SplitString(line, ' ');
+  bool good = false;
+
+  if (pieces.size() > 0)
+    good &= TryParse(pieces[0], &out->address, 16);
+
+  if (pieces.size() > 1)
+    good &= TryParse(pieces[1], &out->data, 16);
```
BhaaLseN, line 263: This looks fishy. `false & anything` will never become `true`.

Source/Core/Core/GeckoCodeConfig.cpp
```diff
@@ -251,4 +250,21 @@ void SaveCodes(IniFile& inifile, const std::vector<GeckoCode>& gcodes)
...
+
+  out->original_line = line;
```
BhaaLseN, line 265: Should we trust the caller that `out` was initialized and points to something? Just to be on the safe side.

JosJuice: Is there anything reasonable to do if it doesn't point to valid memory?

Source/Core/Core/GeckoCodeConfig.cpp
JosJuice: Oops, yes. It should start off being `true`.

Source/Core/Core/GeckoCodeConfig.cpp
BhaaLseN: Fair enough. I mean, you should probably `return false` when `out == nullptr`. Not sure if you could even detect an invalid pointer. Probably should've worded it clearer in first place (wrt `out` being `null`.

JosJuice: I don't think it's worth checking for, since there's no case where passing in `nullptr` would be useful for the caller.

BhaaLseN: _Intentionally_, yes. But your choice, if you're certain it usually ends up being fine.

Source/Core/Core/GeckoCodeConfig.cpp
```diff
@@ -251,4 +250,21 @@ void SaveCodes(IniFile& inifile, const std::vector<GeckoCode>& gcodes)
   inifile.SetLines("Gecko_Enabled", enabled_lines);
   inifile.SetLines("Gecko_Disabled", disabled_lines);
 }
+
+bool DeserializeLine(const std::string& line, GeckoCode::Code* out)
+{
+  std::vector<std::string> pieces = SplitString(line, ' ');
```
AdmiralCurtiss, line 256: I'm not sure if this is ever really a problem in practice, but this function has quite different behavior compared to the old `ss >> std::hex >> new_code.address >> new_code.data;` code. Leading whitespace was previously okay but is now a failure, and any non-space whitespace anywhere near the code parts also no longer works.

If you're okay with breaking behavior anyway I would just go ahead and make this function return `std::optional<GeckoCode::Code>`. The other code above can just construct its own `GeckoCode::Code` with just the `original_line` set in the failure case.

Source/Core/Core/PatchEngine.cpp
```diff
@@ -45,6 +46,53 @@ const char* PatchTypeAsString(PatchType type)
...
+    if (!TryParse(items[3], &entry.comparand))
+      return std::nullopt;
+    entry.conditional = true;
+  }
+
+  const auto iter = std::find(s_patch_type_strings.begin(), s_patch_type_strings.end(), items[1]);
+  const ptrdiff_t type = std::distance(s_patch_type_strings.begin(), iter);
+  if (static_cast<size_t>(type) == s_patch_type_strings.size())
+    return std::nullopt;
+  entry.type = static_cast<PatchType>(type);
```
AdmiralCurtiss, line 77: I think it'd be more readable to just check for `iter == s_patch_type_strings.end()` here.

JosJuice: Yeah, now that I think about it outside of the context of just trying to translate the old code, I agree.

Source/Core/Core/GeckoCodeConfig.cpp
JosJuice: Hmm... I hadn't thought about that change in behavior. I think it should be fine for INI files (and having Gecko INI parsing be consistent with AR INI parsing is nice), but I'm unsure if the stuff we download from geckocodes.org might contain odd whitespace. However, we're not actually calling this from the geckocodes.org downloading function, so it should be fine.

Good point about how the code above could handle getting an optional. I've made it do that.

Source/Core/Core/GeckoCodeConfig.cpp
```diff
@@ -251,4 +254,23 @@ void SaveCodes(IniFile& inifile, const std::vector<GeckoCode>& gcodes)
...
 }
+
+std::optional<GeckoCode::Code> DeserializeLine(const std::string& line)
+{
+  std::vector<std::string> items = SplitString(line, ' ');
+
+  GeckoCode::Code code;
+  code.original_line = line;
+
+  if (items.size() < 3)
```
AdmiralCurtiss, line 265: `items.size() < 2`, right?

JosJuice: That's what I get for using PatchEngine as a reference without thinking...

AdmiralCurtiss (approved):
Looks good to me now. Untested though.

PR #10012, opened 2021-08-09
UberShaderVertex: Fix Tony Hawk Pro Skater 4

Pokechu22:
Fixes https://bugs.dolphin-emu.org/issues/12620

The changed code did not match the corresponding code in VertexShaderGen.  Some parts of the sky have 2 color channels in each vertex, while others only have 1, despite only color channel 0 being used and XFMEM_SETNUMCHAN being set to 1 for both of them.  The old code (from #4601) caused channel 0 to be set to channel 1 if the vertex contained both color channels but the number of channels was set to 1, which is wrong.

This code was also a bit funky because the [per-pixel-lighting code below](https://github.com/dolphin-emu/dolphin/blob/128e1029ddf09258dd476e8c5cc9fda9e85c2936/Source/Core/VideoCommon/VertexShaderGen.cpp#L463-L483) also affects things, but the same code existed in ubershaders.  I've tidied up the redundancies there.

### DIFF
--- a/Source/Core/VideoCommon/UberShaderVertex.cpp
+++ b/Source/Core/VideoCommon/UberShaderVertex.cpp
@@ -214,20 +214,36 @@ ShaderCode GenVertexShader(APIType api_type, const ShaderHostConfig& host_config
   if (num_texgen > 0)
     GenVertexShaderTexGens(api_type, num_texgen, out);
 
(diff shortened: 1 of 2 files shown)


JMC47:
Seems to fix the issue.

PR #10013, opened 2021-08-09
WiimoteReal: Remove WiimoteScannerDarwin

OatmealDome:
Title.

I think this should be done for several reasons:

- It pegs my CPU at 95% for scanning even when Dolphin is idle
- ``WiimoteScannerHidapi`` works fine on macOS (in fact, in my testing, ``hidapi`` always connects a Wiimote first before the Darwin-specific code...)
- Less macOS code to maintain


Techjar:
Is there maybe some systems/configurations which won't work with hidapi?

OatmealDome:
hidapi uses the standard macOS IOKit HID APIs which have been basically unchanged in a while, so I don't see a reason why it wouldn't work on some Macs. It should be compatible with at least macOS 10.9 Mavericks given when the version in Externals was last updated (Dolphin requires 10.12 Sierra minimum).

I think it should be compatible with all configurations.

Techjar:
Alright. I'm just trying to think of any reason why we were keeping the Darwin code around in the first place. Having 2 functionally similar scanners seems a tad silly.

lioncash (approved):

elsaxo:
> Title.
> 
> I think this should be done for several reasons:
> 
> * It pegs my CPU at 95% for scanning even when Dolphin is idle
> * `WiimoteScannerHidapi` works fine on macOS (in fact, in my testing, `hidapi` always connects a Wiimote first before the Darwin-specific code...)
> * Less macOS code to maintain

PR #10015, opened 2021-08-10
GameINI:  Add missing Vertex Rounding to GIQ.ini

JMC47:
This is necessary to actually fully fix the game at higher resolutions and was forgotten in the earlier batch of INI changes.


Pokechu22 (approved):
Confirmed that this is the only game modified in #10005 that is missing `VertexRounding`.  LGTM.

PR #10016, opened 2021-08-10
GameINI: Rayman Arena - Add Disable Culling Patch that fixes "Rise and Shrine" Hang

JMC47:
So funny story: the author of the "widescreen" patch for Rayman Arena (which is a patch ported from gamemasterplc's widescreen code Rayman 3) told me that it fixes the infamous hang in Rise and Shrine.  Now, Rayman Arena is a serious bastard of a game, and has hanged in Dolphin on this map as far back as I could get the game to run in Dolphin.  I thought to myself:

"How in the world could a widescreen patch work-around the bug???"

So I tried it.  And it worked.  So I began studying the code and found out it wasn't exactly a widescreen patch, it was a no culling patch.  But here's the thing: Rayman Arena already has almost no culling.  The only thing it culls is a few character models and item boxes at the start of the match.  The impact of the no culling cheat is actually very subtle, even when looking at statistics about what the game is rendering.

Before:
image
After:
image

The only difference is that the AI opponent is rendering.  In the past, @leoetlino and I have experimented with trying to adjust the timings of Rayman Arena to work in Dolphin and found that the *only* way to fully avoid the hang is to set Dolphin's Emulated GPU Clock to 1% or lower in Single Core.  And that leaves the game completely unplayable.  Users have gone to another solution: changing the emulated CPU clock to 150% or above and mashing to skip the cutscene at the start of this level where the hang occurs.

The thing that truly confuses me is that if you're playing without the patch, the level *never even starts rendering* as far as I can tell.  In order to get those statistics, I overclocked the CPU to 150% to get it to render a couple of frames before freezing, as at the base clock it will freeze while the loading screen is still up.

In terms of performance, the No Culling Patch is so lightweight that I see literally no difference in performance, and honestly I kind of think it performs slightly better but that may be confirmation bias creeping in.

For now, I'm just pushing this pull request to try and get some more eyes on yet another Ubisoft game doing something.  Does anyone have any idea why such a minor different matters when doing *wild* changes to Dolphin's timings don't?  This patch is useful as it gives us a side by side look with the failure state vs a successful state with minimal changes to the game or dolphin compared to previous efforts, but until we understand why the game is failing, I'm uncomfortable merging it.

### DIFF
--- a/Data/Sys/GameSettings/GYRE41.ini
+++ b/Data/Sys/GameSettings/GYRE41.ini
@@ -1,0 +1,14 @@
+# GYRE41 - Rayman Arena
+
+[OnFrame]
+$Disable Culling to Fix Rise and Shrine Hang
+0x8005EE90:dword:0x38600001
+0x8005EE94:dword:0x4E800020
+0x8005EF08:dword:0x38600001
+0x8005EF0C:dword:0x4E800020
+
+# This extremely minor patch that disables culling
+# in this game.  For some reason, this culling function
+# causes the "Rise and Shrine" hang in Dolphin.
+# There is no noticeable side-effects unless Dolphin's
+# built-in Widescreen Hack is enabled.


leoetlino:
Worth noting that the game hasn't actually crashed -- it's just sleeping forever. Its only thread never wakes up after a call to GXDrawDone (IIRC, it's been a while since I looked at the game). This is very likely to be a GPU timing issue.

RibShark:
To narrow it down further, it is the 2nd half of the codes (8005EF08 onwards) that is "fixing" this, so whatever culling is being disabled by the first 2 codes doesn't seem to affect this.

JMC47:
Addressed the comments.

I've disabled the hack by default.  I've also clarified it's a hang, not a crash in the comments.

AdmiralCurtiss:
Should we merge this?

Pokechu22:
I don't think the mostly-blank `GYR.ini` needs to exist, and the patch name is still "Disable Culling to Fix Rise and Shrine Crash" when it should maybe say "hang" like the comment, but I think it's fine to merge it otherwise (I have done no testing and haven't investigated this game myself)

JMC47:
I'll fix this in a bit then.

Pokechu22 (approved):

icup321:
Heads up, I just noticed that the game ID is wrong so it doesn't get detected by Dolphin, it's supposed to be G**RY**E41, not GYRE41.

JMC47:
lol oops.  I'll fix in the morning

PR #10020, opened 2021-08-11
D3D12: Transition compute textures to UAV state

K0bin:
When doing GPU texture decoding Dolphin never transitions the decoding texture to RESOURCE_STATE_UNORDERED_ACCESS despite writing to it in a compute shader. It gets created with `RESOURCE_STATE_PIXEL_SHADER_RESOURCE` so there isn't any implicit state promotion going on either.

This could fix https://bugs.dolphin-emu.org/issues/12094. I can't reproduce that bug though, it seems to only affect AMD GPUs.
Nvidia GPUs ignore the VkImageLayout with Vulkan so the reason for why it works on Nvidia might be something similar.


Pokechu22:
@dolphin-emu-bot rebuild

Tilka:
Confirmed that it fixes the bug.

K0bin:
Great!

PR #10022, opened 2021-08-12
DolphinQt: Add entitlement to allow debugger to attach in Debug builds

MerryMage:
In order to attach a debugger to a process with a Hardened Runtime (e.g. on an M1 mac), one needs the `com.apple.security.get-task-allow` entitlement.

Normally this entitlement is injected by XCode when running from XCode (assuming the XCode attribute `CODE_SIGN_INJECT_BASE_ENTITLEMENTS` is enabled), but this doesn't happen when building from a makefile or when you manually specify entitlements directly to codesign like this.

### DIFF
--- a/Source/Core/DolphinQt/CMakeLists.txt
+++ b/Source/Core/DolphinQt/CMakeLists.txt
@@ -527,12 +527,12 @@ if(APPLE)
     # Code sign make file builds
     add_custom_command(TARGET dolphin-emu
       POST_BUILD COMMAND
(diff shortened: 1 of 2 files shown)


JosJuice:
Is there a reason why we don't want to allow this for release builds too?

Tilka:
Debug features might make exploits easier and are probably disabled by default for a reason. Dolphin isn't exactly trying hard to be secure when running untrusted code but no need to add more weak points.

lioncash (approved):

PR #3841, opened 2016-05-14
Force Neo Geo games to use LLE Audio

RisingFog:
Another fix for [issue 9315](https://bugs.dolphin-emu.org/issues/9315).

---
This change is [Reviewable](https://reviewable.io/reviews/dolphin-emu/dolphin/3841)

PR #10024, opened 2021-08-13
GameSettings: force DSPLLE in remaining Neo Geo games

Tilka:
These two games were missed in PR #3841.

### DIFF
--- a/Data/Sys/GameSettings/ECH.ini
+++ b/Data/Sys/GameSettings/ECH.ini
@@ -1,0 +1,20 @@
+# ECHJJ8, ECHEJ8, ECHPJ8 - The King of Fighters '98
+
+[Core]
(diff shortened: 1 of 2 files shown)


Pokechu22:
Wouldn't it be better to just put this in `E.ini`?  Or is E used for both Neo Geo games and something else?

JMC47:
There was a definite reason we didn't put it in e.ini.  I don't remember what game overlapped.

JosJuice:
I looked it up, and it seems like E also covers "Virtual Console Arcade".

Tilka:
I've merged everything into three files.

JosJuice:
Dolphin doesn't support two-character game INIs.

Tilka:
Reverted.

PR #10026, opened 2021-08-14
Various GameINI Updates

PatrickFerry:
Mostly fixes for Wii games # - B

Some fixes to dual core issues, EFB Copies to Texture and Ram, Defer EFB Copies to RAM issues and Texture Cache Accuracy issues.

### DIFF
--- a/Data/Sys/GameSettings/RNO.ini
+++ b/Data/Sys/GameSettings/RNO.ini
@@ -1,7 +1,8 @@
 # RNOJ01, RNOP01 - Another Code R Kioku no Tobira
 
 [Core]
-# Values set here will override the main Dolphin settings.
+# Prevents save game corruption.
+SyncGPU = True
 
 [OnLoad]
 # Add memory patches to be loaded once on boot here.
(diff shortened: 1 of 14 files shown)


Data/Sys/GameSettings/RNO.ini
```diff
@@ -17,3 +18,5 @@
 [Video_Hacks]
 EFBToTextureEnable = False
 ImmediateXFBEnable = False
+# Multiple choice options displays some graphic corruption
+# DeferEFBCopies = False
```
Rumi-Larry, line 22: The "#" should be removed

PatrickFerry: I did it on purpose since it hurts performance a good deal but this issue is minor and only happens for brief moments a few times during the multiple choice selections in the game

nf6429: Wouldn't it be good to mention that in the comment? Nobody would know the reasoning for the # otherwise.

PR #10028, opened 2021-08-14
VolumeVerifier: Use correct IOS filename for development discs.

GerbilSoft:
Retail-signed discs use the format: IOS56-64-v5661.wad
Debug-signed discs use the format:  firmware.64.56.22.29.wad

Debug-signed discs usually have a 128 version of the firmware as well, since some devkits have 128 MB MEM2. (Retail has 64 MB.)

This fixes part of VolumeVerifier for debug discs. It's not validating the signatures properly. I attempted to add the debug root key (which is different from retail), and that helped, but ESDevice::VerifyContainer() was still getting an error from mbedtls, even though the public key appeared to be correctl

### DIFF
--- a/Source/Core/DiscIO/VolumeVerifier.cpp
+++ b/Source/Core/DiscIO/VolumeVerifier.cpp
@@ -630,14 +630,17 @@ bool VolumeVerifier::CheckPartition(const Partition& partition)
     // IOS9 is the only IOS which can be assumed to exist in a working state on any Wii
     // regardless of what updates have been installed. At least Mario Party 8
     // (RM8E01, revision 2) uses IOS9 without having it in its update partition.
-    bool has_correct_ios = tmd.IsValid() && (tmd.GetIOSId() & 0xFF) == 9;
+    const u64 ios_ver = tmd.GetIOSId() & 0xFF;
+    bool has_correct_ios = tmd.IsValid() && ios_ver == 9;
 
     if (!has_correct_ios && tmd.IsValid())
     {
       std::unique_ptr<FileInfo> file_info = filesystem->FindFileInfo("_sys");
       if (file_info)
       {
-        const std::string correct_ios = "IOS" + std::to_string(tmd.GetIOSId() & 0xFF) + "-";
+        const std::string ios_ver_str = std::to_string(ios_ver);
+        const std::string correct_ios =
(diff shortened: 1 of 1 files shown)


Pokechu22:
@dolphin-emu-bot rebuild

Source/Core/DiscIO/VolumeVerifier.cpp
```diff
@@ -630,14 +630,16 @@ bool VolumeVerifier::CheckPartition(const Partition& partition)
     // IOS9 is the only IOS which can be assumed to exist in a working state on any Wii
     // regardless of what updates have been installed. At least Mario Party 8
     // (RM8E01, revision 2) uses IOS9 without having it in its update partition.
-    bool has_correct_ios = tmd.IsValid() && (tmd.GetIOSId() & 0xFF) == 9;
+    const u8 ios_ver = tmd.GetIOSId() & 0xFF;
```
lioncash, line 633:
```suggestion
    const u64 ios_ver = tmd.GetIOSId() & 0xFF;
```

Otherwise this implicitly truncates to u8

GerbilSoft: A u8 is all that's needed here, since the IOS slot number is limited to 8-bit. (It's the low 8 bits of the title ID.) I could remove the `& 0xFF`, since that's implied by the truncation.

lioncash: Right, but using `u64` here preserves the type, has no implicit truncations taking place, and works totally fine. If u8 is used, its preferable to use a static_cast to make the conversion explicit as far as the type system cares, otherwise this can cause warnings.

Dentomologist:
@dolphin-emu-bot rebuild

JosJuice:
I don't have any development discs to verify this with, but I'll trust you that this is correct. However you'll need to fix the lint error before this can be merged.

JosJuice:
@dolphin-emu-bot rebuild

GerbilSoft:
Updated to switch to u64 and fix the lint error.

I have several RVT-R images, all of which have WADs in the update partition similar to the following:

```
firmware.128.56.21.29.wad
firmware.64.56.21.29.wad
```

These are both the same IOS version (equivalent to retail IOS56 v5405). The 64 version is for systems with 64 MB MEM2 (RVT-R Reader), while the 128 version is for systems with 128 MB MEM2 (RVT-H Reader, NDEV).

JosJuice (approved):

PR #10031, opened 2021-08-16
DSPInterpreter: Fix IsLess

Pokechu22:
`IsLess` would incorrectly return true if both `SR_OVERFLOW` and `SR_SIGN` are set, as `(sr & SR_OVERFLOW) != (sr & SR_SIGN)` becomes `SR_OVERFLOW != SR_SIGN` which is true as the two masks are different.  This broke in e651592ef56521f6db71f8a671fe8c20b948d338.

This issue only affected the DSP LLE Interpreter, and not the DSP LLE JIT.

I've also included a simple test case for this.  `ax0.l` (on the top left) is set to 0 if the instruction following `IFL` does not execute and to 1 if it is executed.

<details><summary>On console</summary>

Screenshot 2021-08-15 16-41-00
</details>
<details><summary>On Dolphin with JIT</summary>

OHBCHB_2021-08-15_17-23-54
</details>
<details><summary>On Dolphin before this change</summary>

OHBCHB_2021-08-15_16-50-00
</details>
<details><summary>On Dolphin with this change</summary>

OHBCHB_2021-08-15_16-58-44
</details>

### DIFF
--- a/Source/Core/Core/DSP/Interpreter/DSPInterpreter.cpp
+++ b/Source/Core/Core/DSP/Interpreter/DSPInterpreter.cpp
@@ -250,10 +250,7 @@ bool Interpreter::CheckCondition(u8 condition) const
   const auto IsCarry = [this] { return IsSRFlagSet(SR_CARRY); };
   const auto IsOverflow = [this] { return IsSRFlagSet(SR_OVERFLOW); };
   const auto IsOverS32 = [this] { return IsSRFlagSet(SR_OVER_S32); };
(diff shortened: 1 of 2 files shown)


phire (approved):

lioncash:
The commit message itself should probably contain the PR description (difference images aren't necessary though), so that the context of what's being fixed is more obvious (and stores that info directly in the repository).

LGTM otherwise though.

lioncash (approved):

PR #10032, opened 2021-08-16
DSP masking and sign extension fixes

Pokechu22:
The `cr` and `prod.h` registers are 8 bits and do not experience sign extension.  Also, the `SR_100` bit always seems to read back as 0, regardless of what's written to it `sr`.  Lastly, `ac0.h` and `ac1.h` do experience sign extension; this is already implemented, but the implementation worked incorrectly on the interpreter when the whole `acc0` or `acc1` register was updated and then `ac0.h` or `ac1.h` were read back directly.

<details><summary>Results</summary>

Apologies for how unreadable this is.  Think of it more as a table of thumbnails than something you can directly compare.

| Test | Hardware | Old JIT | Old Int | New JIT | New Int |
|------|----------|---------|---------|---------|---------|
| Less | Screenshot 2021-08-15 19-53-12 | OHBCHB_2021-08-15_19-55-44_Old_JIT | OHBCHB_2021-08-15_19-56-50_Old_Int | OHBCHB_2021-08-15_19-59-04_New_JIT | OHBCHB_2021-08-15_19-59-18_New_Int |
| FFFF | Screenshot 2021-08-15 22-10-38_FFFF | OHBCHB_2021-08-15_21-43-23_Old_JIT_FFFF | OHBCHB_2021-08-15_21-45-24_Old_Int_FFFF | OHBCHB_2021-08-15_21-47-10_New_JIT_FFFF | OHBCHB_2021-08-15_21-49-24_New_Int_FFFF |
| 0000 | Screenshot 2021-08-15 22-11-12_0000 | OHBCHB_2021-08-15_21-43-43_Old_JIT_0000 | OHBCHB_2021-08-15_21-45-26_Old_Int_0000 | OHBCHB_2021-08-15_21-47-12_New_JIT_0000 | OHBCHB_2021-08-15_21-49-28_New_Int_0000 |
| 007F | Screenshot 2021-08-15 22-11-26_007F | OHBCHB_2021-08-15_21-43-47_Old_JIT_007F | OHBCHB_2021-08-15_21-45-29_Old_Int_007F | OHBCHB_2021-08-15_21-47-14_New_JIT_007F | OHBCHB_2021-08-15_21-49-31_New_Int_007F |
| 0080 | Screenshot 2021-08-15 22-11-42_0080 | OHBCHB_2021-08-15_21-43-49_Old_JIT_0080 | OHBCHB_2021-08-15_21-45-31_Old_Int_0080 | OHBCHB_2021-08-15_21-47-16_New_JIT_0080 | OHBCHB_2021-08-15_21-49-33_New_Int_0080 |
| 0100 | Screenshot 2021-08-15 22-11-56_0100 | OHBCHB_2021-08-15_21-43-52_Old_JIT_0100 | OHBCHB_2021-08-15_21-45-32_Old_Int_0100 | OHBCHB_2021-08-15_21-47-18_New_JIT_0100 | OHBCHB_2021-08-15_21-49-36_New_Int_0100 |

For the less test, the JIT behaves the same as it did before and as it does now, as it handled sign extension properly.  Its result is not correct for the `sr` register, but that's a separate issue (the JIT seems to not update flags for overflow properly) which will be handled later.  The interpreter now exactly matches hardware.

For the other tests, the old JIT and old interpreter had the same incorrect behavior, and the new JIT and the new interpreter have the same behavior that matches hardware.
</details>

I'm not completely confident on the code for the JIT, which I based based on the code for `DSP_REG_ACH0`.  It seems to work, but the TODO about immediate values is a bit odd.

### DIFF
--- a/Source/Core/Core/DSP/Interpreter/DSPInterpreter.cpp
+++ b/Source/Core/Core/DSP/Interpreter/DSPInterpreter.cpp
@@ -687,7 +687,7 @@ void Interpreter::OpWriteRegister(int reg_, u16 val)
 
   switch (reg)
   {
-  // 8-bit sign extended registers. Should look at prod.h too...
+  // 8-bit sign extended registers.
   case DSP_REG_ACH0:
   case DSP_REG_ACH1:
     // sign extend from the bottom 8 bits.
@@ -720,10 +720,10 @@ void Interpreter::OpWriteRegister(int reg_, u16 val)
     state.r.wr[reg - DSP_REG_WR0] = val;
     break;
   case DSP_REG_CR:
-    state.r.cr = val;
+    state.r.cr = val & 0x00ff;
     break;
   case DSP_REG_SR:
-    state.r.sr = val;
+    state.r.sr = val & ~SR_100;
     break;
   case DSP_REG_PRODL:
     state.r.prod.l = val;
@@ -732,7 +732,8 @@ void Interpreter::OpWriteRegister(int reg_, u16 val)
     state.r.prod.m = val;
     break;
   case DSP_REG_PRODH:
-    state.r.prod.h = val;
+    // Unlike ac0.h and ac1.h, prod.h is not sign-extended
+    state.r.prod.h = val & 0x00ff;
     break;
   case DSP_REG_PRODM2:
     state.r.prod.m2 = val;
--- a/Source/Core/Core/DSP/Interpreter/DSPInterpreter.cpp
+++ b/Source/Core/Core/DSP/Interpreter/DSPInterpreter.cpp
@@ -394,13 +394,14 @@ s16 Interpreter::GetAXHigh(s32 reg) const
 s64 Interpreter::GetLongAcc(s32 reg) const
 {
   const auto& state = m_dsp_core.DSPState();
-  return static_cast<s64>(state.r.ac[reg].val << 24) >> 24;
+  return static_cast<s64>(state.r.ac[reg].val);
 }
 
 void Interpreter::SetLongAcc(s32 reg, s64 value)
 {
   auto& state = m_dsp_core.DSPState();
-  state.r.ac[reg].val = static_cast<u64>(value);
+  // 40-bit sign extension
+  state.r.ac[reg].val = static_cast<u64>((value << (64 - 40)) >> (64 - 40));
 }
 
 s16 Interpreter::GetAccLow(s32 reg) const
@@ -690,8 +691,8 @@ void Interpreter::OpWriteRegister(int reg_, u16 val)
   // 8-bit sign extended registers.
   case DSP_REG_ACH0:
   case DSP_REG_ACH1:
-    // sign extend from the bottom 8 bits.
-    state.r.ac[reg - DSP_REG_ACH0].h = (u16)(s16)(s8)(u8)val;
+    // Sign extend from the bottom 8 bits.
+    state.r.ac[reg - DSP_REG_ACH0].h = static_cast<s8>(val);
     break;
 
   // Stack registers.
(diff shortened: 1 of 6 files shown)


shuffle2:
kinda unrelated, but have you seen https://github.com/dolphin-emu/dolphin/blob/master/Source/Core/Core/DSP/Interpreter/DSPIntMultiplier.cpp#L104 /  know example that triggers the error the comment is talking about? Wanted to check it out before (but don't have hw to test on... :) )

Pokechu22:
I don't, but I'm guessing it's similar to the rounding that happens with e.g. `clrl` (which seems to be round-to-even), probably complicated by the fact that there are _two_ middle product registers (and Dolphin doesn't currently emulate the way the two product registers are written during multiplication; I don't think it's been documented anywhere and I haven't looked into how it works).

Source/Core/Core/DSP/Interpreter/DSPInterpreter.cpp
```diff
@@ -394,12 +394,16 @@ s16 Interpreter::GetAXHigh(s32 reg) const
...
 {
   const auto& state = m_dsp_core.DSPState();
-  return static_cast<s64>(state.r.ac[reg].val << 24) >> 24;
+  return static_cast<s64>(state.r.ac[reg].val);
 }
 
 void Interpreter::SetLongAcc(s32 reg, s64 value)
 {
   auto& state = m_dsp_core.DSPState();
+  if (value & 0x0080'0000'0000)
```
lioncash, line 403: Prefer making the comparison explicit

```suggestion
  if ((value & 0x0080'0000'0000) != 0)
```

Source/Core/Core/DSP/Interpreter/DSPInterpreter.cpp
```diff
@@ -394,12 +394,16 @@ s16 Interpreter::GetAXHigh(s32 reg) const
...
+    value |= 0xffff'ff00'0000'0000;
+  else
+    value &= 0x00ff'ffff'ffff;
```
lioncash, line 406:
```suggestion
    value &= 0x0000'00ff'ffff'ffff;
```

To be consistent length-wise with the branch above it

Source/Core/Core/DSP/Jit/x64/DSPJitRegCache.cpp
```diff
@@ -772,6 +772,56 @@ void DSPJitRegCache::PutReg(int reg, bool dirty)
...
+      {
+        // register is already shifted correctly
+        // (if at all)
+
+        // Zero extend from the bottom 8 bits.
+        m_emitter.MOVZX(16, 8, oparg.GetSimpleReg(), oparg);
+      }
+      else if (oparg.IsImm())
+      {
+        // TODO: Immediates?
```
lioncash, line 789: Perhaps there should be assertions here and in the case below, so if these are ever actually hit they don't silently do nothing

lioncash (commented):
Echoing what @Tilka said on IRC in that we really should be making the tests for the unit-testing system that runs for every PR instead of for one that needs to manually be run

Source/Core/Core/DSP/Jit/x64/DSPJitRegCache.cpp
Pokechu22: Done (also for the previously existing `DSP_REG_ACH0` case).

It's worth noting that `LRI $prod.h,  #0xffff` doesn't hit this case.

Source/Core/Core/DSP/Jit/x64/DSPJitRegCache.cpp
```diff
@@ -772,6 +773,58 @@ void DSPJitRegCache::PutReg(int reg, bool dirty)
...
+        ASSERT(false);
+      }
+      else
+      {
+        // Clear SR_100, which always reads back as 0
+        m_emitter.AND(32, m_regs[reg].loc, Gen::Imm32(~SR_100));
```
Tilka, line 824: Are you sure about the 32 here?

Pokechu22: The code failed to assemble when I used 16, but I think that was because I was using `Gen::Imm32` instead of `Gen::Imm16`.  `m_emitter.AND(16, m_regs[reg].loc, Gen::Imm16(~SR_100))` seems to work correctly.

PR #10034, opened 2021-08-16
JitArm64: Fix paired loadstore instruction decoding

JosJuice:
One of those fun "how did this ever work" bugs.


lioncash (approved):

JosJuice:
Okay, I figured out how this ever worked. The indexed variants are all mapped to `FallBackToInterpreter`... Should've been obvious in hindsight, but whatever.

So this PR didn't actually fix anything, but it did bring us one step closer to implementing the indexed variants.

PR #10035, opened 2021-08-17
JitArm64: dcbx loop detection for improved performance when invalidating large memory regions

JosJuice:
Follow-up to PR #10007.


AdmiralCurtiss:
In the commit message of the last commit you say `bits 5-10`, but I think it's only `bits 5-9` that are actually relevant here.

x64 changes look reasonable to me, can't say much about the ARM code...

JosJuice:
I meant 5 inclusive, 10 exclusive. I'm not sure if that's really clear from how I wrote it, though.

AdmiralCurtiss:
Yeah that's not clear, maybe write `[5-10[` or `[5-10)` then.

JosJuice:
Changed it to 5-9.

leoetlino (approved):

PR #10037, opened 2021-08-17
Bump minimum macOS to 10.13 High Sierra

OatmealDome:
As discussed [here](https://forums.dolphin-emu.org/Thread-unable-to-open-newest-version-of-dolphin-in-os-x-sierra?pid=520790#pid520790) and on IRC.

The minimum macOS version supported by Qt 5.14 is 10.13 High Sierra.

### DIFF
--- a/BuildMacOSUniversalBinary.py
+++ b/BuildMacOSUniversalBinary.py
@@ -64,7 +64,7 @@ DEFAULT_CONFIG = {
 
     # Minimum macOS version for each architecture slice
     "arm64_mac_os_deployment_target":  "11.0.0",
-    "x86_64_mac_os_deployment_target": "10.12.0",
+    "x86_64_mac_os_deployment_target": "10.13.0",
 
     # CMake Generator to use for building
     "generator": "Unix Makefiles",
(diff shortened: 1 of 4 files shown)


Dentomologist (commented):
A couple small additions:

There are references to 10.12 in a comment on lines 11 and 13 of [root]/CMakeLists.txt.

FileUtil.cpp has a __builtin_available check for 10.12 or later I believe we can remove.

OatmealDome:
OK, I think I got everything.

MayImilae:
Btw here is the stuff I found while researching this topic on the Qt side. I never found any post or anything explicitly declaring that macOS 10.12 was dropped, however, the documentation says it well enough.

Qt 5.13 documentation showing macaOS 10.12 as the minimum supported macOS version. - https://doc.qt.io/archives/qt-5.13/supported-platforms.html#macos 

Qt 5.14 documentation showing macOS 10.13 as the minimum supported macOS version - https://doc.qt.io/qt-5.14/supported-platforms.html#macos

Latest Qt documentation (currently 5.15) showing macOS 10.13 as the current minimum supported macOS version. - https://doc.qt.io/qt-5/supported-platforms.html#macos

lioncash (approved):

PR #10038, opened 2021-08-17
DSPLLE: Split SRS into SRS and SRSH and implement conditional variants of RTI

Pokechu22:
* The `SRS` instruction works differently from the `LRS` instruction.  `LRS` supports, in order, `$ax0.l`, `$ax1.l`, `$ax0.h`, `$ax1.h`, `$ac0.l`, `$ac1.l`, `$ac0.m`, and `$ac1.m`.  `SRS`, however, goes `$ac0.h`, `$ac1.h`, invalid, invalid, `$ac0.l`, `$ac1.l`, `$ac0.m`, `$ac1.m`.  As such I've split `SRS` into `SRSH` and `SRS`.  `SRS` with ax registers was used in the free rom, so I've replaced that.  It does seem that `SRSH` appears in the official rom, so this change may improve accuracy.

    The behavior of these instructions can be verified with the `srs_test.ds` program.  What was previously `SRS @0x00, $AX0.H` or `SRS @0x00, $AX1.H` does not actually make any stores, and `SRS @0x00, $AX0.L` or `SRS @0x00, $AX1.L` actually stores `$ac0.h`/`$ac1.h` (including the sign extension).
* The `RTI` instruction supports conditions.  I highly doubt anything actually uses this, because conditionally returning from an interrupt is rarely useful.  Testing this required rejiggering some of the DSPSpy base code to allow custom interrupt handlers to be used.

    The behavior of these instructions can be verified with the `rti_test.ds` program.  Dolphin previously ignored the `RTINZ` and `RTIZ` instructions and always had 0x3333 in `$ax1.l`.  Now, Dolphin correctly shows 0x1111 or 0x2222 depending on whether `$ax1.h` is 0 (i.e. it shows, in order, 0x2222, 0x1111, 0x2222), matching hardware.  I haven't actually tested the other conditional variants, and am assuming that they all match the same conditional variants used by other instructions.

Since the free DSP rom needed to be updated, I've also included Tilka/Dolphin@84a7cdaed509daa045da95895dd0905740a6827d which makes some other rom-related changes that were not considered important enough for a separate pull request.

Since the old free DSP roms no longer work correctly, I've changed the on-screen display warnings to a `AskYesNoFmtT`.  The message says "If you select "No", audio might be garbled.", which is not strictly accurate (the invalid `SRS` instructions were actually in ROM code used by the GBA uCode, and using the old ROM with this PR seems to only break GBA transfers and not audio in general).  However, it would only affect users who manually copied the free ROM into their user `GC` folder; users with the official DSP ROM or with no DSP ROM should not see this message.

I will update the gamecube DSP manual for these new instructions in a separate pull request.  I already have a large number of manual updates prepared, and would prefer to include this with them.

The `test_main` change causes this PR to conflict with #10032; although Git will not report a merge conflict, they should not be merged at the same time.  It's an easy fix (`reg_mask_test.ds` will just need `test_main:` added to it), but it can't be performed until one of the PRs is merged.


Source/Core/Core/DSP/DSPCore.cpp
```diff
@@ -75,22 +78,13 @@ static bool VerifyRoms(const SDSP& dsp)
...
+  if (rom_idx >= 1 && rom_idx <= 5)
   {
-    Host::OSD_AddMessage("You are using a free DSP ROM made by the Dolphin Team.", 8000);
-    Host::OSD_AddMessage("All Wii games will work correctly, and most GameCube games", 8000);
-    Host::OSD_AddMessage("should also work fine, but the CARD UCode will not work.", 8000);
+    if (AskYesNoFmtT("You are using an old free DSP ROM made by the Dolphin Team.\n"
+                     "Due to emulation inaccuracy improvements, this ROM longer works correctly.\n"
+                     "Would you like to stop now to fix the problem?\n"
+                     "If you select \"No\", audio might be garbled."))
+      return false;
```
BhaaLseN, line 87: Do we want to include some info on _what_ the user could do to fix the problem (for example, delete/replace the file?)

Also, since the `if` condition spans over multiple lines, our code style wants curlies here for readability.

Source/Core/Core/DSP/DSPTables.cpp
```diff
@@ -48,7 +48,22 @@ const std::array<DSPOPCTemplate, 214> s_opcodes =
...
+  {"RTINZ",    0x02f4, 0xffff,    1, 0, {},                                                                                     false, true, false, true, false}, // return from interrupt if not zero
+  {"RTIZ",     0x02f5, 0xffff,    1, 0, {},                                                                                     false, true, false, true, false}, // return from interrupt if zero
+  {"RTINC",    0x02f6, 0xffff,    1, 0, {},                                                                                     false, true, false, true, false}, // return from interrupt if not carry
+  {"RTIC",     0x02f7, 0xffff,    1, 0, {},                                                                                     false, true, false, true, false}, // return from interrupt if carry
+  {"RTIx8",    0x02f8, 0xffff,    1, 0, {},                                                                                     false, true, false, true, false}, // return from interrupt if TODO
+  {"RTIx9",    0x02f9, 0xffff,    1, 0, {},                                                                                     false, true, false, true, false}, // return from interrupt if TODO
```
BhaaLseN, line 60: Chances are those two follow the remaining conditional opcode convention and check for above/below `s32` range (basically bit 4 of the status register).

Plus the following two for a yet unknown pair of checks.

Source/Core/Core/DSP/Interpreter/DSPIntBranch.cpp
```diff
@@ -103,13 +103,17 @@ void Interpreter::ret(const UDSPInstruction opc)
...
 }
 
-// RTI
+// RTIcc
 // 0000 0010 1111 1111
 // Return from exception. Pops stored status register $sr from data stack
 // $st1 and program counter PC from call stack $st0 and sets $pc to this
 // location.
-void Interpreter::rti(const UDSPInstruction)
+// This instruction has a conditional form, but it is not used by any official ucode.
```
BhaaLseN, line 111: I wonder if this is worth a game quirk report? Probably not?

Source/Core/Core/DSP/DSPTables.cpp
Pokechu22: The "TODO" part for x8 and x9 is found with all of the instructions; I just copied that part from RET.  I can change it to "not over s32" and "over s32" respectively (for all of the conditional instructions), but I'll leave the xA and xB ones as `TODO`.

To be clear: The code currently assumes the same conditional behavior for RTIcc as all of the other conditional instructions (it uses `CheckCondition` for this); x8 through xB just don't have a label in DSPTables.

If you have a good summary for what this does:

https://github.com/dolphin-emu/dolphin/blob/a36855c9832847202f266e959584edd278363876/Source/Core/Core/DSP/Interpreter/DSPInterpreter.cpp#L256-L258

then I'd be happy to add a short description and a clearer mnemonic for all of them.

Source/Core/Core/DSP/Interpreter/DSPIntBranch.cpp
Pokechu22: I don't think it is.  There's a pretty well-defined set of uCodes, and it's unlikely that there's an undiscovered one that uses a conditional form of RTI.  (I didn't actually look through all of the uCodes, though, since I don't have all of them on hand.)

Source/Core/Core/DSP/DSPTables.cpp
BhaaLseN: Ah, tbf I didn't look at the others; just at the relation to the conditional instructions table in the docs.
In that case I'm all for consistency (and not just because I don't have any better name for them :P ) to leave them that way.

Pokechu22: I think I'll leave x8 and x9 alone for now too.  Based on 095f81e2f91844fe697f48d6bffd1f9476190f0e (and a quick check) they aren't used by any official uCode.  If further research indicates a clearer use for them, I'll create a separate PR.

Source/Core/Core/DSP/Interpreter/DSPIntBranch.cpp
Tilka: I have all but two ucodes and none of them use conditional rti.

Source/DSPSpy/tests/andc_ls_test.ds
```diff
@@ -14,7 +14,7 @@ include "dsp_base.inc"
 
 	lri $AC0.M, #0x0004
 	lri $AC1.M, #0x1234
-	andc'ls $AC1.M : $AX0.L, $AC0.M
+	cw 0x3d80 ; andc'ls $AC1.M : $AX0.L, $AC0.M
```
Tilka, line 17: What's going on here?

Pokechu22: `andc'ls` isn't a valid instruction, due to the way extension opcodes work when the first nybble is 3 (the main opcode is 9 bits, and the extension opcode is 7 bits).  That bit pattern actually encodes `lsrnr`.  I think this test is what originally found that 9-bit opcode behavior.

I ran into that when I tried to compile all of the tests to make sure they still worked after the `test_main` change.

Source/Core/Core/DSP/DSPCore.cpp
Pokechu22: Done.  I've also changed "emulation inaccuracy improvements" to "emulation accuracy improvements".

AdmiralCurtiss: "emulation inaccuracy improvements" sounds very strange, I'd drop the "in-" prefix.

BhaaLseN (approved):
Code seems good, but I haven't tested it.

Tilka:
What if:
- try to load ROM from file, warn if its hash doesn't match the original ROM
- otherwise assemble free ROM from embedded text

This way we don't need to update hashes all the time.

Pokechu22:
That could work, but I didn't actually find the hashes too hard to update.  I modified the hashes mismatch message to include the hashes and got them that way (and then I changed the message back).

I feel like it would be useful to have the hash comparison even if we were assembling the free ROM from text, as a sanity check to make sure the assembly worked well.  So maybe it'd be better to just always include the hashes in the mismatch message?

Tilka:
I just feel silly maintaining a changelog and a list of file hashes. That's what Git is for.

Source/Core/Core/DSP/DSPCore.cpp
```diff
@@ -69,28 +72,29 @@ static bool VerifyRoms(const SDSP& dsp)
...
-    Host::OSD_AddMessage("should also work fine, but the GBA/CARD UCodes will not work.", 8000);
-  }
-  else if (rom_idx == 4)
+  if (rom_idx >= 1 && rom_idx <= 5)
   {
-    Host::OSD_AddMessage("You are using a free DSP ROM made by the Dolphin Team.", 8000);
-    Host::OSD_AddMessage("All Wii games will work correctly, and most GameCube games", 8000);
-    Host::OSD_AddMessage("should also work fine, but the CARD UCode will not work.", 8000);
+    if (AskYesNoFmtT("You are using an old free DSP ROM made by the Dolphin Team.\n"
+                     "Due to emulation accuracy improvements, this ROM longer works correctly.\n\n"
```
Rumi-Larry, line 89:
```suggestion
                     "Due to emulation accuracy improvements, this ROM no longer works correctly.\n\n"
```

Rumi-Larry:
> 
> 
> I just feel silly maintaining a changelog and a list of file hashes. That's what Git is for.

IMO, given the small size of it. keeping this documentation here is fine. Otherwise there would need to be another repository just for the replacement ROMS, which would have been a good idea when they were being created, but now creating a new repository for them seems silly too.

PR #10040, opened 2021-08-18
Interpreter: Simplify mcrfs implementation

JosJuice:


Source/Core/Core/PowerPC/Interpreter/Interpreter_SystemRegisters.cpp
```diff
@@ -549,34 +549,12 @@ void Interpreter::isync(UGeckoInstruction inst)
...
-    FPSCR.VXVC = 0;
-    break;
-  case 5:
-    FPSCR.VXSOFT = 0;
-    FPSCR.VXSQRT = 0;
-    FPSCR.VXCVI = 0;
-    break;
-  }
+  u32 shift = 4 * (7 - inst.CRFS);
+  u32 fpflags = (FPSCR.Hex >> shift) & 0xF;
```
BhaaLseN, line 553: Can we (or rather: do we want to) mark those as `const` since we're not changing them afterwards?
```suggestion
  const u32 shift = 4 * (7 - inst.CRFS);
  const u32 fpflags = (FPSCR.Hex >> shift) & 0xF;
```

BhaaLseN (commented):
I wonder if that implementation is really considered _simpler_ than before. It is _shorter_ though. 😅

Source/Core/Core/PowerPC/Interpreter/Interpreter_SystemRegisters.cpp
```diff
@@ -549,34 +549,12 @@ void Interpreter::isync(UGeckoInstruction inst)
...
-    FPSCR.VXSOFT = 0;
-    FPSCR.VXSQRT = 0;
-    FPSCR.VXCVI = 0;
-    break;
-  }
+  const u32 shift = 4 * (7 - inst.CRFS);
+  const u32 fpflags = (FPSCR.Hex >> shift) & 0xF;
+
+  // If any exception bits were read, clear them
+  FPSCR.Hex &= ~((0xF << shift) & (FPSCR_FX | FPSCR_ANY_X));
```
leoetlino, line 556: Could you explain why the mask is AND'd with `FPSCR_FX | FPSCR_ANY_X`?

JosJuice: If we read a bit that represents whether a certain exception has happened, we need to clear that bit. But if we read some other bit, like the rounding mode, FPSCR, or whether a certain exception should be enabled, that bit shouldn't be cleared.

leoetlino (approved):

PR #10041, opened 2021-08-18
Interpreter: Make signedness and narrowing conversions explicit

lioncash:
Makes our conversions between different signs and cases where truncations would occur to be explicit to indicate that they're intentional and, well, make implicit conversions explicit for reading purposes, which I think is preferable for the interpreter, given it essentially partially doubles as a reading reference for faster CPU backends to base behavior off of in tandem with other documenation. This also resolves numerous compiler warnings when stricter conversion warnings are enabled.

Note that we have a bunch of seemingly... old history tracing code in Interpreter.cpp, if this isn't particularly useful, I'm fine with completely removing it, since it'd make the interpreter's Run() member function a little more compact and nicer to read without the ifdefs.

### DIFF
--- a/Source/Core/Core/PowerPC/Interpreter/Interpreter_Branch.cpp
+++ b/Source/Core/Core/PowerPC/Interpreter/Interpreter_Branch.cpp
@@ -7,7 +7,6 @@
 #include "Core/HLE/HLE.h"
 #include "Core/PowerPC/Interpreter/ExceptionUtils.h"
 #include "Core/PowerPC/Interpreter/Interpreter.h"
-#include "Core/PowerPC/MMU.h"
 #include "Core/PowerPC/PowerPC.h"
 
 void Interpreter::bx(UGeckoInstruction inst)
@@ -15,10 +14,12 @@ void Interpreter::bx(UGeckoInstruction inst)
   if (inst.LK)
     LR = PC + 4;
 
+  const auto address = u32(SignExt26(inst.LI << 2));
+
   if (inst.AA)
-    NPC = SignExt26(inst.LI << 2);
+    NPC = address;
   else
-    NPC = PC + SignExt26(inst.LI << 2);
+    NPC = PC + address;
 
   m_end_block = true;
 }
@@ -29,11 +30,11 @@ void Interpreter::bcx(UGeckoInstruction inst)
   if ((inst.BO & BO_DONT_DECREMENT_FLAG) == 0)
     CTR--;
 
-  const bool true_false = ((inst.BO >> 3) & 1);
-  const bool only_counter_check = ((inst.BO >> 4) & 1);
-  const bool only_condition_check = ((inst.BO >> 2) & 1);
+  const bool true_false = ((inst.BO >> 3) & 1) != 0;
+  const bool only_counter_check = ((inst.BO >> 4) & 1) != 0;
+  const bool only_condition_check = ((inst.BO >> 2) & 1) != 0;
   const u32 ctr_check = ((CTR != 0) ^ (inst.BO >> 1)) & 1;
-  const bool counter = only_condition_check || ctr_check;
+  const bool counter = only_condition_check || ctr_check != 0;
   const bool condition =
       only_counter_check || (PowerPC::ppcState.cr.GetBit(inst.BI) == u32(true_false));
 
@@ -42,10 +43,12 @@ void Interpreter::bcx(UGeckoInstruction inst)
     if (inst.LK)
       LR = PC + 4;
 
+    const auto address = u32(SignExt16(s16(inst.BD << 2)));
+
     if (inst.AA)
-      NPC = SignExt16(inst.BD << 2);
+      NPC = address;
     else
-      NPC = PC + SignExt16(inst.BD << 2);
+      NPC = PC + address;
   }
 
   m_end_block = true;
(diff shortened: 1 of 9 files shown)


Source/Core/Core/PowerPC/Interpreter/Interpreter_Branch.cpp
```diff
@@ -7,18 +7,19 @@
...
 #include "Core/PowerPC/Interpreter/Interpreter.h"
-#include "Core/PowerPC/MMU.h"
 #include "Core/PowerPC/PowerPC.h"
 
 void Interpreter::bx(UGeckoInstruction inst)
 {
   if (inst.LK)
     LR = PC + 4;
 
+  const auto address = u32(SignExt26(s16(inst.LI << 2)));
```
JosJuice, line 17: `s16` here seems wrong.

Source/Core/Core/PowerPC/Interpreter/Interpreter_FloatingPoint.cpp
```diff
@@ -22,9 +22,9 @@ enum class RoundingMode
   TowardsNegativeInfinity = 0b11
 };
 
-static void SetFI(UReg_FPSCR* fpscr, int FI)
+void SetFI(UReg_FPSCR* fpscr, u32 FI)
```
JosJuice, line 25: Shouldn't this be static? I can't find it in any header file.

Source/Core/Core/PowerPC/Interpreter/Interpreter_SystemRegisters.cpp
```diff
@@ -487,50 +487,66 @@ void Interpreter::mtspr(UGeckoInstruction inst)
 
 void Interpreter::crand(UGeckoInstruction inst)
 {
-  PowerPC::ppcState.cr.SetBit(inst.CRBD, PowerPC::ppcState.cr.GetBit(inst.CRBA) &
-                                             PowerPC::ppcState.cr.GetBit(inst.CRBB));
+  const u32 a = PowerPC::ppcState.cr.GetBit(inst.CRBA);
+  const u32 b = PowerPC::ppcState.cr.GetBit(inst.CRBB);
+
+  PowerPC::ppcState.cr.SetBit(inst.CRBD, a & b);
```
JosJuice, line 493: Is there a reason why `GetBit` and `SetBit` are using `u32` instead of `bool`? It may be cleaner to change the definitions of those functions than to start using `u32` here.

phire:
> Note that we have a bunch of seemingly... old history tracing code in Interpreter.cpp, if this isn't particularly useful, I'm fine with completely removing it

It might theoretically be useful, but only if someone decides to make major changes to dolphin's JITs in the future. Not sure if it's worth keeping around just for that eventuality.

Though, any removal should be done in another PR after merging this one.

Source/Core/Core/PowerPC/Interpreter/Interpreter_FloatingPoint.cpp
lioncash: It's within an anonymous namespace, so it already has internal linkage. ConvertToInteger below doesn't have `static` either.

Source/Core/Core/PowerPC/Interpreter/Interpreter_Branch.cpp
lioncash: Good catch!

Source/Core/Core/PowerPC/Interpreter/Interpreter_SystemRegisters.cpp
lioncash: Using bool here can result in false-positives (e.g. `bool & bool` and `bool | bool` are moderately common logic bugs, so compilers and analysis tools will usually warn about these or have facilities that find them).

FWIW, these functions were fine as they were conversion-wise, I just hoisted the accessors out of the `SetBit` call to make them a little nicer to read, given the accessors themselves are quite long.

JosJuice: Alright, let's keep this as it is in this PR then.

Source/Core/Core/PowerPC/Interpreter/Interpreter_Integer.cpp
```diff
@@ -10,10 +10,10 @@
 
 void Interpreter::Helper_UpdateCR0(u32 value)
 {
-  s64 sign_extended = (s64)(s32)value;
-  u64 cr_val = (u64)sign_extended;
-  cr_val = (cr_val & ~(1ull << PowerPC::CR_EMU_SO_BIT)) |
-           ((u64)PowerPC::GetXER_SO() << PowerPC::CR_EMU_SO_BIT);
+  const s64 sign_extended = s64{s32(value)};
```
BhaaLseN, line 13: Just wondering, since I see a mix of classic constructors and uniform initialization (going both ways and non-mixed) across this PR, is there any significant difference (or specific reason) why you sometimes use `s64{s32(value)}` (uniform initialization outside, ctor inside), `u32(s32{inst.SIMM_16})` (the opposite; later in this file in `addic`) or just `u32(s32(rGPR[inst.RA]) * inst.SIMM_16)` (only ctor; later in this file in `mulli`)?

At first glance they seem like they should be interchangable in all locations and we could use one syntax for consistency; but I might be missing something...?

lioncash: I use the brace-form where the conversion is not narrowing (expanding SIMM_16 from its bit-field type to `int` doesn't have a loss of information, since it's already a signed type, so it's permissible to use that form of initializing).

On the other hand, if I were to do `u32{SIMM_16}`, this would cause a compiler warning because it's not a lossless conversion (its a sign-conversion), which will cause a compilation warning (and since we build with warnings as errors, it'll cause a compilation error).

It's just a way to allow the compiler to enforce caring about scenarios where conversions aren't lossless (which, in most cases, we actually want here)

BhaaLseN: Got it, thanks 👍

leoetlino (approved):
> Reviewed 9 of 9 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/dolphin-emu/dolphin/10041)* status: all files reviewed, 1 unresolved discussion (waiting on @lioncash)

PR #5744, opened 2017-07-03
Remove non-integer IRs

JosJuice:
Non-integer internal resolutions lead to many problems that we can't do much about – mostly graphical, but in Paper Mario TTYD's case, there are problems like [crashes](https://forums.dolphin-emu.org/Thread-paper-mario-thousand-year-door-crashing--49298) and even [save corruption](https://bugs.dolphin-emu.org/issues/10171). This PR removes non-integer internal resolutions from Dolphin entirely.

One note about the implementation of this PR: I've changed how Dolphin stores the EFB scale setting in RAM because it simplifies code significantly in several places. Now 0 means auto (integral), 1 means 1x, 2 means 2x, and so on. Do we want to apply this change to the INIs as well (so that values make more sense) or keep using the old numbering system in INIs (to be backwards compatible)? For now, I've just made it use the new numbering system with a new setting name, because I wasn't sure where in the new config code I could put code for translating between the two numbering systems.

PR #10042, opened 2021-08-18
Support for custom resolution (1.5x, 1.65x, 1.75x...)

Gamer64ytb:
This function causes values ​​of 100, 110, 115 ... to be used instead of 1, 2, 3, 4 ..., in this way you can use resolutions 1.5, 1.65, 1.75 and more. So that it does not conflict with the option, go to the arrays.xml and replace the resolution option with this:

<string-array name="internalResolutionEntries" translatable="false">
<item>1x Native (640x528)</item>
<item>1.5x (640x720) BUG</item>
<item>2x Native (1280x1056) for 720p</item>
<item>3x Native (1920x1584) for 1080p</item>
<item>4x Native (2560x2112)</item>
<item>5x Native (3200x2640)</item>
<item>6x Native (3840x3168) for 4K</item>
</string-array>
<integer-array name="internalResolutionValues" translatable="false">
<item>100</item>
<item>150</item>
<item>200</item>
<item>300</item>
<item>400</item>
<item>500</item>
<item>600</item>
</integer-array>

### DIFF
--- a/Source/Core/VideoCommon/RenderBase.cpp
+++ b/Source/Core/VideoCommon/RenderBase.cpp
@@ -314,17 +314,17 @@ void Renderer::RenderToXFB(u32 xfbAddr, const MathUtil::Rectangle<int>& sourceRc
 
 unsigned int Renderer::GetEFBScale() const
 {
-  return m_efb_scale;
+  return m_efb_scale / 100.0f;
 }
 
 int Renderer::EFBToScaledX(int x) const
 {
-  return x * static_cast<int>(m_efb_scale);
+  return x * m_efb_scale / 100.0f;
 }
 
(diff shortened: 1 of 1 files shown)


JosJuice:
No. There was extensive discussion back when this was removed (see PR #5744), and I don't see any clear argument for why it should be brought back now.

Second, if we are to do this, we should not do it by dividing the `InternalResolution` value by 100. That makes no sense, and it also kills the ability to carry a settings file across builds.

JosJuice:
For future reference, fixing arrays.xml and the corresponding DolphinQt code is something the author of the pull request would be expected to do. A pull request that just changes the core code and leaves the GUI broken will not get merged.

PR #10043, opened 2021-08-18
Raise program exception on floating point exceptions

JosJuice:
Untested!


JMC47:
I guess I should note that this doesn't work yet.

Source/Core/Core/PowerPC/Jit64/Jit_SystemRegisters.cpp
```diff
@@ -185,6 +187,30 @@ FixupBranch Jit64::JumpIfCRFieldBit(int field, int bit, bool jump_if_set)
...
+  OR(32, R(fpscr), R(tmp1));
+
+  // fpscr.FEX = ((fpscr >> 22) & (fpscr & FPSCR_ANY_E)) != 0
+  MOV(32, R(tmp1), R(fpscr));
+  MOV(32, R(tmp2), R(fpscr));
+  SHR(32, R(tmp1), Imm8(22));
+  AND(32, R(tmp2), Imm32(FPSCR_ANY_E));
+  TEST(32, R(tmp1), R(tmp2));
+  XOR(32, R(tmp1), R(tmp1));
+  SETcc(CC_NZ, R(tmp1));
```
Sintendo, line 209: This isn't right. XOR also updates the zero flag, so the result from TEST is lost.

Sintendo (changes requested):

Source/Core/Core/PowerPC/Jit64/Jit_SystemRegisters.cpp
JosJuice: Thanks, I wouldn't have considered that, with how used I am to AArch64 :)

I was going to put the XOR before the TEST, but I couldn't clobber any of the three registers until after the TEST. Do you know of a solution that wouldn't require an extra register?

JosJuice: Actually, you know what, I could probably get rid of the XOR. We will be shifting tmp1 left by 30 bits after the SETcc, so what the upper three bytes are set to doesn't matter.

Sintendo: I was going to suggest `MOVZX(32, 8, R(tmp1), R(tmp1));`, but your idea is better! 👍

Source/Core/Core/PowerPC/PPCTables.h
```diff
@@ -64,7 +64,9 @@ enum InstructionFlags : u64
   FL_IN_FLOAT_C_BITEXACT = (1ull << 31),  // The output is based on the exact bits in frC.
   FL_IN_FLOAT_AB_BITEXACT = FL_IN_FLOAT_A_BITEXACT | FL_IN_FLOAT_B_BITEXACT,
   FL_IN_FLOAT_BC_BITEXACT = FL_IN_FLOAT_B_BITEXACT | FL_IN_FLOAT_C_BITEXACT,
-  FL_PROGRAMEXCEPTION = (1ull << 32),  // May generate a system exception.
+  FL_PROGRAMEXCEPTION = (1ull << 32),  // May generate a program exception (not floating point).
+  FL_FLOAT_EXCEPTION = (1ull << 33),   // May generate a program exception (floating point).
+  FL_FLOAT_DIV = (1ull << 33),  // May generate a program exception (FP) due to division by 0.
```
BhaaLseN, line 69: It seems a bit weird that both are `1 << 33`, is this correct?

JMC47:
While it was painfully slow, enabling the floating point exceptions for Call of Duty: Finest Hour also fixed the crash there, but playing through that game with it on might be worse than it crashing.  Actually, playing the game at all is worse than it crashing.

JosJuice:
JMC has confirmed on IRC that both of the affected games are fixed by just handling divide by zero exceptions (like the game INI commit enables).

JMC47 (approved):
I've completed both known games that use this feature.

PR #4123, opened 2016-08-14
Jit: FIFO optimization improvements

hthh:
This introduces speculative constants, allowing FIFO writes to be
optimized in more places.

It also clarifies the guarantees of the FIFO optimization, changing
the location of some of the checks and potentially avoiding redundant
checks.

I wrote this a while ago, but this should finally fix the quadratic trampoline generation problem (which I originally tried to fix in #1218 and almost got fixed in #1230). This saves a fair amount of JIT and trampoline cache memory.

---
This change is [Reviewable](https://reviewable.io/reviews/dolphin-emu/dolphin/4123)

PR #10046, opened 2021-08-20
JitArm64: FIFO optimization improvements

JosJuice:
JitArm64 port of PR #4123.

### DIFF
--- a/Source/Core/Core/PowerPC/JitArm64/Jit.cpp
+++ b/Source/Core/Core/PowerPC/JitArm64/Jit.cpp
@@ -344,8 +344,9 @@ void JitArm64::IntializeSpeculativeConstants()
         STR(IndexType::Unsigned, DISPATCHER_PC, PPC_REG, PPCSTATE_OFF(pc));
         MOVP2R(ARM64Reg::X8, &JitInterface::CompileExceptionCheck);
         MOVI2R(ARM64Reg::W0, static_cast<u32>(JitInterface::ExceptionType::SpeculativeConstants));
-        BLR(ARM64Reg::X8);
-        B(dispatcher_no_check);
+        // Write dispatcher_no_check to LR for tail call
+        MOVP2R(ARM64Reg::X30, dispatcher_no_check);
+        BR(ARM64Reg::X8);
         SwitchToNearCode();
       }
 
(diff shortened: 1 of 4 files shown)


theofficialgman:
I pulled this PR and rebased on master upstream
Doesn't seem to lead to any measurable performance improvements in game when tested on TX1 (already cpu thread limited), but it also doesn't bring any regressions so thats good

JosJuice:
@dolphin-emu-bot rebuild

dvessel:
Hope this isn’t annoying. Going through old PR’s with these numbers. Figured out a jank way to measure frame times while I do other things. 

No real changes here. Apple M1, f-zero intro and a few seconds of game play.

| builds | samples | min  | max | median | stddev |
| ---: | --- | --- | --- | --- | --- |
| master first run | 14361.000 | 1.258 | 906.872 | 7.038 | 9.610 |
| master second run | crashed | ??? | ??? | ??? | ??? |
| master third run | 14361.000 | 1.203 | 878.735 | 6.914 | 9.367 |
| master fourth run | 14361.000 | 1.237 | 891.097 | 6.989 | 9.512 |
| 10046 first run | 14361.000 | 1.213 | 908.541 | 6.998 | 9.603 |
| 10046 second run | 14361.000 | 1.020 | 886.503 | 6.840 | 9.448 |
| 10046 third run | 14361.000 | 1.121 | 917.706 | 6.852 | 9.653 |
| 10046 fourth run | 14361.000 | 1.177 | 891.943 | 6.822 | 9.440 |

Source/Core/Core/PowerPC/JitArm64/Jit.cpp
```diff
@@ -806,6 +850,12 @@ bool JitArm64::DoJit(u32 em_address, JitBlock* b, u32 nextPC)
   gpr.Start(js.gpa);
   fpr.Start(js.fpa);
 
+  if (js.noSpeculativeConstantsAddresses.find(js.blockStart) ==
+      js.noSpeculativeConstantsAddresses.end())
```
Pokechu22, line 854: C++20 has [`unordered_set::contains`](https://en.cppreference.com/w/cpp/container/unordered_set/contains) which might be easier to read (though it probably has no performance difference). (The same applies to the x64 JIT.)

JosJuice: I think I'd prefer to do it like this for now and then move them over all at once. (This isn't the first place in JitArm64 where we're using this idiom.)

Pokechu22: That's fine. (I also noticed that the x64 JIT uses camelCase in its `IntializeSpeculativeConstants`, which is something that can be fixed elsewhere).

Source/Core/Core/PowerPC/JitArm64/Jit.cpp
```diff
@@ -344,8 +344,8 @@ void JitArm64::IntializeSpeculativeConstants()
         STR(IndexType::Unsigned, DISPATCHER_PC, PPC_REG, PPCSTATE_OFF(pc));
         MOVP2R(ARM64Reg::X8, &JitInterface::CompileExceptionCheck);
         MOVI2R(ARM64Reg::W0, static_cast<u32>(JitInterface::ExceptionType::SpeculativeConstants));
-        BLR(ARM64Reg::X8);
-        B(dispatcher_no_check);
+        MOVP2R(ARM64Reg::X30, dispatcher_no_check);
+        BR(ARM64Reg::X8);
```
Pokechu22, line 348: I'd find a comment saying something like "X30 is LR; perform a tail call to dispatcher_no_check after the exception check" useful here (and in the other change in this commit).

Pokechu22 (approved):
Untested, but looks good to me as a port of #4123.

PR #10025, opened 2021-08-14
DolphinQt: Minor consistency fixes for the Settings window

mbc07:
During the (ongoing) review of the Brazilian Portuguese translation, I noticed various minor consistency issues with the main Settings window, and this PR fixes all of them. Here's a breakdown of what changed (with screenshots):

### General tab
* Fixed Fallback region dropdown list misalignment in comparison to the other dropdown lists from this tab

  <details>
  <summary>Before</summary>

  GeneralBefore
  </details>
  <details>
  <summary>After</summary>

  GeneralAfter
  </details>

### Audio tab
* Fixed inconsistent usage of the ms and percent units (e.g. **20ms** and **100 %** instead of **20 ms** and **100%**)
* Removed ms abbreviation from the latency tooltip, to make it consistent with the buffer size tooltip where it's not abbreviated

  <details>
  <summary>Before</summary>

  AudioBefore
  </details>
  <details>
  <summary>After</summary>

  AudioAfter
  </details>

### GameCube tab
* Made the dropdown lists use all available horizontal space, making them consistent with the dropdown lists of other Dolphin tabs/windows using similar single column layouts
* Reworded the Skip Main Menu tooltip from "Put Main Menu roms in User/GC/{region}." to "Put IPL ROMs in User/GC/\<region>." (not shown in the screenshot)

  <details>
  <summary>Before</summary>

  GCBefore
  </details>
  <details>
  <summary>After</summary>

  GCAfter
  </details>

### Broadband Adapter settings dialog
The bulk of this PR is there. Initial goal was to just hide the useless help button from the window title and to make the XLink Kai URL clickable. This dialog was previously implemented with a simple `QInputDialog::getText` function call, but it couldn't achieve the clickable URL with that since the dialog generated by the function is very simple and not much customizable, so I had to implement a new one from scratch. 

Since I had already got my hands dirty at this point, I also implemented input validation for the MAC address (didn't add any for the XLink Kai address since you apparently can use URLs instead of just IP addresses -- didn't test that, though). Also, both BBA types now has a clickable URL, XLink Kai adapters point to [the URL that was already there](https://www.teamxlink.co.uk/wiki/Dolphin) on master, while TAP-based adapters now point to the relevant [Dolphin Wiki page](https://wiki.dolphin-emu.org/index.php?title=Broadband_Adapter). Oh, the window titles were also reworded as the previous ones were too big to be fully displayed.

  <details>
  <summary>Before (TAP Adapter)</summary>

  BBABefore
  </details>
[...]

PR #10047, opened 2021-08-21
DolphinQt: Remove Windows dialog help buttons

Dentomologist:
Removes the Windows help button from a handful of dialogs. I left the XLinkKai and Ethernet dialogs alone because they're taken care of in https://github.com/dolphin-emu/dolphin/pull/10025.

Also throws a few consts around.

Fixes https://bugs.dolphin-emu.org/issues/11536.

As noted in that report there's an option to globally disable the help button, but that would require us to bump the minimum Qt version from 5.9 to 5.10. Once we do bump the version we'll want to go through and remove all the hints.

### DIFF
--- a/Source/Core/DolphinQt/Debugger/WatchWidget.cpp
+++ b/Source/Core/DolphinQt/Debugger/WatchWidget.cpp
@@ -250,9 +250,11 @@ void WatchWidget::OnClear()
 
 void WatchWidget::OnNewWatch()
 {
-  QString text = QInputDialog::getText(this, tr("Input"), tr("Enter address to watch:"));
+  const QString text =
+      QInputDialog::getText(this, tr("Input"), tr("Enter address to watch:"), QLineEdit::Normal,
+                            QString{}, nullptr, Qt::WindowCloseButtonHint);
   bool good;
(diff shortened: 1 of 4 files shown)


Source/Core/DolphinQt/Debugger/WatchWidget.cpp
```diff
@@ -250,9 +250,11 @@ void WatchWidget::OnClear()
 
 void WatchWidget::OnNewWatch()
 {
-  QString text = QInputDialog::getText(this, tr("Input"), tr("Enter address to watch:"));
+  const QString text =
+      QInputDialog::getText(this, tr("Input"), tr("Enter address to watch:"), QLineEdit::Normal,
+                            QString{}, nullptr, Qt::WindowCloseButtonHint);
```
Dentomologist, line 255: The first three new parameters are just default arguments.

Qt::WindowCloseButtonHint overrides the Qt::WindowFlags parameter with the only bit we want set, clearing the offending Qt::WindowContextHelpButtonHint.

iwubcode (approved):

Techjar:
I think we have a subclass for this.

lioncash (approved):

PR #10048, opened 2021-08-21
Improvements to DSP manual

Pokechu22:
A bunch of improvements to the DSP manual, bringing it in line with Dolphin's own code (and changes from #10032, #10038, and #10045) and making it easier to use.

For convenience, here is the newly compiled manual: [GameCube_DSP_Users_Manual.pdf](https://github.com/dolphin-emu/dolphin/files/7026050/GameCube_DSP_Users_Manual.pdf)

### DIFF
--- a/docs/DSP/GameCube_DSP_Users_Manual/GameCube_DSP_Users_Manual.tex
+++ b/docs/DSP/GameCube_DSP_Users_Manual/GameCube_DSP_Users_Manual.tex
@@ -1837,8 +1837,7 @@ There are two pairs of conditions that work similarly: \texttt{EQ}/\texttt{NE} a
   \end{DSPOpcodeDescription}
 
   \begin{DSPOpcodeOperation}
-    res = ($acD.hm - I) | $acD.l
-    FLAGS(res)
+    FLAGS($acD - (I << 16))
     $pc += 2
   \end{DSPOpcodeOperation}
 \end{DSPOpcode}
@@ -1858,7 +1857,7 @@ There are two pairs of conditions that work similarly: \texttt{EQ}/\texttt{NE} a
   \end{DSPOpcodeDescription}
 
   \begin{DSPOpcodeOperation}
-    FLAGS($acD - #I)
+    FLAGS($acD - (I << 16))
     $pc++
   \end{DSPOpcodeOperation}
 \end{DSPOpcode}
--- a/docs/DSP/GameCube_DSP_Users_Manual/GameCube_DSP_Users_Manual.tex
+++ b/docs/DSP/GameCube_DSP_Users_Manual/GameCube_DSP_Users_Manual.tex
@@ -616,17 +616,12 @@ Hardware registers occupy the address space at \Address{0xFFxx} in DSP memory sp
 \begin{tabular}{|l|l|l|}
 \hline
 \textbf{Address} & \textbf{Name}    & \textbf{Description}          \\ \hline
-\multicolumn{3}{|l|}{\textit{Mailboxes}}                            \\ \hline
-\Address{0xFFFE} & \Register{CMBH}  & CPU Mailbox H                 \\ \hline
-\Address{0xFFFF} & \Register{CMBL}  & CPU Mailbox L                 \\ \hline
-\Address{0xFFFC} & \Register{DMBH}  & DSP Mailbox H                 \\ \hline
-\Address{0xFFFD} & \Register{DMBL}  & DSP Mailbox L                 \\ \hline
 \multicolumn{3}{|l|}{\textit{DMA Interface}}                        \\ \hline
-\Address{0xFFCE} & \Register{DSMAH} & Memory address H              \\ \hline
-\Address{0xFFCF} & \Register{DSMAL} & Memory address L              \\ \hline
-\Address{0xFFCD} & \Register{DSPA}  & DSP memory address            \\ \hline
 \Address{0xFFC9} & \Register{DSCR}  & DMA control                   \\ \hline
 \Address{0xFFCB} & \Register{DSBL}  & Block size                    \\ \hline
+\Address{0xFFCD} & \Register{DSPA}  & DSP memory address            \\ \hline
+\Address{0xFFCE} & \Register{DSMAH} & Memory address H              \\ \hline
+\Address{0xFFCF} & \Register{DSMAL} & Memory address L              \\ \hline
 \multicolumn{3}{|l|}{\textit{Accelerator}}                          \\ \hline
 \Address{0xFFD4} & \Register{ACSAH} & Accelerator start address H   \\ \hline
 \Address{0xFFD5} & \Register{ACSAL} & Accelerator start address L   \\ \hline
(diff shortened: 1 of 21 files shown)


Pokechu22:
This PR has almost all of the documentation I want in it.  The only things that it's missing are some hardware testing of the initialization process (I documented it in this PR, but only based on what libogc does and not how hardware reacts; I don't currently plan on investigating it further though) and an opcode map since it would help with visualizing what instructions are valid.  I [did already make an opcode map](https://imgur.com/a/fciCtWA), but it turns out that typesetting a table like that is a nightmare so I didn't do it.  [Here's the spreadsheet I used to make the map](https://github.com/dolphin-emu/dolphin/files/7026061/DSP_Opcodes.ods).

docs/DSP/GameCube_DSP_Users_Manual/GameCube_DSP_Users_Manual.tex
```diff
@@ -1030,7 +1030,7 @@ \section{Bit meanings}
 \section{Conditional opcodes}
 
 Conditional opcodes are executed only when the condition described by their encoded conditional field has been met.
-The groups of conditional instructions are: \Opcode{CALL}, \Opcode{JMP}, \Opcode{IFcc}, \Opcode{RET}, \Opcode{JMPR}, and \Opcode{CALLR}.
+The groups of conditional instructions are: \Opcode{CALL}, \Opcode{JMP}, \Opcode{IFcc}, \Opcode{RET}, \Opcode{RTI}, \Opcode{JMPR}, and \Opcode{CALLR}.
```
Tilka, line 1049: Link the conditional versions instead?

docs/DSP/GameCube_DSP_Users_Manual/GameCube_DSP_Users_Manual.tex
```diff
@@ -198,6 +198,23 @@
...
+    \hline
+    OS & LZ & TB & S32 & S  & AZ & O  & C \\
+    \hline
+    #1 & #2 & #3 & #4  & #5 & #6 & #7 & #8 \\
+    \hline
+  \end{tabular}
+  \end{description}
+}
+
+\newcommand{\DSPOpcodeFlagsUnchanged}[0]{\DSPOpcodeFlags{-}{-}{-}{-}{-}{-}{-}{-}}
```
BhaaLseN, line 216: I think we should put an explaination somewhere (perhaps in the Opcode syntax chapter, or maybe next to Conditional opcodes?) on what the used symbols (which, as far as I could see, were: "-", "X" and "0").

I'm _assuming_ "X" is "affected by the operation and set accordingly", "0" is "cleared/set to zero" and "-" is "unaffected"; but thats even knowing the DSP.

docs/DSP/GameCube_DSP_Users_Manual/GameCube_DSP_Users_Manual.tex
```diff
@@ -3603,144 +4905,199 @@ \section{Alphabetical list of extended opcodes}
...
-            &          &    &                                         \\
-LSL16       &          & *  & 1111 000r xxxx xxxx                     \\
-MADD        &          & *  & 1111 001s xxxx xxxx                     \\
-LSR16       &          & *  & 1111 010r xxxx xxxx                     \\
-MSUB        &          & *  & 1111 011s xxxx xxxx                     \\
-ADDPAXZ     &          & *  & 1111 10ar xxxx xxxx                     \\
-CLRL        &          & *  & 1111 110r xxxx xxxx                     \\
-MOVPZ       &          & *  & 1111 111r xxxx xxxx
+\begin{longtable}{llr}
+Instruction & Opcode & Page \\ \hline
```
BhaaLseN, line 4914: Can we make this table header repeat on a page break?

docs/DSP/GameCube_DSP_Users_Manual/GameCube_DSP_Users_Manual.tex
```diff
@@ -586,87 +650,223 @@ \chapter{Hardware interface}
...
-\Address{0xFFD8} & \Register{ACCAH} & Accelerator current address H \\ \hline
-\Address{0xFFD9} & \Register{ACCAL} & Accelerator current address L \\ \hline
-\Address{0xFFDD} & \Register{ACDAT} & Accelerator data              \\ \hline
-\multicolumn{3}{|l|}{\textit{Interrupts}}                           \\ \hline
-\Address{0xFFFB} & \Register{DIRQ}  & IRQ request                   \\ \hline
+\textbf{Address} & \textbf{Name}      & \textbf{Description}          \\ \hline
+\multicolumn{3}{|l|}{\textit{ADPCM Coefficients}}                     \\ \hline
+\Address{0xFFA0} & \Register{COEF\_A1\_0} & A1 Coefficient \# 0 \\ \hline
+\Address{0xFFA1} & \Register{COEF\_A2\_0} & A2 Coefficient \# 0 \\ \hline
+& & \\ \hline
```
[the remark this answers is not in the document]
Pokechu22: I tried using `\vdots` but it looked somewhat weird due to them being taller (and left-aligned):

image

I just now tried `\multicolumn{3}{|c|}{$\vdots$}` and that looks better (though still a bit odd):

image

docs/DSP/GameCube_DSP_Users_Manual/GameCube_DSP_Users_Manual.tex
```diff
@@ -646,13 +646,15 @@ \subsection{DMA}
...
 
 \begin{RegisterBitDescriptions}
-\RegisterBitDescription{15--0}{d}{W}{}
+\RegisterBitDescription{2}{t}{R}{Transfer currently in progress if set}
+\RegisterBitDescription{1}{i}{R/W}{\begin{tabular}[c]{@{}l@{}}\Value{1} - DMA to/from IMEM\\ \Value{0} - DMA to/from DMEM\end{tabular}}
+\RegisterBitDescription{0}{d}{R/W}{\begin{tabular}[c]{@{}l@{}}\Value{1} - DMA to CPU from DSP\\ \Value{0} - DMA from CPU to DSP\end{tabular}}
 \end{RegisterBitDescriptions}
 
-\RegisterBitOverview{0xFFCB}{DSBL}{DSP Address}{dddd dddd dddd dddd}
+\RegisterBitOverview{0xFFCB}{DSBL}{Block size}{dddd dddd dddd dddd}
```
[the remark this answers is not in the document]
Pokechu22: That'd be consistent with the long name in DSPTables too:

https://github.com/dolphin-emu/dolphin/blob/17a01c894b7720e141c13638af639169a3c66f4b/Source/Core/Core/DSP/DSPTables.cpp#L378

docs/DSP/GameCube_DSP_Users_Manual/GameCube_DSP_Users_Manual.tex
```diff
@@ -858,43 +1046,68 @@ \section{Bit meanings}
...
+\texttt{0b0101}       & \texttt{Z}   & Zero                         & \Code{\$sr.Z != 0}            \\ \hline
+\texttt{0b0110}       & \texttt{NC}  & Not carry                    & \Code{\$sr.C == 0}            \\ \hline
+\texttt{0b0111}       & \texttt{C}   & Carry                        & \Code{\$sr.C != 0}            \\ \hline
+\texttt{0b1000}       & \texttt{x8}  & Below s32                    & \Code{\$sr.AS == 0}           \\ \hline
+\texttt{0b1001}       & \texttt{x9}  & Above s32                    & \Code{\$sr.AS != 0}            \\ \hline
+\texttt{0b1010}       & \texttt{xA}  &                              & \Code{((\$sr.AS != 0)  ||  (\$sr.TB != 0)) \&\& (\$sr.Z == 0)} \\ \hline
+\texttt{0b1011}       & \texttt{xB}  &                              & \Code{((\$sr.AS == 0) \&\& (\$sr.TB == 0))  ||  (\$sr.Z != 0)} \\ \hline
+\texttt{0b1100}       & \texttt{LNZ} & Not logic zero               & \texttt{(\$sr \& 0x40) == 0}  \\ \hline
+\texttt{0b1101}       & \texttt{LZ}  & Logic zero                   & \texttt{(\$sr \& 0x40) != 0}  \\ \hline
+\texttt{0b1110}       & \texttt{O}   & Overflow                     & \texttt{(\$sr \& 0x2) != 0}   \\ \hline
```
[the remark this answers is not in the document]
Pokechu22: I think I just forgot to.  There wasn't a name for the `LZ` before an earlier change, but I think I had given it one before I did these.

docs/DSP/GameCube_DSP_Users_Manual/GameCube_DSP_Users_Manual.tex
Pokechu22: It seems like [it's possible to do that using `\endhead`](https://tex.stackexchange.com/a/11381).  It's also [possible to use `\\*`](https://tex.stackexchange.com/a/468627) instead of `\\`, which tells latex to avoid breaking pages there (and thus to prefer breaking pages on `\OpcodeRowSkip` which will still use `\\`), which means I don't need to manually add an extra one to try and get good page wrapping (and if someone were to format it for letter paper instead of A4 paper, it'd handle it in a reasonably intelligent manner).

docs/DSP/GameCube_DSP_Users_Manual/GameCube_DSP_Users_Manual.tex
Pokechu22: I added a section on flags that documents the way carry and overflow work, and put that information there as well.

PR #10052, opened 2021-08-23
JitArm64: Indexed paired loadstores workaround, attempt 2

JosJuice:
PR #10050 did not make the issue go away.

### DIFF
--- a/Source/Core/Core/PowerPC/JitArm64/JitArm64_LoadStorePaired.cpp
+++ b/Source/Core/Core/PowerPC/JitArm64/JitArm64_LoadStorePaired.cpp
@@ -39,6 +39,7 @@ void JitArm64::psq_lXX(UGeckoInstruction inst)
   gpr.Lock(ARM64Reg::W0, ARM64Reg::W1, ARM64Reg::W2, ARM64Reg::W30);
   fpr.Lock(ARM64Reg::Q0, ARM64Reg::Q1);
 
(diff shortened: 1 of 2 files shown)


JosJuice:
This PR does not have the issue, according to both JMC and a forum user. I don't know why calling `gpr.R(inst.RA)` in cases where we don't need it works around the issue, but I guess it adds up with the theory that it's some kind of register cache issue. I'm still planning to look into the actual cause of the issue, but want to merge this early so that we don't leave this broken for too long.

Source/Core/Core/PowerPC/JitArm64/JitArm64_LoadStorePaired.cpp
```diff
@@ -47,11 +48,11 @@ void JitArm64::psq_lXX(UGeckoInstruction inst)
   if (inst.RA || update)  // Always uses the register on update
   {
     if (indexed)
-      ADD(addr_reg, gpr.R(inst.RA), gpr.R(inst.RB));
+      ADD(addr_reg, arm_addr, gpr.R(inst.RB));
```
degasus, line 51: Both gcc and MSVC called R(RB) before R(RA). So the order of register has changed. I doubt this is an issue here, but it can affect other (regcache) issues.

JosJuice: The instruction variants that used RB were disabled by PR 10050, without affecting the issue.

PR #10053, opened 2021-08-24
IOS/Network/KD: Minor tidy-up changes

lioncash:
Initially this was only intended to change the maps in NetKDRequest's getter functions, but snowballed into this overall set of changes as possible improvements were noticed.

Each commit is relatively self-isolated from the surrounding changes to make for ease of piece-by-piece review.


Source/Core/Core/IOS/Network/KD/NetKDRequest.cpp
```diff
@@ -21,6 +22,126 @@
...
+  RVT = 0,
+  RVV = 0,
+  RVL = 1,
+  RVD = 2,
+  Unknown = 7
+};
+
+u8 GetAreaCode(std::string_view area)
+{
+  static constexpr std::array<std::pair<std::string_view, u8>, 13> regions{{
```
BhaaLseN, line 38: Doesn't this need `utility` (for `std::pair`) and `string_view` (for, well...`std::string_view`) included?

BhaaLseN (approved):
Code LGTM (aside from that one remark, which seems like IWYU), untested.

leoetlino (approved):
> Reviewed 1 of 1 files at r1, 2 of 2 files at r2, 1 of 1 files at r3, 1 of 1 files at r4, 1 of 1 files at r5, 1 of 1 files at r6, 3 of 3 files at r7, 1 of 1 files at r8, 2 of 2 files at r9, 2 of 2 files at r10, 1 of 1 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/dolphin-emu/dolphin/10053)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @lioncash)

PR #10055, opened 2021-08-24
JitArm64: Codegen space reuse

JosJuice:
JitArm64 port of PR #8765. Fixes https://bugs.dolphin-emu.org/issues/11717.

### DIFF
--- a/Source/Core/Core/PowerPC/JitArm64/JitArm64Cache.cpp
+++ b/Source/Core/Core/PowerPC/JitArm64/JitArm64Cache.cpp
@@ -12,6 +12,12 @@ JitArm64BlockCache::JitArm64BlockCache(JitBase& jit) : JitBaseBlockCache{jit}
 {
 }
 
+void JitArm64BlockCache::Init()
+{
+  JitBaseBlockCache::Init();
+  ClearRangesToFree();
+}
+
 void JitArm64BlockCache::WriteLinkBlock(Arm64Gen::ARM64XEmitter& emit,
                                         const JitBlock::LinkData& source, const JitBlock* dest)
 {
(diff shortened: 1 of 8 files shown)


JosJuice:
I have tested that this works when playing games a short while, but not that it works when playing games for the length of time that would be needed to trigger a cache clear. @JMC47, would you like to test?

Source/Core/Core/PowerPC/JitArm64/JitArm64Cache.cpp
```diff
@@ -60,7 +66,7 @@ void JitArm64BlockCache::WriteLinkBlock(const JitBlock::LinkData& source, const
 {
   const Common::ScopedJITPageWriteAndNoExecute enable_jit_page_writes;
   u8* location = source.exitPtrs;
-  ARM64XEmitter emit(location);
+  ARM64XEmitter emit(location, location + 12);
```
Rumi-Larry, line 69: Is there a reason the number 12 was chosen? If so a comment explaining it couldn't hurt

Source/Core/Core/PowerPC/JitArm64/JitArm64Cache.cpp
```diff
@@ -69,9 +75,35 @@ void JitArm64BlockCache::WriteLinkBlock(const JitBlock::LinkData& source, const
 void JitArm64BlockCache::WriteDestroyBlock(const JitBlock& block)
 {
   // Only clear the entry points as we might still be within this block.
-  ARM64XEmitter emit(block.checkedEntry);
+  ARM64XEmitter emit(block.checkedEntry, block.normalEntry + 4);
```
Rumi-Larry, line 78: Ditto, but for the number 4.

Source/Core/Core/PowerPC/JitArm64/JitArm64Cache.cpp
JosJuice: The number of instructions here is always 3, as documented here: https://github.com/dolphin-emu/dolphin/blob/3dac4ec31cdf166419682d37c1c1f0f4e4e9c6d7/Source/Core/Core/PowerPC/JitArm64/JitArm64Cache.cpp#L26

Source/Core/Core/PowerPC/JitArm64/JitArm64Cache.cpp
JosJuice: I don't think this one needs any explaining if you look at the loop condition below.

brujo5:
From 5.0-15524  "JitArm64: Codegen space reuse" the games stutter all the time and become unplayable on android, tested with prince of persia forgotten arenas and others. and possibly the cause of the crash in ultimate spiderman in the cinematic due to the stutter introduced from this version. All versions before this one work perfectly that I have tried it several times. Thx.

DatEmulationGuy:
Same chip as @brujo and I found that F1 2002 will stutter like crazy in 15524 and probably every build after it. Game ran very good in older builds on the same chip

JMC47:
I was messing with this, it looks like some bug is causing the game to repeatedly clear the code-cache.  This will probably be fine once we figure out why that's happening.

jlagaman:
Same issue still seems to be occurring on Harry Potter PoA, on Dolphin 5.0-15586

JMC47:
We need to bisect within the commits, but compiling is being annoying.

JosJuice:
@brujo5, @DatEmulationGuy, @jlagaman: Please test PR #10267.

brujo5:
@JosJuice It work perfect as before.Thx.

PR #10056, opened 2021-08-24
DolphinQt: Minor localization changes

lioncash:
A few tweaks to the existing strings to make them a little more consistent


JosJuice:
LGTM, but let's wait with merging until after the next beta build is out (assuming we are going to do a beta build this month like I heard plans about).

PR #10060, opened 2021-08-27
DSPInterpreter: fix overflow flag in addpaxz/mulac/mulxac/mulcac

Tilka:
Verified on hardware (no test files because I'm lazy). For addpaxz I also tested whether $axS is rounded similarly to $prod. It's not, so I'm inclined to believe that the TODO is bogus, but I haven't done exhaustive testing. (At the time the comment was added, $prod was already being rounded the same way it is today.)


Pokechu22 (commented):
Code looks good (I haven't tested it though).  However, the flags should probably be `x-xx xxxx` and not `--xx xxxx`, since if the overflow bit is set, the overflow sticky bit is probably also set.  (`UpdateSR64Add` does that automatically, so this is only a documentation thing).

The GFDL is also a bit annoying about wanting an updated history section.  Unless it's likely that you'll make more changes to the documentation soon, you probably should bump the version number.

Tilka:
> the flags should probably be x-xx xxxx and not --xx xxxx

Fixed.

PR #10064, opened 2021-08-27
Gekko: Migrate register structs over to Common::BitField

lioncash:
Provides the same behavior, while also preventing undesirable type conversions and promotions on bitfield members in certain cases.

e.g.

```cpp
union Thing
{
  u32 hex = 0;

  struct
  {
    u32 value_1 : 5;
    u32 padding : 27;
  };
};

void SomeFunction(const Thing& thing)
{
  // Assume the encoding stores some value - 1, so we add to get the full value
  const u32 data = thing.value_1 + 1;
}
```

The intuitive (and completely expected most of the time) reading of this would be to think that `value_1` undergoes no type promotions or conversions, since everything is unsigned, however, this line:

```cpp
  const u32 data = thing.value_1 + 1;
```

Actually results in `value_1` being converted to an `int`, incremented, and then assigned to data, implicitly converting it back to a `u32`

`UGQR` was already using this as well, so this keeps the register layouts consistent as well.

### DIFF
--- a/Source/Core/Core/PowerPC/Gekko.h
+++ b/Source/Core/Core/PowerPC/Gekko.h
@@ -643,13 +643,11 @@ union UReg_MMCR0
 // MMCR1 - Monitor Mode Control Register 1 format
 union UReg_MMCR1
 {
-  u32 Hex;
-  struct
-  {
-    u32 : 22;
-    u32 PMC4SELECT : 5;
-    u32 PMC3SELECT : 5;
-  };
+  BitField<0, 22, u32> reserved;
+  BitField<22, 5, u32> PMC4SELECT;
+  BitField<27, 5, u32> PMC3SELECT;
+
+  u32 Hex = 0;
 };
 
 // Write Pipe Address Register
--- a/Source/Core/Core/PowerPC/Gekko.h
+++ b/Source/Core/Core/PowerPC/Gekko.h
@@ -736,12 +736,10 @@ union UReg_THRM12
 
 union UReg_THRM3
 {
-  struct
-  {
-    u32 E : 1;      // Enable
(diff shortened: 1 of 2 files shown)


Source/Core/Core/PowerPC/Gekko.h
```diff
@@ -692,14 +692,12 @@ union UReg_DMAL
...
-    u32 VS : 1;
-    u32 BL : 11;  // Block length (aka block size mask)
-    u32 : 4;
-    u32 BEPI : 15;
-  };
+  BitField<0, 1, u32> VP;
+  BitField<1, 1, u32> VS;
+  BitField<2, 11, u32> BL;  // Block length (aka block size mask)
+  BitField<13, 4, u32> reserved;
+  BitField<17, 14, u32> BEPI;
```
AdmiralCurtiss, line 699: This `14` should be `15`.

(This extremely broke the fifoci builds, haha...)

leoetlino (approved):
> Reviewed 1 of 1 files at r1, 1 of 1 files at r2, 2 of 2 files at r3, 1 of 1 files at r4, 1 of 1 files at r5, 1 of 1 files at r6, 1 of 1 files at r7, 1 of 1 files at r8, 1 of 1 files at r9, 1 of 1 files at r10, 1 of 1 files at r11, 1 of 1 files at r12, 1 of 1 files at r13, 1 of 1 files at r14, 1 of 1 files at r15, 1 of 1 files at r16, 1 of 1 files at r17, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/dolphin-emu/dolphin/10064)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @lioncash)

PR #8026, opened 2019-04-23
Fix the screen going black on pause after changing windows

Pokechu22:
This solves the screen going black if the game is paused and then the output is switched to.  It also helps with games that stop outputting frames (e.g. the homebrew crash screen); it now continues to display the previous frame instead of rendering black.

WX used to keep the last output frame visible, but Qt replaced it with black as soon as the window was focused again (or rather, as soon as it repainted itself).  Telling Qt not to do this is sufficient to keep the last frame visible; this is done by setting `WA_PaintOnScreen` and making `paintEngine` return `nullptr`.  ([More info](https://zeduckmaster.frama.io/2016/how-to-create-a-custom-rendering-in-a-qt5-widget/)).

Note that the whole `SetFillBackground` system exists because of #6519 (which fixes [bug 10958](https://bugs.dolphin-emu.org/issues/10958)).  I've confirmed that it is still fixed with this change (and that removing `SetFillBackground` and always setting the attributes to true causes the issue to come back).

This does _not_ store the frame and manually re-render it.  So, it doesn't help with e.g. screenshots, or resizing the window (which for me causes weird effects but doesn't cause it to go black anymore).

PR #10067, opened 2021-08-29
DolphinQt: Remove WA_PaintOnScreen attribute from RenderWidget

Henrik0x7F:
The attribute prevents the render widget from being resized with the mouse and I see no reason why it was there in the first place.

### DIFF
--- a/Source/Core/DolphinQt/RenderWidget.cpp
+++ b/Source/Core/DolphinQt/RenderWidget.cpp
@@ -96,7 +96,6 @@ RenderWidget::RenderWidget(QWidget* parent) : QWidget(parent)
 
   // We need a native window to render into.
   setAttribute(Qt::WA_NativeWindow);
-  setAttribute(Qt::WA_PaintOnScreen);
 }
 
 QPaintEngine* RenderWidget::paintEngine() const


leoetlino:
This was added in https://github.com/dolphin-emu/dolphin/pull/8026 to fix various issues. Are you sure removing the attribute does not cause those issues to reappear?

Henrik0x7F:
> This was added in #8026 to fix various issues. Are you sure removing the attribute does not cause those issues to reappear?

On first sight everything worked fine. Now that I checked again I noticed the RenderWidget sometimes turns black when emulation is paused. I personally prefer being able to resize the window but that decision is up to you

leoetlino:
What platform are you on? On Linux I'm able to resize the render widget correctly even with the attribute set.

OpenGL possibly being broken on macOS is somewhat more problematic.

Henrik0x7F:
I'm on KDE X11 Linux. I'm able to maximize the window but not resize it by dragging the border. Removing the attribute fixed that even though WA_PaintOnScreen usually doesn't have any effect on window sizing. Since you don't have this problem on Linux I'll open a bug report

Henrik0x7F:
Issue opened: https://bugs.dolphin-emu.org/issues/12649

PR #4723, opened 2017-01-23
Set up IOS RAM information area with real values.

booto:
Set up the IOS memory information area with real values based on the required IOS.

PR #10068, opened 2021-08-29
IOS: Add missing writes of low MEM1 constants

leoetlino:
On a real Wii, these constants are normally written by the system menu
(maybe even as part of common SDK code?)

However, they're cleared by IOS whenever a PPC title is launched.
IOS memsets 0x0-0x3fff and then manually writes some constants
in low MEM1. PR #4723 added most of the writes in the 0x31xx region
but left out the four writes to the legacy constant region.

Previously Dolphin didn't actually clear 0-0x3fff so those constants
would stick around after a system menu execution.
011f7789e0 exposed those missing writes.


JMC47:
Is this still being updated?

leoetlino:
No. This PR is complete.

PR #10069, opened 2021-08-29
Update clang-format to version 12

leoetlino:
12 is the latest release and various editors (VS, VSCode and CLion)
ship with clang-format 12 by default.

Will fail on the lint builder until it's updated


leoetlino:
installed clang-format 12 on the Ubuntu container so this should be good to go

iwubcode (approved):

PR #10072, opened 2021-08-31
DSPInterpreter: fix off-by-one errors in cycle counting

Tilka:
Previously, when calling `RunCycles(1)` we would actually run two instructions. (We currently assume that all instructions take one cycle.)


Pokechu22 (approved):

Pokechu22:
Note that the recompiler was not affected, as it uses `JA` (jump above/unsigned greater than):

https://github.com/dolphin-emu/dolphin/blob/ffdc8538a162b1ca56e3943c5a78c02358053b2a/Source/Core/Core/DSP/Jit/x64/DSPEmitter.cpp#L456-L460

I don't know if the rest of the recompiler properly handles cycle counts though.

PR #10074, opened 2021-08-31
MMU: Tidy up PTE-related code

lioncash:
This moves the split PTE definition unions into Gekko.h and eliminates the need for several macro definitions in the MMU code, since we can use the register unions to access the relevant info.

This also fixes the layout of the SDR1 union, which has been incorrect (from a little-endian POV) since it was introduced.

Each commit should be fairly self-isolated for review

### DIFF
--- a/Source/Core/Core/PowerPC/MMU.cpp
+++ b/Source/Core/Core/PowerPC/MMU.cpp
@@ -1308,34 +1308,39 @@ static TranslateAddressResult TranslatePageAddress(const u32 address, const XChe
   // TLB cache
   // This catches 99%+ of lookups in practice, so the actual page table entry code below doesn't
   // benefit much from optimization.
-  u32 translatedAddress = 0;
-  TLBLookupResult res = LookupTLBPageAddress(flag, address, &translatedAddress, wi);
+  u32 translated_address = 0;
+  const TLBLookupResult res = LookupTLBPageAddress(flag, address, &translated_address, wi);
   if (res == TLBLookupResult::Found)
+  {
     return TranslateAddressResult{TranslateAddressResultEnum::PAGE_TABLE_TRANSLATED,
-                                  translatedAddress};
+                                  translated_address};
+  }
 
-  u32 sr = PowerPC::ppcState.sr[EA_SR(address)];
+  const auto sr = UReg_SR{ppcState.sr[EA_SR(address)]};
 
-  if (sr & 0x80000000)
+  if (sr.T != 0)
     return TranslateAddressResult{TranslateAddressResultEnum::DIRECT_STORE_SEGMENT, 0};
 
   // TODO: Handle KS/KP segment register flags.
 
   // No-execute segment register flag.
-  if ((flag == XCheckTLBFlag::Opcode || flag == XCheckTLBFlag::OpcodeNoException) &&
-      (sr & 0x10000000))
+  if ((flag == XCheckTLBFlag::Opcode || flag == XCheckTLBFlag::OpcodeNoException) && sr.N != 0)
   {
     return TranslateAddressResult{TranslateAddressResultEnum::PAGE_FAULT, 0};
   }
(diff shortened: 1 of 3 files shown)


Source/Core/Core/PowerPC/MMU.cpp
```diff
@@ -1134,68 +1134,6 @@ TranslateResult JitCache_TranslateAddress(u32 address)
...
-#define PTE1_H (1 << 6)
-#define PTE1_API(v) ((v)&0x3f)
-
-#define PTE2_RPN(v) ((v)&0xfffff000)
-#define PTE2_R (1 << 8)
-#define PTE2_C (1 << 7)
-#define PTE2_WIMG(v) (((v) >> 3) & 0xf)
-#define PTE2_PP(v) ((v)&3)
-
-// Hey! these duplicate a structure in Gekko.h
```
JosJuice, line 1171: This comment is still valid. `UReg_PTE` is duplicated by `UPTE_Lo` and `UPTE_Hi`. (Or is there a reason why we want to duplicate them in this way?)

lioncash: Technically we can get rid of UReg_PTE, since it's not used anywhere, if it's preferable to remove it then I can get rid of it. Given our current code, it's only ever operated on in terms of upper and lower halves.

JosJuice: If it's not used, I think it would make sense to remove it.

Source/Core/Core/PowerPC/MMU.cpp
```diff
@@ -1300,83 +1241,96 @@ static TLBLookupResult LookupTLBPageAddress(const XCheckTLBFlag flag, const u32
...
+  ppcState.tlb[0][entry_index].Invalidate();
+  ppcState.tlb[1][entry_index].Invalidate();
 }
 
+union EffectiveAddress
+{
+  BitField<0, 4, u32> SR;
+  BitField<4, 6, u32> API;
+  BitField<4, 16, u32> page_index;
+  BitField<20, 12, u32> offset;
```
JosJuice, line 1278: Are you sure this is the right order?

lioncash: Good catch

Source/Core/Core/PowerPC/MMU.cpp
lioncash: Added a commit that removes it

Source/Core/Core/PowerPC/MMU.cpp
```diff
@@ -1300,83 +1241,96 @@ static TLBLookupResult LookupTLBPageAddress(const XCheckTLBFlag flag, const u32
...
-  tlbe_i.tag[0] = TLBEntry::INVALID_TAG;
-  tlbe_i.tag[1] = TLBEntry::INVALID_TAG;
+  ppcState.tlb[0][entry_index].Invalidate();
+  ppcState.tlb[1][entry_index].Invalidate();
 }
 
+union EffectiveAddress
+{
+  BitField<0, 12, u32> offset;
+  BitField<12, 6, u32> API;
```
JosJuice, line 1276:
```suggestion
  BitField<22, 6, u32> API;
```

JosJuice (approved):

leoetlino (approved):
> Reviewed 1 of 1 files at r1, 2 of 2 files at r2, 2 of 2 files at r3, 2 of 2 files at r4, 2 of 2 files at r5, 1 of 1 files at r6, 1 of 1 files at r7, 1 of 1 files at r8, 1 of 1 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/dolphin-emu/dolphin/10074)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @lioncash)

PR #10076, opened 2021-08-31
SI: Migrate bitfields to Common::BitField

lioncash:
Migrates the SI register union bitfields over to Common::BitField for consistency with the rest of the codebase.

While we're in the same area, we can simplify a trivial conditional

### DIFF
--- a/Source/Core/Core/HW/SI/SI.cpp
+++ b/Source/Core/Core/HW/SI/SI.cpp
@@ -11,6 +11,7 @@
 #include <memory>
 #include <sstream>
 
+#include "Common/BitField.h"
 #include "Common/ChunkFile.h"
 #include "Common/CommonTypes.h"
 #include "Common/Logging/Log.h"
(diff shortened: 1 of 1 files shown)


leoetlino (commented):
> Reviewed 1 of 1 files at r1, 1 of 1 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/dolphin-emu/dolphin/10076)* status: all files reviewed, 1 unresolved discussion (waiting on @lioncash)

___
*[Source/Core/Core/HW/SI/SI.cpp, line 492 at r2](https://reviewable.io/reviews/dolphin-emu/dolphin/10076#-MjUGJklCfMcMhk0zBtx:-MjUGJkm6oTUePpbUVim:b-qk9km5) ([raw file](https://github.com/dolphin-emu/dolphin/blob/2fb789120d7cda8c27a56ddc930ee1d4bf2083fd/Source/Core/Core/HW/SI/SI.cpp#L492)):*
> ```cpp
> 
>   // In and out for the 4 SI channels.
>   for (u32 i = 0; i < MAX_SI_CHANNELS; ++i)
> ```

Won't this cause a sign comparison warning (because of MAX_SI_CHANNELS possibly being signed, since the underlying type isn't specified explicitly)?

lioncash:
@leoetlino Ah, true, good catch. Amended the commit.

leoetlino (approved):
> Reviewed 1 of 1 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/dolphin-emu/dolphin/10076)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @lioncash)

PR #10079, opened 2021-09-01
MemoryWidget: Add negative offset search support

sepalani:
This PR allows to search/replace value from the Memory widget using a negative offset.

### DIFF
--- a/Source/Core/DolphinQt/Debugger/MemoryWidget.cpp
+++ b/Source/Core/DolphinQt/Debugger/MemoryWidget.cpp
@@ -3,6 +3,7 @@
 
 #include "DolphinQt/Debugger/MemoryWidget.h"
 
+#include <limits>
 #include <optional>
 #include <string>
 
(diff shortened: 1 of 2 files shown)


JMC47:
I ran into this the other day.  Thought it was really stupid that we couldn't do negative offset.  +1

Source/Core/DolphinQt/Debugger/MemoryWidget.cpp
```diff
@@ -446,7 +446,7 @@ void MemoryWidget::OnSearchAddress()
   bool good_addr, good_offset;
   // Returns 0 if conversion fails
   const u32 addr = m_search_address->text().toUInt(&good_addr, 16);
-  const u32 offset = m_search_offset->text().toUInt(&good_offset, 16);
+  const s32 offset = m_search_offset->text().toInt(&good_offset, 16);
```
lioncash, line 449: note that without casts below, this introduces an implicit sign/unsigned mismatch in terms of arithmetic. Ditto for the other cases as well

sepalani:
@lioncash 
Done. I changed my approach and used the `GetTargetAddress` method to compute the address and perform sanity checks. It also fixes bugs like being able to search values even if the address is invalid.

Dentomologist (commented):
This is a good feature to have, and it also lets you write values without needing an offset which was annoying.

sepalani:
@Dentomologist Done.

I addressed the offset checks by doing it both way and using `s16` instead to prevent nasal demons.

Dentomologist:
Making the offsets s16 fixes the UB but limits them to +- 32k; I don't know how often people need offsets larger than that, but I would guess that it's not so rare as to not be worth supporting.

Source/Core/DolphinQt/Debugger/MemoryWidget.h
```diff
@@ -57,6 +57,7 @@ class MemoryWidget : public QDockWidget
   void OnDumpFakeVMEM();
 
   std::vector<u8> GetValueData() const;
+  u32 GetTargetAddress(bool* good_address, bool* good_offset) const;
```
lioncash, line 60: Any reason this doesn't return a struct like:

```cpp
struct SomeName
{
  u32 address = 0;
  bool is_good_address = false;
  bool is_good_offset = false;
};
```
?

This would get rid of the out parameters and allow structured bindings.

It also gets rid of the uninitialized reads that are present in the existing code from forgetting to initialize the bools before calling the function (since `|=` is used in the function)

Dentomologist: The struct sounds good to me.

Pedantic: good_address and good_offset are initialized by QString's toUInt and toShort functions before the `|=`s, but the fact that isn't obvious is a point in favor of the struct.

sepalani:
@lioncash 
@Dentomologist 
Done.

AdmiralCurtiss:
This is still limited to +/- 32k offsets.

sepalani:
@AdmiralCurtiss 
Should be fixed, now.

AdmiralCurtiss (approved):
Looks good now.

PR #10080, opened 2021-09-01
MemoryInterface: Migrate bitfields over to Common::BitField

lioncash:
Continues the migration of low hanging fruit over to Common::BitField

While we're in the area, we can get rid of some trivial implicit sign conversions

### DIFF
--- a/Source/Core/Core/HW/MemoryInterface.cpp
+++ b/Source/Core/Core/HW/MemoryInterface.cpp
@@ -3,6 +3,7 @@
 
 #include "Core/HW/MemoryInterface.h"
 
+#include "Common/BitField.h"
 #include "Common/ChunkFile.h"
 #include "Common/CommonTypes.h"
 #include "Core/HW/MMIO.h"
@@ -51,7 +52,7 @@ enum
 
 union MIRegion
 {
-  u32 hex;
+  u32 hex = 0;
   struct
   {
     u16 first_page;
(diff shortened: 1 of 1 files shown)


Source/Core/Core/HW/MemoryInterface.cpp
```diff
@@ -169,15 +163,15 @@ void RegisterMMIO(MMIO::Mapping* mmio, u32 base)
...
+  // The naming is confusing here: the register contains the lower part of
   // the address (hence MI_..._LO but this is still the high part of the
   // overall register.
   mmio->Register(base | MI_PROT_ADDR_LO, MMIO::DirectRead<u16>(&g_mi_mem.prot_addr.hi),
                  MMIO::DirectWrite<u16>(&g_mi_mem.prot_addr.hi));
   mmio->Register(base | MI_PROT_ADDR_HI, MMIO::DirectRead<u16>(&g_mi_mem.prot_addr.lo),
                  MMIO::DirectWrite<u16>(&g_mi_mem.prot_addr.lo));
 
-  for (int i = 0; i < 10; ++i)
+  for (u32 i = 0; i < g_mi_mem.timers.size(); ++i)
```
iwubcode, line 174:
```
for (std::size_t i = 0; i < g_mi_mem.timers.size(); ++i)
```
?

lioncash: This would make the formed addresses 64-bit, which would cause truncations

iwubcode: Ok, I guess then cast?

```
for (u32 i = 0; i < static_cast<u32>(g_mi_mem.timers.size()); ++i)
```

lioncash: I mean, it's fine as is, since they're the same signedness and there's no loss of information.

leoetlino (approved):
> Reviewed 1 of 1 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/dolphin-emu/dolphin/10080)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @lioncash)

PR #10083, opened 2021-09-02
Disable Dual Core by default for multiple Heavy Iron games

JoeyBallentine:
BFBB tends to crash when dual-core is enabled. Interestingly enough, this does not happen on 5.0 stable (actually, the loads are more accurate to an actual Gamecube on stable as well, but that's a different issue).

The main issue is that randomly while playing the game with dual-core enabled, the game will crash with a FIFO error. Disabling dual-core fixes the issue. I have seen many people ask why BFBB is giving them a FIFO error randomly, and every time they are able to fix it by disabling dual-core. Therefore, I believe it should be disabled by default for this game.

### DIFF
--- a/Data/Sys/GameSettings/GGV.ini
+++ b/Data/Sys/GameSettings/GGV.ini
@@ -2,6 +2,8 @@
 
 [Core]
 # Values set here will override the main Dolphin settings.
+# Dual Core mode causes FIFO error
+CPUThread = False
 
 [OnLoad]
 # Add memory patches to be loaded once on boot here.
--- a/Data/Sys/GameSettings/GIC.ini
+++ b/Data/Sys/GameSettings/GIC.ini
@@ -2,6 +2,8 @@
 
 [Core]
 # Values set here will override the main Dolphin settings.
+# Dual Core mode causes FIFO error
+CPUThread = False
 
 [OnLoad]
 # Add memory patches to be loaded once on boot here.
--- a/Data/Sys/GameSettings/GIH.ini
+++ b/Data/Sys/GameSettings/GIH.ini
@@ -2,6 +2,8 @@
 
 [Core]
 # Values set here will override the main Dolphin settings.
+# Dual Core mode causes FIFO error
+CPUThread = False
 
 [OnLoad]
 # Add memory patches to be loaded once on boot here.
--- a/Data/Sys/GameSettings/GIQ.ini
+++ b/Data/Sys/GameSettings/GIQ.ini
@@ -2,6 +2,8 @@
 
 [Core]
 # Values set here will override the main Dolphin settings.
+# Dual Core mode causes FIFO error
+CPUThread = False
 
 [OnLoad]
 # Add memory patches to be loaded once on boot here.
--- a/Data/Sys/GameSettings/GQP.ini
+++ b/Data/Sys/GameSettings/GQP.ini
@@ -2,6 +2,8 @@
 
 [Core]
 # Values set here will override the main Dolphin settings.
+# Dual Core mode causes FIFO error
+CPUThread = False
 
 [OnLoad]
 # Add memory patches to be loaded once on boot here.


JMC47:
Something that might be worth doing before we merge this is bisecting the various issues you're having so we knew *what* caused these differences.  We could merge disabling dualcore as a hotfix, but I'd like to know what changed the behavior, if it was a dualcore regression or something unavoidable, does SyncGPU do enough, etc.

JoeyBallentine:
Sure. What do you recommend I try?

JMC47:
I might be best to merge this as a hotfix, to make sure the game is working in time for the progress report.  But, for longer term maintenance...

1:  I'd like to figure out which build regressed dualcore.  It's possible we could fix it and then restore the settings back to normal if it's some kind of bug... and even if it's not, it'd be nice to know if a change broke things.
2:  Try enabling SyncGPU in the game properties page (right click the game in the gamelist to find that menu) and see if SyncGPU is better performance than single core and if it fixes the crashing.

JMC47:
@dolphin-emu-bot rebuild

JMC47:
Since this game is featured in the next progress report, I'd like to make sure that it's running correctly.  Do you know how much of a performance loss Single Core is?  I don't know how demanding this game is in general, but I *always* use single core and never saw any slowdown.  That's also why I missed that dualcore was broken.

JoeyBallentine:
I never really noticed a performance drop specifically from single core. I do know that a load in the game (going from downtown to the hub) used for a speedrun trick has a slightly longer loading time on dev than on stable. On stable, the timing for the trick is 8.7 seconds, and on dev its somewhere around 9.2 seconds. I have no idea if this is from single core or not. The rest of the game's performance seems unchanged from single vs dual. I'd do some timing of dual core vs single core loads but I know the game will randomly crash if I play it with dual core on.

JMC47:
Loading and whatnot won't be affected.  I'm talking about maximum possible performance, for low-end computers and whatnot.   My main concern is making sure the game is stable and still has reasonable performance. 

As for the timings being slower on the newer builds, are you sure that it's not that older builds were simply too fast?  Dolphin's disc timings have seen a ton of improvements since Dolphin 5.0 to try to accurately reflect how a real GameCube/Wii drive loads.  Dolphin usually ends up within the timing range of an original Wii drive, as that's what we used to create our timing models.

Something that would make things way too slow would be if you're using an nkit, as they move files to the inside of the disc where the timings are slower due to CAV.  If there's a significant difference between the timing of the trick in Dolphin and on console, feel free to write a separate issue report for that.

JosJuice:
Loading times normally have nothing to do with single core vs dual core. If you want to measure performance and your computer is running the game at full speed, you should disable the speed limit in Dolphin's settings.

JoeyBallentine:
My bad -- I tend to think of performance in terms of load times from speedrunning.

JMC47:
Haha, it's fine.  I just want to make sure that the game is running as correctly as possible for next beta.

JoeyBallentine:
I don't really have a low end machine to test on, So I'll try disabling the speed limit and see if I can get a valid comparison. Should I just be comparing FPS values?

JMC47:
Yeah, just compare FPS values.  You can use a savestate to get an identical situation between single core and dualcore, as savestates will be compatible.

JoeyBallentine:
Sounds good. I'll try that.

JMC47:
If the crashing is too common, taking a performance hit is fine.  It's just that I'd like to know how big of a performance hit that we're talking about before we merge it.

JoeyBallentine:
Looks like its about a 50 fps drop using single core.
**dual core:** 197
**single core:** 148

JoeyBallentine:
And the crashing seems to be extremely common. Seems like every new bfbb speedrunner who plays on dolphin gets some form of the FIFO crash, and this always fixes it

JoeyBallentine:
I'm currently trying to do a survey of people to see if anyone *doesn't* get this issue. I'll let you know if I get anything useful back

JoeyBallentine:
Got a few useful responses:

> ive always had errors off in the past so idk
 i know my game will "freeze" on the menu
 where i have to restart the game
 but only ever the main menu
 never happened ingame
 idk ive never turned off dual core lol
 ive only ever had it in menu
 since using dolphin since 2016
 from the bubbles to the menu options to the new game screen
 but never once ingame

> it happens occasionally sometimes
 i dont really know of a fix other than trying again and getting into menu as soon as possible lol
 **ME: when it freezes does it do a regular freeze or does it do repeating audio?**
 the shit where it fades in repeatedly?
**ME: yeah thats probably the same error then, and you just arent seeing the panic handler popup**
 that happens on the menu though is what i mean
 like halfway down the bubble transition thing usually

> I had it in game a couple of times, usually would happen if I stopped giving inputs/went afk in between runs, but it happened once or twice in a run I think too
 Was a not so great laptop

> **ME: Hey btw on dolphin you got that FIFO issue right?**
Ye
**ME: And turning off dual core fixed it?**
I'm pretty sure

And then some various times people have asked how to fix it:

> can i please not use a development build of dolphin for the bfbb randomizer and other games in general because I keep on getting the FIFO Thing that says overflowed CPU, and I press Ignore for this session and it just freezes in place which is very annoying.
 well I am getting FIFO player on GatherPipe error still when I load up on the title screen
just froze again, even with panic handlers off, but this time while I was gonna make a save file

^ this person got back to me and said this:
> okay, the first time I loaded up with dual core off fixed it for now, also I adjusted my cpu clock to 15% to increase fps on the game, but the result it kinda stutters a bit.

> Person 1: so I just lost a run cuz my dolphin crashed in the middle of museum 40 minutes in. The error read "FIFO is overflowed by gatherpipe! CPU thread is too fast!"
Person 2: ​If that happens again, clicking “yes” will dismiss it. Then just wait a few seconds and click on the emulation window and it should be fine.
Sometimes you’ll get locked at 0fps from those but it’s uncommon. But if the 0fps and 60vps lock happens you have to reset emulation.
Waiting a bit after clicking yes seems to decrease chances of that happening (for me)
Person 1: Whenever I press yes on the error most of the time it just brings up 50 other errors so idk
Would rather try to just prevent it from happening in general
I'll see if turning off dual core help
Person 2: I just keep spamming yes until it goes away, if it doesn’t then do configure > advanced > enable CPU clock override and then lowering that slider seems to make it happen less often

> Emulator already ran into a problem before I could get to the title screen
[insert picture of FIFO overflow error]

> (this one is all the way back from 2018) FIFO is overflown by GatherPipe, CPU thread is too fast!  anyone have a quick fix for this dolphin error?
if not i'll come back later

None of the rest of these people ever got back on if they fixed it, even after being told to turn dual core off. I know there's been more times this has been asked, I just can't find them becuase they got lost in the depths of discord

JMC47:
3/4th performance for stability isn't a bad trade imo.

JMC47 (approved):
Appears to not be a regression, but I'd rather the game be stable.

JoeyBallentine:
Yeah it seems like this has been an issue since at least 2016.

I haven't tried it with SyncGPU yet. I can test that out later and see if I get better performance and if I no longer get the crash. The only thing is I'm not 100% sure I can replicate the conditions of the crash with certainty that its fixed, I'll just have to play the game for a while and assume it's fixed if it hasn't crashed yet

JMC47:
We can enable Single Core in beta until then.  If this is a serious issue that the speedrunners run into, then I'd rather things be able and correct than fast.

JoeyBallentine:
@JMC47 I've been told this problem actually occurs on all heavy iron games. Might be worth setting them to single core as well

JMC47:
If you want, you could modify the other ones that use this engine and git commit --amend

I don't know if their Scooby Doo game is exactly the same, but SpongeBob Movie game and the two incredibles games are nearly identical to Battle For Bikini Bottom in how they render.

JoeyBallentine:
I'm pretty sure Scooby is nearly identical as well. I'll add those games to this PR and rename it

nf6429:
If you want I can test the 2 games in 1 version of Bikini Bottom and SpongeBob Movie. When I tried the BB compilation, it did have dualcore errors._

JoeyBallentine:
Sure, tbh I didn't even know that existed

JoeyBallentine:
image
I can now confirm that Sync GPU + Dual Core still causes this, all I had to do was sit on the main menu for a while

JoeyBallentine:
FWIW I've had single core running in unlimited FPS mode for around a half hour now, no crash.

JMC47:
The "Two games in One" version if BFBB is 100% the same as one of the PAL releases, so that isn't necessary.

nf6429:
Yeah your right, i tested it, had instant menu issues with BB, less issues with SM, but still some.

PR #10084, opened 2021-09-03
Identify Triforce games in gamelist

Zopolis4:
Right now, this is moreso laying the groundwork, but I'm submitting it so my branch wont get out of date and to test for regressions early. The readme is unedited because I don't think compatibility has progressed far enough to deserve that.

Nevertheless, I'll include a short writeup of all of the changes, with screenshots included (when applicable).

Firstly, adding support for .bin images. The Triforce images dolphin accepts are extracted from their original GD-ROM's and decrypted. While some of these images have a .GCM extension, most have .BIN extensions, and this change allows dolphin to detect these images without having to change their original extension.

Secondly, adding a proper Triforce platform and boot.id parsing. Previously, dolphin would identify Triforce games as Gamecube games. This rectifies that, via reading the boot.id. Triforce games have a different, internal header than Gamecube and Wii games, located in the root of the disc section of the fst as boot.id. Regular GC/Wii games use the boot.bin header, located in the root of the sys section of the fst, and at the start of the disc image. This header is useless for triforce games, as it usually contains default values (RELSAB, Sample Game Name), or a single name for all Virtua Striker games.

Thirdly, adding a proper Triforce title database linked to the four-lettter triforce ID found in boot.id.

I'm combining these changes into one image, as they mainly affect the gamelist.

Master:
image

This PR:
image

Fourthly, implementing proper reading of DI commands and support for inquiry-type DI commands. Right now, dolphin reads Triforce DI commands as encrypted _and_ the wrong endian. As you can imagine, this is not very helpful. These changes allow Triforce games to "boot", instead of throwing a panic alert about an invalid DI command then several invalid read/write panic alerts.

Fifthly, adding partial support for the EXI side of the Triforce baseboard. This allows games to boot a bit further, with one group even displaying on-screen error text!

I'm also combining both of these changes into one image, as the EXI changes are mostly supplemental.

Master:
image

This PR:
image

I've also added the Triforce IPL hash to Dolphin's database, because it makes sense to document it and we might end up using it in LLE emulation.

And finally, fixing the Triforce GameINIs. Nothing much here, just that most of the GameINIs for Triforce games were spurious or contained incorrect information. No image here, no need for one.


Source/Core/DiscIO/VolumeTri.h
```diff
@@ -0,0 +1,108 @@
...
+enum class Country;
+class FileSystem;
+enum class Language;
+enum class Region;
+enum class Platform;
+
+class VolumeTri : public VolumeDisc
+{
+public:
+  VolumeTri(std::unique_ptr<BlobReader> reader);
```
lioncash, line 31:
```suggestion
  explicit VolumeTri(std::unique_ptr<BlobReader> reader);
```

Source/Core/Core/HW/EXI/EXI_DeviceBaseboard.cpp
```diff
@@ -0,0 +1,56 @@
...
+  {
+    m_command = byte;
+  }
+  else
+  {
+    switch (m_command)
+    {
+    case init:
+    {
+      constexpr std::array<u8, 4> ID = {0x06, 0x04, 0x10, 0x00};
```
lioncash, line 40:
```suggestion
      static constexpr std::array<u8, 4> ID = {0x06, 0x04, 0x10, 0x00};
```

Source/Core/Core/HW/EXI/EXI_DeviceBaseboard.h
```diff
@@ -0,0 +1,32 @@
...
+class CEXIBaseboard : public IEXIDevice
+{
+public:
+  CEXIBaseboard();
+  void SetCS(int CS) override;
+  bool IsPresent() const override;
+  void DoState(PointerWrap& p) override;
+
+private:
+  enum
```
lioncash, line 21:
```suggestion
  enum class Command
```

Source/Core/Core/HW/EXI/EXI_DeviceBaseboard.cpp
```diff
@@ -0,0 +1,56 @@
...
+#include "Core/HW/EXI/EXI_DeviceBaseboard.h"
+
+#include "Common/Assert.h"
+#include "Common/ChunkFile.h"
+#include "Common/CommonTypes.h"
+#include "Common/Logging/Log.h"
+
+namespace ExpansionInterface
+{
+CEXIBaseboard::CEXIBaseboard()
```
lioncash, line 13:
```suggestion
CEXIBaseboard::CEXIBaseboard() = default;
```

Source/Core/Core/HW/EXI/EXI.cpp
```diff
@@ -62,6 +62,21 @@ void AddMemoryCards(int i)
 
   g_Channels[i]->AddDevice(memorycard_device, 0);
 }
+
+void AddSP1Device()
+{
+  TEXIDevices sp1_device;
```
lioncash, line 68:
```suggestion
  TEXIDevices sp1_device = EXIDEVICE_BASEBOARD;
```

Then you only need to handle the else case instead of both the if and else.

Source/Core/Core/HW/DVD/DVDInterface.cpp
```diff
@@ -858,15 +863,49 @@ void ExecuteCommand(ReplyType reply_type)
...
+          case 0x80000000:
+            memset(Memory::GetPointer(s_DIMAR), 0, s_DICMDBUF[2]);
+            break;
+          // Media board status (2)
+          case 0x80000020:
+            memset(Memory::GetPointer(s_DIMAR), 0, s_DICMDBUF[2]);
+            break;
+          // Media board status (3)
+          case 0x80000040:
+            memset(Memory::GetPointer(s_DIMAR), 0xFFFFFFFF, s_DICMDBUF[2]);
```
lioncash, line 889:
```suggestion
            memset(Memory::GetPointer(s_DIMAR), 0xFF, s_DICMDBUF[2]);
```

Since memset only writes on a by-byte basis.

Source/Core/Core/HW/DVD/DVDInterface.cpp
```diff
@@ -858,15 +863,49 @@ void ExecuteCommand(ReplyType reply_type)
...
+            // DIMM size
+            Memory::Write_U32(0x20, s_DIMAR);
+            // GCAM signature
+            Memory::Write_U32(0x4743414D, s_DIMAR + 4);
+            break;
+          // Firmware status (1)
+          case 0x80000120:
+            memset(Memory::GetPointer(s_DIMAR), 0x01010101, s_DICMDBUF[2]);
```
lioncash, line 897:
```suggestion
            memset(Memory::GetPointer(s_DIMAR), 0x01, s_DICMDBUF[2]);
```

Source/Core/Core/HW/DVD/DVDInterface.cpp
```diff
@@ -858,15 +863,49 @@ void ExecuteCommand(ReplyType reply_type)
...
+            break;
+          // Firmware status (2)
+          case 0x80000140:
+            memset(Memory::GetPointer(s_DIMAR), 0x01010101, s_DICMDBUF[2]);
```
lioncash, line 901:
```suggestion
            memset(Memory::GetPointer(s_DIMAR), 0x01, s_DICMDBUF[2]);
```

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/utils/FileBrowserHelper.java
```diff
@@ -30,7 +30,7 @@
 public final class FileBrowserHelper
 {
   public static final HashSet<String> GAME_EXTENSIONS = new HashSet<>(Arrays.asList(
-          "gcm", "tgc", "iso", "ciso", "gcz", "wbfs", "wia", "rvz", "wad", "dol", "elf"));
+          "gcm", "tgc", "bin", "iso", "ciso", "gcz", "wbfs", "wia", "rvz", "wad", "dol", "elf"));
```
lioncash, line 33: Should we really be supporting `.bin`? This is an already super generic extension that's been overloaded into oblivion for dumps

Source/Core/DiscIO/VolumeTri.cpp
```diff
@@ -0,0 +1,241 @@
...
+
+  m_file_system = [this]() -> std::unique_ptr<FileSystem> {
+    auto file_system = std::make_unique<FileSystemGCWii>(this, PARTITION_NONE);
+    return file_system->IsValid() ? std::move(file_system) : nullptr;
+  };
+
+  m_converted_banner = [this] { return LoadBannerFile(); };
+}
+
+VolumeTri::~VolumeTri()
```
lioncash, line 45:
```suggestion
VolumeTri::~VolumeTri() = default;
```

Source/Core/DiscIO/VolumeTri.h
```diff
@@ -0,0 +1,108 @@
...
+  ~VolumeTri();
```
lioncash, line 32:
```suggestion
  ~VolumeTri() override;
```

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/utils/FileBrowserHelper.java
Zopolis4: Many triforce games are in .bin formats by default, which is why I added it

Source/Core/Core/HW/EXI/EXI.cpp
```diff
@@ -62,6 +62,21 @@ void AddMemoryCards(int i)
 
   g_Channels[i]->AddDevice(memorycard_device, 0);
 }
+
+void AddSP1Device()
+{
+  TEXIDevices sp1_device = EXIDEVICE_BASEBOARD;
+  if (SConfig::GetInstance().bTri)
+  {
+    sp1_device = EXIDEVICE_BASEBOARD;
```
lioncash, line 71: this if-else can be collapsed down to the else case, since it's already assigned.

Source/Core/Core/HW/EXI/EXI_DeviceBaseboard.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+#include "Common/Assert.h"
+#include "Common/ChunkFile.h"
+#include "Common/CommonTypes.h"
+#include "Common/Logging/Log.h"
+
+namespace ExpansionInterface
+{
+CEXIBaseboard::CEXIBaseboard() = default;
+{
```
lioncash, line 14: the braces need to be removed, since it's defaulted

Source/Core/Core/HW/EXI/EXI_DeviceBaseboard.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+void CEXIBaseboard::TransferByte(u8& byte)
+{
+  if (m_position == 0)
+  {
+    m_command = byte;
+  }
+  else
+  {
+    switch (m_command)
```
lioncash, line 36:
```suggestion
    switch (static_cast<Command>(m_command))
```

Source/Core/Core/HW/EXI/EXI_DeviceBaseboard.cpp
```diff
@@ -0,0 +1,56 @@
...
+    {
+    case init:
```
lioncash, line 38:
```suggestion
    case Command::init:
```

Source/Core/DiscIO/VolumeTri.cpp
```diff
@@ -0,0 +1,241 @@
...
+  m_file_system = [this]() -> std::unique_ptr<FileSystem> {
+    auto file_system = std::make_unique<FileSystemGCWii>(this, PARTITION_NONE);
+    return file_system->IsValid() ? std::move(file_system) : nullptr;
+  };
+
+  m_converted_banner = [this] { return LoadBannerFile(); };
+}
+
+VolumeTri::~VolumeTri() = default;
+{
```
lioncash, line 46: The braces need to be removed

Source/Core/DiscIO/VolumeGC.cpp
```diff
@@ -129,11 +131,25 @@ const BlobReader& VolumeGC::GetBlobReader() const
...
+{
+  return m_reader;
+}
+
 Platform VolumeGC::GetVolumeType() const
 {
   return Platform::GameCubeDisc;
 }
 
+bool VolumeGC::IsTriforceGame() const
```
iwubcode, line 144: I know nothing about this code but it seems a bit odd to me that this exists in the `VolumeGC` class.  Would be nice if this could somehow be a static on the `VolTri` class instead.

Source/Core/DiscIO/VolumeTri.h
```diff
@@ -0,0 +1,108 @@
...
+{
+class BlobReader;
+enum class BlobType;
+enum class Country;
+class FileSystem;
+enum class Language;
+enum class Region;
+enum class Platform;
+
+class VolumeTri : public VolumeDisc
```
iwubcode, line 28: Could probably mark this `final`

Source/Core/DiscIO/VolumeTri.h
```diff
@@ -0,0 +1,108 @@
...
+{
+public:
+  explicit VolumeTri(std::unique_ptr<BlobReader> reader);
+  ~VolumeTri() override;
```
iwubcode, line 32: For new code, I recommend implementing the [rule of five](https://en.cppreference.com/w/cpp/language/rule_of_three)

Source/Core/DiscIO/VolumeGC.cpp
Zopolis4: Should be possible, that would mean Gamecube games are assigned to VolumeTri by default, and it would also make things uglier when I implement boot.id parsing. This code isn't mine though, so I'm not 100% sure how to go about it

iwubcode: Well, I was hoping you could do something like:

```
if VolumeTri::IsTriforceGame(reader)
  return std::make_unique<VolumeTri>(std::move(reader));
  else
   return std::make_unique<VolumeGC>(std::move(reader));
```

That way keeping all the Trifoce logic separated in its own volume class and not polluting the Gamecube stuff.

But, honestly, I don't really know much about this code so it'd be better to wait for someone else to comment on it :)

Source/Core/DiscIO/VolumeTri.cpp
```diff
@@ -0,0 +1,239 @@
+// Copyright 2008 Dolphin Emulator Project
```
mbc07, line 1: Since it's a new file, you should probably change the year to 2021

Source/Core/DiscIO/VolumeTri.h
```diff
@@ -0,0 +1,108 @@
+// Copyright 2008 Dolphin Emulator Project
```
mbc07, line 1: Same here

Source/Core/DiscIO/VolumeTri.h
Zopolis4: Mark the VolumeTri class final?

iwubcode: Yes

Source/Core/Core/HW/DVD/DVDInterface.cpp
```diff
@@ -822,6 +822,11 @@ void ExecuteCommand(ReplyType reply_type)
   if (static_cast<DICommand>(s_DICMDBUF[0] >> 24) != DICommand::RequestError)
     SetDriveError(DriveError::None);
 
+  if (DVDThread::GetDiscType() == DiscIO::Platform::Triforce)
+  {
+    s_DICMDBUF[0] <<= 24;
```
BhaaLseN, line 827: This seems like an odd place to do this, with the exact value being read (and shifted right by 24) immediately before _and_ afterwards. Depending on what you want to achieve with this, it should probably a lot earlier, a lot later, or maybe even not at all.

Source/Core/Core/Boot/Boot_BS2Emu.cpp
```diff
@@ -163,6 +163,12 @@ bool CBoot::RunApploader(bool is_wii, const DiscIO::VolumeDisc& volume)
   // return
   PC = PowerPC::ppcState.gpr[3];
 
+  if (volume.GetVolumeType() == DiscIO::Platform::Triforce)
+  {
+    char* ptr = (char*)(Memory::GetPointer(0));
+    memset(ptr, 0, 12);
```
BhaaLseN, line 169: We have `Memory::Memset`, maybe you could use that instead?

Also: is there any struct we have that corresponds to the 12 bytes being null'ed, or can we at least name that magic number?

Source/Core/Core/HW/DVD/DVDInterface.cpp
```diff
@@ -858,15 +863,49 @@ void ExecuteCommand(ReplyType reply_type)
...
-
+      if (DVDThread::GetDiscType() == DiscIO::Platform::Triforce)
+      {
+        if (dvd_offset & 0x80000000)
+        {
+          switch (dvd_offset)
+          {
+          // Media board status (1)
+          case 0x80000000:
+            memset(Memory::GetPointer(s_DIMAR), 0, s_DICMDBUF[2]);
```
BhaaLseN, line 881: You can probably do this instead, both here and for the following ones:
```suggestion
            Memory::Memset(s_DIMAR, 0, s_DICMDBUF[2]);
```

Source/Core/DiscIO/Volume.cpp
```diff
@@ -91,9 +92,15 @@ static std::unique_ptr<VolumeDisc> CreateDisc(std::unique_ptr<BlobReader>& reade
...
+  {
+    auto volume = std::make_unique<VolumeGC>(std::move(reader));
+    if (volume->IsTriforceGame())
+      return std::make_unique<VolumeTri>(std::move(volume->GetBlobReaderPtr()));
+    else
+      return volume;
+  }
 
-  // No known magic words found
+    // No known magic words found
```
BhaaLseN, line 103: That spacing is probably unintended.

Source/Core/DiscIO/VolumeGC.cpp
```diff
@@ -27,10 +27,11 @@
 #include "DiscIO/FileSystemGCWii.h"
 #include "DiscIO/Filesystem.h"
 #include "DiscIO/Volume.h"
+#include "DiscIO/VolumeTri.h"
 
 namespace DiscIO
 {
-VolumeGC::VolumeGC(std::unique_ptr<BlobReader> reader) : m_reader(std::move(reader))
+  VolumeGC::VolumeGC(std::unique_ptr<BlobReader> reader) : m_reader(std::move(reader))
```
BhaaLseN, line 34: Same here. Lint should yell at you, and clang-format would fix it for you.

Source/Core/Core/HW/EXI/EXI_DeviceBaseboard.cpp
```diff
@@ -0,0 +1,54 @@
...
+    }
+  }
+  m_position++;
+}
+
+void CEXIBaseboard::DoState(PointerWrap& p)
+{
+  p.Do(m_position);
+  p.Do(m_command);
+}
```
BhaaLseN, line 53: Normally we change the state version when updating any `DoState`, but I'm not sure if we have to do this here (when the hardware actually didn't exist prior.)
https://github.com/dolphin-emu/dolphin/blob/4b8b53ac732645f6c0fd3f4b11733cb3d86c9c41/Source/Core/Core/State.cpp#L76

Source/Core/Core/HW/DVD/DVDInterface.cpp
Zopolis4: Triforce games send the DI commands in a different endian, so this swaps the endian back

Source/Core/Core/Boot/Boot_BS2Emu.cpp
Zopolis4: I wouldn't call it a magic number per se, this is blanking the session key for the encryption of triforce DI commands

Source/Core/DiscIO/VolumeTri.h
Zopolis4: Why only VolumeTri? Wouldn't VolumeGC be eligible as well? Or any of the other classes in DiscIO?

iwubcode: Yes they would.  `VolumeTri` is new code, the others would need to happen in a cleanup review most likely.

Source/Core/DiscIO/Enums.h
```diff
@@ -15,6 +15,7 @@ namespace DiscIO
 enum class Platform
 {
   GameCubeDisc = 0,
+  Triforce,
```
AdmiralCurtiss, line 18: Is there a reason why this is just `Triforce` instead of `TriforceDisc` (to match `GameCubeDisc` and `WiiDisc`)?

JMC47: Maybe because TriForce games can be discs or cartridges?

AdmiralCurtiss: And those two formats are identical data/format-wise?

Source/Core/DiscIO/VolumeGC.h
```diff
@@ -76,6 +78,11 @@ class VolumeGC : public VolumeDisc
                                                     // (only one for BNR1 type)
   };
 
+  struct BootID
+  {
+    u32 id;  // BTID
+  };
+
```
AdmiralCurtiss, line 85: This is used in exactly one place, and it doesn't exactly improve clarity there over just reading into a `u32 boot_id`. Do you really need this struct?

Source/Core/DiscIO/VolumeGC.cpp
```diff
@@ -129,11 +131,25 @@ const BlobReader& VolumeGC::GetBlobReader() const
...
   return Platform::GameCubeDisc;
 }
 
+bool VolumeGC::IsTriforceGame() const
+{
+  constexpr u32 BTID_MAGIC = 0x44495442;
+  BootID triforce_header;
+  ReadFile(*this, PARTITION_NONE, "boot.id", reinterpret_cast<u8*>(&triforce_header),
+           sizeof(BootID));
+  return triforce_header.id == BTID_MAGIC;
```
AdmiralCurtiss, line 150: `triforce_header` is potentially uninitialized if the `ReadFile()` fails. You should check the return value of `ReadFile()`.

Source/Core/DiscIO/Enums.h
Zopolis4: I'm not really sure, given nobody has extracted a correct image from the NAND dumps.

Source/Core/DiscIO/VolumeGC.h
Zopolis4: It will be used later to implement reading the true id's from boot.id, but I can remove it until I implement that

Source/Core/Core/HW/DVD/DVDInterface.cpp
BhaaLseN: 1. Shouldn't this be done _before_ line 822 checks for `DICommand::RequestError`?
2. This doesn't swap endinanness, it just shifts the value and discards 3 bytes. You might wanna use `Common::swap32` instead then (especially for clarity on what is going on.)
3. It seems weird that it uses a different endianness compared to the GC, but if this is really the case, I'd like to see a comment highlighting this (mainly because _I_ didn't know about this, and I'm fairly certain other devs also don't)

Source/Core/Core/HW/DVD/DVDInterface.cpp
```diff
@@ -858,15 +863,49 @@ void ExecuteCommand(ReplyType reply_type)
...
+            // GCAM signature
+            Memory::Write_U32(0x4743414D, s_DIMAR + 4);
+            break;
+          // Firmware status (1)
+          case 0x80000120:
+            memset(Memory::GetPointer(s_DIMAR), 0x01, s_DICMDBUF[2]);
+            break;
+          // Firmware status (2)
+          case 0x80000140:
+            memset(Memory::GetPointer(s_DIMAR), 0x01, s_DICMDBUF[2]);
```
BhaaLseN, line 901: I didn't want to mark all those, since they're next to each other; but you only changed the first one - so please be consistent here and use `Memory::Memset` as well.

Source/Core/Core/HW/DVD/DVDInterface.cpp
Zopolis4: Here's a rough outline of what happens (with the session key blanked out), using F-Zero AX (Rev C) as an example:
```
Triforce game sends read command: 0x****00a8
Command is left shifted: a8000000
Command is right shifted: a8 (Command is now normal)
```
I will add a comment though, and yeah it probably should be before 822.

Source/Core/DiscIO/VolumeGC.h
Zopolis4: I wasn't able to figure out how to implement ReadFile without using a struct, so that's why it's there.

Source/Core/Core/HW/DVD/DVDInterface.cpp
Zopolis4: Using Common::swap32 means that the randomised bytes arent zeroed out, leading to commands being incorrectly interpreted as read subcommands, which causes issues.

BhaaLseN: Then it _really_ needs a comment when the bytes are _supposed to be_ zeroed out.

Source/Core/DiscIO/Enums.h
Zopolis4: But yes, the extracted images should be identical in theory

Source/Core/DiscIO/VolumeTri.h
```diff
@@ -0,0 +1,108 @@
...
+    std::vector<u32> image_buffer;
+    u32 image_height = 0;
+    u32 image_width = 0;
+  };
+
+  ConvertedGCBanner LoadBannerFile() const;
+  ConvertedGCBanner ExtractBannerInformation(const GCBanner& banner_file, bool is_bnr1) const;
+
+  static const size_t BNR1_SIZE = sizeof(GCBanner) - sizeof(GCBannerInformation) * 5;
+  static const size_t BNR2_SIZE = sizeof(GCBanner);
```
BhaaLseN, line 99: Those look like they could be `constexpr`, or am I missing something here?

Source/Core/DiscIO/VolumeTri.h
```diff
@@ -0,0 +1,108 @@
...
+  u64 GetSize() const override;
+  bool IsSizeAccurate() const override;
+  u64 GetRawSize() const override;
+  const BlobReader& GetBlobReader() const override;
+
+  std::array<u8, 20> GetSyncHash() const override;
+
+private:
+  static const u32 GC_BANNER_WIDTH = 96;
+  static const u32 GC_BANNER_HEIGHT = 32;
```
BhaaLseN, line 57: Any reason we don't make those `constexpr`?

Zopolis4: VS complains about `constexpr` not being valid here when I replace them, and it messes up the functions below

Source/Core/DiscIO/VolumeTri.h
Zopolis4: VS throws the same "`  constexpr` is not valid here" error

JosJuice: It should be possible to use `constexpr` here. You might not have gotten the syntax right.

Source/Core/DiscIO/VolumeGC.cpp
Zopolis4: It cant be static because it uses `this`, but I've moved it to VolumeTri

Source/Core/Core/HW/EXI/EXI_DeviceBaseboard.h
```diff
@@ -0,0 +1,32 @@
...
+  enum class Command
+  {
+    init = 0x00
+  };
+
+  // STATE_TO_SAVE
+  u32 m_position = 0;
+  u32 m_command = 0;
+
+  void TransferByte(u8& _uByte) override;
```
lioncash, line 30:
```suggestion
  void TransferByte(u8& byte) override;
```

So this matches the implementation

Source/Core/Core/Boot/Boot_BS2Emu.cpp
```diff
@@ -163,6 +163,11 @@ bool CBoot::RunApploader(bool is_wii, const DiscIO::VolumeDisc& volume)
   // return
   PC = PowerPC::ppcState.gpr[3];
 
+  if (volume.GetVolumeType() == DiscIO::Platform::Triforce)
+  {
+    Memory::Memset(0, 0, 12);
```
lioncash, line 168: Maybe add an explanatory comment about why this is necessary. This looks pretty bizarre in isolation

Source/Core/Core/HW/DVD/DVDInterface.cpp
```diff
@@ -858,15 +864,49 @@ void ExecuteCommand(ReplyType reply_type)
...
           DVDINTERFACE,
           "Read: DVDOffset={:08x}, DMABuffer = {:08x}, SrcLength = {:08x}, DMALength = {:08x}",
           dvd_offset, s_DIMAR, s_DICMDBUF[2], s_DILENGTH);
 
       if (s_drive_state == DriveState::ReadyNoReadsMade)
         SetDriveState(DriveState::Ready);
-
+      if (DVDThread::GetDiscType() == DiscIO::Platform::Triforce)
+      {
+        if (dvd_offset & 0x80000000)
```
lioncash, line 876:
```suggestion
        if ((dvd_offset & 0x80000000) != 0)
```

Source/Core/Core/HW/EXI/EXI_DeviceBaseboard.h
Zopolis4: Oops, not sure how I let that get carried over.

ReNardier:
i might be wrong, but since this is something that was a thing in some really old versions of dolphin, shouldn't this be called "Reimplement preliminary Triforce support" instead of simply "Implement preliminary Triforce support"?

Zopolis4:
> i might be wrong, but since this is something that was a thing in some really old versions of dolphin, shouldn't this be called "Reimplement preliminary Triforce support" instead of simply "Implement preliminary Triforce support"?

I thought about it, but it implements it in quite a different way, and its not a reimplementation of **preliminary** support per se. Besides, the code dates back to the google code era of dolphin, so I don't think its much of a reimplementation

JosJuice:
> i might be wrong, but since this is something that was a thing in some really old versions of dolphin, shouldn't this be called "Reimplement preliminary Triforce support" instead of simply "Implement preliminary Triforce support"?

I don't think it matters, personally.

Source/Core/Core/CMakeLists.txt
```diff
@@ -181,6 +181,8 @@ add_library(core
   HW/EXI/EXI_DeviceAD16.h
   HW/EXI/EXI_DeviceAGP.cpp
   HW/EXI/EXI_DeviceAGP.h
+	HW/EXI/EXI_DeviceBaseboard.cpp
```
mzji, line 184:
```suggestion
  HW/EXI/EXI_DeviceBaseboard.cpp
```

Dentomologist:
@dolphin-emu-bot rebuild

Pokechu22:
@dolphin-emu-bot rebuild

Pokechu22:
@dolphin-emu-bot rebuild

Pokechu22:
@dolphin-emu-bot rebuild

Pokechu22:
@dolphin-emu-bot rebuild

Zopolis4:
I'm not going to make any more feature commits to this PR, so are there any requested changes blocking this from getting merged?

Zopolis4:
I'm working on a few changes to the DiscIO code, so consider this PR a WIP again until they're done.

Zopolis4:
Ok, changes finished.

Source/Core/DiscIO/VolumeDisc.cpp
```diff
@@ -95,6 +95,15 @@ bool VolumeDisc::IsNKit() const
   return ReadSwapped<u32>(0x200, PARTITION_NONE) == NKIT_MAGIC;
 }
 
+bool VolumeDisc::IsTriforceGame() const
+{
+  constexpr u32 BTID_MAGIC = 0x42544944;  // "BTID"
+  return (ReadSwapped<u32>(0x00080000, PARTITION_NONE) == BTID_MAGIC) ||
```
GXTX, line 101: Could you explain these hardset read offsets?

Zopolis4: Those are the locations of the boot.id files in all the triforce games. Previously, it checked using the filesystem, but that broke when I tried to merge it into VolumeGC. This way is also cleaner and shorter, if slightly less ideal.

Source/Core/DiscIO/Enums.h
Rumi-Larry: I think the specifier for the Gamecube doesn't hurt, but for Triforce is different

Zopolis4: I mean the thing is the images dolphin is using arent the original disc dumps, theyve been extracted and decrypted from GD-ROMs to gamecube images, and the same thing would happen with the nand carts.

Data/Sys/GameSettings/GVSJ8P.ini
```diff
@@ -1,5 +1,4 @@
-# GVS46E, GVS46J - Virtua Striker 4 ver. 2006
-# Because Triforce games have weird IDs, properties are inherited from GVS.ini (Virtua Striker 3 ver. 2002)!
+# GVSJ8P, Virtua Striker 2002 (GC), Virtua Striker 2002 (Triforce), Virtua Striker 4, Virtua Striker 4 ver. 2006
```
PatrickFerry, line 1: You have deleted the GVS.ini but added the GC title here? I suspect you were to rename this file to GVS.ini

The title comment should read like this after you've renamed this file
```suggestion
# GVSP8P, GVSE8P, GVSJ8P - Virtua Striker 2002 (GC), Virtua Striker 2002 (Triforce), Virtua Striker 4, Virtua Striker 4 ver. 2006
```

Zopolis4: The current GVS.ini is for the triforce version regardless, at least according to the name, which is why I deleted it.

PatrickFerry: A 3 letter GameID will apply to all GameIDs who's first 3 letters match it. So the GVS.ini applies to these Triforce games and all regions of the GameCube Virtua Striker 2002.

GVSJ8P.ini will only apply to these Triforce games and the Japanese release of GameCube title Virtua Striker 2002 called "Virtual Striker 3 Ver. 2002" in that region.

**The problem is FPRF won't be set for either the NTSC or PAL versions of Virtua Striker 2002 if the only GameID you've left is GVSJ8P.ini**

PatrickFerry: Is that intentional? Is it only the Triforce Virtua Striker requiring FPRF?

Zopolis4: I have no idea, these are really old .ini files.

JosJuice: In that case, you should keep GVS.ini.

JMC47:
@dolphin-emu-bot rebuild

Pokechu22:
@dolphin-emu-bot rebuild

Dentomologist:
@dolphin-emu-bot rebuild

JosJuice:
@dolphin-emu-bot rebuild

JosJuice:
@dolphin-emu-bot rebuild

Source/Core/DiscIO/VolumeGC.cpp
```diff
@@ -40,6 +41,17 @@ VolumeGC::VolumeGC(std::unique_ptr<BlobReader> reader) : m_reader(std::move(read
   };
 
   m_converted_banner = [this] { return LoadBannerFile(); };
+
+  constexpr u32 BTID_MAGIC = 0x44495442;
+  BootID triforce_header;
+  auto tmp_fs = std::make_unique<FileSystemGCWii>(this, PARTITION_NONE);
```
JosJuice, line 47: You should use `GetFileSystem` here instead of manually creating a `FileSystemGCWii`, to ensure that the work done here can be reused later.

Source/Core/DiscIO/VolumeGC.cpp
```diff
@@ -40,6 +41,17 @@ VolumeGC::VolumeGC(std::unique_ptr<BlobReader> reader) : m_reader(std::move(read
...
+  if (tmp_fs)
+  {
+    const u64 file_size = ReadFile(*this, PARTITION_NONE, tmp_fs->FindFileInfo("boot.id").get(),
```
JosJuice, line 50: Missing `nullptr` check for `FindFileInfo` return value.

Source/Core/Core/HW/EXI/EXI_DeviceBaseboard.cpp
```diff
@@ -0,0 +1,55 @@
...
+  if (cs)
+    m_position = 0;
+}
+
+bool CEXIBaseboard::IsPresent() const
+{
+  return true;
+}
+
+void CEXIBaseboard::TransferByte(u8& byte)
```
JosJuice, line 26: Why `&`?

Source/Core/Core/HW/DVD/DVDInterface.cpp
```diff
@@ -854,15 +861,49 @@ void ExecuteCommand(ReplyType reply_type)
...
+          // Firmware status (1)
+          case 0x80000120:
+            Memory::Memset(s_DIMAR, 0x01, s_DICMDBUF[2]);
+            break;
+          // Firmware status (2)
+          case 0x80000140:
+            Memory::Memset(s_DIMAR, 0x01, s_DICMDBUF[2]);
+            break;
+          default:
+            PanicAlertFmtT("Unknown Media Board Read");
```
JosJuice, line 902:
```suggestion
            PanicAlertFmtT("Unknown Media Board Read");
```

Neither users nor translators will be able to make much sense of this string.

JosJuice:
What is the copyright status of the added image files?

Source/Core/Core/HW/EXI/EXI_DeviceBaseboard.cpp
JosJuice: It should be updated.

Zopolis4:
> What is the copyright status of the added image files?

I believe they fall under fair use, but they are copyrighted by Nintendo (and possibly the other companies involved). They're a higher resolution image of what displays when you boot the Triforce.

Source/Core/Core/HW/EXI/EXI_DeviceBaseboard.cpp
JosJuice: Nevermind, this is an overridden function. It'll have to be that way then.

Source/Core/Core/HW/DVD/DVDInterface.cpp
Zopolis4: Fair enough, although I don't think it should be a panic alert now that I think about it.

Source/Core/DiscIO/VolumeGC.cpp
Zopolis4: Not really sure how to implement that, would something like this work?

    
    if (FindFileInfo("boot.id").get() != nullptr && file_size >= 4 && triforce_magic == BTID_MAGIC)
      m_is_triforce = true;`

JosJuice: Store the result of FindFileInfo in a variable, then check the variable against nullptr, then pass the variable to ReadFile.

Source/Core/DiscIO/VolumeGC.h
```diff
@@ -103,6 +103,9 @@ class VolumeGC : public VolumeDisc
   Common::Lazy<std::unique_ptr<FileSystem>> m_file_system;
 
   std::unique_ptr<BlobReader> m_reader;
+
+  u32 triforce_magic; // "BTID"
```
JosJuice, line 107: This should be a local variable instead.

Source/Core/Core/HW/DVD/DVDInterface.cpp
```diff
@@ -814,6 +814,13 @@ void ExecuteCommand(ReplyType reply_type)
   DIInterruptType interrupt_type = DIInterruptType::TCINT;
   bool command_handled_by_thread = false;
 
+  // Swap's endian of Triforce DI commands, and zeroes out random bytes to prevent unknown read
```
JosJuice, line 817:
```suggestion
  // Swaps endian of Triforce DI commands, and zeroes out random bytes to prevent unknown read
```

Source/Core/DiscIO/VolumeGC.cpp
```diff
@@ -40,6 +41,19 @@ VolumeGC::VolumeGC(std::unique_ptr<BlobReader> reader) : m_reader(std::move(read
...
   m_converted_banner = [this] { return LoadBannerFile(); };
+
+  constexpr u32 BTID_MAGIC = 0x44495442;
+  auto tmp_fs = GetFileSystem(PARTITION_NONE);
+  if (tmp_fs)
+  {
+    std::unique_ptr<FileInfo> file_info = tmp_fs->FindFileInfo("boot.id");
+    if (!file_info)
+      return;
+    const u64 file_size = ReadFile(*this, PARTITION_NONE, tmp_fs->FindFileInfo("boot.id").get(),
```
JosJuice, line 52:
```suggestion
    const u64 file_size = ReadFile(*this, PARTITION_NONE, file_info.get(),
```

JosJuice:
@dolphin-emu-bot rebuild

JosJuice:
@dolphin-emu-bot rebuild

Source/Core/DiscIO/VolumeGC.h
Zopolis4: Note-- I've fixed this, but I use the struct in pt.2 [here](https://github.com/dolphin-emu/dolphin/pull/10207/files#diff-c817905baf4f25b70eee8d32352e8db45de47d3a5432eaa281bbc805dc25fe6cR80-R85), and that'll be quite a bit harder to work around. Should I keep it like this and try and find a way to rewrite the other code, or should i move this back to a struct, seeing as it'll be used in the next pr.

JosJuice:
Because you've added an entry in the middle of the Platform enum, you must make a change to the Android code, otherwise the Android enum will be out of sync with the C++ enum and Wii disc games will start showing up as digital Wii games on Android. The simplest adjustment that can be made is the following (preferably squash this into the commit that changes the Platform enum):

```
index f561feb250..3a3a9716d6 100644
--- a/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/ui/platform/Platform.java
+++ b/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/ui/platform/Platform.java
@@ -31,6 +31,9 @@ public enum Platform

   public static Platform fromNativeInt(int i)
   {
+    // TODO: Proper support for Triforce games
+    if (i >= 1)
+      i--;
     // TODO: Proper support for DOL and ELF files
     boolean in_range = i >= 0 && i < values().length;
     return values()[in_range ? i : WIIWARE.value];
```

I think it's best to wait with adding more proper support for the Triforce platform to the Android code. Android's game list is split into three parts based on the platform, and adding Triforce there would be a very user-visible change... It's best to not do that before games actually can run.

Zopolis4:
So are you saying that I should implement the code you posted or scrap all of the android code until games can run?

JosJuice:
> So are you saying that I should implement the code you posted or scrap all of the android code until games can run?

That you should implement the code I posted.

> Although, unless android would create a default space for triforce games

If we were to follow the pattern of how the other platforms work, it would create a default space for Triforce games. That's why my suggestion is to just add those two lines instead of adding proper support for the Triforce platform on Android.

Zopolis4:
Fair enough.

Source/Core/Core/HW/DVD/DVDInterface.cpp
```diff
@@ -863,15 +870,49 @@ void ExecuteCommand(ReplyType reply_type)
...
+          // Firmware status (2)
+          case 0x80000140:
+            Memory::Memset(s_DIMAR, 0x01, s_DICMDBUF[2]);
+            break;
+          default:
+            ERROR_LOG_FMT(DVDINTERFACE, "Unknown Media Board Read");
+            break;
+          }
+        }
+      }
```
shuffle2, line 915: in triforce case, it will still `ExecuteReadCommand`, is that intentional?

Source/Core/Core/HW/DVD/DVDInterface.cpp
```diff
@@ -863,15 +870,49 @@ void ExecuteCommand(ReplyType reply_type)
...
-
+      if (DVDThread::GetDiscType() == DiscIO::Platform::Triforce)
+      {
+        if ((dvd_offset & 0x80000000) != 0)
+        {
+          switch (dvd_offset)
+          {
+          // Media board status (1)
+          case 0x80000000:
+            Memory::Memset(s_DIMAR, 0, s_DICMDBUF[2]);
```
shuffle2, line 888: should `s_DICMDBUF[2]` have any bounds/alignment applied throughout here?

Source/Core/Core/HW/DVD/DVDInterface.cpp
Zopolis4: Yes-- that area of code is just to deal with some triforce-specific commands, triforce games still use many regular DI commands.

shuffle2: i mean that even if the command is handled in one of the triforce-specific switch cases, it will still try to `ExecuteReadCommand`

Zopolis4: It'll still try to `ExecuteReadCommand` that same command?

shuffle2: right - is that intentional?

Zopolis4: No

Source/Core/Core/HW/DVD/DVDInterface.cpp
Zopolis4: What do you mean?

shuffle2: `s_DICMDBUF[2]` is a `u32` that can be fully controlled by software sending the cmd, so probably it is at least limited to some upper bound (either based on target buffer size or just some bitwise truncation). otherwise it could cause a 4GB write through `s_DIMAR` (and well, maybe that _does_ happen, idk how this device behaves).

Source/Core/Core/HW/DVD/DVDInterface.cpp
Zopolis4: Nevermind, that is intentional.

Zopolis4:
Note-- I deleted/moved to pt.2 the media board commands because while they'd presumably work as they're supposed to, no games reach them on this branch, and I should hardware test them as well. That, and something about the old implementation smells off.

sepalani (changes requested):

Source/Core/Core/HW/EXI/EXI_DeviceBaseboard.cpp
```diff
@@ -0,0 +1,55 @@
+// Copyright 2013 Dolphin Emulator Project
```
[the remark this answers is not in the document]
Zopolis4: The copyright is correct, this is derived from an older file from the old implementation.

Source/Core/Core/HW/EXI/EXI_DeviceBaseboard.h
```diff
@@ -0,0 +1,27 @@
...
+  void SetCS(int CS) override;
+  bool IsPresent() const override;
+  void DoState(PointerWrap& p) override;
+
+private:
+  // STATE_TO_SAVE
+  u32 m_position = 0;
+  u32 m_command = 0;
+
+  void TransferByte(u8& byte) override;
```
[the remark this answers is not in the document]
Zopolis4: Yeah all of the EXI code uses `TransferByte(u8& byte)`

Zopolis4:
I've added two commits, would people mind reviewing them, `Implement Triforce ID parsing` especially?

naari3:
will this PR and #10207 not be reviewed? I want to do dynamic analysis of Mariokart Arcade GP 2 and want to use external tools, but they did not work on the 4 series. Therefore, I want to use a relatively new version of Dolphin, but it is sad that this great PR has been inactive for so long.

Zopolis4:
I am willing to rebase this PR if people have a use for it, it has been inactive because I have finished all the features I am going to implement on this branch and there does not seem to be a merge in sight anytime soon.

naari3:
@Zopolis4 It would be great if those PRs get merged! At least I would be happy.

naari3:
It's great to see that this PR conflict has been resolved and is now reviewable again, great work!

JMC47:
I would suggest changing the PR name to something more accurate.  This isn't really triforce support in the sense that users would see it as.  I think I would prefer something along the lines of "properly identify TriForce games" or "differentiate TriForce games from GameCube games"

I do like that this change stops them from crashing.  Most of the changed lines seem to be with INI deletions and some upkeep that I don't see breaking anything.

Data/Sys/ApprovedInis.json
```diff
@@ -932,10 +932,6 @@
     "title": "Region Select",
     "AD12237401ABE9FE4A545AADB5C5AE10355E2076": "RSAPatch"
   },
-  "RELJAB": {
-    "title": "SegaBoot",
-    "130F3594CAB57B85616F95C7126F4748AAC5867D": "DI Seed Blanker"
-  },
```
JosJuice, line 938: If you update this file, you also need to update `APPROVED_LIST_HASH`, otherwise RetroAchievements will break:

https://github.com/dolphin-emu/dolphin/blob/f8bf35e6f01e0165e2877031bf19cd17e71e61e2/Source/Core/Core/AchievementManager.h#L82-L84

Ideally we should have a unit test to catch this, but we currently don't.

Source/Core/Core/TitleDatabase.cpp
```diff
@@ -56,6 +56,9 @@ void TitleDatabase::AddLazyMap(DiscIO::Language language, const std::string& lan
   m_title_maps[language] = [language_code]() -> Map {
     return LoadMap(File::GetSysDirectory() + "wiitdb-" + language_code + ".txt");
   };
+  m_trititle_maps[language] = [language_code]() -> Map {
+    return LoadMap(File::GetSysDirectory() + "tritdb-" + language_code + ".txt");
```
JosJuice, line 60: Any particular reason why Triforce has been shortened to "tri" in "tritdb" and "tri_id"? Normally we don't use these kinds of abbreviations. Especially the file name might be difficult to change later.

Source/Core/Core/TitleDatabase.cpp
```diff
@@ -91,8 +94,32 @@ TitleDatabase::TitleDatabase()
 TitleDatabase::~TitleDatabase() = default;
 
 const std::string& TitleDatabase::GetTitleName(const std::string& gametdb_id,
+                                               const std::string& tri_id,
                                                DiscIO::Language language) const
 {
+  if (tri_id != "" || NULL)
```
JosJuice, line 100: If you want this code to not run, please use `ifdef`, or use `false` instead of `NULL`.

Source/Core/DiscIO/VolumeGC.cpp
```diff
@@ -77,6 +95,18 @@ std::string VolumeGC::GetGameTDBID(const Partition& partition) const
...
+std::string VolumeGC::GetTriID() const
+{
+  if (m_is_triforce)
+  {
+    return (std::string(m_tri_id.data(), m_tri_id.size()));
+  }
+  else
+  {
+    return "";
+  }
```
JosJuice, line 107:
```suggestion
  if (m_is_triforce)
    return std::string(m_tri_id.data(), m_tri_id.size());
  else
    return "";
```

Source/Core/Core/TitleDatabase.cpp
Zopolis4: At the time I think I did it because it lined up neatly with `wiitdb`, but I'm happy to change it if you'd prefer the non-abbreviated version.

Source/Core/DiscIO/VolumeGC.cpp
Zopolis4: Fixed.

Source/Core/Core/TitleDatabase.cpp
Zopolis4: not as far as i can tell, no.

fixed.

Source/Core/Core/TitleDatabase.cpp
JosJuice: Yes, I would prefer it non-abbreviated.

Data/Sys/ApprovedInis.json
Zopolis4: Fixed.

Source/Core/Core/TitleDatabase.cpp
Zopolis4: Fixed.

JosJuice (approved):

JMC47:
The requested change from Sepalani is from 2022 and has been addressed.  This looks good to be merged after a quick test to make sure normal games are booting fine.

PR #10085, opened 2021-09-04
Fix all uninitialized variable warnings (C26495)

Pokechu22:
This fixes all occurrences of [C26495](https://docs.microsoft.com/en-us/cpp/code-quality/c26495?view=msvc-160) within Dolphin code, by adding initializers for everything.  I have mixed feelings about this approach, but it seems like there isn't any simpler approach (a default constructor won't work for instance).

See https://github.com/dolphin-emu/dolphin/pull/9956#issuecomment-911211718 for an example of where a value not being initialized caused problems.


Source/Core/Core/BootManager.cpp
```diff
@@ -64,42 +64,42 @@ struct ConfigCache
...
+  bool bFPRF = false;
+  bool bAccurateNaNs = false;
+  bool bMMU = false;
+  bool bLowDCBZHack = false;
+  bool bDisableICache = false;
+  bool m_EnableJIT = false;
+  bool bSyncGPU = false;
+  int iSyncGpuMaxDistance = false;
+  int iSyncGpuMinDistance = false;
+  float fSyncGpuOverclock = false;
```
sepalani, line 87: Shouldn't this be initialised with `0.0f` instead of false?

sepalani (commented):
Some remarks:
 - Is there a reason to initialise some integers/bool/float types with `{}` where as in other places it uses an assignation `=`?
 - Can't braces be used everywhere to keep it consistent?
 - Should float be initialised with `0.0f` rather than `0` or `false`?

Source/Core/Core/DSP/DSPCore.h
```diff
@@ -279,10 +279,10 @@ struct DSP_Regs
 struct DSPInitOptions
 {
   // DSP IROM blob, which is where the DSP boots from. Embedded into the DSP.
-  std::array<u16, DSP_IROM_SIZE> irom_contents;
+  std::array<u16, DSP_IROM_SIZE> irom_contents = {};
 
   // DSP DROM blob, which contains resampling coefficients.
-  std::array<u16, DSP_COEF_SIZE> coef_contents;
+  std::array<u16, DSP_COEF_SIZE> coef_contents = {};
```
BhaaLseN, line 285: Those two seem to be outliers, other code uses `std::array{}` while this uses `std::array = {}`. I think we should be consistent, if the compiler lets us.

Source/Core/Core/DSP/Jit/x64/DSPJitRegCache.h
```diff
@@ -170,8 +170,8 @@ class DSPJitRegCache
   void MovToMemory(size_t reg);
   void FlushMemBackedRegs();
 
-  std::array<DynamicReg, 37> m_regs;
-  std::array<X64CachedReg, 16> m_xregs;
+  std::array<DynamicReg, 37> m_regs = {};
+  std::array<X64CachedReg, 16> m_xregs = {};
```
BhaaLseN, line 174: Same down here.

Source/Core/Core/DSP/Jit/x64/DSPJitRegCache.cpp
```diff
@@ -194,7 +194,7 @@ DSPJitRegCache::DSPJitRegCache(DSPEmitter& emitter)
 
 DSPJitRegCache::DSPJitRegCache(const DSPJitRegCache& cache)
     : m_regs(cache.m_regs), m_xregs(cache.m_xregs), m_emitter(cache.m_emitter),
-      m_is_temporary(true), m_is_merged(false)
+      m_is_temporary(true), m_is_merged(false), m_use_ctr(0)
```
BhaaLseN, line 197: Thoughts on moving everything that uses a constant value into the header (with direct initialization)?

Source/Core/Core/Debugger/Dump.cpp
```diff
@@ -9,7 +9,7 @@
 #include "Common/CommonTypes.h"
 #include "Common/IOFile.h"
 
-CDump::CDump(const std::string& filename) : m_pData(nullptr)
+CDump::CDump(const std::string& filename) : m_size(0), m_pData(nullptr)
```
BhaaLseN, line 12: Same here for members that are initialized to constant values

Source/Core/Core/FifoPlayer/FifoDataFile.h
```diff
@@ -93,13 +94,13 @@ class FifoDataFile
...
-  u32 m_XFMem[XF_MEM_SIZE];
-  u32 m_XFRegs[XF_REGS_SIZE];
-  u8 m_TexMem[TEX_MEM_SIZE];
-  u32 m_ram_size_real;
-  u32 m_exram_size_real;
+  std::array<u32, BP_MEM_SIZE> m_BPMem;
+  std::array<u32, CP_MEM_SIZE> m_CPMem;
+  std::array<u32, XF_MEM_SIZE> m_XFMem;
+  std::array<u32, XF_REGS_SIZE> m_XFRegs;
+  std::array<u8, TEX_MEM_SIZE> m_TexMem;
```
BhaaLseN, line 101: Do we also want those to be initialized?

Source/Core/Core/HW/WiiSaveStructs.h
```diff
@@ -91,8 +91,8 @@ class Storage
       File = 1,
       Directory = 2,
     };
-    u8 mode, attributes;
-    Type type;
+    u8 mode = 0, attributes = 0;
```
BhaaLseN, line 94: Would you mind splitting those two into two lines?

Source/Core/Core/PowerPC/MMU.h
```diff
@@ -49,7 +49,15 @@ struct TryReadResult
   // the actual value that was read
   T value;
 
+#ifdef _MSC_VER
+#pragma warning(push)
+// Disable warning for uninitialized member variables
```
BhaaLseN, line 54: Can you add a reason _why_ this warning is temporarily disabled for this ctor? (Or: simply address it, so the suppression isn't needed)

Source/Core/Core/PowerPC/PPCAnalyst.h
```diff
@@ -138,23 +138,23 @@ using CodeBuffer = std::vector<CodeOp>;
...
-  u32 m_num_instructions;
+  u32 m_num_instructions = 0;
 
   // Some basic statistics about the block.
-  BlockStats* m_stats;
+  BlockStats* m_stats = nullptr;
 
   // Register statistics about the block.
-  BlockRegStats *m_gpa, *m_fpa;
+  BlockRegStats *m_gpa = nullptr, *m_fpa = nullptr;
```
BhaaLseN, line 151: Can we break those up into two lines?

Source/Core/VideoCommon/GXPipelineTypes.h
```diff
@@ -35,10 +35,18 @@ struct GXPipelineUid
   // and this map lookup can happen every draw call. However, as using memcmp() will also compare
   // any padding bytes, we have to ensure these are zeroed out.
   GXPipelineUid() { std::memset(static_cast<void*>(this), 0, sizeof(*this)); }
+#ifdef _MSC_VER
+#pragma warning(push)
+// Disable warning for uninitialized member variables
```
BhaaLseN, line 40: Can you add a reason to this comment, or address it so it isn't necessary?
My guess is that VS doesn't recognize the `memcpy` as initialization?

Source/Core/VideoCommon/TextureCacheBase.h
```diff
@@ -89,28 +89,28 @@ class TextureCacheBase
...
-    u64 id;
+    u64 id = 0;
 
     bool reference_changed = false;  // used by xfb to determine when a reference xfb changed
 
-    unsigned int native_width,
-        native_height;  // Texture dimensions from the GameCube's point of view
-    unsigned int native_levels;
+    // Texture dimensions from the GameCube's point of view
+    unsigned int native_width = 0, native_height = 0;
```
BhaaLseN, line 112: Can we break those up into two lines?

BhaaLseN (commented):
Changes seem mostly fine, but I'm a bit worried about consistency. Some places use assignments, others use uniform initialization for the same thing; and mixing those two all over the place may lead to more inconsistencies down the road (depending on what file one looks at for inspiration).
Then again, this is likely something we should first address in the code style docs before we force one over the other on a whim.

I do wonder about those `int i = false`, `float f = false` etc. though that sepalani pointed out, seems like copy&paste gone wrong.

Source/Core/Core/PowerPC/MMU.h
Pokechu22: It's because this class is kinda like `std::optional`.  Probably a better solution would be to replace it with `std::optional` to guarantee nobody can touch the other fields.

Source/Core/VideoCommon/GXPipelineTypes.h
Pokechu22: Yep, it's because it doesn't recognize memcpy (but it _does_ recognize memset?)

Source/Core/Core/BootManager.cpp
Pokechu22: Yes, these are copy-paste errors.  I think initializing a float with `0` instead of `0.0f` is fine, but `false` is definitely silly.

Source/Core/Core/PowerPC/MMU.h
Pokechu22: I've replaced `TryReadResult` and `TryWriteResult` with `std::optional<ReadResult>` and `std::optional<WriteResult>`.

Source/Core/Core/DSP/Jit/x64/DSPJitRegCache.h
```diff
@@ -170,14 +170,14 @@ class DSPJitRegCache
...
 
-  std::array<DynamicReg, 37> m_regs;
-  std::array<X64CachedReg, 16> m_xregs;
+  std::array<DynamicReg, 37> m_regs{};
+  std::array<X64CachedReg, 16> m_xregs{};
 
   DSPEmitter& m_emitter;
-  bool m_is_temporary;
-  bool m_is_merged;
+  bool m_is_temporary = true;
```
sepalani, line 177: Depending on the constructor `m_is_temporary` is set to `true` or `false`. Won't setting it here produce side effect?

Source/Core/DolphinQt/Config/Mapping/IOWindow.cpp
```diff
@@ -207,7 +207,7 @@ class InputStateLineEdit : public QLineEdit
 
 private:
   std::function<ControlState()> m_state_evaluator;
-  bool m_should_paint_state_indicator;
+  bool m_should_paint_state_indicator = true;
```
sepalani, line 210: Shouldn't it be initialised to false?

Source/Core/Common/IOFile.h
```diff
@@ -57,6 +58,18 @@ class IOFile
...
   }
 
+  template <typename T, std::size_t N>
+  bool ReadArray(std::array<T, N>& elements, size_t* num_read = nullptr)
+  {
+    return ReadArray(elements.data(), elements.size(), num_read);
+  }
+
+  template <typename T, std::size_t N>
+  bool WriteArray(std::array<T, N>& elements)
```
sepalani, line 68: Can a const reference be used?

Source/Core/Core/DSP/Jit/x64/DSPJitRegCache.h
Pokechu22: Per [cppreference](https://en.cppreference.com/w/cpp/language/constructor), starting in C++11,

> If a non-static data member has a default member initializer and also appears in a member initializer list, then the member initializer is used and the default member initializer is ignored

So, it's fine, but maybe unnecessary.

Source/Core/DolphinQt/Config/Mapping/IOWindow.cpp
Pokechu22: I'm not sure.  I based it on this code:

https://github.com/dolphin-emu/dolphin/blob/4b8b53ac732645f6c0fd3f4b11733cb3d86c9c41/Source/Core/DolphinQt/Config/Mapping/IOWindow.cpp#L670-L689

which sets it to `true` in the default case.  But maybe `false` would be better since the default should be empty.

Source/Core/VideoCommon/VideoConfig.h
```diff
@@ -58,173 +58,173 @@ struct VideoConfig final
...
 
   // General
-  bool bVSync;
-  bool bVSyncActive;
-  bool bWidescreenHack;
-  AspectMode aspect_mode;
-  AspectMode suggested_aspect_mode;
-  bool bCrop;  // Aspect ratio controls.
-  bool bShaderCache;
+  bool bVSync{};
```
iwubcode, line 61: Throughout most of the files you used `bool x = false` or `int x = 0`.  Here you use `bool x{}` and `int x{}`.  Can we make this consistent?

Source/Core/VideoCommon/ShaderCache.h
```diff
@@ -211,16 +211,16 @@ class ShaderCache final
...
     };
     std::map<Uid, Shader> shader_map;
     LinearDiskCache<Uid, u8> disk_cache;
   };
-  ShaderModuleCache<VertexShaderUid> m_vs_cache;
-  ShaderModuleCache<GeometryShaderUid> m_gs_cache;
-  ShaderModuleCache<PixelShaderUid> m_ps_cache;
-  ShaderModuleCache<UberShader::VertexShaderUid> m_uber_vs_cache;
-  ShaderModuleCache<UberShader::PixelShaderUid> m_uber_ps_cache;
+  ShaderModuleCache<VertexShaderUid> m_vs_cache{};
```
iwubcode, line 219: If you move `m_api_type` value to the header.  I think you can then remove or default `ShaderCache()`.  Then the `{}` for all of these 'complex types' can probably go away?

Pokechu22: They aren't needed even without it; I'll remove it.  I think I added them before the `bool pending = false;` in `ShaderModuleCache::Shader`.  (Note that it takes quite some time for the analysis to run, and the analysis doesn't always include all of the warnings if there's more than one for a given file, so I may be initializing things that didn't need to be explicitly initialized).

Source/Core/VideoCommon/VideoConfig.h
Pokechu22: I've changed them to `= false` or `= 0`.

Source/Core/AudioCommon/OpenALStream.h
```diff
@@ -68,9 +68,9 @@ class OpenALStream final : public SoundStream
   Common::Flag m_run_thread;
 
   std::vector<short> m_realtime_buffer;
-  std::array<ALuint, OAL_BUFFERS> m_buffers;
-  ALuint m_source;
-  ALfloat m_volume;
+  std::array<ALuint, OAL_BUFFERS> m_buffers{};
+  ALuint m_source = 0;
+  ALfloat m_volume = 1;
```
iwubcode, line 73: I know nothing about this system.  Is 1 intentional here?

Pokechu22: It's used here and here:

https://github.com/dolphin-emu/dolphin/blob/05f86b9f5b6e7f65aaefcc29e348cc2628bed34f/Source/Core/AudioCommon/OpenALStream.cpp#L259-L260

https://github.com/dolphin-emu/dolphin/blob/05f86b9f5b6e7f65aaefcc29e348cc2628bed34f/Source/Core/AudioCommon/OpenALStream.cpp#L147-L153

And the docs say 1.0 leaves volume unchanged:

https://github.com/dolphin-emu/dolphin/blob/05f86b9f5b6e7f65aaefcc29e348cc2628bed34f/Externals/OpenAL/include/al.h#L184-L196

Source/Core/Core/PowerPC/MMU.h
```diff
@@ -37,49 +37,38 @@ u32 HostRead_Instruction(u32 address);
...
-  TryReadResult(bool translated_, T&& value_)
-      : success(true), translated(translated_), value(std::move(value_))
-  {
-  }
-  TryReadResult(bool translated_, const T& value_)
-      : success(true), translated(translated_), value(value_)
-  {
-  }
-  explicit operator bool() const { return success; }
+  ReadResult(bool translated_, T&& value_) : translated(translated_), value(std::move(value_)) {}
```
iwubcode, line 49: I may be wrong here.  But I think since this is a `T&&` this line should be:

`ReadResult(bool translated_, T&& value_) : translated(translated_), value(std::forward<T>(value_)) {}`

The reason is T&& can be either an lvalue or an rvalue.  But move is only applicable on rvalues.

(I know you didn't write this code originally but just thought I'd mention it)

Source/Core/Core/PowerPC/MMU.h
```diff
@@ -93,36 +82,37 @@ void HostWrite_U64(u64 var, u32 address);
...
-  bool success;
-
   // whether the address had to be translated (given address was treated as virtual) or not (given
   // address was treated as physical)
   bool translated;
 
-  TryWriteResult() : success(false) {}
-  TryWriteResult(bool translated_) : success(true), translated(translated_) {}
-  explicit operator bool() const { return success; }
+  WriteResult(bool translated_) : translated(translated_) {}
```
iwubcode, line 91: Surprised this wasn't `explicit`

Source/Core/Core/PowerPC/MMU.h
Pokechu22: Changed.

Source/Core/Core/PowerPC/MMU.h
Pokechu22: Changed.

iwubcode (approved):

sepalani (commented):
> Reviewed 1 of 2 files at r1, 110 of 110 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/dolphin-emu/dolphin/10085)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Pokechu22)

sepalani (commented):
I don't think making this PR stale for too long is a good idea because it will be a pain to maintain.

PR tested on Windows debug build with some games. LGTM besides these 2 nitpicks.

delroth:
@dolphin-emu-bot rebuild

Source/Core/Common/IOFile.h
```diff
@@ -57,6 +58,18 @@ class IOFile
     return m_good;
   }
 
+  template <typename T, std::size_t N>
+  bool ReadArray(std::array<T, N>& elements, size_t* num_read = nullptr)
```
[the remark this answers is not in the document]
Pokechu22: I was mostly modeling it around `ChunkFile` (the magic code used for save-states), but probably it's better to not do that.  `ChunkFile` _needs_ to be magic, but IOFile doesn't.

Pokechu22:
Rebased and updated.  I also had to replace `TryReadResult` with `std::optional` in CheatSearch since #9886 was merged; that commit should be re-reviewed.

sepalani (approved):
LGTM

Source/Core/VideoCommon/VideoConfig.h
```diff
@@ -58,173 +58,174 @@ struct VideoConfig final
...
+  bool bEFBEmulateFormatChanges = false;
+  bool bSkipEFBCopyToRam = false;
+  bool bSkipXFBCopyToRam = false;
+  bool bDisableCopyToVRAM = false;
+  bool bDeferEFBCopies = false;
+  bool bImmediateXFB = false;
+  bool bSkipPresentingDuplicateXFBs = false;
+  bool bCopyEFBScaled = false;
+  int iSafeTextureCache_ColorSamples = 0;
   float fAspectRatioHackW, fAspectRatioHackH;
```
BhaaLseN, line 130: Did we miss those two, or are they set elsewhere?

Plus, it would probably be a good idea to make this into two lines while we're at it.

Pokechu22: They're set in the constructor.  But so are a few other things which I did initialize - I'll move those values out of the constructor.

Pokechu22:
Note: The fifoci comment is because for some reason it compared with 5.0-15257 even though this PR was based on 5.0-15259 (#10138).  The [more recent comparision](https://fifo.ci/version/23e81202162c52ee9854feb9b72c3aa0424fc793/) has no differences (though the same things still fail to render, which is expected), but fifoci doesn't delete old comments if there are no changes.

shuffle2:
@Pokechu22  I just looked for uninit vars the same way again. Among a few we don't really care about (but probably should just initialize and let the compiler deal with dead store elimination), there's also: 
```
Severity	Code	Description	Project	File	Line	Suppression State	Detail Description
Warning	C6001	Using uninitialized memory 'scanline'.	DolphinLib	C:\src\dolphin\Source\Core\VideoBackends\Software\EfbInterface.cpp	615	
```
image


It has correctly detected that if `source_rect.left >= source_rect.right`, then body of `for (int i = 1, x = left; x < right; i++, x++)` won't be executed, and `scanline` won't be initialized. Is that a condition we know never can happen (particularly the equals case)? Should something be done about it?

Pokechu22:
It's impossible for `left >= right` since the copy size is specified as a width/height (and those are unsigned, with a value of 0 indicating 1 pixel).  See this code (which also has a rather dubious comment):

https://github.com/dolphin-emu/dolphin/blob/e62771856092cd9bbc25c8c94c433ed0e55a9f3f/Source/Core/VideoCommon/BPStructs.cpp#L225-L228

So, even if a game for some reason decided that the XFB should be as thin as possible, it'd still render correctly (at least here; I don't know what hardware will actually do when you try to put a 1-pixel wide framebuffer to screen, but that'd be a VideoInterface problem I think).

shuffle2:
OK, thanks for pointing out caller.

unrelated, but interesting that their source is 10 bit integers...wonder what happens on 10bit overflow or so (which we wouldn't detect afaict) :p

Pokechu22:
Well, the max texture size is 1024 by 1024, which is the biggest size for a 10-bit value, but the EFB itself is 640 by 528, so a full-sized one would include out of bounds data.  I suspect that an overflow would just wrap around (that's roughly what happens without of bounds scissor values, though that also has wrapping at 2048 for other aspects), but I don't know what actually happens in the 640-1023 region (though apparently that sometimes _does_ happen with Mario Kart Wii).  Dolphin currently clamps it to stay within the 640 by 528 region, I guess leaving out-of-bounds memory untouched, so that overflow wouldn't cause any problems.

PR #10086, opened 2021-09-04
JitArm64: Fix W0 being present twice in register cache

JosJuice:
Fixes a regression from PR #7155.

The GPR `allocation_order` is initialized with only 28 elements, so the 29th element ends up getting zero initialized. Very sneaky bug...


AdmiralCurtiss:
Oh wow this is mean. I'm guessing this is the actual cause of the loadstore JIT bug?

JosJuice:
Nope! I found this when looking for that bug, but this does not seem to fix that bug.

lioncash (approved):

PR #10091, opened 2021-09-08
DolphinQt: Allow to set the GCI folder in the GameCube config tab

aldelaro5:
This allows to configure the path to the GCI folder to use in the settings.

I talked about this to @JMC47 before and they told me the setting was already present in the config backend, but the UI wasn't exposing it so anyone could implement it.

To note: I am not entirely sure about adding the `memcardfolder` bool here, if there's a better way to refactor this, lmk.

EDIT: converted to draft as this change is more complicated than anticipated, but I probably will still try unless @Techjar beats me to it


Techjar:
I don't think this is how we want this to work. It would make more sense to rename the setting and store the actual path here, rather than keeping the conditional logic at the usage.

jordan-woyak:
This PR seems dead so I'm going to close it for now. But we can re-open it upon request.

PR #10094, opened 2021-09-11
GameINI: Disable JIT branch following for Namco Museum.

AdmiralCurtiss:
See https://bugs.dolphin-emu.org/issues/12658 for details.

This restores the performance back to 5.0 levels, though it's still not 100% perfect. Probably the best we can do for now though.


JMC47:
Confirmed.

PR #7950, opened 2019-03-31
BPStructs: Gracefully handle out-of-range EFB copies

stenzek:
Since the copy X and Y coordinates/sizes are 10-bit, the game can configure a copy region up to 1024x1024. Hardware tests have found that the number of bytes written does not depend on the configured stride, instead it is based on the size registers, writing beyond the length  of a single row. The data written for the pixels which lie outside the EFB bounds does not wrap around instead returning different colors based on the pixel format of the EFB.

This suggests it's not based on coordinates, but instead on memory addresses. The effect of a within-bounds size but out-of-bounds offset (e.g. offset 320,0, size 640,480) are the same.

As it would be difficult to emulate the exact behavior of out-of-bounds reads, instead of writing the junk data, we don't write anything to RAM at all for over-sized copies, and clamp to the EFB borders for over-offset copies.

Dolphin now matches my hardware tests (apart from the open-bus-like behavior trashing the next rows of textures).

This was previously causing errors in some arcade virtual console games, which perform these out-of-range copies.

PR #10095, opened 2021-09-11
BPStructs: fix out-of-range EFB copy clamping

phire:
Previous code from #7950 only clamps correctly when the efb copies left and top coordinates are (0, 0)

Now we should handle all situations.

Spyro: A hero's tail is an example of a game that does an oversized EFB copy with a non-zero origin.

### DIFF
--- a/Source/Core/VideoCommon/BPStructs.cpp
+++ b/Source/Core/VideoCommon/BPStructs.cpp
@@ -249,8 +249,8 @@ static void BPWritten(const BPCmd& bp)
       srcRect.bottom = std::clamp<int>(srcRect.bottom, 0, EFB_HEIGHT);
     }
 
-    u32 copy_width = srcRect.GetWidth();
-    u32 copy_height = srcRect.GetHeight();
+    const u32 copy_width = srcRect.GetWidth();
+    const u32 copy_height = srcRect.GetHeight();
 
     // Check if we are to copy from the EFB or draw to the XFB
     const UPE_Copy PE_copy = bpmem.triggerEFBCopy;
--- a/Source/Core/VideoCommon/BPStructs.cpp
+++ b/Source/Core/VideoCommon/BPStructs.cpp
@@ -3,6 +3,7 @@
 
 #include "VideoCommon/BPStructs.h"
 
+#include <algorithm>
 #include <cmath>
 #include <cstring>
 #include <string>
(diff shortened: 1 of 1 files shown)


Source/Core/VideoCommon/BPStructs.cpp
```diff
@@ -239,22 +239,24 @@ static void BPWritten(const BPCmd& bp)
...
       // Adjust the copy size to fit within the EFB. So that we don't end up with a stretched image,
       // instead of clamping the source rectangle, we reduce it by the over-sized amount.
-      if (copy_width > EFB_WIDTH)
+      if (u32(srcRect.right) > EFB_WIDTH)
       {
-        srcRect.right -= copy_width - EFB_WIDTH;
-        copy_width = EFB_WIDTH;
+        u32 adjustment = srcRect.right - EFB_WIDTH;
+        srcRect.right -= adjustment;
+        copy_width -= adjustment;
```
iwubcode, line 253: I doubt this happens often (if at all?) but if we had a left position of say EFB_WIDTH - 5.  And a right position of EFB_WIDTH + 10 then this value will be negative which seems wrong.

Instead maybe we should recalculate:  `copy_width = srcRect.GetWidth()`

(same for height)

phire: I'm not sure you have done your math correctly. Plugging in the numbers you suggest, I've got:

    EFB_WIDTH = 640
    left = 635
    right = 650
    copy_width = 15
   
    adjustment = 650 - 640 = 10
    right = 650 - 10 = 640
    copy_width = 15 - 10 = 5

And width stays as a positive number.

iwubcode: Yeah, my apologies, I must have looked at this too quickly.  You are essentially just clamping the bounds and touching up the dimension.  Only time you'd get negative is if the whole rectangle was out of bounds.

Assuming I'm not having another crazy moment, I do think this would read a bit better:

```
srcRect.right = EFB_WIDTH;
copy_width = srcRect.GetWidth();
```

phire: You make a good point, I'm going to rewrite this PR to be easier to read.

phire: New version should be much simpler to understand.

Source/Core/VideoCommon/BPStructs.cpp
```diff
@@ -237,27 +238,20 @@ static void BPWritten(const BPCmd& bp)
...
-      {
-        srcRect.bottom -= copy_height - EFB_HEIGHT;
-        copy_height = EFB_HEIGHT;
-      }
+      // Clamp the copy region to fit within EFB. So that we don't end up with a stretched image.
+      srcRect.right = std::clamp<int>(srcRect.right, 0, EFB_WIDTH);
+      srcRect.bottom = std::clamp<int>(srcRect.bottom, 0, EFB_HEIGHT);
     }
 
+    u32 copy_width = srcRect.GetWidth();
```
iwubcode, line 252: Minor but these can now be `const`

iwubcode (approved):

PR #10097, opened 2021-09-13
JitArm64: Fix fresx/frsqrtex unlocking W0 too early

JosJuice:
`fpr.RW` may clobber GPRs that are not locked. Fixes https://bugs.dolphin-emu.org/issues/12656.


Pokechu22:
Also fixes https://bugs.dolphin-emu.org/issues/12665, apparently.

PR #10098, opened 2021-09-14
Common/FileUtil: Strip trailing path separator in ScanDirectoryTree().

AdmiralCurtiss:
Noticed this during debugging. This causes an inconsistency between the path returned in the 'root' directory node and the path returned in all child directory nodes.

Before:
before

Note how the `rootEntry.physicalName` ends in a trailing slash but all child directories do not, and how all child names have a `//` after `files`.

After:
after

I hope no one is relying on this behavoir somewhere...

### DIFF
--- a/Source/Core/Common/FileUtil.cpp
+++ b/Source/Core/Common/FileUtil.cpp
@@ -483,8 +483,16 @@ bool CreateEmptyFile(const std::string& filename)
 }
 
 // Recursive or non-recursive list of files and directories under directory.
-FSTEntry ScanDirectoryTree(const std::string& directory, bool recursive)
+FSTEntry ScanDirectoryTree(std::string directory, bool recursive)
 {
+#ifdef _WIN32
+  if (!directory.empty() && (directory.back() == '/' || directory.back() == '\\'))
+    directory.pop_back();
+#else
+  if (!directory.empty() && directory.back() == '/')
+    directory.pop_back();
+#endif
+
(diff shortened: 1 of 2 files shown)


AdmiralCurtiss:
Someone should check the behavior on Linux and Android too, I only tested Windows.

Source/Core/Common/FileUtil.cpp
```diff
@@ -483,8 +483,18 @@ bool CreateEmptyFile(const std::string& filename)
 }
 
 // Recursive or non-recursive list of files and directories under directory.
-FSTEntry ScanDirectoryTree(const std::string& directory, bool recursive)
+FSTEntry ScanDirectoryTree(const std::string& input_directory, bool recursive)
 {
+  std::string directory = input_directory;
```
JosJuice, line 488: If you always need to copy the string, it's better to not use a reference to begin with:

```suggestion
FSTEntry ScanDirectoryTree(std::string directory, bool recursive)
{
```

(The main advantage is when the argument is an xvalue – in that case this lets you skip the copy.)

AdmiralCurtiss: Fair point.

leoetlino (approved):

pssandhu:
This has fixed an issue I was having on Linux with the dynamic input textures feature not detecting a gameid.txt within the input textures folder. Thanks!

PR #10100, opened 2021-09-14
GLInterface: Remove unneeded wglShareLists call

Pokechu22:
When RenderDoc is attached, `wglShareLists` fails for some reason (see baldurk/renderdoc#2361).  `wglCreateContextAttribsARB` has a parameter for the share context, so there's no reason to use a separate `wglShareLists` call.

Note that using OpenGL in Dolphin with RenderDoc still has a separate issue with this patch where the screen appears black initially.  This has been fixed on RenderDoc's side, but it seems to be possible to work around it by starting a game/fifolog, then closing it, then starting it again.

This patch is from https://github.com/baldurk/renderdoc/issues/2361#issuecomment-919342173.


Pokechu22:
@dolphin-emu-bot rebuild

PR #10101, opened 2021-09-15
SI_DeviceKeyboard: Shorten MapKeys()

lioncash:
We can wrap this logic into a lambda to avoid copy-pasted checking of over 400+ lines of code.

### DIFF
--- a/Source/Core/Core/HW/SI/SI_DeviceKeyboard.cpp
+++ b/Source/Core/Core/HW/SI/SI_DeviceKeyboard.cpp
@@ -3,6 +3,7 @@
 
 #include "Core/HW/SI/SI_DeviceKeyboard.h"
 
+#include <array>
 #include <cstring>
 
 #include "Common/ChunkFile.h"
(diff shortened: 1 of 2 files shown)


Source/Core/Core/HW/SI/SI_DeviceKeyboard.cpp
```diff
@@ -100,520 +100,115 @@ void CSIDevice_Keyboard::DoState(PointerWrap& p)
...
-  }
-  if (key_status.key5x & KEYMASK_ENTER)
-  {
-    key[keys_held++] = KEY_ENTER;
-    if (keys_held >= MAX_KEYS_HELD)
-      return;
-  }
+      key[keys_held++] = keys[i];
+      if (keys_held >= MAX_KEYS_HELD)
+        return true;
```
leoetlino, line 188: Wouldn't it make more sense to return false here, since this is kind of like an error?

lioncash: I used true here to represent "should we stop reading further keys", since I think it's easier to misread

`if (!thing)` compared to `if (thing)`

I can change it though if it's preferable the other way.

leoetlino: ah, I guess that makes sense

leoetlino (approved):

PR #10102, opened 2021-09-15
NetPlayClient: Split out OnData() into smaller functions

lioncash:
Over time OnData() has become a huge function-long case statement that attempts to manage numerous packet-related behaviors, which makes it a little difficult to reliably ensure certain handling doesn't interfere with another case's. It's also mildly annoying to navigate due to its size.

To make it a little easier to read and find specific behavior, we can break the relevant pieces of code out into their own functions.

Almost all of this change is just code motion


leoetlino (approved):
> Reviewed 2 of 2 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/dolphin-emu/dolphin/10102)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @lioncash)

PR #10104, opened 2021-09-16
Cheats Manager UX improvements.

AdmiralCurtiss:
Fixes some annoyances pointed out by @Dentomologist back in #9886.

Specifically:

- The Cheats Manager option is no longer mysteriously greyed out if no game is running or if cheats are disabled and is instead always active.
- If no game is running the code tabs are greyed out instead of completely missing.
- If cheats are disabled and the "Warning: Cheats are disabled! [Configure Dolphin]" header thing shows (as the equivalent tabs in the game properties), the button now actually works -- it didn't before. Granted, it was rather hard to even reach this window in that situation (you'd have to start a game, open the window, close the game, then disable cheats without closing the window), but either way, it works correctly now.
- This also now allows running memory searches even if cheats are disabled, which is just a nice convenience thing.


iwubcode (approved):
Looks reasonable, untested.

leoetlino (approved):

PR #10106, opened 2021-09-17
BPStructs: Ignore malformed efb copies

phire:
Regression from #10095

Apparently Mario Kart Wii in wifi mode can generate a deformed EFB copy of size 4x4 at offset (328,1020) with stride 64.

This seems to be completely malformed. Lets just skip it all together.


JMC47:
Fixes an issue with Wiimmfi and Mario Kart Wii Online.

PR #10108, opened 2021-09-18
BPStructs: ensure side effects are same

phire:
I Missed this side effect when fixing crash in Mario Kart Wii wifi

Hopefully this fixes the bug in Paper Mario


JMC47:
Fixes the issue in Paper Mario: TTYD

PR #8231, opened 2019-07-06
DolphinQt: Don't update debug widgets when hidden

JosJuice:
Saves on CPU usage when pausing/unpausing with the debugger disabled. This is especially important when using frame advance rapidly.

PR #10111, opened 2021-09-18
DolphinQt: Don't update NetworkWidget when hidden

JosJuice:
Like PR #8231 but for NetworkWidget (which was added later).


sepalani (approved):
LGTM.

PR #10114, opened 2021-09-19
Core/DiscIO: Extract disc and partition constants to DiscUtils.h.

AdmiralCurtiss:
I noticed a lot of these were duplicated across various files, figured it makes sense to collect them in one place.

### DIFF
--- a/Source/Core/DiscIO/DiscUtils.h
+++ b/Source/Core/DiscIO/DiscUtils.h
@@ -29,6 +29,26 @@ constexpr u32 PARTITION_UPDATE = 1;
 constexpr u32 PARTITION_CHANNEL = 2;  // Mario Kart Wii, Wii Fit, Wii Fit Plus, Rabbids Go Home
 constexpr u32 PARTITION_INSTALL = 3;  // Dragon Quest X only
 
+constexpr u32 DISCHEADER_ADDRESS = 0;
+constexpr u32 DISCHEADER_SIZE = 0x440;
+constexpr u32 BI2_ADDRESS = 0x440;
+constexpr u32 BI2_SIZE = 0x2000;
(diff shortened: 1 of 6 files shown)


Source/Core/DiscIO/DiscUtils.h
```diff
@@ -29,6 +29,26 @@ constexpr u32 PARTITION_UPDATE = 1;
 constexpr u32 PARTITION_CHANNEL = 2;  // Mario Kart Wii, Wii Fit, Wii Fit Plus, Rabbids Go Home
 constexpr u32 PARTITION_INSTALL = 3;  // Dragon Quest X only
 
+constexpr u32 PARTITION_DISCHEADER_ADDRESS = 0;
+constexpr u32 PARTITION_DISCHEADER_SIZE = 0x440;
+constexpr u32 PARTITION_BI2_ADDRESS = 0x440;
+constexpr u32 PARTITION_BI2_SIZE = 0x2000;
+constexpr u32 PARTITION_APPLOADER_ADDRESS = 0x2440;
```
JosJuice, line 36: I think it would make more sense if the constants for bi2 and the apploader didn't have partition in the name, since they're also used on GameCube discs (where there are no partitions).

Though, hmm... We do want this to be consistent with `PARTITION_DISCHEADER`, ideally. So feel free to ignore this comment if you think this way is better.

AdmiralCurtiss: Just having it as `DISCHEADER_ADDRESS` felt weird next to `NONPARTITION_DISCHEADER_ADDRESS`, but eh, I don't really mind either way.

AdmiralCurtiss: How about this?

JosJuice (approved):

leoetlino (approved):

PR #10116, opened 2021-09-20
StickWidget: Inscribe stick circle within a square

malleoz:
This PR creates a light gray box around ellipses in the StickWidget. This mimics the old GUI and is useful as it allows TASers to identify the edges of the interactive stick area when clicking outside of the circle.

5.0
image

5.0-15224
image

This PR
image

### DIFF
--- a/Source/Core/DolphinQt/TAS/StickWidget.cpp
+++ b/Source/Core/DolphinQt/TAS/StickWidget.cpp
@@ -47,6 +47,9 @@ void StickWidget::paintEvent(QPaintEvent* event)
 
   const int diameter = std::min(width(), height()) - PADDING * 2;
 
+  // inscribe the StickWidget inside a square
+  painter.fillRect(PADDING, PADDING, diameter, diameter, Qt::lightGray);
+
   painter.setBrush(Qt::white);
   painter.drawEllipse(PADDING, PADDING, diameter, diameter);
 


JosJuice:
Your solution for the alignment feels like a bit of a hack. Keep in mind that Dolphin is translated into many languages, some of which may have significantly longer (or shorter) strings than English. The proper solution would be to use something like a QGridLayout, but I suppose that's not the easiest thing to do with how `CreateSliderValuePairLayout` is structured... Should we maybe not have a `CreateSliderValuePairLayout` function that creates a layout for just one row? Either way, I think this question is complex enough that this matter should be split out into a different PR, so that the ellipse/square commit can get merged faster.

Please also remove the merge commit from this PR.

JosJuice:
@dolphin-emu-bot rebuild

JosJuice:
Okay, second commit removed from the PR. You should update the title of the PR to match the new scope of the PR, though.

The change that this PR makes looks a little strange visually (maybe a black border around it all would help?), but the reasoning behind it makes sense, and the code is fine.

@JMC47, what do you think about the visuals?

JMC47:
Having the gray box around the joysticks; they don't care as much about the shell of the controller unless they're trying to test the viability of console strats.  I think this is a good change.

JosJuice (approved):

PR #10118, opened 2021-09-22
NetPlayProto: Remove lots of casts to MessageId when inserting enum values into packets

lioncash:
`MessageId` is kind of used as a wildcard type to insert identifiers into packets. Ideally, sf::Packet would just automatically handle attempted insertions of enum types by converting given identifiers into their underlying primitive type and then insert their values into the packet. This would get rid of the need to have the `MessageId` type alias and allow enums to be self-describing in terms of how they'll be stored.

Unfortunately `sf::Packet` doesn't handle enum values out-of-the-box, so we can make our own wrappers to make working with them a little more convenient and also a little bit less verbose


Source/Core/Core/NetPlayClient.cpp
```diff
@@ -833,9 +833,9 @@ void NetPlayClient::OnStartGame(sf::Packet& packet)
 
     for (auto& device : m_net_settings.m_EXIDevice)
     {
-      int tmp;
+      ExpansionInterface::TEXIDevices tmp;
       packet >> tmp;
-      device = static_cast<ExpansionInterface::TEXIDevices>(tmp);
+      device = tmp;
```
JosJuice, line 838: Wouldn't it be possible to get rid of the temporary variable here and in a few other places now?

lioncash: Ah, true, good catch

Source/Core/Core/NetPlayServer.cpp
```diff
@@ -986,10 +986,10 @@ unsigned int NetPlayServer::OnData(sf::Packet& packet, Client& player)
 
   case MessageID::GameStatus:
   {
-    u32 status;
+    SyncIdentifierComparison status;
     packet >> status;
 
-    m_players[player.pid].game_status = static_cast<SyncIdentifierComparison>(status);
+    m_players[player.pid].game_status = status;
```
JosJuice, line 992: This one as well

JosJuice: Ah, wait, the `status` variable is also later. Feel free to ignore.

PR #10119, opened 2021-09-22
TASInputWindow: Switch parent depending on m_rendering_to_main, fixed window size

malleoz:
In the current state of Dolphin dev, TAS Input window behavior is a bit inconvenient. Clicking on the game window will cause TAS Input to hide behind the game window. Instead, this PR will swap the TAS Input window(s) parent to the render widget on game bootup. If "Render to Main Window" is disabled, then this has the added benefit, compared to Dolphin 5.0, of not obstructing menu navigation.

I did not set a parent on the window object creation, as AdoptTASInput() is always called via ShowTASInput().

I also enabled fixed size for the TAS Input windows. It is unnecessary to be able to resize these windows. However, the only concern I have is that stick widgets are quite small (and misshapen in the case of WiiMote with no extensions). I believe another PR should be created to reshape these layouts to make room for enlarged stick widgets.

### DIFF
--- a/Source/Core/DolphinQt/MainWindow.cpp
+++ b/Source/Core/DolphinQt/MainWindow.cpp
@@ -956,6 +956,7 @@ void MainWindow::FullScreen()
   }
   else
   {
+    AdoptAllTASInputWindows(m_render_widget);
     m_render_widget->showFullScreen();
   }
 }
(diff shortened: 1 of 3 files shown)


JosJuice:
@dolphin-emu-bot rebuild

Dentomologist:
@dolphin-emu-bot rebuild

PR #10121, opened 2021-09-23
DolphinQt: Add option for Never Hide Mouse Cursor

malleoz:
Adds a toggle in Config->Interface to always keep the mouse cursor visible. It seemed unusual that this wasn't possible before in DolphinQt.

Given that "Always Hide Mouse Cursor" and "Never Hide Mouse Cursor" are mutually exclusive, I've also added functionality to disable one when the other is selected.

Also added an override when loading .ini config so that if both are selected, Never Hide takes priority.


Rumi-Larry:
Wouldn't it be better if this was a radio button menu?
"Cursor hiding:
Default
Always
 Never"

malleoz:
> Wouldn't it be better if this was a radio button menu?
> "Cursor hiding:
> Default
> Always
> Never"

This seems like a much cleaner approach and will make end-user experience less confusing. I will apply this change later. Thanks for your suggestion!

malleoz:
Radio buttons have been added. I don't know if this was the cleanest way to implement the functionality (particularly my use of the enum).

mbc07:
Small nitpick but I think it's important to have "cursor" in the Group Box title, the current "Hide Mouse" title can be a bit ambiguous. I would probably go with "Mouse Cursor Hiding"...

Rumi-Larry:
I think the header could read "Mouse Cursor Visibility" but the options "Always" and "Never" should be swapped. 
Instead of "Delay" that option could be called "On Inactivity" with the description reading "Mouse cursor hides after a certain time of inactivity and reappears when it moves again"

malleoz:
> I think the header could read "Mouse Cursor Visibility" but the options "Always" and "Never" should be swapped.
> Instead of "Delay" that option could be called "On Movement" with the description reading "Mouse cursor hides after a certain time of inactivity and reappears when it moves again"

I agree again. I do think Delay was a bit ambiguous and switching the header and the radio button label gives a better explanation.

Although, the mouse cursor, when hidden after the timer expires, does NOT reappear on movement. It actually reappears on the next click...

That being said, should the behavior instead by changed to make the cursor visible on movement, or keep it on click?

Source/Core/Core/ConfigManager.h
```diff
@@ -14,6 +14,9 @@
 #include "Common/Common.h"
 #include "Common/CommonTypes.h"
 
+// X.h defines Always to be 2, which causes problems with enum HideCursor
+#undef Always
```
mbc07, line 18: Undefining something defined by a system library just to avoid a name conflict seems very risky and is a potential source of bugs. Why not rename `Always` from `enum ShowCursor` to something else instead?

malleoz: Someone had suggested I mimic what another file has done to undefine X.h's `Always`, but that scenario had quite a number of enums so it was probably a more reasonable instance to use it. For this enum it does make sense to change it and I'll do so.

JMC47:
@dolphin-emu-bot rebuild

Source/Core/DolphinQt/Settings/InterfacePane.cpp
```diff
@@ -169,23 +170,36 @@ void InterfacePane::CreateInGame()
...
-      tr("Will immediately hide the Mouse Cursor when it hovers on top of the Render Widget, "
-         "otherwise "
-         "there is a delay.\nIf \"Lock Mouse Cursor\" is enabled, it will hide on Mouse locked"));
+  auto* mouse_groupbox = new QGroupBox(tr("Mouse Cursor Visibility"));
+  auto* m_vboxlayout_hide_mouse = new QVBoxLayout;
+  mouse_groupbox->setLayout(m_vboxlayout_hide_mouse);
+
+  m_radio_cursor_visible_movement = new QRadioButton(tr("On Movement"));
+  m_radio_cursor_visible_movement->setToolTip(
+      tr("Mouse Cursor hides after inactivity and returns upon Mouse Cursor movement."));
```
Filoppi, line 180: In the "On Movement" and "Never" mode, is it like before where the cursor became invisible only when on top of the render widget and always staeyd visible outside of it (e.g. if playing in a window or with multiple monitors)?
Or has that changed? If so, shouldn't there be yet another setting "only hide when on top of render widget"?
Please make sure this works well with and without mouse lockign?

malleoz: Correct, the cursor only becomes invisible when on top of the render widget and when it is inactive. I don't think an additional settings should be added to change that behavior. It doesn't seem significant.

I did test mouse locking and it is working as intended with my changes.

MayImilae:
UI looks good to me. However I have a small question - with the On Movement setting, how long after movement is stopped until the mouse cursor disappears?

malleoz:
> UI looks good to me. However I have a small question - with the On Movement setting, how long after movement is stopped until the mouse cursor disappears?

It is currently set to disappear after 3 seconds of inactivity which is the default behavior prior to this PR.

PR #10122, opened 2021-09-23
HotkeyScheduler: Check for certain hotkeys in game list

malleoz:
There are certain hotkeys that we absolutely want to be able to use without being in-game. Presently, no hotkeys are recognized unless we are in-game.

I've identified and moved the following hotkeys to be checked before the HotkeyScheduler checks to see if the Core is running:

- _Open_
- _Exit_
- _Start Recording_
- _Refresh Game List_

Note that _Play Recording_ should also be implemented here, however it looks like there is no signal for a PlayRecording() function, so this will have to be handled in a later PR once that signal is created and implemented.

### DIFF
--- a/Source/Core/DolphinQt/HotkeyScheduler.cpp
+++ b/Source/Core/DolphinQt/HotkeyScheduler.cpp
@@ -169,11 +169,24 @@ void HotkeyScheduler::Run()
 
       HotkeyManagerEmu::GetStatus(true);
 
-      if (!Core::IsRunningAndStarted())
-        continue;
-
+      // Open
       if (IsHotkey(HK_OPEN))
         emit Open();
+
+      // Refresh Game List
+      if (IsHotkey(HK_REFRESH_LIST))
+        emit RefreshGameListHotkey();
+
+      // Recording
+      if (IsHotkey(HK_START_RECORDING))
+        emit StartRecording();
+
+      // Exit
+      if (IsHotkey(HK_EXIT))
+        emit ExitHotkey();
+
+      if (!Core::IsRunningAndStarted())
+        continue;
 
       // Disc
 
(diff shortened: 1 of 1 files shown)


lioncash:
You may want to consider putting the PR description into the commit itself, so the context of why the change was made is included in the repository as well, once merged

malleoz:
> You may want to consider putting the PR description into the commit itself, so the context of why the change was made is included in the repository as well, once merged

Done. Will keep this in mind for the future. I didn't consider that the PR notes will not be easily visible once merged. Thanks!

Source/Core/DolphinQt/HotkeyScheduler.cpp
```diff
@@ -169,12 +169,25 @@ void HotkeyScheduler::Run()
 
       HotkeyManagerEmu::GetStatus(true);
 
-      if (!Core::IsRunningAndStarted())
-        continue;
-
+      // Open Disc
```
JosJuice, line 172: This one is a bit weirdly named. It should be Open File or Open Game. (Not sure if we have any reason to prefer either of those two, like a matching string elsewhere in DolphinQt)

malleoz: I used Open Disc because a few lines later, EjectDisc() and ChangeDisc() are grouped by the comment "Disc". Fair point though, I have changed to just Open.

JosJuice: There is the important difference that Open also works with things like WAD files and DFF files, whereas eject disc and change disc actually only work with disc images.

JosJuice:
@dolphin-emu-bot rebuild

JosJuice:
@leoetlino Would `HK_TRIGGER_SYNC_BUTTON` also make sense to enable when no game is running?

leoetlino:
No, that does nothing when no game is running

JosJuice (approved):

lioncash (approved):

PR #10124, opened 2021-09-24
HotkeyScheduler: add PlayRecording hotkey support

malleoz:
This hotkey was not previously implemented in the scheduler, even though it's present in the hotkey menu.

### DIFF
--- a/Source/Core/DolphinQt/HotkeyScheduler.cpp
+++ b/Source/Core/DolphinQt/HotkeyScheduler.cpp
@@ -186,7 +186,13 @@ void HotkeyScheduler::Run()
         emit ExitHotkey();
 
       if (!Core::IsRunningAndStarted())
+      {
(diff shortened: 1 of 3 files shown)


JosJuice:
@dolphin-emu-bot rebuild

JosJuice (approved):

malleoz:
I found there was unintended behavior when pressing Play Recording while a game was already running. Given that this was already not possible in previous dev versions (and quite frankly I don't see why it would be of much help), I have chosen to not recognize the hotkey if a game is running.

I've also identified a further hotkey issue, where hotkeys only work in the games list if "Hotkeys Require Window Focus" is disabled. A future PR should handle this, as it's a bit out of scope of the problem this PR aimed to solve.

leoetlino (approved):

PR #10125, opened 2021-09-25
TASInputWindow: always recognize hotkeys

malleoz:
This is a quality-of-life fix for those using TAS Input to record Tool-Assisted recordings.

Previously, hotkeys would not be recognized when TAS Input windows are in focus, unless "Hotkeys Require Window Focus" was disabled. Disabling this setting obviously causes hotkeys to register while other applications are in focus.

TASers do not expect to have to tab into the game renderer in order to frame advance. If they share my belief, then it is inconvenient if hotkeys are registered when tabbed out into other applications.

As such, this commit allows hotkeys to register always when TAS Input windows are in focus.

I believe I have added a `Host_TASInputHasFullFocus()` function for every GUI backend, but please let me know if I have missed one.

NOTE: this will conflict with #10130 and will need to be resolved upon its merge.

### DIFF
--- a/Source/Android/jni/MainAndroid.cpp
+++ b/Source/Android/jni/MainAndroid.cpp
@@ -158,6 +158,11 @@ bool Host_RendererHasFullFocus()
   return true;
 }
 
+bool Host_TASInputHasFullFocus()
+{
+  return false;
+}
+
 bool Host_RendererIsFullscreen()
 {
   return false;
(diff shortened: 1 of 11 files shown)


Pokechu22:
@dolphin-emu-bot rebuild

Avasam:
I'd love to see this feature merged. Having to bounce between both windows is really annoying. (and this is different from allowing hotkeys in background)
There's two currently open related issues: https://bugs.dolphin-emu.org/issues/11578 and https://bugs.dolphin-emu.org/issues/12675

PR #10127, opened 2021-09-26
HLE Riivolution patch support

AdmiralCurtiss:
Figured this is far enough along now that I'd like some reviews/feedback.

This implements the ability to parse and apply Riivolution XML files by rebuilding the game's FST before launch using DirectoryBlob.

A new subfolder in the user directory (`Load/Riivolution`) has been added to act as the virtual SD card root that Riivolution patches would normally search files in. Patch XMLs in `Load/Riivolution/riivolution` will also be parsed automatically when using this feature, mimicking how Riivolution loads XML files in `(SD root)/riivolution`.

To launch a game using this feature, highlight it in the game list, then select `Tools` -> `Start with Riivolution...` after placing the desired patch files in `Load/Riivolution`. Then select the patches you'd like to apply, and click `Start`.

So for example, to launch Newer Super Mario Bros. Wii you would do the following:

- Download the mod [from its official website](https://newerteam.com/wii/).
- Extract the `NewerSMBW` and `riivolution` folders into the `Load/Riivolution` directory. ([example](https://user-images.githubusercontent.com/4522237/134794133-6729c83b-a2da-485c-b775-b28637910d20.png))
- Launch Dolphin, highlight your vanilla New Super Mario Bros. game in the game list, and select `Tools` -> `Start with Riivolution...`. ([example](https://user-images.githubusercontent.com/4522237/134794136-094c5c8d-fd74-47d8-a512-d223c446c287.png))
- For this patch the defaults are already correct, so just hit `Start` in the patch selection window. ([example](https://user-images.githubusercontent.com/4522237/134794138-b9603b8f-ec91-4a50-98a8-3fb6da55fb6d.png))
- And enjoy!

step4

I'm open to suggestions for improving the user-experience, though I don't think the current way is terrible.

TODO:
- [ ] Support for Savegame redirects.
- [ ] Protection against path traversal attacks.
- [ ] Support for Ocarina-style patches.
- [ ] Progress bar while patching for larger patches?


sepalani:
I'm fine with the current UI, it has a similar look and feel compared to Riivolution UI.

I tested this PR with [UMH3](https://www.youtube.com/watch?v=tM8pAzWuxLM)+[MH3SP-patcher](https://github.com/sepalani/MH3SP-patcher) Riivolution files. I also used [AltWFC](https://github.com/polaris-/dwc_network_server_emulator) and [MH3SP](https://github.com/sepalani/MH3SP) as server replacements.

Regarding MH3SP-patcher, some memory patches are working, the search ones using `value` do, [the one](https://github.com/sepalani/MH3SP-patcher/blob/master/riivolution/MH3SP.xml#L50) using `valuefile` doesn't seem to work. I checked with the Memory widget and the value wasn't replaced.

It also seems the UMH3 title screen replacement isn't working. They look like this:
```XML
<file disc="/04/title_bg_eng.tpl" external="mh3t/textures/title/0.tpl" />
```

@AdmiralCurtiss 
I haven't tried this PR on Newer yet. I can if you want me to.

JosJuice (changes requested):
> Reviewed 2 of 2 files at r1, 2 of 2 files at r2, 1 of 1 files at r3, 2 of 2 files at r4, 2 of 2 files at r5, 2 of 2 files at r6, 2 of 2 files at r7.
> *[Reviewable](https://reviewable.io/reviews/dolphin-emu/dolphin/10127)* status: 0 of 24 files reviewed, 2 unresolved discussions (waiting on @AdmiralCurtiss and @JosJuice)

*[a discussion](https://reviewable.io/reviews/dolphin-emu/dolphin/10127#-MkWTtTf3NRp_zv6iUcf:-MkWTtTg-9Rq7R7qXP9Z:b52egi) (no related file):*
I think the system of having to first select the game in the game list and then go to the Tools menu will be unintuitive for users. Currently the only actions in Dolphin that a user needs to do this for are certain TAS actions like starting recording a movie from boot, and I've seen it happen several times that TASers who are new to Dolphin don't understand why Start Recording is greyed out and what they're supposed to do in order to start recording. Having "Start with Riivolution" as a right-click option on games would be better.

---
___
*[Source/Core/DiscIO/DirectoryBlob.cpp, line 554 at r5](https://reviewable.io/reviews/dolphin-emu/dolphin/10127#-MkVwFse5JJjAMY3bJ38:-MkVwFsfCjnEFcTLX5Br:b-nsjy8j) ([raw file](https://github.com/dolphin-emu/dolphin/blob/31e298c857b0069cc8b0a15be6a686565abef671/Source/Core/DiscIO/DirectoryBlob.cpp#L554)):*
> ```cpp
> 
>     const u64 data_size = partitions[i].partition.GetDataSize();
>     const u64 partition_offset = partition_address + PARTITION_DATA_OFFSET;
> ```

I would suggest naming this `partition_data_offset` if it doesn't mess up the line widths too much. I try to always use "partition" when referring to the beginning of the whole partition and "partition data" when referring to the start of the encrypted+hashed area, whereas "address" and "offsets" are basically synonyms.

iwubcode:
> So for example, to launch Newer Super Mario Bros. Wii you would do the following:

This is very cool @AdmiralCurtiss !  I can think of some usability improvements.

1) Could these mods work like TexturePacks?  If we could organize by gameid (or using the gameid.txt) we could have a more generic "enhancements" system in the future that works across the board for all our internal enhancements.
2) Maybe add a "right click" to open up the mod folder?
3) Instead of the "start with riivolution", could we just be able to right click on the game and select these in the same way we support "patches"?  Again in the future, all these systems could be combined in a more intuitive interface.  This would also mimic other emulators (ex: yuzu) and would support multiple mods at once.

I will give this a whirl soon.  I have a lot of translations I want to check out.

AdmiralCurtiss:
> Could these mods work like TexturePacks? If we could organize by gameid (or using the gameid.txt) we could have a more generic "enhancements" system in the future that works across the board for all our internal enhancements.

I did it like this so that the patches are as close as how you would install and use them on a real Wii -- which for most mods just means extracting the patch archive into the Riivolution folder (instead of the SD card root) and you're good. I'm not against 'future-proofing' this for a possible enhancement system, but I have no idea what the best way to accomplish that would be.

> Maybe add a "right click" to open up the mod folder?

Where?

> Instead of the "start with riivolution", could we just be able to right click on the game and select these in the same way we support "patches"? Again in the future, all these systems could be combined in a more intuitive interface. This would also mimic other emulators (ex: yuzu) and would support multiple mods at once.

Multiple mods at once are already supported here -- as they are on real Riivolution, for that matter! It will load all matching XMLs at once and you can just pick and choose what you want. I'm not the biggest fan of handling this like game cheats/patches because that would make it very easy to accidentally leave a mod on when you want to play the vanilla game, but ymmv.

iwubcode:
> Multiple mods at once are already supported here -- as they are on real Riivolution, for that matter! It will load all matching XMLs at once and you can just pick and choose what you want.

Any idea how it handles conflicts?

> I'm not the biggest fan of handling this like game cheats/patches because that would make it very easy to accidentally leave a mod on when you want to play the vanilla game, but ymmv.

Fair.  I was looking at this the opposite way.  Thinking that if I most often want to play with the mod, I have to remember to use this sub-menu to launch it.  I can't just double click on it.  Plus, it's very possible I want to trigger it from the command line (ex: frontend launchers).  Having a separate option for that might be a bit klunky (just my opinion of course).

>  I'm not against 'future-proofing' this for a possible enhancement system, but I have no idea what the best way to accomplish that would be.

If we could just move these files one level down, ex add a gameid layer, that'd be nice.  Just to make it clearer what mods belong to what game.

> Where?

I was thinking right click on the game itself.  If we had a gameid layer, we could open up the gameid folder itself (and if it doesn't exist create it before opening).  This would make it easy to see what mods were associated with each game.

Even if we didn't have a gameid layer, I think it'd be intuitive to add it to the game context menu.

---

Riivolution is for Wii retail only right?  just curious, is there anything similar for GC/Wiiware?  Wondering if `Load/Mods` would be more future proof?

JosJuice:
> If we could just move these files one level down, ex add a gameid layer, that'd be nice. Just to make it clearer what mods belong to what game.

Riivolution XML files already identify what game ID they're for. (See the `IsValidForGame` function in this PR.)

I think the main result of requiring a game ID folder structure would be to make this feature harder for users to set up. Getting them to name their texture pack folders correctly is already hard as it is... (Or at least it was before the texture pack folder naming system was improved. I'm don't think I actually see that many problems with it nowadays.)

iwubcode:
Thanks @JosJuice .  If we have a way to find all the mods for a given game (ex: ability to list all enhancements in the UI for a given gameid), I think we're at least future-proofed in that sense.  Updated my comments.

iwubcode:
One other comment, if we do keep the current interface, please switch to having "Start" as the default button.  That way after clicking to begin the game, I can just hit enter to play.

JMC47:
Was pretty painless to setup for the most part.  I think the biggest issue is figuring out to highlight the game if you're not used to recording TASes.

Zopolis4:
Does this have a priority order like real riivolution? Also, is the intended method to copy the riivolution folders for each respective mod into the main riivolution folder in `Load/Riivolution/riivolution`?

AdmiralCurtiss:
Does Riivolution have a priority system? I haven't seen anything along those lines but maybe I've just overlooked it. Right now I'm not doing any conflict handling or similar, and I was under the impression Riivolution didn't either.

Zopolis4:
Riivolution handles conflicts via the vertical order of the mods, with the higher mods overriding the other mods, iirc.

AdmiralCurtiss:
And how is that order decided (especially across multiple XML files)...?

Zopolis4:
Dragging and dropping with wiimote, I think. Not sure about the backend, but the source code probably shows.

AdmiralCurtiss:
Oh, I didn't realize that was possible! Okay, thanks, I'll research the exact details.

AdmiralCurtiss:
@sepalani Can you try the Monster Hunter patches again? The XML files didn't set any root patch, which I assumed meant 'relative to SD root' but seems to instead mean 'relative to folder containing xml'. They also used Ocarina patches, which were unimplemented before -- I've implemented them now, I think.

sepalani:
@AdmiralCurtiss 
That seems to work now:

RMHE08_2021-09-27_13-07-49
RMHE08_2021-09-27_13-14-52

AdmiralCurtiss:
@Zopolis4 Are you absolutely sure Riivolution on console has a priority feature? I just spent some time messing with it and was unable to find any such thing. Additionally, if I enable multiple patches for the same file, they just seem to get applied in the order they're listed in the XML file, later patches for any given file or file region overwriting whatever happens to be there at the time -- which is identical to what I'm doing here.


One thing I ran into that I didn't realize before is that apparently the offset in mid-file patches is limited to 4 byte alignments, getting aligned down (that is, `address & ~3`) when it's in-between. I should add that.

Also, I'm still not fully treating external paths accurately. If they start with a '/' they seemingly are relative to the SD card root, regardless of what is given as the 'root' parameter. Which makes some sense, I guess.

Fixes for both of these will follow shortly...

Zopolis4:
I may be wrong, I'm mainly going off of memory here. I trust your judgment on this, especially as I could not find any relevant mentions of priority in riivolution's source code.

iwubcode:
> later patches for any given file or file region overwriting whatever happens to be there at the time -- which is identical to what I'm doing here.

I need to look through the code but how do you signify a later patch??  These are just files in the directory, are they loaded in name order?

AdmiralCurtiss:
I mean, like, if you did this:

~~~
<wiidisc version="1" root="_test">
  <id game="SUKE" />
  <options>
    <section name="Sec1">
      <option id="A" name="A" default="1">
        <choice name="a">
          <patch id="a" />
        </choice>
      </option>
      <option id="B" name="B" default="0">
        <choice name="b">
          <patch id="b" />
        </choice>
      </option>
      <option id="C" name="C" default="0">
        <choice name="c">
          <patch id="c" />
        </choice>
      </option>
    </section>
  </options>
  <patch id="a">
    <file disc="/msg/US_English/TitleMenu.msbt" external="_A" />
  </patch>
  <patch id="b">
    <file disc="/msg/US_English/TitleMenu.msbt" external="_B" />
  </patch>
  <patch id="c">
    <file disc="/msg/US_English/TitleMenu.msbt" external="_C" />
  </patch>
</wiidisc>
~~~

If you enable all three of these patches, file `_C` ends up being used, presumably because it patches them over the data in order. Similarly, if you assume those replacement files are all, say, 8 bytes long and did something like

~~~
  <patch id="a">
    <file disc="/msg/US_English/TitleMenu.msbt" external="_small1" offset="0x71c" resize="false" />
  </patch>
  <patch id="b">
    <file disc="/msg/US_English/TitleMenu.msbt" external="_small2" offset="0x724" resize="false" />
  </patch>
  <patch id="c">
    <file disc="/msg/US_English/TitleMenu.msbt" external="_small3" offset="0x720" resize="false" />
  </patch>
~~~

Then you end up with the first 4 bytes of `_small1`, followed by the full contents of `_small3`, followed by the last 4 bytes of `_small2` -- presumably because it overlays the three patches in that order.

With folder patches and recursive file replacement this gets more convoluted to test, but I'd presume it's similar?

AdmiralCurtiss:
If you mean the order the XMLs are being loaded and how that would affect the results, very good question. I'm not sure anyone ever fully considered this logic.

iwubcode:
> Which still looks to me that it's just whatever order the files happen to be stored in the FAT on the SD card, so completely arbitrary. Unfortunate!

Does that mean filename ordering?

--

I question if we should be HLE'ing the actual riivolution loading as opposed to us owning the patch ordering and riivolution just being one type of patch.  As I said before, we really should call the folder inside 'Load' **mods**, not make it specific to riivolution.  This allows us to support other patches in the future (ex: the [Pokemon Mystery Dungeon](https://www.romhacking.net/translations/5518/) translation patches are bps because riivolution only applies to Wii retail discs).

AdmiralCurtiss:
> Does that mean filename ordering?
Depends on the code that writes the FAT in the first place I guess? I think in practice you can make no assumptions, though of course we could define a logic for it for consistency (even if that may be inconsistent with actual Riivolution).

---

I do agree that supporting other types of patches makes sense, but the specifics of how you'd actually like that to be implemented are rather nebulous to me. I'm also not sure if it really makes sense to mix together multiple types of patches in the same place in the file system. Maybe make mockup or something so I have a better idea what you want here?

AdmiralCurtiss:
Like, if you're saying you want something like

~~~
├ Load/
 ├ Textures/
 ├ Mods/
  ├ (gameid)/
   ├ mod1/
   │├ mainfest.json <-- defines the name, type of patch and so on
   │├ patch.bps
   ├ mod2/
   │├ mainfest.json
   │├ riivolution.xml
   │├ files/
   etc.
~~~

Then yes, totally agreed that that's what we should eventually strive for. But! This completely breaks drop-in compatibility with existing Riivolution patches, and we can't exactly expect the average user to figure out how to define this for their 7 year old mod they downloaded. So unless we go full-hog and immediately make a big 'Mods' GUI for this with importing and everything -- which I'll be honest, I'm not particularly in the mood for at the moment -- this is a bit of a showstopper until we actually have the infrastructure to support this. Which, hey, if you want to code this up, be my guest, that's definitely the better long-term solution. It just kinda blows the scope of this PR way up if you ask me.

iwubcode:
> So unless we go full-hog and immediately make a big 'Mods' GUI for this with importing and everything -- which I'll be honest, I'm not particularly in the mood for at the moment 

I'm sorry if I gave the wrong impression.  I'm not saying we need a big mods GUI (though I have plans for something like this) or other crazy changes but just that we make sure we future-proof ourselves a bit, as Riivolution unfortunately isn't a silver bullet for our modding needs.

> Like, if you're saying you want something like ...

Yeah, something like that would work.  I'm leaning toward something similar to our TexturePack system and have some ideas for the UI but we'll see.  I finally had a chance to glance at the code for this Riivolution support and I'm not sure we need any significant changes made.  Should have done that to start :)

At its core, we need a way to determine what mods are available for a gameid.  The manifest.json approach you mention is the best scenario for additional metadata but I think we can also look for any `.xml` that exist as well and make the assumption those are riivolution files.  Not as much metadata is available but we can at least get the mod name.

When adding a second patch format, I think it'd be rather simple to implement with a `std::variant`, instead of:

```
DiscIO::Riivolution::ApplyPatchesToMemory(boot->riivolution_patches);
```

we could do (pseudocode):

```
for (const auto& patch : boot->patches)
{
  if constexpr (std::is_same_v<RiivolutionPatch> patch)
  {
     DiscIO::Riivolution::ApplyPatchToMemory(patch)
  }
  else if (std::is_same_v<BPSPatch> patch)
  {
    DiscIO::BPS::ApplyPatchToMemory(patch);
  }
}
```
(something similar for the files)

but that's something we can do when we get that mythical second patch format.  I'd assume those patches are applied in order, so that's also how we could support 'priority' in the future.

Only change I'd ask for now is to change the folder to "Mods" or something similar.

AdmiralCurtiss:
> Only change I'd ask for now is to change the folder to "Mods" or something similar.

...If you're sure. I'm just warning you, we're probably polluting the future intended mod directory structure with this.

iwubcode:
> ...If you're sure. I'm just warning you, we're probably polluting the future intended mod directory structure with this.

No I'm not sure.  I'm just one person with an idea and a passion for Wii games.  :)

I can expound on my idea for those interested.  There might be issues with it.


My idea is to adopt something similar to TexturePacks, where each mod has a folder under the root ("Mods" folder) with a gameid.json.  We search for all of these files and each one is considered a "mod" for the game.  It dictates the format (so how to apply the mod), possibly files involved (if we want to be explicit?) and metadata that can be used for a GUI. (ex: path for structure, title, author, etc).  Possibly options for more dynamic mods.  Json files will take priority but Riivolution mods (based on xml) will also be in the list (distributors of the Riivolution mods can provide a json for more details but aren't required).

Dolphin allows you to change the priority order of those mods, to handle conflicts.  That data would be stored in a separate file that is loaded on startup of Dolphin (for the UI) and boot of a game.

The UI would provide a detail of all game mods, sorted by game for the user to interact with.

Thoughts?

AdmiralCurtiss:
Yeah that all sounds great and I'd love to see that! But... that conflicts with making our makeshift Riivolution SD card root the `Mods` directory, doesn't it?

Like, maybe this is more illustrative of what I mean. Here's how my `Load/Riivolution` and `Load/Riivolution/riivolution` folders look after adding just a handful of patches for testing purposes:

riivolution-topdir
riivolution-subdir

If I were using these patches on a real Wii, this exactly mirrors the SD card root and the `riivolution` directory, respectively.

Now... do you want to fit your tidy mod folder file structure into this mess? Or do we just from one version to the next go, nope, sorry, mods have to be structured differently now? Neither seems like a particularly good option to me. Or do you have a different idea on how to migrate this later?

iwubcode:
> Yeah that all sounds great and I'd love to see that! But... that conflicts with making our makeshift Riivolution SD card root the Mods directory, doesn't it?

If you're ok with a structure like:

```
├ Load/
 ├ Textures/
 ├ Mods/
   ├ bps_mod1/
   │├ gameid.json <-- defines the name, type of patch and so on
   │├ patch.bps
   ├ ips_mod2/
   │├ gameid.json
   │├ riivolution.xml
   │├ files/
  ├ NewerSMBW/
   │├ ...
  ├ riivolution/
   │├ codes/
   │├ config/
   │├ NewerSMBW.xml
   │├ ....xml
   etc.
```

Part of my idea for the mod system is to be rather flexible.  So rather than have an arbitrary layout where individual folders reference individual mods, we just search for json or xml files and each of those files signifies a mod (right now that location for xmls is hardcoded to `riivolution`: https://github.com/dolphin-emu/dolphin/pull/10127/files#diff-474eed864cee8bd3b827d6ed8ca82475bac5299d19d20ab842461226e3a62f9dR83 ).  The file location and/or the file details signify what the layout looks like (maybe a bps mod has 3 bps files that need to be applied and those are next to the json in some directory).

> Now... do you want to fit your tidy mod folder file structure into this mess? Or do we just from one version to the next go, nope, sorry, mods have to be structured differently now? Neither seems like a particularly good option to me

Yeah, I thought of the same thing.

I don't know if there's any other option **if** you want to mimic the riivolution SDcard layout and riivolution patches are intended to be included in the mod system.  Even if we exclude riivolution patches, there might be other patch formats that are ugly in different ways.  Unless we develop a RVZ-style Dolphin specific patch format and a way to convert between it and other formats (bps, riivolution, ips, etc), I don't see an alternative.  It just doesn't seem very necessary to me to do all that work just to keep a clean folder structure.  Ideally, the user should rarely see that folder (maybe to install/unpack a mod?), everything should be managed through the UI (could even include deletion)

sepalani:
@iwubcode 
> EDIT: I guess I should add, this also is assuming we don't want to have separate folders for separate patch types. I guess **a third option is we could do `Riivolution`, `BPS`, etc top level directories**. And then Dolphin collects them all. That'd keep each folder as clean as possible. The negative is the user has to put different mods in different folders.

I think that's the only way around it, IMHO. It's also what this PR is almost doing, AFAICT. The path `Load/Riivolution` can be changed to `Load/Mods/Riivolution` if that's the issue.

Riivolution patches has to be put in `SD:\\Riivolution` folder which will be scanned for any XML files by the Riivolution app to load patches. These patches can search the files to be injected **anywhere** in the SD card including at the root of the SD card (outside the Riivolution folder).

This PR's `Load/Riivolution` is the root of the SD card and will search for XML files in `Load/Riivolution/Riivolution`. Due to Riivolution's behaviour, we have no easy way to make it cleaner or isolate Riivolution patches per gameid while patches can load files from any folders on the SD card (even outside the Riivolution folder). Moreover, some patches can work on some specific regions like (PAL+NTSC-U, All regions, Japan only, etc.) which will be complicated to handle.

JosJuice:
Personally I think we should go with having a "Riivolution" folder the way this PR originally implemented it, for the reasons described by AdmiralCurtiss. If we do want to support multiple mod types later, we would have both the option of having a separate folder for each mod type and also the option of coming up with some fancy design for how a unified "Mods" folder would work (we do not have a working design for this right now!) while leaving in the ability to load from the Riivolution folder as a legacy feature just for mods the user added in the past.

iwubcode:
> This PR's Load/Riivolution is the root of the SD card and will search for XML files in Load/Riivolution/Riivolution. Due to Riivolution's behaviour, we have no easy way to make it cleaner or isolate Riivolution patches per gameid while patches can load files from any folders on the SD card (even outside the Riivolution folder).

Just to point out.  There's no reason Dolphin needs to be restricted by this behavior.  We own the file loading process, we aren't calling a riivolution app that handles it, we're HLE'ing it.  We hardcoded the `riivolution` folder as being the place we look for the underlying xmls.  Obviously you want to support that since users who are familiar with riivolution will expect that system but we don't need to force that behavior.

> If we do want to support multiple mod types later, we would be able to choose later between the option of having a separate folder for each mod type and the option of coming up with some fancy design for how a unified "Mods" folder would work while leaving in the ability to load from the Riivolution folder as a legacy feature just for mods the user added in the past. Naming the Riivolution folder "Mods" would not let us do either of these cleanly,

I don't personally like the idea of "leaving in the ability to load from the Riivolution folder as a legacy feature just for mods the user added in the past" as a general software design practice because that means long-term technical debt.

If we do separate folders, we should stick with separate folders.  We shouldn't introduce a "Mods" folder, that would just be confusing to users.  Riivolution files are also mods.

---------

I don't see a reason to not do the "Mods" folder other than it being a bit messy in the default `riivolution` layout.  However, we are talking about a hypothetical scenario at this point (unless someone adds bps, ips, xdelta, etc support), so I don't want to delay what otherwise is a really fantastic feature.

@AdmiralCurtiss - my apologies, please go ahead and switch the root back to `Riivolution`

RoadrunnerWMC:
This is awesome! I have a UI suggestion, though:

Unless you're speedrunning, playing a mod like Newer will normally be split up across many sessions. For most games in Dolphin, this is as convenient as imaginable -- launch Dolphin, click the game you want to play, and click "Play" to pick up where you left off. Note that this is also the existing UX flow for playing Newer in Dolphin with existing "ISO builder" tools, so it's a reasonable thing to compare to.

image

With this PR, Newer is much easier to set up the first time, but you somewhat pay for this convenience in terms of a few extra required clicks every time you want to launch it: right click, "Start with Riivolution", edit settings if necessary (after reading all the discussion above, it's unclear to me whether Dolphin remembers your previous settings here), "Start".

Would it be possible to show Riivolution patches in the game list directly somehow?

----

Considerations:

* Dolphin needs to remember Riivolution-patch-related settings. I think this is important even if nothing else proposed in this post happens.
* A Riivolution mod by itself doesn't know exactly which game to apply itself to. The XML specifies the game ID as a filter, but if you have multiple copies of one game (for example, currently I have five distinct New Super Mario Bros. Wiis in my Dolphin game list), that's still ambiguous.
* What if the user tries to launch the game before they've configured it?

An intuitive but *bad* design would be to have Dolphin add an entry to the game list for each Riivolution path it recognizes. This doesn't work because it would prevent the user from selecting multiple patches at once, which is a feature of the current UI design. Instead, I propose the following:

## Design Proposal

An "Add to Game List" button is added to the current "Start with Riivolution" dialog. This adds a new "child" entry in the game list underneath the base ROM. Right-clicking it will show a "Configure..." option, which shows the config dialog again.

This would require changing the game list to support displaying a tree structure instead of just a flat list, which I realize may be infeasible (I'm not too familiar with Dolphin's UI code). Probably something sort of like this:

image

If the tree design turns out to be impossible or too difficult, a compromise could be to just add a new "Riivo" type to the "Platform" column in the game list. This version has some drawbacks, though -- it's unclear if Riivolution patches should be sorted in the list like any other game or be permanently stuck underneath their parent game, and if they do sort like other games, how can the user tell what the parent game is?

Another issue is that it's not really obvious how to populate the list entry. After all, it represents a Riivolution configuration, not an individual patch, so it doesn't even have an intrinsic name, much less a banner. There could be a prompt to enter a name when you click "Add to Game List," but the banner problem is harder to solve. If a patch uses save redirection and includes a custom save banner, Dolphin could show the banner from there, but this doesn't help for other patches, and it also doesn't suggest a method by which the user could override the banner (and other metadata) like they can for any other type of game.

Anyway, if these more challenging design points can be worked out, I think a system like this would go a long way towards improving the long-term user experience.

Zopolis4:
I like WMC's suggestion, and had myself been thinking about it for a while. Another way to implement it: have a "Add Current Config" (To Game List) button at the bottom of the start with riivolution window, which adds the current mod setup to the gamelist in a single step, although maybe with some configuring for names/banners/gameID's.

RoadrunnerWMC:
Yeah, that's essentially what I was suggesting

JMC47:
I'd rather not go rewriting parts of Dolphin's gamelist to support stuff like this until after the initial merge.  What your suggesting would be nice, but it's also a lot of work and I don't want to hold up the initial merge over what is usability.

iwubcode:
Just to add a counter opinion, I'm not a huge fan of blowing out the game list (my gamelist is already large enough).  I still stand by a mod enable/disable feature with a _list._ of mods.  This matches every other emulator (Yuzu, Cemu, etc).  The additional entries in the game list doesn't sound like it would handle multiple mods enabled at once (unless I'm misunderstanding).

Outside of a mod list in the game properties window, I'd prefer a "remember this" checkbox or something similar that saved your current mod layout so when you double clicked or launched with the command line, you'd get the mod(s) in question.

RoadrunnerWMC:
> The additional entries in the game list doesn't sound like it would handle multiple mods enabled at once (unless I'm misunderstanding).

You are misunderstanding. The idea is that you would use the current "Start with Riivolution" option, select/configure the patch(es) you want, and then click a button that saves that configuration as a new game list entry. So each game list entry isn't one Riivolution patch, it's one Riivolution *configuration* that can include any number of patches.

I agree with Zopolis that having Dolphin remember only one Riivolution configuration per game isn't really a great idea because it's not uncommon for people to switch between several mods of the same game.

And I also agree with everyone that it makes sense to defer this to a later PR in order to get this one merged sooner.

Meatball132:
Hey, this is really neat! I'm a developer working on NewerSMBW right now (for an update). There's a Riivolution quirk that needs to be addressed for Newer to work properly I thought I should let you know about.

If you have disc transfer rate emulation on, switching between world maps stutters horribly and causes extremely high load times, because it's streaming the world map assets and the music simultaneously from disc. This wasn't a problem on real hardware, because files loaded by Riivolution weren't _on_ a disc. It doesn't make the game unplayable, but it is extraordinarily annoying, especially for players that don't know why it's happening.

In other words, for correct behaviour, Dolphin should _not_ emulate disc transfer rates for Riivolution patch files. Perhaps this could be an option, at least (defaulting to "don't emulate disc speeds")?

AdmiralCurtiss:
Yes, I am aware of this, we talked about this a bit on IRC. I'll need to find a good method to forward the disc speed for 'unpatched' disc files but use a different, likely constant, speed for files loaded from the patch.

Pokechu22:
I haven't really looked into how this is implemented, but I think the normal DVD transfer speed itself is probably fine &mdash; it's the seek times for the emulated disc drive head that are probably causing issues (especially if it's seeking between two files).

One option would be to place the Riivolution-generated files far after the normal end of the disc, and then modify `DVDInterface::ScheduleReads` to not perform the seeking logic when it's out of bounds (instead acting as if the data is already in the disc drive cache, like with `bFastDiscSpeed`).  The logic at the end that updates the buffer positions should probably also be skipped (so reads that don't hit a Riivolution file and instead use the disc drive get accurate emulation timings).

Meatball132:
> I haven't really looked into how this is implemented, but I think the normal DVD transfer speed itself is probably fine — it's the seek times for the emulated disc drive head that are probably causing issues (especially if it's seeking between two files).

Yeah, I believe that's the particular issue as well, thank you for clarifying.

Meatball132:
Another suggestion, although I'm not sure how feasible this is with how it's currently implemented:

You are able to browse for an XML outside of `Load/Riivolution`, but even if you do select one elsewhere it'll still treat `Load/Riivolution` like the root of the SD card. I think it should be able to use `[SelectedXML]/..` instead, if it can't find the required files. See the following image:
image
If I placed these files in an arbitrary location on my computer, and selected the XML inside that "riivolution" folder, I would expect it to be able to find the NewerSMBW folder as well just fine. But with the current behaviour it just crashes. This is _fine_, but in my personal opinion it is a slight annoyance.

I can definitely see why this might be difficult to do, though, especially depending on how it's implemented here.

Rumi-Larry:
> 
> 
> Yes, I am aware of this, we talked about this a bit on IRC. I'll need to find a good method to forward the disc speed for 'unpatched' disc files but use a different, likely constant, speed for files loaded from the patch.

I mean, presumably if you are using Riivolution, you aren't using disk media to play anyway, so emulating disk speeds makes no sense. I think making it disable the emulation of disk speeds when using this is warranted.

JosJuice:
> I mean, presumably if you are using Riivolution, you aren't using disk media to play anyway, so emulating disk speeds makes no sense.

Doesn't Riivolution **require** you to use a disc? At least that's what the documentation I can find says...

Pokechu22:
Yeah, it requires the disc for any files not in the patch (which there will almost always be some of).  It actually flashes the disc drive light when a read hits the disc drive, so you can see how often that is.

Rumi-Larry:
> 
> 
> Yeah, it requires the disc for any files not in the patch (which there will almost always be some of). It actually flashes the disc drive light when a read hits the disc drive, so you can see how often that is.

Holy shit, my bad! I thought that modders would avoid disk formats like the plague.

AdmiralCurtiss:
> File search path

If you want it to find the files when they are placed in an arbitrary location, make sure they're referenced in the XML with a relative path rather than an absolute one, eg:

```
<wiidisc version="1" shiftfiles="true" root="NewerSMBW" log="true">
	(...)
	<patch id="newercore">
		(...)
		<folder external="Maps" disc="/Maps" create="true" />
		<folder external="Stages" disc="/Stage" create="true" />
		(...)
	</patch>
</wiidisc>
```

newerr

I suppose we could make the root search directory manually settable, but I feel like that could cause confusion for the general userbase... ymmv. Maybe an extra 'root' field at the bottom of the window wouldn't be so bad if it does the current thing by default?

JosJuice:
> One option would be to place the Riivolution-generated files far after the normal end of the disc, and then modify `DVDInterface::ScheduleReads` to not perform the seeking logic when it's out of bounds (instead acting as if the data is already in the disc drive cache, like with `bFastDiscSpeed`). The logic at the end that updates the buffer positions should probably also be skipped (so reads that don't hit a Riivolution file and instead use the disc drive get accurate emulation timings).

I'm not sure if this is a great idea. Keep in mind that many Wii games will fail to boot with `ERROR #001` if out of bounds reads succeed. (I guess there is the possibility of reverting PR #9577, but then we're adding hacks just to get other hacks working...)

Meatball132:
> Note that I don't allow you to access files that are outside either the `Load/Riivolution` folder (for absolute paths) or the folder that contains the xml itself (for relative paths) to prevent someone reading your personal files with cleverly formed Riivolution XMLs.
Ah, well if you're definitely not going to read outside the folder containing the XML then I don't think there's much of a reason to be able to select the root manually.

The reason I asked is because Riivolution mods go in a `riivolution` folder on the SD card, so the root would always be a folder up from the XML; this is also how mods tend to be released, so it would just be more convenient to be able to extract your mods into their own folders instead of forcing it to be `Load/Riivolution`.

AdmiralCurtiss:
No that would be fine, actually! You'd set the root to the 'one folder up' and it would work like you expect.

Meatball132:
That's not what I meant; like, take Newer for example. It sets the root to `/NewerSMBW`. So Riivolution will look for the directory `NewerSMBW` on the root of the SD card, and set its new root to that. If you just unzipped a Newer download somewhere, then it won't work, because it will look for "NewerSMBW" inside of `Load/Riivolution` instead of the folder `riivolution` is contained in.

This would apply to anything that uses absolute paths rather than relative ones, because practically, "root" = "a folder above where the XML is" to Riivolution.

So to rephrase, my suggestion is "make the root a folder above where the XML is", because that is where it would realistically be on a real Wii, and mods are packaged according to this so unless you went and edited the XML you would HAVE to put it in `Load/Riivolution`.

IonicPixels:
also about the root folder, on some games, take Super Mario Galaxy for example. it wont always load the files in Load/Riivolution, and will only read the ASM (if any) in the XML.

Meatball132:
> 
> 
> also about the root folder, on some games, take Super Mario Galaxy for example. it wont always load the files in Load/Riivolution, and will only read the ASM (if any) in the XML.

I mean, if Riivolution isn't loading custom files (is that what you meant?), then the root isn't relevant anyway.

IonicPixels:
> > also about the root folder, on some games, take Super Mario Galaxy for example. it wont always load the files in Load/Riivolution, and will only read the ASM (if any) in the XML.
> 
> I mean, if Riivolution isn't loading custom files (is that what you meant?), then the root isn't relevant anyway.

well, if you are on a real Wii, for example, and this also happens in AdmiralCurtiss's explaination on how to use this, Riivolution takes the root and reads the folder names in the XML, and replaces all the files in game with the files in the folder temporarily.

in other words: if you have "<folder external="/SMGTKGS/ObjectData" disc="/ObjectData" recursive="true" create="true" />" then all the files in the original games ObjectData with the modded ObjectData until the game is closed.

AdmiralCurtiss:
You're misunderstanding my use of 'root', We have a physical root directory (which is the top directory of the SD card) and a logical root directory (which is the path specified in the XML itself), the latter of which if when set absolute is relative to where the former is. So if, hypothetically speaking, you extract your unmodified NewerSMBW zip into, `C:\Users\me\Downloads\Newer` (so that you now have the asset folder `C:\Users\me\Downloads\Newer\NewerSMBW` and the file `C:\Users\me\Downloads\Newer\riivolution\NewerSMBW.xml`), you would then set the theoretical 'SD card root' option in the 'Start with Riivolution Patches' dialog to `C:\Users\me\Downloads\Newer`. Since the `root="/NewerSMBW"` is an absolute path, it is then resolved relative to the configured root of `C:\Users\me\Downloads\Newer` to arrive at `C:\Users\me\Downloads\Newer\NewerSMBW` for the directory the rest of the patches reference. We could even potentially auto-assign that root upon loading the XML, I guess?

Complicated? Maybe. But that's what happens when you try to HLE a thing that assumes an existing filesystem and want it to be configurable, I'm afraid.

phire:
There really should be some way to use this functionality from Dolphin's command line. No real reason to make it clean for this PR, but I think it should at least be possible. 

Perhaps the user should be able to supply a list of Choices/Options on the command line.

IonicPixels:
> You're misunderstanding my use of 'root', We have a physical root directory (which is the top directory of the SD card on real hardware) and a logical root directory (which is the path specified in the XML itself), the latter of which when set absolute is relative to where the former is. So if, hypothetically speaking, you extract your unmodified NewerSMBW zip into, `C:\Users\me\Downloads\Newer` (so that you now have the asset folder `C:\Users\me\Downloads\Newer\NewerSMBW` and the file `C:\Users\me\Downloads\Newer\riivolution\NewerSMBW.xml`), you would then set the theoretical 'SD card root' option in the 'Start with Riivolution Patches' dialog to `C:\Users\me\Downloads\Newer`. Since the `root="/NewerSMBW"` is an absolute path, it is then resolved relative to the configured root of `C:\Users\me\Downloads\Newer` to arrive at `C:\Users\me\Downloads\Newer\NewerSMBW` for the directory the rest of the patches reference. We could even potentially auto-assign that root upon loading the XML, I guess?
> 
> Complicated? Maybe. But that's what happens when you try to HLE a thing that assumes an existing filesystem and want it to be configurable, I'm afraid.

hmmm on other mods it loads the StageData from the XML just fine, even when I have an unmodified XML. It sometimes does that when its in `Load\Riivolutuin\mod name`

AdmiralCurtiss:
Not you, my comment was meant for @Meatball132. If you don't specify a root directory at all in the XML it is indeed just relative to the XML itself.

IonicPixels:
> There really should be some way to use this functionality from Dolphin's command line. No real reason to make it clean for this PR, but I think it should at least be possible.
> 
> Perhaps the user should be able to supply a list of Choices/Options on the command line.

that is a very good idea and would be extremely useful at times, especially when using Dolphin without its UI

IonicPixels:
> Not you, my comment was meant for @Meatball132. If you don't specify a root directory at all in the XML it is indeed just relative to the XML itself.

oh ok

Meatball132:
> 
> 
> You're misunderstanding my use of 'root', We have a physical root directory (which is the top directory of the SD card on real hardware) and a logical root directory (which is the path specified in the XML itself), the latter of which when set absolute is relative to where the former is. So if, hypothetically speaking, you extract your unmodified NewerSMBW zip into, `C:\Users\me\Downloads\Newer` (so that you now have the asset folder `C:\Users\me\Downloads\Newer\NewerSMBW` and the file `C:\Users\me\Downloads\Newer\riivolution\NewerSMBW.xml`), you would then set the theoretical 'SD card root' option in the 'Start with Riivolution Patches' dialog to `C:\Users\me\Downloads\Newer`. Since the `root="/NewerSMBW"` is an absolute path, it is then resolved relative to the configured root of `C:\Users\me\Downloads\Newer` to arrive at `C:\Users\me\Downloads\Newer\NewerSMBW` for the directory the rest of the patches reference. We could even potentially auto-assign that root upon loading the XML, I guess?
> 
> Complicated? Maybe. But that's what happens when you try to HLE a thing that assumes an existing filesystem and want it to be configurable, I'm afraid.

I'm not, I think you are misunderstanding me. If the XML is at `C:\Users\me\Downloads\Newer\riivolution\NewerSMBW.xml` and it can't find the "logical root" at `C:\Users\me\Documents\Dolphin Emulator\Load\Riivolution\NewerSMBW` then it should look at where the XML is, go back a folder (to `C:\Users\me\Downloads\Newer`) and treat that as the new physical root directory - because the XML is in the riivolution folder, like on the SD, the SD root would be a level back from that no matter what assuming it is in the same position as it would relative to the OTHER files, as it would be on the SD card.

Because the XML is in `root/riivolution`. `riivolution` is always inside the root. So a level up would be the actual physical root, and that's where the other files that normally go on the SD card are packaged in mods.

Like, you literally extract the files from Newer's zip directly onto the root of the SD card, so a folder where you extract Newer's zip to should also be treated as the root.

Sorry for not being clearer :(

RoadrunnerWMC:
> Holy shit, my bad! I thought that modders would avoid disk formats like the plague.

To clarify this a bit: it doesn't require a disc *format* (as in, ISO file), but rather a literal physical disc to be inserted into the console's disc slot.

The only things needed on your SD card are the specific game files you want to replace, and a configuration XML. Riivolution is intentionally designed so that playing mods requires having a physical copy of the original game -- the authors had a very strong anti-piracy stance. :)

Meatball132:
Hm, yet another small thing: Riivolution, I believe, saves your configuration by default (in `SD:/riivolution/config`). If you have like 20 mods for a game and the default setting for them is all "enabled" then with this you currently have to, every single time you start with Riivolution, disable all of the ones you don't want. Would be especially nice if a real Riivolution configuration worked on Dolphin too.

IonicPixels:
> Hm, yet another small thing: Riivolution, I believe, saves your configuration in `SD:/riivolution/config` (disclaimer I didn't check whether or not this was true but I think it is lol). If you have like 20 mods for a game and the default setting for them is all "enabled" then with this you currently have to, every single time you start with Riivolution, disable all of the ones you don't want. Would be especially nice if a real Riivolution configuration worked on Dolphin too.

yes it saves it there

IonicPixels:
> Hm, yet another small thing: Riivolution, I believe, saves your configuration in `SD:/riivolution/config` (disclaimer I didn't check whether or not this was true but I think it is lol). If you have like 20 mods for a game and the default setting for them is all "enabled" then with this you currently have to, every single time you start with Riivolution, disable all of the ones you don't want. Would be especially nice if a real Riivolution configuration worked on Dolphin too.

I agree with this, I kind of forgot that you made this comment. however yes, the actual Riivolution saves to `SD:/riivolution/config` and as you said it would be really nice, it would also be nice if you could make it only enabled forever until the user disables it on certain game directories though.

iwubcode:
> The Riivolution window remembering your last selected patches

Just a thought.  We already have this concept of a "profile" system for inputs.  Feels like the json file you have for the command-line support could be thought of as a profile.  It's a system your average user is at least familiar with.  Having this as a profile would also mean you could rather easily load various 'flavors' of mods (with different options set).  If you went with this approach, I can think of a few other nice to haves:

* Drag and drop the json file on the Dolphin window should launch the game with those mods (as if launched from command line)
* The UI (that exists today?) should allow for loading and saving the profile, just like inputs
* We could add a game setting that sets the default "profile" (just like we can with inputs)

AdmiralCurtiss:
@Meatball132 Dolphin now acts more sensibly when you load an XML from an arbitrary location, so your use case of 'extract NewerSMBW.zip whereever and point Dolphin at it' should work now. It should also now remember your last selected patches in the same format actual Riivolution does.

@iwubcode My idea was to add a button next to Start that's like, 'save this config as a preset' which generates the .json file that can then be command line booted and shows up in the game list and for Netplay and the like. Having this as part of a profile-like system might work too. Though I would prefer to put this in a follow-up PR, since this one is getting a bit bloated...

AdmiralCurtiss:
@JosJuice  I did *most* of your things, except:

- The FSTBuilderNode filename I just made store UTF8 and convert to Shift-JIS in BuildFST(). This is also better for the Riivolution filename comparisons, which were in a bit of an undefined encoding state before.
- I'm not confident that treating the game partition like a normal partition actually works, because I need its header for SetNonpartitionDiscHeader(), and I'm not sure if that's safe to move after SetPartition()... Not to mention I still need to know which partition is the game partition *anyway* for the FST patching.

AdmiralCurtiss:
@JosJuice Did most of your things, except:
- Switching `IOFile` to `FileInfo` in RiivolutionPatcher when we only need the existence and filesize of a file made Dolphin crash for me for unclear reasons, I'll need to investigate this further.
- The potential existing `main.dol` in the root dir is not done yet.
- The more-than-two-dot directory traversal is not done yet -- need to test if Riivolution itself even allows this first.

@leoetlino Did all of your things. For the record, `ParseGameModDescriptorFile()` takes `const std::string&` because it's immediately passed to an IOFile, which does not have a string_view constructor.

AdmiralCurtiss:
As for everyone else, I've modified the memory patching logic to me more accurate to the convoluted mess Riivolution itself appears to use. It may be useful to re-test all mods that use Search or Ocarina memory patches and report if there are any issues.

N-I-N-0:
Hi 👋

I am a New Super Mario Bros. Wii modder and I noticed that this (really cool) feature seems to be slightly incomplete. Multiple mods make use of the ability to write to (e.g.) the SD card which is a feature provided by Riivolution using something called ProxiIOS. This contains code for mounting the SD card and writing to it using FAT, allowing mods such as Newer Super Mario Bros. Wii to interact with the SD card. In NewerSMBW this is relavant for replay recordings you can see on the title screen of such mods. However, at the moment this can only be done on a real Wii due to this behaviour not being implemented in Dolphin.

A tutorial on how to do replay recording in Newer can be found here: https://horizon.miraheze.org/wiki/Recording_Replays

Information below may be incomplete or partially wrong, since I don't want to claim I have a full understanding of how Riivolution works.
Regardless, to go a bit more in-depth about how this seems to work: Riivolution patches strings to certain devices such as "/dev/usb/hid" to strings like "/dev/usb/hi0" in the game code and runs an altered IOS that can handle IOCTLV calls with those new devices opened as well as intercepting IOS_Open calls to new files on the filesystem.
The patching of the game happens here: https://github.com/AerialX/rawksd/blob/4ef79e2e234d7c2f19d030c2d9440852b5c73b7a/launcher/source/launcher.cpp#L442
The ProxiIOS code is spread accross many files but the part handling the filesystem access (e.g. over SD/USB and even network) is handled here it seems: https://github.com/AerialX/rawksd/blob/4ef79e2e234d7c2f19d030c2d9440852b5c73b7a/filemodule/source/module.cpp#L17

The relevant code mods use to actually write files to e.g. a SD card can be found here: https://github.com/Newer-Team/NewerSMBW/blob/cw/Kamek/src/replay.S#L143

When the same code is run on Dolphin, we are sadly incapable of getting it to write to an inserted virtual SD card, forcing people to use a Wii or other means to obtain such recordings, highly slowing down development speed of such projects in this regard.

With this I want to ask whether there is any chance the altered behaviour Riivolution provides for games could be implemented into Dolphin as well, so that mods relying on this feature would work the same on Dolphin. If this feature could be used reliably both on real hardware and Dolphin, mods be able to implement new features such as creating screenshots stored on a SD card based on this! @AdmiralCurtiss

RoadrunnerWMC:
([Issue about this missing feature on the bug tracker](https://bugs.dolphin-emu.org/issues/13145))

PR #10128, opened 2021-09-26
TASCheckBox: Show checked/unchecked on turbo button press

malleoz:
In the current state of Dolphin dev, a tristate is used in which the checkbox has a square to represent the controller button is in turbo mode. This is what it looks like to turbo currently.

Alt Text

It is better to be able to visualize what inputs are going to be pressed as you frame advance.

This PR references the OnFrameEnd() function in Core.cpp. It then runs a Host function to emit a signal which is picked up by GCTASInputWindow and WiiTASInputWindow. Then, by referencing the turbo press/release durations, it determines whether to check or uncheck each input that is in turbo mode.

This is the new functionality.

Alt Text

I'd like to clarify what might "appear" like a one frame delay compared to TAS Input and the game renderer... I feel that it is more beneficial to, at any given time, reference TAS Input to see what inputs will be activated upon advancing into the next frame. Thus, if a button becomes unchecked on frame 0, this represents that the input will deactivate in-game on frame 1.

I noticed an issue where setting a turbo, closing TAS Input window, frame advancing, then re-opening TAS Input window would cause there to be an offset; even though the input might toggle on frame 2, the TAS Input window might not reflect that until frame 4.

Rather than fix that issue in this PR, I intend to reset all aspects of TAS Input windows upon window close. I will do so in a different PR, since that's a bit out-of-scope from what this PR intends to add.


JMC47:
@dolphin-emu-bot rebuild

JMC47:
@dolphin-emu-bot rebuild

phire:
Bold for turbo is a little non-obvious.

PR #10129, opened 2021-09-27
QFileDialog: differentiate between recording open/save

malleoz:
Previously, the file dialog window was ambiguous between saving or loading a .dtm. This commit simply gives a bit more context to differentiate the two windows.


leoetlino (approved):
LGTM (other than the lint failure)

PR #10130, opened 2021-09-27
Core: fix hotkeys not being recognized on game list

malleoz:
Previously I had PR #10122 merged in which the HotkeyScheduler was changed to check for certain hotkeys before checking if the core is running and started.

However, those hotkeys do not pass the focus_passes check. It appears I mistakenly had "Hotkeys Require Window Focus" disabled and failed to identify that...

As such, I've added a case to pass the focus check if the core is not running and the main window has focus.

This appears to resolve the issue.

I was not able to identify any negative implications of this change. Please let me know if you find this is not the case.

NOTE: this will conflict with #10125 and will need to be resolved upon its merge.

### DIFF
--- a/Source/Android/jni/MainAndroid.cpp
+++ b/Source/Android/jni/MainAndroid.cpp
@@ -152,6 +152,11 @@ bool Host_RendererHasFocus()
   return true;
 }
 
+bool Host_MainWindowHasFocus()
+{
+  return false;
+}
+
 bool Host_RendererHasFullFocus()
 {
   // Mouse cursor locking actually exists in Android but we don't implement (nor need) that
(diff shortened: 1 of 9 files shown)


malleoz:
In order for this commit to work on other OSes besides Windows, I'm having trouble figuring out how to determine if the MainWindow is in focus... I'm getting unresolved external symbols when I try to use focusInEvent() and focusOutEvent(), which probably isn't even a good approach either... Some advice would be appreciated.

Pokechu22:
@dolphin-emu-bot rebuild

Pokechu22:
@dolphin-emu-bot rebuild

PR #10138, opened 2021-09-30
Fix AbortOnPanicAlert with PanicAlertFmt

Pokechu22:
PR #10066 added functionality to call std::abort when a panic alert occurs; however, that PR only implemented it for `MsgAlert` and not `MsgAlertFmtImpl`, meaning that the functionality was not used with `PanicAlertFmt` (only `PanicAlert`, which is not used frequently).

### DIFF
--- a/Source/Core/Common/MsgHandler.cpp
+++ b/Source/Core/Common/MsgHandler.cpp
@@ -106,20 +106,10 @@ std::string GetStringT(const char* string)
   return s_str_translator(string);
 }
 
-// This is the first stop for gui alerts where the log is updated and the
-// correct window is shown
-bool MsgAlert(bool yes_no, MsgType style, const char* format, ...)
+static bool ShowMessageAlert(std::string_view text, bool yes_no, MsgType style)
 {
-  // Read message and write it to the log
   const char* caption = GetCaption(style);
-  char buffer[2048];
-
-  va_list args;
-  va_start(args, format);
-  CharArrayFromFormatV(buffer, sizeof(buffer) - 1, s_str_translator(format).c_str(), args);
-  va_end(args);
-
-  ERROR_LOG_FMT(MASTER_LOG, "{}: {}", caption, buffer);
+  ERROR_LOG_FMT(MASTER_LOG, "{}: {}", caption, text);
 
   // Panic alerts.
   if (style == MsgType::Warning && s_abort_on_panic_alert)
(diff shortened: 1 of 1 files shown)


Pokechu22:
`fifa-street` and `medabots-crash` now fail to render.  `fifa-street` is failing here:

https://github.com/dolphin-emu/dolphin/blob/a7224b299c609a5d5ee803610f1fbedea5fb0a30/Source/Core/VideoCommon/CPMemory.h#L108-L124

The normal format and the format of both colors is set to 5, which is invalid (the other components are set to 4, indicating float).  `medabots-crash` has invalid pointer errors.  So, them failing to render is an expected result and should be fixed by resolving the underlying bugs.

phire (approved):

phire:
Do you want to fix the FIFO_CI failures before merging this?

Pokechu22:
I don't think they have an easy fix, so I'm fine with leaving them erroring for now until I have time to research them further.

iwubcode (approved):

Source/Core/Common/MsgHandler.cpp
```diff
@@ -106,20 +106,9 @@ std::string GetStringT(const char* string)
   return s_str_translator(string);
 }
 
-// This is the first stop for gui alerts where the log is updated and the
-// correct window is shown
-bool MsgAlert(bool yes_no, MsgType style, const char* format, ...)
+static bool ShowMessageAlert(const char* caption, const char* text, bool yes_no, MsgType style)
```
leoetlino, line 109: Could we take a `std::string_view` instead of a `const char*` for `text`? (so fmt doesn't have to recalculate the string length unnecessarily)

Pokechu22:
@dolphin-emu-bot rebuild

Source/Core/Common/MsgHandler.cpp
```diff
@@ -131,26 +120,36 @@ bool MsgAlert(bool yes_no, MsgType style, const char* format, ...)
...
-
-  // Don't ignore questions, especially AskYesNo, PanicYesNo could be ignored
-  if (s_msg_handler != nullptr &&
-      (s_alert_enabled || style == MsgType::Question || style == MsgType::Critical))
-  {
-    return s_msg_handler(caption, message.c_str(), yes_no, style);
-  }
 
-  return true;
+  return ShowMessageAlert(caption, message.c_str(), yes_no, style);
```
Pokechu22, line 153: The use of `message.c_str()` means the previous `string_view` change doesn't do anything.  Will replace with just `message`.

leoetlino (approved):

PR #10140, opened 2021-09-30
DTM: add timebase as desync-prevention method

malleoz:
**NOTE: This adds extra data to the .dtm format. In the event this PR is merged, DTMs from prior revisions will become pretty much incomprehensible after the header.**

This PR aims to implement an idea introduced by JMC47. By storing the timebase on each input poll and comparing it with the timebase occurred during playback, we can quickly observe the exact point in-game that caused a desync in movie playback.

On every input poll, the `u64 timebase` is read by `Movie::InputUpdate()` and stored to `s_temp_input`. On state load, the dtm is compared with the savestate movie and looks for mismatched timebases to indicate that a desync has occured. On desync detection, a panic alert gives a warning to not continue recording beyond this point in the movie. More on this later.

During playback, to allow TASers to pinpoint the exact frame that they experience a desync, the timebase is checked on every input poll. On mismatch detected (once per playback session, as otherwise this warning would occur every frame going forward), the user is prompted with a warning dialog advising them to not continue.

A few comments:

**Why do we need to write the entire u64 timebase? Why not just the lowest byte, which would catch >99% of desyncs?**
Yes, u64 is overkill in this aspect. But, quite a number of TASers often like to hex edit, that is, manually modify the dtm file. Yes, most hex editors allow you to resize the number of byte columns shown, but because the header is 0x100 in size, input sequences of 8 bytes + the size of the timebase will become misaligned unless timebase+inputs = 2<sup>n</sup> bytes long. If timebase is < 8 bytes long, then this would make hex editing much more tedious and less intuitive.

From a filesize perspective, my 2 hour TTYD TAS is only 9MB and zips to 53KB. The filesize increase of 100% due to adding timebase thus should not be significant enough to warrant reducing the number of bytes we save.

**Why write timebase on every input poll?**
Well the true granularity of data in the .dtm is input polls, which constitute 8 bytes each. It seems like a very poor idea to try to write the timebase on every frame, when there may not always be 2 input polls per frame. After all, it's much easier to explain "each input poll is 16 bytes", rather than "each frame is on average 16 bytes, followed by 8 bytes". Thus, this means the overall file layout stays uniform.

**If a TASer wants to hex edit, then their timebase will obviously become incorrect at some point.**
I imagine that the best way to handle this is the fact that the user can always press to ignore the warning for the current session, continue onward, and then overwrite the movie with latest changes. In doing so, the movie will now reflect the correct timebases.

There's been a bit of a discussion about a DTM 2.0 version, which begs the question: should this PR be merged before there is finalized discussion about what other changes should be made to the format? Is it okay to make these changes independently? Just food for thought.

_A change like this would truly introduce a life-saving safety net for TASers. I would love to see this added in the near future._

I'd appreciate any questions you may have, and any cleanup suggestions. Thanks!

### DIFF
--- a/Source/Core/Core/Movie.cpp
+++ b/Source/Core/Core/Movie.cpp
@@ -70,9 +70,6 @@
 #include "VideoCommon/VideoBackendBase.h"
 #include "VideoCommon/VideoConfig.h"
 
-// The chunk to allocate movie data in multiples of.
-#define DTM_BASE_LENGTH (1024)
-
 namespace Movie
 {
 using namespace WiimoteCommon;
@@ -278,6 +275,11 @@ void InputUpdate()
     s_totalInputCount = s_currentInputCount;
     s_totalTickCount += CoreTiming::GetTicks() - s_tickCountAtLastInput;
     s_tickCountAtLastInput = CoreTiming::GetTicks();
+
+    u64 timebase = SystemTimers::GetFakeTimeBase();
+    s_temp_input.resize(s_currentByte + sizeof(u64));
+    memcpy(&s_temp_input[s_currentByte], &timebase, sizeof(u64));
+    s_currentByte += sizeof(u64);
   }
 }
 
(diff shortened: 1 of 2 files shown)


Source/Core/Core/Movie.cpp
```diff
@@ -72,6 +72,8 @@
 
 // The chunk to allocate movie data in multiples of.
 #define DTM_BASE_LENGTH (1024)
+#define DTM_GC_FRAME_LENGTH (sizeof(ControllerState) + sizeof(u64))
+#define DTM_WII_FRAME_LENGTH (size + sizeof(u64))
```
lioncash, line 76:
```suggestion
static constexpr size_t DTMWiiFrameLength(size_t size) {
  return size + sizeof(u64);
}
```

Otherwise this is implicitly dependent on `size` and can break if code is ever moved around in the future (this function should also be placed in the Movie namespace)

Source/Core/Core/Movie.cpp
```diff
@@ -72,6 +72,8 @@
 
 // The chunk to allocate movie data in multiples of.
 #define DTM_BASE_LENGTH (1024)
+#define DTM_GC_FRAME_LENGTH (sizeof(ControllerState) + sizeof(u64))
```
lioncash, line 75: This and DTM_BASE_LENGTH can be converted into `constexpr size_t` constants (and can also be moved into the Movie namespace)

Dentomologist:
@dolphin-emu-bot rebuild

Source/Core/Core/Movie.cpp
malleoz: Changes applied. Also noting that I am removing `DTM_BASE_LENGTH` (I didn't create it), as it is never referenced.

Source/Core/Core/Movie.h
```diff
@@ -138,6 +138,13 @@ static_assert(sizeof(DTMHeader) == 256, "DTMHeader should be 256 bytes");
 
 #pragma pack(pop)
 
+static constexpr size_t DTMWiiFrameLength(size_t size)
+{
+  return size + sizeof(u64);
+}
+
+static constexpr size_t DTMGCFrameLength = sizeof(ControllerState) + sizeof(u64);
```
lioncash, line 146:
```suggestion
constexpr size_t DTMGCFrameLength = sizeof(ControllerState) + sizeof(u64);
```

Source/Core/Core/Movie.cpp
```diff
@@ -1161,7 +1185,26 @@ void PlayController(GCPadStatus* PadStatus, int controllerID)
   }
 
   memcpy(&s_padState, &s_temp_input[s_currentByte], sizeof(ControllerState));
-  s_currentByte += sizeof(ControllerState);
+<<<<<<< HEAD
```
lioncash, line 1188: Seems like an incomplete rebase occurred here

malleoz: You would be correct... Fixed and force pushing.

JosJuice:
> There's been a bit of a discussion about a DTM 2.0 version, which begs the question: should this PR be merged before there is finalized discussion about what other changes should be made to the format? Is it okay to make these changes independently? Just food for thought.

My vague ideas for DTM 2.0 include turning it into a zip file that can contain various things. We could store settings as an INI file instead of having to deal with a manually designed struct (this way adding new settings would be almost effortless), include memory cards and NAND data, SRAM data, and so on. This would be a rather big change and I'm fine with releasing the change made in this PR as a kind of "DTM 1.1" in the meantime.

malleoz:
Accidentally left in incorrect incrementation to s_currentByte in PlayRecording() which occurred when I was fixing the rebase issue that lioncash pointed out above. This is resolved now.

PR #6325, opened 2018-01-25
Remote debugging

revel8n:
Added extended support for connecting to Dolphin and debugging remotely through other applications.
- Extended breakpoint support to keep track of last reason execution paused
- Added GDBThread class, an extended form of the GDBStub implementation that was already present, in order to expose more elements of the emulation state to remote clients as well as integrate with both Interpreter and JIT runtimes
- Added TCPGecko class, an implementation of the USBGecko protocol over TCP, integrated more directly with the emulator

Wanted to see if this was an interesting addition for others, as well as possibly discuss any improvements to the implementation to better integrate with Dolphin in either case.

In particular:
- Currently remote clients can access the full memory range, even some addresses that cause PanicAlerts (left that way so as to not compromise those checks).  Only real drawback of this is the MessageBox loop until the read it finished.  Can probably work around this to just return what it would have without the PanicAlerts in place, or simply limit the remote process to the acceptable ranges of memory and read/write larger chunks of data at once instead of byte by byte.

There are two companion projects:
- dolphindbg (https://github.com/revel8n/dolphindbg) which is a plugin for IDA Pro that allows connections to the GDB interface
- geckowii (https://github.com/revel8n/geckowii) modified version of Gecko.Net which supports connecting over TCP, and works with the TCP interface provided

i hope to add plugins for other debuggers as well as extend the plugins to work with actual hardware where applicable.

Any thoughts or other information you have would be greatly appreciated.

PR #10142, opened 2021-10-01
Rework the entire logic of the GDB stub

aldelaro5:
The GDB stub is an old and very unmaintained feature that can be very useful when paired with reversing tools like ghidra, but being unmaintained, it didn't worked very well (and only worked on the interpreter).

This PR does the ground work to rework it entirely to at least a usable state. Here's a summary of the changes:

- The stub process commands logic now runs every so often while the emulation is running to process potential ctrl+c commands
- The main gdb stalling logic is now in the CPU::Run instead of the interpreter whenever gdb still has control (basically, after a ctrl+c or breakpoint, but not user pause). This effectively means all CPU backend are now supported.
- If active, boot to pause is effectively forced until we get a connection and a continue command (this behavior may be changed in another PR, but the focus of this one is to just make it work).
- Completely rework the breakpoint processing logic to use the same functions than the integrated debugger instead of tracking them separately since the integrated debugger already works very well
- Add support for the T command that checks if a thread is alive (this is something ghidra wanted for mgba and it's trivial to support so might as well)
- Allow the stop button to kill the gdb connexion with it
- Remove the cmake option and the ifdefs, the code is now always enabled
- Rework the logs (upgraded some debug to info and added one when we break due to GDB)
- HUGE refactor of the code as it was far from respecting the current coding style

To note: the hardest part was figuring out how to coordinate this with the core safely: I am not experienced with multithreading stuff so when reviewing, I recommend to pay close attention to how this interacts with the core and to make sure there's no race or anything unsafe I am doing here. This pr is mostly to make sure the approach done here is correct for potential further improvements.

For now, I can vouch for a gdb shell to work (mostly tested on the one provided by devkitppc, but gdb-multiarch seemed to work well), but ghidra seems to have issues of its own that doesn't seem to be in the scope of Dolphin.


Source/Core/Core/PowerPC/GDBStub.cpp
```diff
@@ -564,103 +494,106 @@ static void gdb_write_register()
...
-  DEBUG_LOG_FMT(GDB_STUB, "gdb: added {} breakpoint: {:08x} bytes at {:08x}", type, bp->len,
-                bp->addr);
+  if (type == GDB_BP_TYPE_X)
+  {
+    PowerPC::breakpoints.Add(addr);
+    INFO_LOG_FMT(GDB_STUB, "gdb: added {} breakpoint: {:08x} bytes at {:08x}", type, len, addr);
+  }
+  else
+  {
+    TMemCheck newMemcheck;
```
lioncash, line 576:
```suggestion
    TMemCheck new_memcheck;
```

Source/Core/Core/PowerPC/GDBStub.cpp
```diff
@@ -714,108 +647,125 @@ static void gdb_remove_bp()
...
+    len = (len << 4) | Hex2char(s_cmd_bfr[i++]);
 
-  gdb_bp_remove(type, addr, len);
-  gdb_reply("OK");
+  RemoveBreakpoint(type, addr, len);
+  SendReply("OK");
 }
 
-void gdb_handle_exception()
+void ProcessCommands(bool loopUntilContinue)
```
lioncash, line 668:
```suggestion
void ProcessCommands(bool loop_until_continue)
```

Source/Core/Core/PowerPC/GDBStub.cpp
```diff
@@ -826,123 +776,87 @@ void gdb_init(u32 port)
...
-  return gdb_bp_check(addr, GDB_BP_TYPE_W);
-}
-
-int gdb_bp_a(u32 addr)
-{
-  if (sock == -1)
-    return 0;
-
-  return gdb_bp_check(addr, GDB_BP_TYPE_A);
+  char bfr[128];
```
lioncash, line 857:
```suggestion
  char bfr[128] = {};
```

The memset below can be removed

Source/Core/Core/PowerPC/GDBStub.cpp
```diff
@@ -826,123 +776,87 @@ void gdb_init(u32 port)
...
+  memset(bfr, 0, sizeof bfr);
+  sprintf(bfr, "T%02x%02x:%08x;%02x:%08x;", signal, 64, PC, 1, GPR(1));
```
lioncash, line 859:
```cpp
fmt::format_to(bfr, T{:02x}{:02x}:{:08x};{:02x}:{:08x}, signal, 64, PC, 1, GPR(1));
```

Source/Core/Core/PowerPC/GDBStub.h
```diff
@@ -26,18 +25,17 @@ typedef enum
...
-void gdb_init(u32 port);
-void gdb_init_local(const char* socket);
-void gdb_deinit();
-bool gdb_active();
-void gdb_break();
+const s64 GDB_UPDATE_CYCLES = 100000;
 
-void gdb_handle_exception();
-int gdb_signal(u32 signal);
+void UpdateCallback(u64 userdata, s64 cyclesLate);
```
lioncash, line 30:
```suggestion
void UpdateCallback(u64 userdata, s64 cycles_late);
```

Source/Core/Core/PowerPC/GDBStub.h
```diff
@@ -26,18 +25,17 @@ typedef enum
...
-int gdb_bp_a(u32 addr);
+void Init(u32 port);
+void InitLocal(const char* socket);
+void Deinit();
+bool IsActive();
+bool HasControl();
+void TakeControl();
 
-bool gdb_add_bp(u32 type, u32 addr, u32 len);
+void ProcessCommands(bool loopUntilContinue);
```
lioncash, line 39:
```suggestion
void ProcessCommands(bool loop_until_continue);
```

Warepire:
> * If active, boot to pause is effectively forced until we get a connection and a continue command (this behavior may be changed in another PR, but the focus of this one is to just make it work).

This is how gdbserver feature of gdb work when used for remote debugging, if this functionality is not in the way it might be a good idea to keep, as it will work like most gdb users expect out of the box.

aldelaro5:
> > * If active, boot to pause is effectively forced until we get a connection and a continue command (this behavior may be changed in another PR, but the focus of this one is to just make it work).
> 
> This is how gdbserver feature of gdb work when used for remote debugging, if this functionality is not in the way it might be a good idea to keep, as it will work like most gdb users expect out of the box.

The idea that me and @phire were talking about were you would start the emu normally WITHOUT the stub, but have the port in the settings and you would use some sort of menu option to start it so it would effectively stall AFTER the emulation started. This is actually how mgba works: you can start and stop it on demand.

It's fine for now if it only works on boot, but I was just saying we might improve it in another pr to not force it on boot.

aldelaro5:
Ok I addressed the review, fixed compilation and also added a race condition fix I found while testing on windows.

phire:
> Does this PR make #6325 obsolete?

Looks like it mostly obsoletes it, and does so with much less code and without an entire thread.

PR #10143, opened 2021-10-02
Add option for setting the PNG zlib compression level

Pokechu22:
When using large internal resolutions, it can take a while for the image to be compressed.  This PR adds a setting to change the zlib compression level, which enables a trade-off between smaller files and faster saving.  (Since PNG is lossless, the image quality is always the same).

Libpng notes this about the compression levels:

https://github.com/dolphin-emu/dolphin/blob/890a5ed99ad4130d68b3e7228c3398445c33cdc4/Externals/libpng/png.h#L1500-L1509

Libpng presumably uses zlib's default of 6 by default:

https://github.com/dolphin-emu/dolphin/blob/890a5ed99ad4130d68b3e7228c3398445c33cdc4/Externals/zlib/zlib.h#L235-L239 

Though libpng also does have a "fast" option which uses a value of 3:

https://github.com/dolphin-emu/dolphin/blob/890a5ed99ad4130d68b3e7228c3398445c33cdc4/Externals/libpng/pngwrite.c#L2104-L2116

Here are some results using frame 0 of the [Melty Molten Galaxy fifolog](https://fifo.ci/media/dff/MeltyMoltenGalaxy.dff) with arbitrary mipmap detection enabled at auto aspect ratio and an internal resolution of 16x (producing a 12968 by 7296 image):

|Level|Size (bytes)|Time (s)|Size diff|Time diff|
|:---:|-----------:|--------|--------:|--------:|
|  0  | 284,310,113|0:02.942|100%/100%|1.0x/1.0x|
|  1  |  60,348,200|0:04.437|21.2%/21%|1.5x/1.5x|
|  2  |  57,772,001|0:05.307|20.3%/96%|1.8x/1.2x|
|  3  |  54,902,451|0:07.170|19.3%/95%|2.4x/1.4x|
|  4  |  49,811,855|0:07.917|17.5%/91%|2.7x/1.1x|
|  5  |  48,492,485|0:10.111|17.1%/97%|3.4x/1.3x|
|  6  |  47,381,039|0:17.710|16.7%/98%|6.0x/1.8x|
|  7  |  46,835,815|0:29.212|16.5%/99%|9.9x/1.6x|
|  8  |  45,792,636|1:15.136|16.1%/98%|25.5x/2.6x|
|  9  |  45,346,752|2:18.179|15.9%/99%|47.0x/1.8x|

The size diff and time diff values are comparisons to level 0 followed by comparisons to the previous level.

### DIFF
--- a/Source/Core/Common/Image.cpp
+++ b/Source/Core/Common/Image.cpp
@@ -10,6 +10,9 @@
 
 #include "Common/CommonTypes.h"
 #include "Common/IOFile.h"
+#include "Common/ImageC.h"
+#include "Common/Logging/Log.h"
+#include "Common/Timer.h"
 
 namespace Common
 {
@@ -39,23 +42,28 @@ bool LoadPNG(const std::vector<u8>& input, std::vector<u8>* data_out, u32* width
   return true;
 }
 
-bool SavePNG(const std::string& path, const u8* input, ImageByteFormat format, u32 width,
-             u32 height, int stride)
+static void WriteCallback(png_structp png_ptr, png_bytep data, size_t length)
 {
-  png_image png = {};
-  png.version = PNG_IMAGE_VERSION;
-  png.width = width;
-  png.height = height;
+  std::vector<u8>* buffer = static_cast<std::vector<u8>*>(png_get_io_ptr(png_ptr));
+  buffer->insert(buffer->end(), data, data + length);
+}
+
+bool SavePNG(const std::string& path, const u8* input, ImageByteFormat format, u32 width,
+             u32 height, int stride, int level)
+{
+  Common::Timer timer;
+  timer.Start();
 
   size_t byte_per_pixel;
+  int png_format;
   switch (format)
   {
   case ImageByteFormat::RGB:
-    png.format = PNG_FORMAT_RGB;
+    png_format = PNG_FORMAT_RGB;
     byte_per_pixel = 3;
     break;
   case ImageByteFormat::RGBA:
-    png.format = PNG_FORMAT_RGBA;
+    png_format = PNG_FORMAT_RGBA;
     byte_per_pixel = 4;
     break;
   default:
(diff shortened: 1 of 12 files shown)


Source/Core/Common/Image.cpp
```diff
@@ -39,23 +41,52 @@ bool LoadPNG(const std::vector<u8>& input, std::vector<u8>* data_out, u32* width
...
+
+bool SavePNG0(png_structp png_ptr, png_infop info_ptr, int png_format, u32 width, u32 height,
+              int level, std::vector<u8>* buffer, const std::vector<const u8*>& rows) noexcept
+{
+#ifdef _MSC_VER
+#pragma warning(push)
+// Disable warning for setjmp in C++
+#pragma warning(disable : 4611)
+#endif
+  if (setjmp(png_jmpbuf(png_ptr)) != 0)
```
shuffle2, line 58: i seem to remember purposefully getting rid of setjmp usage in libpng-related code before; is it really required?

shuffle2: The side effects (and reason for warning) can be pretty subtle and easy to get wrong.

Pokechu22: The code previously used the simplified libpng API (`png_image_write_to_memory`); as far as I can tell there is no way to set the compression level with that API.

`png_image_write_to_memory` [calls](https://github.com/dolphin-emu/dolphin/blob/890a5ed99ad4130d68b3e7228c3398445c33cdc4/Externals/libpng/pngwrite.c#L2242) `png_safe_execute`, which [uses](https://github.com/dolphin-emu/dolphin/blob/890a5ed99ad4130d68b3e7228c3398445c33cdc4/Externals/libpng/pngerror.c#L945) `setjmp` internally.  So `setjmp` was being used before, though not directly in Dolphin code.

I did try to isolate the `setjmp` call to its own function and have the C++ objects be constructed and destructed outside of that function, but I'm not sure if that's guaranteed to be safe (either by the standard or in practice due to inlining).

AdmiralCurtiss: Since I'm the one who ported this over to the simplified API: IMO if you want to use the complex API, we should make a separate C (not C++) compilation unit and use the `setjmp` there. Otherwise this is inviting danger for anyone who touches the code in the future.

shuffle2: If there's a way to write it in C and have the write callback also be C that would be nice. It's not a huge deal to me either way but just a bit of unfortunate code pattern by libpng :/

Pokechu22: I've moved `SavePNG0` into a C file (which also required disabling precompiled headers for that file).

According to [cppreference on `longjmp`](https://en.cppreference.com/w/cpp/utility/program/longjmp),

> No destructors for automatic objects are called. If replacing of `std::longjmp` with `throw` and `setjmp` with `catch` would execute a non-trivial destructor for any automatic object, the behavior of such `std::longjmp` is undefined. 

Since `longjmp` isn't called by the callback, I think it's fine for the callback to be in C++ code.  I also don't want to try and port the old buffer-resizing code to C, partially because I think it'd be fiddly to do and partially because if the buffer ends up being too small, I don't want to run the compression process a second time (though that probably won't happen often for higher compression levels).

shuffle2: The reason I suggested that the callback be C is because std::vector::insert can throw (which comes with it's own problems but might also cause dtors to be hit on paths you wouldn't expect). But, dolphin doesn't care about such exceptions at all anyway, so 🤷‍♂️. It's fine how it is imo but just FYI

leoetlino (approved):

PR #10144, opened 2021-10-02
Port Main.DSP to OnionConfig

malleoz:
While trying to work on adding audiodump support for CLI, I was alerted that it was important to first try moving the DSP configs to the new OnionConfig before continuing, as that makes it substantially easier to write clean code to add such a feature.

This PR aims to allow for Dolphin to only rely on OnionConfig for DSP-related settings.

I am not very familiar with the new config, so please let me know if I missed any concepts in this PR. Notably, I was unsure what to do in BootManager.cpp `BootCore()` in the case of netplay... Seems like it uses its own SConfig which checks for DSP JIT.


Source/Core/Core/ConfigLoaders/IsSettingSaveable.cpp
```diff
@@ -72,6 +72,20 @@ bool IsSettingSaveable(const Config::Location& config_location)
...
+      &Config::MAIN_DSP_JIT.GetLocation(),
+      &Config::MAIN_DSP_CAPTURE_LOG.GetLocation(),
+      &Config::MAIN_DUMP_AUDIO_SILENT.GetLocation(),
+      &Config::MAIN_DUMP_UCODE.GetLocation(),
+      &Config::MAIN_AUDIO_BACKEND.GetLocation(),
+      &Config::MAIN_AUDIO_VOLUME.GetLocation(),
+      &Config::MAIN_AUDIO_MUTED.GetLocation(),
+#ifdef _WIN32
+      &Config::MAIN_WASAPI_DEVICE.GetLocation(),
+#endif
```
JosJuice, line 87: Instead of adding all of these, you could add "DSP" to the list of sections at line 29.

Also, I think you've forgotten MAIN_DSP_HLE.

Source/Core/Core/ConfigLoaders/NetPlayConfigLoader.cpp
```diff
@@ -47,6 +47,7 @@ class NetPlayConfigLayerLoader final : public Config::ConfigLayerLoader
     layer->Set(Config::MAIN_FALLBACK_REGION, m_settings.m_FallbackRegion);
     layer->Set(Config::MAIN_ALLOW_SD_WRITES, m_settings.m_AllowSDWrites);
     layer->Set(Config::MAIN_DSP_JIT, m_settings.m_DSPEnableJIT);
+    layer->Set(Config::MAIN_DSP_HLE, m_settings.m_DSPHLE);
```
JosJuice, line 50: There's already a line that sets this.

JosJuice:
@dolphin-emu-bot rebuild

Source/Core/Core/ConfigLoaders/IsSettingSaveable.cpp
JosJuice: Actually, speaking of DSP HLE, you don't seem to have removed that one from ConfigManager.cpp. Please be consistent with whether you're intending to port that one over or not. If it's not going to be in the new config system, then you shouldn't be adding calls to `Config::Get(Config::MAIN_DSP_HLE)`.

JosJuice:
Also, the new config system is not officially called OnionConfig. But this is just me being opinionated and not something that necessarily should block the PR :)

Source/Core/Core/Config/MainSettings.cpp
```diff
@@ -129,13 +128,18 @@ const Info<bool> MAIN_DISABLE_SCREENSAVER{{System::Main, "Display", "DisableScre
 // Main.DSP
 
 const Info<bool> MAIN_DSP_CAPTURE_LOG{{System::Main, "DSP", "CaptureLog"}, false};
+const Info<bool> MAIN_DSP_HLE{{System::Main, "DSP", "DSPHLE"}, true};
```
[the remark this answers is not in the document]
malleoz: It was originally in core prior to this commit, but I thought, given that fact that MAIN_DSP_JIT was under DSP category, that it was fitting to move HLE to this section too. I might be misunderstanding something though.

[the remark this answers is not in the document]
JosJuice: Changing which INI section a setting is stored in breaks existing configs, which should not be done unless there's some particularly important reason. Had you been adding a new setting then I agree that putting it in DSP makes sense, but this little mismatch isn't important enough to warrant breaking configs.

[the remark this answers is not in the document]
malleoz: Ah I see. I will move it back to Core, where it was originally.

Source/Core/Core/Config/MainSettings.h
```diff
@@ -9,6 +9,8 @@
 #include "Common/Config/Config.h"
 #include "DiscIO/Enums.h"
 
+#define BACKEND_NULLSOUND _trans("No Audio Output")
```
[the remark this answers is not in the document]
malleoz: It's referenced in AudioCommon.cpp which was originally using it from the old config file... But I now realize I should've moved all the others over... Not sure why I didn't.

Source/Core/Core/ConfigLoaders/IsSettingSaveable.cpp
malleoz: Added DSP to line 29, thanks for that cleanup. I had forgotten MAIN_DSP_HLE, so I've added that in `s_setting_saveable` under Main.Core.

You're right, I hadn't removed it from ConfigManager.cpp but I've done so now. There was an SConfig function dependent on bDSPHLE and bDPL2Decoder, so I ported both over now.

Source/Core/Core/Config/MainSettings.h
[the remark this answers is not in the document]
JosJuice: Ah, I didn't think of the fact that it was in ConfigManager.h originally. Yes, in that case I think it would make sense to move them all here.

malleoz:
Implemented Jos's review notes. Kept `MAIN_DSP_HLE` in Core so as to not break previous config files, finished porting over DSP HLE to use the new config, removed references to the new config from `ConfigCache`, and ported over `MAIN_DSP_THREAD` too.

Source/Core/AudioCommon/AudioCommon.cpp
```diff
@@ -241,7 +242,7 @@ void DecreaseVolume(unsigned short offset)
 
 void ToggleMuteVolume()
 {
-  bool& isMuted = SConfig::GetInstance().m_IsMuted;
+  bool isMuted = Config::Get(Config::MAIN_AUDIO_MUTED);
```
JosJuice, line 245: This doesn't do the same thing anymore. (Note how the reference is assigned to in the next line.) Same with `currentVolume` in two places above.

malleoz: Ahh, I understand why a reference was used here now. Will fix this.

PR #10146, opened 2021-10-02
DolphinQt/Android: Add warning when converting NKit files

JosJuice:
Yes, that's right! It's time to add even more NKit warnings, because users still don't understand what NKit is or how it works!

More specifically, some users seem to be under the impression that converting an NKit file to for instance RVZ using Dolphin's convert feature will result in a normal RVZ file, when it in fact results in an NKit RVZ file (since NKit is not a container format in the sense that GCZ/WIA/RVZ/WBFS/CISO is, but rather a kind of trimmed ISO). I can hardly blame users for not knowing this, because it's not intuitive unless you know the technical details of how NKit works.

### DIFF
--- a/Source/Core/DolphinQt/ConvertDialog.cpp
+++ b/Source/Core/DolphinQt/ConvertDialog.cpp
@@ -329,6 +329,21 @@ void ConvertDialog::Convert()
     }
   }
 
+  if (std::any_of(m_files.begin(), m_files.end(), std::mem_fn(&UICommon::GameFile::IsNKit)))
+  {
+    if (!ShowAreYouSureDialog(
+            tr("Dolphin can't convert NKit files to non-NKit files. Converting an NKit file in "
+               "Dolphin will result in another NKit file.\n"
+               "\n"
+               "If you want to convert an NKit file to a non-NKit file, you can use the same "
+               "program as you originally used when converting the file to the NKit format.\n"
+               "\n"
+               "Do you want to continue anyway?")))
+    {
+      return;
+    }
+  }
+
   QString extension;
   QString filter;
   switch (format)
--- a/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/fragments/ConvertFragment.java
+++ b/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/fragments/ConvertFragment.java
@@ -27,6 +27,7 @@ import java.io.File;
 import java.util.ArrayList;
 
 import androidx.annotation.NonNull;
+import androidx.annotation.StringRes;
 import androidx.appcompat.app.AlertDialog;
 import androidx.fragment.app.Fragment;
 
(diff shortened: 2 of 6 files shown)


Rumi-Larry:
Why not just completely disable this conversion?

JosJuice:
It does have a purpose if you have .nkit.iso and you want to convert it to something that is still NKit but takes up less space, like .nkit.gcz (which by the way is something that's explicitly supported in the NKit app). Not that I know why people would want to convert to something which still is NKit, but well, I don't think people should use NKit in general either and yet they do it :)

I suppose another reason is that I don't entirely like arbitrary limitations in Dolphin's functionality. An .nkit.iso file *is* in fact just an ISO file, and Dolphin is supposed to be able to convert ISO files. So far, everything I've done in Dolphin in regards to NKit has only been to add various kinds of warnings, not to change how Dolphin actually behaves. Though, this is a rather weak argument, I must admit.

Source/Core/DolphinQt/ConvertDialog.cpp
```diff
@@ -329,6 +329,21 @@ void ConvertDialog::Convert()
...
   }
 
+  if (std::any_of(m_files.begin(), m_files.end(), std::mem_fn(&UICommon::GameFile::IsNKit)))
+  {
+    if (!ShowAreYouSureDialog(
+            tr("Dolphin can't convert NKit files to non-NKit files. Converting an NKit file in "
+               "Dolphin will result in another NKit file.\n"
+               "\n"
+               "If you want to convert an NKit file to a non-NKit file, you can use the same "
+               "program as you originally used when converting the file to the NKit format.\n"
```
Rumi-Larry, line 339: I suggest the following wording: 

```suggestion
            tr("Dolphin cannot properly convert an NKit file to another file format."
               "Trying to, will convert it to a hybrid between NKit and the target format,"
               "that may not behave as expected.\n"
               "\n"
               "If you want to do a proper conversion, please use another program.\n"
```

Rumi-Larry:
> 
> 
> It does have a purpose if you have .nkit.iso and you want to convert it to something that is still NKit but takes up less space, like .nkit.gcz (which by the way is something that's explicitly supported in the NKit app). Not that I know why people would want to convert to something which still is NKit, but well, I don't think people should use NKit in general either and yet they do it :)
> 
> I suppose another reason is that I don't entirely like arbitrary limitations in Dolphin's functionality. An .nkit.iso file _is_ in fact just an ISO file, and Dolphin is supposed to be able to convert ISO files. So far, everything I've done in Dolphin in regards to NKit has only been to add various kinds of warnings, not to change how Dolphin actually behaves. Though, this is a rather weak argument, I must admit.

How difficult would it be to make it possible to do proper conversion of NKit files within Dolphin? It seems like the users really want NKit files no matter the cost.

Source/Core/DolphinQt/ConvertDialog.cpp
JosJuice: > Dolphin cannot properly convert an NKit file to another file format

It can properly convert an NKit file in one container format to an NKit file in another container format.

> Trying to, will convert it to a hybrid between NKit and the target format

Combining NKit with a compressed container format is not some kind of strange hybrid. .nkit.gcz is the default output of the NKit app.

> that may not behave as expected

I'd rather be clear about in what way it doesn't behave as expected, since it's simple to explain: The file will still be an NKit file. The problems you will have are the exact same as with any other NKit file.

Rumi-Larry: That's very complicated stuff, in which case I cannot think of a better wording than the one you are using. Still, it doesn't read particularly well.

JosJuice:
Much harder than the formats we already support. I'm not willing to implement it myself.

To be frank here, the only people who use NKit files and don't already know how to use the NKit app are pirates. Even if it wasn't for the NKit format being as complicated as it is, I'm not going to implement features that only benefit pirates (apart from the feature of Dolphin yelling at them so that they don't come asking us for support with problems caused by the poor decisions of warez sites).

Source/Core/DolphinQt/ConvertDialog.cpp
JosJuice: With that said, I'm willing to drop the second paragraph if there's more than one person who doesn't think we should include what I wrote there. (I would prefer dropping it over going with your variant, since your variant doesn't really provide any information that isn't already implied by the first paragraph.)

PR #10148, opened 2021-10-03
DolphinQt: Add i18n comment for "Keep addresses where value in memory"

JosJuice:
This string is extremely likely to be mistranslated without the proper context. Actually, it's probably impossible to translate this string in a good way to some languages, but I'm not sure how to solve that. Let's at least add an i18 comment for now.

### DIFF
--- a/Source/Core/DolphinQt/CheatSearchWidget.cpp
+++ b/Source/Core/DolphinQt/CheatSearchWidget.cpp
@@ -159,6 +159,11 @@ void CheatSearchWidget::CreateWidgets()
 
   auto* value_layout = new QHBoxLayout();
 
+  // i18n: This label is followed by a dropdown where the user can select things like "is equal to"
+  // or "is less than or equal to", followed by another dropdown where the user can select "any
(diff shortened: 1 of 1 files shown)


Source/Core/DolphinQt/CheatSearchWidget.cpp
```diff
@@ -159,6 +159,11 @@ void CheatSearchWidget::CreateWidgets()
 
   auto* value_layout = new QHBoxLayout();
 
+  // i18n: This label is followed by a dropdown where the user can select things like "is equal to"
+  // or "is less than or equal to", followed by another dropdown where the user can "any value",
```
leoetlino, line 163:
```suggestion
  // or "is less than or equal to", followed by another dropdown where the user can choose "any value",
```
?

AdmiralCurtiss:
Maybe combining the label with the dropdown would make it better? But yeah I admit this is not particularly localization-friendly...

JosJuice:
I'm not sure if there are any languages we support where that actually would help. My main concern is with languages where the verb comes at the end of the sentence.

leoetlino (approved):

PR #10149, opened 2021-10-03
Add Game ID, timecode to dspdump and dtkdump filenames

malleoz:
Framedump filenames store the game ID and timecode to represent when the file was created. I felt that this should be consistent with audiodumps too, so I have added identical formatting for the dspdump and dtkdump filenames.

Only issue I can see here is that the timecode may not match the framedump. It looks like the audiodump files are created before the framedump files. Does it matter if the timecodes don't line up? Perhaps not, as it will still be easy to identify what the corresponding video/audio files are for a session.


JosJuice:
@dolphin-emu-bot rebuild

Source/Core/AudioCommon/AudioCommon.cpp
```diff
@@ -201,8 +204,15 @@ void SendAIBuffer(const short* samples, unsigned int num_samples)
...
-  std::string audio_file_name_dtk = File::GetUserPath(D_DUMPAUDIO_IDX) + "dtkdump.wav";
-  std::string audio_file_name_dsp = File::GetUserPath(D_DUMPAUDIO_IDX) + "dspdump.wav";
+  std::time_t start_time = std::time(nullptr);
+
+  std::string path_prefix = File::GetUserPath(D_DUMPAUDIO_IDX) + SConfig::GetInstance().GetGameID();
+
+  std::string base_name =
+      fmt::format("{}_{:%Y-%m-%d_%H-%M-%S}", path_prefix, *std::localtime(&start_time));
+
+  std::string audio_file_name_dtk = fmt::format("{}_{}", base_name, "dtkdump.wav");
```
lioncash, line 214:
```suggestion
  const std::string audio_file_name_dtk = fmt::format("dtkdump.wav_{}", base_name);
```

similarly for dspdump.wav below

malleoz: Fixed. Just noting that I'm doing `fmt::format("{}_dtkdump.wav", base_name)` and not `fmt::format("dtkdump.wav_{}", base_name)` as that would append the base after the file extension.

malleoz:
@dolphin-emu-bot rebuild

PR #10152, opened 2021-10-05
WiiTASInputWindow: Update on extension change

malleoz:
Previously, TAS Input window would not be updated in response to a WiiMote extension change. The change to the window would only be applied on the next bootup of Dolphin.

Since TASInputWindows are maintained in MainWindow, we need MainWindow to receive a signal. This signal has to be emitted as a result of execution in WiimoteEmuGeneral(). Holding this to be true, referencing the sender proved to be a bit challenging for me when trying to stay solely in DolphinQt rather than using `Host.h`.

Here are the parent-child relationships:
- MainWindow has child `ControllersWindow* m_controllers_window`.
- ControllersWindow has child `WiimoteControllersWidget* m_wiimote_controllers`.
- `WiimoteControllersWidget::OnWiimoteConfigure()` has **local** child `MappingWindow* window`.
- `MappingWindow::SetMappingType()` has **local** child `WiimoteEmuGeneral* widget`.
- WiimoteEmuGeneral has `OnAttachmentSelected()` which is called on extension change.

Starting from MainWindow, we can navigate from `m_controllers_window` to `m_wiimote_controllers`. We are unable to get to WiimoteControllersWidget's `window` because it is only local to `OnWiimoteConfigure()`.

Now we have to try working our way up the relationships.

`OnAttachmentSelected()` can get its MappingWindow parent, which can get its WiimoteControllersWidget parent.

Since MainWindow can get to WiimoteControllersWidget, and WiimoteEmuGeneral was able to get to WiimoteControllersWidget, this means I am connected to the object emitting the signal, thus the signal can be recognized.

This seems to be a bit cleaner of a solution than I had before, but do let me know if I can do anything to improve further. For one, I don't understand if it is better/worse to use `static_cast<WiimoteControllersWidget*>(parent())` instead of `qobject_cast<WiimoteControllersWidget*>(parent())`.

For the TAS Input GUI, `m_wii_tas_input_windows[port]->resize(newSize);` was used as otherwise switching from an extension which has a larger TAS Input window to an extension which has a smaller TAS Input window would cause the window to remain enlarged.

### DIFF
--- a/Source/Core/DolphinQt/Config/Mapping/WiimoteEmuGeneral.cpp
+++ b/Source/Core/DolphinQt/Config/Mapping/WiimoteEmuGeneral.cpp
@@ -98,6 +98,8 @@ void WiimoteEmuGeneral::OnAttachmentSelected(int extension)
 
   ce_extension->SetSelectedAttachment(extension);
 
+  GetParent()->OnExtensionChanged(GetPort(), extension);
+
   ConfigChanged();
   SaveSettings();
 }
--- a/Source/Core/DolphinQt/Config/ControllersWindow.cpp
+++ b/Source/Core/DolphinQt/Config/ControllersWindow.cpp
@@ -41,3 +41,8 @@ void ControllersWindow::ConnectWidgets()
 {
   connect(m_button_box, &QDialogButtonBox::rejected, this, &QDialog::reject);
 }
+
+WiimoteControllersWidget* ControllersWindow::GetWiimoteControllers()
+{
+  return m_wiimote_controllers;
+}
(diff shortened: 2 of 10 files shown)


Source/Core/DolphinQt/Config/Mapping/WiimoteEmuGeneral.cpp
```diff
@@ -89,6 +90,9 @@ void WiimoteEmuGeneral::OnAttachmentChanged(int extension)
   GetParent()->ShowExtensionMotionTabs(extension == WiimoteEmu::ExtensionNumber::NUNCHUK);
 
   m_extension_widget->ChangeExtensionType(extension);
+
+  // Update TAS Input window
+  Host_UpdateWiiTASInputExt(GetPort(), extension);
```
JosJuice, line 95: This is a bit weird. Why do we have a host function that's only called from DolphinQt and only implemented in DolphinQt? You could probably just use a Qt signal directly instead of going through Host.h.

malleoz: When you put it like that I realize this was a silly approach. Will fix later today

PR #10155, opened 2021-10-08
Disable hotkeys on static QFileDialog calls

malleoz:
Previously, if you had "Hotkeys Require Window Focus" disabled, you could repeatedly use the "Open" hotkey, for example, to stack File Open windows over top of each other over and over.

This PR just adds a wrapper function to toggle hotkeys for QFileDialogs. Thank you to lioncache for the approach.


malleoz:
Hopefully this was done in a proper way. Works on my end perfectly!

PR #10156, opened 2021-10-08
Movie: Warn user on dual core recording

malleoz:
Dual Core is not stable for movie recordings. TASes should never use this setting.

If a given TASer is not aware of this fact, we should present them with a warning to avoid heartbreak later down the line.

### DIFF
--- a/Source/Core/Core/BootManager.cpp
+++ b/Source/Core/Core/BootManager.cpp
@@ -252,7 +252,17 @@ bool BootCore(std::unique_ptr<BootParameters> boot, const WindowSystemInfo& wsi)
     IniFile::Section* dsp_section = game_ini.GetOrCreateSection("DSP");
     IniFile::Section* controls_section = game_ini.GetOrCreateSection("Controls");
 
-    core_section->Get("CPUThread", &StartUp.bCPUThread, StartUp.bCPUThread);
+    bool bCPUThreadPresent =
+        core_section->Get("CPUThread", &StartUp.bCPUThread, StartUp.bCPUThread);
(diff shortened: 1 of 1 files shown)


AdmiralCurtiss:
This should probably be a `ModalMessageBox::warning()` instead so it shows even if you have panic alerts disabled.

Also the `if` statement because the body is multiline so you should place `{}` around it.

malleoz:
Good points on both of those. I agree, this is significant enough of an issue that the user should be warned regardless of their panic alert setting.

malleoz:
Looks like `ConfigManager` defaults `bCPUThread` to true if not present in the ini. So I don't think it's going to be possible to distinguish between` bCPUThread` being true and the game's .ini setting it to true.

JosJuice:
To be clear, I'm not asking that the check should distinguish between the case where dual core is set globally and the case where dual core is set in a game INI. We can show the same warning for both. What I'm asking for is that if the global setting and the game INI setting differ, we check the game INI setting.

malleoz:
Note that, because this occurs in BootManager, I'm not able to make it a `ModalMessageBox`. I personally think most TASers would keep panic handlers on, as otherwise desync warnings don't show themselves.

JosJuice:
If I'm reading the new logic right, the message will not show up if the game INI forces dual core to true. Is this intentional?

Avasam:
It took me a while to understand why I couldn't use movie recording (frozen or black frame), for a while I thought it was just broken. Having at the very least a warning (one of those yellow toast messages) would've been very useful and saved me a lot of headscratching.

PR #7786, opened 2019-02-05
WIP: Broadband Adapter over Tcp (cross-compatible!)

0xFEEDC0DE64:
In this PR I implement a second Broadband adapter which uses tcp client/server architecture to communicate with other dolphin instances.

Should be easier to setup on Windows/Linux/Mac/Android than tun/tap vpn interfaces.

Server comes as standalone headless cmd line program or in the official DolphinQt gui (Tools -> Broadband Adapter Server)

Status: Implementation largly finished, doing cleanups and fixing MSVC build

image

image

PR #1826, opened 2015-01-03
Relicense Dolphin under GPLv2+

Tilka:
~~**Don't merge this yet!**~~

Right now, Dolphin has multiple license conflicts:

1. On Android, we link to libraries which are licensed under Apache License 2.0 which is incompatible with GPLv2 but compatible with GPLv3.
2. Several libraries that we always link are GPLv3 which is incompatible with GPLv2.

To resolve these issues, we are attempting to relicense from "GPLv2" to "GPLv2 or later" aka "GPLv2+". (In practice, Dolphin as a whole would only be usable under GPLv3 but parts of its source code could be reused under GPLv2.) All past and present contributors have been asked to agree. Some couldn't be reached because their email bounced, some have refused, some have additional demands, some haven't answered. New contributions are already covered by GPLv2+ since December 1st, 2014.

TODO (everyone):
- [x] Get ~~100%~~>95% agreement, either by managing to contact the remaining contributors or by dropping/rewriting their code.
- [x] Make sure all [licenses of linked code](https://github.com/Tilka/dolphin/blob/gplv2%2B/Externals/licenses.md) are compatible with GPLv3.
- ~~Implement credits in Dolphin as required by some licenses. @JMC47 suggested adding a menu entry that starts emulation of a homebrew program.~~ Postponed.
- ~~Create a contributor credits page on the website to meet demands of one contributor. For all I care, this could simply be a link to [the contributors list on GitHub](https://github.com/dolphin-emu/dolphin/graphs/contributors).~~ No copyrightable code remaining. We'll do something anyway.
- [x] Update any new references from GPLv2 to GPLv2+.

Alternatives:
- Drop Android support. This would be a shame.
- Ignore the license issue and possibly get sued or taken down from the app store.

In terms of commits, [current progress](https://docs.google.com/spreadsheets/d/1QeZYHduULTSZWUgMg_uUghIWqsCuTvvW5Z2YCfHM7oM/edit) is at 96.53%.

PR #10157, opened 2021-10-09
Tools: Add IDA Save/Load Dolphin Map Scripts for 7.x

dreamsyntax:
IDA 7.4+ breaks compatibility with IDA 6.x style scripts.
All IDA 7.0+ versions are compatible with IDA 7.x style scripts.

Updated unsupported calls per official hex-rays document:
https://hex-rays.com/products/ida/support/ida74_idapython_no_bc695_porting_guide.shtml

* Kept legacy scripts and made folders IDA 6.X and IDA 7.X in Tools/IDA - intentionally specifying its the program again rather than allowing assumption of script version.
* Added cancel check to save/load for all scripts (including 6.x scripts)
* Updated copyright and license info to reflect current status

### DIFF
--- a/Tools/IDA/IDA 6.x/LoadDolphinMap.py
+++ b/Tools/IDA/IDA 6.x/LoadDolphinMap.py
@@ -1,0 +1,62 @@
+# Copyright 2018 Dolphin Emulator Project
+# Licensed under GPLv2+
+# Refer to the LICENSES/GPL-2.0-or-later.txt file included.
+
+from collections import namedtuple
+
+
+DolphinSymbol = namedtuple("DolphinSymbol", [
+    "section", "addr", "size", "vaddr", "align", "name"
+])
+
+
+def load_dolphin_map(filepath):
+    with open(filepath, "r") as f:
+        section = ""
+        symbol_map = []
+        for line in f.readlines():
+            t = line.strip().split(" ", 4)
+            if len(t) == 3 and t[1] == "section" and t[2] == "layout":
+                section = t[0]
+                continue
+            if not section or len(t) != 5:
+                continue
+            symbol_map.append(DolphinSymbol(section, *t))
+        return symbol_map
+
+
+def ida_main():
+    import idc
+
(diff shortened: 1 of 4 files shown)


dreamsyntax:
> 
> 
> I might be wrong but AFAIK the convention of this repository is to not change the copyright year of existing files...

All I could find is
https://github.com/dolphin-emu/dolphin/blob/master/Contributing.md#dolphin-coding-style--licensing

Which does not specify this. If someone can confirm I'll change the 6.x scripts, though they were modified/changed with the extra check.

JosJuice:
It is our policy that the copyright year is based on when the file was created, not when it was last modified. Though indeed, I don't think I can see any place where this policy is written down other than review comments on PRs.

Pokechu22:
It's mentioned at https://github.com/dolphin-emu/dolphin/pull/1826#issuecomment-105214385 and https://github.com/dolphin-emu/dolphin/pull/7786#discussion_r254452368 for instance.

phire:
Theoretically it could be updated if the file has substantially changed, but you have only made minor changes to those files here.

dreamsyntax:
> 
> 
> It is our policy that the copyright year is based on when the file was created, not when it was last modified. Though indeed, I don't think I can see any place where this policy is written down other than review comments on PRs.

Gotcha, will update.


> 
> 
> Theoretically it could be updated if the file has substantially changed, but you have only made minor changes to those files here.

For clarity sake, since the original files were moved those should stay as 2018.
The 'new' files (which are just copies + sdk/func call changes) should also be 2018? Or should those be 2021 since they did not priorly exist?

JMC47:
@dolphin-emu-bot rebuild

JosJuice:
> The 'new' files (which are just copies + sdk/func call changes) should also be 2018? Or should those be 2021 since they did not exist prior?

I would've set them to 2021, but I don't think it really matters. It's a bit arbitrary. Let's leave it as it is unless someone else objects.

JosJuice:
Yes, ideally. But since it was missed in that PR, I'm fine with leaving it as it is in this PR, so as to not put a bunch of unrelated changes in the same PR.

PR #10159, opened 2021-10-10
Core: Use log variant of PrintCallstack, not print variant

JosJuice:
There are two reasons for this.

1. Using Dolphin's logging system lets the user decide whether the printout should go to the terminal, the GUI, or a file. `fmt::print` always prints to stdout... unless you're on Android, in which case it does nothing at all, because Android disables stdout.

2. The Windows version of Dolphin crashes when you use `fmt::print`. Yes, really. The crash happens because a call to `std::fprint` in `fmt::v7::detail::fwrite_fully` returns that less characters were written than requested, which fmt handles by throwing an exception. (As always, Dolphin does not use exception handling.) I'm not sure why `std::fprint` is doing this, but since switching away from using `fmt::print` is a good idea due to the previous point anyway, I'd say it's best to just switch.

### DIFF
--- a/Source/Core/Core/Debugger/Debugger_SymbolMap.cpp
+++ b/Source/Core/Core/Debugger/Debugger_SymbolMap.cpp
@@ -97,28 +97,6 @@ bool GetCallstack(std::vector<CallstackEntry>& output)
   return true;
 }
 
-void PrintCallstack()
-{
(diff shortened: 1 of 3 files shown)


Adamillo:
Can confirm that this PR fixes the crash

PR #10165, opened 2021-10-14
Fix switch warning in GameListModel

fpdotmonkey:
`-Wswitch` warnings were coming from GameListModel during build because of
unhandled branches. Adding a default branch brings some quiet to the otherwise
noisy act of compilation.


Dentomologist:
This warning was resolved in #9721, although it used `default: break;` instead of `default: return QVariant();` as suggested by phire above.

AdmiralCurtiss:
Probably fine to close this then, if there's still issues just make a new PR.

PR #10170, opened 2021-10-16
Enable deb package generation with cpack

ooshlablu:
This adds the variables necessary for `cpack` to generate a deb package right from this project, and enables the generation of the dependencies section of the deb via `dpkg-shlibdeps` (installed via the `dpkg-dev` package in Debian, Ubuntu, and Mint).

To build a debian package using this pull request:
```sh
## Follow the usual steps from the docs, but create a branch using this PR
git clone https://github.com/dolphin-emu/dolphin.git dolphin-emu
cd ./dolphin-emu
git fetch origin pull/10170/head:pr10170
git checkout pr10170
git submodule update --init
mkdir Build && cd Build

## Specify the package contact (Required for building debs with cpack)
cmake -DCPACK_PACKAGE_CONTACT="Your Name Here" ..
make -j$(nproc)

## instead of 'make install' generate the deb
cpack -G DEB
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -931,6 +931,10 @@ set(CPACK_RPM_PACKAGE_LICENSE GPL-2.0)
 # TODO: CPACK_NSIS_*
 # TODO: Use CPack components for DSPSpy, etc => cpack_add_component
 
+# Debian Specific. Install dpkg-dev for automatic deps generation
+set(CPACK_DEBIAN_PACKAGE_SHLIBDEPS ON)
+set(CPACK_DEBIAN_PACKAGE_SECTION "Games")
+
 set(CPACK_SET_DESTDIR ON)
 set(CPACK_SOURCE_GENERATOR "TGZ;TBZ2;ZIP")
 set(CPACK_SOURCE_IGNORE_FILES  "\\\\.#;/#;.*~;\\\\.swp;/\\\\.git")


ooshlablu:
> I'm not sure if we should be setting `CPACK_PACKAGE_CONTACT`. That should be done by the person actually building the package.
> 
> We don't want to end up as the contact for packages done by someone else.

Yeah, I'm open to suggestions. It is required to generate deb packages with `cpack`. I'll throw my name in there if necessary :-)

phire:
Yes, but it shouldn't be baked in.

Modify your package building instructions to tell the maintainer to add their name during configuration:

```
cmake .. -DCPACK_PACKAGE_CONTACT="Your Name Here"
```

Rumi-Larry:
Ammending your commit to fix mistakes or for minor changes, is always better than making a new one. Just fyi, because you will need to squash those commits anyway.

Drex-C137:
_So that it doesn't happen to anyone else again_, here is a list of dependencies that I had to install so that the project would compile **without errors**. I did this on **Linux Mint 21.3** (Ubuntu / Debian based)

`sudo apt install build-essential libgl1-mesa-dev libx11-dev libudev-dev libegl1 libxi-dev libevdev-dev qtbase5-dev qtbase5-private-dev `

PR #10174, opened 2021-10-17
DolphinQt: Add "Reset Ignore Panic Handler" MenuBar Option

dreamsyntax:
### Overview:
* Adds a "Reset Ignore Panic Handler" button to the Options subgroup in the MenuBar; It will only appear if Debug mode is enabled
* The button will do nothing if Config -> Interface -> Use Panic Handlers is disabled

Originally, I was going to make this a toggle like Automatic Start and Boot to Pause. Since Main.cpp calls SetEnabledAlert directly and there is no way to call MenuBar qActions from Main.cpp, I opted to go with a stateless button. User usage of the button also made little sense to have a toggle, since Config -> Interface already has a "Use Panic Handlers".

### Why this is useful:
When a person debugging runs into an exception like this, sometimes there are 5000+ messages and the only choice is to choose "Ignore for this session". There really should be a "pause" option for those scenarios, but that's coming in another PR. This prevents any further pop-up messages about exceptions. The person debugging then has no way to re-enable the Ignore flag[*note1] without completely restarting Dolphin. With this change now there is a clear way to do it under the Options sub-menu.

mem-exception

reset-ignore-button-fix

[*note1] - You could disable panic handlers under Interface and re-enable them, but the state of the config at this point is desynced under that scenario.

### DIFF
--- a/Source/Core/DolphinQt/MenuBar.cpp
+++ b/Source/Core/DolphinQt/MenuBar.cpp
@@ -165,6 +165,7 @@ void MenuBar::OnDebugModeToggled(bool enabled)
   // Options
   m_boot_to_pause->setVisible(enabled);
   m_automatic_start->setVisible(enabled);
+  m_reset_ignore_panic_handler->setVisible(enabled);
   m_change_font->setVisible(enabled);
 
   // View
@@ -542,6 +543,13 @@ void MenuBar::AddOptionsMenu()
 
   connect(m_automatic_start, &QAction::toggled, this,
           [](bool enable) { SConfig::GetInstance().bAutomaticStart = enable; });
+
+  m_reset_ignore_panic_handler = options_menu->addAction(tr("Reset Ignore Panic Handler"));
+
+  connect(m_reset_ignore_panic_handler, &QAction::triggered, this, []() {
+    if (Config::Get(Config::MAIN_USE_PANIC_HANDLERS))
+      Common::SetEnableAlert(true);
+  });
 
   m_change_font = options_menu->addAction(tr("&Font..."), this, &MenuBar::ChangeDebugFont);
 }
--- a/Source/Core/DolphinQt/MenuBar.h
+++ b/Source/Core/DolphinQt/MenuBar.h
@@ -234,6 +234,7 @@ private:
   // Options
   QAction* m_boot_to_pause;
   QAction* m_automatic_start;
+  QAction* m_reset_ignore_panic_handler;
   QAction* m_change_font;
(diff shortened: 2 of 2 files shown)


dreamsyntax:
> 
> 
> Hmm. To be direct, I don't understand why this should exist when creating a pause for the panic handler does the same thing but better.
> 
> EDIT: OH, I think I misunderstood. Still an idea though.
> 
> So Dolphin already has the ability to turn panic handling off and on at demand. So why not turn panic handling off until you reach what you need, then turn it back again? Wouldn't that achieve the same result? 

We cannot programmatically 'reach what you need' as every game/context is different.

> To make it easier, you could add an option in the drop down for debug mode to make it more accessible, and/or add a hotkey.

I'm not opposed to this, but I don't think it would be widely used as a hotkey. The use cases I see this being used is validating/tracking down buffer overflows, memory leaks, or validating unstable mods/codes easier.
I do plan to rewrite how panic alerts are handled entirely though, maybe Ignoring types rather than ignoring all could be a thing I'll consider as well.

dreamsyntax:
@phire is there any other requirements to get this through? Or are we good to merge this?
I've been using this in my integration test for my regular workflows and find it useful.

dreamsyntax:
Bumping this, this is a minor change that adds value to those using debugging

AdmiralCurtiss:
This is a very simple change that's only visible in the debugging UI and potentially useful, so I'm for it.

iljatanger:
Hello dear developers and thank you very much for your labour and such a great emulator.

Started from the release 15474, the performance in many games (SMG, SMG2, Metroid Prime, Super Mario bros.) etc got significantly lower (2-3 times!).

this is 100% tested on many releases (about 20 times). 

Initially I tried to update Dolphin on MacBook Air M1 from the beta version 15445 to the newest one 16101, because I had artifacts (black and transparent squares in the center of screen in Super Smash Bros. Brawl). And I was shocked what the performance became so poor. If at 15445  3X in Mario Galaxy and Metroid Prime was playable, now it became much worse, so only 1X and 2x (hardly) playable.  So after that I installed many releases and found, that such visible, significant performance drop had started from this release. The size of the release also was changed started from this version 15474 (38.8mb instead of 34.9 before, so I suggest it was some global recompiling of the code).

so now I stopped on version 15472 as the latest playable one. But unfortunately, I still have the same visual artifacts in Super Smash Bros...

Could you ever be so kind to check whether there is some mistake in this release starting from 15474?

I have macbook Air M1 16/256 on BigSur

thank you very much in advance

PR #10175, opened 2021-10-18
Core/DolphinQt: Add Pause on Panic Option for DSI Exceptions and Unknown Instruction

dreamsyntax:
### Overview
* Adds `Pause on Panic` to Config -> Advanced

advance-pane

* JitBase.cpp change is necessary because otherwise the JIT splices would break on the next block rather than the instruction that threw the exception.
* Defaults to OFF since this has the same performance hit that enabling MMU does.
* Currently, this is only affecting panics for invalid read/writes. Ex `stw r4, r5` where r5 is a non-valid address (ex 99232100)
* Additionally, I also check for the flag in Interpreter.cpp for Unknown Instruction. I do not want to default this to pause/break as this will negatively impact a user who runs into an Unknown Instruction that doesn't crash Dolphin or the emulation. This will be updated with the Panic message rewrite future PR.

### Why
* When debugging games/homebrew if a read/write exception occurs, there is no easy way to observe the callstack and registers at time of fault. With this, the emulation will pause on any such exception while enabled.

### TODO / Before Merge (All Done)
* ~~Does not yet save the checked state on program close.~~ | Done
* ~~Verify the name "Pause on Panic Handles" is okay- open to different names.~~ | Done (changed to Pause on Panic Handler)
* ~~Expand to all PanicMessage types if reasonable~~ Opted to only account for the most common scenarios DSI & Unknown Instruction. A future PR will rework the Yes/No/Ignore entirely, which will follow this https://github.com/dolphin-emu/dolphin/pull/10175#issuecomment-946149157. 
* ~~Do I need to account for this PR: https://github.com/dolphin-emu/dolphin/pull/10162 ?~~ Nope, thanks JosJuice

### DIFF
--- a/Source/Core/DolphinQt/Settings/AdvancedPane.cpp
+++ b/Source/Core/DolphinQt/Settings/AdvancedPane.cpp
@@ -68,6 +68,12 @@ void AdvancedPane::CreateLayout()
       "Enables the Memory Management Unit, needed for some games. (ON = Compatible, OFF = Fast)"));
   cpu_options_group_layout->addWidget(m_enable_mmu_checkbox);
 
+  m_pause_on_panic_checkbox = new QCheckBox(tr("Pause on Panic"));
+  m_pause_on_panic_checkbox->setToolTip(
+      tr("Pauses the emulation if a Read/Write or Unknown Instruction panic occurs.\nEnabling will "
+         "affect performance.\nThe performance impact is the same as having Enable MMU on."));
+  cpu_options_group_layout->addWidget(m_pause_on_panic_checkbox);
+
   auto* clock_override = new QGroupBox(tr("Clock Override"));
   auto* clock_override_layout = new QVBoxLayout();
   clock_override->setLayout(clock_override_layout);
@@ -180,6 +186,9 @@ void AdvancedPane::ConnectLayout()
   connect(m_enable_mmu_checkbox, &QCheckBox::toggled, this,
           [](bool checked) { Config::SetBaseOrCurrent(Config::MAIN_MMU, checked); });
 
+  connect(m_pause_on_panic_checkbox, &QCheckBox::toggled, this,
+          [](bool checked) { Config::SetBaseOrCurrent(Config::MAIN_PAUSE_ON_PANIC, checked); });
+
   m_cpu_clock_override_checkbox->setChecked(Config::Get(Config::MAIN_OVERCLOCK_ENABLE));
   connect(m_cpu_clock_override_checkbox, &QCheckBox::toggled, [this](bool enable_clock_override) {
     Config::SetBaseOrCurrent(Config::MAIN_OVERCLOCK_ENABLE, enable_clock_override);
@@ -245,6 +254,9 @@ void AdvancedPane::Update()
 
   m_enable_mmu_checkbox->setChecked(Config::Get(Config::MAIN_MMU));
   m_enable_mmu_checkbox->setEnabled(!running);
+
+  m_pause_on_panic_checkbox->setChecked(Config::Get(Config::MAIN_PAUSE_ON_PANIC));
+  m_pause_on_panic_checkbox->setEnabled(!running);
 
   QFont bf = font();
   bf.setBold(Config::GetActiveLayerForConfig(Config::MAIN_OVERCLOCK_ENABLE) !=
--- a/Source/Core/Core/Config/MainSettings.cpp
+++ b/Source/Core/Core/Config/MainSettings.cpp
@@ -168,6 +168,7 @@ const Info<bool> MAIN_WIIMOTE_ENABLE_SPEAKER{{System::Main, "Core", "WiimoteEnab
 const Info<bool> MAIN_CONNECT_WIIMOTES_FOR_CONTROLLER_INTERFACE{
     {System::Main, "Core", "WiimoteControllerInterface"}, false};
 const Info<bool> MAIN_MMU{{System::Main, "Core", "MMU"}, false};
+const Info<bool> MAIN_PAUSE_ON_PANIC{{System::Main, "Core", "PauseOnPanic"}, false};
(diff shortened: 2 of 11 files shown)


Dentomologist:
@dolphin-emu-bot rebuild

dreamsyntax:
> 
> 
> I know you are also working on pausing for other panic alert types. Do you actually want to have a single option that enables pausing on panic handlers?
> 
> Or would it be smarter to have the feature enabled by default, giving the user the choice between ignore/stop/break) for every (supported) panic alert except for MMU exceptions (which require either memchecks or interpeter).

If we go this route, then I think we should change the behavior of the Panic Messages entirely.

Unknown Instruction (ill instruction) for example:
unknown-panic
Current Dolphin behavior is as follows:
Yes / Ignore / No - What am I saying Yes/No to?
Yes - the program simply skips the bad instruction. If another panic happens later another pop-up message will appear.
Ignore - the program does the same as Yes, but also setting a flag to no longer pop up panic messages.
No - The program crashes. 

Ideally, I would propose:
mockup
Not sure how feasible since Yes/Ignore/No is shared across a bunch of scenarios.

> Also, should it be called Pause on Panic, or Break on Panic, with explicit mention of debugging in the tooltip?
I don't think its necessary to say Break / mention debugger, as having this on would affect users in non-debug as well. If it was on by default, then even if they chose "Ignore" earlier and later run into exceptions all they would see is the emulation paused, with seemingly no reason.

> Talking about the tooltip, you probably want to mention that the cost is the same as enabling MMU.
Yes, or if MMU is on there is no additional cost.

dreamsyntax:
I've updated this per suggestions.
I would rather get the most common exceptions (read/write and unknown instr) out now. I've updated the hint to properly reflect that.

In another PR I will look at reworking the Panic MessageBox entirely, where we can then have behaviors like in my mock #2 image https://github.com/dolphin-emu/dolphin/pull/10175#issuecomment-946149157 , but this is a bigger effort that will affect a lot code. I

dreamsyntax:
Going to rebase and resolve the conflicts.
Did some testing on latest dev dolphin, the idea behind this change is still needed even with improvements to the exception handler (it is still very easy to crash to desktop in latest dev)

dreamsyntax:
Migrated to new config format

old:
https://github.com/dolphin-emu/dolphin/pull/10175/commits/1acced43dee8ede59fbfbd9106a2d854c0ca87ea

sepalani (approved):
Code LGTM, untested.

MasterofGalaxies:
This would be very useful for me. Hope it gets merged soon.

Source/Core/DolphinQt/Settings/AdvancedPane.cpp
```diff
@@ -68,6 +68,12 @@ void AdvancedPane::CreateLayout()
       "Enables the Memory Management Unit, needed for some games. (ON = Compatible, OFF = Fast)"));
   cpu_options_group_layout->addWidget(m_enable_mmu_checkbox);
 
+  m_pause_on_panic_checkbox = new QCheckBox(tr("Pause on Panic"));
+  m_pause_on_panic_checkbox->setToolTip(
+      tr("Pauses the emulation if a Read/Write or Unknown Instruction panic occurs.\nEnabling will "
```
sepalani, line 73: This description is misleading, I came across invalid read/write from `WriteToHardware` and `ReadToHardware` with the associated `Unable to resolve read/write address` panic alerts and it didn't break the execution. You should probably reword it to emphasis that it breaks on unknown instructions and DSI exceptions instead.

sepalani (changes requested):
Other than that, tested and work as intended.

Source/Core/DolphinQt/Settings/AdvancedPane.cpp
dreamsyntax: Ideally, I would like to extend this to all of the write/read exceptions. Is there specific areas I should look into for this?

JosJuice: The invalid read/writes (they are not exceptions) that sepalani is referring to are these:

https://github.com/dolphin-emu/dolphin/blob/349c484ec803ccba96bd334da5de771cfd107029/Source/Core/Core/PowerPC/MMU.cpp#L246
https://github.com/dolphin-emu/dolphin/blob/349c484ec803ccba96bd334da5de771cfd107029/Source/Core/Core/PowerPC/MMU.cpp#L400

dreamsyntax: Added these. As a caveat, these continue to re-appear when pressing "OK".
If picking "Ignore" it works as intended. The "Reset Panic Handler" drop down from Tools can be used in conjunction as a work around.

MasterofGalaxies:
I think this would be better if it weren't persistent across launches. I definitely would use this if it got merged, but not every time I use Dolphin. Maybe it could go under Options alongside Boot to Pause?

dreamsyntax:
> I think this would be better if it weren't persistent across launches. I definitely would use this if it got merged, but not every time I use Dolphin. Maybe it could go under Options alongside Boot to Pause?

I'm not opposed to this, but seeing as its similar perf hit to MMU I think it makes sense to have the context under config->advanced. Open to other opinions though.

MasterofGalaxies:
Any news?

sepalani (approved):
This PR still LGTM. You might want to squash the 2 commits into a single one with a proper commit message.

dreamsyntax:
> This PR still LGTM. You might want to squash the 2 commits into a single one with a proper commit message.

Done, rebased on latest master + squashed

MasterofGalaxies:
This PR looks ready to merge

MasterofGalaxies:
Can someone merge this or indicate why it’s not being merged

sepalani (approved):
Still LGTM.

PR #10178, opened 2021-10-20
Improvements for converting multiple files

flagrama:
When converting multiple files, it is easy to cancel prematurely by typing and
starting the process back up can involve either checking the already converted
files and skipping selecting them, or clicking "No" many times on the warning
dialog asking you if you want to replace the existing file.

The "No To All" button will skip creating all files that already exist without
asking for further user input, but will still create converted files that do
not already exist in the selected directory.

image

One potential issue I see is that it makes "No to All" the default selection. Unsure if that is desired, and if not, how to fix it.

### DIFF
--- a/Source/Core/DiscIO/FileBlob.cpp
+++ b/Source/Core/DiscIO/FileBlob.cpp
@@ -41,18 +41,18 @@ bool PlainFileReader::Read(u64 offset, u64 nbytes, u8* out_ptr)
 }
 
 bool ConvertToPlain(BlobReader* infile, const std::string& infile_path,
-                    const std::string& outfile_path, CompressCB callback)
+                    const std::string& tempfile_path, CompressCB callback)
 {
   ASSERT(infile->IsDataSizeAccurate());
 
-  File::IOFile outfile(outfile_path, "wb");
-  if (!outfile)
+  File::IOFile tempfile(tempfile_path, "wb");
+  if (!tempfile)
   {
     PanicAlertFmtT(
         "Failed to open the output file \"{0}\".\n"
         "Check that you have permissions to write the target folder and that the media can "
         "be written.",
-        outfile_path);
+        tempfile_path);
     return false;
   }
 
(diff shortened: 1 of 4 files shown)


Pokechu22:
@dolphin-emu-bot rebuild

Rumi-Larry:
> Qt automatically selects the ordering of the buttons depending on what's customary on the user's operating system. I don't think there's sufficient reason to interfere with that.

That's very interesting. Thank you!

flagrama:
Rebased onto master. Should get rid of FIFO diffs since this doesn't touch any rendering otherwise.

JosJuice:
@dolphin-emu-bot rebuild

Source/Core/DiscIO/WIABlob.cpp
```diff
@@ -1702,6 +1702,10 @@ ConversionResultCode WIARVZFileReader<RVZ>::RunCallback(size_t groups_written, u
 
   const float completion = static_cast<float>(bytes_read) / iso_size;
 
+  // Progress Dialog will close prematurely if 100% is reached
+  if (completion >= 0.99f)
+    return callback(text, 0.99f) ? ConversionResultCode::Success : ConversionResultCode::Canceled;
```
JosJuice, line 1707: Could this be done in DolphinQt instead of in DiscIO? The behavior that you're trying to work around is in Qt, after all.

Source/Core/DiscIO/FileBlob.cpp
```diff
@@ -93,23 +93,16 @@ bool ConvertToPlain(BlobReader* infile, const std::string& infile_path,
...
     }
   }
 
-  if (!success)
-  {
-    // Remove the incomplete output file.
-    outfile.Close();
-    File::Delete(outfile_path);
-  }
-
```
JosJuice, line 112: If you want to get rid of this, you'll have to update the Android code in addition to the DolphinQt code.

JosJuice:
> Was unsure if I should still pass the `outfile_path` and use it for the error messages, or just change the error messages to indicate that the file is a temp file in the same location as the final file will be.

The main purpose of showing the path to the user is to indicate which game the problem occurred for, so perhaps it would be useful. (Or we could show just the file name instead of the whole path.)

Source/Core/DiscIO/WIABlob.cpp
flagrama: I'll take a look at that. It doesn't seem to be a trivial fix though.

flagrama:
That makes sense. I'll see if I can find a way to still inform the user which game it was.

Source/Core/DiscIO/FileBlob.cpp
flagrama: That's a good point. I only removed it because QTemporaryFile handles that automatically. I'll probably just restore this for now since it shouldn't make any difference for the Qt code.

PR #10187, opened 2021-10-26
Support for GameModDescriptor files in Game List.

AdmiralCurtiss:
Someone more familiar with how the game list cache works please check if this is correct.

TODO:
- [x] Sensible Netplay hash calculation.
- [x] The temporary Netplay NAND interacts *very poorly* with savegame redirection.

### DIFF
--- a/Source/Core/Core/WiiRoot.cpp
+++ b/Source/Core/Core/WiiRoot.cpp
@@ -35,8 +35,18 @@ namespace Core
 namespace FS = IOS::HLE::FS;
 
 static std::string s_temp_wii_root;
+static std::string s_temp_redirect_root;
 static bool s_wii_root_initialized = false;
 static std::vector<IOS::HLE::FS::NandRedirect> s_nand_redirects;
+
+// When Temp NAND + Redirects are both active, we need to keep track of where each redirect path
+// should be copied back to after a successful session finish.
+struct TempRedirectPath
+{
+  std::string real_path;
+  std::string temp_path;
+};
+static std::vector<TempRedirectPath> s_temp_nand_redirects;
 
 const std::vector<IOS::HLE::FS::NandRedirect>& GetActiveNandRedirects()
 {
@@ -175,6 +185,28 @@ static void InitializeDeterministicWiiSaves(FS::FileSystem* session_fs,
         WARN_LOG_FMT(CORE, "Failed to copy Mii database to the NAND");
       }
     }
+
+    const auto& netplay_redirect_folder = boot_session_data.GetWiiSyncRedirectFolder();
+    if (!netplay_redirect_folder.empty())
+      File::CopyDir(netplay_redirect_folder, s_temp_redirect_root + "/");
+  }
+}
+
+static void MoveToBackupIfExists(const std::string& path)
+{
+  if (File::Exists(path))
+  {
+    const std::string backup_path = path.substr(0, path.size() - 1) + ".backup" DIR_SEP;
+    WARN_LOG_FMT(IOS_FS, "Temporary directory at {} exists, moving to backup...", path);
+
+    // If backup exists, delete it as we don't want a mess
+    if (File::Exists(backup_path))
+    {
+      WARN_LOG_FMT(IOS_FS, "Temporary backup directory at {} exists, deleting...", backup_path);
+      File::DeleteDirRecursively(backup_path);
+    }
+
+    File::CopyDir(path, backup_path, true);
   }
 }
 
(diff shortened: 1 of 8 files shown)


JMC47:
Tested.  Launches correctly from Gamelist and Netplay.

Source/Core/DiscIO/Blob.cpp
```diff
@@ -49,6 +49,8 @@ std::string GetName(BlobType blob_type, bool translate)
     return "WIA";
   case BlobType::RVZ:
     return "RVZ";
+  case BlobType::MOD_DESCRIPTOR:
+    return "Mod";
```
JosJuice, line 53:
```suggestion
    return translate_str("Mod");
```

Source/Core/UICommon/GameFile.cpp
```diff
@@ -163,6 +165,34 @@ GameFile::GameFile(std::string path) : m_file_path(std::move(path))
...
+        m_gametdb_id = proxy.GetGameTDBID();
+        m_title_id = proxy.GetTitleID();
+        m_maker_id = proxy.GetMakerID();
+        m_region = proxy.GetRegion();
+        m_country = proxy.GetCountry();
+        m_platform = proxy.GetPlatform();
+        m_revision = proxy.GetRevision();
+        m_disc_number = proxy.GetDiscNumber();
+        if (ReadPNGBanner(descriptor->banner))
+          m_volume_banner = std::move(m_pending.custom_banner);
```
JosJuice, line 191: I would suggest putting the `ReadPNGBanner` call in `CustomBannerChanged` instead of the constructor so that the game list will finish populating quicker. (`CustomBannerChanged` is called by `UpdateAdditionalMetadata` after all GameFiles have been constructed and inserted into the game list. Perhaps its name doesn't do a great job of explaining what it does...)

Source/Core/UICommon/GameFile.cpp
```diff
@@ -579,23 +611,83 @@ std::string GameFile::GetNetPlayName(const Core::TitleDatabase& title_database)
...
 {
-  std::array<u8, 20> hash{};
+  std::optional<std::array<u8, 20>> hash;
 
   if (m_platform == DiscIO::Platform::ELFOrDOL)
   {
-    std::string buffer;
-    if (File::ReadFileToString(m_file_path, buffer))
-      mbedtls_sha1_ret(reinterpret_cast<unsigned char*>(buffer.data()), buffer.size(), hash.data());
+    hash = MixHash(hash, GetFileHash(m_file_path));
```
JosJuice, line 656: Can't you skip the `MixHash` call here? `hash` is always empty at this point.

Source/Core/Core/NetPlayCommon.cpp
```diff
@@ -187,6 +217,37 @@ bool DecompressPacketIntoFile(sf::Packet& packet, const std::string& file_path)
...
+
+  sf::Uint64 size;
+  packet >> size;
+  for (size_t i = 0; i < size; ++i)
+  {
+    std::string name;
+    packet >> name;
+    bool is_folder;
+    packet >> is_folder;
+    std::string path = fmt::format("{}/{}", folder_path, name);
```
JosJuice, line 233: Path traversal vulnerability.

(I wish there was a convenient API for escaping/filtering dangerous characters... SQL got this down a pretty long time ago with prepared statements, but for paths there doesn't really seem to be anything in major OSes/libraries.)

AdmiralCurtiss: Checking for '/' and '\' in `name` is probably good enough here?

AdmiralCurtiss: Oh no wait, it can't be all dots either... bleh.

Source/Core/UICommon/GameFile.cpp
```diff
@@ -624,23 +625,83 @@ std::string GameFile::GetNetPlayName(const Core::TitleDatabase& title_database)
...
+
+static std::optional<std::array<u8, 20>> GetFileHash(const std::string& path)
+{
+  std::string buffer;
+  if (!File::ReadFileToString(path, buffer))
+    return std::nullopt;
+  return GetHash(buffer);
+}
+
+static std::optional<std::array<u8, 20>> MixHash(const std::optional<std::array<u8, 20>>& lhs,
```
JosJuice, line 651: What do you think about making lhs and the return type non-optional? I feel like it would clean things up.

AdmiralCurtiss: You... think so? I feel like it would make `GameFile::GetSyncHash()` way more complicated, actually.

JosJuice: Can't you initialize the hash to an all-zero vector, remove the `value_or` call at the end, and keep the rest as it is?

AdmiralCurtiss: https://gist.github.com/AdmiralCurtiss/7932ee2a8caf818f5e8559c8090d6b4b I guess? I don't really feel like either version is better than the other honestly.

JosJuice: Ah, I forgot that `GetFileHash` returned an optional. I guess it's a wash then. Feel free to ignore my suggestion.

JosJuice:
Like with the original Riivolution PR: I would appreciate if @leoetlino could look at the save redirection part, but the rest LGTM.

Source/Core/Core/WiiRoot.cpp
```diff
@@ -312,29 +336,45 @@ void InitializeWiiFileSystemContents(
...
   }
-  else if (save_redirect)
+
+  if (save_redirect)
   {
     const u64 title_id = SConfig::GetInstance().GetTitleID();
     std::string source_path = Common::GetTitleDataPath(title_id);
-    if (!File::IsDirectory(save_redirect->m_target_path))
+
+    DiscIO::Riivolution::SavegameRedirect evaluated_redirect;
```
leoetlino, line 354: Why not modify `save_redirect` directly? You could avoid to need for a separate DiscIO::Riivolution::SavegameRedirect variable, a separate move and it's more error prone to have both `save_redirect` and `evaluated_redirect` in the same scope. Too easy to accidentally use the wrong one...

If you use `save_redirect`, you'd only have to overwrite the SavegameRedirect's string in the `if (s_temp_nand)` instead of having to create an entirely new SavegameRedirect and initialize each field (even fields which haven't changed, like `m_clone`) manually.

leoetlino (commented):
Mostly LGTM, just one minor complaint.

Source/Core/Core/WiiRoot.cpp
AdmiralCurtiss: Yeah, that's fair.

leoetlino (approved):

leo60228:
Is there a way to set the Maker in addition to the name and banner?

leo60228:
An HBC-style XML file works, but I feel like it'd make sense to have a field in the JSON.

AdmiralCurtiss:
Yeah we can probably add that, should be easy enough.

PR #10188, opened 2021-10-26
Externals: Update imgui to 1.85

Pokechu22:
This PR updates Dear ImGui to v1.85 (from 1.70).  There shouldn't be any behavior differences, but this update is useful for some other stuff I'm working on (in particular, it adds `BeginDisabled` and `EndDisabled`, and also causes a scroll bar to show up if a window is too big and `ImGuiWindowFlags_AlwaysAutoResize` is in use (before, the window would just not render properly)).

### DIFF
--- a/Externals/imgui/CMakeLists.txt
+++ b/Externals/imgui/CMakeLists.txt
@@ -1,6 +1,7 @@
 set(SRCS
   imgui.cpp
   imgui_draw.cpp
+  imgui_tables.cpp
   imgui_widgets.cpp
 )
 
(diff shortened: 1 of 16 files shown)


Pokechu22:
That issue has been fixed in 5.0-15427 since a clean build was performed for that.  Technically, this PR didn't cause the issue (everything would work properly if you built it locally), but instead the build system not rebuilding files that depended on imgui caused it.  That's still something that should be fixed, but it's not clear exactly why the build system isn't rebuilding it.

BParks21:
I see. Ok then.

PR #10189, opened 2021-10-27
RenderBase: replace FPS: 60.00 to FPS: 60

Gamer64ytb:
This makes the FPS counter much more stable and is unnecesary have the ".00" extra.


leoetlino:
Having the format jump between XX (for integer values) and XX.XX seems even more distracting than always showing 2 decimals, though, which would defeat the stated goal of making the FPS counter "much more stable".

Considering the mostly negative reception and the fact that there are few good reasons for this change other than personal preference, I am closing this PR because it is unlikely to be merged.

> But... Why? Games REALLY like having their smooth fps. Even 59.30 could confuse some games that are especially picky

For the record, that doesn't make *any* sense — this PR only changes the user-visible FPS counter. There is no way this can actually affect emulation.

PR #10191, opened 2021-10-28
Fix stale imgui library issues

Pokechu22:
5.0-15421 updated Dear ImGui, but it seems like this is causing issues such as https://bugs.dolphin-emu.org/issues/12717 - also visible with the "Show Statistics" option under advanced graphics settings (the size is wrong and mouse interactions don't work).  This may be caused by the wrong version of the headers being used, causing a mismatch between version of a struct seen in the header file and the actual version the library is using.

@Dentomologist mentioned that RenderWidget was using `#include "imgui.h"` instead of `#include <imgui.h>`, and it does seem like [it wasn't rebuilt for 5.0-15421](https://dolphin.ci/#/builders/6/builds/1440/steps/3/logs/stdio); only `Version.cpp`, `NetPlayChatUI.cpp`, `RenderBase.cpp`, and `ShaderCache.cpp` were rebuilt.  However, `Host.cpp`, `OnScreenDisplay.cpp`, `NetPlayGolfUI.cpp`, and `Statistics.cpp` also include `<imgui.h>` and they weren't rebuilt, so I don't think this is the core cause (changing the file just caused it to be rebuilt, hiding the issue).

### DIFF
--- a/Source/Core/DolphinQt/RenderWidget.cpp
+++ b/Source/Core/DolphinQt/RenderWidget.cpp
@@ -19,7 +19,7 @@
 #include <QTimer>
 #include <QWindow>
 
-#include "imgui.h"
+#include <imgui.h>
 
 #include "Core/Config/MainSettings.h"
 #include "Core/ConfigManager.h"


leoetlino:
Well, even if this doesn't fix the issue, we should still be using angle brackets rather than quotes for the include :)

PR #10194, opened 2021-10-31
Fix some Riivolution file patching edge cases.

AdmiralCurtiss:
This fixes the booting of Wii Sports Resort Storm Island, [as reported on the forums](https://forums.dolphin-emu.org/Thread-wii-sports-resort-storm-island-freezes-at-intro-screen-riivolution).

### DIFF
--- a/Source/Core/DiscIO/RiivolutionPatcher.cpp
+++ b/Source/Core/DiscIO/RiivolutionPatcher.cpp
@@ -366,25 +366,29 @@ static FSTBuilderNode* FindFileNodeInFST(std::string_view path, std::vector<FSTB
                            create_if_not_exists);
 }
 
-static void FindFilenameNodesInFST(std::vector<DiscIO::FSTBuilderNode*>* nodes_out,
-                                   std::string_view filename, std::vector<FSTBuilderNode>* fst)
+static DiscIO::FSTBuilderNode* FindFilenameNodeInFST(std::string_view filename,
+                                                     std::vector<FSTBuilderNode>& fst)
 {
-  for (FSTBuilderNode& node : *fst)
+  for (FSTBuilderNode& node : fst)
   {
     if (node.IsFolder())
     {
-      FindFilenameNodesInFST(nodes_out, filename,
-                             &std::get<std::vector<FSTBuilderNode>>(node.m_content));
+      DiscIO::FSTBuilderNode* result = FindFilenameNodeInFST(filename, node.GetFolderContent());
+      if (result)
+        return result;
     }
     else if (node.m_filename == filename)
     {
-      nodes_out->push_back(&node);
+      return &node;
     }
   }
+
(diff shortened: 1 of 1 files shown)


Source/Core/DiscIO/RiivolutionPatcher.cpp
```diff
@@ -364,148 +364,109 @@ static FSTBuilderNode* FindFileNodeInFST(std::string_view path, std::vector<FSTB
...
+    const auto combine_paths = [](std::string_view a, std::string_view b) {
+      if (a.empty())
+        return std::string(b);
+      if (b.empty())
+        return std::string(a);
+      if (StringEndsWith(a, "/"))
+        a.remove_suffix(1);
+      if (StringBeginsWith(b, "/"))
+        b.remove_prefix(1);
+      return std::string(a) + "/" + std::string(b);
```
iwubcode, line 430: NBD but I think `fmt::format("{}/{}", a, b)` may be a bit more performant

AdmiralCurtiss:
Hold on, this apparently breaks Newer...

Source/Core/DiscIO/RiivolutionPatcher.cpp
AdmiralCurtiss: Good point, yes.

Source/Core/DiscIO/RiivolutionPatcher.cpp
```diff
@@ -364,148 +364,109 @@ static FSTBuilderNode* FindFileNodeInFST(std::string_view path, std::vector<FSTB
                            create_if_not_exists);
 }
 
-static void FindFilenameNodesInFST(std::vector<DiscIO::FSTBuilderNode*>* nodes_out,
-                                   std::string_view filename, std::vector<FSTBuilderNode>* fst)
+static DiscIO::FSTBuilderNode* FindFilenameNodeInFST(std::string_view filename,
+                                                     std::vector<FSTBuilderNode>& fst)
 {
-  for (FSTBuilderNode& node : *fst)
+  for (FSTBuilderNode& node : fst)
```
iwubcode, line 370: Are you able to make this `const` ?

AdmiralCurtiss: No, because this is potentially the return value and thus must be mutable.

AdmiralCurtiss:
I genuinely have no idea why FifoCI seems to think this affects rendering. Help?

Source/Core/DiscIO/RiivolutionPatcher.cpp
iwubcode: Yeah, it looks like all the usages were just getting data on the item but I wasn't 100% sure.  NBD either way

leoetlino (approved):

PR #4481, opened 2016-12-01
[HLE] Variable Argument Lists support

sepalani:
I originally planned to rewrite the ```GetStringVA``` but each times I tried I figured out there are so many tricky things to handle and it will make code redundancy with so many "hacks" and makes it ugly really quickly. So, I had the idea to rewrite an implementation of varargs following the [SVR4 ABI](http://refspecs.linux-foundation.org/elf/elfspec_ppc.pdf).

It pays attention to registers parity for doubleword, stack alignment depending on the type and will grab values from registers or the stack when necessary.

Concerning the [long double](https://en.wikipedia.org/wiki/Long_double) support, it works and doesn't at the same time. DevkitPPC interpret them as double so it's no big deal and will work. Some compilers such as GCC use double-double arithmetic that we have no easy way to support or represent.

The varargs support will help to make that printf's like function better. Then it might help in the far future to do tricky things like grabbing a struct/class from the memory (the programmer will need to pay attention to the endianness though) and use them in its own code.

This PR needs tons of tests before being merged.

---
This change is [Reviewable](https://reviewable.io/reviews/dolphin-emu/dolphin/4481)

PR #10198, opened 2021-11-02
Breakpoints: Add custom log message

sepalani:
This PR add custom log messages to (M)BPs. This is still a work in progress and I'm open to suggestions regarding the design of this. I'm currently using fmt's named arguments. I'm also catching fmt's exceptions when the formatting fails.

TODO:
 - [ ] Add PairedSingle custom formatter
 - [ ] Add string support for integers (via `PowerPC::HostGetString`)
 - [ ] Convert some registers to the right type when the conversion specifiers is specified (ex: paired single)
 - [ ] Add variadic string conversion specifiers `v`/`V` (via `HLE_GeneralDebugPrint`/`HLE_GeneralDebugVPrint`)
 - [ ] Add a help button to document the message format


Source/Core/Core/Debugger/HLEFormatter.cpp
```diff
@@ -0,0 +1,72 @@
...
+  const HLENamedArgs args = GetHLENamedArgs();
+  try
+  {
+    return fmt::vformat(message, args);
+  }
+  catch (const fmt::format_error& e)
+  {
+    ERROR_LOG_FMT(MEMMAP, "Failed to format message: {}", e.what());
+    return message.data();
+  }
```
BhaaLseN, line 70: Do they have a non-throwing version as well? Since we build without exceptions and all that...

sepalani: I haven't found one, yet. Based on the source: 
 - the exception is thrown using the macro `FMT_THROW` in the `throw_format_error` function,
 - that macro might call `FMT_ASSERT` instead of `throw` if exception is disabled (i.e. FMT_EXCEPTIONS is false),
 - if NDEBUG is set, `fmt::detail::ignore_unused` should noop.

I'll try to see if I can meet these requirements in the Windows build and get rid of the throwing version.

sepalani:
This PR is ready to be reviewed and tested. If wrong formatters are used, it shouldn't raise an exception nor crash. I tested it on Windows but Linux should be tested just in case.

sepalani:
This PR should probably wait https://github.com/dolphin-emu/dolphin/pull/10367 to be merged to make sure it doesn't break if it needs adjustments.

Source/Core/Core/Debugger/HLEFormatter.cpp
Pokechu22: I believe it's possible to redefine `FMT_THROW` and `FMT_ASSERT`, but that could be a bit fiddly.  It'd be nice to have them hooked in to Dolphin's panic alert system (in most cases) though.

sepalani: Sadly I don't we can because both `FMT_THROW` and `FMT_ASSERT` are used in `constexpr` functions _(in `fmt/core.h`)_ and our panic alert system isn't `constexpr`.

sepalani:
Rebased and still ready to be reviewed & tested.

Some screenshots:
image
image

Source/Core/Core/PowerPC/BreakPoints.cpp
```diff
@@ -113,6 +118,7 @@ void BreakPoints::Add(u32 address, bool temp, bool break_on_hit, bool log_on_hit
   bp.break_on_hit = break_on_hit;
   bp.log_on_hit = log_on_hit;
   bp.address = address;
+  bp.message = message;
```
iwubcode, line 132: `bp.message = std::move(message);`

Source/Core/DolphinQt/Debugger/BreakpointWidget.cpp
```diff
@@ -388,16 +390,17 @@ void BreakpointWidget::AddBP(u32 addr)
   AddBP(addr, false, true, true);
 }
 
-void BreakpointWidget::AddBP(u32 addr, bool temp, bool break_on_hit, bool log_on_hit)
+void BreakpointWidget::AddBP(u32 addr, bool temp, bool break_on_hit, bool log_on_hit,
+                             std::string message)
 {
-  PowerPC::breakpoints.Add(addr, temp, break_on_hit, log_on_hit);
+  PowerPC::breakpoints.Add(addr, temp, break_on_hit, log_on_hit, message);
```
iwubcode, line 396: `PowerPC::breakpoints.Add(addr, temp, break_on_hit, log_on_hit, std::move(message));`

Source/Core/DolphinQt/Debugger/BreakpointWidget.cpp
```diff
@@ -408,6 +411,7 @@ void BreakpointWidget::AddAddressMBP(u32 addr, bool on_read, bool on_write, bool
   check.is_break_on_write = on_write;
   check.log_on_hit = do_log;
   check.break_on_hit = do_break;
+  check.message = message;
```
iwubcode, line 414: `check.message = std::move(message);`

Source/Core/DolphinQt/Debugger/BreakpointWidget.cpp
```diff
@@ -430,6 +434,7 @@ void BreakpointWidget::AddRangedMBP(u32 from, u32 to, bool on_read, bool on_writ
...
```
iwubcode, line 437: `check.message = std::move(message);`

Source/Core/DolphinQt/Debugger/NewBreakpointDialog.cpp
```diff
@@ -161,6 +165,7 @@ void NewBreakpointDialog::accept()
   // Actions
   bool do_log = m_do_log->isChecked() || m_do_log_and_break->isChecked();
   bool do_break = m_do_break->isChecked() || m_do_log_and_break->isChecked();
+  const std::string message = m_log_message->text().toStdString();
```
iwubcode, line 170: If you remove the `const` you can move the message in the below calls

Source/Core/Core/Debugger/HLEFormatter.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+HLENamedArgs GetHLENamedArgs()
+{
+  auto args = HLENamedArgs();
+  std::string name;
+
+  for (int i = 0; i < 32; i++)
+  {
+    // General purpose registers (int)
+    name = "r" + std::to_string(i);
```
iwubcode, line 307: `name = fmt::format("r{}", i);`

(same with others)

Source/Core/Core/PowerPC/BreakPoints.cpp
```diff
@@ -80,10 +80,11 @@ BreakPoints::TBreakPointsStr BreakPoints::GetStrings() const
 
 void BreakPoints::AddFromStrings(const TBreakPointsStr& bp_strings)
 {
-  for (const std::string& bp_string : bp_strings)
+  for (std::string bp_string : bp_strings)
```
iwubcode, line 83: Is the copy here intentional?

sepalani: Yes, it's needed for the StringPopFrontIf function.

sepalani:
@iwubcode Done.

I also fixed a typo in the help message since `lx` has to be used to handle all the bits of the paired single:
> Paired Single: {f1} = ({f1_1:016**l**x},{f1_2:016**l**x})

Source/Core/Core/PowerPC/BreakPoints.cpp
```diff
@@ -78,6 +78,10 @@ void BreakPoints::AddFromStrings(const TBreakPointsStr& bp_strings)
 
     iss >> std::hex >> bp.address;
     iss >> flags;
+    std::getline(iss, bp.message);
+    if (!bp.message.empty() && bp.message.front() == ' ')
```
iwubcode: Dumb question...is there a reason we only erase the first space?  Should we use `StripSpaces` instead?

Source/VSProps/Base.props
```diff
@@ -63,6 +63,7 @@
       <ExternalTemplatesDiagnostics>true</ExternalTemplatesDiagnostics>
       <DisableAnalyzeExternal>true</DisableAnalyzeExternal>
       <TreatAngleIncludeAsExternal>true</TreatAngleIncludeAsExternal>
+      <PreprocessorDefinitions>FMT_EXCEPTIONS=0;FMT_ASSERT=__noop;%(PreprocessorDefinitions)</PreprocessorDefinitions>
```
iwubcode, line 66: Does this need to be done for the cmake builds as well?

sepalani: It shouldn't be needed since C++ exceptions are only enabled on Windows build.

sepalani: Oh right, I completely forgot that was a thing.

I added it and tested with Visual Studio + CMake.

Source/Core/Core/Debugger/HLEFormatter.cpp
```diff
@@ -22,6 +29,232 @@ struct fmt::formatter<PowerPC::PairedSingle>
...
+
+    switch (base_type)
+    {
+    default:
+    case BaseType::Default:
+    {
+      if (arg.IsFloat())
+      {
+        [[fallthrough]];
+      case BaseType::Double:
```
iwubcode, line 234: I'm shocked this works.  Tbh, I don't find this very readable.

Is this just to avoid some duplication in something like this:

```
switch(base_type)
{
default:
case BaseType::Default:
{
  if (arg.IsFloat())
    return fmt::format_to(ctx.out(), format_string, arg.Value<double>());
  return fmt::format_to(ctx.out(), format_string, arg.Value<s64>());
}
case BaseType::Double:
  return fmt::format_to(ctx.out(), format_string, arg.Value<double>());
case BaseType::Int64:
  return fmt::format_to(ctx.out(), format_string, arg.Value<s64>());
...
}
```

Pokechu22: See https://en.wikipedia.org/wiki/Duff%27s_device for a related thing.

Source/Core/Core/Debugger/HLEFormatter.cpp
```diff
@@ -22,6 +29,232 @@ struct fmt::formatter<PowerPC::PairedSingle>
...
+    }
+
+    // Let {fmt} handle the invalid end iterator
+    const auto error = begin;
+
+    // Search the closing brace
+    const auto format_end = std::find(begin, end, '}');
+    if (format_end == end)
+      return error;
+    base_format = "{:" + std::string(begin, format_end + 1);
```
iwubcode, line 205: I wish fmt  had more recent concatenation benchmarks.  I don't feel strongly about it but regardless, this could be:

`base_format = fmt::format("{:{}", std::string(begin, format_end + 1));`

Also, could you possibly put this down below?

```
if (base_type == BaseType::Pointer)
      base_format = "{:#010x}";
else
      base_format = fmt::format("{:{}", std::string(begin, format_end + 1));
```

sepalani: Good catch. Done.

Source/Core/Core/Debugger/HLEFormatter.cpp
sepalani: @iwubcode 
Yeah, I used this syntax to avoid code duplication. You can think of `switch-case` like some kind of glorified `labels:`/`goto`. This technique is used in some memcpy implementation using Duff's device, there are also some concepts using it as well such as protothread (some kind of stackless/lightweight threads).

iwubcode: I'm not a performance dev so I usually prefer readability unless something is a known hit.  This seems like premature optimization but I have no clue, maybe it could be a hotspot if this is ran often.  Regardless, I had not heard of this before, appreciate you both giving me details about it.  i will read some more.  Thanks!

Source/Core/Core/Debugger/HLEFormatter.cpp
```diff
@@ -252,6 +264,23 @@ struct fmt::formatter<HLEArg>
       return fmt::format_to(ctx.out(), format_string, arg.Value<u32>());
     case BaseType::Uint64:
       return fmt::format_to(ctx.out(), format_string, arg.Value<u64>());
+    case BaseType::VarArgs:
+    {
+      const auto gpr_index = arg.GprIndex();
+      if (!gpr_index || *gpr_index < 3)
```
iwubcode, line 270: What is the significance of three here.  Can you give an example?

Pokechu22: `r3` is the first register used for the first parameter in the eabi calling convention, I think? (With `r4` through `r10` being further parameters, `r1` being the stack pointer, `r2` and `r13` being used for the small data area (i.e. pointers to globals) `r0`, `r11`, `r12` having some other purpose, and `r14`-`r31` being used for local variables (callee-saved)). Not sure if we have a helper for this though.)

iwubcode: Thanks @Pokechu22  for sharing your knowledge!  I see `GetStringVA` also has a default value of 3, so it seems 0, 1, 2 are not valid.  I'm curious why that is, when you mention others outside that range as well.

I think I can take a leap of faith for the review (it follows conventions) but it'd be nice to know for my own understanding.

sepalani: The reasoning behind it is to support format string used by printf (using varargs) and vprintf (using va_list). Dolphin emulates it since https://github.com/dolphin-emu/dolphin/pull/4481 in HLE debug message (HLE OSReport) when printf/vprintf-like function signatures are detected.

GC/Wii/Homebrew applications seems to follow the [SVR4 ABI](http://refspecs.linux-foundation.org/elf/elfspec_ppc.pdf) regarding how parameters should be passed. You can search the keywoard `varargs` and `va_list` in the PDF to find the relevant chapters. [Wiibrew](https://wiibrew.org/wiki/Assembler_Tutorial#Application_Binary_Interface_.28SVR4_ABI.29) also has a section talking about the ABI.

iwubcode (approved):
Lots of logic I can't speak for in terms of correctness but the code looks reasonable.  Untested.

AdmiralCurtiss:
For those cases where you use `StringPopFrontIf()`: It likely doesn't matter much here because the strings aren't super long, but you'd be better off making a `string_view` instead and removing the first character off of that.

Source/Core/Core/Debugger/HLEFormatter.cpp
AdmiralCurtiss: +1 from me for this being very weird and difficult to read.

Source/Core/Core/PowerPC/BreakPoints.cpp
```diff
@@ -78,6 +78,10 @@ void BreakPoints::AddFromStrings(const TBreakPointsStr& bp_strings)
 
     iss >> std::hex >> bp.address;
     iss >> flags;
+    std::getline(iss, bp.message);
```
AdmiralCurtiss, line 93: This breaks if you have a breakpoint with no flags, since the `>>` skips an arbitrary amount of whitespace so `flags` gets the first word of the message. Might need to redesign how you store these or rewrite the parsing logic.

Source/Core/Core/Debugger/HLEFormatter.cpp
sepalani: Alright, I'm going to replace it with a more readable syntax instead.

Source/Core/Core/PowerPC/BreakPoints.cpp
sepalani: This shouldn't normally occur unless someone messes up its config file manually because the BP should have either the `break` or `log` flag set. I can add an additional check if you really care. Ditto for MBP which has the same issue.

AdmiralCurtiss: If it's not possible for a breakpoint to have no flags then fair enough. Might still be worth it in case the UI allows this in the future?

PR #10200, opened 2021-11-04
Add this resolution options 1.5x

JaePlays:
Screenshot_2021-11-04-08-03-29-794_org dolphinemu mmjr


phire:
Merge 10,000 commits, with 1.5 million lines changed. 

No.

phire:
This is not an issue tracker where you can make feature requests. They should go to https://bugs.dolphin-emu.org

PR #10201, opened 2021-11-04
GameINI: Fix Comment Causing Parsing Errors in Mario Party 6/7 INIs.

JMC47:
It appears that this code is a Gecko Code and was mistakenly put as an Action Replay code.  I may be wrong and it may be a different problem, but based on the how the code looks I'm confident this was the problem.  The code didn't crash the game when I moved it to Gecko, and it was causing problems when it was under Action Replay.

### DIFF
--- a/Data/Sys/GameSettings/GP6E01.ini
+++ b/Data/Sys/GameSettings/GP6E01.ini
@@ -441,7 +441,7 @@ $Game - Orb Expansion [Airsola]
 042bd36c 00000000
 042bd37c 00000000
 00000000 40000000
-*(Almost) all Orbs become obtainable on each board. Orb prices are rebalanced.
+# Almost all Orbs become obtainable on each board. Orb prices are rebalanced.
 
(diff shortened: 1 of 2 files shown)


PatrickFerry:
I think you should do the same for GP7E01.ini

JMC47:
GP7E01 isn't complaining as far as I can tell, and I don't know for certain if this is the problem.  I'm not an expert.  If you know for sure it's wrong, I can move it.

PatrickFerry:
I just seen that it was the same creator with the same goal in mind.

An obvious issue to me is the asterisk in the comments for the action replay code aren't valid. The comments under action replay codes would have to be using a hash unlike the gecko codes

JMC47:
I didn't know that; maybe that's the problem and it is an action replay code?

JMC47:
Testing to see if this fixes it properly.

JMC47:
Everything seems to be working on this now.

PR #10204, opened 2021-11-07
VideoCommon: Use the copy filter for EFB copies as well as XFB copies

Pokechu22:
This fixes the pink screens in EA Sports Active.  See https://gist.github.com/Pokechu22/49455f9094ed0ff017da64e3f7aa0404 for details.

This will also need to be implemented for the software renderer, which has its own implementation of EFB copies.  Unfortunately it doesn't seem like there is a quick fix for it, as the relevant code (in [TextureEncoder.cpp](https://github.com/dolphin-emu/dolphin/blob/2c5d11cacebe05450ee4c5d276044e110ed286c1/Source/Core/VideoBackends/Software/TextureEncoder.cpp#L1423), while XFBs are in [EfbInterface.cpp](https://github.com/dolphin-emu/dolphin/blob/2c5d11cacebe05450ee4c5d276044e110ed286c1/Source/Core/VideoBackends/Software/EfbInterface.cpp#L556)) is pretty complicated.

### DIFF
--- a/Source/Core/VideoCommon/BPStructs.cpp
+++ b/Source/Core/VideoCommon/BPStructs.cpp
@@ -275,13 +275,12 @@ static void BPWritten(const BPCmd& bp)
     {
       // bpmem.zcontrol.pixel_format to PixelFormat::Z24 is when the game wants to copy from ZBuffer
       // (Zbuffer uses 24-bit Format)
-      static constexpr CopyFilterCoefficients::Values filter_coefficients = {
-          {0, 0, 21, 22, 21, 0, 0}};
       bool is_depth_copy = bpmem.zcontrol.pixel_format == PixelFormat::Z24;
       g_texture_cache->CopyRenderTargetToTexture(
           destAddr, PE_copy.tp_realFormat(), copy_width, copy_height, destStride, is_depth_copy,
           srcRect, PE_copy.intensity_fmt, PE_copy.half_scale, 1.0f, 1.0f,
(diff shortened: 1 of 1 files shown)


iwubcode (approved):

phire:
I'm super confused as to why Spyro-depth is partially fixed by this PR. I thought I had that bisected to a texture cache issue.

phire (approved):

Pokechu22:
If you look at [fifoci itself](https://fifo.ci/dff/spyro-depth/) the issue did first appear in #8350 so your bisect is probably right.  For reference, here's a diff for this PR compared to prior to that PR: [mesa](https://fifo.ci/compare/6952309-6828086/), [radeon](https://fifo.ci/compare/6952309-6828086/).

With spyro-depth, most EFB copies have the copy filter set as (0, 0, 1, 63, 0, 0, 0), but others have it set to (0, 0, 0, 63, 0, 0, 0), which would be *slightly* darker.  A difference in brightness is likely to make a texture's hash change, so probably this resolved a few of the texture cache collisions simply by chance.

I also don't experience that blackness on my machine when using the fifolog (with vulkan and texture cache set to fast).  Not sure why.

phire:
> If you look at fifoci itself the issue did first appear in #8350 so your bisect is probably right. 

Yeah, it was a known regression of #8350. But further investigation showed that Minimal TMEM was hiding the actual source of the regression, and I tracked it down to cdbd9863b. Or at least I thought I did, there were two related regressions and maybe I didn't confirm that both bisected to that commit.

PR #7568, opened 2018-11-12
RFC: Enable sustained performance mode on Android

JonnyH:
Is this useful? Should we add it to a UI option?

PR #10206, opened 2021-11-07
Android: Enable sustained performance mode

JosJuice:
Rebase of PR #7568, mostly for the sake of producing a build to test with.

### DIFF
--- a/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/activities/EmulationActivity.kt
+++ b/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/activities/EmulationActivity.kt
@@ -130,6 +130,10 @@ class EmulationActivity : AppCompatActivity(), ThemeProvider {
         if (NativeLibrary.IsGameMetadataValid())
             title = NativeLibrary.GetCurrentTitleDescription()
 
+        if (Build.VERSION.SDK_INT >= 24) {
+          getWindow().setSustainedPerformanceMode(true)
+        }
+
         if (skylanderSlots.isEmpty()) {
             for (i in 0..7) {
                 skylanderSlots.add(SkylanderSlot(getString(R.string.skylander_slot, i + 1), i))


Gamemulatorer:
So this is enable by default?

JosJuice:
In this PR as it currently exists, it is always enabled.

DatEmulationGuy:
For me the sustained performance build doesn't seem to help with the throttling because the heavy performance loss begins in similar battery temperatures as with a normal recent dev build. Enabling Game Turbo and using the performance optimisation setting is the only way I found to avoid throttling in Dolphin on my stock MIUI, this also helps in Citra for the record and possibly other emulators and  native Android games and whatnot

Poco X3 Pro with 12.5.3.0 EEA MIUI ROM

DatEmulationGuy:
> For me the sustained performance build doesn't seem to help with the throttling because the heavy performance loss begins in similar battery temperatures as with a normal recent dev build. Enabling Game Turbo and using the performance optimisation setting is the only way I found to avoid throttling in Dolphin on my stock MIUI, this also helps in Citra for the record and possibly other emulators and native Android games and whatnot
> 
> Poco X3 Pro with 12.5.3.0 EEA MIUI ROM

Not sure if this is right but PPSSPP says sustained performance not supported by my OS, that would explain the identical results in my tests

Gamemulatorer:
> > For me the sustained performance build doesn't seem to help with the throttling because the heavy performance loss begins in similar battery temperatures as with a normal recent dev build. Enabling Game Turbo and using the performance optimisation setting is the only way I found to avoid throttling in Dolphin on my stock MIUI, this also helps in Citra for the record and possibly other emulators and native Android games and whatnot
> > Poco X3 Pro with 12.5.3.0 EEA MIUI ROM
> 
> Not sure if this is right but PPSSPP says sustained performance not supported by my OS, that would explain the identical results in my tests

Same happen to duckstaion https://github.com/hrydgard/ppsspp/issues/11018#issuecomment-954416433

DatEmulationGuy:
> > > For me the sustained performance build doesn't seem to help with the throttling because the heavy performance loss begins in similar battery temperatures as with a normal recent dev build. Enabling Game Turbo and using the performance optimisation setting is the only way I found to avoid throttling in Dolphin on my stock MIUI, this also helps in Citra for the record and possibly other emulators and native Android games and whatnot
> > > Poco X3 Pro with 12.5.3.0 EEA MIUI ROM
> > 
> > 
> > Not sure if this is right but PPSSPP says sustained performance not supported by my OS, that would explain the identical results in my tests
> 
> Same happen to DuckStation [hrydgard/ppsspp#11018 (comment)](https://github.com/hrydgard/ppsspp/issues/11018#issuecomment-954416433)

I'm confused, this sustained performance mode was removed from new Android versions or something?

Gamemulatorer:
I tried this and don't feel anything is changes in terms of performance 🤔

Gamemulatorer:
Just related to android performance :)
IMG_20211217_220154
yesterday saw it mentioned in Android documentation (search for affinity in https://developer.android.com/games/optimize ).

Maybe that could help speed up dolphin emu on android 😅

dreamsyntax:
What needs to be tested here / what led to this being re-opened?

JosJuice:
I reopened it because someone on the forums wanted a build, and will probably close it again later. I don't have any plans to merge this unless it turns out this does have some noticeable effect.

dreamsyntax:
Can you please mark this as a draft

JosJuice:
Sure, that makes sense.

landfillbaby:
> Just related to android performance :) IMG_20211217_220154 yesterday saw it mentioned in Android documentation (search for affinity in https://developer.android.com/games/optimize ).
> 
> Maybe that could help speed up dolphin emu on android 😅

it definitely would help on my device since i get a major speed increase to above 100% only during system ui updates (pulling down notification bar, messing with left menu, using touch controls, etc.)

PR #10207, opened 2021-11-08
Implement preliminary Triforce support Pt.2

Zopolis4:
I'm splitting these changes into a new PR because I don't want the old one to get too bloated. This PR will be working on the EXI baseboard connection, and should hopefully get some more games booting. Obviously, this depends on #10084, and when that is merged I'll fully clean up the commit history on this branch. Right now this is just here to get some feedback on the latest changes.

TODO:
- Implement backup support
- Emulate EXI interrupts
- (Squash rework of command handling into original commit?)


Zopolis4:
Marking this as ready for review because it's.... ready for review. Not for merge though, it still depends on #10084 and i'll rebase this after it gets merged. No compatibility improvements here unfortunately, it still depends on some SI stuff. I'll be taking a break for a while, and it'll be quite a while before any SI commits show up, so hopefully I won't have to make a part 3.

Pokechu22:
@dolphin-emu-bot rebuild

Zopolis4:
Note: this is still very much a WIP, seeing as the triforce name code _still doesn't work for some reason_, and I need to implement a few other EXI commands.

Source/Core/Core/CMakeLists.txt
```diff
@@ -232,6 +232,9 @@ add_library(core
   HW/ProcessorInterface.h
   HW/SI/SI_Device.cpp
   HW/SI/SI_Device.h
+  HW/SI/SI_DeviceBaseboard.h
+  HW/SI/SI_DeviceBaseboard.cpp
+  HW/SI/SI_DeviceDanceMat.h
   HW/SI/SI_DeviceDanceMat.cpp
```
Rumi-Larry, line 238: You just added the header for the dancemat again

LPFan12:
What is the pulse on this PR, considering that the original post says that it would come back after Part 1 is merged and that happened a while ago now? Is it still alive or is it effectively cancelled?

Zopolis4:
Unfortunately, circumstances have changed since I originally began working on this, and so it is unlikely I will be picking this back up in the forseeable future.

PR #10209, opened 2021-11-11
Assertion and panic alert improvements

Pokechu22:
* The failed condition and the relevant function are now included in the assert message.  This applies even when using `ASSERT_MSG`; now, all relevant information is shown with that too.
* `ASSERT_MSG` now uses fmt.  I've updated all uses of this (and in some cases added additional details to the message).
* The `CHECK` macro in the D3D backends has been removed and replaced with `ASSERT_MSG`.
* I added an `HRWrap` struct that can be used to format `HRESULT` in messages; a `fmt::formatter` exists for it.  This uses `_com_result` based on [this](https://stackoverflow.com/a/7008111).  This gives _MUCH_ more useful information (compared to both just a hex value, and to no information at all (which was the norm for most uses)) and it should be fairly user-friendly.  ~~I put this in `D3DCommon.h`, but maybe this should be somewhere else so that it could be used in other code.~~  This is now located in `Common/HRWrap.h`, and used by a few places outside of D3D.  D3D11 and D3D12 both have their own variants that call `GetDeviceRemovedReason` when needed.
* The legacy non-format variants of `PanicAlert` have been removed; they're no longer used by anything.
* `ASSERT_MSG`'s first parameter now indicating the log type is now actually used.
* `PanicAlertFmtT` now actually works.  Yes, this was broken.  The values were extracted for translation beforehand, but they were never actually translated at runtime.  You can test this by modifying `dsp_rom.bin` so that it has an invalid hash, and then setting the language to French; the message will still be in English despite [a translation for it existing](https://github.com/dolphin-emu/dolphin/blob/aa6db1e029f513b3eb4b1d418a61dbc8eca20a41/Languages/po/fr.po#L11687-L11710).  This will hopefully make it easier for people to understand these messages (though it may make things a bit harder for support).

<details><summary>Before</summary>

```patch
diff --git a/Source/Core/Core/Core.cpp b/Source/Core/Core/Core.cpp
index 622d7f9d90..9174a0ca94 100644
--- a/Source/Core/Core/Core.cpp
+++ b/Source/Core/Core/Core.cpp
@@ -20,6 +20,7 @@
 
 #include "AudioCommon/AudioCommon.h"
 
+#include "Common/Assert.h"
 #include "Common/CPUDetect.h"
 #include "Common/CommonPaths.h"
@@ -439,6 +440,8 @@ static void FifoPlayerThread(const std::optional<std::string>& savestate_path,
 // See the BootManager.cpp file description for a complete call schedule.
 static void EmuThread(std::unique_ptr<BootParameters> boot, WindowSystemInfo wsi)
 {
+  ASSERT(false);
+  ASSERT_MSG(CORE, false, "Test: %d", 42);
   const SConfig& core_parameter = SConfig::GetInstance();
   CallOnStateChangedCallbacks(State::Starting);
   Common::ScopeGuard flag_guard{[] {
```

would result in these dialogs:

image
image

and this being logged:

```
27:16:705 Common\MsgHandler.cpp:112 E[MASTER]: Warning: An error occurred.

  Line: 443
  File: C:\Users\Pokechu22\source\repos\dolphin\Source\Core\Core\Core.cpp

Ignore and continue?
27:20:730 Common\MsgHandler.cpp:112 E[MASTER]: Warning: Test: 42
```
</details>
<details><summary>After</summary>

```patch
diff --git a/Source/Core/Core/Core.cpp b/Source/Core/Core/Core.cpp
index 622d7f9d90..c447c208e1 100644
--- a/Source/Core/Core/Core.cpp
+++ b/Source/Core/Core/Core.cpp
@@ -20,6 +20,7 @@

 #include "AudioCommon/AudioCommon.h"

+#include "Common/Assert.h"
 #include "Common/CPUDetect.h"
 #include "Common/CommonPaths.h"
@@ -439,6 +440,8 @@ static void FifoPlayerThread(const std::optional<std::string>& savestate_path,
 // See the BootManager.cpp file description for a complete call schedule.
 static void EmuThread(std::unique_ptr<BootParameters> boot, WindowSystemInfo wsi)
 {
+  ASSERT(false);
+  ASSERT_MSG(CORE, false, "Test: {}", 42);
   const SConfig& core_parameter = SConfig::GetInstance();
   CallOnStateChangedCallbacks(State::Starting);
   Common::ScopeGuard flag_guard{[] {
```

would result in these dialogs:

image
image

and this being logged:

```
23:50:874 Common\MsgHandler.cpp:113 E[MASTER]: Warning: An error occurred.

  Condition: false
  File: C:\Users\Pokechu22\source\repos\dolphin\Source\Core\Core\Core.cpp
  Line: 443
  Function: EmuThread

Ignore and continue?
23:59:503 Common\MsgHandler.cpp:113 E[CORE]: Warning: An error occurred.

Test: 42

  Condition: false
  File: C:\Users\Pokechu22\source\repos\dolphin\Source\Core\Core\Core.cpp
  Line: 444
  Function: EmuThread

Ignore and continue?
```
</details>

### DIFF
--- a/Source/Core/Common/Arm64Emitter.cpp
+++ b/Source/Core/Common/Arm64Emitter.cpp
@@ -577,13 +577,14 @@ void ARM64XEmitter::EncodeAddSubImmInst(u32 op, bool flags, u32 shift, u32 imm, 
 
 void ARM64XEmitter::EncodeLogicalImmInst(u32 op, ARM64Reg Rd, ARM64Reg Rn, LogicalImm imm)
 {
-  ASSERT_MSG(DYNAREC, imm.valid, "Invalid logical immediate");
+  ASSERT_MSG(DYNA_REC, imm.valid, "Invalid logical immediate");
 
   // Sometimes Rd is fixed to SP, but can still be 32bit or 64bit.
   // Use Rn to determine bitness here.
   bool b64Bit = Is64Bit(Rn);
 
-  ASSERT_MSG(DYNAREC, b64Bit || !imm.n, "64-bit logical immediate does not fit in 32-bit register");
+  ASSERT_MSG(DYNA_REC, b64Bit || !imm.n,
+             "64-bit logical immediate does not fit in 32-bit register");
 
   Write32((b64Bit << 31) | (op << 29) | (0x24 << 23) | (imm.n << 22) | (imm.r << 16) |
           (imm.s << 10) | (DecodeReg(Rn) << 5) | DecodeReg(Rd));
--- a/Source/Core/Common/Arm64Emitter.cpp
+++ b/Source/Core/Common/Arm64Emitter.cpp
@@ -5,12 +5,13 @@
 
 #include <algorithm>
 #include <array>
-#include <cinttypes>
 #include <cstring>
 #include <optional>
 #include <tuple>
 #include <utility>
 #include <vector>
+
+#include <fmt/format.h>
 
 #include "Common/Align.h"
 #include "Common/Assert.h"
@@ -307,13 +308,12 @@ void ARM64XEmitter::EncodeCompareBranchInst(u32 op, ARM64Reg Rt, const void* ptr
   bool b64Bit = Is64Bit(Rt);
   s64 distance = (s64)ptr - (s64)m_code;
 
-  ASSERT_MSG(DYNA_REC, !(distance & 0x3), "%s: distance must be a multiple of 4: %" PRIx64,
-             __func__, distance);
+  ASSERT_MSG(DYNA_REC, !(distance & 0x3), "Distance must be a multiple of 4: {}", distance);
 
   distance >>= 2;
 
   ASSERT_MSG(DYNA_REC, distance >= -0x40000 && distance <= 0x3FFFF,
-             "%s: Received too large distance: %" PRIx64, __func__, distance);
+             "Received too large distance: {}", distance);
 
   Write32((b64Bit << 31) | (0x34 << 24) | (op << 24) | (((u32)distance << 5) & 0xFFFFE0) |
           DecodeReg(Rt));
@@ -325,13 +325,12 @@ void ARM64XEmitter::EncodeTestBranchInst(u32 op, ARM64Reg Rt, u8 bits, const voi
   u8 b5 = (bits >> 5) & 0x1;
   s64 distance = (s64)ptr - (s64)m_code;
 
-  ASSERT_MSG(DYNA_REC, !(distance & 0x3), "%s: distance must be a multiple of 4: %" PRIx64,
-             __func__, distance);
+  ASSERT_MSG(DYNA_REC, !(distance & 0x3), "distance must be a multiple of 4: {}", distance);
 
   distance >>= 2;
 
-  ASSERT_MSG(DYNA_REC, distance >= -0x3FFF && distance < 0x3FFF,
-             "%s: Received too large distance: %" PRIx64, __func__, distance);
+  ASSERT_MSG(DYNA_REC, distance >= -0x3FFF && distance < 0x3FFF, "Received too large distance: {}",
+             distance);
 
   Write32((b5 << 31) | (0x36 << 24) | (op << 24) | (b40 << 19) |
           ((static_cast<u32>(distance) << 5) & 0x7FFE0) | DecodeReg(Rt));
@@ -341,13 +340,12 @@ void ARM64XEmitter::EncodeUnconditionalBranchInst(u32 op, const void* ptr)
 {
   s64 distance = (s64)ptr - s64(m_code);
 
-  ASSERT_MSG(DYNA_REC, !(distance & 0x3), "%s: distance must be a multiple of 4: %" PRIx64,
-             __func__, distance);
+  ASSERT_MSG(DYNA_REC, !(distance & 0x3), "distance must be a multiple of 4: {}", distance);
 
   distance >>= 2;
 
   ASSERT_MSG(DYNA_REC, distance >= -0x2000000LL && distance <= 0x1FFFFFFLL,
-             "%s: Received too large distance: %" PRIx64, __func__, distance);
+             "Received too large distance: {}", distance);
 
   Write32((op << 31) | (0x5 << 26) | (distance & 0x3FFFFFF));
 }
@@ -359,8 +357,7 @@ void ARM64XEmitter::EncodeUnconditionalBranchInst(u32 opc, u32 op2, u32 op3, u32
 
 void ARM64XEmitter::EncodeExceptionInst(u32 instenc, u32 imm)
 {
-  ASSERT_MSG(DYNA_REC, !(imm & ~0xFFFF), "%s: Exception instruction too large immediate: %d",
-             __func__, imm);
+  ASSERT_MSG(DYNA_REC, !(imm & ~0xFFFF), "Exception instruction too large immediate: {}", imm);
 
   Write32((0xD4 << 24) | (ExcEnc[instenc][0] << 21) | (imm << 5) | (ExcEnc[instenc][1] << 2) |
           ExcEnc[instenc][2]);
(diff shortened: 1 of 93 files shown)


Pokechu22:
@dolphin-emu-bot rebuild

Source/Core/Core/ARDecrypt.cpp
```diff
@@ -476,14 +476,15 @@ void DecryptARCode(std::vector<std::string> vCodes, std::vector<AREntry>* ops)
   else if (!batchdecrypt(uCodes.data(), (u16)vCodes.size() << 1))
   {
     // Commented out since we just send the code anyways and hope for the best XD
-    // PanicAlert("Action Replay Code Decryption Error:\nCRC Check Failed\n\n"
-    // "First Code in Block(should be verification code):\n%s", vCodes[0].c_str());
+    // PanicAlertFmt("Action Replay Code Decryption Error:\nCRC Check Failed\n\n"
+    //               "First Code in Block(should be verification code):\n{}",
```
Rumi-Larry, line 480: Missing space after "Block".

Pokechu22:
I've added another commit, which uses the caller's file and line number for panic alerts instead of the line number of the log call in `ShowMessageAlert` in `MsgAlert.cpp`.

<details><summary>Example</summary>

```
46:45:173 Core\Core.cpp:443 E[MASTER]: Warning: An error occurred.

  Condition: false
  File: C:\Users\Pokechu22\source\repos\dolphin\Source\Core\Core\Core.cpp
  Line: 443
  Function: EmuThread

Ignore and continue?
46:49:162 Core\Core.cpp:444 E[CORE]: Warning: An error occurred.

Test: 42

  Condition: false
  File: C:\Users\Pokechu22\source\repos\dolphin\Source\Core\Core\Core.cpp
  Line: 444
  Function: EmuThread

Ignore and continue?
```
</details>

Source/Core/Common/MsgHandler.h
```diff
@@ -30,22 +31,35 @@ void RegisterStringTranslator(StringTranslator translator);
...
+                  const Args&... args)
+{
+  static_assert(!has_non_positional_args,
+                "Translatable strings must use positional arguments (e.g. {0} instead of {})");
+  static_assert(NumFields == sizeof...(args),
+                "Unexpected number of replacement fields in format string; did you pass too few or "
+                "too many arguments?");
+  // This is purely to verify that things work in the base language (English).
+  // We can't compile-time check translations unfortunately.
+  fmt::make_args_checked<Args...>(format, args...);
```
BhaaLseN, line 60: Is this guaranteed to be called, vs. a compiler simply omitting it? Should we explicitly mark it as ignored/discarded?

Source/Core/Common/x64Emitter.cpp
```diff
@@ -309,7 +309,7 @@ void OpArg::WriteRest(XEmitter* emit, int extraBytes, X64Reg _operandReg,
     s64 distance = (s64)offset - (s64)ripAddr;
     ASSERT_MSG(DYNA_REC,
                (distance < 0x80000000LL && distance >= -0x80000000LL) || !warn_64bit_offset,
-               "WriteRest: op out of range (0x%" PRIx64 " uses 0x%" PRIx64 ")", ripAddr, offset);
```
BhaaLseN, line 312: Do we still need `#include <cinttypes>` after removing this use of `PRIx64`?

Source/Core/Common/MsgHandler.h
Pokechu22: We're calling it purely for the static assertions it has; the result doesn't matter (and in fact, it would be good if the compiler optimized out the runtime-related aspects; it probably does).

BhaaLseN: Yeah, but does this cause the static assert to trigger even when it is optimized out? If so, good, and you can ignore this one :D

Pokechu22: I would be very annoyed if compilers could look at a function and decide that because it has no side-effects at runtime, its `static_assert`s can be ignored.  That's not to say that's not how it works, just that it would be very annoying if that was how it works.

But on further investigation, we can use the arg-list from the English string as a parameter for the non-English string; both `make_format_args` and `make_args_checked` return `format_arg_store`, and `make_args_checked`'s return value doesn't depend on the format string (only the checking logic), so the return value doesn't need to go unused.

Source/Core/Common/x64Emitter.cpp
Pokechu22: Nope.  Same applies to Arm64Emitter.

In fact, all inclusions of `<cinttypes>` are unused other than the ones in MenuBar (where it is used to create title IDs) in RenderBase (where it is used with ImGui, which uses printf strings).  I've added another commit to remove these.

Source/Core/Common/MsgHandler.h
```diff
@@ -41,21 +41,28 @@ bool MsgAlertFmt(bool yes_no, MsgType style, Common::Log::LogType log_type, cons
...
   static_assert(NumFields == sizeof...(args),
                 "Unexpected number of replacement fields in format string; did you pass too few or "
                 "too many arguments?");
-  return MsgAlertFmtImpl(yes_no, style, log_type, format,
-                         fmt::make_args_checked<Args...>(format, args...));
+  static_assert(fmt::is_compile_string<S>::value);
+  // It's only possible for us to compile-time check the English-language string.
+  // make_args_checked uses static_asserts to verify that a string is formattable with the g
+  // arguments.  But it can't do that if the string varies at runtime, so we can't check
+  // translations.  Still, verifying that the English string is correct will help ensure tha
```
sepalani, line 63: Ditto for `ensure tha`

Source/Core/Common/MsgHandler.h
```diff
@@ -41,21 +41,28 @@ bool MsgAlertFmt(bool yes_no, MsgType style, Common::Log::LogType log_type, cons
...
   static_assert(!has_non_positional_args,
                 "Translatable strings must use positional arguments (e.g. {0} instead of {})");
   static_assert(NumFields == sizeof...(args),
                 "Unexpected number of replacement fields in format string; did you pass too few or "
                 "too many arguments?");
-  return MsgAlertFmtImpl(yes_no, style, log_type, format,
-                         fmt::make_args_checked<Args...>(format, args...));
+  static_assert(fmt::is_compile_string<S>::value);
+  // It's only possible for us to compile-time check the English-language string.
+  // make_args_checked uses static_asserts to verify that a string is formattable with the g
```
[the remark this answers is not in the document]
Pokechu22: Yep, I think caused by me copying the comment from a diff when fixing a merge conflict by using the rectangle selection, and I missed the last few columns.  This was supposed to be "the given".

Source/Core/Common/Arm64Emitter.cpp
```diff
@@ -1434,23 +1429,23 @@ void ARM64XEmitter::BFI(ARM64Reg Rd, ARM64Reg Rn, u32 lsb, u32 width)
 {
   u32 size = Is64Bit(Rn) ? 64 : 32;
   ASSERT_MSG(DYNA_REC, lsb < size && width >= 1 && width <= size - lsb,
-             "%s passed lsb %d and width %d which is greater than the register size!", __func__,
+             "{} passed lsb {} and width {} which is greater than the register size!", __func__,
```
JosJuice, line 1432: Any particular reason why you removed the use of `__func__` earlier in the file but stopped here?

Pokechu22: I think it was partially getting distracted partway through, and partly because this message is one where it would need to be reworded more than just removing the function name followed by a colon.  I could get rid of it here too, but the message would need to be altered a bit.

iwubcode (approved):
Wanted to make sure this didn't get lost.  Untested but looks like a huge improvement!

BhaaLseN (approved):

PR #10211, opened 2021-11-12
Retarget .vcxproj files to v143

Zopolis4:
Retarget the .vcxproj files to the v143 toolset. Complimentary to #10210, as currently the user needs to retarget the files each time while building, or install an older sdk (which defeats the entire point of #10210).


shuffle2:
If you are having a build problem, maybe describe it on the issue tracker. The current state (after #10210 was merged) should work fine and use v143 toolset for all projects. I tested on both my machine and the buildbot and it's working fine. This PR should be closed IMO.

mbc07:
Just updated my local fork and it also successfully built with Visual Studio 2022 and SDK 10.0.22000.194, no manual retargets were needed. Did you try cleaning the solution after updating to VS 2022?

Zopolis4:
I made a new clone using VS 2022, but I think I know what the issue was, although I can't check atm.

PR #10213, opened 2021-11-13
Android: Bump NDK and CMake versions

JosJuice:
This should make C++20 and `std::filesystem` work. (Not that we really can use `std::filesystem` much on Android since it doesn't work with scoped storage...)


JosJuice:
Okay, I'll have to do some plumbing on the Android buildbot for this one.

JosJuice:
@dolphin-emu-bot rebuild

lioncash (approved):

PR #10214, opened 2021-11-14
MoltenVK: Don't run fetchDependencies unnecessarily

OatmealDome:
Currently, MoltenVK's ``fetchDependencies`` script is ran on every build. The script clones various git repositories for dependencies and rebuilds each one of them. This causes rebuild times to take significantly longer, which makes development more annoying. To fix this, don't run the script after the initial build if MoltenVK's version hasn't been updated. 

Decreases rebuild times without any changes to source code from ~250s to ~10s on an M1 Mac Mini.

### DIFF
--- a/Externals/MoltenVK/CMakeLists.txt
+++ b/Externals/MoltenVK/CMakeLists.txt
@@ -1,8 +1,10 @@
 include(ExternalProject)
 
+set(MOLTENVK_VERSION "v1.1.5")
+
 ExternalProject_Add(MoltenVK
   GIT_REPOSITORY https://github.com/KhronosGroup/MoltenVK.git
-  GIT_TAG v1.1.5
+  GIT_TAG ${MOLTENVK_VERSION}
 
   CONFIGURE_COMMAND <SOURCE_DIR>/fetchDependencies --macos
 
--- a/Externals/MoltenVK/CMakeLists.txt
+++ b/Externals/MoltenVK/CMakeLists.txt
@@ -6,7 +6,7 @@ ExternalProject_Add(MoltenVK
   GIT_REPOSITORY https://github.com/KhronosGroup/MoltenVK.git
   GIT_TAG ${MOLTENVK_VERSION}
 
-  CONFIGURE_COMMAND <SOURCE_DIR>/fetchDependencies --macos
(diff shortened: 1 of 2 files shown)


Externals/MoltenVK/CMakeLists.txt
```diff
@@ -1,10 +1,12 @@
...
 
+set(MOLTENVK_VERSION "v1.1.5")
+
 ExternalProject_Add(MoltenVK
   GIT_REPOSITORY https://github.com/KhronosGroup/MoltenVK.git
-  GIT_TAG v1.1.5
+  GIT_TAG ${MOLTENVK_VERSION}
 
-  CONFIGURE_COMMAND <SOURCE_DIR>/fetchDependencies --macos
+  CONFIGURE_COMMAND ${CMAKE_CURRENT_LIST_DIR}/configure.sh <DOWNLOAD_DIR> <SOURCE_DIR> ${MOLTENVK_VERSION}
```
leoetlino, line 9: Does using DOWNLOAD_DIR actually work? The [docs](https://cmake.org/cmake/help/latest/module/ExternalProject.html#external-project-definition) indicate that

> This directory is only used by the URL download method, all other download methods use SOURCE_DIR directly instead.

OatmealDome: It does on CMake version 3.21.3, installed from brew.

(I actually wanted to use ``STAMP_DIR`` as it seemed more sane, but for some reason it never replaced ``STAMP_DIR`` in the command with the actual stamp directory path.)

leoetlino: Does using SOURCE_DIR work here? I'm a bit wary of using DOWNLOAD_URL as that's not supposed to work (and the fact it does work with anything other than the URL download method is an implementation detail / undocumented).

OatmealDome: #10215 does a ``git reset`` on this directory to ensure patch files apply correctly, so the version file would be lost between builds.

What about ``LOG_DIR``? It's automatically set to ``STAMP_DIR`` and actually works unlike using ``STAMP_DIR`` directly.

leoetlino: Sorry for the delay -- yeah, using LOG_DIR sounds fine.

OatmealDome: OK, changed it to ``LOG_DIR``.

PR #10218, opened 2021-11-14
[WIP] Add timing hack to fix Splinter Cell: Chaos Theory and Speed Challenge.

JMC47:
Dolphin's GPU timings are awful, and the GPU is way too fast especially in low polygon scenes.  In order to rectify this, I readded the old timings we used before 4.0-5369, but only use it when there's very few vertices being rendered.

This needs heavy testing in order to make sure there aren't performance regressions or any other crashes.  This is known to fix most Action Replay titles, both Advance Game Ports hanging, and Ubisoft's Speed Challenge.  This is very likely to either break/fix fifo sensitive titles that have either regressed or been fixed by various things.

Code was initially an if/else, but JosJuice pointed out to me there was a much cleaner way to do it, resulting in this.

image

### DIFF
--- a/Source/Core/VideoCommon/OpcodeDecoding.cpp
+++ b/Source/Core/VideoCommon/OpcodeDecoding.cpp
@@ -250,7 +250,9 @@ u8* Run(DataReader src, u32* cycles, bool in_display_list)
         src.Skip(bytes);
 
         // 4 GPU ticks per vertex, 3 CPU ticks per GPU tick
-        total_cycles += num_vertices * 4 * 3 + 6;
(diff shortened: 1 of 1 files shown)


JMC47:
I've been testing these FIFOCI differences, and they are just that - fifoci differences.  The actual games are all fine so far.  I haven't tested every single one.

I will rerecord fifologs as needed though.

phire:
FifoCI shouldn't be effected by timing differences at all. 

I think there is a bug there

JMC47:
Something else fixed the Datel Products.

PR #10219, opened 2021-11-15
msbuild: use /external:anglebrackets

shuffle2:
Revert usage of ExternalIncludePath, as that var is specifically
for external includes which do not get scanned for changes.

Instead, use /external:anglebrackets , which causes anything `#include`d with `<>` to be considered "external" (for adjusting warning level only - not affecting out-of-date checking).


JosJuice (approved):
Sounds reasonable.

lioncash (approved):

PR #10220, opened 2021-11-15
RenderBase: Put Opacity on FPS background.

Gamer64ytb:
Sometimes when playing a dialogue can not be seen in full due to the shadow of the FPS, therefore by putting a little opacity it could already be seen perfectly.


Pokechu22:
This ended up being implemented by #11326.

PR #10222, opened 2021-11-16
Fix copy filter clamping

phire:
This fixes horizontal lines in the bloom effect of Spyro: A Hero's Tail which is a regression caused by PR #10204.

<details><summary>Screenshot of regression</summary>

final

</details>

<details><summary>Fixed</summary>

final-fixed

</details>

Spyro uses an 640x80 pixel sub-buffer within the EFB to calculate
it's bloom effects, which it places below the main 640x448 buffer.

<details><summary>EFB layout</summary>

Note: Colors are wrong because the main color buffer uses RGBA6, while the bloom is calculated in RGB8

full-efb

</details>

This allows it to do bloom without backing up part of the EFB to
main memory, as most games do.

But, since some of the sub-buffers used in the bloom effect are taller
than 80 pixels, they need to be sliced up into smaller sub, sub buffers
which get combined later when copied to main memory.

At one point, a 320x224 buffer is broken up into 320x80, 320x64 and
320x80 slices. These are copied out with the copy filter set to a
vertical blur.

Because there was an off-by-one errror in the clamping coordinates,
the bottom line of the color buffer would be blurred into
the top of each slice.

<details><summary>Final combined EFB copy</summary>

merged-copy

</details>
<details><summary>Fixed version</summary>

merged-copy-fixed

</details>

Before #10204 the copy filter wasn't enabled for efb copies, and most
other games don't do this type of slicing.

FIFO CI shows that a few other games are effected, it's always just a minor difference to the top line where there was previously a slight hint of garbage.

### DIFF
--- a/Source/Core/VideoCommon/TextureCacheBase.cpp
+++ b/Source/Core/VideoCommon/TextureCacheBase.cpp
@@ -2685,7 +2685,8 @@ void TextureCacheBase::CopyEFBToCacheEntry(TCacheEntry* entry, bool is_depth_cop
   uniforms.filter_coefficients[2] = filter_coefficients.lower;
   uniforms.gamma_rcp = 1.0f / gamma;
   uniforms.clamp_top = clamp_top ? framebuffer_rect.top * rcp_efb_height : 0.0f;
-  uniforms.clamp_bottom = clamp_bottom ? framebuffer_rect.bottom * rcp_efb_height : 1.0f;
+  int bottom_coord = framebuffer_rect.bottom - 1;
+  uniforms.clamp_bottom = clamp_bottom ? bottom_coord * rcp_efb_height : 1.0f;
   uniforms.pixel_height = g_ActiveConfig.bCopyEFBScaled ? rcp_efb_height : 1.0f / EFB_HEIGHT;
   uniforms.padding = 0;
   g_vertex_manager->UploadUtilityUniforms(&uniforms, sizeof(uniforms));
@@ -2750,7 +2751,8 @@ void TextureCacheBase::CopyEFB(AbstractStagingTexture* dst, const EFBCopyParams&
   encoder_params.y_scale = y_scale;
   encoder_params.gamma_rcp = 1.0f / gamma;
   encoder_params.clamp_top = clamp_top ? framebuffer_rect.top * rcp_efb_height : 0.0f;
-  encoder_params.clamp_bottom = clamp_bottom ? framebuffer_rect.bottom * rcp_efb_height : 1.0f;
+  int bottom_coord = framebuffer_rect.bottom - 1;
+  encoder_params.clamp_bottom = clamp_bottom ? bottom_coord * rcp_efb_height : 1.0f;
   encoder_params.filter_coefficients[0] = filter_coefficients.upper;
   encoder_params.filter_coefficients[1] = filter_coefficients.middle;
   encoder_params.filter_coefficients[2] = filter_coefficients.lower;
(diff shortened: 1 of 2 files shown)


Source/Core/VideoCommon/TextureCacheBase.cpp
```diff
@@ -2685,7 +2685,8 @@ void TextureCacheBase::CopyEFBToCacheEntry(TCacheEntry* entry, bool is_depth_cop
   uniforms.filter_coefficients[2] = filter_coefficients.lower;
   uniforms.gamma_rcp = 1.0f / gamma;
   uniforms.clamp_top = clamp_top ? framebuffer_rect.top * rcp_efb_height : 0.0f;
-  uniforms.clamp_bottom = clamp_bottom ? framebuffer_rect.bottom * rcp_efb_height : 1.0f;
+  int bottom_coord = framebuffer_rect.bottom - 1;
+  uniforms.clamp_bottom = clamp_bottom ? bottom_coord * rcp_efb_height : 1.0f;
```
Pokechu22, line 2689: This is the corresponding code in the software renderer (which only exists in `EncodeXFB`; the equivalent for EFBs doesn't handle clamping or the copy filter currently):

https://github.com/dolphin-emu/dolphin/blob/0b81640dd1ce436519368556a69eb7fc650a60fa/Source/Core/VideoBackends/Software/EfbInterface.cpp#L587-L612

Is it also affected?  Given that `y_prev` can be 0 and `y_next` can be `EFB_HEIGHT` or `source_rect.bottom` I suspect it is... but I'm not 100% certain.  Since it's not implemented for EFBs at all currently this isn't too pressing.

I also need to question the `1.0f` in this code, though; shouldn't it be `(EFB_HEIGHT - 1) * rcp_efb_height`?  Or does that not matter because the texture is clamped elsewhere as well?

phire: Looks like the same bug. The for loop iterates from `top` to `bottom - 1`, yet it clamps at `bottom`.

As for the 1.0f... I'm not sure, but I suspect our whole coordinate system might be off-by-one here.

phire: I've fixed videosoftware.

I'll have to look more deeply into the UV coordinates later.

Pokechu22: The fifoci dffs for software are most visible in [mii-channel](https://fifo.ci/compare/6969196-6968476/) and [nhl-slap](https://fifo.ci/compare/6969256-6968536/); I took a look at mii-channel and it has clamping enabled, so the garbage data going away is correct (it's not an out-of-bounds read that the comment warns about).

Pokechu22: As I previously guessed, the texture is clamped elsewhere: either `GetLinearSamplerState` or `GetPointSamplerState` is used:

https://github.com/dolphin-emu/dolphin/blob/0b81640dd1ce436519368556a69eb7fc650a60fa/Source/Core/VideoCommon/TextureCacheBase.cpp#L2698-L2699

and those clamp u and v:

https://github.com/dolphin-emu/dolphin/blob/0b81640dd1ce436519368556a69eb7fc650a60fa/Source/Core/VideoCommon/RenderState.cpp#L357-L358
https://github.com/dolphin-emu/dolphin/blob/0b81640dd1ce436519368556a69eb7fc650a60fa/Source/Core/VideoCommon/RenderState.cpp#L372-L373

So, it won't go out of bounds, though whether this results in things being off slightly I'm less sure.  With linear filtering I vaguely think it might be correct to use 1.0f but I'm iffy on that.

phire: Yeah, the texture is clamped.

If the coordinate space is off-by-one then it should still produce the correct result for point-sampled copies, but will produce a slightly wrong result for linear sampling.

Pokechu22 (approved):
There's still a chance that there's an off-by-one when linear filtering is enabled, but this is definitely more accurate in most cases.

PR #10223, opened 2021-11-16
GameINI: Fix comments causing errors in GALE01r0.ini

JosJuice:
Same problem as PR #10201... Except instead of just making the comments use #, I'm actually removing some of the codes entirely because they are of questionable value.

Should fix https://bugs.dolphin-emu.org/issues/12737.

### DIFF
--- a/Data/Sys/GameSettings/GALE01r0.ini
+++ b/Data/Sys/GameSettings/GALE01r0.ini
@@ -127,59 +127,7 @@ $P1 - one-half size
 $P1 - one-tenth size
 0447E730 3DCCCCDF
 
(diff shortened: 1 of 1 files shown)


lioncash (commented):
unreal @ those codes

lioncash (approved):
Meant to approve with the previous comment

Avasam:
Curious, I decided to look into it. Despite the questionable name, this is actually a feature of the Master Debug Menu (at least for peach and Zelda).
https://tcrf.net/Super_Smash_Bros._Melee/Master_Debug_Menu#Camera_Controls
image
image

PR #10235, opened 2021-11-20
Netplay: Fix possible Wii save restore race condition between Netplay and CPU threads on game shutdown by making the Wii Save Sync data part of the BootParameters.

AdmiralCurtiss:
Scenario: Two people play Netplay with Wii save syncing enabled. This works in Dolphin by making a temporary Wii NAND, copying the host's save file into there, then running on that temp NAND for the game session. Then the host disconnects from the client so that this piece of code is reached:

https://github.com/dolphin-emu/dolphin/blob/6ca2897d200f898607cb1fe4370ed20b7f6ddd86/Source/Core/Core/NetPlayClient.cpp#L1543-L1549

and then in `StopGame()`:

https://github.com/dolphin-emu/dolphin/blob/6ca2897d200f898607cb1fe4370ed20b7f6ddd86/Source/Core/Core/NetPlayClient.cpp#L2251-L2254

`m_dialog->StopGame();` can return before the shutdown is actually complete -- I'm not entirely sure on the details here, but I believe what happens is that this `emit Stop()` here:

https://github.com/dolphin-emu/dolphin/blob/6ca2897d200f898607cb1fe4370ed20b7f6ddd86/Source/Core/DolphinQt/NetPlay/NetPlayDialog.cpp#L691-L698

Is handled asynchronously by the GUI thread. Which then leads the Netplay thread to call `ClearWiiSyncData();` while the CPU thread is still in the process of shutting down, which:

https://github.com/dolphin-emu/dolphin/blob/6ca2897d200f898607cb1fe4370ed20b7f6ddd86/Source/Core/Core/NetPlayClient.cpp#L2557-L2566

The important line here is the `s_wii_sync_fs.reset();`, which the CPU thread shutdown routine uses to detect whether it should copy back data to the main NAND or not -- if we're the host it is null, if we're the client it is not null, or that's the theory anyway.

https://github.com/dolphin-emu/dolphin/blob/6ca2897d200f898607cb1fe4370ed20b7f6ddd86/Source/Core/Core/WiiRoot.cpp#L339-L345

Anyway, long story short, I fixed this by just putting this data into the BootParameters so the CPU thread lifetime decides when it gets cleared, rather than some global in the Netplay files. This should circumvent this entire mess and have it work regardless of timing.

(Also note: The `"Wii" GC_MEMCARD_NETPLAY` directory being deleted is *not*, as you might assume, the temp NAND -- rather it is the staging directory used for collecting saves when they're sent to the client on netplay initialization. So that part doesn't break anything.)

### DIFF
--- a/Source/Android/jni/MainAndroid.cpp
+++ b/Source/Android/jni/MainAndroid.cpp
@@ -549,8 +549,7 @@ static float GetRenderSurfaceScale(JNIEnv* env)
 }
 
 static void Run(JNIEnv* env, const std::vector<std::string>& paths, bool riivolution,
-                const std::optional<std::string>& savestate_path = {},
-                bool delete_savestate = false)
+                BootSessionData boot_session_data = BootSessionData())
 {
   ASSERT(!paths.empty());
   __android_log_print(ANDROID_LOG_INFO, DOLPHIN_TAG, "Running : %s", paths[0].c_str());
@@ -561,9 +560,8 @@ static void Run(JNIEnv* env, const std::vector<std::string>& paths, bool riivolu
 
   s_have_wm_user_stop = false;
 
-  std::unique_ptr<BootParameters> boot = BootParameters::GenerateFromFile(paths, savestate_path);
-  if (boot)
-    boot->delete_savestate = delete_savestate;
+  std::unique_ptr<BootParameters> boot =
+      BootParameters::GenerateFromFile(paths, std::move(boot_session_data));
 
   if (riivolution && std::holds_alternative<BootParameters::Disc>(boot->parameters))
   {
@@ -638,8 +636,10 @@ Java_org_dolphinemu_dolphinemu_NativeLibrary_Run___3Ljava_lang_String_2ZLjava_la
     JNIEnv* env, jclass, jobjectArray jPaths, jboolean jRiivolution, jstring jSavestate,
     jboolean jDeleteSavestate)
 {
-  Run(env, JStringArrayToVector(env, jPaths), jRiivolution, GetJString(env, jSavestate),
-      jDeleteSavestate);
+  DeleteSavestateAfterBoot delete_state =
+      jDeleteSavestate ? DeleteSavestateAfterBoot::Yes : DeleteSavestateAfterBoot::No;
+  Run(env, JStringArrayToVector(env, jPaths), jRiivolution,
+      BootSessionData(GetJString(env, jSavestate), delete_state));
 }
 
 JNIEXPORT void JNICALL Java_org_dolphinemu_dolphinemu_NativeLibrary_ChangeDisc(JNIEnv* env, jclass,
--- a/Source/Core/Core/Boot/Boot.cpp
+++ b/Source/Core/Core/Boot/Boot.cpp
@@ -112,22 +112,53 @@ static std::vector<std::string> ReadM3UFile(const std::string& m3u_path,
   return result;
 }
 
-BootParameters::BootParameters(Parameters&& parameters_,
-                               const std::optional<std::string>& savestate_path_)
-    : parameters(std::move(parameters_)), savestate_path(savestate_path_)
+BootSessionData::BootSessionData()
 {
 }
 
-std::unique_ptr<BootParameters>
-BootParameters::GenerateFromFile(std::string boot_path,
-                                 const std::optional<std::string>& savestate_path)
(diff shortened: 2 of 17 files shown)


JMC47:
This is probably urgent to merge once it's ready, due to the nature of how this works.

leoetlino (commented):
> Reviewed 9 of 9 files at r1, 5 of 5 files at r2, 5 of 5 files at r3, 6 of 6 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/dolphin-emu/dolphin/10235)* status: all files reviewed, 4 unresolved discussions (waiting on @AdmiralCurtiss)

___
*[Source/Core/Core/Core.cpp, line 483 at r3](https://reviewable.io/reviews/dolphin-emu/dolphin/10235#-Mp33p8H1YwoaY9ad4pA:-Mp33p8H1YwoaY9ad4pB:b-5kjyyh) ([raw file](https://github.com/dolphin-emu/dolphin/blob/0ac01279ba01d57fc7a9b6923164ad9ee72633e1/Source/Core/Core/Core.cpp#L483)):*
> ```cpp
> 
>   BootSessionData boot_session_data = std::move(boot->boot_session_data);
>   const std::optional<std::string> savestate_path = boot_session_data.GetSavestatePath();
> ```

Making a copy of the savestate path seems unnecessary now, so this can be turned into a reference
___
*[Source/Core/Core/Core.cpp, line 621 at r3](https://reviewable.io/reviews/dolphin-emu/dolphin/10235#-Mp343Fx7xIssMEL4NLK:-Mp343Fx7xIssMEL4NLL:b-3wgywq) ([raw file](https://github.com/dolphin-emu/dolphin/blob/0ac01279ba01d57fc7a9b6923164ad9ee72633e1/Source/Core/Core/Core.cpp#L621)):*
> ```cpp
>   // with the correct title context since save copying requires title directories to exist.
>   Common::ScopeGuard wiifs_guard{
>       [&boot_session_data]() { Core::CleanUpWiiFileSystemContents(boot_session_data); }};
> ```

no need for `()`
___
*[Source/Core/Core/WiiRoot.cpp, line 342 at r3](https://reviewable.io/reviews/dolphin-emu/dolphin/10235#-Mp35LBc35xXeZ4gt--6:-Mp35LBc35xXeZ4gt--7:b14dsid) ([raw file](https://github.com/dolphin-emu/dolphin/blob/0ac01279ba01d57fc7a9b6923164ad9ee72633e1/Source/Core/Core/WiiRoot.cpp#L342)):*
> (Also note: The "Wii" GC_MEMCARD_NETPLAY directory being deleted is not, as you might assume, the temp NAND -- rather it is the staging directory used for collecting saves when they're sent to the client on netplay initialization. So that part doesn't break anything.)

I'm a bit confused -- is this clearing the netplay temp NAND or the save staging area?
___
*[Source/Core/DolphinQt/NetPlay/NetPlayDialog.h, line 39 at r2](https://reviewable.io/reviews/dolphin-emu/dolphin/10235#-Mp33I-g7A9WVMspkiC7:-Mp33I-g7A9WVMspkiC8:bbi29hg) ([raw file](https://github.com/dolphin-emu/dolphin/blob/a8a5f6e763e2f4a4770bcf7ef68a063b39040740/Source/Core/DolphinQt/NetPlay/NetPlayDialog.h#L39)):*
> ```objc
> public:
>   explicit NetPlayDialog(const GameListModel& game_list_model,
>                          std::function<void(const std::string& path,
> ```

`using StartGameCallback = ...` to make this more readable (and avoid having to write the entire thing a second time for m_start_game_callback)?

AdmiralCurtiss:
Ah oops, yes, staging area, not temp NAND, I was confused on this myself at first and it looks like the function name is a leftover from that. Probably not the right place for this either now that I think about it, that should be somewhere in the Netplay code...

AdmiralCurtiss:
Alright, redid this a bit so the staging directory is deleted in a callback set from the Netplay code. I think this is a lot cleaner!

leoetlino (approved):
> Reviewed 12 of 12 files at r5, 5 of 5 files at r6, 5 of 5 files at r7, 6 of 6 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/dolphin-emu/dolphin/10235)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @AdmiralCurtiss)

PR #10236, opened 2021-11-20
Cleanup and Update N64 Virtual Console INIs

JMC47:
This Pull Request updates N64 INIs based on what I know about various behaviors since the last time we updated it.  This should result in more sane settings for Mario Tennis and Super Smash Bros. while still preventing the out of memory crash due to the tremendous amounts of EFB copies.

### DIFF
--- a/Data/Sys/GameSettings/NAL.ini
+++ b/Data/Sys/GameSettings/NAL.ini
@@ -19,5 +19,9 @@ StereoConvergence = 5000
 # This game creates a large number of EFB copies at different addresses, resulting
 # in a large texture cache which takes considerable time to save.
 SaveTextureCacheToState = False
-# This game crashes very soon after boot if above 1x IR
-InternalResolution = 1
+
+[Video_Hacks]
+# This game crashes very soon after boot if EFB copies are scaled at higher resolutions.
+# This is due to Dolphin running out of memory.  Disabling Scaled EFB copies allows playing
+# at higher resolutions without the game crashing for now.
+EFBScaledCopy = False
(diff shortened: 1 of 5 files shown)


Data/Sys/GameSettings/NAL.ini
```diff
@@ -19,5 +19,9 @@ StereoConvergence = 5000
...
 # in a large texture cache which takes considerable time to save.
 SaveTextureCacheToState = False
-# This game crashes very soon after boot if above 1x IR
-InternalResolution = 1
+
+[Video_Hacks]
+# This game crashes very soon after boot if EFB copies are scaled at higher resolutions.
+# This is due to Dolphin running out of memory.  Disabling Scaled EFB copies allows playing
+# at higher resolutions without the game crashing for now.
+EFBScaledCopy = True
```
iwubcode, line 27: Do you mean `False` here?  The comment and the value don't match I think

Data/Sys/GameSettings/NAT.ini
```diff
@@ -16,5 +16,9 @@
...
 # in a large texture cache which takes considerable time to save.
 SaveTextureCacheToState = False
-# This game crashes very soon after boot if above 1x IR
-InternalResolution = 1
+
+[Video_Hacks]
+# This game crashes very soon after boot if EFB copies are scaled at higher resolutions.
+# This is due to Dolphin running out of memory.  Disabling Scaled EFB copies allows playing
+# at higher resolutions without the game crashing for now.
+EFBScaledCopy = True
```
iwubcode, line 24: Ditto my other comment

Data/Sys/GameSettings/NAL.ini
JMC47: I didn't realize I never pushed this - should be fixed now.

Data/Sys/GameSettings/NAT.ini
JMC47: Fixed

JosJuice (approved):

PR #10238, opened 2021-11-21
Gameini: Enable MMU for some games

PatrickFerry:
This is to enable MMU for a couple games.

"007: From Russia with Love" requires MMU to display in-game menus. Otherwise they are black. Also I have enabled medium texture cache to fix the jittery intro video.

"We Love Golf!" crashes without full MMU. This was apparently fixed in 5.0-540 (Dynamic bat) which is true only if MMU is enabled. Since it's not enabled by default the game should overwrite the MMU setting to work.

### DIFF
--- a/Data/Sys/GameSettings/GLZ.ini
+++ b/Data/Sys/GameSettings/GLZ.ini
@@ -1,0 +1,13 @@
+# GLZP69, GLZE69, GLZD69, GLZF69 - 007: From Russia with Love
+
+[Core]
+# Displays in-game menus
+MMU = True
+
+[Video_Settings]
+SafeTextureCacheColorSamples = 512
+
+[Video_Hacks]
+# Fixes Bloom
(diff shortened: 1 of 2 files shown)


JMC47:
Confirmed and Confirmed.  My real question is WTF is going on with 007: From Russia with Love.  This is now the third game that does this shit.

JMC47:
Fixes the games, so merging once I re-read the INIs.

JMC47:
Why are the bloom fixes commented out?

PatrickFerry:
Because with both the bloom fixes in place I can drop to 50% on an 8th gen i7 at time, like the first indoor location. Whereas it's fullspeed for me without the bloom fixes in the same location.

I would just think it's the more sensible default given bloom is a minor detail.

ajay9634:
Ultimate spiderman is crashing but running at 30fps.

PR #10239, opened 2021-11-21
Android: Add MMU setting to GUI

JosJuice:
Now that we have fast MMU emulation on AArch64, there's no reason to keep excluding this from the GUI like we've been doing.


lioncash (approved):

brujo5:
Hi few users from telegram group say ultimate spiderman now run at 30fps all time but crashes in cutscene sadly.

PR #10240, opened 2021-11-21
GBA: Add import and export save options to context menu.

AdmiralCurtiss:
Since users seem to be very confused about where save files should go for the emulator to pick them up.

### DIFF
--- a/Source/Core/DolphinQt/GBAWidget.cpp
+++ b/Source/Core/DolphinQt/GBAWidget.cpp
@@ -229,6 +229,36 @@ void GBAWidget::DoState(bool export_state)
       false);
 }
 
+void GBAWidget::ImportExportSave(bool export_save)
+{
+  if (!CanControlCore() && !export_save)
+    return;
+  if (!m_core_info.has_rom)
+    return;
+
+  QString save_path = QDir::toNativeSeparators(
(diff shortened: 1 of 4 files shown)


Source/Core/DolphinQt/GBAWidget.cpp
```diff
@@ -229,6 +229,36 @@ void GBAWidget::DoState(bool export_state)
       false);
 }
 
+void GBAWidget::ImportExportSave(bool export_save)
```
iwubcode, line 232: I don't personally have a strong opinion about this but I know some people prefer an enumeration over a boolean

iwubcode (approved):
Untested, LGTM

Source/Core/DolphinQt/GBAWidget.cpp
AdmiralCurtiss: I just copied the existing `DoState()` more or less, but yeah, probably.

JMC47:
This is probably an improvement.  Does this handle that each GBA uses a different save?

AdmiralCurtiss:
It does handle that correctly, yes.

JMC47:
Tested with all 4 GBAs

PR #10241, opened 2021-11-22
Ensure user paths are stored in a consistent manner.

AdmiralCurtiss:
That is:
- All path separators are forward slashes.
- Directory paths end with a separator.
- File paths do not end with a separator.

### DIFF
--- a/Source/Core/DiscIO/NANDImporter.cpp
+++ b/Source/Core/DiscIO/NANDImporter.cpp
@@ -34,10 +34,9 @@ void NANDImporter::ImportNANDBin(const std::string& path_to_bin,
   if (!ReadNANDBin(path_to_bin, get_otp_dump_path))
     return;
 
-  const std::string nand_root = File::GetUserPath(D_WIIROOT_IDX);
+  std::string nand_root = File::GetUserPath(D_WIIROOT_IDX);
+  nand_root.pop_back();  // remove trailing path separator
   m_nand_root_length = nand_root.length();
-  if (nand_root.back() == '/')
-    m_nand_root_length++;
 
   FindSuperblock();
   ProcessEntry(0, nand_root);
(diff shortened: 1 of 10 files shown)


MayImilae:
What happens if the user enters a path incorrectly, say using backslash instead of forwardslash? What is the experience that the user will get when that occurs? Does it just fix it for them or will it alert them?

Also, isn't backslash \ used for paths on windows? I guess this isn't accounting for OS differences like that?

AdmiralCurtiss:
We've already been using `/` as the path separator on all platforms before this (see eg. [here](https://github.com/dolphin-emu/dolphin/blob/cc9a75a87bc6e6c28d05535d3dd8877979aab4d0/Source/Core/Common/CommonPaths.h#L7-L8) or [here](https://github.com/dolphin-emu/dolphin/blob/cc9a75a87bc6e6c28d05535d3dd8877979aab4d0/Source/Core/Common/FileUtil.cpp#L265) or [here](https://github.com/dolphin-emu/dolphin/blob/cc9a75a87bc6e6c28d05535d3dd8877979aab4d0/Source/Core/UICommon/UICommon.cpp#L277-L279) or anywhere we concatenate paths really; Windows does allow this!), this just makes it more explicit and something you can actually rely on.

If the user enters a backslash it will just get silently replaced with a forward slash internally.

Source/Core/DiscIO/NANDImporter.cpp
```diff
@@ -34,10 +34,9 @@ void NANDImporter::ImportNANDBin(const std::string& path_to_bin,
   if (!ReadNANDBin(path_to_bin, get_otp_dump_path))
     return;
 
-  const std::string nand_root = File::GetUserPath(D_WIIROOT_IDX);
+  std::string nand_root = File::GetUserPath(D_WIIROOT_IDX);
+  nand_root.pop_back();  // remove trailing path separator
```
Starsam80, line 38: IIRC it doesn't matter if this path has a trailing separator, the only piece of code that does (or did) care is to make debug logs a little nicer to look at. So I don't think this line is needed but feel free to do it anyways.

PS: Annoyed by my old code in either case, I made a quick refactor that completely removes the weird `_length` variable (and a few other things) at https://github.com/Starsam80/dolphin/commit/8911c3bb42ba8abaa6beffbe7924b0dd622cc77e which I might make a PR for if I ever get enough energy.

leoetlino (approved):

PR #10244, opened 2021-11-24
Delay singlecore gpu interrupts; Fixes Bomberman Jetters in single core mode.

phire:
Fixes Bomberman Jetters in single core mode.

When single core mode pauses the CPU to execute the GPU
FIFO it greedily executes the whole thing. Before this commit,
Finish and Token interrupts would happen instantly, not even
taking into account how long the current FIFO window has
taken to execute. The interrupts would be effectively backdated
to the start of this execution window.

This commit does two things: It pipes the current FIFO window
execution time though to the interrupt scheduling and it enforces
a minimum delay of 500 cycles before an interrupt will be fired.

### DIFF
--- a/Source/Core/VideoCommon/PixelEngine.cpp
+++ b/Source/Core/VideoCommon/PixelEngine.cpp
@@ -277,7 +277,7 @@ static void SetTokenFinish_OnMainThread(u64 userdata, s64 cyclesLate)
 // Raise the event handler above on the CPU thread.
 // s_token_finish_mutex must be locked.
 // THIS IS EXECUTED FROM VIDEO THREAD
-static void RaiseEvent()
+static void RaiseEvent(int cycles_into_future)
 {
   if (s_event_raised)
     return;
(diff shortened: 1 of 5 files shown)


Source/Core/VideoCommon/PixelEngine.cpp
```diff
@@ -277,22 +277,29 @@ static void SetTokenFinish_OnMainThread(u64 userdata, s64 cyclesLate)
...
   if (!SConfig::GetInstance().bCPUThread || Fifo::UseDeterministicGPUThread())
+  {
     from = CoreTiming::FromThread::CPU;
-  CoreTiming::ScheduleEvent(0, et_SetTokenFinishOnMainThread, 0, from);
+
+    // Hack: Dolphin's single-core gpu timings are way too fast. Enforce a minimum delay to give
+    //       games time to setup any interrupt state
+    cycles = std::max(500, cycles_into_future);
+  }
+  CoreTiming::ScheduleEvent(cycles, et_SetTokenFinishOnMainThread, 0, from);
```
phire, line 297: I'm noticing a pre-existing bug here, which I fear I'm making worse.

There is no synchronisation, so if there are many finish or tokens close to each other in the FIFO, they will stomp over each other. The cpu thread will only see a single interrupt, and the token will jump straight to the final value.

This can happen in dual-core mode, randomly if the gpu thread is fast, or the cpu thread is slow.  
Before this PR, this can happen in single core mode, but only within a single FIFO window (or if multiple gather-pipe bursts happen within the same JIT block)
After this PR, this can also happen when multiple gather-pipe bursts happen to within 500 cycles of each other.

I fear that this bug needs be fixed before merging this fix, otherwise we risk causing more regressions.

JMC47:
Fixes Resident Evil 3 and Bomberman Jetters in Single Core.

Fixes Advance Game Port and other Datel Loaders that use the same era loader.  Requires MMU to be off.

JMC47:
We're going to merge this to see if it breaks anything.

PR #10245, opened 2021-11-24
Fix saving RGBA images

Pokechu22:
Fixes [broken texture dumping](https://forums.dolphin-emu.org/Thread-texture-dumping).

`PNG_FORMAT_RGB` and `PNG_COLOR_TYPE_RGB` both evaluate to 2, but `PNG_FORMAT_RGBA` evaluates to 3 while `PNG_COLOR_TYPE_RGBA` evaluates to 6; the bit indicating a palette is 1 while the bit indicating alpha is 4.

This generates an error, but because the default libpng error handler logs to stderr, it wasn't seen.  I've also added custom error handlers that put it into Dolphin's logging system.  With that, the following message now showed up (prior to fixing the above issue):

```
30:33:346 Common\Image.cpp:140 E[FRAMEDUMP]: Failed to save 852 by 480 image to C:/Users/Pokechu22/Documents/Dolphin Emulator/Dump/Textures/OHBCHB/tex1_852x480_502591290a4f95cd_6.png at level 6: 0 warnings, 1 errors
30:33:346 Common\Image.cpp:142 E[FRAMEDUMP]: libpng error: Valid palette required for paletted images
```

### DIFF
--- a/Source/Core/Common/Image.cpp
+++ b/Source/Core/Common/Image.cpp
@@ -8,6 +8,7 @@
 
 #include <png.h>
 
+#include "Common/Assert.h"
 #include "Common/CommonTypes.h"
 #include "Common/IOFile.h"
 #include "Common/ImageC.h"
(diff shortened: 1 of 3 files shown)


iwubcode (approved):
Code LGTM.  Confirmed that it fixes the problem

PR #10246, opened 2021-11-24
DolphinNoGUI: Add CLI tool to verify disc images (RVZ / ISO / NKit)

ssdsnake:
**Preface:**
Still working on testing for Windows. This is my first PR, open to any feedback. Thank you !

**Use case:**
Adds ability to verify and output SHA1/MD5/CRC32 digests of RVZ / ISO / NKit using the same routines as the GUI verifier widget. The tool is intended to be simple with an easy to parse output. This can help speed up RVZ DAT verification workflows with external tooling.

Named "discimagetool" and given a mode switch to be extensible. 

**Example**
Default algorithm selection (SHA1)
```
$ ./dolphin-emu-discimagetool -m verify -i /tmp/test.rvz
SHA1: f2439bbe1ff64133050fbc00574be8478210a958
Problems Found: No
```
Default algorithm selection (SHA1) with simple output
```
$ ./dolphin-emu-discimagetool -m verify --verify_simple_out -i /tmp/test.rvz
f2439bbe1ff64133050fbc00574be8478210a958
```
Only CRC32
```
$ ./dolphin-emu-discimagetool -m verify -i /tmp/test.rvz --verify_algorithm crc32
CRC32: dcd8fa3d
Problems Found: No
```

All algorithms
```
$ ./dolphin-emu-discimagetool -m verify -i /tmp/test.rvz --verify_algorithm all
CRC32: dcd8fa3d
MD5: 03c6f44267aa4a5c42ab20698f88f477
SHA1: f2439bbe1ff64133050fbc00574be8478210a958
Problems Found: No
```

Usage
```
Usage: dolphin-emu-discimagetool [options]... [FILE]...

Options:
  --version             show program's version number and exit
  -h, --help            show this help message and exit
  -m MODE, --mode=MODE  Operating mode [verify]
  -i FILE, --input=FILE
                        Path to disc image FILE
  --verify_simple_out   Print the digest of the selected algorithm and exit.
                        Incompatible with 'all' option.
  --verify_algorithm=VERIFY_ALGORITHM
                        Digest algorithms to compute (default = sha1) [all|crc32|md5|sha1]
```


JosJuice:
@dolphin-emu-bot rebuild

phire:
I'm going to do some bikeshedding.

I'm not a big fan of how long `dolphin-emu-discimagetool -m verify` is to type.

Could we shorten the name to something like `dolphin-tool` or `dolphin-emu-tool`. And can we remove the requirement for the `-m` flag. Since there is no default mode, it's always require. We might as well just take the mode as the first argument, so `dolphin-tool verify`, the same way that a lot of modern cli tools do (for example, git)

ssdsnake:
> I'm going to do some bikeshedding.
> 
> I'm not a big fan of how long `dolphin-emu-discimagetool -m verify` is to type.
> 
> Could we shorten the name to something like `dolphin-tool` or `dolphin-emu-tool`. And can we remove the requirement for the `-m` flag. Since there is no default mode, it's always require. We might as well just take the mode as the first argument, so `dolphin-tool verify`, the same way that a lot of modern cli tools do (for example, git)

Hi, yeah I can find a shorter name. Re: the argument, this was the original plan :)
The challenge was with the used optparse library not have ability to parse the second argument as a subcommand concept
I can pull it out of argv, but was not sure how optparse would react to that. I can give it a shot.

phire:
Another reason to shorten the name is that now that such an interface exists, it's likely to get expanded with other tools, that might not be disc related. For example, NAND manipulation, WAD extraction, memory card manipulation. Maybe even other tools like texture format converters.

ssdsnake:
> Another reason to shorten the name is that now that such an interface exists, it's likely to get expanded with other tools, that might not be disc related. For example, NAND manipulation, WAD extraction, memory card manipulation. Maybe even other tools like texture format converters.

Is there WAD extraction code already existing? I can add it too

phire:
> I can pull it out of argv, but was not sure how optparse would react to that.

You can manually parse it, then adjust `argc` and `argv` to start the parser at the new position. But maybe it might be smarter to enhance or replace optparse with something that better handles this use case.

Tilka:
A while ago I wrote a hacky version of this for my own purposes and I also just called it dolphin-tool, so +1 for that name.

iwubcode:
Is there a reason we don't just have `dolphin-cmdline` or `dolphin-tool`, etc and merge `dolphin-nogui` into that as one executable?

`dolphin-tool` is fine but seems like it's more generic which makes me wonder why `dolphin-nogui` isn't also available.  `dolphin-nogui` isn't released as part of our release which I'd think we'd want to do for this tool if users are going to use it..

ssdsnake:
Will open a new PR addressing all of this feedback (and more)

PR #10247, opened 2021-11-25
Game INI updates

mbc07:
This PR provides a small batch of Game INI updates:

* **ABBA: You Can Dance and Michael Jackson: The Experience:** removed Anisotropic Filtering override as it doesn't cause visual glitches anymore.

* **Epic Mickey 2, Just Dance Kids, The Black Eyed Peas Experience, The Hip Hop Dance Experience:** disabled "Force Texture Filtering" to avoid FMV corruption.

* **Eternal Darkness:** replaced Deterministic GPU with Single Core mode and disabled "Store XFB Copies to Texture Only". Single Core fixes both the game hang after the Edgar Allan Poe quote as well as the aspect ratio distortion when entering/exiting the pause menu (Determinist GPU only prevented the hang), while "Store XFB Copies to Texture Only" fixes the purple transitions at the start/end of a chapter.

* **Just Dance 1, 2, 3, Best Of, Greatest Hits, Summer Party, Wii and Wii 2:** removed Safe Texture Cache and Anisotropic Filtering overrides as they don't cause visual glitches anymore. Enabled "Manual Texture Sampling" as it fixes vertical lines on dancers and loading screens (Just Dance 2, Wii and Summer Party are still affected but this option makes the lines less noticeable).

* **Just Dance Disney Party, Disney Party 2, Kids 2 and Kids 2014:** disabled "Force Texture Filtering" to avoid FMV corruption and enabled Safe Texture Cache to fix missing text and stuttery video playback.

### DIFF
--- a/Data/Sys/GameSettings/GED.ini
+++ b/Data/Sys/GameSettings/GED.ini
@@ -2,7 +2,7 @@
 
 [Core]
 # Values set here will override the main Dolphin settings.
-GPUDeterminismMode = fake-completion
+CPUThread = False
 
 [OnLoad]
 # Add memory patches to be loaded once on boot here.
@@ -13,6 +13,8 @@ GPUDeterminismMode = fake-completion
 [ActionReplay]
 # Add action replay cheats here.
 
+[Video_Hacks]
+XFBToTextureEnable = False
+
 [Video_Settings]
 SafeTextureCacheColorSamples = 512
-
--- a/Data/Sys/GameSettings/S2E.ini
+++ b/Data/Sys/GameSettings/S2E.ini
@@ -13,5 +13,4 @@
 # Add action replay cheats here.
 
 [Video_Enhancements]
-MaxAnisotropy = 0
 ForceFiltering = False
--- a/Data/Sys/GameSettings/S5D.ini
+++ b/Data/Sys/GameSettings/S5D.ini
@@ -1,4 +1,19 @@
 # S5DE41, S5DP41 - Just Dance Disney Party 2 
 
+[Core]
+# Values set here will override the main Dolphin settings.
+
+[OnLoad]
+# Add memory patches to be loaded once on boot here.
+
+[OnFrame]
+# Add memory patches to be applied every frame here.
+
+[ActionReplay]
+# Add action replay cheats here.
+
+[Video_Enhancements]
+ForceFiltering = False
+
 [Video_Settings]
 SafeTextureCacheColorSamples = 0
--- a/Data/Sys/GameSettings/SD2.ini
+++ b/Data/Sys/GameSettings/SD2.ini
@@ -12,9 +12,8 @@
 [ActionReplay]
 # Add action replay cheats here.
 
-[Video_Settings]
-SafeTextureCacheColorSamples = 512
-
 [Video_Enhancements]
-MaxAnisotropy = 0
 ForceFiltering = False
+
+[Video_Hacks]
+FastTextureSampling = False
--- a/Data/Sys/GameSettings/SD2J01.ini
+++ b/Data/Sys/GameSettings/SD2J01.ini
@@ -15,9 +15,8 @@
 [ActionReplay]
 # Add action replay cheats here.
 
-[Video_Settings]
-SafeTextureCacheColorSamples = 512
-
 [Video_Enhancements]
-MaxAnisotropy = 0
 ForceFiltering = False
+
+[Video_Hacks]
+FastTextureSampling = False
--- a/Data/Sys/GameSettings/SDN.ini
+++ b/Data/Sys/GameSettings/SDN.ini
@@ -13,5 +13,7 @@
 # Add action replay cheats here.
 
 [Video_Enhancements]
-MaxAnisotropy = 0
 ForceFiltering = False
+
+[Video_Hacks]
+FastTextureSampling = False
--- a/Data/Sys/GameSettings/SDZ.ini
+++ b/Data/Sys/GameSettings/SDZ.ini
@@ -1,4 +1,5 @@
-# SJDE41, SJDP41, SJDX41, SJDY41, SJDZ41 - Just Dance 3
+# SDZE41 - Just Dance Kids
+# SDZP41 - Dance Juniors [Europe]
 
 [Core]
 # Values set here will override the main Dolphin settings.
@@ -14,6 +15,3 @@
 
 [Video_Enhancements]
 ForceFiltering = False
-
-[Video_Hacks]
-FastTextureSampling = False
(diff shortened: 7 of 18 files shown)


mbc07:
Ping²

Pokechu22:
@dolphin-emu-bot rebuild

PatrickFerry:
Are we okay with disabling FastTextureSampling since it'll have a performance impact?

Are these games that improve with FastTextureSampling throughout the game or only on intro FMVs?

Pokechu22:
Just dance uses FMVs throughout the game for the dancers.  That said, only the first few Just Dance titles use Bink Video if I recall correctly, and those are the ones where manual texture sampling doesn't necessarily fix everything, so it might not necessarily be helpful.

mbc07:
There are 8 Just Dance games that are affected by FastTextureSampling. From those 8, 5 becomes flawless, the remaining 3 aren't completely fixed but the issue is greatly reduced, so I think it's worth enabling that setting by default...

Pokechu22:
I took a look through my notes, and yeah, that does seem to be consistent with what I found.  That said, I _did not_ find Just Dance Disney Party 2 or Just Dance Kids 2014 to use Bink Video (it's not on the box), so I wouldn't expect Force Texture Filtering to do anything for them.  Same deal with Epic Mickey 2 (and the original): I didn't see Bink Video on the box, so Force Texture Filtering shouldn't do anything.  I would also expect any game where Force Texture Filtering _does_ cause problems with FMVs to also be improved by Manual Texture Sampling, so it's odd that you only are enabling it for some.  (It's possible that e.g. Just Dance Kids 2014 needs safe texture cache only, and force texture filtering has no effect on it.)

Also, you're missing a very important game: The Smurfs: Dance Party apparently uses Bink Video too (it's even explicitly noted in [the list](http://www.radgametools.com/binkgames.htm)), and presumably The Smurfs: Party Pack is also affected since it bundles Dance Party.

mbc07:
If you're assuming that only FMVs from games that uses Bink video breaks with Force Texture Filtering, please don't, that's definitely not the case. I've tested every game referenced on this PR separately and only disabled Force Texture Filtering or enabled Manual Texture Sampling when required. I didn't touch the INIs from Smurfs Dance Party and the bundle because I couldn't test those games, but they likely requires Force Texture Filtering disabled as well.

FMVs from Epic Mickey 2 are completely broken when Force Texture Filtering is enabled, same deal with the first Epic Mickey, but it already has Force Texture Filtering disabled on its INI on master. I haven't digged through the filesystem of the game discs so I can't confirm if they really use Bink video or not, but Force Texture Filtering definitely affects the FMVs of both.

About the Just Dance franchise, the Wii games run on 3 different engines (4 if you consider Just Dance 2014 and up, but it's out of the scope of this PR as they're not affected neither by Force Texture Filtering nor Manual Texture Sampling):

* **Jade Engine:** Used only in the first Just Dance. Uses Bink video, is affected by both Force Texture Filtering and Manual Texture Sampling, the latter completely fixes the vertical lines on the dancers.

* **Land Ho! engine (unknown name):** Used in all Just Dance games developed or co-developed  by Land Ho! (The Hip Hop Dance Experience, all Kids games, all Party games except Summer Party). Uses Bink video, all are affected by Force Texture Filtering and some also requires Safe Texture Cache to fix missing text and/or stuttery video playback. They don't require Manual Texture Sampling because they already render correctly without it, even on NVIDIA GPUs.

* **LyN engine:** Used in all spin-offs where Land Ho! weren't involved and on the main series until Just Dance 4. Uses Bink and Mobiclip video (sometimes both, e.g. Just Dance 4, DLCs from Just Dance 2 and 3), all are affected by Force Texture Filtering but only some are affected by Manual Texture Sampling. Earlier spin-offs (based on Just Dance 2) aren't completely fixed by Manual Texture Sampling but the issue is greatly reduced, while later spin-offs (based on Just Dance 3) becomes flawless. Spin-offs where I didn't enable Manual Texture Sampling already renders flawlessly without it, so I left them out. Just Dance 4 is affected by Manual Texture Sampling as well but I left it out because the only Bink video used by the game that exhibits vertical lines is a very short loading animation that appears only once in the entire game (before the "press A to start" screen).

My test methodology was starting a fresh portable copy of Dolphin with the bundled Game INIs (Sys\GameSettings) removed, then enabling the settings touched by this PR one by one until the issues were gone. I've also tested Safe Texture Cache separately of Force Texture Filtering and only enabled both when required. Hope this clears up the raised points...

Pokechu22:
> If you're assuming that only FMVs from games that uses Bink video breaks with Force Texture Filtering, please don't, that's definitely not the case.

No, there are definitely other ones (EA titles, e.g. those using VP6 (though other codecs also have issues)), I know that.  But if Force Texture Filtering is causing issues, then that usually means it's doing something with indirect textures where precision matters and filtering/interpolating textures breaks it.

On the other hand, I remembered one other aspect to this: with Bink (or at least some revisions of Bink), the issue only happens if the video's resolution does not match the resolution the video is being played at.  For instance, Shrek the Third's intro video is 640 by 360, but is played back at 640 by 480, which causes lines in some places on Nvidia only (no lines appear on Intel); however, the logo videos (e.g. activision's logo) are at the correct resolution and thus no lines appear on them.  In all cases, force texture filtering causes issues.  So that's probably the cause of force texture filtering but not manual texture sampling being needed (and that same thing might also apply to mobiclip?)

<details><summary>Images of shrek, for context</summary>

Nvidia:
image
Enlarged:
image
Intel:
image

</details>

If you've actually gone through and tested it, that's probably better than the approach I would have taken of just applying the safest settings to all games with the bink logo on the box or that I have another reason to think uses bink (since there are several hundred of them, though note that only games released after about 2006 are affected since before then color conversion was done on the CPU instead of the GPU).

mbc07:
Is there anything else I can do to move this PR forward? Kinda puzzled on how long a few Game INI changes are taking to get approved...

AdmiralCurtiss:
Probably because it's 15ish different games and no one actually has all of them* to check if the changes are appropriate. And GitHub's UI is not great for signalling partial approvals.

(*okay maybe JMC does but you know what I'm saying, right?)

JMC47:
I think all of the changes make sense, but long term I really want to change how gamesettings are applied.  A lot of these decision are 100% correct, but I'm not sure if they're right for Android, or users that are on AMD that don't need manual sampling, etc.

I don't see a better solution right now.

JMC47 (approved):

PatrickFerry:
This is also somewhat the case with gamesettings applied in the past too, where there is often a tradeoff being made between accuracy and performance.

Without the settings being able to be conditionally set. I'd also say this is positive change on balance.

JMC47:
I'm writing up a proposal to potentially address this long term, but for now I say we just get the games rendering as correctly as possible.

PR #10248, opened 2021-11-27
Fix default input config default device not being loaded/found

Filoppi:
Fixes bug: https://bugs.dolphin-emu.org/issues/12744

Before https://github.com/dolphin-emu/dolphin/commit/e1e3db13baabefa89991388d37db0bb260c4f535 the ControllerInterface m_devices_mutex was "wrongfully" locked for the whole Initialize() call, which included the first device population refresh, this has the unwanted (accidental) consequence of often preventing the different pads (GC Pad, Wii Contollers, ...) input configs from loading until that mutex was released (the input config defaults loading was blocked in EmulatedController::LoadDefaults()), which meant that the devices population would often have the time to finish adding its first device, which would then be selected as default device (by design, the first device added to the CI is the default default device, usually the "Keyboard and Mouse" device).

After the commit mentioned above removed the unnecessary m_devices_mutex calls, the default default device would fail to load (be found) causing the default input mappings, which are specifically written for the default default device on every platform, to not be bound to any physical device input, breaking input on new dolphin installations (until a user tried to customize the default device manually).

Default devices are now always added synchronously to avoid the problem, and so they should in the future (I added comments and warnings to help with that).
The only one which was async was the Windows "Keyboard and Mouse" device, and I took care of that. I also changed devices sorting order to be clearer (and avoid other devices accidental ending up as default) (the sorting order only affect the Qt devices list order and nothing else).

### DIFF
--- a/Source/Core/DolphinQt/MainWindow.cpp
+++ b/Source/Core/DolphinQt/MainWindow.cpp
@@ -319,6 +319,13 @@ void MainWindow::InitControllers()
     return;
 
   g_controller_interface.Initialize(GetWindowSystemInfo(windowHandle()));
+  if (!g_controller_interface.HasDefaultDevice())
+  {
+    // Note that the CI default device could be still temporarily removed at any time
+    WARN_LOG(CONTROLLERINTERFACE,
+             "No default device has been added in time. EmulatedController(s) defaulting adds"
+             " input mappings made for a specific default device depending on the platform");
+  }
   Pad::Initialize();
   Pad::InitializeGBA();
   Keyboard::Initialize();
--- a/Source/Core/InputCommon/ControllerInterface/ControllerInterface.cpp
+++ b/Source/Core/InputCommon/ControllerInterface/ControllerInterface.cpp
@@ -159,6 +159,10 @@ void ControllerInterface::RefreshDevices(RefreshReason reason)
   // with their own PlatformPopulateDevices().
   // This means that devices might end up in different order, unless we override their priority.
   // It also means they might appear as "disconnected" in the Qt UI for a tiny bit of time.
+  // This helps the emulation and host thread to not stall when repopulating devices for any reason.
+  // Every platform that adds a device that is meant to be used as default device should try to not
+  // do it async, to not risk the emulated controllers default config loading not finding a default
+  // device.
 
 #ifdef CIFACE_USE_WIN32
   ciface::Win32::PopulateDevices(m_wsi.render_window);
--- a/Source/Core/InputCommon/ControllerInterface/CoreDevice.cpp
+++ b/Source/Core/InputCommon/ControllerInterface/CoreDevice.cpp
@@ -258,10 +258,18 @@ std::vector<std::string> DeviceContainer::GetAllDeviceStrings() const
   return device_strings;
 }
 
+bool DeviceContainer::HasDefaultDevice() const
+{
+  std::lock_guard lk(m_devices_mutex);
+  // Devices are already sorted by priority
+  return !m_devices.empty() && m_devices[0]->GetSortPriority() >= 0;
+}
+
 std::string DeviceContainer::GetDefaultDeviceString() const
 {
   std::lock_guard lk(m_devices_mutex);
-  if (m_devices.empty())
+  // Devices are already sorted by priority
+  if (m_devices.empty() || m_devices[0]->GetSortPriority() < 0)
     return "";
 
   DeviceQualifier device_qualifier;
--- a/Source/Core/InputCommon/ControllerInterface/CoreDevice.h
+++ b/Source/Core/InputCommon/ControllerInterface/CoreDevice.h
@@ -136,6 +136,7 @@ public:
   // Use this to change the order in which devices are sorted in their list.
   // A higher priority means it will be one of the first ones (smaller index), making it more
   // likely to be index 0, which is automatically set as the default device when there isn't one.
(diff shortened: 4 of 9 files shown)


Source/Core/InputCommon/ControllerInterface/ControllerInterface.cpp
```diff
@@ -92,6 +94,24 @@ void ControllerInterface::Initialize(const WindowSystemInfo& wsi)
...
+  // so that we can then safely load the default input configs for different controllers,
+  // as otherwise their default device would end up being empty and the default key bindings would
+  // not be linked to any device (by design, the first device added to the CI is the default default
+  // device, and it's usually the "keyboard and mouse" on every OS, which should always be present).
+  if (wait_for_first_device)
+  {
+    // Somebody might have triggered InvokeDevicesChangedCallbacks() without actually
+    // adding any device, so check against that and reset the event.
+    // In some extremely rare occasions devices might change in between the two checks,
+    // but at worse we'll wait a few seconds.
```
JosJuice, line 108: It should be possible to fix this race condition properly. Waiting for five seconds sounds rather annoying.

A `Common::Event` is essentially a wrapper around a mutex and a condition variable. The key to avoiding the race condition is to take the mutex before you check `m_devices.empty()` and keep holding it until you're done with the `wait_for` call. I don't think it's possible to do this using the `Common::Event` wrapper class, so my suggestion is that you use a raw `std::condition_variable` in the following way:

1. Lock `m_devices_mutex`
2. Check `m_devices.empty()` (do it anew – don't just read the old `devices_empty` variable, since we've unlocked `m_devices_mutex` since then)
3. If it was empty, call `wait_for` on the condition variable
4. Unlock `m_devices_mutex`

This would also require that the code that calls notify on the condition variable has `m_devices_mutex` locked. I suppose it could make sense to put the notify call in `AddDevice` instead of in `InvokeDevicesChangedCallbacks`?

Filoppi: Should be all safe now.

JosJuice (approved):
I suppose this solution works too. Personally I like to use a raw condition variable in cases like this, but I know there are others that prefer `Common::Event`, so let's not make that a blocker.

PatrickFerry:
I had to build myself since the buildbot seems to be having trouble with building for Windows,

It works to resolved the issue.

dezraj:
Hello! I've downloaded Dolphin 5.0-15611 (most recent version for Windows) from build-bot and the issue is still there.

Is there a build already available with this fix or should I wait?

JosJuice:
Once this pull request says "Merged" instead of "Open", the change will be available in a build at https://dolphin-emu.org/. For now, either download a test build from this page, or wait.

Source/Core/DolphinQt/MainWindow.cpp
```diff
@@ -319,6 +319,13 @@ void MainWindow::InitControllers()
     return;
 
   g_controller_interface.Initialize(GetWindowSystemInfo(windowHandle()));
+  if (!g_controller_interface.HasDefaultDevice())
+  {
+    // Note that the CI default device could be still temporarily removed at any time
+    WARN_LOG(CONTROLLERINTERFACE,
+             "No default divice has been added in time. EmulatedController(s) defaulting"
+             " adds bindings made for a specific default device per platform");
```
JosJuice, line 327: The second sentence here doesn't make sense to me.

Also, `device` is misspelled in the first sentence.

Source/Core/InputCommon/ControllerInterface/ControllerInterface.cpp
```diff
@@ -159,6 +159,9 @@ void ControllerInterface::RefreshDevices(RefreshReason reason)
   // with their own PlatformPopulateDevices().
   // This means that devices might end up in different order, unless we override their priority.
   // It also means they might appear as "disconnected" in the Qt UI for a tiny bit of time.
+  // This helps the emulation and host thread to not stall when repopulating devices for any reason.
+  // Every platform that adds a device that is meant to be used as default device should try to not
+  // do it async, to not risk the emulated controllers defaulting not finding a default device.
```
JosJuice, line 164: "defaulting not finding a default device" doesn't make sense grammatically, but with this one I at least do understand what the comment is supposed to mean.

Source/Core/InputCommon/ControllerInterface/CoreDevice.cpp
```diff
@@ -258,10 +258,20 @@ std::vector<std::string> DeviceContainer::GetAllDeviceStrings() const
   return device_strings;
 }
 
+bool DeviceContainer::HasDefaultDevice() const
+{
+  std::lock_guard lk(m_devices_mutex);
+  // Devices are already sorted by priority
+  if (m_devices.empty() || m_devices[0]->GetSortPriority() < 0)
+    return false;
+  return true;
```
JosJuice, line 267:
```suggestion
  return !m_devices.empty() && m_devices[0]->GetSortPriority() >= 0;
```

Filoppi: Addressed all

Source/Core/InputCommon/ControllerInterface/ControllerInterface.cpp
JosJuice: Ah, I see now, you meant "emulated controllers' defaulting". A little confusing using the word "defaulting" like a noun, but I suppose it works.

PR #10249, opened 2021-11-27
Support utvideo lossless codec

vadosnaprimer:
Updated ffmpeg to v4.4. I wonder if it makes sense to commit ffmpeg source as a submodule to make future updating even more straightforward. Yes you can use that build script on any ffmpeg version and it should work fine.

To enable utvideo codec you need this in `GFX.ini`

```
UseFFV1 = False
DumpCodec = utvideo
```

Probably better approach would be to have a dropdown menu to select between ffv1/utvideo/mpeg4, but I've never worked with Qt GUI.


JosJuice:
@dolphin-emu-bot rebuild

vadosnaprimer:
I have a feeling @RisingFog will end up getting pinged :D

RisingFog:
@vadosnaprimer yeah, because you just pinged me with that message :P

On the surface, I don't see any inherent issues with the PR since it's just a simple dependency update.

I can't really comment on making it visible in the settings UI, or if it should be made a submodule. I'll leave those comments to others who are actually active in Dolphin development.

As an aside, I do think that the goal should eventually be allowing full customization of dump options, including the possibility of muxing audio and video into a single file in real-time. However, that's something that should be looked at down the line.

JMC47:
Currently there is a problem with muxing video/audio into a single file.  An accuracy change to reflect a bug in the GameCube's hardware has us changing the Hz slightly, and that has been bisected as the cause of the audio files no longer lining up with the video files.  So something needs to be done with that before we allow for muxing into one file realtime.  I might submit a PR reverting the change so it's not blocking anything.

AdmiralCurtiss:
IMO if you can make it a submodule, do so.

Zinfidel:
Just ran some highly-unscientific tests on this branch. Dumping 3 seconds of 3840x2948 video with FFV1 results in 1 FPS and a 1.1GB file. Dumping the same 3 seconds using UTVideo results in 3 FPS and a 1.4GB file. The section of video is a fully-3D scene from Spyro, not big blocks of static color or anything.

Dumping hour+ movies at 4K+ sizes is pretty painful, so a 3x speedup here is huge. File is ~30% bigger, and based on what I've seen of FFV1 in the past the margin may be even larger in a real scenario, but 30% isn't a problem at least for me.

vadosnaprimer:
> IMO if you can make it a submodule, do so.

Done.

So what's the fate of this PR?

AdmiralCurtiss:
Please squash this (lots of redundant commits here) and make sure it builds on all builders.

I also question having binaries in the repo when the source is right there next to it. Can we just build this instead?

vadosnaprimer:
> I also question having binaries in the repo when the source is right there next to it. Can we just build this instead?

Only if msys2 is installed, not sure it everyone wants it to become a hard dependency.

JosJuice:
@dolphin-emu-bot rebuild

vadosnaprimer:
Build and lint should be good now.

shuffle2:
1. Please revert the newline changes to .vcxproj files.
2. Remove the submodule. It would make sense to include mainline ffmpeg repo as submodule if we build from source, but we don't.
3. Probably a better idea (if you want to go submodule route) is to have a submodule with the binaries and the build script / any required dolphin patches in it (similar to how Qt is handled).

ffmpeg build doesn't require msys. It can be built with WSL (_highly_ preferable to msys or whatever - but don't want either to become a dependency, anyway). See previous discussions:
https://github.com/dolphin-emu/dolphin/pull/8547
https://github.com/dolphin-emu/dolphin/pull/8495#pullrequestreview-334316841
And a repo I've made before:
https://github.com/shuffle2/FFmpeg/tree/dolphin

vadosnaprimer:
What if I commit my script and files it build, and if someone refactors the ffmpeg setup, it can be changed in the future? An important bit is having this documented and easily reproducible.

vadosnaprimer:
Is it already integrated into dolphin and documented on its side?

AdmiralCurtiss:
This should be fixed with #10530, so I'm closing this.

PR #10250, opened 2021-11-27
DolphinQt: Add i18n comment for "Save Game"

JosJuice:
It's very easy to assume the "save" in this string is a verb.


lioncash (approved):

Miksel12:
Wouldn't 'Game Save' make more sense?

AdmiralCurtiss:
I agree that the nomenclature here is a bit weird, but that's what mGBA standalone calls them so I wanted to remain consistent:

image

PR #9660, opened 2021-04-21
VideoCommon: Fix scissorOffset, handle negative value correctly

ezio1900:
VideoCommon: Fix scissorOffset, handle negative value correctly
VideoBackends: Fix Software Clipper.PerspectiveDivide function, use bpmem.scissorOffset instead of hard code 342

Origin pull request: [#9656](https://github.com/dolphin-emu/dolphin/pull/9656)
Can some one help me to reopen #9656, I found a way to reset the commits.

PR #10251, opened 2021-11-28
Rework scissor handling

Pokechu22:
This fixes [the rendering issue in Major Minor's Majestic March](https://bugs.dolphin-emu.org/issues/12223).

There are some situations where the emulated scissor configuration results in multiple real scissors being needed.  I couldn't find an easy way to implement that on the hardware backends, but I did implement it in software.  For the hardware backends, I pick the biggest one.  This affects Major Minor, but only briefly during scrolling loading screens (right near the end and only for ~4 frames).  Compare frames [204](https://user-images.githubusercontent.com/8334194/145466628-d64f39a7-0c74-432b-a021-e892654b42a8.png) and [205](https://user-images.githubusercontent.com/8334194/145466693-8fc27e5f-b6f8-4a73-886b-08503bd411a2.png) in major_minor_loading_long.dff.  Here's what's drawn behind the loading text on those same frames (this is via the software renderer): [204](https://user-images.githubusercontent.com/8334194/145092021-4e9dcbac-2787-44de-a6f7-d55bb0fb6b4d.png) and [205](https://user-images.githubusercontent.com/8334194/145092093-506dd5a7-bab1-4f7d-a9b5-067301195569.png).

Here's [my hardware test](https://gist.github.com/Pokechu22/51f6f53862b0298bafa6cddca111b699), and [a video of that test on hardware](https://youtu.be/nUcioBIpVJw) (this is probably something that can be changed into something for the hwtests repo, but I haven't gotten to that yet).  The generated test results are [also in that gist](https://gist.github.com/Pokechu22/51f6f53862b0298bafa6cddca111b699#file-results-7z).

The software renderer generates mostly matching results for `Scissor_info_vo.log` - the colors don't match exactly, but it finds the exact same rectangles.  The hardware renderers give accurate results, but only have on rectangle.  Neither of them gives completely accurate results for `Scissor_info_all.log`, but that's more because of weird behavior with clipping, not with the scissor.  The hardware renderers reject anything outside of the viewport, while the software renderer now sometimes draws things outside of the viewport (based on the xfmem configuration); this fixes the [nintendo channel lettering](https://bugs.dolphin-emu.org/issues/12562) on the software renderer.  However, there seems to be additional oddities with clipping that haven't been fixed yet (even when clipping is enabled on hardware, stuff outside of the viewport can exist, and even with clipping disabled you only are allowed to go so far outside of the viewport). I don't currently plan on investigating clipping further at this time.

I also created an imgui-based tool to visualized the scissor rectangles, and have created some videos based on it.  "Old" and "current" refer to behavior before and after PR #9660, "nooff" refers to having the scissor offset function completely disabled, and "new" is this PR.

<details><summary>Title screen (major_minor_title_long.dff)</summary>

https://user-images.githubusercontent.com/8334194/143727979-83a0c4d6-1af2-4caf-a8ee-3d61d17b6b51.mp4

https://user-images.githubusercontent.com/8334194/143727989-e6305a35-2b2e-4627-abc9-664a015afd72.mp4

https://user-images.githubusercontent.com/8334194/143728003-2e2174af-d080-47f3-9a80-5447f67d0725.mp4

https://user-images.githubusercontent.com/8334194/145466817-4c861f97-36fe-4792-82df-103a92e12540.mp4

</details>
<details><summary>Loading screen (major_minor_loading_long.dff)</summary>

https://user-images.githubusercontent.com/8334194/143727699-f0beb442-bb84-48c2-bf21-6ac1e377e6d9.mp4

https://user-images.githubusercontent.com/8334194/143727680-e0a46e25-86bc-4ae0-805f-3a36773c0a16.mp4

https://user-images.githubusercontent.com/8334194/143727694-17c89ab9-ae64-4892-9915-edfae10d184b.mp4

https://user-images.githubusercontent.com/8334194/145466856-e4ae8e87-0a9a-41c0-a9c2-8a7494723256.mp4

</details>
<details><summary>Retry screen (major_minor_retry_long.dff)</summary>

https://user-images.githubusercontent.com/8334194/143728062-2186e177-c650-4672-9059-9b0070282f04.mp4

https://user-images.githubusercontent.com/8334194/143728072-3a5230a9-8ea6-4dc1-858b-5addff241e28.mp4

https://user-images.githubusercontent.com/8334194/143728080-25c400f4-3742-46bb-a631-7f3d13760132.mp4

https://user-images.githubusercontent.com/8334194/145466909-344d5e99-71bc-4e24-8646-2f7debc1fa6e.mp4

</details>
<details><summary>Super Mario Galaxy boss roars</summary>

Old:  
SMGOld

Current:  
SMGCurrent

No off (+ second screenshot with some objects disabled to see what's going on better):  

New:  

</details>

I also have some notes [here](https://gist.github.com/Pokechu22/5f83afb548bef8d75d3575d1c02bd518), but these probably aren't worth reading, as they're messy and currently unfinished.  An un-squashed version of this code (which also includes hack I used to framedump with imgui included) is [here](https://github.com/pokechu22/dolphin/tree/archive/negative-scissor-presquash1).

### DIFF
--- a/Source/Core/VideoCommon/BPFunctions.cpp
+++ b/Source/Core/VideoCommon/BPFunctions.cpp
@@ -7,6 +7,7 @@
 #include <cmath>
 #include <string_view>
 
+#include "Common/Assert.h"
 #include "Common/CommonTypes.h"
 #include "Common/Logging/Log.h"
 
@@ -37,48 +38,172 @@ void SetGenerationMode()
   g_vertex_manager->SetRasterizationStateChanged();
 }
 
-void SetScissor()
+int ScissorRect::GetArea() const
 {
-  /* NOTE: the minimum value here for the scissor rect is -342.
-   * GX SDK functions internally add an offset of 342 to scissor coords to
-   * ensure that the register was always unsigned.
-   *
-   * The code that was here before tried to "undo" this offset, but
-   * since we always take the difference, the +342 added to both
-   * sides cancels out. */
+  return rect.GetWidth() * rect.GetHeight();
+}
 
-  /* NOTE: With a positive scissor offset, the scissor rect is shifted left and/or up;
-   * With a negative scissor offset, the scissor rect is shifted right and/or down.
-   *
-   * GX SDK functions internally add an offset of 342 to scissor offset.
-   * The scissor offset is always even, so to save space, the scissor offset register
-   * is scaled down by 2. So, if somebody calls GX_SetScissorBoxOffset(20, 20);
-   * the registers will be set to ((20 + 342) / 2 = 181, 181).
-   *
-   * The scissor offset register is 10bit signed [-512, 511].
-   * e.g. In Super Mario Galaxy 1 and 2, during the "Boss roar effect",
-   * for a scissor offset of (0, -464), the scissor offset register will be set to
-   * (171, (-464 + 342) / 2 = -61).
-   */
-  s32 xoff = bpmem.scissorOffset.x * 2;
-  s32 yoff = bpmem.scissorOffset.y * 2;
+int ScissorResult::GetViewportArea(const ScissorRect& rect) const
+{
+  int x0 = std::clamp<int>(rect.rect.left + rect.x_off, viewport_left, viewport_right);
+  int x1 = std::clamp<int>(rect.rect.right + rect.x_off, viewport_left, viewport_right);
 
-  MathUtil::Rectangle<int> native_rc(bpmem.scissorTL.x - xoff, bpmem.scissorTL.y - yoff,
-                                     bpmem.scissorBR.x - xoff + 1, bpmem.scissorBR.y - yoff + 1);
-  native_rc.ClampUL(0, 0, EFB_WIDTH, EFB_HEIGHT);
+  int y0 = std::clamp<int>(rect.rect.top + rect.y_off, viewport_top, viewport_bottom);
+  int y1 = std::clamp<int>(rect.rect.bottom + rect.y_off, viewport_top, viewport_bottom);
 
-  auto target_rc = g_renderer->ConvertEFBRectangle(native_rc);
+  return (x1 - x0) * (y1 - y0);
+}
+
+// Compare so that a sorted collection of rectangles has the best one last, so that if they're drawn
+// in order, the best one is the one that is drawn last (and thus over the rest).
+// The exact iteration order on hardware hasn't been tested, but silly things can happen where a
+// polygon can intersect with itself; this only applies outside of the viewport region (in areas
+// that would normally be affected by clipping).  No game is known to care about this.
+bool ScissorResult::IsWorse(const ScissorRect& lhs, const ScissorRect& rhs) const
+{
+  // First, penalize any rect that is not in the viewport
+  int lhs_area = GetViewportArea(lhs);
+  int rhs_area = GetViewportArea(rhs);
+
+  if (lhs_area != rhs_area)
+    return lhs_area < rhs_area;
+
+  // Now compare on total areas, without regard for the viewport
+  return lhs.GetArea() < rhs.GetArea();
+}
+
+namespace
+{
+// Dynamically sized small array of ScissorRanges (used as an heap-less alternative to std::vector
+// to reduce allocation overhead)
+struct RangeList
+{
+  static constexpr u32 MAX_RANGES = 9;
+
+  u32 m_num_ranges = 0;
+  std::array<ScissorRange, MAX_RANGES> m_ranges{};
+
+  void AddRange(int offset, int start, int end)
+  {
+    DEBUG_ASSERT(m_num_ranges < MAX_RANGES);
+    m_ranges[m_num_ranges] = ScissorRange(offset, start, end);
+    m_num_ranges++;
+  }
+  auto begin() const { return m_ranges.begin(); }
+  auto end() const { return m_ranges.begin() + m_num_ranges; }
+
+  u32 size() { return m_num_ranges; }
+};
+
+static RangeList ComputeScissorRanges(int start, int end, int offset, int efb_dim)
+{
+  RangeList ranges;
+
+  for (int extra_off = -4096; extra_off <= 4096; extra_off += 1024)
+  {
+    int new_off = offset + extra_off;
+    int new_start = std::clamp(start - new_off, 0, efb_dim);
+    int new_end = std::clamp(end - new_off + 1, 0, efb_dim);
+    if (new_start < new_end)
+    {
+      ranges.AddRange(new_off, new_start, new_end);
+    }
+  }
+
+  return ranges;
+}
+}  // namespace
+
+ScissorResult::ScissorResult(const BPMemory& bpmemory, const XFMemory& xfmemory)
+    : ScissorResult(bpmemory,
+                    std::minmax(xfmemory.viewport.xOrig - xfmemory.viewport.wd,
+                                xfmemory.viewport.xOrig + xfmemory.viewport.wd),
+                    std::minmax(xfmemory.viewport.yOrig - xfmemory.viewport.ht,
+                                xfmemory.viewport.yOrig + xfmemory.viewport.ht))
+{
+}
+ScissorResult::ScissorResult(const BPMemory& bpmemory, std::pair<float, float> viewport_x,
+                             std::pair<float, float> viewport_y)
+    : scissor_tl{.hex = bpmemory.scissorTL.hex}, scissor_br{.hex = bpmemory.scissorBR.hex},
+      scissor_off{.hex = bpmemory.scissorOffset.hex}, viewport_left(viewport_x.first),
+      viewport_right(viewport_x.second), viewport_top(viewport_y.first),
+      viewport_bottom(viewport_y.second)
+{
+  // Range is [left, right] and [top, bottom] (closed intervals)
+  const int left = scissor_tl.x;
+  const int right = scissor_br.x;
+  const int top = scissor_tl.y;
+  const int bottom = scissor_br.y;
+  // When left > right or top > bottom, nothing renders (even with wrapping from the offsets)
+  if (left > right || top > bottom)
+    return;
+
+  // Note that both the offsets and the coordinates have 342 added to them internally by GX
+  // functions (for the offsets, this is before they are divided by 2/right shifted). This code
+  // could undo both sets of offsets, but it doesn't need to since they cancel out when subtracting
+  // (and those offsets actually matter for the left > right and top > bottom checks).
+  const int x_off = scissor_off.x << 1;
+  const int y_off = scissor_off.y << 1;
+
+  RangeList x_ranges = ComputeScissorRanges(left, right, x_off, EFB_WIDTH);
+  RangeList y_ranges = ComputeScissorRanges(top, bottom, y_off, EFB_HEIGHT);
+
+  m_result.reserve(x_ranges.size() * y_ranges.size());
+
+  // Now we need to form actual rectangles from the x and y ranges,
+  // which is a simple Cartesian product of x_ranges_clamped and y_ranges_clamped.
+  // Each rectangle is also a Cartesian product of x_range and y_range, with
+  // the rectangles being half-open (of the form [x0, x1) X [y0, y1)).
+  for (const auto& x_range : x_ranges)
+  {
+    DEBUG_ASSERT(x_range.start < x_range.end);
+    DEBUG_ASSERT(x_range.end <= EFB_WIDTH);
+    for (const auto& y_range : y_ranges)
+    {
+      DEBUG_ASSERT(y_range.start < y_range.end);
+      DEBUG_ASSERT(y_range.end <= EFB_HEIGHT);
+      m_result.emplace_back(x_range, y_range);
+    }
+  }
+
+  auto cmp = [&](const ScissorRect& lhs, const ScissorRect& rhs) { return IsWorse(lhs, rhs); };
+  std::sort(m_result.begin(), m_result.end(), cmp);
+}
+
+ScissorRect ScissorResult::Best() const
+{
+  // For now, simply choose the best rectangle (see ScissorResult::IsWorse).
+  // This does mean we calculate all rectangles and only choose one, which is not optimal, but this
+  // is called infrequently.  Eventually, all backends will support multiple scissor rects.
+  if (!m_result.empty())
+  {
+    return m_result.back();
+  }
+  else
+  {
+    // But if we have no rectangles, use a bogus one that's out of bounds.
+    // Ideally, all backends will support multiple scissor rects, in which case this won't be
+    // needed.
+    return ScissorRect(ScissorRange{0, 1000, 1001}, ScissorRange{0, 1000, 1001});
+  }
+}
+
+ScissorResult ComputeScissorRects()
+{
+  return ScissorResult{bpmem, xfmem};
+}
+
+void SetScissorAndViewport()
+{
+  auto native_rc = ComputeScissorRects().Best();
+
+  auto target_rc = g_renderer->ConvertEFBRectangle(native_rc.rect);
   auto converted_rc =
       g_renderer->ConvertFramebufferRectangle(target_rc, g_renderer->GetCurrentFramebuffer());
   g_renderer->SetScissorRect(converted_rc);
-}
 
-void SetViewport()
-{
-  const s32 xoff = bpmem.scissorOffset.x * 2;
-  const s32 yoff = bpmem.scissorOffset.y * 2;
-  float raw_x = xfmem.viewport.xOrig - xfmem.viewport.wd - xoff;
-  float raw_y = xfmem.viewport.yOrig + xfmem.viewport.ht - yoff;
+  float raw_x = (xfmem.viewport.xOrig - native_rc.x_off) - xfmem.viewport.wd;
+  float raw_y = (xfmem.viewport.yOrig - native_rc.y_off) + xfmem.viewport.ht;
   float raw_width = 2.0f * xfmem.viewport.wd;
   float raw_height = -2.0f * xfmem.viewport.ht;
   if (g_ActiveConfig.UseVertexRounding())
--- a/Source/Core/VideoCommon/BPFunctions.cpp
+++ b/Source/Core/VideoCommon/BPFunctions.cpp
@@ -153,17 +153,6 @@ void SetViewport()
     far_depth = 1.0f - min_depth;
   }
 
-  // Clamp to size if oversized not supported. Required for D3D.
-  if (!g_ActiveConfig.backend_info.bSupportsOversizedViewports)
-  {
-    const float max_width = static_cast<float>(g_renderer->GetCurrentFramebuffer()->GetWidth());
-    const float max_height = static_cast<float>(g_renderer->GetCurrentFramebuffer()->GetHeight());
-    x = std::clamp(x, 0.0f, max_width - 1.0f);
-    y = std::clamp(y, 0.0f, max_height - 1.0f);
-    width = std::clamp(width, 1.0f, max_width - x);
-    height = std::clamp(height, 1.0f, max_height - y);
-  }
-
   // Lower-left flip.
   if (g_ActiveConfig.backend_info.bUsesLowerLeftOrigin)
     y = static_cast<float>(g_renderer->GetCurrentFramebuffer()->GetHeight()) - y - height;
--- a/Source/Core/VideoCommon/BPFunctions.h
+++ b/Source/Core/VideoCommon/BPFunctions.h
@@ -7,16 +7,131 @@
 
 #pragma once
 
-#include "Common/MathUtil.h"
+#include <utility>
+#include <vector>
 
-struct BPCmd;
+#include "Common/MathUtil.h"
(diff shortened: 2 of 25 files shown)


Source/Core/VideoCommon/BPFunctions.h
```diff
@@ -7,16 +7,129 @@
...
+  float viewport_right;
+  float viewport_top;
+  float viewport_bottom;
+
+  // Actual result
+  std::vector<ScissorRect> m_result;
+
+  ScissorRect Best() const;
+
+  constexpr bool ScissorMatches(const ScissorResult& other) const
```
lioncash, line 100: These functions marked constexpr can never actually be used in a constexpr context, since the constructors aren't constexpr

Source/Core/VideoCommon/BPFunctions.h
```diff
@@ -7,16 +7,129 @@
...
+  ScissorResult(const BPMemory& bpmem, const XFMemory& xfmem);
+  ~ScissorResult() = default;
+  ScissorResult(const ScissorResult& other)
+      : scissor_tl{.hex = other.scissor_tl.hex}, scissor_br{.hex = other.scissor_br.hex},
+        scissor_off{.hex = other.scissor_off.hex}, viewport_left{other.viewport_left},
+        viewport_right{other.viewport_right}, viewport_top{other.viewport_top},
+        viewport_bottom{other.viewport_bottom}, m_result{other.m_result}
+  {
+  }
+  ScissorResult& operator=(const ScissorResult& other)
```
lioncash, line 54: Should this and the move constructor have an:

```cpp
if (this == &other)
  return *this;
```

check in each of them?

JMC47:
Holy shit.  I'd usually write something more on topic, but the level of depth and difficulty of this problem was astounding.  Kudos to finally fixing it.

Does this per-chance affect that other game that rendered black over the screen?  Or did that end up being separate?

Pokechu22:
Looking more closely at the fifoci results above, different parts of this caused different changes.  It's worth noting that several of the software renderer fixes come from the "fix scissor rectangle always being block-aligned" change; this was a bug in the software renderer that _was not_ present in the hardware backends and anything that was fixed by that should have already worked outside of the software renderer.

The issue effectively was like this code existing (except this would affect all backends):

```patch
diff --git a/Source/Core/VideoCommon/BPStructs.cpp b/Source/Core/VideoCommon/BPStructs.cpp
index ec5ab6b348..4317be33dd 100644
--- a/Source/Core/VideoCommon/BPStructs.cpp
+++ b/Source/Core/VideoCommon/BPStructs.cpp
@@ -131,6 +131,10 @@ static void BPWritten(const BPCmd& bp)
   case BPMEM_SCISSORTL:      // Scissor Rectable Top, Left
   case BPMEM_SCISSORBR:      // Scissor Rectable Bottom, Right
   case BPMEM_SCISSOROFFSET:  // Scissor Offset
+    bpmem.scissorTL.x = bpmem.scissorTL.x & ~1;
+    bpmem.scissorTL.y = bpmem.scissorTL.y & ~1;
+    bpmem.scissorBR.x = bpmem.scissorBR.x | 1;
+    bpmem.scissorBR.y = bpmem.scissorBR.y | 1;
     VertexShaderManager::SetViewportChanged();
     GeometryShaderManager::SetViewportChanged();
     return;
```

Here are the individual differences and an explanation for why they changed:

* `dbz-depth`: The game uses a scissor of `[0, 639) X [0, 447)` with an offset of 0 for the start of the frame, and changed that scissor to `[0, 640) X [0, 448)` in object 313 (the start of a bloom effect).  The first scissor causes black borders on the bottom and right of the screen; the black borders bleed inward from the bloom later on.  The hardware renderers already handle this.  I think this is probably a game bug and not a deliberate effect.
* `jb-shadow`:   When drawing shadows, a scissor of `[1, 127) X [1, 127)` with offset 0 is used, and then an EFB copy of `[0, 128) X [0, 128)` is made.  When the shadows are drawn, the texture from that EFB copy is set to clamp, so the pixels on the border get drawn for anything out of bounds.  Thus those pixels should be transparent; the scissor is used to achieve this.  The hardware renderers already handle this.  Compare [before](https://user-images.githubusercontent.com/8334194/143811266-aef0cb72-8115-4ca6-a45a-57fe7ea53bce.png) and [after](https://user-images.githubusercontent.com/8334194/143811274-c1279053-bf85-4d87-ac74-3bd91ffc1f02.png) (where red is originally transparent, and black is originally white).
* `nfsu-reflections`: A viewport of `[343, 407) X [343, 407)` is used (normally the viewport has a top-left of (342, 342)).  The scissor is adjusted accordingly to use `[343, 405) X [343, 405)` (with raw coordinates).  This would trigger the block-alignment issue.  No scissor offset is used, though.  I don't exactly know what's going on with rendering here, but the difference is probably caused by this.  The hardware renderers already handled it correctly.  This fifolog is somewhat broken visually already, and probably the game needs to be tested running normally (though that will be annoying when using the software renderer).
* `nhl-slap`: A stupid number of different scissor rectangle configurations are used here, 34 total (33 of which are for the UI; the game itself only uses one which covers the whole screen).  Number 17 is relevant here, using `[464, 343) X [598, 389)`.  The block alignment issue meant that one extra column of pixels was rendered on the grey highlight behind "P. Marleau".  The hardware renderers already handled this correctly.
* `sms-water`: D.E.B.S. uses a scissor of `[45, 390) X [499, 415)`, which means the block alignment issue would affect it.  Hardware was unaffected.  I was unable to reproduce the change to the water on my machine.  However, some parts of the scene do have clipping disabled, including object 350 (which comes right after the water renders).
* `sw3-dt`: The bloom effect draws a black rectangle using a scissor of `[0, 82) X [0, 58)`.  It then uses a scissor of `[1, 81) X [1, 57)` when drawing a small copy of the screen.  In both cases, no scissor offset is used.  Finally, it makes an EFB copy of the screen region `[0, 82) X [0, 58)`, meaning a black border is included.  This is what produces the vignette.  Unlike dbz, this seems to be a deliberate choice.  The hardware renderers already handled this correctly, but the fifolog also experiences vertex explosions after the first frame, so this should be tested by running the game normally too.

* `nintendo-channel` is fixed because of clipdisable.
* `burnout2-vehicletextures`: Very slight change, apparently introduced by "Use new scissor logic" - I think it might be floating-point rounding stuff since the scissor offset is no longer subtracted from vertex coordinates.  I'm not entirely sure though.
* `rs3-bumpmapping`: This was affected by both "Use new scissor logic" and the clipdisable change.  I think this is also a floating-point rounding difference.  I'm also not sure about this one.
* `djhero2-blend`: This was caused by the clipdisable change.  I think it's an integer overflow, as the floor triangles are fairly big: [image](https://user-images.githubusercontent.com/8334194/143812044-ec98c2a0-79da-47e0-b1bb-8a78b5c5f3d6.png) and [freelook image](https://user-images.githubusercontent.com/8334194/143812042-da00cd21-b1a5-418b-8ec1-d850afdcb348.png).  This is a regression, but I think more investigation into clipping would be needed to fix it, so I don't know if it needs to be dealt with just yet.

This does not fix Gormiti: The Lords of Nature (the game that has the black screen), but I didn't expect it to either.  And, `smg-roar` (which uses the scissor offset) continues to render correctly (which is good).

Source/Core/VideoCommon/BPFunctions.cpp
```diff
@@ -36,48 +37,149 @@ void SetGenerationMode()
...
+  // Range is [left, right] and [top, bottom] (closed intervals)
+  const int left = scissor_tl.x;
+  const int right = scissor_br.x;
+  const int top = scissor_tl.y;
+  const int bottom = scissor_br.y;
+  // When left > right or top > bottom, nothing renders (even with wrapping from the offsets)
+  if (left > right || top > bottom)
+    return;
+
+  // Note that both the offsets and the coordinates have 342 added to them internally
```
neobrain, line 116: "internally" could refer to either hardware behavior or GX APIs. I'd vouch for leaving the mention of GX in the comment.

Seems like a worthwhile rewording other than that :)

Source/Core/VideoCommon/BPFunctions.cpp
```diff
@@ -36,48 +37,149 @@ void SetGenerationMode()
...
+  int rhs_area = GetViewportArea(rhs);
+
+  if (lhs_area != rhs_area)
+    return lhs_area < rhs_area;
+
+  // Now compare on total areas, without regard for the viewport
+  return lhs.GetArea() < rhs.GetArea();
+}
+
+static std::vector<ScissorRange> ComputeScissorRanges(int start, int end, int offset, int efb_dim)
```
neobrain, line 74: Not sure if this could cause any problems whatsoever, but allocating memory through std::vector for this seems wasteful (note that std::vector has no small object optimization).

I'd just go with an `std::array<ScissorRange>` where unused elements are "null" ranges, or alternatively a pair of that array with an element count.

Source/Core/VideoCommon/BPFunctions.h
```diff
@@ -7,16 +7,132 @@
...
+      return false;
+    return true;
+  }
+
+private:
+  ScissorResult(const BPMemory& bpmem, std::pair<float, float> viewport_x,
+                std::pair<float, float> viewport_y);
+
+  int GetViewportArea(const ScissorRect& rect) const;
+  bool IsWorse(const ScissorRect& lhs, const ScissorRect& rhs) const;
```
neobrain, line 128: Should this be a static member function?

Source/Core/VideoCommon/BPFunctions.cpp
```diff
@@ -36,48 +37,149 @@ void SetGenerationMode()
...
+    DEBUG_ASSERT(x_range.end <= EFB_WIDTH);
+    for (const auto& y_range : y_ranges)
+    {
+      DEBUG_ASSERT(y_range.start < y_range.end);
+      DEBUG_ASSERT(y_range.end <= EFB_HEIGHT);
+      m_result.emplace_back(x_range, y_range);
+    }
+  }
+
+  auto cmp = [&](const ScissorRect& lhs, const ScissorRect& rhs) { return IsWorse(lhs, rhs); };
```
neobrain, line 145: If `IsWorse` is made static, this lambda isn't needed and `IsWorse` can just be passed directly to `std::sort`.

Source/Core/VideoCommon/BPFunctions.h
```diff
@@ -7,16 +7,132 @@
 
 #pragma once
 
-#include "Common/MathUtil.h"
+#include <set>
```
neobrain, line 10: This header appears to be unused.

Source/Core/VideoCommon/BPMemory.h
```diff
@@ -1050,24 +1050,62 @@ struct fmt::formatter<LPSize>
...
 };
 
-union X12Y12
+union ScissorPos
+{
+  // The top bit is ignored, and not included in the mask used by SDK functions
+  // (though libogc includes it for the bottom coordinate (only) for some reason)
+  // The SDK also adds 342 to this value.
+  BitField<0, 11, u32> y;
+  BitField<0, 12, u32> y_full;
```
neobrain, line 1059: The `_full` members could use a short comment explaining what distinguishes them from the non-`_full` ones, if that's information we know.

Source/Core/VideoCommon/BPFunctions.h
```diff
@@ -7,16 +7,132 @@
...
+  }
+
+  MathUtil::Rectangle<int> rect;
+  int x_off;
+  int y_off;
+
+  int GetArea() const;
+};
+
+struct ScissorResult
```
neobrain, line 43: I haven't seen any code documentation that explains why there are multiple scissor rects and how one is better than others among them. Maybe a docstring for this struct would be a good place to have this? (Discard if there is documentation already and I just happened to miss it)

Source/Core/VideoCommon/BPFunctions.h
Pokechu22: It can't be static, because it needs to call `GetViewportArea` which is non-static.

Pokechu22:
After a bit of investigating, I've figured out why djhero2-blend is broken by the clipdisable change.  Here are 3 relevant input vertices:

|x|y|z|
|-|-|-|
|0|-32765|0|
|10451|-32765|-3170|
|10711|-32765|-2130|

Their projected positions become these:

|x|y|z|w|
|-|-|-|-|
|-1393.6|318.1|10.0|-22.7|
|143.8|97.8|10.1|331.6|
|136.5|-105.5|-10.1|451.1|

Which then becomes this after the perspective division:

|x|y|z|
|-|-|-|
|20269.4|-2618.1|24150130|
|800.8|637.3|16265513|
|758.8|623.3|16399608|

As explained [here](https://stackoverflow.com/a/41087445), negative w is problematic for clipping; you'd expect a negative x to remain off-screen in the negative direction, but now it's a large positive value (and y is negative, and z is bigger than it should be).  For now I think the easiest workaround is to still do clipping if any w value is negative, until further hardware testing explains how hardware actually works when `disable_clipping_detection` is used.

Source/Core/VideoCommon/BPMemory.h
Pokechu22: There isn't any information other than the top bit being ignored - `y` is `y_full` with that bit ignored.  But the comment doesn't currently make that clear.

Source/Core/VideoCommon/BPFunctions.cpp
Pokechu22: I ended up going for my own struct which wraps `std::array`, so that range-based for loops could still work.

Source/Core/VideoCommon/BPFunctions.h
Pokechu22: I've added some documentation here; let me know how it looks.

Source/Core/VideoCommon/BPFunctions.h
```diff
@@ -7,16 +7,154 @@
...
+// Although the GameCube/Wii have only one scissor configuration and only one viewport
+// configuration, some values can result in multiple parts of the screen being updated.
+// This can happen if the scissor offset combined with the bottom or right coordinate ends up
+// exceeding 1024; then, both sides of the screen will be drawn to, while the middle is not.
+// Major Minor's Majestic March causes this to happen during loading screens and other scrolling
+// effects, though it draws on top of one of them.
+// This can also happen if the scissor rectangle is particularly large, but this will usually
+// involve drawing content outside of the viewport, which Dolphin does not currently handle.
+//
+// The hardware backends can currently only use one viewport and scissor rectangle, so we need to
```
neobrain, line 53: Could we disable scissoring on the host altogether and instead implement it manually in the pixel shaders instead to fully support this case?

Source/Core/VideoCommon/BPFunctions.h
neobrain: Looks great, thanks!

Source/Core/VideoCommon/BPFunctions.cpp
```diff
@@ -36,48 +37,171 @@ void SetGenerationMode()
...
+  if (lhs_area != rhs_area)
+    return lhs_area < rhs_area;
+
+  // Now compare on total areas, without regard for the viewport
+  return lhs.GetArea() < rhs.GetArea();
+}
+
+namespace
+{
+// Avoid overhead of std::vector
```
neobrain, line 76:
```suggestion
// Dynamically sized small array of ScissorRanges (used as an heap-less alternative to std::vector to reduce allocation overhead)
```

Source/Core/VideoCommon/BPFunctions.cpp
neobrain: IsWorse can't be made static as discussed above, closing.

Source/Core/VideoCommon/BPFunctions.h
Pokechu22: I don't think so.  Masking two different sides of the screen could be done using a pixel shader pretty easily, but I don't think moving things to different sides is possible in just a pixel shader.

Most graphics APIs do support multiple viewports and scissors internally, though, so I think it might be possible to implement it with that.  I just haven't looked into the details.

I added some additional images with the software renderer where the loading text is removed which might help visualize it: Frame 204, [ImGui](https://user-images.githubusercontent.com/8334194/143727516-8601edbe-7d7c-42e6-867a-19bc1ab1f055.png) and [SW](https://user-images.githubusercontent.com/8334194/145092021-4e9dcbac-2787-44de-a6f7-d55bb0fb6b4d.png); Frame 205, [ImGui](https://user-images.githubusercontent.com/8334194/143727517-4f08454b-7457-4a43-9996-f5646c4d6469.png) and [SW](https://user-images.githubusercontent.com/8334194/145092093-506dd5a7-bab1-4f7d-a9b5-067301195569.png).

neobrain: Hmm not sure I'm following. Do I understand you correctly there are two components that to such scissor "overflows"?
1. Pixels on the very left and on the very right are rendered, leaving a big gap in the middle
2. The left subarea is not a true scissor rect, but instead content that ends up on the left subarea is rendered *as if* it had been rendered to the right of the right subarea, i.e. that content wraps around

Reading the comment it was easy to see 1 being the case, but 2 comes at a surprise. Granted, I may have glanced over the comment too quickly. If I didn't just miss it, I think it's a detail worth mentioning though.

Pokechu22: I think I just haven't explained things very well (I didn't do a writeup beforehand, other than my messy notes); it definitely could be clearer.

Both 1 and 2 are correct.  But I think there might be a clearer explanation.

The scissor rectangle itself _doesn't_ overflow.  You can give it any set of values `x0`, `y0`, `x1`, and `y1` all in the range `[0, 2047]` and things will render correctly: if `x0 > x1` or `y0 > y1`, nothing shows up; otherwise, pixels between those ranges will be rasterized.  This is after the viewport has been applied, so normally you want the viewport and scissor to match (and both the scissor and viewport have 342 added to their origins by GX functions).  (The viewport is also based on floats, and uses a center+radius format, which is a bit confusing, but that can be ignored here).  For instance, a 640 by 480 viewport and scissor would have the scissor coordinates 342, 342, 981, 821, and a viewport for 342, 342, 982, 822.

You can still increase the size of the scissor, and things behave normally for any size (even 0, 0, 2047, 2047), except that you'll start getting stuff that's outside of the viewport and instead in the clipper guardband.  I suspect that if you change the viewport size to something larger, everything will behave normally, but in my testing I always used the same viewport.

The scissor offset is then used to choose where rasterized pixels end up in the EFB.  Usually, both `x_off` and `y_off` set to 171, which becomes an offset of 342 when multiplied by 2.  So the pixels that past the scissor test (those with x between 342 and 981 and y between 342 and 821) are offset by 342 into the EFB, ending up with x between 0 and 639 and y between 0 and 479.

The scissor offset step is the thing that can overflow.  Textures on the GameCube/Wii are at max 1024 by 1024, and I guess operations on the EFB wrap around at that point.  Suppose `x_off` was 427, which becomes an offset of 854 = 342 + 512.  Then, a scissor with `x0` of 342 and `x1` of 981 ends up at -512 through 127.  -512 becomes +512, and 512 is less than the EFB's width of 640, so the pixels at the left of the viewport (x=0 without the offset of 342) end up at x=512, and pixels at the right of the viewport (x=639 without the offset of 342) end up at x=127; x=512-639 and x=0-127 end up being drawn to and x=128-511 don't get drawn.

You _can_ also get overflowing with a default `x_off` of 342 if the scissor has `x1` - `x0` > 1024; for instance, in a scissor with `x0` of 342 and `x1` of 2005 (639+342+1024), pixels with x between 342 and 2005 all pass the scissor test and then both x between 342-981 and x between 1366-2005 map to 0-639 on screen.  With the default viewport, one of these is far out of bounds and will usually be clipped (always clipped in Dolphin, since it seems like we clip precisely to the viewport), but with clipping disabled (or presumably a large enough viewport) you can get primitives that draw over themselves, producing [weird results](https://i.imgur.com/6xwybM2.png).

All of this is also made more confusing by the fields for `x0`, `y0`, `x1`, and `y1` being large enough to store `[0, 4095]` despite actually only going for `[0, 2047]` and `x_off`/`y_off` being large enough to store `[0, 2046]` (in jumps of 2). but instead only covering `[0, 1022]` (still in jumps of 2).

Hopefully this is enough to at least understand why things are happening.  I'm not sure how to condense this down to a simpler, more useful explanation.

Pokechu22:
The [difference for pbr-sfx](https://fifo.ci/compare/7082705-7080875/) is new since that fifolog was just added.  Although at first glance this looks like it's a regression, this is actually how the game behaves on real hardware (and also how it behaves in Dolphin's other backends); see [issue 12629](https://bugs.dolphin-emu.org/issues/12629) and [my writeup on the subject](https://gist.github.com/Pokechu22/e9fa9037fe21312ff32475638b78ba27).  PR #10279 includes a game patch that fixes this behavior.

Source/Core/VideoCommon/BPFunctions.h
neobrain: Heh, what a mess. Not sure I can help simplify this any further, but thanks for the explanation :)

JMC47:
This hasn't been looked at in a while and fixes a few games.  Let's see if it breaks any.

PR #10252, opened 2021-11-29
DolphinTool: Add CLI tool subsystem + commands for verifying and converting RVZ/ISO/WIA/GCZ

ssdsnake:
**Preface:**
Still having trouble with Windows. May reach out for help
Evolution of https://github.com/dolphin-emu/dolphin/pull/10246

**Use case:**
Adds a command line "tool" project and shell. Initial commands include (a) ability to verify and output SHA1/MD5/CRC32 digests of RVZ / ISO / NKit  and (b) ability to convert between RVZ/ISO/WIA/GCZ formats; both using the same routines as the GUI verifier widget. 

**Usage and Examples**
Basic Subcommand Shell
```
$ ./dolphin-tool 
usage: tool COMMAND -h

commands supported: [convert, verify]
```

Verify Mode Usage
```
$ ./dolphin-tool verify -h
Usage:  verify [options]...

Options:
  -h, --help            show this help message and exit
  -i FILE, --input=FILE
                        Path to disc image FILE.
  -a ALGORITHM, --algorithm=ALGORITHM
                        Optional. Compute and print the digest using the
                        selected algorithm, then exit. [crc32|md5|sha1]
```

Full Verification
```
$ ./dolphin-tool verify -i /tmp/test.rvz
CRC32: dcd8fa3d
MD5: 03c6f44267aa4a5c42ab20698f88f477
SHA1: f2439bbe1ff64133050fbc00574be8478210a958
Problems Found: No
```

Simple Verification
```
$ ./dolphin-tool verify -i /tmp/test.rvz -a sha1
f2439bbe1ff64133050fbc00574be8478210a958
```

Convert Mode Usage
```
$ ./dolphin-tool convert -h
Usage: convert [options]... [FILE]...

Options:
  -h, --help            show this help message and exit
  -i FILE, --input=FILE
                        Path to disc image FILE.
  -o FILE, --output=FILE
                        Path to the destination FILE.
  -f FORMAT, --format=FORMAT
                        Container format to use. Default is RVZ. [iso|gcz|wia|rvz]
  -s, --scrub           Scrub junk data as part of conversion.
  -b BLOCK_SIZE, --block_size=BLOCK_SIZE
                        Block size for GCZ/WIA/RVZ formats, as an integer. RVZ
                        suggested: 131072 (128KiB)
  -c COMPRESSION, --compression=COMPRESSION
                        Compression method to use when converting to WIA/RVZ.
                        Suggested: Zstandard (zstd). [none|zstd|bzip|lzma|lzma2]
  -l COMPRESSION_LEVEL, --compression_level=COMPRESSION_LEVEL
                        Level of compression for the selected method, if it is
                        not 'none'. 5 is suggested for 'zstd'.)
```

Convert Mode Example
```
$ ./dolphin-tool convert -i /tmp/test.rvz -o /tmp/test_proof.rvz -f rvz -b 131072 -c zstd -l 22
Success

$ sha1sum /tmp/*.rvz
852abeb852867748ba451016523fd28163622c79  /tmp/test_proof.rvz
852abeb852867748ba451016523fd28163622c79  /tmp/test.rvz
```


phire (approved):

phire:
I'm happy with this design. 

My only minor nitpick is that I think` dolphin-tool` is a better command name than `dolphin-emu-tool`. But I'm not going to fight for it.

Source/Core/DolphinTool/ConvertCommand.cpp
```diff
@@ -0,0 +1,276 @@
...
+    return 1;
+  }
+
+  // --scrub
+  const bool scrub = static_cast<bool>(options.get("scrub"));
+
+  if (scrub && format == DiscIO::BlobType::PLAIN)
+    std::cerr << "Warning: Scrubbing does not save space when converting to ISO unless using "
+                 "external compression."
+              << std::endl;
```
JosJuice, line 99:
```suggestion
  if (scrub && format == DiscIO::BlobType::PLAIN)
  {
    std::cerr << "Warning: Scrubbing does not save space when converting to ISO unless using "
                 "external compression."
              << std::endl;
  }
```

Source/Core/DolphinTool/ConvertCommand.cpp
```diff
@@ -0,0 +1,276 @@
...
+
+  if (!scrub && format == DiscIO::BlobType::GCZ &&
+      volume->GetVolumeType() == DiscIO::Platform::WiiDisc && !volume->IsDatelDisc())
+    std::cerr << "Warning: Converting Wii disc images to GCZ without scrubbing may not offer space "
+                 "advantages over ISO."
+              << std::endl;
```
JosJuice, line 105:
```suggestion
  if (!scrub && format == DiscIO::BlobType::GCZ &&
      volume->GetVolumeType() == DiscIO::Platform::WiiDisc && !volume->IsDatelDisc())
  {
    std::cerr << "Warning: Converting Wii disc images to GCZ without scrubbing may not offer space "
                 "advantages over ISO."
              << std::endl;
  }
```

Source/Core/DolphinTool/ConvertCommand.cpp
```diff
@@ -0,0 +1,276 @@
...
+      return good;
+    });
+    break;
+
+  default:
+    ASSERT(false);
+    break;
+  }
+
+  if (!success.get())
```
JosJuice, line 207: If you're going to call `get` on the future immediately, there's no reason to use `std::launch::async`.

Source/Core/DolphinTool/ConvertCommand.cpp
```diff
@@ -0,0 +1,276 @@
...
+{
+  if (format == DiscIO::BlobType::GCZ)
+  {
+    // Refer to ConvertDialog.cpp
+    if (!(file_size % block_size == 0 &&
+          file_size % (block_size * ConvertCommand::GCZ_BLOCKS_PER_BUFFER) != 0))
+    {
+      std::cerr << "Error: Invalid block size for GCZ. File size must be an integer multiple of "
+                   "the block size "
+                   "and must not be an integer multiple of the block size multiplied by 32.";
```
JosJuice, line 228: If the user *explicitly* enters something which old versions of Dolphin have trouble with, I think it makes more sense to just show a warning instead of aborting. The command-line interface is more of a power user interface than DolphinQt after all.

Source/Core/DolphinTool/ConvertCommand.cpp
```diff
@@ -0,0 +1,276 @@
...
+  }
+
+  for (int good_block_size = ConvertCommand::MIN_BLOCK_SIZE;
+       good_block_size <= ConvertCommand::MAX_BLOCK_SIZE; good_block_size *= 2)
+  {
+    if (block_size == good_block_size)
+      return true;
+  }
+
+  return false;
```
JosJuice, line 240: Perhaps consider having just a warning in this case as well, but I care less about this case than the previous case.

JosJuice: Ah, but you would still have to filter out sizes smaller than 32 KiB for RVZ. And if you allow sizes larger than 2 MiB for WIA/RVZ, you should ideally also implement checking if the size is an integer multiple of 2 MiB rather than a power of two... So maybe you should just leave this part as it is.

JosJuice: One thing I did notice just now is that you don't seem to be filtering out sizes smaller than 2 MiB for WIA. Am I missing where it is, or is it not implemented?

Source/Core/DolphinTool/VerifyCommand.cpp
```diff
@@ -0,0 +1,180 @@
...
+}
+
+std::optional<DiscIO::VolumeVerifier::Result>
+VerifyCommand::VerifyVolume(std::shared_ptr<DiscIO::VolumeDisc> volume, bool enable_crc32,
+                            bool enable_md5, bool enable_sha1)
+{
+  if (!volume)
+    return std::nullopt;
+
+  DiscIO::VolumeVerifier verifier(*volume, false, {enable_crc32, enable_md5, enable_sha1});
```
JosJuice, line 150: Do we not want to provide an option for Redump verification?

Source/Core/DolphinTool/VerifyCommand.cpp
```diff
@@ -0,0 +1,180 @@
...
+
+  auto future = std::async(std::launch::async,
```
JosJuice, line 152: Same here regarding `std::launch::async` as for the conversion code.

Source/Core/CMakeLists.txt
```diff
@@ -9,6 +9,7 @@ add_subdirectory(VideoBackends)
 
 if(ENABLE_NOGUI)
   add_subdirectory(DolphinNoGUI)
+  add_subdirectory(DolphinTool)
```
iwubcode, line 12: This seems odd to me, why is this only built if `ENABLE_NOGUI` is set?

Rumi-Larry: Would one need it outside that context?

Pokechu22: The nogui version is fairly niche, with the main use I can think of being fifoci.  It's not built for windows by default and not included in the updater or on the download page.  I could see wanting to build the CLI tool but not the nogui version (and probably, the CLI tool should be distributed for windows, especially since DSPTool currently is).

I suspect part of the reason why the nogui version isn't included by default is that it's ~10 MB, with the main dolphin EXE being ~14 MB (not counting the Qt DLLs); DSPTool is pretty small (~270 KB), and if the CLI tool is small there wouldn't be much of a reason to exclude it.

ssdsnake: Should it maybe get its own flag e.g. ENABLE_TOOL?

Pokechu22: Yeah, I think that a separate flag would be good (though perhaps `ENABLE_CLI_TOOL` would be a clearer name?)

Source/Core/DolphinTool/ConvertCommand.cpp
ssdsnake: My bad good call.

Source/Core/DolphinTool/ConvertCommand.cpp
ssdsnake: Hi, thanks for the tips. Should be much better now. Check IsDiscImageBlockSizeCompatible() in DiscUtils

Source/Core/DolphinTool/VerifyCommand.cpp
ssdsnake: I did not implement this, at least for V1 of this project, for 2 reasons 
1 - Didn't want this to be accidentally used to abuse Redump server in a batch setting
2 - Had issues setting up a temp directory to cache the DAT, ended up being a rabbit hole for now

Pokechu22:
@dolphin-emu-bot rebuild

Pokechu22:
@dolphin-emu-bot  rebuild

Pokechu22:
@dolphin-emu-bot rebuild

Source/Core/DolphinQt/ConvertDialog.cpp
```diff
@@ -257,6 +243,7 @@ void ConvertDialog::OnFormatChanged()
   m_block_size->setEnabled(m_block_size->count() > 1);
   m_compression->setEnabled(m_compression->count() > 1);
 
+  // Block scrubbing of Datel discs when using non-RVZ container
```
JosJuice: This comment is misleading. We block scrubbing when we have a Datel disc **or** when we have an RVZ container.

Source/Core/DolphinTool/ConvertCommand.cpp
```diff
@@ -33,24 +33,24 @@ int ConvertCommand::Main(const std::vector<std::string>& args)
...
       .action("store")
-      .help("Block size for GCZ/WIA/RVZ formats. RVZ Default = 131072 (128KiB)");
+      .help("Block size for GCZ/WIA/RVZ formats, as an integer. RVZ suggested: 131072 (128KiB)");
 
-  parser->add_option("-c", "--compress")
+  parser->add_option("-c", "--compression")
       .type("string")
       .action("store")
-      .help("Compression method to use when converting to WIA/RVZ. Default is Zstandard (zstd). "
+      .help("Compression method to use when converting to WIA/RVZ. Suggested: Zstandard (zstd). "
```
JosJuice:
```suggestion
      .help("Compression method to use when converting to WIA/RVZ. RVZ suggested: Zstandard (zstd)."
```

Since WIA doesn't support zstd.

Source/Core/DolphinTool/ConvertCommand.cpp
```diff
@@ -93,16 +93,26 @@ int ConvertCommand::Main(const std::vector<std::string>& args)
   // --scrub
   const bool scrub = static_cast<bool>(options.get("scrub"));
 
+  if (scrub && format != DiscIO::BlobType::RVZ && volume->IsDatelDisc())
```
JosJuice: Doesn't make sense. Should be:

```suggestion
  if (scrub && (format == DiscIO::BlobType::RVZ || volume->IsDatelDisc()))
```

Or, you could make the RVZ part just a warning. Scrubbing RVZ does work correctly, it's just that it saves almost no space. (Keep the Datel part an error.)

Source/Core/DolphinTool/ConvertCommand.cpp
```diff
@@ -121,35 +131,61 @@ int ConvertCommand::Main(const std::vector<std::string>& args)
...
       std::cerr << "Error: Block size is invalid" << std::endl;
       return 1;
     }
   }
 
+  if (DiscIO::IsGCZBlockSizeLegacyCompatible(block_size_o.value(), volume->GetSize()))
+  {
+    std::cerr << "Warning: For GCZs, the file size must be an integer multiple of "
+                 "the block size "
+                 "and must not be an integer multiple of the block size multiplied by 32.";
```
JosJuice, line 148: This should have a mention of older versions of Dolphin. Right now it sounds like it just won't work at all.

Source/Core/DiscIO/DiscUtils.cpp
```diff
@@ -198,4 +199,39 @@ u64 GetBiggestReferencedOffset(const Volume& volume, const std::vector<Partition
...
+}
+
+bool IsDiscImageBlockSizeCompatible(const int block_size, const DiscIO::BlobType format)
+{
+  switch (format)
+  {
+  case DiscIO::BlobType::WIA:
+    if (block_size < WIA_MIN_BLOCK_SIZE)
+      return false;
+  case DiscIO::BlobType::GCZ:
```
JosJuice, line 218:
```suggestion
  case DiscIO::BlobType::WIA:
    if (block_size < WIA_MIN_BLOCK_SIZE)
      return false;
  [[fallthrough]];
  case DiscIO::BlobType::GCZ:
```

To make the intent clear (and avoid warnings).

Source/Core/DiscIO/DiscUtils.h
```diff
@@ -49,6 +50,24 @@ constexpr u32 WII_NONPARTITION_DISCHEADER_SIZE = 0x100;
 constexpr u32 WII_REGION_DATA_ADDRESS = 0x4E000;
 constexpr u32 WII_REGION_DATA_SIZE = 0x20;
 
+// Default block size for GCZ/RVZ images
+constexpr int GCZRVZ_DEFAULT_BLOCK_SIZE = 0x20000; // 128 KiB
```
JosJuice: I would prefer `GCZ_RVZ` over `GCZRVZ`.

Source/Core/DiscIO/DiscUtils.h
```diff
@@ -49,6 +50,24 @@ constexpr u32 WII_NONPARTITION_DISCHEADER_SIZE = 0x100;
...
+
+// Because DVD timings are emulated as if we can't read less than an entire ECC block at once
+// (32 KiB - 0x8000), there is little reason to use a block size smaller than that.
+constexpr int GCZRVZ_MIN_BLOCK_SIZE = 0x8000; // 32 KiB
+
+// For performance reasons, blocks shouldn't be too large.
+// 2 MiB (0x200000) was picked because it is the smallest block size supported by WIA.
+constexpr int GCZRVZ_MAX_BLOCK_SIZE = 0x200000; // 2 MiB
```
JosJuice: This applies to WIA in the exact same way as it applies to GCZ and RVZ, so the name should be changed.

Perhaps consider naming it something like `RECOMMENDED_MAX_BLOCK_SIZE`, since larger block sizes do in fact work fine, to better contrast it with `WIA_MIN_BLOCK_SIZE` (which in fact is a strict limit).

Source/Core/DolphinTool/ConvertCommand.cpp
```diff
@@ -121,35 +131,61 @@ int ConvertCommand::Main(const std::vector<std::string>& args)
...
 
-    if (!CheckBlockSize(block_size_o.value(), volume->GetSize(), format))
+    if (!DiscIO::IsDiscImageBlockSizeCompatible(block_size_o.value(), format))
     {
       std::cerr << "Error: Block size is invalid" << std::endl;
       return 1;
     }
   }
 
+  if (DiscIO::IsGCZBlockSizeLegacyCompatible(block_size_o.value(), volume->GetSize()))
```
JosJuice, line 144:
```suggestion
  if (!DiscIO::IsGCZBlockSizeLegacyCompatible(block_size_o.value(), volume->GetSize()))
```

Source/Core/DiscIO/DiscUtils.h
```diff
@@ -59,4 +78,7 @@ std::optional<u64> GetFSTSize(const Volume& volume, const Partition& partition);
 
 u64 GetBiggestReferencedOffset(const Volume& volume);
 u64 GetBiggestReferencedOffset(const Volume& volume, const std::vector<Partition>& partitions);
+
+bool IsGCZBlockSizeLegacyCompatible(const int block_size, const u64 file_size);
+bool IsDiscImageBlockSizeCompatible(const int block_size, const DiscIO::BlobType format);
```
JosJuice, line 83:
```suggestion
bool IsGCZBlockSizeLegacyCompatible(int block_size, u64 file_size);
bool IsDiscImageBlockSizeCompatible(int block_size, DiscIO::BlobType format);
```

Source/Core/DiscIO/DiscUtils.cpp
```diff
@@ -198,4 +199,39 @@ u64 GetBiggestReferencedOffset(const Volume& volume, const std::vector<Partition
...
+  case DiscIO::BlobType::RVZ:
+  default:
+    if (block_size < GCZRVZ_MIN_BLOCK_SIZE)
+      return false;
```
JosJuice, line 222: This rejects `GCZ_FALLBACK_BLOCK_SIZE` for GCZ (and other extra small or extra large valid block sizes), which doesn't make much sense if this function is supposed to be generally useful (not just useful to one specific caller).

Source/Core/DiscIO/DiscUtils.h
```diff
@@ -49,6 +50,24 @@ constexpr u32 WII_NONPARTITION_DISCHEADER_SIZE = 0x100;
...
+
+// For performance reasons, blocks shouldn't be too large.
+// 2 MiB (0x200000) was picked because it is the smallest block size supported by WIA.
+constexpr int GCZRVZ_MAX_BLOCK_SIZE = 0x200000; // 2 MiB
+
+// If we didn't find a good block size, pick the block size which was hardcoded
+// in older versions of Dolphin. That way, at least we're not worse than older versions.
+constexpr int GCZ_FALLBACK_BLOCK_SIZE = 0x4000;
+
+// This is the smallest block size supported by WIA. For performance, larger sizes are avoided.
```
JosJuice, line 68: The `For performance, larger sizes are avoided.` part of the comment isn't especially relevant in this context. Probably best to remove it. We already have a comment further up about how "For performance reasons, blocks shouldn't be too large".

CMakeLists.txt
```diff
@@ -32,6 +32,10 @@ if(NOT WIN32 AND NOT APPLE AND NOT HAIKU)
   option(ENABLE_EGL "Enables EGL OpenGL Interface" ON)
 endif()
 
+if(NOT ANDROID)
+  option(ENABLE_CLI_TOOL "Enable dolphin-tool, a CLI tool for power users" ON)
```
JosJuice, line 36: Specifying "for power users" doesn't add anything useful in my opinion.

Maybe "Enable dolphin-tool, a CLI tool for managing disc images"?

Source/Core/DiscIO/DiscUtils.h
```diff
@@ -49,6 +50,24 @@ constexpr u32 WII_NONPARTITION_DISCHEADER_SIZE = 0x100;
 constexpr u32 WII_REGION_DATA_ADDRESS = 0x4E000;
 constexpr u32 WII_REGION_DATA_SIZE = 0x20;
 
+// Default block size for GCZ/RVZ images
+constexpr int GCZRVZ_DEFAULT_BLOCK_SIZE = 0x20000;  // 128 KiB
```
JosJuice, line 54: I would prefer `GCZ_RVZ` over `GCZRVZ`.

Source/Core/DiscIO/DiscUtils.h
```diff
@@ -49,6 +50,24 @@ constexpr u32 WII_NONPARTITION_DISCHEADER_SIZE = 0x100;
...
+
+// Because DVD timings are emulated as if we can't read less than an entire ECC block at once
+// (32 KiB - 0x8000), there is little reason to use a block size smaller than that.
+constexpr int GCZRVZ_MIN_BLOCK_SIZE = 0x8000;  // 32 KiB
+
+// For performance reasons, blocks shouldn't be too large.
+// 2 MiB (0x200000) was picked because it is the smallest block size supported by WIA.
+constexpr int GCZRVZ_MAX_BLOCK_SIZE = 0x200000;  // 2 MiB
```
JosJuice, line 62: This applies to WIA in the exact same way as it applies to GCZ and RVZ, so the name should be changed.

Perhaps consider naming it something like `RECOMMENDED_MAX_BLOCK_SIZE`, since larger block sizes do in fact work fine, to better contrast it with `WIA_MIN_BLOCK_SIZE` (which in fact is a strict limit).

Source/Core/DolphinTool/ConvertCommand.cpp
```diff
@@ -0,0 +1,281 @@
...
+  }
+
+  // --scrub
+  const bool scrub = static_cast<bool>(options.get("scrub"));
+
+  if (scrub && (format == DiscIO::BlobType::RVZ || volume->IsDatelDisc()))
+  {
+    std::cerr << "Error: Scrubbing an RVZ container or Datel disc is not supported";
+    return 1;
+  }
```
JosJuice, line 103: Consider making the RVZ part just a warning. Scrubbing RVZ does work correctly, it's just that it saves almost no space. (Scrubbing Datel, on the other hand, most certainly doesn't work and should remain an error.)

Source/Core/DolphinTool/ConvertCommand.cpp
```diff
@@ -0,0 +1,281 @@
...
+      std::cerr << "Error: Compression format must be set for WIA or RVZ" << std::endl;
+      return 1;
+    }
+
+    if (compression_o.value() == DiscIO::WIARVZCompressionType::Zstd &&
+        format == DiscIO::BlobType::WIA)
+    {
+      std::cerr << "Error: WIA does not support Zstd" << std::endl;
+      return 1;
+    }
```
JosJuice, line 172: `ParseCompressionTypeString` can return `DiscIO::WIARVZCompressionType::Purge`. You'll need to reject `DiscIO::WIARVZCompressionType::Purge` for RVZ.

Source/Core/DolphinTool/ConvertCommand.cpp
```diff
@@ -0,0 +1,281 @@
...
+    break;
+
+  default:
+    ASSERT(false);
+    break;
+  }
+
+  if (success)
+  {
+    std::cout << "Success" << std::endl;
```
JosJuice, line 240: It's customary for CLI tools to not print anything on success, so unless a successful run results in any messages being printed other than this one, perhaps we should remove this message?

Source/Core/DolphinTool/VerifyCommand.cpp
JosJuice: Good point. Let's leave it as it is for now.

Source/Core/DolphinTool/ConvertCommand.cpp
ssdsnake: good eye

Pokechu22:
@dolphin-emu-bot rebuild

Source/Core/DiscIO/DiscUtils.h
```diff
@@ -49,6 +49,24 @@ constexpr u32 WII_NONPARTITION_DISCHEADER_SIZE = 0x100;
 constexpr u32 WII_REGION_DATA_ADDRESS = 0x4E000;
 constexpr u32 WII_REGION_DATA_SIZE = 0x20;
 
+// Default block size for GCZ/WIA/RVZ images
```
JosJuice, line 52: This size is impossible to use for WIA, so WIA shouldn't be mentioned here.

Source/Core/DiscIO/DiscUtils.h
```diff
@@ -49,6 +49,24 @@ constexpr u32 WII_NONPARTITION_DISCHEADER_SIZE = 0x100;
...
+constexpr int PREFERRED_BLOCK_SIZE = 0x20000;  // 128 KiB
+
+// Because DVD timings are emulated as if we can't read less than an entire ECC block at once
+// (32 KiB - 0x8000), there is little reason to use a block size smaller than that.
+constexpr int PREFERRED_MIN_BLOCK_SIZE = 0x8000;  // 32 KiB
```
JosJuice, line 57: "32 KiB" is mentioned twice, which is a little unnecessary.

(Also, I'm realizing now that the `-` looks confusingly much like a minus... But since it's from existing code, I'm not going to require you to fix it.)

Source/Core/DiscIO/DiscUtils.h
```diff
@@ -49,6 +49,24 @@ constexpr u32 WII_NONPARTITION_DISCHEADER_SIZE = 0x100;
...
+
+// For performance reasons, blocks shouldn't be too large.
+// 2 MiB (0x200000) was picked because it is the smallest block size supported by WIA.
+constexpr int PREFERRED_MAX_BLOCK_SIZE = 0x200000;  // 2 MiB
```
JosJuice, line 61: Same here about 2 MiB.

Source/Core/DolphinQt/ConvertDialog.cpp
```diff
@@ -156,21 +148,17 @@ void ConvertDialog::OnFormatChanged()
   {
   case DiscIO::BlobType::GCZ:
   {
-    // In order for versions of Dolphin prior to 5.0-11893 to be able to convert a GCZ file
-    // to ISO without messing up the final part of the file in some way, the file size
-    // must be an integer multiple of the block size (fixed in 3aa463c) and must not be
-    // an integer multiple of the block size multiplied by 32 (fixed in 26b21e3).
-
+    // To support legacy versions of dolphin, we have to check the GCZ block size
+    // See DiscIO::IsGCZBlockSizeLegacyCompatible() in DiscUtil.cpp for details
```
JosJuice, line 152: `DiscUtils.cpp`. (Alternatively, if you prefer it, don't mention the filename. `IsGCZBlockSizeLegacyCompatible` is a unique enough function name that it's easy to search for.)

Source/Core/DolphinTool/ConvertCommand.cpp
```diff
@@ -0,0 +1,297 @@
...
+
+  parser->add_option("-s", "--scrub")
+      .action("store_true")
+      .help("Scrub junk data as part of conversion.");
+
+  parser->add_option("-b", "--block_size")
+      .type("int")
+      .action("store")
+      .help("Block size for GCZ/WIA/RVZ formats, as an integer. Suggested value for RVZ: 131072 "
+            "(128KiB)");
```
JosJuice, line 42:
```suggestion
            "(128 KiB)");
```

Source/Core/DolphinTool/ConvertCommand.cpp
```diff
@@ -0,0 +1,297 @@
...
+      volume->GetVolumeType() == DiscIO::Platform::WiiDisc && !volume->IsDatelDisc())
+  {
+    std::cerr << "Warning: Converting Wii disc images to GCZ without scrubbing may not offer space "
+                 "advantages over ISO. Continuing anyway."
+              << std::endl;
+  }
+
+  if (volume->IsNKit())
+    std::cerr << "Warning: Converting an NKit file, output will still be NKit! Continuing anyway."
+              << std::endl;
```
JosJuice, line 128:
```suggestion
  if (volume->IsNKit())
  {
    std::cerr << "Warning: Converting an NKit file, output will still be NKit! Continuing anyway."
              << std::endl;
  }
```

JosJuice (commented):
In general, my old comments have been addressed now. But it seems like there's no longer anything to prevent you from using block sizes that don't work at all. For instance, I'm not sure if GCZ supports block sizes that aren't a power of two (maybe it does? Better check), and RVZ doesn't support block sizes smaller than 2 MiB that aren't a power of two, doesn't support block sizes smaller than 32 KiB at all, and (just like WIA) requires block sizes larger than 2 MiB to be an integer multiple of 2 MiB. This is particularly important to get right, because we never want to end up in a situation where it *looks* like Dolphin converted a file successfully but the file is actually unreadable.

docs/WiaAndRvz.md
```diff
@@ -245,3 +245,23 @@ u32* buffer_ptr;
...
 buffer_ptr++;
 ```
+
+## Block Sizes
+
+When archiving to a disc image, it is important to choose block sizes wisely.
+
+### Requirements
+
+- GCZ: Block size "must" be a power of 2 (It has not been verified that things break if it isn't! But using a power of 2 is at the very least recommended)
```
JosJuice: Details about GCZ shouldn't be in this file.

docs/WiaAndRvz.md
```diff
@@ -245,3 +245,23 @@ u32* buffer_ptr;
...
+
+- WIA: Block size must not be smaller than `2 MiB`. Block size must be an integer multiple of `2 MiB`.
```
JosJuice: This is already written earlier in the file (line 45).

docs/WiaAndRvz.md
```diff
@@ -245,3 +245,23 @@ u32* buffer_ptr;
...
+
+- RVZ: Block size must not be smaller than `32 KiB`. Block sizes smaller than `2 MiB` must be a power of 2. Block sizes larger than `2 MiB` must be an integer multiple of `2 MiB`.
```
JosJuice: This too is already written earlier in the file (line 178-179).

docs/WiaAndRvz.md
```diff
@@ -245,3 +245,23 @@ u32* buffer_ptr;
...
+
+### Recommendations
+
+- Block sizes smaller than `32 KiB` are not recommended (performance doesn't increase anymore after that point)
```
JosJuice: This isn't useful to note for RVZ and WIA, since they don't support sizes smaller than that anyway.

docs/WiaAndRvz.md
```diff
@@ -245,3 +245,23 @@ u32* buffer_ptr;
...
+
+- Block sizes larger than `2 MiB` are not recommended (performance gets worse the larger the block size is)
```
JosJuice: This threshold is subjective. I would prefer to not include it in a document that's supposed to be more about the format itself than Dolphin's implementation of it.

docs/WiaAndRvz.md
```diff
@@ -245,3 +245,23 @@ u32* buffer_ptr;
...
+
+- GCZ block sizes which are not compatible with old Dolphin versions are not recommended (if you don't care about old Dolphin versions, consider RVZ)
```
JosJuice: Details about GCZ shouldn't be in this file.

Source/Core/DiscIO/DiscUtils.cpp
```diff
@@ -208,16 +209,40 @@ bool IsGCZBlockSizeLegacyCompatible(int block_size, u64 file_size)
...
-    if (block_size == good_block_size)
-      return true;
+  case DiscIO::BlobType::GCZ:
+    // Block size "must" be a power of 2
+    if (!MathUtil::IsPow2(block_size))
+      return false;
+
+    break;
+  case DiscIO::BlobType::WIA:
+    // Block size must be less than the minimum and a multiple of it
```
JosJuice, line 223: You're missing a "not" here.

Source/Core/DiscIO/DiscUtils.cpp
```diff
@@ -208,16 +209,40 @@ bool IsGCZBlockSizeLegacyCompatible(int block_size, u64 file_size)
...
+    if (block_size < WIA_MIN_BLOCK_SIZE || block_size % WIA_MIN_BLOCK_SIZE != 0)
+      return false;
+
+    break;
+  case DiscIO::BlobType::RVZ:
+    // Block size must not be smaller than the minimum
+    // Block sizes smaller than the large block size threshold must be a power of 2
+    // Block sizes larger than that threshold must be a multiple of a certain value
```
JosJuice, line 231:
```suggestion
    // Block sizes larger than that threshold must be a multiple of the threshold
```

Source/Core/DiscIO/DiscUtils.h
```diff
@@ -49,24 +50,32 @@ constexpr u32 WII_NONPARTITION_DISCHEADER_SIZE = 0x100;
...
-// This is the smallest block size supported by WIA.
+// 2 MiB (0x200000) is the smallest block size supported by WIA.
 constexpr int WIA_MIN_BLOCK_SIZE = 0x200000;
 
+// 32 KiB (0x8000) is the smallest block size supported by RVZ.
+constexpr int RVZ_MIN_BLOCK_SIZE = 0x8000;
+
+// 2 MiB (0x200000): for RVZ, block sizes larger than 2 MiB must be an integer multiple of 2 MiB.
+constexpr int RVZ_BIG_BLOCK_SIZE_THRESHOLD = 0x200000;
+constexpr int RVZ_BIG_BLOCK_SIZE_LCM = 0x200000;
```
JosJuice, line 77: These two constants should probably just be one constant.

Source/Core/DiscIO/DiscUtils.h
```diff
@@ -49,24 +50,32 @@ constexpr u32 WII_NONPARTITION_DISCHEADER_SIZE = 0x100;
...
+// an entire ECC block at once. Therefore, little reason to choose a smaller block size.
+constexpr int PREFERRED_MIN_BLOCK_SIZE = 0x8000;
 
-// For performance reasons, blocks shouldn't be too large.
 // 2 MiB (0x200000) was picked because it is the smallest block size supported by WIA.
-constexpr int PREFERRED_MAX_BLOCK_SIZE = 0x200000;  // 2 MiB
+// For performance reasons, blocks shouldn't be too large.
+constexpr int PREFERRED_MAX_BLOCK_SIZE = 0x200000;
 
+// 16 KiB (0x4000) for supporting GCZs in versions of Dolphin prior to 5.0-11893
```
JosJuice, line 64: I think it can be a little misleading to state this before you get to the "if we didn't find a good block size" qualifier. Most of the time it's fine to use at least 32 KiB when you want compatibility with older versions of Dolphin.

Source/Core/DolphinQt/ConvertDialog.cpp
```diff
@@ -149,7 +149,7 @@ void ConvertDialog::OnFormatChanged()
   case DiscIO::BlobType::GCZ:
   {
     // To support legacy versions of dolphin, we have to check the GCZ block size
-    // See DiscIO::IsGCZBlockSizeLegacyCompatible() in DiscUtil.cpp for details
+    // See IsGCZBlockSizeLegacyCompatible() for details
```
JosJuice, line 152: Nitpick (feel free to ignore): The `DiscIO::` prefix does make sense to include.

Source/Core/DiscIO/DiscUtils.h
```diff
@@ -61,9 +61,9 @@ constexpr int PREFERRED_MIN_BLOCK_SIZE = 0x8000;
 // For performance reasons, blocks shouldn't be too large.
 constexpr int PREFERRED_MAX_BLOCK_SIZE = 0x200000;
 
+// If we didn't find a good GCZ block size, pick the block size which was hardcoded
+// in legacy versions. That way, at least we ensure compatibility.
```
JosJuice, line 65: This does not ensure compatibility. (Yes, old versions of Dolphin could create files which couldn't be correctly read by itself.) The "not worse than older versions" wording was chosen for a reason.

JosJuice:
LGTM if you squash the commits.

JosJuice:
@dolphin-emu-bot rebuild

JosJuice:
@dolphin-emu-bot rebuild

JosJuice (approved):

shuffle2:
is there a reason to have this as a separate binary instead of adding the commandline options to a common location? It seems both DolphinQt and DolphinNoGUI could house this functionality with no need for separate tool...?

ssdsnake:
They serve different purposes and bring up larger environments. I didn't want to clutter those use cases or risk breaking CLI workflows for a poweruser tool.

TomTurbine:
Hate to necro this but seems the most appropriate place to ask this.

When it comes to verifying your games, it is possible to do so against the ReDump datafiles without having to convert them from RVZ back to ISO? 

And if possible, any way to include this feature in a future release of the Dolphin tool? Would gladly use another program to edit a dat file to what I need and then load it into your software if it allowed users to audit their collections against it.

Pokechu22:
For RVZ, yes, you can verify without converting (it internally converts back while verifying, the same way it internally converts back when the game tries to read data for the disc). For other formats I'm not sure whether the same guarantee applies or not (it'd depend on whether it was scrubbed for instance).

TomTurbine:
I understand it can verify and give you the check sum values, but I can't find a rom manager that supports using it except 1 which seemed to give me issues using it.

Was asking if there would be any way to have it verify the games against a dat file like from ReDump or potentially build in support for doing so in the future?

Basically a process of load in the dat file that you either just got or customized yourself to match your games, load in your games and have it run them to verify them. Will show you any that are missing or corrupted.

mbc07:
Dolphin can verify any supported ISO format against the Redump DATs, but on the GUI you can only do this manually, one by one. Exposing this feature via DolphinTool would likely allow batch verifying many ISOs with some basic CLI automation, but AFAICT that's not possible at the moment...

TomTurbine:
Thank you for the reply.

Maybe this could be a good future feature to add but given all that just happened with steam, I am figuring this will be low on the priority lists if ever on them.

Would be nice to just load a dat file and have it check them for you. I found another tool named "OxyMoron" or something like that that claims to support it but I just couldn't seem to get it going and the other program I found extracted it to a hard drive in ISO then scanned it and then deleted it which can seem excessive to audit.

AdmiralCurtiss:
You should be able to get the hash out of DolphinTool and then forward/pipe that into something that checks it against the dat, eg.

```
>dolphintool verify -i "e:\Game Images\Wii\RPBP01.rvz" -a sha1
8ae47b225b154f11d29e8e39aceb2f037739f430
```

If you need something more fancy than that then you can always open a PR or add a feature request on our bug tracker.

alucryd:
@TomTurbine oxyromon author here, RVZ should be working fine (it does here), feel free to open an issue if you're having issues with it.

TomTurbine:
Thank you Alucryd, I am good for now, but will check back into it again later.

I think my biggest issues I had was mostly the use of command-line. I used to be pretty good at the stuff back in the 90s and early 2000's but that skill set is pretty rusty nowadays.

alucryd:
No problem. I'm slowly making the web interface more than just for visualization, the long term goal is to bring it on par with the CLI.

PR #10253, opened 2021-11-29
InputCommon: fix InputConfig::LoadConfig() not always replacing emu controllers values

Filoppi:
If InputConfig::LoadConfig() was called once with a non empty/customized config,
then called again after manually deleting the config (dolphin calls LoadConfig() every time it opens the mapping widget),
the second load would fail to clear the values on any non first EmulatedController and would instead keep the
previous config values despite it being deleted (while it would instead correctly default the first EmulatedController).

This is not a big bug though the code is better now.


Source/Core/InputCommon/InputConfig.cpp
```diff
@@ -150,22 +150,18 @@ bool InputConfig::LoadConfig(InputClass type)
   }
   else
   {
-    // Only load the default profile for the first controller,
-    // otherwise they would all share the same mappings and default device
+    // Only load the default profile for the first controller and clear the others,
+    // otherwise they would all share the same mappings on the same and default device
```
JosJuice, line 154: The addition of "on the same" doesn't make sense to me.

Filoppi: On the same device, which is also the default device.

JosJuice: I see what you mean then... But I think it just makes the sentence more confusing.

Rumi-Larry: It could read "with the same default device".

Filoppi: Addressed

PR #10254, opened 2021-11-30
TASInputWindow: Shrink width for stick input slider value boxes

malleoz:
The value boxes for TAS Input stick sliders was unnecessarily wide, which would cause the aspect_ratio widget containing the stick to be severely restricted in size... Thus, this change ends up making the stick a lot bigger.

image


JosJuice:
It would be good to test this with some different OSes/fonts to see if 40 really is big enough. I agree with making it smaller than 60 regardless, though.

leoetlino (changes requested):
The text boxes are too *narrow* on Linux with this PR.

image

Maybe consider using layout stretching for the stick widget instead of setting a maximum, hardcoded width for the textbox.

PR #1812, opened 2015-01-02
zfreeze

phire:
This has been a long journey but zfreeze is finished.
This PR contains code from neobrain, NanoByte011 and phire.

Changes since #1767 
 * Don't cull CULLALL primitives so early because they are often used as reference planes.
 * Convert CalculateZSlope to screenspace coordinates.
 * Convert Pixelshader to screenspace coordinates (instead of worldspace  xy coordinates, which is totally wrong)
 * Divide depth by 2^24 instead of clamping to 0.0-1.0 as was done before.
 * Adjust from DirectX/OpenGL screenspace coordinates to gamecube screenspace coordinates, by compensating for EFB scale. (and flipping the vertical axis in the case of OpenGL) Zfreeze now works in most games.
 * Make sure early depth isn't forced on when zfreeze is enabled, fixing fast depth and gimmick courts in Mario Power Tennis **(NanoByte)**
 * Improved shader constant uniform buffer savestate code.
 * Move Common code into VideoCommon
 * OpenGL: don't send CULLALL vertexes to GPU Memory.

## Known Issues:
 - [x] The Delfino Plaza court in Mario Power Tennis has zfighting issues. These may be unfixable.
 - [ ] It should still be possible to generate test cases with visible reference triangles which aren't rendered correctly with this approach. Fortunately, all known retail games use invisible reference planes.

## Todo:
 - [x] Don't update the constants uniform buffer before every drawcall.
 - [x] Cleanup CULLALL mode code path.
 - [x] Make sure this isn't broken on anyone's video card.
 - [x] Savestate Issues
 - [x] Handle per-vertex matrix indexes for the reference plane, or leave a comment justifying that we don't need to.

PR #10255, opened 2021-11-30
Software: Fix zfreeze with CullMode::All

Pokechu22:
CullMode::All is often used with zfreeze to draw a triangle for the reference plane without it showing up (see the [note on `GX_SetCoPlanar`](https://github.com/devkitPro/libogc/blob/70323c888069f6bdd032b2e0b644a5cbefbfe179/gc/ogc/gx.h#L2904-L2905)).  However, this was broken in the software renderer.  I've fixed it by re-using the infrastructure in VertexManagerBase used by the hardware backends, but this is a bit jank.

### DIFF
--- a/Source/Core/VideoBackends/Software/Rasterizer.cpp
+++ b/Source/Core/VideoBackends/Software/Rasterizer.cpp
@@ -20,15 +20,81 @@ namespace Rasterizer
 {
 static constexpr int BLOCK_SIZE = 2;
 
+struct SlopeContext
+{
+  SlopeContext(const OutputVertexData* v0, const OutputVertexData* v1, const OutputVertexData* v2,
+               s32 x0, s32 y0)
+      : x0(x0), y0(y0)
+  {
+    // adjust a little less than 0.5
+    const float adjust = 0.495f;
+
+    xOff = ((float)x0 - v0->screenPosition.x) + adjust;
+    yOff = ((float)y0 - v0->screenPosition.y) + adjust;
+
+    dx10 = v1->screenPosition.x - v0->screenPosition.x;
+    dx20 = v2->screenPosition.x - v0->screenPosition.x;
+    dy10 = v1->screenPosition.y - v0->screenPosition.y;
+    dy20 = v2->screenPosition.y - v0->screenPosition.y;
+  }
+  s32 x0;
+  s32 y0;
+  float xOff;
+  float yOff;
+  float dx10;
+  float dx20;
+  float dy10;
+  float dy20;
+};
+
+struct Slope
+{
+  Slope() = default;
+  Slope(float f0, float f1, float f2, const SlopeContext& ctx) : f0(f0)
+  {
+    float delta_20 = f2 - f0;
+    float delta_10 = f1 - f0;
+
+    //        x2 - x0    y1 - y0    x1 - x0    y2 - y0
+    float a = delta_20 * ctx.dy10 - delta_10 * ctx.dy20;
+    float b = ctx.dx20 * delta_10 - ctx.dx10 * delta_20;
+    float c = ctx.dx20 * ctx.dy10 - ctx.dx10 * ctx.dy20;
+
+    dfdx = a / c;
+    dfdy = b / c;
+
+    x0 = ctx.x0;
+    y0 = ctx.y0;
+    xOff = ctx.xOff;
+    yOff = ctx.yOff;
+  }
+
+  // These default values are used in the unlikely case that zfreeze is enabled when drawing the
+  // first primitive.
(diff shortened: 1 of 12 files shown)


Pokechu22:
It seems like there was a second issue: the slope used with zfreeze should use offsets from the screen's origin, but instead it was using offsets from the triangle's origin.  This is what caused the triangles in the sky in RS2: half of them had the first vertex at the top of the triangle, and the other half had the first vertex at the bottom of the triangle.

phire:
> but this is a bit jank.

Indeed. A better fix is to disable the code in vertex manager which filters out `CULL_ALL` draw calls and forward them to video software.

Source/Core/VideoBackends/Software/Rasterizer.cpp
```diff
@@ -265,11 +301,46 @@ static void BuildBlock(s32 blockX, s32 blockY)
...
+  }
+
+  for (unsigned int i = 0; i < bpmem.genMode.numtexgens; i++)
+  {
+    for (int comp = 0; comp < 3; comp++)
+    {
+      TexSlopes[i][comp] = Slope(v0->texCoords[i][comp] * w[0], v1->texCoords[i][comp] * w[1],
+                                 v2->texCoords[i][comp] * w[2], ctx);
+    }
+  }
```
phire, line 334: For performance reasons, it might be worth splitting this function in half, one for depth, one for color/uv slopes and then only calculating the depth slope for culled triangles. 

In #10260 I got a speedup for delaying the calculation of color and uv values for each pixel until the depth passed for the quad. I wonder if it might be worth delaying the calculation of these slopes until the first pixel quad passes depth testing.

Pokechu22: I've renamed the function to `UpdateZSlope` and only update the zslope in it now (while the other slopes happen on their own); this does mean that `SlopeContext` is created twice though (one in `UpdateZSlope` and one in `DrawTriangleFrontFace`).

I feel like trying to delay the calculation of the slopes would cause more problems than it would fix, as unlike computing the slope values for each pixel, you only need to calculate the slopes once, so you'd need to check whether the slopes had already been computed for each block, which seems like it'd almost always be slower than computing the slopes once.

Source/Core/VideoBackends/Software/Rasterizer.cpp
```diff
@@ -20,28 +20,92 @@ namespace Rasterizer
...
+    float c = ctx.dx20 * ctx.dy10 - ctx.dx10 * ctx.dy20;
+
+    dfdx = a / c;
+    dfdy = b / c;
+
+    x0 = ctx.x0;
+    y0 = ctx.y0;
+    xOff = ctx.xOff;
+    yOff = ctx.yOff;
+  }
```
Pokechu22, line 70: The equivalent code in the hardware backends looks like this:

https://github.com/dolphin-emu/dolphin/blob/e92e90d1474d028ba63a3e11b5528e30ebc72c85/Source/Core/VideoCommon/VertexManagerBase.cpp#L574-L592

Apart from some of the excessive minus signs and other confusion that also originally existed in `InitSlope`, there's a divide by zero check that doesn't exist in the software renderer.  Does this actually matter?

The check was added in commit add59b3bea032e331363dcf28361f8e0de65e43d, but https://github.com/dolphin-emu/dolphin/pull/1812#discussion_r22443801 says it shouldn't happen with zfreeze specifically.

Source/Core/VideoBackends/Software/Rasterizer.cpp
phire: It's something you would only do if profiling proved that it was worth it.

It's pretty cheap to check if the slopes have been setup for this triangle yet, a single load, a compare and a branch. What's potentially expensive is the extra branch misspredicticts. But branch prediction is really good on modern CPUs, taking branch history into account.... It's really hard to tell without profiling.

Actually, I've just realized such an optimisation will be made irrelevant by my plans to put a multithreadding boundary here, running the depth testing on the main thread to get timing information, then sending the depth test mask along with the triangles to multiple worker threads where the color will be calculated.

Source/Core/VideoBackends/Software/Rasterizer.cpp
phire: I think my comment is saying this based on the fact that games that use zfreeze should be sane enough to use non-degenerate triangles for their z reference planes.

No idea what happens on real hardware. These are zero-sized triangles that won't even get rasertized, right? Which is why they don't have a valid slope. It's entirely possible they have already been discarded by this point in the pipeline.

phire (approved):

PR #10256, opened 2021-11-30
Renderbase: Show rerecord count

malleoz:
This PR adds a "Show Rerecord Count" toggle under the Movie menu. When enabled, this causes the RenderBase to show `s_rerecords` in the Movie window.

Note that the rerecord count will show 0 if a movie is not being recorded or playing back. I'm curious if anyone has any thoughts about if this should be hidden? Perhaps check for if `IsMovieActive()`? Or perhaps if the rerecord count doesn't update, it's a healthy reminder to TASers that they aren't actively recording a movie (I sometimes accidentally forget to check if I am in read-only mode or not).

### DIFF
--- a/Source/Core/Core/Movie.cpp
+++ b/Source/Core/Core/Movie.cpp
@@ -202,6 +202,15 @@ std::string GetRTCDisplay()
   return format_time.str();
 }
 
+// NOTE: GPU Thread
+std::string GetRerecords()
+{
+  if (IsMovieActive())
+    return fmt::format("Rerecords: {}", s_rerecords);
+
+  return "Rerecords: N/A";
+}
+
 void FrameUpdate()
 {
   s_currentFrame++;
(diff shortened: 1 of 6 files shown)


Source/Core/Core/Movie.cpp
```diff
@@ -202,6 +202,12 @@ std::string GetRTCDisplay()
   return format_time.str();
 }
 
+// NOTE: GPU Thread
+std::string GetRerecords()
+{
+  return std::format("Rerecords: {}", s_rerecords);
```
Pokechu22, line 208: Dolphin targets C++17 currently, where `std::format` doesn't exist; however, `std::format` is just a standardization of the fmt library, which is used by dolphin.  So `return fmt::format("Rerecords: {}", s_rerecords);` should work fine.

The visual studio solution doesn't restrict it to C++17, which is probably why you didn't run into this in development.

malleoz: Ah thank you, I didn't realize this. I will keep this in mind for the future.

Pokechu22 (approved):
Code looks good to me.  I think it's fine to keep it visible when a movie isn't active, though perhaps "Rerecords: N/A" would also work.  Maybe the read-only status would also be useful to include in the movie window, though?

malleoz:
> Code looks good to me. I think it's fine to keep it visible when a movie isn't active, though perhaps "Rerecords: N/A" would also work. Maybe the read-only status would also be useful to include in the movie window, though?

I've decided to check `IsMovieActive()` within `Movie::GetRerecords()` to show N/A if no movie is currently active. This should suffice for now. To show the status of read-only can be its own PR in the future I figure.

phire (approved):

PR #10257, opened 2021-11-30
BS2Emu: Stop clearing the RTC flags

Pokechu22:
The system menu does clear the RTC flags, but we currently aren't updating the cache file, and since we clear them the system menu doesn't know to update the cache either.  This means that launching a game via the system menu, and then launching a game directly and exiting via HOME will result in the system menu using an outdated cache and displaying the old game.  This causes it to fail to launch the game on the disc channel (since it doesn't match the cache), resulting in it resetting (though it will ignore the cache after resetting).  Not clearing the cache avoids this issue.

### DIFF
--- a/Source/Core/Core/Boot/Boot_BS2Emu.cpp
+++ b/Source/Core/Core/Boot/Boot_BS2Emu.cpp
@@ -461,8 +461,13 @@ bool CBoot::EmulatedBS2_Wii(const DiscIO::VolumeDisc& volume,
     state->discstate = 0x01;
   });
 
-  // The system menu clears the RTC flags
-  ExpansionInterface::g_rtc_flags.m_hex = 0;
+  // The system menu clears the RTC flags.
+  // However, the system menu also updates the disc cache when this happens; see
+  // https://wiibrew.org/wiki/MX23L4005#DI and
+  // https://wiibrew.org/wiki//title/00000001/00000002/data/cache.dat for details. If we clear the
+  // RTC flags, then the system menu thinks the disc cache is up to date, and will show the wrong
+  // disc in the disc channel (and reboot the first time the disc is opened)
+  // ExpansionInterface::g_rtc_flags.m_hex = 0;
 
   // While reading a disc, the system menu reads the first partition table
   // (0x20 bytes from 0x00040020) and stores a pointer to the data partition entry.


AdmiralCurtiss:
Oh that's an extremely long-standing issue, good find.

Pokechu22:
It's worth noting that the issue was _partially_ fixed by #8394 - that implemented the RTC flag system, and set it when discs were changed.  After it was merged, using "change disc" while a game was running (or while a channel was running, more commonly) would correctly result in the system menu clearing the cache.  But I guess I never thought of the more common case of launching a game directly and then never touching "change disc" but instead going directly to the wii menu.

phire:
What's the downside to not clearing the flags? Does anything else break?

Pokechu22:
The downside is that the system menu will _always_ think that the disc changed, even if you used the same disc.  But that is more due to how dolphin sets the flag in the first place; it doesn't keep track of what disc is inserted across runs, so each time it's as if the disc was inserted right before emulation started (even if using the default disc function), which sets the flag.  

Clearing the flag and updating the cache file would be slightly preferable (but somewhat complicated), while not clearing the flag and not updating the cache (which is what this PR does) is fine too (at least if I recall correctly).  Clearing the flag without updating the cache (the old behavior) is just a bad idea and not something that can happen on console.

phire:
That's pretty small as far as downsides go.

JMC47:
Yeah, that seems fairly safe.

phire (approved):
That's pretty small as

Pokechu22:
I guess it's worth also adding that in either case, with how dolphin currently works, launching the system menu directly won't use the disc cache (since the flags are still set in that case, and are always set due to the disc being immediately inserted on startup).

If you launch a game via the system menu, and then exit back to the system menu without changing discs, the cache will still be used properly.  And the cache will also be used properly after this PR if you launch a game directly, then go to the system menu (which will not use the cache), then launch the game from the system menu, and then go back to the system menu again (which will now use the cache).  On a real console, you generally can't launch a game without going through the system menu, but if you were to do so you would run into the same behavior of the cache not being updated and the flag not being cleared.

Pokechu22:
Oh, I did forget one other thing: games actually do look at the RTC flags, but only in [OSShutdownSystem](https://gist.github.com/Pokechu22/fac0b3ef6a26724c0065784db9ed86a1)/OSRebootSystem/OSReturnToMenu, and only to update [state.dat](https://wiibrew.org/wiki//title/00000001/00000002/data/state.dat).  The use in games is the reason why I know they are called the RTC flags: debug symbols give the names `__OSGetRTCFlags` and `__OSClearRTCFlags`.  That said, this shouldn't make any difference for the system menu.

One other oddity is that if the eject flag is set when the system menu launches, it will eject the disc and then shut down.  This normally happens if the console is turned on by pressing the eject button (though bootmii as boot2 can make it a bit weirder).  However, the code in OSReturnToMenu means that this flag is cleared when exiting games, so there's no problem here.

PR #10258, opened 2021-12-02
FifoPlayer: Clear EFB before starting playback

Pokechu22:
This fixes the bad rendering on the first frame when using the software renderer: the software renderer's Z buffer started out at 0, but most games clear it to 0xffffff instead; this means that things don't render correctly except for in the regions where the screen was cleared by an EFB copy earlier in the frame.


phire (approved):

phire:
Seems obvious in retrospect. Actually fixes a lot of the "first-frame" issues in fifoci

PR #10259, opened 2021-12-02
SW Rasterizer : Add simple cycle timings.

phire:
More of an estimate than anything accurate, but it's enough to get Speed Challenge working on the software backend.

The timings of this commit:
 * Each rasterized 2x2 quad takes one cycle for per TEV stage.
 * Each triangle takes a minimum of two cycles.
 * Vertices take 4 cycles. For each draw, maximum of the vertices or
   rasterizer cycles will be used.
 * TMEM misses are not counted.
 * Indirect stages are not counted
 * There is no performance advantage to early depth fails


JMC47:
It does work for Speed Challenge, but Rayman Arena gets closer but still freezes ~8 frames into the cutscene on stage 2 instead of before it loads.

Pokechu22:
The fifoci differences still happen when this is rebased atop #10258.  I believe that we determined that some of them were caused by the more frequent flushings for the vertex loader (one for each primitive command), combined with memory updates in (some?) fifologs being located too late (at the last primitive command in an object, instead of the first), meaning that the object doesn't have e.g. texture data until the last primitive command.  This would be an issue that only applies to the fifoplayer and not actual games, but it's something that should be fixed (if possible) before merging this.

Pokechu22:
This change also breaks the software renderer's "dump objects" function, which previously saved a screenshot of the EFB on every batch of draw commands (sometimes resulting in fewer results than the number of objects shown in the FIFO player), but now saves one for every single draw command (resulting in WAY more results than the number of objects shown in the FIFO player).  Perhaps it would be best to get rid of that function and instead put something in the FIFO player itself that saves the images (so that it can also be used with other backends).  (Note that I ran into this myself in #9718 before I removed the "Provide raw primitive data to SetupUnit" commit.)

PR #10260, opened 2021-12-02
SW Rasterizer: Do early depth as quads before UVs and LODs

phire:
This both improves timing accuracy, and improve performance by about 3%


Source/Core/VideoBackends/Software/Rasterizer.cpp
```diff
@@ -71,6 +71,37 @@ void SetTevReg(int reg, int comp, s16 color)
   tev.SetRegColor(reg, comp, color);
 }
 
+// Does the depth test for all four pixels in the block
+// Returns true if at least one pixel passes
+static bool DepthTestBlock(s32 block_x, s32 block_y)
+{
+  int passCount = 0;
```
Pokechu22, line 78: I think this comment would benefit from something explicitly stating that this function also updates `pixel.mask` for all pixels in the block, as otherwise I wondered why it didn't return early as soon as one pixel passed.

Source/Core/VideoBackends/Software/Rasterizer.h
```diff
@@ -29,6 +29,7 @@ struct RasterBlockPixel
 {
   float InvW;
   float Uv[8][2];
+  bool mask;
```
Pokechu22, line 32: I don't like the way that basically all uses of this need to be negated.  I also don't really like the name; it seems slightly ambiguous to me.  Maybe `should_draw` could be used instead (along with getting rid of the negations?)

PR #10261, opened 2021-12-02
SoftwareRasterizer: Timings + Earlier depth

phire:
Draft PR that includes both #10259 and #10260. Both of those should be merged before this one. 

Only difference is that the second commit now has improved timings for 2x2 quads that fail early depth.

### DIFF
--- a/Source/Core/VideoBackends/Software/Rasterizer.cpp
+++ b/Source/Core/VideoBackends/Software/Rasterizer.cpp
@@ -265,8 +265,8 @@ static void BuildBlock(s32 blockX, s32 blockY)
   }
 }
 
-void DrawTriangleFrontFace(const OutputVertexData* v0, const OutputVertexData* v1,
-                           const OutputVertexData* v2)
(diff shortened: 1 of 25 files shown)


Source/Core/VideoBackends/Software/Rasterizer.cpp
```diff
@@ -465,7 +482,40 @@ void DrawTriangleFrontFace(const OutputVertexData* v0, const OutputVertexData* v
...
+
+      // If early depth-testing is enabled, and all pixels in the block fail, we can safely skip
+      // the entire block.
+      if (bpmem.UseEarlyDepthTest())
+      {
+        EfbInterface::IncPerfCounterPixelCount(PQ_ZCOMP_INPUT_ZCOMPLOC);
+        if (!DepthTestBlock(x, y))
+        {
+          continue;
+          early_depth_blocks++;
```
phire, line 494: That's not how c++ works.

PR #10262, opened 2021-12-02
SW Rasterizer: CMPR texture decoding optimisations

phire:
Doesn't have all commits yet, adding them incrementally to get FIFOCI runs


phire:
Looks like I have some errors on FIFOCI, but I'm not entirely sure, nor am I sure what commit they come from.

I'm also considering removing `SW Rast: Bake wrap mode and tmem into SampleMip` It adds a lot of complexity, without making an impact on benchmarks, at least on my computer.

Pokechu22:
> Looks like I have some errors on FIFOCI, but I'm not entirely sure, nor am I sure what commit they come from.

They come from `Decode CMPR texels in quads`.  The assumption that `(s & 3) != 3` means that `s2` is in the same block as `s` is wrong; for instance, `s` might be 4 and `s2` might be 3; this can happen if the original coordinate was -4 and mirror wrapping is in use (ztp-grass is a good test for this, as the tree (object 411) has mirror wrapping on texture 0 and the texture is drawn in a big manner).

Using `(s & ~3) == (s2 & ~3)` (and the same for `t`) seems to fix it:

```patch
diff --git a/Source/Core/VideoCommon/TextureDecoder_Common.cpp b/Source/Core/VideoCommon/TextureDecoder_Common.cpp
index 52ca1c04b9..27e0829fa5 100644
--- a/Source/Core/VideoCommon/TextureDecoder_Common.cpp
+++ b/Source/Core/VideoCommon/TextureDecoder_Common.cpp
@@ -739,7 +739,9 @@ void TexDecoder_DecodeTexelQuad(u8* dst, const u8* src, int s, int t, int s2, in
   case TextureFormat::CMPR:
   {
     int widthBlks = (imageWidth >> 3) + 1;
-    if ((s & 3) != 3 && (t & 3) != 3)
+    bool same_block_s = (s & ~3) == (s2 & ~3);
+    bool same_block_t = (t & ~3) == (t2 & ~3);
+    if (same_block_s && same_block_t)
     {
       // All samples are from the same CMPR block, so we only need a single load and color lookup
       auto data = TexDecoder_LoadCmprTexel(src, s, t, widthBlks);
@@ -749,7 +751,7 @@ void TexDecoder_DecodeTexelQuad(u8* dst, const u8* src, int s, int t, int s2, in
       *((u32*)(dst + 8)) = TexDecoder_CMPRColorSelect(data, colors, s, t2);
       *((u32*)(dst + 12)) = TexDecoder_CMPRColorSelect(data, colors, s2, t2);
     }
-    else if ((s & 3) != 3)
+    else if (same_block_s)
     {
       // There are different blocks on the top and bottom.
       auto data = TexDecoder_LoadCmprTexel(src, s, t, widthBlks);
@@ -761,7 +763,7 @@ void TexDecoder_DecodeTexelQuad(u8* dst, const u8* src, int s, int t, int s2, in
       *((u32*)(dst + 8)) = TexDecoder_CMPRColorSelect(data, colors, s, t2);
       *((u32*)(dst + 12)) = TexDecoder_CMPRColorSelect(data, colors, s2, t2);
     }
-    else if ((t & 3) != 3)
+    else if (same_block_t)
     {
       // There are different blocks on the left and right.
       auto data = TexDecoder_LoadCmprTexel(src, s, t, widthBlks);
```

> I'm also considering removing `SW Rast: Bake wrap mode and tmem into SampleMip`

I agree, it's probably worth removing that one given the lack of performance improvements and the overall decreased readability and increased complexity.

phire:
Rebased on current master, removed `SW Rast: Bake wrap mode and tmem into SampleMip` and applied Pokechu22, suggested patch.

Source/Core/VideoCommon/TextureDecoder_Common.cpp
```diff
@@ -546,61 +566,48 @@ static u32 TexDecoder_DecodePixelCMPR(cmprTexelData data, int s, int t)
...
+
+  // It's quite expensive to derive the four colors of each block.
+  // So we cache then, indexed by the two colors.
+
+  u32 cacheSet = cmprCacheSet(data.color1, data.color2);
+  u32 cacheTag = cmprCacheTag(data.color1, data.color2);
+
+  // our cache is initialized to all zeros, and we don't have space for a valid bit.
+  // This means tag 0 will end up as valid.
+  // A simple solution is to just never hit on tag zero
```
phire, line 597: An alternative solution which I didn't implement was to simply pre-populate the cache with the correct colours for tag zero 

Tag zero is a nonsense block color anyway, with three solid blacks plus a solid black with 0% alpha

Source/Core/VideoBackends/Software/TextureSampler.cpp
```diff
@@ -112,64 +112,109 @@ void Sample(s32 s, s32 t, s32 lod, bool linear, u8 texmap, u8* sample)
...
   const TexMode0& tm0 = texUnit.texMode0;
   const TexImage0& ti0 = texUnit.texImage0;
   const TexTLUT& texTlut = texUnit.texTlut;
-  const TextureFormat texfmt = ti0.format;
-  const TLUTFormat tlutfmt = texTlut.tlut_format;
+  state.texfmt = ti0.format;
+  state.tlutfmt = texTlut.tlut_format;
 
-  const u8* imageSrc;
-  const u8* imageSrcOdd = nullptr;
```
Pokechu22, line 126: `state.pointer_odd` should probably be set to `nullptr` here just to be safe, as otherwise it doesn't get updated from the previous value.

Source/Core/VideoBackends/Software/TextureSampler.cpp
```diff
@@ -112,64 +112,109 @@ void Sample(s32 s, s32 t, s32 lod, bool linear, u8 texmap, u8* sample)
...
+  // move texture pointer to mip location
+  if (mip)
+  {
+    image_width_minus_1 >>= mip;
+    image_height_minus_1 >>= mip;
+    s >>= mip;
+    t >>= mip;
+
+    imageSrc += state.mipOffsets[std::min(maxMips, mip) - 1];
+  }
```
Pokechu22, line 217: Theoretically you could remove the `if` statement here since shifting by 0 does nothing, although you'd need to make `mipOffsets[0]` be 0 (possibly that's why the size is `maxMips + 1` (see earlier comment)?).  I don't know if that'd make much of a difference though.

Source/Core/VideoCommon/TextureDecoder_Common.cpp
```diff
@@ -434,55 +435,206 @@ static inline u32 DecodePixel_Paletted(u16 pixel, TLUTFormat tlutfmt)
...
+  u32 offset = (base + blkOff) << 1;
+  return *(u16*)(src + offset);
+}
+
+static u32 TexDecoder_LoadTexel32(const u8* src, int s, int t, int imageWidth)
+{
+  u16 sBlk = s >> 2;
+  u16 tBlk = t >> 2;
+  u16 widthBlks = (imageWidth >> 2) + 1;
+  u32 base = (tBlk * widthBlks + sBlk) << 5;  // shift by 5 is correct
```
Pokechu22, line 496: To clarify this comment: the shift by 5 is correct because there are two 4x4 blocks, one after the other, the first containing AR and the second containing GB.  It *might* be clearer to instead have the code below use `u32 offset = ((base << 1) + blkOff) << 1`, but I'm not sure that really helps.

Source/Core/VideoCommon/TextureDecoder_Common.cpp
```diff
@@ -434,55 +435,206 @@ static inline u32 DecodePixel_Paletted(u16 pixel, TLUTFormat tlutfmt)
...
+  u32 blkOff = (blkT << 2) + (blkS << 1);
+
+  u32 offset = base + blkOff;
+
+  cmprTexelData ret;
+  std::memcpy(&ret, src + offset, sizeof(ret));
+  return ret;
+}
+
+static constexpr int CacheHashBits = 10;
```
Pokechu22, line 535: `CACHE_HASH_BITS`

Source/Core/VideoCommon/TextureDecoder_Common.cpp
```diff
@@ -434,55 +435,206 @@ static inline u32 DecodePixel_Paletted(u16 pixel, TLUTFormat tlutfmt)
...
+  return color1 | (color2 << 16);
+}
+
+struct CMPRCacheEntry
+{
+  u32 tag = {};
+  u32 colors[4] = {};
+};
+
+static CMPRCacheEntry cmprCache[1 << CacheHashBits] = {};
```
Pokechu22, line 553: `s_cmpr_cache`.  IMO it would also be a good idea to add a comment noting that this cache is only an implementation detail for performance and not an actual requirement for implementing the texture format.

Source/Core/VideoCommon/TextureDecoder_Common.cpp
```diff
@@ -434,55 +435,206 @@ static inline u32 DecodePixel_Paletted(u16 pixel, TLUTFormat tlutfmt)
...
+
+static constexpr u32 cmprCacheSet(u32 color1, u32 color2)
+{
+  return (color1 * color2 * 0x9E3779B1) >> (32 - CacheHashBits);
+}
+
+static constexpr u32 cmprCacheTag(u32 color1, u32 color2)
+{
+  return color1 | (color2 << 16);
+}
```
Pokechu22, line 545: `color1` and `color2` should be `u16` and then casted to u32 in these functions (or the `cmprTexelData` should directly be passed).  With them as `u32` it isn't clear how this avoids collisions (e.g. if `color1` were `0xffff0000` and `color2` were `0x0000ffff` that'd be the same as `color2` being 0), but with them as `u16` it's clear that the returned value completely covers all inputs.

Source/Core/VideoCommon/TextureDecoder_Common.cpp
Pokechu22: Pre-populating it seems like a better solution to me, and wouldn't be that difficult; it'd be something like this:

```C++
static CMPRCacheEntry s_cmpr_cache[1 << CACHE_HASH_BITS] = {{cmprCacheTag(0, 0),
                                                             {
                                                                 MakeRGBA(0, 0, 0, 255),
                                                                 MakeRGBA(0, 0, 0, 255),
                                                                 MakeRGBA(0, 0, 0, 255),
                                                                 MakeRGBA(0, 0, 0, 0),
                                                             }}};
// Pre-populate tag 0 (for input colors of (0, 0)) with the correct value, since we don't have a
// valid bit and thus couldn't tell if tag 0 has a valid value.  Games aren't likely to use this
// though, since it only has 3 solid blacks and a transparent black.
static_assert(cmprCacheSet(0, 0) == 0, "Tag 0 must be at set 0 for above initialization to work");
```

I'm also not 100% sure if the "unlikely to use this" thing is true; wouldn't tag 0 be used whenever a game has a transparent black section of a texture, which doesn't seem particularly uncommon to me?  (Or is there another way to represent transparent black?)

Source/Core/VideoCommon/TextureDecoder_Common.cpp
```diff
@@ -434,55 +435,206 @@ static inline u32 DecodePixel_Paletted(u16 pixel, TLUTFormat tlutfmt)
...
+  }
+
+  return cmprCache[cacheSet].colors;
+}
+
+static u32 TexDecoder_CMPRColorSelect(cmprTexelData data, u32* colors, int s, int t)
+{
+  int ss = s & 3;
+  int tt = t & 3;
+  int shift = tt * 8 + (6 - (ss << 1));
```
Pokechu22, line 611: This would be clearer as `((tt * 4) + (3 - ss)) * 2`, which hopefully would compile to the same thing.

Source/Core/VideoCommon/TextureDecoder_Common.cpp
Pokechu22: Here's a possibility for how to refactor this:

```C++
namespace
{
static constexpr u32 CACHE_HASH_BITS = 10;

class CMPRColorCache
{
private:
  static constexpr u32 GetCacheSet(u16 color1, u16 color2)
  {
    return (u32(color1) * u32(color2) * 0x9E3779B1) >> (32 - CACHE_HASH_BITS);
  }

  static constexpr u32 GetCacheTag(u16 color1, u16 color2)
  {
    return u32(color1) | (u32(color2) << 16);
  }

  static constexpr std::array<u32, 4> ComputeColors(u16 color1, u16 color2)
  {
    int blue1 = Convert5To8(color1 & 0x1F);
    int blue2 = Convert5To8(color2 & 0x1F);
    int green1 = Convert6To8((color1 >> 5) & 0x3F);
    int green2 = Convert6To8((color2 >> 5) & 0x3F);
    int red1 = Convert5To8((color1 >> 11) & 0x1F);
    int red2 = Convert5To8((color2 >> 11) & 0x1F);

    u32 c0 = MakeRGBA(red1, green1, blue1, 255);
    u32 c1 = MakeRGBA(red2, green2, blue2, 255);
    u32 c2, c3;

    if (color1 > color2)
    {
      c2 = MakeRGBA(DXTBlend(red2, red1), DXTBlend(green2, green1), DXTBlend(blue2, blue1), 255);
      c3 = MakeRGBA(DXTBlend(red1, red2), DXTBlend(green1, green2), DXTBlend(blue1, blue2), 255);
    }
    else
    {
      // color[3] is the same as color[2] (average of both colors), but transparent.
      // This differs from DXT1 where color[3] is transparent black.
      c2 = MakeRGBA((red1 + red2) / 2, (green1 + green2) / 2, (blue1 + blue2) / 2, 255);
      c3 = MakeRGBA((red1 + red2) / 2, (green1 + green2) / 2, (blue1 + blue2) / 2, 0);
    }

    return {c0, c1, c2, c3};
  }

  DOLPHIN_NO_INLINE void Miss(cmprTexelData data, u32 cache_set, u32 cache_tag)
  {
    const u16 color1 = Common::swap16(data.color1);
    const u16 color2 = Common::swap16(data.color2);

    entries[cache_set].colors = ComputeColors(color1, color2);
    entries[cache_set].tag = cache_tag;
  }

public:
  constexpr CMPRColorCache()
  {
    // Since we don't have a dedicated valid bit, and we zero-initialize everything, we need to
    // pre-populate tag 0 (corresponding to colors 0 and 0) with the correct data.
    constexpr u32 zero_set = GetCacheSet(0, 0);
    constexpr u32 zero_tag = GetCacheTag(0, 0);
    static_assert(zero_tag == 0);
    entries[zero_set].tag = zero_tag;
    entries[zero_set].colors = ComputeColors(0, 0);
  }

  u32* LookUp(cmprTexelData data)
  {
    // It's quite expensive to derive the four colors of each block.
    // So we cache then, indexed by the two colors.
    const u32 cache_set = GetCacheSet(data.color1, data.color2);
    const u32 cache_tag = GetCacheTag(data.color1, data.color2);

    if (entries[cache_set].tag != cache_tag)
    {
      Miss(data, cache_set, cache_tag);
    }

    return entries[cache_set].colors.data();
  }

private:
  struct CacheEntry
  {
    u32 tag = {};
    std::array<u32, 4> colors = {};
  };
  std::array<CacheEntry, 1 << CACHE_HASH_BITS> entries = {};
};

static CMPRColorCache s_cmpr_cache;
}  // namespace
```

Source/Core/VideoCommon/TextureDecoder_Common.cpp
phire: It produces different code. 

**clang x86_64:** Same number of instructions, but preforms 30% slower (according to MCA) due to dependencies.  
**gcc x86_64:** Same number of instructions, preforms the same according to MCA  
**clang arm64:** One extra instruction (MCA isn't working for arm on godbolt)
**gcc arm64:** One extra instruction (on top of GCC already producing a function that was one instruction longer)

Source/Core/VideoCommon/TextureDecoder_Common.cpp
Pokechu22: Just to follow up on this: I found a case where this does indeed happen: in Mario Kart Wii, at the end of Toad's Factory, there's a mesh-like texture in a pit (`tex1_256x256_m_dec381016fad805a_14.png`), and it fits the c1 = 0, c2 = 0 case. There probably are a bunch of other examples, but this is the first one I actually found where this ends up being the case.

PR #10264, opened 2021-12-03
FifoPlayer: Use early memory updates by default (test)

Pokechu22:


phire:
I don't think this is going to work.

https://fifo.ci/compare/7038774-7032240/ and https://fifo.ci/compare/7038830-7032296/ are two examples where this results in significantly worse results. 

I think early memory updates tries to apply all updates at the beginning of the frame, and will fail with anything that actually updates the content of memory in the middle of a frame.

PR #10265, opened 2021-12-05
RiivolutionParser: Create config xml directory if it does not exist yet.

AdmiralCurtiss:
Apparently I forgot to do this and never noticed, oops.


elsaxo1988:
AdmiralCurtiss:create-riivolution-config-dir

JosJuice (approved):

JosJuice:
Actually, on second thought... Could you put the `CreateFullPath` call inside `WriteConfigFile` instead? That way you'll fix it on Android too.

AdmiralCurtiss:
Sure, I guess.

JosJuice (approved):

PR #10267, opened 2021-12-08
JitArm64: Allocate 64 MB for farcode

JosJuice:
My change in 867cd99 caused farcode to fill up much more than before. Most games still ran fine, but certain games would have multiple cache clears per minute, on top of being overall slow.

Maybe there's something prettier we can do about this than just allocating more RAM, but we have the resource budget for making Dolphin use more RAM, so I consider increasing the size of the cache to be a good solution at least for the time being.

At least for Prince of Persia: The Forgotten Sands, 48 MB isn't enough to stop the cache clears, but 64 MB is. (I only played the game for a few minutes when testing, though.)


brujo5:
Confirmed this PR Fixed slowndowns for me. Thx

PR #10065, opened 2021-08-28
VideoBackends / VideoCommon: expose ability to view shader/texture debug names

iwubcode:
This gives the ability to view an application defined name for our Shaders/Textures in applications like RenderDoc.  Currently nothing takes advantage of this but a future feature will.  Breaking it out to make it easier to review.

PR #10269, opened 2021-12-09
VideoCommon: Add names for textures and shaders

Pokechu22:
This continues the work started in #10065.  Names are now present for all shaders other than the generated pixel/vertex shaders used for emulation and for all textures except the ones used in emulation (as there are a lot of these, and I haven't thought of a good way to create unique names for them).  Names are also not created for shaders created from a binary, but that only applies to shaders used for emulation.

### DIFF
--- a/Source/Core/VideoCommon/TextureCacheBase.h
+++ b/Source/Core/VideoCommon/TextureCacheBase.h
@@ -5,6 +5,7 @@
 
 #include <array>
 #include <bitset>
+#include <fmt/format.h>
 #include <map>
 #include <memory>
 #include <optional>
@@ -69,6 +70,23 @@ struct EFBCopyParams
   bool depth;
   bool yuv;
   bool copy_filter;
+};
+
+template <>
+struct fmt::formatter<EFBCopyParams>
+{
+  constexpr auto parse(format_parse_context& ctx) { return ctx.begin(); }
+  template <typename FormatContext>
+  auto format(const EFBCopyParams& uid, FormatContext& ctx)
+  {
+    std::string copy_format;
+    if (uid.copy_format == EFBCopyFormat::XFB)
+      copy_format = "XFB";
+    else
+      copy_format = fmt::to_string(uid.copy_format);
+    return format_to(ctx.out(), "format: {}, copy format: {}, depth: {}, yuv: {}, copy filter: {}",
+                     uid.efb_format, copy_format, uid.depth, uid.yuv, uid.copy_filter);
+  }
 };
 
 // Reduced version of the full coefficient array, with a single value for each row.
--- a/Source/Core/VideoCommon/TextureConverterShaderGen.h
+++ b/Source/Core/VideoCommon/TextureConverterShaderGen.h
@@ -3,7 +3,11 @@
 
 #pragma once
 
+#include <fmt/format.h>
+#include <string>
+
 #include "Common/CommonTypes.h"
+
 #include "VideoCommon/ShaderGenCommon.h"
 #include "VideoCommon/TextureDecoder.h"
 
@@ -34,3 +38,23 @@ TCShaderUid GetShaderUid(EFBCopyFormat dst_format, bool is_depth_copy, bool is_i
                          bool scale_by_half, bool copy_filter);
 
 }  // namespace TextureConversionShaderGen
+
+template <>
+struct fmt::formatter<TextureConversionShaderGen::UidData>
+{
+  constexpr auto parse(format_parse_context& ctx) { return ctx.begin(); }
+  template <typename FormatContext>
+  auto format(const TextureConversionShaderGen::UidData& uid, FormatContext& ctx)
+  {
+    std::string dst_format;
+    if (uid.dst_format == EFBCopyFormat::XFB)
+      dst_format = "XFB";
+    else
+      dst_format = fmt::to_string(uid.dst_format);
+    return format_to(ctx.out(),
+                     "dst_format: {}, efb_has_alpha: {}, is_depth_copy: {}, is_intensity: {}, "
+                     "scale_by_half: {}, copy_filter: {}",
+                     dst_format, uid.efb_has_alpha, uid.is_depth_copy, uid.is_intensity,
+                     uid.scale_by_half, uid.copy_filter);
+  }
+};
--- a/Source/Core/VideoCommon/FramebufferManager.cpp
+++ b/Source/Core/VideoCommon/FramebufferManager.cpp
@@ -3,6 +3,7 @@
 
 #include "VideoCommon/FramebufferManager.h"
 
+#include <fmt/format.h>
 #include <memory>
 
 #include "Common/ChunkFile.h"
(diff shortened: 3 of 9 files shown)


iwubcode (approved):

iwubcode:
Very nice work @Pokechu22 .  I'd also like to expose a way to do event regions:  https://renderdoc.org/docs/how/how_annotate_capture.html#application-provided-marker-regions but haven't had time to implement that yet.

Do note there is a bug in d3d12's compute shader implementation that causes a crash.  I have a fix for that, I need to push it up.

iwubcode:
> all textures except the ones used in emulation (as there are a lot of these, and I haven't thought of a good way to create unique names for them)

Do you mean the ones in the texture cache?  I'd suggest using [GetFullName()](https://github.com/dolphin-emu/dolphin/blob/master/Source/Core/VideoCommon/TextureInfo.h#L31) which is the common name texture pack authors are used to seeing .  Though not sure how much of a perf impact that'd have (?)

Pokechu22:
`GetFullName()` would be a good choice (the possible performance issue (if there is one) could be worked around by only calling it if `g_ActiveConfig.bEnableValidationLayer` is set, since that also enables other debug functions).  However, what we have is an instance of `TextureConfig`, not `TextureInfo`, so `GetFullName()` isn't available.

------

For what it's worth, I've also made the following change:

<details><summary>Patch</summary>

```patch
diff --git a/Source/Core/VideoCommon/ShaderCache.cpp b/Source/Core/VideoCommon/ShaderCache.cpp
index 257650713c..27edf8d3dc 100644
--- a/Source/Core/VideoCommon/ShaderCache.cpp
+++ b/Source/Core/VideoCommon/ShaderCache.cpp
@@ -422,8 +422,7 @@ std::unique_ptr<AbstractShader> ShaderCache::CompileVertexShader(const VertexSha
 {
   const ShaderCode source_code =
       GenerateVertexShaderCode(m_api_type, m_host_config, uid.GetUidData());
-  return g_renderer->CreateShaderFromSource(ShaderStage::Vertex, source_code.GetBuffer(),
-                                            "Dolphin vertex shader");
+  return g_renderer->CreateShaderFromSource(ShaderStage::Vertex, source_code.GetBuffer());
 }

 std::unique_ptr<AbstractShader>
@@ -431,16 +430,14 @@ ShaderCache::CompileVertexUberShader(const UberShader::VertexShaderUid& uid) con
 {
   const ShaderCode source_code =
       UberShader::GenVertexShader(m_api_type, m_host_config, uid.GetUidData());
-  return g_renderer->CreateShaderFromSource(ShaderStage::Vertex, source_code.GetBuffer(),
-                                            "Dolphin vertex UberShader");
+  return g_renderer->CreateShaderFromSource(ShaderStage::Vertex, source_code.GetBuffer());
 }

 std::unique_ptr<AbstractShader> ShaderCache::CompilePixelShader(const PixelShaderUid& uid) const
 {
   const ShaderCode source_code =
       GeneratePixelShaderCode(m_api_type, m_host_config, uid.GetUidData());
-  return g_renderer->CreateShaderFromSource(ShaderStage::Pixel, source_code.GetBuffer(),
-                                            "Dolphin pixel shader");
+  return g_renderer->CreateShaderFromSource(ShaderStage::Pixel, source_code.GetBuffer());
 }

 std::unique_ptr<AbstractShader>
@@ -448,8 +445,7 @@ ShaderCache::CompilePixelUberShader(const UberShader::PixelShaderUid& uid) const
 {
   const ShaderCode source_code =
       UberShader::GenPixelShader(m_api_type, m_host_config, uid.GetUidData());
-  return g_renderer->CreateShaderFromSource(ShaderStage::Pixel, source_code.GetBuffer(),
-                                            "Dolphin pixel UberShader");
+  return g_renderer->CreateShaderFromSource(ShaderStage::Pixel, source_code.GetBuffer());
 }

 const AbstractShader* ShaderCache::InsertVertexShader(const VertexShaderUid& uid,
```

</details>

This is because giving names for those results in RenderDoc always using them, but this results in ambiguity, as the names are always the same (all of them show up as "Dolphin vertex shader"); not supplying a name results in "Shader Module 440" or similar showing up, with a unique number.  So, until detailed names are available for those generated shaders, it's probably best not to apply any name.

iwubcode:
> GetFullName() would be a good choice (the possible performance issue (if there is one) could be worked around by only calling it if g_ActiveConfig.bEnableValidationLayer is set, since that also enables other debug functions). However, what we have is an instance of TextureConfig, not TextureInfo, so GetFullName() isn't available.

We do have a `TextureInfo` as it is passed into `GetTexture()` so it could be used here:  https://github.com/dolphin-emu/dolphin/blob/master/Source/Core/VideoCommon/TextureCacheBase.cpp#L1550  if passed down.  There are other places like the reinterpret and the palette where we are only given the config or the cache entry pointer, so we'd have to save that off somehow but at least for "normal" textures, it could be done rather easily.

> This is because giving names for those results in RenderDoc always using them, but this results in ambiguity, as the names are always the same (all of them show up as "Dolphin vertex shader"); not supplying a name results in "Shader Module 440" or similar showing up, with a unique number. So, until detailed names are available for those generated shaders, it's probably best not to apply any name.

Could we use uid possibly?  It's fine leaving it as, not sure how useful the context of knowing it was a vertex/pixel/uber shader was (vertex/pixel is something you can get from just looking at the pipeline).

Pokechu22:
> Could we use uid possibly?

That's definitely possible, and I've already done something similar for geometry shaders... it's just that the UID has a lot of information in it, so it would result in a very large name (the names created here are already somewhat long (I think the EFB to VRAM or maybe EFB to RAM one ends up being multiple lines (though in a large font)), and including everything in the UID would result in a stupid-long name probably).  It might be fine to just use the number of texgens and such (the information that ends up in the comment at the top of the shader), but I think it'd be better to address this in a later PR.

Source/Core/VideoCommon/TextureCacheBase.h
```diff
@@ -71,6 +72,23 @@ struct EFBCopyParams
...
+{
+  constexpr auto parse(format_parse_context& ctx) { return ctx.begin(); }
+  template <typename FormatContext>
+  auto format(const EFBCopyParams& uid, FormatContext& ctx)
+  {
+    std::string_view copy_format;
+    if (uid.copy_format == EFBCopyFormat::XFB)
+      copy_format = "XFB";
+    else
+      copy_format = fmt::to_string(uid.copy_format);
```
leoetlino, line 86: This seems unsafe: `copy_format` is a string_view, and fmt::to_string returns a std::string which is destructed after the assignment, so I think `copy_format` would end up being a dangling view.

Source/Core/VideoCommon/TextureConverterShaderGen.h
```diff
@@ -34,3 +37,23 @@ TCShaderUid GetShaderUid(EFBCopyFormat dst_format, bool is_depth_copy, bool is_i
...
+{
+  constexpr auto parse(format_parse_context& ctx) { return ctx.begin(); }
+  template <typename FormatContext>
+  auto format(const TextureConversionShaderGen::UidData& uid, FormatContext& ctx)
+  {
+    std::string_view dst_format;
+    if (uid.dst_format == EFBCopyFormat::XFB)
+      dst_format = "XFB";
+    else
+      dst_format = fmt::to_string(uid.dst_format);
```
leoetlino, line 52: Same dangling view issue here.

leoetlino (changes requested):

leoetlino (approved):

brujo5:
android build crash when launch game with opengl and vulkan since this version. thx

PR #10270, opened 2021-12-10
Treewide: Adjust order of includes

Pokechu22:
This PR adjusts includes in several files to match [contributing.md's guidelines](https://github.com/dolphin-emu/dolphin/blob/master/Contributing.md#headers).  The main thing that I was looking for was putting the header for the source file at the top of the list.

I also found and fixed several other minor issues, as doing this involved looking through all sources files in the project.  Apart from the issues with dedicated commits, there are the following:

- Video backends: the "main" file now includes the backend's VideoBackend.h file at the top (as that's effectively the header for the source file)
- DXTexture.h and DXPipeline.h were missing an include for D3DBase.h, meaning they didn't work if they were included first.
- ARM/x64-specific files include the common header (e.g. vertex loaders, CPUDetect)
- Cases where multiple files are used to implement something declared in one header use that header (e.g. JIT/interpreter)
- ENetUtil and DolphinFileDialog used file-relative includes before
- EXI_DeviceIPL had an unused include of DVDInterface.h
- EmuSubroutines included WiimoteReal.h when it only needed HW/Wiimote.h


Source/Core/InputCommon/ControllerInterface/DInput/DInputJoystick.cpp
```diff
@@ -1,5 +1,7 @@
 // Copyright 2010 Dolphin Emulator Project
-// SPDX-License-Identifier: GPL-2.0-or-later
+// SPDX-License-Identifier:
```
JosJuice, line 2: Something went wrong here.

Source/Core/Common/GenericFPURoundMode.cpp
```diff
@@ -1,9 +1,11 @@
+
```
Warepire, line 1: Empty line snuck in here.

JosJuice (approved):

BhaaLseN (approved):

PR #10271, opened 2021-12-10
Merge CRC32.h into Hash.h and remove MD5.h

Pokechu22:
The main point of this is to have one obvious location where all hash-related code is.  I've also used the wrapper around zlib's `crc32` in other places (it was first added for the GameCube IPL rom in 3d662e746bef6651dcdd0bc985c9e6d6518f006d).


Source/Core/Common/Hash.h
```diff
@@ -4,14 +4,28 @@
 #pragma once
 
 #include <cstddef>
+#include <functional>
+#include <string_view>
 
 #include "Common/CommonTypes.h"
 
+namespace DiscIO
```
lioncash, line 12: Common shouldn't be dependent on DiscIO

lioncash (changes requested):

Source/Core/Common/Hash.h
Pokechu22: MD5.cpp already was in common and depended on DiscIO.

JosJuice: The only function in MD5.cpp seems to have NetPlayClient.cpp as its only caller. I would suggest moving it either to there or to DiscIO.

Source/Core/Core/NetPlayClient.cpp
```diff
@@ -2437,6 +2436,43 @@ bool NetPlayClient::DoAllPlayersHaveGame()
   });
 }
 
+static std::string MD5Sum(const std::string& file_path, std::function<bool(int)> report_progress)
```
JosJuice, line 2439: `#include <functional>`

Source/Core/Core/NetPlayClient.cpp
```diff
@@ -2437,6 +2437,43 @@ bool NetPlayClient::DoAllPlayersHaveGame()
...
+  }
+
+  std::array<u8, 16> output;
+  mbedtls_md5_finish_ret(&ctx, output.data());
+
+  // Convert to hex
+  for (u8 n : output)
+    output_string += fmt::format("{:02x}", n);
+
+  return output_string;
```
iwubcode, line 2474: `return fmt::format("{:02x}", fmt::join(output, ""));` ?

Pokechu22: Changed - I don't think this is necessarily more performant, but it's also not going to be significantly worse.  (It's formatting an array of 16 characters, and this happens once per netplay session.)

Source/Core/Core/NetPlayClient.cpp
```diff
@@ -2462,6 +2462,43 @@ bool NetPlayClient::DoAllPlayersHaveGame()
...
 }
 
+static std::string MD5Sum(const std::string& file_path, std::function<bool(int)> report_progress)
+{
+  std::string output_string;
+  std::vector<u8> data(8 * 1024 * 1024);
+  u64 read_offset = 0;
+  mbedtls_md5_context ctx;
+
+  std::unique_ptr<DiscIO::BlobReader> file(DiscIO::CreateBlobReader(file_path));
```
leoetlino, line 2471: Missing include for DiscIO::CreateBlobReader (DiscIO/Blob.h).

leoetlino (changes requested):

leoetlino (approved):

PR #10272, opened 2021-12-11
Android: Rework intent handling to work under scoped storage

JosJuice:
It seems like Android frontend apps don't work with Dolphin anymore now that we're using scoped storage. With this PR, I'm providing a way for frontend apps to supply content URIs to Dolphin, since content URIs work with scoped storage. Frontend apps will probably need to be modified to make use of this, but there are already other emulators like PPSSPP that work the same way, so hopefully it's not a too big thing to ask them to do.

### DIFF
--- a/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/utils/StartupHandler.java
+++ b/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/utils/StartupHandler.java
@@ -2,8 +2,11 @@
 
 package org.dolphinemu.dolphinemu.utils;
 
+import android.content.ClipData;
 import android.content.Context;
+import android.content.Intent;
 import android.content.SharedPreferences;
+import android.net.Uri;
 import android.os.Bundle;
 import android.preference.PreferenceManager;
 import android.text.TextUtils;
@@ -15,6 +18,7 @@ import org.dolphinemu.dolphinemu.activities.EmulationActivity;
 
 import java.time.Instant;
 import java.time.temporal.ChronoUnit;
+import java.util.Objects;
 
 public final class StartupHandler
 {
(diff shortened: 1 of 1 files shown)


Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/utils/StartupHandler.java
```diff
@@ -33,27 +37,60 @@ public static void HandleInit(FragmentActivity parent)
...
+    {
+      String[] uris = new String[clipData.getItemCount()];
+      for (int i = 0; i < uris.length; i++)
       {
-        String start_file = extras.getString("AutoStartFile");
-        if (!TextUtils.isEmpty(start_file))
-        {
-          start_files = new String[]{start_file};
-        }
+        uris[i] = Objects.toString(clipData.getItemAt(i).getUri());
```
ssdsnake, line 68: Hi, Here's an idiomatic form that should work on API level 24+
```
  return IntStream.range(0, clipData.getItemCount()).boxed()
          .map(ClipData::::getItemAt)
          .collect(Collectors.toList())
          .toArray(String[]::new);
```

Proof / sample:
```
import java.util.Arrays;
import java.util.List;
import java.util.stream.Collectors;
import java.util.stream.IntStream;

public class Main {

    public static void main(String[] args) {
        List<String> test = List.of("mario", "rosalina", "luigi", "bowser");

        // You would return this
        String[] result = IntStream.range(0, test.size()).boxed()
                .map(test::get)
                .collect(Collectors.toList())
                .toArray(String[]::new);

        // Just to see the output
        Arrays.stream(result).forEach(System.out::println);
    }
}
```

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/utils/StartupHandler.java
```diff
@@ -33,27 +37,60 @@ public static void HandleInit(FragmentActivity parent)
     if (TvUtil.isLeanback(parent))
       TvUtil.scheduleSyncingChannel(parent);
 
-    String[] start_files = null;
-    Bundle extras = parent.getIntent().getExtras();
-    if (extras != null)
+    String[] games_to_launch = getGamesFromIntent(parent.getIntent());
+    if (games_to_launch != null && games_to_launch.length > 0)
```
ssdsnake, line 41: camel case maybe? (cause Java :D)

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/utils/StartupHandler.java
ssdsnake: Appears minimum is 21 (Android 5 / 2014) though in build.gradle . Disregard if not posssible

JosJuice: I will skip this due to the minSdkVersion.

Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/utils/StartupHandler.java
JosJuice: Done.

PR #10273, opened 2021-12-12
FifoPlayer: Wait after clearing the screen

Pokechu22:
Continues #10258.  If we don't wait, then the copy will be performed at a later time, which may overwrite memory updates.  This was causing problems for me when messing with memory updates, since the EFB copy would end up overwriting memory that had been written by the memory update.  I think this may have been the cause of some of the differences in #10264, and will submit a second PR to check.

### DIFF
--- a/Source/Core/Core/FifoPlayer/FifoPlayer.cpp
+++ b/Source/Core/Core/FifoPlayer/FifoPlayer.cpp
@@ -214,6 +214,9 @@ public:
   void Init() override
   {
     IsPlayingBackFifologWithBrokenEFBCopies = m_parent->m_File->HasBrokenEFBCopies();
+    // Without this call, we deadlock in initialization in dual core, as the FIFO is disabled and
+    // thus ClearEfb()'s call to WaitForGPUInactive() never returns
+    CPU::EnableStepping(false);
 
     m_parent->m_CurrentFrame = m_parent->m_FrameRangeStart;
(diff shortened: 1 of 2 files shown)


Pokechu22:
#10274 shows that this improves early memory updates (though early memory updates still isn't a good solution).  So although this produces no visible fifoci differences, it is an improvement.

iwubcode:
I ran some fifo logs and all of them ran as a black screen.  Master shows them correctly.

Pokechu22:
It seems like that happens with dual core (but not single core).  However. fifoci uses dual core if I recall correctly, and [it ran without issues](https://fifo.ci/version/379add904e1074459f6a653a307722aa0d0beb18/) so I'm not sure quite what's going on.

Pokechu22:
Fixed.  I needed to add this to the fifoplayer's `CPUCore::Init`:

https://github.com/dolphin-emu/dolphin/blob/86f2b39d660673bb3ae993dc2d71e34b36880e98/Source/Core/Core/FifoPlayer/FifoPlayer.cpp#L217-L219

(Also, I was incorrect in my thought that FifoCI uses dual core; I misinterpreted what `CPUThread = False` meant in configs.  `CPUThread = False` actually means "don't use a separate CPU thread from the GPU thread", i.e. "don't use dual core".)

iwubcode (approved):
Dual core and single core now work.  Code changes make sense

PR #10274, opened 2021-12-12
FifoPlayer: Wait after clearing the screen + use early memory updates by default (test)

Pokechu22:
Combination of #10273 and #10264, for testing.


Pokechu22:
This fixes the random black textures in #10264, but early memory updates still seems to cause a lot of issues.  In particular, vertex arrays have issues ([dckr-char](https://fifo.ci/compare/7062880-7062396/) is a good example), and some things from the last frame are applied to earlier frames ([nsmbw-coins](https://fifo.ci/compare/7062884-7062400/) is a good example).

PR #10275, opened 2021-12-14
Jit64: Use farcode for exception exit in twX.

AdmiralCurtiss:
See https://forums.dolphin-emu.org/Thread-new-5-0-15445-causes-jump-target-too-far-away-needs-force5bytes-true

I didn't do it yet, but it might be worth bisecting what change exactly caused the distance to become too long?

### DIFF
--- a/Source/Core/Core/PowerPC/Jit64/Jit_Integer.cpp
+++ b/Source/Core/Core/PowerPC/Jit64/Jit_Integer.cpp
@@ -2559,9 +2559,11 @@ void Jit64::twX(UGeckoInstruction inst)
       fixups.push_back(f);
     }
   }
-  FixupBranch dont_trap = J();
 
+  if (!fixups.empty())
   {
+    SwitchToFarCode();
+
     RCForkGuard gpr_guard = gpr.Fork();
     RCForkGuard fpr_guard = fpr.Fork();
 
@@ -2577,9 +2579,9 @@ void Jit64::twX(UGeckoInstruction inst)
     fpr.Flush();
 
     WriteExceptionExit();
-  }
 
-  SetJumpTarget(dont_trap);
+    SwitchToNearCode();
+  }
 
   if (!analyzer.HasOption(PPCAnalyst::PPCAnalyzer::OPTION_CONDITIONAL_CONTINUE))
   {


JosJuice:
I added the following extra instruction when fixing True Crime. Maybe that managed to push it over the edge...

https://github.com/dolphin-emu/dolphin/blob/f4e2d369c075f44f037289a1ccb78274c7d5cb39/Source/Core/Core/PowerPC/Jit64/Jit_Integer.cpp#L2574

Do you think it would make sense to make `twX` use farcode? Or is the exception path commonly taken?

AdmiralCurtiss:
I have not even the slightest idea how often this path is taken. Would be nice if we had metrics like that... But yes, if it's somewhat rare that's probably not a bad idea.

JosJuice:
`tw` and `twi` are the only instructions that generate `ProgramExceptionCause::Trap`, so one way to measure how common it is could be to add an if statement here that outputs a message to the log if `SRR1 == ProgramExceptionCause::Trap`: https://github.com/dolphin-emu/dolphin/blob/b997048cfee870449e875a8e6752288013c10853/Source/Core/Core/PowerPC/PowerPC.cpp#L486

saulfabregwiivc:
@AdmiralCurtiss @JosJuice 

I tested this build with this pull request on Windows x64 [[pr-10275-win-x64](https://dolphin.ci/#builders/21/builds/5508)] **(i saw that is based in Dolphin 5.0-15605) x64** and the problem of "Jump target too far away, needs force5Bytes = true" is now gone, now the Not64 emulator works again as a charm, as i specified in my thread with old 5.0-15262.

Will you bring this fix for a new development or beta release?

Sintendo:
Do we know of any other titles using tw/twi? [Issue 7253](https://bugs.dolphin-emu.org/issues/7253) mentions Monopoly, but I couldn't hit this instruction for any title that I happened to have at hand.

JMC47:
Does forcing 5 bytes mean that we lose performance?  Does this cost performance in general?

JosJuice:
It will make the code size a little larger. In some cases, this can cause a small performance drop due to less code being able to fit in the icache.

AdmiralCurtiss:
The difference should be minuscule in practice though, especially since apparently this instruction is barely used by games.

AdmiralCurtiss:
I hacked myself some metrics and didn't encounter a single exceptional exit while playing around, so yes moving this to farcode is probably a good idea.

JosJuice (approved):

PR #10276, opened 2021-12-14
Android: Fix opening system file manager

JosJuice:
Turns out that most phones ship with a special Google version of DocumentsUI instead of just using the AOSP version, despite the two being pretty similar as far as I can tell. This change makes us check for both package names instead of just the AOSP package name.

### DIFF
--- a/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/activities/UserDataActivity.java
+++ b/Source/Android/app/src/main/java/org/dolphinemu/dolphinemu/activities/UserDataActivity.java
@@ -59,22 +59,33 @@ public class UserDataActivity extends AppCompatActivity implements View.OnClickL
   {
     try
     {
-      startActivity(getFileManagerIntent());
+      // First, try the package name used on "normal" phones
+      startActivity(getFileManagerIntent("com.google.android.documentsui"));
     }
     catch (ActivityNotFoundException e)
     {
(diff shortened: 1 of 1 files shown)


ssdsnake:
I'm happy to help test this, I don't recall how to get an APK but I have a ton of bored droids
I _think_ Samsung has a special intent too for their file manager but I don't know that it's worth exploring if the Google/AOSP one opens up fine

lioncash (approved):

mbc07:
This PR fixes [issue 12755](https://bugs.dolphin-emu.org/issues/12755)...

JosJuice:
> I *think* Samsung has a special intent too for their file manager but I don't know that it's worth exploring if the Google/AOSP one opens up fine

Unless the Samsung file manager supports accessing Android/data/ and there are phones where the Samsung file manager is installed but not easily for the user to access, there's no reason to add a button for it. Because this is the reason why we have a button for the Google file manager – on some phones this button is the only way to access Android/data/ without installing extra apps.

mrdeathjr28:
hi in my case dont work, when try select game folder in file manager app crash maybe can be related with this:

https://bugs.dolphin-emu.org/issues/12764#change-743465

JosJuice:
I don't see how that could be related to this pull request (or to the issue report you linked for that matter). Please create a new issue report for it at https://bugs.dolphin-emu.org/.

PR #10277, opened 2021-12-14
MemArena: Cleanup & interface changes for safer memory mapping.

AdmiralCurtiss:
See #9544 and #9834.

Note that there are no functional changes in this PR. (or should be, anyway)

### DIFF
--- a/Source/Core/Common/MemArena.h
+++ b/Source/Core/Common/MemArena.h
@@ -15,18 +15,87 @@ namespace Common
 {
 // This class lets you create a block of anonymous RAM, and then arbitrarily map views into it.
 // Multiple views can mirror the same section of the block, which makes it very convenient for
-// emulating
-// memory mirrors.
-class MemArena
+// emulating memory mirrors.
+class MemArena final
 {
 public:
+  MemArena();
+  ~MemArena();
+  MemArena(const MemArena&) = delete;
+  MemArena(MemArena&&) = delete;
+  MemArena& operator=(const MemArena&) = delete;
+  MemArena& operator=(MemArena&&) = delete;
+
+  ///
+  /// Allocate the singular memory segment handled by this MemArena. This will be the actual
+  /// 'physical' available memory for this arena. After allocation, it can be interacted with using
+  /// CreateView() and ReleaseView(). Used to make a mappable region for emulated memory.
+  ///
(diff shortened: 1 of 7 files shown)


Source/Core/Common/MemArena.h
```diff
@@ -15,18 +15,57 @@ namespace Common
...
-  // This finds 1 GB in 32-bit, 16 GB in 64-bit.
-  static u8* FindMemoryBase();
+  // Reserve a memory region, but don't allocate anything within it.
+  // This is used to create our 'fastmem' memory area for the emulated game code to access directly.
+  // 'memory_size': Size in bytes of the memory region to reserve.
+  // Returns a pointer to the memory region, or nullptr on failure.
+  u8* ReserveMemoryRegion(size_t memory_size);
+
+  // Release a memory region previously reserved with ReserveMemoryRegion().
+  void ReleaseMemoryRegion();
```
Pokechu22, line 55: The use of "<em>a</em> memory region" implies that `ReserveMemoryRegion` can be called multiple times; perhaps it should be "the memory region" (or "the memory region, if any") instead?  (The implementations in the other PRs seem to indicate this is how it should work.)

Pokechu22 (approved):

Source/Core/Common/MemArena.h
```diff
@@ -15,18 +15,76 @@ namespace Common
...
+  MemArena(MemArena&&) = delete;
+  MemArena& operator=(const MemArena&) = delete;
+  MemArena& operator=(MemArena&&) = delete;
+
+  // Allocate the singular memory segment handled by this MemArena. This will be the actual
+  // 'physical' available memory for this arena. After allocation, it can be interacted with using
+  // CreateView() and ReleaseView(). This is used to make a mappable region for emulated memory.
+  //
+  // Parameters:
+  //  - 'size' is the amount of bytes that should be allocated in this region.
```
lioncash, line 34: You may want to turn these comments into doxygen style comments with `///` or `/**` and use `@param` here

clang has a -Wdocumentation warning level that can flag anything if the parameter names change in the future through refactoring and the comment falls out of sync

AdmiralCurtiss: I matched it to the format used in `BitUtils.h`, I assume that's good?

PR #10279, opened 2021-12-16
TextureConverterShaderGen: Set alpha to 1 on intensity formats if EFB lacks alpha

Pokechu22:
We were already doing this for non-intensity formats, but it seems like the same applies to intensity formats.  Note that the software renderer already sets alpha to 1 for both intensity and regular formats.

~~This appears to fix [Rygar: The Battle of Argus's white screens](https://bugs.dolphin-emu.org/issues/12763), but I can't tell if this fix is actually correct (as the game's bloom effect doesn't seem to work properly still); I also haven't hardware tested this behavior (but it seems like it's probably correct to me).  It seems to roughly match the results I was getting with the hardware FIFO player, though.~~  The bloom effect works properly in areas where it is enabled (which does not include the area that was experiencing the white screen), and I've confirmed with the hardware FIFO player that I get matching results in areas with bloom enabled.  I haven't done a standalone hardware test confirming this behavior, but it seems like it is almost certainly correct.

### DIFF
--- a/Data/Sys/GameSettings/RPBE01.ini
+++ b/Data/Sys/GameSettings/RPBE01.ini
@@ -1,0 +1,10 @@
+# RPBE01 - Pokemon Battle Revolution
+
+[OnFrame]
+# Fixes black pixels on the bottom/right of the screen, and black in
+# some move effects that apply distortion to the screen
+# This fixes a bug in the base game, not an emulation issue
+# See https://gist.github.com/Pokechu22/e9fa9037fe21312ff32475638b78ba27
+$Fix black screen effects
+0x80244A94:dword:0x39080000
+0x80244A9C:dword:0x38030000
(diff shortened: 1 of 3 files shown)


Miksel12:
This also fixes https://bugs.dolphin-emu.org/issues/11875

Pokechu22:
I've added a patch that fixes https://bugs.dolphin-emu.org/issues/12629 (a game bug in Pokémon Battle Revolution resulting in black appearing in some screen effects), since it's related to the offset issue ([11875](https://bugs.dolphin-emu.org/issues/11875)).

I've also done a writeup explaining both the offset issue and the patch I made: <https://gist.github.com/Pokechu22/e9fa9037fe21312ff32475638b78ba27>.  I plan on doing a second writeup examining how the bloom effect works with Rygar: The Battle of Argus and why that effect didn't work in Dolphin.  (Note that some additional testing has already shown that the bloom effect _does_ work in some cases with this patch; the game tries to disable the bloom effect in other situations, including the one for the fifolog I was working with before, which is why I initially thought it was completely broken).

Pokechu22:
And here's the writeup of Rygar's bloom effect: https://gist.github.com/Pokechu22/f9c6f22388f1c37570a546b80fdd7ddc

iwubcode (approved):
Seems reasonable, untested.

PR #10280, opened 2021-12-17
Core / DolphinQt: make WFS directory configurable

iwubcode:
While the wfs path is only used for one game (DQX), it is a *huge* game.  It's nice to be able to load the files from another location easily.

Fixes:  https://bugs.dolphin-emu.org/issues/10472

### DIFF
--- a/Source/Core/DolphinQt/Settings/PathPane.cpp
+++ b/Source/Core/DolphinQt/Settings/PathPane.cpp
@@ -111,6 +111,17 @@ void PathPane::BrowseSDCard()
   }
 }
 
+void PathPane::BrowseWFS()
+{
+  const QString dir = QDir::toNativeSeparators(DolphinFileDialog::getExistingDirectory(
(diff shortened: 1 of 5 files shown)


AdmiralCurtiss:
Doesn't seem like the shown path in the GUI updates when setting a new folder via the '...' button?

Source/Core/DolphinQt/Settings/PathPane.cpp
```diff
@@ -111,6 +111,17 @@ void PathPane::BrowseSDCard()
   }
 }
 
+void PathPane::BrowseWFS()
+{
+  const QString dir = QDir::toNativeSeparators(DolphinFileDialog::getExistingDirectory(
+      this, tr("Select WFS Path"), QString::fromStdString(Config::Get(Config::MAIN_WFS_PATH))));
+  if (!dir.isEmpty())
+  {
+    m_load_edit->setText(dir);
```
AdmiralCurtiss, line 120: Oh, it's because of this here, this should be `m_wfs_edit`.

AdmiralCurtiss (approved):
Seems reasonable.

PR #10281, opened 2021-12-18
LightingShaderGen: Make s_lighting_struct not inline

Pokechu22:
This generated warnings on the freebsd builder ([example](https://dolphin.ci/#/builders/13/builds/5539)).

### DIFF
--- a/Source/Core/VideoCommon/LightingShaderGen.h
+++ b/Source/Core/VideoCommon/LightingShaderGen.h
@@ -36,13 +36,13 @@ struct LightingUidData
   u32 light_mask : 32;     // 4x8 bits
 };
 
-constexpr inline char s_lighting_struct[] = "struct Light {\n"
-                                            "\tint4 color;\n"
-                                            "\tfloat4 cosatt;\n"
-                                            "\tfloat4 distatt;\n"
-                                            "\tfloat4 pos;\n"
-                                            "\tfloat4 dir;\n"
-                                            "};\n";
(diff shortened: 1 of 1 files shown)


Source/Core/VideoCommon/LightingShaderGen.h
```diff
@@ -36,13 +36,13 @@ struct LightingUidData
...
 };
 
-constexpr inline char s_lighting_struct[] = "struct Light {\n"
-                                            "\tint4 color;\n"
-                                            "\tfloat4 cosatt;\n"
-                                            "\tfloat4 distatt;\n"
-                                            "\tfloat4 pos;\n"
-                                            "\tfloat4 dir;\n"
-                                            "};\n";
+static constexpr inline char s_lighting_struct[] = "struct Light {\n"
```
lioncash, line 39: This can be simplified to:

```suggestion
constexpr char s_lighting_struct[] = "struct Light {\n"
```

and it should also resolve the warnings

Pokechu22: I guess that also means it won't be inlined into each object file... but that's a good thing; there's no reason for it to be duplicated 4 times for each uber/specialized vertex/pixel shader (especially since each one only uses it as `out.Write("{}", s_lighting_struct);`)

iwubcode (approved):

iwubcode:
Someday soon we can make this a `string_view`  :)

PR #10282, opened 2021-12-19
Android: Make WFS directory configurable

JosJuice:
The Android version of PR #10280. No scoped storage support, though...


JMC47:
Technically works, but moving it outside of the user directory is not possible, and if you're running without Scoped Storage, there's no point.  Still worth having.

PR #10285, opened 2021-12-23
ControllerEmu: Hide the cursor if the input gate is disabled

Pokechu22:
Before, if the window was unfocused and background input was disabled, the cursor would become visible and move to the center of the screen.  Now, the cursor is set as hidden.

This is particularly useful for Super Paper Mario, where having the cursor on screen puts the game into a separate mode.  Without this change, even if Hide was bound to something like <code>!\`M\`</code>, the cursor would become visible when the window lost focus unless background input is enabled; this makes recording fifologs very annoying.


JMC47:
Seems to work at a glance.  I had this issue for a long time.

PR #10286, opened 2021-12-24
GeomeryShaderGen: Set gl_ClipDistance on Vulkan as well as OpenGL

Pokechu22:
Fixes https://bugs.dolphin-emu.org/issues/12548.

Both Vulkan and OpenGL use `gl_ClipDistance` when backend depth clamping is enabled.  However, the geometry shader was only copying it for OpenGL, not for Vulkan.  Since stereoscopic 3D uses a geometry shader, this was most visible there, but I suspect it could also affect line/point effects (I was unable to confirm this though).

Here's the relevant vertex shader code (skipping the non-geomerty-shader path):

https://github.com/dolphin-emu/dolphin/blob/6dae4043cb967925ae28cce92d9c100330a5d17e/Source/Core/VideoCommon/VertexShaderGen.cpp#L561-L568
https://github.com/dolphin-emu/dolphin/blob/6dae4043cb967925ae28cce92d9c100330a5d17e/Source/Core/VideoCommon/VertexShaderGen.cpp#L582-L599

### DIFF
--- a/Source/Core/VideoCommon/GeometryShaderGen.cpp
+++ b/Source/Core/VideoCommon/GeometryShaderGen.cpp
@@ -331,21 +331,19 @@ static void EmitVertex(ShaderCode& out, const ShaderHostConfig& host_config,
   if (wireframe && first_vertex)
     out.Write("\tif (i == 0) first = {};\n", vertex);
 
-  if (api_type == APIType::OpenGL)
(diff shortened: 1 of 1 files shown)


iwubcode (approved):
LGTM.  Tested and confirmed that the bug was fixed, didn't notice any regressions

PR #10287, opened 2021-12-24
GameINI: Disable XFB to Texture Only with Resident Evil 0

OatmealDome:
Opening and closing menus have annoying purple flashing when XFB to Texture Only is enabled.

### DIFF
--- a/Data/Sys/GameSettings/GBZ.ini
+++ b/Data/Sys/GameSettings/GBZ.ini
@@ -1,0 +1,5 @@
+# GBZE08, GBZP08, GBZJ08 - Resident Evil 0
+
+[Video_Hacks]
+# Fixes purple screens when transitioning between menus.
+XFBToTextureEnable = False


JMC47:
Do you think this flickering is a seizure problem?  If not, do we know for sure that it'd be worth enabling this for Android where they're more performance constrained?

OatmealDome:
The flashing is just a couple continuous frames of bright purple every time you open, close, or go to a new menu, so it's probably not seizure inducing.

Personally, I find it quite annoying since Resident Evil 0 (like many classic Resis) has a lot of inventory management so you have to open and close the inventory often. Not sure if Android users would think the same if it drops performance, though.

https://user-images.githubusercontent.com/11504941/147391580-49b8c771-4f04-408a-83bf-af79c8c93281.mov

dezraj:
Hello! I'm not sure if this is the right place for this, but I'm also seeing this very same purple flashing frames in Extreme G 3 loading screens. It's not as severe as in RE0 but it is quite noticeable too.

Perhaps you are already aware of that or should I open an issue for investigating this?

Anyway, thank you so much for your work.

PR #10288, opened 2021-12-24
Jit64: Make paired loads always safe

JosJuice:
It's always a good sign when the comments say "this will definitely crash" and "I don't know if this is for a good reason".

Fixes https://bugs.dolphin-emu.org/issues/12762.


JMC47:
Fixes Rygar: Battle of Argus

merryhime (approved):

JMC47:
This fixes backpatch errors in other titles, such as Disney Infinity/Cars 2 when MMU is disabled, allowing them to be played without MMU.  No they are not any faster.

PR #10292, opened 2021-12-25
PixelShaderGen: Use LOD bias when sampling texture on Metal and OpenGL ES

OatmealDome:
#9998, take two.

In Twilight Princess, the iris of Midna's eyes disappears as the camera moves further away from her model. This is done by using two mipmaps: one with her iris, and one with just the pale yellow of her sclera. The game sets the distance at which the transition between the two mipmaps occurs by setting the LOD bias value to a large negative value.

To emulate LOD bias in the Vulkan backend, ``mipLodBias`` in the ``VkSamplerCreateInfo`` for the texture's sampler is set. Unfortunately, Metal does not support setting LOD bias in a sampler, so `mipLodBias` is ignored by MoltenVK. On OpenGL ES, the sampler parameter for setting LOD bias (``GL_TEXTURE_LOD_BIAS``) is unsupported, unlike desktop OpenGL.

This causes Midna's iris to only appear when the camera is very close to her.

While manual texture sampling does solve this issue, it can be less performant at higher resolutions and on low-power platforms (i.e. Android). As a separate workaround, we can pass through a bias value in the ``texture()`` call in the fragment shader, which both OpenGL ES and Metal support.

Fixes https://bugs.dolphin-emu.org/issues/11568 and https://bugs.dolphin-emu.org/issues/11993..

<details>
<summary>Screenshots (macOS)</summary>

Broken:

image

Fixed:

image

</details>

### DIFF
--- a/Source/Core/VideoBackends/OGL/OGLRender.cpp
+++ b/Source/Core/VideoBackends/OGL/OGLRender.cpp
@@ -513,6 +513,9 @@ Renderer::Renderer(std::unique_ptr<GLContext> main_gl_context, float backbuffer_
     // ARB_get_texture_sub_image (unlikely, except maybe on NVIDIA), we can use that instead.
     g_Config.backend_info.bSupportsDepthReadback = g_ogl_config.bSupportsTextureSubImage;
 
+    // GL_TEXTURE_LOD_BIAS is not supported on GLES.
+    g_Config.backend_info.bSupportsLodBiasInSampler = false;
+
     if (GLExtensions::Supports("GL_EXT_shader_framebuffer_fetch"))
     {
       g_ogl_config.SupportedFramebufferFetch = EsFbFetchType::FbFetchExt;
--- a/Source/Core/VideoCommon/PixelShaderGen.cpp
+++ b/Source/Core/VideoCommon/PixelShaderGen.cpp
@@ -648,7 +648,19 @@ uint WrapCoord(int coord, uint wrap, int size) {{
               "  float3 coords = float3(float(uv.x) / size_s, float(uv.y) / size_t, layer);\n");
     if (api_type == APIType::OpenGL || api_type == APIType::Vulkan)
     {
-      out.Write("  return iround(255.0 * texture(tex, coords));\n}}\n");
+      if (!host_config.backend_sampler_lod_bias)
+      {
+        out.Write("  uint texmode0 = samp_texmode0(texmap);\n"
+                  "  float lod_bias = {} / 256.0f;\n"
+                  "  return iround(255.0 * texture(tex, coords, lod_bias));\n",
+                  BitfieldExtract<&SamplerState::TM0::lod_bias>("texmode0"));
+      }
+      else
+      {
+        out.Write("  return iround(255.0 * texture(tex, coords));\n");
+      }
+
+      out.Write("}}\n");
     }
     else if (api_type == APIType::D3D)
     {
--- a/Source/Core/VideoCommon/ShaderGenCommon.h
+++ b/Source/Core/VideoCommon/ShaderGenCommon.h
@@ -172,6 +172,7 @@ union ShaderHostConfig
   BitField<23, 1, bool, u32> enable_validation_layer;
   BitField<24, 1, bool, u32> manual_texture_sampling;
   BitField<25, 1, bool, u32> manual_texture_sampling_custom_texture_sizes;
+  BitField<26, 1, bool, u32> backend_sampler_lod_bias;
 
   static ShaderHostConfig GetCurrent();
 };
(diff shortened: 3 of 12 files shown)


pizuz:
Works as intended.

Source/Core/VideoBackends/OGL/OGLRender.cpp
```diff
@@ -513,6 +513,9 @@ Renderer::Renderer(std::unique_ptr<GLContext> main_gl_context, float backbuffer_
     // ARB_get_texture_sub_image (unlikely, except maybe on NVIDIA), we can use that instead.
     g_Config.backend_info.bSupportsDepthReadback = g_ogl_config.bSupportsTextureSubImage;
 
+    // GL_TEXTURE_LOD_BIAS is not supported on GLES.
+    g_Config.backend_info.bSupportsLodBiasInSampler = false;
+
```
Pokechu22, line 518: It looks like you're only ever setting it to false for GLES; you're never setting it to true for regular OpenGL.  That's probably the cause of the fifoci differences on ogl-lin-mesa.

OatmealDome: I believe I set it to true by default in ``OGLMain.cpp``?

Pokechu22: Ah, yes, I missed that.  And FifoCI does seem to be using regular OpenGL (see [the build log](https://dolphin.ci/#/builders/16/builds/5337/steps/7/logs/stdio), which says `Dolphin [refs/pull/10292/head] 5.0-15686 | FifoPlayer SC | OpenGL | HLE | FPS: 1 - VPS: 1 - 0%` instead of `OpenGL ES`).

Source/Core/VideoCommon/PixelShaderGen.cpp
```diff
@@ -648,7 +648,18 @@ uint WrapCoord(int coord, uint wrap, int size) {{
...
+        out.Write("  uint texmode0 = samp_texmode0(texmap);\n"
+                  "  float lod_bias = {} / 256.0f;\n",
+                  BitfieldExtract<&SamplerState::TM0::lod_bias>("texmode0"));
+      }
+      else
+      {
+        out.Write("  float lod_bias = 0.0f;\n");
+      }
+
+      out.Write("  return iround(255.0 * texture(tex, coords, lod_bias));\n}}\n");
```
Pokechu22, line 662: I feel like it would be cleaner to generate one of two different texture calls, one that has the bias and one that doesn't, instead of declaring `lod_bias` in two places.  There would still be the closing bracket, which needs to happen in both cases, but keeping a separate `Write` call for that after the if statement would work.

Source/Core/VideoCommon/ShaderGenCommon.h
```diff
@@ -169,9 +169,10 @@ union ShaderHostConfig
   BitField<20, 1, bool, u32> backend_shader_framebuffer_fetch;
   BitField<21, 1, bool, u32> backend_logic_op;
   BitField<22, 1, bool, u32> backend_palette_conversion;
-  BitField<23, 1, bool, u32> enable_validation_layer;
-  BitField<24, 1, bool, u32> manual_texture_sampling;
-  BitField<25, 1, bool, u32> manual_texture_sampling_custom_texture_sizes;
+  BitField<23, 1, bool, u32> backend_sampler_lod_bias;
+  BitField<24, 1, bool, u32> enable_validation_layer;
+  BitField<25, 1, bool, u32> manual_texture_sampling;
+  BitField<26, 1, bool, u32> manual_texture_sampling_custom_texture_sizes;
```
Pokechu22, line 175: The order of these fields is important (they're reflected in the shader cache filenames, e.g. `D3D-specialized-pipeline-00000000-61FF40.cache`/`D3D-specialized-pipeline-00000000-61FF60.cache`/`D3D-specialized-pipeline-00000000-E1FF40.cache`).  If you add a new field, you should put it at the end.

OatmealDome: Fixed, thanks... Thought this order didn't matter and tried to keep backend stuff together

Pokechu22 (approved):

brujo5:
someone posted this in forum

Version 5.0-15676 fixed the problem of lack of shadows in Rayman 3 in OpenGL ES, but in version 5.0-15706, which is part 2 of this fix in this API, games break when using OpenGL ES and this message appears as this image capture:

image https://forums.dolphin-emu.org/attachment.php?aid=19950

https://forums.dolphin-emu.org/Thread-games-crashes-on-opengl-es-from-5-0-15706?pid=523231#pid523231

OatmealDome:
https://github.com/dolphin-emu/dolphin/pull/10314 should fix this crash. Sorry about that.

JosJuice:
Already fixed. See PR #10314

brujo5:
working  but for some reason hybrid shader not working with opengl

JosJuice:
That issue is being tracked as https://bugs.dolphin-emu.org/issues/12778

brujo5:
hybrid shader work for me with vulkan

PR #10293, opened 2021-12-25
Config: Port CPU overclock setting to new config.

AdmiralCurtiss:
Mostly so I can get a review on if this is the correct way to do it to avoid losing perf in hot code, in particular the config change callback.

### DIFF
--- a/Source/Core/Core/CoreTiming.cpp
+++ b/Source/Core/Core/CoreTiming.cpp
@@ -78,6 +78,10 @@ Globals g;
 
 static EventType* s_ev_lost = nullptr;
 
+static size_t s_registered_config_callback_id;
+static float s_config_OC_factor;
+static float s_config_OC_inv_factor;
+
 static void EmptyTimedCallback(u64 userdata, s64 cyclesLate)
 {
 }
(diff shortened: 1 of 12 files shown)


Source/Core/Core/CoreTiming.cpp
```diff
@@ -148,6 +151,18 @@ void Shutdown()
...
+void RefreshConfig()
+{
+  if (!s_has_registered_config_callback)
+  {
+    Config::AddConfigChangedCallback([]() { Core::RunAsCPUThread([]() { RefreshConfig(); }); });
+    s_has_registered_config_callback = true;
+  }
+
+  s_config_OC_factor =
+      Config::Get(Config::MAIN_OVERCLOCK_ENABLE) ? Config::Get(Config::MAIN_OVERCLOCK) : 1.0f;
```
JosJuice, line 163: Perhaps consider also having an `s_config_OC_factor_inverted` to let us skip a division in `Advance`. (Maybe not all that important, but hey, Advance does run fairly often...)

JosJuice (commented):
The implementation seems reasonable. But have we tested that there actually is any noticeable performance loss when doing this without the second commit?

AdmiralCurtiss:
re perf, as always this is very unscientific but idling in the Melee main menu I get roughly:

before: ~390 FPS
without last commit: ~370 FPS
with last commit: ~390 FPS

JosJuice:
Alright, it definitely sounds like we should go with the caching then, considering it's a fairly simple thing to add.

JosJuice (approved):

PR #10294, opened 2021-12-26
Claim USB interface before using it

devfonks:
As the commit message states:
```
I've been having this strange problem recently, where my GC Adapter (of
the Hyperkin sort, in WiiU mode) has suddenly stopped working when it used
to work before. I'm not sure of the exact point it stopped, but it's been
an intermittent (and now persistent) problem for a while.

I've tried tweaking my udev rule, checking libusb versions, the works.

Then, I noticed in the CI log in Dolphin itself, as long as my adapter
was plugged in, it was vomiting out a bunch of these lines in a loop:

08:08:860 InputCommon/GCAdapter.cpp:90 E[CI]: adapter libusb read failed: err=LIBUSB_ERROR_TIMEOUT
08:08:868 InputCommon/GCAdapter.cpp:90 E[CI]: adapter libusb read failed: err=LIBUSB_ERROR_IO

Then, I checked dmesg, and lo and behold:

[47380.057244] usb 3-3: usbfs: process 72071 (GC Adapter Scan) did not claim interface 0 before use

So this is an attempt to fix this problem. I'm not familiar with this
project's source at all and this may well break other USB functions.
Reprimand me as necessary, I'd much rather learn the hard way than have
it go missed.

Signed-off-by: devfonks <nkeck72@netzero.net>
```
The commit I've made fixes one of the two problems. The kernel log message has gone away, so I've at least managed to make the USB handling code a little nicer to the kernel and libusb. However, I still am having USB troubles with my own device. Since that's irrelevant to this I'll leave those details out (this is a PR and not an issue report), but in a sort of happy-accident way I've managed to fix a problem that may or may not have actually been a problem. :)

To do this, I've simply moved the interface claim above to directly after the kernel driver check. This places the claim before the first use, which is where the Nyko fix is. I assume that's okay? Do the Nyko adapters need that hack before being claimed for some reason?


Pokechu22:
@dolphin-emu-bot rebuild

Source/Core/InputCommon/GCAdapter.cpp
```diff
@@ -322,14 +332,7 @@ static bool CheckDeviceAccess(libusb_device* device)
     return false;
   }
 
-  ret = libusb_claim_interface(s_handle, 0);
```
leoetlino, line 325: It looks like the if block just above this one ("this split is needed so that we don't avoid claiming the interface when...") should also be moved.

leoetlino (changes requested):
The second commit should be squashed into the first one.

leoetlino:
We're supposed to claim an interface before doing any kind of I/O according to the libusb docs, so this change seems reasonable to me (even though libusb and/or the kernel may try to auto-claim as a failsafe). Claiming an interface does not send any request to the device so I don't think this will break anything.

Source/Core/InputCommon/GCAdapter.cpp
devfonks: Gotcha, will update and push. Currently away from my desktop so will take a bit to get here.

devfonks: In the latest commit, I've moved that condition check to immediately after the kernel detach would happen. This should guard the interface claim from being avoided like the comment describes, unless I misread something.

I'm buildtesting as I speak, since the bot doesn't autotrust :)

devfonks: After latest commit, the build goes fine on my end. I assume that where I've moved this to will work, so I'll resolve, but if it needs to be someplace else feel free to drop another comment.

devfonks:
I squashed those commits as requested, but I didn't realize that github let you do that on the web interface. I did this on the command line via reset, squash, and merge; so the requested changes won't auto-resolve. d'oh.

leoetlino:
It looks like the squashing went wrong because there are still 3 commits in your PR.

Could you try:

```
git remote add upstream https://github.com/dolphin-emu/dolphin
git fetch upstream
git rebase -i upstream/master
```

You should see an editor pop up with something like:

```
pick 41791e21d Claim USB interface before using it
pick xxxxxxxxxx ...................
pick xxxxxxxxxx ...................
```

Change every "pick" into "f" except for your first commit ("Claim USB interface before using it").

devfonks:
Will do when I get back to my computer.

devfonks:
I forgot I had this open and it looks like it got fixed upstream anyway (or else something else is creating a merge conflict).

PR #10296, opened 2021-12-26
Config: Port GameList settings to new config system.

AdmiralCurtiss:


Source/Core/Core/Config/MainSettings.cpp
```diff
@@ -202,4 +202,50 @@ const Info<bool> MAIN_ANALYTICS_ENABLED{{System::Main, "Analytics", "Enabled"},
...
+const Info<bool> MAIN_GAMELIST_LIST_UNKNOWN{{System::Main, "GameList", "ListUnknown"}, true};
+const Info<int> MAIN_GAMELIST_LIST_SORT{{System::Main, "GameList", "ListSort"}, 3};
+const Info<int> MAIN_GAMELIST_LIST_SORT_SECONDARY{{System::Main, "GameList", "ListSortSecondary"},
+                                                  0};
+const Info<bool> MAIN_GAMELIST_COLUMN_PLATFORM{{System::Main, "GameList", "ColumnPlatform"}, true};
+const Info<bool> MAIN_GAMELIST_COLUMN_DESCRIPTION{{System::Main, "GameList", "ColumnDescription"},
+                                                  false};
+const Info<bool> MAIN_GAMELIST_COLUMN_BANNER{{System::Main, "GameList", "ColumnBanner"}, true};
+const Info<bool> MAIN_GAMELIST_COLUMN_TITLE{{System::Main, "GameList", "ColumnTitle"}, true};
+const Info<bool> MAIN_GAMELIST_COLUMN_MAKER{{System::Main, "GameList", "ColumnNotes"}, true};
```
JosJuice, line 235: Something went wrong here.

JosJuice: Ah, no, the maker column and the notes column are in fact the same thing, and this is how the naming worked in the old config too... Ah well, let's leave it like this.

JosJuice (approved):

Source/Core/Core/Config/MainSettings.cpp
AdmiralCurtiss: Yeah I got tripped by this too for a second, not sure why it's like this but we tend to avoid changing INI keys for compatibilty, so I guess we'll leave it...

PR #10299, opened 2021-12-27
GDB Stub: add support for various query packets

aldelaro5:
These tends to get requested from either pure GDB or Ghidra. They reduce the verbosity of the communications. The QSupported packet is also important to implement for future proofing too.

### DIFF
--- a/Source/Core/Core/PowerPC/GDBStub.cpp
+++ b/Source/Core/Core/PowerPC/GDBStub.cpp
@@ -307,12 +307,20 @@ static void SendReply(const char* reply)
 
 static void HandleQuery()
 {
-  DEBUG_LOG_FMT(GDB_STUB, "gdb: query '{}'", CommandBufferAsString() + 1);
+  DEBUG_LOG_FMT(GDB_STUB, "gdb: query '{}'", CommandBufferAsString());
 
-  if (!strcmp((const char*)(s_cmd_bfr + 1), "TStatus"))
-  {
-    return SendReply("T0");
-  }
+  if (!strncmp((const char*)(s_cmd_bfr), "qAttached", strlen("qAttached")))
+    return SendReply("1");
+  if (!strcmp((const char*)(s_cmd_bfr), "qC"))
+    return SendReply("QC1");
+  if (!strcmp((const char*)(s_cmd_bfr), "qfThreadInfo"))
+    return SendReply("m1");
+  else if (!strcmp((const char*)(s_cmd_bfr), "qsThreadInfo"))
+    return SendReply("l");
+  else if (!strncmp((const char*)(s_cmd_bfr), "qThreadExtraInfo", strlen("qThreadExtraInfo")))
+    return SendReply("00");
+  else if (!strncmp((const char*)(s_cmd_bfr), "qSupported", strlen("qSupported")))
+    return SendReply("swbreak+;hwbreak+");
 
   SendReply("");
 }


Source/Core/Core/PowerPC/GDBStub.cpp
```diff
@@ -309,10 +309,18 @@ static void HandleQuery()
 {
   DEBUG_LOG_FMT(GDB_STUB, "gdb: query '{}'", CommandBufferAsString() + 1);
 
-  if (!strcmp((const char*)(s_cmd_bfr + 1), "TStatus"))
-  {
-    return SendReply("T0");
-  }
```
Pokechu22, line 315: Did you mean to get rid of this?  It seems to be a valid packet ([`qTStatus`](https://sourceware.org/gdb/onlinedocs/gdb/Tracepoint-Packets.html#index-qTStatus-packet))

Source/Core/Core/PowerPC/GDBStub.cpp
```diff
@@ -309,10 +309,18 @@ static void HandleQuery()
...
+  if (!strcmp((const char*)(s_cmd_bfr + 1), "Attached"))
+    return SendReply("1");
```
Pokechu22, line 313: The [documentation](https://sourceware.org/gdb/onlinedocs/gdb/General-Query-Packets.html#index-qAttached-packet) says that the syntax is <code>qAttached:<var>pid</var></code>.  To my understanding, `strcmp` returns a nonzero value if the strings are of different lengths (i.e. `strcmp("qAttached:1", "qAttached")` is nonzero).  So I'm not sure if this will do anything.  You seem to use `strncmp` for this below; perhaps it should be used here?

Source/Core/Core/PowerPC/GDBStub.cpp
```diff
@@ -309,10 +309,18 @@ static void HandleQuery()
...
+  if (!strcmp((const char*)(s_cmd_bfr + 1), "C"))
+    return SendReply("QC1");
```
Pokechu22, line 315: The [documentation](https://sourceware.org/gdb/onlinedocs/gdb/General-Query-Packets.html#index-qC-packet) says that the syntax is <code>QC <var>thread-id</var></code>, but it also says near the top:

> We include spaces in some of the templates for clarity; these are not part of the packet’s syntax. No GDB packet uses spaces to separate its components.

I assume that this is correct, but I'd like to make sure.

Source/Core/Core/PowerPC/GDBStub.cpp
```diff
@@ -309,10 +309,18 @@ static void HandleQuery()
...
+  if (!strcmp((const char*)(s_cmd_bfr + 1), "fThreadInfo"))
+    return SendReply("m1");
+  else if (!strcmp((const char*)(s_cmd_bfr + 1), "sThreadInfo"))
+    return SendReply("l");
+  else if (!strncmp((const char*)(s_cmd_bfr + 1), "ThreadExtraInfo", 15))
+    return SendReply("00");
```
Pokechu22, line 321: This looks correct per [the documentation](https://sourceware.org/gdb/onlinedocs/gdb/General-Query-Packets.html#index-qfThreadInfo-packet), but Dolphin does actually have a way of viewing the thread list for a running game (see view &rarr; threads).  I think it'd be useful to include this there (with [`qThreadExtraInfo`](https://sourceware.org/gdb/onlinedocs/gdb/General-Query-Packets.html#index-qThreadExtraInfo-packet) showing the info in the table), though that doesn't need to happen in this PR.  Perhaps if that's done, the thread's address would be used as its thread ID, but I don't really know how threads are implemented.

Source/Core/Core/PowerPC/GDBStub.cpp
```diff
@@ -309,10 +309,18 @@ static void HandleQuery()
...
+  else if (!strncmp((const char*)(s_cmd_bfr + 1), "Supported", static_cast<size_t>(9)))
+    return SendReply("swbreak+;hwbreak+");
```
Pokechu22, line 323: The implementation looks correct per [the documentation](https://sourceware.org/gdb/onlinedocs/gdb/General-Query-Packets.html#index-qSupported-packet), but I don't think the `static_cast<size_t>(9)` part makes sense; using just `9` is fine.

Source/Core/Core/PowerPC/GDBStub.cpp
```diff
@@ -309,10 +309,18 @@ static void HandleQuery()
 {
   DEBUG_LOG_FMT(GDB_STUB, "gdb: query '{}'", CommandBufferAsString() + 1);
```
Pokechu22, line 310: Just as an opinion thing, I feel it would be clearer to not get rid of the first character (both here and in the `strcmp` calls) to make it more obvious what the command names are for searching in the documentation.

Source/Core/Core/PowerPC/GDBStub.cpp
aldelaro5: This is correct, you actually see this nomenclature everywhere.

Btw, you can actually log from GDB's side what it sends and receive by using `set debug remote on` before connecting which can be done with `target remote localhost:port` where `port` is the one you set in the INI. (you might want to do `set architecture powerpc:750` and `set endian big` before though)

Source/Core/Core/PowerPC/GDBStub.cpp
aldelaro5: Should I do an strlen instead?

Realized it looked a little ugly (the size_t cast was to fix some warnings I THINK)

Source/Core/Core/PowerPC/GDBStub.cpp
aldelaro5: I did because TStatus is about a feature called tracepoint which we do not support.

Source/Core/Core/PowerPC/GDBStub.cpp
Pokechu22: I see other places using it with just an int, presumably without warnings.  I don't think strlen is necessary (but it [can be inlined at compile time](https://stackoverflow.com/a/26021702), so there isn't a performance cost, though the string would have to be duplicated?)

Source/Core/Core/PowerPC/GDBStub.cpp
aldelaro5: So the pid seems to be optional because it sends `qAttached` on my end, but since we are never going to create a new process on attach, I might as well do it sure.

Source/Core/Core/PowerPC/GDBStub.cpp
aldelaro5: Thanks for telling me, it won't happen in this pr because to put it simply: it's not that simple, there's a WHOLE HOST of complications I have to figure out if we start to support multithreading in gdb

Source/Core/Core/PowerPC/GDBStub.cpp
Pokechu22: Ah, the documentation does explicitly say that:

> When the multiprocess protocol extensions are supported (see multiprocess extensions), <var>pid</var> is an integer in hexadecimal format identifying the target process. Otherwise, GDB will omit the <var>pid</var> field and the query packet will be simplified as ‘`qAttached`’.

I think it is appropriate to respond with an error if a PID is specified, then.

aldelaro5: Well, we never reply with multiprocess+ in qSupported, I don't think adding the if for that is worth when if that even happens, the problem would be in qSupported.

Source/Core/Core/PowerPC/GDBStub.cpp
Pokechu22: Ah, I guess it does make sense to remove it then.

Pokechu22 (approved):
Everything looks good here, but I haven't done any testing.

PR #10300, opened 2021-12-27
GDB Stub: fix thread report requests

aldelaro5:
We are always reporting thread 1 existing and also fix an issue with a wrong memcmp size.

### DIFF
--- a/Source/Core/Core/PowerPC/GDBStub.cpp
+++ b/Source/Core/Core/PowerPC/GDBStub.cpp
@@ -327,16 +327,16 @@ static void HandleQuery()
 
 static void HandleSetThread()
 {
-  if (memcmp(s_cmd_bfr, "Hg0", 3) == 0 || memcmp(s_cmd_bfr, "Hc-1", 4) == 0 ||
-      memcmp(s_cmd_bfr, "Hc0", 3) == 0 || memcmp(s_cmd_bfr, "Hc1", 3) == 0)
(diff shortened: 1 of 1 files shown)


Source/Core/Core/PowerPC/GDBStub.cpp
```diff
@@ -319,16 +319,14 @@ static void HandleQuery()
 
 static void HandleSetThread()
 {
-  if (memcmp(s_cmd_bfr, "Hg0", 3) == 0 || memcmp(s_cmd_bfr, "Hc-1", 4) == 0 ||
-      memcmp(s_cmd_bfr, "Hc0", 3) == 0 || memcmp(s_cmd_bfr, "Hc1", 3) == 0)
+  if (memcmp(s_cmd_bfr, "Hc-1", 4) == 0 || memcmp(s_cmd_bfr, "Hc1", 3) == 0)
```
leoetlino, line 322: Can you explain why Hc0 was removed?

> Several packets and replies include a thread-id field to identify a thread. Normally these are positive numbers with a target-specific interpretation, formatted as big-endian hex strings. A thread-id can also be a literal ‘-1’ to indicate all threads, or ‘0’ to pick any thread.

Is "any thread" not a valid option?

aldelaro5: huh, I actually didn't read that part, I assumed 0 was wrong because gdb wasn't actually inquiring about it, but if the specs says it's fine, sure, I'll put it back (also I actually thought this was a dash and not a -1...oops)

PR #10301, opened 2021-12-27
BBA: Fix TAP detection

RA-Kooi:
Current code expects the component ID of the tap adapter to be equal to `tap0901`, however on my machine (Windows 10, latest TAP adapter) the component ID is `root\tap0901`. This should be a fix that is backwards compatible with "old"(?) setups.

### DIFF
--- a/Source/Core/Core/HW/EXI/BBA/TAP_Win32.cpp
+++ b/Source/Core/Core/HW/EXI/BBA/TAP_Win32.cpp
@@ -66,11 +66,16 @@ bool IsTAPDevice(const TCHAR* guid)
 
         if (status == ERROR_SUCCESS && data_type == REG_SZ)
         {
-          if (!_tcscmp(component_id, TAP_COMPONENT_ID) && !_tcscmp(net_cfg_instance_id, guid))
+          TCHAR* const component_id_sub = _tcsstr(component_id, TAP_COMPONENT_ID);
+
+          if (component_id_sub)
           {
-            RegCloseKey(unit_key);
-            RegCloseKey(netcard_key);
-            return true;
+            if (!_tcscmp(component_id_sub, TAP_COMPONENT_ID) && !_tcscmp(net_cfg_instance_id, guid))
+            {
+              RegCloseKey(unit_key);
+              RegCloseKey(netcard_key);
+              return true;
+            }
           }
         }
       }


Source/Core/Core/HW/EXI/BBA/TAP_Win32.cpp
```diff
@@ -31,6 +31,7 @@ bool IsTAPDevice(const TCHAR* guid)
     HKEY unit_key;
     TCHAR component_id_string[] = _T("ComponentId");
     TCHAR component_id[256];
+    TCHAR* component_id_sub;
```
Warepire, line 34: This can be declared within the if-statement that uses it.

Source/Core/Core/HW/EXI/BBA/TAP_Win32.cpp
```diff
@@ -66,11 +67,16 @@ bool IsTAPDevice(const TCHAR* guid)
 
         if (status == ERROR_SUCCESS && data_type == REG_SZ)
         {
-          if (!_tcscmp(component_id, TAP_COMPONENT_ID) && !_tcscmp(net_cfg_instance_id, guid))
+          component_id_sub = _tcsstr(component_id, TAP_COMPONENT_ID);
+
+          if(component_id_sub)
```
Warepire, line 72: Missing space before the `(`, did you run clang-format? https://github.com/dolphin-emu/dolphin/blob/master/Contributing.md#checking-and-fixing-formatting-issues

Warepire (commented):
Logic wise this looks fine to me, some minor comments.

Source/Core/Core/HW/EXI/BBA/TAP_Win32.cpp
Pokechu22: Note that if you use Visual Studio, you can press Control+K followed by Control+D to auto-format the current document using the clang-format rules (without needing to install clang-format separately).

Source/Core/Core/HW/EXI/BBA/TAP_Win32.cpp
RA-Kooi: That was my initial idea, but seeing everything defined at the top level of each function scope, I figured it was part of the style.

Source/Core/Core/HW/EXI/BBA/TAP_Win32.cpp
RA-Kooi: I use vim, but let me fix this real quick.

JosJuice:
@dolphin-emu-bot rebuild

Source/Core/Core/HW/EXI/BBA/TAP_Win32.cpp
RA-Kooi: I declared it within the if-statement now.

RA-Kooi:
I fixed all comments.

JosJuice:
@dolphin-emu-bot rebuild

PR #10302, opened 2021-12-27
DolphinQt: Fix generation of /external:I flags

RA-Kooi:
Mainly concerns to building with Ninja, as that's what I tested it with.
Originally it would only prepend the first path with `/external:I`, however all paths in the list have to be prepended with `/external:I`.
The MS documentation seems to support this, as it makes no mention of it accepting a list.

This is probably the worst way to implement this, I am unfamiliar with CMake.  
As such you probably don't want to merge this, but I'll push it as a reminder for whoever is going to work on CMake support for Windows.


iwubcode (approved):

Pokechu22:
@dolphin-emu-bot rebuild

PR #10303, opened 2021-12-27
RenderWidget: Add include to Windows.h

RA-Kooi:
Fixes errors when building with CMake (Ninja).  
Currently it depends on PCH support, but it's broken currently.  
Not depending on PCH is good practice anyway.

Not sure if `Windows.h` is the most lightweight include though, but it at least fixes the error of there not being a target defined.

### DIFF
--- a/Source/Core/DolphinQt/RenderWidget.cpp
+++ b/Source/Core/DolphinQt/RenderWidget.cpp
@@ -36,8 +36,7 @@
 #include "VideoCommon/VideoConfig.h"
 
 #ifdef _WIN32
-#include <WinUser.h>
-#include <windef.h>
+#include <Windows.h>
 #endif
 
 RenderWidget::RenderWidget(QWidget* parent) : QWidget(parent)


Pokechu22:
@dolphin-emu-bot rebuild

Pokechu22:
The specific error:

```
9>RenderWidget.cpp
9>C:\Program Files (x86)\Windows Kits\10\Include\10.0.22000.0\um\winnt.h(169,1): fatal error C1189: #error:  "No Target Architecture"
9>C:\Program Files (x86)\Windows Kits\10\Include\10.0.22000.0\um\winnt.h(169,1): fatal error C1189: #error "No Target Architecture"
9>C:\Program Files (x86)\Windows Kits\10\Include\10.0.22000.0\um\winnt.h(169,1): fatal error C1189: ^
```

According to [this SO post](https://stackoverflow.com/a/4845893) you can just include `windows.h` on its own, and that includes `windef.h`.  Note that Dolphin defines `WIN32_LEAN_AND_MEAN` so `windows.h` is much more lightweight.

As a related note, Dolphin seems to be inconsistent about specifying `windows.h` vs `Windows.h` (but that doesn't matter since Windows uses a case-insensitive filesystem; you can even use `wInDoWs.H`).  The current inconsistency isn't something you need to deal with, and it's probably best to stick with `Windows.h` since that file also uses `WinUser.h` instead of `winuser.h`.

Source/Core/DolphinQt/RenderWidget.cpp
```diff
@@ -37,6 +37,7 @@
 #include "VideoCommon/VideoConfig.h"
 
 #ifdef _WIN32
+#include <Windows.h>
 #include <WinUser.h>
```
leoetlino, line 41: Could you remove these two lines? As @Pokechu22 said, including `<Windows.h>` [automatically includes](https://en.wikipedia.org/wiki/Windows.h#Child_header_files) winuser.h and windef.h.

leoetlino (changes requested):

Source/Core/DolphinQt/RenderWidget.cpp
RA-Kooi: Fixed.

leoetlino (approved):

PR #10305, opened 2021-12-27
Config: Port FifoPlayer setting to new config system.

AdmiralCurtiss:
A quick and simple one.


JosJuice (approved):

Pokechu22 (approved):
LGTM.

It'd be nice to get this setting exposed on the UI somewhere (I occasionally use it with dump frames or the software renderer's dump objects function, where looping endlessly leads to an excessive amount of files; this was particularly relevant for debugging issues with the first frame of FIFO playback, though those are mostly solved now). That's something I could handle in a separate PR, though.

PR #10308, opened 2021-12-27
GDB Stub: validate the address exists before reading/writting to it

aldelaro5:
This prevented unknown pointer spams when you connect with ghidra because it seems to want to probe memory that happened to be invalid, this shouldn't happen, we should just return an error and not panic.

### DIFF
--- a/Source/Core/Core/PowerPC/GDBStub.cpp
+++ b/Source/Core/Core/PowerPC/GDBStub.cpp
@@ -536,9 +536,10 @@ static void ReadMemory()
 
   if (len * 2 > sizeof reply)
     SendReply("E01");
+
+  if (!PowerPC::HostIsRAMAddress(addr))
+    return SendReply("E00");
   u8* data = Memory::GetPointer(addr);
-  if (!data)
-    return SendReply("E0");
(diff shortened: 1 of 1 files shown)


Source/Core/Core/PowerPC/GDBStub.cpp
```diff
@@ -560,9 +561,9 @@ static void WriteMemory()
     len = (len << 4) | Hex2char(s_cmd_bfr[i++]);
   INFO_LOG_FMT(GDB_STUB, "gdb: write memory: {:08x} bytes to {:08x}", len, addr);
 
+  if (!PowerPC::HostIsRAMAddress(addr))
+    return SendReply("E0");
```
leoetlino, line 565: Shouldn't this be E00?

aldelaro5: yes, done.

lioncash: This still seems to be unaddressed

aldelaro5: > This still seems to be unaddressed

woops, done.

PR #10309, opened 2021-12-28
Gdb stub: fix watchpoints

aldelaro5:
This fixes 2 issues with watchpoints:

1: z4 packets would get ignored because of a bogus check for bogus types
2: read only watchpoint and write only watchpoint were reversed in the enum

### DIFF
--- a/Source/Core/Core/PowerPC/GDBStub.cpp
+++ b/Source/Core/Core/PowerPC/GDBStub.cpp
@@ -58,6 +58,8 @@ enum class BreakpointType
   Access,
 };
 
+constexpr u32 NUM_BREAKPOINT_TYPES = 4;
+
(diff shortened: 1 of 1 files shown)


Source/Core/Core/PowerPC/GDBStub.cpp
```diff
@@ -626,7 +626,7 @@ static void HandleRemoveBreakpoint()
   u32 type, addr, len, i;
 
   type = Hex2char(s_cmd_bfr[1]);
-  if (type >= 4)
+  if (type > 4)
```
leoetlino, line 629: Could you turn this magic number into a named constant? You could add `constexpr u32 NUM_BREAKPOINT_TYPES = 4;` right after the BreakpointType enum (and then use that constant in HandleAddBreakpoint too).

PR #10312, opened 2021-12-28
GameINI: Replace all uses of SyncGPU with single core

JosJuice:
At least on Android, SyncGPU has really bad performance compared to single core. It's less stable too – I get fatal GPU desyncs in Pokémon XD using the default settings.


JMC47:
SyncGPU is currently bugged and should probably be removed at this point.  It's just added complication.

lioncash (approved):

PR #10315, opened 2021-12-28
OGL: Use GL_EXT_texture_lod_bias on OpenGL ES if available

OatmealDome:
This adds support for [EXT_texture_lod_bias](https://www.khronos.org/registry/OpenGL/extensions/EXT/EXT_texture_lod_bias.txt) on OpenGL ES.

### DIFF
--- a/Source/Core/Common/GL/GLExtensions/EXT_texture_lod_bias.h
+++ b/Source/Core/Common/GL/GLExtensions/EXT_texture_lod_bias.h
@@ -1,0 +1,10 @@
+/*
+** Copyright (c) 2013-2015 The Khronos Group Inc.
+** SPDX-License-Identifier: MIT
+*/
+
+#include "Common/GL/GLExtensions/gl_common.h"
+
+#define GL_TEXTURE_FILTER_CONTROL_EXT 0x8500
+#define GL_TEXTURE_LOD_BIAS_EXT 0x8501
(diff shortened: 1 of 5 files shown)


Source/Core/VideoCommon/PixelShaderGen.cpp
```diff
@@ -648,14 +648,14 @@ uint WrapCoord(int coord, uint wrap, int size) {{
               "  float3 coords = float3(float(uv.x) / size_s, float(uv.y) / size_t, layer);\n");
     if (api_type == APIType::OpenGL || api_type == APIType::Vulkan)
     {
-      if (!host_config.backend_sampler_lod_bias)
+      /*if (!host_config.backend_sampler_lod_bias)
```
OatmealDome, line 651: The shader fallback path has been disabled for testing... Will drop this when ready

JMC47:
It doesn't seem like any of our phones actually support the extension to test it lol.

OatmealDome:
Was Jos's phone supported? I think they said their screen matched the screenshots taken on macOS, but maybe I misunderstood

JosJuice:
My phone isn't supported. What I meant is that this PR matches the known broken rendering and master matches the known good rendering.

OatmealDome:
Closing because of poor device support and minimal performance difference

PR #10316, opened 2021-12-28
UICommon: Append trailing slash to DOLPHIN_EMU_USERPATH

OatmealDome:
This fixes a really annoying issue. If you define a path in ``DOLPHIN_EMU_USERPATH`` and don't put a trailing slash, all the users directories will be created incorrectly.

For example, if ``DOLPHIN_EMU_USERPATH`` is ``/path/to/user``, folders like ``/path/to/userWii`` and ``/path/to/userConfig`` are created.

To fix this, I append a ``DIR_SEP`` to the variable in case it isn't suffixed by one already.

### DIFF
--- a/Source/Core/UICommon/UICommon.cpp
+++ b/Source/Core/UICommon/UICommon.cpp
@@ -308,7 +308,7 @@ void SetUserDirectory(const std::string& custom_path)
 #if defined(__APPLE__) || defined(ANDROID)
     if (env_path)
     {
-      user_path = env_path;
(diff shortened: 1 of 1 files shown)


JosJuice:
You should probably make the same adjustment to the Linux path.

OatmealDome:
I missed that, thanks.

AdmiralCurtiss:
I think this is also fixed by #10241?

leoetlino:
Closing because this should be fixed by #10241 -- we now always normalize trailing slashes.

PR #10319, opened 2021-12-29
WIP: Gdb stub target descriptor

aldelaro5:
DO NOT MERGE (blocked by #10298 and #10299 , I will rebase to master once these are merged, but rn, this is based on working versions of these prs).

This adds support for gdb's xml target descriptor which allows us to define explicitly the layout of the registers. In particular:

- Change the view of the FPR so that we can see both ps field OR the entire thing as a double
- Add the registers exclusive to this variant of PowerPC
- Set the architecture and the osabi

In theory, the only 2 things you will need to setup before connecting after this is setting the endian to big (which sadly, this isn't supported by gdb, it's pulled from the symbols and even if you load some like I did with a ghidra script, you have to RESET IT to auto for it to pick up the new endianness) and define a dummy proc mapping (which we can address by supporting qXfer:memory-map:read so gdb will know the actual mapping).

NOTE: this has issues with Ghidra because it doesn't seem it's able to understand the view that I put with the FPR. on GDB, this shows up like this:

``` f0             {ps = {ps0 = 0x0, ps1 = 0x0}, double = 0x0} {ps = {ps0 = 0, ps1 = 0}, double = 0}``` (the first part is raw, the second part is formatted)

I'll try to see how I can address the Ghidra problem before removing the wip tag.

### CODE REVIEW

File: Source/Core/Core/PowerPC/GDBStub.cpp [hunk 1]
```diff
@@ -320,8 +565,10 @@ static void HandleQuery()
   else if (!strncmp((const char*)(s_cmd_bfr), "qThreadExtraInfo", strlen("qThreadExtraInfo")))
     return SendReply("00");
   else if (!strncmp((const char*)(s_cmd_bfr), "qSupported", strlen("qSupported")))
-    return SendReply("swbreak+;hwbreak+");
-
+    return SendReply("swbreak+;hwbreak+;qXfer:features:read+");
+  else if (!strncmp((const char*)(s_cmd_bfr),
+                    "qXfer:features:read:target.xml:", strlen("qXfer:features:read:target.xml:")))
+    ProcessXFerCommand(target_xml);
```
BhaaLseN, line 571: One one hand: the condition spans multiple lines, so our code style dictates you add curlies (both here and everywhere else).

On the other hand though: Thoughts on moving this to a helper method, where the string isn't duplicated? Getting something wrong on the `strlen` call is fairly easy and even the `strcmp` ones are probably better off using `strncmp` (or any appropriate equivalent) anyways.

(plus it might get the line length down enough so the curlies don't have to be there.)

aldelaro5: The thing is that I was in progress of adding pr #10321 which WILL need the curly anyway for other reasons (I realised they both are close enough that they should be the same pr, but 2 different commits)

I'll do the helper thing though, but just know that the pr isn't yet ready for review, I'll move it out of draft when it is

Source/Core/Core/PowerPC/GDBStub.cpp
```diff
@@ -305,23 +521,90 @@ static void SendReply(const char* reply)
...
+  else if (IsQuery("qXfer:features:read:target.xml:"))
+  {
+    ProcessXFerCommand(target_xml, strlen("qXfer:features:read:target.xml:"));
+  }
+  else if (IsQuery("qXfer:memory-map:read::"))
+  {
+    const char* memoryMapXml = target_memory_map_xml_noMMU;
+    if (SConfig::GetInstance().bMMU)
+      memoryMapXml = target_memory_map_xml_withMMU;
+    ProcessXFerCommand(memoryMapXml, strlen("qXfer:memory-map:read::"));
```
Warepire, line 606: Should the query commands be constants? `qXfer:memory-map:read::` is a rather complicated string and used twice, which may or may not be a problem in the future. (easy to forget changing one but not the other for example).

aldelaro5: We could just constantize the xfer ones, they are especially long and checking other queries, it's probably the longest we'll deal with. Is that ok?

Warepire: Works for me!

aldelaro5:
Had to force push to fix sending an empty reply even if it's sucessful.

Source/Core/Core/PowerPC/GDBStub.cpp
```diff
@@ -66,6 +67,224 @@ constexpr int MACH_O_POWERPC_750 = 9;
...
+    "<memory-map version=\"1.0\">"
+    "<memory type=\"ram\" start=\"0x7E000000\" length=\"0x2000000\"/>"
+    "<memory type=\"ram\" start=\"0x80000000\" length=\"0x1800000\"/>"
+    "<memory type=\"ram\" start=\"0x90000000\" length=\"0x4000000\"/>"
+    "</memory-map>";
+
+const char* TARGET_MEMORY_MAP_XML_WITH_MMU =
+    "<memory-map version=\"1.0\">"
+    "<memory type=\"ram\" start=\"0x0\" length=\"0x10000000\"/>"
+    "</memory-map>";
```
Pokechu22, line 83: Is this correct?  Based on [wiibrew](https://wiibrew.org/wiki/Memory_map) MEM1 physically covers `0x00000000`-`0x017FFFFF`, and MEM2 physically covers `0x10000000`-`0x13FFFFFF`.  This mapping seems like it won't cover MEM2 at all.

Source/Core/Core/PowerPC/GDBStub.cpp
```diff
@@ -66,6 +67,224 @@ constexpr int MACH_O_POWERPC_750 = 9;
...
 
+const char* QUERY_XFER_TARGET = "qXfer:features:read:target.xml:";
+const char* QUERY_XFER_MEMORY_MAP = "qXfer:memory-map:read::";
+
+const char* TARGET_MEMORY_MAP_XML_NO_MMU =
+    "<memory-map version=\"1.0\">"
+    "<memory type=\"ram\" start=\"0x7E000000\" length=\"0x2000000\"/>"
+    "<memory type=\"ram\" start=\"0x80000000\" length=\"0x1800000\"/>"
+    "<memory type=\"ram\" start=\"0x90000000\" length=\"0x4000000\"/>"
+    "</memory-map>";
```
Pokechu22, line 78: I think this is a good use for [raw string literals](https://en.cppreference.com/w/cpp/language/string_literal):

```C++
const char* TARGET_MEMORY_MAP_XML_NO_MMU =
    R"(<memory-map version="1.0">
<memory type="ram" start="0x7E000000" length="0x2000000"/>
<memory type="ram" start="0x80000000" length="0x1800000"/>
<memory type="ram" start="0x90000000" length="0x4000000"/>
</memory-map>)";
```

Note that this is not 100% equivalent, as raw string literals also include the newlines, but I think that's fine.

Source/Core/Core/PowerPC/GDBStub.cpp
```diff
@@ -320,25 +539,91 @@ static void WriteHostInfo()
...
   DEBUG_LOG_FMT(GDB_STUB, "gdb: query '{}'", CommandBufferAsString());
 
-  if (!strncmp((const char*)(s_cmd_bfr), "qAttached", strlen("qAttached")))
+  if (IsQuery("qAttached"))
+  {
     return SendReply("1");
-  if (!strcmp((const char*)(s_cmd_bfr), "qC"))
+  }
+  else if (IsQuery("qC"))
+  {
```
Pokechu22, line 593: Note that the old behavior here was checking if the string exactly matched `qC`, not if it started with `qC`.  I don't know if this is a problem.

PR #10321, opened 2021-12-29
WIP Gdb stub: memory map

aldelaro5:
DO NOT MERGE (blocked by #10299 , I will rebase to master once it's merged).

This allows gdb to get information about the memory mapping which can be seen with `info mem`. The mapping is more precise with MMU off and works for most cases while if MMU is on, it provides a default 32 bit mapping.

It was found with Ghidra to reduce the verbosity of the communication with MMU off which can be helpful.

As for why we provide a dummy mapping with MMU off, it's because it's not possible to tell GDB to refetch the mappings and the game can change them at any time. With MMU off however, it should work most of the time with some assumptions.


aldelaro5:
Closing since this is now in #10319

PR #10323, opened 2021-12-30
Config: Port Movie settings to new config system.

AdmiralCurtiss:


Source/Core/VideoCommon/RenderBase.cpp
```diff
@@ -1670,7 +1672,7 @@ void Renderer::FrameDumpThreadFunc()
 
         // Stop frame dumping if we fail to start.
         if (!frame_dump_started)
-          SConfig::GetInstance().m_DumpFrames = false;
+          Config::SetCurrent(Config::MAIN_MOVIE_DUMP_FRAMES, false);
```
AdmiralCurtiss, line 1675: This one I wasn't sure whether I should be using `SetBase()`, `SetCurrent()`, or `SetBaseOrCurrent()`. Probably doesn't matter too much though.

JosJuice: The behavior in master where it effectively sets the base layer is actually really confusing, because the checkbox in the GUI doesn't update until you restart Dolphin, and so if you want to restart emulation and retry dumping, you have to realize that you have to go and turn the framedumping checkbox off and then on again even though it looks like it's already on. So I agree that `SetCurrent` makes sense.

JosJuice (approved):

PR #10326, opened 2021-12-30
Config: Port BluetoothPassthrough settings to new config system.

AdmiralCurtiss:
This one is untested because I don't have a bluetooth adapter handy at the moment. Someone should test if it still works as expected.

### DIFF
--- a/Source/Core/Core/IOS/USB/Bluetooth/BTReal.cpp
+++ b/Source/Core/Core/IOS/USB/Bluetooth/BTReal.cpp
@@ -24,7 +24,7 @@
 #include "Common/Network.h"
 #include "Common/StringUtil.h"
 #include "Common/Swap.h"
-#include "Core/ConfigManager.h"
+#include "Core/Config/MainSettings.h"
 #include "Core/Core.h"
 #include "Core/HW/Memmap.h"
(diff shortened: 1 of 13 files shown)


Source/Core/Core/IOS/USB/Bluetooth/BTReal.cpp
```diff
@@ -556,7 +558,7 @@ void BluetoothRealDevice::SaveLinkKeys()
   std::string config_string = oss.str();
   if (!config_string.empty())
     config_string.pop_back();
-  SConfig::GetInstance().m_bt_passthrough_link_keys = config_string;
+  Config::SetBase(Config::MAIN_BLUETOOTHPASSTHROUGH_LINK_KEYS, config_string);
```
AdmiralCurtiss, line 561: Is `SetBase()` fine here or should this be `SetBaseOrCurrent()`? I can't really imagine a case where this is overridden by anything, but...

leoetlino: SetBase sounds fine to me. We always want this to be persisted.

leoetlino (approved):

PR #10327, opened 2021-12-30
Config: Port USBPassthrough setting to new config system.

AdmiralCurtiss:
Had to redesign this a bit due to the `std::set<std::pair<>>` nature of the setting... I think this is fine but I'm open to other ideas.

### DIFF
--- a/Source/Core/Core/ConfigManager.cpp
+++ b/Source/Core/Core/ConfigManager.cpp
@@ -91,7 +91,6 @@ void SConfig::SaveSettings()
   SaveGeneralSettings(ini);
   SaveInterfaceSettings(ini);
   SaveCoreSettings(ini);
-  SaveUSBPassthroughSettings(ini);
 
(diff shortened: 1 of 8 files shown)


Source/Core/Core/ConfigManager.cpp
```diff
@@ -195,23 +195,10 @@ void SConfig::SaveCoreSettings(IniFile& ini)
...
-  if (!devices_string.empty())
-    devices_string.pop_back();
-
-  section->Set("Devices", devices_string);
-}
-
 void SConfig::LoadSettings()
 {
   Config::Load();
+  IOS::HLE::LoadUSBWhitelistFromConfig();
```
AdmiralCurtiss, line 201: I'm... not sure where to put this, honestly. This doesn't seem like a good place for it, but I don't know of a better one.

leoetlino (approved):

PR #10330, opened 2021-12-31
GDB Stub: add support for lldb's qHostInfo

aldelaro5:
This along with #10319 is actually all it takes to make lldb work correctly and it's very easy to implement so might as well add it.

NOTE: this is specific to lldb, more info here: https://github.com/llvm/llvm-project/blob/main/lldb/docs/lldb-gdb-remote.txt#L867

In my test though, it seems to mix up the packets with read memory, but I have a suspicion that it's an unrelated issue that I will be trying to address because from dolphin's side, it is replying correctly.

### DIFF
--- a/Source/Core/Core/PowerPC/GDBStub.cpp
+++ b/Source/Core/Core/PowerPC/GDBStub.cpp
@@ -28,6 +28,7 @@ typedef SSIZE_T ssize_t;
 #include "Common/Event.h"
 #include "Common/Logging/Log.h"
 #include "Common/SocketContext.h"
+#include "Common/StringUtil.h"
 #include "Core/Core.h"
 #include "Core/HW/CPU.h"
 #include "Core/HW/Memmap.h"
@@ -59,6 +60,9 @@ enum class BreakpointType
 };
 
 constexpr u32 NUM_BREAKPOINT_TYPES = 4;
+
+constexpr int MACH_O_POWERPC = 18;
+constexpr int MACH_O_POWERPC_750 = 9;
 
 const s64 GDB_UPDATE_CYCLES = 100000;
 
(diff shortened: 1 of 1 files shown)


Source/Core/Core/PowerPC/GDBStub.cpp
```diff
@@ -319,6 +335,8 @@ static void HandleQuery()
     return SendReply("l");
   else if (!strncmp((const char*)(s_cmd_bfr), "qThreadExtraInfo", strlen("qThreadExtraInfo")))
     return SendReply("00");
+  else if (!strncmp((const char*)(s_cmd_bfr), "qHostInfo", strlen("qHostInfo")))
```
leoetlino, line 338: I don't like this chunk of C code (`(const char*)(s_cmd_bfr)` should be assigned to a local std::string_view variable to avoid repeating the ugly casts, strncmp and string literals), but I guess we can clean this up in a follow-up PR.

Source/Core/Core/PowerPC/GDBStub.cpp
```diff
@@ -305,6 +312,15 @@ static void SendReply(const char* reply)
   }
 }
 
+static void WriteHostInfo()
+{
+  return SendReply(
+      StringFromFormat(
```
leoetlino, line 318: Could we use fmt instead of StringFromFormat? The latter should be considered deprecated since we are migrating to fmt.

leoetlino (changes requested):

Source/Core/Core/PowerPC/GDBStub.cpp
aldelaro5: So in #10319 , we are moving these to a helper method which still has the stenxmp. Because I did it in that pr, it might not be a great idea to do it here.

Source/Core/Core/PowerPC/GDBStub.cpp
aldelaro5: I am not familiar with fmt, I just found this function when checking if we had a printf, how does that one work?

Source/Core/Core/PowerPC/GDBStub.cpp
leoetlino: Yeah, I'm not saying we should change this in this PR

Source/Core/Core/PowerPC/GDBStub.cpp
leoetlino: `<fmt/format.h>` is already included so you just need to do `fmt::format("cputype:{};cpusubtype:{};...", MACH_O_POWERPC, MACH_O_POWERPC_750);`. It's similar to Python string formatting.

leoetlino:
This needs a rebase fwiw.

leoetlino (approved):

PR #10332, opened 2022-01-01
Config: Port remaining General settings to new config system.

AdmiralCurtiss:


Source/Core/Core/Config/MainSettings.cpp
```diff
@@ -163,6 +164,58 @@ const Info<std::string> MAIN_RESOURCEPACK_PATH{{System::Main, "General", "Resour
...
+        {Config::System::Main, "General", fmt::format("ISOPath{}", current_path_idx)}, ""};
+    Config::SetBase(iso_path_info, p);
+    ++current_path_idx;
+  }
+
+  for (size_t i = current_path_idx; i < old_size; ++i)
+  {
+    Config::Info<std::string> iso_path{
+        {Config::System::Main, "General", fmt::format("ISOPath{}", i)}, ""};
+    Config::SetBase(iso_path, "");  // TODO: This actually needs a Config::Erase().
```
AdmiralCurtiss, line 214: We don't seem to have a way to actually remove a key from the new config system, so I've left this like this for now.

AdmiralCurtiss:
Also note: `IsSettingSaveable` already contained this section key and no specific in- or exclusions, so I didn't need to change anything there.

Source/Core/Core/Config/MainSettings.cpp
```diff
@@ -163,6 +164,58 @@ const Info<std::string> MAIN_RESOURCEPACK_PATH{{System::Main, "General", "Resour
...
+const Info<int> MAIN_ISO_PATH_COUNT{{System::Main, "General", "ISOPaths"}, 0};
+
+std::vector<std::string> GetIsoPaths()
+{
+  size_t count = MathUtil::SaturatingCast<size_t>(Config::Get(Config::MAIN_ISO_PATH_COUNT));
+  std::vector<std::string> paths;
+  paths.reserve(count);
+  for (size_t i = 0; i < count; ++i)
+  {
+    Config::Info<std::string> iso_path_info{
```
leoetlino, line 181: Can we move this into a helper function (e.g. `MakeISOPathConfigInfo(size_t idx)` or something like that)? This pattern shows up several times.

Source/Core/Core/Config/MainSettings.cpp
leoetlino: You could call DeleteKey() on the relevant layer but yeah I think this is fine for now.

Source/Core/Core/Config/MainSettings.cpp
AdmiralCurtiss: Yup, did that.

PR #10333, opened 2022-01-02
Concatenate green and blue values for color texgens

Pokechu22:
YAGCD [says](http://hitmen.c02.at/files/yagcd/yagcd/chap5.html#sec5.11.4) (for `COLOR_STRGBC0`/`COLOR_STRGBC1`): "Color texgen: (s,t)=(r,g:b) (g and b are concatenated)".  I don't know if anything actually cares about this, and I haven't done any hardware testing to confirm this behavior (or what concatenation actually means here; I've assumed it's a multiplication by 256), but it's worth implementing at least as a test.

### DIFF
--- a/Source/Core/VideoBackends/Software/TransformUnit.cpp
+++ b/Source/Core/VideoBackends/Software/TransformUnit.cpp
@@ -425,14 +425,14 @@ void TransformTexCoord(const InputVertexData* src, OutputVertexData* dst)
     break;
     case TexGenType::Color0:
       ASSERT(texinfo.inputform == TexInputForm::AB11);
-      dst->texCoords[coordNum].x = (float)dst->color[0][0] / 255.0f;
-      dst->texCoords[coordNum].y = (float)dst->color[0][1] / 255.0f;
+      dst->texCoords[coordNum].x = dst->color[0][0] / 255.0f;
+      dst->texCoords[coordNum].y = dst->color[0][1] / 255.0f + dst->color[0][2] / (255.0f * 256.0f);
       dst->texCoords[coordNum].z = 1.0f;
       break;
     case TexGenType::Color1:
       ASSERT(texinfo.inputform == TexInputForm::AB11);
-      dst->texCoords[coordNum].x = (float)dst->color[1][0] / 255.0f;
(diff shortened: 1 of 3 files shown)


Pokechu22:
The [FifoCI results](https://fifo.ci/version/5b45a2d334141d97b419a873a783dda0d9f8fa75/) (particularly [thps4-shadow](https://fifo.ci/compare/7177530-7170958/)) clearly indicate that `float3(o.colors_0.r, o.colors_0.g + o.colors_0.b * 256, 1)` is wrong.  I'll try `float3(o.colors_0.r, o.colors_0.g + o.colors_0.b / 256, 1)` now.  (That has the added benefit of staying in the range of 0-1, which would make sense since texture coordinates are usually multiplied by the texture size later on).

Note that thps4-shadow experienced green prior to #10012, but in [different places](https://fifo.ci/compare/6654262-6649809/).

Pokechu22:
Hmm, now none of the diffs are really interesting.  The one with the largest number of changed pixels is thps4-shadow, but the actual difference is negligible, and what the game seems to be doing (transforming RGBA values by having color 0 and color 1 be offset by 2 bytes so that a color 0 texgen uses r/g and color 1 uses b/a, and those values can be converted by a texture and then used to color things... at least as far as I can tell) doesn't look like it's supposed to use the concatenated values.  Hardware testing is needed for this.

shuffle2:
I think I've mentioned it before, but generally I wouldn't trust YAGCD (without hw tests)

Pokechu22:
Yep, I did a [quick test](https://gist.github.com/Pokechu22/a69beade3eae1ca87b4acf775f62dbb0) (based on the lesson08 example, except without completely broken tex coord handling and blending, see gist revisions for changes).

With the color as r=1, g=2, b=255:

Screenshot 2022-01-08 12-26-58

With the color as r=1, g=2, b=0:
Screenshot 2022-01-08 12-27-38

With this PR, b=255 would have resulted in the texture repeating 3 times (well, 2 + 255/256 times), but clearly that doesn't happen on real hardware.  So this is probably something done on early hardware that they decided against for the final hardware.

